PR #190, opened 2023-06-21
fix(satellite): do not expose CLI methods in migrators index

kevin-dp:
This PR fixes a problem with the current main where non NodeJS applications crash because `clients/typescript/src/migrators/index.ts` exported a CLI related function that uses NodeJS to read and write files. But NodeJS is not available in the browser and hence the app crashes.

This PR fixes this problem by moving all CLI related functions that use NodeJS specific methods to the `clients/typescript/src/cli` package. This package should only be used in dev mode when building Electric applications but should never be imported by the rest of the typescript client. Later, we will probably move the CLI package out of the typescript client to a package of its own.

### DIFF
--- a/clients/typescript/src/cli/index.ts
+++ b/clients/typescript/src/cli/index.ts
@@ -1,1 +1,1 @@
-export { buildMigrations } from './migrator'
+export { buildMigrations, loadMigrations } from './migrator'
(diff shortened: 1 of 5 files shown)


icehaunter (approved):
Good catch!

PR #191, opened 2023-06-21
feat: remove Electric.Postgres.SchemaRegistry

magnetised:


components/electric/lib/electric/postgres/postgres_manager.ex
```diff
@@ -157,18 +156,14 @@ defmodule Electric.Replication.PostgresConnectorMng do
...
-           tables <- Client.query_replicated_tables(conn, publication),
-           :ok <- Client.close(conn) do
-        tables
-        |> Enum.map(&Map.delete(&1, :columns))
-        |> then(&SchemaRegistry.put_replicated_tables(publication, &1))
-
-        Enum.each(tables, &SchemaRegistry.put_table_columns({&1.schema, &1.name}, &1.columns))
+           :ok <- OidDatabase.save_oids(oids) do
+        Logger.info(
+          "Migrated #{origin} to extension version #{versions |> Enum.reverse() |> List.first()}"
```
icehaunter, line 165:
```suggestion
          "Migrated #{origin} to extension version #{List.last(versions)}"
```

components/electric/lib/electric/postgres/replication.ex
```diff
@@ -4,6 +4,48 @@ defmodule Electric.Postgres.Replication do
...
+      identity?: false
+    ]
+
+    @type name() :: Postgres.name()
+
+    @type t() :: %__MODULE__{
+            name: name(),
+            type: atom(),
+            type_modifier: integer(),
+            identity?: boolean() | nil
```
icehaunter, line 23: I still mildly prefer `part_of_identity?` as a gentle reminder that there may be more that one column with this set to `true`

components/electric/lib/electric/postgres/extension/schema_cache.ex
```diff
@@ -173,29 +267,93 @@ defmodule Electric.Postgres.Extension.SchemaCache do
...
-  #          {{:ok, table}, state} <-
-  #            {Schema.fetch_table(schema, {ns, table}), state} do
-  #       {Schema.primary_keys(table), state}
-  #     end
+  # 1. we call refresh tables
+  # 2. pg **synchronously** queries the replication publication (electric)
+  #    for the list of replicated tables
+  # 3. the TcpServer calls this process to get the list of electrified tables
+  # 4. this process is waiting for the `REFRESH SUBSCRIPTION` call to finish
+  # 5. deadlock
```
icehaunter, line 324: Wow this must've been an absolute pain to debug

icehaunter (approved):
Amazing work, and all makes total sense! I have one tiniest suggestion, and a request: can you remove `@callback` statements from the `client.ex` module? We're not using that behaviour anywhere right now.

Oh, and one more thing: please add a changeset

components/electric/lib/electric/postgres/replication.ex
magnetised: That's an excellent point 👍🏻

components/electric/lib/electric/postgres/extension/schema_cache.ex
magnetised: oh yeah

components/electric/lib/electric/postgres/postgres_manager.ex
```diff
@@ -157,18 +156,12 @@ defmodule Electric.Replication.PostgresConnectorMng do
...
-           OidDatabase.save_oids(oids),
-           tables <- Client.query_replicated_tables(conn, publication),
-           :ok <- Client.close(conn) do
-        tables
-        |> Enum.map(&Map.delete(&1, :columns))
-        |> then(&SchemaRegistry.put_replicated_tables(publication, &1))
-
-        Enum.each(tables, &SchemaRegistry.put_table_columns({&1.schema, &1.name}, &1.columns))
+           :ok <- OidDatabase.save_oids(oids) do
+        Logger.info("Migrated #{origin} to extension version #{List.last(versions)}")
```
alco, line 164: It's a bit weird to have two log entries within the same logical block of code. Could you merge them into one?

components/electric/lib/electric/postgres/extension/schema_cache.ex
```diff
@@ -42,8 +62,20 @@ defmodule Electric.Postgres.Extension.SchemaCache do
     Electric.name(__MODULE__, origin)
   end
 
+  @spec name(Electric.reg_name()) :: Electric.reg_name()
+  def name(ref) when is_tuple(ref) do
+    ref
+  end
+
   def instance() do
```
alco, line 70: This function is never called.

components/electric/lib/electric/postgres/extension/schema_cache.ex
```diff
@@ -154,15 +229,34 @@ defmodule Electric.Postgres.Extension.SchemaCache do
...
     {:reply, {:ok, state.origin}, %{state | backend: backend, current: {version, schema}}}
   end
 
   def handle_call({:relation_oid, type, schema, name}, _from, state) do
     {:reply, SchemaLoader.relation_oid(state.backend, type, schema, name), state}
   end
 
-  def handle_call({:primary_keys, schema, name}, _from, state) do
-    {:reply, SchemaLoader.primary_keys(state.backend, schema, name), state}
+  # TODO: we shouldn't need this once vaxine is out of the read path
```
alco, line 240: Vaxine is already out of the read path.

components/electric/lib/electric/postgres/extension/schema_cache.ex
```diff
@@ -173,29 +267,93 @@ defmodule Electric.Postgres.Extension.SchemaCache do
...
 
-  def handle_call({:refresh_subscription, name}, _from, state) do
-    {:reply, SchemaLoader.refresh_subscription(state.backend, name), state}
+    {:reply, :ok, state}
+  end
+
+  @impl GenServer
+  # task process done
+  def handle_info({ref, result}, %{refresh_task: %{ref: ref}} = state) when is_reference(ref) do
+    Logger.debug("task_complete: #{inspect(result)}")
```
alco, line 350: This log entry lacks context to be undestood in the log stream. I guess it can be removed?

components/electric/lib/electric.ex
```diff
@@ -33,6 +33,19 @@ defmodule Electric do
     end
   end
 
+  @spec reg_or_locate(reg_name(), any()) :: :ok | {:error, :already_registered, {pid(), any()}}
+  def reg_or_locate({:via, :gproc, name}, value) do
+    this = self()
+
+    case :gproc.reg_or_locate(name, value) do
```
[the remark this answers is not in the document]
magnetised: `:gproc.reg_or_locate/2` does not spawn a process, that's `:gproc.reg_or_locate/3` - this `/2` version simply checks if the given name is registered and if not registers the current process under that name. hence the first clause of the case matches if nothing was registered, as `self()` becomes the process registered. 

I could do the locate+reg separately but this function does it all.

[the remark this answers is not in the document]
alco: Wow, I didn't expect `reg_or_locate/3` to be different from `reg_or_locate/2` in that regard. My bad, sorry.

components/electric/lib/electric/postgres/extension/schema_cache.ex
alco: @magnetised I was commenting on the `instance` function but it appears that `name` is not called outside of the module either. So `instance` and `@instance` can both be removed while the remaining clauses of `name` can be made private.

alco (approved):
Nice! Left one final comment on unused and  unnecessarily public functions.

components/electric/lib/electric/postgres/extension/schema_cache.ex
```diff
@@ -42,8 +60,12 @@ defmodule Electric.Postgres.Extension.SchemaCache do
     Electric.name(__MODULE__, origin)
   end
 
-  def instance() do
-    Electric.name(__MODULE__, :instance)
+  @spec name(Connectors.origin()) :: boolean()
```
alco, line 63:
```suggestion
  @spec ready?(Connectors.origin()) :: boolean()
```

PR #192, opened 2023-06-21
feat(satellite): ignore already applied migrations

kevin-dp:
This PR introduces a safeguard in Satellite that applies incoming migrations only if they were not already applied.

### DIFF
--- a/clients/typescript/src/migrators/bundle.ts
+++ b/clients/typescript/src/migrators/bundle.ts
@@ -56,15 +56,15 @@ export class BundleMigrator implements Migrator {
     // If this is the first time we're running migrations, then the
     // migrations table won't exist.
     const tableExists = `
-      SELECT count(name) as numTables FROM sqlite_master
+      SELECT 1 FROM sqlite_master
         WHERE type = 'table'
           AND name = ?
     `
-    const [{ numTables }] = await this.adapter.query({
+    const tables = await this.adapter.query({
       sql: tableExists,
       args: [this.tableName],
     })
-    if (numTables == 0) {
+    if (tables.length === 0) {
       return []
     }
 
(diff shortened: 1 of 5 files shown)


clients/typescript/src/migrators/bundle.ts
```diff
@@ -113,4 +113,31 @@ export class BundleMigrator implements Migrator {
...
+  /**
+   * Applies the provided migration only if it has not yet been applied.
+   * @param migration The migration to apply.
+   * @returns A promise that resolves to a boolean
+   *          that indicates if the migration was applied.
+   */
+  async applyIfNotAlready(migration: StmtMigration): Promise<boolean> {
+    const versionExists = `
+      SELECT count(version) as numVersions FROM ${this.tableName}
+        WHERE version = ?
```
alco, line 126: This looks a bit over-engineered. Wouldn't the following simpler query work?

```
const rows = await this.adapter.query({
  sql: "SELECT 1 FROM ${this.tableName} WHERE version = ?",
  args: [migration.version],
})
const shouldApply = rows.length == 0
```

---

Sidenote: shouldn't `this.tableName` also be passed as an argument to avoid interpolating any values into the SQL string?

alco (approved):

clients/typescript/src/migrators/bundle.ts
kevin-dp: Agreed, will simplify it this way. I based myself on a similar piece of code that already existed, will modify that one too.

icehaunter: > Sidenote: shouldn't this.tableName also be passed as an argument to avoid interpolating any values into the SQL string?

Binding a paramater as a table/column name is usually impossible in SQL, both in PostrgeSQL and [SQLite](https://stackoverflow.com/questions/1274432/sqlite-parameters-not-allowing-tablename-as-parameter)

PR #196, opened 2023-06-22
feat(satellite) add support for shape subscriptions on satellite client

balegas:


clients/typescript/src/util/subscriptions.ts
```diff
@@ -0,0 +1,403 @@
...
+      return
+    }
+
+    const inflight = this.inFlight[subscriptionId]
+    delete this.inFlight[subscriptionId]
+    for (const shapeReq of inflight) {
+      const shapeRequestOrResolved = shapeReq as ShapeRequestOrDefinition
+
+      if (shapeReqToUuid[shapeReq.requestId]) {
+        if (!this.subToShapes[subscriptionId]) {
```
kevin-dp, line 103: this if test can be simplified: `this.subToShapes[subscriptionId] = this.subToShapes[subscriptionId] ?? []`

clients/typescript/src/util/subscriptions.ts
```diff
@@ -0,0 +1,403 @@
...
+    this.subToShapes = subToShapes
+  }
+}
+
+// server sends tags, but no timestamp. Is this correct?
+// initial connect sends pending oplog, clears it (wait for server ack) and then can start making subscriptions.
+// ensure database is up-with-schema before receiving shape, otherwise need to send relation alongside and handle that.
+// client would process migrations if they are sent before subscription begin
+
+export class SubscriptionsDataCache extends EventEmitter {
```
kevin-dp, line 155: Let's move this to its own file in a new `shapes` folder :)

clients/typescript/src/util/subscriptions.ts
```diff
@@ -0,0 +1,403 @@
...
+  setState(
+    inFlight: SubcriptionShapeRequests,
+    subToShapes: SubcriptionShapeDefinitions
+  ) {
+    this.inFlight = inFlight
+    this.subToShapes = subToShapes
+  }
+}
+
+// server sends tags, but no timestamp. Is this correct?
```
kevin-dp, line 150: seem to be notes while implementing. Don't forget to remove before merging

clients/typescript/src/util/subscriptions.ts
```diff
@@ -0,0 +1,403 @@
...
+      this.subscriptionError(
+        SatelliteErrorCode.UNEXPECTED_SUBSCRIPTION_STATE,
+        `received subscription request but a subscription is already being delivered`
+      )
+    }
+    shapeRequestIds.forEach((rid) => this.remainingShapes.add(rid))
+  }
+
+  subscriptionResponse({ subscriptionId }: SatSubsResp) {
+    if (this.remainingShapes.size == 0) {
```
kevin-dp, line 183: This if test is duplicated on several places (e.g. in `subscriptionRequest`).
Could refactor it to a function that takes the error message as argument.

clients/typescript/src/util/subscriptions.ts
```diff
@@ -0,0 +1,403 @@
...
+        changes: res.transaction.map((t) =>
+          this.proccessShapeDataOperations(t, relations)
+        ),
+      },
+      shapeReqToUuid: res.shapeReqToUuid,
+    }
+
+    this.requestedSubscription = undefined
+    this.remainingShapes = new Set()
+    this.inDelivery = undefined
```
kevin-dp, line 252: Don't we also need to reset `currentShapeRequestId` ?
If yes, we can call `this.reset()`

clients/typescript/src/satellite/process.ts
```diff
@@ -190,14 +222,28 @@ export class SatelliteProcess implements Satellite {
...
-    const connectionPromise = this._connectAndStartReplication()
+    const connectionPromise = this._connectAndStartReplication(opts)
     return { connectionPromise }
   }
 
   async _setAuthState(authState: AuthState): Promise<void | Error> {
     this._authState = authState
   }
 
+  async _garbageCollectShapeHandler(shapeDef: ShapeDefinition): Promise<void> {
```
balegas, line 233: Need to pass an array and defer foreign key check, otherwise will be hell to drop a table

clients/typescript/src/util/subscriptions.ts
kevin-dp: Oh i see now, this will already have been done by `shapeDataEnd`. Still, may be easier to call `reset` here, don't think we care about 1 extra assignment.

clients/typescript/src/util/subscriptions.ts
```diff
@@ -0,0 +1,403 @@
...
+    this.requestedSubscription = undefined
+    this.remainingShapes = new Set()
+  }
+  isDelivering(): boolean {
+    return this.inDelivery != undefined
+  }
+
+  subscriptionRequest(shapeRequestIds: string[]) {
+    if (this.remainingShapes.size != 0) {
+      this.reset()
```
kevin-dp, line 173: no need to call `reset` because that's the first thing `subscriptionError` does.
Same comment applies on line 184

clients/typescript/src/util/types.ts
```diff
@@ -142,8 +157,40 @@ export enum AckType {
...
+export type SubscriptionDeliveredCallback = (data: SubscriptionData) => void
+export type SubscriptionErrorCallback = (error: SatelliteError) => void
+
 export type ConnectivityState =
   | 'available'
   | 'connected'
   | 'disconnected'
   | 'error'
+
+export type SubscribeResponse = {
```
kevin-dp, line 169: I would move all shape-related types to a separate `types.ts` file in `shapes` folder

clients/typescript/src/util/types.ts
```diff
@@ -142,8 +157,40 @@ export enum AckType {
...
+>
+
+export type ShapeSelect = {
+  tablename: string
+}
+
+export type SubscriptionData = {
+  subscriptionId: string
+  data: SubscriptionDataTransaction
+  shapeReqToUuid: { [req: string]: string }
```
kevin-dp, line 195: `Record<string, string>`

clients/typescript/test/satellite/process.test.ts
```diff
@@ -1195,5 +1204,215 @@ test('garbage collection is triggered when transaction from the same origin is r
...
+  try {
+    const conn = await satellite.start(authState)
+    await conn.connectionPromise
+    t.fail('start should throw')
+  } catch (e: any) {
+    t.is(e.code, SatelliteErrorCode.INVALID_POSITION)
+  }
+})
+
+test('apply shape data and persist subscription', async (t) => {
```
kevin-dp, line 1243: I'm wondering if we should make these tests serial because they are actually modifying a shared SQLite DB. And if these tests run in parallel this may perhaps cause issues.

clients/typescript/test/satellite/server_ws_stub.ts
```diff
@@ -128,6 +134,60 @@ export class SatelliteWSServerStub {
                 ])
               )
             }
+
+            if (msgType == getTypeFromString(SatSubsResp.$type)) {
```
kevin-dp, line 138: I know that you copied how it is done above, but this is way too much code duplication for my likings. Essentially every if test is doing the same, only the type cast is different and the type. This could be refactored:
- an array containing all message types
- loop over the array and for each message type do the if test

kevin-dp (changes requested):
Overall a great PR, most code is very readable :-)
I left some comments none of which are major. 
All minor refactorings, improvements, commenting issues.

clients/typescript/src/satellite/index.ts
```diff
@@ -47,8 +54,15 @@ export interface Satellite {
...
 
-  start(authConfig: AuthConfig): Promise<ConnectionWrapper>
+  start(
+    authConfig: AuthConfig,
+    opts?: SatelliteReplicationOptions
+  ): Promise<ConnectionWrapper>
   stop(): Promise<void>
+  subscribe(
+    shapeDefinitions: ClientShapeDefinition[]
+  ): Promise<void | SatelliteError>
```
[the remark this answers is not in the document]
icehaunter: That's not it. @balegas, this is unfortunately just wrong, and it's wrong everywhere in the client: you're assuming that the `raise` you're actually doing in `unsubscribe` (e.g. in the implementation in `process.ts`) will be somehow reflected in the type. Extremely unfortunately, it won't - so including the `SatelliteError` in return signature anywhere is wrong. The `raise`-ed error can never end up in `.then()` handler or show up as a result of `await ...` - it'll get propagated to `.catch` (untyped) or reraised to be caught in `try ... catch (e: unknown)` (also untyped). We're not returning errors, we're (correctly) `raise`-ing them, but TypeScript has no ways to define that behaviour (no `throws` from Java) (explanation as to why this is the case can be found [here](https://github.com/microsoft/TypeScript/issues/13219#issuecomment-1515037604)). Plese remove them from signatures here, and @kevin-dp, we can remove them anywhere we encounter

clients/typescript/src/satellite/client.ts
```diff
@@ -910,7 +910,7 @@ export class SatelliteClient extends EventEmitter implements Client {
     })
   }
 
-  private sendMessage(request: SatPbMsg) {
+  private sendMessage<T extends SatPbMsg>(request: T) {
```
kevin-dp, line 913: Isn't this equivalent in this case? Adding a generic with a subtyping constraint is useful if you don't want to lose type information for instance if the return type would be the same T as the one passed in. But in this case i don't see an added benefit of doing this.

clients/typescript/src/util/proto.ts
```diff
@@ -95,16 +81,18 @@ export function getTypeFromCode(code: number): string {
   return codemapping[code] ?? ''
 }
 
-export function getTypeFromString(string_type: string): number {
-  // FIXME
-  // return (msgtypemapping)[string_type]![0] ?? ''
-  return (msgtypemapping as any)[string_type]![0] ?? ''
+export function getTypeFromString(string_type: string): number | undefined {
+  return msgtypemapping[string_type]?.[0] ?? undefined
```
kevin-dp, line 85: `?? undefined` seems very weird, unless its purpose is to turn `null` into `undefined`?
If so, add a comment documenting this. Otherwise, just return `msgtypemapping[string_type]?.[0]`

kevin-dp (approved):
@icehaunter Looks good. Left two minor comments.

clients/typescript/src/util/subscriptions.ts
```diff
@@ -0,0 +1,403 @@
...
+  private subToShapes: SubcriptionShapeDefinitions
+
+  constructor() {
+    super()
+
+    this.inFlight = {}
+    this.subToShapes = {}
+  }
+
+  subscribeToUnsubscribeEvent(callback: GarbageCollectShapeHandler) {
```
icehaunter, line 71: Please specify return types on all non-private functions. And it would be nice to also have them specified on private. Having explicit return types makes it safer for us to update implementation detail - we have a clear contract with the world outside of the function and can adhere to it. If the return type is inferred, then any internal changes may alter the return type, unintentionally type-breaking the place of call instead of place of change

icehaunter (changes requested):
I agree with @kevin-dp's comments, and just a small ones of mine

clients/typescript/src/migrators/schema.ts
```diff
@@ -14,7 +14,7 @@ export const data = {
         //`-- Somewhere to track migrations\n`,
         `CREATE TABLE IF NOT EXISTS ${migrationsTable} (\n  id INTEGER PRIMARY KEY AUTOINCREMENT,\n  version TEXT NOT NULL UNIQUE,\n  applied_at TEXT NOT NULL\n);`,
         //`-- Initialisation of the metadata table\n`,
-        `INSERT INTO ${metaTable} (key, value) VALUES ('compensations', 0), ('lastAckdRowId','0'), ('lastSentRowId', '0'), ('lsn', ''), ('clientId', '');`,
+        `INSERT INTO ${metaTable} (key, value) VALUES ('compensations', 0), ('lastAckdRowId','0'), ('lastSentRowId', '0'), ('lsn', ''), ('clientId', ''), ('subscriptions', '');`,
```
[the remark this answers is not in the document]
balegas: Let me share the rationale behind the decision I made on this subject:
- It is very easy to manipulate the in-memory data-structure, but would lot more work to manipulate each individual subscription directly on the database.
- Therefore, database is just there to provide persistence
- We subscribe concurrently without hitting the storage. Subscription state is stored only when we apply the subscription, which happens serially. 
- The size of the object is relatively small so that we don't care about the cost comp/io (and it is client-side)

Also note that subscriptions will be an occasional operation, so I would not try to optimise so much for concurrency over simplicity. I think doing it from the database only makes sense if we have more complex flows.

clients/typescript/src/satellite/client.ts
```diff
@@ -18,6 +18,13 @@ import {
...
   SatAuthHeader,
   SatAuthHeaderPair,
+  SatSubsResp,
+  SatSubsReq,
+  SatSubsError,
+  SatSubsDataBegin,
+  SatSubsDataEnd,
+  SatShapeDataBegin,
+  SatShapeDataEnd,
 } from '../_generated/protocol/satellite'
```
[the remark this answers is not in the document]
balegas: Will look to it

clients/typescript/src/satellite/index.ts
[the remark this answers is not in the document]
balegas: I will fix that. I think was me that introduced the bad pattern early and everyone copied it (including me), even knowing that it doesn't make sense :D

clients/typescript/test/satellite/process.test.ts
```diff
@@ -1195,5 +1204,215 @@ test('garbage collection is triggered when transaction from the same origin is r
...
+                sql: `SELECT tags FROM _electric_shadow`,
+              })
+              t.is(shadowRows.length, 2)
+
+              const subsMeta = await satellite._getMeta('subscriptions')
+              const subsObj = JSON.parse(subsMeta)
+              t.is(Object.keys(subsObj).length, 2)
+              res()
+            } catch (e) {
+              t.fail()
```
icehaunter, line 1344: First, you can just `return` the promise you're building for a clearer intention instead of building and then awaitng it. 
Second, you can `reject(e)` here from the promise so that Jest reports a proper reason why this failed instead of a generic "failed" state

clients/typescript/src/util/proto.ts
```diff
@@ -23,14 +27,26 @@ const msgtypetuples: MappingTuples = {
...
+  SatShapeDataEnd: [18, Pb.SatShapeDataEnd],
 }
 
 const msgtypemapping = Object.fromEntries(
   Object.entries(msgtypetuples).map((e) => [getFullTypeName(e[0]), e[1]])
 )
 
 const codemapping = Object.fromEntries(
-  Object.entries(msgtypetuples).map((e) => [e[1][0], getFullTypeName(e[0])])
+  // FIXME
```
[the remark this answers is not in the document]
icehaunter: I addressed this in a later commit

clients/typescript/src/satellite/mock.ts
```diff
@@ -95,6 +120,81 @@ export class MockRegistry extends BaseRegistry {
...
 
 export class MockSatelliteClient extends EventEmitter implements Client {
+  relations: RelationsCache
+  constructor() {
+    super()
+
+    this.relations = {}
+  }
+
+  setRelations(relations: RelationsCache) {
```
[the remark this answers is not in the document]
balegas: Note that is for the mock, so not much problem with what is public.
I need this to set the relations ahead of streaming operations for a subscription, otherwise tests would also need to handle migrations etc... in fact, maybe I need to do that too.

@icehaunter this was something I wanted to raise during the call. I don't know how migrations and consequently relations are delivered if needed for subscriptions. I assume that the necessary messages will be delivered before.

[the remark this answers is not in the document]
balegas: - [ ] test missing relation path

clients/typescript/src/satellite/process.ts
```diff
@@ -264,7 +404,19 @@ export class SatelliteProcess implements Satellite {
...
+        if (
+          error.code == SatelliteErrorCode.BEHIND_WINDOW &&
+          opts?.clearOnBehindWindow
+        ) {
+          return this._handleSubscriptionError(error).then(() =>
+            this._connectAndStartReplication()
+          )
+        }
+
+        if (error.code && throwErrors.find((te) => te == error.code)) {
```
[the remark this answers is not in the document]
balegas: always learning

clients/typescript/src/satellite/process.ts
```diff
@@ -122,9 +145,18 @@ export class SatelliteProcess implements Satellite {
     )
 
     this.relations = {}
+
+    // TODO: load from database
+    this.subscriptions = new InMemorySubscriptionsManager()
+    this.subscriptions.subscribeToUnsubscribeEvent(
```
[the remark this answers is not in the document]
balegas: Fair point

clients/typescript/src/satellite/process.ts
```diff
@@ -230,6 +281,93 @@ export class SatelliteProcess implements Satellite {
...
+  _handleSubscriptionData(subsData: SubscriptionData) {
+    this.subscriptions.subscriptionDelivered(subsData)
+    if (subsData.data) {
+      this._applySubscriptionData(subsData.data.changes)
+    }
+  }
+
+  // only works if there is no previous subscription for table
+  // TODO: throw errors properly or define conflcit behaviour
+  async _applySubscriptionData(changes: DataChange[]) {
```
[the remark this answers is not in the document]
balegas: Will leave relevant info on comment

balegas:
Thanks a lot for all the comments. I will handle them with pleasure.

clients/typescript/src/satellite/client.ts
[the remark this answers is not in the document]
balegas: Need to touch all occcurence because it's ```import * as ``

clients/typescript/src/satellite/mock.ts
```diff
@@ -95,6 +120,81 @@ export class MockRegistry extends BaseRegistry {
 }
 
 export class MockSatelliteClient extends EventEmitter implements Client {
+  relations: RelationsCache
```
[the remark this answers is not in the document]
balegas: There was some strange reordering here. Fixed it now.

clients/typescript/src/util/subscriptions.ts
```diff
@@ -0,0 +1,403 @@
...
+}
+
+type SubcriptionShapeRequests = {
+  [k: string]: ShapeRequest[]
+}
+
+type SubscriptionDataInternal = {
+  subscriptionId: string
+  transaction: SatTransOp[]
+  shapeReqToUuid: { [req: string]: string }
```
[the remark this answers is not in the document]
balegas: There are a few of these

clients/typescript/src/util/subscriptions.ts
balegas: In general, I might have not paid attention to this, but the interface defines the return types. Agree that for convetion we might just specify the return types in the implementation anyways, for public and private.

clients/typescript/src/util/subscriptions.ts
balegas: Just calling reset instead :)

clients/typescript/src/util/subscriptions.ts
balegas: I prefer the monotony of the if statements if I cannot generize the pattern.

clients/typescript/src/util/types.ts
balegas: We (wrongly) overlaod the type Record in this file

clients/typescript/src/satellite/process.ts
```diff
@@ -230,6 +281,93 @@ export class SatelliteProcess implements Satellite {
...
+      SatelliteErrorCode.INTERNAL,
+      'unsubscribe shape not supported'
+    )
+    // return this.subscriptions.unsubscribe(subscriptionId)
+  }
+
+  _handleSubscriptionData(subsData: SubscriptionData) {
+    this.subscriptions.subscriptionDelivered(subsData)
+    if (subsData.data) {
+      this._applySubscriptionData(subsData.data.changes)
```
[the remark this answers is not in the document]
balegas: This function is going to be called through the event emitter, which doesn't await on completion. So, it doesn't matter if we await or not. However, given that this function could eventually be used elsewhere (but unexpectedly), it should still be async.  

Regarding the use of await in the return statement: the pattern I typically use is to return the result instead of awaiting and returng.
- I think it doesn't make a big difference because if the caller doesn't await, awaiting in the function has no effect either. I'd like to hear your toughts to understand better what should we adopt as a pattern.

balegas (commented):
Addressed all suggestions, closed those that I think require no second review or action

clients/typescript/src/util/types.ts
kevin-dp: I see.. We could let the IDE refactor our own `Record` type to a slightly different name to solve this. I can also live with leaving it like this for now

clients/typescript/test/satellite/process.test.ts
kevin-dp: Yep overlooked that! Ignore this comment :)

clients/typescript/src/migrators/schema.ts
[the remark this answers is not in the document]
kevin-dp: I thought it would be simpler to have 1 row per subscription (e.g. adding a subscription is adding a row vs reading, modifying the structure, writing it back) but i trust you when you say that actually the inverse is simpler.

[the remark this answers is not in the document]
balegas: It felt easier to me to reconstruct the in-memory state of the subscription manager by serialising/deserializng the entire thing, instead of reading/writting multiple columns and reconstructing it back from the query result.

To clarify we manage the in-flight subscriptions in-memory. If there is a crash, we don't recover in-flight subscriptions. We only persist the subscriptions object when a subscription is delivered.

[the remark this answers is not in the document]
kevin-dp: Yeah this will probably be even more true in the future when those structures are more complex. Thanks for clarifying!

icehaunter (approved):

PR #201, opened 2023-06-26
feat(cli): npm migration script

kevin-dp:
This PR introduces a NodeJS migration script that fetches all migrations from Electric and uses them to upgrade the application. It is written as a NodeJS program to ensure cross-platform compatibility. The `package.json` is extended with a `bin` property allowing users to write `npx electric <command>`. At the moment, we support a single command (`migrate`), so `npx electric migrate`. When calling `npx electric` it invokes the NodeJS script, and that script than looks at the additional parameter `migrate` and invokes the right handler for that. So in the future, we can extend this CLI with support for other commands.


icehaunter (approved):
Looks good & very readable! I have two reservations:
1. I'm not sure about the `migrate` subcommand - I can't say I agree fully that it reflects what's being done. Maybe smth like `pull-migrations` or other? @thruflo will definitely have an opinion
2. I don't love error handling by raises: on minified built code it yields this:
image
Which is hardly nice. How about wrapping the `handler(...commandArgs)` in a try-catch and just printing `ERROR: ...` with a correct non-zero exit code?

kevin-dp:
> Looks good & very readable! I have two reservations:
> 
>     1. I'm not sure about the `migrate` subcommand - I can't say I agree fully that it reflects what's being done. Maybe smth like `pull-migrations` or other? @thruflo will definitely have an opinion
> 
>     2. I don't love error handling by raises: on minified built code it yields this:
>        image
>        Which is hardly nice. How about wrapping the `handler(...commandArgs)` in a try-catch and just printing `ERROR: ...` with a correct non-zero exit code?

Yes, we can change the command name, just need to agree.
Regarding the error throwing, i figured out the same thing yesterday while working on the `init` command.
Had changed it there and was going to apply similar changes here.

icehaunter (approved):

PR #205, opened 2023-06-26
feat!: Stream Postgres replication log to Satellites

icehaunter:
This PR implements direct streaming of changes from Postgres to Satellites bypassing Vaxine.

It's implemented by utilizing a special WAL cache: a process, that stores WAL segments coming from PG, and the Satellites know and track their position in this WAL cache. This is because Postgres really doesn't like a lot of logical readers, so we have just one, and do the fan-out from the cache. Clients may also be in different positions in the stream, which leads to the WAL Cache interface providing options to start from an arbitrary point. 

## Notes for review
1. The E2E test may or may not be broken. They never were intended to work with this set of changes, so please disregard their status.
2. The ETS-backed WAL cache implementation is a bit rough, but it's to get it going at all. We're likely to replace it with an improved implementation a bit later (there's a PR by defnull that implements the improvements, but I decided to merge that PR after we have a baseline working 
3. The LSN position that Satellite stores is opaque, however with the coming shapes/initial sync logic, client may need to be able to compare those LSNs. For this reason it's a string-serialized number instead of `:erlang.term_to_binary/1`.

### DIFF
--- a/components/electric/config/config.exs
+++ b/components/electric/config/config.exs
@@ -35,6 +35,9 @@ config :electric, Electric.Replication.Postgres,
   pg_client: Electric.Replication.Postgres.Client,
   producer: Electric.Replication.Postgres.LogicalReplicationProducer
 
+config :electric, Electric.Postgres.CachedWal.Api,
+  implementation: Electric.Postgres.CachedWal.EtsBacked
+
 config :electric, Electric.StatusPlug, port: 5050
 
 config :electric, Electric.Satellite.Auth, provider: {Electric.Satellite.Auth.Insecure, []}
--- a/components/electric/lib/electric/postgres/cached_wal/api.ex
+++ b/components/electric/lib/electric/postgres/cached_wal/api.ex
@@ -19,6 +19,11 @@ defmodule Electric.Postgres.CachedWal.Api do
   @callback request_notification(wal_pos()) :: {:ok, await_ref()} | {:error, term()}
   @callback cancel_notification_request(await_ref()) :: :ok
 
+  @callback serialize_wal_position(wal_pos()) :: binary()
+  @callback parse_wal_position(binary()) :: {:ok, wal_pos()} | :error
+
+  @default_implementation Application.compile_env!(:electric, [__MODULE__, :implementation])
+
   @doc """
   Convert a "public" LSN position to an opaque pointer for the cached WAL.
 
@@ -28,7 +33,7 @@ defmodule Electric.Postgres.CachedWal.Api do
   database directly to catch up.
   """
   @spec get_wal_position_from_lsn(module(), lsn()) :: {:ok, wal_pos()} | {:error, :lsn_too_old}
-  def get_wal_position_from_lsn(module, lsn) do
+  def get_wal_position_from_lsn(module \\ @default_implementation, lsn) do
     module.get_wal_position_from_lsn(lsn)
   end
 
@@ -42,7 +47,7 @@ defmodule Electric.Postgres.CachedWal.Api do
   """
   @spec next_segment(module(), wal_pos()) ::
           {:ok, segment(), new_position :: wal_pos()} | :latest | {:error, :lsn_too_old}
-  def next_segment(module, wal_pos) do
+  def next_segment(module \\ @default_implementation, wal_pos) do
     module.next_segment(wal_pos)
   end
 
@@ -54,7 +59,7 @@ defmodule Electric.Postgres.CachedWal.Api do
   as soon as a new segment becomes available in the cache.
   """
   @spec request_notification(module(), wal_pos()) :: {:ok, await_ref()} | {:error, term()}
-  def request_notification(module, wal_pos) do
+  def request_notification(module \\ @default_implementation, wal_pos) do
     module.request_notification(wal_pos)
   end
 
@@ -62,7 +67,17 @@ defmodule Electric.Postgres.CachedWal.Api do
   Cancel a notification request issued previously by `request_notification/2`.
   """
   @spec cancel_notification_request(module(), await_ref()) :: :ok
-  def cancel_notification_request(module, await_ref) do
+  def cancel_notification_request(module \\ @default_implementation, await_ref) do
     module.cancel_notification_request(await_ref)
+  end
+
+  @spec parse_wal_position(module(), binary()) :: {:ok, wal_pos()} | :error
+  def parse_wal_position(module \\ @default_implementation, bin) do
+    module.parse_wal_position(bin)
+  end
+
+  @spec serialize_wal_position(module(), wal_pos()) :: binary()
+  def serialize_wal_position(module \\ @default_implementation, wal_pos) do
+    module.serialize_wal_position(wal_pos)
   end
 end
--- a/components/electric/lib/electric/postgres/cached_wal/api.ex
+++ b/components/electric/lib/electric/postgres/cached_wal/api.ex
@@ -1,0 +1,68 @@
+defmodule Electric.Postgres.CachedWal.Api do
+  @moduledoc """
+  Behavior for accessing cached wal
+  """
+  @type lsn :: Electric.Replication.Lsn
+
+  @typedoc "Position in the cached write-ahead log"
+  @type wal_pos :: term()
+
+  @typedoc "Notification reference no notify when new wal segment is available"
+  @type await_ref :: reference()
+
+  @typedoc "Wal segment, where segment is just an abstraction term within Electric"
+  @type segment :: Electric.Replication.Changes.Transaction.t()
+
+  @callback get_wal_position_from_lsn(lsn()) :: {:ok, wal_pos()} | {:error, term()}
+  @callback next_segment(wal_pos()) ::
+              {:ok, segment(), new_position :: wal_pos()} | :latest | {:error, term()}
+  @callback request_notification(wal_pos()) :: {:ok, await_ref()} | {:error, term()}
+  @callback cancel_notification_request(await_ref()) :: :ok
+
+  @doc """
+  Convert a "public" LSN position to an opaque pointer for the cached WAL.
+
+  Opaque pointer can be used with this API to request further segments.
+  There could be a case where lsn is already too old (i.e. out of the cached window),
+  in which case an error will be returned, and the client is expected to query source
+  database directly to catch up.
+  """
+  @spec get_wal_position_from_lsn(module(), lsn()) :: {:ok, wal_pos()} | {:error, :lsn_too_old}
+  def get_wal_position_from_lsn(module, lsn) do
+    module.get_wal_position_from_lsn(lsn)
+  end
+
+  @doc """
+  Get the next segment from the cached WAL from the current position.
+
+  If there's a next segment available, returns it along with the new position for the next read,
+  otherwise returns an atom `:latest`. There could be a case where lsn is already too old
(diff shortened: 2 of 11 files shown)


components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
```diff
@@ -0,0 +1,209 @@
...
+    GenStage.start_link(__MODULE__, opts, genstage_opts)
+  end
+
+  def clear_cache(stage) do
+    GenStage.cast(stage, :clear)
+  end
+
+  @impl Api
+  def get_wal_position_from_lsn(lsn) do
+    with {:ok, table} <- ETS.Set.wrap_existing(@ets_table_name) do
```
magnetised, line 57: Could we not replace these `wrap_existing` calls with a defined-at-compile-time `%ETS.Set{}` struct? 

Feel like the checks and metadata lookup done by that call are redundant.

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
```diff
@@ -0,0 +1,209 @@
...
+      {num, ""} -> {:ok, num}
+      _ -> :error
+    end
+  end
+
+  # Internal API
+
+  @impl GenStage
+  def init(opts) do
+    set = ETS.Set.new!(name: @ets_table_name, ordered: true)
```
magnetised, line 102: so the 'defined at compile time' value would just be this struct...

components/electric/lib/electric/postgres/cached_wal/producer.ex
```diff
@@ -0,0 +1,92 @@
...
+  @impl GenStage
+  def handle_demand(demand, state) do
+    state
+    |> Map.update!(:demand, &(&1 + demand))
+    |> send_events_from_cache()
+  end
+
+  @impl GenStage
+  def handle_info({:cached_wal_notification, _, :new_segments_available}, state) do
+    send_events_from_cache(state)
```
magnetised, line 73:
```suggestion
    send_events_from_cache(state, [])
```
removing the empty function decl would be clearer in this case IMHO

components/electric/lib/electric/replication/satellite_connector.ex
```diff
@@ -28,17 +28,15 @@ defmodule Electric.Replication.SatelliteConnector do
   def init(init_arg) do
     name = init_arg.name
     producer = init_arg.producer
-    vaxine_opts = init_arg.vaxine_opts
+    # vaxine_opts = init_arg.vaxine_opts
```
magnetised, line 31: why not just delete this comment

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -394,7 +395,7 @@ defmodule Electric.Satellite.Protocol do
 
   @spec initiate_subscription(String.t(), any(), OutRep.t()) :: OutRep.t()
   def initiate_subscription(client, lsn, out_rep) do
-    {:via, :gproc, vaxine_producer} = Vaxine.LogProducer.get_name(client)
+    {:via, :gproc, vaxine_producer} = Producer.name(client)
```
magnetised, line 398: `vaxine_producer`?

components/electric/lib/electric/postgres/cached_wal/producer.ex
```diff
@@ -0,0 +1,92 @@
...
+  def name(param) do
+    {:via, :gproc, {:n, :l, {__MODULE__, param}}}
+  end
+
+  @impl GenStage
+  def init(opts) do
+    {:producer,
+     %{
+       cached_wal_module:
+         Keyword.get(opts, :cached_wal_module, Electric.Postgres.CachedWal.EtsBacked),
```
magnetised, line 30: there are hardcoded assumptions later on that the cached-wal module is the one set in `config.exs`. I think the default here should be retrieved from the `config_env!` and I would question whether is should be configured on a per-producer level at all while those hard-coded assumptions exist

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -24,9 +24,9 @@ defmodule Electric.Satellite.Serialization do
   """
   @spec serialize_trans(Transaction.t(), term(), relation_mapping()) ::
           {[%SatOpLog{}], [Changes.relation()], relation_mapping()}
-  def serialize_trans(%Transaction{} = trans, vx_offset, known_relations) do
+  def serialize_trans(%Transaction{} = trans, offset, known_relations) do
     tm = DateTime.to_unix(trans.commit_timestamp, :millisecond)
-    lsn = :erlang.term_to_binary(vx_offset)
+    lsn = Electric.Postgres.CachedWal.Api.serialize_wal_position(offset)
```
magnetised, line 29: see comment on `cached_wal/producer.ex` above -- this is the hard-coded implementation assumption (or at least hard-coded at compilation time, taking into account the configuration).

it wouldn't be impossible to thread through the cached wal implementation choice into this serialisation function if we choose to make the backend configurable at runtime.

magnetised (commented):
love how understandable this is. great work

see my comments re compile- vs runtime-configuration

components/electric/lib/electric/replication/satellite_connector.ex
icehaunter: Cleaned up in a later PR, I've `git commit --fixup`-ed a lot of changes, but not all. The naming and comments are absolutely addressed in the entire PR chain, but I may have let some things slip with regards to where in the chain.

icehaunter: I won't address it here as to avoid introducing rebase conflicts, but thanks for noticing!

components/electric/lib/electric/satellite/protocol.ex
icehaunter: Same comment about naming & PRs

components/electric/lib/electric/postgres/cached_wal/producer.ex
icehaunter: I personally disagree, as an empty list as an argument to a function named `send_...` reads at a glance like we're sending an empty list

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
icehaunter: We can't move that call to compile time, but I can remove `ETS.Set` calls here

components/electric/lib/electric/postgres/cached_wal/producer.ex
icehaunter: I agree, that this module should by default be pulled from `config.exs`, however I'll stand on per-producer configurability. The reason here is that it makes the producer and the cached wal modules a bit more testable. I don't think in production we're going to switch out caches mid-flight, but for testing it's useful to be able to feed it whatever

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
icehaunter: Addressed in latest commit

components/electric/lib/electric/postgres/cached_wal/producer.ex
icehaunter: And done, now it also pulls from the same config

components/electric/config/config.exs
```diff
@@ -35,6 +35,9 @@ config :electric, Electric.Replication.Postgres,
   pg_client: Electric.Replication.Postgres.Client,
   producer: Electric.Replication.Postgres.LogicalReplicationProducer
 
+config :electric, Electric.Postgres.CachedWal.Api,
+  implementation: Electric.Postgres.CachedWal.EtsBacked
```
alco, line 39: Not sure if it's changed in a later PR but the Elixir convention is to use the name `adapter` for a module that implements a behaviour.

components/electric/lib/electric/postgres/cached_wal/api.ex
```diff
@@ -0,0 +1,84 @@
+defmodule Electric.Postgres.CachedWal.Api do
+  @moduledoc """
+  Behavior for accessing cached wal
+  """
+  @type lsn :: Electric.Replication.Lsn
```
alco, line 5:
```suggestion
  @type lsn :: Electric.Replication.Lsn.t()
```

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
```diff
@@ -0,0 +1,204 @@
...
+  """
+  def get_name(origin \\ :default) do
+    {:via, :gproc, {:n, :l, {__MODULE__, origin}}}
+  end
+
+  @doc """
+  Start the cache. See module docs for options
+  """
+  def start_link(opts) do
+    genstage_opts = Keyword.take(opts, [:name])
```
alco, line 47: The ability to configure the process name looks superfluous. Since the ETS table name is hard-coded, only a single such process can run at the same time, so it can be named `Electric.Postgres.CachedWal.EtsBacked` without the ability to override this name.

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
```diff
@@ -0,0 +1,204 @@
...
+    GenStage.start_link(__MODULE__, opts, genstage_opts)
+  end
+
+  def clear_cache(stage) do
+    GenStage.cast(stage, :clear)
```
alco, line 52: Why not use the same name for both the public function and the message tag? That would make it easier to jump between the two in the code.

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
```diff
@@ -0,0 +1,204 @@
...
+    # This doesn't do anything with notification requests, but this function is not meant to be used in production
+    {:noreply, [], %{state | current_cache_count: 0, last_seen_wal_pos: 0}}
+  end
+
+  @impl GenStage
+  @spec handle_events([Transaction.t()], term(), state()) :: {:noreply, [], any}
+  def handle_events(events, _, state) do
+    events
+    |> Stream.each(& &1.ack_fn.())
+    |> Stream.reject(&Enum.empty?(&1.changes))
```
alco, line 145: When I was testing my initial sync implementation, I saw a few empty transactions passing throught this `handle_events()` callback. I don't know what actions created them but I just want to note here that ignoring such transactions means the cached WAL may have an outdated view of the actual PG LSN even after all messages from the PG replication stream have been processed.

alco (approved):

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
icehaunter: That's fair, but we also don't want to inflate the cache by empty transactions and send them to all subscribers. Let us leave it like this for this PR, and update it when we run into issues

icehaunter: Added a comment to address

components/electric/lib/electric/postgres/cached_wal/producer.ex
magnetised: i just find it harder to parse the empty function head. I'd reserve those for public functions with more complex pattern matching requirements. 

perhaps the function just has the wrong name? maybe `flush_events_cache` or something.

not a biggie

PR #208, opened 2023-06-26
chore(pg): Remove all remnants of Vaxine

icehaunter:
This is the final PR of PG-based conflict resolution. It just removes all mentions of Vaxine in general, as well as the old migration system. If reviewers know what else would be good to remove from codebase, it's a great opportunity to tell me.

Closes VAX-647


components/electric/lib/electric/satellite/ws_server.ex
```diff
@@ -154,7 +154,7 @@ defmodule Electric.Satellite.WsServer do
     end
   end
 
-  # Consumer (Vaxine) has reported that this lsn has been stored successfully
+  # Consumer (SatelliteCollectorConsumer) has reported that this lsn has been stored successfully
```
alco, line 157: What is SatelliteCollectorConsumer?

alco (approved):

components/electric/lib/electric/replication/postgres/slot_server.ex
```diff
@@ -229,14 +218,18 @@ defmodule Electric.Replication.Postgres.SlotServer do
...
 
-    vx_offset = get_vx_offset(state.slot_name, start_lsn)
-    Logger.debug("Got vx offset #{inspect(vx_offset)} for start_lsn #{inspect(start_lsn)}")
-
-    GenStage.async_subscribe(self(), [
-      {:to, {:via, :gproc, state.producer_name}},
-      {:cancel, :temporary}
-      | producer_info()
-    ])
+    # TODO: We're currently not using this position in a the `SatelliteCollectorProducer`, but we should
```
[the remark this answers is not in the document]
icehaunter: That's a "fan-in" producer, added in https://github.com/electric-sql/electric/pull/206/commits/1ab4fa173d94ff2a02de68585b0db88dbc99aff8, which is meant to feed PG a merged stream of changes from all Satellites

components/electric/lib/electric/satellite/ws_server.ex
icehaunter: That's a "fan-in" consumer, added in https://github.com/electric-sql/electric/commit/1ab4fa173d94ff2a02de68585b0db88dbc99aff8, that's needed since Satellite `ws_server` process is structured in such a way that it initiates a `SatStartReplicationReq` request to fetch changes from Satellite to the server only after it receives demand from a GenStage consumer. This particular consumer is very simple: it generates demand, and then forwards any events to `SatelliteCollectorProducer`, which in turn acts as GenStage producer for `SlotServer` that sends these events to PostgreSQL.

Before you ask, I couldn't have structured this as a `:producer_consumer`, since `:producer_consumer` never generates it's own demand, and I found it harder to reason about a single producer-consumer that we need to subscribe explicitly to each Satellite ws process, and the demand would flow up from `SlotServer`. If you want, we can take that discussion to the appropriate PR of #206

magnetised (approved):
Satisfying.

PR #209, opened 2023-06-26
feat(cli): init command to set up Electric app

kevin-dp:
This PR extends the Electric Node CLI with an `init` command to initialise an Electric app.
This script creates a `.electric/@config/index.mjs` file that contains the necessary configurations variables required by the current state of the Satellite client.
The script can be used as follows:
```shell
npx electric init <appId>
# or
npx electric init <appId> --force # overrides existing config files
```


icehaunter (approved):
Love the readability. Just two small nitpicks

alco (approved):
:+1:

PR #210, opened 2023-06-26
feat(electric): Initial sync from PG

alco:
Closes VAX-669.

This PR implements a "full initial sync" that sends the migration history as well as all data from electrified tables to a fresh client in one go. Subsequent improvements are coming up as defined by VAX-732, VAX-733, and VAX-734.

Note that the commits authored by Ilia in this PR overlap with the changes made in https://github.com/electric-sql/electric/pull/205. Once that PR is merged, I will rebase mine on top.

### DIFF
--- a/components/electric/lib/electric/satellite/protocol.ex
+++ b/components/electric/lib/electric/satellite/protocol.ex
@@ -191,45 +191,34 @@ defmodule Electric.Satellite.Protocol do
   # Satellite client request replication
   def process_message(
         %SatInStartReplicationReq{lsn: client_lsn, options: opts} = msg,
-        %State{in_rep: _in_rep, out_rep: out_rep} = state
+        %State{} = state
       ) do
-    with {:ok, lsn} <- validate_lsn(client_lsn, opts) do
-      out_rep =
-        case Enum.member?(opts, :SYNC_MODE) do
-          true ->
-            sync_batch_size = msg.sync_batch_size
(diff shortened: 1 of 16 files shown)


components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -207,8 +209,21 @@ defmodule Electric.Satellite.Protocol do
 
       Metrics.satellite_replication_event(%{started: 1})
 
-      out_rep = initiate_subscription(state.client_id, lsn, out_rep)
-      {[%SatInStartReplicationResp{}], %State{state | out_rep: out_rep}}
+      state =
+        if lsn == %Lsn{segment: 0, offset: 0} do
```
icehaunter, line 213: Tests are failing since alias is gone after rebase.

However, it's important to note that I'm not using `%Lsn{}` struct any more here. You should probably depend on `lsn == :start_from_first` if we expect the client to send `:FIRST_LSN` option here, otherwise we may expect that to be a literal integer 0, but that's implementation detail of the cache. Since we expect client-sent lsn do be semi-opaque until they reach the cached wal module, it's better to expect `:start_from_first`

icehaunter:
I'd love a rebase on #208 (which is the latest PR into my "remove vaxine" branch), as e2e tests pass there. You can then add at least one e2e test covering this functionality with actual NodeJS Satellite client

icehaunter (approved):

PR #212, opened 2023-06-28
feat(cli): remove config file

kevin-dp:
This PR removes the need for a configuration file, instead configuration options are passed directly to the `electrify` call.
The migrations are no longer part of those options and are instead bundled together with the database description object `dbSchema` in the generated client.

Thus, this PR also modifies the CLI's `generate` command to generate a migration file that lives in the same folder as the generated client. The generated client imports the migrations and passes them to the `dbSchema` object.


alco (commented):
:clap:

PR #216, opened 2023-06-28
feat(electric)!: Revamp Electric's auth implementation

alco:
Joken provides a richer set of features of which declarative claim validation and support for easily configurable asymmetric signing algorithms are the main highlights.

It's also more actively maintained and orders of magnitude more popular, judging by the download stats on hex.pm.

There's no need for any authentication-related configuration to happen at compile time. We'll be moving it to runtime config which will allow us to

  1. Validate user-provided config at application start.

  2. Convert the user-provided config into a clean map that is then easier to work with in the modules that implement auth.

The `metadata` field can be used to add any keys to Logger calls without also white-listing them in the Logger configuration (which is easy to forget).

Fixes VAX-663, VAX-665, VAX-666

BREAKING CHANGE: Old-style insecure authentication will no longer work with these server changes. Old-style JWT tokens produced by Electric Console are also not compatible with these changes.


components/electric/config/runtime.exs
```diff
@@ -7,6 +7,46 @@
...
+        {Auth.Insecure, auth_config}
+
+      "jwt" ->
+        auth_config =
+          [
+            alg: System.get_env("SATELLITE_AUTH_JWT_ALG"),
+            key: System.get_env("SATELLITE_AUTH_JWT_KEY"),
+            namespace: System.get_env("SATELLITE_AUTH_JWT_NAMESPACE"),
+            iss: System.get_env("SATELLITE_AUTH_JWT_ISS"),
+            aud: System.get_env("SATELLITE_AUTH_JWT_AUD")
```
icehaunter, line 36: Please add all available environment variables to README

alco: I didn't think it was necessary to mention optional configuration parameters in README. They're all going to be documented in our new docs ([preview](https://deploy-preview-4--wip-electric-sql-docs.netlify.app/docs/usage/auth/jwt#configuration-settings)).

icehaunter: Sure, but we're an open-source project. I find it very very useful when all possible configuration params are mentioned in README, and details are in proper docs

icehaunter: Also your docs link mentions some-but-not-all configuration params, and I really like a full reference when I'm setting up a service to be ran. I also would love for this readme to show up on DockerHub, where it's also going to be very useful to have all configuration options immediately listed

icehaunter (changes requested):

components/electric/config/runtime.exs
alco: Some open-source projects choose to use GitHub exclusively for documentation and for those projects it makes sense to have the full API or configuration reference in the README file.

But we have the bulk of our docs published on a dedicated website and, from what I've seen, API reference for the various components of our system will be part of that. In this context, duplicating reference docs in the README and on the docs website is just asking for trouble. At some point one will remember to update them in one place but forget to do it in the other one.

My opinion is that our READMEs should be slimmed down in the lead up to the public v2.0 release as we're making the docs website more comprehensive, not the other way around.

alco: Regarding your last comment, I don't which configuration options are missing from the docs preview. These are all the options that are relevant to auth and they're all mentioned on the docs website:
 * SATELLITE_AUTH_MODE
 * SATELLITE_AUTH_JWT_ALG
 * SATELLITE_AUTH_JWT_KEY
 * SATELLITE_AUTH_JWT_NAMESPACE
 * SATELLITE_AUTH_JWT_ISS
 * SATELLITE_AUTH_JWT_AUD

components/electric/test/electric/satellite/auth/jwt_test.exs
```diff
@@ -0,0 +1,288 @@
...
+        public_key = Path.join("test/fixtures/keys", public_key_filename) |> File.read!()
+        private_key = Path.join("test/fixtures/keys", private_key_filename) |> File.read!()
+
+        signer = Joken.Signer.create(alg, %{"pem" => private_key})
+        {:ok, token, _} = Joken.encode_and_sign(claims, signer)
+
+        config = build_config!(alg: alg, key: public_key, namespace: @namespace)
+        assert {alg, {:ok, %Auth{user_id: "12345"}}} == {alg, validate_token(token, config)}
+      end
+    end
```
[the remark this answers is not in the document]
alco: I wouldn't be so sure about Joken's test suite, it only tests a few things related to the implementation of its `Joken.Signer` struct. You can say that it's `jose`'s responsibility to test this because that's the underlying library that Joken wraps but then there won't be any tests verifying the full end-to-end flow from our Elixir code to `jose` unless we write those tests ourselves.

Think of these as end-to-end tests that verify our claims in the documentation. For example, we support the `ES512` (signing algorithm because it's one of the algorithms defined in JWT spec) and provide hints on generating the key pair compatible with it. Without testing a token signed using this algorithm and an appropriate key, we can only hope it'll work as expected. Better verify in tests than rely on hope, IMO.

Especially when it comes to public key and elliptic curve cryptography, keys may be encoded in different formats, so it's best to have a tested happy path if only to be able to look it up when we get user feedback complaining about a certain JWT not getting validated by Electric.

components/electric/config/runtime.exs
```diff
@@ -7,6 +7,46 @@
...
 import Config
 
+alias Electric.Satellite.Auth
+
+auth_provider =
+  if config_env() == :test do
+    auth_config =
+      Auth.JWT.build_config!(
+        alg: "HS256",
+        key: Base.decode64!("AgT/MeUiP3SKzw5gC6BZKXk4t1ulnUvZy2d/O73R0sQ="),
```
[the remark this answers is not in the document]
alco: No reason, this code was taken from the old implementation verbatim. I have updated it to use a less opaque-looking key.

components/electric/test/electric/satellite/auth/jwt_test.exs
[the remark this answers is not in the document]
icehaunter: OK, sounds good

components/electric/config/runtime.exs
icehaunter: > Regarding your last comment, I don't which configuration options are missing from the docs preview. These are all the options that are relevant to auth and they're all mentioned on the docs website:
> 
>     * SATELLITE_AUTH_MODE
> 
>     * SATELLITE_AUTH_JWT_ALG
> 
>     * SATELLITE_AUTH_JWT_KEY
> 
>     * SATELLITE_AUTH_JWT_NAMESPACE
> 
>     * SATELLITE_AUTH_JWT_ISS
> 
>     * SATELLITE_AUTH_JWT_AUD

Yeah, the docs mention all of them, but I'm saying that that's not enough to actually run the container. I believe readme should be sufficient if light on the details. Please please let's add them to the list -- that list was written by me when I was setting up our infra + console (now defunct), and found it very useful to immediately know what variables are available to configure the container. The Docs are very good at explaining the details, but I insist on having enough information for configuring the container in README

alco: Added.

icehaunter (approved):

PR #217, opened 2023-07-04
feat(CLI): watch flag for generate command

kevin-dp:
This PR introduces an optional `--watch` flag for the CLI's `generate` command.
When running with the `--watch` flag, the process periodically polls Electric's migration endpoint for new migrations. Use it as follows:
```sh
npx electric-sql generate --watch
```

We modified the Electric http endpoint to return a 204 if there are no new migrations.

### DIFF
--- a/clients/typescript/src/cli/migrations/migrate.ts
+++ b/clients/typescript/src/cli/migrations/migrate.ts
@@ -1,4 +1,5 @@
 import path from 'path'
+import * as z from 'zod'
 import * as fs from 'fs/promises'
 import { createWriteStream } from 'fs'
 import http from 'http'
@@ -13,6 +14,7 @@ const appRoot = path.resolve() // path where the user ran `npx electric migrate`
 export const defaultOptions = {
   service: process.env.ELECTRIC_URL ?? 'http://localhost:5050',
   out: path.join(appRoot, 'src/generated/client'),
+  watch: false,
 }
 
 export type GeneratorOptions = typeof defaultOptions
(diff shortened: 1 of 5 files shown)


clients/typescript/src/cli/migrations/migrate.ts
```diff
@@ -34,6 +36,119 @@ type DataSourceDescription = {
...
+  }
+}
+
+/**
+ * Periodically polls Electric's migration endpoint
+ * to check for new migrations. Invokes `_generate`
+ * when there are new migrations.
+ */
+async function watchMigrations(opts: Omit<GeneratorOptions, 'watch'>) {
+  const pollingInterval = 1000 // ms
```
thruflo, line 53: Might be nice to be configurable.

thruflo (approved):
Great stuff =D

Looking at [the handler](https://github.com/electric-sql/electric/blob/main/components/electric/lib/electric/plug/migrations.ex#L21
), it looks pretty simple to just return 204 if the `migrations` for version are empty. E.g.:

```elixir
  get "/" do
    conn = fetch_query_params(conn)

    with {:ok, dialect} <- get_dialect(conn),
         {:ok, migrations} when migrations != [] <- get_migrations(conn),
         {:ok, body} <- migrations_zipfile(migrations, dialect) do
      conn
      |> put_resp_content_type("application/zip", nil)
      |> send_resp(200, body)
    else
      {:error, reason} ->
        json(conn, 204, %{})
      {:error, reason} ->
        json(conn, 403, %{error: to_string(reason)})
    end
  end
```

kevin-dp:
@thruflo aha, yeah 204 seems a good approach! Will try it out

thruflo (approved):
Awesome stuff, thank you @kevin-dp :)

One note: it's probably worth adding a test to [the Elixir side](https://github.com/electric-sql/electric/blob/main/components/electric/test/electric/plug_test.exs) for the 204 behaviour.

PR #221, opened 2023-07-05
feat(examples):linearlite without vaxine or local config

paulharter:


examples/linearlite/src/pages/Board/IssueItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+              <div className="flex flex-col">
+                <span className="text-xs font-normal text-gray-500 uppercase">
+                  {issue.id}
+                </span>
+                <span className="mt-1 text-sm font-medium text-gray-700 line-clamp-2 overflow-ellipsis">
+                  {issue.title}
+                </span>
+              </div>
+              <div className="flex-shrink-0">
+                {issue.owner ? (
```
kevin-dp, line 59: instead of returning a different avatar depending on whether an owner is defined (which btw will never be the case becasue that column does not exist in the DB) it would be better to let the Avatar component handle the case where name and/or avatarUrl are missing. That would simplify this code:
```typescript jsx
<Avatar
  name={issue.owner?.name}
  avatarUrl={issue.owner?.avatar}
/>
```

examples/linearlite/src/pages/Board/IssueCol.tsx
```diff
@@ -0,0 +1,65 @@
+import StatusIcon from '../../components/StatusIcon';
+import React, { memo } from 'react';
+import { Droppable, DroppableProvided } from 'react-beautiful-dnd';
+import { BsThreeDots as MoreIcon } from 'react-icons/bs';
+import { GoPlus as AddIcon } from 'react-icons/go';
+// import { Issue } from 'types/issue';
```
kevin-dp, line 6: You have to import Issue since you are using it below, but i think you want to `import {issue as Issue} from '../../generated/client'`

examples/linearlite/src/components/IssueModal.tsx
```diff
@@ -0,0 +1,213 @@
...
+import PriorityMenu from './contextmenu/PriorityMenu';
+import StatusMenu from './contextmenu/StatusMenu';
+import PriorityIcon from './PriorityIcon';
+import StatusIcon from './StatusIcon';
+
+interface Props {
+  isOpen: boolean;
+  onDismiss?: () => void;
+}
+function getPriorityString(priority: string) {
```
kevin-dp, line 30: The argument can probably be typed more precisely as `Priority` (cf. `types/issue.ts`) instead of `string`. Also these constants are already strings, e.g. `Priority.HIGH` is the string `'high'. So we could replace this entire switch by:
```typescript
function getPriorityString(priority: Priority) {
  return priority.charAt(0).toUpperCase() + priority.slice(1)
}
```

Note that the only functional difference with this code and yours is that `Priority.NO_PRIORITY` is defined as `'no_priority'` and you're here turning it into `'Priority'`. If that's really needed you could special case it in my code above but i would argue that we should seek to unify them, e.g. just use 'no_priority'.

Also, i'm not sure if these priorities really need to be uppercased here or if we can just use them as how they are defined in `Priority`.

examples/linearlite/src/components/IssueModal.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+  const handleSubmit = () => {
+    if (title === '') {
+      showWarning('Please enter a title before submiting', 'Title required');
+      return;
+    }
+
+    db.issue.create({data: {
+        title: title,
+        name: "Hello",
```
kevin-dp, line 62: why are we hardcoding the name of the issue?

examples/linearlite/src/components/IssueModal.tsx
```diff
@@ -0,0 +1,213 @@
...
+    //     id: undefined,
+    //     priority: priority,
+    //     status: status,
+    //     description: description,
+    //   })
+    // );
+
+    // clear state
+    // close modal
+    if (onDismiss) onDismiss();
```
kevin-dp, line 83: can be simplified to `onDismiss?.()`

examples/linearlite/src/components/contextmenu/PriorityMenu.tsx
```diff
@@ -0,0 +1,83 @@
...
+  onSelect,
+}: Props) {
+  const [keyword, setKeyword] = useState('');
+
+  const handleSelect = (priority: string) => {
+    setKeyword('');
+    if (onSelect) onSelect(priority);
+  };
+  let statusOpts = [
+    [NoPriorityIcon, 'No priority', Priority.NO_PRIORITY],
```
kevin-dp, line 35: the strings that are used here are different from the others.
for example, `Priority.NO_PRIORITY` is defined as `'no_priority'` but i have seen usages of the string `'Priority'` for  `Priority.NO_PRIORITY`  in `getPriorityString` in `issueModal.tsc` and now we there is a 3rd version which is `'No priority'`. Not sure what these are used for but if these are used for anything else than showing to the user this will yield subtle bugs

kevin-dp (changes requested):
Amazing to see a LinearLite app running on top of Electric =D 

I went through all files and left comments. I'm not sure which parts you wrote/modified and which parts stem from the linearlite app you forked so i may have left comments on parts that were not written by you.

I have a few general comments:
- The formatting across the project is inconsistent and also inconsistent with the formatting we adhere to in the typescript client. Since there are many files, best to set up prettier to automatically format all code.
- There are quite some typescript errors, could you please fix them?
I left a comment in `tsconfig.json` where you should add a line in order to be able to run the typescript compiler using `tsc`.
- There are commented pieces of code and logging statements meant for debugging purposes a bit everywhere. Could you remove those please? :-)

balegas:
Closing this PR, referring it from #VAX-942

PR #222, opened 2023-07-05
feat(electric)!: implement partial replication via subscriptions

icehaunter:
This implements partial replication using subscription. Client-side part of this has been done in #196, so this implements the server-side. This implementation is a breaking change, as clients that do not issue subscriptions will never receive any data, unlike previously. 

## Subscription logic

When a subscription request comes in, we need to do two things to send the initial snapshot: request the data and figure out where to insert it in the replication stream, especially with other subscriptions already active.

We achieve the correct functionality with two techniques: `REPEATABLE READ` transactions and stalling the outgoing replication stream if we've reached the point of insertion before we actually have the data for it. 

### Insertion point

Under `REPEATABLE READ` read-only transactions, we can confidently say, that [`xmin`](https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-PG-SNAPSHOT-PARTS) of the current snapshot (i.e. the last transaction id that has been committed and visible) is definitely the point after which the have not seen any transactions. That means that while processing the replication stream, if we know the `xmin` of a subscription initial data request is `X`, then we can safely send any transactions where `id < X`.

### Pause logic

When subscription initial data is requested, we start the aforementioned `REPEATABLE READ` transaction from a separate process and wait for `xmin` to be relayed back to the main process. This means the WebSocket process if frozen until we know the insertion point (which should be very fast) to avoid accidentally skipping past the insertion point. There could be a case where the client send multiple subscription requests while the first one's data is still pending. In that case we use the knowledge that the `xmin` of the next subscription request is going to be larger or equal to the previous one, so we enqueue the new insertion point to be handled afterwards.

After the insertion point is known, there are four possible scenarios:

1. We've been sending the replication stream, but we've reached a point where transaction id >= `xmin` for the given subscription. That means we cannot continue until we are able to send the subscription data. We thus pause the outgoing stream and enqueue all events that are meant to be sent and wait for the initial data to arrive. Any additional events that come in while the stream is pause are also enqueued. We may already have the data for the pending subscription (see scenario 4), in which case we can send it immediately and continue sending the events until the next insertion point.
2. We've received the subscription data while the stream is "paused", and we're paused waiting for this subscription. In that case we're immediately sending the data over to the client and then continuing sending the events, starting from the enqueued ones, until the next insertion point.
3. We've received the subscription data while the stream is "paused", and we're paused at an insertion point for a different subscription. In that case we just store the data to be sent when we reach the correct insertion point.
4. We've received the subscription data while the stream is ongoing. That means we've definitely not reached the insertion point for this subscription, so we store the data  to be sent when we reach the correct insertion point.

### Filtering logic

The entire "insertion point" logic exists to support correct filtering of the outgoing transactions: we should send any operation that aren't part of any subscriptions. That rule works with initial data seeing only transactions with ids < `xmin`: any changes pertaining to the subscription before `xmin` will be filtered out, and after we send the initial data "at" `xmin`, any changes after will be sent through and the client will see the correct picture - both initial state and the streaming changes. It means that only subscriptions for which the data has been sent are considered "established" and are used to check if the changes fall into one of the subscriptions.


alco (commented):
Monumental work! Good job.

It's difficult at times to wrap your head around the logic split between WsServer, Protocol, and InitialSync. But I guess that's just the complexity of our requirements showing through.

One question: is the race condition in the initial sync of transactions a deal breaker for this PR? Reason I'm asking is because I've added a few more changes to the Initial Sync PR. You may want to rebase on top of that or I can rebase my additional changes on top of this PR. Either way works for me but I would like to get my PR merged as is to unblock you and address the race condition separately, based on the changes in this PR.

components/electric/lib/electric/replication/shapes.ex
```diff
@@ -0,0 +1,117 @@
...
+
+  @doc """
+  Remove all changes from a transaction which do not belong to any of the
+  requested shapes.
+
+  May result in a transaction with no changes.
+  """
+  @spec filter_changes_from_tx(Transaction.t(), [term()]) :: Transaction.t()
+  def filter_changes_from_tx(%Transaction{changes: changes} = tx, shapes) do
+    %{tx | changes: Enum.filter(changes, &should_change_be_send?(&1, shapes))}
```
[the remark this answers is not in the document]
icehaunter: I'm not sure if I agree that the name is awkward, I like it :)
I can change it to `change_belongs_to_any_shape?` maybe?

components/electric/lib/electric/replication/shapes/shape_request.ex
```diff
@@ -0,0 +1,135 @@
...
+  @spec rows_to_records_with_tags([tuple()], [String.t(), ...], term(), String.t()) :: [
+          Changes.NewRecord.t()
+        ]
+  defp rows_to_records_with_tags(rows, col_names, relation, origin) when is_list(rows) do
+    for row_tuple <- rows do
+      [tags | values] =
+        row_tuple
+        |> Tuple.to_list()
+        |> Enum.map(fn
+          :null -> nil
```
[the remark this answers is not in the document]
icehaunter: So you're both correct and not exactly correct. We indeed don't have a proper casting function that would convert strings coming from `epgsql` to booleans, integers, and so on - but we don't need it. Both `Changes.NewRecord` and Protobuf's `SatOpInsert{}` expect only strings as values in the fields - the protocol coming from PG to us, and our Satellite protobuf can only work with string representations of all values. Thus, it's on us just to make sure that Satellite deserializes the typed values correctly on the client (be that boolean, integer, or others). Whether that translation layer exists - I'm pretty sure it's missing, as we've been using only basic integer/string types so far as that's the only types SQLite has.

components/electric/lib/electric/satellite/ws_server.ex
```diff
@@ -317,13 +383,14 @@ defmodule Electric.Satellite.WsServer do
     end
   end
 
-  defp binary_frames(pb_msgs) when is_list(pb_msgs) do
-    Enum.map(pb_msgs, fn pb_msg -> binary_frame(pb_msg) end)
-  end
+  defp binary_frames(pb_msgs, acc \\ [])
```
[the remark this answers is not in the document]
icehaunter: I also found the original code confusing (it also had a bug), but I addressed it differently in the latest commit in a way that should be optimal: https://github.com/electric-sql/electric/pull/222/commits/0b1bfdc4fe4181ab171c81c1884830e10674e97b

components/electric/test/electric/replication/initial_sync_test.exs
```diff
@@ -23,7 +23,7 @@ defmodule Electric.Replication.InitialSyncTest do
       pg_connector_opts: pg_connector_opts
     } do
       # Verify that the cached LSN is not going to catch up since we don't have any electrified tables.
-      assert :error == conn |> fetch_current_lsn() |> wait_for_cached_lsn_to_catch_up()
+      assert :error == conn |> fetch_current_lsn() |> wait_for_cached_lsn_to_catch_up(false)
```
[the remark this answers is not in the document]
icehaunter: The test was flake because of a timing issue. The change updates the timeout at the top, and makes the error message much more explicit and immediate (the other, not this test, failed, with just `:error didn't match :ok`). I prefer the tests fail a bit louder

components/electric/lib/electric/replication/initial_sync.ex
```diff
@@ -55,6 +57,70 @@ defmodule Electric.Replication.InitialSync do
...
+              {:ok, data} -> {:cont, [{request.id, data} | results]}
+              {:error, reason} -> {:halt, {:error, reason}}
+            end
+          end)
+          |> case do
+            {:error, reason} ->
+              send(parent, {:subscription_init_failed, subscription_id, reason})
+
+            results ->
+              send(parent, {:subscription_data, subscription_id, wal_pos, results})
```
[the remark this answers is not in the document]
icehaunter: I was aware of that, and it shouldn't - all results together are bookended by `SatSubsDataBegin` and `SatSubsDataEnd`, so the client knows that it just needs to consume everything up to the `SatSubsDataEnd`. So the client will be able to map back results for each shape.

components/electric/lib/electric/replication/initial_sync.ex
```diff
@@ -55,6 +57,70 @@ defmodule Electric.Replication.InitialSync do
...
+
+          Enum.reduce_while(requests, [], fn request, results ->
+            case Shapes.ShapeRequest.query_initial_data(request, conn, schema, origin) do
+              {:ok, data} -> {:cont, [{request.id, data} | results]}
+              {:error, reason} -> {:halt, {:error, reason}}
+            end
+          end)
+          |> case do
+            {:error, reason} ->
+              send(parent, {:subscription_init_failed, subscription_id, reason})
```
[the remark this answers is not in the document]
icehaunter: Yeah, that needs to be handled

components/electric/lib/electric/replication/shapes.ex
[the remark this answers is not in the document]
alco: I like this one more.

alco (approved):

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -504,4 +677,61 @@ defmodule Electric.Satellite.Protocol do
...
   end
+
+  defp current_shapes(%State{subscriptions: subscriptions, out_rep: out_rep}) do
+    subscriptions
+    |> Enum.reject(fn {id, _shapes} -> OutRep.subscription_pending?(id, out_rep) end)
+    |> Enum.flat_map(fn {_, shapes} -> shapes end)
+  end
+
+  defp query_subscription_data(requests, state) do
+    id = Utils.uuid4()
```
alco, line 688: Aren't we supposed to use the `subscription_id` obtained from the client's `SatSubsReq` message?

```
# satellite.proto

// (Consumer) Request for new subscriptions
message SatSubsReq {
    // a client-generated identifier to track the subscription
    string subscription_id = 1;    <== should be the same ID as in SatSubsResp

    // Shape requests
    repeated SatShapeReq shape_requests = 2;
}

// (Producer) Response for a subscription request
message SatSubsResp {
    // identifier of the subscription this response refers to
    string subscription_id = 1;    <== should be the same ID as in SatSubsReq

    [...]
```

icehaunter: Good catch, didn't update the assertion in tests

components/electric/lib/electric/satellite/ws_server.ex
```diff
@@ -215,13 +216,43 @@ defmodule Electric.Satellite.WsServer do
...
+    Logger.error(
+      "Couldn't retrieve initial data for subscription #{subscription_id}, with reason #{inspect(reason)}"
+    )
+
+    error = %SatSubsDataError{
+      subscription_id: subscription_id,
+      message: "Could not retrieve initial data for subscription"
+    }
+
+    # In any case we need to remove the subscription itself since it'll
```
alco, line 243: Sentence cut off prematurely.

components/electric/test/electric/satellite/ws_server_test.exs
```diff
@@ -591,8 +654,11 @@ defmodule Electric.Satellite.WsServerTest do
 
   defp build_events(changes, lsn) do
     [
-      {%Changes.Transaction{changes: List.wrap(changes), commit_timestamp: DateTime.utc_now()},
-       lsn}
+      {%Changes.Transaction{
+         changes: List.wrap(changes),
+         commit_timestamp: DateTime.utc_now(),
+         xid: lsn
```
alco, line 660: If this is intentional, it needs an explanation of why `lsn` is assigned to `xid`.

components/electric/test/support/setup_helpers.ex
```diff
@@ -24,12 +24,15 @@ defmodule ElectricTest.SetupHelpers do
         end
       )
 
+    stmts =
+      Regex.split(~r/((?:[^;'"]*(?:"(?:\\.|[^"])*"|'(?:\\.|[^'])*')[^;'"]*)+)|;/, sql, trim: true)
```
alco, line 28: Could you add a comment explaining this splitting behaviour in plain English?

e2e/tests/3.5_fresh_node_satellite_receives_all_migrations_during_initial_sync.lux
```diff
@@ -64,18 +64,6 @@
...
-    \{"\$$type":"Electric.Satellite.v[0-9_]+.SatTransOp","begin":\{.*\}\},\
-    \{"\$$type":"Electric.Satellite.v[0-9_]+.SatTransOp","insert":\{.*\}\},\
-    \{"\$$type":"Electric.Satellite.v[0-9_]+.SatTransOp","insert":\{.*\}\},\
-    \{"\$$type":"Electric.Satellite.v[0-9_]+.SatTransOp","insert":\{.*\}\},\
-    \{"\$$type":"Electric.Satellite.v[0-9_]+.SatTransOp","insert":\{.*\}\},\
-    \{"\$$type":"Electric.Satellite.v[0-9_]+.SatTransOp","commit":\{.*\}\}
-
-  ?apply incoming changes for LSN: \d+
-
-  -Received message
```
alco, line 78: This line needs to remain to verify that no data transactions are delivered to the client.

e2e/tests/3.6_node_satellite_does_sync_on_subscribe.lux
```diff
@@ -0,0 +1,58 @@
...
+[include _satellite_macros.luxinc]
+
+[invoke setup]
+
+[invoke setup_client 1 "electric_1" 5133]
+
+[shell pg_1]
+    [local sql=
+        """
+        CREATE TABLE public.items (
```
alco, line 12: Does this table need to have so many columns? We should strive to keep the test code as a simple as possible and only define things that are relevent to the test code.

In this test, I only see `id` and `content` columns referenced.

e2e/tests/3.7_node_satellite_can_delete_freshly_synced_rows.lux
```diff
@@ -0,0 +1,53 @@
...
+[include _satellite_macros.luxinc]
+
+[invoke setup]
+
+[invoke setup_client 1 "electric_1" 5133]
+
+[shell pg_1]
+    [local sql=
+        """
+        CREATE TABLE public.items (
```
alco, line 12: Same feedback here as in my previous comment.

alco (approved):
Great work :+1:

e2e/tests/3.6_node_satellite_does_sync_on_subscribe.lux
icehaunter: The TS client's Primsma schema expects the columns to exist to work (it fails on querying by trying to query unknown columns otherwise)

PR #223, opened 2023-07-06
feat(satellite) improve subscribe request

balegas:
- Distinguish between subscribe request errors and subscription data errors
- Client-side generated subscription ids
- (Piggy-back) load subscriptions manager state from storage

### DIFF
--- a/clients/typescript/src/satellite/shapes/cache.ts
+++ b/clients/typescript/src/satellite/shapes/cache.ts
@@ -3,6 +3,7 @@ import {
   SatShapeDataBegin,
   SatSubsDataBegin,
   SatSubsDataError,
+  SatSubsReq,
   SatSubsResp,
   SatTransOp,
 } from '../../_generated/protocol/satellite'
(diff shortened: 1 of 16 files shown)


clients/typescript/src/satellite/shapes/cache.ts
```diff
@@ -205,11 +214,31 @@ export class SubscriptionsDataCache extends EventEmitter {
     throw error
   }
 
-  subscriptionError(msg: SatSubsError): never {
+  // It is safe to reset the cache state without throwing.
+  // However, if message is unexpected, we emit the error
+  subscriptionError(): void {
+    if (this.remainingShapes.size == 0 || !this.requestedSubscription) {
```
kevin-dp, line 220: better use strict equality `this.remainingShapes.size === 0`

clients/typescript/src/satellite/shapes/manager.ts
```diff
@@ -97,4 +101,10 @@ export class InMemorySubscriptionsManager
   serialize(): string {
     return JSON.stringify(this.subToShapes)
   }
+
+  // TODO: input validation
```
kevin-dp, line 105: could use zod if the validations are not straightforward

kevin-dp (approved):
Great work!
Found everything very readable

PR #225, opened 2023-07-08
fix(electric): Initial sync of migrations

alco:
Closes VAX-734.

These changes are based on https://github.com/electric-sql/electric/pull/222 so that PR needs to be merged first.

This PR fixes the issue described in the ticket as follows:

  * `Extension.migration_history()` is updated to include `txid` and `txts` of migration transactions in its return value.
  * Those `txid` and `txts` are included in the `Transaction` structs created by InitialSync.
  * The `txid` of the most recent migration is stored in WsServer's state.
  * When WsServer receives a transaction from CachedWal.Producer, it checks whether it is a migration and whether its `xid` is greater than the `txid` stored in its state. If not, the transaction is dropped because it was already included in the initial sync previously.

NOTE: If a newly connected client doesn't do initial sync (i.e. its a resuming client), the `txid` in WsServer's state will be 0, so every subsequent migration transaction will be delivered to the client.

Remaining work:

  * ~[ ] Write an e2e test that verifies migration transaction flitering.~ Addressed separately in https://github.com/electric-sql/electric/pull/231.
  * [x] Wait for #222 to get merged and rebase this PR on top of `main`.

### DIFF
--- a/components/electric/lib/electric/postgres/extension.ex
+++ b/components/electric/lib/electric/postgres/extension.ex
@@ -33,8 +33,6 @@ defmodule Electric.Postgres.Extension do
   @electrified_index_table electric.(@electrified_index_relation)
 
   @all_schema_query ~s(SELECT "schema", "version", "migration_ddl" FROM #{@schema_table} ORDER BY "version" ASC)
-  @migration_history_query ~s(SELECT "version", "schema", "migration_ddl" FROM #{@schema_table} ORDER BY "version" ASC)
-  @partial_migration_history_query ~s(SELECT "version", "schema", "migration_ddl" FROM #{@schema_table} WHERE "version" > $1 ORDER BY "version" ASC)
   @current_schema_query ~s(SELECT "schema", "version" FROM #{@schema_table} ORDER BY "id" DESC LIMIT 1)
   @schema_version_query ~s(SELECT "schema", "version" FROM #{@schema_table} WHERE "version" = $1 LIMIT 1)
   # FIXME: VAX-600 insert into schema ignoring conflicts (which I think arise from inter-pg replication, a problem
@@ -49,6 +47,32 @@ defmodule Electric.Postgres.Extension do
   @publication_name "electric_publication"
   @slot_name "electric_replication_out"
   @subscription_name "electric_replication_in"
+
+  defp migration_history_query(after_version) do
+    where_clause =
+      if after_version do
+        ~s|WHERE #{@version_table}.txid > (SELECT txid FROM #{@version_table} WHERE version = $1)|
+      else
+        # Dummy condition just to keep the $1 parameter in the query.
+        ~s|WHERE $1::text IS NULL|
+      end
+
+    ~s"""
+    SELECT
+      #{@version_table}.txid,
+      #{@version_table}.txts,
+      #{@version_table}.version,
+      #{@schema_table}.schema,
+      #{@schema_table}.migration_ddl
+    FROM
+      #{@version_table}
+    JOIN
+      #{@schema_table} USING (version)
+    #{where_clause}
+    ORDER BY
+      #{@version_table}.txid ASC
+    """
+  end
 
   def schema, do: @schema
   def ddl_table, do: @ddl_table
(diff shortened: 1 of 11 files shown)


components/electric/lib/electric/replication/initial_sync.ex
```diff
@@ -24,36 +23,32 @@ defmodule Electric.Replication.InitialSync do
...
         end
 
-      build_transaction(connector_opts, records, lsn, commit_timestamp)
+      transaction = %Transaction{
+        xid: txid,
+        changes: records,
+        commit_timestamp: txts,
+        origin: origin,
+        publication: publication,
+        lsn: 0,
```
alco, line 47: Not sure if it'd be appropriate to fetch the latest LSN and use that instead of 0.

The only problem with using 0 is that a client who connects for the first time, receives migration transactions but no data and then reconnects, will go through the initial sync of migrations again (though it won't receive duplicate transactions the second time around once https://github.com/electric-sql/electric/pull/226 is merged.

A potential problem with putting the latest LSN into these migration transactions could be inadvertent filtering of a subsequent migration coming from the cached WAL after the initial sync has finished.

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -50,6 +48,32 @@ defmodule Electric.Postgres.Extension do
...
+      #{@version_table}.version,
+      #{@schema_table}.schema,
+      #{@schema_table}.migration_ddl
+    FROM
+      #{@version_table}
+    JOIN
+      #{@schema_table} USING (version)
+    #{where_clause}
+    ORDER BY
+      #{@version_table}.txid ASC
```
alco, line 73: We could have chosen to sort by version instead but there's a potential problem with versions: app developer may set any version for their migration (via `SELECT electric.migration_version('...")`) and we cannot ensure that those versions are monotonically increasing from older migrations to newer ones. In other words, a new migration may end up with a version that sorts _before_ one or more previously applied migrations.

Although that should be avoided in practice because it may caused other issues, even outside Electric, but just to be on the safe side, we sort by txid instead. Because migration transactions are applied serially, we can use their `txid`s as a safe, monotonically increasing criterion for sorting.

kevin-dp: Interesting, the CLI does depend on the migration versions being in chronological order because it fetches a zip file containing the migrations from Electric. So in this zip file it does not know the order of migrations so it has to sort the migrations based on their version. This problem will be fixed later when Electric exposes the migrations in JSON format instead of a zip file.

alco: Great point!

I mean, we could assume that versions are ordered correctly (i.e. migrations are always applied in version order) and avoid this complication with reliance on txid. I'm going to flag it up as a discussion point with the team.

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -515,6 +538,18 @@ defmodule Electric.Satellite.Protocol do
...
 
+  # TODO(alco): add a top-level field to %Transaction{} to indicate whether it is a migration or not.
+  # Or define a separate struct for migrations.
+  defp skip_migration_tx?(
+         %Transaction{xid: xid, changes: [change | _]},
+         last_seen_migration_xid
+       )
+       when is_migration_relation(change.relation) and xid <= last_seen_migration_xid do
+    true
+  end
```
icehaunter, line 549: I think this is wrong, since we explicitly decided to allow "mixed" transactions (ones with both DDL and DML statements) This shouldn't skip transaction wholesale, it should filter out migrations from changes if they've already been seen

icehaunter (changes requested):
One important question from me, otherwise I think it's good

components/electric/lib/electric/satellite/protocol.ex
alco: This function is meant to skip migrations coming from cached WAL during a very narrow time window: when a migration transaction has been committed while a client is performing initial sync, such that the client receives it as part of initial sync, and then the client subscribes to changes from cached WAL at an LSN that precedes this last transaction's LSN. If this happens, this same transaction will be coming in once again from the cached WAL producer.

At any other time, including a new client connection that doesn't go through the initial sync, this function should be a no-op. I could do a better job of communicating that in the code. I think what I need to add is this: once the current txid coming from the cached WAL producer matches `state.out_rep.last_seen_migration_xid` stored just after the initial sync, the `skip_migration_tx?()` function should explicitly be made a no-op.

Now, regarding whether it's appropriate to drop the full transactions if it is to be skipped. Right after the initial sync, the client may not yet have any active subcriptions. So it doesn't matter if we drop the whole transaction, including its DML.

If the client manages to send a SatSubReq before the duplicate transaction comes in from the cached WAL producer, the client will get the data for the shapes it's subscribing to via the regular path of fetching initial shape data. So the duplicate transaction has to be dropped anyhow because the client will have already received the data part of it.

icehaunter (approved):

PR #226, opened 2023-07-08
feat(satellite,electric): Rely on client-supplied schema version to push only missing transactions to the client

alco:
Closes VAX-733.

These changes are based on https://github.com/electric-sql/electric/pull/218 so that PR needs to be merged first.

This PR updates the client to include its local `schemaVersion` in the `SatInStartReplicationReq` message which the server then uses to only fetch migrations that the client is missing.

Remaining work:

  * ~[ ] Write an e2e test that verifies the server only sends migrations that the client doesn't already have.~ It seemed like a good idea initially, but after struggling with implementing proper setup for discarding client's local state, I've abandoned it. Currently, this functionality is covered with TypeScript tests on the client and Elixir tests on the server.
  * [x] Wait for https://github.com/electric-sql/electric/pull/218 to get merged and rebase this PR on top of main.

### DIFF
--- a/clients/typescript/src/migrators/bundle.ts
+++ b/clients/typescript/src/migrators/bundle.ts
@@ -52,7 +52,7 @@ export class BundleMigrator implements Migrator {
     return unapplied.length
   }
 
-  async queryApplied(): Promise<MigrationRecord[]> {
+  async migrationsTableExists(): Promise<boolean> {
     // If this is the first time we're running migrations, then the
     // migrations table won't exist.
     const tableExists = `
@@ -64,18 +64,40 @@ export class BundleMigrator implements Migrator {
       sql: tableExists,
       args: [this.tableName],
     })
-    if (tables.length === 0) {
+
+    return tables.length > 0
+  }
+
+  async queryApplied(): Promise<MigrationRecord[]> {
+    if (!(await this.migrationsTableExists())) {
       return []
     }
 
-    // The migrations table exists, so let's query the version of
-    // the previously applied migrations.
     const existingRecords = `
       SELECT version FROM ${this.tableName}
         ORDER BY id ASC
     `
     const rows = await this.adapter.query({ sql: existingRecords })
     return rows as unknown as MigrationRecord[]
+  }
+
+  // Returns the version of the most recently applied migration
+  async querySchemaVersion(): Promise<string | undefined> {
+    if (!(await this.migrationsTableExists())) {
+      return
+    }
+
+    const schemaVersion = `
+      SELECT version FROM ${this.tableName}
+        ORDER BY version DESC
+        LIMIT 1
+    `
+    const rows = await this.adapter.query({ sql: schemaVersion })
+    if (rows.length === 0) {
+      return
+    }
+
+    return (rows[0] as MigrationRecord).version
   }
 
   async validateApplied(
--- a/clients/typescript/src/migrators/bundle.ts
+++ b/clients/typescript/src/migrators/bundle.ts
@@ -87,8 +87,11 @@ export class BundleMigrator implements Migrator {
       return
     }
 
+    // The hard-coded version '0' below corresponds to the version of the internal migration defined in `schema.ts`.
+    // We're ignoring it because this function is supposed to return the application schema version.
     const schemaVersion = `
       SELECT version FROM ${this.tableName}
+        WHERE version != '0'
         ORDER BY version DESC
         LIMIT 1
     `
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -479,7 +479,7 @@ export class SatelliteProcess implements Satellite {
 
   async _connectAndStartReplication(
     opts?: SatelliteReplicationOptions
-  ): Promise<void | SatelliteError> {
+  ): Promise<void> {
     Log.info(`connecting and starting replication`)
 
     if (!this._authState) {
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -491,12 +491,18 @@ export class SatelliteProcess implements Satellite {
       await this.client.connect()
       await this.client.authenticate(authState)
 
+      const schemaVersion = await this.migrator.querySchemaVersion()
+
       // Fetch the subscription IDs that were fulfilled
       // such that we can resume and inform Electric
       // about fulfilled subscriptions
       const subscriptionIds = this.subscriptions.getFulfilledSubscriptions()
 
-      await this.client.startReplication(this._lsn, subscriptionIds)
+      await this.client.startReplication(
+        this._lsn,
+        schemaVersion,
+        subscriptionIds
+      )
     } catch (error: any) {
       if (
         error.code == SatelliteErrorCode.BEHIND_WINDOW &&
--- a/clients/typescript/test/satellite/client.test.ts
+++ b/clients/typescript/test/satellite/client.test.ts
@@ -208,6 +208,28 @@ test.serial('replication start sends FIRST_LSN', async (t) => {
   })
 })
 
+test.serial('replication start sends schemaVersion', async (t) => {
+  await connectAndAuth(t.context)
+  const { client, server } = t.context
+
+  return new Promise(async (resolve) => {
+    server.nextResponses([
+      (data?: Buffer) => {
+        const msgType = data!.readUInt8()
+        t.assert(
+          msgType == getTypeFromString(Proto.SatInStartReplicationReq.$type)
+        )
+
+        const req = decode(data!) as Proto.SatInStartReplicationReq
+        t.assert(req.schemaVersion === '20230711')
+
+        resolve()
+      },
+    ])
+    await client.startReplication(new Uint8Array(), '20230711')
+  })
+})
+
 test.serial('replication start failure', async (t) => {
   await connectAndAuth(t.context)
   const { client, server } = t.context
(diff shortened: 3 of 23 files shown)


kevin-dp (approved):
Good work, looks good to me! Just make sure the e2e tests pass before we merge

clients/typescript/test/satellite/client.test.ts
```diff
@@ -208,6 +214,57 @@ test.serial('replication start sends FIRST_LSN', async (t) => {
...
 
+test.serial('replication start sends empty schemaVersion when there are no migrations', async(t) => {
+  await connectAndAuth(t.context)
+  const { satellite, client, server } = t.context
+  console.log("satellite: ", satellite)
+
+  const createTable = {
+    table: {
+      name: 'NewTable',
+      columns: [{name: 'id', sqliteType: 'text', pgType: undefined } as Proto.SatOpMigrate_Column],
```
alco, line 225: I took this code from `test/satellite/process.migration.test.ts` but the TypeScript compiler keeps finding typing issues if I omit explicit casting.

clients/typescript/test/satellite/client.test.ts
```diff
@@ -208,6 +214,57 @@ test.serial('replication start sends FIRST_LSN', async (t) => {
...
+
+  const migrationTx = {
+    origin: 'remote',
+    commit_timestamp: Long.fromNumber((new Date).getTime()),
+    changes: [createTable],
+    lsn: new Uint8Array(),
+    // starts at 3, because the app already defines 2 migrations
+    // (see test/support/migrations/migrations.js)
+    // which are loaded when Satellite is started
+    migrationVersion: '3',
```
alco, line 244: I took this code from `test/satellite/process.migration.test.ts` but it doesn't appear the test database has any migrations in it like the comment above this line claims.

clients/typescript/test/satellite/client.test.ts
```diff
@@ -208,6 +214,57 @@ test.serial('replication start sends FIRST_LSN', async (t) => {
...
+
+  return new Promise(async (resolve) => {
+    server.nextResponses([
+      (data?: Buffer) => {
+        const msgType = data!.readUInt8()
+        if (
+          msgType == getTypeFromString(Proto.SatInStartReplicationReq.$type)
+        ) {
+          const req = decode(data!) as Proto.SatInStartReplicationReq
+          t.deepEqual(req.schemaVersion, undefined)
```
alco, line 258: I wanted to verify this condition first, then apply a migration to the local DB and verify that `schemaVersion` is no longer `undefined`. I didn't get to that last part yet due to issues described in my other comments above.

clients/typescript/src/migrators/bundle.ts
```diff
@@ -78,6 +81,25 @@ export class BundleMigrator implements Migrator {
...
 
+  // Returns the version of the most recently applied migration
+  async querySchemaVersion(): Promise<string | undefined> {
+    if (!(await this.migrationsTableExists())) {
+      return
+    }
+
+    const schemaVersion = `
+      SELECT version FROM ${this.tableName}
+        ORDER BY id DESC
```
alco, line 92: I'm ordering by `id` here to ensure the same order in which the migrations were applied. In principle, we could order by `version` instead, but there's a possibility (even though it's vanishingly small) that migrations are applied to the PG database in a different order than the one defined by their versions. See https://github.com/electric-sql/electric/pull/225/files#r1257996658 for additional context.

clients/typescript/test/satellite/client.test.ts
kevin-dp: The problem is that `satellite._applyTransaction()` expects an object of type `Transaction`:
```
export type Transaction = {
  commit_timestamp: Long
  lsn: LSN
  changes: Change[]
  // This field is only set by transactions coming from Electric
  origin?: string
  migrationVersion?: string // the Postgres version number if this is a migration
}
```

The `changes` is of type `Change[]` which is:
```
export type Change = DataChange | SchemaChange

export type SchemaChange = {
  table: MigrationTable // table affected by the schema change
  migrationType: SatOpMigrate_Type
  sql: string
}

export type MigrationTable = Omit<SatOpMigrate_Table, '$type'>
```

So, we don't need to provide a `$type` field on `SatOpMigrate_Table` but inside `SatOpMigrate_Table` the `columns` and `fks` properties are of type `SatOpMigrate_Column[]` and `SatOpMigrate_ForeignKey[]` respectively. Both of these types expect a `$type` field, which you are not providing in your `columns` property.

However, our code does not actually use these `$type` fields so we can remove them from the type.
Changing the `MigrationTable` type as follows solves the issue:
```
export type MigrationTable = Omit<SatOpMigrate_Table, '$type' | 'columns' | 'fks'> & {
  columns: Array<Omit<SatOpMigrate_Column, '$type'>>
  fks: Array<Omit<SatOpMigrate_ForeignKey, '$type'>>
}
```

clients/typescript/test/satellite/client.test.ts
kevin-dp: In the `process.migration.test.ts` we call `await satellite.start(authState)` before every test.
Satellite then calls `migrator.up()` which loads the migrations defined in `test/support/migrations/migrations.js` and hence the DB contains those 2 migrations.

However, in your test, those migrations are never loaded, so you can use `migrationVersion: '1'`.

clients/typescript/test/satellite/client.test.ts
kevin-dp: i would replace the if test by an assertion, `t.assert(msgType == getTypeFromString(Proto.SatInStartReplicationReq.$type))`.
Also, `deepEqual` is useful to compare objects for structural equality instead of referential equality.
In this case, you're comparing primitive values so you can just do `t.assert(req.schemaVersion === undefined)`.
No need to call `t.pass()` here.

icehaunter:
It seems like relevant e2e tests are still failing

alco:
@icehaunter Fixed.

components/electric/lib/electric/satellite/ws_server.ex
```diff
@@ -189,7 +189,8 @@ defmodule Electric.Satellite.WsServer do
   # client has connected which needs to perform the initial sync of migrations and the current database state before
   # subscribing to the replication stream.
   def websocket_info({:perform_initial_sync_and_subscribe, msg}, %State{} = state) do
-    migration_transactions = InitialSync.migrations_since(nil, state.pg_connector_opts)
+    %SatInStartReplicationReq{schema_version: schema_version} = msg
+    migration_transactions = InitialSync.migrations_since(schema_version, state.pg_connector_opts)
```
icehaunter, line 193: Not sure where to leave this comment, but:

I'm not sure, I'd rather deny connection if the supplied migration version isn't known. It means the client is either ahead and we can't do replication or misbehaving (or DB has been reset) and we can't do replication. That means I would rather do a "is a known migration?" check on connection against the cache, not in this "continuation" handler

components/electric/test/electric/satellite/ws_server_test.exs
```diff
@@ -338,6 +339,24 @@ defmodule Electric.Satellite.WsServerTest do
       end)
     end
 
+    test "no migrations are delivered as part of initial sync if PG has no electrified tables",
```
icehaunter, line 342: I don't see a test to check if the proper migrations are sent if `schema_version` is present in ReplicationReq

clients/typescript/src/satellite/process.ts
```diff
@@ -475,7 +475,7 @@ export class SatelliteProcess implements Satellite {
     }
   }
 
-  async _connectAndStartReplication(
+  _connectAndStartReplication(
     opts?: SatelliteReplicationOptions
   ): Promise<void | SatelliteError> {
```
icehaunter, line 480: Just while we're here

```suggestion
  ): Promise<void> {
```

clients/typescript/src/satellite/process.ts
```diff
@@ -488,7 +488,10 @@ export class SatelliteProcess implements Satellite {
     return this.client
       .connect()
       .then(() => this.client.authenticate(authState))
-      .then(() => this.client.startReplication(this._lsn))
+      .then(() => this.migrator.querySchemaVersion())
```
icehaunter, line 491: optional, feel free to skip: I'd appreciate a rewrite of this chain to `await`s

icehaunter (changes requested):
Looks nice, but I have one important question and some nitpicks

clients/typescript/src/satellite/process.ts
alco: It's curious that I had this call as an `await` first but Kevin suggested that I include it in this chain of `then`s. But now Kevin is rewriting all of it as a sequence of `await` calls in https://github.com/electric-sql/electric/pull/247/commits/4ea8afecfadfbdcb163e8c223517b94e1e9a1341#diff-4ae478716dc5aae8eca5e417ee51fbb261dc105b923b67eff253f2a25ae42d95L490-R508.

icehaunter: Lol. it's either all `await` or all `then`, not a good practice to mix unless it's mostly awaits and a one-off `then`. Don't worry about this then for now, whoever merges second will get to resolve the conflict :D

components/electric/test/electric/satellite/ws_server_test.exs
alco: This is a good point. Added one test. Writing it turned out to be a huge pain in the ass because of the whole mocking of things that are involved in maintaining the schema cache and migration versions.

PR #228, opened 2023-07-10
feat(DAL, Satellite): add sync shape functionality to DAL

kevin-dp:
This PR extends the DAL with functionality to sync shapes.
The signature of this `sync` method is:
```
sync<T extends SyncInput<Include>>(i?: T): Promise<Sub>

interface SyncInput<Include> {
  include?: Include
}

type Sub = {
  dataReceived: Promise<void>
}
```

The returned promise resolves when the subscription request is accepted by Electric.
The value to which it resolves is an object containing a `dataReceived` property that is again a promise.
The `dataReceived` promise resolves when the subscription request is fulfilled, i.e. all data has been loaded.

Example usage:
```
// Sync only the `Post` table
const { dataReceived } = await db.Post.sync()
await dataReceived

// Sync the `Post` table and the tables for the related `author` and `profile` fields
const { dataReceived } = await Post.sync({
      include: {
        author: {
          include: {
            profile: true
          }
        }
      }
    })

await dataReceived
```

Note that Satellite cannot handle concurrent subscription requests.
Hence, it is important to always await the nested promise before making a new subscription request!!
For example:

```
await db.Post.sync() // problematic
await db.User.sync() // this subscription request may be issued before the previous one was fulfilled!
```

**For reviewers:** the DAL has a shape manager abstraction that keeps track of which shapes have been synced etc. I chose to expose a single shared shape manager instance which every DAL table uses. As such that instance does not need to be passed to every table separately. Not sure if that is a good idea, let me know what you think.

### DIFF
--- a/clients/typescript/src/client/input/findInput.ts
+++ b/clients/typescript/src/client/input/findInput.ts
@@ -1,5 +1,3 @@
-//export type SelectInput<T> = { [field in keyof T]?: boolean }
-
 import {
   NarrowInclude,
   NarrowOrderBy,
--- a/clients/typescript/src/client/input/syncInput.ts
+++ b/clients/typescript/src/client/input/syncInput.ts
@@ -1,0 +1,3 @@
+export interface SyncInput<Include> {
+  include?: Include
+}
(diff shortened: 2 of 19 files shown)


icehaunter (approved):

thruflo:
Hey @kevin-dp just a note on this (prompted by documenting it): I'm not a huge fan of `dataReceived` as a variable name here, just because it's a spelling mistake waiting to happen.

Perhaps:

```tsx
const { resolveStatus } = await db.Post.sync()
```

?

Also, perhaps `Sub` should be the beginning of a `Shape` interface?

PR #230, opened 2023-07-10
fix(electric): tag serialization wasn't working assigning `null` to tags

icehaunter:
Internal representation of tags on PG is done in a way that any PG-originating tag has it's origin component set to NULL, and then NULL origins are ordered higher than any strings. When PG-originating tags are sent to Satellites, we preprocess tags to replace NULL values with pg origin name. But the inverse path has an error, whereby on a return trip the pg origin in observed tags is not set back to NULL before being set to PG like it should.

Closes VAX-762

### DIFF
--- a/components/electric/test/electric/postgres/shadow_table_transformation_test.exs
+++ b/components/electric/test/electric/postgres/shadow_table_transformation_test.exs
@@ -22,7 +22,7 @@ defmodule Electric.Postgres.ShadowTableTransformationTest do
     "_modified_columns_bit_mask"
   ]
 
-  describe "split_change_into_main_and_shadow/3" do
+  describe "split_change_into_main_and_shadow/4" do
     test "INSERT operation gets converted and split correctly" do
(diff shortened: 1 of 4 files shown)


components/electric/test/electric/postgres/shadow_table_transformation_test.exs
```diff
@@ -110,6 +113,36 @@ defmodule Electric.Postgres.ShadowTableTransformationTest do
                  ShadowTableTransformation.serialize_pg_array(["f", "f"])
              }
     end
+
+    test "origins in tags are set to nil correctly" do
```
alco, line 117: I can't put this test name to the code inside the test. Where does it check that origins in tags are set to nil?

icehaunter: Updated the body to reflect the name

alco (approved):

PR #231, opened 2023-07-10
chore(electric): Add an e2e test that verifies migration deduplication

alco:
Follow-up to https://github.com/electric-sql/electric/pull/225 but based on changes https://github.com/electric-sql/electric/pull/226.

This is a rough implementation that I'm not too happy about. Would love to hear ideas for a simpler approach to testing that the server discards repeat migrations after initial sync.

### DIFF
--- a/components/electric/lib/electric/satellite/ws_server.ex
+++ b/components/electric/lib/electric/satellite/ws_server.ex
@@ -207,7 +207,10 @@ defmodule Electric.Satellite.WsServer do
     migration_transactions = InitialSync.migrations_since(schema_version, state.pg_connector_opts)
     {msgs, state} = Protocol.handle_outgoing_txs(migration_transactions, state)
 
-    max_txid = Enum.max_by(migration_transactions, fn {tx, _offset} -> tx.xid end, fn -> 0 end)
+    max_txid =
+      migration_transactions
+      |> Enum.map(fn {tx, _offset} -> tx.xid end)
+      |> Enum.max(fn -> 0 end)
 
     state =
       state
--- a/components/electric/lib/electric/satellite/ws_server.ex
+++ b/components/electric/lib/electric/satellite/ws_server.ex
@@ -189,7 +189,8 @@ defmodule Electric.Satellite.WsServer do
   # client has connected which needs to perform the initial sync of migrations and the current database state before
   # subscribing to the replication stream.
   def websocket_info({:perform_initial_sync_and_subscribe, msg}, %State{} = state) do
-    migration_transactions = InitialSync.migrations_since(nil, state.pg_connector_opts)
+    %SatInStartReplicationReq{schema_version: schema_version} = msg
+    migration_transactions = InitialSync.migrations_since(schema_version, state.pg_connector_opts)
     {msgs, state} = Protocol.handle_outgoing_txs(migration_transactions, state)
 
     lsn = CachedWal.Api.get_current_position()
--- a/components/electric/lib/electric/satellite/ws_server.ex
+++ b/components/electric/lib/electric/satellite/ws_server.ex
@@ -189,11 +189,18 @@ defmodule Electric.Satellite.WsServer do
   # client has connected which needs to perform the initial sync of migrations and the current database state before
   # subscribing to the replication stream.
   def websocket_info({:perform_initial_sync_and_subscribe, msg}, %State{} = state) do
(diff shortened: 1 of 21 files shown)


protocol/satellite.proto
```diff
@@ -112,6 +112,8 @@ message SatInStartReplicationReq {
         // Asks receiver to start replication from the last position in the log,
         // whatever this position is. Used for tests only.
         LAST_LSN = 4;
+        // This option is only used by some E2E tests.
+        PAUSE_DURING_INITIAL_SYNC = 5;
```
icehaunter, line 116: I really worry that we're exposing such a function to actual clients... Doesn't seem like it should backfire, but it might...

components/electric/lib/electric/satellite/ws_server.ex
```diff
@@ -520,4 +536,17 @@ defmodule Electric.Satellite.WsServer do
       Map.update!(state, :out_rep, &OutRep.remove_next_pause_point/1)
     )
   end
+
+  # This function is only used in some e2e tests.
+  defp wait_until_lsn_advances(lsn) do
+    CachedWal.Api.request_notification(lsn)
+
+    receive do
+      {:cached_wal_notification, _, :new_segments_available} ->
```
icehaunter, line 545: you probably should use the `ref` returned by `request_notification`. Selective receive when one of the elements is a ref is faster

e2e/tests/3.8_node_satellite_no_duplicate_migrations_after_initial_sync.lux
```diff
@@ -0,0 +1,79 @@
...
+
+  ?Received message \{"\$$type":"Electric.Satellite.v[0-9_]+.SatRelation","schemaName":"public",.*"tableName":"users"
+  ?Received message \{"\$$type":"Electric.Satellite.v[0-9_]+.SatOpLog","ops":\[\
+    \{"\$$type":"Electric.Satellite.v[0-9_]+.SatTransOp","begin":\{.*\}\},\
+    \{"\$$type":"Electric.Satellite.v[0-9_]+.SatTransOp","migrate":\{.*,"version":"$migration2_vsn",.*,"table":\{.*,"name":"users"
+
+# Verify that the server has filtered out the duplicate migration
+[shell electric]
+  ?Filtering transaction %Electric.Replication.Changes.Transaction\{\
+     ack_fn: nil, \
```
icehaunter, line 60: Map key ordering is not guaranteed, and is explicitly "can change from run to run" under OTP26. We can use the new "sort map keys" option on inspect if we move there, but this test is likely to fail after rebase on main w/ OTP25

e2e/tests/3.8_node_satellite_no_duplicate_migrations_after_initial_sync.lux
```diff
@@ -0,0 +1,79 @@
...
+# Make sure the server keeps streaming migrations to the client after the initial sync is done.
+[shell pg_1]
+  [local sql=
+    """
+    ALTER TABLE users ADD COLUMN foo TEXT DEFAULT 'bar';
+    """]
+  [invoke migrate_pg $migration3_vsn $sql]
+
+[shell satellite_1]
+  ?Received message \{"\$$type":"Electric.Satellite.v[0-9_]+.SatRelation","schemaName":"public",.*"tableName":"users"
```
icehaunter, line 73: I don't think we should match on this, as the client doesn't use the `SatRelation` messages for anything other than parsing the incoming stream (as it should). We should only care about the `migrate`s here

icehaunter (changes requested):
I don't know, it's good that the e2e highlighted an actual error, but I'm very much unsure I like the addition to the proto.

Can we cover the same case on a unit-test level and not do the e2e if it's so hard to set up properly? Or, if you really insist on e2e being present, let's at least use our Elixir WS client and just verify that we get correct messages instead of adding all this logic to the TS client just to pass a single e2e option. I definitely don't see any reason to use Node client just to check if the correct message is sent as you're doing here.

alco:
@icehaunter I've opened [a new pull request](https://github.com/electric-sql/electric/pull/255) that adds an e2e test using SatelliteWsClient. I tried writing a unit test first but that failed due to the amount of mocking we're using there. I had to have the real DB schema, SchemaCache, etc. up and running to be able to verify migration transaction deduplication.

PR #233, opened 2023-07-11
chore: fix types in tests

icehaunter:
- Removes `| SatelliteError` from return types in Satellite client (we properly raise the errors, they shouldn't have been in the type)
- Uses proper AVA test type casting to not have `t.context as ContextType` in every test


kevin-dp (approved):
Great stuff! 🙏

PR #234, opened 2023-07-11
feat(electric): support unsubscribe messages

icehaunter:
Adds support for the `SatUnsubsReq` message, that stops ongoing subscriptions. That means the client will no longer receive data for that particular subscription, and any yet-unfulfilled subscription listed in the unsubscribe request will get immediately cancelled and their data discarded.

Note: E2E tests are missing because although `unsubscribe` call can be issued by the client, we're yet to expose a public interface for it to e2e test.

### DIFF
--- a/components/electric/test/test_helper.exs
+++ b/components/electric/test/test_helper.exs
@@ -3,7 +3,7 @@ if System.get_env("INTEGRATION") do
 else
   Mox.defmock(Electric.Replication.MockPostgresClient, for: Electric.Replication.Postgres.Client)
   ExUnit.configure(capture_log: true, exclude: :integration, timeout: 15_000)
-  Logger.configure(level: :debug)
+  Logger.configure(level: :info)
 end
 
 File.rm(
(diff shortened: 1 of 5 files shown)


components/electric/test/test_helper.exs
```diff
@@ -3,7 +3,7 @@ if System.get_env("INTEGRATION") do
 else
   Mox.defmock(Electric.Replication.MockPostgresClient, for: Electric.Replication.Postgres.Client)
   ExUnit.configure(capture_log: true, exclude: :integration, timeout: 15_000)
-  Logger.configure(level: :info)
+  Logger.configure(level: :debug)
```
alco, line 6: Did you intentionally commit this change? From what I recall, it makes log output quite noisy, so may not be appropriate as the default value.

alco (approved):

components/electric/test/test_helper.exs
icehaunter: no, by accident, good catch

PR #235, opened 2023-07-11
client: rename dbSchema to schema.

thruflo:
Hey, mind if I rename this? It just makes all the initial instantiation examples much sweeter to be:

```tsx
await electrify(conn, schema, config)
```

As opposed to:

```tsx
await electrify(conn, dbSchema, config)
```

I'll obviously update the docs. I think this catches all the current uses, certainly in the monorepo.


kevin-dp (approved):
Indeed, that looks better 💯

PR #236, opened 2023-07-11
feat(electric): allow subscription resume on reconnect

icehaunter:
Allow subscription resume on reconnect by accepting a list of subscriptions the client wants to continue. The subscriptions are stored **in memory** for now, which aligns with how we allow subscriptions to resume -- only if still within the cached WAL window. WAL cache is in memory, so storing subscriptions with a "longer" lifetime makes little sense for now. If the client ever tries to connect outside the window, we immediately consider them fallen behind and remove saved subscriptions.

### DIFF
--- a/components/electric/lib/electric/satellite/subscription_manager.ex
+++ b/components/electric/lib/electric/satellite/subscription_manager.ex
@@ -9,6 +9,8 @@ defmodule Electric.Satellite.SubscriptionManager do
   use GenServer
   require Logger
 
+  alias Electric.Replication.Shapes.ShapeRequest
+
   @ets_table :saved_subscription_ids
 
   def start_link(_) do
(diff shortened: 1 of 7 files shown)


e2e/tests/_shared.luxinc
```diff
@@ -53,6 +53,29 @@
     ?rec \[\d\]: %Electric.Satellite.V\d+.SatInStartReplicationResp\{
 [endmacro]
 
+[macro client_session_continue user_id session_id position subscription_ids]
+    [invoke log "user ${user_id}: session ${session_id}: start"]
+    [local client_id=client_${user_id}_${session_id}]
+    !auth_config = Electric.Satellite.Auth.JWT.build_config!( \
```
alco, line 59: This has been renamed to `Auth.Secure` in main.

components/electric/lib/electric/satellite/subscription_manager.ex
```diff
@@ -0,0 +1,81 @@
...
+  @ets_table :saved_subscription_ids
+
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @doc """
+  Saves the given shape requests under given subscription id for this client
+  """
+  @spec save_subscription(String.t(), String.t(), [term()]) :: :ok
```
alco, line 21: Any reason we can't make the 3rd argument's type be `ShapeRequest.t()`?

alco (approved):
:+1:

components/electric/lib/electric/satellite/subscription_manager.ex
icehaunter: We can, I just forgot

PR #237, opened 2023-07-11
feat(Satellite): support concurrent subscription requests

kevin-dp:
To allow concurrent subscription requests this PR modifies the Satellite client such that it keeps a data cache per subscription request. As such, it is possible to request new subscriptions even if some other subscriptions are still fetching the data.

For example, you can now request new subscriptions as soon as the previous one is acknowledged.

```
await db.Post.sync()
await db.User.sync()
await db.Profile.sync() 
```

Whereas previously, you had to await the subscription to be completely fulfilled:

```
const { dataReceived } = await db.Post.sync()
await dataReceived
const { dataReceived: recv2 } = await db.User.sync()
await recv2
const { dataReceived: recv3 } = await db.Profile.sync() 
await recv3
```

Note that we still require subscription requests to be issues one at a time.
i.e. this may lead to concurrency issues:
```
db.Post.sync()
db.User.sync()
```

## For reviewers

To support new subscription requests while others subscriptions are still being fetched, we introduce a new class `SubscriptionCaches` that creates a data cache per subscription request. When receiving messages for a subscription request, it forwards it to the corresponding data cache.

When a subscription is fulfilled or fails it is GCed by the data cache which calls the `gc` function we pass into it when instantiating it:
```typescript
const gc = () => {
    // GC all information about this subscription
    // (is called by the cache when the subscription
    //  is fulfilled or when the subscription fails)
    delete this.caches[subscriptionId]
    if (this.currentDelivery === subscriptionId)
      this.currentDelivery = undefined
  }
const cache = new SubscriptionDataCache(subsRequest, this.emitter, gc)
```

The `gc` function is called by `SubscriptionDataCache.reset()` on fulfillment or on failure.


clients/typescript/src/satellite/shapes/cache.ts
```diff
@@ -191,23 +186,29 @@ export class SubscriptionsDataCache extends EventEmitter {
...
       this.inDelivery.transaction.push(op)
     }
   }
 
-  internalError(code: SatelliteErrorCode, msg: string): never {
-    this.reset()
+  internalError(
+    code: SatelliteErrorCode,
+    msg: string,
+    subId?: SubscriptionId
```
icehaunter, line 209: Since this is optional, but always filled from `this.inDelivery.subscriptionId` if `this.inDelivery` at all exists, then maybe we should remove that third argument from each place-of-call and try setting it here?

Smth like
```typescript
subId ??= this.inDelivery?.subscriptionId
```

which makes all the third argument calls apart from ones in `subscriptionResponse` and `subscriptionDataBegin` redundant and makes the changes nice and small

clients/typescript/test/satellite/client.test.ts
```diff
@@ -795,31 +795,26 @@ test.serial('subscription incorrect protocol sequence', async (t) => {
...
+        client.unsubscribeToSubscriptionEvents(success, error)
+        res()
+      }
+
+      client.subscribeToSubscriptionEvents(success, error)
+      client.subscribe(subscriptionId, [shapeReq, shapeReq])
+    })
+    await promise
+  }
+  t.pass()
```
icehaunter, line 817: So I love your simplification, but I think this can actually be a little more AVA-native:

```suggestion
  t.plan(testCases.length) // Expect exactly this amount of assertions
  for (const next of testCases) {
    server.nextResponses(next)

    const promise = new Promise<void>((res, rej) => {
      const success = () => {
        t.fail('expected the client to fail on an invalid message sequence')
        rej()
      }

      const error = () => {
        client.unsubscribeToSubscriptionEvents(success, error)
        t.pass()
        res()
      }

      client.subscribeToSubscriptionEvents(success, error)
      client.subscribe(subscriptionId, [shapeReq, shapeReq])
    })
    await promise
  }
```

icehaunter (changes requested):
Small nitpicks, but overall lgtm

icehaunter (approved):

PR #239, opened 2023-07-12
feat(Satellite, DAL): rename `dataReceived` and `Sub`

kevin-dp:
Renames `dataReceived` to `synced` and `Sub` to `ShapeSubscription`.
I did not rename `Sub` to `Shape` because there is already a `Shape` type.
So i think it makes more sense that the client passes a `Shape` to sync and that they get back a `ShapeSubscription`.
```typescript
sync(shape: Shape): Promise<ShapeSubscription>
```


thruflo (approved):
Nice, thanks :)

Issue #132, opened 2023-04-25
Possible issues on the Typescript client


js2702:
We've been developing (@davidmartos96) a Dart implementation of the electric client (Available here https://github.com/SkillDevs/electric_dart). We found some issues/ had some concerns and some questions from the client implementation. We will post them all here in different sections, you can later organize them as you wish, or discard them.

# Issues

## Electric tables sqlite types
Some metadata tables that Electric creates (oplogTable, metaTable, etc) use the column type STRING. Which does not exist. 
The affinity for the column defaults as numeric. We had some issues here where we encountered a number instead of a string. 
This works in Typescript but might be a source of problems in the future. 

Here is an example: https://github.com/electric-sql/electric/blob/a52c752236699d8dfb289cbb10f5a15b9bf2c9b8/clients/typescript/src/migrators/schema.ts#L20

## Union return types
Some functions in the satellite client and process have a return type of `Promise<void | Error>` or `Promise<void | SatelliteError>`. This is fine but we found some places where the promise is rejected instead of returning the Error. It may be confusing to have a header which returns an union type and the promise actually fails. 

The implementation method.
https://github.com/electric-sql/electric/blob/a52c752236699d8dfb289cbb10f5a15b9bf2c9b8/clients/typescript/src/satellite/client.ts#L239

The interface method.
https://github.com/electric-sql/electric/blob/a52c752236699d8dfb289cbb10f5a15b9bf2c9b8/clients/typescript/src/satellite/index.ts#L60

## Unused property
https://github.com/electric-sql/electric/blob/a52c752236699d8dfb289cbb10f5a15b9bf2c9b8/clients/typescript/src/satellite/process.ts#L71

## Throttle cleanup
The following lodash throttled never gets cleaned up. The output of throttle has a method cancel.
https://github.com/electric-sql/electric/blob/a52c752236699d8dfb289cbb10f5a15b9bf2c9b8/clients/typescript/src/satellite/process.ts#L113
Maybe it could be instantiated in the start and cleaned up in stop?

## Connectivity change subscription cleanup
This subscription never gets unsubscribed, maybe it could be unsubscribe right before subscribing again?
https://github.com/electric-sql/electric/blob/a52c752236699d8dfb289cbb10f5a15b9bf2c9b8/clients/typescript/src/satellite/process.ts#L150

## Typo?
This probably should be refreshToken.
[...]

PR #240, opened 2023-07-12
Modify test to un-silence wrong expectation

davidmartos96:
This has been mentioned in #132 , but since most of those have already been fixed, I include this PR to isolate it and leave it for discussion.

The `t.is(lsn2, '2')` is never executed in the original code, and the test passes, but the assert actually fails.

### DIFF
--- a/clients/typescript/test/satellite/process.test.ts
+++ b/clients/typescript/test/satellite/process.test.ts
@@ -1179,7 +1179,7 @@ test('handling connectivity state change stops queueing operations', async (t) =
   const lsn1 = await satellite._getMeta('lastSentRowId')
   t.is(lsn1, '1')
 
-  satellite._connectivityStateChange('connected')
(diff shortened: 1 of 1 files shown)


icehaunter:
@balegas, do you have some time to dig in if this is a incorrectly written test or an actual bug?

clients/typescript/test/satellite/process.test.ts
```diff
@@ -1181,10 +1181,9 @@ test('handling connectivity state change stops queueing operations', async (t) =
 
   satellite._connectivityStateChange('connected')
```
balegas, line 1182: This should fix the test: what triggers the reconnect is the connections becoming available

```suggestion
  satellite._connectivityStateChange('available')
```

davidmartos96: Done!

icehaunter (approved):

PR #241, opened 2023-07-12
Expand serialization tests with float and integer values

davidmartos96:
I added some additional serialization test cases, as only Strings and null values were being tested.

Related to this, does Electric support Blob columns? I haven't found any example in the tests, so I'm not sure if it's supported or not.


icehaunter:
@davidmartos96, as to BLOBs -- I'm not sure. It should (i.e. they are not treated differently) as long as Satellite and Postgres agree on string serialization of the blob (i.e. so that when the blob gets on replication stream to PG it will get deserialized properly), and I cannot say I verified if they do agree

PR #245, opened 2023-07-12
auth: drop the SATELLITE_ prefix from env vars.

thruflo:
Given that auth mode env vars are mandatory one way or the other, I'd rather we presented:

```shell
docker run \
    -e "AUTH_MODE=insecure" \
    -e "DATABASE_URL=postgresql://..." \
    -e "ELECTRIC_HOST=..." \
    -p 5050:5050 \
    -p 5133:5133 \
    electricsql/electric
```

Than:

```shell
docker run \
    -e "DATABASE_URL=postgresql://..." \
    -e "ELECTRIC_HOST=..." \
    -e "SATELLITE_AUTH_MODE=insecure" \
    -p 5050:5050 \
    -p 5133:5133 \
    electricsql/electric
```

Because it's shorter, the auth mode is the mode for the whole electric service and we're not proposing to auth anything else.


alco (approved):
Beautiful!

I've pushed a fix to table formatting in the README.md file to make it readable when not rendered.

PR #246, opened 2023-07-12
chore: updated Elixir to 1.15 and OTP to 25.3

icehaunter:


alco (approved):

components/electric/config/dev.exs
```diff
@@ -8,12 +8,12 @@ config :electric, Electric.Replication.Connectors,
   postgres_1: [
     producer: Electric.Replication.Postgres.LogicalReplicationProducer,
     connection: [
-      host: 'localhost',
+      host: ~c"localhost",
```
alco, line 11: Weird rule. We could set the `normalize_charlists_as_sigils` option instead and retain the `'charlist'` syntax.

icehaunter: Eh, I like the explicitness, especially for non-elixir devs. Using different quotes leading to a substantially differing behaviour is indeed confusing

PR #252, opened 2023-07-13
fix: replaced JWT library with web-compatible one

icehaunter:


kevin-dp:
Could you replace the hardcoded tokens in the examples and call this `authToken` instead please? :-)

PR #255, opened 2023-07-13
chore(electric): Bug fixes and a new test for initial sync

alco:
Supersedes https://github.com/electric-sql/electric/pull/231.


icehaunter (changes requested):
Looks good, and nice that you found actual bugs! I'm still hesitant about the pause (although much less), and I have one more nitpick about the e2e

PR #256, opened 2023-07-17
chore(protocol): Fix typos and make corrections in the protocol definition

alco:


clients/typescript/src/_generated/protocol/satellite.ts
```diff
@@ -109,8 +109,8 @@ export enum SatErrorResp_ErrorCode {
   AUTH_FAILED = 2,
   REPLICATION_FAILED = 3,
   INVALID_REQUEST = 4,
-  PROTO_VSN_MISSMATCH = 5,
-  SCHEMA_VSN_MISSMATCH = 6,
+  PROTO_VSN_MISMATCH = 5,
```
icehaunter, line 112: Since this isn't backwards-compatible fix, how about a protobuf version bump?

icehaunter (changes requested):

clients/typescript/src/_generated/protocol/satellite.ts
alco: I think I was wrong when I called this a breaking change. The numbers for these options are unchanged, so even if an old client connects to a new server, both will be sending `5` on the wire even though each will decode it into different symbolic constants.

icehaunter: Yeah, you're correct actually

icehaunter (approved):

PR #268, opened 2023-07-19
Cleanup after `useEffect`s.

thruflo:


thruflo:
Hey @icehaunter so this implements some cleanup and unsubscribing. I haven't gone into the satellite client or shapes. A starting point for those would be to look at the constructor and start method for the satellite process.

It also occurs to me that the `ensureStarted` approach to starting the satellite process could keep use a bookkeeping approach to stop the process when the last client using it is cleanedup.

I.e.:

- electrified client 1 calls ensureStarted('dbName') -> starts satellite process for 'dbName'
- electrified client 2 calls ensureStarted('dbName') -> noop because started

If the satellite process kept a "used by" list of [client1, client2], then if the clients somehow call `registry.stopUsing('dbName')` on cleanup, the registry could keep a list of the clients and stop the process when no clients are using.

This would need to be threaded through by looking at the return value of electrify, i.e.: an App calls `const db = electrify(conn, schema, config)` in a useEffect -- that return value would need to change (e.g.: by adding a cleanup function to `db` which is currently an ElectricNamespace) so the useEffect could return a cleanup function which calls, along the lines of:

```tax
useEffect(() => {
  let unmount
  
  const init = async () => {
    const { db, cleanup } = async electrify(conn, schema, config)
     
    unmount = cleanup 
  }

  return unmount
}, []
```

PR #269, opened 2023-07-19
fix(electric): Put the current LSN on migration transactions delivered during initial sync

alco:
Fixes VAX-771.


components/electric/lib/electric/satellite/ws_server.ex
```diff
@@ -200,13 +200,15 @@ defmodule Electric.Satellite.WsServer do
...
-    migration_transactions = InitialSync.migrations_since(schema_version, state.pg_connector_opts)
-    {msgs, state} = Protocol.handle_outgoing_txs(migration_transactions, state)
 
-    max_txid =
-      migration_transactions
-      |> Enum.map(fn {tx, _offset} -> tx.xid end)
-      |> Enum.max(fn -> 0 end)
+    {migration_transactions, max_txid} =
+      schema_version
+      |> InitialSync.migrations_since(state.pg_connector_opts)
```
icehaunter, line 206: I did the same fix locally, but then though: could we just make `migrations_since/2` a little more protocol-agnostic and just return transactions without the offset pair? I think it would be cleaner even if requires adjusting tests due to changing function return value

icehaunter (changes requested):

icehaunter:
Also a changeset please

components/electric/lib/electric/satellite/ws_server.ex
alco: Refactored it a bit. Had to force-push after rebasing on top of https://github.com/electric-sql/electric/pull/267, so I rewrote the original commit with the updated implementation.

icehaunter (approved):

components/electric/lib/electric/satellite/ws_server.ex
icehaunter: Thanks! Approved

PR #273, opened 2023-07-19
fix(electric): use magic write to ensure subscription data is sent 

icehaunter:


components/electric/lib/electric/postgres/extension/migrations/20230715000000_utilities_table.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Electric.Postgres.Extension.Migrations.Migration_20230715000000_UtilitiesTable do
+  alias Electric.Postgres.Extension
+
+  @behaviour Extension.Migration
+
+  @impl true
+  def version, do: 20_230_715_000_000
```
alco, line 7:
```suggestion
  def version, do: 2023_07_15_00_00_00
```

components/electric/lib/electric/replication/initial_sync.ex
```diff
@@ -67,11 +67,20 @@ defmodule Electric.Replication.InitialSync do
...
      results when they are ready. That message **has** to be sent ASAP since if we send the results
      at the end, we might have already skipped the point where the data is relevant.
-  2. `{:subscription_data, subscription_id, observed_lsn, data}` is when we've collected all the data.
-     The `observed_lsn` here is a cheat-code to sidestep an issue where there were no transactions.
-     It's required, because if we haven't received any transactions while querying, we don't have
-     any transactions to trigger sending this data in `Electric.Satellite.WsServer`, so we're using
-     this LSN point only for equality checking if we haven't moved at all.
+  2. `{:subscription_data, subscription_id, data}` is when we've collected all the data.
+
+  One more this this function is expected to do is to perform any write to Postgres into a table that is a part
```
alco, line 72:
```suggestion
  One more thing this function is expected to do is to record any write to Postgres in a table that is a part
```

components/electric/lib/electric/replication/initial_sync.ex
```diff
@@ -67,11 +67,20 @@ defmodule Electric.Replication.InitialSync do
...
+  of our publication (i.e. the Electric will see it) and the table should probably part of the `electric` schema.
```
alco, line 73:
```suggestion
  of our publication (i.e. the Electric will see it) and the table should probably be part of the `electric` schema.
```

components/electric/lib/electric/replication/initial_sync.ex
```diff
@@ -113,11 +123,22 @@ defmodule Electric.Replication.InitialSync do
...
   end
+
+  defp perform_magic_write(opts) do
+    Connectors.get_connection_opts(opts, replication: false)
+    |> Client.with_conn(fn conn ->
+      {:ok, 1} =
+        :epgsql.squery(
+          conn,
+          "UPDATE #{Extension.utilities_table()} SET content = '{}' WHERE id = 'magic write'"
+        )
```
alco, line 141: This should proably be a public function in the `Extension` module. That's where we've been keeping all internal statements up to this point.

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -31,6 +31,7 @@ defmodule Electric.Postgres.Extension do
   @schema_table electric.("schema")
   @electrified_tracking_table electric.(@electrified_table_relation)
   @electrified_index_table electric.(@electrified_index_relation)
+  @utilities_table electric.("utilities")
```
magnetised, line 34: `utilities` feels like the wrong name. could we use something like `transaction_marker`  or `transaction_flush`

magnetised (changes requested):
can we change the table name. not saying my name ideas are particularly stellar, but it feels early to go for "utils"... 

other than that glad the idea worked and required so little code... did the modified tests fail before the addition of this?

components/electric/lib/electric/postgres/extension.ex
icehaunter: I kinda thought it's about time for `utilities` table, maybe we could store more info there?

icehaunter: Although sure, it's not like we're short on table names

components/electric/lib/electric/replication/initial_sync.ex
icehaunter: Done

components/electric/lib/electric/postgres/extension.ex
icehaunter: Done

PR #275, opened 2023-07-20
fix: correctly disable triggers on shape GC

icehaunter:
Shapes don't use fully qualified table names, but the functions disabling the triggers expect FQ names. So we didn't disable the triggers before. Ideally FQ names should be represented by a different type (e.g. a tuple), but that's for another day

### DIFF
--- a/.changeset/odd-timers-tickle.md
+++ b/.changeset/odd-timers-tickle.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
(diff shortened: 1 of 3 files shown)


alco (approved):

thruflo:
> Ideally FQ names should be represented by a different type (e.g. a tuple)

Just to note that there is already a `QualifiedTablename` type in `utils/types`.

icehaunter:
@thruflo 
> Just to note that there is already a QualifiedTablename type in utils/types.
yeah, just noticed myself, but it is not applied consistently and I don't want to dig right now through all call chains to see if what works now is qualified-but-string or unqualified. I'll get back to it

PR #286, opened 2023-07-24
fix: make use of `subscriptionId` to differentiate RPC responses

icehaunter:
This is more of a bandaid than a proper fix - proper fix involves adding full RPC logic to the protocol, but this isn't the time


clients/typescript/src/satellite/client.ts
```diff
@@ -1011,9 +1014,22 @@ export class SatelliteClient extends EventEmitter implements Client {
...
-      })
+      const handleRpcResp = (resp: T) => {
+        if (distinguishOn) {
+          // TODO: remove this comment when RPC types are fixed
+          // @ts-ignore this comparison is valid because we expect the same field to be present on both request and response, but it's too much work at the moment to rewrite typings for it
+          if (resp[distinguishOn] === request[distinguishOn]) {
+            return resolve(resp)
+          } else {
+            // This WAS an RPC response, but not the one we were expecting, waiting more
+            this.once('rpc_response', handleRpcResp)
```
alco, line 1025: I'm not sure how `this.once()` works. It sets a unique handler for the given event, correct?

Here, we're discarding the response we've just received. But some other place in the code must be expecting it, no?

icehaunter: no, `.once` will just run this function once, on the next time this event is emitted. It does not affect any other listeners

icehaunter: See https://nodejs.org/api/events.html#emitteronceeventname-listener

clients/typescript/src/satellite/client.ts
```diff
@@ -1011,9 +1014,22 @@ export class SatelliteClient extends EventEmitter implements Client {
         return reject(error)
       })
 
-      this.once('rpc_response', (resp: T) => {
-        return resolve(resp)
-      })
+      const handleRpcResp = (resp: T) => {
+        if (distinguishOn) {
```
alco, line 1018: Since we know the value of `distinguishOn` outside of the closure, we can move the condition outside of the closure too.

```typescript
if (distinguishOn) {
  const handleRpcResp = (resp: T) => {
    if (resp[distinguishOn] === request[distinguishOn]) {
      return resolve(resp)
    } else {
      this.once('rpc_response', handleRpcResp)
    }
  }
  this.once('rpc_response', handleRpcResp)
} else {
  this.once('rpc_response', (resp: T) => resolve(resp))
}
```

Makes it easier to wrap your head around the case when `distinguishOn` is undefined. Also makes it easier to spot the "async recursion" in the first branch of `if`.

alco (approved):

PR #278, opened 2023-07-20
fix: make sure to deduplicate subscribe calls on the client

icehaunter:

PR #287, opened 2023-07-24
fix: make sure the subscription request is saved early

icehaunter:
Continuation of #278, adding the test and covering additional behaviour

### DIFF
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -344,6 +344,8 @@ export class SatelliteProcess implements Satellite {
     }))
 
     const subId = this.subscriptionIdGenerator()
+    this.subscriptions.subscriptionRequested(subId, shapeReqs)
+
     // store the resolve and reject
     // such that we can resolve/reject
     // the promise later when the shape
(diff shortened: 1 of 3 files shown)


clients/typescript/src/satellite/process.ts
```diff
@@ -357,18 +359,19 @@ export class SatelliteProcess implements Satellite {
     const { subscriptionId, error }: SubscribeResponse =
       await this.client.subscribe(subId, shapeReqs)
     if (subId !== subscriptionId) {
+      delete this.subscriptionNotifiers[subId]
```
balegas, line 362: seeing `this.subscriptionNotifiers` for the first time, why don't we follow pattern we follow in other places and handle notifications from the manager directly? This would avoid handling subscriptions-related events in two places

clients/typescript/src/satellite/process.ts
```diff
@@ -344,6 +344,8 @@ export class SatelliteProcess implements Satellite {
     }))
 
     const subId = this.subscriptionIdGenerator()
+    this.subscriptions.subscriptionRequested(subId, shapeReqs)
```
balegas, line 347: the client should be smart enough not preemptively avoid making duplicate requests.

I understand we first add intelligence on the server, but let's add an improvement ticket for this, so we don't lose track

balegas (approved):
Looks good. Minor comment about changes that were committed before.

clients/typescript/src/satellite/process.ts
balegas: I notice this is done already

PR #288, opened 2023-07-24
Add precise typings for _getMeta

### LINKED ISSUE #259: [client] _getMeta wrong typings
After some recent changes, there is a call on _setMetaStatement with the value null.

https://github.com/electric-sql/electric/blob/8f68c1b49c78909ef1355dadbf16c9a320bb2e23/clients/typescript/src/satellite/process.ts#L443C40-L443C40 

But at the same time getMeta returns string and not string|null. Wouldn't this need to be changed to string|null? And maybe _setMetaStatement could receive string|null instead of SqlValue.

klavinski:
This small change now guarantees the result type from `_getMeta` depending on the entry key:

image

To ensure this result, `_setMeta` also checks its inputs:

image

Here, we read `Argument of type '"nonsense"' is not assignable to parameter of type '"" | `${string}-${string}-${string}-${string}-${string}`'.`, _i.e._, it is not a valid `Uuid`.

While editing the `Uuid` type, I also noticed that the module was declared empty in `declarations.d.ts`. I put the official typings instead.

Closes #259.

### DIFF
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -1325,7 +1325,7 @@ export class SatelliteProcess implements Satellite {
   private async _getClientId(): Promise<Uuid> {
     const clientIdKey = 'clientId'
 
-    let clientId = (await this._getMeta(clientIdKey))
+    let clientId = await this._getMeta(clientIdKey)
 
     if (clientId === '') {
       clientId = uuid() as Uuid
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -1,5 +1,16 @@
 import throttle from 'lodash.throttle'
 
+type Uuid = `${string}-${string}-${string}-${string}-${string}`
+
+type MetaEntries = {
+  clientId: Uuid | '',
+  compensations: number,
+  lastAckdRowId: `${ number }`,
+  lastSentRowId: `${ number }`,
+  lsn: string | null,
+  subscriptions: string,
+}
+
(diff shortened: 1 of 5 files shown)


clients/typescript/src/satellite/process.ts
```diff
@@ -1293,16 +1310,16 @@ export class SatelliteProcess implements Satellite {
...
-    return rows[0].value as string
+    return rows[0].value
   }
 
-  private async _getClientId(): Promise<string> {
+  private async _getClientId(): Promise<Uuid> {
     const clientIdKey = 'clientId'
 
-    let clientId: string = await this._getMeta(clientIdKey)
+    let clientId = await this._getMeta(clientIdKey) as Uuid | ''
```
icehaunter, line 1319: I feel like this type cast does the work you did a disservice. Let's either use a `clientId` string directly or add `as const` after it's assignment to the variable so that this type cast is no longer needed. I think using a "magic" string without a variable declaration is absolutely fine because your type checks will guard against typos anyway.

icehaunter (changes requested):
Thank you very much for the PR! That's great work, and great types. Love the precision we're getting from this.

One tiny change to make it flawless :)

clients/typescript/src/satellite/process.ts
klavinski: Good catch, thanks! I replaced `:string` with `as Uuid | ''` in the beginning, and forgot to remove it. Types still work without this assertion.

icehaunter (approved):

PR #293, opened 2023-07-26
fix: properly send all pending txs instead of discarding

icehaunter:
`.splice(0)` without a second argument removes all elements from the array, returning them. We essentially cleared the array and did nothing with removed elements.

The changes in the test are from the fact that the outbound transactions starts filling up only after the third transaction: first write calls the debounced function immediately, second write falls into the debounce window, and third write gets accumulated with the second and then doesn't get sent.

### DIFF
--- a/.changeset/fast-forks-reply.md
+++ b/.changeset/fast-forks-reply.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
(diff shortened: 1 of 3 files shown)


alco (approved):
:+1:

thruflo (approved):

balegas:
Me writing JavaScript for the first time. Thanks and sorry for that.

PR #294, opened 2023-07-26
fix: follow the `include`s when returning affected tables in `live*`

icehaunter:


thruflo:
This is awesome, thanks :) and this will fix my hack posted in #user-feedback.

One other requirement noted in the Linear issue description that's not addressed by the include tree is also detecting when a where clause uses a related table. For example, you could query on a relation without including it. I'd suggest we address that in future when implementing proper relation query support.

thruflo (approved):

PR #298, opened 2023-07-26
client: fix buggy connectivity hook.

thruflo:
The problem was that existing satellite processes were not reporting their existing connectivity status to the connectivity hook. This was confusing the hook into always starting in a disconnected state, despite the connection being connected.

I've left one note in the code about a future refactor. I'm not convinced by the current code structure and connectivity state machine. But suggesting we merge this as we need the fix urgently.

### DIFF
--- a/clients/typescript/src/client/model/client.ts
+++ b/clients/typescript/src/client/model/client.ts
@@ -1,6 +1,4 @@
 import { ElectricNamespace } from '../../electric/namespace'
-import { DatabaseAdapter } from '../../electric/adapter'
-import { Notifier } from '../../notifiers'
 import { DbSchema, TableSchema } from './schema'
 import { liveRaw, raw, Table } from './table'
 import { Row, Statement } from '../../util'
@@ -59,10 +57,14 @@ export class ElectricClient<
 > extends ElectricNamespace {
   private constructor(
     public db: ClientTables<DB> & RawQueries,
-    adapter: DatabaseAdapter,
-    notifier: Notifier
+    electric: ElectricNamespace
   ) {
-    super(adapter, notifier)
+    super(electric.adapter, electric.notifier)
+
+    Object.defineProperty(this, 'isConnected', {
+      get: () => electric.isConnected,
+      set: (_value) => { /* noop */ }
+    })
   }
 
   // Builds the DAL namespace from a `dbDescription` object
@@ -98,6 +100,6 @@ export class ElectricClient<
       liveRaw: liveRaw.bind(null, electric.adapter),
     }
 
-    return new ElectricClient(db, electric.adapter, electric.notifier)
+    return new ElectricClient(db, electric)
   }
 }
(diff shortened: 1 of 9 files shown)


thruflo:
p.s.: someone please tell me the right command to run prettier properly!

alco:
@thruflo Run `pnpm format` inside the `clients/typescript` directory.

clients/typescript/src/client/model/client.ts
```diff
@@ -59,10 +57,16 @@ export class ElectricClient<
...
   ) {
-    super(adapter, notifier)
+    super(electric.adapter, electric.notifier)
+
+    Object.defineProperty(this, 'isConnected', {
+      get: () => electric.isConnected,
+      set: (_value) => {
+        /* noop */
+      },
+    })
```
icehaunter, line 69: This feels very weird and indirect to me. 

So before this change, the `const electric = new ElectricNamespace()` variable in `src/satellite/index.ts` was used in a mildly weird way: it was used for the constructor, and then only `.adapter` and `.notifier` properties got used from that instance. This `ElectricClient` instance extends `ElectricNamespace` - so it will be initialized again with the same notifier, but now you're building a closure to capture a previous namespace and use `isConnected` value from there instead of using `isConnected` value derived from the `extends ElectricNamespace`. I honestly see no reason at all for this override: if it's just so you can run `electric.setIsConnected(satellite.connectivityState)` in `src/satellite/index.ts` then you can remove this property and call `namespace.setIsConnected` a couple lines down in the same file.

I don't know why we need that kind of cognitive load - figuring out how/why this is working took a minute. 

I think the correct change would be to:
1. Replace the `electric: ElectricNamespace` parameter in `create` with explcit `adapter` and `notifier`. Same for this constructor.
2. Don't create a "transient" `ElectricNamespace` instance in `src/satellite/index.ts` at all - it doesn't get used and the same constructor code runs perfectly well here in the `super` call.
3. Move `electric.setIsConnected(satellite.connectivityState)` to instead reference the created `ElectricClient` instance directly.

That way it's cleaner, we're setting up one less listener, have only once source of truth instance, and don't need an ugly `definePropery` call in constructor

icehaunter (changes requested):

clients/typescript/src/client/model/client.ts
thruflo: Totally. Be very good to simplify this.

As a left field thought: perhaps the electric client should just have a reference to the satellite process?

clients/typescript/src/frameworks/react/hooks/useConnectivityState.ts
```diff
@@ -1,61 +1,98 @@
...
 import { ElectricContext } from '../provider'
 
+type RetVal = {
+  connectivityState: ConnectivityState
+  setConnectivityState: (state: ConnectivityState) => void
+  toggleConnectivityState: () => void
+}
+type HookFn = () => RetVal
+
+const STATES: {
```
kevin-dp, line 15: Why not using a TypeScript enumeration for this?

kevin-dp (approved):

icehaunter (approved):

PR #300, opened 2023-07-30
chore: rewritten `_performSnapshot` function

icehaunter:
Previous implementation executed 3N+2 queries for N oplog entries (`UPDATE` timestamps, `SELECT` updated, then, per oplog entry, `SELECT` shadow row, `UPDATE` oplog entry, `UPDATE` or `DELETE` shadow row).

This implementation always executes exactly 4 queries to achieve the same. This lowers the amount of times we're traversing C or WASM boundary and offloads more work to SQLite.

This shouldn't affect the performance of current "lightweight" apps, but is likely to show significant improvements over 100+ oplog entries

### DIFF
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -1,7 +1,7 @@
 import throttle from 'lodash.throttle'
 
 import { AuthConfig, AuthState } from '../auth/index'
-import { DatabaseAdapter, RunResult } from '../electric/adapter'
+import { DatabaseAdapter } from '../electric/adapter'
 import { Migrator } from '../migrators/index'
 import {
   AuthStateNotification,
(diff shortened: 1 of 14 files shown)


clients/typescript/src/satellite/process.ts
```diff
@@ -695,17 +691,113 @@ export class SatelliteProcess implements Satellite {
...
+        GROUP BY namespace, tablename, primaryKey
+        HAVING rowid = max(rowid) AND optype = 'DELETE'
+      )
+    `,
+      args: [timestamp.toISOString(), this._lastAckdRowId],
+    }
+
+    // Execute the four queries above in a transaction, returning the results from the first query
+    // We're dropping down to this transaction interface because `runInTransaction` doesn't allow queries
+    const oplogEntries = (await this.adapter.transaction<OplogEntry[]>(
```
balegas, line 783: Raising @kevin-dp attention to this

balegas (approved):
Thanks fro addressing the issues raised during the call.

(add the changelog plz :D)

clients/typescript/src/satellite/process.ts
icehaunter: I think it would be hard to have a generic interface for queries in `runInTransaction`, and the callbacks are required for some drivers to keep the transaction scope

PR #303, opened 2023-08-01
fix: move a dependency from dev to prod & bump wa-sqlite

icehaunter:


thruflo:
Have you verified that the intro examples still work with wa-sqlite master? That was the thing I had to previously revert to get the sync to work.

kevin-dp (approved):
Right, the generate script indeed requires better-sqlite3, good catch!
I didn't notice that they actually merged their breaking-changes branch into master.
Again, great catch!

icehaunter:
@thruflo I haven't tested the examples, but I have tested my own apps that they still indeed work
@kevin-dp you're most welcome to test, but current intro examples require some wrangling to start

thruflo:
Let’s verify that it’s not a regression. Master worked for me initially until it didn’t. I can check later if that’s not too late.

thruflo:
Equally, it may be useful to reproduce the prev error I was getting in case it helps flush out somewhere we’re swallowing an error? (Or did we already flush that out?)

icehaunter:
@thruflo don't worry, I'll verify

PR #304, opened 2023-08-01
fix(generator): casing model names also in references

kevin-dp:
This PR ensures that references to other models are also cased properly.


clients/typescript/src/cli/migrations/migrate.ts
```diff
@@ -318,7 +329,37 @@ async function pascalCaseTableNames(prismaSchema: string): Promise<void> {
...
+
+    if (insideModel && ln.trim().startsWith('}')) {
+      insideModel = false
+      return ln
     }
+
+    if (insideModel) {
+      // the regex below matches two identifiers separated by a space
+      // (first identifier is the column name, second one is its type)
+      const reg = /(\s*\w+\s+)(\w+)/
```
icehaunter, line 355:
```suggestion
      const reg = /^(\s*\w+\s+)(\w+)/
```

This should probably be applied only at the start of any given string

clients/typescript/src/cli/migrations/migrate.ts
```diff
@@ -318,7 +329,37 @@ async function pascalCaseTableNames(prismaSchema: string): Promise<void> {
...
+  })
+
+  // Go over the schema again but now
+  // replace references to the old table names
+  // by the new model name when we are inside
+  // the definition of a model
+  let insideModel = false
+  lines = lines.map((ln) => {
+    if (ln.trim().startsWith('model ')) {
+      insideModel = true
```
icehaunter, line 343: Suggestion: If you did a `.flatMap`, you could insert `@@map` annotations right about here

clients/typescript/test/cli/migrations/migrate.test.ts
```diff
@@ -0,0 +1,107 @@
...
+model UserProfile {
+  @@map("user_profile")
+  id     Int    @id
+  bio    String
+  userId Int    @unique
+  user   User?  @relation(fields: [userId], references: [id])
+}
+`
+
+test('migrator correctly PascalCases model names', (t) => {
```
icehaunter, line 102: Thanks for the test!!!

icehaunter (changes requested):
Looks good, and thanks for the test! A couple small comments

icehaunter (approved):

PR #306, opened 2023-08-02
client: remove legacy absurd-sql references.

thruflo:
Noticed some stray absurd-sql / worker stuff still in the code.


clients/typescript/package.json
```diff
@@ -138,6 +138,7 @@
   "dependencies": {
     "async-mutex": "^0.4.0",
     "base-64": "^1.0.0",
+    "better-sqlite3": "^8.4.0"
```
kevin-dp, line 141: misses a comma at the end which makes everything fail

icehaunter (approved):

PR #307, opened 2023-08-02
fix: respect `electric_user_id` filtering on initial sync

icehaunter:
This is a somewhat temporary fix to allow for partial replication based on user id. This assumes user id is kinda immutable per device for the on-device database lifetime (as does a lot of our code), so no effort has been made to ensure row deduplication. Should tide us over until proper DDLX


alco (approved):
:+1:

PR #311, opened 2023-08-02
fix: don't send rows to clients that have already seen this particular table

icehaunter:
This is a very simplified enabling implementation so that we remove a foot-gun of intersecting shapes while we only support unfiltered full-table sync (with one immutable filter of user id).

### DIFF
--- a/components/electric/lib/electric/replication/shapes/shape_request.ex
+++ b/components/electric/lib/electric/replication/shapes/shape_request.ex
@@ -97,37 +97,41 @@ defmodule Electric.Replication.Shapes.ShapeRequest do
          origin,
          filtering_context
        ) do
-    table = Enum.find(schema.tables, &(&1.name.schema == schema_name && &1.name.name == name))
-    columns = Enum.map_join(table.columns, ", ", &~s|main."#{&1.name}"|)
-    {:ok, pks} = Schema.primary_keys(table)
-    pk_clause = Enum.map_join(pks, " AND ", &~s|main."#{&1}" = shadow."#{&1}"|)
+    if filtering_context[:sent_tables] && MapSet.member?(filtering_context[:sent_tables], rel) do
+      {:ok, []}
+    else
+      table = Enum.find(schema.tables, &(&1.name.schema == schema_name && &1.name.name == name))
+      columns = Enum.map_join(table.columns, ", ", &~s|main."#{&1.name}"|)
+      {:ok, pks} = Schema.primary_keys(table)
+      pk_clause = Enum.map_join(pks, " AND ", &~s|main."#{&1}" = shadow."#{&1}"|)
 
-    ownership_column = Ownership.id_column_name()
+      ownership_column = Ownership.id_column_name()
 
-    where_clause =
-      if filtering_context[:user_id] && Enum.any?(table.columns, &(&1.name == ownership_column)) do
-        escaped = String.replace(filtering_context[:user_id], "'", "''")
+      where_clause =
+        if filtering_context[:user_id] && Enum.any?(table.columns, &(&1.name == ownership_column)) do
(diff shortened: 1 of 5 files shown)


components/electric/lib/electric/replication/shapes/shape_request.ex
```diff
@@ -97,37 +97,41 @@ defmodule Electric.Replication.Shapes.ShapeRequest do
...
+    else
+      table = Enum.find(schema.tables, &(&1.name.schema == schema_name && &1.name.name == name))
+      columns = Enum.map_join(table.columns, ", ", &~s|main."#{&1.name}"|)
+      {:ok, pks} = Schema.primary_keys(table)
+      pk_clause = Enum.map_join(pks, " AND ", &~s|main."#{&1}" = shadow."#{&1}"|)
+
+      ownership_column = Ownership.id_column_name()
+
+      where_clause =
+        if filtering_context[:user_id] && Enum.any?(table.columns, &(&1.name == ownership_column)) do
```
alco, line 111: Just a nitpick: I would personally assign `filtering_context[:sent_tables]` and `filtering_context[:user_id]` to local variables before the `if`, both to make the lines shorter and avoid the pitfalls of code repetition (potential for typos, slight performance hit).

components/electric/lib/electric/replication/shapes/shape_request.ex
```diff
@@ -97,37 +97,41 @@ defmodule Electric.Replication.Shapes.ShapeRequest do
...
-    where_clause =
-      if filtering_context[:user_id] && Enum.any?(table.columns, &(&1.name == ownership_column)) do
-        escaped = String.replace(filtering_context[:user_id], "'", "''")
-
-        # We're using explicit interpolation here instead of extended query, because we need all columns regardless of type to be returned as binaries
-        "WHERE #{ownership_column} = '#{escaped}'"
-      else
-        ""
+    if filtering_context[:sent_tables] && MapSet.member?(filtering_context[:sent_tables], rel) do
+      {:ok, []}
```
alco, line 101: This branch effectively makes the function a noop. I we can make the code simpler by moving this condition check outside of the `query_full_table()` function and leave the latter unchanged.

Here's what I mean:

```elixir
  def query_initial_data(%__MODULE__{} = request, conn, schema, origin, filtering_context \\ %{}) do
    table_already_sent_fn =
      if sent_tables = filtering_context[:sent_tables] do
        fn table -> MapSet.member?(sent_tables, table) end
      else
        fn _ -> false end
      end

    Enum.reduce_while(request.included_tables, {:ok, []}, fn table, {:ok, acc} ->
      if table_already_sent_fn.(table) do
        {:cont, {:ok, acc}}
      else
        case query_full_table(conn, table, schema, origin, filtering_context) do
          {:ok, results} ->
            {:cont, {:ok, acc ++ results}}
[...]
```

I know we're going to have more filtering logic in the future, so this may be a moot point. I just wanted to point out that wrapping the whole function body into an `if` is a smell; usually a better approach is to move the `if` outside of the function and make the latter have one less level of indentation.

alco (commented):
Looks nice.

No tests?

components/electric/lib/electric/replication/shapes/shape_request.ex
icehaunter: I agree about trying to avoid indentation, but I explicitly wanted to "show" and remind myself that on implementation each shape needs to be able to use context information. So logically, it falls better inside the inner function then the outside. Hence my decision to put it there

alco (approved):

components/electric/lib/electric/replication/shapes/shape_request.ex
icehaunter: Added tests

PR #312, opened 2023-08-03
feat(DAL): made argument of liveMany optional

kevin-dp:
Small changes to make the argument of `liveMany` optional such that we can write `db.items.liveMany()` instead of `db.items.liveMany({})`.

Also introduces some additional unit tests for the query builder to test functionality of the DAL.


thruflo (approved):
Awesome!

PR #313, opened 2023-08-03
fixt(client): Correctly subtract 1 second instead of 1000 in the JWT "iat" claim

davidmartos96:
Context: https://discord.com/channels/933657521581858818/1079688869852753981/1136612631218765904


alco (approved):

davidmartos96:
@alco I've included an additional minor change. 
It's a change we've also needed to apply to the Dart client tests, as they could be flaky.
If we first delete the tests database, it's possible that Satellite tries to use it in an async callback when it has been deleted already.

davidmartos96:
It could be one reason why this failed https://github.com/electric-sql/electric/actions/runs/5750170432/job/15586397973

PR #314, opened 2023-08-03
feat(examples): starter template

kevin-dp:
This PR introduces a package containing a NodeJS script to bootstrap Electric apps.


starter/template/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Backend
+
+Build and start the backend:
+
+```sh
+pnpm backend:start
+```
+
+The above command builds the latest Electric image and starts a fresh Postgres DB and the Electric sync service.
```
thruflo: The primary way to use this starter is to install via `npx create-electric-app`. Thus the source code is not available. And we don't want to force a developer to build electric from scratch to run an app.

So we need to use the published `electricsql/electric` image. Not build from source.

starter/template/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Now, migrate Postgres such that it contains the necessary tables for your app to work:
+```sh
+pnpm backend:migrate
```
thruflo: Could we call this `db:migrate`? It's just a clearer noun to understand what's happening.

starter/template/README.md
```diff
@@ -0,0 +1,47 @@
...
+```sh
+pnpm install
+```
+
+## Backend
+
+Build and start the backend:
+
+```sh
+pnpm backend:start
```
thruflo: I think the use-your-own PG could be as simple as a `--database-url` option or similar. Presumably the complexity comes from the LOGICAL_PUBLISHER_HOST, i.e.: how the database connects to the Electric. If necessary, we could mandate that `--database-url` and `--logical-publisher-host` are both provided if one is. Or we default to something we hope will work like host.docker.internal or whatever it is.

starter/template/README.md
```diff
@@ -0,0 +1,47 @@
...
+pnpm backend:migrate
+```
+
+## Frontend
+
+In order for your app to interact with the backend, you need to generate an Electric client:
+```sh
+npx electric-sql generate
+```
+This command generates an Electric client that is up-to-date with the data model on Postgres.
```
thruflo, line 32: I'd drop this line -- less conversation, more action.

starter/template/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Now, you can build and run the app:
+```sh
+pnpm build && pnpm start
```
thruflo: Why build and start, why not just start?

starter/template/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "electric-sql-web-example",
+  "version": "0.1.0",
+  "main": "builder.js",
+  "author": "Valter Balegas",
+  "license": "Apache-2.0",
```
thruflo, line 6: They can choose their own license. Suggest remove.

starter/template/src/Example.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+import { dbSchema, Electric } from './generated/client'
+import { electrify, ElectricDatabase } from 'electric-sql/wa-sqlite'
+import { makeElectricContext, useLiveQuery } from 'electric-sql/react'
+
+const { ElectricProvider, useElectric } = makeElectricContext<Electric>()
+
+const config = {
+  auth: {
+    token: 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJsb2NhbC1kZXZlbG9wbWVudCIsInR5cGUiOiJhY2Nlc3MiLCJ1c2VyX2lkIjoidGVzdC11c2VyIiwiaWF0IjoxNjg3ODc3OTQ1LCJleHAiOjE2OTc4ODE1NDV9.L5Ui2sA9o5MeYDuy67u9lBV-2FzpOWL9dKcitRvgorg',
```
thruflo: We have a function to generate insecure `user_id`s now. Let's use it rather than hardcoding like this.

starter/template/src/Example.tsx
```diff
@@ -0,0 +1,95 @@
+import React, { useEffect, useState } from 'react'
+import './Example.css'
+
+import { dbSchema, Electric } from './generated/client'
```
thruflo: I think this is `schema` now, right?

starter/template/src/Example.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const ExampleComponent = () => {
+  const { db } = useElectric()!
+  const { results } = useLiveQuery(db.items.liveMany({})) // select all items
+
+  const addItem = async () => {
+    await db.items.create({
+      data: {
+        value: crypto.randomUUID(),
+        // uncomment the line below after migration
```
thruflo: I think this to be removed?

starter/template/src/Example.tsx
```diff
@@ -0,0 +1,95 @@
...
+        //other_value: crypto.randomUUID(),
+      }
+    })
+  }
+
+  const clearItems = async () => {
+    await db.items.deleteMany() // delete all items
+  }
+
+  // After the migration, comment out this code and uncomment code block below
```
thruflo: I think this after migration stuff is out of scope for the starter?

starter/template/src/declarations.d.ts
thruflo, line 1: We don't need this any more. Also remove from any deps and lock files.

starter/template/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Whenever the data model in Postgres changes, the Electric client must be updated accordingly.
+To this end, generate an updated Electric client:
+```sh
+yarn client:generate
+```
+
+Alternatively, you can run the above command in watch mode such that the client is re-generated automatically when the data model changes:
+```sh
+yarn client:generate -- --watch
```
thruflo, line 51: To sanity check: is this the correct syntax?

starter/template/local-stack/docker-compose.yaml
```diff
@@ -0,0 +1,40 @@
...
+    extends:
+      file: "./compose/docker-compose-templates.yaml"
+      service: electric_template
+    environment:
+      <<: *preshared-auth-variables
+      DATABASE_URL: postgresql://postgres:password@postgres_1:5432/electric
+      ELECTRIC_HOST: electric_1
+    ports:
+      - 5050:5050
+      - 5133:5133
```
thruflo, line 36: Also `- 5433:5433`.

starter/template/local-stack/docker-compose.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - pg_data:/var/lib/postgresql/data
+
+  electric_1:
+    extends:
+      file: "./compose/docker-compose-templates.yaml"
+      service: electric_template
+    environment:
+      <<: *preshared-auth-variables
+      DATABASE_URL: postgresql://postgres:password@postgres_1:5432/electric
+      ELECTRIC_HOST: electric_1
```
thruflo, line 33: This is now `LOGICAL_PUBLISHER_HOST`.

kevin-dp: Yes in Electric `0.5.0-next.5` but latest published docker image is `0.5.0-next.4` which still uses `ELECTRIC_HOST`.

starter/template/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "electric-sql-web-example",
```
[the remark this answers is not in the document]
kevin-dp: it's indeed the wrong default name but it is being overwritten by the chosen app name anyway (`starter/src/index.ts`, L46).

starter/template/README.md
kevin-dp: The `--` separator is needed when passing argument to a yarn/npm command. However, this is no longer needed as we introduced a separate `client:watch` command.

starter/template/local-stack/docker-compose.yaml
kevin-dp: What runs on 5433? The `electric/local-stack` does not expose port 5433 either, should it?

starter/template/public/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=360,initial-scale=1.0" />
+    <title>Web Example - ElectricSQL</title>
```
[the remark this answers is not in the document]
kevin-dp: Do you want it to be `ElectricSQL starter template` or the name of the app chosen by the dev?

starter/template/local-stack/docker-compose.yaml
thruflo: It's the default `LOGICAL_PUBLISHER_PORT` https://main--wip-electric-sql-docs.netlify.app/docs/api/service#core-config

starter/template/local-stack/docker-compose.yaml
thruflo: Ok, perhaps have both with a note to remove the legacy one?

starter/template/public/index.html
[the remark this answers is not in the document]
thruflo: Probably the latter.

.changeset/wicked-students-suffer.md
thruflo, line 1: Is this a dupe?

examples/starter/template/backend/compose/dot_envrc
```diff
@@ -0,0 +1,3 @@
+# Docker images for the local stack
+export POSTGRESQL_IMAGE=postgres:14-alpine
+export ELECTRIC_IMAGE=electricsql/electric:0.5.0-next.4
```
thruflo, line 3: Should this be `electric:next`?

examples/starter/template/backend/compose/postgres/init.sql
```diff
@@ -0,0 +1,2 @@
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
thruflo, line 1: Is this necessary for Electric? Because the example doesn't use a uuid data type.

thruflo:
See https://github.com/electric-sql/docs/pull/34

examples/starter/template/backend/compose/dot_envrc
kevin-dp: Right, should be `electric:latest` actually (we don't have a `next` tag on docker hub).

thruflo (approved):

PR #316, opened 2023-08-07
Add environment variable to configure ssl on pg connection

magnetised:


alco:
@magnetised correct me if I'm wrong please: this setting is supposed to affect Electric's behaviour as a DB client when it is connecting to an external DB hosted by Crunchy. If Crunchy only works over SSL, I believe we could provide a better experience by emulating Postgres' `sslmode=prefer` behaviour by default: that is, try connecting over SSL first and if that fails, fall back to unencrypted connection.

We can achieve that by setting `ssl=true` in our Postgres connector config unconditionally. When epgsql sees the `ssl` flag in connection opts, it will attempt to connect over SSL but fall back to unencrypted connection, unless `ssl=required`. In that latter case, it will fail the connection attempt with an error. See https://github.com/epgsql/epgsql/blob/7ba52768cf0ea7d084df24d4275a88eef4db13c2/src/commands/epgsql_cmd_connect.erl#L119-L142.

---

Tangential: the connection opts we're using for epgsql have no effect on the options we pass to postgres in order to establish a reverse replication connection. I'm looking at [these](https://github.com/electric-sql/electric/blob/87ab51ea1cbd1dab32ca2deb1f851239f8bbda7b/components/electric/lib/electric/replication/postgres/tcp_server.ex#L195-L201) [two](https://github.com/electric-sql/electric/blob/87ab51ea1cbd1dab32ca2deb1f851239f8bbda7b/components/electric/lib/electric/replication/postgres/tcp_server.ex#L365-L369) functions and I'm not sure whether we do or do not support SSL for incoming connections from Postgres. Does Crunchy have an opinion about it? Or a way to configure it?

icehaunter:
@alco we definitely don't support SSL on connection from PG (or clients) to ourselves

thruflo:
Let's raise an issue for that ...

icehaunter:
Changelog please

alco (approved):
:+1: 
Don't forget to update the docs too - https://main--wip-electric-sql-docs.netlify.app/docs/api/service#core-config

PR #317, opened 2023-08-07
examples: handle unmount in web-wa-sqlite example.

thruflo:


kevin-dp (approved):
Looks good :-)
Just a minor comment to make the definition of the `items` constant more idiomatic.

PR #319, opened 2023-08-07
fix(satellite): vax-807 cannot catch up to the servers current state

balegas:
After server is gone, it loses information about clients. Right now it is throwing a BEHIND_WINDOW error when a client tries to reconnect and server doesn't find any info regarding that client. It's arguable if this is the appropriate error to throw, but it highlighted a problem with failures to reconnect on error.

1. the client would not trigger subscriptions cleanup because of improper configuration
2. it would send unsubscribe requests to the server without first starting replication, which would result in an error
3. there was de-referencing error to a variable that was out of scope

After fixing all that, application was connecting but not catching up with the server yet. The problem was that we had cleared subscriptions and did not request them again. There is a question here: is re-subscribing an internal procedure of the client, or should it be tied to the lifecycle of the application and the app should decide what to do... the second approach raises new issues because we would need to appropriately call sync based on reporting the errors to the app. I've taken the first approach for now.

I'd like to know your feedback on the approach

### DIFF
--- a/clients/typescript/src/satellite/client.ts
+++ b/clients/typescript/src/satellite/client.ts
@@ -853,18 +853,16 @@ export class SatelliteClient extends EventEmitter implements Client {
     const messageOrError = toMessage(data)
     if (Log.getLevel() <= 1 && !(messageOrError instanceof SatelliteError))
       Log.debug(`[proto] recv: ${msgToString(messageOrError)}`)
-    if (messageOrError instanceof Error) {
-      this.emit('error', messageOrError)
-    } else {
-      const handler = this.handlerForMessageType[messageOrError.$type]
-      try {
-        const response = handler.handle(messageOrError)
-        if (handler.isRpc) {
-          this.emit('rpc_response', response)
-        }
-      } catch (error) {
-        Log.warn(`uncaught errors while processing incoming message: ${error}`)
+    try {
+      const message: SatPbMsg = toMessage(data)
+      const handler = this.handlerForMessageType[message.$type]
+      const response = handler.handle(message)
+      if (handler.isRpc) {
+        this.emit('rpc_response', response)
       }
+    } catch (error) {
+      Log.warn(`uncaught errors while processing incoming message: ${error}`)
+      this.emit('error', error)
     }
   }
 
@@ -1148,12 +1146,12 @@ function serializeNullData(): Uint8Array {
   return typeEncoder.text('')
 }
 
-export function toMessage(data: Uint8Array): SatPbMsg | SatelliteError {
+export function toMessage(data: Uint8Array): SatPbMsg {
   const code = data[0]
   const type = getTypeFromCode(code)
   const obj = getObjFromString(type)
   if (obj == undefined) {
-    return new SatelliteError(
+    throw new SatelliteError(
       SatelliteErrorCode.UNEXPECTED_MESSAGE_TYPE,
       `${code})`
     )
--- a/clients/typescript/src/satellite/client.ts
+++ b/clients/typescript/src/satellite/client.ts
@@ -242,8 +242,7 @@ export class SatelliteClient extends EventEmitter implements Client {
       this.close()
     }
 
-    const initializing = emptyPromise()
-    this.initializing = initializing
+    this.initializing = emptyPromise()
 
     const connectPromise = new Promise<void>((resolve, reject) => {
       // TODO: ensure any previous socket is closed, or reject
@@ -291,7 +290,7 @@ export class SatelliteClient extends EventEmitter implements Client {
     return backOff(() => connectPromise, retryPolicy).catch((e) => {
       // We're very sure that no calls are going to modify `this.initializing` before this promise resolves
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      initializing.reject(e)
+      this.initializing?.reject(e)
       throw e
     })
   }
@@ -854,7 +853,7 @@ export class SatelliteClient extends EventEmitter implements Client {
     if (Log.getLevel() <= 1 && !(messageOrError instanceof SatelliteError))
       Log.debug(`[proto] recv: ${msgToString(messageOrError)}`)
     try {
-      const message: SatPbMsg = toMessage(data)
+      const message = toMessage(data)
       const handler = this.handlerForMessageType[message.$type]
       const response = handler.handle(message)
       if (handler.isRpc) {
@@ -862,7 +861,12 @@ export class SatelliteClient extends EventEmitter implements Client {
       }
     } catch (error) {
       Log.warn(`uncaught errors while processing incoming message: ${error}`)
-      this.emit('error', error)
+      if (error instanceof SatelliteError) {
+        this.emit('error', error)
+      } else {
+        // This is an unexpected runtime error
+        throw error
+      }
     }
   }
 
--- a/clients/typescript/src/satellite/client.ts
+++ b/clients/typescript/src/satellite/client.ts
@@ -238,7 +238,8 @@ export class SatelliteClient extends EventEmitter implements Client {
   connect(
     retryHandler?: (error: any, attempt: number) => boolean
   ): Promise<void> {
-    this.initializing = emptyPromise()
+    const initializing = emptyPromise()
+    this.initializing = initializing
 
     const connectPromise = new Promise<void>((resolve, reject) => {
       // TODO: ensure any previous socket is closed, or reject
@@ -286,7 +287,7 @@ export class SatelliteClient extends EventEmitter implements Client {
     return backOff(() => connectPromise, retryPolicy).catch((e) => {
       // We're very sure that no calls are going to modify `this.initializing` before this promise resolves
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      this.initializing!.reject(e)
+      initializing.reject(e)
       throw e
     })
   }
@@ -516,7 +517,10 @@ export class SatelliteClient extends EventEmitter implements Client {
   }
 
   unsubscribe(subIds: string[]): Promise<UnsubscribeResponse> {
-    if (this.inbound.isReplicating !== ReplicationStatus.ACTIVE) {
+    if (
+      this.inbound.isReplicating !== ReplicationStatus.ACTIVE &&
+      this.inbound.isReplicating !== ReplicationStatus.SERVER_ERROR
+    ) {
       return Promise.reject(
         new SatelliteError(
           SatelliteErrorCode.REPLICATION_NOT_STARTED,
--- a/clients/typescript/src/satellite/client.ts
+++ b/clients/typescript/src/satellite/client.ts
@@ -815,6 +815,9 @@ export class SatelliteClient extends EventEmitter implements Client {
   }
 
   private handleError(error: SatErrorResp) {
+    // TODO: this causing intermittent issues in tests because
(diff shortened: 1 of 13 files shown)


clients/typescript/src/satellite/index.ts
```diff
@@ -60,10 +58,7 @@ export interface Satellite {
 
   connectivityState?: ConnectivityState
 
-  start(
-    authConfig: AuthConfig,
-    opts?: SatelliteReplicationOptions
-  ): Promise<ConnectionWrapper>
+  start(authConfig: AuthConfig): Promise<ConnectionWrapper>
```
balegas, line 61: flattened options. Not necessary for now, I guess.
Happy to revert to avoid changing api and types but need to thread through behindWindow flag in multiple places

clients/typescript/src/satellite/process.ts
```diff
@@ -566,19 +559,17 @@ export class SatelliteProcess implements Satellite {
...
-    Log.error('Encountered a subscription error: ' + satelliteError.message)
+    Log.error('encountered a subscription error: ' + satelliteError.message)
 
     this._lsn = undefined
     await this.adapter.runInTransaction(
       this._setMetaStatement('lsn', null),
       this._setMetaStatement('subscriptions', this.subscriptions.serialize())
     )
 
-    await this.client.unsubscribe(ids)
```
balegas, line 579: this causes an error because replication state hasn't been changed. However, we don't need to call unsubscribe because we're clearing local state and next time we connect, server will throw away client data.

It's a small optimisation to ensure we clear subscriptions early. Don't mind doing that.

clients/typescript/src/satellite/process.ts
```diff
@@ -638,18 +627,19 @@ export class SatelliteProcess implements Satellite {
     } catch (error: any) {
       if (
         error.code == SatelliteErrorCode.BEHIND_WINDOW &&
-        opts?.clearOnBehindWindow
+        this.opts?.clearOnBehindWindow
       ) {
-        return this._handleSubscriptionError(error).then(() =>
-          this._connectAndStartReplication()
-        )
+        await this._handleSubscriptionError(error)
```
balegas, line 632: this seems to be the right procedure on the client, but we still missing re-subscriptions from the app.

clients/typescript/src/util/types.ts
```diff
@@ -35,6 +35,7 @@ export class SatelliteError extends Error {
 }
 
 export enum SatelliteErrorCode {
+  CONNECTION_FAILED,
```
balegas, line 38: Error thrown if client can't reach Electric. Opened VAX-830 because client should long poll to try to reconnect to server. I think I wrote that code, but doesn't seem to be working

clients/typescript/src/satellite/process.ts
```diff
@@ -638,18 +627,19 @@ export class SatelliteProcess implements Satellite {
...
+        await this.client.close()
```
balegas, line 633: connectivity hook does this, but it might race, so better do it twice, I think. Although doesn't seem to create a lot of trouble not to do it

thruflo:
Will these changes will also fix https://linear.app/electric-sql/issue/VAX-831/replication-not-active-error-when-cant-reach-backend ?

clients/typescript/src/satellite/client.ts
```diff
@@ -284,7 +285,7 @@ export class SatelliteClient extends EventEmitter implements Client {
     return backOff(() => connectPromise, retryPolicy).catch((e) => {
       // We're very sure that no calls are going to modify `this.initializing` before this promise resolves
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      this.initializing!.reject(e)
```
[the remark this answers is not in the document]
icehaunter: Why not do this then?

```suggestion
      this.initializing?.reject(e)
```

icehaunter (changes requested):
Great work! I have some nitpicks/suggestions and one possible bug highlighted

clients/typescript/src/satellite/client.ts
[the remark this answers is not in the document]
balegas: I'm afraid there is a larger issue here that is not being addressed. 
- I feel like the behind window exception should not reach here. Maybe it doesn't any longer, I can check
- but, initializing could only be undefined at this point if we closed before. In that case, I think we should reject on close (and provide error context to close).

clients/typescript/src/satellite/client.ts
```diff
@@ -836,20 +843,18 @@ export class SatelliteClient extends EventEmitter implements Client {
...
+      Log.info(`Received message ${JSON.stringify(message)}`)
+
+      const handler = this.handlerForMessageType[message.$type]
+      const response = handler.handle(message)
+      if (handler.isRpc) {
+        this.emit('rpc_response', response)
       }
+    } catch (error) {
+      Log.warn(`uncaught errors while processing incoming message: ${error}`)
+      this.emit('error', error)
```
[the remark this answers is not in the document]
balegas: I agree, its swallowing too much info. Will apply the suggestion

clients/typescript/src/satellite/client.ts
```diff
@@ -969,7 +974,7 @@ export class SatelliteClient extends EventEmitter implements Client {
   }
 
   private sendMessage<T extends SatPbMsg>(request: T) {
-    Log.debug(`Sending message ${JSON.stringify(request)}`)
+    Log.info(`Sending message ${JSON.stringify(request)}`)
```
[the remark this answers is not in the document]
balegas: Right, unintended.

Let me raise here that I would like very much to review the patters we use in the e2e for matching messages. Right now is impossible to print messages that have data, because prints all the bytes. I wanted to just print the type, but found out that e2e relies on more info on this strings.

I was thinking two approaches, but both are a bit processing heavy:
- parse the message and hide all the byte arrays (I tried it, but I think this needs some deep traversal to do right)
- have specific prints for each messages

clients/typescript/src/satellite/process.ts
```diff
@@ -316,6 +313,7 @@ export class SatelliteProcess implements Satellite {
       }, stmts)
 
     await this.adapter.runInTransaction(...stmts)
+    tablenames.forEach((t) => this.notifier.actuallyChanged(t, []))
```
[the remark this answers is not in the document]
balegas: Thanks for finding this, I think this is also some remnants of my root cause analysis. I'll try to check hwo things work without that line -- which should be doing nothing atm.

clients/typescript/src/satellite/process.ts
```diff
@@ -560,35 +558,60 @@ export class SatelliteProcess implements Satellite {
...
   }
 
+  async _handleBehindWindow(): Promise<void> {
+    Log.warn(
+      'client cannot resume replication from server, resetting replication state'
+    )
+    const subscriptionIds = this.subscriptions.getFulfilledSubscriptions()
+    const shapeDefs: ClientShapeDefinition[] = subscriptionIds
+      .map((subId) => this.subscriptions.shapesForActiveSubscription(subId))
+      .filter((s): s is ShapeDefinition[] => s != undefined)
```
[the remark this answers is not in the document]
balegas: I've been suggested to change this, and it's a think that I should just change my default.

for my edification, what is the difference if we're comparing with undefined?

balegas (commented):
Thanks, will review all suggestions shortly

clients/typescript/src/satellite/process.ts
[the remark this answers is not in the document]
icehaunter: Yeah, that line is a no-op right now

clients/typescript/src/satellite/process.ts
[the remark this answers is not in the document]
icehaunter: Functionally, `null == undefined` but `null !== undefined`, but practically since `==` (and `!=`) does implicit type conversions before doing the comparison, it's a really good idea to never use `==` at all. It can hide errors (although less so with typescript, but if you let `any` sneak in which can happen you're in for a runtime treat). It's a good thing to never use it so you don't forget it in a critical place later. ESLint even has a  rule to just disallow non-strict equality checks: https://eslint.org/docs/latest/rules/eqeqeq.

icehaunter (approved):

clients/typescript/src/satellite/process.ts
```diff
@@ -656,30 +656,32 @@ export class SatelliteProcess implements Satellite {
...
+          error.code == SatelliteErrorCode.BEHIND_WINDOW &&
+          this.opts?.clearOnBehindWindow &&
+          !currentlyReconnecting
+        ) {
+          return await this._handleBehindWindow()
+        }
+
+        throw error
+      }
     } catch (error: any) {
```
icehaunter, line 677: This feels like a very weird and uncommon pattern for Typescript. I don't understand how it's better than doing that `if` with `return` within a `catch` clause. Getting the error as a `Result` and then immediately throwing it JUST to catch it?

clients/typescript/src/satellite/client.ts
```diff
@@ -636,23 +645,25 @@ export class SatelliteClient extends EventEmitter implements Client {
...
     if (this.inbound.isReplicating == ReplicationStatus.STARTING) {
       if (resp.err) {
         this.inbound.isReplicating = ReplicationStatus.STOPPED
-        this.emit('error', startReplicationErrorToSatelliteError(resp.err))
+        return { error: startReplicationErrorToSatelliteError(resp.err) }
       } else {
         this.inbound.isReplicating = ReplicationStatus.ACTIVE
       }
     } else {
-      this.emit(
```
icehaunter, line 642: Why aren't we doing `this.emit` anymore and it's fine?

icehaunter (changes requested):
Hey, I'm a bit unsure of very suddenly using a `Result ` monad in just this one place with no visible benefit. Can we discuss the reason for this approach?

clients/typescript/src/satellite/process.ts
balegas: nothing against that suggestion. Applied it.

clients/typescript/src/satellite/client.ts
balegas: RPC errors where being handled inconsistently. Some would emit 'error', others would reply with a response object with an error field.
In practice, in the first case we would handle the response in a catch clause, in the latter we would process the response with error.
I consolidated on the latter.

clients/typescript/src/satellite/client.ts
```diff
@@ -1009,15 +1038,17 @@ export class SatelliteClient extends EventEmitter implements Client {
       }, this.opts.timeout)
 
       // reject on any error
-      this.once('error', (error: SatelliteError) => {
+      const errorListener = (error: SatelliteError) => {
```
balegas: In fact we can't ensure this is a SatelliteError. Reverting

balegas (commented):
Thanks, applied your suggestion and replied to your question

icehaunter (changes requested):
This is the final discussion comment from me, although I somewhat feel like we're complicating this more than it should be. What do you think? If you feel like this is nitpicking - feel free to merge, we can give it another pass at a later date, but it seems like the current complicated implementation may hide some bugs - I believe I may have spotted at least two

icehaunter (approved):

balegas:
@icehaunter thanks for bearing with me on all the back-and-forth with this one

PR #320, opened 2023-08-08
chore(electric): Add server-side validation of row values incoming from Satellite clients

alco:
Closes VAX-813.


components/electric/lib/electric/postgres/postgres_manager.ex
```diff
@@ -160,6 +163,10 @@ defmodule Electric.Replication.PostgresConnectorMng do
            {:ok, _} <-
              Client.create_subscription(conn, subscription, publication, electric_connection),
            {:ok, oids} <- Client.query_oids(conn),
+           oids =
```
magnetised, line 166: can we split this stage into a separate function and keep this `with` block as clean as possible, in fact why split this bit out of the `OidDatabase` module? could easily just pass in the `conn_config` and keep all this messiness out of this "manager"

components/electric/lib/electric/postgres/postgres_manager.ex
```diff
@@ -169,4 +176,21 @@ defmodule Electric.Replication.PostgresConnectorMng do
       end
     end)
   end
+
+  defp assoc_oids_with_postgrex_extensions(oids, conn_config) do
+    # Here we're using Postgrex's private APIs to get hold of its typename-oid mapping table that includes
+    # Postgrex.Extension.* module for each of the supported Postgres types.
+    # We're associating the extension module with type oid in order to be able to validate values of the type on demand.
+    type_server_key = {conn_config[:host], conn_config[:port], conn_config[:database]}
+    type_server = Postgrex.TypeSupervisor.locate(Postgrex.DefaultTypes, type_server_key)
```
magnetised, line 185: how stable are these APIs? 😰 

well done for digging them out though!

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -344,66 +344,82 @@ defmodule Electric.Satellite.Serialization do
...
+    decode_record(row, columns, false)
   end
 
-  defp row_to_map(_columns, nil, false) do
+  defp decode_record(nil, _columns, false) do
     raise "protocol violation, empty row"
   end
 
-  defp row_to_map(_columns, nil, true), do: nil
+  defp decode_record(nil, _columns, true), do: nil
```
magnetised, line 389: this 3rd argument boolean would be clearer as an atom, say `:nil_valid`, `:nil_ok` or something

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -344,66 +344,82 @@ defmodule Electric.Satellite.Serialization do
...
+    row |> validate_values!(column_types) |> row_to_map(column_names)
+  end
 
-    {row, _, []} =
-      Enum.reduce(columns, {%{}, bitmask_list, values}, fn
-        column, {map0, [0 | bitmask_list0], [value | values0]} ->
-          {Map.put(map0, column, value), bitmask_list0, values0}
+  defp validate_values!(%SatOpRow{nulls_bitmask: bitmask, values: values} = row, column_types) do
+    Enum.each(Enum.zip([column_types, values, bitmask_to_boolean(bitmask)]), fn
+      {_type, _val, true = _null?} ->
```
magnetised, line 399: i think the need to label the boolean with a `_null?` var suggests that maybe `bitmask_to_boolean` should be `bit_mask_to_null` and return `:null` rather than true/false

magnetised (changes requested):
few nitpicks but otherwise great

components/electric/lib/electric/postgres/postgres_manager.ex
alco: Last time the `Postgrex.Types` module saw a behaviour-changing update was in 2021. And both `Postgrex.TypeSupervisor` and `Postgrex.TypeServer` were last updated in 2019.

components/electric/lib/electric/satellite/validation.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Electric.Satellite.Validation do
+  @spec assert_type!(String.t(), atom) :: :ok | no_return
+  def assert_type!(val, col_type) do
```
icehaunter, line 3: Could you add a docstring here?

components/electric/lib/electric/satellite/validation.ex
```diff
@@ -0,0 +1,36 @@
...
+    postgrex_ext = Electric.Postgres.OidDatabase.postgrex_ext_for_name(col_type)
+
+    # encode_value() will raise if val is invalid. We're not interested in the return value itself.
+    _ =
+      val
+      |> parse_column_value(col_type)
+      |> Postgrex.DefaultTypes.encode_value(postgrex_ext)
```
icehaunter, line 10: So correct me if I'm wrong, but it feels like we're relying on multiple internal Postgrex APIs essentially just to validate ranges on types. I'm not sure how much I agree with this, given that we're doing the actual hard lifting below, in `parse_column_value/2`, to do a first pass of validation for data coming form Satellite. If we were calling `.decode_value()` of some kind on Postgres, I would have been fine with it, but pulling in a bunch of code to do very little seems a bit unreasonable. Especially when we're going to need to evaluate row constraints on parsed values anyway - it seems like maybe keeping most code in our codebase would be reasonably OK?

My argument is fully based on the notion that we're already doing heavy lifing on our side in parsing, and we're going to do that even more: our "intermediate representation" of a value that SQLite can interact with is going to likely differ from PG representaiton (or we're going to do quite a bit of conversions on the client, which is also OK). In particular, think of array serialization: SQLite can only benefit from JSON arrays, while PG can have the same array as `int4[]` and we're going to need to marry that up somehow. Seems to me that most complicated logic will live "on our side" anyway, so I don't see a need to drop down to Postgrex just to check if int is indeed int4.

WDYT?

alco:
@magnetised Great feedback! I've implemented your suggestions.

magnetised (approved):
magic

icehaunter (changes requested):
Good work, cool to see a bit more rigidity in the system! I have one slightly philosophical question.

components/electric/lib/electric/satellite/validation.ex
alco: You're absolutely right!

I may have been carried away by thoughts of using more of Postgrex further down the road, e.g. as a more Elixir-native way of interfacing with Postgres compared to epgsql. I was also thinking we could start using Postgrex's encoding/decoding functionality to switch to the binary encoding of row data in the logical replication stream.

But all of that is irrelevant to the current task. And as I've learned by now, Postgrex does not do everything we need to be doing in terms of validation. For example, it doesn't do any validation for sized `char` and `varchar` types, it simply lets Postgres report an error if the user data exceeds the length limit.

I have removed Postgrex and rethought the idea of validation a bit, making more emphasis on decoding values coming from Satellite, rather than validating them. This will be mirror by the corresponding encoding of some types as can be seen already in the upcoming PR that adds support for `timestamp` types - https://github.com/electric-sql/electric/pull/330.

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -392,43 +392,60 @@ defmodule Electric.Satellite.Serialization do
...
+  end
 
-  defp apply_nulls_bitmask([_ | vals], <<1::1, bitmask::bitstring>>),
-    do: [nil | apply_nulls_bitmask(vals, bitmask)]
+  @int2_range -32768..32767
+  @int4_range -2_147_483_648..2_147_483_647
+  @int8_range -9_223_372_036_854_775_808..9_223_372_036_854_775_807
 
-  defp apply_nulls_bitmask([], _), do: []
+  defp assert_integer_in_range!(int, :int2) when int in @int2_range, do: :ok
```
magnetised, line 448: this function will raise a matcherror (I think) when an integer is out of range -- wouldn't it be helpful to raise some `"integer column #{column_name} value #{value} is outside the range of a #{type}"` type exception

components/electric/test/electric/satellite/ws_validations_test.exs
```diff
@@ -152,13 +152,16 @@ defmodule Electric.Satellite.WsValidationsTest do
         ctx.db,
         vsn,
         "public.foo",
-        "CREATE TABLE public.foo (id TEXT PRIMARY KEY, f4 REAL, f8 DOUBLE PRECISION)"
```
alco: I haven't found a way to reliably tell if a given float number can fit in a 4-byte float without loss of precision. For now, I'm leaving support only for double precision floats.

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -392,43 +392,60 @@ defmodule Electric.Satellite.Serialization do
     raise "protocol violation, empty row"
   end
 
-  defp decode_record(%SatOpRow{} = row, columns, _) do
-    column_types = Enum.map(columns, &String.to_existing_atom(&1.type))
-    column_names = Enum.map(columns, & &1.name)
-    row |> validate_values!(column_types) |> row_to_map(column_names)
+  defp decode_record(%SatOpRow{nulls_bitmask: bitmask, values: values}, columns, _) do
+    values
+    |> apply_nulls_bitmask(bitmask)
```
magnetised, line 397: this is very much in the hot path of the system, could we do the same thing in a single pass?

magnetised (commented):
actually much simpler without postgrex huh. I'm not particularly worried about the performance tweak but I think a more helpful error in the out-of-range-integer case is worth it

in fact, should this validation step raise at all? don't we want to feed back to the client rather than crash the process?

alco:
@magnetised 

> in fact, should this validation step raise at all? don't we want to feed back to the client rather than crash the process?

The validation is initiated via the call to `Serialization.deserialize_trans()` from `Electric.Satellite.Protocol`. The call is wrapped in a `try` block there, so any exception is transformed into an `INVALID_REQUEST` response to the client. No processes crash as a result of this.

On the server, we do get the exception logged with all the details. In the example with `int2`, the log entry is pretty clear about which function clause failed to match (that one that has `:int2` as the second argument).

As for the client, we don't have a proper mechanism for reporting such critical errors to the user. From the perspective of the client app, it did its write locally and moved on. It doesn't track how this write is then synced to the server in background and results in an error response from the server. We want to avoid having this is a valid error state in practice because once a local write has succeeded, it should eventually get synced to the server.

What I think would be useful to add eventually is a way to report validation faliures to the app developer.

components/electric/lib/electric/satellite/serialization.ex
alco: Do you mean squashing this function pipeline into a single loop that does everything: check for nulls, validate types and produce the {name, value} tuple?

magnetised:
> On the server, we do get the exception logged with all the details. In the example with `int2`, the log entry is pretty clear about which function clause failed to match (that one that has `:int2` as the second argument).

ok, seems vaguely irresponsible but ok... ;)

> As for the client, we don't have a proper mechanism for reporting such critical errors to the user. From the perspective of the client app, it did its write locally and moved on. It doesn't track how this write is then synced to the server in background and results in an error response from the server. We want to avoid having this is a valid error state in practice because once a local write has succeeded, it should eventually get synced to the server.

that's a much bigger fish. gotcha

components/electric/lib/electric/satellite/serialization.ex
alco: Updated.

PR #321, opened 2023-08-08
fix(electric): Adorn electric-generated trigger functions with SECURITY DEFINER

alco:
As explained in the official docs[1]:

  > SECURITY INVOKER indicates that the function is to be executed
  > with the privileges of the user that calls it. That is the default.
  > SECURITY DEFINER specifies that the function is to be executed
  > with the privileges of the user that owns it.

[1]: https://www.postgresql.org/docs/current/sql-createfunction.html

While testing the scenario described in VAX-806, I was hitting multiple permission denials, such as

    ERROR:  permission denied for schema electric
    LINE 3:             __current_tag electric.tag;

and, after granting the demo user access to the "electric" schema,

    ERROR:  permission denied for table shadow__public__sliders
    CONTEXT:  SQL statement "INSERT INTO electric.shadow__public__sliders (_last_modified, _tag, _tags, _resolved,
                                              id, _tag_value, _tag_demo_id, _tag_demo_name, _tag_electric_user_id)

The root cause of those permission issues was the fact that the demo user didn't have access to all the internal types and tables that live under the "electric" schema.

Instead of granting the demo user access to everything, we can alter the definitions of trigger functions to ensure they'll run with the priveleges of the user who created them, that is, Electric's "super user".

I found this solution at https://github.com/orgs/supabase/discussions/306#discussioncomment-138424

### DIFF
--- a/components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers/trigger_function_installers.sql
+++ b/components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers/trigger_function_installers.sql
@@ -36,7 +36,7 @@ BEGIN
     EXECUTE format($injected$
         CREATE OR REPLACE FUNCTION electric.%1$I()
         RETURNS TRIGGER
-        LANGUAGE PLPGSQL AS
+        LANGUAGE PLPGSQL SECURITY DEFINER AS
         $function$
         DECLARE
             shadow_row electric.%2$I%%ROWTYPE;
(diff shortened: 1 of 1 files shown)


magnetised (changes requested):
It's always worth reading every word of the pg docs isn't it? good find!

I hadn't see VAX-806 but while deploying the intro-backend app I'd hit very similar problems. Basically because we need to create event triggers we need to run electric with root-level db access. Because of this if you try to connect your app with a lower priv user, you just get errors when e.g. electrifying things. 

Could you apply the same change to the other functions we declare, at least the ones reading and writing to the tables in the electric schema (so the various ddlx functions are fine as-is).

Even if we move to the proxy, so will require a much lower-level privilege level for the electric db connection, I suspect we'll hit the same problem

thruflo:
@alco very nice 👍 

@magnetised hang on, does this somehow allow us to work with lower permissions? So e.g.: the Electric service could have low perms, whilst the migration application user needs higher? That would be *awesome*!!!!

magnetised (approved):
Approving with other work tbc in separate pr

PR #330, opened 2023-08-09
feat(electric, satellite): Add support for timestamp and timestamptz column types

alco:
Closes VAX-818.

### DIFF
--- a/components/electric/lib/electric/satellite/serialization.ex
+++ b/components/electric/lib/electric/satellite/serialization.ex
@@ -399,7 +399,7 @@ defmodule Electric.Satellite.Serialization do
   end
 
   defp op_to_change(%SatOpInsert{row_data: row_data, tags: tags}, columns) do
-    %NewRecord{record: decode_record(row_data, columns), tags: tags}
+    %NewRecord{record: decode_record!(row_data, columns), tags: tags}
   end
 
   defp op_to_change(
@@ -407,7 +407,7 @@ defmodule Electric.Satellite.Serialization do
          columns
        ) do
     %UpdatedRecord{
-      record: decode_record(row_data, columns, :allow_nulls),
+      record: decode_record!(row_data, columns, :allow_nulls),
       old_record: nil,
       tags: tags
     }
@@ -418,8 +418,8 @@ defmodule Electric.Satellite.Serialization do
          columns
        ) do
     %UpdatedRecord{
-      record: decode_record(row_data, columns),
-      old_record: decode_record(old_row_data, columns),
+      record: decode_record!(row_data, columns),
+      old_record: decode_record!(old_row_data, columns),
       tags: tags
     }
   end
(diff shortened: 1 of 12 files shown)


thruflo:
So you think we need to serialise the values in the DAL?

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -25,6 +27,7 @@ defmodule Electric.Satellite.Serialization do
       int2 int4 int8
       float8
       text
+      timestamp timestamptz
```
icehaunter, line 30: This function is weird placed here, because it's used in the `20230605141256_electrify_function.ex` migration, but that migration is not reran on Electric upgrade, meaning in-place upgrades are not going to expand functionality to allow these new columns. I think we either need a special function that runs on every startup against PG and updates the `electrify` function to use the new types, or we need to add a new migration to do that for each type PR

components/electric/test/electric/satellite/serialization_test.exs
```diff
@@ -1,94 +1,144 @@
...
+        rescue
+          _ -> :ok
+        else
+          val -> flunk("Expected decode_record() to raise but it returned #{inspect(val)}")
+        end
+      end)
+    end
 
-  test "test row deserialization with long bitmask" do
-    mask = <<0b1101000010000000::16>>
```
icehaunter, line 42: Could you please preserve the long bimask test? Bitmask format (direction, namely) discrepancy between server and client was a source of weird bugs very early in development and I'd have this double-checked

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -462,6 +493,21 @@ defmodule Electric.Satellite.Serialization do
     val
   end
 
+  def decode_column_value(val, :timestamp) do
```
icehaunter, line 496: nitpick: I would appreciate it if you renamed `decode_column_value` and `decode_record` functions to `decode_column_value!` and `decode_record!` because it took me a minute to recall that missed matches here are intended behaviour

icehaunter (changes requested):

icehaunter:
Sorry, VSCode Github plugin submitted the review without the top comment

The PR looks good to me! I don't have any issues with timestamp implementation and details, but some auxiallary questions. The "no migration" problem seems quite important

components/electric/lib/electric/satellite/serialization.ex
alco: It's a very good point. But since it's not directly related to an implementation of any specific type, I'm going to take the liberty of addressing it separately, just prior to merging the PRs that add support for new types.

I've addressed your other comments.

icehaunter: Ok, sounds good

icehaunter (approved):

components/electric/lib/electric/satellite/serialization.ex
alco: Garry has proposed to define all PG functions when Electric starts to avoid issues with putting them in migrations --  https://linear.app/electric-sql/issue/VAX-1016/change-migration-mechanism-for-pg-functions.

Before that is implemented, I can make the type-checking function defined at runtime to address the immediate issue of adding new types in a backwards-compatible manner. But that's better done after all the current type PRs are merged to avoid another cascading rebase fest.

PR #332, opened 2023-08-10
chore(starter): improved output and commands

kevin-dp:
This PR hides output from the underlying processes when running `npx create-electric-app my-app` as well as in the output of the yarn commands. Also introduces a new command `db:connect` that connects to the PG database using the `DATABASE_URL` env variable. If that variable is not defined it assumes that we're running the entire stack and it will connect to the PG that is running in Docker.

### DIFF
--- a/examples/starter/src/index.ts
+++ b/examples/starter/src/index.ts
@@ -6,6 +6,9 @@ import { spawn } from 'child_process'
 import * as fs from 'fs/promises'
 import { fileURLToPath } from 'url'
 import path from 'path'
+import ora from 'ora'
+
+const spinner = ora('Creating project structure').start()
 
 // The first argument will be the project name
 const projectName = process.argv[2]
@@ -69,7 +72,7 @@ if (name) {
 }
 
 // Run `yarn install` in the project directory to install the dependencies
-console.log('Installing dependencies (may take some time) ...')
+spinner.text = 'Installing dependencies (may take some time) ...'
 const proc = spawn('yarn install', [], { stdio: ['ignore', 'ignore', 'pipe'], cwd: projectDir, shell: true })
 
 let errors: Uint8Array[] = []
@@ -78,6 +81,7 @@ proc.stderr.on('data', (data) => {
 })
 
 proc.on('close', (code) => {
+  spinner.stop()
   if (code === 0) {
     console.log(`⚡️ Your ElectricSQL app is ready at \`./${projectName}\``)
   }
@@ -85,6 +89,6 @@ proc.on('close', (code) => {
     console.error(Buffer.concat(errors).toString())
     console.log(`⚡️ Could not install project dependencies. Nevertheless the template for your app can be found at \`./${projectName}\``)
   }
-  
+
   console.log(`Navigate to your app folder \`cd ${projectName}\` and follow the instructions in the readme.`)
 })
--- a/examples/starter/src/index.ts
+++ b/examples/starter/src/index.ts
@@ -90,5 +90,5 @@ proc.on('close', (code) => {
     console.log(`⚡️ Could not install project dependencies. Nevertheless the template for your app can be found at \`./${projectName}\``)
   }
 
-  console.log(`Navigate to your app folder \`cd ${projectName}\` and follow the instructions in the readme.`)
+  console.log(`Navigate to your app folder \`cd ${projectName}\` and follow the instructions in the README.md.`)
 })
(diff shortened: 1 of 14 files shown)


examples/starter/src/index.ts
```diff
@@ -69,13 +69,22 @@ if (name) {
...
-    console.log(`Success! Your ElectricSQL app is ready at \`./${projectName}\``)
+    console.log(`⚡️ Your ElectricSQL app is ready at \`./${projectName}\``)
   }
   else {
-    console.log(`Could not install project dependencies. Nevertheless the template for your app can be found at \`./${projectName}\``)
+    console.error(Buffer.concat(errors).toString())
+    console.log(`⚡️ Could not install project dependencies. Nevertheless the template for your app can be found at \`./${projectName}\``)
   }
+  
+  console.log(`Navigate to your app folder \`cd ${projectName}\` and follow the instructions in the readme.`)
```
thruflo, line 89: Nit, suggest: In the `README.md`.

examples/starter/template/README.md
```diff
@@ -12,6 +12,8 @@ Open a new tab in your terminal. Navigate back to the same folder. Apply the mig
 ```shell
 yarn db:migrate
 ```
+The above command applies all sql files in the migrations folder so make sure they are idempotent if you run the command multiple times.
+Or, alternatively, use proper migration tooling.
```
thruflo, line 16: Aggressive ;)

I suggest we reorganise this a touch: minimise the info in the primary flow. So just skip this here. Instead, as discussed with @balegas we can add a more info section at the bottom that says "for info about X, Y and Z see the <link to running the examples> page".

The primary flow is zero friction / minimum cognitive overhead to dopamine, i.e.: get the thing running with the least complexity possible.

Moving the extra stuff to the docs means we only have to maintain it once.

examples/starter/template/db/connect.js
thruflo, line 1: For me, I'd like `yarn db:psql` -- but I'm happy to go with connect if you prefer for nice noun:verb consistency.

examples/starter/template/db/migrate.js
```diff
@@ -38,15 +22,13 @@ const apply = async (fileName) => {
...
-    const fileNames = fs.readdirSync(MIGRATIONS_DIR)
-    for (const file of fileNames) {
-      await apply(file)  
+  const fileNames = fs.readdirSync(MIGRATIONS_DIR)
+  for (const file of fileNames) {
+    if (path.extname(file) === '.sql') {
+      await apply(file)
     }
   }
+  console.log('⚡️ Database is migrated.')
```
thruflo, line 31: Suggest `Database migrated.` I think the zap is good for the end of the initial generate command.

thruflo (approved):

examples/starter/template/db/migrate.js
```diff
@@ -1,34 +1,18 @@
 const fs = require('fs')
 const path = require('path')
-const parseArgs = require('minimist')
-const shell = require('shelljs')
+const { DATABASE_URL } = require('util.js')
```
balegas, line 3:
```suggestion
const { DATABASE_URL } = require('./util.js')
```
Fixing this I was able to run commands with DATABASE_URL

balegas (approved):

examples/starter/template/package.json
```diff
@@ -7,6 +7,7 @@
     "backend:stop": "docker compose -f backend/compose/docker-compose.yaml down --volumes",
     "electric:start": "node ./backend/startElectric.js",
     "db:migrate": "node ./db/migrate.js",
+    "db:connect": "node ./db/connect.js",
```
balegas, line 10: below, match "@types/node": "^16.9.1", with node version

examples/starter/template/package.json
```diff
@@ -7,6 +7,7 @@
     "backend:stop": "docker compose -f backend/compose/docker-compose.yaml down --volumes",
```
balegas, line 7: Should we start the services as a daemon?

Not a strong opinion, but is a bit annoying that I do ```yarn backend:start``` and I lose control of that shell. Again, raising it, but not a strong opinion.

thruflo: I think it's nice to see the output.

balegas:
@kevin-dp I noticed that the generated example code is not properly type-checked.
See https://github.com/electric-sql/electric/pull/333/files#diff-4e27ade820362150f9a28493ea449906cd1cf3d591863323f133674bb9f9b3a8R9

examples/starter/template/README.md
```diff
@@ -12,6 +12,8 @@ Open a new tab in your terminal. Navigate back to the same folder. Apply the mig
 ```shell
 yarn db:migrate
```
balegas, line 13: - [ ] document ```DATABASE_URL``` in notes or elsewhere

PR #336, opened 2023-08-16
chore: add better logging to client proto requests/responses

icehaunter:
Before we JSON-logged messages that are sent/received, which made inspecting the contents quite complex, since text contents weren't decoded. This makes the logging mildly more palatable.

Before:
```
Received message {"$type":"Electric.Satellite.v1_4.SatOpLog","ops":[{"$type":"Electric.Satellite.v1_4.SatTransOp","begin":{"$type":"Electric.Satellite.v1_4.SatOpBegin","commitTimestamp":{"low":-38279406,"high":393,"unsigned":true},"transId":"","lsn":{"type":"Buffer","data":[50,53,53,53,54,50,50,52]},"origin":"c0cbe095-4a2c-4236-8d67-a8a8f6d3edfc","isMigration":false}},{"$type":"Electric.Satellite.v1_4.SatTransOp","insert":{"$type":"Electric.Satellite.v1_4.SatOpInsert","relationId":16632,"rowData":{"$type":"Electric.Satellite.v1_4.SatOpRow","nullsBitmask":{"type":"Buffer","data":[40]},"values":[{"type":"Buffer","data":[116,101,115,116,95,105,100,95,49]},{"type":"Buffer","data":[]},{"type":"Buffer","data":[]},{"type":"Buffer","data":[]},{"type":"Buffer","data":[]},{"type":"Buffer","data":[49,48]}]},"tags":["c0cbe095-4a2c-4236-8d67-a8a8f6d3edfc@1692178835218"]}},{"$type":"Electric.Satellite.v1_4.SatTransOp","commit":{"$type":"Electric.Satellite.v1_4.SatOpCommit","commitTimestamp":{"low":-38279406,"high":393,"unsigned":true},"transId":"","lsn":{"type":"Buffer","data":[50,53,53,53,54,50,50,52]}}}]}
```

After:
```
recv: #SatOpLog{ops: [#Begin{lsn: MjU1NjMxMTI=, ts: 1692201176274, migration?: false}, #Insert{for: 16632, tags: [098a99d1-fb70-442b-8aaa-3b3e0abba452@1692201176274], new: ["test_id_1", "brand new content", ∅, "", ∅, "10"]}, #Insert{for: 16641, tags: [098a99d1-fb70-442b-8aaa-3b3e0abba452@1692201176274], new: ["other_test_id_1", "", "test_id_1"]}, #Commit{lsn: MjU1NjMxMTI=}]}
```

### DIFF
--- a/clients/typescript/src/util/proto.ts
+++ b/clients/typescript/src/util/proto.ts
@@ -360,7 +360,7 @@ function opToString(op: Pb.SatTransOp): string {
   if (op.begin)
     return `#Begin{lsn: ${base64.fromBytes(
       op.begin.lsn
-    )}, ts: ${op.begin.commitTimestamp.toString()}, migration?: ${
+    )}, ts: ${op.begin.commitTimestamp.toString()}, isMigration: ${
       op.begin.isMigration
     }}`
   if (op.commit) return `#Commit{lsn: ${base64.fromBytes(op.commit.lsn)}}`
--- a/clients/typescript/src/util/proto.ts
+++ b/clients/typescript/src/util/proto.ts
@@ -2,6 +2,8 @@ import * as Pb from '../_generated/protocol/satellite'
 import * as _m0 from 'protobufjs/minimal'
 import { SatelliteError, SatelliteErrorCode } from './types'
 import { ShapeRequest } from '../satellite/shapes/types'
+import { base64, typeDecoder } from './common'
+import { getMaskBit } from './bitmaskHelpers'
 
 type GetName<T extends { $type: string }> =
   T['$type'] extends `Electric.Satellite.v1_4.${infer K}` ? K : never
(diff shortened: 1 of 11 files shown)


clients/typescript/src/util/proto.ts
```diff
@@ -269,3 +271,132 @@ export function shapeRequestToSatShapeReq(
   }
   return shapeReqs
 }
+
+export function msgToString(message: SatPbMsg): string {
+  switch (message.$type) {
+    case 'Electric.Satellite.v1_4.SatAuthReq':
```
alco, line 277: What's going to happen when we bump the protocol version? Seems like this function will start returning `undefined` and someone will have to find-and-replace all of these by hand. Can we make it protocol-version-agnostic?

clients/typescript/src/util/proto.ts
```diff
@@ -269,3 +271,132 @@ export function shapeRequestToSatShapeReq(
...
+    case 'Electric.Satellite.v1_4.SatOpLog':
+      return `#SatOpLog{ops: [${message.ops.map(opToString).join(', ')}]}`
+  }
+}
+
+function opToString(op: Pb.SatTransOp): string {
+  if (op.begin)
+    return `#Begin{lsn: ${base64.fromBytes(
+      op.begin.lsn
+    )}, ts: ${op.begin.commitTimestamp.toString()}, migration?: ${
```
alco, line 363: I'm a huge fan of the general direction you've taken to make these log messages more concise and with less punctuation noise. But I'm against renaming things because that adds to the cognitive load of us--mere mortals--and makes project-wide source code search less precise.

What's your rationale for changing the  `isMigration` field's name here?

alco (commented):
Loving all the red diff lines in e2e tests :heart:

clients/typescript/src/util/proto.ts
balegas: if hard to make proto agnostic, add a test that forces us to update this function on a protocol bump

clients/typescript/src/util/proto.ts
```diff
@@ -269,3 +271,132 @@ export function shapeRequestToSatShapeReq(
   }
   return shapeReqs
 }
+
+export function msgToString(message: SatPbMsg): string {
```
balegas, line 275: I'd like this to have a more structured representation, using a map or something, that makes it more readable than going through the switch structure.

balegas (approved):
Thanks for addressing this, which will allow us to have much more compact and readable logging on the client.

I'd like this PR to have a pass in order to make it more 'generic', but I'm happy to move forward and give it another go later.

clients/typescript/src/util/proto.ts
icehaunter: I wouldn't, to be honest. This is essentially a debug function. I explicitly don't want to make it generic, since we have a very much finite set of messages on the protocol, and debug representation can be just that. I really don't want to encourage using this function for anything more than debug logging. If we need a better intermediate representation of messages - sure, at some point, maybe. I'd rather not over-complicate this function for no reason.

clients/typescript/src/util/proto.ts
icehaunter: No reason, just prefer elixir-style question mark suffixing. I've changed it to `isMigration`.

clients/typescript/src/util/proto.ts
icehaunter: This case will break types on protocol version update. TS is smart enough to check for case exhaustiveness, and any new messages or changes to exisiting ones will be immediately visible, so I don't think that's a problem

alco (approved):

PR #340, opened 2023-08-21
feat: enable compensations

icehaunter:
Compensations are stored as a special-form updates that don't specify previous row values. Previously, such updates were treated as inserts by the server, but now they are converted to an update that doesn't mark any columns as modified. This allows PG to just use the tag for reviving the row as needed.

This is covered by an e2e test.

Closes VAX-801

---
Review notes: ignore commit messages :D


clients/typescript/test/satellite/merge.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      oldRow: JSON.stringify({ id: 1, value: 'TEST' }),
+      clearTags: JSON.stringify(['common@100000000']),
+    },
+  ]
+
+  const merged = mergeEntries('local', local, 'remote', remote)
+
+  // Merge should resolve into the UPSERT for this row, since the remote DELETE didn't observe this local update
+  t.like(merged, { 'main.public': { [pk]: { optype: 'UPSERT' } } })
+  t.deepEqual(merged['main.public'][pk].tags, ['local@100001000'])
```
icehaunter, line 39: @balegas, it took me some time to untangle the merge code, and I'm still unsure about the logic. This test passes, as in matches what's happening. Is this what we intend to happen? The shadow row tags to match this local insert, as "latest" applied operation, since it should also win on the server, right?

balegas: This test implies something that we should clarify: in this exact example the resulting row shall be `{id:1, value: 'TEST'}`.

Why:
- The local row has some `value` which has been created before arriving at this scenario
- The local change did not modify that value
- The incoming operation is a delete, but also is sending a new `value`. 
- The timestamp for the incoming operation is higher than any operation in the oplog, therefore all it's modified values shall be applied to the local row

From looking into prop `changes` in `merged`, I interpret that we will not take value `TEST`
```
{"main.public":{"{\"id\":1}":{"namespace":"main","tablename":"public","primaryKeyCols":{"id":1},"optype":"UPSERT","changes":{"id":{"value":1,"timestamp":100001000}},"fullRow":{"id":1,"value":"TEST"},"tags":["local@100001000"]}}} 
```

Why taking column values if incoming op is a `DELETE`?
- The semantics is that on `INSERT` || `DELETE`, we still do LWW for the row columns

Is it possible to have a `DELETE` with modified columns?
- In theory yes: a sequence of `UPDATE` ... `DELETE` can be squashed into a single `DELETE`
- In practice, if we don't squash operations, then the `UPDATE` would always precede the `DELETE`, which would guarantee that `local` would already have applied value `TEST` by the time it receives the `DELETE`

My take: I think we should assume that `DELETE` can include column changes and follow LWW, but if we have absolute guarantee that currently we preserve all statements and shadow ops in a transaction we can move ahead. I am flagging this mainly because I am not sure what we actually do after the trigger implementation in server, but you shallbe able to confirm this.

- note (or double-check) that the server should produce the same behaviour!

balegas (commented):
Sharing this comment while I continue with review

clients/typescript/test/satellite/merge.test.ts
icehaunter: First of all, yeah, I can add a clause for this test to show that the resulting row is indeed `value: test`. It's true, I've observed it under `console.log` output while debugging this test. I gathered that this is what's expected.

icehaunter: Second, server behaves mildly differently, in that for a given DELETE a modified columns bitmask is forced to all false values regardless of DELETE contents. I think it makes sense of the server, since client's delete may not have observed all updates, and if it hasn't then it will just be resolved to a no-op under add-wins semantics.

icehaunter: Generally the server will always send any operations changing the columns prior to sending the DELETE, so taking or not taking values from a DELETE should not matter

balegas (approved):
Checked this with integrity example, skimmed through the changes.
@icehaunter would be good to go over the merge function together. Let's try to do that this week.

PR #344, opened 2023-08-21
fix(satellite): try reconnecting when server is gone

balegas:
Before we were moving to connectivity state 'disconnected' when an error occurred on server. We changed that to move to 'available' and fixed issues with reconnection path so that client can recover on intermittent Electric errors.

A follow up to this task is VAX-835

- [x] need to add changelog


clients/typescript/src/satellite/client.ts
```diff
@@ -241,63 +258,67 @@ export class SatelliteClient extends EventEmitter implements Client {
...
         })
-        resolve()
-      })
 
-      this.socket.onceError((error) => {
-        this.socket = undefined
-        this.notifier.connectivityStateChanged(this.dbName, 'disconnected')
-        reject(error)
-      })
+        this.socket.onceError((error) => {
```
icehaunter, line 295: This listener is best removed in the `onceConnect` to not call it on the first error

icehaunter: Maybe that was the reason for weird behaviour? unexpected `onceError` call?

clients/typescript/test/satellite/client.test.ts
```diff
@@ -92,7 +92,8 @@ test.serial('connection backoff success', async (t) => {
   }
 
   try {
-    await client.connect(retry)
+    client['connectionRetryHandler'] = retry
+    await client.connect()
```
icehaunter, line 96: You should add `t.plan(1)` before this to make sure the `t.pass()` in the retry function is called the correct amount of times

clients/typescript/src/sockets/node.ts
```diff
@@ -44,14 +55,7 @@ export class WebSocketNode extends EventEmitter implements Socket {
...
   onError(cb: (error: Error) => void): void {
-    this.on('error', () =>
-      cb(
-        new SatelliteError(
-          SatelliteErrorCode.CONNECTION_FAILED,
-          'failed to establish connection'
-        )
-      )
-    )
+    this.on('error', cb)
```
balegas, line 58: this change (and lines 28-36) was required to be able to remove the passed cb

clients/typescript/src/satellite/client.ts
```diff
@@ -1089,13 +1098,13 @@ export class SatelliteClient extends EventEmitter implements Client {
         return reject(error)
       }
 
-      this.once('error', errorListener)
+      this.once('rpc_error', errorListener)
```
balegas, line 1101: separate 'error' from 'rpc_error'

clients/typescript/src/sockets/react-native.ts
```diff
@@ -30,14 +25,14 @@ export class WebSocketReactNative implements Socket {
       }
     }
 
-    this.socket.onerror = (event: any) => {
```
balegas, line 33: this was wrong. popping error callbacks everytime.

clients/typescript/src/sockets/react-native.ts
```diff
@@ -91,6 +86,18 @@ export class WebSocketReactNative implements Socket {
   }
 
   onceError(cb: (error: Error) => void): void {
-    this.errorCallbacks.push(cb)
+    this.onceErrorCallbacks.push(cb)
+  }
+
+  removeErrorListener(cb: (error: Error) => void): void {
```
balegas, line 92: made generic implementation for remove listener

clients/typescript/src/sockets/web.ts
```diff
@@ -30,25 +26,24 @@ export class WebSocketWeb implements Socket {
 
     // event doesn't provide much
     this.socket.addEventListener('error', () => {
-      while (this.errorCallbacks.length > 0) {
```
balegas, line 33: same as above

PR #346, opened 2023-08-22
fix(electric): Fix pattern matching in the recently added e2e test

alco:


alco:
@icehaunter I have fixed the pattern matching problem. But the actual compensation mechanism is now failing because it doesn't pass the [recently added][1] null check.

[1]: https://github.com/electric-sql/electric/blob/7a38cfefd3f2fb9be14b7d12dc48bd413d2980dc/components/electric/lib/electric/satellite/serialization.ex#L379

The compensation mechanism produces this message that the client tries to sync to the server:
```
%Electric.Satellite.V14.SatOpUpdate{
  relation_id: 2,
  row_data: %Electric.Satellite.V14.SatOpRow{
    nulls_bitmask: "|",
    values: ["test_id_1", "", "", "", "", ""],
    __uf__: []
  },
  old_row_data: nil,
  tags: ["1bee507a-9a72-4f96-92ae-78d3187708d8@1692695602923"],
  __uf__: []
}
```
The `nulls_bitmask` looks as follows in binary: `{0, 1, 1, 1, 1, 1, 0, 0}`, i.e. the value for the 2nd column `content` in `row_data` is null but the column itself has the `not null` constraint which fails to validate [here][2].

[2]: https://github.com/electric-sql/electric/blob/7a38cfefd3f2fb9be14b7d12dc48bd413d2980dc/components/electric/lib/electric/satellite/serialization.ex#L422-L424

icehaunter:
@alco can we special-case `SatOpUpdate` so if there is `old_row_data` filled, it's validated, and if not, only primary keys are validated? PG hooks will use only PKs in that case (we're marking all columns as not modified, hence their values unused), so we can safely drop/ignore non-pk columns for updates where `old_row_data: nil`?

icehaunter (approved):

PR #347, opened 2023-08-22
docs: Include new SSL-mode env var in deployment docs

magnetised:


alco (approved):

docs/api/service.md
```diff
@@ -34,6 +34,7 @@ Everything in the table below that doesn't have a default value is required to r
...
 | `DATABASE_URL`                                                  | PostgreSQL connection URL for the database.                                                                                                                                                                                                                                                                                                                                       |
+| `DATABASE_REQUIRE_SSL`<p>&nbsp;&nbsp;(`false`)</p>              | Set to `yes` or `true` to require SSL for the connection to the database. Note that you can always configure SSL for the connection by adding `sslmode=require` to [the `DATABASE_URL` parameters](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS).                                                                                               |
```
thruflo, line 37: Nice. Just wonder if the wording could be clearer on the "or use sslmode=require" -- i.e.: is it clear that you can use either method and that if you use one you don't need to use the other? Also what happens when they disagree, i.e.: you set `DATABASE_REQUIRE_SSL=true` and `sslmode=never`?

thruflo (approved):

docs/api/service.md
```diff
@@ -34,7 +34,7 @@ Everything in the table below that doesn't have a default value is required to r
...
 | `DATABASE_URL`                                                  | PostgreSQL connection URL for the database.                                                                                                                                                                                                                                                                                                                                       |
-| `DATABASE_REQUIRE_SSL`<p>&nbsp;&nbsp;(`false`)</p>              | Set to `yes` or `true` to require SSL for the connection to the database. Note that you can always configure SSL for the connection by adding `sslmode=require` to [the `DATABASE_URL` parameters](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS).                                                                                               |
```
magnetised: @thruflo does this address your concerns?

thruflo (approved):
👍

PR #349, opened 2023-08-22
feat(electric): Add support for DATE data type in electrified tables

alco:
Closes VAX-855.

### DIFF
--- a/components/electric/lib/electric/satellite/serialization.ex
+++ b/components/electric/lib/electric/satellite/serialization.ex
@@ -24,8 +24,9 @@ defmodule Electric.Satellite.Serialization do
   @spec supported_pg_types :: [atom]
   def supported_pg_types do
     ~w[
-      int2 int4 int8
+      date
       float8
+      int2 int4 int8
       text
       timestamp timestamptz
       varchar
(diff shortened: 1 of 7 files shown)


components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -479,6 +480,18 @@ defmodule Electric.Satellite.Serialization do
     val
   end
 
+  def decode_column_value(<<y1, y2, y3, y4, ?-, m1, m2, ?-, d1, d2>> = val, :date)
+      when y1 in ?0..?9 and y2 in ?0..?9 and y3 in ?0..?9 and y4 in ?0..?9 and
+             ((m1 - ?0) * 10 + (m2 - ?0)) in 1..12 and ((d1 - ?0) * 10 + (d2 - ?0)) in 1..31 do
+    _ = Date.from_iso8601!(val)
```
icehaunter, line 486: Why can't we just use this function and not do the char-by-char match?

icehaunter (approved):
Just a small question, otherwise LGTM

components/electric/lib/electric/satellite/serialization.ex
alco: That's because `Date` is more permissive than Postgres' input validation. For example, `Date` can take negative dates or dates with year 0. Neither is supported in SQLite or Postgres. But it's still useful to rely on `Date.from_iso8601!` to check for correct month number and number of days in a month, including leap years, as opposed to reimplementing the whole thing ourselves.

I've explained this in a newer commit in the timestamp PR - https://github.com/electric-sql/electric/pull/330/files#diff-11ee61155c8d413582313d21317149ae3fc66074cfbf4d345f293152f179f40fR530-R541.

I'm also going to push an adjustment to this PR that will clean up the pattern matching to improve the readibility.

PR #353, opened 2023-08-23
client: revise auth token functions.

thruflo:
Previously we had an `authToken` function exported from `electric-sql/auth` that:

1. was a mock/test function with mock data in it
2. used in the examples
3. importing the Jose library

Plus in the examples, we had separate code copied around used to generate insecure tokens.

The PR:

1. provides user-usable `secureAuthToken(claims, iss, key)` and `insecureAuthToken(claims)` functions
2. renames the authToken function to `mockSecureAuthToken`
3. updates the docs and examples accordingly -- examples *should* use the insecure function. The backend stack for the examples is configured in insecure mode and this is what it is for.
4. moves the secure functions to an optional import path, so that we remove the Jose JWT signing library from users' JS builds if they are not using it (generally they won't be -- tokens will come from the server side)


clients/typescript/package.json
```diff
@@ -139,6 +139,7 @@
     "**/*": "prettier --write --ignore-unknown"
   },
   "dependencies": {
+    "Base64": "^1.2.0",
```
icehaunter, line 142: Why? `btoa` exists literally everywhere since IE9 and node 16. We don't support either.
 https://caniuse.com/?search=btoa

icehaunter (changes requested):

clients/typescript/package.json
thruflo: It’s not in NodeJS.

icehaunter: https://nodejs.org/docs/latest-v16.x/api/globals.html#btoadata

thruflo: Node has Buffer. Can feature detect and use one or the other. But this is 700 bytes and just works https://www.npmjs.com/package/Base64

icehaunter: Every node since [16.0](https://nodejs.org/docs/latest-v20.x/api/globals.html#btoadata) has a global `btoa`, as does every browser since IE9. Open a fresh node console and run `btoa('test')` - it also just works. I just don't see a reason to add a dependency however small for something consistently available in the standard library on all supported platforms

thruflo: Right, ok, so btoa is available everywhere. I’ll remove dep 👍

clients/typescript/src/auth/mock.ts
```diff
@@ -0,0 +1,11 @@
+import { secureAuthToken } from './util'
+
+export function mockSecureAuthToken(
+  iss?: string,
+  key?: string
+): Promise<string> {
+  const mockIss = iss ?? 'dev.electric-sql.com'
+  const mockKey = key ?? 'integration-tests-signing-key-example'
+
+  return secureAuthToken({ user_id: 'test-user' }, mockIss, mockKey)
```
alco: I know we've been lax about this, but we probably should be using a UUID for `user_id` in our examples and docs. Once we have UUID support in main and documented, we'll likely start using it in our examples for `user_id` columns.

clients/typescript/src/declarations.d.ts
```diff
@@ -1 +1,2 @@
+declare module 'Base64'
```
alco, line 1:
```suggestion
```

alco (approved):
This is brilliant! :clap:

thruflo:
Question: I've updated the examples to use the new function API within this PR. Whereas in the tabId PRs I've seperated.

Should I revert the changes to the examples (and docs?!) and extract to a separate PR? Or can we stage manage merging this as part of a release so things match up (quickly)?

I'm aware that this bears on the published version vs workspace:* discussion. However assuming we are sticking with pinned versions in the examples, what's the best development methodology here? It's a bit tricky either way I think.

clients/typescript/src/auth/mock.ts
thruflo: Yup, I suggest let's do a pass of updating examples to use UUIDs when we switch the examples to use UUID columns.

thruflo:
What about docs I wonder. For example, if we somehow connect the website submodule update to the release process.

N.b.: the starter template needs the pinned version in the deps. But that's the *published* starter template. So it's OK to update code prior to release.

With the other examples, it occurs to me that we could possibly consider publishing to a git repo? So i.e.: on release we update the code in an examples repo?

It's pretty fiddly to keep track of docs and example changes in a separate PR and know when it's OK to merge.

icehaunter:
Yeah, "freezing" a version into a separate repo is definitely a possibility

alco:
Another option is to have a dedicated Git branch for each release and use it in URLs that are referring to example code. So, when a new release is published, a matching branch is created for it which then stays unchanged unless there are hotfixes that need to be applied to the code. Development continues in `main` where it's OK to break compatibility between example code and _published_ package versions.

For a usage example of this strategy in the wild look at Elixir - https://github.com/elixir-lang/elixir/branches

PR #356, opened 2023-08-23
fix: malformed psql command in docs

icehaunter:
Before:
```
psql "postgresql://generated_user:generated_passwd@core_user:core_password@postgres:54321/electric"
```

After:
```
psql "postgresql://generated_user:generated_passwd@postgres:54321/electric"
```


thruflo (approved):
Good catch, thanks 🙏

PR #357, opened 2023-08-23
fix: client reconnection didn't work properly on server going away

balegas:
This turned into a big refactoring with great impact on re-connection strategies and error handling. I can do a walkthrough on the most impacting changes.

- [ ] Should be ran against website examples by someone else
- [ ] push changeset


clients/typescript/src/sockets/node.ts
```diff
@@ -16,23 +16,35 @@ export class WebSocketNode extends EventEmitter implements Socket {
...
+      )
+    }
+
     this.socket = new WebSocket(opts.url)
     this.socket.binaryType = 'nodebuffer'
 
     this.socket.on('open', () => this.emit('open'))
     this.socket.on('message', (data) => this.emit('message', data))
-    this.socket.on('error', (error) => this.emit('error', error))
+    this.socket.on('error', (_unusedError) =>
```
icehaunter, line 32: Let's add a `Log.error` here, just so we have some visibility over the socker closing?

clients/typescript/src/satellite/process.ts
```diff
@@ -404,6 +433,10 @@ export class SatelliteProcess implements Satellite {
     // this could especially happen in unit tests
     this.subscriptionNotifiers[subId] = emptyPromise()
 
+    if (this.initializing) {
+      await this.initializing.promise
+    }
```
icehaunter, line 438:
```suggestion
    await this.initializing?.promise
```

Awaiting undefined is a no-op, so we can do it like that.

clients/typescript/src/satellite/process.ts
```diff
@@ -628,8 +685,36 @@ export class SatelliteProcess implements Satellite {
...
+      )
+
+      this.initializing?.reject(error)
+      this.notifier.connectivityStateChanged(this.dbName, 'disconnected')
+      throw error
+    })
+  }
+
+  // NO DIRECT CALLS TO CONNECT
+  async _connect(): Promise<void> {
```
icehaunter, line 716: Then make it private :D


```suggestion
  private async _connect(): Promise<void> {
```

clients/typescript/src/satellite/process.ts
```diff
@@ -638,8 +723,19 @@ export class SatelliteProcess implements Satellite {
 
     try {
       await this.client.connect()
-      await this.client.authenticate(authState)
+      const authResp = await this.client.authenticate(authState)
+
+      if (authResp.error) {
+        throw authResp.error
+      }
```
icehaunter, line 730: nitpick: and this is why `authenticate` should also throw :P

clients/typescript/src/satellite/process.ts
```diff
@@ -404,6 +433,10 @@ export class SatelliteProcess implements Satellite {
     // this could especially happen in unit tests
     this.subscriptionNotifiers[subId] = emptyPromise()
 
+    if (this.initializing) {
+      await this.initializing.promise
```
icehaunter, line 437: I may be wrong, but my gut feeling is that this may need some care in that in the sequence of events
1. `await start()`
2. `await subscribe()`

If `initializing` is rejected, the requested subscription gets stuck because it wasn't cleaned up with `this.subscriptions.subscriptionCancelled`

Another point of interest: is it possible to call this method before `start` was called or had a chance to reach `_connectWithBackoff`? Probably not if `electrify` was awaited, right? Just thinking of actual use in React apps.

clients/typescript/src/satellite/process.ts
```diff
@@ -96,8 +100,23 @@ type MetaEntries = {
...
 }
 
+type ConnectRetryHandler = (error: Error, attempt: number) => boolean
+
+const connectRetryHandler: ConnectRetryHandler = (_error, _attempts) => {
+  // for simplicity, always retry
+  return true
+}
+
+const connectionRetryPolicy: Partial<IBackOffOptions> = {
```
icehaunter, line 110: nitpick: We may want to pull this into a class property just so it can be overridden in tests easily

icehaunter (commented):
I really like the approach you've taken. Code feels quite a bit cleaner. Big props! I have some questions, but nothing too critical I think

icehaunter (changes requested):
I really like the approach you've taken. Code feels quite a bit cleaner. Big props! I have some questions, but nothing too critical I think

clients/typescript/src/sockets/node.ts
balegas: error will always trigger logging on the process, but a process might no longer be listening. So that is a good idea.

added logging on the client listener.

clients/typescript/src/satellite/process.ts
balegas: sure :). I had to remove a bunch of internal calls to connect too.

clients/typescript/src/satellite/process.ts
balegas: the initializing is horrible. Specially the double-throw. We need to fix this.

I think this remains a source of bugs. added to triage.

clients/typescript/src/sockets/web.ts
```diff
@@ -2,77 +2,124 @@ import { ConnectionOptions, Data, Socket, SocketFactory } from '.'
...
-  private errorCallbacks: ((error: Error) => void)[]
+  private connectCallbacks: (() => void)[] = []
+  private errorCallbacks: ((error: SatelliteError) => void)[] = []
+  private onceErrorCallbacks: ((error: SatelliteError) => void)[] = []
 
-  constructor() {
-    this.connectCallbacks = []
-    this.errorCallbacks = []
+  // event doesn't provide much
+  private errorListener = () => {
```
icehaunter, line 19: This likely should be a method and not a property if this is never reassigned

clients/typescript/src/sockets/web.ts
```diff
@@ -2,77 +2,124 @@ import { ConnectionOptions, Data, Socket, SocketFactory } from '.'
...
+      const cb = this.onceErrorCallbacks.pop()
+      if (cb)
+        cb(new SatelliteError(SatelliteErrorCode.SOCKET_ERROR, 'socket error'))
+    }
+  }
+
+  private connectListener = () => {
+    while (this.connectCallbacks.length > 0) {
+      this.connectCallbacks.pop()!()
+    }
```
icehaunter, line 34: This can be written slightly more error-resistant without the `!`
```suggestion
    let callback: (() => void) | undefined
    while (callback = this.connectCallbacks.pop()) {
      callback()
    }
```

clients/typescript/src/util/common.ts
```diff
@@ -69,3 +70,37 @@ export function emptyPromise<T = void>() {
...
+
+export type Waiter = {
+  waitOn: () => Promise<void>
+  resolve: () => Promise<void>
+  reject: (error: SatelliteError) => Promise<void>
+  finished: boolean
+}
+
+export function getWaiter(): Waiter {
+  {
```
icehaunter, line 82: Is this extra block level intentional?..

clients/typescript/src/util/common.ts
```diff
@@ -69,3 +70,37 @@ export function emptyPromise<T = void>() {
...
+        finished = true
+        resolve()
+      },
+
+      reject: async (error) => {
+        finished = true
+        waiting ? reject(error) : resolve()
+      },
+
+      finished,
```
icehaunter, line 103: I don't think this works? Will the object store a reference to closed variable here? I may be wrong though

clients/typescript/src/satellite/client.ts
```diff
@@ -316,15 +297,7 @@ export class SatelliteClient extends EventEmitter implements Client {
...
-    this.initializing?.reject(
-      new SatelliteError(
-        SatelliteErrorCode.INTERNAL,
-        'Socket is closed by the client while initializing'
-      )
-    )
-    this.initializing = undefined
+
     if (this.socket != undefined) {
       this.socket!.closeAndRemoveListeners()
```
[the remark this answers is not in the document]
balegas: let's add that lint rule when a good moment (and that other one to avoid testing conditions on fuction definitions :D)

clients/typescript/src/satellite/process.ts
```diff
@@ -554,27 +582,44 @@ export class SatelliteProcess implements Satellite {
...
 
-    if (shapeDefs) {
+    if (shapeDefs.length > 0) {
       this.subscribe(shapeDefs)
     }
   }
 
+  async _resetClientStateAndReconnect() {
+    return this._resetClientState().then(this._reconnect.bind(this))
+  }
```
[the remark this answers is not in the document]
balegas: yeah, should have not mixed approaches

clients/typescript/src/satellite/process.ts
```diff
@@ -1169,6 +1295,11 @@ export class SatelliteProcess implements Satellite {
     await this.notifyChangesAndGCopLog(opLogEntries, origin, commitTimestamp)
   }
 
+  _handleAck: AckCallback = async (lsn: LSN, type: AckType) => {
```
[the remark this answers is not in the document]
balegas: out of place

clients/typescript/src/util/common.ts
balegas: oh, this was silly.... All the bug fixing without proper tests and just confirming our own intuition is recipe for disaster.

icehaunter (approved):
Looks good, great work!

PR #358, opened 2023-08-24
util: add a uniqueTabId function.

thruflo:
Returns a unique ID for each browser tab. This can be used to construct a tab scoped DB name. This allows data to sync between tabs by using one named SQLite database file per tab. This is sub-optimal compared with sharing a database but avoids concurrent database access.

Handles duplicate tabs with some potential for false positives. False positives result in a new tabID which means a new database name and thus additional data transfer and storage.

Uses window.sessionStorage. Some browsers disable access to sessionStorage even when saying its available (e.g.: as a result of disabling third party cookies o_O). So we handle this by falling back to an in-memory tabId singleton -- which means each page load syncs data into a new DB.

Other platforms can pass in a storage implementation or default to the in-memory tabId singleton.


clients/typescript/src/util/tab.ts
```diff
@@ -0,0 +1,120 @@
...
+// False positives result in a new tabID which means a new database name
+// and thus additional data transfer and storage.
+//
+// Uses window.sessionStorage. Some browsers disable access to sessionStorage
+// even when saying its available (e.g.: as a result of disabling third party
+// cookies o_O). So we handle this by falling back to an in-memory tabId
+// singleton -- which means each page load syncs data into a new DB.
+//
+// Other platforms can pass in a storage implementation or default to the
+// in-memory tabId singleton.
```
icehaunter, line 46: This should be a docstring so it shows up as a tooltip when used.
```typescript
/**
 * Returns a unique ID for each browser tab.
 *
 * This can be used to construct tab scoped DB name. 
 */
export const ...

File: clients/typescript/src/util/tab.ts [hunk 2]
```diff
@@ -0,0 +1,120 @@
...
+export const uniqueTabId = (opts: Options = {}): ReturnValue => {
+  const key = opts.key ?? defaultKey
+
+  // Use sessionStorage as the default storage if available.
+  let defaultStorage: Storage
+  try {
+    defaultStorage = window.sessionStorage
```
icehaunter, line 53: Could you link me to a reference that shows that this assignment would immediately throw on access like that? Sources I could find try setting-deleting a key to detect support. If this doesn't throw, then it's just assigning `undefined` to the property

clients/typescript/src/util/tab.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+  const storage: Storage = opts.storage ?? defaultStorage
+
+  let defaultNavEntries: NavEntry[]
+  try {
+    defaultNavEntries = window.performance.getEntriesByType('navigation')
+  } catch {
+    defaultNavEntries = []
+  }
+  const navEntries = opts.navEntries ?? defaultNavEntries
```
icehaunter, line 69: Can't we build defaults only if the option isn't explicitly provided? Seems a tiny bit wasteful to try accessing SessionStograge, or nav entries if they get immediately discarded

clients/typescript/src/util/tab.ts
```diff
@@ -0,0 +1,120 @@
...
+    return {
+      tabId: existingTabId as string,
+      hasExisting: true,
+      usingExisting: true,
+    }
+  }
+
+  // If it's not a duplicate tab then use it.
+  const navEntry =
+    navEntries && navEntries.length ? navEntries[0] : { type: null }
```
icehaunter, line 94:
```suggestion
  const navEntry = navEntries[0] ?? { type: null }
```

Maybe a comment as to why we're only interested in the first element?

icehaunter (changes requested):
Generally looks good to me, just a couple of nitpicks. The main one being a request to use a proper docstring

clients/typescript/src/util/tab.ts
thruflo:
```
➜  electric git:(thruflo/tab-id) node
Welcome to Node.js v18.17.1.
Type ".help" for more information.
> const foo = window.sessionStorage
Uncaught ReferenceError: window is not defined
>
```

icehaunter: Ah, you're guarding against `window`, not against `sessionStorage` not being available. Ok, makes sense.  and then `try-catch` guards on `setItem`/`getItem`.

thruflo:
Addressed the review comments.

icehaunter (approved):

PR #364, opened 2023-08-24
fix(client): incorrect truthy value

davidmartos96:
This if statement is actually checking if the function exists, which is true always.

It might be beneficial to add this Typescript lint, it would have caught it:
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unnecessary-condition.md


icehaunter:
Good point on the lint, we'll add quite some rules in that next iteration! I'll close this PR, since this change is also made in #357 which generally does quite a bit of work to validate and rework re-connection behaviours

PR #365, opened 2023-08-24
fix(client): toMessage call incorrectly used as union

davidmartos96:
Now `toMessage` no longer returns an union type, it throws, so the `messageOrError instanceof SatelliteError` logic would not be properly checked.
The Log.debug condition can be moved to the try/catch so that both success and errors can be handled correctly.


icehaunter:
Thanks! Great catch!

PR #370, opened 2023-08-24
fix(client): Remove .only in test

davidmartos96:


icehaunter (approved):

icehaunter:
Thanks! I'm bad at catching these after a debugging session.

davidmartos96:
@icehaunter This lint rule could be useful. I'm not sure if it's compatible with ava though.
https://www.npmjs.com/package/eslint-plugin-no-only-tests

icehaunter:
Yeah, good should, I'll look into that!

PR #374, opened 2023-08-25
examples: add expo and react native examples.

thruflo:
Port over updated `expo` and `react-native` examples from the legacy https://github.com/electric-sql/examples repo (see https://github.com/electric-sql/examples/pull/24).

I've tested successfully on iOS. Currently there is an issue with Android support due I think to the way we're using the driver, as per https://linear.app/electric-sql/issue/VAX-893/android-support-current-broken

Both apps now use the latest v0.5 and have been updated to match the web styling. React Native was also updated to v0.72.4, as per https://react-native-community.github.io/upgrade-helper/?from=0.70.1&to=0.72.4

## Expo

<a href="https://github.com/electric-sql/electric/assets/60015/b134ac5c-9bb6-466f-bb0b-c6865202a063">expo-grab</a>

## React Native

<a href="https://github.com/electric-sql/electric/assets/60015/bb641d4d-bc72-4357-b612-15d1d2bd2017">rn-grab</a>


alco:
Tried following Expo example README up to but excluding starting the iOS simulator (I'm on Linux). Everything worked as expected. The only weird thing was when I hit Ctrl-C in the terminal where I had the local stack running via `yarn backend:start`, the output was:
```
```
with no terminal prompt following it. But when I hit Enter, I saw my terminal prompt and nothing leftover in `docker ps` output, as expected.

alco:
More specifically, I tested the following commands:
* `yarn backend.start`
* `yarn db:psql`
* `yarn db:migrate`
* `yarn client:generate`
* `yarn client:watch`

Just did the same for the RN example, can confirm it's working too.

There was a difference in the output from `yarn db:migrate`. When I executed it in the RN example, this warning got printed:
```
bigIntMode currently defaults to "number" but will default to "bigint" in the next major version of @databases/pg. Set it explicitly to disable this warning.
Applying /home/alco/tmp/electric/examples/react-native/db/migrations/create_items_table.sql
```
It didn't interfere with the migration process which finished successfully.

Saw the same behaviour as with the Expo example when I hit Ctrl-C in the terminal window that had `yarn backend:start` running.

alco:
There's one thing I forgot to mention about the Expo example: when I ran `yarn` in it, I was presented with a choice of a version for the `@types/react-native` package. I hit Enter and it proceeded fine from there.
Screenshot from 2023-08-28 13-03-27

I didn't get that when I was testing the RN example.

thruflo:
Yup, I get the same output / experience as you.

examples/react-native/backend/compose/.envrc
```diff
@@ -0,0 +1,4 @@
+# Docker images for the local stack
+export POSTGRESQL_IMAGE=postgres:14-alpine
+export ELECTRIC_IMAGE=electricsql/electric:latest
+export APP_NAME=web_wa_sqlite
```
alco, line 4: Wrong APP_NAME. This var is used as the database name in docker-compose.yaml.

examples/react-native/backend/compose/docker-compose.yaml
```diff
@@ -0,0 +1,52 @@
...
+    volumes:
+      - pg_data:/var/lib/postgresql/data
+
+  electric:
+    image: "${ELECTRIC_IMAGE:-electricsql/electric:latest}"
+    init: true
+    environment:
+      DATABASE_URL: postgresql://postgres:password@postgres:5432/${APP_NAME:-electric}
+      LOGICAL_PUBLISHER_HOST: electric
+      # Currently published version (electric-sql@0.5.0-next.4)
```
alco, line 39: This is no longer relevant.

examples/react-native/db/migrations/create_items_table.sql
```diff
@@ -0,0 +1,8 @@
+-- Create a simple items table.
```
alco, line 1: I think this file needs a comment explaining that its contents will be sent to Postgres. It may be difficult for newcomers to wrap their head around our starter template setup where you execute `yarn` commands to control the backend and apply Postgres migrations. Every little hint helps.

examples/expo/backend/compose/.envrc
```diff
@@ -0,0 +1,4 @@
+# Docker images for the local stack
+export POSTGRESQL_IMAGE=postgres:14-alpine
+export ELECTRIC_IMAGE=electricsql/electric:latest
+export APP_NAME=web_wa_sqlite
```
alco, line 4: Wrong app name.

examples/expo/backend/compose/docker-compose.yaml
```diff
@@ -0,0 +1,52 @@
...
+    volumes:
+      - pg_data:/var/lib/postgresql/data
+
+  electric:
+    image: "${ELECTRIC_IMAGE:-electricsql/electric:latest}"
+    init: true
+    environment:
+      DATABASE_URL: postgresql://postgres:password@postgres:5432/${APP_NAME:-electric}
+      LOGICAL_PUBLISHER_HOST: electric
+      # Currently published version (electric-sql@0.5.0-next.4)
```
alco, line 39: Same as RN.

examples/expo/db/migrations/create_items_table.sql
```diff
@@ -0,0 +1,8 @@
+-- Create a simple items table.
```
alco, line 1: Same as RN.

alco (approved):
Great work!

examples/react-native/backend/compose/.envrc
thruflo: Good catch. I'm going to set all the basic items examples to an app name of `electric`. There's no reason why they need different ones. It means that the scripts will work across them if trying to e.g.: run both expo and web against the same backend services, etc.

alco: May I suggest to use the name `electric_examples` instead? We're already using the name `electric` for internal tables, I think we should avoid calling anything aimed to being modified by users `electric`.

There's a downside to reusing the same DB name for all examples: they have conflicting migrations. E.g. both Expo and RN examples create `items` table, so only the one that is run first will succeed.

thruflo: We can use `basic` for these basic items demos. The pro of same DB name is you can sync between them. The con is the migrations clash.

alco: I think using the same DB for multiple examples is a more advanced use case than setting up each example as a standalone app.

We expect users to follow the instructions in the README and those have been fine-tuned to provide everything out of the box for every example independently of users. It wasn't a pleasant experience for me when I stopped the backend for the Expo app and then started looking into the RN app only to see a conflict error when I tried to apply its migrations. That happened because the DB volume was preserved from the Expo example.

IMO it's important to avoid such clashes. More advanced users should be able to run the local stack and then connect example apps to it instead of starting a separate backend for each example. But that should not negatively affect the out-of-the-box experience of each individual example.

thruflo: Fair enough, I agree. Have reverted 👍

PR #375, opened 2023-08-28
fix: make sure that if slot server dies, it can continue from a known point

icehaunter:
Original issue is that when PG cuts the TCP connection that is sourcing the replication from Electric, the process structure wasn't implemented in a way that allowed correct continuation of the replication stream to PG without losing operations by the clients.

1. I've moved in `OffsetStorage.save_pg_position` into the loop because the loop crashes on a closed connection when trying to call the `send()` function that matches on `:ok`
2. When PG reconnects, it will tell us the latest LSN it persisted - we use the saved mapping to restore that position in the cache and now we can continue from there.
3. If all `send()` calls have succeeded we notify our producer that the Satellite-sent txs can be garbage collected

There are still two issues after this PR:
1. We're calling `ack` on the Satellite txs within this producer, which is a problem on the server restart, as they've not yet been persisted. Proper fix would be to listen for a ping message from PG and ACK only then, but there are yet more issues with that approach.
2. If the a "bad" txn is last in the batch, the PG will close the connection, but only after the last `send()` call, so we will have called garbage collection too early and may lose a write. This requires a more systematic fix.

### DIFF
--- a/.changeset/tough-actors-divide.md
+++ b/.changeset/tough-actors-divide.md
@@ -1,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
+Fixed a bug with lost writes when PG dropped the replication connection
--- a/components/electric/lib/electric/replication/postgres/slot_server.ex
+++ b/components/electric/lib/electric/replication/postgres/slot_server.ex
@@ -159,6 +159,7 @@ defmodule Electric.Replication.Postgres.SlotServer do
        slot_name: slot,
        origin: origin,
        producer_name: producer,
+       producer_pid: nil,
        opts: Map.get(replication_opts, :opts, []),
        # Under the current implementation, this function is always going to be
        # `ShadowTableTransformation.split_change_into_main_and_shadow/4`,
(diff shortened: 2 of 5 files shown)


alco (approved):

alco:
> There are still two issues after this PR:
> 
>  1. We're calling `ack` on the Satellite txs within this producer, which is a problem on the server restart, as they've not yet been persisted. Proper fix would be to listen for a ping message from PG and ACK only then, but there are yet more issues with that approach.

Could you elaborate on the "more issues" part? Does following this approach require more research on when would be the best time to ack clients' txs?

>  2. If the a "bad" txn is last in the batch, the PG will close the connection, but only after the last `send()` call, so we will have called garbage collection too early and may lose a write. This requires a more systematic fix.

What is a `"bad" txn`?

icehaunter:
@alco 
> "more issues"

Yes, the problem with acknowledging the clients' txn on a write from PG is that this strategy requires more persistence. Consider:
1. Client writes TX1,
2. Electric receives TX1 and stores it in this producer, 
3. We send TX1 to PG, splitting the TX into individual operations with their own LSN numbers, saving the mapping from LSN number to an index in producer,
4. Electric restarts, losing context in producer
5. Electric starts up, we have two issues: 
    1. Clients reconnect sending their unacked txs again - we can hold until PG connects and tells us its last observed LSN and immediately ACK txs that PG has actually seen, but
    2. How do we "restore" the "actual" LSN of each client-sent transaction? That was transient state in the producer which is now lost. We were long aware of that issue, but were postponing it.

> "bad txn"

Anything PG call error out on. Currently, the buggy compensation op with null values that causes an error in PG stream. I don't consider this issue of too  high priority, the aforementioned restart case is much harder to reason about

PR #376, opened 2023-08-28
fix: make sure modified bitmask is respected within one tx

icehaunter:


balegas (approved):
Approved. I'd like to come back at this later.

PR #381, opened 2023-08-29
Fix starter build.

thruflo:


examples/starter/package.json
```diff
@@ -18,7 +18,8 @@
   },
   "devDependencies": {
     "@tsmodule/tsmodule": "42",
-    "@types/node": "^18.8.4"
+    "@types/node": "^18.8.4",
+    "typescript": "^5.1.3"
```
balegas, line 22: Not a huge difference but we use  `"typescript": "^4.9"`in the client

thruflo: We use `"typescript": "^5.1.3"` in the other examples.

icehaunter (approved):
Lgtm

PR #385, opened 2023-08-29
fix(satellite): snapshot deleted all shadow entries

### LINKED ISSUE #315: Can't delete two entries in a row
Hello,
We are experiencing an issue in the todos app. 

Steps to reproduce:

* Create 2 todos
* Delete 1
* Delete the remaining one

The first delete works correctly, but the second makes the item reappear instantly.
We cannot reproduce it in the wa-sqlite example. Might have to do with the example having only one column.

Maybe this is a regression from 232f7a5bbbce92c291db0d5d9612c182ac52c677

It's a similar issue as #258 . The delete operation is being sent without an empty tags value `[]`.

As a side note, it might be worth to include an up to date version of the todos example in the monorepo or make it as a default instead of the simplified `items` database from the wa-sqlite example. There were other instances where we encountered issues with the todos app and not with the wa-sqlite example.

icehaunter:
Malformed the `WHERE EXISTS` query - it didn't use columns from the shadow table despite being a sub-query that should have, so it returned `1` for all rows in the shadow table and thus deleted all the rows. New implementation does a correct primary key join to find the rowids of all shadow rows and then deletes only them.

Big thanks to @davidmartos96 for providing a ready-made test, made debugging this a breeze! 

Closes #315

### DIFF
--- a/.changeset/sharp-hairs-crash.md
+++ b/.changeset/sharp-hairs-crash.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Fixed incorrect snapshotting issue which led to weird behaviours


thruflo:
Awesome 😎

balegas (approved):
Another one bites the dust!

PR #386, opened 2023-08-29
chore(electric): Do not lose client writes when Electric restarts

alco:
Closes VAX-911.

Notes:

* @icehaunter I remember your suggestion to use a connection pool in WsServer. I want to address that in a follow-up PR to add one pool for use both by the LSN fetching function and initial sync. Both are initiated by a WsServer process.
* @magnetised I don't fully understand the need for SchemaCache to implement the SchemaLoader behaviour. It makes the definition somewhat circular when SchemaCache itself calls functions from SchemaLoader. If the goal is to have an abstraction on top of epgsql, then we can abstract just that behind an adapter strategy and make SchemaCache a standalone module that does not implement any behaviour.


balegas:
I like that this issue is number 911

components/electric/lib/electric/postgres/extension/migrations/20230829000000_acknowledged_client_lsns_table.ex
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  @impl true
+  def down(_), do: []
+
+  @impl true
+  def create_table_ddl do
+    """
+    CREATE TABLE #{Extension.acked_client_lsn_table()} (
+      client_id TEXT PRIMARY KEY,
```
alco, line 37: It's probably a good idea to constrain this column to some fixed length. But I don't what value to pick.

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -247,10 +233,18 @@ defmodule Electric.Postgres.Extension do
...
-      Migrations.Migration_20230715000000_UtilitiesTable
+      Migrations.Migration_20230715000000_UtilitiesTable,
+      Migrations.Migration_20230829000000_AcknowledgedClientLsnsTable
     ]
   end
 
+  def create_table_ddls do
+    for migration_module <- migrations(),
+        function_exported?(migration_module, :create_table_ddl, 0) do
+      migration_module.create_table_ddl()
```
magnetised, line 244: what about if a migration  needs to create > 1 replicated table, shouldn't `create_table_ddl/0` return a list

components/electric/lib/electric/postgres/extension/migrations/20230829000000_acknowledged_client_lsns_table.ex
```diff
@@ -0,0 +1,42 @@
...
+      @trigger_sql,
+      Extension.add_table_to_publication_sql(table)
+    ]
+  end
+
+  @impl true
+  def down(_), do: []
+
+  @impl true
+  def create_table_ddl do
```
magnetised, line 34: I think this needs a different name, e.g. `replicated_tables_ddl` because you only need to use if if you're adding the internal table to the replication publication.

components/electric/lib/electric/replication/postgres/migration_consumer.ex
```diff
@@ -110,6 +110,16 @@ defmodule Electric.Replication.Postgres.MigrationConsumer do
           true
 
         %{relation: relation} = change when is_extension_relation(relation) ->
+          if relation == Extension.acked_client_lsn_relation() do
```
magnetised, line 113: could we replace this if with another guard clause e.g. `is_client_lsn_relation/1` so that we keep all the relations that we act on at the top level

components/electric/lib/electric/replication/satellite_collector_producer.ex
```diff
@@ -36,8 +40,33 @@ defmodule Electric.Replication.SatelliteCollectorProducer do
   @impl GenStage
   def handle_call({:store_incoming_transactions, transactions}, _, state) do
     transactions
-    |> Stream.each(& &1.ack_fn.())
     |> Stream.reject(&Enum.empty?(&1.changes))
+    |> Stream.map(fn tx ->
+      # NOTE(alco): !!! Potential data race warning !!!
```
magnetised, line 45: perhaps lower the scare-factor of this comment, which starts off "warning!!" of an un-handled scenario leading to bugs then goes on to say that actually it's fine, n.p.

docs/api/service.md
```diff
@@ -30,29 +30,28 @@ The Electric application is configured using environment variables.
...
+| `HTTP_PORT`<p>&nbsp;&nbsp;(`5133`)</p>              | Port for HTTP connections. Includes client websocket connections on `/ws`, and other functions on `/api`                                                                                                                                                                                                                        |
 
 ### Authentication config
 
 When `AUTH_MODE=secure` (the default), the `AUTH_JWT_ALG` and `AUTH_JWT_KEY` options are also required.
 
 When `AUTH_MODE=insecure`, all other authentication options can be omitted.
 
-| Variable                                                        | Description                                                                                                                                                                                                                                                                                                                                                                       |
```
magnetised, line 48: it's annoying that when you change a line in these tables, the whole thing gets reformatted, making it hard to see what's changed. we should perhaps remove the alignment of the table markers so we can track what's going on (I managed to make a one-line change by turning off markdown mode in vim and just deleting the line, which stopped any re-formatting on save)

components/electric/lib/electric/postgres/extension/schema_loader.ex
```diff
@@ -26,7 +26,7 @@ defmodule Electric.Postgres.Extension.SchemaLoader do
   @callback migration_history(state(), version() | nil) ::
               {:ok, [Migration.t()]} | {:error, term()}
   @callback known_migration_version?(state(), version()) :: boolean
-  @callback electrified_tables(state()) :: {:ok, [table()]} | {:error, term()}
```
magnetised, line 29: I don't like the removal of the electrified tables callback. I get that we need a separate callback that lists all tables in the replication publication (only some of which are "electrified" in the sense of application data that's been added by the developer) but removing the call to list those application-level tables seems like a mistake. 

perhaps we should keep the electrified_tables call, but remove the shadow table list from it and move those shadow tables to the new "all replicated tables" function, which would be purer.

also, this `internal_schema` call is purely an extension managed thing. the purpose of the schema_loader is to abstract the storage and retrieval of schema info, but the internal schema is defined purely in code so every implementation will just call the underlying `Extension` function.

so perhaps there's a need for a `replicated_tables` function which takes a loader impl as an arg and returns electrified tables (from the loader) + shadow tables (which is again a functional thing not stored anywhere) + extension internal stuff.

magnetised (changes requested):
Since we're having to mess with an otherwise purely application-level and electric-level split that we'd managed to maintain up til now, I think it's worth doing the work to make it consistent at a slightly deeper level in terms of how we build the list of replicated tables, in particular which are managed by the schema (and schema loader backed by pg plus migration machinery) and which are or can be purely functional/generated based on that and our extension infra.

docs/api/service.md
alco: You can switch to the Rich diff view for this value to see it rendered and nicely diffed.

alco: Screenshot from 2023-09-05 13-53-32

components/electric/lib/electric/postgres/bytea.ex
```diff
@@ -0,0 +1,31 @@
...
+
+  Because we're using the text format for tuple values in the logical replication stream between Postgres and
+  Electric, we have to encode the raw binary into one of Postgres' input encodings for BYTEA, of which the Hex
+  encoding is the simpler one.
+
+  https://www.postgresql.org/docs/current/datatype-binary.html#id-1.5.7.12.9
+  """
+
+  def to_postgres_hex(bin) do
+    for <<bbbb::4 <- bin>>, into: "\\x", do: <<to_hex_digit(bbbb)>>
```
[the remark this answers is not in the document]
alco: This function takes an Elixir value and produces a new value that is then sent to Postgres over the logical replication stream. So the encoded value needs to use Postgres syntax for bytea values, either Hex-encoded or Escape-encoded - https://www.postgresql.org/docs/current/datatype-binary.html#id-1.5.7.12.9.

That is, as long as we're using the text encoding for tuple data. I'm researching the possibility to switch to binary encoding but haven't made any progress to speak of on that front.

components/electric/lib/electric/postgres/extension/schema_loader.ex
alco: I get what you're saying with regards to splitting electrified user tables and our internal replicated tables + shadow tables of the former. But this here is a bit of a separate issue.

I remove `electrified_tables` from SchemaLoader because it had never been used, even before my addition of internal schema. If you do a code search with `main` checked out, you'll see that `electrified_tables()` is only ever called as `SchemaCache.Global.electrified_tables()` in the `Postgres.TcpServer` module. `SchemaCache.Global` simply delegates to `SchemaCache.electrified_tables()` which has its own bespoke `handle_call(:electrified_tables)` callback that doesn't mention `SchemaLoader`. So I treat this function as something specific to `SchemaCache` and not part of the `SchemaLoader` abstraction, similar to the `SchemaCache.relation/*` family of functions.

Let me know if you would like to preserve the `electrified_tables` function separately from what I implemented as `logical_publication_tables` here and:
* restore the `SchemaLoader` definition for it
* or just keep it local to `SchemaCache`

I'm thinking of defining a separate function to just return the replicated tables that are part of the internal schema and invoke that in `Postgres.TcpServer` in addition to `electrified_tables()`.

As for adjusting the abstraction implemented by `SchemaCache`, `SchemaLoader`, and `ShadowTableTransformation`, I'm not ready to take that on and that would be out of scope of this PR.

alco:
@magnetised address your feedback and implemented renaming proposals. I had to rebase on top of main to resolve conflicts but I didn't touch old commits. You can catch up with the changes made after your review by looking at https://github.com/electric-sql/electric/pull/386/commits/23d5b9df8f06a92a03c9795d11254d70a035a12e and later commits.

Thanks!

magnetised (approved):
thanks for addressing those concerns, some of which were slightly metaphysical.

PR #387, opened 2023-08-29
fix: run GC before applying the incoming txn

icehaunter:
Problem was that GC was called after the application, so when a transaction did a round-trip to the server and back, it would then be included in the conflict resolution against itself, which is a big problem if the tx was altered on the server-side conflict resolution process. Removing these transactions first ensures we don't have this problem.

### DIFF
--- a/clients/typescript/test/migrators/builder.test.ts
+++ b/clients/typescript/test/migrators/builder.test.ts
@@ -120,7 +120,7 @@ test('read migration meta data', async (t) => {
   t.deepEqual(versions, ['20230613112725_814', '20230613112735_992'])
 })
 
-test('load migration from meta data', async (t) => {
+test.skip('load migration from meta data', async (t) => {
   const db = new Database(':memory:')
   const migration = makeMigration(parseMetadata(migrationMetaData))
   const electric = await electrify(db, new DbSchema({}, [migration]), {
--- a/clients/typescript/test/migrators/builder.test.ts
+++ b/clients/typescript/test/migrators/builder.test.ts
@@ -14,6 +14,7 @@ import Database from 'better-sqlite3'
 import { electrify } from '../../src/drivers/better-sqlite3'
 import path from 'path'
 import { DbSchema } from '../../src/client/model'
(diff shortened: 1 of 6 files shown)


clients/typescript/test/migrators/builder.test.ts
```diff
@@ -120,7 +120,7 @@ test('read migration meta data', async (t) => {
   t.deepEqual(versions, ['20230613112725_814', '20230613112735_992'])
 })
 
-test('load migration from meta data', async (t) => {
+test.skip('load migration from meta data', async (t) => {
```
alco, line 123: Did you intend to commit this?

icehaunter: Yes, that test tries to actually connect to the local electric and fails if it's actually running and a different migration was applied. I don't like actually-connecting test in the suite, so I skipped it for now, to come back at some point

balegas: This is causing some uncaught exceptions. We need to set a mock socket.
I think this might do the trick

```suggestion
  const electric = await electrify(
    ...
    { socketFactory: new MockSocketFactory() }
  )
```

clients/typescript/test/satellite/process.test.ts
```diff
@@ -376,6 +376,96 @@ test('take snapshot and merge incoming wins', async (t) => {
...
+  }
+  await satellite._applyTransaction(secondIncomingTx)
+
+  const [{ value: value2 }] = await adapter.query({
+    sql: 'SELECT value FROM parent WHERE id = 1',
+  })
+  t.is(
+    value2,
+    'incoming',
+    'Applying the resolved write from the round trip should be a no-op'
```
balegas, line 465: [no-op] this is well said, we could even drop the operation, but we don't because it should not affect correctness applying it.

balegas (approved):
Thanks. Looks good. Left a suggestion.

clients/typescript/test/migrators/builder.test.ts
icehaunter: Applied

PR #389, opened 2023-08-30
chore(electric): Remove unused SYNC_MODE option from SatInStartReplicationReq

alco:
We have no logic on the client handling it. There is a bit of server-side code that checked for the presence of SYNC_MODE but since that option is never used, we can simplify the code by removing the conditional branches.

### DIFF
--- a/components/electric/lib/electric/satellite/protocol.ex
+++ b/components/electric/lib/electric/satellite/protocol.ex
@@ -544,7 +544,7 @@ defmodule Electric.Satellite.Protocol do
     if CachedWal.Api.lsn_in_cached_window?(lsn) do
       case restore_subscriptions(msg.subscription_ids, state) do
         {:ok, state} ->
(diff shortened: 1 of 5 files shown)


components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -572,14 +562,7 @@ defmodule Electric.Satellite.Protocol do
     end
   end
 
-  defp report_lsn(satellite, _pid, nil, lsn) do
```
alco, line 575: This function is called by `process_message(%SatOpLog{} = msg, ...)` which, in turn, is called after the server has sent `SatInStartReplicationReq` to the client. By this time, `in_rep.sync_batch_size` is set to 1 and doesn't ever change afterwards.

icehaunter (approved):
Always good to remove some parts of code!

PR #391, opened 2023-08-30
docs: write reference -> architecture page.

thruflo:


thruflo:
Merging prior to deploy on main URL. Happy to amend if there are suggestions in another PR.

PR #393, opened 2023-08-31
chore: replace Ranch/Cowboy with ThousandIsland/Bandit

icehaunter:
Reasons are twofold:
1. I find the codebase of Bandit & ThousandIsland much more approachable, which makes me a more confident in debugging edge cases around protocol handling. That's on top of better performance.
2. This allows us to very cleanly move our websocket server into the same plug pipeline other services are on, allowing us to use plugs very nicely - for example to consistently fill `remote_ip` field behind proxies.


alco:
Ranch and cowboy are still in `mix.lock`. Where in the codebase are their remaining uses?

icehaunter:
Good catch @alco, fixed it. Lost the result of running `mix deps.unlock ranch cowboy` somewhere in rebasing

alco:
I wanted to check if we can get rid of `cowlib` and ran `mix deps.unlock --unused` on my machine, with your branch checked out. I got this output:
```
$ mix deps.unlock --unused
Unlocked deps:
* connection
* hpax
* mime
* plug
* plug_crypto
* telemetry_metrics_prometheus_core
* websock
```

Doesn't look right, considering that we're using Plug and websock is used by Bandit.

After a closer look at mix.lock, I saw that cowlib is a dependency of `gun` which we use in `SatelliteWsClient`. I wish we could tag it as a test-only dependency but no dice, since we're using `SatelliteWsClient` in E2E tests as well.

alco:
@icehaunter I had not run `mix deps.get` prior to unlocking unused deps. Now I too only see `connection` and `telemetry_metrics_prometheus_core` as unused. Weird that it mattered, since `mix deps.get` does not modify mix.exs nor mix.lock.

Re Mint: I think Req would be a better option based on [this article](https://andrealeopardi.com/posts/breakdown-of-http-clients-in-elixir/). But that's out of scope of everything.

alco (approved):

icehaunter:
@alco req is a good wrapper over Mint, but doesn't do websockets at all

PR #394, opened 2023-08-31
feat: unify status and websocket ports

icehaunter:
Replaced the `5050` separate port for HTTP calls with handling everything on `5133`, serving websockets on `/ws`. Makes setting up Electric easier, and less finicky to configure the difference between getting the migrations and connecting the clients.


icehaunter:
YEAH @thruflo, renaming to be more generic seems good. `HTTP_PORT`?

thruflo:
Yup, or even `PORT` or maybe `SERVICE_PORT`?

The description para could read “the port to expose the sync service on, with the websocket Satellite protocol at /ws and the HTTP status API at /status” or similar.

docs/integrations/deployment/k8s.md
```diff
@@ -42,8 +42,6 @@ spec:
         - name: AUTH_JWT_KEY
           value: "..."
         ports:
-        - name: status-api
-          containerPort: 5050
         - name: satellite-ws
```
alco, line 45: This name needs to be updated.

alco (approved):

PR #396, opened 2023-09-01
docs: clearer docs on data types, constraints and limitations.

thruflo:
Firstly this adds explicit pages in usage/data-modelling on:

- data type support
- constraints

Then it goes around and tidies up the various copy around limitations.


thruflo:
Thanks @alco have addressed comments 👍

PR #397, opened 2023-09-04
chore(client): enable foreign_keys pragma on startup

kevin-dp:
Closes VAX-880.
No need to also enable the compensations because they are already enabled by the base migration:
- defined here with compensations set to 1: https://github.com/electric-sql/electric/blob/main/clients/typescript/src/migrators/schema.ts#L20
- and applied here: https://github.com/electric-sql/electric/blob/main/clients/typescript/src/migrators/bundle.ts#L38

### DIFF
--- a/clients/typescript/src/electric/index.ts
+++ b/clients/typescript/src/electric/index.ts
@@ -21,6 +21,11 @@ export interface ElectrifyOptions {
   notifier?: Notifier
   socketFactory?: SocketFactory
   registry?: Registry
+  prepare?: (connection: DatabaseAdapter) => Promise<void>
+}
+
+const defaultPrepare = async (connection: DatabaseAdapter) => {
+  await connection.run({ sql: 'PRAGMA foreign_keys = ON;' })
 }
 
 /**
@@ -38,7 +43,8 @@ export const electrify = async <DB extends DbSchema<any>>(
   opts?: Omit<ElectrifyOptions, 'adapter' | 'socketFactory'>
 ): Promise<ElectricClient<DB>> => {
   setLogLevel(config.debug ? 'TRACE' : 'WARN')
-  await adapter.run({ sql: 'PRAGMA foreign_keys = ON;' })
+  const prepare = opts?.prepare ?? defaultPrepare
+  await prepare(adapter)
 
   const configWithDefaults = hydrateConfig(config)
   const migrator =
--- a/clients/typescript/src/electric/index.ts
+++ b/clients/typescript/src/electric/index.ts
@@ -21,6 +21,13 @@ export interface ElectrifyOptions {
   notifier?: Notifier
   socketFactory?: SocketFactory
   registry?: Registry
+  /**
+   * Function that prepares the database connection.
+   * If not overridden, the default prepare function
+   * enables the `foreign_key` pragma on the DB connection.
+   * @param connection The database connection.
+   * @returns A promise that resolves when the database connection is prepared.
+   */
   prepare?: (connection: DatabaseAdapter) => Promise<void>
 }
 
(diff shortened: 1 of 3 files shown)


thruflo:
N.b.: let's also remove the dupe calls from the examples. N.n.b: this may require another PR, i.e.: merge this, release library, then merge PR relying on it.

kevin-dp:
> N.b.: let's also remove the dupe calls from the examples. N.n.b: this may require another PR, i.e.: merge this, release library, then merge PR relying on it.

Yep. This makes me think, do we really want the examples to run against the latest published version?
Having the examples run against the ts client in the mono repo would simplify the worklow and simplify evolution of the examples together with the codebase. Because now, examples cannot be evolved until next release which is also cumbersome for our own testing purposes.

thruflo:
Yup, we had the same discussion whilst you were away. Another option was versioning the published docs and examples. No solution is ideal. We need to pick one and work with it.

kevin-dp:
> Yup, we had the same discussion whilst you were away. Another option was versioning the published docs and examples. No solution is ideal. We need to pick one and work with it.

Yes. If we want examples against latest published version, i would advocate for using a separate repo as it does not make much sense to put it in the mono repo. But i'm sure you've discussed this. We can catch up on this.

balegas:
If we start managing the next release PR through Linear or with a open PR here we can track pending changes. I suspect @icehaunter might have some input on this too.

clients/typescript/src/satellite/process.ts
```diff
@@ -205,6 +205,7 @@ export class SatelliteProcess implements Satellite {
   }
 
   async start(authConfig: AuthConfig): Promise<ConnectionWrapper> {
+    await this.adapter.run({ sql: 'PRAGMA foreign_keys = ON;' })
```
balegas, line 208: Note that foreign keys need to be enabled on a connection-basis. Even though this ensures foreign keys will be on, atm when they need to be. I think this should be done closer to when the connection is open.

Would also be nice that this configuration change would be done through a initial configuration step and not just ad-hoc command.

balegas (changes requested):

clients/typescript/src/satellite/process.ts
balegas: I think the right pattern is that there might be several things we want to configure on sqlite and setting foreign keys is one of those steps

kevin-dp: I see, if there are several things to configure should we be the ones doing it though?
If we will provide a flag for several configuration options we may as well just let programmers configure the DB directly. So maybe for now best to configure it with foreign_keys pragma because we need it, and if later we need additional options we can come back to this discussion.

balegas: I’m not thinking so far in the future. Just making sure we have a step in init that is to set default configs and fk is one of them and is not just an ad hoc statement somewhere

kevin-dp:
@balegas i moved it to the base electrify call as that is the earliest point where we can get to the user-defined database connection. Let me know if that is fine.

thruflo:
Just to chip in: I think that's a great place for it. That's where we take a standard conn and electrify it, which thus includes making sure it's got the right settings etc.

That said, perhaps you could extract to a `prepare(conn)` function that's called at the start of electrify *and* make that function overridable, in case drivers need custom setup?

thruflo (approved):
Looks good 👍

clients/typescript/src/electric/index.ts
```diff
@@ -21,6 +21,11 @@ export interface ElectrifyOptions {
   notifier?: Notifier
   socketFactory?: SocketFactory
   registry?: Registry
+  prepare?: (connection: DatabaseAdapter) => Promise<void>
```
balegas, line 24: It would be better to document that if the dev overrides `prepare`, the default properties are expected to be set.

balegas (approved):

clients/typescript/src/electric/index.ts
kevin-dp: Would you prefer if we always call the default prepare that sets the foreign key pragma, followed by calling the provided prepare function (if any)?

balegas: I thought of that but couldn't decide what's better, because the dev might now some stuff is already set or not. Let's not overdo it. documenting it is just fine. A long time before anyone overrides that function

PR #400, opened 2023-09-05
chore: add protocol version negotiation on websocket connect

icehaunter:
1. Figures out compatibility between the client and the server:
    1. Protocol version for client is taken from major & minor versions of the package
    2. Protocol version for the server is it's package version verbatim
    3. Client cannot be ahead of the server - meaning that minor package version of the client cannot run ahead of the server
    4. We can just add a lower bound of supported client if/when we do more major protocol logic breaks, that way the server will always be able to tell the clients if they are compatible or not
2. Small refactoring of sockets on the client to remove an unnecessary level of abstraction
3. Replace `:gun`

## Replacing `:gun`
Reasoning for this change is twofold:
1. `:gun` requires an explicit separate process that implements a subprotocol IF the server replies with a subprotocol and immeditaly closes the connection otherwise
2. Current `SatelliteWsClient` module was written using a bare `:proc_lib` without any GenServer niceties or anything, making most of that file actually irrelevant to what it was doing. I extracted the generic websocket logic, and also extracted "builders" - we had a bunch of now-unused functions in the `SatelliteWsClient`, and other functions that built protocol messages were a bit hard to find and reason about. All of those were used only in e2e test, so I extracted all that.

Now I have 3 Elixir modules: generic websocket client over Mint, particular implmentation that is relevant to use, and a `ProtocolHelpers` module that's used by e2e code to prepare objects


components/electric/lib/electric/plug/satellite_websocket_plug.ex
```diff
@@ -13,21 +15,70 @@ defmodule Electric.Plug.SatelliteWebsocketPlug do
...
+    else
+      send_resp(conn, 400, "Bad request")
+    end
+  end
 
-      Logger.metadata(
-        remote_ip: conn.remote_ip |> :inet.ntoa() |> to_string(),
-        instance_id: Electric.instance_id()
-      )
+  defp check_if_subprotocol_present(%Plug.Conn{state: :sent} = conn), do: conn
```
alco, line 50: Why do we need to match on `state: :sent`?

components/electric/lib/satellite/satellite_ws_client.ex
```diff
@@ -29,6 +29,8 @@ defmodule Electric.Test.SatelliteWsClient do
     connect(host, port)
   end
 
+  @satellite_vsn "satellite.#{Electric.vsn().major}.#{Electric.vsn().minor}"
```
alco, line 32: I think we should have our main brand name in the protocol. Perhaps `electric/satellite` or `electric/protobuf`?

alco (approved):

components/electric/lib/electric/plug/satellite_websocket_plug.ex
icehaunter: Because those are chained functions, and each one may send an error via `send_resp(conn, 400, "error")`, meaning all further checks should not try sending their error "again"

components/electric/lib/satellite/satellite_ws_client.ex
icehaunter: websocket subprotocol header that I'm using here cannot contain any "separator characters" (i.e. `/`, `:`, etc). I'll switch to just `electric` instead

components/electric/lib/electric/plug/satellite_websocket_plug.ex
alco: This is just a nitpick but I think it'd be more idiomatic to use `with` instead of a function pipeline here. There's really no need to have the `conn` struct to be returned from each successive function. We could extract the version in `check_if_subprotocol_present` and return it as part of a tagged tuple `{:ok, satellite_vsn}` which would then be passed to `check_if_vsn_compatible` instead of `conn`.

I see the appeal of having error handling local to each function but using `with` and dealing with different error cases in its `else` clause is something I often see in Elixir code.

balegas (approved):
Looks good to me. I don't have a strong opinion on the impact of the approach

components/electric/lib/mint/web_socket_client.ex
```diff
@@ -0,0 +1,351 @@
...
+  def setup_ws_connection(protocol, host, port, path \\ "/", ws_headers \\ [])
+      when protocol in [:ws, :wss] do
+    http_protocol = if protocol == :ws, do: :http, else: :https
+
+    Logger.debug(
+      "Trying to establish websocket connection to #{protocol}://#{host}:#{port}#{path}"
+    )
+
+    with {:ok, conn} <- Mint.HTTP.connect(http_protocol, host, port, log: true),
+         {:ok, conn, ref} <- Mint.WebSocket.upgrade(protocol, conn, "/ws", ws_headers),
```
alco, line 296: Shouldn't this use `path` instead of the literal `"/ws"`?

alco (approved):

components/electric/lib/electric/application.ex
```diff
@@ -15,15 +15,20 @@ defmodule Electric.Application do
...
     children =
       children ++
         unless Application.get_env(:electric, :disable_listeners, false) do
           [
             # Satellite websocket connections are served from this router
-            {Bandit, plug: Electric.Plug.Router, port: ws_server_port()}
+            {Bandit,
+             plug: Electric.Plug.Router,
+             port: ws_server_port(),
+             thousand_island_options: [num_acceptors: 5]}
```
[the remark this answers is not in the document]
icehaunter: Good catch, I lowered it from original 100 to have a saner process count view in observer while testing something. I'll remove it

components/electric/lib/electric/plug/satellite_websocket_plug.ex
```diff
@@ -13,21 +18,74 @@ defmodule Electric.Plug.SatelliteWebsocketPlug do
...
+        Logger.metadata(
+          remote_ip: conn.remote_ip |> :inet.ntoa() |> to_string(),
+          instance_id: Electric.instance_id()
+        )
+
+        protocol_vsn = "#{conn.assigns.satellite_vsn.major}.#{conn.assigns.satellite_vsn.minor}"
+        Logger.debug("Upgrading connection for client with version #{protocol_vsn}")
+
+        conn
+        |> put_resp_header("sec-websocket-protocol", @protocol_prefix <> "#{protocol_vsn}")
```
alco, line 39: Code style nitpick:

```suggestion
        client_vsn = conn.assigns.satellite_vsn
        protocol_vsn = "#{client_vsn.major}.#{client_vsn.minor}"
        Logger.debug("Upgrading connection for client with protocol version #{protocol_vsn}")

        conn
        |> put_resp_header("sec-websocket-protocol", @protocol_prefix <> protocol_vsn)
```

components/electric/lib/mint/web_socket_client.ex
```diff
@@ -0,0 +1,351 @@
...
+                |> Mint.WebSocketClient.process_callback_response(conn_state)
+                |> case do
+                  {:noreply, state} -> {:cont, {:ok, state}}
+                  {:stop, reason, _} -> {:halt, {:stop, reason}}
+                end
+              end)
+
+            {:error, reason} ->
+              Mint.HTTP.close(state.conn)
+              {:stop, {:error, reason}}
```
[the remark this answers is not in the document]
icehaunter: Yeah, there the `reason` in `{:stop, reason, state}` can be any valid reason for a stop, but here `handle_connection/3`returns `{:error, reason}` and we want to use that tuple as the stop reason

components/electric/lib/electric/plug/satellite_websocket_plug.ex
icehaunter: I personally like the approach of piped checks like `Ecto.Changeset` does with it's functions, and I thought I may want to do more with the conn before closing the connection, hence this pattern. I appreciate the perspective and updated this to use `with`.

PR #402, opened 2023-09-05
chore(electric): Clean up Logger calls in server-side auth logic

alco:


components/electric/lib/electric/plug/satellite_websocket_plug.ex
```diff
@@ -16,10 +16,7 @@ defmodule Electric.Plug.SatelliteWebsocketPlug do
     if Bandit.WebSocket.Handshake.valid_upgrade?(conn) do
       ws_opts = if List.first(conn.path_info) == "compress", do: [compress: true], else: []
 
-      Logger.metadata(
-        remote_ip: conn.remote_ip |> :inet.ntoa() |> to_string(),
```
icehaunter, line 20: Why are you removing the remote IP here? This is a valuable piece of metadata for debugging problems of users in production

icehaunter (approved):

components/electric/lib/electric/plug/satellite_websocket_plug.ex
alco: IP addresses are user-identifiable information protected under GDPR. Our customers might not like it
when they use client IP addresses in the logs of their on-premises installation of Electric.

icehaunter: I didn't know IPs fell under GDPR. Go ahead then.

alco: @icehaunter I have done a bit more research on this issue and it appears the consensus is: logging IP addresses id fine. See https://law.stackexchange.com/a/28609.

But there are caveats.

I removed it ealier in this PR because I thought that adding IP addresses to logs should be made as part of the same decision that determines our log retention policy. Since we're not tying the logged IP addresses to users' personal information managed by customer apps, we cannot clean them up as part of the "delete all user data" procedure that app developers have in place to conform to GDPR requirements. Therefore, logging IP addresses by default is probably better avoided in production builds of Electric, though we may reintroduce it as a configurable option for better DX later on.

PR #405, opened 2023-09-07
chore: remove version from the proto package name

icehaunter:
Important note that I removed keeping unknown fields in serverside protobuf, because we're not doing roundtrips of the same message so those unknown fields are useless


clients/typescript/test/migrators/support/migrations/20230613112725_814/metadata.json
```diff
@@ -3,6 +3,6 @@
   "ops": [
     "GocBCgVzdGFycxISCgJpZBIEVEVYVBoGCgR0ZXh0EhoKCmF2YXRhcl91cmwSBFRFWFQaBgoEdGV4dBIUCgRuYW1lEgRURVhUGgYKBHRleHQSGgoKc3RhcnJlZF9hdBIEVEVYVBoGCgR0ZXh0EhgKCHVzZXJuYW1lEgRURVhUGgYKBHRleHQiAmlkChIyMDIzMDYxMzExMjcyNV84MTQS1QES0gFDUkVBVEUgVEFCTEUgInN0YXJzIiAoCiAgImlkIiBURVhUIE5PVCBOVUxMLAogICJhdmF0YXJfdXJsIiBURVhUIE5PVCBOVUxMLAogICJuYW1lIiBURVhULAogICJzdGFycmVkX2F0IiBURVhUIE5PVCBOVUxMLAogICJ1c2VybmFtZSIgVEVYVCBOT1QgTlVMTCwKICBDT05TVFJBSU5UICJzdGFyc19wa2V5IiBQUklNQVJZIEtFWSAoImlkIikKKSBXSVRIT1VUIFJPV0lEOwo="
   ],
-  "protocol_version": "Electric.Satellite.v1_4",
+  "protocol_version": "Electric.Satellite",
```
alco, line 6: Looks like this is missing the actual protocol version.

clients/typescript/test/migrators/support/migrations/20230613112735_992/metadata.json
```diff
@@ -3,6 +3,6 @@
   "ops": [
     "Gk4KBWJlZXJzEhIKAmlkEgRURVhUGgYKBHRleHQSFwoHc3Rhcl9pZBIEVEVYVBoGCgR0ZXh0GhQKB3N0YXJfaWQSBXN0YXJzGgJpZCICaWQKEjIwMjMwNjEzMTEyNzM1Xzk5MhLVARLSAUNSRUFURSBUQUJMRSAiYmVlcnMiICgKICAiaWQiIFRFWFQgTk9UIE5VTEwsCiAgInN0YXJfaWQiIFRFWFQsCiAgQ09OU1RSQUlOVCAiYmVlcnNfc3Rhcl9pZF9ma2V5IiBGT1JFSUdOIEtFWSAoInN0YXJfaWQiKSBSRUZFUkVOQ0VTICJzdGFycyIgKCJpZCIpLAogIENPTlNUUkFJTlQgImJlZXJzX3BrZXkiIFBSSU1BUlkgS0VZICgiaWQiKQopIFdJVEhPVVQgUk9XSUQ7Cg=="
   ],
-  "protocol_version": "Electric.Satellite.v1_4",
+  "protocol_version": "Electric.Satellite",
```
alco, line 6: Same here.

components/electric/lib/electric/plug/migrations.ex
```diff
@@ -91,7 +91,7 @@ defmodule Electric.Plug.Migrations do
               version: version,
               ops: &1,
               format: "SatOpMigrate",
-              protocol_version: PB.get_long_proto_vsn()
+              protocol_version: "Electric.Satellite"
```
alco, line 94: This change looks intentional but it doesn't make sense to see `Electric.Satellite` as the protocol version.

components/electric/test/electric/plug_test.exs
```diff
@@ -103,7 +103,7 @@ defmodule Electric.PlugTest do
               %{
                 "format" => "SatOpMigrate",
                 "ops" => [op1, _op2, _op3],
-                "protocol_version" => "Electric.Satellite.v1_4",
+                "protocol_version" => "Electric.Satellite",
```
alco, line 106: Same here.

alco (approved):

clients/typescript/test/migrators/support/migrations/20230613112725_814/metadata.json
icehaunter: So this protocol version was always just the full protobuf version, but I'm unsure if putting in "actual" protocol version carries any meaning here - I'm just unsure how to make this information actionable on the client. I'm open to suggestions tho

alco: The client compares this value against the value of the generated `protobufPackage` variable when loading bundled migrations. This is done as a sanity check before the client decodes the migration `ops` using the generated protobuf definition for `SatOpMigrate`.

This check is not strictly necessary, IMO. If a protocol version is upgraded on the server, that doesn't affect client code for decoding migrations, so the check does nothing in this scenario. And when the client's protocol definitions are generated from a newer protocol version, old bundled migrations should still work. I mean, we should probably maintain that backwards compatibility.

icehaunter: Yeah, I'm not sure that check makes a lot of sense, especially when applied to already-stored migrations. I think backwards compatibility is more important here, and I'm not sure what we gain with this check. I still would like to keep the field around for now, since removing may be a more breaking change and I don't want to reason about that in this PR

PR #406, opened 2023-09-07
chore: switch ping/pong to ws-level

icehaunter:
Utilize ws-level ping/pong frames for heartbeat. We technically stop using protocol level ping/pong messages here, but they were not used anyway, and are fully removed in the next PR


e2e/tests/03.09_fresh_node_satellite_receives_only_missing_migrations_during_initial_sync.lux
```diff
@@ -36,11 +36,8 @@
   ??[proto] send: #SatInStartReplicationReq{lsn: , schema: $migration1_vsn
   ??[proto] recv: #SatInStartReplicationResp{}
 
-  # The client receives only the second migration.
+  # The client receives only the second migration. Since the client sends migrations in order, we're sure that we haven't seen migraiton 1
```
alco, line 39:
```suggestion
  # The client receives only the second migration. Since the client sends migrations in order, we're sure that we haven't seen migration 1
```

alco (approved):

PR #409, opened 2023-09-07
chore: pull array chunking into utils

icehaunter:


clients/typescript/src/satellite/process.ts
```diff
@@ -1205,52 +1205,15 @@ export class SatelliteProcess implements Satellite {
...
-        await processChunk(changeType(previousChange))
-      }
-
-      if (idx === changes.length - 1) {
-        // we're at the last change
-        // process this chunk
-        const thisChange = changes[idx]
-        await processChunk(changeType(thisChange))
+    // Chunk incoming changes by their types, and process each chunk one by one
+    for (const [dataChange, chunk] of chunkBy(
```
kevin-dp, line 1209: Much better! :100

clients/typescript/src/util/arrays.ts
```diff
@@ -0,0 +1,40 @@
...
+    *[Symbol.iterator]() {
+      if (arr.length === 0) return
+
+      let currentChunkValue: K = mapper(arr[0], 0, arr)
+      let newChunkValue: K
+      let currentChunk: T[] = [arr[0]]
+
+      for (let idx = 1; idx < arr.length; ++idx) {
+        newChunkValue = mapper(arr[idx], idx, arr)
+        if (idx === 0 || currentChunkValue! === newChunkValue) {
```
kevin-dp, line 25: `idx` will never be 0.
Also `currentChunkValue! === newChunkValue` does not seem to be syntactically correct.

```suggestion
        if (currentChunkValue === newChunkValue) {
```

clients/typescript/src/util/arrays.ts
```diff
@@ -0,0 +1,40 @@
...
+          currentChunk.push(arr[idx])
+          currentChunkValue = newChunkValue
```
kevin-dp, line 27: This is only needed when processing the first element of the array.
Since we do that separately on L19 there is no need to do this here because `newChunkValue === currentChunkValue`

kevin-dp (approved):

clients/typescript/src/util/arrays.ts
icehaunter: Yeah, that and `idx === 0` is a layover from when first element was processed in loop, but I didn't like it so I pulled the first element out of the loop

clients/typescript/src/util/arrays.ts
icehaunter: It is correct, it's a TS assertion that this value is not undefined followed by a strict equality check. Also a holdover

PR #412, opened 2023-09-07
chore(electric): Clean up the response returned from the Status API

alco:
We should review the usage of connectors in the code base and refactor those bits of code now that we have a single connector. In this PR I'm only addressing the issue raised on Linear because it made sense to me to make the Status API return a human-readable explanation of the server's status.

### DIFF
--- a/components/electric/lib/electric/plug/status.ex
+++ b/components/electric/lib/electric/plug/status.ex
@@ -5,25 +5,20 @@ defmodule Electric.Plug.Status do
   use Plug.Router
   import Plug.Conn
 
-  plug(:match)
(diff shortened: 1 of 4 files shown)


icehaunter (approved):

thruflo (approved):
All looks good to me 👍 

Just to note: in https://linear.app/electric-sql/issue/VAX-903/status-api-leaks-database-structure-to-the-public-internet I was referring to leaking the actual table structure. Nothing about status messages etc.

alco:
@thruflo Were you referring to the migrations API then?

PR #413, opened 2023-09-08
feat(electric): Add support for BOOLEAN data type in electrified tables

alco:
Closes VAX-821.

### DIFF
--- a/clients/typescript/src/util/common.ts
+++ b/clients/typescript/src/util/common.ts
@@ -18,11 +18,13 @@ setGlobalUUID(
 )
 
 export const typeDecoder = {
+  bool: bytesToBool,
   number: bytesToNumber,
   text: bytesToString,
 }
 
 export const typeEncoder = {
+  bool: boolToBytes,
   number: numberToBytes,
   text: (string: string) => new TextEncoder().encode(string),
 }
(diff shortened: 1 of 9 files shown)


clients/typescript/src/util/common.ts
```diff
@@ -38,6 +40,23 @@ export const base64 = {
 
 export const DEFAULT_LOG_POS = numberToBytes(0)
 
+const trueByte = 116 // the character 't'
+const falseByte = 102 // the character 'f'
```
icehaunter, line 44:
```suggestion
const trueByte = 't'.charCodeAt(0)
const falseByte = 'f'.charCodeAt(0)
```

I slightly prefer deriving the constant as to avoid the typos

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -475,6 +476,11 @@ defmodule Electric.Satellite.Serialization do
   """
   @spec decode_column_value(binary, atom) :: binary
 
+  def decode_column_value(val, :bool) do
+    true = val in ["t", "f"]
+    val
+  end
```
icehaunter, line 482: For consistency with date PR:
```suggestion
  def decode_column_value(val, :bool) when val in ["t", "f"], do: val
```

icehaunter (approved):
LGTM! Just a couple code style nitpicks

components/electric/lib/electric/satellite/serialization.ex
alco: I'm actually siwtching to doing pattern matching inside the function body in the date PR. There's a likelihood that we'll need a catch-all clause for a particular "electrify dangerously" James has mentioned recently. So it's better for a given known type to proceed into the function body and fail there rather than rely on FunctionClauseError to do the right thing.

PR #414, opened 2023-09-08
chore(electric): Explicitly set those configuration parameters that affect formatting of values of certain types

alco:
Closes VAX-974.


icehaunter (approved):
LGTM, good find on setting them like this

PR #421, opened 2023-09-12
feat: Migration and introspection proxy

magnetised:
Also includes this new library: https://github.com/electric-sql/pg_protocol

Notes for review:

1. The bulk of this work is under the Electric.Postgres.Proxy namespace. 
2. The proxy is started by `Electric.Replication.PostgresConnectorSup` as part of the normal bringup after pg init

## Impactful changes to existing functionality

1. the `ddl_commands` table that is replicated into electric and triggers the actual migration of the internal schema no longer has a `version` string. 

   This is because the version of the migration may not be known at the point of inserting the row, since the proxy tries to capture any migration version set by a framework and this may not happen until after the migration commands have been run. 
   Instead the version is got via a separate lookup to the migration_versions table using the {txid, txts} timestamp of the migration tx

2. The event trigger is completely gone. I've removed the creation of the event trigger completely (rather than leave as-is and just remove later) because this will cause perms problems. I have added a "drop event trigger" stanza in order to bring any existing installs up-to-date (though this may cause perms problems -- not sure about that)

3. No event trigger means that the maintenance of the shadow tables is now managed by the proxy via  a new procedure `alter_shadow_table` which is triggered when the proxy detects a migration affecting an electrified table.
4. The electrify function still does all the work of verifying table schema -- the proxy just calls it
5. This PR includes an implementation of the DDLX commands, with associated sql procedures. This will probably change in the future, for now they stay as is, later prs will disable them temporarily while we work on the permissions inplementation.
6. The changes to the metadata tables in pg are done using a valid migration, so existing installs should carry on working.


thruflo:
Wow. That’s a PR.

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -162,14 +177,40 @@ defmodule Electric.Postgres.Extension do
...
+
+  @doc """
+  Given a db row which points to a compound transaction id, returns the version
+  for that transaction.
+  """
+  @spec tx_version(conn(), %{binary() => integer() | binary()}) ::
+          {:ok, String.t()} | {:error, term()}
+  def tx_version(conn, %{"txid" => txid, "txts" => txts}) do
+    with {:ok, _cols, rows} <-
+           :epgsql.equery(conn, @tx_version_query, [txid, to_integer(txts)]) do
```
alco, line 190: Why is the `to_integer()` cast needed here? The `txts` we're getting from the DB should already be of type `int8`, no?

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -162,14 +177,40 @@ defmodule Electric.Postgres.Extension do
...
+        [] ->
+          {:error, "No version found for tx txid: #{txid}, txts: #{txts}"}
+
+        [{version}] ->
+          {:ok, version}
+      end
+    end
+  end
+
+  def get_tx_version(_conn, row) do
```
alco, line 201:
```suggestion
  def tx_version(_conn, row) do
```

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -182,51 +223,68 @@ defmodule Electric.Postgres.Extension do
...
-        txts: decode_epgsql_timestamp(txts_tuple),
+        txts: txts,
+        timestamp: decode_epgsql_timestamp(timestamp),
         schema: Proto.Schema.json_decode!(schema_json),
         stmts: stmts
       }
     end)
   end
 
   @table_is_electrifed_query "SELECT count(id) AS count FROM #{@electrified_tracking_table} WHERE schema_name = $1 AND table_name = $2 LIMIT 1"
```
alco, line 238: Now that I'm looking it at, why does this query use `count()` instead of `EXISTS`?
```
SELECT EXISTS(SELECT 1 FROM #{@electrified_tracking_table} WHERE schema_name = $1 AND table_name = $2)
```

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -537,4 +572,7 @@ defmodule Electric.Postgres.Extension do
     RETURNING content->>0;
     """)
   end
+
+  defp to_integer(i) when is_integer(i), do: i
+  defp to_integer(s) when is_binary(s), do: String.to_integer(s)
```
alco, line 577: This looks like an expensive cast to be done at such a low level. Would it be a lot of effort to fix the types further up the call stack?

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -182,51 +223,68 @@ defmodule Electric.Postgres.Extension do
...
+      {:ok, count == 1}
+    end
   end
 
-  @electrifed_index_query "SELECT id, table_id  FROM #{@electrified_index_table} ORDER BY id ASC"
-  def electrified_indexes(conn) do
-    with {:ok, _, rows} <- :epgsql.equery(conn, @electrifed_index_query, []) do
-      {:ok, rows}
+  @index_electrified_query """
+  SELECT COUNT(pci.oid)
```
alco, line 248: Same comment about the user of `count()` as above. Also, this doesn't look like a cheap query. Could the run time become an issue as more and more of a Postgres schema gets electrified? In prod people may have lots of indexes, many of which are just leftovers from past optimization attempts.

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -247,8 +305,12 @@ defmodule Electric.Postgres.Extension do
       Migrations.Migration_20230512000000_conflict_resolution_triggers,
       Migrations.Migration_20230605141256_ElectrifyFunction,
       Migrations.Migration_20230715000000_UtilitiesTable,
+      Migrations.Migration_20230814170123_RenameDDLX,
+      Migrations.Migration_20230814170745_ElectricDDL,
       Migrations.Migration_20230829000000_AcknowledgedClientLsnsTable,
-      Migrations.Migration_20230918115714_DDLCommandUniqueConstraint
+      Migrations.Migration_20230918115714_DDLCommandUniqueConstraint,
+      Migrations.Migration_20230921161045_DropEventTriggers,
+      Migrations.Migration_20230921161418_ProxyCompatibility
```
alco, line 313: The order of migrations needs to be updated. There's already a published release that includes `Migrations.Migration_20230829000000_AcknowledgedClientLsnsTable`, so we can't just insert new migrations before it.

alco (commented):
Here's the first batch of review comments, so as not to overwhelm us both.

I'll come back with another batch later.

components/electric/lib/electric/postgres/extension/schema_cache.ex
```diff
@@ -272,36 +296,42 @@ defmodule Electric.Postgres.Extension.SchemaCache do
...
+    end)
+  end
 
-    {:reply, result, state}
+  def handle_call({:table_electrified?, sname, tname}, _from, state) do
+    # delegate this call directly to the extension metadata tables to avoid race conditions
+    # that can happen between an 'electrify table' call and the receipt of the
+    # migration via the replication stream - it's important that this function 
+    # be consistent with the state of the db, not our slightly laggy view on it
+    {:reply, SchemaLoader.table_electrified?(state.backend, {sname, tname}), state}
```
alco, line 309: Given the comment above this line, we shouldn't have this handled by `SchemaCache` at all. Rather the caller should call `SchemaLoader`, or better yet `Extension` directly.

But let's leave it here for now, it can be improved in a followup.

components/electric/mix.exs
```diff
@@ -66,7 +66,11 @@ defmodule Electric.MixProject do
       {:thousand_island, "~> 1.0-pre"},
       {:mint_web_socket, "~> 1.0"},
       {:mint, "~> 1.5"},
-      {:req, "~> 0.4"}
+      {:req, "~> 0.4"},
+      {:pg_protocol, github: "electric-sql/pg_protocol"},
+      {:nimble_parsec, "~> 1.3"},
+      {:postgrex, "~> 0.17", only: [:dev, :test]},
+      {:ecto_sql, "~> 3.10", only: [:dev, :test]}
```
alco, line 73: Why are these needed in :dev? How can the code that relies on those in :dev work in :prod?

e2e/common.mk
```diff
@@ -93,11 +97,17 @@ start_satellite_client_%:
 
 DOCKER_PREFIX:=$(shell basename $(CURDIR))
 docker-psql-%:
-	docker exec -it -e PGPASSWORD=password ${DOCKER_PREFIX}-$*-1 psql -h $* -U postgres -d electric
+	docker exec -it -e PGPASSWORD=password ${DOCKER_PREFIX}-$*-1 psql -h $* -p ${PG_PORT} -U postgres -d electric
```
alco, line 100: Where is `PG_PORT` defined?

e2e/common.mk
```diff
@@ -93,11 +97,17 @@ start_satellite_client_%:
...
+
+docker-proxy:
+	docker exec -it -e PGPASSWORD=${PG_PROXY_PASSWORD} ${DOCKER_PREFIX}-${PG_HOST}-1 psql -h ${PROXY_HOST} -p ${PG_PROXY_PORT} -U electric -d electric
```
alco, line 103: Where are `PROXY_HOST` and `PG_HOST` defined?

alco: Not sure if we've already used this convention for other "env" vars. But it confused me for a second here. Maybe document in this Makefile that those variables are always provided from the outside, when `make docker-proxy` is invoked?

e2e/services_templates.yaml
```diff
@@ -6,6 +6,8 @@ services:
       AUTH_JWT_ALG: HS256
       AUTH_JWT_KEY: integration-tests-signing-key-example
       ELECTRIC_TELEMETRY: disabled
+      PG_PROXY_PASSWORD:
+      PG_PROXY_PORT:
```
alco, line 10: These get overriden later, right? We don't need them here probably.

e2e/tests/compose.yaml
```diff
@@ -50,3 +53,12 @@ services:
     extends:
       file: ../services_templates.yaml
       service: satellite_client
+
+  prisma_example_1:
+    image: "prisma_example:local"
+    environment:
+      PG_PROXY_PASSWORD:
+      PG_PROXY_PORT:
```
alco, line 61: Do these get overriden also?

components/electric/lib/electric/postgres/pg_query.ex
```diff
@@ -0,0 +1,21 @@
...
+      m_t.(m)
+
+    [m | rest], f ->
+      {:|, [], [m_t.(m), f.(rest, f)]}
+  end
+
+  {:ok, modules} = :application.get_key(:pg_query_ex, :modules)
+  modules = Enum.reject(modules, &(&1 in [PgQuery, PgQuery.Parser]))
+
+  @type t() :: unquote(typespec.(modules, typespec))
```
alco, line 20:
```suggestion
  @moduledoc false

  # Introspect the :pg_query_ex application to get a list of all protocol buf structs
  # and generate a type spec that covers all of them.

  [mod | struct_modules] =
    Application.spec(:pg_query_ex, :modules)
    |> List.delete(PgQuery)
    |> List.delete(PgQuery.Parser)

  pg_query_structs_type =
    Enum.reduce(struct_modules, quote(do: %unquote(mod){}), fn name, quoted ->
      quote do
        %unquote(name){} | unquote(quoted)
      end
    end)

  @type t :: unquote(pg_query_structs_type)
```

alco (commented):
2nd batch of review comments.

components/electric/lib/electric/ddlx/command/common.ex
```diff
@@ -0,0 +1,38 @@
+defmodule Electric.DDLX.Command.Common do
+  def sql_repr(value) when value == nil do
```
alco, line 2:
```suggestion
  def sql_repr(nil) do
```

components/electric/lib/electric/ddlx/command/common.ex
```diff
@@ -0,0 +1,38 @@
...
+    "'#{escape_quotes(value, [])}'"
+  end
+
+  def sql_repr(values) when is_list(values) do
+    [
+      "ARRAY[",
+      values |> Stream.map(&sql_repr/1) |> Enum.intersperse(", "),
+      "]"
+    ]
+    |> IO.iodata_to_binary()
```
alco, line 16: Since the whole result is converted to binary in the end, I think something like this would be more efficient:
```suggestion
    array_str = for val <- values, into: "ARRAY[", do: sql_repr(val) <> ", "
    # cut the trailing comma off before appending the closing bracket
    :binary.part(array_str, 0, byte_size(array_str) - 2) <> "]"
```

components/electric/lib/electric/ddlx/command/common.ex
```diff
@@ -0,0 +1,38 @@
+defmodule Electric.DDLX.Command.Common do
+  def sql_repr(value) when value == nil do
+    "null"
+  end
+
+  def sql_repr(value) when is_binary(value) do
+    "'#{escape_quotes(value, [])}'"
```
alco, line 7:
```suggestion
    "'" <> :binary.replace(value, "'", "''", [:global]) <> "'"
```

components/electric/lib/electric/ddlx/command/revoke.ex
```diff
@@ -0,0 +1,41 @@
...
+
+  defstruct @keys
+
+  defimpl Command do
+    import Electric.DDLX.Command.Common
+
+    def pg_sql(revoke) do
+      [
+        """
+        CALL electric.revoke(#{sql_repr(revoke.privilege)}, #{sql_repr(revoke.on_table)}, #{sql_repr(revoke.role)}, #{sql_repr(revoke.column_names)}, #{sql_repr(revoke.scope)});
```
alco, line 30: This could benefit from multi-line formatting.

components/electric/lib/electric/ddlx/command.ex
```diff
@@ -0,0 +1,25 @@
...
+  @spec tag(t()) :: String.t()
+  def tag(command)
+end
+
+defimpl Electric.DDLX.Command, for: List do
+  def pg_sql(commands) do
+    Enum.flat_map(commands, &Electric.DDLX.Command.pg_sql/1)
+  end
+
+  # only return the first table name - not sure what the right answer is here TBH
```
alco, line 17: If this is ambiguous, can't we raise here instead to catch erroneous uses?

components/electric/lib/electric/ddlx/command.ex
```diff
@@ -0,0 +1,25 @@
+defprotocol Electric.DDLX.Command do
+  @spec pg_sql(t()) :: [String.t()]
```
alco, line 2: Why does this return a list of strings?

components/electric/lib/electric/ddlx/ddlx.ex
```diff
@@ -0,0 +1,36 @@
...
+  @spec is_ddlx(String.t()) :: boolean()
+  def is_ddlx(statement) do
+    Parser.is_ddlx(statement)
+  end
+
+  @doc """
+  Turns DDLX statements into Electric.DDLX.Commands.Command structs
+  """
+  @spec ddlx_to_commands(String.t() | [String.t()]) ::
+          {:ok, Command.t()} | {:ok, [Command.t()]} | {:error, String.t()}
```
alco, line 20: Is this ambiguity useful to keep? You know, we could `List.wrap()` the return value to simplify the return type.

components/electric/lib/electric/ddlx/ddlx.ex
```diff
@@ -0,0 +1,36 @@
...
+      Parser.parse(statement)
+    else
+      {:error, "not recognised"}
+    end
+  end
+
+  @doc """
+  Turns Electric.DDLX.Commands.Command structs into PostgreSQL
+  """
+  @spec command_to_postgres(Command.t() | [Command.t()]) :: String.t() | [String.t()]
```
alco, line 32: And maybe this spec could also be simplified as a result of implementing my suggestion above?

components/electric/lib/electric/ddlx/readme.md
```diff
@@ -0,0 +1,104 @@
...
+- sql
+- parse
+- commands
+
+
+### sql
+
+The SQL statements that install DDLX tables and functions into Postgres. They are actually in a single file in `sql/init_ddlx.sql`.
+
+You can use this on it own without `parse` or `commands` you want. They are really just utilities to help call the sql.
```
alco, line 14:
```suggestion
You can use this on its own without `parse` or `commands` if you want. They are really just utilities to help call the sql.
```

components/electric/lib/electric/ddlx/readme.md
```diff
@@ -0,0 +1,104 @@
...
+
+These are a set of parsers that read DDLX's extended PostgreSQL language statements and return command structs.
+
+See [https://github.com/electric-sql/docs/blob/main/documentation/docs/api/ddlx.mdx]() for the syntax of DDLX that they read.
+
+This is WIP and not everything works yet.
+
+This can be accessed through `Electric.DDLX.ddlx_to_commands()`.
+
+These is also an extra utility function `Electric.DDLX.is_ddlx()` to help check if a SQL statement is actuallt a DDLX statement.
```
alco, line 56:
```suggestion
These is also an extra utility function `Electric.DDLX.is_ddlx()` to help check if a SQL statement is actually a DDLX statement.
```

components/electric/lib/electric/ddlx/readme.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Functions internals
+
+
+### electric.enable
+currently does nothing - is the new 'electrify'
```
alco, line 62: That's not true. It calls `electric.electrify()`.

components/electric/lib/electric/ddlx/readme.md
```diff
@@ -0,0 +1,104 @@
...
+
+### electric.disable
+currently does nothing - is the new 'unelectrify'
+
+### electric.assign
+Adds and entry to the `electric.assignments` and creates a set of resources used by the assignment:
```
alco, line 68:
```suggestion
Adds a new entry to `electric.assignments` and creates a set of resources used by the assignment:
```

components/electric/lib/electric/ddlx/readme.md
```diff
@@ -0,0 +1,104 @@
...
+
+- 1 x join table
+- 2 x functions 
+- 3 x triggers
+
+When a table referenced by the assignment is created or modified the triggers run a function that updates both the join table
```
alco, line 74:
```suggestion
When a table referenced by the assignment is created or modified, the triggers run a function that both updates the join table
```

components/electric/lib/electric/ddlx/readme.md
```diff
@@ -0,0 +1,104 @@
...
+and writes the role into the `electric.roles` table.
+
+The row in the the join table has foreign keys to:
```
alco, line 77:
```suggestion
The row in the join table has foreign keys to:
```

components/electric/lib/electric/ddlx/sql/init_ddlx.sql
```diff
@@ -0,0 +1,720 @@
+-- noinspection SqlNoDataSourceInspectionForFile
+
+-- CUT HERE install uuid
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```
alco, line 4: We can get rid of this by using `gen_random_uuid()` everywhere. It's a builtin function since Postgres 13 -- https://github.com/prisma/docs/issues/2209.

components/electric/lib/electric/ddlx/parse/assign_parser.ex
```diff
@@ -0,0 +1,51 @@
...
+    },
+    %Element{required: true, type: "kv", options: ["to"], name: "user", valueType: :name},
+    %Element{required: false, type: "kv", options: ["if"], name: "if", valueType: :collection}
+  ]
+
+  use Electric.DDLX.Parse.Common
+
+  def matches(statement) do
+    String.starts_with?(statement, "electric assign")
+  end
```
alco, line 25: It's probably more a matter of style than performance, but I'd use pattern matching here as a more efficient and idiomatic construct.
```suggestion
  def matches(statement) do
    match?("electric assign" <> _, statement)
  end
```

components/electric/test/electric/ddlx/ddlx_commands_test.exs
```diff
@@ -0,0 +1,371 @@
...
+
+  def query(conn, query) do
+    with {:ok, cols, rows} <- :epgsql.equery(conn, query, []) do
+      {:ok, cols, Enum.map(rows, &Tuple.to_list/1) |> Enum.map(&null_to_nil/1)}
+    end
+  end
+
+  defp null_to_nil(row) do
+    Enum.map(row, fn
+      :null -> nil
```
alco, line 31: BTW, this can be configured in :epgsql at conncetion time. See https://github.com/epgsql/epgsql#connect.

components/electric/test/electric/ddlx/parse_test.exs
```diff
@@ -0,0 +1,442 @@
+defmodule DDLXParserTest do
+  use ExUnit.Case
```
alco, line 2:
```suggestion
  use ExUnit.Case, async: true
```

components/electric/test/electric/ddlx/parse_test.exs
```diff
@@ -0,0 +1,442 @@
...
+      tokens = Parser.get_tokens(input, re)
+
+      assert tokens == [
+               {:keyword, "hello"},
+               {:name, "from"},
+               {:name, "very stupid.old man"}
+             ]
+    end
+
+    test "can create tokens with quoted names3" do
```
alco, line 57: Um, why have these crude test names if those different cases could be written as a single table test?

components/electric/test/electric/ddlx/parse_test.exs
```diff
@@ -0,0 +1,442 @@
...
+
+      assert tokens == [
+               {:keyword, "hello"},
+               {:name, "from"},
+               {:name, "very stupid.man"}
+             ]
+    end
+  end
+
+  describe "Can parse electrid ddlx" do
```
alco, line 70:
```suggestion
  describe "Can parse electric ddlx" do
```

components/electric/test/electric/ddlx/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start()
```
alco, line 1: This looks like it was carried over from a different Mix project but is no longer needed here.

components/electric/lib/electric/ddlx/parse/common.ex
```diff
@@ -0,0 +1,185 @@
+defmodule Electric.DDLX.Parse.Common do
+  @doc """
+    uses this to template a compiled regex into each parser with the common bits from below and their own keywords
```
alco, line 3:
```suggestion
  Use this to template a compiled regex into each parser with the common bits from below and their own keywords
```

components/electric/lib/electric/ddlx/parse/common.ex
```diff
@@ -0,0 +1,185 @@
...
+    end
+  end
+
+  def scope_and_role(role_name) do
+    if String.contains?(role_name, ":") do
+      String.split(role_name, ":")
+      |> List.to_tuple()
+    else
+      {"__global__", role_name}
+    end
```
alco, line 26: I'm just nitpicking, but...
```suggestion
    case String.split(role_name, ":") do
      [schema, role] -> {schema, role}
      [role] -> {"__global__", role]
    end
```

components/electric/lib/electric/ddlx/parse/common.ex
```diff
@@ -0,0 +1,185 @@
...
+  end
+
+  def schema_and_table(table_name, default_schema) do
+    if String.contains?(table_name, ".") do
+      String.split(table_name, ".")
+      |> List.to_tuple()
+    else
+      {default_schema, table_name}
+    end
```
alco, line 35:
```suggestion
    case String.split(table_name, ".") do
      [schema, table] -> {schema, table}
      [table] -> {default_schema, table]
    end
```

components/electric/lib/electric/ddlx/parse/common.ex
```diff
@@ -0,0 +1,185 @@
...
+      ~S"\((?<collection>.+)\)",
+      ~S"'(?<string>[^']+)'",
+      ~S"(?<name>[\p{L}_0-9\$.\/🚀]+)\W"
+    ]
+
+    values_captures = Enum.join(value_types, "|")
+    keyword_captures = Enum.join(keywords, "|")
+    regexstring = "(?i)(?<keyword>#{keyword_captures})|#{values_captures}"
+    {:ok, re} = Regex.compile(regexstring, "u")
+    re
```
alco, line 58:
```suggestion
    Regex.compile!(regexstring, "u")
```

components/electric/lib/electric/ddlx/parse/common.ex
```diff
@@ -0,0 +1,185 @@
...
+  end
+
+  def get_value(values, value_name) do
+    case values[value_name] do
+      nil ->
+        nil
+
+      {_, value} ->
+        String.trim(value)
+    end
```
alco, line 68: Assuming we define `maybe_trim` in this module as
```
defp maybe_trim(nil), do: nil
defp maybe_trim(str), do: String.trim(str)
```

```suggestion
    get_in(values, [value_name, Access.elem(1)]) |> maybe_trim()
```

components/electric/lib/electric/ddlx/parse/common.ex
```diff
@@ -0,0 +1,185 @@
...
+  end
+
+  def get_value_type(values, value_name) do
+    case values[value_name] do
+      nil ->
+        nil
+
+      {value_type, _} ->
+        value_type
+    end
```
alco, line 78:
```suggestion
    get_in(values, [value_name, Access.elem(0)])
```

components/electric/lib/electric/ddlx/parse/unelectrify_parser.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  def matches(statement) do
+    String.starts_with?(statement, "unelectrify")
+  end
+
+  def make_from_values(values) do
+    {schema_name, table_name} = schema_and_table(get_value(values, "table"), @default_schema)
+
+    {
+      :ok,
```
alco, line 28: I haven't seen a single instance of returning an error tuple from `make_from_values()`. Looks like returning a tagged tuple from these is superfluous.

alco (commented):
Part 3 of N.

components/electric/lib/electric/postgres/extension.ex
magnetised: These `index_electrified?` and `table_electrified?` queries are only used by the proxy during a migration that affects an existing table, so performance isn't really a problem.

components/electric/lib/electric/postgres/extension.ex
magnetised: These are not in any critical path

components/electric/lib/electric/postgres/extension.ex
```diff
@@ -22,36 +22,47 @@ defmodule Electric.Postgres.Extension do
...
   @subscription_name "electric_replication_in"
 
+  # https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-PG-SNAPSHOT
+  # pg_current_xact_id() -> xid8
+  # The internal transaction ID type .. xid8 ... [id] a 64-bit type xid8 that
+  # does not wrap around during the life of an installation
+  @txid_type "xid8"
+  # use an int8 for the txts timestamp column because epgsql has very poor
+  # support for timestamp columns :(
+  @txts_type "int8"
```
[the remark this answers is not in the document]
magnetised: The conversion turns a timestamp into some number of microseconds, which is pg's max resolution for timestamps, so no resolution lost and would prefer to avoid messing with epgsql any further TBH

components/electric/lib/electric/postgres/extension.ex
magnetised: i wouldn't be so sure. this row data comes from the replication stream which tends to encode as binary

components/electric/lib/electric/postgres/extension.ex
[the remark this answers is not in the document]
alco: Fair enough. I have now found the explicit conversion code that I missed on my first look.

components/electric/lib/electric/postgres/extension.ex
alco: Good point.

components/electric/lib/electric/postgres/extension.ex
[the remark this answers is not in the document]
alco: Fair enough. I have now found the explicit conversion code that I missed on my first look.

components/electric/lib/electric/ddlx/parse/element.ex
```diff
@@ -0,0 +1,118 @@
+defmodule Electric.DDLX.Parse.Element do
+  alias __MODULE__
+
+  @type element_value_type :: :name | :collection | :path | :string
+
+  @type t() :: %Element{
+          required: boolean,
+          options: [String.t()],
+          type: String.t(),
+          valueType: element_value_type,
```
alco, line 10: A bit of a casing convention mishap here.

components/electric/lib/electric/ddlx/parse/element.ex
```diff
@@ -0,0 +1,118 @@
...
+  defstruct [
+    :required,
+    :options,
+    :type,
+    :name,
+    :valueType
+  ]
+
+  def read(element, tokens) do
+    if length(tokens) == 0 do
```
alco, line 28:
```suggestion
    if tokens == [] do
```

components/electric/lib/electric/ddlx/parse/parser.ex
```diff
@@ -0,0 +1,108 @@
...
+  alias Electric.DDLX.Parse.DisableParser
+  alias Electric.DDLX.Parse.UnelectrifyParser
+  alias Electric.DDLX.Parse.AssignParser
+  alias Electric.DDLX.Parse.UnassignParser
+  alias Electric.DDLX.Parse.GrantParser
+  alias Electric.DDLX.Parse.RevokeParser
+  alias Electric.DDLX.Parse.SQLiteParser
+  alias Electric.DDLX.Parse.EnableParser
+  alias Electric.DDLX.Parse.Element
+  alias Electric.DDLX.Command
```
alco, line 13: Such lists are easier to parse (by eye) when they are sorted.

```suggestion
  alias Electric.DDLX.Command
  alias Electric.DDLX.Parse.AssignParser
  alias Electric.DDLX.Parse.DisableParser
  alias Electric.DDLX.Parse.ElectrifyParser
  alias Electric.DDLX.Parse.Element
  alias Electric.DDLX.Parse.EnableParser
  alias Electric.DDLX.Parse.EnableParser
  alias Electric.DDLX.Parse.GrantParser
  alias Electric.DDLX.Parse.RevokeParser
  alias Electric.DDLX.Parse.SQLiteParser
  alias Electric.DDLX.Parse.UnassignParser
  alias Electric.DDLX.Parse.UnelectrifyParser
```

components/electric/lib/electric/ddlx/parse/parser.ex
```diff
@@ -0,0 +1,108 @@
...
+    GrantParser,
+    RevokeParser,
+    SQLiteParser,
+    EnableParser
+  ]
+
+  @quoted_re ~r/\"(?<quoted>[^\"]+)\"/u
+
+  def is_ddlx(statement) do
+    !(parser_for_statement(statement) == nil)
```
alco, line 30:
```suggestion
    not is_nil(parser_for_statement(statement))
```

components/electric/lib/electric/ddlx/parse/parser.ex
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  def parse(statement) do
+    statement = String.trim_leading(statement)
+
+    parser = parser_for_statement(statement)
+
+    if parser == nil do
+      nil
+    else
```
alco, line 40:
```suggestion
    if parser do
```

components/electric/lib/electric/ddlx/parse/parser.ex
```diff
@@ -0,0 +1,108 @@
...
+      tokens = get_tokens(statement, parser.token_regex)
```
alco, line 41:
```suggestion
      tokens = get_tokens(statement, parser.token_regex())
```

components/electric/lib/electric/ddlx/parse/parser.ex
```diff
@@ -0,0 +1,108 @@
...
+
+      results =
+        Enum.reduce_while(
+          parser.elements,
```
alco, line 45:
```suggestion
          parser.elements(),
```

components/electric/lib/electric/postgres/dialect/postgresql.ex
```diff
@@ -0,0 +1,155 @@
...
+  end
+
+  def expression(%Expression.Value{type: :BOOLEAN} = v) do
+    v.value
+  end
+
+  # def expression(%Expression.Value{type: :STRING, value: nil}) do
+  #   "''"
+  # end
+
```
alco, line 105:
```suggestion
```

components/electric/test/electric/ddlx/ddlx_commands_test.exs
```diff
@@ -0,0 +1,371 @@
...
+        column_names: ["*"],
+        scope: "projects",
+        using_path: nil,
+        check_fn: nil
+      }
+
+      sql = Electric.DDLX.command_to_postgres(grant1)
+
+      query(conn, sql)
+      query(conn, sql)
```
alco, line 116: This doesn't look like an assertion.

components/electric/lib/electric/postgres/extension/migrations/20230921161418_proxy_compatibility.ex
```diff
@@ -0,0 +1,107 @@
...
+        ALTER COLUMN txid SET DEFAULT #{schema}.current_xact_id();
+      """,
+      """
+      ALTER TABLE #{version_table} DROP CONSTRAINT IF EXISTS migration_unique_fk;
+      """,
+      """
+      ALTER TABLE #{ddl_table} ALTER COLUMN txts TYPE #{txts_type} USING (extract(epoch from txts) * 1000000)::#{txts_type};
+      """,
+      """
+      ALTER TABLE #{ddl_table} DROP COLUMN version;
```
alco, line 76: The version column is still referenced in `functions/create_active_migration.sql.eex` and `Electric.Replication.InitialSync.migrations_since()`. The latter returns a list of migrations which are then expected to have this `version` key present, I believe. Perhaps that's not being tested properly by unit tests.

components/electric/test/electric/satellite/serialization_test.exs
```diff
@@ -320,9 +323,9 @@ defmodule Electric.Satellite.SerializationTest do
             record: %{
               "id" => "6",
               "query" => "create table something_else (id uuid primary key);",
-              "txid" => "749",
-              "txts" => "2023-04-20 19:41:56.236357+00",
-              "version" => version
+              "txid" => "100",
+              "txts" => "200"
+              # "version" => version
```
alco, line 328:
```suggestion
```

components/electric/test/electric/satellite/serialization_test.exs
```diff
@@ -332,9 +335,9 @@ defmodule Electric.Satellite.SerializationTest do
             record: %{
               "id" => "7",
               "query" => "create table other_thing (id uuid primary key);",
-              "txid" => "749",
-              "txts" => "2023-04-20 19:41:56.236357+00",
-              "version" => version
+              "txid" => "100",
+              "txts" => "200"
+              # "version" => version
```
alco, line 340:
```suggestion
```

components/electric/test/electric/satellite/serialization_test.exs
```diff
@@ -344,19 +347,9 @@ defmodule Electric.Satellite.SerializationTest do
...
-          %Electric.Replication.Changes.UpdatedRecord{
-            relation: {"electric", "migration_versions"},
-            old_record: nil,
-            record: %{
-              "txid" => "749",
-              "txts" => "2023-04-20 19:41:56.236357+00",
-              "version" => version
+              "txid" => "100",
+              "txts" => "200"
+              # "version" => version
```
alco, line 352:
```suggestion
```

components/electric/test/support/mock_schema_loader.ex
```diff
@@ -199,7 +350,81 @@ defmodule Electric.Postgres.MockSchemaLoader do
...
+
+    key = tx_key(row)
+
+    # we may not explicitly configure the mock loader with txids
+    case Map.fetch(opts[:txids] || %{}, key) do
+      :error ->
+        # we only use the txid which MUST be set to the version because
+        # the mocking system has no way to propagate transaction ids through
+        # -- the txid/txts stuff is an implementation detail of the proxy system
+        # FIXME: re-factor the proxy impl to cache actions until it has a version
```
alco, line 413: Is this FIXME only pertinent to the test code? Or should we track it as a Linear issue?

components/electric/test/support/injector_test/test_scenario.ex
```diff
@@ -0,0 +1,574 @@
...
+    # {:ok, {sname, tname}} = Parser.table_name(query)
+    #
+    # struct(M.NoticeResponse,
+    #   severity: "NOTICE",
+    #   code: "00000",
+    #   message: "Migration affecting electrified table #{inspect(sname)}.#{inspect(tname)}",
+    #   detail: "Capturing migration: #{query}",
+    #   schema: sname,
+    #   table: tname
+    # )
```
alco, line 93: Leftovers?

components/electric/test/support/injector_test/test_scenario.ex
```diff
@@ -0,0 +1,574 @@
...
+      # any close->closecomplete pair should just come through the system
+      # (client->server->client) untouched
+      # %M.Close{type: "S", name: name},
+      %M.Parse{query: sql, name: name},
+      %M.Describe{name: name},
+      %M.Flush{}
+    ]
+  end
+
+  # defp _random_name() do
```
alco, line 114: If this corresponds to the `random_name()` call on line 100, the naming is inconsistent.

components/electric/test/support/injector_test/test_scenario.ex
```diff
@@ -0,0 +1,574 @@
+defmodule Electric.Postgres.Proxy.TestScenario do
```
alco, line 1: Wow, there's a lot of custom support test code in this directory. I won't be able to properly conceptualize all of it in my head. Just skimming.

components/electric/test/electric/postgres/proxy/ecto_test.exs
```diff
@@ -0,0 +1,64 @@
+defmodule Electric.Postgres.Proxy.EctoTest do
+  use ExUnit.Case, async: false
+
+  alias Electric.Postgres.Extension
+  alias Electric.Postgres.Extension.SchemaLoader
+
+  import Electric.Postgres.TestConnection
+
+  @moduletag capture_log: true
```
alco, line 9: Redudant here, like in the other files?

components/electric/lib/electric/postgres/proxy/injector/operation.ex
```diff
@@ -0,0 +1,980 @@
...
+
+        {
+          Enum.concat(
+            op.commands,
+            [%Operation.FakeExecute{msgs: msgs, tag: tag}]
+          ),
+          state
+        }
+
+        # {op.commands ++ [%Operation.FakeExecute{msgs: msgs, tag: tag}], state}
```
alco, line 974:
```suggestion
```

components/electric/test/electric/postgres/proxy/injector_test.exs
```diff
@@ -0,0 +1,618 @@
+defmodule Electric.Postgres.Proxy.InjectorTest do
+  use ExUnit.Case, async: true
+
+  alias Electric.Postgres.Proxy.Injector
+  alias Electric.Postgres.Proxy.TestScenario
+  alias Electric.Postgres.Extension.SchemaLoader
+  alias Electric.Postgres.MockSchemaLoader
+
+  @moduletag capture_log: true
```
alco, line 9: Redundant?

components/electric/lib/electric/postgres/proxy/parser.ex
```diff
@@ -0,0 +1,615 @@
...
+    quote do
+      def unquote(function)(unquote(pattern) = var!(stmt)) when unquote(guards) do
+        _ = var!(rest)
+        _ = var!(stmt)
+        unquote(block)
+      end
+    end
+  end
+
+  defp build_match(chars) do
```
alco, line 23: Please include a code sample in a comment that shows what Elixir code it produces. Perhaps there's a way to write it using `quote do end`?

components/electric/lib/electric/postgres/proxy/parser.ex
```diff
@@ -0,0 +1,615 @@
...
+    {:<<>>, [],
+     Enum.map(chars, fn {_c, i} -> {:"::", [], [{:"c#{i}", [], Elixir}, 8]} end) ++
+       [{:"::", [], [{:var!, [], [{:rest, [], Elixir}]}, {:binary, [], Elixir}]}]}
+  end
+
+  defp build_guards([{c, i}]) do
```
alco, line 29: Please include a code sample in a comment that shows what Elixir code it produces. Perhaps there's a way to write it using `quote do end`?

components/electric/lib/electric/postgres/proxy/parser.ex
```diff
@@ -0,0 +1,615 @@
...
+  defkeyword :capture?, "BEGIN", trailing: false do
+    {true, :begin}
+  end
+
+  defkeyword :capture?, "ALTER" do
+    case object(rest) do
+      "table" ->
+        {true, {:alter, "table"}}
+
+      _other ->
```
alco, line 230: Would this be the proper place to implement capture of `ALTER TYPE`? We'll need it for enum support, I believe. Not saying it should be implemented now.

components/electric/lib/electric/postgres/proxy/parser.ex
```diff
@@ -0,0 +1,615 @@
...
+    end
+  end
+
+  defp parse_with_electric_syntax(query, type) do
+    case parse_and_split(query, type) do
+      {:ok, stmts} ->
+        {:ok, stmts}
+
+      {:error, error} ->
+        smuggle_electric_syntax(query, type, error)
```
alco, line 376: Could you explain a bit what this is about and how it works at a high level in a comment?

alco (approved):
:+1: :medal_sports: :muscle: :rocket: :weight_lifting:

components/electric/lib/electric/postgres/proxy/parser.ex
magnetised: no. in fact I meant to delete this code (will do that now). 

see the `QueryAnalyser` code for how we recognise statements we need to do something with: https://github.com/electric-sql/electric/blob/garry/vax-907-implement-a-proxy-for-capturing-migrations/components/electric/lib/electric/postgres/proxy/query_analyser.ex

components/electric/test/support/injector_test/test_scenario.ex
magnetised: yeah. turned out unit-testing the thing in an easy to understand way was half the battle

components/electric/lib/electric/postgres/proxy/injector.ex
```diff
@@ -0,0 +1,165 @@
...
+    "$__" <> (:crypto.strong_rand_bytes(6) |> Base.encode16(case: :lower)) <> "__$"
+  end
+
+  def migration_version do
+    DateTime.utc_now()
+    |> Calendar.strftime("%Y%m%d%H%M%S_%f")
+  end
+
+  def debug(msg) do
+    Logger.debug(msg, component: Electric.Postgres.Proxy)
```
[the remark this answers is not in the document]
magnetised: Yup, leftover, no idea why I did that

components/electric/lib/electric/postgres/extension/migrations/20230921161418_proxy_compatibility.ex
```diff
@@ -0,0 +1,107 @@
...
+      ALTER TABLE #{ddl_table}
+        ADD CONSTRAINT ddl_table_unique_migrations
+        UNIQUE (txid, txts, query);
+      """,
+      # functions that are being re-defined as procedures:
+      """
+      DROP FUNCTION IF EXISTS capture_ddl;
+      """,
+      """
+      DROP FUNCTION IF EXISTS create_active_migration;
```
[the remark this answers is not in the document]
magnetised: good catch

components/electric/lib/electric/postgres/extension/migrations/20230921161418_proxy_compatibility.ex
magnetised: The version history query gets its data from the version table, not the ddl commands table

components/electric/lib/electric/postgres/extension/migrations/20230605141256_electrify_function.ex
```diff
@@ -41,39 +28,18 @@ defmodule Electric.Postgres.Extension.Migrations.Migration_20230605141256_Electr
...
-      CREATE INDEX electrified_tracking_table_name_oid ON #{electrified_tracking_table} (oid);
-      """,
-      """
-      CREATE TABLE #{electrified_index_table} (
-          id          oid NOT NULL PRIMARY KEY,
-          table_id    int8 NOT NULL REFERENCES #{electrified_tracking_table} (id) ON DELETE CASCADE,
-          created_at  timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP
-      );
+      CREATE INDEX #{Extension.electrified_tracking_relation()}_name_idx ON #{electrified_tracking_table} (schema_name, table_name);
+      CREATE INDEX #{Extension.electrified_tracking_relation()}_oid_idx ON #{electrified_tracking_table} (oid);
```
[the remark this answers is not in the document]
magnetised: yeah, thought it was a terrible name... I'll revert in case the name change causes problems (though don't see why it should)

components/electric/test/electric/replication/electrification_test.exs
```diff
@@ -64,18 +64,17 @@ defmodule Electric.Replication.ElectrificationTest do
             }} = wait_for_next_cached_wal_tx(lsn)
 
     # Try electrifying the table again
-    assert {:error,
-            {:error, :error, _, :raise_exception, "table public.foo is already electrified", _}} =
-             :epgsql.squery(conn, "CALL electric.electrify('public.foo')")
+    assert {:ok, [], []} = :epgsql.squery(conn, "CALL electric.electrify('public.foo')")
```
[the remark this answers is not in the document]
magnetised: intentional idempotency

components/electric/test/test_helper.exs
```diff
@@ -1,7 +1,12 @@
 if System.get_env("INTEGRATION") do
   ExUnit.configure(capture_log: true, exclude: :test, include: :integration)
 else
-  ExUnit.configure(capture_log: true, exclude: :integration, timeout: 15_000)
+  ExUnit.configure(
+    capture_log: true,
+    exclude: [:integration, :sql_generation],
```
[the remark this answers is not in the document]
magnetised: badly named tag. I added some ideas for tests that make sure the proxy just passes through stuff it doesn't recognise, but to do that properly I'd need to expand the sql generation stuff to cover all sorts of random stuff, which I didn't see as a priority. it would be a lot of work to fill out the "query space" so probably not worth it. I'll replace those disabled tests with some simpler, less ambitious, test of random sql. note that because we parse the sql that comes through, you can't just write literally random stuff without getting a bunch of errors

components/electric/lib/electric/ddlx/parse/unelectrify_parser.ex
magnetised: actually the `UnassignParser` can return an error tuple

components/electric/lib/electric/ddlx/parse/common.ex
magnetised: I actually wrote a proper parser for quoted / namespaced table names, so I'll just use that

components/electric/lib/electric/ddlx/command/common.ex
magnetised: speed is not critical for this, think the existing version is more readable tbh

e2e/services_templates.yaml
magnetised: you need these otherwise they don't get passed in...

e2e/common.mk
magnetised: passed as an arg at the call site

components/electric/mix.exs
magnetised: I'm always hitting problems with test-only deps, especially when importing their formatting rules, so this is mostly for dev-mode stuff, not production. i'll mark as test-only to remove ambiguities

components/electric/lib/electric/postgres/extension/schema_cache.ex
magnetised: I think i have a solution to the confusing schemaloader vs schemacache issue

magnetised: https://linear.app/electric-sql/issue/VAX-1193/resolve-issues-with-schemaloader-vs-schemacache

components/electric/lib/electric/postgres/extension.ex
magnetised: actually we can. the code doesn't look for ordering just application. so this will work

components/electric/lib/electric/postgres/extension.ex
magnetised: can we consider this stuff out of scope for this pr? I didn't touch this query... honest

components/electric/lib/electric/postgres.ex
```diff
@@ -75,4 +79,18 @@ defmodule Electric.Postgres do
   def json_types, do: @json_types
   def bool_types, do: @bool_types
   def uuid_types, do: @uuid_types
+
+  def supported_types do
```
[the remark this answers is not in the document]
magnetised: i disagree, the module is electric.postgres so the context is electric's view of postgres.

components/electric/lib/electric/application.ex
```diff
@@ -10,6 +10,7 @@ defmodule Electric.Application do
     children = [
       Electric.Telemetry,
       Electric.Postgres.OidDatabase,
+      Electric.Postgres.Proxy.SASL.SCRAMLockedCache,
```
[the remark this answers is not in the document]
magnetised: that cache is shared between connections so really it lives outside that supervisor

components/electric/lib/electric/postgres/proxy/sasl.ex
```diff
@@ -0,0 +1,175 @@
+defmodule Electric.Postgres.Proxy.SASL do
+  @moduledoc """
+  Does the SASL authentication dance.
+
+  See: https://www.postgresql.org/docs/current/sasl-authentication.html
+
+  Ripped almost entirely, without shame, [from
```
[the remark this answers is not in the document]
magnetised: the only bit we use without modification is the `Electric.Postgres.Proxy.SASL.SCRAMLockedCache` module, so I don't think it's worth it. perhaps if/when we replace epgsql with postgrex we can just use that active instance instead of our own

components/electric/lib/electric/postgres/extension.ex
alco: Sure. And I'm sorry for the barrage of questions and suggestions. It's just that once something is merged, it becomes difficult to make even small changes to it because there's always something more important planned. That's why I'm using the review time as an opportunity to make these small changes but only if you--the author of the PR--deem it appropriate.

components/electric/lib/electric/postgres/extension.ex
alco: But why make things more confusing than they need to be? I.e. why have the order of application different from the order in this list?

This is what it looks like when I have an existing schema from before this PR and then start Electric with the changes in this PR:
```
[localhost] electric:electric=# select * from electric.schema_migrations ;
    version     │        inserted_at         
────────────────┼────────────────────────────
 20230328113927 │ 2023-10-09 14:00:18.888432
 20230424154425 │ 2023-10-09 14:00:18.888432
 20230512000000 │ 2023-10-09 14:00:18.888432
 20230605141256 │ 2023-10-09 14:00:18.888432
 20230715000000 │ 2023-10-09 14:00:18.888432
 20230829000000 │ 2023-10-09 14:00:18.888432
 20230918115714 │ 2023-10-09 14:00:18.888432
 20230814170123 │ 2023-10-09 14:07:21.765562
 20230814170745 │ 2023-10-09 14:07:21.765562
 20230921161045 │ 2023-10-09 14:07:21.765562
 20230921161418 │ 2023-10-09 14:07:21.765562
(11 rows)
```

components/electric/lib/electric/postgres/extension/schema_cache.ex
alco: I've done some work in a local branch towards freeing SchemaCache of any loading behavior. IMO it shouldn't have any, that only muddies the water with respect to how it communicates with the DB and what is actually cached vs what is always fetched at the latest version.

There is a place for a pure SchemaCache that gets updates through explicit update calls and a separate SchemaLoader abstraction that goes out to the DB every time a function is called on it.

components/electric/README.md
```diff
@@ -127,15 +130,15 @@ See [our official docs](https://electric-sql.com/docs/usage/auth) to learn about
 
 Migrations are semi-automatically managed by the Postgres source. Once Postgres has been initialized by Electric (i.e. Electric had connected to it at least once), you will have two functions available in your SQL:
```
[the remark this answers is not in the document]
magnetised: These docs are out of date. Am going to update with some kind of proxy-specific stuff

components/electric/lib/electric/postgres/extension.ex
magnetised: I get what you're saying but if we apply this policy it means that branches will get into a race condition where migrations developed independently and merged out-of-order will have to be edited to force this (arbitrary) ordering. Yes, **some** existing installations could potentially have a migration version order that doesn't match the application order, but is this really a problem?

components/electric/lib/electric/postgres.ex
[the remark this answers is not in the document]
alco: Fair enough.

components/electric/lib/electric/postgres/dialect/sqlite.ex
```diff
@@ -38,6 +40,15 @@ defmodule Electric.Postgres.Dialect.SQLite do
   @json_types Postgres.json_types()
   @bool_types Postgres.bool_types()
   @uuid_types Postgres.uuid_types()
+  @float_types Postgres.float_types()
+  @text_types Postgres.text_types()
+  @binary_types Postgres.binary_types()
+  @datetime_types Postgres.datetime_types()
+  @json_types Postgres.json_types()
+  @bool_types Postgres.bool_types()
+  @uuid_types Postgres.uuid_types()
```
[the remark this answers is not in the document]
alco: This remains unaddressed.

components/electric/lib/electric/postgres/extension.ex
alco: > branches will get into a race condition where migrations developed independently and merged out-of-order will have to be edited to force this (arbitrary) ordering

Yes, that's exactly what needs to happen: when one person's pull request is merged, the other person with concurrent changes needs to rebase and rename. Like when two separate branches add a new E2E test with the same number; you can't get away with sloppiness in that case because duplicate numbers stand out more prominently than out-of-order timestamps in migration names. But I don't think it's OK to be sloppy with the migrations either.

components/electric/lib/electric/ddlx/command/common.ex
alco: Speed is never critical until someone starts throwing production workloads at it. But readability is important so I'll give you a pass on this ;)

components/electric/lib/electric/ddlx/readme.md
alco: This remains unaddressed.

components/electric/lib/electric/ddlx/readme.md
alco: This remains unaddressed.

components/electric/lib/electric/postgres/extension.ex
magnetised: Appreciate that, think that in this case there's enough to think about though.... 😨

PR #423, opened 2023-09-12
feat: replace ad-hoc RPC with a proper one

icehaunter:


clients/typescript/src/satellite/RPC.ts
```diff
@@ -0,0 +1,204 @@
...
+/**
+ * Build an RPC responder to reply to server-sent RPC requests.
+ *
+ * The responder function itself just correctly wraps the result or error in
+ * a SatRpcResponse object, and then sends it.
+ *
+ * @param send function to send the response to the server
+ * @returns function that builds and sends the RPC response
+ */
+export const rpcRespond =
```
kevin-dp, line 148: Is there any benefit of having this as an arrow function bound to constant instead of defining a named function?
Personally, I find a function more readable but this is a very minor comment.

clients/typescript/src/satellite/client.ts
```diff
@@ -202,6 +156,16 @@ export class SatelliteClient extends EventEmitter implements Client {
     this.outbound = this.resetReplication()
 
     this.subscriptionsDataCache = new SubscriptionsDataCache()
+    this.rpcClient = new RPC(
+      this.sendMessage.bind(this),
+      this.opts.timeout,
+      Log
+    )
+
+    this.service = withRpcRequestLogging(
```
kevin-dp, line 165: Could we give this a better name?
This is basically the core of the RPC as this object gives you the remove procedures you can invoke.
How would one of these look:
```
await this.remote.startReplication(...)
// or
await this.electric.startReplication(...) // i like how this looks, but calling it `electric` is perhaps a bit over the top
```

clients/typescript/src/satellite/client.ts
```diff
@@ -602,7 +580,41 @@ export class SatelliteClient extends EventEmitter implements Client {
...
+      )
+    } else if (message.method === 'stopReplication') {
+      const decoded = SatInStopReplicationReq.decode(message.message)
+      responder(
+        message,
+        await this.handlerForRpcRequests[message.method](decoded)
+      )
+    } else {
+      Log.warn(
+        `Server has sent an RPC request with a method client does not support: ${message.method}`
```
kevin-dp, line 603: Minor improvement to the error message:

```suggestion
        `Server has sent an RPC request with a method that the client does not support: ${message.method}`
```

clients/typescript/src/satellite/RPC.ts
icehaunter: I personally also prefer functions, but this a function that immediately returns a function, which IMO is slightly more readable as an arrow function

clients/typescript/src/satellite/client.ts
icehaunter: Service is the preferred name for this, per Protobuf itself. `this.rpc` maybe?

kevin-dp (approved):

components/electric/lib/electric/satellite/protobuf.ex
```diff
@@ -2,88 +2,101 @@ defmodule Electric.Satellite.Protobuf do
...
 
   if Enum.any?(Map.values(@mapping), &(&1 in @reserved)) do
     raise "Cannot use a reserved value as the message tag"
   end
 
+  if elem =
+       Enum.find(Enum.frequencies(Map.values(@mapping)), &match?({_, count} when count > 1, &1)) do
+    {key, count} = elem
+    raise "Cannot have duplicating value tags: #{key} is used #{count} times"
+  end
```
alco, line 47:
```suggestion
  with {tag, messages} <- find_messages_with_duplicate_value_tag(@mapping) do
    raise "Cannot have duplicate value tags: #{tag} is used #{length(messages)} times, namely, by #{inspect messages}"
  end
```

Elsewhere in the module:

```elixir
defp find_messages_with_duplicate_value_tag(map) when is_map(map) do
  map
  |> Enum.group_by(&elem(&1, 1), &elem(&1, 0))
  |> Enum.find(fn {_tag, msgs} -> match?([_, _ | _], msgs) end)
end
```

alco (approved):

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -382,6 +375,104 @@ defmodule Electric.Satellite.Protocol do
...
+    resp = %SatRpcResponse{method: method, request_id: req.request_id}
+
+    case PB.decode_rpc_request(method, req.message) do
+      {:ok, decoded} ->
+        case handle_rpc_request(decoded, state) do
+          {:reply, result, state} ->
+            {%{resp | result: {:message, rpc_encode(result)}}, state}
+
+          {:force_unpause, result, state} ->
+            {:force_unpause, %{resp | result: {:message, result}}, state}
```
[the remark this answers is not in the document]
icehaunter: Good catch!

e2e/tests/_shared.luxinc
```diff
@@ -70,24 +70,22 @@
     ?Connection established with protocol vsn
     ?Auth passed
     ?Subscribed
-    ?+rec \[\d\]: %Electric.Satellite.SatInStartReplicationReq\{
-    ?rec \[\d\]: %Electric.Satellite.SatInStartReplicationResp\{
+    ?+rec \[\d\]: %Electric.Satellite.SatRpcRequest\{(?:.*)method: "startReplication"
+    ?rec \[\d\]: %Electric.Satellite.SatRpcResponse\{(?:.*)method: "startReplication"
```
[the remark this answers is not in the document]
icehaunter: sort_maps is OTP26 tho, right?

components/electric/lib/electric/satellite/protobuf.ex
icehaunter: Good idea, but doesn't work since this is module-level compile-time check - we can't use locally defined functions here.

icehaunter: This doesn't work, unfortunately, as this is a compile-time check and thus cannot use locally-defined functions, only from other modules, and I don't want to create that yet

e2e/tests/_shared.luxinc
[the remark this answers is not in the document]
alco: Yeah, but prior to OTP 26 keys for small maps are always ordered alphabetically. So I'm not sure why `(?:.*) us neede here.

components/electric/lib/electric/satellite/protobuf.ex
alco: Oh, I missed that. An anonymous function would work at the module level.

PR #425, opened 2023-09-13
Fix messageString

davidmartos96:
Minor issue, where SatSubsReq is being logged instead of SatSubsResp


icehaunter (approved):
Well spotted!

PR #427, opened 2023-09-13
feat(electric): Add support for TIME data type in electrified tables

alco:
Closes VAX-857.


alco:
@magnetised Looping you in for this PR just to see if you maybe have a different take on the implementation from Ilia's.

.changeset/shy-days-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/electric": patch
```
magnetised, line 2: shouldn't this also include the client?

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -511,6 +512,25 @@ defmodule Electric.Satellite.Serialization do
     val
   end
 
+  def decode_column_value!(val, :time) do
```
magnetised, line 515: this serialisation module is getting biiig. might be worth splitting the column value support into a separate module

magnetised (approved):
this looks absolutely fine to me. great stuff

.changeset/shy-days-whisper.md
alco: The only change done on the client is to not crash when receiving a value of type TIME. All the user-facing changes are being implemented in DAL by Kevin.

components/electric/lib/electric/satellite/serialization.ex
alco: True! Good point. I'll refactor it after merging this and the Bytea utility module from https://github.com/electric-sql/electric/pull/386 into main.

PR #430, opened 2023-09-15
fix(electric): Monitor PostgresConnectorSup and restart it whenever Postgres is up

alco:
Fixes VAX-844.

### DIFF
--- a/.changeset/light-stingrays-listen.md
+++ b/.changeset/light-stingrays-listen.md
@@ -1,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
+Fix the problem where Postgres was failing to re-establish a replication connection to Electric after a restart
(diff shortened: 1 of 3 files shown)


icehaunter (approved):
Very good job! A patch-level changelog and good to go IMO. I ran into this as well in dealing with telemetry (namely, trying to emit a "connection ended" event)

alco:
@icehaunter @magnetised There's still room for improvement. When Postgres restarts and is ready to accept connections, Electric will still wait until its backoff timer fires before it attempts to reconnect to Postgres. A quicker response can be achieved by emitting a message from `Electric.Replication.Postgres.TcpServer.handle_connection()` and catching it in `Electric.Replication.PostgresConnectorMng` to cancel the backoff timer and try connecting immediately.

We don't have a decoupled way of doing such an emit/catch exchange without the two modules becoming dependent on one another, do we?

PR #431, opened 2023-09-15
chore(electric): Define __validate_table_column_types() at startup to use the up-to-date list of supported types

alco:
Closes VAX-1033.

### DIFF
--- a/components/electric/lib/electric/postgres/extension.ex
+++ b/components/electric/lib/electric/postgres/extension.ex
@@ -3,7 +3,7 @@ defmodule Electric.Postgres.Extension do
   Manages our pseudo-extension code
   """
 
-  alias Electric.Postgres.{Extension.Migration, Schema, Schema.Proto}
+  alias Electric.Postgres.{Schema, Schema.Proto, Extension.Functions, Extension.Migration}
   alias Electric.Utils
 
   require Logger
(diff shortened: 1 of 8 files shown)


components/electric/lib/electric/postgres/extension.ex
```diff
@@ -267,6 +278,8 @@ defmodule Electric.Postgres.Extension do
       create_migration_table(txconn)
 
       with_migration_lock(txconn, fn ->
+        define_functions(txconn)
```
magnetised, line 281: I think it's better to move the function definitions *after* the schema migrations. This would allow for any run-once preliminary setup to happen before defining the functions.

alco: That would mean taking the definition of the `electrify` function out of an internal migration and moving it to runtime as well. I wanted to avoid such drastic changes in this PR and only limit it to the type-checking function.

I can add a note explaining that this call needs to be moved elsewhere as part of VAX-1016.

icehaunter (approved):
I think you've taken a great approach with limiting the scope. I was worried we'd be redefining `electrify` on every run, but redefining just this function seems reasonable. Lovely  work!

components/electric/lib/electric/postgres/extension/functions/validate_table_column_types.sql.eex
```diff
@@ -0,0 +1,47 @@
...
+      time
+      timestamp timestamptz
+      uuid
+      varchar
+    ]a
+    |> Enum.map(&"'#{&1}'")
+    |> Enum.join(",")
+%>
+
+CREATE OR REPLACE FUNCTION electric.__validate_table_column_types(table_name text)
```
icehaunter, line 18: I'm going to merge this PR in front of launch so that we have a better upgrade story, but `electric` here should probably be taken from `@schema` in Extension module, as in all other places.

PR #432, opened 2023-09-15
fix(electric): Fix the {:error, :relation_missing} error that regularly causes E2E tests to fail on CI

alco:
At least with this change in place, I can no longer reproduce the error. Whereas before it would take me 2-3 runs to see it happen.

FIxes VAX-980.


magnetised (approved):
So we had the same problem -- glad it fixed everything

alco:
@magnetised Nope, still happening. This is from a recent CI run on `main` - https://github.com/electric-sql/electric/actions/runs/6405168943/job/17387182502.

PR #434, opened 2023-09-15
fix: Update migrate_pg lux macro to actually wait for migration to finish

davidmartos96:
This is an attempt to reduce the flakiness of the E2E tests.
Sometimes the following exception is thrown in the electric service

image

It happens when calling the macro migrate_pg sequentially, like in the test 03.10

image

My guess by looking at the macro and the "Unable to retrieve oid for table" message is that Electric is looking at the migration before it has been applied. The macro only waits for the message `Applying migration $version`, but we could match `Saved schema version $version` to know when it finished.

https://github.com/electric-sql/electric/blob/cf4ee7c2344b2391e65cb6aa4e63314728776656/e2e/tests/_shared.luxinc#L110

I'm not experienced at all with lux, but we've been using it for the Dart client as well and we learned a few things :sweat_smile: 
I hope this small change makes it more robust.

### DIFF
--- a/e2e/tests/_shared.luxinc
+++ b/e2e/tests/_shared.luxinc
@@ -107,7 +107,7 @@
     ?$psql
     [my old=$LUX_SHELLNAME]
     [shell electric]
-        ??[info] Applying migration $version
+        ??[info] Saved schema version $version
     [shell $old]
 [endmacro]
 


alco (approved):
Good find! :+1: 

I don't believe that error is related to `migate_pg`, though. I've seen it happen a lot of times and spent some time investigating. It has something to do with the internals of Electric itself.

Even when matching on `Applying migration` in E2E tests, the two `migrate_pg` calls are executed in the same PG shell, so they are execute in serially and not concurrently. In Electric, migrations are also processed sequentially. But yourchange is quite useful regardless, since it won't let the test proceed until a later line if the migration fails, making for a better localized error location reported by Lux.

davidmartos96:
Thanks! Good to hear it's being looked into.

PR #435, opened 2023-09-18
fix (starter): Remove --no-warnings flag

kevin-dp:


thruflo:
Can I just sanity check that you've tested an e2e starter flow on ubuntu?

alco:
I've googled to see if there's a way to apply the "no-warnings" setting once Node is launched. There appear to be two working solutions:
* call `process.removeAllListeners('warning')` as explained here -- https://github.com/nodejs/node/issues/32876#issuecomment-616709931 (found via https://github.com/nodejs/node/issues/10802#issuecomment-616728458)
* spawn a child Node process (letting the OS resolve the path to the executable) and pass the `--no-warnings` flag to it -- https://stackoverflow.com/a/56095929

kevin-dp:
> Can I just sanity check that you've tested an e2e starter flow on ubuntu?

I have:
- Started a Docker container using node image: `docker run -it node /bin/bash`
- Inside the container i cloned the github repo, went to the starter template, built it, linked it
- Then still from within the container, i tested `create-electric-app foo`, runs without problems.

That should confirm that the problem is gone.
I have not ran the example app in the Docker container because i don't know how to achieve that (as it would essentially have to run in the browser).

kevin-dp:
> I've googled to see if there's a way to apply the "no-warnings" setting once Node is launched. There appear to be two working solutions:
> 
>     * call `process.removeAllListeners('warning')` as explained here -- [Node warnings bubble up and in bin scripts can't be disabled  nodejs/node#32876 (comment)](https://github.com/nodejs/node/issues/32876#issuecomment-616709931) (found via [Cannot disable warnings when node is launched via a shell script. nodejs/node#10802 (comment)](https://github.com/nodejs/node/issues/10802#issuecomment-616728458))
> 
>     * spawn a child Node process (letting the OS resolve the path to the executable) and pass the `--no-warnings` flag to it -- https://stackoverflow.com/a/56095929

Yes, might be useful in the feature if we really do need to pass some flags.
However, in this case, i think the current solution (read file then parse it as JSON) is perfectly fine.

thruflo:
LGTM 👍

alco (approved):

PR #444, opened 2023-09-19
feat (DAL): timestamp support

kevin-dp:


e2e/satellite_client/src/client.ts
```diff
@@ -66,17 +66,80 @@ export const syncTable = async (electric: Electric, table: string) => {
 }
 
 export const get_tables = async (electric: Electric) => {
-  return electric.db.raw({ sql: `SELECT name FROM sqlite_master WHERE type='table';` })
+  return await electric.db.raw({ sql: `SELECT name FROM sqlite_master WHERE type='table';` })
```
alco, line 69: `get_tables` is called as `await get_tables()` in `_satellite_macros.luxinc`. From what I understand, returning a promise from an async function is more idiomatic, JavaScript will simply return it as is to the caller. It's only when you're returning a plain value from an async function will JavaScript wrap it in a promise such that it can be awaited on by the caller.

alco: I would replace all occurrences of `return await` in this file with just `return`.

e2e/satellite_client/src/client.ts
```diff
@@ -66,17 +66,80 @@ export const syncTable = async (electric: Electric, table: string) => {
...
+  const datetime = await electric.db.datetimes.findUnique({
+    where: {
+      id: id
+    }
+  })
+  console.log(`Found date time?:\n${JSON.stringify(datetime, undefined, 2)}`)
+  return datetime
+}
+
+export const read_timestamp = async (electric: Electric, id: string, expectedCreatedAt: string, expectedUpdatedAt: string) => {
```
alco, line 117: The naming of `get_timestamp` and `read_timestamp` is somewhat repetitive. What do you think about calling this `assert_local_timestamp` to make it more descriptive?

e2e/satellite_client/src/prisma/schema.prisma
```diff
@@ -0,0 +1,45 @@
+datasource db {
```
alco, line 1: Did you write this schema by hand and then generated a client from it? Smart!

When I was writing E2E tests for new types, I would run the generator to bundle migrations for new tables with the e2e satellite client. Your approach looks much nicer.

e2e/tests/03.14_node_satellite_can_sync_dates_and_times.lux
```diff
@@ -29,22 +29,14 @@
...
-    ??t: '23:05:17.999999'
+    [invoke node_await_read_datetime "001" "2023-08-23" "1970-01-01 11:00:59"]
+    [invoke node_await_read_datetime "002" "0001-01-01" "1970-01-01 00:59:03.110"]
+    [invoke node_await_read_datetime "003" "Feb 29 6000" "1970-01-01 23:05:17.999999"]
 
-    [invoke node_await_insert_extended_into "datetimes" "{id: '004', d: '1999-12-31'}"]
-    [invoke node_await_insert_extended_into "datetimes" "{id: '005', t: '00:00:00.000'}"]
+    !await client.write_datetime(db, { id: '004', d: new Date('1999-12-31'), t: new Date('1999-12-31 14:46:05.42') })
+    ?$node
+    !await client.write_datetime(db, { id: '005', d: new Date('2023-09-26'), t: new Date('2023-09-26 00:00:00.000') })
```
alco, line 38: I would use different dates for `d` and `t` here just to avoid the appearance of these two fields being related in any way.

alco (commented):
I have only reviewed the changes to e2e tests, they look great to me! Brilliant work! :+1:

e2e/satellite_client/src/prisma/schema.prisma
kevin-dp: Yes, i wrote the schema by hand and then ran the generator.
I decided to introduce the generator as a dependency because it will be needed by anyone that wants to do the same in the future.

pnpm-lock.yaml
```diff
@@ -117,6 +117,9 @@ importers:
         specifier: ^3.20.2
         version: 3.21.4
     devDependencies:
+      '@electric-sql/prisma-generator':
+        specifier: workspace:*
+        version: link:../../generator
```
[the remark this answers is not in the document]
kevin-dp: Yes, previously we committed the generated client so we did not add this dependency. However, this dependency is useful to have when we want to re-generate the client in dev mode. Perhaps we should stop pushing the generated client and change the GH action such that it invokes the generator.

e2e/satellite_client/src/client.ts
kevin-dp: Sure, can do. If returning the promise directly, no need to declare it as `async`.

kevin-dp:
> @kevin-dp what's the relation between this PR and #537?

This is the predecessor of #537 and is now made obsolete by it.
Closing this one.

PR #445, opened 2023-09-19
examples: add library version to scoped db names.

thruflo:
This updates the web and intro examples to scope the local database name to the library version. The result is that users who previously had a local database with a previous library version will use a new database after the library is updated.

The intention is to prevent some local-db incompatibility bugs with users that return to the website.

Note that the implementation uses a json file import. Is there any potential compatibility issue with this?

### DIFF
--- a/examples/introduction/src/electric.ts
+++ b/examples/introduction/src/electric.ts
@@ -1,3 +1,4 @@
+import { version } from 'electric-sql/package.json'
 import { makeElectricContext } from 'electric-sql/react'
 import { globalRegistry } from 'electric-sql/satellite'
 import { ElectricDatabase, electrify } from 'electric-sql/wa-sqlite'
@@ -22,9 +23,9 @@ export const { ElectricProvider, useElectric } = makeElectricContext<Electric>()
 
 export const initElectric = async (name: string = 'intro') => {
   const { tabId } = uniqueTabId()
-  const tabScopedDbName = `${name}-${tabId}.db`
+  const scopedDbName = `${name}-${version}-${tabId}.db`
 
-  const conn = await ElectricDatabase.init(tabScopedDbName, '/')
+  const conn = await ElectricDatabase.init(scopedDbName, '/')
 
   const config = {
     auth: {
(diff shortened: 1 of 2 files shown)


thruflo:
@icehaunter we want this on the website before launch.

thruflo:
One way or the other -- I saw various conflicting advice on the internet about how to do the json import with Typescript. Alternatively, we can parse the json in the build step?

icehaunter:
@thruflo I required our TS library version in runtime in the library itself, so now we have this magic self-updating file: https://github.com/electric-sql/electric/blob/main/clients/typescript/src/version.ts. Do you maybe want to expose that as a cleaner alternative?

thruflo:
Perfect :) this is in the latest release?

icehaunter:
@thruflo even slightly before, since 0.5.3

thruflo:
@icehaunter 

```
➜  web-wa-sqlite git:(thruflo/add-library-version-to-example-dbnames) ✗ yarn start
yarn run v1.22.19
warning ../../../../../package.json: No license field
$ SERVE=true npm run build

> @examples/web-wa-sqlite@0.1.0 build
> node copy-wasm-files.js && node builder.js

✘ [ERROR] Could not resolve "electric-sql/version"

    src/Example.tsx:3:28:
      3 │ import { LIB_VERSION } from 'electric-sql/version'
        ╵                             ~~~~~~~~~~~~~~~~~~~~~~

  The module "./dist/version/index.js" was not found on the file system:

    node_modules/electric-sql/package.json:62:11:
      62 │     "./*": "./dist/*/index.js"
         ╵            ~~~~~~~~~~~~~~~~~~~
```

Just testing the version import locally; looks like it falls foul of our package entry points config.

alco:
> One way or the other -- I saw various conflicting advice on the internet about how to do the json import with Typescript. Alternatively, we can parse the json in the build step?

We had a JSON import in the starter template until recently. It prints a warning but works, apparently. See https://github.com/electric-sql/electric/pull/435/files#diff-df9c66c1fa4ff2155189b8f37386563faf2b1ec52888779f6d69fc1f35e16636.

thruflo:
I think that's a node script.

thruflo:
This should work now with the new electric-sql rep published. I've tested successfully locally.

PR #446, opened 2023-09-19
feat (toolbar): Adding new top level package for web developer toolbar

paulharter:


toolbar/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "build": "node toolbar-build.js && tsc --emitDeclarationOnly --outDir dist"
+  },
+  "engines": {
+    "node": ">=16.11.0"
+  },
+  "dependencies": {
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  },
+  "devDependencies": {
```
icehaunter, line 31: Please add `prettier`. Reference config can be found in `clients/typescript/.prettierrc.json`

thruflo:
Just to note on the packaging stuff, does it make sense to have the example packaging in the toolbar folder? As opposed to keep the toolbar a lighter package and include it as a dependency in an/all-of-the example(s)?

Or if we need a proper app to dev/test the toolbar, maybe that goes in a subfolder that isn't included in the package list?

toolbar/package.json
paulharter: I've added prettier and copied the conf from the client

paulharter:
> Just to note on the packaging stuff, does it make sense to have the example packaging in the toolbar folder? As opposed to keep the toolbar a lighter package and include it as a dependency in an/all-of-the example(s)?
> 
> Or if we need a proper app to dev/test the toolbar, maybe that goes in a subfolder that isn't included in the package list?

This is how it works. The web app in the folder is for dev only and isn't built into the package. The package itself is quite minimal and can be added to any of the examples. I can try to tease apart the dev app code from the toolbar code for clarity. Maybe I can get code reloading etc to work properly with a clean example app and the toolbar as a dependency then I could dump the dev web app altogether. I tried this before and couldn't get it to work.

paulharter:
I've moved the dev app into a sub folder for tidiness but the structure is the same. There is a dev app that is used during development in `dev-app` but the only stuff included in the build into `dist` is the actual toolbar stuff.

I have added the toolbar to the web-wa-sqlite example. It shows if you do `export DEBUG_MODE=true `. At the moment it is included with a relative file reference in the devDependencies. At some point we need to add this package to our automated build for node. After talking to @kevin-dp I think all thats required is to add an initial build to out npm registry and the current machinery will then pick it up.

paulharter:
@thruflo I've got the SQLite console working with codemirror. And sort of got the running of DAL queries working too but was blocked by the way `useElectric` is written, so I've parked that for now.

There are a couple of outstanding build issues:

- The lib I used to get `codemirror` working with React 18 has a bad use of `global` which I'm having to patch in the build script until I can work round it better
- For some reason I can't fathom the the export of the in the component is broken but my original adding it with a function works ok. I'll try to fix this next week.

We could merge this now with these two issues or wait til fix them properly next week.

thruflo:
Awesome :) I'll have a play with it. Sounds like it's worth resolving the build issues before merging -- no mad rush.

thruflo:
(Maybe also sync on the `useElectric` thing -- I'm curious)

examples/web-wa-sqlite/src/Example.tsx
```diff
@@ -9,10 +9,17 @@ import { authToken } from './auth'
 import { DEBUG_MODE, ELECTRIC_URL } from './config'
 import { Electric, Items as Item, schema } from './generated/client'
 
+
 import './Example.css'
 
+// toolbar imports
```
thruflo, line 15: Ideally it would be good to keep the imports in this example as clean / minimal as possible. I'd suggest add a toolbar.tsx src file and have that encapsulate these details and just expose the component.

Also presumably the css import should not be required here (should be imported and thus bundled by the toolbar lib).

Also you're mixing double quote and single quite and the import spacing is inconsistent. Syntax wise you want:

```ts
import { globalRegistry } from 'electric-sql/satellite'
import { AddToolbar, TypescriptApi } from '@electric-sql/debug-toolbar'
```

This should be normalised by prettier, so have you run the listing here?

examples/web-wa-sqlite/src/Example.tsx
```diff
@@ -37,8 +44,9 @@ export const Example = () => {
       if (!isMounted) {
         return
       }
-
       setElectric(electric)
+      AddToolbar(TypescriptApi(globalRegistry))
```
thruflo, line 48: As planned, let's remove this from here and in the JSX below just render `<DebugToolbar />` where that has been setup to use the `TypescriptApi(globalRegistry)` by the separate `toolbar.tsx` file.

Also btw, the capitalisation on `AddToolbar(TypescriptApi(...))` jumps out. If these are functions, why are they starting with an upper case letter?

examples/web-wa-sqlite/src/Example.tsx
```diff
@@ -61,6 +70,7 @@ export const Example = () => {
 
 const ExampleComponent = () => {
   const { db } = useElectric()!
+  // const { db } = useContext(ElectricContext)
```
thruflo, line 73: Stray comment.

examples/web-wa-sqlite/src/index.tsx
```diff
@@ -4,3 +4,6 @@ import App from './App'
 
 const root = ReactDOM.createRoot(document.getElementById("root") as HTMLElement)
 root.render(<App />)
+
```
thruflo, line 7: Extra spaces.

pnpm-lock.yaml
```diff
@@ -320,6 +320,61 @@ importers:
         specifier: ^5.1.3
         version: 5.1.3
 
+  examples/web-wa-sqlite:
```
thruflo, line 323: Why is this added here?

pnpm-workspace.yaml
```diff
@@ -4,3 +4,4 @@ packages:
   - 'examples/starter'
   - 'generator'
   - 'components/electric'
+  - 'toolbar'
```
thruflo, line 7: This is a wider question but looking at the monorepo structure, I wonder why generator and toolbar are not in components.

toolbar/README.md
thruflo, line 1: This is the web example README. Obviously need to write the toolbar README.

toolbar/builder.js
thruflo, line 1: Can I sanity check: are we exporting types with this build? Generally, we've used ts-module for the lib dependency and esbuild for the examples (which we're going to switch to site). If this works and is simple great but let's make sure it provides the same affordances as the ts-module build.

toolbar/dev-app/backend/compose/.envrc
```diff
@@ -0,0 +1,4 @@
+# Docker images for the local stack
+export POSTGRESQL_IMAGE=postgres:14-alpine
+export ELECTRIC_IMAGE=electricsql/electric:latest
+export APP_NAME=web-wa-sqlite
```
thruflo, line 4: Wrong app name.

toolbar/dev-app/backend/compose/docker-compose.yaml
thruflo, line 1: A general point is: do we really need a duplicated backend here? We have the `local-stack` folder in the monorepo for development backends. This folder is not meant to be an example. Could we just remove this and rely on the local-stack to run behind the demo app?

I'm not sure but generally duping all this stuff everywhere is suboptimal. You need it for examples but given this isn't an example, it's maybe simpler to leverage the existing local-stack?

toolbar/dev-app/public/index.html
```diff
@@ -0,0 +1,14 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=360,initial-scale=1.0" />
+    <title>Web Example - ElectricSQL</title>
```
thruflo, line 7: Wrong title.

toolbar/dev-app/public/index.js
thruflo, line 1: Is this not a built artefact? What is this file?

toolbar/dev-app/public/index.js.map
thruflo, line 1: Ditto.

toolbar/dev-app/public/robots.txt
thruflo, line 1: a) do we need it? b) is this actually valid syntax, with the empty value for Disallow? (If not and other code uses it, let's fix that).

toolbar/dev-app/src/App.tsx
```diff
@@ -0,0 +1,18 @@
...
+import './App.css'
+import './style.css'
+
+import { Example } from './Example'
+
+export default function App() {
+  return (
+    <div className="App">
+      <header className="App-header">
+        TOOLBAR DEV PAGE
```
thruflo, line 12: Classy

toolbar/dev-app/src/Example.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { useEffect, useState, useContext } from 'react'
+
+import { makeElectricContext, useLiveQuery } from 'electric-sql/react'
+import { genUUID, uniqueTabId } from 'electric-sql/util'
+import { ElectricDatabase, electrify } from 'electric-sql/wa-sqlite'
+import { authToken } from './auth'
+import { DEBUG_MODE, ELECTRIC_URL } from './config'
+import { Electric, Items as Item, schema } from './generated/client'
+import { ElectricContext } from 'electric-sql/react'
```
thruflo, line 9: Just OCD on ordering of the imports -- you have two lines far apart importing from `electric-sql/react` and I'd move the satellite import above the util import on line 4.

toolbar/dev-app/src/Example.tsx
```diff
@@ -0,0 +1,115 @@
...
+      const { tabId } = uniqueTabId()
+      const tabScopedDbName = `electric-${tabId}.db`
+
+      const conn = await ElectricDatabase.init(tabScopedDbName, '')
+      const electric = await electrify(conn, schema, config)
+
+      if (!isMounted) {
+        return
+      }
+      AddToolbar(TypescriptApi(globalRegistry))
```
thruflo, line 41: a) let's move to the component approach as per prev comments on web example; b) with this as is, the spacing is a bit loose -- I would have it as:

```ts
  if (!isMounted) {
    return
  }

  setElectric(electric)

  AddToolbar(TypescriptApi(globalRegistry))
}
```

toolbar/dev-app/src/Example.tsx
```diff
@@ -0,0 +1,115 @@
...
+    return () => {
+      isMounted = false
+    }
+  }, [])
+
+  if (electric === undefined) {
+    return null
+  }
+
+
```
thruflo, line 57: Extra spaces.

toolbar/dev-app/src/Example.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <ElectricProvider db={electric}>
+      <ExampleComponent />
+      {/*<ElectricToolbar api={TypescriptApi(globalRegistry)} />*/}
+    </ElectricProvider>
+  )
+}
+
+const ExampleComponent = () => {
+  //const { db } = useElectric()!
+  const { db } = useContext(ElectricContext)
```
thruflo, line 69: What's the issue with `useElectric`?

toolbar/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@electric-sql/debug-toolbar",
+  "version": "0.1.0",
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.esm.js",
+  "types": "dist/index.d.ts",
+  "author": "ElectricSQL",
+  "license": "Apache-2.0",
+  "scripts": {
```
thruflo, line 9: Does it make sense to have the backend and demo-app related scripts in the package.json here. The package.json is copied into the dist. Could the library package.json be clean and the demo-app package.json have the db:migrate client:generate scripts etc.?

toolbar/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "engines": {
+    "node": ">=16.11.0"
+  },
+  "dependencies": {
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-codemirror2": "^7.2.1",
+    "codemirror": "^5.57.0"
+  },
+  "devDependencies": {
```
thruflo, line 33: As per above on scripts, if you cut out the dev dependencies to the demo-app package.json then this would maybe allow you to remove from here? I.e.: are there dev dependencies to build and test the toolbar that live here and dev dependencies to run the demo app that belong in the demo app?

toolbar/src/client/api/api-typescript.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  getSatelliteNames(): string[] {
+    return Object.keys(this.globalRegistry.satellites)
+  }
+
+  getSatelliteStatus(name: string): string {
+    let sat = this.globalRegistry.satellites[name]
+    if (sat === undefined) {
+      return ''
```
thruflo, line 21: Is returning an empty string the right approach here?

toolbar/src/client/api/api-typescript.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  resetDB(dbName: string): void {
+    const DBDeleteRequest = window.indexedDB.deleteDatabase(dbName)
+    DBDeleteRequest.onsuccess = function () {
+      console.log('Database deleted successfully')
+    }
+    // the indexedDB cannot be deleted if the database connection is still open,
+    // so we need to reload the page to close any open connections.
+    // On reload, the database will be recreated.
+    window.location.reload()
```
thruflo, line 40: This is obviously web specific. But the adapter/strategy pattern with ApiTypescript(globalRegistry) just specifies typescript. We use Typescript for web and mobile drivers.

toolbar/src/client/api/api-typescript.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  queryDB(dbName: string, sql: string): Promise<Row[]> {
+    let sat = this.globalRegistry.satellites[dbName]
+    if (sat === undefined) {
+      return new Promise((resolve, _reject) => {
```
thruflo, line 46: Perhaps this could be an async function to naturally return a promise, so you can just return `[]` here.

toolbar/src/client/api/api-typescript.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  constructor(globalRegistry: GlobalRegistry) {
+    this.globalRegistry = globalRegistry
+  }
+
+  getSatelliteNames(): string[] {
+    return Object.keys(this.globalRegistry.satellites)
+  }
+
+  getSatelliteStatus(name: string): string {
```
thruflo, line 18: These functions that take a name, could they instead take a satellite instance? That way you move the check whether satellite for dbname exists to a single place and then make the implementation of this api simpler?

toolbar/src/index.css
```diff
@@ -0,0 +1,160 @@
+#electric-toolbar {
```
thruflo, line 1: I guess it makes sense to only have one toolbar at a time and for it to always be at the bottom. But does this need to be the concern of the core toolbar component, i.e.: could the toolbar have normal positioning and then there's a wrapper that deals with the fixed / absolute etc.?

toolbar/src/index.css
```diff
@@ -0,0 +1,160 @@
...
+}
+
+.Toolbar {
+  all: revert;
+  border-color: lightgrey;
+  border-width: 1px;
+  border-style: solid;
+  border-radius: 5px;
+  height: fit-content;
+  font-family: 'Helvetica Neue', Helvetica, sans-serif;
```
thruflo, line 17: This is not our brand font.

toolbar/src/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+export type { ToolbarApiBase } from './client/api/api-base'
+
+export type ToolbarProps = {
+  api: ToolbarApiBase
+}
+
+export default function ElectricToolbar({ api }: ToolbarProps) {
+  setApi(api)
+  const [hidden, setHidden] = useState(true)
+  const [dbNames, setDbNames] = useState(['mary', 'mungo', 'midge'])
```
thruflo, line 31: these default values are presumably a temporary dev thing

thruflo (changes requested):

thruflo:
Note from `ppm install` at repo root:

```
toolbar
└─┬ react-codemirror2 7.2.1
  └── ✕ unmet peer react@">=15.5 <=16.x": found 18.2.0
```

Seems `react-codemirror2` expects React unto 16.x.

thruflo:
There's a missing `zod` dev rep:

```
➜  toolbar git:(paul/toolbar) ✗ yarn start
yarn run v1.22.19
warning ../../../../package.json: No license field
$ SERVE=true npm run build-dev

> @electric-sql/debug-toolbar@0.1.0 build-dev
> node ./dev-app/copy-wasm-files.js && node dev-app/builder.js

✘ [ERROR] Could not resolve "zod"

    dev-app/src/generated/client/index.ts:1:18:
      1 │ import { z } from 'zod';
        ╵                   ~~~~~

  You can mark the path "zod" as external to exclude it from the bundle, which will remove this
  error.

error Command failed with exit code 1.
➜  toolbar git:(paul/toolbar) ✗
```

thruflo:
I get an error trying to edit the SQL command. Specifically, I can move the cursor around but then when I try to e.g.: type backspace to delete a character I get:

```
[Error] TypeError: string.split is not a function. (In 'string.split(/\r\n?|\n/)', 'string.split' is undefined)
	(anonymous function) (index.js:18585)
	callCallback2 (index.js:5600)
	dispatchEvent
	invokeGuardedCallbackDev (index.js:5625)
	invokeGuardedCallback (index.js:5659)
	reportUncaughtErrorInDEV (index.js:18584)
	captureCommitPhaseError (index.js:21480)
	commitLayoutMountEffects_complete (index.js:19904)
	commitLayoutEffects_begin (index.js:19891)
	commitLayoutEffects (index.js:19842)
	commitRootImpl (index.js:21275)
	commitRoot (index.js:21199)
	performSyncWorkOnRoot (index.js:20817)
	performSyncWorkOnRoot
	flushSyncCallbacks (index.js:11061)
	(anonymous function) (index.js:20549)
[Error] The above error occurred in the <Controlled2> component:

Controlled2@http://localhost:3002/index.js:48143:32
div
div
div
SQLTab@http://localhost:3002/index.js:67439:25
div
div
ToolbarTabs@http://localhost:3002/index.js:67561:30
div
ElectricToolbar@http://localhost:3002/index.js:67641:31

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://reactjs.org/link/error-boundaries to learn more about error boundaries.
	logCapturedError (index.js:15962)
	(anonymous function) (index.js:15982)
	callCallback (index.js:12437)
	commitUpdateQueue (index.js:12454)
	commitLayoutEffectOnFiber (index.js:19009)
	commitLayoutMountEffects_complete (index.js:19902)
	commitLayoutEffects_begin (index.js:19891)
	commitLayoutEffects (index.js:19842)
	commitRootImpl (index.js:21275)
	commitRoot (index.js:21199)
	performSyncWorkOnRoot (index.js:20817)
	performSyncWorkOnRoot
	flushSyncCallbacks (index.js:11061)
	(anonymous function) (index.js:20549)
[Error] TypeError: string.split is not a function. (In 'string.split(/\r\n?|\n/)', 'string.split' is undefined)
	flushSyncCallbacks (index.js:11071)
	(anonymous function) (index.js:20549)
```

thruflo:
I have a few comments on the layout and naming:

- [ ] status, could perhaps be connection; and would be great to use the connectivity toggle here to allow user to toggle on and off
- [ ] IndexedDB, firstly this is web driver config specific; secondly maybe we want to call this Reset; thirdly, perhaps connectivity, reset, re-sync, clear local changes, etc. could all be in a single panel together
- [ ] SQLite, not quite sure of the right name, maybe shell? on the layout / results, I think we want a nice table view for the results with column names; this really needs full width, so I think the input box has to be above the results, unless maybe at very large widths

On the last point, generally we want this to work with mobile too, so we have to take care of how functional the UI is at smaller sizes. And it might be nice to make resizable -- although I know the extension panel will handle this so maybe ignore. But right now as an overlay, it hides the page content underneath it, rather than allowing that to scroll.

examples/web-wa-sqlite/src/Example.tsx
paulharter: Accessing the globalRegistry needs to be done from the client code not the library I played with this a lot - its to do with when to global singleton gets created

examples/web-wa-sqlite/src/Example.tsx
paulharter: I am having trouble building this as a library. The function version of the root component gets compiled into a JXS clas object which then breaks the React 18 hooks - I need to dig into how React compiles more

toolbar/src/client/api/api-typescript.ts
paulharter: I wanted to avoid doing this and keep the actual api agnostic of the client types as much as possible so it can potentially be reused with other clients

toolbar/src/index.css
paulharter: This is useful to get the css reset working nicely - this id specificity overrides class ones

toolbar/src/index.tsx
paulharter: A dev thing but not temporary - there for developing against without a real client

paulharter: oops not those ones!

toolbar/src/index.css
paulharter: What is? OpenSauceOne?

pnpm-workspace.yaml
icehaunter: I agree btw, I feel like toolbar definitely belongs in components

paulharter:
@thruflo can you have another look at this now? It's still using the separate root element added into the dom. I couldn't get it to build as an exported react component, due to complexities of loading react. (The codemirror react library I'm using has a `require("react")` which makes me marking react as external causes a runtime error for dynamic imports, but if I don't I get issues with two reacts clashing and disallowing my hooks.)

I have published it to the npm registry yet - this might have to wait til next week.

toolbar/src/tabs.tsx
```diff
@@ -0,0 +1,65 @@
...
+    )
+  }
+}
+
+export type ToolbarTabsProps = {
+  dbName: string
+}
+
+export default function ToolbarTabs({ dbName }: ToolbarTabsProps): JSX.Element {
+  const [active, setActive] = useState('status')
```
kevin-dp, line 37: can be typed more precisely:
```
useState<'status' | 'db' | 'sql'>('status')
```

Then we can remove the default branch in the `renderComp` function.

toolbar/.gitignore
```diff
@@ -0,0 +1,11 @@
+dist
+dev-app/dist
+node_modules
+dev-app/public/wa-sqlite-async.wasm
```
kevin-dp, line 4: dev-app can go if we remove it

toolbar/package.json
kevin-dp: Agree. I would even argue that we don't want to package a demo-app with the toolbar. Demo apps should be found in `examples`.

kevin-dp (changes requested):
Great work!
I left quite some comments but nothing major.
Also, prettier reports some formatting issues when running `npm run check-styleguide` both in the toolbar and in linearlite.

examples/web-wa-sqlite/src/Example.tsx
```diff
@@ -11,8 +11,15 @@ import { Electric, Items as Item, schema } from './generated/client'
 
 import './Example.css'
 
+// toolbar imports
+import { globalRegistry } from "electric-sql/satellite";
+import AddToolbar, { TypescriptApi } from '@electric-sql/debug-toolbar'
+import '@electric-sql/debug-toolbar/dist/index.cjs.css'
+
 const { ElectricProvider, useElectric } = makeElectricContext<Electric>()
```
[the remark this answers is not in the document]
paulharter: The global registry has to be imported from `electric-sql/satellite` in the developers own code to get the right version of the global

examples/web-wa-sqlite/builder.js
```diff
@@ -82,6 +82,7 @@ let buildParams = {
   plugins: [inlineImage()],
 };
```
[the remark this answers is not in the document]
paulharter: Did I add that there? if so I shouldn't have. removed

toolbar/src/client/api/index.ts
```diff
@@ -0,0 +1,11 @@
+import { ToolbarApiBase } from './api-base'
+
+let toolbar_api: ToolbarApiBase
+
+export function setApi(api: ToolbarApiBase) {
```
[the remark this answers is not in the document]
paulharter: The api isn't stateful and it is a global. It's holding a particular version of the base interface. Putting it here is a bit messy, I could attach it to the window? or create a new registry object but that seem a bit heavyweight.

toolbar/src/client/api/api-base.ts
```diff
@@ -0,0 +1,10 @@
+export type SqlValue = string | number | null | Uint8Array | bigint
```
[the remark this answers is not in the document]
paulharter: Importing the types from the utils file pulls in a lot of other unnecessary things to the api, can we refactor the types out of the utils file into a file with just the types?

toolbar/src/client/api/api-base.ts
```diff
@@ -0,0 +1,10 @@
...
+export type Row = { [key: string]: SqlValue }
+
+export interface ToolbarApiBase {
+  getSatelliteNames(): string[]
+  getSatelliteStatus(name: string): string
+  resetDB(dbName: string): void
+  queryDB(dbName: string, sql: string): Promise<Row[]>
```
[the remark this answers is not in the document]
paulharter: I want to try and keep the api as free as possible of types in the client. As its an api

toolbar/src/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      // </div>
+    )
+  }
+}
+
+export function TypescriptApi(globalRegistry: GlobalRegistry) {
+  return new ToolbarApiTypescript(globalRegistry)
+}
+
+export function DummyApi() {
```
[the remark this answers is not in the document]
paulharter: This is for development and very useful

toolbar/src/client/api/api-base.ts
```diff
@@ -0,0 +1,10 @@
+export type SqlValue = string | number | null | Uint8Array | bigint
+export type Row = { [key: string]: SqlValue }
+
+export interface ToolbarApiBase {
```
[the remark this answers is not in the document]
paulharter: how about 'ApiInterface` and `ApiTypescript` ?

toolbar/src/client/api/api-typescript.ts
```diff
@@ -0,0 +1,71 @@
+import { ToolbarApiBase } from './api-base'
+import { GlobalRegistry } from 'electric-sql/satellite'
+export type SqlValue = string | number | null | Uint8Array | bigint
+export type Row = { [key: string]: SqlValue }
+// import * as ts from "typescript"
+
+export class ToolbarApiTypescript implements ToolbarApiBase {
```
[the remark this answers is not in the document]
paulharter: I've gone with `ToolbarTypescript`

toolbar/src/client/api/api-typescript.ts
```diff
@@ -0,0 +1,71 @@
+import { ToolbarApiBase } from './api-base'
+import { GlobalRegistry } from 'electric-sql/satellite'
+export type SqlValue = string | number | null | Uint8Array | bigint
```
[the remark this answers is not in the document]
paulharter: i've moved these imports into a `types.ts` file

toolbar/src/client/api/api-base.ts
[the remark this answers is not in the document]
paulharter: I've gone with `ToolbarInterface` and `ToolbarTypescript` for these tow and moved the dummy one into the dev-app

toolbar/src/index.tsx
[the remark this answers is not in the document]
paulharter: I've moved this into the now absent dev app

toolbar/src/client/api/api-base.ts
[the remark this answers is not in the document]
paulharter: I've followed your suggestions for adding complex types int the api definition, but I have some reservations about this as it reduces the portability - but were not there yet so maybe it won't cause the problems I envisage.

toolbar/src/client/api/api-base.ts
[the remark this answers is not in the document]
paulharter: This works ok now with the `types.ts`

toolbar/src/client/api/index.ts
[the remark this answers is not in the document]
paulharter: I've moved this into a prop that is passed down through to the tab components

examples/web-wa-sqlite/src/Example.tsx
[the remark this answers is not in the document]
paulharter: I haven't found a way of tidying up the `globalRegistry` import

examples/linearlite/src/App.tsx
```diff
@@ -11,6 +11,11 @@ import LeftMenu from './components/LeftMenu'
 import { ElectricProvider, initElectric, dbName, DEBUG } from './electric'
 import { Electric } from './generated/client'
 
+// toolbar imports
+import { globalRegistry } from "electric-sql/satellite";
+import AddToolbar, { TypescriptApi } from '@electric-sql/debug-toolbar'
```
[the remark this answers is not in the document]
paulharter: I've gone with `ToolbarTypescript`as there may be other

toolbar/src/client/api/api-base.ts
```diff
@@ -0,0 +1,10 @@
+export type SqlValue = string | number | null | Uint8Array | bigint
+export type Row = { [key: string]: SqlValue }
+
```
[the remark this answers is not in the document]
paulharter: I've made the calls and file names match

toolbar/src/index.css
paulharter: I've added this via a CDN

paulharter:
@kevin-dp I've pretty much done all the changes as you suggested except:

- I couldn't get the globalRegistry import from the toolbar code so am still passing it in
- I called the api for the typescript client `ToolbarTypescript` rather than `Toolbar`

I have also added the OpenSourceOne font via a CDN

I tried using the pnpm workspace for local dev but you still have go through the build step for every change, it usable but annoying. I have added instructions on setting this up locally in the readme, but I will keep my local dev-app for convenience.

examples/linearlite/src/App.tsx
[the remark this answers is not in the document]
kevin-dp: I don't really get why we are naming the toolbar class after the programming language its implemented in.
What do you mean when you say there could be another? If it is in another language it won't be in this package so it will be imported from somewhere else do i don't see the need to differentiate its name by the programming language.

toolbar/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "codemirror": "^5.65.15",
+    "react": "^18.2.0",
+    "react-codemirror2": "^7.3.0",
+    "react-dom": "^18.2.0"
+  },
+  "devDependencies": {
+    "@types/node": ">=16.11.0",
+    "@types/react": "^18.0.18",
+    "@types/react-dom": "^18.0.11",
+    "electric-sql": "latest",
```
kevin-dp, line 27: Is this a real dev dependency?
It's probably a remnant of the local dev app you use.
But since we don't checkout the dev app we probably don't need this dev dependency.

toolbar/src/api/types.ts
```diff
@@ -0,0 +1 @@
+export type { Row, Statement, ConnectivityState } from 'electric-sql/dist/util'
```
kevin-dp, line 1: We should not import/re-export things from `electric-sql/dist`, `dist` is the distribution folder.
This should be `electric-sql/util`. I would not bother re-exporting it here, you can import it directly from electric in whichever file you need it. Please double check that there are no other places where you import from `electric-sql/dist`.

toolbar/src/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+          </select>
+        </header>
+        <ToolbarTabs dbName={dbName} api={api} />
+      </div>
+      // </div>
+    )
+  }
+}
+
+export function typescriptApi(
```
kevin-dp, line 71: i don't like `typescript` being in the name of this function.

kevin-dp (changes requested):
I left some new comments.
Also, it seems that there are some comments you did not address (e.g. the comments in `toolbar/src/index.tsx`).

Also i just noticed that there are no unit tests.
Perhaps we should add unit tests for the toolbar to check that when you create a toolbar it can access the satellites, and correctly query and reset the DB. We can use the `better-sqlite3` driver in these unit tests. That will give some assurance that the global registry mechanism works.

There is a conflict on the lock file so you will need to update it and push it.

Don't forget to generate a changeset for your PR.

toolbar/package.json
paulharter: I think it need them all - at least it doesn't build without them.

kevin-dp: in `src/api/toolbar-typescript.ts` the global registry is being imported:
```ts
import { GlobalRegistry } from 'electric-sql/satellite'
```

So `electric-sql` needs to be a dependency rather than a dev dependency.
Perhaps it should even be a peer dependency.

kevin-dp: Noticing now that it is listed as a peer-dependency. Why does it need to be a dev dependency also then? Is it being used in some test somewhere?

toolbar/package.json
```diff
@@ -28,7 +28,8 @@
     "esbuild": "^0.16.17",
     "esbuild-plugin-inline-image": "^0.0.9",
     "prettier": "3.0.3",
-    "typescript": "^4.4.3"
+    "typescript": "^4.4.3",
+    "zod": "^3.21.1"
```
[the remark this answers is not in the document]
paulharter: no you're right

toolbar/src/index.tsx
paulharter: I've changed this to `clientApi`

toolbar/src/api/types.ts
paulharter: I've done these import directly now

kevin-dp: but you're still doing `import { Row, Statement, ConnectivityState } from 'electric-sql/dist/util'`.
You should not import from `dist` but just `electric-sql/util`. Same for other imports where you import from `dist`.

paulharter:
https://github.com/electric-sql/electric/pull/686

PR #450, opened 2023-09-20
feat: suffix slot name with db name

icehaunter:
This change allows running electric per db, since slots have to be instance-unique, not database-unique. This is not related to horizontally scaling Electrics against a single PG instance.

### DIFF
--- a/components/electric/lib/electric/replication/connectors.ex
+++ b/components/electric/lib/electric/replication/connectors.ex
@@ -79,7 +79,7 @@ defmodule Electric.Replication.Connectors do
   @spec get_replication_opts(config()) :: replication_opts()
   def get_replication_opts(config) do
     origin = origin(config)
-    database_name = get_connection_opts(config).database
+    database_name = get_in(config, [:connection, :database]) || "test"
 
     config
     |> Keyword.fetch!(:replication)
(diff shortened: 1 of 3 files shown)


components/electric/lib/electric/replication/connectors.ex
```diff
@@ -79,11 +79,12 @@ defmodule Electric.Replication.Connectors do
   @spec get_replication_opts(config()) :: replication_opts()
   def get_replication_opts(config) do
     origin = origin(config)
+    database_name = get_connection_opts(config).database
 
     config
     |> Keyword.fetch!(:replication)
     |> Map.new()
-    |> Map.put(:slot, Extension.slot_name())
+    |> Map.put(:slot, Extension.slot_name() <> "_#{database_name}")
```
magnetised, line 87: guess we're assuming well-formed db names here. are we sure we use quoted names for this puppy all the way down?

magnetised (approved):
on the assumption of simple `^[a-z_]+$` db names this is fine

components/electric/lib/electric/replication/connectors.ex
alco, line 87: I don't know for sure but the DB name may need to be escaped before concatenation.

alco (approved):

PR #452, opened 2023-09-20
Fix migration version in the README

alco:
As reported in the [#bug-reports channel in Discord](https://discord.com/channels/933657521581858818/1079688869852753981/1153973432497750108), we have a [client-side validation](https://github.com/electric-sql/electric/blob/5f44c1a54b06b4e851d695478a284440c9900309/clients/typescript/src/migrators/bundle.ts#L21) that only allows numbers and the underscore to be used in migration versions.

Here I'm also implementing the relaxation of the migration version validation that was proposed along with the bug report.

### DIFF
--- a/.changeset/good-cooks-watch.md
+++ b/.changeset/good-cooks-watch.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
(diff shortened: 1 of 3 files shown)


icehaunter:
Changelog please

alco:
Fixes VAX-1059.

PR #466, opened 2023-09-20
chore(examples): pull latest Electric image

kevin-dp:
This PR completes VAX-1057.
Modified the `package.json` of the different examples to pull the latest Electric image from docker hub.

The starter template is a bit more special though as we should ensure that a project created with `npx create-electric-app` continues to work even if new breaking versions of Electric are published. So, we want to freeze the typescript dependencies as well as the Electric image. Therefore we:
- use `yarn upgrade --exact electric-sql` and `yarn upgrade --exact @electric-sql/prisma-generator` in order to turn `latest` version to that concrete version number in the `package.json`
- use `docker pull electricsql/electric:latest` to pull the latest Electric image
- then inspect that pulled image to get its hash and fix the Electric image to that hash in `backend/compose/.envrc`

Discussion: I have not yet addressed what to do on `yarn upgrade`.
Should we add a `postinstall` hook to the `package.json` of all examples in order to call `yarn electric:pull` ?
(except in the starter because we want to keep it fixed there)


balegas (approved):
I think this makes sense.
Pinning the versions I think is a good initial approach, but we might want to improve it later.

The upside of this approach is that generated projects won't break when updating deps.
The downside is that users won't get updates for minor/patch version, which could potentially work with current pinned docker image.
 - we need to define a convention for releases and make sure that we keep compatibility say for patch releases (but would be a complication right now)
 -  in the future we should document how to upgrade electric client and how to find the right image. This could be a script that get minor electric dep and matches the latest image  with same minor version (I'll move this to a separate task)

PR #467, opened 2023-09-20
chore(electric): Mention extra_apps in mix.exs that are needed to run Observer

alco:
Regarding the num_acceptors configuration, I lowered the number of acceptors for the Postgres TCP server in `prod`. I also lowered the numbers for both Postgres and HTTP listeners in `dev` to make the Applications tab in Observer usable.

### DIFF
--- a/components/electric/config/runtime.dev.exs
+++ b/components/electric/config/runtime.dev.exs
@@ -2,6 +2,10 @@ import Config
 
 config :logger, level: :debug
 
+config :electric,
+  num_http_acceptors: 2,
+  num_pg_acceptors: 2
+
 auth_provider = System.get_env("AUTH_MODE", "secure") |> Electric.Satellite.Auth.build_provider!()
 config :electric, Electric.Satellite.Auth, provider: auth_provider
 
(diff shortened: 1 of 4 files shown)


icehaunter:
As you can see from e2e, adding `:wx` to `extra_applications`  is not the best idea in docker. You can still load :observer just fine with a local incantation loading the app, so I'm not sure we should make this change. Maybe just write the incantation into a function in utils module

icehaunter:
Smth like
```elixir
if Mix.env() != :prod do
def start_observer() do
  Mix.ensure_application!(:wx)
  Mix.ensure_application!(:runtime_tools)
  Mix.ensure_application!(:observer)
  observer.start()
end
end

alco:
@icehaunter It's convenient to have the ability to run Observer in prod. I'm not sure if `:wx` and `:observer` deps are needed for that, but I think `:runtime_tools` definitely needs to be included in Electric itself.

Let me play a bit more with this on the side and redo the setup once I have more info.

Thanks for taking a look!

PR #470, opened 2023-09-20
fix(docs): working prisma link

arve0:
Link to https://www.prisma.io instead of https://github.com/electric-sql/electric/edit/main/generator/www.prisma.io.


thruflo (approved):
Awesome, thanks 👍

Issue #457, opened 2023-09-20
[VAX-1058] Handle indexes in the generator


balegas:
The generator breaks if Pg table has an index as reported [here](https://discord.com/channels/933657521581858818/1079688869852753981/1153841371938045972).

* [Reproducible error repo](https://discord.com/channels/933657521581858818/1079688869852753981/1153841745319182447).

Reported by @hugodutka

balegas:
@kevin-dp when analysing root cause, please make sure necessary changes are valid if we move introspection to Pg. We can handle this issue after that, if necessary.

kevin-dp:
I was able to reproduce the bug and found the cause.
The ts-client fetches the migrations from Electric and then builds them and bundles them in the client.

When building the migrations, the ts-client expects that every `SatOpMigrate` message contains a `table` field of type `SatOpMigrate_Table`. However, that field is not present if the migration only contains an index creation.

I can see 2 solutions:
1) The TS-client checks whether the message contains a table field. If not, it ignores the migration altogether
2) Electric makes sure to also send the table information for index creations

Maybe we even want to do both.
1) will be needed anyway because the Satellite protocol defines the `SatOpMigrate.table` field to be optional.
So it must correctly handle the absence of that field. Unless we would make that field required in the protocol.

magnetised:
@kevin-dp why do we need the table information in every migration? a create index operation doesn't affect any table structure, so what's the requirement for table information?

kevin-dp:
@magnetised Totally right. The current ts-client naively thinks that every migration changes at least some table structure and thus needs that information to add/modify any triggers. I will remove that assumption from the client and simply not try to build any triggers for it if the table information is absent.

PR #471, opened 2023-09-21
fix (client): solve migration bug with indexes

kevin-dp:
This PR solves the index bug described in #457.
The bug was caused by the client migration code which tried to build table triggers even if the table information is missing. The code was modified not to build any triggers if the table information is missing. The table information may be missing e.g. when only creating an index.

### DIFF
--- a/.changeset/funny-terms-knock.md
+++ b/.changeset/funny-terms-knock.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Do not build triggers if table information is missing.
--- a/.changeset/funny-terms-knock.md
+++ b/.changeset/funny-terms-knock.md
@@ -2,4 +2,4 @@
 "electric-sql": patch
 ---
 
-Do not build triggers if table information is missing.
+Fix the bug where the client would crash/stop working/stop syncing if it received a migration containing a new index creation.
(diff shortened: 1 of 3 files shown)


.changeset/funny-terms-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Do not build triggers if table information is missing.
```
alco, line 5: I may be nitpicking here but I think these changelog entries should make sense to anyone looking at the release notes when they are upgrading to a new version.

As is, this description does not contain any more info than the commit message itself.

I would maybe rephrase it as

> Fix the bug where the client would crash/stop working/stop syncing if it received a migration containing a new index creation.

clients/typescript/src/migrators/builder.ts
```diff
@@ -74,6 +74,9 @@ export function makeMigration(migration: MetaData): Migration {
     .map((op) => op.stmts.map((stmt) => stmt.sql))
     .flat()
   const tables = migration.ops
+    // if the operation did not change any table
+    // then ignore it as we don't have to build triggers for it
+    .filter((op) => op.table !== undefined)
```
balegas, line 79: @kevin-dp it's approved. up to you to add a test before merging

PR #476, opened 2023-09-21
Add comment about possibly receiving empty subsData

davidmartos96:
On the Dart client we discovered a bug, where only the second time the app is open after an initial sync, electric will try to apply Inserts again to the local database.

We discovered the problem was due to interpreting `if (subsData.data) {`  as `if (subsData.data.length !== 0)` because the `data` property cannot be null/undefined. 

The problem is that not calling this after the initial sync, makes electric not store the subscription and lsn to the metadata table, so a second open will try to refetch again and apply again, causing UNIQUE constraint errors.

What I did is to remove the if statement, which can be confusing due to the empty arrays being truthy and because executing that code is necessary even when subsData is empty.


davidmartos96:
For context, when the data is empty, these are the statements that get executed:

image

kevin-dp:
LGTM.
@balegas what's your take?

balegas:
It makes total sense to always call `_applySubscriptionData` because that is where we save the metadata info.

I'm sorry for taking so long to look into this.

balegas (approved):

balegas:
@davidmartos96 needs a changeset file :)

davidmartos96:
@balegas Done!

PR #483, opened 2023-09-22
chore(electric): Remove the sending of the :lsn_ack message that nobody receives

alco:
This is a leftover from the work done on VAX-911 (GitHub PR #386). It should have been removed during a rebase of the PR, since pings and acks had already been removed from main by then, see
https://github.com/electric-sql/electric/pull/408/files#diff-b0cea52cb9ec25d24558bed7297b456f66bd57a5080ef259d568bc3bffc05093L144-L149.


magnetised (approved):
hard to review the why without more context but the how is definitely sound

PR #490, opened 2023-09-23
fix : broken link in frontend integrations

sebws:
link in docs has a typo in it, breaking the link


thruflo (approved):
Good spot, thanks for this 🙏

PR #494, opened 2023-09-25
chore(electric): Move the definition of create_active_migration() out of migrations

alco:
At least I think it was an accidental change when the updated version of the `create_active_migration()` function was introduced in https://github.com/electric-sql/electric/pull/439/files#diff-578136eb5d1250bc92bc4ccdde42fdeea6bed399bb1f22f4d621074813070932R35.

### DIFF
--- a/components/electric/lib/electric/postgres/extension/functions.ex
+++ b/components/electric/lib/electric/postgres/extension/functions.ex
@@ -34,6 +34,12 @@ defmodule Electric.Postgres.Extension.Functions do
 
   @function_names function_names
 
+  # https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-PG-SNAPSHOT
+  # pg_current_xact_id() -> xid8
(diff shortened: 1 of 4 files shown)


magnetised (commented):
@alco Actually I made this change deliberately to inform the user who's making the migration that it was captured properly by electric -- useful output in the case where there's a problem, at least there's some output telling us that the migration was captured so we can eliminate that doubt. if we just raise a DEBUG then there's no information we can use when debugging without entering into a dialogue asking for the user to query our metadata tables etc.

alco:
@magnetised Thanks for the context! I see no problem leaving it as is then.

I would still like to merge the change that extracts the updated function definition from an extension migration and runs the function definition at run time. It serves as a nice show case of how we can define all future functions outside of extension migrations from here on out.

magnetised (commented):
super. thanks

magnetised (approved):

PR #499, opened 2023-09-26
feat(examples): New Ionic demo app - Appointments

samwillis:
This is a demo combining Electric with Ionic - a simple Appointment/Booking app.

The Ionic CLI was used to start an Ionic app, I then manually added the scaffolding from our starter. This process can be seen in the first two commits.

**I have purposely squashed this into three commits showing the initialisation of the project, I think we should rebase and merge rather than squash and commit so that this history is maintained and can be linked to**

It now uses the latest Electric with the proxy, and exercises the new `uuid`, `timestamp` and `bool` type support.

Testing this PR requires running against a locally built Electric, ts-client, and generator.

- modify your app to use the local Electric image (because it has not yet been published to Docker hub)
i.e. in ionic-demo/backend/compose/.envrc change the `ELECTRIC_IMAGE` env var to be `electric:local-build`.
- modify your app to use our packed ts-client:
change the electric-sql dependency in ionic-demo/package.json to be `"electric-sql": "file:<path to electric mono  repo>/clients/typescript/electric-sql-0.6.4-2.tgz"`,
change the @electric-sql/prisma-generator dependency in ionic-demo/package.json to be `"@electric-sql/prisma-generator": "file:<path to electric mono  repo>/generator/electric-sql-prisma-generator-1.0.2-regex.tgz"`,
- now run the app:
```
npm install
npm run backend:start
npm run db:migrate
npm run client:generate
npm run start
```

You can follow the instruction in the readme to build and run on iOS or Android.

Android support depends on https://github.com/electric-sql/electric/pull/571


samwillis:
I'm suspicious there is a bug in the underlying capacitor/sqlite plugin (not our driver).

I have had to modify our capacitor/sqlite driver to not use `beginTransaction()` and `.commitTransaction()` and instead do explicit `BEGIN` and `COMMIT`, I was getting errors about concurrent transactions, even after adding a mutex to all calls matching the style from our wa-sqlite driver.

iOS was fine, it was only apparent on Android...

This probably needs further investigation and Andy changes to the diver splitting into a separate PR.

samwillis:
I believe I have confirmed that there is a bug in the underlying Capacitor SQLite plugin, issue report here: https://github.com/capacitor-community/sqlite/issues/478

examples/ionic-demo/android/app/src/test/java/com/getcapacitor/myapp/ExampleUnitTest.java
```diff
@@ -0,0 +1,18 @@
...
+import static org.junit.Assert.*;
+
+import org.junit.Test;
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+public class ExampleUnitTest {
```
kevin-dp, line 12: Do we need these example unit tests?

samwillis: These are all from the Ionic CLI generator. There is no custom code in the test, iOS, or android directories.

examples/ionic-demo/cypress/e2e/test.cy.ts
```diff
@@ -0,0 +1,6 @@
+describe('My First Test', () => {
```
kevin-dp, line 1: Do we need this test?

examples/ionic-demo/android/app/src/test/java/com/getcapacitor/myapp/ExampleUnitTest.java
kevin-dp: So we can remove the test, right?

samwillis: Yes, we can. However, we may not want to, it depends on how much we want to alter the generated code from the Ionic Capacitor template. Understand completely it looks like unnecessary cruft, but someone may want to checkout out demo and use it as the basis for building an app. Leaving in all the scaffolding, even if we don't use it, could be helpful for people taking the example as basis to work on.

kevin-dp:
It would be useful for other reviewers to add the steps that are needed to run the app.
Because the app relies on the proxy but the docker-compose uses the published Electric image so it will not work out of the box for reviewers. So need to change .envrc to use local build of Electric. Also, after doing that `backend:start` won't work because Electric complains that `PG_PROXY_PASSWORD` is not set. So you probably forgot to add that to the compose file.

examples/ionic-demo/src/pages/New.tsx
```diff
@@ -0,0 +1,132 @@
...
+        const optionEnd = new Date(
+          date.getFullYear(),
+          date.getMonth(),
+          date.getDate(),
+          h,
+          m + 15,
+        )
+        let disabled = false
+        if (optionStart < new Date()) {
+          // In the past
```
kevin-dp, line 68: Couldn't it be ongoing?

examples/ionic-demo/src/pages/New.tsx
```diff
@@ -0,0 +1,132 @@
...
+          disabled = true
+        }
+        // Check for clashes
+        if (results && !disabled) {
+          disabled = results.some((appointment) => {
+            const appointmentStart = new Date(appointment.start)
+            const appointmentEnd = new Date(appointment.end)
+            if (
```
kevin-dp, line 76: replace L76-82 by `return optionStart.getTime() < appointmentEnd.getTime() &&
              optionEnd.getTime() > appointmentStart.getTime()`

examples/ionic-demo/src/pages/New.tsx
```diff
@@ -0,0 +1,132 @@
...
+      },
+      orderBy: {
+        start: 'asc',
+      },
+    }),
+  )
+
+  const timeOptions = useMemo(() => {
+    const timeOptions = []
+    for (let h = 8; h < 18; h++) {
```
kevin-dp, line 48: Not sure why the grid is hardcoded to be between 8h and 18h.
Does this mean it cannot be customized to other hours?
Can i see events that fall out of that time range?

examples/ionic-demo/android/app/src/test/java/com/getcapacitor/myapp/ExampleUnitTest.java
kevin-dp: Right, can see that use case

kevin-dp (approved):
Awesome app!
I left some comments but nothing major.
Please extract the changes to the driver to a separate PR before merging though.

examples/ionic-demo/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "client:generate": "npx electric-sql generate",
+    "client:watch": "npx electric-sql generate --watch",
+    "db:migrate": "node ./db/migrate.js",
+    "db:psql": "node ./db/connect.js",
+    "reset": "npm run backend:down && npm run backend:up && sleep 5 && npm run db:migrate && npm run client:generate",
+    "electric:start": "node ./backend/startElectric.js",
+    "prepare-wa-sqlite": "cp ./node_modules/wa-sqlite/dist/wa-sqlite-async.mjs ./public/wa-sqlite-async.mjs && cp ./node_modules/wa-sqlite/dist/wa-sqlite-async.wasm ./public/wa-sqlite-async.wasm",
+    "dev": "vite",
+    "start": "npm run prepare-wa-sqlite && npm run dev",
+    "build": "npm run prepare-wa-sqlite && tsc && vite build",
```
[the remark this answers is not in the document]
samwillis: Ah, yes. The version you tested linked against the workspace electric rather than the published one, due to the bug with Android and my changes. I will change to link to the published version.

That fixes this and the problem you got `running npm run start`

On `electric-sql/version`, I believe that was introduced in v0.6. We should fix types types.

examples/ionic-demo/src/pages/New.tsx
samwillis: I did this for simplicity, in a "real" scheduling app the "owner" would be able to set the business hours. In the calendar view, yes you can see events outside this window. A "customer" can only book during business hours.

(Note that I have done nothing to address time zones in this app, that would also need solving for a "read" app)

PR #500, opened 2023-09-26
chore(electric): Use a PG connection pool instead of opening a new connection for every query

alco:
Closes VAX-1035.

* [ ] Test error handling inside ConnectionPool: failing query, transaction rollback, transaction isolation level change, etc.
* [ ] Investigate the new kind of E2E test failures - https://github.com/electric-sql/electric/actions/runs/6323249430/job/17170446899?pr=500
* [ ] Investigate the new failures in ElectrificationTest that are caused by the initial call to `CachedWal.Api.current_position()` returning `nil`.

### DIFF
--- a/components/electric/lib/electric/postgres/connection_pool.ex
+++ b/components/electric/lib/electric/postgres/connection_pool.ex
@@ -32,11 +32,19 @@ defmodule Electric.Postgres.ConnectionPool do
     NimblePool.checkout!(
       pool,
       :checkout,
-      fn _pool, conn ->
-        {fun.(conn), :ok}
-      end,
+      fn _pool, conn -> {execute_fun_with_conn(fun, conn), :ok} end,
       @pool_timeout
     )
+  end
+
+  defp execute_fun_with_conn(fun, conn) do
+    try do
+      fun.(conn)
+    rescue
+      e ->
+        Logger.error(Exception.format(:error, e, __STACKTRACE__))
+        {:error, e}
+    end
   end
 
   @spec name(Connectors.origin()) :: Electric.reg_name()
--- a/components/electric/lib/electric/postgres/connection_pool.ex
+++ b/components/electric/lib/electric/postgres/connection_pool.ex
@@ -18,7 +18,7 @@ defmodule Electric.Postgres.ConnectionPool do
       worker: {__MODULE__, conn_config},
       # only connect when required, not immediately
       lazy: true,
-      pool_size: 4,
+      pool_size: 20,
       worker_idle_timeout: 30_000,
       name: name(Connectors.origin(conn_config))
     )
--- a/components/electric/lib/electric/postgres/connection_pool.ex
+++ b/components/electric/lib/electric/postgres/connection_pool.ex
@@ -1,0 +1,89 @@
+defmodule Electric.Postgres.ConnectionPool do
+  @moduledoc false
+
+  @behaviour NimblePool
+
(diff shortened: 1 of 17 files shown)


components/electric/lib/electric/postgres/connection_pool.ex
```diff
@@ -0,0 +1,89 @@
...
+    Logger.debug("Starting SchemaLoader pg connection: #{inspect(conn_config)}")
+    # NOTE: use `__connection__: conn` in tests to pass an existing connection
+    {:ok, conn} =
+      case Keyword.fetch(conn_config, :__connection__) do
+        {:ok, conn} ->
+          {:ok, conn}
+
+        :error ->
+          conn_config
+          |> Connectors.get_connection_opts(replication: false)
```
magnetised, line 60: I think this `replication: false` may be causing problems. previously the default connection params would include create a connection in replica mode whereas with this pool approach you're always connecting in non-replica mode.

components/electric/lib/electric/replication/postgres_manager.ex
```diff
@@ -151,15 +157,13 @@ defmodule Electric.Replication.PostgresConnectorMng do
...
 
-  defp start_subscription(%State{conn_config: conn_config, repl_config: rep_conf} = state) do
-    case Client.with_conn(
-           conn_config,
-           fn conn ->
-             Client.start_subscription(conn, rep_conf.subscription)
-           end
-         ) do
+  defp start_subscription(%State{origin: origin, repl_config: rep_conf}) do
+    ConnectionPool.checkout!(origin, fn conn ->
```
magnetised, line 161: so I think this is where the `replication: false` bites you. might be an argument that you don't need the connection pool here as this is a long-running connection. it's the transient connections, which are all replication: false, which need the pool

components/electric/lib/electric/replication/postgres_manager.ex
```diff
@@ -181,14 +185,14 @@ defmodule Electric.Replication.PostgresConnectorMng do
       "Attempting to initialize #{origin}: #{conn_config.username}@#{conn_config.host}:#{conn_config.port}"
     )
 
-    Client.with_conn(conn_config, fn conn ->
```
magnetised, line 184: this is replication: true but I'm not sure it needs to be... there's things we weren't able to do in the extension migration stuff because the connection was replica: false but not sure if turning off replication causes other problems

magnetised (commented):
Think I've found the cause of the new issues

components/electric/lib/electric/replication/postgres_manager.ex
alco: It's good that you mention this, the replication option has slipped my mind while I was working on the pool. Unfortunately, I don't think that's the problem here. Both tyimes PostgresManager needs to connect to the DB, it runs regular SQL statements: `ALTER SUBSCRIPTION`, `SELECT`. The long-running replication connection is set up in `LogicalReplicationProducer.init()`, I haven't touched that code.

I took a closer look at the failing E2E tests. There are two types of recurring issues there: (1) missing output to match on (but I think it's the test code that's wrong there) and (2) "Received subscribe response for unknown subscription" error.

The former must have been caused by different scheduling of concurrent processes due to the introduction of the connnection pool, I can fix that in the test code. The latter is an actual new problem I need to look more closely at. Probably caused by the same connection being used to run different queries.

PR #502, opened 2023-09-27
fix(examples/starter): Fix the issue where ConnectionPool would be disposed of after running the 1st migration

alco:
Fixes VAX-1113.

@kevin-dp How can we unit-test this?

### DIFF
--- a/examples/starter/template/db/migrate.js
+++ b/examples/starter/template/db/migrate.js
@@ -14,7 +14,7 @@ const apply = async (fileName) => {
   const filePath = path.join(MIGRATIONS_DIR, fileName)
   console.log('Applying', filePath)
 
-  await db.tx(
+  db.tx(
     (tx) => tx.query(
       sql.file(filePath)
     )


kevin-dp:
Thanks for catching this!
I don't understand though why awaiting it twice doesn't work.

Minor comment: I would keep `await` on line 17 and remove the one on line 28 such that we do not need to remember to await everytime we call the async function. What do you think?

alco:
@kevin-dp 

> Minor comment: I would keep `await` on line 17 and remove the one on line 28 such that we do not need to remember to await everytime we call the async function. What do you think?

A different way to look at it is this: the function `apply()` is defined as `async`, so someone looking to call it would assume it can be `await`ed on, no? I'm completely out of my depth here, so I don't really have any preference. Maybe the whole approach could be restructured such that each migration is evaluated into a Promise and then there's a single `Promise.all` call to resolve them all instead of doing `await` in a loop.

kevin-dp:
I think it is best to keep awaiting in a loop because there may be cases where a migration depends on one or more previous migrations. So we need to ensure that they are applied in order. I know that this is currently not yet the case, we could apply the migrations in the lexicographical order of their filenames. We can make an issue for that.

For now, i took another look at the code and i think that the cleanest solution that circumvents the problem would be to not define apply as an async function:

```
const apply = (fileName) => {
  const filePath = path.join(MIGRATIONS_DIR, fileName)
  console.log('Applying', filePath)

  return db.tx(
    (tx) => tx.query(
      sql.file(filePath)
    )
  )
}
```

What do you think?

kevin-dp (approved):

alco:
Superseded by https://github.com/electric-sql/electric/pull/520.

Issue #489, opened 2023-09-22
[VAX-1074] Database does not exist / support running the generator twice


thruflo:
Running the generator twice to create two different projects runs into a database name not found error.

As per report at [https://discord.com/channels/933657521581858818/1079688869852753981/1154842122571223091](https://discord.com/channels/933657521581858818/1079688869852753981/1154842122571223091)

<sub>From [SyncLinear.com](https://synclinear.com) | [VAX-1074](https://linear.app/electric-sql/issue/VAX-1074/database-does-not-exist-support-running-the-generator-twice)</sub>

barbalex:
I am having a similar issue which makes it impossible (?) for me to start working with electric.

What I have done:
```
npx create-electric-app@latest capturing_electric_2
cd capturing_electric_2
npm i
npm run backend:start
npm run db:migrate
```

The result:
```
PS C:\Users\alexa\capturing_electric_2> npm run backend:start

> capturing_electric_2@0.1.0 backend:start
> docker compose --env-file backend/compose/.envrc -f backend/compose/docker-compose.yaml up

[+] Running 2/2
 ✔ Container compose-postgres-1  Created                                                                                                                                                                                                                                  0.4s 
 ✔ Container compose-electric-1  Recreated                                                                                                                                                                                                                                0.2s 
Attaching to compose-electric-1, compose-postgres-1
compose-postgres-1  | 
compose-postgres-1  | PostgreSQL Database directory appears to contain a database; Skipping initialization
compose-postgres-1  | 
compose-postgres-1  | 
compose-postgres-1  | 2023-09-25 12:02:39.353 GMT [1] LOG:  starting PostgreSQL 14.9 on x86_64-pc-linux-musl, compiled by gcc (Alpine 12.2.1_git20220924-r10) 12.2.1 20220924, 64-bit
compose-postgres-1  | 2023-09-25 12:02:39.353 GMT [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
compose-postgres-1  | 2023-09-25 12:02:39.353 GMT [1] LOG:  listening on IPv6 address "::", port 5432
compose-postgres-1  | 2023-09-25 12:02:39.382 GMT [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
compose-postgres-1  | 2023-09-25 12:02:39.389 GMT [22] LOG:  database system was shut down at 2023-09-21 10:04:56 GMT
[...]

PR #505, opened 2023-09-28
feat(starter): starter improvements to avoid port clashes

### LINKED ISSUE #497: [VAX-1087] Prefix starter template container name with app name.
As per [https://github.com/electric-sql/electric/issues/489](https://github.com/electric-sql/electric/issues/489) et al, we currently have a clash of container names from multiple invocations of the starter template.

One fix is to prefix the container name in the generated docker compose with the app name. Basically to stop Docker from reusing the same container / volume.

<sub>From [SyncLinear.com](https://synclinear.com) | [VAX-1087](https://linear.app/electric-sql/issue/VAX-1087/prefix-starter-template-container-name-with-app-name)</sub>

kevin-dp:
This PR improves the starter to enable several (independent) Electric applications to run on the same machine and thus fixes #497 and #489. Also addressed VAX-1114.

The changes are as follows:
- Avoid volume clashes by putting project name on the app name
- Avoid web server port clashes by supporting a custom port to be used and fix it everywhere in the template
- Avoid Electric port clashes by supporting a custom port to be used and fix it everywhere in the template
- Introduced 2 modes for the starter: fast mode and interactive mode
  - Fast mode: app name and ports (optional) are provided as command line arguments
  - Interactive mode: no arguments to the script, app name and ports are asked via prompts
- Changed start to check that the chosen port for Electric is free. If not, prompt user if they want to change the port
- Script to change the project’s configuration with other Electric and web server ports. Exposed as a new `ports:configure` command.
- Changed `backend:start` to check for port clashes and if there is a port clash suggest to run `ports:configure`.
- Changed `yarn start` and `client:generate` to check that the Electric port the app is configured to use is the app's own Electric service (e.g. catches problem where project 1 is running Electric on port 5133 and another project is accidentally configured to also connect to port 5133).
- Modified build script to use forward requests to right esbuild server. This fixes a bug where if an esbuild server was already running, the new app would forward its requests to the esbuild server of the other app.

tldr: it should now be possible to run several Electric applications. Port clashes should always be detected and reported with a suggestion to reconfigure the app with other ports.


examples/starter/template/backend/startElectric.js
```diff
@@ -1,35 +1,72 @@
...
 
-shell.exec(
+const res = shell.exec(
   `docker run \
       -e "DATABASE_URL=${db}" \
       -e "LOGICAL_PUBLISHER_HOST=localhost" \
       -e "AUTH_MODE=insecure" \
-      -p 5133:5133 \
+      -p ${electricPort}:5133 \
       -p 5433:5433 ${electric}`
```
kevin-dp, line 57: Do we still need to expose this port?
The compose file is not binding to 5433 so i'm assuming it is not necessary and we could remove it here?

kevin-dp, line 57: Do we still need to expose this port?
The compose file is not binding to 5433 so i'm assuming it is not necessary here either and we can remove it?

examples/starter/template/backend/startElectric.js
```diff
@@ -1,35 +1,72 @@
...
+    default:
+      error(`Unrecognized option: '${flag}'.`)
+  }
 }
-else if (process.argv.length !== 2) {
-  console.log('Wrong number of arguments provided. Only one optional argument `-db <Postgres connection url>` is supported.')
+
+function parseElectricPort(port) {
+  // checks that the number is between 0 and 65535
+  const portRegex = /^([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$/
```
alco, line 37: It would have been much simpler to convert the port to a number and check that it is within bounds.

examples/starter/template/backend/startElectric.js
alco: 5433 is the logical publisher port. I think it is being exposed here to support setup where people are running their own postgres on the host while Electric itself is run inside a Docker container.

thruflo (approved):
Looks great 👍

examples/starter/template/backend/startElectric.js
alco: Also, why is the port parsing code implemented multiple times: here, in examples/starter/src/index.ts, in examples/starter/template/change-ports.js?

examples/starter/template/backend/startElectric.js
kevin-dp: Oh yes, i see. Perhaps i'll leave a comment to remember.
No the script for running only Electric is this separate file that executes `docker run` because there are some differences (PG url can be passed as an argument, and 5433 is not exposed in compose file).

examples/starter/template/backend/startElectric.js
kevin-dp: > Also, why is the port parsing code implemented multiple times: here, in examples/starter/src/index.ts, in examples/starter/template/change-ports.js?

Because those are separate packages.
`examples/starter/src/index.ts` lives in the `create-electric-app` package, whereas, `examples/starter/template/change-ports.js` lives in the template which will become part of the generated app (thus another package).

kevin-dp:
Btw: at least someone else than me should try this out to see that everything works fine.
Try to break it with port clashes etc. and check that they get caught and that we can fix them :-)

samwillis:
A few quick thoughts, `yarn` is not installed with vanilla node, I know that predates this PR. We should use `npm run`.

*(Aside: I feel strongly that all starters, examples and docs need to just use plain npm. Anyone that uses yarn or pnpm knows how to rewrite a command for it.)*

If we do change the starter to use Vite as discussed on Discord earlier, the code for setting the dev web server port is not needed. Vite will enumerate ports until it finds one free and print it. (still need logic for electric)

I also feel that having the change port command rewrite the package.json and other committed files, isn't right. Different devs on will want to use different ports. Local config should not be in committed files.

Should we not have an optional *not committed* `.electric` file that holds things like port numbers for the local project environment?

kevin-dp:
> A few quick thoughts, `yarn` is not installed with vanilla node, I know that predates this PR. We should use `npm run`.
> 
> _(Aside: I feel strongly that all starters, examples and docs need to just use plain npm. Anyone that uses yarn or pnpm knows how to rewrite a command for it.)_

I had a similar opinion but we decided to stick with yarn everywhere a while ago.
If needed we can re-discuss it and change this.
But that will be a separate PR.

> If we do change the starter to use Vite as discussed on Discord earlier, the code for setting the dev web server port is not needed. Vite will enumerate ports until it finds one free and print it. (still need logic for electric)

Yep but should be a separate PR, so for now best to do this such that the starter works with several apps.

> I also feel that having the change port command rewrite the package.json and other committed files, isn't right. Different dev will want to use different ports. Local config should not be in committed files.
> 
> Should we not have an optional _not committed_ `.electric` file that holds things like port numbers for the local project environment?

This has been a bit of a rabbit hole. We have discussed and iterated on several strategies this week.
We decided to stick with this approach of fixing the ports everywhere in the files.
If there is a better solution we can discuss and implement it later but we should probably move forward with other features first.

alco:
Notes from my testing:
1. The output of `yarn backend:start` is no longer colored. Also, there are now lines that look like this

       Network app_1_default  Creating
       Network app_1_default  Created

   Whereas before the line would be updated in-place. It seems that the command is now treating the terminal as a dumb one and skips employing advanced features such as line rewriting and the use of ANSI escapes for colors.
2. After starting the backend and running `yarn db:migrate`, when I then run `yarn client:generate` it fails:

       /bin/sh: line 1: electric:check: command not found
       error Command failed with exit code 127

   The `package.json` file includes a definition for the `"electric:check"` script. The problem is probably with the way `"client:generate"` invokes it:

       "client:generate": "electric:check && npx electric-sql generate [...]"

   I managed to make it work by prefixing the command with `yarn`:

       "client:generate": "yarn electric:check && npx electric-sql generate [...]"

alco:
Re point 1 above, here's a side-by-side diff between `main` and this PR:
Screenshot from 2023-09-29 12-32-16

Weirdly enough, when I run `yarn backend:down`, the output is colored in both versions.

kevin-dp:
@alco Output of `yarn backend:start` is no longer colored because we now invoke `docker compose up` from a script that uses shelljs to invoke it and shelljs still does not support coloring...

alco:
> @alco Output of `yarn backend:start` is no longer colored because we now invoke `docker compose up` from a script that uses shelljs to invoke it and shelljs still does not support coloring...

I have found this package which is a lightweight wrapper of `shelljs` that supposedly adds coloring. Could we use it instead?
https://github.com/shelljs/shelljs/issues/86#issuecomment-939378750

examples/starter/template/backend/startElectric.js
kevin-dp: After discussing this with @alco I decided to keep the regex as it is more strict.
e.g. `Number("000123") = 123` which i don't like.

kevin-dp:
@alco 

2. Thanks! Not sure how this went unnoticed.
5. Indeed, the esbuild server runs on port 8000 by default. If it is already taken it will take another available port with preference for ports in the range 8000 to 8009. Not sure why this matters though? I made sure that the app's `builder.js` script uses whatever port that was chosen by esbuild.
6. This is actually good! It did detect the problem and wanted to raise it as an error but somehow the `error` function is not yet initialized (because it is an arrow function assigned to a constant, i will turn it into a real function to fix this).

I will also give the wrapper for shelljs a try.

kevin-dp:
> > @alco Output of `yarn backend:start` is no longer colored because we now invoke `docker compose up` from a script that uses shelljs to invoke it and shelljs still does not support coloring...
> 
> I have found this package which is a lightweight wrapper of `shelljs` that supposedly adds coloring. Could we use it instead? [shelljs/shelljs#86 (comment)](https://github.com/shelljs/shelljs/issues/86#issuecomment-939378750)

Yep had also seen it but was not sure about its stability and maturity though.
So I now gave it a try and it works flawlessly.

alco (approved):
Top notch! :trophy: 

I've pushed one commit as a suggestion, feel free to keep it or remove it.

Issue #507, opened 2023-09-28
Client unable to sync up local writes to boolean or float columns


alco:
This is a known issue that will be addressed in one of the upcoming patch releases.

The problem happens for tables that are electrified and then included in the client as a bundled migration by running `yarn db:migrate` or `npx electric-sql generate`, depending on your setup.

When a new record with a boolean column is then created on the client, it will break the replication stream such that no further local writes will be able to sync up. A workaround for testing sync with boolean columns is to electrify a table in Postgres and then use `electric.db.sql.raw({sql: "INSERT ... INTO ...", args: [...]})`.

For float columns, the same error state is reached when trying to sync up a whole number, i.e. one that does not have any places after the decimal point. There's no workaround for this issue but we're working on releasing a fix soon.

Related to VAX-842 and VAX-1078.

gorbak25:
@alco We had this issue and we also were attributing this issue to floats. The problem is not with floats but with CamelCased columns. A table with a float column named "asdf" will work fine but a table with a "CreatedAt" float column fails to sync to postgres.

balegas:
@alco can we close this issue after your changes?

alco:
@balegas booleans still aren't working.

alco:
This has been fixed in version 0.7.0.

PR #508, opened 2023-09-29
fix(satellite): Ensure decimal point in real number encoding

alco:
Fixes VAX-1078.

Related: https://github.com/electric-sql/electric/issues/507


clients/typescript/src/util/common.ts
```diff
@@ -66,6 +67,18 @@ export function numberToBytes(i: number) {
   )
 }
 
+export function realToBytes(num: number) {
+  let num_str = ''
+  if (Math.trunc(num) === num) {
+    // num is an integer, we need to explicitly append the ".0" to it.
+    num_str += num + '.0'
```
kevin-dp, line 74: No need to use `+=`, regular assignment will do just fine

kevin-dp (approved):
LGTM 👍

clients/typescript/src/util/common.ts
alco: Hey, thanks for this suggestion. I've refactored this bit to make it look even cleaner.

PR #509, opened 2023-09-29
chore(electric): Reject tables that have columns with DEFAULT expressions

alco:
Fixes VAX-1100.

* [x] Add tests
* [x] Update the type-checking function to use `%I`
* [x] Create a follow-up issue in Linear.

### DIFF
--- a/components/electric/lib/electric/postgres/extension/functions/validate_table_column_defaults.sql.eex
+++ b/components/electric/lib/electric/postgres/extension/functions/validate_table_column_defaults.sql.eex
@@ -1,8 +1,8 @@
 -- This function validates each column of the table that's being electrified
 -- and aborts electrification if any column has DEFAULT expression.
 
-CREATE OR REPLACE FUNCTION <%= @schema %>.__validate_table_column_defaults(table_name text)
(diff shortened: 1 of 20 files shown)


gorbak25:
@alco I think this PR might be too strict. I think a better approach would be to remove default expressions from sqlite and keep them in progress. The prisma client can ensure those fields with default expressions are not optional on the client side.

components/electric/lib/electric/postgres/extension/functions/validate_table_column_defaults.sql.eex
```diff
@@ -0,0 +1,27 @@
+-- This function validates each column of the table that's being electrified
+-- and aborts electrification if any column has DEFAULT expression.
+
+CREATE OR REPLACE FUNCTION <%= @schema %>.__validate_table_column_defaults(table_name text)
```
magnetised, line 4: should be a procedure since it doesn't return anything, and also do include the `SECURITY DEFINER` clause

magnetised (approved):
Looks good apart from function->procedure switch

PR #511, opened 2023-09-29
chore(electric): Clean up templated SQL routines

alco:
This is a cleanup PR that includes the following changes:
* Remove some of the dead SQL code from Extension migration modules. The code used to define routines but they are now defined at run time.
* Import the Extension module into Functions to make all functions from the former available automatically to all templated SQL routines. This removes the need to maintain a map of assigns by hand.
* Fix the wildcard pattern in Functions to include subdirectories. Prior to this change, the functions from `lib/electric/postgres/extension/functions/ddlx/` weren't included in the module.


components/electric/lib/electric/postgres/extension/functions.ex
```diff
@@ -3,24 +3,28 @@ defmodule Electric.Postgres.Extension.Functions do
...
 
-  function_names =
+  function_paths =
     for path <- sql_files do
       @external_resource path
 
+      fpath = Path.relative_to(path, Path.expand("./functions", __DIR__))
       name = path |> Path.basename(".sql.eex") |> String.to_atom()
-      _ = EEx.function_from_file(:def, name, path, [:assigns])
+      _ = EEx.function_from_file(:def, name, path, [])
```
magnetised, line 21: since we're looking at this module, why are we using `function_from_file`  rather than `compile_file`? I don't see the need for these sql functions to be available as actual functions. we could use a simple map and `compile_file` and keep the same functionality as this (with the `by_name/1` invocation). 

We could even lazily parse and evaluate the templates on-demand to reduce the compiled code.

magnetised (approved):
😓 thanks for the hard work on this one! good changes all in all, but I do question the need for the functions-as-functions approach considering the required lifetime of these blobs of rendered sql.

components/electric/lib/electric/postgres/extension/functions.ex
alco: Nice idea about lazy evaluation. To make that work, we'll have to move the template files into `priv` because as is they are not bundled with the Elixir release. That was part of the motivation to compile them all into the Elixir module, just so we can forget about the template files themselves and move on.

alco: @magnetised I've updated this PR to load SQL functions at run time. The templates had to be moved to `priv/sql_function_templates` to make them available in releases.

components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers.ex
```diff
@@ -19,11 +19,6 @@ defmodule Electric.Postgres.Extension.Migrations.Migration_20230512000000_confli
   def up(schema) do
     [
       @contents["electric_tag_type_and_operators"],
-      @contents["utility_functions"],
```
magnetised, line 22: I'm guessing that this is ok and referencing functions that don't exist is not a problem -- if so then good to get rid of the duplication

magnetised (approved):
lovely stuff. Shame git didn't track the renames.

components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers.ex
alco: It depends on the context. If you have a function that calls another function, it is fine for the former to be defined _before_ the latter.

On the other hand, if you have a trigger definition that references a function, the function has to be defined _prior_ to the evaluation of the trigger def.

PR #515, opened 2023-10-02
fix (generator): correctly capitalise table named model

kevin-dp:
This PR fixes the bug for a table named "model", described by a user on [Discord](https://discord.com/channels/933657521581858818/933657523049885698/1156342498453639260).
We were parsing the Prisma schema in order to capitalise model names but if the model name itself was "model" we would capitalise the "model" keyword instead of the model name "model". This is fixed by replacing the exact match in the string instead of replacing the first occurrence of the table name.

Also extended the unit test to check this case.

### DIFF
--- a/clients/typescript/src/cli/migrations/migrate.ts
+++ b/clients/typescript/src/cli/migrations/migrate.ts
@@ -309,7 +309,8 @@ export function doPascalCaseTableNames(lines: string[]): string[] {
   const replacements: Map<string, string> = new Map() // maps table names to their PascalCased model name
   const modelNameToDbName: Map<string, string> = new Map() // maps the PascalCased model names to their original table name
 
-  const getModelName = (ln: string) => ln.match(/^\s*model\s+(\w+)/)?.[1]
+  const modelRegex = /^\s*model\s+(\w+)\s*{/
+  const getModelName = (ln: string) => ln.match(modelRegex)?.[1]
 
   lines.forEach((ln, idx) => {
     const tableName = getModelName(ln)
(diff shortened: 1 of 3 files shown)


clients/typescript/test/cli/migrations/migrate.test.ts
```diff
@@ -97,11 +101,17 @@ model UserProfile {
...
+  @@map("model")
+  id Int @id
+}
 `
 
 test('migrator correctly PascalCases model names', (t) => {
   const newSchema = doPascalCaseTableNames(
     lowerCasePrismaSchema.split(/\r?\n/)
   ).join('\n')
+  console.log(expectedPrismaSchema)
```
alco, line 115: Looks like leftover debug code.

clients/typescript/src/cli/migrations/migrate.ts
```diff
@@ -320,7 +321,10 @@ export function doPascalCaseTableNames(lines: string[]): string[] {
         : capitaliseFirstLetter(tableName) // always capitalise first letter
 
       // Replace the model name on this line
-      const newLn = ln.replace(tableName, modelName)
+      //const newLn = ln.replace(tableName, modelName)
```
alco, line 324: Looks like leftover code.

alco (approved):
:+1:

clients/typescript/src/cli/migrations/migrate.ts
kevin-dp: Oops, was too fast. Thanks for catching!

PR #516, opened 2023-10-02
fix (satellite): avoid applying migrations twice

kevin-dp:
This PR fixes the bug where a migration would be applied twice in the following scenario:
- migrate PG while client app is running
- re-generate client
- restart app

On restart the app would try to re-apply the migration it has previously received through the replication stream.

Normally this should not occur because we keep track of which migrations have been applied. However, we forgot to store the migration version in the transaction (on Satellite) which caused it to be processed as a regular transaction instead of a migration ([cf. this if test in the code](https://github.com/electric-sql/electric/blob/main/clients/typescript/src/satellite/process.ts#L1230)) and hence we did not mark it as applied.


alco (commented):
Why aren't there changes to any test code here? Could you add a client test using a mock SatOpLog message to verify that it stores the migration version after processing it?

kevin-dp:
> Why aren't there changes to any test code here? Could you add a client test using a mock SatOpLog message to verify that it stores the migration version after processing it?

Right, i was thinking that a proper test would be an e2e test to send a migration at runtime and then re-generate the client, etc. But we currently don't have the test infrastructure to re-generate client code etc.
But it is indeed a good idea to write a unit test that mocks an incoming migration and checks that it is stored in the DB.
Thanks for the feedback, will do that!

kevin-dp:
@alco Actually this is a bit difficult to test in a unit test because it requires testing interaction between two modules (the Satellite client and the Satellite process). So it would require setting up a Satellite process with a mock Satellite client with a mock socket.. which is a bit of a hassle for a one line code change. Note that we already have unit tests that check that the Satellite process correctly applies migrations given that the transaction information it gets from the Satellite client is correct. So i added a unit test to the Satellite client to check that the transaction information it constructs contains all the necessary information.

alco (approved):
Great work! :+1:

PR #518, opened 2023-10-03
fix(satellite): clients overload the server on fatal errors 

balegas:
Some errors would trigger a reconnect loop on the server. This patch modifies error handling for preventing the client from reconnecting in a loop on FATAL errors.

Initially, we thought about DROPping the database when FATAL errors occur, but I've decided not to do that as part of this patch. The main reason is that FATAL errors should not occur in production if not because of a bug or re-deploying the backend database without refreshing the client code. In both cases, using some recovery tool is a more sensible way of addressing any potential issues -- which we're currently working on.

Meanwhile, we point the user to the logs for learning how to address the issue and raise that a recovery tool is underway.

This patch additionally exposes the backoff configuration to the application through the Electrify config.

### DIFF
--- a/clients/typescript/src/config/index.ts
+++ b/clients/typescript/src/config/index.ts
@@ -1,6 +1,8 @@
-import { IBackOffOptions } from 'exponential-backoff'
 import { AuthConfig } from '../auth/index'
-import { satelliteDefaults } from '../satellite/config'
+import {
+  ConnectionBackoffOptions as ConnectionBackOffOptions,
+  satelliteDefaults,
+} from '../satellite/config'
 
 export interface ElectricConfig {
   auth: AuthConfig
@@ -29,7 +31,7 @@ export interface ElectricConfig {
   /**
    * Optional backoff options for connecting with Electric
    */
-  backoffOpts?: Omit<IBackOffOptions, 'retry'>
+  connectionBackOffOptions?: ConnectionBackOffOptions
 }
 
 export type HydratedConfig = {
@@ -40,7 +42,7 @@ export type HydratedConfig = {
     ssl: boolean
   }
   debug: boolean
-  backoffOpts: Omit<IBackOffOptions, 'retry'>
+  connectionBackOffOptions: ConnectionBackOffOptions
 }
 
 export type InternalElectricConfig = {
@@ -51,7 +53,7 @@ export type InternalElectricConfig = {
     ssl: boolean
   }
   debug?: boolean
-  backoffOpts?: Omit<IBackOffOptions, 'retry'>
+  connectionBackOffOptions?: ConnectionBackOffOptions
 }
 
 export const hydrateConfig = (config: ElectricConfig): HydratedConfig => {
@@ -76,12 +78,30 @@ export const hydrateConfig = (config: ElectricConfig): HydratedConfig => {
     ssl: sslEnabled,
   }
 
-  const backoffOpts = config.backoffOpts || satelliteDefaults.backoffOpts
+  const {
+    delayFirstAttempt,
+    jitter,
+    maxDelay,
+    numOfAttempts,
+    startingDelay,
+    timeMultiple,
+  } =
+    config.connectionBackOffOptions ??
+    satelliteDefaults.connectionBackOffOptions
+
+  const connectionBackOffOptions = {
+    delayFirstAttempt,
+    jitter,
+    maxDelay,
+    numOfAttempts,
+    startingDelay,
+    timeMultiple,
+  }
 
   return {
     auth,
     replication,
     debug,
-    backoffOpts,
+    connectionBackOffOptions,
   }
 }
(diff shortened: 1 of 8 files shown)


kevin-dp:
@balegas i believe `npx prettier --write .` also works on tests if you invoke it from `clients/electric`

clients/typescript/src/config/index.ts
```diff
@@ -24,6 +26,10 @@ export interface ElectricConfig {
    * Defaults to `false`.
    */
   debug?: boolean
+  /**
+   * Optional backoff options for connecting with Electric
+   */
+  backoffOpts?: Omit<IBackOffOptions, 'retry'>
```
kevin-dp, line 32: We're using the `Omit<IBackOffOptions, 'retry'>` type in several places.
May be worth creating a type alias for it in order to improve maintainability.

clients/typescript/src/config/index.ts
```diff
@@ -68,9 +76,12 @@ export const hydrateConfig = (config: ElectricConfig): HydratedConfig => {
     ssl: sslEnabled,
   }
 
+  const backoffOpts = config.backoffOpts || satelliteDefaults.backoffOpts
```
kevin-dp, line 79: If you don't want users to use the `retry` option, make sure to remove it from the object here.
Perhaps don't mutate the original object, that may be unexpected for the user. Instead, it may be better to create a copy of only the options we support.

clients/typescript/src/satellite/process.ts
```diff
@@ -671,30 +670,55 @@ export class SatelliteProcess implements Satellite {
...
-    }).catch((e) => {
-      // We're very sure that no calls are going to modify `this.initializing` before this promise resolves
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    }
 
-      const error = new SatelliteError(
-        SatelliteErrorCode.CONNECTION_FAILED_AFTER_RETRY,
-        `Failed to connect to server after exhausting retry policy. Last error thrown by server: ${e.message}`
-      )
+    await backOff(async () => {
```
kevin-dp, line 683: You're aware that `backOff` will only run the function for the first time after the chosen `startingDelay` has elapsed? In the default config you chose `1000` so it will only try to connect for the first time after 1 second. Not sure if we want that. I would expect that we try to connect immediately and only afterwards introduce a backoff. This is a bit annoying to do, so perhaps lower the default starting delay?

clients/typescript/src/satellite/process.ts
```diff
@@ -735,54 +764,32 @@ export class SatelliteProcess implements Satellite {
...
+      }
 
-        Log.warn(
-          `Couldn't start replication with reason: ${error.message}. resetting client state and retrying`
-        )
-        return this._resetClientStateAndReconnect()
+      // Some errors could be fixed by dropping local database entirely
+      // We propagate a fatal error, so the application can decide how to handle it
+      if (isThrowable(error) || isFatal(error)) {
+        throw isFatal(error) ? wrapFatalError(error) : error
```
kevin-dp, line 783: A bit weird to have a ternary inside an if test.
I would split it into several branches:

```
if (isThrowable(error)) {
  throw error
}
else if (isFatal(error) {
  throw wrapFatalError(error)
}
```

clients/typescript/test/satellite/process.test.ts
```diff
@@ -70,7 +70,8 @@ test('setup starts a satellite process', async (t) => {
 test('start creates system tables', async (t) => {
   const { adapter, satellite, authState } = t.context
 
-  await satellite.start(authState)
+  const conn = await satellite.start(authState)
+  await conn.connectionPromise
```
kevin-dp, line 74: Why do the tests now need to also `await conn.connectionPromise` while this was not the case before?
I don't see any changes to the `start` function that explain this.

kevin-dp (approved):
I only have some minor comments, otherwise it LGTM 👍

clients/typescript/src/satellite/config.ts
```diff
@@ -37,6 +41,14 @@ export const satelliteDefaults: SatelliteOpts = {
   pollingInterval: 2000,
   minSnapshotWindow: 40,
   clearOnBehindWindow: true,
+  connectionBackOffOptions: {
+    delayFirstAttempt: false,
+    startingDelay: 0,
```
kevin-dp, line 46: Have you tested this? The docs say:
> The startingDelay is multiplied by the timeMultiple to increase the delay between reattempts.
So i'm afraid you're always going to multiple by 0 so it will not back off at all.

balegas: ups.. my bad. was just thinking on the first delay thing and did not thought through

balegas (commented):
Sorry forgot so send comments

clients/typescript/src/satellite/process.ts
kevin-dp: oops I had missed the `delayFirstAttempt` option.

PR #520, opened 2023-10-03
chore(examples/starter): Fix DB migrations and subcommand execution

alco:
This is a follow-up to https://github.com/electric-sql/electric/pull/505 and https://github.com/electric-sql/electric/pull/502. I'm addressing a few different concerns in the same PR only because we don't have automated tests for the starter and redoing the same manual testing steps multiple times is error-prone.

Here's a brief description of changes.

## Removed `shelljs-live`

This reverts https://github.com/electric-sql/electric/pull/505/commits/a8764b0181171c9ab4360caf028b1c0b0dcec554.

Turns out `shelljs-live` is not a drop-in replacement for `shelljs`. It only uses shelljs's config but implements a different API for executing a subprocess that only returns the status code to the caller. This broke our error handling in `startCompose.js` and `startElectric.js` that expects the return value to have the `stderr` field. I discovered this problem when I executed `yarn backend:up` instead of `yarn backend:start`:

<details>
<summary>Console output</summary>

```
$ yarn backend:up
yarn run v1.22.19
warning package.json: No license field
warning ../../../package.json: No license field
$ yarn backend:start --detach
warning package.json: No license field
warning ../../../package.json: No license field
$ node ./backend/startCompose.js --detach
[+] Running 5/5
 ✔ Network app_default         Created                                                                                                                                                                          0.1s 
 ✔ Volume "app_pg_data"        Created                                                                                                                                                                          0.0s 
 ✔ Volume "app_electric_data"  Created                                                                                                                                                                          0.0s 
 ✔ Container app-postgres-1    Started                                                                                                                                                                          0.1s 
 ✔ Container app-electric-1    Started                                                                                                                                                                          0.1s 
/home/alco/code/tmp/app/backend/startCompose.js:11
if (res.code !== 0 && res.stderr.includes('port is already allocated')) {
                                 ^

TypeError: Cannot read properties of undefined (reading 'includes')
    at Object.<anonymous> (/home/alco/code/tmp/app/backend/startCompose.js:11:34)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:86:12)
    at node:internal/main/run_main_module:23:47

Node.js v18.18.0
error Command failed with exit code 1.
error Command failed with exit code 1.
```

</details>

It does mean we lose colorized output from `yarn backend:start`. But I believe it can be restored by getting rid of `shelljs` completely and using `cross-spawn` like `shelljs-live` does.

## Replaced ad-hoc migration runner with @databases/pg-migrations

This package does everything we need: it can run multiple migrations and keeps track of already applied migrations. Here's sample output where I run it to apply a migration, then run it again to demonstrate the idempotency, then I add a new migration file and run it one more time:

Screenshot from 2023-10-03 17-40-10

The way `@databases/pg-migrations` keeps track of applied migrations is a bit noisy:
```
[localhost] postgres:app=# \d
                          List of relations
 Schema │                 Name                  │   Type   │  Owner   
────────┼───────────────────────────────────────┼──────────┼──────────
 public │ atdatabases_migrations_applied        │ table    │ postgres
 public │ atdatabases_migrations_applied_id_seq │ sequence │ postgres
 public │ atdatabases_migrations_version        │ table    │ postgres
 public │ foo                                   │ table    │ postgres
 public │ items                                 │ table    │ postgres
(5 rows)
```

But I don't think it's a problem for the starter template. Once we introduce integrations with backend frameworks, we'll replace this migrations runner with framework-specific ones.

**NOTE**: GitHub shows a very hard-to-read diff for `db/migrate.js`. I suggest viewing the file instead because it has completely new content compared to what was there before.

## Reset terminal color back to default after printing an error message

Our colorized error messages where missing the "reset" ANSI escape which was causing the output following the error message to also be colored, namely, the `Done` line printed by `yarn`.


kevin-dp (approved):
Good job! Thanks for taking care of this @alco 🥇

kevin-dp:
I guess we can close #502 now as this replaces it.

PR #523, opened 2023-10-04
chore(electric): Log failure reason when client fails to establish a WebSocket connection

alco:
This should provide more useful debug logs than the current

    [info] Get /ws
    [info] Sent 400 in 153 us


magnetised (approved):
👍🏻

PR #527, opened 2023-10-05
Tauri example postgres

tachyonicbytes:
A part of the tauri example. This is a WIP, more changes are needed.

To load a dump of data into the db, download this  https://www.dropbox.com/s/oqzy4igb89fi953/data.zip?dl=1 and extract it to `./examples/tauri-postgres/db/data/`. then run:

```
ISSUES_TO_LOAD=2000 pnpm run db:load-data
```


KyleAMathews:

PR #529, opened 2023-10-05
feat(electric): VAX-1053 detection of prisma migration version

magnetised:
and general fixes for prisma compatibility


e2e/prisma_example/migrations-1/prisma/migrations/20230915175207_electric/migration.sql
```diff
@@ -0,0 +1,3 @@
+--ALTER TABLE "public"."Items" ENABLE ELECTRIC;
```
alco, line 1: Why doesn't this work in this context?

alco (approved):
:muscle:

e2e/prisma_example/migrations-1/prisma/migrations/20230915175207_electric/migration.sql
magnetised: https://linear.app/electric-sql/issue/VAX-1172/ddlx-parser-fails-to-parse-table-name-publicitems

:(

PR #537, opened 2023-10-09
feat(DAL)!: data type support for time-related types

kevin-dp:
This PR introduces data type support for PG's `time`, `timetz`, `timestamp`, `timestamptz`, and `date` type on the client*. It fixes VAX-965, VAX-966, VAX-858, VAX-856, and VAX-841 and its subissues: VAX-991, VAX-992, VAX-993, VAX-994, VAX-995, VAX-996, VAX-997.

*Although the client can handle the `timetz` type, it is not supported by Electric.

**BREAKING CHANGE:** This PR introduces breaking changes to the format of the generated Electric client.
The fields of a table now consist of a map from column name to PG column type whereas previously it was an array of column names.

In a nutshell, this PR introduces the following changes:
- Modifies the CLI to introspect the PG database via Electric's proxy
- Modifies the generator to parse the Prisma schema (we got by introspecting the PG DB) and extend the DMMF with the type attributes of fields
- Modifies the generator to expose PG column types on the generated Electric client
- Modifies DAL to transform JS date objects to SQLite strings encoding the date/time/timestamp based on the PG type information in the generated Electric client
- Modifies DAL to transform SQLite strings into JS date objects when reading from the SQLite DB based on the PG type information in the generated Electric client

Testing this PR requires running an example app against a locally built Electric, ts-client, and generator.
- build the starter (`cd examples/starter && pnpm install && pnpm build && npm link`)
- create an app `create-electric-app my-app`
- modify your app to use the local Electric image (because it has not yet been published to Docker hub)
   i.e. in `my-app/backend/compose/.envrc` change the `ELECTRIC_IMAGE` env var to be `electric:local-build`.
- modify your app to use our packed ts-client:
    - change the electric-sql dependency in `my-app/package.json` to be `"electric-sql": "file:<path to electric mono 
    - change the `@electric-sql/prisma-generator` dependency in `my-app/package.json` to be `"@electric-sql/prisma-generator": "file:<path to electric mono 
- now run the app:
    - `yarn`
    - `yarn backend:start`
    - `yarn db:psql`
        - Now, create a table containing the different date/time types:
           ```
           CREATE TABLE public.timestamps (
             id TEXT PRIMARY KEY,
             created_at TIMESTAMP NOT NULL DEFAULT now(),
             updated_at TIMESTAMPTZ NOT NULL DEFAULT now(),
             time TIME NOT NULL DEFAULT now(),
             date DATE NOT NULL DEFAULT now()
           );
           ALTER TABLE public.timestamps ENABLE ELECTRIC;
           ```
    - `yarn client:generate`
    - Now modify `Example.tsx` to read from and insert into the `timestamps` table (the modified `Example.tsx` file is provided at the bottom of this PR)
    - `yarn start`

Modified `Example.tsx` file for testing purposes:
```
import React, { useEffect, useState } from 'react'

import { makeElectricContext, useLiveQuery } from 'electric-sql/react'
import { genUUID, uniqueTabId } from 'electric-sql/util'
import { ElectricDatabase, electrify } from 'electric-sql/wa-sqlite'

import { authToken } from './auth'
import { DEBUG_MODE, ELECTRIC_URL } from './config'
import { Electric, Timestamps as Timestamp, schema } from './generated/client'

import './Example.css'

const { ElectricProvider, useElectric } = makeElectricContext<Electric>()

export const Example = () => {
  const [ electric, setElectric ] = useState<Electric>()

  useEffect(() => {
    let isMounted = true

    const init = async () => {
      const config = {
        auth: {
          token: authToken()
        },
        debug: DEBUG_MODE,
        url: ELECTRIC_URL
      }

      const { tabId } = uniqueTabId()
      const tabScopedDbName = `electric-${tabId}.db`

      const conn = await ElectricDatabase.init(tabScopedDbName, '')

      if (!isMounted) {
        return
      }

      setElectric(electric)
    }

    init()

    return () => {
      isMounted = false
    }
  }, [])

  if (electric === undefined) {
    return null
  }

  return (
    <ElectricProvider db={electric}>
      <ExampleComponent />
    </ElectricProvider>
  )
}

const ExampleComponent = () => {
  const { db } = useElectric()!
  const { results } = useLiveQuery(
    db.timestamps.liveMany()
  )

  useEffect(() => {
    const syncItems = async () => {
      // Resolves when the shape subscription has been established.
      const shape = await db.timestamps.sync()

      // Resolves when the data has been synced into the local database.
      await shape.synced
    }

    syncItems()
  }, [])

  const addItem = async () => {
    const now = new Date()
    await db.timestamps.create({
      data: {
        id: genUUID(),
        created_at: now,
        updated_at: now,
        time: now,
        date: now,
      }
    })
  }

  const clearItems = async () => {
    await db.timestamps.deleteMany()
  }

  const items: Timestamp[] = results ?? []

  return (
    <div>
      <div className="controls">
        <button className="button" onClick={ addItem }>
          Add
        </button>
        <button className="button" onClick={ clearItems }>
          Clear
        </button>
      </div>
      {items.map((item: Timestamp, index: number) => (
        <p key={ index } className="item">
          <code>{ item.id + ' - ' + item.created_at + ' - ' + item.updated_at + ' - ' + item.time + ' - ' + item.date }</code>
        </p>
      ))}
    </div>
  )
}
```


clients/typescript/test/client/generated/client/libquery_engine-darwin-arm64.dylib.node
alco, line 1: Just noticed this. Probably accidental addition to Git? Otherwise, it would only work on macOS obviously.

kevin-dp: oops indeed! good catch

clients/typescript/src/client/conversions/sqlite.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    case PgDateType.PG_TIMETZ:
+      // Returns UTC time
+      return extractDateAndTime(v).time
+  }
+}
+
+// Deserialises a SQLite compatible date string into a `Date` object
+function deserialiseDate(v: string, pgType: PgDateType): Date {
+  const parse = (v: any) => {
```
alco, line 87: This can be defined as a module-level function. Any specific reason why it's a closure?

clients/typescript/src/client/conversions/sqlite.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+}
+
+/**
+ * Corrects the provided `Date` such that
+ * the current date is set as UTC date.
+ * e.g. if it is 3PM in GMT+2 then it is 1PM UTC.
+ *      This function would return a date in which it is 3PM UTC.
+ */
+function ignoreTimeZone(v: Date): Date {
```
alco, line 116: We could instead conditionally call either `toISOString()` or `toString()` where appropriate. That would alleviate the need to do arithmetic on milliseconds and would fully offload timezone handling to the JavaScript runtime.

alco (approved):
LGTM! :+1: 

I have skipped the changes `generator/src` as those are beyond my level of understanding.

I would suggest @icehaunter to have a look at the TS code as well.

clients/typescript/src/client/conversions/sqlite.ts
```diff
@@ -0,0 +1,137 @@
...
+ * and such that values that are read from the SQLite DB can be converted into JS values.
+ * For example, a `Date` value representing a Postgres timestamp can be converted to a string that can be stored in SQLite.
+ * When reading from the SQLite database, the string can be parsed back into a `Date` object.
+ */
+
+export enum PgBasicType {
+  PG_BOOL = 'BOOLEAN',
+  PG_SMALLINT = 'INT2',
+  PG_INT = 'INT4',
+  PG_FLOAT = 'FLOAT8',
```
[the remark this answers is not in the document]
alco: I see you've fixed it in the next PR - https://github.com/electric-sql/electric/pull/542/commits/5fb56065273d5c64a279087887f39f5283c45e22#diff-0e52d53e61005f9ad23e9a24992613360a8f77cfaab3911d9809a2c759a79af8. So feel free to ignore my comment here.

clients/typescript/src/client/conversions/sqlite.ts
alco: Since you've moved this code to another file in the next PR, it makes sense to make changes there (if you're planning to act on my comments here).

clients/typescript/src/client/model/builder.ts
```diff
@@ -25,6 +25,11 @@ export class Builder {
     private shapeManager: IShapeManager
   ) {}
 
+  // TODO: in create, updateInternal, deleteInternal, and findWhere
```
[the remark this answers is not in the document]
kevin-dp: This is a leftover todo. It has been implemented. Will remove.

clients/typescript/src/client/conversions/sqlite.ts
kevin-dp: @alco  We can't use `toISOString()` or `toString()` for this purpose. We want to transform the provided Date object into the same date and time in UTC (e.g. input 3PM GMT+2 -> we want output 3PM UTC). However, `toISOString()` would return a string where it is `1PM UTC` and `toString` would leave it in GMT+2 time zone.

alco: Got it. My bad, I didn't fully grasp the intent here.

davidmartos96 (changes requested):
Could we add a test to the `datatype.test.ts` file for the case of TIMESTAMP when a Date object with offset is provided?

```ts
test.serial('support timestamp type - input date with offset', async (t) => {
  const date = new Date('2023-08-07 18:28:35.421+05')

  const res = await tbl.create({
    data: {
      id: 1,
      timestamp: date,
    },
  })

  // ??????????
})
```

Should the function `ignoreTimeZone` return `18:28:35.421` here or `13:28:35.421`?

kevin-dp:
> Could we add a test to the `datatype.test.ts` file for the case of TIMESTAMP when a Date object with offset is provided?
> 
> ```ts
> test.serial('support timestamp type - input date with offset', async (t) => {
>   const date = new Date('2023-08-07 18:28:35.421+05')
> 
>   const res = await tbl.create({
>     data: {
>       id: 1,
>       timestamp: date,
>     },
>   })
> 
>   // ??????????
> })
> ```
> 
> Should the function `ignoreTimeZone` return `18:28:35.421` here or `13:28:35.421`?

When using `timestamp` (i.e. without time zone) we are storing the local date and time and anyone reading the timestamp should get the same date and time local to them. So if i'm in GMT+2 and it is 3PM and i store the timestamp then anyone reading it should also get 3PM in their timezone.

So when writing `const date = new Date('2023-08-07 18:28:35.421+05')` to the database as in your example, the timestamp everyone should read depends on the time zone of the user that wrote it to the database. I'm in GMT+2 so for me `'2023-08-07 18:28:35.421+05'` is `2023-08-07 15:28:35.421` local time and that is what gets stored in the DB. Hence, anyone reading my write will get `2023-08-07 15:28:35.421` in their timezone. That's exactly what `serialiseDate` (L65 in `src/client/conversions/sqlite.ts`) does: `ignoreTimeZone(new Date('2023-08-07 18:28:35.421+05')).toISOString()` results in `"2023-08-07T15:28:35.421Z"`.

davidmartos96:
@kevin-dp 

> So if i'm in GMT+2 and it is 3PM and i store the timestamp then anyone reading it should also get 3PM in their timezone.

Ok, this statement and the following review comment is what created the confusion https://github.com/electric-sql/electric/pull/590#discussion_r1371301971.

We were under the assumption that the value stored in SQLite is the date in UTC, without the T and the Z. So for your example, 1PM UTC would be stored, and anyone would read 1PM UTC converted to the corresponding local times.

We thought TIMESTAMP in Postgres could be use to store a moment in time, same for any client, and interpreted always as UTC.
Source: https://stackoverflow.com/a/5876276
In particular, this statement from the Stackoverflow:
> Without a time zone as part of the value, the obvious default time zone is UTC, so it is rendered for that time zone.

alco:
> Without a time zone as part of the value, the obvious default time zone is UTC, so it is rendered for that time zone.

@davidmartos96 I don't think that characterisation of the `TIMSTAMP` type is correct. A better way to think about it and `TIMESTAMPTZ` is as follows:
* `TIMESTAMP` stores a timestamp as basically text. You can think of it as a photo of a wall clock (that also shows the date). No matter where in the world one looks at the photo, it'll show the same time regardless.
* `TIMESTAMPTZ` stores an objective moment in time: it's the same moment in time no matter where in the world one reads it. If this moment happened 2 hours ago for me in GMT+3, for someone in GMT+2 it happened 1 hour ago.

Here's an example `psql` session that demonstrates this distinction:
```
=# create table foo (id text primary key, t timestamp, tz timestamptz);
CREATE TABLE

=# set timezone to 'Europe/Brussels';
SET
=# show timezone
electric-# ;
    TimeZone     
─────────────────
 Europe/Brussels
(1 row)

=# insert into foo values ('1', now(), now());
INSERT 0 1
=# insert into foo values ('2', '2023-10-25 17:28:00', '2023-10-25 17:28:00');
INSERT 0 1
=# select * from foo;
 id │             t              │              tz               
────┼────────────────────────────┼───────────────────────────────
 1  │ 2023-10-25 16:27:48.505401 │ 2023-10-25 16:27:48.505401+02
 2  │ 2023-10-25 17:28:00        │ 2023-10-25 17:28:00+02
(2 rows)

=# set timezone to 'UTC';
SET
=# show timezone;
 TimeZone 
──────────
 UTC
(1 row)

=# select * from foo;
 id │             t              │              tz               
────┼────────────────────────────┼───────────────────────────────
 1  │ 2023-10-25 16:27:48.505401 │ 2023-10-25 14:27:48.505401+00
 2  │ 2023-10-25 17:28:00        │ 2023-10-25 15:28:00+00
(2 rows)
```

See how the values in `t` stay the same regardless of the client's time zone? We've implemented what we think is equivalent behaviour in DAL: the `TIMESTAMP` value is simply fed into `new Date(...)` as is, with no time zone conversion.

If what you're interested is a moment in time, rather than a snapshot of wallclock time, you should use `TIMESTAMPTZ`.

davidmartos96:
@alco Thanks a lot! That really helps.
It was a misinterpretation with how Postgres works with those values. We don't have much experience with time operations in Postgres, we've always used dates stored as INT8 as milliseconds since Unix epoch.

I guess Timestamp is not the most clear name :sweat_smile:

alco:
> we've always used dates stored as INT8 as milliseconds since Unix epoch.

I think it's fair to say that this is almost exactly how Postgres stores TIMESTAMPTZ values internally: as an offset in microseconds since UNIX epoch which is defined in UTC.

Issue #532, opened 2023-10-06
[VAX-1187] Modify db:psql to connect to PG via Docker container


kevin-dp:
The current `db:psql` command relies on the machine's local `psql` command to connect to the PG DB that runs inside a Docker container. Thus it requires users to have postgres installed on their machine in order for the `psql` command to be found. We can lift this requirement by modifying the `db:psql` command such that it connects to the container that runs PG and then within that terminal run the psql command.

<sub>From [SyncLinear.com](https://synclinear.com) | [VAX-1187](https://linear.app/electric-sql/issue/VAX-1187/modify-dbpsql-to-connect-to-pg-via-docker-container)</sub>

PR #538, opened 2023-10-09
fix(starter): [VAX-1187] Modify db:psql to connect to PG via Docker container

alco:
I've defined a utility function to run `docker compose` without shelljs, and used that for both `backend/startCompose.js` and `db/connect.js`.

The benefit is that we get output coloring back.

The drawback is inability to capture the output and inspect it for error messages. But the current code that's supposed to find the port reuse error in the output already does not work as expected on Linux because the error message printed there is `address already in use`.

### DIFF
--- a/.changeset/brave-ways-hide.md
+++ b/.changeset/brave-ways-hide.md
@@ -1,0 +1,7 @@
+---
+"create-electric-app": patch
+---
+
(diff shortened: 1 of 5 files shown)


kevin-dp (approved):
Excellent! I like the dockerCompose abstraction.

balegas:
Closes #532

PR #541, opened 2023-10-10
feat(electric): Allow for user-set migration version to override automatic

magnetised:
- always assign migration transactions a version
- allow for user-assigned versions to override the automatic version
- prevent proxy-generated versions from overriding user-assigned versions


components/electric/lib/electric/postgres/extension/functions/assign_migration_version.sql.eex
```diff
@@ -1,6 +1,11 @@
 
-CREATE OR REPLACE PROCEDURE <%= @schema %>.assign_migration_version(_version text) SECURITY DEFINER AS $function$
+CREATE OR REPLACE PROCEDURE <%= @schema %>.assign_migration_version(_version text, _priority int) SECURITY DEFINER AS $function$
```
alco, line 2: The `_priority` argument should also be `int2` to match the column type.

components/electric/lib/electric/postgres/proxy/injector/operation.ex
```diff
@@ -547,16 +547,23 @@ defmodule Operation.AssignMigrationVersion do
...
+          {generate_version(state), 0, state}
       end
     end
 
     defp migration_version(%Operation.AssignMigrationVersion{version: version}, state)
          when is_binary(version) do
-      {version, state}
+      # this version is coming from the current transaction, so give it the
+      # highest priority of all these options
+      {version, 4, state}
```
alco, line 566: We need these priorities defined as constants or functions with descriptive names. It's easy to get the magic numbers mixed up.

alco (approved):

components/electric/lib/electric/postgres/extension/functions/assign_migration_version.sql.eex
magnetised: i did that at first, but it means you have to remember to cast the argument to `int2` at every call site, which is a pain.

PR #542, opened 2023-10-10
feat(DAL): support for booleans

kevin-dp:
This PR introduces support for booleans in the DAL.


clients/typescript/src/client/conversions/datatypes/date.ts
alco, line 1: What is date stuff doing in the PR for booleans?

kevin-dp: This PR is based on the date PR because that date PR introduced the machinery for doing SQLite <-> PG conversions which we also need for booleans. In this PR i refactored it a bit and moved conversions for each data type to a dedicated file (so i moved the date conversions to their own file and created a file for boolean conversions).

alco (approved):

kevin-dp:
@davidmartos96 Pasting here my message from Discord for anyone that may bump into this.

Let me summarise:
- To support PG types like boolean, float, etc. that are not supported by SQLite we need to encode them somehow.
- To encode the values, the client needs to known the original PG type of a column
- But, the Relation info cannot be trusted because that information is obtained in 2 ways:
  - When generating the app, we bundle the migrations by fetching them from Electric. The migrations that Electric provides us (through `SatRelation` messages) contains the SQLite type of the column instead of the PG type. This is problematic.
  - When a migration arrives at runtime, the `SatRelation` message contains the PG type.
- So unfortunately, we cannot rely on the Relation info as the bundled migrations will contain SQLite types and so we lost the original PG type.
- Instead, we now introspect the PG DB directly. So the Prisma schema now contains annotations with the original PG type of the columns.
- We expose that information in the generated client and use it in the client to do the SQLite <-> PG conversions.

davidmartos96 (changes requested):

PR #544, opened 2023-10-10
chore(electric): Apply own review feedback from the Proxy PR and other fixes

alco:
This applies the feedback that was missed in https://github.com/electric-sql/electric/pull/421. I've also made two more cleanups:
* got rid of the need to `create extension uuid-ossp` by using the builtin `gen_random_uuid()` function
* consistently applied the mapping from PG's `NULL` to Elixir's `nil` in all epgsql connections


e2e/prisma_example/migrations-1/prisma/schema.prisma
```diff
@@ -14,13 +14,13 @@ datasource db {
 // https://www.prisma.io/docs/guides/migrate/developing-with-prisma-migrate/add-prisma-migrate-to-a-project#baseline-your-production-environment
 
 model entries {
-  id        String  @id @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  id        String  @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
```
magnetised, line 17: how does this play with your disallow-defaults pr?

magnetised (approved):
cool. didn't notice the comments re epgsql or the uuid stuff tbh.

e2e/prisma_example/migrations-1/prisma/schema.prisma
alco: This PR is actually based on the branch that rejects tables with DEFAULT clauses. As you can see, all tests are passing. These `entries` and `owned_entries` tables aren't electrified, at least not as part of the migrations in the `migrations-1/prisma` directory. I didn't dive in to investigate this nuance, to be honest.

components/electric/lib/electric/replication/postgres_manager.ex
```diff
@@ -134,9 +134,11 @@ defmodule Electric.Replication.PostgresConnectorMng do
         {:DOWN, ref, :process, pid, reason},
         %State{pg_connector_sup_monitor: ref} = state
       ) do
-    Logger.warning(
```
davidmartos96, line 137: @alco Minor typo I have noticed by chance when overlooking the code :sweat_smile: 
"shutdwon"

Issue #243, opened 2023-07-12
[VAX-971] client: Leaky resources (Notifier subscriptions, EventEmitter listeners...)


davidmartos96:
Here I list some places in the client code which leak subscriptions to the notifier, as they are never cleared up. These are the ones we found while working on the Dart client.

- [x] `ElectricNamespace` 
https://github.com/electric-sql/electric/blob/23e6ff1bff414c4bb38d660cc4faf22742ce62b5/clients/typescript/src/electric/namespace.ts#L18
Would it make sense to have a `dispose`/`close` method in the `ElectricClient` class? It wouldn't close the database connection, but it would clean up internal resources used by Electric.

- [x] `Client` listeners https://github.com/electric-sql/electric/blob/1b24995731a663661064bf478bb73da8560168f7/clients/typescript/src/satellite/process.ts#L342
Those listeners are not cleaned up individually nor by calling `removeAllListeners` on the `EventEmitter`.

- [x] `AuthState` changes https://github.com/electric-sql/electric/blob/23e6ff1bff414c4bb38d660cc4faf22742ce62b5/clients/typescript/src/satellite/process.ts#L188
Possible solution: Since the subscription id is being reused, maybe there could be a `dispose` method in the `Satellite` interface.

- [x] `ConnectivityState` changes https://github.com/electric-sql/electric/blob/23e6ff1bff414c4bb38d660cc4faf22742ce62b5/clients/typescript/src/satellite/process.ts#L204
Possible solution: Unsubscribe in the `stop` method.

- [x] Throttle snapshot https://github.com/electric-sql/electric/blob/23e6ff1bff414c4bb38d660cc4faf22742ce62b5/clients/typescript/src/satellite/process.ts#L151
`lodash` throttle output has a method cancel.
Possible solution: Instantiate the throttle in the `start` method and clean it up in `stop`

<sub>[VAX-971](https://linear.app/electric-sql/issue/VAX-971/client-leaky-resources-notifier-subscriptions-eventemitter-listeners)</sub>

balegas:
Hei, we’ll look into this and report here

davidmartos96:
Another design question. Should the ElectricClient/Namespace `close` also close the Satellite process from the registry? Or is it something that the user would need to call additionally?

balegas:
I think close would be a destructive call and self-deregister from the registry.
I haven't yet got my head in the space for working on this, but is on my list :)

balegas:
[...]

PR #548, opened 2023-10-11
chore(satellite): VAX-1208: improve client events type safety

balegas:
Type check events emitted by the client. The consequence is that client no longer extends the eventEmitter, which shouldn't from the start.

Additionally, I added a shutdown mode to satellite.stop that clears event handlers assigned to the client. Not that it was a great harm because the client is scoped to the satellite instance. This allows to tick one more checkbox in #243

### DIFF
--- a/clients/typescript/src/satellite/client.ts
+++ b/clients/typescript/src/satellite/client.ts
@@ -125,6 +125,8 @@ interface SafeEventEmitter {
   removeListener(event: 'outbound_started', callback: () => void): void
   // TODO: missing removeListeners
 
+  removeAllListeners(): void
+
   listenerCount(event: 'error'): number
 }
 
@@ -151,6 +153,7 @@ export class SatelliteClient implements Client {
   private allowedMutexedRpcResponses: Array<keyof Root> = []
 
   private dbDescription: DbSchema<any>
+  private isDown: boolean = false
 
   private handlerForMessageType: { [k: string]: IncomingHandler } =
     Object.fromEntries(
@@ -216,15 +219,21 @@ export class SatelliteClient implements Client {
   }
 
   async connect(): Promise<void> {
-    if (!this.isClosed()) {
-      this.close()
+    if (this.isDown) {
+      throw new SatelliteError(
+        SatelliteErrorCode.UNEXPECTED_STATE,
+        'client has already shutdown'
+      )
+    }
+    if (!this.isDisconnected()) {
+      this.disconnect()
     }
 
     return new Promise<void>((resolve, reject) => {
       this.socket = new this.socketFactory(PROTOCOL_VSN)
 
       const onceError = (error: Error) => {
-        this.close()
+        this.disconnect()
         reject(error)
       }
 
@@ -241,7 +250,7 @@ export class SatelliteClient implements Client {
         this.socket.onMessage(this.socketHandler)
         this.socket.onError((error) => {
           if (this.emitter.listenerCount('error') === 0) {
-            this.close()
+            this.disconnect()
             Log.error(
               `socket error but no listener is attached: ${error.message}`
             )
@@ -249,7 +258,7 @@ export class SatelliteClient implements Client {
           this.emitter.emit('error', error)
         })
         this.socket.onClose(() => {
-          this.close()
+          this.disconnect()
           if (this.emitter.listenerCount('error') === 0) {
             Log.error(`socket closed but no listener is attached`)
           }
@@ -271,7 +280,7 @@ export class SatelliteClient implements Client {
     })
   }
 
-  close() {
+  disconnect() {
     this.outbound = this.resetReplication(this.outbound.last_lsn)
     this.inbound = this.resetReplication(this.inbound.last_lsn)
 
@@ -283,8 +292,14 @@ export class SatelliteClient implements Client {
     }
   }
 
-  isClosed(): boolean {
+  isDisconnected(): boolean {
     return !this.socketHandler
+  }
+
+  shutdown(): void {
+    this.disconnect()
+    this.emitter.removeAllListeners()
+    this.isDown = true
   }
 
   async startReplication(
@@ -978,7 +993,7 @@ export class SatelliteClient implements Client {
 
   private sendMessage<T extends SatPbMsg>(request: T) {
     if (Log.getLevel() <= 1) Log.debug(`[proto] send: ${msgToString(request)}`)
-    if (!this.socket || this.isClosed()) {
+    if (!this.socket || this.isDisconnected()) {
       throw new SatelliteError(
         SatelliteErrorCode.UNEXPECTED_STATE,
         'trying to send message, but client is closed'
--- a/clients/typescript/src/satellite/client.ts
+++ b/clients/typescript/src/satellite/client.ts
@@ -219,7 +219,7 @@ export class SatelliteClient implements Client {
     }
   }
 
-  async connect(): Promise<void> {
+  connect(): Promise<void> {
     if (this.isDown) {
       throw new SatelliteError(
         SatelliteErrorCode.UNEXPECTED_STATE,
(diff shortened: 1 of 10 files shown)


balegas:
The stop interface on process is not perfect right now, but it would require a bit more work.

clients/typescript/src/satellite/client.ts
```diff
@@ -100,9 +103,35 @@ const subscriptionError = [
...
+  emit(
+    event: 'transaction',
+    transaction: Transaction,
+    ackCb: () => void
+  ): boolean
+  emit(event: 'outbound_started', lsn: LSN): boolean
+
+  removeListener(event: 'error', callback: ErrorCallback): void
+  removeListener(event: 'outbound_started', callback: () => void): void
+  // TODO: missing removeListeners
```
kevin-dp, line 123: the ones for relation and transaction?
Is this work for a future issue?

clients/typescript/src/satellite/client.ts
```diff
@@ -100,9 +103,35 @@ const subscriptionError = [
   SatelliteErrorCode.SHAPE_DELIVERY_ERROR,
 ]
 
-export class SatelliteClient extends EventEmitter implements Client {
+interface SafeEventEmitter {
```
kevin-dp, line 106: Let's move `SafeEventEmitter` to `util/types.ts`

clients/typescript/src/satellite/client.ts
```diff
@@ -182,15 +208,21 @@ export class SatelliteClient extends EventEmitter implements Client {
   }
 
   async connect(): Promise<void> {
```
kevin-dp, line 210: this function is not actually doing any `await` and explicitly returns a promise so we can remove the `async` keyword.

clients/typescript/src/satellite/client.ts
```diff
@@ -249,8 +281,14 @@ export class SatelliteClient extends EventEmitter implements Client {
     }
   }
 
-  isClosed(): boolean {
-    return !this.socketHandler
+  isConnected(): boolean {
+    return this.socketHandler !== undefined
```
kevin-dp, line 285: Could do `return !!this.socketHandler`

clients/typescript/src/satellite/client.ts
```diff
@@ -249,8 +281,14 @@ export class SatelliteClient extends EventEmitter implements Client {
...
+  }
+
+  shutdown(): void {
+    this.disconnect()
+    this.emitter.removeAllListeners()
+    this.isDown = true
   }
 
   async startReplication(
```
kevin-dp, line 294: We can remove the `async` keyword here because we are always returning a promise and not using `await`.

clients/typescript/src/satellite/client.ts
```diff
@@ -329,18 +367,14 @@ export class SatelliteClient extends EventEmitter implements Client {
...
 
-  subscribeToTransactions(
-    callback: (transaction: Transaction) => Promise<void>
-  ) {
-    this.on('transaction', async (txn, ackCb) => {
-      // move callback execution outside the message handling path
-      await callback(txn)
-      ackCb()
+  subscribeToTransactions(callback: TransactionCallback) {
+    this.emitter.on('transaction', (txn, ackCb) => {
```
kevin-dp, line 371: perhaps make the callback async?
```
async (txn, ackCb) => {
  await callback(txn)
  ackCb()
}
```

clients/typescript/src/satellite/process.ts
```diff
@@ -331,7 +331,7 @@ export class SatelliteProcess implements Satellite {
   }
 
   // Unsubscribe from data changes and stop polling
-  async stop(): Promise<void> {
+  async stop(shutdown?: boolean): Promise<void> {
```
kevin-dp, line 334: This method has a slightly different type in the `Satellite` interface:
```
stop(shutdown: boolean): Promise<void>
```

The argument is not optional in the interface but is optional here.

clients/typescript/src/satellite/registry.ts
```diff
@@ -150,7 +150,7 @@ export abstract class BaseRegistry implements Registry {
     const satellites = this.satellites
     const satellite = satellites[dbName]
     if (satellite !== undefined) {
-      const stoppingPromise = satellite.stop().then(() => {
+      const stoppingPromise = satellite.stop(true).then(() => {
```
kevin-dp, line 153: This function is marked as `async` but is not using the `async/await` functionality.
Either remove the `async` keyword or use `async/await`.

clients/typescript/test/satellite/client.test.ts
```diff
@@ -295,10 +287,13 @@ test.serial('receive transaction over multiple messages', async (t) => {
   server.nextRpcResponse('stopReplication', [stop])
 
   await new Promise<void>(async (res) => {
-    client.on('transaction', (transaction: DataTransaction) => {
-      t.is(transaction.changes.length, 3)
-      res()
-    })
+    client.subscribeToTransactions(
+      (transaction: Transaction) =>
+        new Promise(() => {
```
kevin-dp, line 292: This promise never gets resolved?
The call to `res()` resolves the outer promise.

kevin-dp: This would be solved and be a bit cleaner perhaps if we use `async/await``:
```
async (transaction: Transaction) => {
  t.is(transaction.changes.length, 3)
  res()
}
```

clients/typescript/src/satellite/client.ts
balegas: No, just capturing the type

clients/typescript/test/satellite/client.test.ts
```diff
@@ -295,10 +287,13 @@ test.serial('receive transaction over multiple messages', async (t) => {
   server.nextRpcResponse('stopReplication', [stop])
 
   await new Promise<void>(async (res) => {
```
kevin-dp, line 289: passing an async function to a Promise constructor is an [anti-pattern](https://stackoverflow.com/a/43050114/2828147).

clients/typescript/test/satellite/client.test.ts
```diff
@@ -652,18 +653,22 @@ test.serial('default and null test', async (t) => {
...
-        transaction.changes[0].record['content'] as any
-      )
-      t.is(
-        record['text_null'] as any,
-        transaction.changes[0].record['text_null'] as any
-      )
-      res()
-    })
+    client.subscribeToTransactions(
+      // FIXME
```
kevin-dp, line 657: outdated comment?

clients/typescript/test/satellite/client.test.ts
```diff
@@ -652,18 +653,22 @@ test.serial('default and null test', async (t) => {
...
+      (transaction: any) =>
+        new Promise(() => {
```
kevin-dp, line 659: let's turn it into an async function

kevin-dp (approved):
Good job, i like the improved typings for the event emitter :-)
I left some minor comments.

kevin-dp (approved):
Good job, i like the improved typings for the event emitter :-)
I left some minor comments.

kevin-dp:
Don't forget to generate a changeset

clients/typescript/src/satellite/client.ts
kevin-dp: So this is something you are still going to do in this PR?

clients/typescript/test/satellite/client.test.ts
balegas: I changed to the async pattern.

clients/typescript/test/satellite/client.test.ts
balegas: made it more clear. Not touching the test :P

balegas (commented):
Addressed comments

kevin-dp (approved):

PR #549, opened 2023-10-11
chores(website): remove local stack

balegas:
Any reason for keeping this around?


balegas:
README is a but outdated

thruflo:
Have you checked that no one is using it? I remember @icehaunter was previously, for example.

icehaunter:
I personally see less reason to flat out remove it than to keep it and reference it from example scripts, because I think we're starting to get some duplication in each example setup. Maybe make local stack a little more generic and then softlink into examples? 

I was using local stack, but it's too much hassle I can maintain a docker-compose file on my own PC

balegas:
What would be the changes to the example? 
I'm okay with keeping the local-stack if we have a use for it, but we need to update documentation for it.

balegas:
@icehaunter if the decision is to keep this please make sure that documentation is update with latest instructions and documentation references.

alco (approved):

alco:
@icehaunter @balegas By this point, local stack is unusable with examples. Bringing it up to date conflicts with the starter template and with the ongoing work on improving the experience of running a local Electric backend as championed by Sam (see "Front end tooling improvement project" in Slab).

Issue #524, opened 2023-10-04
Domain type support


gorbak25:
Hi!
In my app the postgres database is append only for everything, due to it we want to use UUIDv7 not UUIDv4. 
I'm trying to enforce in the data model that everybody uses UUIDv7 not UUIDv4. I've created a domain type based on uuid for that:
```
-- <GORBAK_CUSTOM>
CREATE OR REPLACE FUNCTION get_uuid_version(_uuid uuid)
RETURNS INTEGER
LANGUAGE SQL
IMMUTABLE PARALLEL SAFE STRICT LEAKPROOF
AS $$
  SELECT get_byte(uuid_send(_uuid), 6) & 240 >> 4;
$$;
CREATE DOMAIN uuidv7 AS uuid CHECK (get_uuid_version(VALUE) = 7);
-- </GORBAK_CUSTOM>
```

Unfortunately electric doesn't consider that a type might be a domain type and that it has a base type:
https://github.com/electric-sql/electric/blob/2b24fa273d9597a3e143f278d756e67221c1d848/components/electric/lib/electric/postgres/extension/functions/validate_table_column_types.sql.eex#L27C20-L27C29
I suggest replacing
```
    FOR _col_name, _col_type, _col_typmod, _col_type_pretty IN
        SELECT attname, typname, atttypmod, format_type(atttypid, atttypmod)
            FROM pg_attribute
            JOIN pg_type on atttypid = pg_type.oid
            WHERE attrelid = table_name::regclass AND attnum > 0 AND NOT attisdropped
            ORDER BY attnum
    LOOP
```
with something along the lines of 
```
    FOR _col_name, _col_type, _col_typmod, _col_type_pretty IN
        SELECT attname, (CASE typtype = 'd' THEN (SELECT t1.typname from pg_type AS t1 WHERE t1.oid=pg_type.typbasetype) ELSE typname END), atttypmod, format_type(atttypid, atttypmod)
            FROM pg_attribute
            JOIN pg_type on atttypid = pg_type.oid
            WHERE attrelid = table_name::regclass AND attnum > 0 AND NOT attisdropped
            ORDER BY attnum
    LOOP
```

icehaunter:
Hey, thanks for the suggestion!

[...]

PR #564, opened 2023-10-12
feat(electric): Enable electrification of tables with columns of user-defined PG ENUM types

alco:
Closes VAX-1040, VAX-1041.

### DIFF
--- a/components/electric/lib/electric/postgres/extension.ex
+++ b/components/electric/lib/electric/postgres/extension.ex
@@ -142,8 +142,8 @@ defmodule Electric.Postgres.Extension do
         [] ->
           {:error, "no schema with version #{inspect(version)}"}
 
-        [{schema, version}] ->
-          with {:ok, schema} <- Proto.Schema.json_decode(schema) do
+        [{schema_json, version}] ->
+          with {:ok, schema} <- decode_schema_json(schema_json, conn) do
             {:ok, version, schema}
           end
       end
@@ -156,12 +156,32 @@ defmodule Electric.Postgres.Extension do
         [] ->
           {:ok, nil, Schema.new()}
 
-        [{schema, version}] ->
-          with {:ok, schema} <- Proto.Schema.json_decode(schema) do
+        [{schema_json, version}] ->
+          with {:ok, schema} <- decode_schema_json(schema_json, conn) do
             {:ok, version, schema}
           end
       end
     end
+  end
+
+  defp decode_schema_json(json, conn) do
+    with {:ok, schema} <- Proto.Schema.json_decode(json) do
+      {:ok, maybe_load_schema_types(schema, conn)}
+    end
+  end
+
+  # If the schema has no types, it must have been saved prior to the introduction of the `types` field on
+  # `Proto.Schema`. Load types at runtime as a workaround.
+  # This branch will become dead when another migration is processed by `MigrationConsumer`, causing a version
+  # of the schema with the `types` field included to be saved to the DB.
+  defp maybe_load_schema_types(%{types: type_map} = schema, conn) when map_size(type_map) == 0,
+    do: load_types_for_schema(schema, conn)
+
+  defp maybe_load_schema_types(schema, _conn), do: schema
+
+  defp load_types_for_schema(schema, conn) do
+    {column_type_names, types} = fetch_table_column_types(conn, schema.tables)
+    Schema.patch_column_types(schema, column_type_names, types)
   end
 
   def save_schema(conn, version, %Proto.Schema{} = schema, stmts) do
--- a/components/electric/lib/electric/postgres/extension.ex
+++ b/components/electric/lib/electric/postgres/extension.ex
@@ -3,6 +3,8 @@ defmodule Electric.Postgres.Extension do
   Manages our pseudo-extension code
   """
 
+  import Electric.Postgres.OidDatabase.PgType
+
   alias Electric.Postgres.{Schema, Schema.Proto, Extension.Functions, Extension.Migration}
   alias Electric.Replication.Postgres.Client
   alias Electric.Utils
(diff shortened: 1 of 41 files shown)


components/electric/lib/electric/postgres/extension.ex
```diff
@@ -566,4 +588,85 @@ defmodule Electric.Postgres.Extension do
...
+
+  A "canonical type name" is a string value of what is either a built-in type name or a schema-qualified user-defined
+  type. It is basically `PgType.type_name() |> to_string()`.
+  """
+  @spec fetch_table_column_types(conn, [%Proto.Table{}]) ::
+          {Schema.table_column_to_type_name(), Schema.type_name_to_extended_info()}
+  def fetch_table_column_types(conn, tables) do
+    table_column_types =
+      for %{name: name} = table <- tables do
+        {:ok, column_types} = Client.query_table_column_types(conn, table.oid)
```
magnetised, line 610: I disagree with this approach. We discussed it previously IIRC and my point was that having an immutable, point-in-time schema with dynamically looked up enum type values was not the right way to go as we lose the ability to view the schema -- including the definitions of any enum types -- at a particular revision

magnetised (changes requested):
Sorry, I disagree with the approach here. I think we should treat enum types much like the rest of the schema information, ie. changes come as ddl over the replication stream and are serialised into the db as an immutable struct.

The `ddlgen_create` function is able to provide the sql to create a type:

```
=# CREATE TYPE public.mood AS ENUM ('sad', 'ok', 'happy');
=# select electric.ddlgen_create('public.mood'::regtype);
       ddlgen_create
----------------------------
 CREATE TYPE mood AS ENUM (+
  'sad',                   +
  'ok',                    +
  'happy'                  +
 );                        +
```

you already have enum support in the schema proto, so it's just a question of patching the `electrify` function to detect custom types and prepend their creation sql to the front of the create table ddl, then adding support for create type and alter type to the schema update.

Also, separately, we need to add capture and validation of changes to enum types to the proxy code, so that we can refuse to remove or alter existing enum states as well as capture the `ALTER TYPE name ADD VALUE [ IF NOT EXISTS ] new_enum_value` statements (and decide if we support changing the ordering...). I can help you with that stuff

alco:
> I think we should treat enum types much like the rest of the schema information, ie. changes come as ddl over the replication stream

There are pros and cons to this, right?

It makes perfect sense from the point of view of maintaining consistency with how we process table electrification. It's dffierent syntax for different "kinds" of types, e.g. `CREATE TYPE ... AS ENUM` vs `CREATE DOMAIN ... AS ... CHECK ...` but that's just life. What we need to get serious about here is patching ddlgen up to use schema-qualified identifiers as we'll have to support schemas different from "public" one way or another.

On the other hand, Postgres has different kinds of types, such as base, enum, [domain](https://github.com/electric-sql/electric/issues/524), etc. It is reasonable to say we want to keep the information we store about those different kinds of types consistent. Hence the need for a generic query that can fetch the info from the DB that is not included in the SQL we get from ddlgen.

> Also, separately, we need to add capture and validation of changes to enum types to the proxy code, so that we can refuse to remove or alter existing enum states

I thought the Proxy rejected any `ALTER TYPE` in its current form? Intercepting those and updating the electrified schema is outside of scope for this PR. I'll circle back to you on this at a later time.

alco:
> There are pros and cons to this, right?

Just to add to this, ddlgen itself is a big hack that queries system tables in order to build what looks like the SQL used for creation of a given database object. It’s great that we can use it to ship a migration to clients using SQL syntax. But, in my opinion, it still remains to be seen whether parsing this generated SQL to recover some of the info from system catalogs that ddlgen used to build said SQL in the first place is the right approach for the server.

magnetised:
> > There are pros and cons to this, right?
> 
> Just to add to this, ddlgen itself is a big hack that queries system tables in order to build what looks like the SQL used for creation of a given database object. It’s great that we can use it to ship a migration to clients using SQL syntax. But, in my opinion, it still remains to be seen whether parsing this generated SQL to recover some of the info from system catalogs that ddlgen used to build said SQL in the first place is the right approach for the server.

Though I appreciate what you're saying here -- it does seem a bit silly -- what ddlgen is doing for us is doing the work of looking in the system catalogs and then serialising that information in a format that's consistent with the format we already have to handle in the case of DDL commands coming through the migrations via the proxy (e.g. `ALTER TABLE...` ). Since we will always need to parse and understand those, having a way of serialising an existing table structure into SQL DDL as well actually makes things easier for us, as there's only one format we need to understand, which is SQL.

if we changed the output of the ddlgen queries (which we'd have to run anyway to get the required info) then we'd need two pathways into the schema update mechanism, one in our custom serialisation format (for the electrification step), one in direct SQL DDL (for captured migrations), which seems like a lot of work for very little gain.

magnetised:
> I thought the Proxy rejected any `ALTER TYPE` in its current form? Intercepting those and updating the electrified schema is outside of scope for this PR. I'll circle back to you on this at a later time.

It does.

Re-reading the title of this pr you're right -- if it's only electrification then the proxy stuff is unimportant. my bad for assuming the title of this pr was "enum support" or equivalent.

alco:
> Since we will always need to parse and understand those

This is what I'm trying to understand. From my perspective, it's not a must, it was simply a choice that we made. We could just as well use custom queries in both cases: when electrifying a table and when we intercept DDL statements in the proxy, after letting those statements execute in Postgres. I'm not saying we should switch to that approach, just pointing out that it's a game of trade-offs. Our choice to parse SQL has a trade-off in that we still have to query the DB for table OIDs _separately_ from parsing SQL. The other trade-off is that we will need to patch up ddlgen to use schema-qualified identifiers if we want to support non-public schemas.

With that out of the way, I take your feedback to heart but I'm not 100% sure regarding the additional work needed in this PR to bring it to a mergeable state. As far as I understand:

* [ ] Modify the "electrify" procedure to insert the enum type definition into `ddl_commands` along with the table creation statement.
* [ ] In MigrationConsumer, replace the fetching of type info from the DB with parsing the type definition SQL
* [ ] (?) Get rid of fetching the type OID from the DB (?)
* [ ] (?) Revert the "canonicalisation" of type names that is done in `OidDatabase.PgType.type_name()` and assume "public" schema (?)

I don't like this last item because it makes future work of making Electric schema-aware a bit more difficult. It is easier to gradually start adding schema-awareness than do it all at once later.

magnetised:
> We could just as well use custom queries in both cases: when electrifying a table and when we intercept DDL statements in the proxy, after letting those statements execute in Postgres

compared to parsing the ddl  statements that's much harder. basically my point is that ddlgen is doing those queries for us and serialising the result into a format that's easy for us to use (and re-use in the case of say postgres in the client).

anyway. it's all tradeoffs as you say and i'm happy with the tradeoff as is. 

patching ddlgen to remove use of the search path I think should be relatively straightforward. never nice to have a big blob of code that you didn't write but have to support but I'm not worried about that.

> (?) Get rid of fetching the type OID from the DB (?)

that's what the `oid_loader` function does in the schema modification code. we just have to update it to support enums etc.

> Revert the "canonicalisation" of type names that is done in OidDatabase.PgType.type_name() and assume "public" schema (?)

is this a requirement because ddlgen will produce non-qualified type names and so we lose the ability to lookup types?

do you want me to update ddlgen to always schema-qualify names?

other than that I think your todos are right

PR #570, opened 2023-10-16
chore(electric): Update E2E tests to use the proxy_1 shell for migrations

alco:


magnetised (approved):
thanks for finding those

PR #571, opened 2023-10-16
chore(DAL): Bump capacitor-community/sqlite version to enable Android support

samwillis:
There was a bug in capacitor-community/sqlite that prevented our driver working on Android, this is fixed in the latest version. This sets the min version to that current published version.


thruflo:
Brilliant 🚀

kevin-dp (approved):
LGTM, don't forget to add a changeset.

PR #574, opened 2023-10-17
feat(DAL): support float8

kevin-dp:
This PR introduces client-side support for float8 values.
The e2e test fails because validation on Electric does not yet support `Infinity`, `-Infinity`, and `NaN` values.
Once Electric supports those values the e2e test should turn green.


alco (approved):

davidmartos96 (changes requested):
One thing we noticed that is broken with this PR is the merge logic of the oplog when using Infinite FLOAT8 values.

That's because `JSON.stringify` is called as is and infinite values are turned into `null`

https://github.com/electric-sql/electric/blob/cc35c12d4d95aaabfc5cd1689234657f69e35b09/clients/typescript/src/satellite/oplog.ts#L285 


JSON.stringify behavior for Infinite values
https://stackoverflow.com/questions/16644597/json-stringify-converting-infinity-to-null

davidmartos96 (changes requested):
Another issue related to the FLOAT8 type addition is the use of the `json_object` function on the SQLite oplog TRIGGERs. It can return an unparsable JSON object on SQLite version 3.41.0 and lower.

More info here:
https://discord.com/channels/933657521581858818/1163829658236760185

kevin-dp:
> One thing we noticed that is broken with this PR is the merge logic of the oplog when using Infinite FLOAT8 values.
> 
> That's because `JSON.stringify` is called as is and infinite values are turned into `null`
> 
> https://github.com/electric-sql/electric/blob/cc35c12d4d95aaabfc5cd1689234657f69e35b09/clients/typescript/src/satellite/oplog.ts#L285
> 
> JSON.stringify behavior for Infinite values https://stackoverflow.com/questions/16644597/json-stringify-converting-infinity-to-null

Thanks again for this very valuable input. Same problem for `NaN` also.
We could fix this by providing a custom transformer to `JSON.stringify` that transforms infinity values into a string representation. Care needs to be taken when parsing to differentiate between the string "+Infinity" and the infinity value encoded as "+Infinity" (can be differentiated based on the PG type). Instead of doing this special casing we could also re-use the serialise and deserialise functions already implemented in the satellite client. Although this means that it will actually do a SQLite -> PG transformation of the values when serialising and a PG -> SQLite transformation of the values when deserialising.

kevin-dp:
Note to self: add tests of edge-cases (NaN, +Inf, -Inf) to `serialization.test.ts`

kevin-dp:
> One thing we noticed that is broken with this PR is the merge logic of the oplog when using Infinite FLOAT8 values.
> 
> That's because `JSON.stringify` is called as is and infinite values are turned into `null`
> 
> https://github.com/electric-sql/electric/blob/cc35c12d4d95aaabfc5cd1689234657f69e35b09/clients/typescript/src/satellite/oplog.ts#L285
> 
> JSON.stringify behavior for Infinite values https://stackoverflow.com/questions/16644597/json-stringify-converting-infinity-to-null

I committed a fix for this which consists of a custom serialiser and deserialiser.
The serialiser transforms `NaN`, `+Infinity`, and `-Infinity` into a string representation.
Then, the deserialiser transforms those strings back into their numeric value if the value belongs to a column whose type is a float or real number.

kevin-dp:
> Another issue related to the FLOAT8 type addition is the use of the `json_object` function on the SQLite oplog TRIGGERs. It can return an unparsable JSON object on SQLite version 3.41.0 and lower.
> 
> More info here: https://discord.com/channels/933657521581858818/1163829658236760185

I still have this problem on SQLite 3.41.2.
Can confirm that it is fixed from version 3.42 and on.

kevin-dp:
> Another issue related to the FLOAT8 type addition is the use of the `json_object` function on the SQLite oplog TRIGGERs. It can return an unparsable JSON object on SQLite version 3.41.0 and lower.
> 
> More info here: https://discord.com/channels/933657521581858818/1163829658236760185

@davidmartos96 we found a way to work around this bug. The key is to cast the value to TEXT such that infinity values become `"Inf"`/`"-Inf"` and then upon deserialisation of an oplog entry make sure to convert it back into its numerical value `Infinity`/`-Infinity`. I've implemented the workaround, we only do the cast to TEXT for columns of type REAL (see [triggers.ts](https://github.com/electric-sql/electric/pull/574/commits/55af617e24f68cc81c92df547a8aa516f3fcd640#diff-528bca9ac6df31ffaf14ba0bc196657e0791fc518b80e37a44498fde394ba9efR275)). I've also added unit tests for this workaround.

PR #575, opened 2023-10-17
bug(satellite): potential race condition in _performSnapshot()

balegas:
`_performSnapshot()` must be executed serially. All calls to are protected with a mutex and we have an assertion on the function to ensure we enforce serial execution. However, sometimes we'd still see the assertion fail.

We are not setting properly the `performingSnapshot` variable when exceptions occur in `_performSnapshot()`. Although this probably constitutes a bug, it happens. The exceptions being thrown might also be swallowed sometimes, which made more difficult to track the error (and potentially still hiding other error).

### DIFF
--- a/.changeset/chilled-cameras-swim.md
+++ b/.changeset/chilled-cameras-swim.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+handle exceptions properly in \_performSnapshot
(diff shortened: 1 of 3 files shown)


balegas:
I apologize to the reviewers for the horrid diff :D

davidmartos96:
@balegas Not bad at all to review when hiding whitespace diffs :wink: 

image

kevin-dp (approved):
Good catch!

PR #578, opened 2023-10-18
feat(docs): Framework docs for Phoenix

magnetised:
plus update the beers-stars example to use the proxy


local-stack/docker-compose.yaml
```diff
@@ -31,9 +31,12 @@ services:
       <<: *preshared-auth-variables
       DATABASE_URL: postgresql://postgres:password@postgres_1:5432/electric
       LOGICAL_PUBLISHER_HOST: electric_1
+      PG_PROXY_PORT: "65432"
```
alco, line 34: Same feedback here as above.

docs/integrations/backend/phoenix.md
```diff
@@ -5,7 +5,80 @@ description: >-
...
+end
+```
+
+Configure Ecto to use the `ProxyRepo` for its tasks. Because in most cases queries through the proxy will just pass unmodified to the backing Postgresql server, it's ok that the proxy repo will be used for other Ecto mix tasks apart from applying migrations.
+
+```elixir
+# config/config.exs
+
+config :my_app,
+  ecto_repos: [MyApp.ProxyRepo]
```
thruflo, line 33: Assuming that this doesn't make your code generally use the ProxyRepo, maybe a note to that effect? I was eyeballing this code feeling nervous.

docs/usage/data-modelling/migrations.md
```diff
@@ -39,7 +41,60 @@ See the [Limitations](#limitations) section below and the [Roadmap](../../refere
 
 ## Migration Proxy
```
thruflo, line 42: Lower case `p`.

docs/usage/data-modelling/migrations.md
```diff
@@ -39,7 +41,60 @@ See the [Limitations](#limitations) section below and the [Roadmap](../../refere
...
+Postgresql migrations are applied to Postgresql via a proxy server integrated into the Electric application.
+
+This proxy server serves various purposes:
+
+- It allows the use of the [DDLX syntax](../../api/ddlx.md) for managing your tables and access permissions,
+- It captures migrations applied to Electrified tables in order to propagate those DDL changes to the client schemas,
+- It validates migrations to electrified tables to ensure that changes to the schema are supported by Electric (e.g. validating the types of any added columns, ensuring that only additive migrations are applied, etc), and
+- It provides an endpoint for schema introspection to allow Electric to return its view of the underlying Postgres database to the data access library.
+
+Migrations not passed through the proxy endpoint will not be captured by Electric and will cause problems as Electric's view of the Postgresql schema will be out of sync with the actual table schema.
```
thruflo, line 53: In my understanding, the user can still apply migrations to non-electrified tables without using the proxy and that's fine, right? If so, I think this sentence could be clearer. E.g.:

> All migrations using DDLX syntax or changing electrified tables must go through the migration proxy. Migrations not using DDLX syntax and only affecting non-electrified tables can go through the proxy but do not need to.

docs/usage/data-modelling/migrations.md
```diff
@@ -39,7 +41,60 @@ See the [Limitations](#limitations) section below and the [Roadmap](../../refere
...
+- It validates migrations to electrified tables to ensure that changes to the schema are supported by Electric (e.g. validating the types of any added columns, ensuring that only additive migrations are applied, etc), and
+- It provides an endpoint for schema introspection to allow Electric to return it's view of the underlying postgres database to the data access library.
+
+Migrations not passed through the proxy endpoint will not be captured by Electric and will cause problems as Electric's view of the Postgresql schema will be out of sync with the actual table schema.
+
+### Configuring and connecting to the Postgresql migration proxy
+
+There are two environment variables that configure the proxy in Electric:
+
+- `PG_PROXY_PORT` (default `65432`). This is the TCP port that you should connect to in order to pass through the migration proxy. Since the proxy speaks fluent Postgres, you can connect to it via any Postgres, compatible tool, e.g. `psql -U electric -p 65432 electric`
```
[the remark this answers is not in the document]
thruflo: Do we need this to be a separate port or could it be unified with the `HTTP_PORT` using routing that falls back to the TCP proxy? If so, this would make it simpler to deploy Electric.

thruflo (commented):
Looks great. I've left a few comments.

I'll add commits to this branch to update the instructions for the other frameworks.

docs/usage/data-modelling/migrations.md
alco: I would refrain from instructing users to avoid the proxy for some migrations. That has implications with regards to which user owns the tables, plus it's more complicated to configure.

docs/usage/data-modelling/migrations.md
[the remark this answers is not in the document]
thruflo: Yeah (2) is a fair point ;)

Just annoying to lose the deployment simplicity (assuming we remove the logical publisher port by switching to direct writes). But I can’t see a way round it.

docs/usage/data-modelling/migrations.md
magnetised: I agree with @alco  here - it's conceptually simpler and less error-prone to just always apply migrations via the proxy

docs/integrations/backend/phoenix.md
magnetised: This config is only used by ecto tasks -- in your code you generally directly reference a repo module. I've certainly never referenced this config and have only had to set it when my ecto tasks fail. 

i.e. https://hexdocs.pm/ecto/Ecto.html#module-mix-tasks-and-generators

> Repo resolution
>
> Ecto requires developers to specify the key :ecto_repos in their application configuration before using tasks like ecto.create and ecto.migrate. For example:

examples/beer-stars/backend/compose.yaml
```diff
@@ -35,11 +35,14 @@ services:
...
     environment:
       DATABASE_URL: "postgresql://postgres:password@postgres:5432/beer-stars"
-      ELECTRIC_HOST: electric
-      LOGICAL_PUBLISHER_HOST: electric
-      AUTH_MODE: insecure
+      ELECTRIC_HOST: "electric"
+      LOGICAL_PUBLISHER_HOST: "electric"
+      AUTH_MODE: "insecure"
+      PG_PROXY_PASSWORD: "proxy-password"
+      PG_PROXY_PORT: "65432"
```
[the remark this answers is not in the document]
magnetised: I was going for explicit over implicit, allowing our configurations to act as examples for users but rather than add the complexity here I'll just remove it

alco (approved):

Nezteb:
When does this get deployed? https://electric-sql.com/docs/integrations/backend/phoenix 😄

thruflo:
Yes, the build was broken because other pages (and actually also GitHub READMEs) were pointing to the removed Running the examples page.

I've re-enabled the page, removing the incorrect sections, here: https://github.com/electric-sql/electric/pull/614

@alco can you approve that and when merged the build will run. Note that to test locally, run `yarn build` in the website repo having fetched and switched the submodules/electric to the branch you want to test and it will verify there are no broken links / that the CI build will work.

alco:
@thruflo thanks! I wasn't aware of the local build for docs.

PR #533, opened 2023-10-06
chore (starter): expose configurable proxy port

kevin-dp:
#529 introduces a proxy that runs on a new port (65432 in Electric's Docker container).
This PR does the following:
- exposes the port in the Docker compose files and in the `docker run` commands
- modifies the `db:psql` and `db:migrate` commands to go through the proxy (i.e. via that port)
- modifies the starter to make that port configurable

For now, in order to test this PR, you need to checkout this PR and then:
- build Electric (`cd components/electric && make docker-build`) 
- build the starter (`cd examples/starter && pnpm install && pnpm build && npm link`)
- create an app `create-electric-app my-app`
- modify your app to use the local Electric image (because it has not yet been published to Docker hub)
   i.e. in `my-app/backend/compose/.envrc` change the `ELECTRIC_IMAGE` env var to be `electric:local-build`.

PR #579, opened 2023-10-19
chore(examples): Use the updated starter template as the basis for web-wa-sqlite

alco:
This PR updates the `web-wa-sqlite` example by
* first removing it and generating a brand-new app with `create-electric-app web-wa-sqlite`, using the latest starter from https://github.com/electric-sql/electric/pull/533 (this is the first commit)
* backporting example-app-specific changes on top of this new app (2nd commit)

As you can see by the size of the [second commit](https://github.com/electric-sql/electric/pull/579/commits/754c68ff3263d9a2973f29304894c0acc6796b1b), `web-wa-sqlite` is essentially a starter app with some generated files added to it.

Do we need to keep it checked into source control? I'm thinking we could point people to the starter instead. The difference in user experience would be the one between
* cloning the monorepo, `cd` into the example directory and running `yarn`
* running `npx create-electric-app@latest`, `cd` into the created directory

One upside of going with the starter by default is that it has nicer user experience for handling port conflicts. For example, if I have something listening on port `65432` and try to run `yarn backend:start` inside `examples/web-wa-sqlite`, I get this error:
```
$ yarn backend:start
[...]
Error response from daemon: driver failed programming external connectivity on endpoint web_wa_sqlite-electric-1 (48abcf4b8c03f550c4c4c3c0723b336ac2cba0f97f73f5e82feaf3bffe993de7): Error starting userland proxy: listen tcp4 0.0.0.0:65432: bind: address already in use
 Failed to start the Electric backend. Check the output from `docker compose` above.
If the error message mentions a port already being allocated or address being already in use,
execute `yarn ports:configure` to run Electric on another port. 
Done in 1.58s.
```

Whereas when create a new app using `create-electric-app`, I get a chance to assign different ports upfront if any of the ones we require are currently taken:
```
$ create-electric-app web-wa-sqlite-2
Port 5133 for Electric is already in use.
prompt: Hit Enter to keep it or enter a different port number:  (5133) 5134
Port 65432 for Electric's proxy is already in use.
prompt: Hit Enter to keep it or enter a different port number:  (65432) 65433
Installing dependencies (may take some  time) ...
```

One reason for having the `web-wa-sqlite` example explicitly checked in could be that it's more representative of actual user apps because it has additional metadata such as the license field. However, in my testing, I'm still seeing "No license field" warnings:
```
$ cd examples/web-wa-sqlite

$ yarn backend:up
yarn run v1.22.19
warning ../../../../../package.json: No license field
$ yarn backend:start --detach
warning ../../../../../package.json: No license field
$ node ./backend/startCompose.js --detach
[+] Running 2/2
 ✔ Container web_wa_sqlite-postgres-1  Started
 ✔ Container web_wa_sqlite-electric-1  Started
Done in 1.24s.

$ yarn db:migrate
yarn run v1.22.19
warning ../../../../../package.json: No license field
$ node ./db/migrate.js
Connecting to proxy at postgresql://electric@localhost:65432/web_wa_sqlite
warning ../../../../../package.json: No license field
[...]
```


balegas:
In my perspective, while we don't have a common approach to handle starter changes on all examples, is just a small annoyance that we have to do one more. However, I think it's important to automate that step.

@thruflo I believe you'll have strong opinions on this.

alco:
I agree, this is a step we need to go through for all examples this time, to bring them all up to date. But going forward, we should re-evaluate the starter's approach as @samwillis suggested in VAX-1253.

samwillis:
All looks good to me. Only thing to note is that I think these files have previously been accidentally added to the repo, that can be deleted:
- public/index.js
- public/index.js.map
- public/wa-sqlite-async.mjs

That second commit gets even smaller!

We are going to address the issues with upgrading examples as part of "Front end tooling improvement" project:
- https://linear.app/electric-sql/project/front-end-tooling-improvement-a4fed15f6b1d/VAX
- https://electric-sql.slab.com/posts/front-end-tooling-improvement-project-4esosdsr

samwillis (approved):

PR #583, opened 2023-10-23
fix(electric): New extension migrations fail to apply on top of Electric v0.6.4 schema

magnetised:
Hard to test without using docker images and the reproduction strategy in VAX-1245, have created an issue to track the need for a test of this migration path: https://linear.app/electric-sql/issue/VAX-1248/e2e-test-to-validate-upgrade-path-from-release-to-release

### DIFF
--- a/.changeset/funny-pianos-hang.md
+++ b/.changeset/funny-pianos-hang.md
@@ -1,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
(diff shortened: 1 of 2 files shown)


alco (approved):
Brilliant!

With regards to testing, we'll need to become more restrictive in terms of the Postgres role we use. I think that'll fall out naturally of the upcoming work to survey the PG permissions needed at different stages of Electric operations.

PR #584, opened 2023-10-23
feat!: baseline backend support for where clauses

icehaunter:
Multiple considerations in this PR:
1. WHERE clauses cannot "span relations": when a table is where-claused, we disallow any children to reference it within the same shape to guarantee that the client cannot request data that will put it into inconsistent state
2. Not all types are covered, any quite few functions are present. You're welcome to request more or ask about more edge cases, but the code is generally written to be defensive - the client cannot use undefined functions, and we're not using client query input before validating we can evaluate it.
3. The function overload resolution is copied from PG's C code, so it references as-yet unsupported features (e.g. polymorphic functions using pseudo types like `anyelement`
4. Function definition currently doesn't support named argument calls or defaults. I'll add support for them when the requirement comes. 

Closes VAX-1260

Superseded by #899


components/electric/lib/electric/replication/shapes/shape_request.ex
```diff
@@ -18,33 +18,113 @@ defmodule Electric.Replication.Shapes.ShapeRequest do
...
+
+  This function will fail if the request contains a where statement that's not
+  present in the `where_statements` map.
   """
-  @spec from_satellite_request(%SatShapeReq{}) :: t()
-  def from_satellite_request(%SatShapeReq{request_id: id, shape_definition: shape}) do
+  @spec from_satellite_request(%SatShapeReq{}, %{String.t() => Eval.Expr.t()}) :: t()
+  def from_satellite_request(
+        %SatShapeReq{request_id: id, shape_definition: shape},
+        where_statements
```
alco, line 115: Should rather be `where_expressions` or `where_clauses`? Also goes for the docstring above.

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -492,7 +492,7 @@ defmodule Electric.Satellite.Serialization do
     val
   end
 
-  def decode_column_value!(val, :float8) do
+  def decode_column_value!(val, type) when type in [:float8, :numeric] do
```
alco, line 495: `:numeric` stands for arbitrary-precision numbers, doesn't it? If you try to convert a stringified number that doesn't fit in a 64-bit floating point, `String.to_float()` will fail with ArgumentError.

components/electric/lib/electric/replication/eval/env/known_functions.ex
```diff
@@ -0,0 +1,75 @@
...
+  defpostgres "*integral_type* & *integral_type* -> *integral_type*", delegate: &Bitwise.band/2
+  defpostgres "*integral_type* | *integral_type* -> *integral_type*", delegate: &Bitwise.bor/2
+  defpostgres "*integral_type* # *integral_type* -> *integral_type*", delegate: &Bitwise.bxor/2
+
+  ## String functions
+
+  defpostgres "text || text -> text" do
+    # Can't do `delegate: &Kernel.<>/2`, because it's a macro that gets converted to local function call
+    def text_concat(t1, t2) when is_binary(t1) and is_binary(t2) do
+      <<t1::binary, t2::binary>>
```
alco, line 59:
```suggestion
      t1 <> t2
```

alco (commented):
This is the first batch of comments after I have reviewed 21/33 files. I'll do another round of review later today.

components/electric/lib/electric/replication/eval/env/basic_types.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Electric.Replication.Eval.Env.BasicTypes do
+  @moduledoc """
+  This module describes basic types, their categories, and if they are preferred within said category.
```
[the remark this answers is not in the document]
icehaunter: This info is not available in PG docs; it's in the PG itself, queryable via `pg_type` table - that's the "true" source of info. I can add a note

components/electric/lib/electric/replication/eval/env/explicit_casts.ex
```diff
@@ -0,0 +1,78 @@
+defmodule Electric.Replication.Eval.Env.ExplicitCasts do
+  @moduledoc """
+  Postgres has explicit casts, achievable via `::type` calls.
+  This module defines implementations for a subset of known ones.
+  "function name" column here is a function name in this module.
```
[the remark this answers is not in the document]
icehaunter: Then it's not currently defined in Electric and thus currently unsupported

components/electric/lib/electric/replication/eval/env/known_functions.ex
```diff
@@ -0,0 +1,75 @@
...
+  alias Electric.Replication.PostgresInterop.Casting
+  alias Electric.Replication.Eval.Env.BasicTypes
+
+  ## "input" functions
+
+  defpostgres "int2(text) -> int2", delegate: &Casting.parse_int2/1
+  defpostgres "int4(text) -> int4", delegate: &Casting.parse_int4/1
+  defpostgres "int8(text) -> int8", delegate: &Casting.parse_int8/1
+  defpostgres "float4(text) -> float4", delegate: &Casting.parse_float8/1
+  defpostgres "float8(text) -> float8", delegate: &Casting.parse_float8/1
```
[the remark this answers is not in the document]
icehaunter: Hmm, first of all - good point. Issue is, that means we need to overload mathematical operators to not be just Kernel functions in order to support inf/nan behaviour like PG does. Not happy about that, but definitely possible

components/electric/lib/satellite/protocol_helpers.ex
```diff
@@ -12,11 +12,16 @@ defmodule Satellite.ProtocolHelpers do
...
-      {id, tables: tables} ->
+      {id, tables} ->
         %SatShapeReq{
           request_id: to_string(id),
           shape_definition: %SatShapeDef{
-            selects: tables |> Enum.map(&%SatShapeDef.Select{tablename: &1})
+            selects:
+              tables
+              |> Enum.map(fn
+                "" <> table -> %SatShapeDef.Select{tablename: table}
```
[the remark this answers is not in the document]
alco: `table when is_binary(table)` would be more idiomatic if less clever.

What else can it be? If a binary or a tuple are the only two options, just switch it with the other case clause and remove the pattern matching.

components/electric/lib/electric/replication/eval/env/known_functions.ex
alco: I get the impossibility to delegate part. But `<>` is still the string concatenation operator in Elixir. It's more idiomatic to use it over `<<t1:: binary, t2 :: binary>>`. All the checks you mention happen at compile time so they are of no concern to us.

components/electric/lib/electric/replication/eval/env.ex
```diff
@@ -0,0 +1,321 @@
...
+  def get_type_category(_, :unknown), do: :unknown
+  def get_type_category(_, {:composite, _}), do: :composite
+  def get_type_category(_, :composite), do: :composite
+  def get_type_category(_, name) when name in @common_polymorphic_types, do: :pseudo
+  def get_type_category(_, name) when name in @simple_polymorphic_types, do: :pseudo
+
+  def get_type_category(%__MODULE__{known_basic_types: types}, type) when is_map_key(types, type),
+    do: types[type].category
+
+  def get_type_category(_, type), do: raise({:unknown_type, type})
```
alco, line 176:
```
iex(1)> raise {:unknown_type, :text}
** (ArgumentError) raise/1 and reraise/2 expect a module name, string or exception as the first argument, got: {:unknown_type, :text}
    iex:1: (file)
```

components/electric/lib/electric/replication/eval/parser.ex
```diff
@@ -0,0 +1,497 @@
...
+
+  defmodule Ref do
+    defstruct [:path, :type, location: 0]
+  end
+
+  defmodule Func do
+    defstruct [:args, :type, :implementation, :name, strict?: true, immutable?: true, location: 0]
+  end
+
+  @valid_types ~w[
```
alco, line 25: This duplicates the list defined in `Electric.Postgres.supported_types()`. If those two are to remain in sync, I would advise removing this attribute in favor of the one from `Electric.Postgres`. Define the guard `is_supported_type` in the latter module and use it in this one.

components/electric/lib/electric/replication/eval/parser.ex
```diff
@@ -0,0 +1,497 @@
...
+
+      :error ->
+        with {:ok, [%{args: [^from_type]} = impl]} <- Map.fetch(env.funcs, out_func) do
+          {:ok, impl}
+        end
+    end
+  end
+
+  defp find_cast_function(%Env{} = env, from_type, to_type) do
+    # I know this looks like a no-op `with`, but I want this function to encapsulate the access
```
alco, line 339: I don't know what you meant by this comment, but the `with` block below:
```
        with {:ok, {module, fun}} <- Map.fetch(env.explicit_casts, {from_type, to_type}) do
          {:ok, {module, fun}}
        end
```
is exactly equivalent to
```
Map.fetch(env.explicit_casts, {from_type, to_type})
```

components/electric/test/electric/replication/eval/runner_text.exs
```diff
@@ -0,0 +1,43 @@
...
+
+    test "should not apply strict functions to nil values" do
+      assert {:ok, nil} =
+               ~S|"test" + 1|
+               |> Parser.parse_and_validate_expression!(%{["test"] => :int4})
+               |> Runner.execute(%{["test"] => nil})
+    end
+
+    test "should return error on invalid function application instead of crashing" do
+      assert {:error, %{args: ["test", 1]}} =
```
alco, line 37: Not sure where this error is further processed when encountered in non-test code but it could be helpful for debugging to include the immediate operator/function in the error message along with the arguments.

components/electric/test/electric/replication/eval/parser_test.exs
```diff
@@ -0,0 +1,259 @@
...
+  alias Electric.Replication.Eval.Env
+  alias Electric.Replication.Eval.Expr
+
+  @int_to_bool_casts %{
+    {:int4, :bool} => {ExplicitCasts, :int4_to_bool},
+    {:bool, :int4} => {ExplicitCasts, :bool_to_int4}
+  }
+
+  describe "parse_and_validate_expression/3 basics" do
+    test "should correctly parse constants" do
```
alco, line 16: This whole module looks great! Just one nitpick: in Elixir we name tests as "correctly parses constants" instead of prefixing everything with "should".

alco (approved):
:muscle:

components/electric/lib/electric/replication/eval/env/implicit_casts.ex
```diff
@@ -0,0 +1,61 @@
...
+                  |> String.split("## List of implicit casts")
+                  |> List.last()
+                  |> String.split("---- |\n")
+                  |> List.last()
+                  |> String.split("\n", trim: true)
+                  |> Enum.flat_map(fn line ->
+                    [type, targets] =
+                      line
+                      |> String.split("|", trim: true)
+                      |> Enum.map(&String.trim/1)
```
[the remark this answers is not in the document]
icehaunter: I know but I'd rather keep the duplication to be explicit. This is meant just to parse the moduledoc exactly above, don't want to make it "generic"

components/electric/lib/electric/replication/eval/env/explicit_casts.ex
```diff
@@ -0,0 +1,78 @@
...
+                  |> Enum.flat_map(fn line ->
+                    [from, to, function_name] =
+                      line
+                      |> String.split("|", trim: true)
+                      |> Enum.map(&String.trim/1)
+
+                    if function_name != "" do
+                      [
+                        {{String.to_atom(from), String.to_atom(to)},
+                         {__MODULE__, String.to_atom(function_name)}}
```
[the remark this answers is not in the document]
icehaunter: Yup, works

components/electric/lib/electric/replication/shapes/shape_request.ex
```diff
@@ -18,33 +18,113 @@ defmodule Electric.Replication.Shapes.ShapeRequest do
...
   @doc """
-  Check if the given change belongs to this shape.
+  Check if the given record belongs to this shape.
   """
-  @spec change_belongs_to_shape?(t(), Changes.change()) :: boolean()
-  def change_belongs_to_shape?(%__MODULE__{} = shape, change) do
-    change.relation in shape.included_tables
+  @spec record_belongs_to_shape?(t(), Changes.relation(), Changes.record()) :: boolean()
+  def record_belongs_to_shape?(%__MODULE__{} = shape, relation, record) do
+    if relation in shape.included_tables do
```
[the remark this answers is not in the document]
icehaunter: Yeah, good catch, but I feel `with` is going to be less readable. I've added `else` but I won't change anything else yet because this code will be deleted in the shapes PR

components/electric/lib/electric/replication/shapes/shape_request.ex
```diff
@@ -18,33 +18,113 @@ defmodule Electric.Replication.Shapes.ShapeRequest do
...
+            {false, false} -> :not_in
+            {false, true} -> :move_in
+            {true, false} -> :move_out
+            {true, true} -> :in
+          end
+        else
+          :error ->
+            # Failed to apply parse the record into refs
+            Logger.warning("""
+            Could not convert the string record values to internal types
```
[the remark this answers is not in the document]
icehaunter: True & good point, but won't address since this is deleted in next PRs

components/electric/lib/electric/satellite/serialization.ex
icehaunter: Good point, however we kinda don't support numeric yet, but the PG functions are mostly defined over numeric and not `float8`, so this is better for now for function interop

components/electric/lib/satellite/protocol_helpers.ex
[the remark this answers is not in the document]
icehaunter: since this is a helper, this is meant as a guard against improper usage when you're writing tests quickly instead of having a weird fail later

e2e/tests/02.07_where_clauses_on_subscriptions.lux
```diff
@@ -0,0 +1,44 @@
+[doc Where clauses on subscription function correctly]
+[global fail_pattern=[Ee][Rr][Rr][Oo][Rr]]
+[include _shared.luxinc]
+
+[invoke setup]
+[invoke electrify_table entries]
+
+[global user_id_1=1]
+[newshell user_1_ws1]
+    -$fail_pattern|(bad sentinel value)
```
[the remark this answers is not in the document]
icehaunter: Why? This is an explicit regex grouping

components/electric/lib/electric/replication/postgres_interop/casting.ex
```diff
@@ -0,0 +1,84 @@
...
+      iex> like?("helloo", "hell_")
+      false
+
+      iex> like?("helloo", "%o_")
+      true
+
+      iex> like?("HELLO", "hello")
+      false
+
+      iex> like?("HELLO", "hello", true)
```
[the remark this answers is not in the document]
icehaunter: sure but you're not meant to use that, there's an explicit `ilike` :)

components/electric/lib/electric/replication/eval/env/known_functions.ex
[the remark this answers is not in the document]
icehaunter: We can pull in `Decimal` module. It has Inf, Nan, and can work with arbitrary precision. I'll open a follow-up issue

components/electric/test/electric/replication/eval/runner_text.exs
icehaunter: It does, it returns a `Electric.Replication.Eval.Parser.Func` struct which has a lot of info including location in the source string

components/electric/test/electric/replication/eval/parser_test.exs
icehaunter: that's very nitpicky :D
I like reading `should` tests a little more, but I'll keep this in mind

components/electric/test/electric/replication/eval/parser_test.exs
```diff
@@ -0,0 +1,259 @@
...
+    end
+
+    test "should work with unknown constants" do
+      assert {:ok, %Expr{eval: result}} = Parser.parse_and_validate_expression("'test'")
+      assert %Const{value: "test", type: :text} = result
+    end
+
+    test "should correctly parse type casts on constants" do
+      assert {:error, "At location 0: unknown cast from type int4 to type bool"} =
+               Parser.parse_and_validate_expression("1::boolean", %{}, Env.empty())
```
magnetised, line 28: this expression works in pg, `select 1::bool` -> `t`. any particular reason to be different?

components/electric/test/electric/replication/shapes_test.exs
```diff
@@ -27,7 +55,88 @@ defmodule Electric.Replication.ShapesTest do
...
+          insert(%{"value" => "1"}),
+          insert(%{"value" => "11"}),
+          delete(%{"value" => "1"}),
+          delete(%{"value" => "11"})
+        ])
+
+      assert %{
+               changes: [
+                 %NewRecord{record: %{"value" => "11"}},
+                 %DeletedRecord{old_record: %{"value" => "11"}}
```
magnetised, line 73: can we please rename these change structs to Insert, Delete and Update.... 😅

components/electric/test/electric/replication/shapes_test.exs
```diff
@@ -137,8 +246,116 @@ defmodule Electric.Replication.ShapesTest do
...
+          [
+            "CREATE TABLE public.entries (id uuid PRIMARY KEY);",
+            "CREATE TABLE public.parent (id uuid PRIMARY KEY, value TEXT);",
+            "CREATE TABLE public.child (id uuid PRIMARY KEY, parent_id uuid REFERENCES public.parent(id));"
+          ]
+        }
+      ])
+    end
+
+    test "should fail on tables are targets of FK within same request", %{origin: origin} do
```
magnetised, line 263:
```suggestion
    test "should fail on tables which are targets of FK within same request", %{origin: origin} do
```
?

magnetised (approved):
Incredible work. I have a few comments but I suspect you have good answers...

components/electric/test/electric/replication/eval/parser_test.exs
icehaunter: It actually does. That cast is present and available on filled `Env` struct, however this is specifically an empty environment to not "test" against a cast we'll add later on

components/electric/lib/electric/replication/shapes.ex
```diff
@@ -112,4 +181,33 @@ defmodule Electric.Replication.Shapes do
...
   end
+
+  # Used in `Enum.reduce_while/3`
+  @spec parse_where_clause(%SatShapeDef.Select{}, SchemaVersion.t(), %{
+          String.t() => Eval.Expr.t()
+        }) ::
+          {:cont, %{String.t() => Eval.Expr.t()}} | {:halt, {atom(), String.t()}}
+  defp parse_where_clause(%{tablename: table, where: where}, schema, acc) do
+    refs =
+      SchemaVersion.table!(schema, {"public", table})
```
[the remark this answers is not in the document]
icehaunter: Because protobuf selects don't specify a schema. I should pull the default from the `Extension` module though, but any better suggestions?

components/electric/test/electric/replication/shapes_test.exs
icehaunter: Yes but won't accept since this test is changed in later PR - I don't want more conflicts

components/electric/test/electric/replication/shapes_test.exs
icehaunter: at some point!

components/electric/lib/electric/replication/eval/env/known_functions.ex
```diff
@@ -0,0 +1,75 @@
...
+  defpostgres "@ *numeric_type* -> *numeric_type*", delegate: &:erlang.abs/1
+  defpostgres "*integral_type* & *integral_type* -> *integral_type*", delegate: &Bitwise.band/2
+  defpostgres "*integral_type* | *integral_type* -> *integral_type*", delegate: &Bitwise.bor/2
+  defpostgres "*integral_type* # *integral_type* -> *integral_type*", delegate: &Bitwise.bxor/2
+
+  ## String functions
+
+  defpostgres "text || text -> text" do
+    # Can't do `delegate: &Kernel.<>/2`, because it's a macro that gets converted to local function call
+    def text_concat(t1, t2) when is_binary(t1) and is_binary(t2) do
```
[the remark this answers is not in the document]
icehaunter: We are! Each `defpostgres`  builds an object that has a `strict?` field, which, as in PG, if `true` doesn't even run a function if one of the arguments is `NULL`. We also have `immutable` property, which allows us to pre-run functions on constants (as is PG, again)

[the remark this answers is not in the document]
magnetised: gotcha. I actually read various bits of nil/null handling and didn't join the dots.

components/electric/lib/electric/replication/shapes.ex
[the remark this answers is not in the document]
magnetised: > Because protobuf selects don't specify a schema 

what exactly do you mean here? this is the shape api right? I think we need to support namespaced tables here. You can use `Electric.Postgres.NameParser.parse/2` to turn e.g. `Schema."Name"` to `{"schema", "Name"}`

Happy for that to be in a separate issue though, no reason for it to block this PR

Issue #581, opened 2023-10-21
Electron- Error: crypto.getRandomValues() not supported


Rajatkalsotra:
Tried this for an electron app (electron-vite), but getting this error when trying to run electrify(db, schema, config) . 

PG db schema:

CREATE TABLE IF NOT EXISTS "todos" (
    "id" TEXT NOT NULL,
    "title" TEXT NOT NULL,    
    CONSTRAINT "todo_pkey" PRIMARY KEY ("id")
);

Error message:

applying migration: 0
applying migration: 20231019152134_192
Error: crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported

Discord link: https://discord.com/channels/933657521581858818/1164503440773353516

Rajatkalsotra:
Thanks @balegas

PR #585, opened 2023-10-23
remove genUUID and unused useRandom hook

balegas:
Use of `crypto` was creating problems with some platforms. This was also redundant because we also define a `uuid()` function.

- [ ] https://github.com/electric-sql/website/pull/65


thruflo:
`genUUID` is used in a bunch of places. Examples, docs, etc.

balegas:
yep, pushed another commit right now. Checking if examples are still working (website is okay)

balegas:
Tested with Expo and React native to double-check that `uuid()` polyfills are working on mobile environments.

Had to fiddle with React Native to get it working. I initially thought that this was because of upgrading to Sonoma, but apparently is a global issue: https://github.com/CocoaPods/CocoaPods/issues/12081

I got things working for me. @kevin-dp @thruflo could one of you check that React Native is also broken for you.  If yes, I'll provide instructions.

balegas:
- [ ] revert unintended formatting changes in sessions file

thruflo:
Attempting to verify. If I can get an 8GB iOS Simulator upgrade down over pub wifi ...

thruflo:
React Native works fine for me.

thruflo:
Screenshot 2023-10-24 at 11 45 03

balegas:
Cool. No hassle installing ruby libraries?

thruflo:
Only with my own Xcode tooling.

balegas:
Will be merging this since changes are not related with the instructions for React Native failing for me. Created a new issue to review React Native instructions with more recent version of cocoa pods.

balegas:
@thruflo still need an approval

balegas:
> @balegas how can i run this?
> 
> Because how it is configured it will use the published Electric image. If i do that i get errors. Should i run against a local build? But that's a bit tricky as it requires all the changes to expose the necessary ports.

@kevin-dp not sure if I missed something, but the answer is the same way you tested applications before, because nothing changed.

kevin-dp:
> > @balegas how can i run this?
> > Because how it is configured it will use the published Electric image. If i do that i get errors. Should i run against a local build? But that's a bit tricky as it requires all the changes to expose the necessary ports.
> 
> @kevin-dp not sure if I missed something, but the answer is the same way you tested applications before, because nothing changed.

I don't think this works since we introduced the proxy and now the latest published image also uses the proxy but this app is not using the proxy.

balegas:
@samwillis pulling you in on this one after conversation yesterday.
I opened this PR because `genUUID()` was causing problems and I decided to converge code to using `uuid()`. 

As you point out, we're overriding a global variable, so maybe we need to handle this better. We do that to ensure we use the right dep for the different envs. If we have one that works across that would simplify code. I think initially we decided not to go with a js-based approach because of higher collision rate. I'm happy with that tradeoff.

If you wish to fix this as part of your refactoring, I'm very happy. Otherwise, leave a comment here so we can pick it up later.

balegas:
I'm, closing this PR and requesting #581 to be solved in VAX-1352

PR #587, opened 2023-10-23
fix(electric): @migrations version capture failing intermittently

magnetised:


kevin-dp:
Let's have a reviewer on this PR and merge the reviewed version into kevindp/vax-1180-expose-proxy-port to avoid having to rebase on every change.

components/electric/lib/electric/postgres/proxy/injector.ex
```diff
@@ -79,7 +79,34 @@ defmodule Electric.Postgres.Proxy.Injector do
...
+      %{pending_messages: pending} -> {pending ++ msgs, %{state | pending_messages: []}}
+    end
+    |> sync_messages()
+  end
+
+  defp sync_messages({msgs, state}) do
+    {out, pending} =
+      Enum.flat_map_reduce(msgs, [], fn %type{} = msg, buf ->
+        if type in [M.Sync, M.Flush, M.Query] do
+          {[Enum.reverse([msg | buf])], []}
```
alco, line 95: This reverses the list every time a new message of one of those 3 types is encountered. It makes it quite hard to parse the algorithm encoded by this code.

If we expect to see only one message of one of those 3 types in the `msgs` list, it'd be more appropriate to use `Enum.split_while()`:

```
{out, [sync_flush_query|pending]} = Enum.split_while(msgs, fn %type{} -> type not in [...] end)
{List.flatten([out, sync_flush_query]), pending}
```

If we expect to see more than one occurrence of Sync, Flush, or Query, then this code splits `msgs` at the last occurrence, right? I think I have an idea for how to make this intent read a bit more clearly.

alco (approved):

components/electric/lib/electric/postgres/proxy/injector.ex
alco: I know this is 200% less elegant than your implementation, but I think it is easier to comprehend.

```elixir
  def split_at_last_occurrence(list, fun), do: split_at_last_occurrence(list, fun, [], 0, [])

  defp split_at_last_occurrence([], _fun, acc, last_chunk_len, residue) do
    # discard the incomplete last chunk from the main list before reversing
    list = acc |> Enum.drop(last_chunk_len) |> Enum.reverse()
    {list, residue}
  end

  defp split_at_last_occurrence([h | t], fun, acc, last_chunk_len, residue) do
    if fun.(h) do
      # this could be the split point. reset the chunk length counter to have the correct value
      # of the last chunk's length once we reach the end of the list. t will be the residue then
      split_at_last_occurrence(t, fun, [h | acc], 0, t)
    else
      # just keep going over the list while holding on to the previously computed residue
      split_at_last_occurrence(t, fun, [h | acc], last_chunk_len + 1, residue)
    end
  end
```

```
defmodule M do ... end

iex> l = [:a, :b, :sync, :c, :d, :query, :e, :f, :g, :flush, :h, :i]
iex> M.split_at_last_occurrence l, & &1 in [:sync, :query, :flush]
{[:a, :b, :sync, :c, :d, :query, :e, :f, :g, :flush], [:h, :i]}

alco: The `residue` argument in my implementation can be removed and computed from `acc` by replacing `Enum.drop()` with `Enum.split()`. Storing the value in a variable is simply a micro-optimization to save on one `Enum.reverse()` call.

magnetised: yeah, wasn't happy with the result here but didn't want to spend too much time on it. gonna merge and if you feel the urge work away

PR #588, opened 2023-10-24
chore(electric): Rewrite the splitting algorithm in Proxy.Injector to be more concise

alco:
This is a follow-up to https://github.com/electric-sql/electric/pull/587#discussion_r1368945549.


alco:
@magnetised The list of commits here is changing due to ongoing rebases. All I wanted you to take a look at is the last commit -- that's the only one that's new.

magnetised (approved):
tweaky, love it

PR #590, opened 2023-10-24
chore(examples): Update Linearlite to use the Proxy and new types

samwillis:
This updates linearlite to use the Proxy and new types.

Of the new times we are using `timestamp` and 'uuid'.

The only difference between the files from the starter and their copies in this example is that they have been renamed from `*.js` to `*.cjs`. This example has `"type": "module"` set in `package.json`, by renaming the scripts to `*.cjs` they continue to work unmodified. Without this they would have to be converted to ESM with "import" rather than "require" syntax.

`builder.cjs` is included in the project but not used to build the app, it uses Vite. It is inspected for port definitions by other scripts from the starter and so cannot be deleted without modifying them.

Testing this PR requires running the app against a locally built Electric, ts-client, and generator.
- modify linearlite to use the local Electric image (because it has not yet been published to Docker hub)
   i.e. in `linearlite/backend/compose/.envrc` change the `ELECTRIC_IMAGE` env var to be `electric:local-build`.
- modify linearlite to use our packed ts-client:
    - change the electric-sql dependency in `linearlite/package.json` to be `"electric-sql": "file:<path to electric mono 
    - change the `@electric-sql/prisma-generator` dependency in `linearlite/package.json` to be `"@electric-sql/prisma-generator": "file:<path to electric mono 
- install dept
   - `npm install`
- Download [data.zip](https://github.com/electric-sql/electric/files/13215996/data.zip) and extract to db/data
- now run the app: 
   - `npm run reset` to start the docker images, run migrations, load data, generate client
   - `npm run start` to start the dev server


examples/linearlite/db/migrations/01-create_tables.sql
```diff
@@ -5,8 +5,8 @@ CREATE TABLE IF NOT EXISTS "issue" (
     "description" TEXT NOT NULL,
     "priority" TEXT NOT NULL,
     "status" TEXT NOT NULL,
-    "modified" TEXT NOT NULL,
-    "created" TEXT NOT NULL,
+    "modified" TIMESTAMP NOT NULL,
```
alco, line 8: `TIMESTAMPTZ` is a better choice for storing a point in time because it's interpreted correctly in different time zones by the clients. `TIMESTAMP`, on the other hand, works more like a photo of a wall clock, so every client will see the same time but will think that it's local to their time zone.

thruflo: Could we not also use UUIDs for the primary key IDs?

samwillis: Currently the ids from the imported dataset are ints (as strings), only new issues/comments get uuids. 

A future change would be to assign a UUID to each imported issue, maybe deterministically, an maybe something to then show as a demo of how to do it.

alco (approved):

examples/linearlite/package.json
```diff
@@ -1,24 +1,25 @@
...
-    "backend:start": "docker compose --env-file backend/compose/.envrc -f backend/compose/docker-compose.yaml up",
-    "backend:stop": "npm run backend:down",
-    "backend:up": "docker compose --env-file backend/compose/.envrc -f backend/compose/docker-compose.yaml up --detach",
-    "client:generate": "npx electric-sql generate",
-    "client:watch": "npx electric-sql generate --watch",
-    "db:migrate": "node ./db/migrate.js",
-    "db:psql": "node ./db/connect.js",
+    "backend:start": "node ./backend/startCompose.cjs",
+    "backend:stop": "docker compose --env-file backend/compose/.envrc -f backend/compose/docker-compose.yaml stop",
+    "backend:up": "node ./backend/startCompose.cjs --detach",
```
kevin-dp, line 9: could we call `backend:start` and pass the `--detach` option?
Like so: `npm run backend:start --detach`

examples/linearlite/package.json
```diff
@@ -1,24 +1,25 @@
...
     "db:load-data": "node ./db/load_data.js",
     "db:process-data": "node ./db/process_data.js",
     "db:truncate": "node ./db/truncate.js",
     "db:reload-data": "npm run db:truncate && npm run db:load-data",
+    "ports:configure": "node ./change-ports.js",
+    "electric:start": "node ./backend/startElectric.cjs",
+    "electric:check": "node ./check-electric-is-running.cjs",
     "reset": "npm run backend:down && npm run backend:up && sleep 5 && npm run db:migrate && npm run client:generate && npm run db:load-data",
-    "electric:start": "node ./backend/startElectric.js",
     "prepare-wa-sqlite": "cp ./node_modules/wa-sqlite/dist/wa-sqlite-async.mjs ./public/wa-sqlite-async.mjs && cp ./node_modules/wa-sqlite/dist/wa-sqlite-async.wasm ./public/wa-sqlite-async.wasm",
```
kevin-dp, line 23: We should use the dedicated script for this here: `copy-wasm-files.cjs`

kevin-dp (commented):
I had a look at the code, great work!
Left some minor comments.

Will test the PR now based on your instructions.

kevin-dp:
I followed your steps to run the app but it fails on `npm run reset` because the issue data does not seem to be present.
Should it be bundled with the app or should it be fetched by a script somehow?
This is the error i get:

```
Error: ENOENT: no such file or directory, open '/Users/kevin/Documents/Electric/development/electric/examples/linearlite/db/data/issues.json'
```

PR #591, opened 2023-10-25
DDLX parser does not correctly handle uppercase identifiers

magnetised:
also fixes VAX-1220


components/electric/test/electric/ddlx/parser_test.exs
```diff
@@ -0,0 +1,595 @@
...
+  alias Electric.DDLX.Parser
+  alias Electric.DDLX.Command
+
+  describe "ENABLE ELECTRIC" do
+    test "parse enable" do
+      sql = "ALTER TABLE things ENABLE ELECTRIC;"
+      {:ok, result} = Parser.parse(sql)
+
+      assert result == %Command.Enable{
+               table_name: {"public", "things"}
```
alco, line 14: I'm asking this without fully grasping the data flow in the new parser: do we have to assign a default schema to the result of parsing if one is missing in the input?

In practice, the upstream connection from Electric to Postgres may have a different schema set in its `search_path`. That's fully under control of the developer and outside of Electric's control. If the default schema used in a particular Postgres instance is, say, `my_app`, then the SQL statements into which Electric "injects" the `public` schema will fail.

alco: We should at least document it as a limitation on the website if it's too much work to make Electric work with a default schema that's named something other than `public`.

components/electric/lib/electric/ddlx/parser/build.ex
```diff
@@ -0,0 +1,90 @@
...
+    if value1 == value2 do
+      {:ok, value1}
+    else
+      {:error,
+       "#{name1} must equal #{name2}: got #{name1}: #{inspect(value1)} #{name2}: #{inspect(value2)}"}
+    end
+  end
+
+  def split_role_def(role_def, opts) do
+    # TODO: validate that none of these are the empty string
```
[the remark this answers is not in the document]
magnetised: remnant -- did to that validation.. thanks

components/electric/test/electric/ddlx/parser_test.exs
magnetised: good catch, something that's been lurking in my mind. there are a *lot* of places where we have to assume a default schema and at the moment this is hard coded as `public`. I'll add a ticket to make that configurable, which I don't  think is particularly hard TBH

magnetised: https://linear.app/electric-sql/issue/VAX-1273/allow-for-configuration-of-a-default-schema-other-than-public

Issue #473, opened 2023-09-21
[VAX-893] Android support current broken.


thruflo:
Was just testing the expo and RN examples in Android. Both throw this error:

`Queries can be performed using SQLiteDatabase query or rawQuery methods only.`

There's a few results online, e.g.: [https://www.b4x.com/android/forum/threads/sqlite-and-pragma.16410/](https://www.b4x.com/android/forum/threads/sqlite-and-pragma.16410/)

Needs more investigation but seems like we need to change the functions we're using to execute SQL statements. This currently blocks anything working on Android.

<sub>From [SyncLinear.com](https://synclinear.com) | [VAX-893](https://linear.app/electric-sql/issue/VAX-893/android-support-current-broken)</sub>

balegas:
We've identified the cause. There is a fix already available for Expo, but we need to investigate RN driver

kevin-dp:
Additional details can be found here:
[](https://github.com/expo/expo/issues/24283)

balegas:
We now have Expo working, but still missing React Native

kevin-dp:
> We now have Expo working, but still missing React Native

Actually, expo-sqlite pushed a workaround but we are not yet using it.

kevin-dp:
This was fixed in https://github.com/electric-sql/electric/pull/592 and is now released in electric-sql v0.7.0.

PR #592, opened 2023-10-25
fix(client): modify expo-sqlite driver to work on Android

kevin-dp:
This PR addresses #473 for expo applications running on Android.
The crux of the problem was that we execute SQL queries containing a `RETURNING` clause but the sqlite driver for Android did not support this. To address this issue, we use the new `execRawQuery` which is provided by expo-sqlite as a workaround (cf. [https://github.com/expo/expo/pull/24320](https://github.com/expo/expo/pull/24320)). However, this requires rewriting our database adapter for expo-sqlite such that it uses `execRawQuery` instead of expo-sqlite's transactional support. As a result, we now have to manage transactions manually using `BEGIN` and `COMMIT`/`ABORT` statements. This is very similar to the way how our database adapter for wa-sqlite works. Therefore, i decided to extract that implementation into an abstract `DatabaseAdapter` class (cf. `clients/typescript/src/drivers/generic/database.ts`) and have the expo-sqlite and wa-sqlite adapters extend it.


thruflo (approved):
Looks good to me 👍

samwillis:
Looks good to me too.

My one comment would be to maybe name the new generic adapter something like `ManualMutexDatabaseAdapter` to make it more explicit it's not a generic adapter for all use cases?

We may want to refactor the capacitor adaptor to use this too (in another PR)

kevin-dp:
> Looks good to me too.
> 
> My one comment would be to maybe name the new generic adapter something like `ManualMutexDatabaseAdapter` to make it more explicit it's not a generic adapter for all use cases?
> 
> We may want to refactor the capacitor adaptor to use this too (in another PR)

Yes exactly. I used `DatabaseAdapter` because that's how we call every database adapter and we differentiate based on the folder in which they are defined. So perhaps we rename the folder from "generic" to "manualMutex" ?

Issue #569, opened 2023-10-15
deploying to fly


spence:
hey guys - awesome tech! 

this was almost exactly what i was looking for (pg interface with a logically-replicated sqlite + crdt on globally distributed instances). i've been playing around with [nqlite](https://github.com/voxoco/nqlite) (nats as a sqlite active-active replicator + yjs), but it's not a robust solution. ideally, i'd like to quickly instantiate many (thousands) of scoped sqlite from pg + replicate crdt-based writes per instance during runtime.

i ran into a few issues deploying electric to fly but ran out of time before i could get things running.

1. `:epgsql.connect` &  ipv6: there's a check for ECTO_IPV6 for the electric endpoints, but on fly, my pg instance is also only available via ipv6.  i added:
```ex
    connection_opts = config
    |> Keyword.fetch!(:connection)
    |> new_map_with_charlists()
    |> set_replication(replication?)
    |> Map.put_new(:tcp_opts, [:inet6])
```

2. composing LOGICAL_PUBLISHER_HOST from fly envs. the ideal hostname is:
```ex
"#{env.FLY_MACHINE_ID}.vm.#{env.FLY_APP_NAME}.internal"
```
but this is a bit tricky to setup since fly config nor fly secrets support interpolating from existing values. i couldn't use `[[processes]]` to inject env vars since dockerfile sets the ENTRYPOINT. i ended up forking electric. 

3. i'm now showing an error i don't quite understand, but this is where i left it
```ex
:error, "55006",
:object_in_use,
"replication slot \"electric_replication_out_test\" is active for PID 22227",
[file: "slot.c", line: "516", routine: "ReplicationSlotAcquire", severity: "ERROR"]
```

cheers

alco:
Hey @spence. Thanks a lot for sharing the info about Fly and your troubleshooting steps! We'll look into  polishing the deployment experience promptly.

Could you elaborate a bit on the second point? Did you fork Electric to make it do the environment variable interpolation in the above string that you're passing to it as a literal `LOGICAL_PUBLISHER_HOST="#{env.FLY_MACHINE_ID}...`?

spence:
@alco yeah exactly:

```ex
  fly_machine_id =
    System.get_env("FLY_MACHINE_ID") ||
      raise("Required environment variable FLY_MACHINE_ID is not set")
  fly_app_name =
    System.get_env("FLY_APP_NAME") ||
      raise("Required environment variable FLY_APP_NAME is not set")
[...]

PR #594, opened 2023-10-25
feat(electric): Add support for IPv6 networks (e.g. the 6PN private network on Fly.io)

alco:
Closes VAX-1263. Fixes VAX-1266.

This was originally reported in https://github.com/electric-sql/electric/issues/569.


magnetised (approved):
excellent tcp listen options mangling

PR #596, opened 2023-10-26
bug(satellite): Cleanup startingPromise in registry when the Satellite process fails to start

davidmartos96:
When the satellite process fails (local state is invalid or out of date, invalid migrations, etc...), if the app cleans up the state and retries again (without a browser refresh), it currently will keep failing, because the `startingPromise` is being cached and not released when the promise rejects.

### DIFF
--- a/.changeset/shiny-steaks-raise.md
+++ b/.changeset/shiny-steaks-raise.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Cleanup startingPromise in registry when the Satellite process fails to start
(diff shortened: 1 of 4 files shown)


davidmartos96:
@balegas Hello Valter, could you review this PR? I believe you would be the most appropriate by looking at the change history in the file. Thank you! :smile: 

There is also this PR pending from some weeks ago :sweat_smile: 
https://github.com/electric-sql/electric/pull/476

balegas:
It looks right and test proves it. We haven't looked much into the registry yet and we have plans to address the lifecycle questions with processes that @davidmartos96 has raised, but we haven't dedicated the time.

I'll pull in @kevin-dp to give his ok on code.

kevin-dp (approved):
Thank you for this contribution!
Great work, nothing to say :)

PR #597, opened 2023-10-26
chore(electric): Rename leftover ddlx_to_commands() call

alco:


magnetised (approved):
how did I miss that? thanks for catching

PR #598, opened 2023-10-30
Add test to verify pg types are used correctly when serializing

davidmartos96:
This tests the behavior of the `getColumnType` function which uses the DbSchema Postgres types and fallbacks to the type name in the Relation.
Bools are inferred as INTEGER from SQLite, but Satellite encodes them in a special way, not like integers.

There is also a small bugfix in the tests dbDescription, which doesn't affect the test suite nor the new test.

### DIFF
--- a/clients/typescript/src/util/relations.ts
+++ b/clients/typescript/src/util/relations.ts
@@ -1,4 +1,3 @@
-
 import { SatRelation_RelationType } from '../_generated/protocol/satellite'
 import { DatabaseAdapter } from '../electric/adapter'
 import { SatelliteOpts } from '../satellite/config'
--- a/clients/typescript/src/util/relations.ts
+++ b/clients/typescript/src/util/relations.ts
@@ -1,0 +1,69 @@
+
+import { SatRelation_RelationType } from '../_generated/protocol/satellite'
+import { DatabaseAdapter } from '../electric/adapter'
+import { SatelliteOpts } from '../satellite/config'
+import { Relation, RelationsCache } from './types'
+
+// TODO: Improve this code once with Migrator and consider simplifying oplog.
+export async function inferRelationsFromSQLite(
+  adapter: DatabaseAdapter,
+  opts: SatelliteOpts
+): Promise<{ [k: string]: Relation }> {
+  const tableNames = await _getLocalTableNames(adapter, opts)
+  const relations: RelationsCache = {}
+
+  let id = 0
+  const schema = 'public' // TODO
+  for (const table of tableNames) {
+    const tableName = table.name
+    const sql = 'SELECT * FROM pragma_table_info(?)'
+    const args = [tableName]
+    const columnsForTable = await adapter.query({ sql, args })
+    if (columnsForTable.length == 0) {
+      continue
+    }
+    const relation: Relation = {
+      id: id++,
+      schema: schema,
+      table: tableName,
+      tableType: SatRelation_RelationType.TABLE,
+      columns: [],
+    }
+    for (const c of columnsForTable) {
+      relation.columns.push({
+        name: c.name!.toString(),
+        type: c.type!.toString(),
+        isNullable: Boolean(!c.notnull!.valueOf()),
+        primaryKey: Boolean(c.pk!.valueOf()),
+      })
+    }
+    relations[`${tableName}`] = relation
+  }
+
+  return Promise.resolve(relations)
+}
+
+async function _getLocalTableNames(
+  adapter: DatabaseAdapter,
+  opts: SatelliteOpts
+): Promise<{ name: string }[]> {
+  const notIn = [
+    opts.metaTable.tablename.toString(),
+    opts.migrationsTable.tablename.toString(),
+    opts.oplogTable.tablename.toString(),
+    opts.triggersTable.tablename.toString(),
+    opts.shadowTable.tablename.toString(),
+    'sqlite_schema',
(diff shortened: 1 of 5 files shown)


kevin-dp:
Thanks for the PR!
I think the unit test belongs in `client.test.ts` or even better in the dedicated `serialization.test.ts`

kevin-dp (commented):
It would be good if we also test the fallback.
i.e. have a second unit test for serialising a row of a table that is not known in the `DbSchema` but for which the relations holds the PG type (because at runtime when receiving a migration those SatRelation messages do contain PG types and thus a PG type will be stored in the relations cache).

davidmartos96:
@kevin-dp  That would involve hardcoding the SQLite inferred relations, whereas the test here, because it calls `start`, is loading the relations that Satellite loads when starting using `getLocalRelations`. 
I wasn't sure how to/where to define it without hardcoding the SQLite relation.

kevin-dp:
> @kevin-dp That would involve hardcoding the SQLite inferred relations, whereas the test here, because it calls `start`, is loading the relations that Satellite loads when starting using `getLocalRelations`. I wasn't sure how to/where to define it without hardcoding the SQLite relation.

Sure that is a good thing.
You are already mocking the db description, so for the fallback you could also mock the relation that is passed to `serializeRow` and `deserializeRow`.

davidmartos96:
> > @kevin-dp That would involve hardcoding the SQLite inferred relations, whereas the test here, because it calls `start`, is loading the relations that Satellite loads when starting using `getLocalRelations`. I wasn't sure how to/where to define it without hardcoding the SQLite relation.
> 
> Sure that is a good thing. You are already mocking the db description, so for the fallback you could also mock the relation that is passed to `serializeRow` and `deserializeRow`. Otherwise it will get complex because you will have to mimick a `SatRelation` message arriving on the satellite client (although do-able in the client tests i don't think it is actually needed for testing the serialization).

Sorry, I should have quoted the message. I was replying to why the test is in process.test.ts
What I'm thinking about is moving out the `getLocalRelations` funtion from the process into util `relations.ts`, and receive a DatabaseAdapter via parameter. That way this could be moved into serialization/client tests and work with a memory database.

1. Load memory db
2. Create table statements
3. Infer SQLite relations (without using the process)
4. Serialize/Deserialize assertions 

What do you think?

kevin-dp:
> > > @kevin-dp That would involve hardcoding the SQLite inferred relations, whereas the test here, because it calls `start`, is loading the relations that Satellite loads when starting using `getLocalRelations`. I wasn't sure how to/where to define it without hardcoding the SQLite relation.
> > 
> > 
> > Sure that is a good thing. You are already mocking the db description, so for the fallback you could also mock the relation that is passed to `serializeRow` and `deserializeRow`. Otherwise it will get complex because you will have to mimick a `SatRelation` message arriving on the satellite client (although do-able in the client tests i don't think it is actually needed for testing the serialization).
> 
> Sorry, I should have quoted the message. I was replying to why the test is in process.test.ts What I'm thinking about is moving out the `getLocalRelations` funtion from the process into util `relations.ts`, and receive a DatabaseAdapter via parameter. That way this could be moved into serialization/client tests and work with a memory database.
> 
>     1. Load memory db
> 
>     2. Create table statements
> 
>     3. Infer SQLite relations (without using the process)
> 
>     4. Serialize/Deserialize assertions
> 
> 
> What do you think?

Yes sounds good!

clients/typescript/src/util/relations.ts
```diff
@@ -0,0 +1,68 @@
...
+        name: c.name!.toString(),
+        type: c.type!.toString(),
+        isNullable: Boolean(!c.notnull!.valueOf()),
+        primaryKey: Boolean(c.pk!.valueOf()),
+      })
+    }
+    relations[`${tableName}`] = relation
+  }
+
+  return Promise.resolve(relations)
```
kevin-dp, line 42: Now that we are touching this code i'm noticing some stylistic issues with the original code.
Let's fix it now as we are touching that code anyway. We can just `return relations` here.

clients/typescript/src/util/relations.ts
```diff
@@ -0,0 +1,68 @@
...
+    'sqlite_sequence',
+    'sqlite_temp_schema',
+  ]
+
+  const tables = `
+      SELECT name FROM sqlite_master
+        WHERE type = 'table'
+          AND name NOT IN (${notIn.map(() => '?').join(',')})
+    `
+  return (await adapter.query({ sql: tables, args: notIn })) as {
```
kevin-dp, line 65: Same here, we can clean up this code:
```ts
function _getLocalTableNames(
  adapter: DatabaseAdapter,
  opts: SatelliteOpts
): Promise<Array<{ name: string }>> {
  ...
  return adapter.query({ sql: tables, args: notIn }) as Array<{ name: string }>
}
```

kevin-dp (approved):
Great work! Thank you for this contribution 💯 
I left some very minor comments which we should fix and then we can merge it

kevin-dp:
I had to rebase my branches on main which yields some conflicts here.
Normally a rebase onto will fix all conflicts automatically:
```
git rebase --update-refs --onto kevindp/vax-996-generator-expose-pg-type-info bdf4f465f215eebb52df971866e8563cdfe93ead davidmartos96:pg-types-test
```

davidmartos96:
@kevin-dp Done!

PR #602, opened 2023-10-30
chore(electric): Run Electric with an unprivileged PG role in dev

alco:


rodo:
A real need for company who work with public cloud and have already set up their Logical replication.
A question by the way, is the superuser privilege needed only to setup the logical replication ?

PR #603, opened 2023-10-31
chore(docs): Various updates and fixes for the docs

alco:
Closes VAX-1247, VAX-1274.


thruflo (approved):
Lots of nice catches, great stuff 👍

Perhaps we could keep the “no special durability requirements” wording?

alco:
> Perhaps we could keep the “no special durability requirements” wording?

As you wish.

PR #604, opened 2023-10-31
feat (DAL): support for enumerations

kevin-dp:
This PR adds support for enumerations by:
- modifying the generator to store enumeration types as "TEXT" type in the generated client
- modifying satellite to serialise/deserialise enumerations as text
  - since enumeration types are unknown to Satellite it serialises/deserialises them as text

### DIFF
--- a/.changeset/moody-carpets-allow.md
+++ b/.changeset/moody-carpets-allow.md
@@ -1,0 +1,6 @@
+---
+"electric-sql": patch
+"@electric-sql/prisma-generator": patch
+---
+
+Adds client-side support for enumerations.
--- a/clients/typescript/src/satellite/client.ts
+++ b/clients/typescript/src/satellite/client.ts
@@ -1128,15 +1128,6 @@ function deserializeColumnData(
   columnType: PgType
 ): string | number {
   switch (columnType) {
-    case PgBasicType.PG_CHAR:
-    case PgDateType.PG_DATE:
-    case PgBasicType.PG_TEXT:
-    case PgDateType.PG_TIME:
-    case PgDateType.PG_TIMESTAMP:
-    case PgDateType.PG_TIMESTAMPTZ:
-    case PgBasicType.PG_UUID:
-    case PgBasicType.PG_VARCHAR:
-      return typeDecoder.text(column)
     case PgBasicType.PG_BOOL:
       return typeDecoder.bool(column)
     case PgBasicType.PG_INT:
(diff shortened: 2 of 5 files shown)


magnetised (approved):
have we dropped the idea of mapping enums to constants? it was more complicated but nicer

kevin-dp:
> have we dropped the idea of mapping enums to constants? it was more complicated but nicer

I do not completely recall what we discussed about constants. Do you have an example?
The generator maps the enumeration type to PG's `TEXT` type but the generated client will only allow the enumeration's values as input.

For example, if PG defines an enumeration called `Color` with three values `RED`, `GREEN`, and `BLUE` then the generated client will only allow the strings `"RED"`, `"GREEN"`, and `"BLUE"` to be passed in (both at the type level and at runtime). However, since we mapped the enum to `TEXT` type, the ts-client will serialise/deserialise them to/from text, which is what we want.

magnetised:
> > have we dropped the idea of mapping enums to constants? it was more complicated but nicer
> 
> I do not completely recall what we discussed about constants. Do you have an example? The generator maps the enumeration type to PG's `TEXT` type but the generated client will only allow the enumeration's values as input.
> 
> For example, if PG defines an enumeration called `Color` with three values `RED`, `GREEN`, and `BLUE` then the generated client will only allow the strings `"RED"`, `"GREEN"`, and `"BLUE"` to be passed in (both at the type level and at runtime). However, since we mapped the enum to `TEXT` type, the ts-client will serialise/deserialise them to/from text, which is what we want.

IIRC we were discussing the benefits of an enum type like (apologies for the probably-pseudo-typescript):

```
struct MyEnum {
    RED: "RED",
    GREEN: "GREEN",
}
```

which would allow the developer to write `MyEnum.RED` and have that type-checked rather than just using strings which could contain e.g. typos that would throw runtime errors.

We did discuss how that made everything more complicated in terms of migrations, so maybe that's why it dropped off.

kevin-dp:
> > > have we dropped the idea of mapping enums to constants? it was more complicated but nicer
> > 
> > 
> > I do not completely recall what we discussed about constants. Do you have an example? The generator maps the enumeration type to PG's `TEXT` type but the generated client will only allow the enumeration's values as input.
> > For example, if PG defines an enumeration called `Color` with three values `RED`, `GREEN`, and `BLUE` then the generated client will only allow the strings `"RED"`, `"GREEN"`, and `"BLUE"` to be passed in (both at the type level and at runtime). However, since we mapped the enum to `TEXT` type, the ts-client will serialise/deserialise them to/from text, which is what we want.
> 
> IIRC we were discussing the benefits of an enum type like (apologies for the probably-pseudo-typescript):
> 
> ```
> struct MyEnum {
>     RED: "RED",
>     GREEN: "GREEN",
> }
> ```
> 
> which would allow the developer to write `MyEnum.RED` and have that type-checked rather than just using strings which could contain e.g. typos that would throw runtime errors.
> 
> We did discuss how that made everything more complicated in terms of migrations, so maybe that's why it dropped off.

Oh i see. To be honest, there's not much difference because the strings you pass in will still be typed. So if you'd pass "ORANGE" to the enum i described above, the typechecker will complain. But you have a good point, perhaps i could modify the generator such that the generated client also exports the  enum.

PR #606, opened 2023-11-01
chore(electric): Tag Electric's Docker images with a minor version

alco:
This is an idiom in the Docker community that allows consumers of the image to follow all versions of an image up until the next major release.

### DIFF
--- a/components/electric/Makefile
+++ b/components/electric/Makefile
@@ -66,14 +66,20 @@ CACHING_SETTINGS_ELECTRIC := --cache-to type=gha,mode=max,scope=${GITHUB_REF_NAM
 CACHING_SETTINGS_WS_CLIENT := --cache-to type=gha,mode=max,scope=${GITHUB_REF_NAME}-ws-client --cache-from type=gha,scope=${GITHUB_REF_NAME}-ws-client
 endif
 
+# Cut off the patch component of the version and anything else that's trailing it,
+# leaving only the leading X.Y parts of the version.
+ELECTRIC_VERSION_MINOR != echo "${ELECTRIC_VERSION}" | awk -F. '{print $$1 "." $$2}'
+
 docker-build-ci:
 	docker buildx build --load --build-arg ELECTRIC_VERSION=${ELECTRIC_VERSION} \
       -t ${ELECTRIC_IMAGE_NAME}:${ELECTRIC_VERSION} \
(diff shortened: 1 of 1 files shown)


components/electric/Makefile
```diff
@@ -66,14 +66,20 @@ CACHING_SETTINGS_ELECTRIC := --cache-to type=gha,mode=max,scope=${GITHUB_REF_NAM
 CACHING_SETTINGS_WS_CLIENT := --cache-to type=gha,mode=max,scope=${GITHUB_REF_NAME}-ws-client --cache-from type=gha,scope=${GITHUB_REF_NAME}-ws-client
 endif
 
+# Cut off the patch component of the version and anything else that's trailing it,
+# leaving only the leading X.Y parts of the version.
+ELECTRIC_VERSION_MINOR != echo "${ELECTRIC_VERSION}" | awk -F. '{print $$1 "." $$2}'
```
samwillis, line 71: I'm suspicious the default version of Make that on Mac doesn't support the "!=" syntax here.

My quick and dirty test:
```make
ELECTRIC_VERSION = "0.7.0"

ELECTRIC_IMAGE_NAME = "electric"

ELECTRIC_VERSION_MINOR != echo "${ELECTRIC_VERSION}" | awk -F. '{print $1 "." $2}'

test:
	echo "${ELECTRIC_IMAGE_NAME}:${ELECTRIC_VERSION}"
	echo "${ELECTRIC_IMAGE_NAME}:${ELECTRIC_VERSION_MINOR}"
	echo "${ELECTRIC_IMAGE_NAME}:latest"
```

I get:

```
echo ""electric":"0.7.0""
electric:0.7.0
echo ""electric":"
electric:
echo ""electric":latest"
electric:latest
```

If I change to:
```make
ELECTRIC_VERSION_MINOR := $(shell echo "${ELECTRIC_VERSION}" | awk -F. '{print $$1 "." $$2}')
```
I get:

```
echo ""electric":"0.7.0""
electric:0.7.0
echo ""electric":0.7"
electric:0.7
echo ""electric":latest"
electric:latest
```

alco: @samwillis for some reason, in your quick and dirty test you lost the doubling of the dollar signs in the invocation of `awk`.

And you can use `@echo` instead of `echo` to avoid the noise in the output.

samwillis: Yep, oops, that was a copy+past error. With the double "$$" it doesn't work.

PR #611, opened 2023-11-01
chore(examples): Update examples/ionic-demo to use the latest starter template

alco:


samwillis (approved):
All good. Tested and works

PR #612, opened 2023-11-01
chore(examples): Update examples/beer-stars to work with Electric Proxy

alco:


examples/beer-stars/package.json
```diff
@@ -12,7 +12,7 @@
     "backend:up": "yarn electric:pull && docker compose --env-file backend/.envrc -f backend/compose.yaml up --detach postgres electric",
     "build": "rm -rf dist && node builder.js",
     "build:prod": "rm -rf dist && NODE_ENV=production node builder.js",
-    "client:generate": "npx --no-install electric-sql generate --out ./src/generated/client",
+    "client:generate": "electric-sql generate --service http://localhost:5133 --proxy postgresql://prisma:proxy_password@localhost:65432/electric",
```
thruflo, line 15: Just curious -- why username `prisma` with Ecto?

examples/beer-stars/server/lib/beer_stars/release.ex
```diff
@@ -7,6 +7,7 @@ defmodule BeerStars.Release do
   alias Ecto.Migrator
 
   alias BeerStars.Model
+  alias BeerStars.Repo
```
thruflo, line 10: Also curious -- I thought from Garry's instructions that the trick here would be to use either the passed in `repo` argument or explicitly change to use the `ProxyRepo`.

examples/beer-stars/package.json
alco: `prisma` is used to let the proxy know that it needs to switch to the "prisma mode". This is client-specific, has nothing to do with Ecto.

thruflo: Yeah, but why does it need to switch to the prisma mode here?!

alco: That's how our generator works now that it talks to Postgres instead of deriving a Prisma schema from a temporary SQLite table.

The Proxy has a special "prisma mode" that makes it look like only Electrified tables are part of the DB schema. This is needed because Prisma itself doesn't have controls to limit what it introspects.

thruflo: I see, so our generator always needs to connect in prisma mode.

For the record, I think this is pretty fragile. I'd suggest it would be better to have another special username for `generator` that we document must be used when connecting as the generator. Otherwise why would a user know that swapping the username is meaningful to the schema shape the generator will see?

alco: Actually, it might be possible to get Prisma to do what we want it to do by setting up a Postgres role for it with a curated set of privileges. We didn't pursue that option due to apparent complexity but what I'm saying is that having to use a certain role for a certain task is not unusual.

But you probably mean that having this undocumented is fragile. And I agree. And we're likely to change the way it works when we slip off the "superuser' needle and start approach Postgres roles and privileges more systematically. I think then we'll need to have the proxy use the username passed it to authenticate with Postgres, and so the developer will have to be more aware about the roles being used. This is yet to be re-evaluated.

PR #619, opened 2023-11-02
fix (examples): fixes for expo example app

kevin-dp:
Fixes the version of expo-sqlite that is used and the height that is allocated to the items display.


alco (approved):
:+1:

PR #621, opened 2023-11-02
chore (examples): bump Electric versions in react-native example app

kevin-dp:
This PR bumps the version of the Electric image and JS packages that are used by the react-native example app.
I was, however, not able to test it on Android because i can't get the app to run on Android.


examples/react-native/ios/Podfile.lock
```diff
@@ -722,4 +722,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: d3187d26f4fd14b41ad0f108319f2cd79b303f24
 
-COCOAPODS: 1.12.1
+COCOAPODS: 1.11.3
```
alco, line 725: why the downgrade?

alco (approved):

examples/react-native/ios/Podfile.lock
kevin-dp: Don't know why it happened. I followed the readme instructions and it happened to downgrade it. I wasn't sure whether or not i should push this lockfile?

alco: Could it be that you had downgraded your local version of Cocoapods at some point to work around some problem?

kevin-dp: Reverted to 1.12.1

PR #622, opened 2023-11-02
chore(starter): Update version pinning behaviour in the starter template

alco:
JS deps were easy to change.

As for the Docker image, in order to find available tags, we have to reach out to Docker Hub's API. That was a bit over my pay-grade in terms of implementing that quickly in TypeScript as part of the starter. I've added a command-line flag that allows passing the image tag, as demonstration.

Now, we can actually fetch the list of tags from Docker Hub API without authentication and then pick the one that has the same digest as `latest` and whose name matches the `x.y` version format. Here's what I mean:

```
$ curl https://hub.docker.com/v2/namespaces/electricsql/repositories/electric/tags >electric-tags.txt

$ cat electric-tags.txt | jq '.results[]
                              | select(.name == "latest")
                              | {name: .name, digest: .digest}'
{
  "name": "latest",
  "digest": "sha256:adcc8804fb8d299ed38317339467b2fc4792da7cdd5cae69227cfc47363a9897"
}

$ cat electric-tags.txt | jq '.results[]
                              | select(.digest == "sha256:adcc8804fb8d299ed38317339467b2fc4792da7cdd5cae69227cfc47363a9897")
                              | {name: .name, digest: .digest}'
{
  "name": "latest",
  "digest": "sha256:adcc8804fb8d299ed38317339467b2fc4792da7cdd5cae69227cfc47363a9897"
}
{
  "name": "0.7",
  "digest": "sha256:adcc8804fb8d299ed38317339467b2fc4792da7cdd5cae69227cfc47363a9897"
}
{
  "name": "0.7.0",
  "digest": "sha256:adcc8804fb8d299ed38317339467b2fc4792da7cdd5cae69227cfc47363a9897"
}

$ cat electric-tags.txt | jq '.results[]
                              | select(.digest == "sha256:adcc8804fb8d299ed38317339467b2fc4792da7cdd5cae69227cfc47363a9897")
                              | select(.name | test("^[^.]+\\\\.[^.]+$"))
                              | {name: .name, digest: .digest}'
{
  "name": "0.7",
  "digest": "sha256:adcc8804fb8d299ed38317339467b2fc4792da7cdd5cae69227cfc47363a9897"
}
```

I've only used `jq` here to demonstrate the idea. It should be implemented in TypeScript.

### DIFF
--- a/examples/starter/src/index.ts
+++ b/examples/starter/src/index.ts
@@ -17,7 +17,7 @@ const spinner = ora('Validating arguments').start()
 
 const error = (err: string) => {
   spinner.stop()
-  console.error('\x1b[31m', err + '\nnpx create-electric-app [<app-name>] [--electric-port <port>] [--electric-proxy-port <port>] [--webserver-port <port>]', '\x1b[0m')
+  console.error('\x1b[31m', err + '\nnpx create-electric-app [<app-name>] [--electric-image-tag <tag>] [--electric-port <port>] [--electric-proxy-port <port>] [--webserver-port <port>]', '\x1b[0m')
   process.exit(1)
 }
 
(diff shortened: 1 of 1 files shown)


kevin-dp (approved):
LGTM! We can implement the described feature of fetching the tag corresponding to latest in a different PR.

PR #626, opened 2023-11-03
chore(electric): Cleanups in code that deals with Postgres connections

alco:
Among a bunch of minor cleanups this change also removes the operator class for the `electric.tag` type. See the commit messsage for more info.

### DIFF
--- a/components/electric/lib/electric/replication/initial_sync.ex
+++ b/components/electric/lib/electric/replication/initial_sync.ex
@@ -102,7 +102,8 @@ defmodule Electric.Replication.InitialSync do
       origin = Connectors.origin(opts)
       {:ok, _, schema} = Extension.SchemaCache.load(origin)
 
-      :epgsql.with_transaction(
+      Client.with_transaction(
+        "ISOLATION LEVEL REPEATABLE READ READ ONLY",
         conn,
         fn conn ->
           # Do the magic write described in the function docs. It's important that this is
@@ -144,8 +145,7 @@ defmodule Electric.Replication.InitialSync do
             results ->
               send(parent, {:subscription_data, subscription_id, results})
           end
-        end,
-        begin_opts: "ISOLATION LEVEL REPEATABLE READ READ ONLY"
+        end
       )
     end)
   end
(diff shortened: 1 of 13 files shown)


components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers/electric_tag_type_and_operators.sql
```diff
@@ -146,13 +146,3 @@ CREATE OR REPLACE FUNCTION electric.tag_operator_cmp(first electric.tag, second
         WHEN first < second then -1
         ELSE 1
     END;
-
-
-CREATE OPERATOR CLASS btree__electric_tag_ops
```
magnetised, line 151: removed because it isn't used?

magnetised (approved):
nice. that change of default for `:replication` earns you an 🍦

components/electric/lib/electric/replication/initial_sync.ex
```diff
@@ -98,11 +98,12 @@ defmodule Electric.Replication.InitialSync do
         connection: opts,
         telemetry_span: span
       ) do
-    Client.with_conn(Connectors.get_connection_opts(opts, replication: false), fn conn ->
+    Client.with_conn(Connectors.get_connection_opts(opts), fn conn ->
       origin = Connectors.origin(opts)
       {:ok, _, schema} = Extension.SchemaCache.load(origin)
 
-      :epgsql.with_transaction(
+      Client.with_transaction(
```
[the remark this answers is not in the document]
alco: I think it is important to preserve the whole context at this call site. I.e. hiding the "read only" part just to make the function name palatable doesn't cut it for me. Plus, this would be the single place where that function would be used, it's simply not worth it.

components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers/electric_tag_type_and_operators.sql
alco: I hope it isn't! From my understanding, which I documented in the commit message, Postgres would raise an exception if a value of type `electric.tag` was used in a context that requires an operator class to exist.

PR #630, opened 2023-11-06
Proxy crashes when GSSAPI session encryption is requested

magnetised:


alco (approved):
Looks good :+1: 

Just noting that there's another bug waiting to be uncovered here. According to the same [Message Flow page](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-SSL),
> To issue a cancel request, the frontend opens a new connection to the server and sends a **CancelRequest message**, rather than the **StartupMessage message** that would ordinarily be sent across a new connection. The server will process this request and then close the connection. For security reasons, no direct reply is made to the cancel request message.

I'm not suggesting we would handle cancel requests in the proxy but we want to at least log a descriptive error message when a client sends it.

magnetised:
> Looks good 👍
> 
> Just noting that there's another bug waiting to be uncovered here. According to the same [Message Flow page](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-SSL),
> 
> > To issue a cancel request, the frontend opens a new connection to the server and sends a **CancelRequest message**, rather than the StartupMessage message that would ordinarily be sent across a new connection. The server will process this request and then close the connection. For security reasons, no direct reply is made to the cancel request message.
> 
> I'm not suggesting we would handle cancel requests in the proxy but we want to at least log a descriptive error message when a client sends it.


Yeah, handling that would be extremely complicated. But good catch, I'll handle it with some kind of grace

alco:
Beautiful!

PR #633, opened 2023-11-07
feat(DAL): client-side support for JSON type

kevin-dp:
This PR adds client-side support for the JSON type.
The DAL accepts JS values representing JSON and serialises them to a string that is stored in SQLite.
When reading a JSON value from the database the string is deserialised back into a JS value.
A corner case arises when having an optional column of type JSON because we need to differentiate between the database NULL value and the JSON null value. We treat the regular JS null value as a database NULL (to be consistent with how null values are interpreted for other column types) and require users to pass a special JsonNull object in order to store a top-level JSON null value.

Still need to add an E2E test for json values.


e2e/satellite_client/src/generated/client/prismaClient.d.ts
alco, line 1: This file wasn't part of our source tree before this PR. What is it used for?

alco: Oh, I see it's where the `InputJsonValue` type is defined. Can we perhaps have this type defined in `generated/client/index.ts`? Pulling in 10k lines that GitHub cannot even load just for this one line feels excessive to me.

kevin-dp: This file are the typings of the generated Prisma client. Previously it was written to `node_modules/@prisma/...`.
Our generated Electric client uses these typings so i decided to change the generator such that that file is located next to the Electric client. Our client uses more than only the `InputJsonValue` from this file. Almost all types in our Electric client come from this `prismaClient.d.ts` file.

alco: So we need to keep this file checked in until we figure out a way to run the actual generator for the e2e client instead of updating these files by hand?

kevin-dp: Indeed

generator/src/functions/tableDescriptionWriters/writeTableSchemas.ts
```diff
@@ -120,7 +124,7 @@ function pgType(field: ExtendedDMMFField, modelName: string): string {
       return 'DECIMAL'
     case 'Float':
       return 'FLOAT8'
-    case 'JSON':
+    case 'Json':
       return 'JSON'
```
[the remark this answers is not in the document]
kevin-dp: Yes, Prisma defaults to mapping its `Json` type to PG's `JSONB`. If `JSON` is needed than it requires the `@db.Json` annotation.
For the Electric client, it doesn't really matter though because in the JavaScript world there is no such difference as `Json` and `JsonB` as essentially both get parsed using `JSON.parse`. But it is indeed better to still provide the right json type here such that we can differentiate between them, would we need it later.

PR #634, opened 2023-11-07
fix (DAL): transform all data types in input structures

kevin-dp:
This PR fixes the bug reported by Kyle Mathews on Discord: https://discord.com/channels/933657521581858818/1079688869852753981/1169710877201674300

> Possible bug in 0.7. Upgraded and changed column type to boolean now my node.js liveMany query is failing with SQLite error that it can't bind to the type past in. In the better-sqlite adapter, if I log out queries and args passed to the query function then I see a "true" is being passed in which presumably should be 1. If I try to change the query back to use 1, I get a zod error (as I'd expect).

The problem came from the fact that we were only transforming date objects to SQLite strings but not the other values that were added later (booleans, floats, etc.).

### DIFF
--- a/clients/typescript/src/client/conversions/input.ts
+++ b/clients/typescript/src/client/conversions/input.ts
@@ -1,6 +1,6 @@
 import mapValues from 'lodash.mapvalues'
 import { FieldName, Fields } from '../model/schema'
-import { fromSqlite, toSqlite } from './sqlite'
+import { fromSqlite, toSqlite, isDataObject } from './sqlite'
 import { InvalidArgumentError } from '../validation/errors/invalidArgumentError'
 import { mapObject } from '../util/functions'
 import { PgType } from './types'
@@ -333,7 +333,7 @@ function isObject(v: any): boolean {
 function isFilterObject(value: any): boolean {
   // if it is an object it can only be a timestamp or a filter object
   // because those are the only objects we support in where clauses
-  return isObject(value) && !(value instanceof Date)
+  return isObject(value) && !isDataObject(value)
 }
 
 /**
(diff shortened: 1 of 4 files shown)


clients/typescript/src/client/conversions/input.ts
```diff
@@ -318,25 +318,22 @@ function transformFieldsAllowingFilters(
...
   }
+
+  return toSqlite(value, pgType)
+}
+
+function isObject(v: any): boolean {
+  return typeof v === 'object' && !Array.isArray(v) && v !== null
+}
+
+function isFilterObject(value: any): boolean {
```
alco, line 333: I wonder if this is the right place for this definition. If we add support for more object-typed values in the future, for example, an object to represent arbitrary-precision decimals or a JSON value or a custom interval object, this function would treat them as filter objects. So I'm thinking this check needs to be placed close to type conversions so that it's easy to spot when support for a new type is added.

alco (approved):
Looks good. I just have one nitpick.

clients/typescript/src/client/conversions/input.ts
kevin-dp: Yeah i was wondering the same. Reason for putting it here is that filter objects are specific to the input structures of the DAL which is the core of this file. But as you pointed out, the reason we had missed this is because it was in this file and too far from where we implement the conversions.

kevin-dp: I have refactored the implementation of `isFilterObject` to delegate part of the check to `isDataObject` which is defined in `sqlite.ts` (the file where conversion to and from sqlite are implemented). As such it is closer to the relevant piece of code.

alco: Brilliant!

PR #636, opened 2023-11-08
fix(electric): Introspection through proxy yields weird FKs

magnetised:


kevin-dp (approved):

components/electric/lib/electric/postgres/proxy/prisma/query.ex
```diff
@@ -1142,7 +1142,8 @@ defmodule Electric.Postgres.Proxy.Prisma.Query.ForeignKeyV4_8 do
           # index for ordering only and will be stripped
           i,
           # make up an oid that we're unlikely to hit in real life
-          i32(table.oid + 2_000_000),
+          # and make sure every fk oid is different
+          i32(table.oid + 2_000_000 + :erlang.phash2(fk.name, 1_000_000)),
```
alco, line 1146: I think something like this would be more robust, i.e. guarantee that the same OID is never generated twice for the lifetime of the running Erlang VM.
```elixir
def unique_oid do
  2_000_000 + System.unique_integer([:positive, :monotonic]))
end
```

`:erlang.phash2` does not provide that same guarantee.

e2e/tests/04.04_prisma_introspection_fk.lux
```diff
@@ -0,0 +1,73 @@
...
+          "createdAt" TIMESTAMP NOT NULL,
+          "updatedAt" TIMESTAMP NOT NULL,
+          "noteHtml" TEXT,
+          "linkValidatedEdgeId" TEXT,
+          "threadValidatedEdgeId" TEXT,
+          "threadTitle" TEXT,
+          "personContactId" TEXT,
+          "personGivenName" TEXT,
+          "personFamilyName" TEXT,
+          "personDisplayName" TEXT,
```
alco, line 25: I may be nitpicking but to me all these columns are noise and unnecessary congnitive load on the person reading the code if they don't have relevance for the thing being exercised in this test.

alco (approved):

e2e/tests/04.04_prisma_introspection_fk.lux
magnetised: they are noise. will remove

components/electric/lib/electric/postgres/proxy/prisma/query.ex
alco: got it

Issue #605, opened 2023-10-31
The protocol timeout is hardcoded


gorbak25:
https://github.com/electric-sql/electric/blob/3fdb28908b08141e3ee74b4cd1dcb283fd5df70d/clients/typescript/src/satellite/config.ts#L55

When electric requests a lot of data the default timeout which is hard-coded to 3000 is sometimes not enough and I see no way to increase it.
![image](https://github.com/electric-sql/electric/assets/1352848/44a29090-3c01-4f55-a991-715fbaca4f47)
[...]
![image](https://github.com/electric-sql/electric/assets/1352848/b186b338-d962-4b22-a14c-0e2a2f4a0217)

balegas:
We'll expose ```SatelliteClientOpts.timeout```

kevin-dp:
This feature has been implemented in https://github.com/electric-sql/electric/pull/637 which is now merged into main. It will be part of our next patch release.

PR #637, opened 2023-11-08
feat (client): configurable subscription timeout

kevin-dp:
This PR adds an option to electrify to configure the timeout for subscriptions as requested in https://github.com/electric-sql/electric/issues/605

### DIFF
--- a/clients/typescript/src/config/index.ts
+++ b/clients/typescript/src/config/index.ts
@@ -23,7 +23,9 @@ export interface ElectricConfig {
    */
   url?: string
   /**
-   * Timeout (in milliseconds) for RPC requests to fulfill shape subscriptions.
+   * Timeout (in milliseconds) for RPC requests.
+   * Needs to be large enough for the server to have time to deliver the full initial subscription data
+   * when the client subscribes to a shape for the first time.
    */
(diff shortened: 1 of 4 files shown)


clients/typescript/src/config/index.ts
```diff
@@ -19,9 +19,13 @@ export interface ElectricConfig {
...
    *
    * Defaults to:
-   * `http://127.0.0.1:5133`
+   * `http://localhost:5133`
    */
   url?: string
+  /**
+   * Timeout (in milliseconds) for RPC requests to fulfill shape subscriptions.
+   */
+  timeout?: number
```
alco, line 28: This option has a very generic name for something that's only used to fulfill shape subscriptions.
Perhaps update the docstring to say something to the tune of
> Timeout (in milliseconds) for RPC requests.
>
> Needs to be large enough for the server to have time to deliver the full initial subscription data when the client subscribes to a shape for the first time.

alco (approved):

clients/typescript/src/config/index.ts
kevin-dp: Good point, will do!

PR #638, opened 2023-11-08
iff => if

KyleAMathews:
Noticed a typo and then found more of them!

### DIFF
--- a/clients/typescript/src/notifiers/index.ts
+++ b/clients/typescript/src/notifiers/index.ts
@@ -99,8 +99,8 @@ export interface Notifier {
   actuallyChanged(dbName: DbName, changes: Change[]): void
 
   // Reactive hooks then subscribe to "data has actually changed" notifications,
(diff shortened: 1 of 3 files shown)


kevin-dp:
Hey @KyleAMathews, i don't think these are typos, but we really meant iff (if and only if).
Perhaps we have overdone it a bit and used it in too many places (i.e. in places where `if` would do equally well).

KyleAMathews:
Ah .. ok, actually have never heard of that but I'll close.

kevin-dp:
Hi @KyleAMathews, you are not the only one that doesn't know "iff".
It stems from my academic background where iff means <=> (logical equivalence) as compared to if which means => (logical implication). Let's indeed replace them by "if".

thruflo:
This is me being a smart arse and using iff. I’d suggest let’s either simplify to “if” or expand to “if and only if”.

PR #639, opened 2023-11-08
chore(electric,client): Create new protocol op to represent a compensation

magnetised:
For permissions etc we need to always have the `OLD` values available, so for clarity don't send out SatOpUpdate with no old_data values

### DIFF
--- a/.changeset/itchy-carrots-invite.md
+++ b/.changeset/itchy-carrots-invite.md
@@ -1,0 +1,6 @@
+---
+"@core/electric": patch
+"electric-sql": patch
+---
+
+[VAX-1335] Create new protocol op to represent a compensation
--- a/.changeset/itchy-carrots-invite.md
+++ b/.changeset/itchy-carrots-invite.md
@@ -1,6 +1,6 @@
 ---
-"@core/electric": patch
-"electric-sql": patch
+"@core/electric": minor
+"electric-sql": minor
 ---
 
 [VAX-1335] Create new protocol op to represent a compensation
--- a/clients/typescript/src/migrators/triggers.ts
+++ b/clients/typescript/src/migrators/triggers.ts
@@ -118,10 +118,9 @@ export function generateOplogTriggers(
 /**
  * Generates triggers for compensations for all foreign keys in the provided table.
  *
- * Compensation is recorded as a specially-formatted update. It acts as a no-op, with
- * previous value set to NULL, and it's on the server to figure out that this is a no-op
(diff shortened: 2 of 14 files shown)


clients/typescript/src/migrators/triggers.ts
```diff
@@ -167,7 +166,7 @@ function generateCompensationTriggers(table: Table): Statement[] {
               1 == (SELECT value from _electric_meta WHERE key == 'compensations')
       BEGIN
         INSERT INTO _electric_oplog (namespace, tablename, optype, primaryKey, newRow, oldRow, timestamp)
-        SELECT '${fkTableNamespace}', '${fkTableName}', 'UPDATE', json_object(${joinedFkPKs}), json_object(${joinedFkPKs}), NULL, NULL
+        SELECT '${fkTableNamespace}', '${fkTableName}', 'COMPENSATION', json_object(${joinedFkPKs}), json_object(${joinedFkPKs}), NULL, NULL
```
icehaunter, line 169: do we need to handle client updates (i.e. reapply these triggers)?

icehaunter (approved):
LGTM, couple tiny questions

.changeset/itchy-carrots-invite.md
```diff
@@ -0,0 +1,6 @@
+---
+"@core/electric": patch
```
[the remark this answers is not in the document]
magnetised: wasn't sure how we handled the version change now there's no version in the handshake or the protocol

clients/typescript/src/migrators/triggers.ts
magnetised: yes we would, good point. off the top of your head, is there a mechanism in place to trigger (😉 ) the re-creation of the triggers?

.changeset/itchy-carrots-invite.md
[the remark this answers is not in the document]
icehaunter: There is, it's just a bit hidden. It's sent in HTTP headers in WebSocket upgrade request.

clients/typescript/src/migrators/triggers.ts
magnetised: @icehaunter looking at the client it seems that adding re-application of these triggers on boot is a fairly major change. are we comfortable with postponing that feature for the moment considering that this change is backwards compatible with existing trigger implementations?

magnetised: https://linear.app/electric-sql/issue/VAX-1385/internal-schema-migration-for-client-side-db-schema

PR #640, opened 2023-11-09
chore(docs): Update deployment instructions for Fly.io

alco:


docs/integrations/deployment/fly.md
```diff
@@ -5,33 +5,241 @@ description: >-
...
+```
+
+### Generate a type-safe Data Access Layer
+
+Now that the database has one electrified table, we can generate a type-safe client from it. Use the same database connection URL as in the previous step but change the username to `prisma` (this is requried for the schema introspection to work correctly).
+
+```shell
+$ npx electric-sql generate
+      --service http://electric-on-fly-test-app.internal:5133
+      --proxy postgresql://prisma:******@electric-on-fly-test-app.internal:65432/electric
```
thruflo, line 228: Noting that this url ends with `/electric`. Whereas the previous one ends with `/postgres`.

thruflo (approved):
Great stuff, I pushed a few minor copy tweaks.

docs/integrations/deployment/fly.md
alco: Good catch! :eagle: :eye:

PR #641, opened 2023-11-09
fix (client): quote namespaces, table names, and column names in triggers

kevin-dp:
This PR modifies the trigger generation code to quote namespaces, table names, and column names.
This is needed because a namespace, table, or column could be named after a reserved keyword which leads to errors.

### DIFF
--- a/.changeset/witty-bears-shake.md
+++ b/.changeset/witty-bears-shake.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Fix bug where SQL complains about syntax error near "from" due to table names and column names not being quoted in the generated triggers.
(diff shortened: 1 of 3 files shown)


alco:
@kevin-dp we need proper quoting that escapes any internal quotes to really fix this issue and avoid opening up a loophole for SQL injections. I can easily break this quoting by modifying the personTable in test/satellite/common.ts to have
```
  columns: ['id"', '"name"', 'age\'', 'bmi\\'],
```

those are all valid column names in Postgres:
```
[localhost] postgres:electric=# create table foo("id""" text primary key, """name""" text, "age'" int, "bmi\\" real);
CREATE TABLE
[localhost] postgres:electric=# \d foo
                Table "public.foo"
 Column │  Type   │ Collation │ Nullable │ Default 
────────┼─────────┼───────────┼──────────┼─────────
 id"    │ text    │           │ not null │ 
 "name" │ text    │           │          │ 
 age'   │ integer │           │          │ 
 bmi\\  │ real    │           │          │ 
```

alco (approved):

PR #642, opened 2023-11-09
Change the client build to use "tsc", use contemporary "nodenext" settings, enable source maps

samwillis:
This changes the client build to use vanilla "tsc" to build, it also changes the tsconf to:
```
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
```
This requires changing all relative imports to explicitly have the `.js` extension and directory imports to have `/index.js` (there is no sensible way around this and it makes it a "proper TS+ESM package"). It's `.js` rather than `.ts` extensions as TypeScript decided that's how its supposed to be... 

One key advantage of doing this was that "tsmodule" did not have a way to create source maps, I have now turned source maps on. This will hopefully help with users debugging when they find errors in electric.

I used a modified version of this to fix all the imports: https://gist.github.com/silassare/cc3db860edeea07298cb03f64738c997

My version that handles our imports without a line ending ';': 

```js
import * as fs from 'fs'
import * as path from 'path'

// https://gist.github.com/lovasoa/8691344
async function* walk(dir) {
  for await (const d of await fs.promises.opendir(dir)) {
    const entry = path.join(dir, d.name)
    if (d.isDirectory()) {
      yield* walk(entry)
    } else if (d.isFile()) {
      yield entry
    }
  }
}

function resolveImportPath(sourceFile, importPath, options) {
  const sourceFileAbs = path.resolve(process.cwd(), sourceFile)
  const root = path.dirname(sourceFileAbs)
  const { moduleFilter = defaultModuleFilter } = options

  if (moduleFilter(importPath)) {
    const importPathAbs = path.resolve(root, importPath)
    let possiblePath = [
      path.resolve(importPathAbs, './index.ts'),
      path.resolve(importPathAbs, './index.js'),
      importPathAbs + '.ts',
      importPathAbs + '.js',
    ]

    if (possiblePath.length) {
      for (let i = 0; i < possiblePath.length; i++) {
        let entry = possiblePath[i]
        if (fs.existsSync(entry)) {
          const resolved = path.relative(root, entry.replace(/\.ts$/, '.js'))

          if (!resolved.startsWith('.')) {
            return './' + resolved
          }

          return resolved
        }
      }
    }
  }

  return null
}

function replace(filePath, outFilePath, options) {
  const code = fs.readFileSync(filePath).toString()
  const newCode = code.replace(
    /(import|export) (.+?) from ('[^\n']+'|"[^\n"]+")(;|\n)/gs,
    function (found, action, imported, from, end) {
      const importPath = from.slice(1, -1)
      const resolvedPath = resolveImportPath(filePath, importPath, options)

      if (resolvedPath) {
        console.log('\t', importPath, resolvedPath)
        return `${action} ${imported} from '${resolvedPath}'${end}`
      }

      return found
    }
  )

  if (code !== newCode) {
    fs.writeFileSync(outFilePath, newCode)
  }
}

// Then, use it with a simple async for loop
async function run(srcDir, options = defaultOptions) {
  const { sourceFileFilter = defaultSourceFileFilter } = options

  for await (const entry of walk(srcDir)) {
    if (sourceFileFilter(entry)) {
      console.log(entry)
      replace(entry, entry, options)
    }
  }
}

const defaultSourceFileFilter = function (sourceFilePath) {
  return (
    /\.(js|ts)$/.test(sourceFilePath) && !/node_modules/.test(sourceFilePath)
  )
}

const defaultModuleFilter = function (importedModule) {
  return (
    !path.isAbsolute(importedModule) &&
    !importedModule.startsWith('@') &&
    !importedModule.endsWith('.js')
  )
}

const defaultOptions = {
  sourceFileFilter: defaultSourceFileFilter,
  moduleFilter: defaultModuleFilter,
}

// Switch this to test on one file or directly run on a directory.
const DEBUG = false

if (DEBUG) {
  console.log('DEBUG MODE')
  replace(
    '/Users/samwillis/Code/electric/clients/typescript/src/index.ts',
    '/Users/samwillis/Code/electric/clients/typescript/src/index-TEST.ts',
    defaultOptions
  )
} else {
  await run(
    '/Users/samwillis/Code/electric/clients/typescript/',
    defaultOptions
  )
}

```


samwillis:
Replaced by #659

PR #643, opened 2023-11-09
fix (generator): capitalise model names

kevin-dp:
This PR modifies the generator to capitalise the first letter of model names in the generated Prisma schema.
We no longer turn snake_cased model names into PascalCase because this may lead to additional name clashes.
We also let Prisma turn PG table names such as `_foo` into valid model names like `foo`.

### DIFF
--- a/.changeset/quiet-otters-film.md
+++ b/.changeset/quiet-otters-film.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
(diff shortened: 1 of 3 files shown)


icehaunter (approved):
LGTM, although that's another data point pointing to maybe dropping Prisma is the right approach at some point in the future. Dancing around a text-based intermediate representation of a `.prisma` file just to generate some types feels unsustainable long term.

PR #644, opened 2023-11-09
chore (client): fix assertions in unit tests

kevin-dp:
This PR fixes some assertions that were wrongly being used.


samwillis (approved):
All good 👍

PR #645, opened 2023-11-09
chore(client): Move Prisma and our generator to be dependencies of the electric client.

samwillis:
- Prisma is pinned to a known working version.
- Rather than `npx prisma` we resolve the prisma cli script that is a direct dependancy of our client package. This should enable projects using the electric client to install a different version of prisma to the one we use.

### To test:

Build the client:
```sh
cd client/typescript
pnpm install
pnpm build && pnpm pack
```
Update and test example:
```sh
cd ../../examples/[an example]
rm -rf node_modules/
rm package-lock.json  # If there is one
npm uninstall @prisma/client @electric-sql/prisma-generator prisma electric-sql
npm install ../../clients/typescript/electric-sql-0.8.0.tgz
```

Ensure ./backend/compose/.envrc is using the 0.8 tagged Electric Docker image:
```sh
# ./backend/compose/.envrc
export ELECTRIC_IMAGE=electricsql/electric:0.8
```

Then start the example as usual.

---

To do:

- [ ] Test examples - prisma and generator dependencies need removing:
  - [x] Linearlite **working**
  - [ ] expo
  - [x] web-wa-sqlite
  - [x] beer-starts
  - [x] inonic-demo
  - [ ] react-native
- [x] Test that a project can have a conflicting version of Prisma installed

After this is released the examples and starter will need updating to remove the Prisma and generator dependancies.


clients/typescript/Makefile
```diff
@@ -1,3 +1,6 @@
+build_generator:
+	(cd ../../generator && make deps && make build)
```
alco, line 2: This can be rewritten using the following more concise style
```suggestion
	$(MAKE) -C build
```
Note that `build` already depends on `deps` in generator's Makefile.

alco:
@samwillis Nice work, Sam!

I tried following your testing instructions for the beer-stars example. Everything installs and builds without issues. But when I start the app and open it in my web browser, I don't see any UI and the browser console shows this error:
Screenshot from 2023-11-14 16-42-40

samwillis:
> @samwillis Nice work, Sam!
> 
> I tried following your testing instructions for the beer-stars example. Everything installs and builds without issues. But when I start the app and open it in my web browser, I don't see any UI and the browser console shows this error: !

I can't reproduce this error, it works for me on the latest version.

samwillis:
I have tested installing an incompatible Prisma@5.6 version in the web-wa-sqlite example. As you can see in the screen shots below, the project has v5.6 installed, but the electric client has v4.12 installed.

Running the generator worked correctly, and the app works, despite an incompatible Prisma in the project.

image

image

kevin-dp:
> I looked at the changes, seems all good!
> 
> My only comment is a bit more philosophical and is about the generator and prisma dependencies. By introducing them as dependencies, we require all Electric applications to have those dependencies. But those are not needed at runtime. They really are _dev_ dependencies of the app that is using Electric, because the generator and prisma client are only needed in dev mode for generating the Electric client. But yes then people may easily forget to add them as dev dependencies of their project or use an incompatible version... Just wanted to raise this, don't want to block this but something to keep in mind to be sure this is the thing we want to do.
> 
> PS: i've not been able to test this yet. Hope to be able to test it today

This problem probably comes from the fact that our `electric-sql` package in fact contains 2 independent packages... 1 for the CLI tools (generate command) and 1 which is the actual ts-client runtime. If the CLI tools would be in a separate package (e.g. `electric-cli`) then we could move those dependencies to be dependencies of `electric-cli`.

kevin-dp:
> > They really are _dev_ dependencies of the app that is using Electric
> 
> Yes, so this was a conversation I had with @thruflo a while back. Users will still use a bundler when building their apps so it won't be included in the built app, they are only installed if they distribute their apps via NPM which is unlikely.
> 
> We have the same question about the CLI work too, and if that should be a separate package (and then these Prisma and generator dependancies become dependancies of that).
> 
> Ultimately for now it was decided to keep it all in one package, we can always break the CLI out as a separate package later.

Ok good for me! Eventually we will need to split to CLI tooling into a separate package.

alco:
> > @samwillis Nice work, Sam!
> > I tried following your testing instructions for the beer-stars example. Everything installs and builds without issues. But when I start the app and open it in my web browser, I don't see any UI and the browser console shows this error: !
> 
> I can't reproduce this error, it works for me on the latest version.

I can't reproduce that issue either anymore.

Seems to work! Only one observation: when I add `@prisma/client: "^5.6"` or `prisma: "^5.12"` to `dependencies` or `devDependencies` in `beer-stars/package.json` and then run `yarn`, I see those versions in the generated `yarn.lock`. However, I don't see the versions that `electric-sql` pins. Like yourself, I too see the pinned versions inside `node_modules/electric-sql/node_modules` but I was expecting them to appear in the top-level lock file as well. I'm probably misunderstanding something about how `yarn.lock` works.

samwillis:
Related to this: https://github.com/electric-sql/electric/issues/741, this needs testing in monorepos and with yarn a pnpm.

samwillis:
Tested and confirmed this will fix #741

PR #648, opened 2023-11-10
client: Log SQLite version when starting

davidmartos96:
Log the SQLite version that is running on the client.
I feel it could end up being useful, given how there could be features/bugs on different versions of SQLite that run on the multiple supported adapters.

### DIFF
--- a/clients/typescript/src/satellite/config.ts
+++ b/clients/typescript/src/satellite/config.ts
@@ -22,6 +22,8 @@ export interface SatelliteOpts {
   clearOnBehindWindow: boolean
   /** Backoff options for connecting with Electric*/
   connectionBackOffOptions: ConnectionBackoffOptions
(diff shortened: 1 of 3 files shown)


kevin-dp:
Yes this is very useful. I would argue to do this only in debug mode though. Would you mind adding the debug flag to the satellite options and only querying and printing the SQLite version if the debug flag is set?

The debug flag can be added to the `satelliteOpts` object on L220 in `clients/typescript/src/satellite/registry.ts`.
```ts
const satelliteOpts: SatelliteOpts = {
  ...satelliteDefaults,
  connectionBackOffOptions: config.connectionBackOffOptions,
  debug: config.debug,
}
```

davidmartos96:
@kevin-dp Yes, god point. Done.

kevin-dp (approved):

PR #653, opened 2023-11-12
feat(electric): Add support for the int8 column type

alco:


e2e/satellite_client/src/prisma/schema.prisma
```diff
@@ -57,9 +57,10 @@ model Ints {
   id String @id
   i2 Int? @db.SmallInt /// @zod.number.int().gte(-32768).lte(32767)
   i4 Int?              /// @zod.number.int().gte(-2147483648).lte(2147483647)
+  i8 Int?              /// @zod.number.int().gte(-9223372036854775808).lte(9223372036854775807)
```
alco, line 60: @kevin-dp It is odd that we have to specify validations by hand in this Prisma schema. If we change the regular validation code but forget to update this Prisma schema, the tests will no longer be verifying the actual implementation.

kevin-dp: it is odd indeed.. A proper solution would consist of spinning up a postgres DB that has the complete schema, and then run the generate script on that. I remember looking into this to see if this was something i could setup quickly but i was not sure how this can be done in the CI. I would be willing to look into this but it may take some time as i need to figure out which dependencies are missing from the CI scripts etc.

alco: I don't think we need to run the generator on CI. We just need to replace the prisma schema with an SQL one and write a script that will generate a client for e2e tests from that using `npx electric-sql generate`. We can then commit the generated client to Git just like we already do.

But let's not worry about it now. I'll open a separate issue for this.

kevin-dp: Yes, that's a possibility indeed

magnetised (approved):
looks great. good to get those big-ol ints back.

PR #659, opened 2023-11-13
chore(client): Change to tsup for building client

samwillis:
As discussed with @icehaunter this is an alternative to #642 but using ESBuild backed [tsup](https://tsup.egoist.dev).

This uses:
- `tsup` to compile the typescript to js
- `tsc` to generate type declarations
- `./scripts/fix-imports.js` to rewrite all imports in the build to point to full paths with extensions, `tsc` is not able to do this itself. If we used `tsup` to emit the declarations it takes 34 seconds. This setup builds in 6 seconds (`tsc` is the bottleneck). This script is based not he one I used in #642
- `tsx` is used as a loader for the test runner.

The import path rewrite step is required to ensure this is a fully ESM compliant build and can be used in projects that have a `tsconfig.json` `moduleResolution: "nodenext"`. The alternative way to do this is #642.

With this build we also now get source maps in the distribution package, this will help with debugging.

I've also upgraded Typescript from 4.9 to 5.3, this required one change due to `tsc` no longer being able to infer a type in a nested reduce.

---

~~TBH, I'm not sure which is better at this point:~~

- ~~`tsc` takes about 7.2 seconds, and we have to change all the imports to the ugly syntax.~~
- ~~`tsup` takes 60ms to emit the js, but totals **34 seconds** to output the type defs! I'm not sure why it's so incredible slow as it uses tsc to do the typedefs. But we don't have to change any code.*Edit: plus the 7 sec for types*~~

~~ESBuild is awesomely fast, but as we want the type defs anyway I'm unsure that this one is any better.~~


samwillis:
I have changed this to use `tsc` for type defs, rather than `tsup`'s built in system for using `tsc`, so it now builds in 7 seconds.

alco:
@samwillis does the PR title need an update to drop the mention of tsup?

samwillis:
@alco TSUP is still used for building/bundling, it just uses tsc directly.

e2e test are currently failing on this PR due to how were were previously building the client library. Esmodule was doing a fairly simple file by file conversion of the TS to JS, we are then importing some of these files into e2e tests directly even though they are not listed as exports in the package.json.

Usually with libraries built for use in the browser you would do bundling, this could be into a single file or multiple if you have multiple entry points (listed as exports in package.json). That is what tsup is doing (and any other *fast* ts build tools), it bundles the code in to chunks and tree shakes to remove any unused code.

The question I have, what is the purpose of our build:
1. to just translate the ts to js
2. or to bundle/minify for browser distribution

If its only 1 then we should probably just do what I had done in #642 and use tsc, there are no speed improvements to gain from using esbuild as we have to use tsc anyway to build the types.

If we want to do a bundling build, we need to fix the exports so that they all make sense and work for e2e testing.

Currently as we have a generated client all of our uses will be using their own bundling build tool themselves, however if we have an option in future to not have a built client then being able to load the library as a bundle from a CDN would be good.

I started this PR as I just wanted source maps in our built library to help with debugging (both for us and our users), but it's blown up into a more complex question...

(tagging @icehaunter as you may also have thoughts)

kevin-dp (commented):
I don't have strong opinions on this one (except that i do not like having to import JS files) and i don't have much experience with JS/TS build tooling so i will let @icehaunter and @samwillis decide on this one.

clients/typescript/scripts/fix-imports.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  return null
+}
+
+function replace(filePath, outFilePath, options) {
+  const code = fs.readFileSync(filePath).toString()
+  let logging = false
+  const newCode = code.replace(
+    /(import|export) (.+?) from ('[^\n']+'|"[^\n"]+")(;|\n)/gs,
```
icehaunter, line 54: You probably want a line start & line end anchors here

samwillis:
@icehaunter 

> I don't love the complexity of the final setup

Neither do I, I really wanted to find one tool that would do what we wanted, but it doesn't seem to exist (I had hoped tsup would, and I can but takes 36 seconds to build).

Fundamentally, `tsc` is required somewhere to emit type declarations and it does not rewrite imports at all. So we either have to rewrite imports in our codebase, or do it during build.

I went looking how other typescript library projects do:
- Use `moduleResolution: "node/bundler"` in `tsconfig.json` for extension-less and dir imports in their code base
- Use a "fast" build tool
- Emit type declarations for users
- Have full path imports in the built codebase so that it's a "proper" esm package, and will be supported under all configurations.

A good example I could find was Drizzle, they are using `tsup` + `tsc` with a post import build rewrite - although they are using [recast](https://github.com/benjamn/recast) in their [rewrite script](https://github.com/drizzle-team/drizzle-orm/blob/main/drizzle-orm/scripts/fix-imports.ts).

The "simple" setup would be vanilla `tsc` with `moduleResolution: "nodenext"` and having the full path import with '.js' extensions in our codebase. It's no slower, as *we have to use `tsc` anyway*.

alco:
Can we use [Bun](https://bun.sh/docs/runtime/typescript)? It's not 100% node-compatible yet but can be used for building and [being fast](https://twitter.com/bunjavascript/status/1734588929112740150) is its primary selling point.

samwillis:
> Can we use [Bun](https://bun.sh/docs/runtime/typescript)?

I want to try it.

> [being fast](https://twitter.com/bunjavascript/status/1734588929112740150) is its primary selling point.

Yes, however to emit our type declarations we will still have to run `tsc` 😒 , bun just ignores the type annotations.

PR #661, opened 2023-11-13
feat(electric): Add support for the JSONB column type

alco:


components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -512,6 +512,11 @@ defmodule Electric.Satellite.Serialization do
     val
   end
 
+  def decode_column_value!(val, type) when type in [:json, :jsonb] do
+    _ = Jason.decode!(val)
```
magnetised, line 516: shame to have to do a full decode with all those allocations just to test the syntax but for now it makes sense

magnetised (approved):
cool. weirdly much simpler than the int8 support...

components/electric/lib/electric/satellite/serialization.ex
alco: Oh yes, this is tragic. Ideally we'd use a native library that only validates JSON, without any allocations. It's surprisingly difficult to find one because googling for it yields results about validating json schema.

Alternatively, and I don't know how feasible it is, but maybe we could pull the JSON validation code (directly from Postgres)[https://github.com/postgres/postgres/blob/50c67c2019ab9ade8aa8768bfe604cd802fe8591/src/backend/utils/adt/json.c#L1625] and wrap it into a NIF. The maintanence burden of that would be low, especially when taking into account that we currently have uncovered edge cases where `Jason.decode!()` would treat a given JSON as valid while Postgres would reject it.

barbalex:
Did you see that sqlite will soon support jsonb? https://sqlite.org/draft/jsonb.html

PR #664, opened 2023-11-15
feature(client): New client CLI commands

samwillis:
New CLI commands built into the client.

To run the CLI in a project that has electric-sql installed:
```
npx electric-sql
```
This will output the possible commands.

Run `npx electric-sql help [command-name]` for details of how to run a command.

All config args can be loaded as env vars, including from a .env at the root of the project.

Tasks:

- [x] Generalised env var and cli option loading/passing
- [x] Convert `generate` command to use option loading and `commander` package
- [x] "New" Commands:
  - [x] `start` (with additional `--with-postgres` option)
  - [x] `stop`
  - [x] `status`
  - [x] `psql`
  - ~~`configure-ports`~~ (Scrapped)
  - [x] `show-config`
  - [x] `generate --with-migrations`
  - [x] `with-config`
  - [x] `proxy-tunnel`
- [x] Env vars for Electric and Postgres images

Each command is also available as an export of the electric-sql/cli module and can be used in users scripts.

Easiest way to test is in conjunction with #669


davidmartos96:
@samwillis Hello!
When trying out this with #669 I had to "manually" copy the embedded docker files into the dist folder after the client build. I'm not sure if you are aware of this limitation.

At the end of the npm command `build`
`&& cp -r src/cli/docker-commands/docker ./dist/cli/docker-commands/`

samwillis:
@davidmartos96 Oops, thanks for testing and spotting that. I had lost a change in a rebase somewhere... it now copies the docker config into dist again.

clients/typescript/src/cli/config-options.ts
```diff
@@ -0,0 +1,280 @@
...
+    constructedDefault:
+      'postgresql://{ELECTRIC_DATABASE_USER}:{ELECTRIC_DATABASE_PASSWORD}@{ELECTRIC_DATABASE_HOST}:{ELECTRIC_DATABASE_PORT}/{ELECTRIC_DATABASE_NAME}',
+    groups: ['database', 'electric'],
+  },
+  DATABASE_HOST: {
+    doc: dedent`
+      Hostname of the database server.
+    `,
+    valueType: String,
+    defaultVal: () => defaultDbUrlPart('host', 'localhost'),
```
kevin-dp, line 101: Sure this needs to be done lazily? At first glance, it looks like this does not need to be a function. Same for the occurences below.

clients/typescript/package.json
```diff
@@ -121,8 +121,9 @@
   },
   "scripts": {
     "post-version-bump": "node -p \"'export const LIB_VERSION = \\'' + require('./package.json').version + '\\''\" > src/version/index.ts",
-    "dev": "rm -rf dist && mkdir dist && concurrently \"tsup --watch\" \"tsc -p tsconfig.build.json --watch\" \"nodemon --watch ./dist  --delay 250ms ./scripts/fix-imports.js\"",
-    "build": "rm -rf dist && concurrently \"tsup\" \"tsc -p tsconfig.build.json\" && node scripts/fix-imports.js",
+    "dev": "rm -rf dist && npm run build:copy-docker && mkdir dist && concurrently \"tsup --watch\" \"tsc -p tsconfig.build.json --watch\" \"nodemon --watch ./dist  --delay 250ms ./scripts/fix-imports.js\"",
+    "build": "rm -rf dist && npm run build:copy-docker && concurrently \"tsup\" \"tsc -p tsconfig.build.json\" && node scripts/fix-imports.js",
```
kevin-dp, line 125: these commands won't work on Windows

clients/typescript/src/cli/configure/command-configure-ports.ts
```diff
@@ -0,0 +1,88 @@
...
+import { getConfig } from '../config'
+import fs from 'fs'
+import portUsed from 'tcp-port-used'
+import prompts from 'prompts'
+
+export function makeConfigurePortsCommand() {
+  const command = new Command('configure-ports')
+  command
+    .description('Configure the ports used by the ElectricSQL sync service')
+
```
kevin-dp, line 11: blank line not needed

clients/typescript/src/cli/configure/command-configure-ports.ts
```diff
@@ -0,0 +1,88 @@
...
+    .action(async () => {
```
kevin-dp, line 12: I believe we can do `.action(configurePorts)`

kevin-dp: I also feel that this is easier to read:
```ts
export function makeConfigurePortsCommand() {
  return new Command('configure-ports')
    .description('Configure the ports used by the ElectricSQL sync service')
    .action(configurePorts)
}
```

clients/typescript/src/cli/configure/command-configure-ports.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  if (port < 0 || port > 65535) {
+    return 'Port must be between 0 and 65535'
+  } else if (await portUsed.check(port)) {
+    return 'Port is already in use'
+  }
+  return true
+}
+
+function writeEnvLocal(port: number, proxyPort: number) {
```
kevin-dp, line 62: Is this file always written by our scripts? Or could it be that the user modifies the contents of this file?
We could make it a bit more robust by trimming the lines before checking `startsWith`.

clients/typescript/src/cli/configure/command-configure-ports.ts
```diff
@@ -0,0 +1,88 @@
...
+    : []
+  const serviceUrl = `http://localhost:${port}`
+  const newLines = lines.map((line) => {
+    if (line.startsWith('ELECTRIC_HTTP_PORT=')) {
+      return `ELECTRIC_HTTP_PORT=${port}`
+    } else if (line.startsWith('ELECTRIC_PG_PROXY_PORT=')) {
+      return `ELECTRIC_PG_PROXY_PORT=${proxyPort}`
+    } else if (line.startsWith('ELECTRIC_SERVICE=')) {
+      return `ELECTRIC_SERVICE=${serviceUrl}`
+    } else {
```
kevin-dp, line 74: `else` is not needed since all other branches return.
```ts
} else if (line.startsWith('ELECTRIC_SERVICE=')) {
  return `ELECTRIC_SERVICE=${serviceUrl}`
}
return line
```

clients/typescript/src/cli/configure/command-configure-ports.ts
```diff
@@ -0,0 +1,88 @@
...
+      return line
+    }
+  })
+  if (!newLines.find((line) => line.startsWith('ELECTRIC_HTTP_PORT='))) {
```
kevin-dp, line 78: There's a bit of code duplication over the following 3 if statements.
We could have an array of the strings we're looking for and map over the array to search for each one.
Something like:

```ts
const envOptionsMapping = [ ['ELECTRIC_HTTP_PORT', port], ['ELECTRIC_PG_PROXY_PORT', proxyPort], ['ELECTRIC_SERVICE', serviceUrl] ]
envOptionsMapping.forEach(envOption => {
  const [name, value] = envOption
  if (!newLines.find((line) => line.startsWith(`${name}=`))) {
    newLines.push(`${name}=${value}`)
  }
})
```

clients/typescript/src/cli/configure/command-show-config.ts
```diff
@@ -0,0 +1,19 @@
+import { Command } from 'commander'
+import { getConfig } from '../config'
+
+export function makeShowConfigCommand() {
+  const command = new Command('show-config')
```
kevin-dp, line 5: could again just write it all in one return statement

clients/typescript/src/cli/configure/command-with-config.ts
```diff
@@ -0,0 +1,47 @@
+import { Command } from 'commander'
+import { spawnSync } from 'child_process'
+import { getConfig, envFromConfig, ConfigOptionName, Config } from '../config'
+
+export function makeWithConfigCommand() {
+  const command = new Command('with-config')
```
kevin-dp, line 6: single return statement

clients/typescript/src/cli/configure/command-with-config.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function withConfig(command: string, config?: Config) {
+  const _config = config || getConfig()
+  const env = process.env
+  const re = /\{{([A-Z_]+)}}/g
+  const cmd = command
+    .replace(re, (match, envVar) => {
+      const value = envVar.startsWith('ELECTRIC_')
+        ? _config[envVar.slice('ELECTRIC_'.length) as ConfigOptionName] ||
```
kevin-dp, line 26: Do you really mean `||` or would `??` be more appropriate?

clients/typescript/src/cli/docker-commands/command-status.ts
```diff
@@ -0,0 +1,20 @@
+import { Command } from 'commander'
+import { dockerCompose } from './docker-utils'
+
+export function makeStatusCommand() {
+  const command = new Command('status')
```
kevin-dp, line 5: single return statement

clients/typescript/src/cli/docker-commands/command-stop.ts
```diff
@@ -0,0 +1,41 @@
+import { Command } from 'commander'
+import { dockerCompose } from './docker-utils'
+
+type StopCommandArgs = {
+  remove?: boolean
+}
+
+export function makeStopCommand() {
+  const command = new Command('stop')
```
kevin-dp, line 9: single return

clients/typescript/src/cli/migrations/migrate.ts
```diff
@@ -246,29 +286,34 @@ function escapePathForString(inputPath: string): string {
...
+    user: 'prisma', // We use the "prisma" user to put the proxy into introspection mode
+    password: config.PG_PROXY_PASSWORD,
+    host: config.SERVICE_HOST,
+    port: config.PG_PROXY_PORT,
+    dbName: config.DATABASE_NAME,
+  })
   const schema = dedent`
     generator electric {
-      provider      = "node ${escapePathForString(generatorPath)}"
+      provider      = "${escapePathForString(generatorPath)}"
```
kevin-dp, line 304: sure `node` is not needed? iirc it was added because client generation would fail on Windows without it.

clients/typescript/src/cli/migrations/migrate.ts
```diff
@@ -246,29 +286,34 @@ function escapePathForString(inputPath: string): string {
...
       output        = "${escapePathForString(output)}"
       relationModel = "false"
     }
 
     generator client {
       provider = "prisma-client-js"
-      output   = "${escapePathForString(output)}"
+      output   = "${output}"
```
kevin-dp, line 311: why is the output path not escaped? Escaping is needed for Windows.

kevin-dp (dismissed):
Tremendous job you did here! 🥇 
I left some comments, nothing major.
Still need to test the CLI, what's the best way to do this? `npm link` ?

clients/typescript/src/cli/docker-commands/command-start.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (!config.WITH_POSTGRES && !config.DATABASE_URL) {
+      console.error(
+        'You must set --database-url or the ELECTRIC_DATABASE_URL env var when not using --with-postgres.'
+      )
+      process.exit(1)
+    }
+    const startOptions = {
+      detach: opts.detach,
+      withPostgres: !!config.WITH_POSTGRES,
+      pgPort: opts.WITH_POSTGRES ? opts.DATABASE_PORT : undefined,
```
davidmartos96, line 34: The `pgPort` parameters doesn't seem to be used, unless I'm missing something.

Also, does `opts.WITH_POSTGRES` resolve correctly from the `commander` CLI arguments? Or was that supposed to be `config.WITH_POSTGRES`? Same for `DATABASE_PORT`

kevin-dp:
Tested this via the web-wa-sqlite example, works like a charm!

clients/typescript/src/cli/main.ts
```diff
@@ -0,0 +1,42 @@
...
+async function main() {
+  const program = new Command()
+
+  program
+    .name('npx electric-sql')
+    .description('CLI to enable building ElectricSQL projects in TypeScript')
+    .version(LIB_VERSION)
+
+  // Add commands
+  ;[
```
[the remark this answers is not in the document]
samwillis: Blame prettier, it's stuck it here.
If there isn't a semicolon between the previous method call and the opening of the array litral, js thinks you are trying to index into something. If I try to move the semicolon to the end of the previous statment prettier moves it back.

clients/typescript/src/cli/config.ts
```diff
@@ -0,0 +1,188 @@
+import type { Command } from 'commander'
+import { extractDatabaseURL } from './utils'
+import { configOptions } from './config-options'
+
+export interface AnyConfigOption {
+  doc: string
+  valueType: typeof String | typeof Number | typeof Boolean
```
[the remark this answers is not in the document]
samwillis: I've used other runtime apis where you use the `String`, `Number` and `Boolean` constructors to indicate types. Much of a muchness, happy to change if wanted. Only ugly bit is here in the type.

clients/typescript/src/cli/config.ts
```diff
@@ -0,0 +1,188 @@
...
+  valueTypeName?: string
+  shortForm?: string
+  defaultVal?: string | number | boolean | (() => string | number | boolean)
```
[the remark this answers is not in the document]
samwillis: Yep, exactly. we have various derived options.

clients/typescript/src/cli/config.ts
```diff
@@ -0,0 +1,188 @@
...
+export function getConfigValue<K extends ConfigOptionName>(
+  name: K,
+  options?: any
+): ConfigOptionValue<K> {
+  // First check if the option was passed as a command line argument
+  if (options) {
+    const optName = snakeToCamel(name.toLocaleLowerCase())
+    if (options[optName] !== undefined) {
+      return options[optName] as ConfigOptionValue<K>
+    } else if (options[name] !== undefined) {
```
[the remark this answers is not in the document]
samwillis: We export the helpful `getConfig` function that constructs a full config from parsed arguments and any env vars. This could be used by our users.

It excepts both the camel case and uppercase variants:

image

clients/typescript/src/cli/config.ts
```diff
@@ -0,0 +1,188 @@
...
+  return Object.fromEntries(
+    Object.keys(config).map((name) => [
+      name.startsWith('ELECTRIC_') ? name : `ELECTRIC_${name}`,
+      config[name as ConfigOptionName]?.toString(),
+    ])
+  )
+}
+
+function snakeToCamel(s: string) {
+  return s.toLocaleLowerCase().replace(/(_\w)/g, (m) => m[1].toUpperCase())
```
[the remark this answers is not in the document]
samwillis: I've change to match behaviour of `Camelize<>` type

```ts
s.toLocaleLowerCase().replace(/(_+\w)/g, (m) => m.slice(-1).toUpperCase())
```

image
image

clients/typescript/src/cli/config-options.ts
```diff
@@ -0,0 +1,280 @@
...
+- when using WITH_POSTGRES, set the default DATABASE_URL to the postgres container
+*/
+
+// Name will be prefixed with ELECTRIC_ as environment variables.
+export const configOptions = {
+  // *** Client options ***
+  SERVICE: {
+    valueType: String,
+    valueTypeName: 'url',
+    doc: dedent`
```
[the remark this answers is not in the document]
samwillis: No need other than uniformity.

clients/typescript/src/cli/config-options.ts
samwillis: If the user has an env var of `ELECTRIC_DATABASE_URL=postgresql://postgres...` this derives the components from it. It defaults to a "derived" value.

clients/typescript/package.json
samwillis: Do we need them to as this is just for building the client?

clients/typescript/src/cli/configure/command-configure-ports.ts
samwillis: The user is very likly do modify this file themselves. I've written this so that it retains any formatatting that the user may have used.

clients/typescript/src/cli/config.ts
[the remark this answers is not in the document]
kevin-dp: yep, don't worry i can live with it

clients/typescript/src/cli/config-options.ts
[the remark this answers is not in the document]
kevin-dp: let's make it a plain string literal than

clients/typescript/package.json
kevin-dp: Ideally this would work on Windows too such that Windows users can also contribute to the project. I understand that these are some minor commands that will break which any dev that is on Windows should be able to fix so i can live with it for the time being, but ideally we fix this at some point.

clients/typescript/src/cli/configure/command-configure-ports.ts
kevin-dp: But if the user writes
```sh
    ELECTRIC_HTTP_PORT=5000
```
This won't be detected and we are going to append it at the end.
I understand this will work as the appended env var will replace the one of the user but not sure that is better than loosing "formatting"

clients/typescript/src/cli/migrations/migrate.ts
samwillis: That's a oopsie from a rebase.

clients/typescript/src/cli/migrations/migrate.ts
samwillis: Same rebase oopsie

clients/typescript/src/cli/docker-commands/command-start.ts
samwillis: Thanks, good spot. Removed it.

balegas (approved):
great work. This will greatly improve our users onboarding experience.

Thanks to everyone that participated in the review

PR #665, opened 2023-11-15
feat (client): client-side support for int8

kevin-dp:
This PR adds client-side support for int8 aka bigint.
Supporting BigInts is tricky because of inconsistencies in the sqlite drivers.
For example, better-sqlite3 requires BigInts to be explicitly enabled but then *all* integers are returned as BigInts because SQLite does not distinguish between them (they are all just 64bit integers). On the other hand, wa-sqlite returns a regular JS number if the integer fits, otherwise it returns a BigInt.

To avoid these problems, we always cast BigInt to text when reading from the DB and then let the DAL turn it into a BigInt;
i.e., when the Postgres type of a column "foo" is bigint, we read as follows: `SELECT cast(foo as TEXT) AS foo FROM table`. Similarly, we cast returned values: `INSERT INTO table VALUES (9223372036854775807) RETURNING cast(foo as TEXT) AS foo`.

A similar problem occurs in the oplog triggers. These triggers insert an oplog entry for every insert/update/delete. This oplog entry contains the old and new rows as JSON objects. However, JSON does not support BigInts:
```sql
sqlite> SELECT json_valid('{"a": 123n}');
0
```
We also can't write it as a regular number because when parsing the JSON it is read as a number and rounding errors occur:
```js
js> JSON.parse('{"a": 9223372036854775807}')
{ a: 9223372036854776000 }
```

Hence, the triggers also need to cast BigInts to text and the deserialisation function for oplog entries (`deserialiseRow`) needs to correctly deserialise them as BigInts (when the column type is int8/bigint).


clients/typescript/src/client/model/builder.ts
```diff
@@ -243,6 +277,13 @@ export class Builder {
     query: T
   ): T {
     return this._fields.reduce((query, field) => {
+      // if field is of type BigInt cast the result to TEXT
+      // because not all adapters deal well with BigInts
+      // the DAL will convert the string into a BigInt in the `fromSqlite` function from `../conversions/sqlite.ts`.
+      const pgType = this._tableDescription.fields.get(field)
+      if (pgType === PgBasicType.PG_INT8) {
+        return query.returning(`cast(${field} as TEXT) AS ${field}`)
```
alco, line 285: This string should be returned from a function if it's used in multiple places like this. If, say, we add quoting for these field identifiers later, it'll be easier to make the change once and not worry about missing some occurrences that will be left unquoted.

clients/typescript/src/migrators/triggers.ts
```diff
@@ -268,7 +274,10 @@ function joinColsForJSON(
   // casts the value to TEXT if it is of type REAL
   // to work around the bug in SQLite's `json_object` function
   const castIfNeeded = (col: string, targettedCol: string) => {
-    if (colTypes[col] === 'REAL') {
+    const tpes = colTypes[col]
+    const sqliteType = tpes.sqliteType
+    const pgType = tpes.pgType
+    if (sqliteType === 'REAL' || pgType === 'INT8' || pgType === 'BIGINT') {
       return `cast(${targettedCol} as TEXT)`
```
alco, line 281: Seeing this for the first time now, it's surprising we're not using a disciplined approach for interpolating fargments into SQL strings. I know it's out of scope for this PR but I wanted to hear your opinion on it: shouldn't we audit all table column references that are interpolated into strings like this and replace them with a centralized function that can quote all identifiers uniformly?

clients/typescript/test/client/prisma/schema.prisma
```diff
@@ -54,6 +54,7 @@ model DataTypes {
   uuid        String?   @db.Uuid     /// @zod.string.uuid()
   int2        Int?      @db.SmallInt /// @zod.number.int().gte(-32768).lte(32767)
   int4        Int?                   /// @zod.number.int().gte(-2147483648).lte(2147483647)
+  int8        BigInt?
```
alco, line 57: Why doesn't this field have the db native attribute or the validation comment?

alco (commented):
Stellar job! :star: :muscle: 

My only piece of feedback is that I'm not seeing any tests that verify the validation is working for bigints. I.e. trying to write a value that would exceed the range of what a 64-bit integer can hold.

clients/typescript/test/client/prisma/schema.prisma
kevin-dp: 1) It does not have the @db.BigInt attribute because the `BigInt` Prisma type identifies the PG type unambiguously, hence, you can leave it out. Prisma introspection also does not write that attribute.

2) We cannot add a validation comment because of some technical details how it is implemented.
Originally, the zod-prisma-types generator we re-use will create the schema: `z.number().bigint()`
but this only allows users to pass bigints. Like Prisma, we also want to allow users to pass regular integers for big int columns. So we modified the schema to be: `z.union([ z.number().int(), z.number().bigint() ])`. Now, the problem is that we cannot add validators like `gte` and `lte` on a union (only on int or bigint directly). So I had to modify the generator internally to add the `gte` and `lte` validations directly on the schema. Thus the generator will generate this schema for BigInt:
```ts
z.union([
  z.bigint().gte(-9223372036854775808n).lte(9223372036854775807n), 
  z.number().int().gte(Number.MIN_SAFE_INTEGER).lte(Number.MAX_SAFE_INTEGER).transform(BigInt)
])
```
(Also notice the `transform(BigInt)` to turn regular JS numbers into BigInt such that everything we write to the DB are BigInts)

clients/typescript/src/client/model/builder.ts
kevin-dp: You want to introduce a local `getField()` function that returns the field?
something like:
```ts
private returnAllFields<T extends QueryBuilder & ReturningMixin>(
    query: T
  ): T {
    const getField = (field: string) => field 
    return this._fields.reduce((query, field) => {
      // if field is of type BigInt cast the result to TEXT
      // because not all adapters deal well with BigInts
      // the DAL will convert the string into a BigInt in the `fromSqlite` function from `../conversions/sqlite.ts`.
      const pgType = this._tableDescription.fields.get(field)
      if (pgType === PgBasicType.PG_INT8) {
        return query.returning(`cast(${getField()} as TEXT) AS ${getField()}`)
      }
      return query.returning(getField())
    }, query)
  }
}
```

I don't really see the value of adding this now as we merely use the field without any changes.
I do see the value once we do more than just using the field name, e.g. add quoting, but now it is literally the identity function.

clients/typescript/src/migrators/triggers.ts
kevin-dp: We could indeed change all places were we quote explicitly by calling a `quote` function that does the quoting. That improves maintainability if at some later point in time we want to change the way we quote identifiers but i don't think we will ever change the quoting as it seems the only way to quote identifiers is to use double quotes.

kevin-dp:
> Stellar job! ⭐ 💪
> 
> My only piece of feedback is that I'm not seeing any tests that verify the validation is working for bigints. I.e. trying to write a value that would exceed the range of what a 64-bit integer can hold.

Thanks, i added a unit test for values that are out of range :-)

clients/typescript/src/client/model/builder.ts
alco: I was only referring to the act of interpolating a variable into a string which is then passsed as SQL for execution by SQLite. Depending on the context, this can lead to SQL injection attacks or merely client-side fatal errors.

What I'm saying is that we should properly quote all identifiers and values when interpolating them into strings of SQL. It doesn't take much effort for a particular field name to break our client and as a consequence the app that depends on it. Here's a quick example:

```
$ sqlite3
SQLite version 3.36.0 2021-06-18 18:36:39
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> create table foo(cast integer);
sqlite> insert into foo values (1), (2), (3);
sqlite> select * from foo;
1
2
3
sqlite> select cast(cast as text) from foo;
Error: near "as": syntax error
sqlite> select cast("cast" as text) from foo;
1
2
3
```

Perhaps the value in the `field` variable is already quoted? In that case, disregard this comment.

kevin-dp: Indeed, the arguments we pass to the function are already quoted, e.g. `castIfNeeded(col, `"${col}"`)`

PR #673, opened 2023-11-16
Fix ddlx caveats

thruflo:


alco (approved):
Nice changes.

I've pushed a couple of fixes to this branch. One fixes a typo and the other fixes a duplicate link that was used twice within the same admonition.

PR #675, opened 2023-11-16
chore(docs): Document the limitation of only the `public` schema being supported

alco:


docs/usage/data-modelling/migrations.md
```diff
@@ -239,6 +239,10 @@ See <DocPageLink path="integrations/backend" /> and <DocPageLink path="api/ddlx"
 
 There are currently a number of limitations on the data models and migrations that ElectricSQL supports.
 
+### Default schema
+
+Only tables in the default schema named [`public`](https://www.postgresql.org/docs/14/ddl-schemas.html#DDL-SCHEMAS-PUBLIC) can be electrified at the moment.
```
balegas, line 244:
```suggestion
Only tables in the default schema named [`public`](https://www.postgresql.org/docs/14/ddl-schemas.html#DDL-SCHEMAS-PUBLIC) can be electrified at the moment. We're working to lift this restriction.
```

balegas (changes requested):
Thanks for accommodating my suggestion so quickly. 
Before we merge, I'm calling in @kevin-dp to also document the current limitations with table name collisions and any other issues that we've identified so far with regards to generator, and whatever (see VAX-1101 and VAX-1158).

docs/usage/data-modelling/migrations.md
```diff
@@ -239,6 +239,19 @@ See <DocPageLink path="integrations/backend" /> and <DocPageLink path="api/ddlx"
 
 There are currently a number of limitations on the data models and migrations that ElectricSQL supports.
 
+### Table names
+
+To avoid name clashes between tables, we recommend naming your tables such that they match this regex: `[a-z][A-Za-z0-9_]`.
+Internally, any character that is not part of that regex will be removed, e.g. `_myTable` will be stripped to `myTable`.
+This may lead to name clashes between tables, e.g. `_myTable`, `123myTable`, and `myTable` all clash.
+
+In addition, we capitalise the first letter of table names.
```
thruflo, line 248: Where do we capitalise the table name -- in the Postgres schema? In PG and SQLite? Just during the generation process? If not in the actual DDL schemas then perhaps better to rephrase as clashes being case insensitive, so myTable and MyTable clash.

thruflo: Yup, so it's just internal to the client generation process. I suggest just saying the table name comparison is case insensitive.

kevin-dp: That's not entirely true, as we can differentiate `foobar` from `fooBar` but not from `Foobar`.

docs/usage/data-modelling/migrations.md
```diff
@@ -239,6 +239,19 @@ See <DocPageLink path="integrations/backend" /> and <DocPageLink path="api/ddlx"
 
 There are currently a number of limitations on the data models and migrations that ElectricSQL supports.
 
+### Table names
```
balegas, line 242:
```suggestion
### Table names

The client generator sanitises table names (because of an issue in an external library [link]). The rule of thumb is that the generator handles table names as case insensitive and removes any prefix of the table name that is not a letter. For example `_myTable`, `123myTable`, and `myTable` are all handled as the same table name. Electrifying tables that have a name clash causes errors during the generation.
```
- Please move this section to after the default schema section
- Do we have a link for the issue in the library?
- Is it true that it leads to an error? If not, it should.

docs/usage/data-modelling/migrations.md
thruflo: My point is don't make it sound like we're changing your tablenames in the your actual database schema.

docs/usage/data-modelling/migrations.md
kevin-dp: Ok, i moved it and added the link to the issue. Yes, it leads to an error.

kevin-dp (commented):
I applied Valter's suggestions, waiting for his approval to merge.

balegas (approved):

PR #677, opened 2023-11-17
capacitor-sqlite runInTransaction now takes the lock.

gregzo:
[EDIT] Following @kevin-dp 's work on generic adapters, this PR has been repurposed. The concurrency issue is now taken care of by implementing BatchDatabaseAdapter.

## ORIGINAL PR DESCRIPTION
This PR fixes a concurrency issue with the capacitor-sqlite database adapter. The mutex was previously only acquired in the `transaction` method, causing concurrency conflicts when calling `runInTransaction`.

Comments have been added to other public methods of the adapter in order to clarify the intended behaviour of each method. In effect, `run`, `execute` and `query` have ambiguous meanings depending on the layer (electric driver -> capacitor-sqlite -> native SQLite). It is therefore important to make sure that the methods as implemented correspond to what is expected of `electric`'s API.

### DIFF
--- a/.changeset/quick-rats-end.md
+++ b/.changeset/quick-rats-end.md
@@ -2,4 +2,4 @@
 "electric-sql": patch
 ---
 
-Capacitor sqlite DB driver re-imlemented to extend the new generic BatchDatabaseAdapter.
+Capacitor sqlite DB driver re-implemented to extend the new generic BatchDatabaseAdapter.
--- a/.changeset/quick-rats-end.md
+++ b/.changeset/quick-rats-end.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Capacitor sqlite DB driver re-imlemented to extend the new generic BatchDatabaseAdapter.
--- a/clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
+++ b/clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
@@ -41,6 +41,14 @@ export class DatabaseAdapter extends GenericDatabaseAdapter {
     return { rowsAffected: this.#rowsAffected }
   }
 
+  /**
+   *
+   * @returns the number of rows modifierd by the last exec or execBatch call.
+   * Because Capacitor-SQLite does not expose sqlite3_changes, the value returned here is cached
+   * from the previous query's reported value, which is 0 for queries other than INSERT, UPDATE or DELETE.
+   * Calling getRowsModified() right after execBatch should return an accurate aggregated result, regardless
+   * of the type of statements executed in the batch.
+   */
   getRowsModified() {
     return this.#rowsAffected
   }
--- a/clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
+++ b/clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
@@ -43,7 +43,7 @@ export class DatabaseAdapter extends GenericDatabaseAdapter {
 
   /**
    *
-   * @returns the number of rows modifierd by the last exec or execBatch call.
+   * @returns the number of rows modified by the last exec or execBatch call.
    * Because Capacitor-SQLite does not expose sqlite3_changes, the value returned here is cached
    * from the previous query's reported value, which is 0 for queries other than INSERT, UPDATE or DELETE.
    * Calling getRowsModified() right after execBatch should return an accurate aggregated result, regardless
--- a/clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
+++ b/clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
@@ -1,153 +1,47 @@
-import { capSQLiteSet } from '@capacitor-community/sqlite'
-import {
-  DatabaseAdapter as DatabaseAdapterInterface,
-  RunResult,
-  TableNameImpl,
-  Transaction as Tx,
-} from '../../electric/adapter'
-import { Row, SqlValue, Statement } from '../../util'
 import { Database } from './database'
-import { Mutex } from 'async-mutex'
+import { Row, SqlValue } from '../../util/types'
+import { Statement } from '../../util'
+import { BatchDatabaseAdapter as GenericDatabaseAdapter } from '../generic'
+import { capSQLiteSet } from '@capacitor-community/sqlite'
+import { RunResult } from '../../electric/adapter'
 
-export class DatabaseAdapter
-  extends TableNameImpl
-  implements DatabaseAdapterInterface
-{
-  #txMutex: Mutex
+export class DatabaseAdapter extends GenericDatabaseAdapter {
+  readonly db: Database
+  #rowsAffected = 0
 
-  constructor(public db: Database) {
+  constructor(db: Database) {
     super()
-    this.#txMutex = new Mutex()
+    this.db = db
   }
 
-  /**
-   * Executes a single SQLite statement, without taking the lock and without wrapping it in a transaction.
-   * @param statement:
-   * @returns
-   */
-  async run({ sql, args }: Statement): Promise<RunResult> {
+  async exec(statement: Statement): Promise<Row[]> {
     const wrapInTransaction = false
+    const result = await this.db.run(
+      statement.sql,
+      statement.args,
+      wrapInTransaction
+    )
 
-    const result = await this.db.run(sql, args, wrapInTransaction)
-    const rowsAffected = result.changes?.changes ?? 0
-    return { rowsAffected }
+    this.#rowsAffected = result.changes?.changes ?? 0
+
+    return result.changes?.values ? result.changes.values : []
   }
 
-  /**
-   * Runs one or more statements within a transaction block, taking the lock.
-   * @param statements
-   * @returns
-   */
-  async runInTransaction(...statements: Statement[]): Promise<RunResult> {
+  async execBatch(statements: Statement[]): Promise<RunResult> {
     const set: capSQLiteSet[] = statements.map(({ sql, args }) => ({
       statement: sql,
       values: (args ?? []) as SqlValue[],
     }))
 
     const wrapInTransaction = true
-
-    const releaseMutex = await this.#txMutex.acquire()
     const result = await this.db.executeSet(set, wrapInTransaction)
-    releaseMutex()
 
-    const rowsAffected = result.changes?.changes ?? 0
-    // TODO: unsure how capacitor-sqlite populates the changes value (additive?), and what is expected of electric here.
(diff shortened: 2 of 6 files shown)


clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
```diff
@@ -20,27 +20,46 @@ export class DatabaseAdapter
     this.#txMutex = new Mutex()
   }
 
+  /**
+   * Executes a single SQLite statement, without taking the lock and without wrapping it in a transaction.
+   * @param statement:
+   * @returns
```
kevin-dp, line 26: No need for @param and @returns if there is no explanation for them.

clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
```diff
@@ -20,27 +20,46 @@ export class DatabaseAdapter
     this.#txMutex = new Mutex()
   }
 
+  /**
+   * Executes a single SQLite statement, without taking the lock and without wrapping it in a transaction.
```
kevin-dp, line 24: I would not say "without wrapping it in a transaction" because SQLite will always wrap standalone reads/writes in a transaction (cf. https://www.sqlite.org/lang_transaction.html):
> No reads or writes occur except within a transaction. Any command that accesses the database (basically, any SQL command, except a few [PRAGMA](https://www.sqlite.org/pragma.html#syntax) statements) will automatically start a transaction if one is not already in effect. Automatically started transactions are committed when the last SQL statement finishes.

kevin-dp, line 24: The `run` and `query` methods do not take the lock, i believe that is problematic as it may lead to `run`/`query` statements being interleaved in a concurrent transaction. For example, Satellite starts a transaction using `transaction` or `runInTransaction` and the lock is acquired. These methods are async, while waiting for some async call, JS may handle a call to `run`/`query`. Since these methods do not use the lock, they can now become interleaved within the transaction, which should not have been the case (they should be executed in a transaction of their own).

clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
```diff
@@ -20,27 +20,46 @@ export class DatabaseAdapter
...
     const set: capSQLiteSet[] = statements.map(({ sql, args }) => ({
       statement: sql,
       values: (args ?? []) as SqlValue[],
     }))
 
     const wrapInTransaction = true
+
+    const releaseMutex = await this.#txMutex.acquire()
     const result = await this.db.executeSet(set, wrapInTransaction)
+    releaseMutex()
```
kevin-dp, line 51: If `executeSet` fails, an error is thrown and `releaseMutex()` is not called.
This should be wrapped in a try..finally block:
```ts
const releaseMutex = await this.#txMutex.acquire()
try {
  const result = await this.db.executeSet(set, wrapInTransaction)
  const rowsAffected = result.changes?.changes ?? 0
  return { rowsAffected }
} finally {
  releaseMutex()
}
```

clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
```diff
@@ -20,27 +20,46 @@ export class DatabaseAdapter
...
+
     const rowsAffected = result.changes?.changes ?? 0
     // TODO: unsure how capacitor-sqlite populates the changes value (additive?), and what is expected of electric here.
```
kevin-dp, line 54: Electric expects the value to be additive, i.e. to be the total number of changes.

kevin-dp (changes requested):
Hi @gregzo, thanks for your PR!
I left some comments as i believe we also need to lock the `run` and `query` methods.

clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
kevin-dp: Perhaps it would be easier if we-rewrite the adapter to extend the generic adapter and override `runInTransaction` to execute the queries in batch using `executeSet`. Even better, i can perhaps modify the generic adapter to allow concrete adapters to provide an optional `executeSet` method; then the generic adapter can have a default implementation for `runInTransaction` that uses `executeSet` if provided and otherwise resorts to executing the queries manually one after the other in a transaction. That would considerably simplify the implementation of the capacitor-sqlite adapter.

gregzo: Indeed, this is why I did not add locks to `query` and `run` methods. A recursive mutex could work though. But your suggestion to add an optional `executeSet` method sounds much better - transaction complexity will be handled higher up whilst retaining an opportunity for using better optimised driver APIs, 👍.

Should I wait for this enhancement of the generic adapter and rewrite the whole thing then?

kevin-dp: Yes, i will implement it now. And comment here when it is done such that you can rewrite on top of that branch.

kevin-dp: @gregzo Here you go: https://github.com/electric-sql/electric/pull/685 (still needs to be reviewed so is subject to change).

gregzo: @kevin-dp That's beautiful, and very straightforward to implement thanks!

I've resolved all conversations since they're not relevant anymore - my last commit implements your BatchDatabaseAdapter.

I've rebased on `main` before, so it should be ready to merge.

2 things:

1. Capacitor-SQLite returns `numRowsAffected` within the `run` transaction result, but doesn't offer an API to retrieve that value separately. I've simply cached it at the adapter level. Is that safe? Risk of race condition?
2. I've implemented the `exec` method by calling the lib's `run` method, which is meant for single SQL statements and supports binding values. Can you please confirm that `exec` will never be called with multiple statements? In which case, I'd need to map it differently.

clients/typescript/src/drivers/capacitor-sqlite/database.ts
```diff
@@ -8,15 +8,6 @@ type OriginalDatabase = SQLiteDBConnection
 // The relevant subset of the SQLitePlugin database client API
 // that we need to ensure the client we're electrifying provides.
 // TODO: verify which functions we actually need.
-export interface Database
```
kevin-dp, line 11: Looks like the todo on L10 has become obsolete?

clients/typescript/src/drivers/capacitor-sqlite/test.ts
```diff
@@ -9,13 +9,14 @@ import { Notifier } from '../../notifiers/index'
 import { MockNotifier } from '../../notifiers/mock'
 import { MockRegistry } from '../../satellite/mock'
 
-import { DatabaseAdapter } from './adapter'
+import { DatabaseAdapter as CapacitorSQLiteAdapater } from './adapter'
```
kevin-dp, line 12: Typo in `CapacitorSQLiteAdapater` -> `CapacitorSQLiteAdapter`

clients/typescript/src/drivers/capacitor-sqlite/test.ts
```diff
@@ -47,7 +48,7 @@ export const initTestable = async <
   const dal = await electrify(
     dbName,
     dbDescription,
-    adapter,
+    adapter as DatabaseAdapter,
```
kevin-dp, line 51: Why do you need to cast the Capacitor sqlite DB adapter to the better-sqlite3 DB adapter?

kevin-dp (changes requested):
Thanks for your contribution, the PR looks good!
Just some minor things to fix.

.changeset/quick-rats-end.md
```diff
@@ -0,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Capacitor sqlite DB driver re-imlemented to extend the new generic BatchDatabaseAdapter.
```
kevin-dp, line 5: typo in "re-imlemented" -> "re-implemented"

clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
```diff
@@ -1,134 +1,47 @@
...
-    const wrapInTransaction = false // Default is true. electric calls run from within transaction<T> so we need to disable transactions here.
+  async exec(statement: Statement): Promise<Row[]> {
+    const wrapInTransaction = false
+    const result = await this.db.run(
+      statement.sql,
+      statement.args,
+      wrapInTransaction
+    )
+
+    this.#rowsAffected = result.changes?.changes ?? 0
```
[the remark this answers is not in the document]
gregzo: Looking at the Swift implementation, it appears that changes are tracked globally with `sqlite3_total_changes(sqlite3*);`

How it works:
1. The concrete swift imp of Database computes a changes diff by caching `UtilsSQLCipher.dbChanges()` before executing queries. I didn't check all methods in detail, but the two that we use certainly do (`runSQL` and `execSet`). **Database.swift:**
https://github.com/capacitor-community/sqlite/blob/master/ios/Plugin/Database.swift
2. `UtilsSQLCipher.dbChanges` simply maps to `sqlite3_total_changes(sqlite3*);` **UtilsSQLCipher.swift:** https://github.com/capacitor-community/sqlite/blob/master/ios/Plugin/Utils/UtilsSQLCipher.swift)

**`sqlite3_total_changes(sqlite3*)` doc:**
https://github.com/capacitor-community/sqlite/blob/master/ios/Plugin/Database.swift

My understanding is that the current adapter implementation should be fine for `execBatch` (routed to `execSet`), but will indeed return 0 changes in a select query executed through the adapter's `exec`, which maps to `runSQL`, which will compute a diff of total changes of zero even if a previous update transaction did change rows.

If this is a problem, we can selectively route the adapter's `exec` imp to cap-sqlite's `query` instead of `run`, and not update the cached changes value for select queries. That should work fine. Should I give it a go? If so, which util can I use to determine if a query is a single select statement?

clients/typescript/src/drivers/capacitor-sqlite/test.ts
gregzo: Not sure how that happened, good catch thanks!

kevin-dp (approved):

clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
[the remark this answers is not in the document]
gregzo: As per our Discord discussion, the PR should be ready to merge. We've determined that `getTowsAffected()` returning 0 if called after a SELCT query is acceptable for now.

PR #678, opened 2023-11-17
perf: Change useState to useRef/useMemo for values not used for rendering

KyleAMathews:
This is the general rule for state vs. ref — only use state if it's
rendered — which in the case of useLiveQuery is just the results.

Changing e.g. the subscription key shouldn't cause a render.

I noticed a lot of extra renders happening in my components with
useLiveQuery which I traced to this problem.

Without this PR, my component rendered 8 times, with it,
it renders 2 times (starting the query and then showing results).

### DIFF
--- a/.changeset/curvy-bobcats-walk.md
+++ b/.changeset/curvy-bobcats-walk.md
@@ -1,5 +1,0 @@
----
-"electric-sql": patch
----
-
-Improved client events type safety
--- a/.changeset/friendly-oranges-fold.md
+++ b/.changeset/friendly-oranges-fold.md
@@ -1,6 +1,0 @@
----
-"electric-sql": patch
----
-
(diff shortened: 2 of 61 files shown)


KyleAMathews:
New trick I learned in this — put all dependencies in https://github.com/simbathesailor/use-what-changed & it'll show what's changing on each render.

thruflo:
> This is the general rule for state vs. ref — only use state if it's rendered — which in the case of useLiveQuery is just the results.

Right, I didn’t realise you could useRef like this. That’s really good to know!

> Without this PR, my component rendered 8 times, with it, it renders 2 times

So much better!

thruflo (approved):

KyleAMathews:
Flickers when rendering drives me crazy haha. It was going to bug me until I figured out what was going on.

PR #679, opened 2023-11-18
chore(electric): Do not prematurely discard type OIDs and remove unused bits from OidDatabase

alco:
We have been unnecessarily converting type OIDs consumed from the logical replication stream to atomized type names. But when you add custom, user-defined types into the picture, it's easy to lose information such that you won't be able to convert a type name back into an OID.

This change removes the conversion in prepartion of a more thorough revamp of how we maintain type info in Electric that is being done as part of my work on VAX-1040.


components/electric/lib/electric/postgres/oid_database.ex
```diff
@@ -13,70 +17,47 @@ defmodule Electric.Postgres.OidDatabase do
...
-
-  def type_length({:array, _}), do: raise("Can't get type length of an array type")
-
-  def type_length(name) do
-    case :ets.match(@ets_table_name, pg_type(name: name, length: :"$1")) do
-      [[len]] -> len
-      _ -> raise("Unknown name #{name}")
+  @spec pg_type_for_name(atom | binary) :: PgType.t()
+  def pg_type_for_name(name) do
+    case :ets.match_object(@oid_table, pg_type(name: name)) do
```
magnetised, line 46: why `match_object/2`? we could just use `:ets.lookup(@oid_table, name)`, probably not much of a perf  difference but simpler conceptually

components/electric/lib/electric/postgres/oid_database.ex
```diff
@@ -13,70 +17,47 @@ defmodule Electric.Postgres.OidDatabase do
...
-    {:ok, table} =
-      ETS.Set.new(
-        name: @ets_table_name,
-        keypos: 2,
-        protection: :protected,
-        read_concurrency: true
-      )
-
-    ETS.Set.put(table, Electric.Postgres.OidDatabase.Defaults.get_defaults())
+    oid_table = :ets.new(@oid_table, [:named_table, keypos: 2, read_concurrency: true])
```
magnetised, line 53:
```suggestion
    oid_table = :ets.new(@oid_table, [:set, :named_table, keypos: 2, read_concurrency: true])
```

if we're explicit here then we don't have to remember the default type for an ets table

components/electric/lib/electric/postgres/oid_database/pg_type.ex
```diff
@@ -8,46 +8,57 @@ defmodule Electric.Postgres.OidDatabase.PgType do
...
+
+    pg_type(
+      namespace: namespace,
+      name: type_name(namespace, name, kind),
+      oid: String.to_integer(oid),
+      array_oid: array_oid,
+      element_oid: String.to_integer(element_oid),
+      length: String.to_integer(len),
+      kind: kind,
+      is_array: array_oid == 0
```
magnetised, line 51:
```suggestion
      is_array: array_oid == "0"
```
then we can lose the `String.to_integer()` call on line 41

components/electric/test/electric/replication/postgres/migration_consumer_test.exs
```diff
@@ -118,9 +118,9 @@ defmodule Electric.Replication.Postgres.MigrationConsumerTest do
           name: "mistakes",
           replica_identity: :all_columns,
           columns: [
-            %Relation.Column{flags: [:key], name: "id", type: :uuid, type_modifier: nil},
-            %Relation.Column{flags: [], name: "value", type: :text, type_modifier: nil},
-            %Relation.Column{flags: [], name: "price", type: :int4, type_modifier: nil}
+            %Relation.Column{flags: [:key], name: "id", type_oid: 2950, type_modifier: nil},
```
magnetised, line 121: is it worth keeping the type name in the data structure, just for usability when debugging etc? looking at these structs here gives me a strong sense of unease.

magnetised (approved):
nice. have a few minor comments only. i do think that keeping the type name in the data structures -- if possible -- would help us in undefinable ways

components/electric/lib/electric/postgres/oid_database.ex
alco: The ETS table uses `oid` as the key for records but here we are looking up a type by its `name`.

components/electric/test/electric/replication/postgres/migration_consumer_test.exs
alco: I have the cure! It's in this PR nextdoor - https://github.com/electric-sql/electric/pull/680/commits/664ba2e8dcaca0b4695f466dc485582244a14bda#diff-336532676a5a83b08f5d382a0651e4b7978ea3362f49fb5ceb4a6989aab887e5.

components/electric/lib/electric/postgres/oid_database/pg_type.ex
```diff
@@ -8,46 +8,57 @@ defmodule Electric.Postgres.OidDatabase.PgType do
...
-          element_oid: String.to_integer(element_oid),
-          length: String.to_integer(len),
-          type: map_type(type),
-          base_type: String.to_integer(base_type),
-          rel_id: String.to_integer(rel_id),
-          is_array: Map.fetch!(%{"t" => true, "f" => false}, is_array)
-        )
+  def pg_type_from_tuple({namespace, name, oid, array_oid, element_oid, len, typtype}) do
+    kind = kind(typtype)
+    array_oid = String.to_integer(array_oid)
```
alco, line 41:
```suggestion
```

components/electric/lib/electric/postgres/oid_database/pg_type.ex
```diff
@@ -8,46 +8,57 @@ defmodule Electric.Postgres.OidDatabase.PgType do
...
+
+    pg_type(
+      namespace: namespace,
+      name: type_name(namespace, name, kind),
+      oid: String.to_integer(oid),
+      array_oid: array_oid,
```
alco, line 47:
```suggestion
      array_oid: String.to_integer(array_oid),
```

PR #680, opened 2023-11-18
chore(electric): Cleanups in MigrationConsumer; ignore incoming Relation messages

alco:
I have extracted this from my work on VAX-1040.

Per Postgres' logical replication protocol, the publisher must send a Relation message to the subscriber before sending the first insert/update/delete operation for a given table since the current connection was established. In Electric, we don't have a use for those message since we maintain information about the electrified schema up to date outside of the replication connection.

I've also snuck in an optimization for the count of electrified tables that we use in a telemetry event.


components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
```diff
@@ -389,37 +363,25 @@ defmodule Electric.Replication.Postgres.LogicalReplicationProducer do
...
-
-      true ->
-        {true, state}
+  # Limit replication of electric.* tables since they are not expected to be
+  # replicated further from PG.
+  defp maybe_ignore_relation(state, %Relation{} = rel) do
+    # We do not encourage developers to use 'electric' schema, but some
+    # tools like sysbench do that by default, instead of 'public' schema
+    if rel.id in state.ignored_relations or
+         (rel.namespace == "electric" and rel.name in ["migrations", "meta"]) do
```
magnetised, line 372: i wonder about the need for this -- afaik we don't actually create an `electric.meta` table in pg, and the `migrations` table certainly isn't added to the logical replication publication.

this entire clause may be a holdover from the times when we published the entire db over LR. certainly now if any table is in the LR stream it's because we need it later on

components/electric/lib/electric/replication/postgres/migration_consumer.ex
```diff
@@ -175,17 +149,29 @@ defmodule Electric.Replication.Postgres.MigrationConsumer do
   end
 
   defp perform_migration({version, stmts}, state) do
-    {:ok, loader} = apply_migration(version, stmts, state.loader)
+    {:ok, loader, schema} = apply_migration(version, stmts, state.loader)
 
-    {:ok, table_count} = SchemaLoader.count_electrified_tables(loader)
+    {:ok, table_count} = SchemaLoader.count_electrified_tables(schema)
```
magnetised, line 154: This should probably be a call to a (new) function in  `Postgres.Schema` since we have a schema instance and so don't need to use the loader abstraction to get one (in order to then count the tables). 

overloading the "loader" metaphor to do work on an actual schema instance is probably wrong -- we should re-write the `count_electrified_tables` function to call to `load`  via the behaviour, then use the function on the schema module to do the work of counting.

magnetised (commented):
delicious deletions. nice to get rid of the relation messages early, it makes things easier. 

can we re-work the use of `SchemaLoader`  to more cleanly separate loading and using schema instances please. I'm happy to keep the `count_electrified_tables` function in the schema loader, but i don't think we should be overloading it's implementation to handle an actual schema instance.

components/electric/lib/electric/replication/postgres/migration_consumer.ex
alco: You're absolutely right. That's just me being lazy after spending hours fixing tests just to get my stuff to work. I'll move this function to `Schema`.

components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
alco: Really good catch! I haven't found a confirmation that a table called `"meta"` had existed in our internal schema at any point. I'm just going to remove all of this "ignored relations" code.

magnetised (approved):

PR #684, opened 2023-11-20
chore (client): prioritize UI queries over Satellite queries

kevin-dp:
This PR modifies the generic DB adapter and the capacitor-sqlite DB adapter to use a priority-aware mutex such that UI queries have higher priority than Satellite queries.


KyleAMathews:

PR #685, opened 2023-11-20
chore (client): provide generic DB adapters for serial and batch execution of SQL

kevin-dp:
This PR splits the generic database adapter into two separate ones, one for serial execution of SQL queries and one for batched execution of SQL queries.

### DIFF
--- a/clients/typescript/src/drivers/generic/adapter.ts
+++ b/clients/typescript/src/drivers/generic/adapter.ts
@@ -33,7 +33,7 @@ abstract class DatabaseAdapter
   abstract exec(statement: Statement): Promise<Row[]>
 
   /**
-   * @returns The number of rows modified by the last SQL query.
+   * @returns The number of rows modified by the last insert/update/delete query.
    */
   abstract getRowsModified(): number
 
--- a/clients/typescript/src/drivers/generic/adapter.ts
+++ b/clients/typescript/src/drivers/generic/adapter.ts
@@ -10,60 +10,42 @@ import { Mutex } from 'async-mutex'
 import { AnyDatabase } from '..'
 
 /**
- * A generic adapter that manages transactions manually by executing `BEGIN` and `COMMIT`/`ROLLBACK` commands.
+ * A generic database adapter.
  * Uses a mutex to ensure that transactions are not interleaved.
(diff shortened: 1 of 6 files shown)


clients/typescript/src/drivers/generic/adapter.ts
```diff
@@ -130,6 +112,66 @@ export abstract class DatabaseAdapter
...
+    // Uses a mutex to ensure that transactions are not interleaved.
+    const release = await this.txMutex.acquire()
+    let open = false
+    let rowsAffected = 0
+    try {
+      await this.exec({ sql: 'BEGIN' })
+      open = true
+      for (const stmt of statements) {
+        await this.exec(stmt)
+        if (isInsertUpdateOrDeleteStatement(stmt.sql)) {
```
magnetised, line 154: seems like an expensive operation to run on every statement -- what happends if you run `getRowsModified()` on a select? if it returns 0 then we could avoid downcasing every statement

alternatively make `isInsertUpdateOrDeleteStatement` more efficient by trimming the string before lower casing it

magnetised (approved):
nice change. makes me almost miss java

clients/typescript/src/drivers/generic/adapter.ts
icehaunter: We can make `isInsertUpdateOrDeleteStatement` even more efficient with a regex tho
```typescript
function isInsertUpdateOrDeleteStatement(stmt: string): boolean {
  return /^\s*(insert|update|delete)/i.test(stmt)
}
```

icehaunter (approved):
LGTM

clients/typescript/src/drivers/generic/adapter.ts
kevin-dp: > seems like an expensive operation to run on every statement -- what happends if you run `getRowsModified()` on a select? if it returns 0 then we could avoid downcasing every statement
> 
> alternatively make `isInsertUpdateOrDeleteStatement` more efficient by trimming the string before lower casing it

Unfortunately, we can't, because most drivers will use this function https://www.sqlite.org/c3ref/changes.html which returns the number of changes made by the last insert/update/delete. I will improve our docstring for `getRowsModified` to mention that that is the expected behavior. And i will follow @icehaunter's suggestion to use a regex.

PR #688, opened 2023-11-21
fix (client): extend wa-sqlite's critical section to avoid bad interleavings

### LINKED ISSUE #682: critical section for wa-sqlite isn't properly scoped
I don't think the synchronization in this method is correct:

https://github.com/electric-sql/electric/blob/d385d3cf7eb4c8822afefdd6173fad7feb71712a/clients/typescript/src/drivers/wa-sqlite/database.ts#L37-L87

The mutex should protect _all_ the wa-sqlite calls but it is unconditionally released at [line 50](https://github.com/electric-sql/electric/blob/d385d3cf7eb4c8822afefdd6173fad7feb71712a/clients/typescript/src/drivers/wa-sqlite/database.ts#L50).

kevin-dp:
This PR extends the critical section of our `exec` method for wa-sqlite which fixes #682.
The mutex now wraps all wa-sqlite calls to avoid any bad interleavings.

### DIFF
--- a/.changeset/brave-rules-walk.md
+++ b/.changeset/brave-rules-walk.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Fix critical section of wa-sqlite DB adapter to avoid bad interleavings.
--- a/clients/typescript/src/drivers/wa-sqlite/database.ts
+++ b/clients/typescript/src/drivers/wa-sqlite/database.ts
@@ -39,24 +39,19 @@ export class ElectricDatabase {
     // otherwise wa-sqlite may encounter problems such as indices going out of bounds
     const release = await this.#mutex.acquire()
 
-    const str = this.sqlite3.str_new(this.db, statement.sql)
-    let prepared
+    let stmt: number | null = null
     try {
-      prepared = await this.sqlite3.prepare_v2(
+      const str = this.sqlite3.str_new(this.db, statement.sql)
+      const prepared = await this.sqlite3.prepare_v2(
         this.db,
         this.sqlite3.str_value(str)
       )
-    } finally {
-      release()
-    }
 
-    if (prepared === null) {
-      release()
-      return []
-    }
+      if (prepared === null) {
+        return []
+      }
 
-    const stmt = prepared.stmt
-    try {
+      stmt = prepared.stmt
       if (typeof statement.args !== 'undefined') {
         this.sqlite3.bind_collection(
           stmt,
@@ -81,7 +76,9 @@ export class ElectricDatabase {
       }
       return resultToRows(res)
     } finally {
-      await this.sqlite3.finalize(stmt)
+      if (stmt !== null) {
+        await this.sqlite3.finalize(stmt)
+      }
       release()
     }
   }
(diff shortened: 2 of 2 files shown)


alco:
Hey, the technical details of this are way over my head, I just wanted to make a code style suggestion: I would put the code that's currently sits between `try` and `finally` into a separate function to make the `try ... finally` control flow structure itself immediately obvious. 

It could also be useful to add a comment explaining while it all needs to be inside one `try` block followed by one `release()` call, since the previous implementation was incorrectly doing multiple `try` and `release()`.

rhashimoto:
The string allocation on this line appears to leak:

https://github.com/electric-sql/electric/blob/45d55b8b6147634b6b39f87a7a8d07bdb56cc05a/clients/typescript/src/drivers/wa-sqlite/database.ts#L44

It should be balanced with a call to [sqlite3.str_finish()](https://rhashimoto.github.io/wa-sqlite/docs/interfaces/SQLiteAPI.html#str_finish); the finally block would be a good place.

rhashimoto:
You could use the [statements()](https://rhashimoto.github.io/wa-sqlite/docs/interfaces/SQLiteAPI.html#statements) convenience function to simplify your application code a bit by handling the statement preparation boilerplate. It would be something like:

```javascript
const release = await this.#mutex.acquire()
try {
  for await (const stmt of sqlite3.statements(this.db, statement.sql)) {
    if (typeof statement.args !== 'undefined') {
      this.sqlite3.bind_collection(
        stmt,
        statement.args as
          | { [index: string]: SQLiteCompatibleType }
          | SQLiteCompatibleType[]
      )
    }

    const rows: SqlValue[][] = []
    let cols: string[] = []

    while ((await this.sqlite3.step(stmt)) === SQLite.SQLITE_ROW) {
      cols = cols.length === 0 ? this.sqlite3.column_names(stmt) : cols
      const row = this.sqlite3.row(stmt) as SqlValue[]
      rows.push(row)
    }

    const res = {
      columns: cols,
      values: rows,
    }
    return resultToRows(res)  // exit loop after one statement
  }
  return [];  // will get here only if there is no statement
} finally {
  release();
}
```

A mild downside is you'll need to reinstate all the preparation code if you eventually cache prepared statements.

kevin-dp:
> You could use the [statements()](https://rhashimoto.github.io/wa-sqlite/docs/interfaces/SQLiteAPI.html#statements) convenience function to simplify your application code a bit by handling the statement preparation boilerplate. It would be something like:
> 
> ```js
> const release = await this.#mutex.acquire()
> try {
>   for await (const stmt of sqlite3.statements(this.db, statement.sql)) {
>     if (typeof statement.args !== 'undefined') {
>       this.sqlite3.bind_collection(
>         stmt,
>         statement.args as
>           | { [index: string]: SQLiteCompatibleType }
>           | SQLiteCompatibleType[]
>       )
>     }
> 
>     const rows: SqlValue[][] = []
>     let cols: string[] = []
> 
>     while ((await this.sqlite3.step(stmt)) === SQLite.SQLITE_ROW) {
>       cols = cols.length === 0 ? this.sqlite3.column_names(stmt) : cols
>       const row = this.sqlite3.row(stmt) as SqlValue[]
>       rows.push(row)
>     }
> 
>     const res = {
>       columns: cols,
>       values: rows,
>     }
>     return resultToRows(res)  // exit loop after one statement
>   }
>   return [];  // will get here only if there is no statement
> } finally {
>   release();
> }
> ```
> 
> A mild downside is you'll need to reinstate all the preparation code if you eventually cache prepared statements.

Thanks for the suggestion, i had missed the convenience function.
Rewrote it as suggested 😃

balegas (approved):
LGTM!
@rhashimoto thanks for the help and nice seeing you here! 👋

PR #689, opened 2023-11-21
WIP chore(examples): New basic Supabase example demonstrating auth and edge functions

samwillis:
See ./examples/supabase-basic/README.md for how to run.

Note that this depends on a new version of Electric that uses the `sub` claim in the Supabase JWT rather than `user_id`. This will be a separate PR.


geoman2:
Hello
Can anyone replace node js with bun as recently bun added support for supabase, would be a good alternative

PR #691, opened 2023-11-22
chore (client): upgrade version of wa-sqlite

kevin-dp:
Upgrades the ts-client, the starter, and the examples to use wa-sqlite version ^0.9.8 in order to include a critical bug fix released in v0.9.8.


samwillis (approved):
👍 
Nice trick with the #semver on a GitHub package.

PR #692, opened 2023-11-22
chore(electric): Swap to using `sub` claim in jwt, backwards compatible with `user_id`

samwillis:
The standard `sub` "subject" claim is where most auth providers will place any pk for the user, we should default to that with backwards compatibility with our `user_id` claim.

In future we will also want to make this configurable.

Todo:

- [ ] Update docs
- [ ] Update examples


kevin-dp:
Looks good to me, but i will let someone with more knowledge double check.

alco (approved):
Brilliant! :+1:

alco:
@samwillis I have pushed a commit that addresses Elixir formatter's failed check.

PR #695, opened 2023-11-23
chore(examples): New "Checkout" demo using Supabase with event sourcing.

samwillis:
WIP demo Checkout app that works with Supabase, using `pg_net` (the underlying extension of Supabase web hooks) and Supabase Auth. UI is built with the Ionic framework.

Supabase Auth is used to register and sign in a user, the Supabase JWT is then used to establish a connection with Electric.

A Supabase Edge Function is used to process the card payment on submission of the checkout. The card details would be converted to a token by a payment provider and attached to the order record, there is then a trigger on the `order` table. When a new order is inserted it calls an edge function to process the order updating it's status in real time and syncing that with the visitors local checkout.

Requires #692 to enable Supabase Auth, and #698 for direct writes to Supabase Postgres.


balegas:
Closes VAX-1093

balegas (approved):
Because it’s Christmas you get this one for free!

PR #696, opened 2023-11-23
chore (client): use async event emitter in Satellite client

kevin-dp:
## Problem
The Satellite client uses an `EventEmitter` to emit events to which the Satellite process is subscribed. However, the Satellite process registers asynchronous event listeners but the `EventEmitter` calls them synchronously. Hence, async listeners for an event may run concurrently and a new event may be emitted while async listeners of the previous event are still running.

An example in NodeJS:

```ts
const EventEmitter = require('node:events');
const emitter = new EventEmitter();
emitter.on('transaction', (tx) => {
  console.log('listener 1 start: ' + tx);
  return new Promise((resolve) => {
    setTimeout(() => {
      console.log('listener 1 end: ' + tx);
      resolve()
    }, 10)
  })
})

emitter.emit('transaction', 'tx1');
emitter.emit('transaction', 'tx2');
/*
   Prints:
     listener 1 start: tx1
     listener 1 start: tx2
     listener 1 end: tx1
     listener 1 end: tx2
 */

emitter.on('transaction', (tx) => {
  console.log('listener 2 start: ' + tx);
})

emitter.emit('transaction');

/*
   Prints:
     listener 1 start: tx3
     listener 2 start: tx3
     listener 1 end: tx3
 */
```

The first example above shows interleaving of async listeners on an event.
The second example shows interleaving of listeners for distinct events.

## Solution
Implemented an `AsyncEventEmitter` that awaits all listeners before processing the next event. Modified the Satellite client to use the `AsyncEventEmitter` instead of the regular `EventEmitter`. In addition, the `AsyncEventEmitter` is fully typed, inspired by the typings of the [typed-emitter package](https://www.npmjs.com/package/typed-emitter).

Note that, we start all listeners for an event in order **without awaiting**. Hence, async listeners registered for an event may run concurrently. However, we await all listeners before processing the next event.

With the new `AsyncEventEmitter` we get this behavior:

```ts
const emitter = new AsyncEventEmitter();
emitter.on('transaction', (tx) => {
  console.log('listener 1 start: ' + tx);
  return new Promise((resolve) => {
    setTimeout(() => {
      console.log('listener 1 end: ' + tx);
      resolve()
    }, 10)
  })
})

emitter.enqueueEmit('transaction', 'tx1');
emitter.enqueueEmit('transaction', 'tx2');
/*
   Prints:
     listener 1 start: tx1
     listener 1 end: tx1
     listener 1 start: tx2
     listener 1 end: tx2
 */

emitter.on('transaction', (tx) => {
  console.log('listener 2 start: ' + tx);
})

emitter.enqueueEmit('transaction');

/*
   Prints:
     listener 1 start: tx3
     listener 2 start: tx3
     listener 1 end: tx3
 */
```

The first example above shows that events are processed in order only after all listeners have finished.
The second example shows that listeners for a given event are started in order one after the other without awaiting.

Note: we renamed `emit` to `enqueueEmit` to avoid confusion with the classical `EventEmitter`'s `emit` function which has different semantics as it does not await all listeners before processing the next event.


clients/typescript/src/satellite/client.ts
```diff
@@ -105,38 +104,18 @@ const subscriptionError = [
...
-  removeListener(event: 'transaction', callback: TransactionCallback): void
-  removeListener(
-    event: 'outbound_started',
-    callback: OutboundStartedCallback
-  ): void
-
-  removeAllListeners(): void
-
-  listenerCount(event: 'error'): number
+type Events = {
```
balegas, line 107: This looks much better than the code above, but I'd prefer to reuse the defined the callbacks (TransactionCallback, etc) than declaring them again

clients/typescript/src/satellite/mock.ts
```diff
@@ -284,7 +297,7 @@ export class MockSatelliteClient extends EventEmitter implements Client {
     this.replicating = true
     this.inboundAck = lsn
 
-    const t = setTimeout(() => this.emit('outbound_started'), 100)
+    const t = setTimeout(() => this.emit('outbound_started', lsn), 100)
```
balegas, line 300: Curious how you caught this. I think the lsn is no longer used. Could you check that and change the event type if no longer used?

clients/typescript/src/satellite/client.ts
```diff
@@ -93,6 +91,7 @@ import { RPC, rpcRespond, withRpcRequestLogging } from './RPC'
 import { Mutex } from 'async-mutex'
 import { DbSchema } from '../client/model'
 import { PgBasicType, PgDateType, PgType } from '../client/conversions/types'
+import { AsyncEventEmitter } from '../util'
```
balegas, line 94: It misses the await calls, but honestly for me is so weird that now everything become async while before the event emitter was simple. I don't know what to expect anymore. More on this on my overall comment

balegas (commented):
Needs more discussion

clients/typescript/src/satellite/client.ts
kevin-dp: I'm not sure what you mean here. Do you mean to use `TransactionCallback` as the type for the `ackCb` parameter for the `transaction` event below?
`TransactionCallback` is not the same type as `() => void`.

clients/typescript/src/satellite/client.ts
kevin-dp: We discussed this over a call. Now, since the `AsyncEventEmitter` processes events from a queue, we are able to keep the `enqueueEmit` function synchronous while still ensuring that events are processed in order and only after all listeners of the previous event have finished.

clients/typescript/src/satellite/mock.ts
kevin-dp: It wouldn't type check anymore because the emit for this event was expecting an lsn to be passed in. Interestingly, the type checker wouldn't complain about the lsn not being used by the listener... because TypeScript allows passing arguments to functions that don't expect arguments.

clients/typescript/src/satellite/mock.ts
```diff
@@ -136,7 +140,16 @@ export class MockRegistry extends BaseRegistry {
...
 
-export class MockSatelliteClient extends EventEmitter implements Client {
+type Events = {
+  [SUBSCRIPTION_DELIVERED]: (data: SubscriptionData) => void
+  [SUBSCRIPTION_ERROR]: (error: SatelliteError, subscriptionId: string) => void
+  outbound_started: OutboundStartedCallback
+  error: ErrorCallback
+}
+export class MockSatelliteClient
+  extends AsyncEventEmitter<Events>
```
icehaunter, line 150: Curious as to why this is `extends` here but a separate private instance on the real client? I'd generally argue for exposing typed `on` methods on the client itself and doing away with `subscribeToSmth` method pattern, however that might be out of scope here.

icehaunter (approved):
LGTM, nice work. A couple of questions but nothing blocking -- address at your discretion. Only thing I would ask you to acknowledge explicitly is the error handling comment

clients/typescript/src/satellite/mock.ts
kevin-dp: It was already like that before so i left it how it was, only changed the `EventEmitter` to use `AsyncEventEmitter`. Your proposed refactoring would indeed make the code cleaner and more idiomatic but should probably be addressed in a separate PR as you suggest.

clients/typescript/src/util/asyncEventEmitter.ts
```diff
@@ -0,0 +1,252 @@
...
+
+      if (event === 'error' && listeners.length === 0) {
+        this.processing = false
+        throw args[0]
+      }
+
+      // deep copy because once listeners mutate the `this.listeners` array as they remove themselves
+      // which breaks the `map` which iterates over that same array while the contents may shift
+      const ls = [...listeners]
+      const listenerProms = ls.map((listener) => listener(...args))
```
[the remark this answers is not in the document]
kevin-dp: Good point. Will make the function async and await the listener such that sync listeners are promisified.

clients/typescript/src/satellite/mock.ts
balegas: I can shed some light on why mock still extends event emitter. The client used to be like that, but I refactored it to not extend event emitter anymore. I might have overseen the mock.

clients/typescript/src/util/asyncEventEmitter.ts
```diff
@@ -0,0 +1,252 @@
...
+  event: Event
+  args: Arg[]
+}
+
+/**
+ * Implementation of a typed async event emitter.
+ * (Asynchronous) event listeners are called in order and awaited
+ * such that the next listener is called only after the previous one finished.
+ * Typings are inspired by the 'typed-emitter' package.
+ */
```
samwillis, line 15: I'm not sure this description is right, it seems to suggest *all* event listers are awaited before the next is run. I don't think "next listener is called only after the previous one finished" - events are, it's conflating listener and events.

Isn't something more like:

This event emitter maintains a queue of events, the events are processed in order and the next is not started until all async event listeners for that event have returned. Event listers for a single event *will be processed concurrently*, events are processed sequentially.

samwillis (approved):

clients/typescript/src/util/asyncEventEmitter.ts
kevin-dp: Indeed, good catch!

PR #698, opened 2023-11-23
feat(electric): Add a new write-to-pg mode that applies changes directly to Postgres

alco:
When Electric is started in this write mode, instead of creating a subscription in Postgres and streaming client writes to it over a logical replication connection, it will use a regular connection to apply client writes as regular DML statements.

The `direct_writes` mode can be enabled as follows when running Electric from source code:
```
MIX_ENV=prod \
ELECTRIC_WRITE_TO_PG_MODE=direct_writes \
PG_PROXY_PASSWORD=<pwd> \
DATABASE_URL=<url> \
iex -S mix
```

Alternatively, you can check out my branch [`alco/direct-writes-to-pg-dev`](https://github.com/electric-sql/electric/tree/alco/direct-writes-to-pg-dev), start the development database with `make start_dev_env` and start Electric with:
```
AUTH_MODE=insecure iex -S mix
```


KyleAMathews:
Curious what's the goal here? Lower latency? Higher throughput?

thruflo:
> Curious what's the goal here? Lower latency? Higher throughput?

The key thing this unlocks is being able to run with lower permissions. Currently we use inbound logical replication for writing data from Electric into Postgres. This adds an alternative direct writes mode that uses a standard interactive Postgres client. This reduces the permissions needed and, for example, means we can run on vanilla hosted Supabase.

alco:
> Curious what's the goal here? Lower latency? Higher throughput?

Both can be improved later on with this change in place. Direct writes allow us to batch multiple statements into one, keep a cache of prepared statements and generally be more aware of when client updates get committed in Postgres. We also get the freedom to "spread" client updates over a pool of connections, such that multiple updates are committed in parallel, whereas up until now we had to merged them into a single stream of transactions to be fed to Postgres as a stream logical replication messages.

thruflo:
> Both can be improved later on with this change in place.

Interesting -- I was thinking under the assumption that direct writes would be slower, because you would wait for one transaction to be confirmed before sending the next? Is that true and then is it a case that that reduced throughput but that this reduction may be mitigated and actually outweighed by the performance improvements from the techniques you listed?

Is there something here in knowing when transactions can be written in parallel and is this something we already have enough metadata for?

KyleAMathews:
🔥 nice set of optimizations that are coming!

components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers/trigger_function_installers.sql
```diff
@@ -508,60 +508,62 @@ BEGIN
         CREATE OR REPLACE TRIGGER postgres_write__upsert_generate_shadow_rows
         BEFORE INSERT OR UPDATE ON %s
         FOR EACH ROW
+        WHEN (electric.__session_replication_role() <> 'replica')
         EXECUTE PROCEDURE electric.%I();
     $$, full_table_identifier, generated_functions->>'create_shadow_row_from_upsert');
 
-    EXECUTE format($$ ALTER TABLE %s ENABLE TRIGGER postgres_write__upsert_generate_shadow_rows $$, full_table_identifier);
+    EXECUTE format($$ ALTER TABLE %s ENABLE ALWAYS TRIGGER postgres_write__upsert_generate_shadow_rows $$, full_table_identifier);
```
icehaunter, line 515: Why do you need to `ENABLE ALWAYS` this trigger, given that this code path never needs to be triggered on `REPLICA` path regardless of this PR?

components/electric/lib/electric/replication/postgres/writer.ex
```diff
@@ -0,0 +1,217 @@
...
+    :gproc.await(producer_name, 1_000)
+
+    GenStage.async_subscribe(
+      self(),
+      [
+        to: {:via, :gproc, producer_name},
+        cancel: :temporary,
+        starting_from: position
+      ] ++ subscription_opts()
+    )
```
icehaunter, line 45: Why are we doing an `async_subscribe/2` call here instead of appending it to the `init` return value?

components/electric/lib/electric/replication/postgres/writer.ex
```diff
@@ -0,0 +1,217 @@
...
+    lsn = tx.lsn
+    values_sql = encode_values([{client_id, :text}, {lsn, :bytea}])
+
+    """
+    INSERT INTO #{Extension.acked_client_lsn_table()} AS t
+    VALUES (#{values_sql})
+    ON CONFLICT (client_id)
+      DO UPDATE
+        SET lsn = excluded.lsn
+      WHERE t.lsn IS DISTINCT FROM excluded.lsn
```
icehaunter, line 151: Why do we need this `WHERE` clause here?

components/electric/lib/electric/replication/postgres/writer.ex
```diff
@@ -0,0 +1,217 @@
...
+    """
+  end
+
+  defp change_to_statement(%Changes.NewRecord{record: data, relation: table}, relations) do
+    {table_schema, table_name} = table
+    columns = relations[table].columns
+
+    table_sql = quote_ident(table_schema, table_name)
+    columns_sql = map_join(columns, &quote_ident(&1.name))
```
icehaunter, line 160: Why do you need a special function if it already exists as [`Enum.map_join/3`](https://hexdocs.pm/elixir/1.12/Enum.html#map_join/3)?

components/electric/lib/electric/postgres/extension/functions/__session_replication_role.sql.eex
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE FUNCTION <%= @schema %>.__session_replication_role(OUT role text) AS $$
+BEGIN
+  SELECT INTO role current_setting('electric.session_replication_role');
+EXCEPTION WHEN undefined_object THEN
+  SELECT INTO role current_setting('session_replication_role');
```
icehaunter, line 5: It took a little bit of reasoning and searching to understand that these two lines are not related to `electric` schema being in the `search_path`, but rather reference two separate variables, second being PG-defined. I would appreciate a comment somewhere about that

icehaunter (commented):
Looks pretty good! I'm glad this functionality could be achieved in a relatively small code change.

I have multiple questions for particular lines, but also some general questions:

1. Why are you calling it `inbound` replication everywhere? We're configuring the system from the POV of Electric, to which streaming to PG is outbound. I found this confusing and would appreciate this changed as I can see myself stumbling over this later on
2. Why are we keeping direct streaming at all? Perfomance comparison? Direct writes seem like a more generic approach from POV of deployment (also less configuration in general)

Great job all around!

alco:
> I have multiple questions for particular lines, but also some general questions:

Hey, thanks for the review! @thruflo can answer those two general questions better than me but I'll give it a go:

1. You can see from the commit history that I started with `ELECTRIC_WRITE_MODE=streaming|immediate`. The `ELECTRIC_INBOUND_MODE` name was James' idea, I believe. It's a reference to the act of applying _inbound_ client updates via _logical replication_ or _direct writes_ to Postgres.
2. Again, not my idea, but it does let us keep existing users unaffected in case the new mode has edge cases we haven't uncovered yet. The new mode is operationally simpler and doesn't require a superuser role, so we will be making it the default mode soon enough, IMO. Then we'll be able to retire the code that deals with logical replication.

components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers/trigger_function_installers.sql
alco: Makes sense to keep it (an othe triggers like it) as `ENABLE TRIGGER` for a minor performance win. But I find it easier to reason about these triggers when their firing is conditional on a single variable: the return value of `electric.__session_replication_role()`.

Once we retire the logical replication mode, all of these triggers will be firing so it'll become a moot point whether to use `ALWAYS` or omit. For now I had to convert `REPLICA` triggers to `ALWAYS` to support both inbound modes. Converting the non-replica triggers to `ALWAYS` makes things more uniform.

components/electric/lib/electric/replication/postgres/writer.ex
alco: My bad, copy-pasting from SlotServer without thinking.

components/electric/lib/electric/replication/postgres/writer.ex
alco: When this code was executing inside a trigger without this `WHERE` condition, I remember seeing it get into an infinite loop. Or perhaps that was before I added the trigger depth limit. In any case, I'll have to test the behaviour without this clause before I feel comfortable removing it.

components/electric/lib/electric/replication/postgres/writer.ex
alco: I wasn't aware of it.

thruflo:
Yup, as @alco says.

Inbound derives from @balegas who I remember referenced MySQL naming. Which is database centric. Where @icehaunter is understandably thinking from Electric’s POV.

I discussed with @balegas because I wanted us to be intentional about the naming and I didn’t think “immediate” made sense. If there’s a better term than inbound great. But I’d lean towards the Postgres-centric view of discussing inbound vs outbound.

balegas:
I'm happy with any decision.
With regards to supporting both modes, let's keep things around until we learn more.

alco:
> > Both can be improved later on with this change in place.
> 
> Interesting -- I was thinking under the assumption that direct writes would be slower, because you would wait for one transaction to be confirmed before sending the next? Is that true and then is it the case that that therefore reduces throughput but that this reduction may be mitigated and actually outweighed by the performance improvements from the techniques you listed?

I believe that providing direct write performance comparable to logical replication is an easy baseline to achieve. Postgres supports query pipelining which means that even if we feed a single stream of changes, we don't have to wait for Postgres to response to every change before sending the next one. This already looks similar to the logical replication approach. On top of that we can add caching of prepared statements which implies encoding parameters using the binary format. This will result in fewer bytes sent over the wire and lower latency.

> Is there something here in knowing when transactions can be written in parallel and is this something we already have enough metadata for?

I may be speaking out of ignorance, but I think we have enough room for parallelization. At a very basic level, updates to disjoint sets of tables can be parallized. But even if two updates would conflict with each other, Postgres's MVCC should already be providing deterministic ordering of such concurrent updates, and our conflict-resolution logic will keep things consistent.

components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers/trigger_function_installers.sql
alco: I've thought about this a bit more and you're right -- no need to turn any `ENABLE TRIGGER` into `ENABLE ALWAYS TRIGGER`. I've reverted this change for all affected triggers.

icehaunter (approved):

PR #699, opened 2023-11-23
tests(client): Include tests for bigint support in the oplog triggers and the merge logic 

davidmartos96:
In addition to the tests, the following logic is removed because at the oplog level, the Record values cannot be BigInt, as the oplog trigger serializes those values as string. Is the `deserialize` who checks the postgres type to turn it into the correct JS object.
Leaving the type check there might cause confusion in the future.

```js
if (typeof value === 'bigint') {
    return value.toString()
}
```

### DIFF
--- a/clients/typescript/src/satellite/oplog.ts
+++ b/clients/typescript/src/satellite/oplog.ts
@@ -306,9 +306,6 @@ function serialiseRow(row?: Rec): string {
         return '-Inf'
       }
     }
-    if (typeof value === 'bigint') {
-      return value.toString()
-    }
     return value
   })
 }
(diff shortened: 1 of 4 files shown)


kevin-dp (approved):
Indeed, that if check seems obsolete as `serialiseRow` is called only from within `fromTransaction` which in turn is called only on transactions received through the replication stream. Those do not contain `BigInt`s.

davidmartos96:
Great! 😄

kevin-dp:
Turns out the `BigInt` test was not obsolete. It's actually used when serialising records whose PK is a `BigInt`, cf. https://github.com/electric-sql/electric/pull/749. Found this thanks to this bug report: https://linear.app/electric-sql/issue/VAX-1421/using-bigint-for-the-primary-key-column-leads-to-type-errors

PR #700, opened 2023-11-27
chore (client): pin Zod to 3.21.1 to avoid code generator bug in client

fooware:
## Summary
As discussed on Discord with @kevin-dp, it appears that the dependencies for Zod can pull in a version that causes the generated client to contain invalid casts.

image

The issue in Zod + Prisma can be found here:
https://github.com/colinhacks/zod/issues/2184
https://github.com/chrishoermann/zod-prisma-types/issues/98

## Background

For us, the problem manifests itself by adding any kind of basic table relationship and generating a client. Indeed, it appears to also happen in this example code created by a third party:
https://github.com/KyleAMathews/vite-react-router-electric-sql-starter/tree/main

In both that example, and our internal code, it is enough to comment out any relationships and the generated code is fine, or rather doesn't contain the code blocks that are causing the issues.

image

Patching the packages lock file in both cases to resolve Zod to 3.21.1 (as suggested in the linked issue threads) fixed the issue for both our code and the third-party example.

image
image

## Temporary Workaround

Since we are using PNPM we could achieve the same result by adding a forced version override to our root package.json file:
```
  "pnpm": {
    "overrides": {
      "zod": "3.21.1"
    }
  }
```

Using that we could also experiment with different Zod versions, and indeed, going to 3.21.2 or higher reintroduced the problem.

## Suggested Fix

This PR pins the version of Zod to 3.21.1 which is the latest version that did not cause the issue. Hopefully, this will mean that new users will get a local package lock file that resolves Zod to 3.21.1 and a working client generator. That last part is a bit hard for me to verify locally though.

## Caveat

I have only updated the TypeScript Client and the Client Generator, but there are at least two places in this mono repo that explicitly reference higher versions of Zod:
- e2e/satellite_client - `^3.21.4`
- examples/linearlite - `^3.22.2`
- 
I was not comfortable downgrading those dependencies since I do not know _why_ they require a higher version.

### DIFF
--- a/.changeset/happy-lobsters-type.md
+++ b/.changeset/happy-lobsters-type.md
@@ -1,7 +1,5 @@
 ---
 "electric-sql": patch
-"@core/electric": patch
-"create-electric-app": patch
 "@electric-sql/prisma-generator": patch
 ---
 
--- a/.changeset/happy-lobsters-type.md
+++ b/.changeset/happy-lobsters-type.md
@@ -1,0 +1,8 @@
+---
+"electric-sql": patch
+"@core/electric": patch
+"create-electric-app": patch
+"@electric-sql/prisma-generator": patch
+---
+
+Pin Zod to version 3.21.1
--- a/clients/typescript/package.json
+++ b/clients/typescript/package.json
@@ -175,7 +175,7 @@
     "uuid": "^9.0.0",
     "walkjs": "^3.2.4",
     "ws": "^8.8.1",
-    "zod": "^3.20.2"
+    "zod": "3.21.1"
   },
   "devDependencies": {
     "@ikscodes/browser-env": "^1.0.0",
--- a/e2e/satellite_client/package.json
+++ b/e2e/satellite_client/package.json
@@ -21,7 +21,7 @@
     "electric-sql": "workspace:*",
     "jsonwebtoken": "^9.0.0",
     "uuid": "^9.0.0",
-    "zod": "^3.21.4"
+    "zod": "3.21.1"
   },
   "devDependencies": {
     "@electric-sql/prisma-generator": "workspace:*",
--- a/examples/linearlite/package.json
+++ b/examples/linearlite/package.json
@@ -67,7 +67,7 @@
     "uuid": "^9.0.0",
     "vite-plugin-svgr": "^3.2.0",
     "wa-sqlite": "rhashimoto/wa-sqlite#semver:^0.9.8",
-    "zod": "^3.22.2"
+    "zod": "3.21.1"
   },
   "devDependencies": {
     "@databases/pg": "^5.4.1",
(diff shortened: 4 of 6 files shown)


kevin-dp (commented):
Hi @fooware, thanks for your contribution!
The e2e tests and linearlite should indeed also be pinned to version 3.21.1 of Zod.
The fact that they are using a higher version of Zod is an accident.

Could you also generate a changeset for this PR please?
To do that, run `pnpm changeset` at the root of the monorepo.

fooware:
> Hi @fooware, thanks for your contribution! The e2e tests and linearlite should indeed also be pinned to version 3.21.1 of Zod. The fact that they are using a higher version of Zod is an accident.

Done.
 
> Could you also generate a changeset for this PR please? To do that, run `pnpm changeset` at the root of the monorepo.

Done, but please make sure I did it correctly.

Please note that I didn't rebuild any packages here. I assume that a build server handles that during release?

.changeset/happy-lobsters-type.md
```diff
@@ -0,0 +1,8 @@
+---
+"electric-sql": patch
+"@core/electric": patch
```
kevin-dp, line 3: The changeset is not correct. It patches all packages but your PR only touches the typescript client (i.e. `electric-sql`) and the generator (i.e. `@electric-sql/prisma-generator`). It does not touch Electric (`@core/electric`) neither does it touch the starter (`"create-electric-app": patch`).

So the changeset should look like this:
```
---
electric-sql": patch
@electric-sql/prisma-generator": patch
---

Pin Zod to version 3.21.1
```

You probably accidentally selected all packages instead of only the changed packages.

kevin-dp (changes requested):

.changeset/happy-lobsters-type.md
fooware: I assumed it only listed the ones it detected changes in so I accepted the defaults.

I’ll push an updated one tomorrow.

fooware: @kevin-dp, any feedback? I assume I shouldn’t resolve this thread?

kevin-dp: > @kevin-dp, any feedback? I assume I shouldn’t resolve this thread?

We will need a bit more time to see if we can provide a fix for this under the hood such that we don't need to pin the version of Zod to a specific version for all Electric projects. This will require some changes in the generator. I'm hoping to look into it tomorrow and ideally push a fix for it tomorrow. We will keep you posted here.

alco:
@kevin-dp @samwillis does https://github.com/electric-sql/electric/pull/727 address this issue?

samwillis:
Hey @fooware,

Thanks for the help with this, we fixed the issue in #727 with your tip off for the upstream bug report. We will be publishing a version with this fixed this week.

Issue #426, opened 2023-09-13
[VAX-1006] Possible race condition in `process.subscribe()`


js2702:
We sometimes are encountering an issue in the subscribe call. The subscription notifier doesn't have the subscription id after the client.subscribe() is finished. While debugging we noticed that when this happens it's because the subscription data message is received before client.subscribe() finishes. So when trying to read the promise from the subscription notifiers it's undefined.

The subscribe function
https://github.com/electric-sql/electric/blob/553a817770691e7afa3ac07562f443650ef87843/clients/typescript/src/satellite/process.ts#L374

This is what triggers the websocket communication that can end up deleting the subscriptionId before finishing.
https://github.com/electric-sql/electric/blob/553a817770691e7afa3ac07562f443650ef87843/clients/typescript/src/satellite/process.ts#L415

Here is where we get the undefined exception.
https://github.com/electric-sql/electric/blob/553a817770691e7afa3ac07562f443650ef87843/clients/typescript/src/satellite/process.ts#L431

Here is where the subscription is being removed.
https://github.com/electric-sql/electric/blob/553a817770691e7afa3ac07562f443650ef87843/clients/typescript/src/satellite/process.ts#L451

Would it make sense to check after client.subscribe() if the subscription exists? Or is that something that shouldn't happen?

When everything works fine, during client.subscribe() WebSocket only receives SatSubsResp. And after that the rest of the subscription messages are received (SatSubsDataBegin, ..., SatSubsDataEnd). When it fails this messages are received while the client.subscribe() is running.

<sub>[VAX-1006](https://linear.app/electric-sql/issue/VAX-1006/possible-race-condition-in-processsubscribe)</sub>

balegas:
We're working on some improvements to the protocol and will look into this after.

[https://github.com/electric-sql/electric/pull/423](https://github.com/electric-sql/electric/pull/423)

alco:
I'm currently investigating a very similarly looking problem, though I traced it down to different locations in the client code from the original report here. Instead of an undefined exception, I'm getting this:

```
encountered a subscription error: Received subscribe response for unknown subscription bfca735d-5037-4685-b547-74a4946118f1
connectivity state changed disconnected
client disconnected from server
Uncaught:
i [Error]: Received subscribe response for unknown subscription bfca735d-5037-4685-b547-74a4946118f1
[...]

PR #702, opened 2023-11-28
fix (client): fix race condition in process.subscribe that made the client crash

kevin-dp:
This PR fixes the race condition described in #426.
The problem was that `process.subscribe` was creating a promise, storing it, making an async call, fetching the promise and returning it. But, depending on the order in which messages are received from Electric, the async call may resolve after the subscription was already delivered and hence the promise had already been deleted. The fix consists of storing the promise in a local variable inside `process.subscribe` such that it doesn't need to fetch the promise after the async call.

### DIFF
--- a/.changeset/nine-boats-burn.md
+++ b/.changeset/nine-boats-burn.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Fix race condition in process.subscribe that made the client crash.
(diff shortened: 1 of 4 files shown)


magnetised (approved):
LGTM

balegas (approved):
Looks good. Before merging, please do a sanity check of starting the linearlite example and see if there are no unexpected exceptions on server restarts. I'm a bit paranoid with that :D

PR #703, opened 2023-11-28
chore (client): expose a close method on the ElectricNamespace to dispose its resources

kevin-dp:
This PR addresses the first task of #243.
As suggested, the `ElectricNamespace` (and thus also the `ElectricClient` as it extends the namespace) now exposes a `close` method that can be called by the user to dispose all resources used by the Electric client. This `close` method calls the `stop` on the global registry such that it also stops the client's satellite process.

### DIFF
--- a/.changeset/brave-colts-change.md
+++ b/.changeset/brave-colts-change.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Dispose listeners when Electric client is closed.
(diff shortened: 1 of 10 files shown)


balegas (approved):
Thanks for addressing this. A couple of suggestions:
- Maybe we should have a lifecycle test for electric testing that a process is registered after the electrification and deregistered after close. I'm not sure how much code is available to test this, but if not low-hanging I'm happy to skip
- Please double check that the things that are allocated for a process are all de-allocated and that we're not leaking any resources, or not de-registering from global stuff. This might have been the last thing, but let's do another pass before closing #243.

balegas (approved):
Thanks for looking into suggestions

PR #706, opened 2023-11-29
fix(electric): Rewrite PostgreSQL URI parser

alco:
Main improvements:
* support `postgres://` scheme
* default to using the username as the database name if the latter is missing from the URI (default behaviour of `psql`)

Fixes VAX-1264, VAX-1265.


components/electric/lib/electric/utils.ex
```diff
@@ -208,4 +208,167 @@ defmodule Electric.Utils do
...
+    ]
+    |> add_replication_param(query_params["replication"])
+    |> add_ssl_param(query_params["sslmode"])
+    |> Enum.reject(fn {_key, val} -> is_nil(val) end)
+  end
+
+  defp assert_valid_scheme!(scheme) when scheme in ["postgres", "postgresql"], do: :ok
+
+  defp assert_valid_scheme!(_scheme) do
+    raise "Invalid scheme in DATABASE_URL"
```
magnetised, line 322: might be useful to show the invalid protocol in the error, i.e. `"Invalid scheme in DATABASE_URL: #{scheme}"` much like you've done for the other validations

magnetised (approved):
lovely stuff. nice to lose a dependency right.

components/electric/lib/electric/utils.ex
alco: Yup.

PR #710, opened 2023-11-30
fix (generator): let generated client import types using TypeScript's import type

### LINKED ISSUE #707: Svelte needs 'import type'
When using Svelte I need to change the line in the generated client 
`import { TableSchema, DbSchema, Relation, ElectricClient, HKT } from 'electric-sql/client/model'`
to 
`import { type TableSchema, DbSchema, Relation, ElectricClient, type HKT } from 'electric-sql/client/model'`
in index.ts

It seems that Svelte can't resolve typescript type imports automatically:
https://github.com/sveltejs/svelte-preprocess/issues/206

kevin-dp:
This PR fixes #707 by explicitly importing types using `import type` instead of just `import`.

### DIFF
--- a/.changeset/tasty-lizards-boil.md
+++ b/.changeset/tasty-lizards-boil.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/prisma-generator": patch
+---
+
+Import types using import type in generated Electric client.
--- a/generator/src/functions/writeSingleFileImportStatements.ts
+++ b/generator/src/functions/writeSingleFileImportStatements.ts
@@ -19,7 +19,7 @@ export const writeSingleFileImportStatements: WriteStatements = (
   }
 
   writeImport(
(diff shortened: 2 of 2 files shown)


thruflo (approved):
Great 👍 

Can I just sanity check: are there any other places in the code where we should fix this to make imports work for a Svelte developer?

I.e.: if they're importing other types exposed by the main typescript client library?

kevin-dp:
> Great 👍
> 
> Can I just sanity check: are there any other places in the code where we should fix this to make imports work for a Svelte developer?
> 
> I.e.: if they're importing other types exposed by the main typescript client library?

This is only a problem in the project that is using Svelte (not internally in our `electric-sql` package). Therefore, I believe those other places are under the control of the application developer as those imports are in their own files. So it is the developer's responsibility to import them using `import type`.

PR #721, opened 2023-12-04
chore(docs): New guides on deploying to DigitalOcean

alco:
Preview: https://deploy-preview-77--electric-sql-website.netlify.app/docs/integrations/deployment/digital-ocean


thruflo (approved):
Brilliant stuff, thank you for this!

One note is that the link to a readme to then click the one click button is a somewhat unusual way of rendering and clicking a button. Is there a reason for it? Could we not just embed the button?

Another minor note is that I see https://github.com/electric-sql/website/pull/77 but I don't see the DO icon being added to static. We need it for the listing page.

Then my main feedback, which I don't see as blocking merge, and aside from probably refactoring this in future to recipes and more normative doc page (we can do that as content work when we get there) is that this doesn't cover Postgres or static site hosting.

The great potential of DO is to provide hosting for all three. Scenarios:

1. can you create a DO PG when you create the app? If so, can we create a one-click button with and a 1-click button without PG and then with PG can we in any way auto-configure the DATABASE_URL env var?
2. can you configure the Docker droplet to deploy the docker image *and* a PG? And can you again auto set any env vars? If not, we definitely can create a marketplace droplet that does both and I suspect has the DATABASE_URL available automatically

balegas:
Closes VAX-1314

alco:
@thruflo Thanks for the feedback, really on point!

> One note is that the link to a readme to then click the one click button is a somewhat unusual way of rendering and clicking a button. Is there a reason for it? Could we not just embed the button?

That's just a POC to have the button displayed somewhere. Should I put the button right inside the relevant doc section as a start?

> Another minor note is that I see [electric-sql/website#77](https://github.com/electric-sql/website/pull/77) but I don't see the DO icon being added to static. We need it for the listing page.

Good catch! I'll fix that.

>     1. can you create a DO PG when you create the app? If so, can we create a one-click button with and a 1-click button without PG and then with PG can we in any way auto-configure the DATABASE_URL env var?

It's possible to add a managed DB as part of configuring a DO App, so we can also preconfigure it for the 1-click button. One caveat is that it creates a dev DB this way which can be upgraded to a production setup after the app is created. Adding a managed database as part of the app setup process will inject a DATABASE_URL variable into the web component of the app automatically, I believe.

>     2. can you configure the Docker droplet to deploy the docker image _and_ a PG? And can you again auto set any env vars? If not, we definitely can create a marketplace droplet that does both and I suspect has the DATABASE_URL available automatically

From what I'm seeing, it's not possible to create a Droplet and a managed DB at the same time. We'll need to create a custom Marketplace image for this. DATABASE_URL will be injected into the droplet automatically if the developer chooses to created the DB together with the droplet.

---

I'm not too keen on the idea of hosting the static website along with Electric on DO. On the App Platform the two will be competing for the same HTTP port, unless they are two separate apps. In Droplet's case, there won't be a builtin mechanism to publish new builds of the static website. Maybe there's something we could do by marrying DO's Spaces and CDN but it's not something we can bundle with Droplet creation and IMO it won't be nearly as convenient as using Netlify or Vercel.

thruflo:
> Should I put the button right inside the relevant doc section as a start?

Yup :)

> It's possible to add a managed DB as part of configuring a DO App, so we can also preconfigure it for the 1-click button. 

Awesome -- let's have two buttons. One with, one without.

> From what I'm seeing, it's not possible to create a Droplet and a managed DB at the same time. We'll need to create a custom Marketplace image for this. DATABASE_URL will be injected into the droplet automatically if the developer chooses to created the DB together with the droplet.

Makes me very happy because then there's a clear reason why Electric specific droplet(s). Let's do this!

thruflo:
> I'm not too keen on the idea of hosting the static website along with Electric on DO. 

Yes, have you looked at their static site hosting offering with app platform? Maybe this is a third separate button? I would *love* for that to auto-populate the env vars needed for build. But maybe that's not possible.

thruflo:
> We cannot run both the static site and Electric in the same app, so no "full-stack button" targeting app platform.
> 
> I tried setting up a stand-alone next.js sampe as a DO app, the experience is similar to that of Heroku.

Do DO provide a `git|cli push -> build app -> deploy static site` option? If so, perhaps we could add a section to the doc that shows how to do this on DO, once you have the backend deployed?

If not, then let's skip this idea.

alco:
> Do DO provide a `git|cli push -> build app -> deploy static site` option?

Yes.

alco:
@thruflo I've pushed a new commit that adds a few corrections to the draft version you reviewed. The "Deploy to DO" button references the new public repo I created at [electric-sql/deploy-to-do](https://github.com/electric-sql/deploy-to-do/). We can have as many variants of the button as we want, each one can be defined as a separate branch in that repo.

Regarding your proposal to include a managed database in the 1-click install, I couldn't get that to work. Currently awaiting a resolution from DO's support team.

Regarding the static website hosting on DO App Platform, I've created VAX-1422 to track that.

PR #728, opened 2023-12-06
fix (generator): escape client generator path on Windows

kevin-dp:
This PR escapes also the output path for the Prisma client generator in the intermediate Prisma schema when running on Windows.


magnetised (approved):
👍🏻

PR #735, opened 2023-12-07
feat(client): New "npx electric-sql proxy-tunnel" command

samwillis:
@alco this builds on your WIP [alco/pg-proxy-over-websocket](https://github.com/electric-sql/electric/tree/alco/pg-proxy-over-websocket) branch 

To connect to the service, and create a local proxy tunnel:
```sh
npx electric-sql proxy-tunnel --service http://something --local-port 65431
```

then to run migrations:
```sh
yarn run -s pg-migrations apply --database postgres://postgres:proxy_password@localhost:65431 --directory ./db/migrations
```

and the generator:
```sh
npx electric-sql generate --service http://localhost:5133 --proxy postgresql://prisma:proxy_password@localhost:65431/electric
```

### DIFF
--- a/clients/typescript/src/cli/tunnel/handler.ts
+++ b/clients/typescript/src/cli/tunnel/handler.ts
@@ -29,6 +29,10 @@ export async function handleTunnel(...args: string[]) {
  * @returns The parsed arguments.
  */
 function parseTunnelArgs(args: string[]) {
+  if (args.length % 2 !== 0 || args.length > 4) {
+    throw new Error('Invalid number of arguments')
+  }
+
   const opts: Partial<TunnelOptions> = {}
   for (let i = 0; i < args.length; i++) {
     const arg = args[i]
--- a/clients/typescript/src/cli/tunnel/handler.ts
+++ b/clients/typescript/src/cli/tunnel/handler.ts
@@ -1,0 +1,54 @@
+import { tunnel, defaultOptions, type TunnelOptions } from './tunnel'
+
+/**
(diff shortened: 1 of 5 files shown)


alco:
@samwillis I have tested this on my machine and it works flawlessly. I have only one concern: it takes upwards of 30 seconds before it starts listening on the local port. Am I running it wrong when I execute `npx --no-install electric-sql proxy-tunnel` inside `clients/typescript`?

Code-wise, is there anything blocking this from getting merged into `main`?

samwillis:
No need for the `--no-install`, if the user has a project with the `electric-sql` package installed `npx electric-sql proxy-tunnel` will use their installed version when run within the project. My guess is that `--no-install` is making it slow for you.

All ready to merge when #738 is 😄

clients/typescript/src/cli/tunnel/handler.ts
```diff
@@ -0,0 +1,54 @@
...
+function parseTunnelArgs(args: string[]) {
+  const opts: Partial<TunnelOptions> = {}
+  for (let i = 0; i < args.length; i++) {
+    const arg = args[i]
+    switch (arg) {
+      case '--service':
+        opts.serviceUrl = args[++i]
+        break
+      case '--local-port':
+        opts.localPort = parseInt(args[++i])
```
kevin-dp, line 40: The argument at index `i+1` might be missing. We probably want to throw an error if there are missing arguments.

kevin-dp (approved):

PR #738, opened 2023-12-08
feat(electric): Allow the migrations proxy to accept WebSocket connections

alco:
Closes VAX-1417.

This change makes https://github.com/electric-sql/electric/pull/735 possible.


components/electric/lib/electric/postgres/proxy/websocket_server.ex
```diff
@@ -0,0 +1,38 @@
...
+
+  It opens a local TCP connection to the proxy and shuttles data between it and the client's WebSocket connection.
+  """
+
+  @behaviour WebSock
+
+  @impl WebSock
+  def init(_opts) do
+    # TODO(alco): look up Proxy listener's port at runtime
+    proxy_port = 65432
```
magnetised, line 13: assuming that this is the other end of the hard-coded port number in proxy.ex:117? 

could we make this value a function call or something to make the link clearer

magnetised (approved):
crazy stuff! horrifying and yet genius.

components/electric/lib/electric/postgres/proxy/websocket_server.ex
magnetised: i actually don't mind the hard-coded port particularly, it's more that you've got the same "magic" number in two places. if that magic number were hard-coded in a single place it would just make it clearer that the two port numbers are joined.

if you do feel that it should be configurable, the `http:65432` version feels the most natural

alco: I've decided to go with the `PG_PROXY_PORT=http:65432`. I think it makes both the documentation and the implementation cleaner. No more hard-coded proxy port numbers.

PR #745, opened 2023-12-11
Update Types doc to reference main branch

robbwdoering:
This link just threw me for a *huge* loop when I thought it meant I'd have to abandon Electric after spending awhile setting it up, since it references a significantly outdated version of this file. Not sure if this is how you want to handle it though, or just moving up the commit pointer.

Thanks for the great work, excited to try it out!

### DIFF
--- a/docs/usage/data-modelling/types.md
+++ b/docs/usage/data-modelling/types.md
@@ -46,7 +46,7 @@ You are responsible for ensuring the uniqueness of your primary keys. If you som
 - `jsonb`
 - `uuid`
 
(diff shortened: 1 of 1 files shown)


thruflo (approved):
Nice catch, thank you! And sorry for the confusion.

alco:
@robbwdoering Thanks, good catch!

Referencing `main` has proven to be problematic as well, since support for new types in published releases lags behind `main`. I think we'll be better off removing that reference altogether and we'll simply maintain the list of types in the docs up-to-date with the latest _published_ version of Electric.

Issue #740, opened 2023-12-08
PG_PROXY_URL not honored. Should be replaced with ELECTRIC_PROXY_URL


jljorgenson18:
It looks like PG_PROXY_URL is not used at all. It looks like internally electric is actually expecting `ELECTRIC_PROXY_URL` 

Were the docs just not updated or should it be set to Electric Proxy URL?

<img width="981" alt="Captura de Tela 2023-12-08 às 5 31 08 PM" src="https://github.com/electric-sql/electric/assets/5751354/adc39f02-9531-43aa-9b44-076d9f2e6325">

samwillis:
Hey, @jljorgenson18 

Thanks for the report. Yes, this is a documentation error, the correct env var is `ELECTRIC_PROXY_URL`

alco:
Fixed in https://github.com/electric-sql/electric/pull/747.

PR #747, opened 2023-12-11
chore(docs): Update docs to account for all the changes introduced in 0.8.x

alco:
Preview: https://deploy-preview-80--electric-sql-website.netlify.app/docs/api/service#configuration-options

Fixes VAX-1405, VAX-1413, https://github.com/electric-sql/electric/issues/740.


docs/api/clients/typescript.md
```diff
@@ -14,6 +14,83 @@ projects have issues, and users can comment on issues
 (in which case they are said to be the author of that comment).
 The data model for this application can be found on the <DocPageLink path="api/clients/typescript" /> page.
 
+## Configuration
```
samwillis, line 17: This new section is great! I've recorded it to be after "Instantiation" in my downstream PR with a link down to it from a tweeted beginning of the Instantiation section.

samwillis (approved):

docs/usage/data-modelling/validation.md
```diff
@@ -1,16 +1,44 @@
...
 ---
 
 :::caution Work in progress
-User input validation is not currently supported.
+User input validation is under active development. As we progress we will update this documentation.
 :::
 
+## Immutable Primary Keys
+
+Electric rejects updates to a table's primary keys.
```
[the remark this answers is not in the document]
alco: Done :+1:

docs/usage/auth/index.md
```diff
@@ -10,7 +10,7 @@ ElectricSQL provides a JSON Web Token based authentication mechanism for [client
 
 ## Overview
 
-Your application must generate a [valid JWT](./token.md). This must have at least a `user_id` claim, which should be a non-empty string matching the primary key UUID of the authenticated user.
+Your application must generate a [valid JWT](./token.md). This must have at least a `sub` claim (formerly `user_id`), which should be a non-empty string matching the primary key UUID of the authenticated user.
```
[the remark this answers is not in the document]
alco: Good point. I hate it that the claim is referenced by so many places. I've chosen to add a note on the Token format page.

By the way, you didn't include a changeset in the PR that renamed `user_id` to `sub`. Would you mind editing the changelog in `components/electric` by hand to add an entry for it under the 0.8.0 release.

docs/api/service.md
```diff
@@ -31,17 +31,48 @@ The Electric application is configured using environment variables.
...
+| `PG_PROXY_PORT`<p>&nbsp;&nbsp;(`65432`)</p>        | <p>Port number for connections to the [Postgres migrations proxy](../usage/data-modelling/migrations#migrations-proxy).</p><p>If you have Electric deployed behind a restrictive firewall that only allows HTTP/HTTPS connections, you can add the `http:` prefix to the port number. This will enable the tunelling mode in which the migrations proxy will accept WebSocket connections from the proxy tunnel (`npx electric-sql proxy-tunnel`).</p> |
+| `PG_PROXY_PASSWORD`                                | Password to use when connecting to the Postgres proxy via `psql` or any other Postgres client.                                                                                                                                                                                                                                                                                                                                                         |
+
+### Write-to-PG mode
```
thruflo, line 44: I'd suggest a line at the top of this section summarising the sections below. So e.g.:

```
Electric writes data to Postgres using one of two modes: logical replication or direct writes. Which mode you choose to use affects your networking config and the database user permissions required.

::: note
Logical replication was the only mode available prior to v0.8, when we introduce the new direct writes mode.
:::
```

docs/api/service.md
```diff
@@ -31,17 +31,48 @@ The Electric application is configured using environment variables.
...
+
+| Variable                                                              | Description                                                                                                                                                    |
+| --------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ELECTRIC_WRITE_TO_PG_MODE`<p>&nbsp;&nbsp;(`logical_replication`)</p> | <p>The mode to use when syncing client updates to Postgres.</p><p>Allowed values: `logical_replication`, `direct_writes`.</p>                                  |
```
thruflo, line 48: Obviously I'm late to the party on this comment but reading the docs, I really want to see a pathway to direct writes being the default mode. I get keeping logical replication as default as it's the previous behaviour. However, could we perhaps indicate that we plan to switch this default value to direct writes in future?

Because direct writes is all gain no pain from a configuration point of view here, i.e.: why would we default to the option that's more complex to setup and requires higher permissions?

docs/api/service.md
```diff
@@ -31,17 +31,48 @@ The Electric application is configured using environment variables.
...
+         |-----LOGICAL_PUBLISHER_HOST--->|
+```
+
+As a result, you must make sure (in terms of networking / firewalls) not only that Postgres is reachable from Electric but also that Electric is reachable from Postgres. And Electric must know its own address, in order to provide it to Postgres when setting up the logical replication publication that allows writes to be replicated into Postgres.
+
+#### Direct writes mode
+
+The `direct_writes` mode works by establishing a regular connection to Postgres and applying client updates as DML statements. When launched in this mode, Electric does not require any elevated privileges from its Postgres role beyond the `REPLICATION` role attribute. Moreover, the `LOGICAL_PUBLISHER_HOST` and `LOGICAL_PUBLISHER_PORT` options are not used by this mode.
 
 ### Authentication
```
thruflo, line 77: Do we not want to expose an env var for specifying the sub/user_id key? Seems like a good way to future proof / expand compatibility.

docs/usage/auth/index.md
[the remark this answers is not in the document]
thruflo: See my comment above re: making this env var configurable.

thruflo (approved):

docs/api/service.md
samwillis: We disscussed doing it this way first, then exposing a configuration option later.

Should probably looks at sessions at the same time.

docs/api/service.md
alco: The direct writes mode has seen little testing. I caught one issue related to it on CI by chance (VAX-1424). There may be more.

I'm also super pro-simplifying configuration but rewriting too much stuff at once is somewhat error-prone. Let's keep iterating instead of trying to get it perfect all at once.

docs/api/service.md
alco: I've created VAX-1430 so that we don't forget about this.

docs/api/service.md
alco: Added :+1:

PR #749, opened 2023-12-12
fix (client): fix bug with primary keys that are BigInts

kevin-dp:
This PR fixes VAX-1421 where a serialisation error would occur when trying to insert a value in a table whose primary key is of type `BigInt` (i.e. `INT8`). Under this corner case, `primaryKeyToStr` is called on the `BigInt` value (as it is the value of the PK) which in turn calls `serialiseRow` which was not handling `BigInt`s. I also added a unit test that reproduced the error, which is now gone thanks to the fix.


samwillis (approved):
👍

PR #758, opened 2023-12-14
fix(client): Ensure temporary migration folder deletion

msfstef:
When client generation failed, the process exited before having a chance to delete the temporary migrations folder, leaving you with an ever increasing number of failed temporary migration folders.

They were also tracked on git by default so I added them to the starter template gitignore - not the most maintainable solution but helps keep things neat for the starter experience.


kevin-dp (commented):
Implementation looks good!
Though i'm not sure we want to delete the folder if something goes wrong.
Often times this folder is very useful to debug what went wrong as it gives us access to the intermediate Prisma schema for which client generation fails. Perhaps also make that configurable? e.g. by default we always remove the temporary directory but if you enable the flag we don't delete it on error.

kevin-dp:
Btw, you should generate a changeset for each PR, to do so run `pnpm changeset` at the root of the mono repo and follow the steps of the interactive changeset tool.

msfstef:
Added a generic ```--debug``` flag to the generator which can be reused for maintaining more info - figured I'd move around some things to using switch statements for clarity.

Added a changeset, tried to include all changes in it so it's multiline, hope that's ok!

clients/typescript/src/cli/migrations/migrate.ts
```diff
@@ -211,11 +214,20 @@ async function _generate(opts: Omit<GeneratorOptions, 'watch'>) {
     await buildMigrations(migrationsFolder, migrationsFile)
     console.log('Successfully built migrations')
   } catch (e: any) {
+    generationFailed = true
     console.error('generate command failed: ' + e)
-    process.exit(1)
+
+    // Rethrow error - finally statement still runs
+    if (!opts.exitOnError) throw e
```
kevin-dp, line 221: Why is this if test needed? If `opts.exitOnError` is `true` and we would still `throw e` then we would still get into the `finally` block and we would call `process.exit(1)` on L230.

clients/typescript/test/cli/migrations/migrate.test.ts
```diff
@@ -108,9 +113,71 @@ model Model {
...
+    // should fail generaton
+    t.assert(await failedGenerate(false))
+
+    // should clean up temporary folders
+    t.assert((await findMigrationFolder()) === null)
+  }
+)
+
+test.serial(
+  'migrator should retain temporary folder on falure in debug mode',
```
kevin-dp, line 171: typo in this test name `falure` -> `failure`

kevin-dp (approved):
Your changes look good!
Left 2 minor comments but when can merge after that.

kevin-dp:
I see that you merged main into your branch to fix some conflicts.
We usually don't do that but solve conflicts by rebasing on top of main to keep the history linear.

clients/typescript/src/cli/migrations/migrate.ts
msfstef: good point - definitely not needed!

samwillis:
@msfstef I would hold off on modifying the commands too much as #664 refactors them a lot and we will need to add any new arguments there.

PR #759, opened 2023-12-14
chore (client): improved interface for the generic adapter

kevin-dp:
This PR improves the generic database adapter to require `_query` and `_run` methods instead of an `exec` method.

### DIFF
--- a/clients/typescript/src/drivers/generic/adapter.ts
+++ b/clients/typescript/src/drivers/generic/adapter.ts
@@ -5,7 +5,6 @@ import {
   Transaction as Tx,
 } from '../../electric/adapter'
 import { Row, Statement } from '../../util'
-import { isInsertUpdateOrDeleteStatement } from '../../util/statements'
 import { Mutex } from 'async-mutex'
 import { AnyDatabase } from '..'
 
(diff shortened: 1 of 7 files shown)


clients/typescript/src/drivers/generic/adapter.ts
```diff
@@ -28,14 +27,19 @@ abstract class DatabaseAdapter
...
 
   /**
-   * @returns The number of rows modified by the last insert/update/delete query.
+   * Runs a single SQL query against the DB.
+   * The query should be read-only.
+   * @param stmt The SQL statement to execute
+   * @returns The rows read by the query.
    */
-  abstract getRowsModified(): number
+  abstract _query(stmt: Statement): Promise<Row[]>
```
icehaunter, line 42: Does this assume no `RETURNING` clauses ever?

samwillis (approved):
@kevin-dp Looks good, I'm all for explicit readonly and write paths.

One thought, we merged #805 last week that introduced `rawQuery` and `unsafeExec` as the front end *enforced* read only and write apis. It may be a good idea to stick to the same naming, so `_exec` rather than `_run`.

clients/typescript/src/drivers/generic/adapter.ts
kevin-dp: If you're using a `RETURNING` clause you have to use `_query`. My understanding is that `query` is not only for read-only queries but is also for modification statements that return rows use the `RETURNING` clause. That's also what we do internally in the DAL (e.g. the DAL's `create` method does an `INSERT INTO` with a `RETURNING` clause and executes it using `query`).

kevin-dp:
> One thought, we merged #805 last week that introduced `rawQuery` and `unsafeExec` as the front end _enforced_ read only and write apis. It may be a good idea to stick to the same naming, so `_exec` rather than `_run`.

This abstract class `DatabaseAdapter` is extended by developers to add support for a new SQLite driver.
These drivers typically provide `run` and `query` methods so i think that is a bit more appropriate here.
I think it's more important to stick to the naming of the drivers here (rather than DAL naming) because this is concerned with adding driver support (not related to the DAL).

clients/typescript/src/drivers/generic/adapter.ts
icehaunter: Yeah, that's the expected behaviour. If that's the case, remove `read-only` from the doc here pls :pray:

icehaunter (approved):

PR #760, opened 2023-12-14
Adapt cap-sqlite mock.ts to new generic adapter

gregzo:
* Removed todo comment RE return type of cap-sqlite `query`, the result type is an array of objects as expected.
* Adapted `mock.ts` to the new requirements

My tests pass, but am getting 2 other failed tests which are probably tied to env/update issues?

`Error: Cannot find module '/Users/gregoriozanon/Documents/Dev/electric/clients/typescript/node_modules/@electric-sql/prisma-generator/dist/generator.js'. Please verify that the package.json has a valid "main" entry`

Makes those tests fail:
```
✖ test/cli/migrations/handler.test.ts exited with a non-zero exit code: 1
✖ test/cli/migrations/migrate.test.ts exited with a non-zero exit code: 1
```

### DIFF
--- a/clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
+++ b/clients/typescript/src/drivers/capacitor-sqlite/adapter.ts
@@ -21,7 +21,6 @@ export class DatabaseAdapter extends GenericDatabaseAdapter {
       wrapInTransaction
     )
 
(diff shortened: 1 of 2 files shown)


kevin-dp (approved):

gregzo:
Tested locally with our app, all good 👍

PR #761, opened 2023-12-14
chore(electric): Enable Electric to connect to the database over IPv6 or IPv4 without explicit configuration

alco:
At a high level, this change adds a "preflight" step to PostgresConnectorMng that allows us to "post-process" the Postgres connector config at run time before handing it out to any other place in the code that calls `Client.connect()` or `Client.with_conn()`.

The first run-time post-processing being added here is performing a DNS lookup for the database host. We do this prior to establishing an `epgsql` connection, in order to include the right connection options, namely, `tcp_opts: [:inet6]`, when needed.

This allows us to switch the default value of `DATABASE_USE_IPV6` to `true` and remove the need for our users to care about which protocol their database of choice is available over. Previously, one would have to set `DATABASE_USE_IPV6=true` when connecting to Fly Postgres. Now the default value should work for all users all of the time.


magnetised (approved):
beautiful. bet you're bored of writing `connector_config` now.

PR #763, opened 2023-12-15
docs: add supabase event sourcing page.

thruflo:
Tweak docs to align with new Secure Transactions post. And split out the Supabase event sourcing info to a page in the event sourcing section.


samwillis (approved):
👍

PR #762, opened 2023-12-14
chore(electric): Format known errors for easier reading and include hints to guide users towards resolution

alco:
There is a set of Electric's failure modes that we know of in advance. Some have to do with improper configuration, others are specific to a given runtime environment or the database in use.

To help users resolve those known issues on their own, this pull request adds a new module for outputting nicely-formatted errors with additional context and pointers for the users. It also adds explicit handling of a bunch of known errors.

A few examples:

### Server and client SSL configuration mismatch

```
23:19:16.864 pid=<0.380.0> [info] Starting ElectricSQL 0.8.1-12-g363f631 in logical_replication mode.
23:19:16.873 pid=<0.384.0> [debug] Loading flag string from env ELECTRIC_FEATURES
23:19:16.875 pid=<0.384.0> [debug] Got feature flag configuration %{proxy_ddlx_assign: false, proxy_ddlx_grant: false, proxy_ddlx_revoke: false, proxy_ddlx_unassign: false} with default value: false
23:19:16.910 pid=<0.381.0> [info] Running Electric.Plug.Router with Bandit 1.0.0 at :::5133 (http)
23:19:16.914 pid=<0.999.0> origin=postgres_1 [debug] Attempting to initialize postgres_1: postgres@localhost:54321
23:19:16.918 pid=<0.999.0> origin=postgres_1 [info] connect: %{database: ~c"electric", host: ~c"localhost", ipv6: false, port: 54321, ssl: false, timeout: 5000, username: ~c"postgres"}
Interactive Elixir (1.15.2) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> 23:19:16.945 pid=<0.999.0> origin=postgres_1 [error] Initialization of Postgres state failed with reason: {:error, :invalid_authorization_specification}.
▓ ┌────────────────────┐
▓ │  CONNECTION ERROR  │
▓ ┕━━━━━━━━━━━━━━━━━━━━┙
▓ 
▓ Failed to initialize Postgres state:
▓   {:error, :invalid_authorization_specification}
▓
▓ It appears that the database is configured to only accept connections
▓ encrypted with SSL. Make sure you configure Electric with DATABASE_REQUIRE_SSL=true.
23:19:16.946 pid=<0.999.0> origin=postgres_1 [info] schedule retry: 2000
```

### Postgres not configured with wal_level=logical
<details>
  <summary>Before</summary>

The following error and stacktrace would be printed out 3 times in succession due to Elixir's supervisor restarting the process a few times:

```
[...]

Issue #543, opened 2023-10-10
can't have healthcheck when ssl is activate


ulricden:
i try to start a electric-sql server on AWS connect to an RDS Postgresql 15.3

when i try to launch the the service without ssl i have error : 
 ```
 [error] initialization for postgresql failed with reason: {:error, :invalid_authorization_specification}"
 [info] connect: %{database: ~c""database_web"", host: ~c""rds.web.com"", port: 5432, ssl: false, timeout: 5000, username: ~c""username""}"
 [debug] Attempting to initialize postgres_1: username@rds.web.com:5432"
 [info] Running Electric.Plug.Router with Bandit 1.0.0-pre.14 at 0.0.0.0:5133 (http)"
 [notice] :alarm_handler: {:set, {:system_memory_high_watermark, []}}"
```

so i add DATABASE_REQUIRE_SSL = TRUE the i had a new error :

```
"(electric 0.6.4) lib/electric/replication/postgres_manager.ex:92: Electric.Replication.PostgresConnectorMng.handle_continue/2"
"** (MatchError) no match of right hand side value: {:error, {{:shutdown, {:failed_to_start_child, :postgres_producer, {:bad_return_value, {:error, {:error, :error, ""55006"", :object_in_use, ""replication slot \""electric_replication_out_database_web\"" is active for PID 32436"", [file: ""slot.c"", line: ""518"", routine: ""ReplicationSlotAcquire"", severity: ""ERROR""]}}}}}, {:child, :undefined, :sup, {Electric.Replication.PostgresConnectorSup, :start_link, [[origin: ""postgres_1"", producer: Electric.Replication.Postgres.LogicalReplicationProducer, connection: [replication: ""database"", ssl: true, host: ""rds.web.com"", database: ""database_web"", port: 5432, username: ""username"", password: ""*****"", timeout: 5000], replication: [electric_connection: [host: ""electric-tcp.web.com"", port: 5433, dbname: ""electric"", connect_timeout: 5000]]]]}, :temporary, false, :infinity, :supervisor, [Electric.Replication.PostgresConnectorSup]}}}"
"11:39:21.217 pid=<0.2811.0> origin=postgres_1 [error] GenServer #PID<0.2811.0> terminating"
"11:39:21.217 pid=<0.2822.0> [debug] Terminating idle db connection #PID<0.2825.0>"
[...]

PR #767, opened 2023-12-17
chore(electric): Flip the default value of `DATABASE_REQUIRE_SSL` to `true`

alco:
This is a better default for any environment other than, perhaps, dev.

Until now, connecting to a Postgres that was configured to only accept SSL connection would result in a cryptic `invalid_authorization_specification` error (FWIW the same error may be logged when Postgres is configured to only accept unencrypted connection, but this is a rare edge case). That was the case with Crunchy Bridge and [AWS RDS](https://github.com/electric-sql/electric/issues/543#issuecomment-1801897232).

With this new default, more hosted databases should be supported out of the box. One prominent exception is Fly Postgres (at least prior to its takeover by Supabase): since Fly machine all connect to the same encrypted private network, there is no benefit of an additional encryption for database connections on top of that. For cases like this, we need to print a friendly error message instructing the developer to set `DATABASE_REQUIRE_SSL=false` in their Electric configuration. This will be addressed in https://github.com/electric-sql/electric/pull/762.

Note that Electric also needs to be able to verify the database host's authenticity to prevent man-in-the-middle attacks. I've filed VAX-1446 to address this at a later time.

Also, the added burden of setting `DATABASE_REQUIRE_SSL=false` in dev is not lost on me. I'm implementing support for `.env` files in https://github.com/electric-sql/electric/pull/770. That will remove some configuration-related woes in dev.


magnetised (approved):
I'm sure you've thought about this but isn't there an "optimistic" version of this setting that will try ssl then fallback if it doesn't work? wouldn't that work everywhere? in this pr you refuse use of the `sslmode` param but isn't `sslmode=prefer` exactly the behaviour we want? i.e. the client sends an `SSLRequest` message and the server responds with either `S` or `N`

alco:
@magnetised This PR bundles a couple of independent changes:

- enforcement of SSL connections by default. What you're saying about `sslmode=prefer` can be achieved by setting `DATABASE_REQUIRE_SSL=false`. But I don't think it should be the default for production deployments. So far we've roughly followed the "secure by default" approach with `AUTH_MODE` being set to `secure` by default.

  Of course, it can be argued, that a database is probably always hidden from the Internet in a private network in production deployments, but I'm not sure that's valid anymore. Hosted PGs like Supabase and Crunchy Bridge make it easy to connect to them over the public Internet.

- rejection of `DATABASE_URL` if it has the `sslmode` query param. This is because we don't intend to support all of the `sslmode` values that `psql` supports. I remember originally asking for this option to be supported but I've reversed my position on this topic after doing some research into the use of SSL with Postgres and how the [pg_hba.conf file](https://www.postgresql.org/docs/current/auth-pg-hba-conf.html) can be used to restrict incoming connections based on host and encryption rules.

magnetised:
@alco thanks for the clarification.

alco:
FYI I had to implement SSL support in the proxy because it wouldn't be able to establish an upstream connection when the database was configured to require SSL. I'm baffled at the fact that we haven't noticed this until now.

PR #770, opened 2023-12-19
chore(electric): Implement support for .env files in dev and test environments

alco:
This change simplifies Electric's configuration for different environments by allowing the use of `dev.env` and `test.env` files to specify environment-specific variables. Purists may complain that it should be `.env.dev` and `.env.test` instead. I don't have a strong preference here but to me using the environment name before the extension feels more natural.

All user configuration parsing is now consolidated in the `config/runtime.exs` file. `runtime.dev.exs` and `runtime.test.exs` can still be used, if needed, for configuration options that aren't exposed as environment variables.

Error handling for required configuration options will be further improved in https://github.com/electric-sql/electric/pull/773.


magnetised (approved):
great stuf... I think you're right that `.test.env`or `.env.test` are better. basically not having config files things polluting the file tree is good.

or we follow some xdg-spec like thing and add a `.config` directory with `.config/test.env`...

PR #774, opened 2023-12-21
fix(client): simplified SQLite query table name parsing and removed deprecated dependency

msfstef:
I noticed that some queries were working with the ```db.raw``` API but not the ```db.liveRaw``` one, because the latter tried to parse the table names from the query using a deprecated [SQLite AST parser](https://github.com/codeschool/sqlite-parser) which threw errors whenever using queries that SQLite has been supporting for the past 5 years (e.g. windowed queries).

I could not find an up to date, reliable replacement for this parser, but figured since the task is only really to parse the table names for keeping the query reactive, a simple RegExp might do the trick. Given sufficient testing I think this might be a simpler solution with fewer dependencies.

See test cases for examples of queries that would fail to be parsed but that are supported by the SQLite drivers:
```sqlite
SELECT timestamp, value, 
avg(value) OVER (ORDER BY timestamp ROWS BETWEEN 3 PRECEDING AND 3 FOLLOWING) as moving_average
FROM monitoring ORDER BY timestamp
```

Unless we unblock this any kind of efficient windowed query leveraging the power of the SQLite drivers is blocked, so many analytics use cases will run into issues.


samwillis:
This looks sensible to me, and it's removing a depreciated dependancy and enabling important SQL support.

I think the worst case is a false positive on a table that hasn't actually been updated, but the impact of that is tiny.

@balegas you should take a quick look too.

KyleAMathews:
Just ran into this as well 👍

clients/typescript/src/util/parser.ts
```diff
@@ -39,65 +36,31 @@ export const parseTableNames = (
...
-  const statement = ast.statement[0]
-  if (statement.type !== 'statement' || statement.variant !== 'select') {
-    throw 'Query must be a valid SELECT statement.'
+  // NOTE(msfstef): using an SQLite parser to create an AST and
+  // walk down it to find tablenames is a cleaner solution, but
+  // there are no up-to-date parser I could find that would not
+  // block modern queries (e.g. windowed queries).
+  // For the sake of parsing table names, this seems to do the
+  // trick, and with enough test coverage it should be fine
+  const tableNameExp = /(?:FROM|JOIN)\s+([a-zA-Z_][a-zA-Z0-9_$.]*)/gi
```
msfstef, line 45: note: we could extract this globally so it's compiled once but we'd need to reset the "lastIndex" before every use, making it slightly more likely to break in the future in insidious ways - given that the table names only need to be parsed once for every query I don't think the marginal performance gain justifies the maintainability loss

clients/typescript/test/client/model/table.errors.test.ts
```diff
@@ -1117,4 +1117,79 @@ test.serial(
...
+    },
+    {
+      instanceOf: InvalidArgumentError,
+      message:
+        'Cannot use queries that might alter the store - please use read-only queries',
+    }
+  )
+})
+
+test('raw update query throws error for unsupported unsafe queries', async (t) => {
```
kevin-dp, line 1135: perhaps let's also test DELETE.

kevin-dp (approved):
Great work 👍

clients/typescript/test/client/model/table.errors.test.ts
```diff
@@ -1117,4 +1117,79 @@ test.serial(
...
+    },
+    {
+      instanceOf: InvalidArgumentError,
+      message:
+        'Cannot use queries that might alter the store - please use read-only queries',
+    }
+  )
+})
+
 // TODO: check why we broke some of the regular tests
```
kevin-dp, line 1210: not sure if this is a leftover from a previous PR or from this one?

kevin-dp: I see, yep some leftover that shouldn't be there anymore. Feel free to remove it or we do it later in another PR.

msfstef:
@kevin-dp @samwillis turns out that a couple of e2e were using the ```raw``` API to do inserts/updates so they ended up failing.

I've added an ```unsafeRaw``` API that is basically the same as what the ```raw``` one was both to fix the e2e and to allow for the behaviour we discussed - however it does beg the question about whether this will break compatibility for anyone using the raw API to modify the store directly.

samwillis (approved):
looks good to me 👍

kevin-dp (approved):
Awesome! Ready to merge 👍

PR #784, opened 2023-12-27
chore (client): refactored socket implementation

kevin-dp:
This PR refactors the implementation of the sockets for the different drivers.
It unifies their implementations in a common `GenericWebSocket` class.
Concrete socket implementations need to extend this class and implement the `makeSocket(opts: ConnectionOptions): WebSocket` method.

### DIFF
--- a/clients/typescript/src/sockets/genericSocket.ts
+++ b/clients/typescript/src/sockets/genericSocket.ts
@@ -1,0 +1,111 @@
+import EventEmitter from 'events'
+import { ConnectionOptions, Data, Socket } from '.'
+import { SatelliteError, SatelliteErrorCode } from '../util'
+
+type WriteType<SupportBuffer extends boolean> = SupportBuffer extends false
+  ? Data
+  : Data | Buffer
+
+interface MessageEvent<T = any> {
+  data: T
+}
+
+export interface IWebSocket<SupportBuffer extends boolean> {
+  send(data: WriteType<SupportBuffer>): void
+
+  addEventListener(event: 'open', cb: () => void): void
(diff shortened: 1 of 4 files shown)


balegas:
This looks great. It was a quite odd implementation that survived until today.
@kevin-dp  have you tested the react-native implementation?

kevin-dp:
> This looks great. It was a quite odd implementation that survived until today. @kevin-dp have you tested the react-native implementation?

I've tested on the web example app. I've also run the unit tests which use node, so that should be fine too. I was not able to test on react-native because i have a local issue which prevents me from running the react-native example application.

balegas:
There were some complications with the React Native driver. I'm sure I might have tried to apply the reusable pattern, I might have missed that it worked with interface since was quite hard to test at the time.

Let's hold merging this before we can test it with react native.

kevin-dp:
> There were some complications with the React Native driver. I'm sure I might have tried to apply the reusable pattern, I might have missed that it worked with interface since was quite hard to test at the time.
> 
> Let's hold merging this before we can test it with react native.

You tested it on RN and it gives some error or some other problem?
Could you provide some more information about the issues you're experiencing

balegas:
> You tested it on RN and it gives some error or some other problem?

Not this PR. When I worked on this the first time.

kevin-dp:
@balegas I fixed my local issues with RN and tested the refactored sockets. It seems to work fine. I uncovered an unrelated issue with the generator which comes from the fact that the generated client includes `./migrations.js` but RN doesn't seem to like that. If i modify that file and replace the import by `./migrations` it works. Will need to address that in a separate PR.

icehaunter (approved):
LGTM

PR #787, opened 2023-12-28
Fix `/ws` 400 error in the `web-wa-sqlite` starter

ickynavigator:
## Issue
- The example repo failed to sync up with other clients/devices
- There was a `/ws` 400 error thrown in the browser (and in the docker logs) image

ps: I wasn't sure if the migrations were meant to be added to git so I just ignored it (please let me know if I should undo that)
pps: I made a typescript+vite version of this example and would be fine with sending in a PR if it's needed - [linked here](https://github.com/ickynavigator/electric-chat)

### DIFF
--- a/examples/web-wa-sqlite/.gitignore
+++ b/examples/web-wa-sqlite/.gitignore
@@ -3,3 +3,4 @@ node_modules
 public/wa-sqlite-async.wasm
 public/wa-sqlite-async.mjs
 src/generated
+.electric_migrations*
--- a/examples/web-wa-sqlite/backend/compose/.envrc
+++ b/examples/web-wa-sqlite/backend/compose/.envrc
@@ -1,5 +1,5 @@
 export POSTGRESQL_IMAGE=postgres:14-alpine
-export APP_NAME=web_wa_sqlite
+export APP_NAME=electric
 export ELECTRIC_PORT=5133
 export ELECTRIC_PROXY_PORT=65432
-export ELECTRIC_IMAGE=electricsql/electric:0.7
+export ELECTRIC_IMAGE=electricsql/electric:0.8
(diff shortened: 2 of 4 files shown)


examples/web-wa-sqlite/.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules
 public/wa-sqlite-async.wasm
 public/wa-sqlite-async.mjs
 src/generated
+.electric_migrations*
```
kevin-dp, line 6: No need to add this to `.gitignore`, this is a temporary file that gets deleted normally. It would only remain if the client generation failed but there is a pending PR which is about to get merged that will always delete the folder (unless passing an flag that explicitly tells not to delete it on error).

examples/web-wa-sqlite/backend/compose/.envrc
```diff
@@ -1,5 +1,5 @@
 export POSTGRESQL_IMAGE=postgres:14-alpine
-export APP_NAME=web_wa_sqlite
+export APP_NAME=electric
```
kevin-dp, line 2: The app name should be `web_wa_sqlite`.

kevin-dp (changes requested):
Thank you for contributing to Electric!
The 400 error code is because the version of Electric and the version of the TypeScript client didn't match. I'm assuming you were running e.g. version 0.8.2 of the client against Electric 0.7 or vice-versa. Your PR bumps both to version 0.8 which is good.

Regarding vite, @samwillis has an open PR https://github.com/electric-sql/electric/pull/786 that introduces vite for this example.

balegas:
As Kevin explained the issue was caused by a mismatch between client and server versions. Examples use specific versions and we only update them for new releases and all at once.

Thanks for engaging.

examples/web-wa-sqlite/.gitignore
ickynavigator: ahh thanks. that should ix it then

ickynavigator:
thanks @balegas !!!

PR #788, opened 2023-12-29
WIP chore(examples/linearlite) Use Yjs+TipTap for issues, allowing merging on conflicting text edits.

samwillis:
### Clone the Electric monorep and checkout the linearlite_yjs branch:
```sh
git clone https://github.com/electric-sql/electric.git
cd electric
git checkout samwillis/linearlite_yjs
```

### Go to the linearlite example:
```sh
cd examples/linearlite
```

### Install the deps (we use npm here as the example is not part of the pnpm workspace):
```sh
npm install ../../clients/typescript/electric-sql-0.9.2.tgz
```

### Start the ElectricSQL and Postgres servers (requires Docker):
```sh
npm run backend:up
```

### Run the database migrations:
```sh
npm run db:migrate
```

### Generate the ElectricSQL client:
```sh
npm run client:generate
```

### Start the WevRTC signaling server (for Yjs presence):
```sh
npm run backend:webrtc
```

### Start the development web server:
```sh
npm run dev
```

### Start the Tauri app:
```sh
npm run tauri dev
```

PR #790, opened 2024-01-01
feature(client) Modify client notifier to emit info about changed rows

samwillis:
Currently the event from `subscribeToDataChanges` only lists the table that has changed, along with the `rowid`s for oplog entries for local changes. This isn't sufficient to watch for a change of a specific row, and requires a full re-query in most cases.

This modifies the notifier to emit info about each row that has changes, along with where the change originated - local, remote, or "initial" for first sync of a shape.

@balegas could you take a look and see if this is a sensible way of doing this?

It is emiting the `record` and `type` ("inset", "update"...) for each row that's change, and I've also extracted the PK(s) for the row changed which makes it easy to then lookup the row.

Once we move to a multi tab/window/process model this could result in large amounts of data being pushed over the broadcast channel. We probably want to add an optional "filter" to `subscribeToDataChanges` so you can do something like:

```ts
notifier.subscribeToDataChanges((changes) => {
  // Do something
}, {
  tables: ['table1', 'table2']
  filter: {
    id: '123123'  
  }
})
```

(I'm using this in the Linearlite+Yjs PR #788)

### DIFF
--- a/.changeset/polite-insects-chew.md
+++ b/.changeset/polite-insects-chew.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Add row primary key infomation to the ActuallyChanged notification.
--- a/clients/typescript/src/notifiers/event.ts
+++ b/clients/typescript/src/notifiers/event.ts
@@ -11,6 +11,7 @@ import {
   Change,
   ChangeCallback,
   ChangeNotification,
+  ChangeOrigin,
   ConnectivityStateChangeCallback,
   ConnectivityStateChangeNotification,
   Notification,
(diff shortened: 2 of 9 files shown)


balegas:
Can you get someone else to review this PR? Let's try to get it merged

samwillis:
@kevin-dp when you have a moment can you take a look at this?

The key thing it does is change the change events to look like this - `recordChanges` is new:
https://github.com/electric-sql/electric/blob/4240bb69bf61ab99429e007a400113be9f4d71b3/clients/typescript/src/notifiers/index.ts#L18-L28

I'm slightly worried it's "noisy" and when we go multi process it will cause a lot of traffic on the notifier channel. But it's also super useful.

One option would be to just emit the PK of the rows, not `record` and `oldRecord`?

Would be good to either merge or close.

(@balegas you may want to have some input on if this is how we want to do this or not)

samwillis:
Dropped the row data from the change event, this PR now adds row PKs and origin to the change notification - they conform to this:

https://github.com/electric-sql/electric/blob/b3c2698baca79035fb703aed3f91eebd92711c8e/clients/typescript/src/notifiers/index.ts#L18-L32

https://github.com/electric-sql/electric/blob/11069a9048533607a3a2d797e2bee5c1df69b8ff/clients/typescript/src/util/types.ts#L123-L128

msfstef (approved):
Looks good, nice work - very convenient info to have for the changes!

Do we believe this additional data being added to this notifier events + the JSON parsing of the primary keys on every update will have any significant performance effect?

msfstef (approved):

PR #793, opened 2024-01-04
fix(client): fix generation of compensation triggers

kevin-dp:
This PR solves an issue that was introduced when adding type support.
We were reading the column types from the wrong table when generating compensation triggers for foreign keys.

### DIFF
--- a/.changeset/light-pants-greet.md
+++ b/.changeset/light-pants-greet.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Fix bug with foreign keys in generate script.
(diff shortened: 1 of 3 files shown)


alco (approved):
I saw this error myself when debugging an issue reported on Discord:
```
TypeError: Cannot read properties of undefined (reading 'sqliteType')
```

So I used `npm pack` with this branch and replaced the test project's `electric-sql` and `prisma-generator` dependencies with locally built ones. That fixed the issue.

So I'm approving this purely on the basis of having seen it resolve the issue in a real use case.

kevin-dp:
> I saw this error myself when debugging an issue reported on Discord:
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'sqliteType')
> ```
> 
> So I used `npm pack` with this branch and replaced the test project's `electric-sql` and `prisma-generator` dependencies with locally built ones. That fixed the issue.
> 
> So I'm approving this purely on the basis of having seen it resolve the issue in a real use case.

Thanks. Code-wise there is not much to say, the code is the same, just reading the type from the right object.

PR #795, opened 2024-01-04
chore(electric): Upgrade Erlang and Elixir to use the latest published versions

alco:


magnetised (approved):
Do we not want to update the version of elixir/erlang in the docker containers? If we're using the latest version for dev then shouldn't we be testing against that in the e2e and using it for the deployed versions?

alco:
> Do we not want to update the version of elixir/erlang in the docker containers? If we're using the latest version for dev then shouldn't we be testing against that in the e2e and using it for the deployed versions?

Yes! Good catch.

magnetised (approved):
Super.

PR #796, opened 2024-01-04
fix(client): Ensure migrations CLI test always fails

msfstef:
A recently added test that ensures migrations folders are removed when generation fails was using a generation command that was not guaranteed to fail.

The test passed on CI cause there's no electric service running there, but locally it was possible for the generation to succeed and thus the test to fail because a sync service was running on localhost.

I changed the service and proxy address to point to invalid ports so this should not be an issue anymore on either CI or locally.


samwillis (approved):
Looks good to me 👍

PR #797, opened 2024-01-04
fix(e2e): Fix flaky prisma E2E that times out every few runs

msfstef:
I noticed that both locally on a fresh run and on CI, the ```04.02_prisma_introspection_generates_correct_schema.lux``` test was occasionally either throwing a warning about the timeout being close to max or timing out.

It seems to be the only test flaking consistently so I figured it makes sense to increase the timeout slightly to make the e2e results more reliable.

I've also updated the README with some info I needed as I set it up to run the e2e locally.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,8 +1,15 @@
 # Integration tests
 
 Tests dependencies:
-- lux tool
-- docker compose
+- [Lux (LUcid eXpect scripting) ](https://github.com/hawk/lux/blob/master/INSTALL.md)
+- [Docker Compose](https://docs.docker.com/compose/install/)
+
+
+You will also need an Electric Docker image, specified by the environment variables ```ELECTRIC_IMAGE_NAME``` and ```ELECTRIC_IMAGE_TAG```, which defaults to the image ```electric:local-build``` that can be built using:
+```sh
+cd ../components/electric
+make docker-build
+```
 
 ## How to add new test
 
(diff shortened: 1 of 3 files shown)


e2e/common.mk
```diff
@@ -3,7 +3,7 @@ export PROJECT_ROOT=$(shell git rev-parse --show-toplevel)
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 export E2E_ROOT := $(dir $(mkfile_path))
 
-LUX=${E2E_ROOT}/lux/bin/lux
+LUX=${E2E_ROOT}lux/bin/lux
```
msfstef, line 6: the ```dir``` command [includes the last slash](https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html#index-dir), at least for me locally adding another one produced an invalid path and couldn't run the tests

e2e/README.md
```diff
@@ -1,8 +1,15 @@
 # Integration tests
 
 Tests dependencies:
-- lux tool
-- docker compose
+- [Lux (LUcid eXpect scripting) ](https://github.com/hawk/lux/blob/master/INSTALL.md)
+- [Docker Compose](https://docs.docker.com/compose/install/)
+
+
+You will also need an Electric Docker image, specified by the environment variables ```ELECTRIC_IMAGE_NAME``` and ```ELECTRIC_IMAGE_TAG```, which defaults to the image ```electric:local-build``` that can be built using:
```
alco, line 8:
```suggestion
You will also need an Electric Docker image, specified by the environment variables `ELECTRIC_IMAGE_NAME` and `ELECTRIC_IMAGE_TAG`, which defaults to the image `electric:local-build` that can be built using:
```

alco (approved):
Lovely :heart:

PR #802, opened 2024-01-05
chore(electric): Define an extended parse_pg_array() in Electric.Utils

alco:
This was originally written as part of the enum type work but it can be reviewed independently of that.


components/electric/lib/electric/utils.ex
```diff
@@ -201,4 +201,90 @@ defmodule Electric.Utils do
...
+      iex> ~s|{"(\\"2023-06-15 11:18:05.372698+00\\",)"}| |> parse_pg_array()
+      [~s|("2023-06-15 11:18:05.372698+00",)|]
+
+      iex> ~s|{"(\\"2023-06-15 11:18:05.372698+00\\",)","(\\"2023-06-15 11:18:05.372698+00\\",)"}| |> parse_pg_array()
+      [~s|("2023-06-15 11:18:05.372698+00",)|, ~s|("2023-06-15 11:18:05.372698+00",)|]
+
+      iex> ~s|{'foo', 'bar{}', "tableName",'{baz,quux}', '}},,,{{',id, ",,\\"{}"}| |> parse_pg_array()
+      ["'foo'", "'bar{}'", "tableName", "'{baz,quux}'", "'}},,,{{'", "id", ",,\"{}"]
+  """
+  def parse_pg_array("{}"), do: []
```
magnetised, line 219: I think this would be happier as `Electric.Postgres.parse_array` don't you think?

components/electric/lib/electric/utils.ex
```diff
@@ -201,4 +201,90 @@ defmodule Electric.Utils do
   def inspect_relation({schema, name}) do
     "#{inspect(schema)}.#{inspect(name)}"
   end
+
+  @doc ~S"""
+  Parse a Postgres string-serialized array into a list of strings, unwrapping the escapes
+
+  ## Examples
+
+      iex> ~s|{"(\\"2023-06-15 11:18:05.372698+00\\",)"}| |> parse_pg_array()
```
magnetised, line 210:
```suggestion
      iex> ~S|{"(\"2023-06-15 11:18:05.372698+00\",)"}| |> parse_pg_array()
```

magnetised (approved):
Nice

components/electric/lib/electric/utils.ex
alco: I've opened https://github.com/electric-sql/electric/pull/809 where I move disparate PG type parsing/serialization functions under one namespace.

PR #803, opened 2024-01-05
chore(electric): Improve error logging in column value validation

alco:
The idea behind this change is to make sure every pattern-matching failure logs the offending value.

Previously, we could get an error looking like this

    ** (MatchError) Patter match failed: true = false


magnetised (approved):
Lovely

PR #804, opened 2024-01-05
feat(electric): Enable electrification of tables with columns of user-defined PG ENUM types

alco:
Closes VAX-1040, VAX-1041.

Supersedes https://github.com/electric-sql/electric/pull/564.

### DIFF
--- a/components/electric/lib/electric/postgres/extension/schema_cache.ex
+++ b/components/electric/lib/electric/postgres/extension/schema_cache.ex
@@ -195,6 +195,10 @@ defmodule Electric.Postgres.Extension.SchemaCache do
     end
   end
 
+  def enums(origin) do
+    call(origin, :enums)
+  end
+
   defp call(name, msg) when is_binary(name) do
     call(name(name), msg)
   end
@@ -404,6 +408,10 @@ defmodule Electric.Postgres.Extension.SchemaCache do
   def handle_call({:internal_relation, relation}, _from, state) do
     state = load_internal_schema(state)
     {:reply, Schema.table_info(state.internal_schema, relation), state}
+  end
+
+  def handle_call(:enums, _from, state) do
+    load_and_reply(state, fn schema_version -> schema_version.schema.enums end)
   end
 
   @impl GenServer
--- a/components/electric/lib/electric/postgres/extension/schema_cache.ex
+++ b/components/electric/lib/electric/postgres/extension/schema_cache.ex
@@ -16,6 +16,8 @@ defmodule Electric.Postgres.Extension.SchemaCache do
 
   use GenServer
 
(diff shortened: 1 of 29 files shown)


components/electric/lib/electric/postgres/extension/schema_cache.ex
```diff
@@ -108,19 +110,21 @@ defmodule Electric.Postgres.Extension.SchemaCache do
     call(origin, {:known_migration_version?, version})
   end
 
+  # SchemaCache.internal_schema() is not used anywhere in the codebase.
+  #
+  # SchemaLoader.internal_schema() is used in this module to actually load the internal schema from the database. But
+  # since SchemaCache also implements the SchemaLoader behaviour, we're forced to have this noop definition here.
```
magnetised, line 116: Yeah, I think this `internal_schema` function doesn't belong in the schema loader behaviour. add it to the list of things to fix there. https://linear.app/electric-sql/issue/VAX-1193/resolve-issues-with-schemaloader-vs-schemacache

components/electric/lib/electric/postgres/oid_database/pg_type.ex
```diff
@@ -58,6 +69,8 @@ defmodule Electric.Postgres.OidDatabase.PgType do
   # convert domain types to atoms for now.
   defp type_name("electric", name, :DOMAIN), do: String.to_atom("electric." <> name)
 
-  # User-defined and other custom types get schema-qualified names as strings.
+  # User-defined and other custom types are strings containing a schema-qualified name or a plain name, if the type's
+  # schema is "public".
+  defp type_name("public", name, _kind), do: name
```
magnetised, line 74: why not return `"public".$name` here?

magnetised (approved):
Looks fantastic. 👍🏻

components/electric/lib/electric/postgres/extension/schema_cache.ex
alco: This function is only needed in the `logical_replication` mode. If we end up removing it and leaving only the `direct_writes` mode, we'll be able to simply delete it.

By the way, it is tempting to use the SchemaLoader simply as a way to get hold of a pooled DB connection to run some queries. We need a proper pool abstraction a-la Ecto.Repo to be able to run queries anywhere in the code without worrying about spamming the DB with connections.

components/electric/lib/electric/postgres/oid_database/pg_type.ex
alco: Because we don't get the namespace part from `ddlgen_create()`. So when we parse the column types in `MigrationConsumer`,  we can only assume that a custom type's namespace is `public`. We've discussed this issue in the previous version of this PR.

We can clean things up once we update ddlgen functions to always generate schema-qualified names.

PR #805, opened 2024-01-08
feat(client): Add new names for safe and unsafe raw APIs and deprecate old ones

msfstef:
Having split the raw APIs to read-only safe ones and an unsafe one, I've also renamed them:

- `raw` with read-only checks to `rawQuery`
- `liveRaw` to `liveRawQuery`
- `raw` without checks to `unsafeExec`

I've updated all relevant docs I could find as well.

Some exampels (beer-stars, introduction, and linearlite) will be using the old versions of the API as they have their electric-sql version fixed to 0.7.0, so when it's updated those PRs should take care of also fixing the deprecated APIs.

### DIFF
--- a/.changeset/silent-pillows-jog.md
+++ b/.changeset/silent-pillows-jog.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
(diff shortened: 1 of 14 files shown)


samwillis (approved):
Looks good to me 👍 

We could add a `console.warning()` message when it's used, but it's probably not needed.

PR #806, opened 2024-01-08
docs: update sync service and migrations docs.

thruflo:


docs/integrations/backend/prisma.md
```diff
@@ -21,7 +21,7 @@ datasource db {
 You can then use [dotenv-cli](https://www.npmjs.com/package/dotenv-cli) to setup multiple `.env` files. For example, create a `.env.proxy` with a database URL to connect to the proxy:
 
 ```shell
-DATABASE_URL=postgresql://postgres:${PG_PROXY_PASSWORD}@localhost:${PG_PROXY_PORT}/mydb
+DATABASE_URL=postgresql://prisma:${PG_PROXY_PASSWORD}@localhost:${PG_PROXY_PORT}/mydb
```
alco, line 24: This change needs to be reverted.

The only time we need to use the `prisma` user is when passing a proxy URL to the `--proxy` flag of the `generate` command. When Prisma used as a migration tooling, the regular username should be used which is usually `postgres`. Using the `prisma` username in this context will actually ensure Prisma's failure to apply migrations to the DB.

docs/usage/data-modelling/migrations.md
```diff
@@ -197,16 +201,20 @@ See <DocPageLink path="integrations/backend" /> and <DocPageLink path="api/ddlx"
...
 
-This proxy server serves various purposes:
+DDL migrations not applied via the proxy are not captured by Electric. This is fine if they do not use DDLX syntax or do not affect the electrified part of your schema (it's also fine if they affect tables that are electrified later on).
 
-- It allows the use of the [DDLX syntax](../../api/ddlx.md) for managing your tables and access permissions,
-- It captures migrations applied to Electrified tables in order to propagate those DDL changes to the client schemas,
-- It validates migrations to electrified tables to ensure that changes to the schema are supported by Electric (e.g. validating the types of any added columns, ensuring that only additive migrations are applied, etc), and
-- It provides an endpoint for schema introspection to allow Electric to return its view of the underlying Postgres database to the data access library.
+However, if they **do** use DDLX syntax or **do** affect the electrified part of your schema, then they **must** be applied via the proxy. If not, they will be rejected and an error will be raised.
```
alco, line 213: How do we detect such modifications when they sidestep the proxy? I thought we could only do that with the help of event triggers which we no longer use?

docs/usage/data-modelling/migrations.md
```diff
@@ -216,17 +224,13 @@ You can either route all DDL access to your Postgres via the Migrations proxy, o
 
 ### Configuring and connecting to the migrations proxy
 
-There are two environment variables that configure the proxy in Electric:
-
-- `PG_PROXY_PORT` (default `65432`). This is the TCP port that [**Electric sync service**](../../api/service.md) will listen on. You should connect to it in order to pass through the migration proxy. Since the proxy speaks fluent Postgres, you can connect to it via any Postgres-compatible tool, e.g. `psql -U electric -p 65432 electric`
+The [Electric sync service](../../api/service.md) connects to a Postgres database at [`DATABASE_URL`](../../api/service.md#database_url) and, [by default](#using-the-proxy-tunnel), exposes the migrations proxy as a TCP service on `PG_PROXY_PORT`, secured by `PG_PROXY_PASSWORD`. Instead of connecting your migrations tooling (or any interactive session applying migrations) to Postgres directly, connect to the Electric sync service instead.
```
alco, line 227: I think the link on `by default` looks odd and is a confusing context switch. I would remove the link and leave the plain `by default`  text.

docs/usage/data-modelling/migrations.md
```diff
@@ -216,17 +224,13 @@ You can either route all DDL access to your Postgres via the Migrations proxy, o
...
-   Some deployment targets restrict you to a single port for your service. On these platforms, `PG_PROXY_PORT` can be set to a special `http` value. This enables the use of the [Proxy Tunnel](../../api/cli.md#proxy-tunnel). Additionally, to set both the TCP port and enable the Proxy Tunnel, use a value such as `http:65432`.
+[![Connecting to the migrations proxy over a TCP port](../../deployment/_images/tcp-port.png)](../../deployment/_images/tcp-port.jpg)
 
-- `PG_PROXY_PASSWORD` (no default). Access to the proxy is controlled by password (see below for information on the username). You must set this password here and pass it to any application hoping to connect to the proxy.
+The migrations proxy intercepts your migrations, validates and transforms them and proxies them on to the Postgres at `DATABASE_URL`. Since the proxy speaks fluent Postgres, you can connect to it via any Postgres-compatible tool, e.g. `psql -U postgres -p 65432 electric`:
 
-You should be able to connect to the proxy directly using `psql` as outlined above and run any DDLX/migration commands you like. These will be validated, captured, and streamed to any connected clients automatically:
-
-```
+```console
```
alco, line 233: Using ```sql instead would improve the readability of this snippet.

docs/usage/data-modelling/migrations.md
```diff
@@ -240,16 +244,26 @@ electric=# ALTER TABLE public.items ADD COLUMN amount integer;
...
 
-Your framework of choice will need to be configured in order to pass migrations (and _only_ migrations, you shouldn't connect your application to the proxy endpoint for any other purpose) through the proxy rather than directly to the underlying Postgres database.
+### Framework integration
 
-As each framework has different requirements for this, example code for each is provided in the [integrations section](../../integrations/backend/index.md)
+Your framework of choice will need to be configured in order to pass migrations (and _only_ migrations) through the proxy rather than directly to the underlying Postgres database.
+
+This typically involves using a different database connection string for migrations, often by running your migrations command with a different `DATABASE_URL` that connects to the migrations proxy. However, as each framework has different requirements for this, we provide framework specific example code for this is provided in <DocPageLink path="integrations/backend" />.
```
alco, line 253:
```suggestion
This typically involves using a different database connection string for migrations, often by running your migrations command with a different `DATABASE_URL` that connects to the migrations proxy. However, as each framework has different requirements, we provide framework-specific example code for this in <DocPageLink path="integrations/backend" />.
```

alco (commented):
First batch of comments out of two. I'll do another round to look at the Sync Service API updates and database permissions in more detail a bit later.

docs/integrations/backend/prisma.md
thruflo: Right, great to clarify. So the place to document the prisma username thing is in the cli.md#generate command?

docs/usage/data-modelling/migrations.md
thruflo: Right, I thought we still had some triggers in there but you're probably right. @magnetised can you confirm? And thus do we have any protection against schema drift or are we reliant on the user being disciplined?

docs/usage/data-modelling/migrations.md
thruflo: You'd think that but it came out as:

image

Which is worse, I think.

alco: I agree that it's worse. Nevermind then.

docs/integrations/backend/prisma.md
alco: Yes, the docs for the `generate` command should bring the reader's attention to the fact that the username **must** be `prisma`.

thruflo: Thanks, I've added the note.

thruflo:
@alco let me know if good to go or if you still need to look at the the sync service page. Thanks!

alco:
@thruflo still looking at it. I won't hold you back for long, just minor corrections. More significant content updates can be done in as a followup.

docs/api/service.md
```diff
@@ -21,86 +25,322 @@ docker run \
...
+
+:::info Why are there two modes?
+Originally (prior to v0.8) all writes to Postgres were made using logical replication.
+
+In [version 0.8](/blog/2023/12/13/electricsql-v0.8-released#from-superuser-to-supabase), Electric added direct writes mode to reduce the database user permissions required and increase compatibility with Postgres hosting providers.
+
+In future, we may depreciate logical replication and consolidate on direct writes. However, logical replication may have performance advantages and, for now, we've kept both modes available while direct writes mode gains maturity and we learn more about the operational characteristics of both modes.
+:::
+
+## Database user permissions
```
alco, line 173: I'm unuser about adding this section. In its current state, in needs some points to be expanded upon. For example, it's usually not possible to create a new user with `SUPERUSER` in managed Postgreses. As another example, AWS RDS does not have the `REPLICATION` role attribute and uses a custom builtin role instead to achieve the same goal.

Crucially, we don't have a solution yet for people to use one PG user for Electric and a different one for their backend app or any other component that, say, has created some tables in the database before Electric. If there's an existing backend app and the developer introduces Electric and starts running migrations through Electric Proxy, the backend app will not be able to access new table created in this way without additional setup.

IMO, we should only say that the direct writes mode doesn't require a SUPERUSER role but we advise using the default role provided by your managed Postgres of choice or use the same role as your existing app uses, provided that you can GRANT it the `REPLICATION` role attribute.

alco:
@thruflo I'm done with my review. Pushed one new commit to fix a broken link for `LOGICAL_PUBLISHER_PORT`.

docs/api/service.md
thruflo: It's imperative that we add documentation about the role attributes / permissions that your database user needs. Otherwise our users are going to get errors and have to fix them in the dark.

Another way of saying this is that it is our responsibility to know and document the necessary permissions for the database user that Electric connects to Postgres as.

This sub-section is the candidate location for this information, linked to from the installation page. If we need to review and figure out the right contents for this section, perhaps let's hop on a quick call and talk through it?

alco: My idea is to hold off including this section in this PR. I would like to simplify it slightly but I don't want that to block other changes in this PR from getting merged.

We can also log detailed error messages when we get `insufficient_permissions` from Postgres, tailored to specific places in the code where they happen. All I'm saying is that we should take this on as a separate changeset:
- improve error reporting for permission errors
- edit this proposed "Database user permissions" section to try simplifying it
- include the URL to this new documentation section in logged errors to help users find help on their own

alco (approved):
@thruflo Looks like you've missed a couple of my corrections. I've pushed them myself.

Looks great overall! The only remaining contention, IMO, is the section on database permissions.

docs/api/service.md
thruflo: OK, but note that this information is currently in the docs, in a worse form. So I suggest we merge this and then improve it.

thruflo:
Yup, sorry, have pushed my version (with extra `the`s).

PR #807, opened 2024-01-08
fix(e2e): increase base timeout multiplier on CI to reduce flakiness

msfstef:
I've noticed the e2e running satellite is a bit flaky because the docker startup occasionally times out.

We have explicitly specified a larger timeout for most docker startup commands so this should be more consistent and reduce flakiness.

Example of flake logs for reference:
```
2024-01-08T09:28:00.4851831Z test case         : tests/03.01_node_satellite_loads_local_migrations.lux
2024-01-08T09:28:11.5151352Z progress          : (()..:...:.)()((...:...:..):...:.:.:.26????26)23C..:...:.(stop development environment..:.(.:.:.:.:.:.:.:.:.:.:.:.:...:.:...))
2024-01-08T09:28:11.5167772Z result            : FAIL at line 23:26 in shell satellite_1
2024-01-08T09:28:11.5168696Z expected=
2024-01-08T09:28:11.5169046Z 	>
2024-01-08T09:28:11.5177569Z actual match_timeout
2024-01-08T09:28:11.5178463Z 	make start_satellite_client_1
2024-01-08T09:28:11.5179705Z 	make[1]: Entering directory '/home/runner/work/electric/electric/e2e/tests'
2024-01-08T09:28:11.5180750Z 	docker compose -f compose.yaml run \
2024-01-08T09:28:11.5181374Z 		--rm \
2024-01-08T09:28:11.5182049Z 		-e TERM=dumb \
2024-01-08T09:28:11.5182551Z 		satellite_client_1
2024-01-08T09:28:11.5183101Z 	[1A[1B[0G[?25l[+] Creating 1/0
2024-01-08T09:28:11.5184022Z 	 [32m✔[0m Network tests_default  [32mCreated[0m[34m0.1s [0m
2024-01-08T09:28:11.5184832Z 	[?25h
```

### DIFF
--- a/e2e/common.mk
+++ b/e2e/common.mk
@@ -35,6 +35,14 @@ else
 	export ELECTRIC_CLIENT_IMAGE=${ELECTRIC_CLIENT_IMAGE_NAME}:${ELECTRIC_IMAGE_TAG}
 endif
 
+# Any timeouts in the tests, specified in seconds,
+# are multiplied by this to convert to milliseconds.
+# If in CI, double all timeouts to reduce flakiness
+TIMEOUT_MULTIPLIER = 1000
+ifeq ($(CI), true)
+    TIMEOUT_MULTIPLIER = 2000
+endif
+
 lux: ${LUX}
 
 ${LUX}:
@@ -123,7 +131,7 @@ docker-make:
 		make ${MK_TARGET}
 
 single_test:
-	${LUX} --progress doc ${TEST}
+	${LUX} --multiplier ${TIMEOUT_MULTIPLIER} --progress doc ${TEST}
 
 single_test_debug:
 	${LUX} --debug ${TEST}
--- a/e2e/common.mk
+++ b/e2e/common.mk
@@ -12,7 +12,8 @@ ifeq ($(CI), true)
     TIMEOUT_MULTIPLIER = 2000
 endif
 
-LUX=${E2E_ROOT}lux/bin/lux --multiplier ${TIMEOUT_MULTIPLIER}
+LUX_PATH=${E2E_ROOT}lux/bin/lux
+LUX=${LUX_PATH} --multiplier ${TIMEOUT_MULTIPLIER}
 
 DOCKER_REGISTRY  = europe-docker.pkg.dev/vaxine/vaxine-io
 DOCKER_REGISTRY2 = europe-docker.pkg.dev/vaxine/ci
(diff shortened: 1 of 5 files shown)


alco:
@msfstef The default timeout in lux is 10 seconds. From your log snippet it can be seen that the second line has a timestamp roughly +10 seconds from the first one. I suppose, lux saves the last update time of the "prgress" line as is its timestamp in the logs.

So this begs the question of whether your change is actually improving things. It sets the timeout to 5 seconds which is lower than the default. I've also found sporadic timeout changes in the `e2e/tests/_shared.luxinc` file. They all set the timeout to a lower value than 10 seconds.

Maybe I'm missing a place where the default is changed in our test  suite? I would prefer to get rid of those ad-hoc timeout settings and instead pass a [`--timeout` flag](https://github.com/hawk/lux/blob/master/doc/lux.md#timeouts) to `lux` which can be made larger on CI, since it's no uncommon to see flake on CI.

msfstef:
@alco whoops, you're right! I couldn't find where the default timeout was set for lux and didn't think to look at the log timestamps.

I agree that it'd be better if most timeouts were not ad hoc, let me try and shift some things around and remove those.

msfstef:
@alco I've removed the ad hoc timeouts and used lux's "multiplier" option such that any timeouts specified are twice as long on CI - this way any meaningful ad hoc timeout changes still work but will be twice as long on CI.

https://github.com/electric-sql/electric/blob/975cce54d4a5807e757cfe55d6df033493beeeed/e2e/common.mk#L38C9-L44

e2e/common.mk
```diff
@@ -123,7 +131,7 @@ docker-make:
 		make ${MK_TARGET}
 
 single_test:
-	${LUX} --progress doc ${TEST}
+	${LUX} --multiplier ${TIMEOUT_MULTIPLIER} --progress doc ${TEST}
```
alco, line 134: This is better but there are multiple make targets across this file and `Makefile` that invoke lux. I would suggest moving the definition of `TIMEOUT_MULTIPLIER` to the top of this file, just above the definition of `LUX` and then add the `--multiplier` flag inside the definition of `LUX`. That way, all instances of `${LUX}` will pick it up.

alco (approved):

e2e/common.mk
msfstef: Agreed - was worried about polluting any other use of lux but this makes it much cleaner.

PR #809, opened 2024-01-08
chore(electric): Consolidate custom PG types under one module namespace

alco:
Follow-up to https://github.com/electric-sql/electric/pull/802.


magnetised (approved):
boom. you ran a long way with that! nice.

Issue #777, opened 2023-12-21
"iat" claim is invalid unless I add timeout


KyleAMathews:
I'm setting up Electric with clerk.dev and am running into odd validation errors with the clerk jwt where Electric rejects the token for unless I add a timeout.

The error looks like:

`Invalid "iat" claim value: 1703184246`

A timeout after auth of ~300ms consistently fails. If I extend the timeout to 500ms it's pretty consistent (though still occasionally fails).

Is there a way to loosen the constraints here a bit?

alco:
@KyleAMathews Thanks for sharing your experience! We should relax the time resolution for `iat` and `nbf` claim validations to account for immediate use of a token after generation and possible clock skew. I'm thinking of adding a 10-second leeway, probably also to the `exp` claim validation.

PR #812, opened 2024-01-09
chore(electric): Relax JWT timestamp validation by introducing the idea of "clock drift"

alco:
Fixes VAX-1449, #777.

### DIFF
--- a/components/electric/config/runtime.exs
+++ b/components/electric/config/runtime.exs
@@ -60,6 +60,18 @@ config :electric, Electric.Replication.Postgres,
 config :electric, Electric.Postgres.Proxy.Handler.Tracing, colour: true
 
 config :electric,
+  # The default acceptable clock drift is set to 2 seconds based on the following mental model:
+  #
+  #   - assume there's a server that generates JWTs and its internal clock has +1 second drift from UTC
+  #
+  #   - assume that Electric runs on a server that has -1 second clock drift from UTC
+  #
+  #   - when a new auth token is generated and is immediately sent to Electric, the latter will
+  #     see its `iat` date being 2 seconds in the future (minus network and processing latencies)
+  #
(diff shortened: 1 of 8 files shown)


components/electric/config/runtime.exs
```diff
@@ -59,7 +59,9 @@ config :electric, Electric.Replication.Postgres,
 
 config :electric, Electric.Postgres.Proxy.Handler.Tracing, colour: true
 
-config :electric, :telemetry_url, "https://checkpoint.electric-sql.com"
+config :electric,
+  max_clock_drift_seconds: 2,
```
magnetised, line 63: shouldn't this be configurable? hate to add another env var but...

magnetised (approved):
if we're going to hard-code 2 seconds as the max drift, then I think it needs some explanation, but otherwise looks great

components/electric/config/runtime.exs
alco: It is configurable. One can define a new Docker image based on a publish Electric one, override this `runtime.exs` file and modify any settings that we expose as application env parameters in it. For example:

```
FROM electricsql/electric:0.8.2

RUN echo 'config :electric, max_clock_drift_seconds: N' >> releases/0.8.2/runtime.exs
```

Mapping to an environment variable is a different issue. I don't think it's necessary until we see there's demand for it.

davidmartos96:
@alco I'm assuming this PR removes the need of the following "patch" in the client
https://github.com/electric-sql/electric/blob/4fe5c7f687b4c43e9ba17133645e24179d481aad/clients/typescript/src/auth/secure/index.ts#L17

alco:
> @alco I'm assuming this PR removes the need of the following "patch" in the client

Yes. Thanks for the heads-up!

PR #819, opened 2024-01-10
chore(electric): Perform the same column checks we do when electrifying a table for ALTER TABLE ... ADD COLUMN

alco:
Fixes VAX-1283.


components/electric/priv/sql_function_templates/electrify/__validate_table_column_types.sql.eex
```diff
@@ -1,5 +1,11 @@
--- This procedure validates the type of each column of the table that's being electrified
--- to check if the type is supported by Electric.
+-- Validate the type of each column of the table that's being electrified,
```
magnetised, line 1: we can remove all these warnings when we move validation of electrified tables to the proxy code...

magnetised (approved):
Thanks, looks great

PR #821, opened 2024-01-10
fix(generator) Revert #753 as it breaks for some bundlers - will fix this bug in the CLI

samwillis:
Revert "chore(generator) Change generator to add .js extention to imports for nodenext compatibility (#753)"

This reverts commit 22652fb3e7616bc5d48f445cdad2d77a0c99c9a8.


balegas (approved):
Cool. 
- Make sure we open an issue to fix this in the CLI
- Would be nice to create a GH issue documenting what we know currently is not supported

PR #824, opened 2024-01-12
fix(client_cli) Fix argName when option starts with ELECTRIC_

davidmartos96:
There is a small bug in the CLI options generation. The argName is not turned into lowercase when the option starts with ELECTRIC_

` npx electric-sql start --help`

image

@samwillis Related to the new CLI, when looking into adding support for it in `electric_dart` we had some questions.

## 1. Start command

Should there be a `docker compose pull <image>` before starting the docker? The CLI already uses the correct docker image, but it might be important to pull in case the user was using image tag `0.x` (`0.x.0` under the hood) and later there is a patch in the service, so the user might not pull `0.x.1`

## 2. start-command usage from the generate command

The generate command runs the start command when fetching migrations, but when stopping the service, it uses the flag `remove: true`. Why is that? Wouldn't that cause data loss to the user? It would be deleting the docker volumes. Or are we missing something? 

Awesome work on the CLI btw, I think it can really help people to get started without many extra script files :smile:

### DIFF
--- a/clients/typescript/src/cli/config.ts
+++ b/clients/typescript/src/cli/config.ts
@@ -156,7 +156,7 @@ export function addOptionToCommand(
 ) {
   let argName = optionName.toLocaleLowerCase().replace(/_/g, '-')
   if (argName.startsWith('electric-')) {
-    argName = optionName.slice('electric-'.length)
+    argName = argName.slice('electric-'.length)
   }
   let localName: string = optionName
   if (!optionName.startsWith('ELECTRIC_')) {


samwillis (approved):
@davidmartos96, Good spot, thanks!

On item 2, the intention is that `--with-migrations` is used with an ephemeral database thats created at the start and then disposed off after the client has been generated.

Maybe we need to look at ensuring that the container has a unique id (and ports) so that we are certain that its not being used with `start`.

davidmartos96:
@samwillis 
Oh, I see. So the migrations are run against a fresh temporary database. When I saw the command I thought it was supposed to connect to the real database and apply the migrations + generate the code in a single command

davidmartos96:
@samwillis If that's the intention, then it currently acts in an unexpected way. When running `npx cli start --with-postgres` and then `npx cli generate --with-migrations="some command"`, the generate command starts the same electric service and database as the `start` command, so when `generate` finishes and cleans up, it's causing data loss because of the volumes removal.

samwillis:
I think it needs to be documented better under the help command then. Will add that to the list for Monday, along with aiming to have it start an explicitly ephemeral container.

With it we are aiming for what we are calling a "local-only-first" workflow, we go into a little more fail on this in the new web based docs. We've had a number of people ask for a workflow where they don't want the sync service running, and that's what I'm trying to service with that command.

But hearing what you thought it was for, is making me wander if we want to enable that too...

Maybe we actually want both your interpretation and an --ephemeral arg that creates and distributes and transient container.

Thanks so much for testing, it's an enormous help!!

PR #825, opened 2024-01-12
fix(client_cli): Proxy fixes 

davidmartos96:
A couple of minor issues I found:

1. ~~Use the prisma user in the db proxy url, otherwise non elecrified tables are generated.~~ Bad call, the prisma user is correctly being used for the introspection. The proxy url doesn't always need to use the prisma user. We can however log the correct proxy url used when generating the client, because the one in the config differs the one used for the introspection.
https://github.com/electric-sql/electric/blob/7fbf292ef1bcb07a71fd4e2a528d2cda965ac5ce/clients/typescript/src/cli/migrations/migrate.ts#L73
https://github.com/electric-sql/electric/blob/7fbf292ef1bcb07a71fd4e2a528d2cda965ac5ce/clients/typescript/src/cli/migrations/migrate.ts#L303-L309

2. Handle ssl mode when building the db url, so that when running locally, the db and proxy urls are built with `?sslmode=disable`. Some postgres migration clients won't connect without that flag in the url when running without ssl.

### DIFF
--- a/clients/typescript/src/cli/config-options.ts
+++ b/clients/typescript/src/cli/config-options.ts
@@ -34,9 +34,11 @@ export const configOptions = {
     defaultVal: () => {
       const host = getConfigValue('SERVICE_HOST')
       const port = getConfigValue('PG_PROXY_PORT')
+      const user = 'postgres'
       const password = getConfigValue('PG_PROXY_PASSWORD')
       const dbName = getConfigValue('DATABASE_NAME')
-      return `postgresql://postgres:${password}@${host}:${port}/${dbName}`
+      const ssl = getConfigValue('DATABASE_REQUIRE_SSL')
+      return buildDatabaseURL({ host, port, user, password, dbName, ssl })
     },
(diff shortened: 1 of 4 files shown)


samwillis:
Thanks @davidmartos96 for the changes, I've merged them into my branch of various fixes over here: https://github.com/electric-sql/electric/pull/834

I plan to finish this off, and do some more testing, tomorrow.

alco:
@davidmartos96
> Some postgres migration clients won't connect without that flag in the url when running without ssl.

I'd like to know which clients have this.

davidmartos96:
@alco `dbmate` is an example, where it's required to provide `?sslmode=disable` in the URL, and if I remember correctly, `go-migrate` too, which we used previously.

alco:
> @alco `dbmate` is an example, where it's required to provide `?sslmode=disable` in the URL, and if I remember correctly, `go-migrate` too, which we used previously.

Thanks!

PR #826, opened 2024-01-15
fix(client): Validate urls in the Electric config

davidmartos96:
Attempt to prevent / warn users to provide invalid URLs to the config.

Discord source: https://discord.com/channels/933657521581858818/1110941752875028530/1194966674026270831

In this example, the user provides that URL by mistake, probably because of lack of familiarity with Electric/Postgres and the client works correctly. The problem is that now the app is embedded with some credentials.

image

### DIFF
--- a/clients/typescript/src/config/index.ts
+++ b/clients/typescript/src/config/index.ts
@@ -133,6 +133,10 @@ function parseServiceUrl(inputUrl?: string): {
     warnings.push('Username and password are not supported.')
   }
 
+  if (url.pathname !== '/' && url.pathname !== '') {
+    warnings.push('An URL path is not supported.')
+  }
+
   const isSecureProtocol = url.protocol === 'https:' || url.protocol === 'wss:'
   const sslEnabled = isSecureProtocol || url.searchParams.get('ssl') === 'true'
 
--- a/clients/typescript/src/config/index.ts
+++ b/clients/typescript/src/config/index.ts
@@ -71,19 +71,12 @@ export const hydrateConfig = (config: ElectricConfig): HydratedConfig => {
   }
 
   const debug = config.debug ?? false
-  const url = new URL(config.url ?? 'http://localhost:5133')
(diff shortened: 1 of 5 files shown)


kevin-dp:
Parsing the service URL with `new URL(...)` is a great improvement!
Not sure about throwing an error versus logging a warning.
@thruflo and @samwillis may have an opinion on this.

clients/typescript/src/config/index.ts
```diff
@@ -114,3 +107,43 @@ export const hydrateConfig = (config: ElectricConfig): HydratedConfig => {
...
+  }
+
+  const validProtocols = ['http:', 'https:', 'ws:', 'wss:', 'electric:']
+
+  if (!validProtocols.includes(url.protocol)) {
+    throwInvalidServiceUrlError('Invalid url protocol.')
+  }
+
+  if (url.username || url.password) {
+    throwInvalidServiceUrlError('Username and password are not supported.')
```
thruflo, line 130: This type of check is the kind of thing I imagine could cause a problem in some hard to foresee edge case where someone was using credentials in some kind of proxy layer.

if we don't use them, can we not just allow them in case someone else does? And we just proactively validate the parts of the URL that we use?

davidmartos96: In the latest commit I've changed the implementation to only throw when `new URL(...)` fails and for the others log a warning.

I understand not wanting to break existing users, but at the same time, if the Electric client doesn't use those parts, might as well let the user know that it is an unexpected usage. If the users have a custom URL, they could provide a "expected" URL with `https://<ip>:<port>`.

The user that encountered this was entering database credentials in that field, but getting a succesful connection at the same time.

I feel a warning improves the situation without breaking, and the special use case with a proxy layer with credentials could easily silence the warning by constructing the url "correctly".

What do you think?

KyleAMathews:

Issue #491, opened 2023-09-23
Local Only Mode


MentalGear:
First of all: Very impressive work, looking forward to add this into our techstack. Thank you!

I'm currently building a small browser extension and, before I add any remote server sync/storage, I'd like to start small and only save data locally. Is it possible to use electric without a server pg backend  - so it runs completely local ?

kevin-dp:
Hi @MentalGear,
It is perfectly possible to run an Electric client without the Electric backend.
In that case the client will run in offline mode so it won't be able to connect to the backend but will still be fully functional meaning that you can read and write the local SQLite DB.

I will let @thruflo answer your pgvector question as he has a better viewpoint on our roadmap.

thruflo:
Hey, how would you like to use pg_vector?

MentalGear:
Thanks for getting back, I appreciate it!

**Local-Mode**
Would it be possible to outline the setup process of having electric run in "local-mode"? As I understand, normally one needs to have a postgres back-end in place to "electrify" tables - or can it be done without? Would I need to go through the normal setup and then just disconnect the database in code?

As I think about it, it'd be great if you had a page on the docs for this kind of "local-mode starter setup".
Lots of apps start small and only need local storage at first: it would be a huge USP if a project can start out with electric in the initial stack for simple client-side storage and, once more advanced sync features are needed, seamlessly mature into using a backend db without rewriting the entire codebase - just by connecting the database!

**Local Storage Considerations**
Are there any risks of running in local-mode for a prolonged amount of time? (ex: time/space limit). It would be a huge frustration for users if the app would just lose data after x days because of some cache cleaning/storage limitation.

**pg_vector**
I'd like to use a vector search for clustering similar textual data client-side. I guess vector search can't be done directly on electric, therefore I have been exploring using client-side [postgres-wasm](https://github.com/snaplet/postgres-wasm) or "voy-search".
I guess if using postgres-wasm, electric can return vector search results like normal queries ?

kevin-dp:
> Would it be possible to outline the setup process of having electric run in "local-mode"? As I understand, normally one needs to have a postgres back-end in place to "electrify" tables - or can it be done without? Would I need to go through the normal setup and then just disconnect the database in code?

The app will run in local-mode if it fails to reach Electric (i.e. the server).
To setup the app you can indeed go through the regular setup process in order to generate the client and then not run the Electric backend.

[...]

PR #828, opened 2024-01-15
chore (client): detach electrification from connection

kevin-dp:
This PR detaches electrification of a database from connecting to the Electric sync service.
Authentication credentials still need to be passed when calling `electrify` because the client ID is used for generating tags when snapshotting the oplog.

Note that the ts-client will complain if you try to sync a shape before being connected to the Electric sync service.

The flow is as follows:

```ts
await electric.connect()
// only now can one sync shapes
```

This PR unblocks https://github.com/electric-sql/electric/issues/491 as it enables using Electric in local-only mode as one can essentially never call `electric.connect`. In local-only mode, one never syncs any shape.

### DIFF
--- a/.changeset/chilled-avocados-compare.md
+++ b/.changeset/chilled-avocados-compare.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Detach database electrification from connecting to the Electric sync service.
--- a/clients/typescript/src/client/model/client.ts
+++ b/clients/typescript/src/client/model/client.ts
@@ -5,7 +5,12 @@ import { Row, Statement } from '../../util'
 import { LiveResult, LiveResultContext } from './model'
 import { Notifier } from '../../notifiers'
 import { DatabaseAdapter } from '../../electric/adapter'
-import { GlobalRegistry, Registry, Satellite } from '../../satellite'
+import {
+  ConnectionWrapper,
+  GlobalRegistry,
+  Registry,
+  Satellite,
+} from '../../satellite'
 import { ShapeManager } from './shapes'
 
 export type ClientTables<DB extends DbSchema<any>> = {
@@ -101,6 +106,15 @@ export class ElectricClient<
   ) {
     super(dbName, adapter, notifier, registry)
     this.satellite = satellite
+  }
+
+  /**
+   * Connects to the Electric sync service.
+   * This method is idempotent, it is safe to call it multiple times.
+   */
+  async connect(): Promise<ConnectionWrapper> {
+    const connectionPromise = this.satellite.connectWithBackoff()
+    return { connectionPromise }
   }
 
   // Builds the DAL namespace from a `dbDescription` object
--- a/clients/typescript/src/client/model/client.ts
+++ b/clients/typescript/src/client/model/client.ts
@@ -5,12 +5,8 @@ import { Row, Statement } from '../../util'
 import { LiveResult, LiveResultContext } from './model'
 import { Notifier } from '../../notifiers'
 import { DatabaseAdapter } from '../../electric/adapter'
-import {
-  ConnectionWrapper,
-  GlobalRegistry,
-  Registry,
-  Satellite,
-} from '../../satellite'
+import type { ConnectionWrapper } from '../../satellite'
+import { GlobalRegistry, Registry, Satellite } from '../../satellite'
 import { ShapeManager } from './shapes'
 
 export type ClientTables<DB extends DbSchema<any>> = {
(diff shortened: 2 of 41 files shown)


kevin-dp:
As discussed with @balegas, i modified `electrify` to make the `AuthConfig` optional.
If an `AuthConfig` is provided it should only contain a client ID.
If no client ID is provided, the Satellite process generates one.

```ts
const electric = await electrify(conn, schema)
```

After electrifying the database, we can now connect to Electric. Now is the time to provide the JWT:
```ts
const token = /* fetch your JWT somehow */
await electric.connect(token)
```

clients/typescript/src/satellite/process.ts
```diff
@@ -216,10 +218,10 @@ export class SatelliteProcess implements Satellite {
     }
 
     const clientId =
-      authConfig.clientId && authConfig.clientId !== ''
+      authConfig?.clientId && authConfig.clientId !== ''
         ? authConfig.clientId
         : await this._getClientId()
-    await this._setAuthState({ clientId: clientId, token: authConfig.token })
+    await this._setAuthState({ clientId: clientId }) //, token: authConfig.token })
```
icehaunter, line 224: nitpick: `await` on a non-promise expression + comment leftovers

clients/typescript/src/client/model/client.ts
```diff
@@ -71,6 +72,17 @@ export class ElectricClient<
     this.satellite = satellite
   }
 
+  /**
+   * Connects to the Electric sync service.
+   * This method is idempotent, it is safe to call it multiple times.
+   * @param token - The JWT token to use to connect to the Electric sync service.
+   */
+  async connect(token: string): Promise<ConnectionWrapper> {
```
icehaunter, line 80: I don't see a reason to keep this indirection around (other than code compat, but we're alpha, right?). This promise was returned from `Satellite#start` because the pre-connect method was async, and connection itself is separately async, and we wanted to return from the setup immediately instead of waiting for a full connection. This signature makes no sense here on an explicit `connect` method, so IMO `ConnectionWrapper` type can be deleted

icehaunter (changes requested):
LGTM, with one nitpick and one type signature question to finish the cleanup

clients/typescript/src/client/model/client.ts
kevin-dp: Yep i had left it like this to make the minimal required changes. But indeed this no longer makes sense so i will fix this, thanks!

clients/typescript/src/satellite/process.ts
kevin-dp: Thanks, i think this slipped back in on a merge conflict that i wrongly resolved

samwillis:
@kevin-dp Is it worth adding a matching `.disconnect()`?
(you may already have discussed this with @balegas)

kevin-dp:
@icehaunter thanks for your review! I adressed your issues and am therefore re-requesting your review
@samwillis yep we discussed this with @balegas, a `disconnect` method is planned but as a separate PR.
The plan is to make the connectivity state read-only (currently you can disconnect by putting it on `disconnected`) and introduce the `disconnect` method.

icehaunter (approved):

.changeset/chilled-avocados-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+"electric-sql": patch
```
alco, line 2: This looks like a breaking change to me.

People with existing apps will start getting type errors about unknown `token` field in the auth config. And their app will stop syncing until they add an explicit `.connect()` call. Or am I missing something?

kevin-dp: Yes that's true. But i believe the philosophy with versioning for now is to indicate compatibility with Electric, not to indicate breaking changes?

alco: That's correct.

I'm worried about the documentation burden of this change. Not only the API reference, but the Quickstart and maybe a few other pages need to be updated.

Looking at our examples, most of them have `electric-sql` pinned to `"^0.7.0"` so they won't be affected even if users update the deps. Only `web-wa-sqlite` has `electric-sql` pinned to `"^0.9.0"`. Perhaps we should update that version requirement to just `"0.9.0"` to avoid unexpected updates.

@balegas @samwillis We need to have a way to notify users about such changes that require action on the user's part. Any ideas? I think printing a warning into the browser console would be the surest way for a developer to notice that their app no longer works as before.

samwillis (approved):
LGTM

I think @alco is right about a warning. Maybe we set a timeout for 500ms after an electric client is initiated, and if `connect` isn't called emit a warning.

kevin-dp:
@samwillis @alco Not sure about adding a timer to our code to signal that the user may have missed this breaking change. Up till when will we keep that timer in our codebase? Till the next minor release?

kevin-dp:
@samwillis as pointed out by @alco in a private conversation, the auth configuration that is passed to the `electrify` call no longer accepts a token. So for those cases where we break it they will see a type error. So there should be no need for an additional warning. Can i proceed and merge? @samwillis @alco

alco:
@kevin-dp Sounds good :+1:

samwillis:
Go for it 👍

PR #829, opened 2024-01-15
chore(electric): Validate enum values for electrified tables

alco:


components/electric/priv/sql_function_templates/electrify/__validate_table_column_types.sql.eex
```diff
@@ -18,31 +18,58 @@ CREATE OR REPLACE PROCEDURE <%= schema() %>.__validate_table_column_types(table_
...
+        IF _enum_oid IS NOT NULL THEN
+            SELECT array_agg(format('%L', enumlabel) ORDER BY enumsortorder)
+            INTO _invalid_enum_vals
+            FROM pg_enum
+            WHERE enumtypid = _enum_oid AND enumlabel !~* '^[a-z][a-z_0-9]*$';
+
+            IF _invalid_enum_vals != '{}' THEN
+                _invalid_cols =
+                    array_append(_invalid_cols,
+                                 format('%I %s (enum type %s contains unsupported values %s)',
```
magnetised, line 49: I think this constraint on "valid" enum values is odd from a developer perspective, I personally would be very confused. so I think this error message should include a definition of what a supported value looks like. I would be happy with the exactness of the validating regexp, but some english equivalent would do.

components/electric/test/electric/postgres/extension_test.exs
```diff
@@ -457,16 +457,20 @@ defmodule Electric.Postgres.ExtensionTest do
 
     test_tx "rejects invalid column types", fn conn ->
       assert [
+               {:ok, [], []},
                {:ok, [], []},
                {:error, {:error, :error, _, :raise_exception, error_msg, _}}
              ] =
                :epgsql.squery(conn, """
+               CREATE TYPE badenum AS ENUM ('1circle', '_square', 'hello world');
```
magnetised, line 465: could we add a valid value so we're testing that the error message is only listing the invalid values

magnetised (approved):
nice, though I think we need to hold developers' hands a little more around this concept of "valid enum values"

components/electric/priv/sql_function_templates/electrify/__validate_table_column_types.sql.eex
alco: Good point. I will add a URL to https://deploy-preview-93--electric-sql-website.netlify.app/docs/usage/data-modelling/types#supported-data-types at the end of this error message.

PR #830, opened 2024-01-15
fix (client): use WebSocketWeb as the default socket for capacitor

kevin-dp:
This PR fixes the default socket that is used by the capacitor driver to be the `WebSocketWeb` socket.
Note that the cordova driver is still using `MockSocket` as the default.
Not sure what should be the default for the Cordova driver? And whether it's worth spending any time in this since it is not used.

### DIFF
--- a/.changeset/happy-emus-push.md
+++ b/.changeset/happy-emus-push.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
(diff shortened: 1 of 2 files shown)


balegas (approved):
We have fixed in another driver some time ago.

I love this one :)

FYI  I've added this as a blocker for the release @kevin-dp

PR #833, opened 2024-01-15
chore(electric): Consistent Proxy error numbering and formatting

alco:
Follow-up to https://github.com/electric-sql/electric/pull/819.


magnetised (approved):
Neat, thanks. 👍🏻

PR #834, opened 2024-01-15
fix(client) Fix bugs in the CLI

samwillis:
Fixes:
- `generate --with-migrations` now runs in a ephemeral container on random ports and random container name
- #825 Correctly handles ssl for db urls
- Allow the pg proxy to be on an alternative host to the sync service (required for the proxy tunnel)
- configurable container name
- correctly handles `PG_PROXY_PORT` with `http` or `http:123` value

### DIFF
--- a/clients/typescript/src/cli/docker-commands/command-psql.ts
+++ b/clients/typescript/src/cli/docker-commands/command-psql.ts
@@ -4,7 +4,7 @@ import {
   getConfig,
   GetConfigOptionsForGroup,
 } from '../config'
-import { buildDatabaseURL } from '../utils'
+import { buildDatabaseURL, parsePgProxyPort } from '../utils'
 import { dockerCompose } from './docker-utils'
 
 export function makePsqlCommand() {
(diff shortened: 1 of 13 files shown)


clients/typescript/src/cli/docker-commands/command-psql.ts
```diff
@@ -20,23 +20,24 @@ export function makePsqlCommand() {
   return command
 }
 
-export function psql(opts: GetConfigOptionsForGroup<'proxy'>) {
+export function psql(
+  opts: GetConfigOptionsForGroup<'proxy'> & GetConfigOptionsForGroup<'electric'>
```
kevin-dp, line 24: Would be nice if we could write `GetConfigOptionsForGroup<'proxy' | 'electric'>`

clients/typescript/src/cli/docker-commands/command-psql.ts
```diff
@@ -20,23 +20,24 @@ export function makePsqlCommand() {
...
+) {
```
kevin-dp, line 25: Is the TODO on the line below still relevant?

kevin-dp (approved):
Looks all good to me 👍

clients/typescript/src/cli/docker-commands/command-psql.ts
samwillis: Good spot, changed

clients/typescript/src/cli/docker-commands/command-psql.ts
samwillis: Hmm, not really. I don't think we need to make the `electric-sql psql` work when the pg isn't inside an electric-sql managed container, far to much complexity. Removed the TODO.

PR #838, opened 2024-01-16
chore(docs): Update Fly deployment doc

alco:
Preview: https://deploy-preview-101--electric-sql-website.netlify.app/docs/deployment/fly

### DIFF
--- a/docs/deployment/fly.md
+++ b/docs/deployment/fly.md
@@ -28,7 +28,7 @@ Save the following snippet into a file named `fly.toml` somewhere on your comput
 app = "electric-on-fly-test-app"
 
 [build]
-  image = "electricsql/electric"
+  image = "electricsql/electric:latest"
 
 [env]
   AUTH_MODE = "insecure"
--- a/docs/deployment/fly.md
+++ b/docs/deployment/fly.md
@@ -176,3 +176,18 @@ Now go back to [Configure your Fly app](#configure-your-fly-app) above and modif
 
 Fly Postgres does not support encrypted database connections inside its private 6PN network while Electric enforces encryption by default. The above `DATABASE_REQUIRE_SSL` setting changes Electric's behaviour by allowing it to fallback to using unencrypted connections.
 
+## Connecting the client app to Electric running on Fly
+
+If you don't already have an Electric-enabled client app, follow our <DocPageLink path="quickstart" /> guide or clone our [Basic Items](/docs/examples/basic) example app and go through its README. In either case, skip the steps that show how to run Postgres and Electric locally.
+
+Create a file named `.env.local` inside your client app's root directory with the following contents so that [`electric-sql commands`](/docs/api/cli) know how to connect to your instance of Electric sync service running on Fly:
+
+```shell
+ELECTRIC_SERVICE=https://electric-on-fly-test-app.fly.dev/
+
+# This should be the same password as the one used
+# for PG_PROXY_PASSWORD in your fly.toml
+ELECTRIC_PG_PROXY_PASSWORD=proxy_password
+```
+
+If you're wondering why we don't set the `ELECTRIC_DATABASE_URL` variable here, that's because client commands only connect to the Migrations proxy which is a component of Electric sync service. Learn more about the Migrations proxy's role and how it fits into the bigger picture in our <DocPageLink path="deployment/concepts#migrations-proxy" /> guide.
--- a/docs/deployment/fly.md
+++ b/docs/deployment/fly.md
@@ -56,13 +56,15 @@ app = "electric-on-fly-test-app"
 [env]
   AUTH_MODE = "insecure"
   DATABASE_URL = "postgresql://..."
-  # When using Fly Postgres, uncomment the below. Fly Postgres does not
-  # support encrypted connections when connecting over its private 6PN network.
+  # When using Fly Postgres, uncomment the config line below.
+  # Fly Postgres does not support encrypted connections
+  # inside its private 6PN network.
   #DATABASE_REQUIRE_SSL = "false"
   ELECTRIC_WRITE_TO_PG_MODE = "direct_writes"
   PG_PROXY_PASSWORD = "proxy_password"
 
-# The main Internet-facing service of Electric to which clients will be connecting.
+# The main Internet-facing service of Electric
+# to which clients will be connecting.
 [http_service]
   internal_port = 5133
   force_https = true
@@ -74,7 +76,8 @@ app = "electric-on-fly-test-app"
     method = "GET"
     path = "/api/status"
 
-# Service definition for the migrations proxy that should be accessible on a separate TCP port.
+# Service definition for the migrations proxy that runs
+# on a separate TCP port.
 [[services]]
   protocol = "tcp"
   internal_port = 65432
--- a/docs/deployment/fly.md
+++ b/docs/deployment/fly.md
@@ -7,8 +7,8 @@ sidebar_position: 40
 
 You can deploy ElectricSQL to [Fly.io](https://fly.io) and use either an external Postgres or an instance of Fly Postgres as the database for the sync service. This guide covers both scenarios:
 
-  1. Deploying Electric sync service and connecting it to an external Postgres.
-  2. Configuring Fly Postgres and Electric sync service to connect over Fly's private network.
+  1. deploying the Electric sync service and connecting it to an external Postgres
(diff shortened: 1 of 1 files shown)


docs/deployment/fly.md
```diff
@@ -137,52 +168,58 @@ Connection to Postgres is up!
...
-### Apply migrations
+Open `examples/web-wa-sqlite/.env` and replace the URL on the `ELECTRIC_SERVICE=http://localhost:5133` line with your Fly app's URL:
 
-Electric can work alongside any tooling you use to manage database migrations with. See the <DocPageLink path="integrations/backend" /> section of the docs for an overview of the most popular frameworks.
+```
+ELECTRIC_SERVICE=https://electric-on-fly-test-app.fly.dev/
+```
 
-In this demo we'll use `@databases/pg-migrations` as it's already included in the basic example. Make sure you have installed all of the dependencies by running `npm install` once.
+### Apply migrations
```
thruflo, line 177: I think this section should link to the Migrations proxy section of the concepts page. Explain that migrations are applied via the proxy, that the URL of the proxy is the URL of your electric service + the exposed TCP port. And have a similar example for what the value of `ELECTRIC_PROXY` is in the code sample.

Otherwise it's not clear where the ELECTRIC_PROXY value comes from.

I would also suggest the migrations page is a better starting point than the backend integrations for understanding how to apply migrations. So maybe link to that first and then the backend integrations.

docs/deployment/fly.md
```diff
@@ -137,52 +168,58 @@ Connection to Postgres is up!
...
 ### Generate a type-safe client
 
-Now that the database has one electrified table, we can [generate a type-safe client](../usage/data-access/client.md) from it. Use the same database connection URL as in the previous step but change the username to `prisma` (this is required for the schema introspection to work correctly).
+Now that the database has one electrified table, we can [generate a type-safe client](../usage/data-access/client.md) from it:
 
 ```shell
-$ npx electric-sql generate
-      --service https://electric-on-fly-test-app.fly.dev
-      --proxy postgresql://prisma:proxy_password@electric-on-fly-test-app.fly.dev:65432/postgres
+$ npm run client:generate
```
thruflo, line 199: Do we not want to use the CLI directly? Where does the --service URL come from -- again, the env var in the example is example specific, not the primary DX.

docs/deployment/fly.md
```diff
@@ -137,52 +168,58 @@ Connection to Postgres is up!
...
 ### Start the app!
 
-Now you should have everything ready to start the web app and have it connected to the Electric sync service deployed on Fly.
+Now you should have everything ready to start the web app and have it connect to the Electric sync service running on Fly:
 
 ```shell
-$ ELECTRIC_URL='wss://electric-on-fly-test-app.fly.dev' \
-  SERVE=true \
-  npm run build
+$ npm run dev
```
thruflo, line 214: As per my comments on the CLI command vs example, I think there's a challenge here where this doc is half reference, half walkthrough. Ideally, I think we need both, so a terse primary deployment page for Fly that does not replicate all the generate use instructions.

However, perhaps best to ship as more walkthrough as is and then loop back later. It is a positive that this flow ends in a successfully deployed app.

docs/deployment/fly.md
```diff
@@ -11,15 +11,37 @@ The app config needs to include an `http_service` with internal port `5133` and
 
 The environment variables used by Electric are described in <DocPageLink path="api/service" />.
 
-## Deploying Electric
+## Postgres with logical replication
+
```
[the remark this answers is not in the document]
alco: I think it would be excessive to include every possible option in the guide if the aim is to get users up and running as quickly and effortlessly as possible.

I think the proxy tunnel should be used as a last resort. Connecting to the proxy directly is preferable because there are fewer moving parts on that path and smaller chance of hitting a problem.

docs/deployment/fly.md
```diff
@@ -62,6 +87,12 @@ app = "electric-on-fly-test-app"
     handlers = ["pg_tls"]
 ```
 
+:::info Secrets and environment variables
+[Secrets](https://fly.io/docs/reference/secrets/) allow sensitive values, such as credentials, to be passed securely to your Fly app. The secret is encrypted and stored in a vault. It is made available to the app as an environment variable.
+
+We're not using secrets in this example to keep things short and simple. As soon as you're ready to take your Fly app from development to production, make sure to replace the `DATABASE_URL`, `PG_PROXY_PASSWORD` and `AUTH_JWT_KEY` (this latter one is required in the [secure authentication mode](/docs/usage/auth/secure)) environment variables with secrets.
```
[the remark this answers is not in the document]
alco: We had a separate step for setting secrets in the [old guide](https://github.com/electric-sql/electric/commit/a69cf5e4f97eb1bbc1899db28d729e8245bf8f9b?short_path=3046ae3#diff-3046ae322d0bba95f7cf5789cddb5a596f8d64b711b714427907288af9f25048). I took that out of this simplified version as non-essential.

The more callouts, tabs, folding sections we add, the more congnitive load it creates for users and more maintenance burden for us. I wanted to avoid that as much as possible since someone with more advanced needs would presumably also be able to utilizes Fly's advanced capabilities without us holding their hands.

docs/deployment/fly.md
alco: In response to this comment and the one above it, I specifically used an example app in this guide as a way to side-step setting up a web app from scratch. It makes sense to use the same instructions that the example app has in its README, in my opinion.

Other possibilities:
- omit the client setup section at all and just point people to the example app
- replace textual instructions with a short video that demonstrates going through the example app's README and executing those steps
- demonstrate the use of the new CLI directly. But then I don't get your complaints about environment variables in the .env file. It's part of the new CLI, not of example apps. One of the perks of the new CLI is that you can add `electric-sql` as a dependency to your React/Vue/Whatever app, create a `.env` file with your URLs and secrets and execute commands such as `start`, `stop`, `generate` without command-line flags.

thruflo:
I think it's fair enough to minimise the info on how to setup Fly Postgres. That's a task for the Fly docs. I do think we should signpost what Fly is good for at the top of this page and then echo the phrasing we use in the section headings.

I.e.: "You can use Fly deploy the Electric sync service and, optimally, to also deploy Postgres. This guide covers two main scenarios: (1) use Fly to deploy the Electric sync service and connect to an external Postgre and (2) use Fly to deploy both Electric and Postgres and connect them over the internal network."

Then you can have those two sections as top level `##` h2s in the actor nav. Focus the content on the first one (cutting out the client section and instead ending up on showing the URLs that the client, generator and migrations proxy will connect on and linking out to the usage or deployment concepts docs for those). Then have a short second section saying "when using Fly Postgres, go here for the docs and swap out the database url like this". With a warning that Fly Postgres is changing to use Supabase (link) so you may want to wait for that.

With the migrations proxy, fair enough, given Fly supports TCP, let's just document the primary path and minimise branching.

docs/deployment/fly.md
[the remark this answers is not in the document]
thruflo: Sure, we don't have to add this.

docs/deployment/fly.md
thruflo: Yup, I suggest we omit the client setup section and just point people to the Quickstart.

alco: I agree with your last point. I almost wish we had a way to run any example app with one command. Then in our deployment guides we would be able to say something along the following lines:

> Now that the sync service is up and running, navigate to the example app, paste your deployed service's URL and proxy password into the `.env` file, run `npm install` followed by `npm run dev`.

and have all of the steps that are currently manual execute automatically: apply migrations, generate the client and start the dev server.

alco:
@thruflo I have made the changes we've discussed earlier:

  - removed a detailed walkthrough of setting up a client app
  - moved the Fly Postgres section below the primary "connect an external database" guide and extended it
  - added a short section at the bottom mentioning the Quickstart and the Basic example
  - added an overview of the two scenarios at the top
  - added a couple of mentions of the Deployment Concepts guide

You can review the actual changes starting from [this commit](https://github.com/electric-sql/electric/pull/838/commits/fbbc9567ba8118fca412528d0ce603c5176f2c0a). Click `Next` to go the next one. There are 5 new commits in total.

The preview page - https://deploy-preview-101--electric-sql-website.netlify.app/docs/deployment/fly.

thruflo (approved):
This reads great, nice work, thanks 👍

PR #842, opened 2024-01-16
chore: Add missing/forgotten changesets

alco:


magnetised:
@alco wondering if we should add an optional check that shows yellow if the PR doesn't have a changeset.

alco:
> @alco wondering if we should add an optional check that shows yellow if the PR doesn't have a changeset.

Could be done - https://github.com/changesets/changesets#integrating-with-ci

alco:
Cherry-picked each commit individually into `main`.

PR #844, opened 2024-01-17
fix(client): Fix redundant query calls in React live query hook

msfstef:
The `useLiveQuery` React implementation had redundant calls to the adapter and some other redundant calls to subscribe/unsubscribe from changes.

I removed any unnecessary state and references to simplify the binding, and enhanced a couple of tests to ensure there are no redundant query calls.

### DIFF
--- a/clients/typescript/src/frameworks/react/hooks/useLiveQuery.ts
+++ b/clients/typescript/src/frameworks/react/hooks/useLiveQuery.ts
@@ -1,6 +1,5 @@
 import {
   useContext,
-  useRef,
   useEffect,
   useState,
   useCallback,
@@ -19,6 +18,11 @@ export interface ResultData<T> {
   error?: unknown
   results?: T
   updatedAt?: Date
+}
+
+interface QueryResult<T> {
+  result: ResultData<T>
+  tablenames?: QualifiedTablename[]
 }
 
 function successResult<T>(results: T): ResultData<T> {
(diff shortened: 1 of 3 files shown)


clients/typescript/test/frameworks/react.test.tsx
```diff
@@ -47,7 +59,7 @@ const test = anyTest as TestFn<{
 test.beforeEach((t) => {
   const original = new MockDatabase('test.db')
   const adapter = new DatabaseAdapter(original, false)
-  const notifier = new MockNotifier('test.db')
+  const notifier = new MockNotifier('test.db', new EventEmitter())
```
msfstef, line 62: If no emitter is provider, it falls back to a global emitter causing tests to become dependent (some failed depending on what other tests ran in parallel)

clients/typescript/src/frameworks/react/hooks/useLiveQuery.ts
```diff
@@ -124,91 +125,77 @@ function useLiveQueryWithQueryUpdates<Res>(
...
+  // Once we have electric, we then run the query and establish the data
+  // change notification subscription, comparing the tablenames used by the
+  // query with the changed tablenames in the data change notification
+  // to determine whether to re-query or not.
+  //
+  // If we do need to re-query, then we use the saved function to reuse the query
+  useEffect(() => {
+    if (electric?.notifier == undefined) return
+    const unsubscribe = subscribeToDataChanges(electric?.notifier)
+    return unsubscribe
```
samwillis, line 197: I would normally return the function result directly, but I like that this makes it more explicit that we are using the unsubscribe function from `subscribeToDataChanges` as the cleanup function for the effect.

clients/typescript/src/frameworks/react/hooks/useLiveQuery.ts
```diff
@@ -124,91 +125,77 @@ function useLiveQueryWithQueryUpdates<Res>(
...
+      const handleChange = (notification: ChangeNotification): void => {
+        // Reduces the `ChangeNotification` to an array of namespaced tablenames,
+        // in a way that supports both the main namespace for the primary database
+        // and aliases for any attached databases.
+        const changedTablenames = notifier.alias(notification)
+        if (
+          relevantTablenames &&
+          hasIntersection(relevantTablenames, changedTablenames)
+        ) {
+          runLiveQuery().then(updateState)
```
samwillis, line 172: I normally prefer `await` and would avoid `.then()` but what you have here is cleaner.

samwillis (approved):
This looks really great, massively easer to follow.

Have you fully tested this with your recipes app?

clients/typescript/src/frameworks/react/hooks/useLiveQuery.ts
```diff
@@ -124,91 +125,77 @@ function useLiveQueryWithQueryUpdates<Res>(
...
-    }
-  }, [electric, tablenamesKey.current, tablenames.current, runLiveQuery])
+  // Once we have electric, we then run the query and establish the data
+  // change notification subscription, comparing the tablenames used by the
+  // query with the changed tablenames in the data change notification
+  // to determine whether to re-query or not.
+  //
+  // If we do need to re-query, then we use the saved function to reuse the query
+  useEffect(() => {
+    if (electric?.notifier == undefined) return
```
kevin-dp, line 195: nitpick: can we use `===` instead of `==`

kevin-dp (approved):
Great improvement! Thanks 🙏

clients/typescript/src/frameworks/react/hooks/useLiveQuery.ts
msfstef: I wanted to unify the subscription to changes code so that the "first query" and subsequent updates as well as the updating/cancelling methods are relatively compact

clients/typescript/src/frameworks/react/hooks/useLiveQuery.ts
msfstef: I try to avoid it to not have dangling async effects, but in this case the method is itself wrapped in a try/catch block and making the caller async would muddle things up in terms of ordering of operations (subscribing/cancelling)

msfstef:
> This looks really great, massively easer to follow.
> 
> Have you fully tested this with your recipes app?

I have not tested this in the recipes app, it's not merged on main yet, but I've enhanced the existing tests a bit to check for number of calls and ensured nothing else was affected.

I think the next step would be to refactor the "subscribeToDataChanges" to a more generic, framework independent interface and then reuse it to create bindings for the various reactive frameworks.

PR #851, opened 2024-01-18
fix(client) Upgrade Prisma dep to 5.2 in the client, Fixes VAX-1524

samwillis:
This seems to fix VAX-1524, but may need more testing with Prisma schemas.

### DIFF
--- a/.changeset/wise-geckos-cough.md
+++ b/.changeset/wise-geckos-cough.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Upgrade Prisma dependency to 5.2 in the client, Fixes VAX-1524
--- a/clients/typescript/package.json
+++ b/clients/typescript/package.json
@@ -137,7 +137,7 @@
   },
   "dependencies": {
     "@electric-sql/prisma-generator": "workspace:*",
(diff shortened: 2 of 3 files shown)


kevin-dp:
Do you know why a newer version fixes this?
Was it a problem in Prisma itself and they solved it in a later version?

samwillis:
I don't know why this fixes it, and that not ideal.

`NodeAPIQueryEngineLibrary.dmmf` (in the error message) was removed from Prisma in a refactor here: https://github.com/prisma/prisma/commit/8ad76c3e04bd3d037e2258b6219d062d863ee00c#diff-8ae842291b5f8da88622203ebd017c3aee4ee8976c9fdcf4423c82cd8516b622L158

image

That landed in 4.10.

samwillis:
Hmm, actually we had a Prisma version mismatch, I bet that explains it!

```
 "@prisma/client": "4.12.0",
 "prisma": "4.8.1",
```

kevin-dp:
aha makes sense! Thanks for looking into this 💯

kevin-dp (approved):
Approving this PR. I'm fine with pinning the client to 5.2 as this does not mean the user's project must use the same version.

PR #854, opened 2024-01-18
chore (e2e tests): fix flaky test

kevin-dp:
This PR modifies e2e test 03.10 which is flaky and regularly fails on CI.
I figured out that we were matching the Node repl before matching some expected string:

```
[invoke electrify_db "originalDb" "electric_1" 5133 "[]"]
?$node # <-- this match is causing trouble and not needed
-no previous LSN
?starting replication with lsn:
?\[rpc\] send: #SatInStartReplicationReq\{lsn: [a-zA-Z0-9=]+,
??[rpc] recv: #SatInStartReplicationResp
```

But sometimes NodeJS print the expected string before returning so after matching `?$node` it can't match `?starting replication with lsn:` anymore as it was already printed. I mainly had this issue on the CI. Let's see if this fixes it.


alco (approved):
Brilliant!

PR #855, opened 2024-01-18
fix(client) Correctly pass the CONTAINER_NAME to the cli "status" command

samwillis:
This ensures the status command correctly shows the containers for the current project.

### DIFF
--- a/clients/typescript/src/cli/docker-commands/command-status.ts
+++ b/clients/typescript/src/cli/docker-commands/command-status.ts
@@ -7,7 +7,7 @@ export function makeStatusCommand() {
     .description(
       'Show status of the ElectricSQL sync service docker containers'
     )
-    .action(async () => {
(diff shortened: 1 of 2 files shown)


clients/typescript/src/cli/docker-commands/command-status.ts
```diff
@@ -1,14 +1,18 @@
...
+import { getConfig, Config } from '../config'
 import { dockerCompose } from './docker-utils'
 
 export function makeStatusCommand() {
   return new Command('status')
     .description(
       'Show status of the ElectricSQL sync service docker containers'
     )
-    .action(status)
+    .action(async () => {
```
kevin-dp, line 10: Can probably remove `async` as the function does not use async/await.

kevin-dp (approved):

PR #856, opened 2024-01-18
chore(client) Pull latest images for tag when running cli 'start' command

samwillis:
If the pull fails, for example the user is offline, it will continue anyway.

Thanks to @davidmartos96 for the suggestion.

### DIFF
--- a/.changeset/chatty-kings-beg.md
+++ b/.changeset/chatty-kings-beg.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Pull latest images for tag when running cli 'start' command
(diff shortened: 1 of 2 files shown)


alco:
I'd find it suprising in a bad way if I had a project that used to work before and then one sunny day it would stop working due to the automatic upgrade of one of its components.

I'd argue against this change because such behaviour would be unlike anything else developers have to deal with: Docker doesn't pull newer versions of images, npm doesn't try to install new dependency versions when you run `npm run start`, Git doesn't pull the branch from origin when it is checked out locally and so on.

samwillis:
@alco hmm, I think you could be right. Maybe we need a "pull-docker" command... (we want to keep the cli small though!!)

PR #859, opened 2024-01-19
fix(electric): Implement support for AuthenticationMD5Password upstream auth method

alco:
Fixes VAX-1553.

@magnetised I would appreciate any guidance on how to best test support for specific upstream messages in the proxy. Maybe we should also unit-test the SSL upgrade that I recently implemented there.

### DIFF
--- a/components/electric/lib/electric/postgres/proxy/upstream_connection.ex
+++ b/components/electric/lib/electric/postgres/proxy/upstream_connection.ex
@@ -187,6 +187,15 @@ defmodule Electric.Postgres.Proxy.UpstreamConnection do
     %{state | sasl: nil}
   end
 
+  defp handle_backend_msg(%msg_type{} = msg, _state)
+       when msg_type in [M.AuthenticationKerberosV5, M.AuthenticationGSS, M.AuthenticationSSPI] do
+    error_msg = "Proxy's upstream connection requested unsupported authentication method:"
+    error_val = inspect(msg, pretty: true)
+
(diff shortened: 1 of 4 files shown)


components/electric/lib/electric/postgres/proxy/upstream_connection.ex
```diff
@@ -206,4 +226,17 @@ defmodule Electric.Postgres.Proxy.UpstreamConnection do
...
+    :crypto.hash(:md5, iodata) |> bin_to_hex()
+  end
+
+  defp bin_to_hex(""), do: ""
+
+  defp bin_to_hex(<<hi::4, lo::4>> <> rest) do
+    <<hex_char(hi), hex_char(lo)>> <> bin_to_hex(rest)
+  end
+
+  defp hex_char(num) when num in 0..9, do: ?0 + num
```
icehaunter, line 240: why not `Integer.to_string(num, 16)`? Speed hardly matters here for 2 chars?

icehaunter: Actually, sorry, the entire `bin_to_hex` is replaceable with [`Base.encode16/2`](https://hexdocs.pm/elixir/Base.html#encode16/2)

icehaunter (approved):
LGTM with one nitpick, we should probs use a built-in functions

components/electric/lib/electric/postgres/proxy/upstream_connection.ex
alco: Perfect! Thanks!

PR #861, opened 2024-01-19
chore: Rename "prisma" username to "electric-generator"

alco:
Closes VAX-1510.


kevin-dp (approved):
LGTM

icehaunter:
> At this point, merging it would do more harm than good. People will need to update both the server and the client at the same time to keep the client-side CLI working correctly with the new username. There are probably other potential breakages that weren't a concern at the time.

as per @alco, so closing for now

PR #862, opened 2024-01-19
fix(cli): Map http(s) scheme in service URL to ws(s) when starting proxy-tunnel

alco:
Fixes VAX-1544.

### DIFF
--- a/clients/typescript/src/cli/tunnel/command-proxy-tunnel.ts
+++ b/clients/typescript/src/cli/tunnel/command-proxy-tunnel.ts
@@ -80,7 +80,7 @@ function log(...args: any[]) {
 }
 
 function mapHttpToWebSocketInUrl(urlString: string) {
-  let url = new URL(urlString)
+  const url = new URL(urlString)
   switch (url.protocol) {
     case 'https:': {
       url.protocol = 'wss:'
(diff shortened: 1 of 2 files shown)


samwillis (approved):
👍

clients/typescript/src/cli/tunnel/command-proxy-tunnel.ts
```diff
@@ -17,10 +17,14 @@ export function makeProxyTunnelCommand() {
...
       const localPort = parsePort(opts.localPort)
-      proxyTunnel({
-        serviceUrl: config.SERVICE,
-        localPort,
-      })
+
+      try {
+        const serviceUrl = mapHttpToWebSocketInUrl(config.SERVICE)
+        proxyTunnel({ serviceUrl, localPort })
+      } catch (error) {
```
alco, line 24: I had to add this try-catch block because without it the command would fail with this unfriendly message:
```
$ npx electric-sql proxy-tunnel
node:internal/process/promises:289
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Invalid URL scheme in ELECTRIC_SERVICE: httpx://electric-sync-service-ziput.ondigitalocean.app/".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

PR #865, opened 2024-01-22
feat (client): read-only connectivity state and disconnect method

kevin-dp:
This PR makes the connectivity state read-only.
To connect or disconnect the client, the user can now call the `connect` and `disconnect` methods on the `ElectricClient` that is returned from `electrify`.

I added an e2e test that checks that `connect`/`disconnect` work as expected.

Fixes VAX-1474 and VAX-1550.


clients/typescript/src/satellite/process.ts
```diff
@@ -724,6 +690,13 @@ export class SatelliteProcess implements Satellite {
...
   }
 
+  /**
+   * @returns True if a JWT token has been set previously. False otherwise.
+   */
+  hasToken(): boolean {
+    return this._authState?.token !== undefined
+  }
+
   async connectWithBackoff(): Promise<void> {
```
msfstef, line 700: Connecting with a backoff here creates a "waiter" that is resolved once the connection is established (or rejected when connection fails), and the `connect` public API calls this satellite method.

I _think_ that it's possible for someone to call `connect`, and while the `initializing` waiter is unresolved and a connection is being established, the `disconnect` method is called, disconnecting the client temporarily, but without interrupting the `connectWithBackoff` call. So after a `disconnect` call it's still possible for the client to connect because of a previous `connect` call that's still trying.

Am I missing something, or should we be somehow "interrupting" the backoff call (or making it contingent on `initializing` not having been rejected, e.g.:)
```
await backOff(async () => {
    if (this.initializing?.finished()) return
    ... blah blah ...

disconnect(error?: SatelliteError): void {
    this.client.disconnect()
    this._notifyConnectivityState('disconnected', error)
    this._initializing?.reject( ... connection attempt interrupted error .... )
  }

```

clients/typescript/src/frameworks/react/hooks/useConnectivityState.ts
```diff
@@ -73,26 +63,7 @@ const useConnectivityState: HookFn = () => {
...
-
-    setState(nextState)
-  }
-
-  return {
-    connectivityState: state,
-    setConnectivityState: setState,
-    toggleConnectivityState: toggleState,
-  }
+  return state
```
msfstef, line 66: Should we perhaps return `{ connectivityState: state }` in order minimise how much we break any clients who were using this as a read-only hook?

If we want to avoid a breaking change we could also leave the hook as is, make it use the new API from the `electric` instance internally, deprecate this hook, and add a new hook, but might not be worth it for what is probably a relatively low-usage feature so feel free to ignore this.

docs/api/clients/typescript.md
```diff
@@ -147,6 +150,13 @@ In addition, one can execute raw SQL queries using the `electric.db.raw` and `el
 Raw queries should be used with caution as they are unchecked and may cause the sync service to stop if they are ill-formed.
 Therefore, only use raw queries for features that are not supported by our regular API.
 
+## Connectivity methods
+
+The Electric client provides three connectivity methods:
+- `connect(token?: string)`: connects to Electric using the provided token. Can be used to reconnect to Electric in case the connection breaks; if the token is not provided the previous one is used.
+- `disconnect()`: disconnects from Electric. Can be used to go into offline mode.
```
msfstef, line 157: nit: since AFAIK we don't have a clear definition of offline mode, might be worth clarifying that this means that no data is replicated/sent upstream while disconnected?

clients/typescript/src/satellite/index.ts
```diff
@@ -65,8 +65,10 @@ export interface Satellite {
 
   start(authConfig: AuthConfig): Promise<void>
   stop(shutdown?: boolean): Promise<void>
-  setToken(token: string): void
+  setToken(token?: string): void
+  hasToken(): boolean
```
msfstef, line 69: question: would a failed connection set the token to "undefined" or would it leave the invalid token there? i.e. would `hasToken` always return true once a token has been provided even if it ceases to be valid?

msfstef (commented):
Looks good! Left a couple of questions and comments

docs/api/clients/typescript.md
kevin-dp: Not sure we need to explicitly tell this, it seems quite obvious that no more data is exchanged with Electric if you explicitly disconnect from Electric.

clients/typescript/src/frameworks/react/hooks/useConnectivityState.ts
samwillis: Along with the braking changes in #828, I'm not worried about this. It's going to be a lesser used API. We just document the change.

clients/typescript/src/satellite/process.ts
kevin-dp: This could indeed happen, but perhaps that is to be expected?
At the time the client called `disconnect` they were not connected in the first place.
So i'm not sure what you would expect to happen at that point.
Having `disconnect` cancel the waiter would be a possibility.

kevin-dp: Consensus seems to be that `disconnect` should cancel the waiter. Will implement it as per your suggestion, thanks!

kevin-dp: I pushed a commit that does this. Will ask @icehaunter to review that commit because there are some subtleties here wrt promise rejections and them being caught/uncaught. I added a unit test to check that `disconnect` cancels the connection.

msfstef (approved):
Looks good, nice work!

Just to confirm, from what I could tell `disconnect` synchronously cancels the connection and cleans up resources so there's no case of a user calling `disconnect` and a subsequent `connect` and there being a race condition where it ends up cancelling the connect call, right? (basically the reverse case of the client connecting after a `disconnect` call)

kevin-dp:
> Looks good, nice work!
> 
> Just to confirm, from what I could tell `disconnect` synchronously cancels the connection and cleans up resources so there's no case of a user calling `disconnect` and a subsequent `connect` and there being a race condition where it ends up cancelling the connect call, right? (basically the reverse case of the client connecting after a `disconnect` call)

Indeed, `disconnect` is synchronous :-)

clients/typescript/src/satellite/index.ts
kevin-dp: Indeed, the token is not removed even if we would fail to connect because e.g. the token is invalid.

clients/typescript/src/satellite/process.ts
```diff
@@ -819,9 +799,29 @@ export class SatelliteProcess implements Satellite {
...
+        ? new SatelliteError(
+            SatelliteErrorCode.CONNECTION_CANCELLED_BY_DISCONNECT,
+            `Connection cancelled by 'disconnect'`
+          )
+        : undefined)
     this.client.disconnect()
-    this._notifyConnectivityState('disconnected', error)
+    this._notifyConnectivityState('disconnected', opts?.error)
+    if (opts?.issuedByClient) {
+      this.cancelConnectionWaiter(error!)
```
icehaunter, line 823: I'm not a fan off this call signature

Reasoning and questions:
1. If `disconnect` is meant to always be called with at least one key, then `{ error: SatelliteError } | { issuedByClient: true }` argument type seems more precise
2. I would appreciate a comment on why  we only call `cancelConnectionWaiter` only if it's issued by client. If we can call it always, then we should and then we can simplify the call signature down to just `error?: SatelliteError` without wrapping and pass cancelled by disconnect expliticly in the only place of call (tests can duplicate, nbd)
3. why `_notifyConnectivityState` doesn't use built `error` variable?

kevin-dp: 1. Sometimes it's called internally in the process without any arguments, sometimes it's called internally with an error, and sometimes it's called (indirectly) by the client in which case we pass only the `issuedByClient` flag. Would you prefer this to be explicit, like so: `opts?: { error: SatelliteError } | { issuedByClient: true }` ? The straightforward signature would be `error?: SatelliteError, issuedByClient?: true` but then the (indirect) call by the client needs to provide `undefined` for the error. Perhaps that's fine as it affects only 1 call. Another way is to have it overloaded, i.e. the type be `errorOrClientFlag: SatelliteError | true` and then check whether it is an error or a boolean. wdyt?

2. The code as it was before this PR, would only reject the promise in `connectionWithBackoff` if we failed to connect after X retries. I extended the code to cancel the retries and reject the promise if the user explicitly disconnects while we're still trying to connect. If we always call `cancelConnectionWaiter` to reject the promise then the behavior becomes different from before, because in the process there can be internal calls to `disconnect` which before would not have rejected that promise. I tried this and some unit tests then start throwing `uncaught promise rejection` errors. But perhaps those are due to the unit test not having a catch on the call to connect. Still, it would mean a change in behavior (wrt what we had until now) which we need to agree on if that is what we want?

3. Yep can use `error`

icehaunter: I would honestly split the signature then. Make `disconnect(error?: SatelliteError)` as-is and instead of `cancelConnectionWaiter`, do `clientDisconnect()` that will build the error, call `disconnect`, and then  unconditionally do what's done in `cancelConnectionWaiter` code. Cleaner, more explicit, keeps the behaviour as-is

PR #867, opened 2024-01-22
feat(client): Add expo-sqlite/next driver

msfstef:
Adds the [expo-sqlite (next)](https://docs.expo.dev/versions/latest/sdk/sqlite-next/) driver on top of the regular one.

I can update the docs as well but I was unsure if we wanted it as a separate entry to the regular Expo one or if I should just add an additional section for using `next` instead of the regular one.

Also, is there a process/tooling in place for us to do a proper test for some of these drivers (e.g. spin up android/ios machines in CI and run basic e2e tests) to ensure they work?

### DIFF
--- a/clients/typescript/src/drivers/expo-sqlite-next/database.ts
+++ b/clients/typescript/src/drivers/expo-sqlite-next/database.ts
@@ -2,5 +2,5 @@ import { SQLiteDatabase } from 'expo-sqlite/next'
 
 export type Database = Pick<
   SQLiteDatabase,
-  'getAllAsync' | 'runAsync' | 'databaseName'
+  'getAllAsync' | 'withTransactionAsync' | 'runAsync' | 'databaseName'
 >
--- a/clients/typescript/src/drivers/expo-sqlite-next/database.ts
+++ b/clients/typescript/src/drivers/expo-sqlite-next/database.ts
@@ -2,5 +2,5 @@ import { SQLiteDatabase } from 'expo-sqlite/next'
 
 export type Database = Pick<
   SQLiteDatabase,
-  'getAllAsync' | 'withTransactionAsync' | 'runAsync' | 'databaseName'
+  'getAllAsync' | 'runAsync' | 'databaseName'
 >
--- a/clients/typescript/src/drivers/expo-sqlite-next/database.ts
+++ b/clients/typescript/src/drivers/expo-sqlite-next/database.ts
@@ -1,0 +1,6 @@
+import { SQLiteDatabase } from 'expo-sqlite/next'
+
+export type Database = Pick<
+  SQLiteDatabase,
+  'getAllAsync' | 'runAsync' | 'databaseName'
+>
--- a/clients/typescript/src/drivers/expo-sqlite-next/mock.ts
+++ b/clients/typescript/src/drivers/expo-sqlite-next/mock.ts
@@ -1,0 +1,30 @@
+import { Database } from './database'
+import {
+  SQLiteBindParams,
+  SQLiteVariadicBindParams,
+  SQLiteRunResult,
+} from 'expo-sqlite/next'
+
+export class MockDatabase implements Database {
+  constructor(public databaseName: string) {}
+  getAllAsync<T>(source: string, params: SQLiteBindParams): Promise<T[]>
+  getAllAsync<T>(
+    source: string,
+    ...params: SQLiteVariadicBindParams
+  ): Promise<T[]>
+  getAllAsync(_source: unknown, _params?: unknown): Promise<any[]> {
+    return Promise.resolve([{ i: 0 }])
+  }
+
+  runAsync(source: string, params: SQLiteBindParams): Promise<SQLiteRunResult>
+  runAsync(
+    source: string,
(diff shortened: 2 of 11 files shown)


clients/typescript/src/drivers/expo-sqlite-next/mock.ts
```diff
@@ -0,0 +1,30 @@
...
+} from 'expo-sqlite/next'
+
+export class MockDatabase implements Database {
+  constructor(public databaseName: string) {}
+  getAllAsync<T>(source: string, params: SQLiteBindParams): Promise<T[]>
+  getAllAsync<T>(
+    source: string,
+    ...params: SQLiteVariadicBindParams
+  ): Promise<T[]>
+  getAllAsync(_source: unknown, _params?: unknown): Promise<any[]> {
```
kevin-dp, line 15: Can we type the method implementation as follows?
```ts
getAllAsync<T>(_source: string, _params?: unknown): Promise<T[]>
```

And similarly for `runAsync`:
```ts
runAsync(_source: string, _params?: unknown): Promise<SQLiteRunResult>
```

kevin-dp (approved):
Well done, looks good to me!
Left a minor comment.

msfstef:
> Could you take a look at the situation with the providing of a native uuid function (tagging @balegas as he has context). It used to be done with a global variable that broke very easily, and I quickly changed it a month or so back to the `setUUIDImpl` function. However, I'm not convinced we need this, JS should be plenty quick enough to generate UUIDs, we shouldn't need to provide a "native" uuid function for mobile platforms.
> 
> Same goes for the `fastestsmallesttextencoderdecoder` polyfil, it would be good to just check if that really is still needed.

I agree we should probably sort this out without arbitrary polyfills, but I think this might be best suited for a separate PR as all of our react native drivers currently do this.

In terms of this PR, is adding expo-sqlite/next as a separate driver ok? or should we replace the old one?

Additionally, this driver does not support BigInt (look at the adapter implementation) - is that ok with us? I could try and open a PR to expo to support that as well if it's not too hard.

Finally, should this PR also document the new driver or do we generally do those tasks separately?

balegas:
> I agree we should probably sort this out without arbitrary polyfills, but I think this might be best suited for a separate PR as all of our react native drivers currently do this.

Let's do the uuid stuff in separate PR if we can please. This should be a quick one. For context, it was a bad advise from someone that was worried about the precision of JS-generated uuids. I should have just followed with js-generated.

> Finally, should this PR also document the new driver or do we generally do those tasks separately?

We have the monorepo, we should do documentation in same PR and we should look after each other to make sure we update documentation when necessariy.

>  might just open a PR to add it on their side as a valid type

This also sounds good to me

msfstef:
I ended up rewriting the adapter using the base interface in order to make use of the driver's transaction APIs - using our generic adapters I was running into issues with transactions being started within transactions for this driver, I'll have a deeper look into that in another PR.

This meant I had to repeat some logic like the mutexes, but I wanted to make sure I couldn't run into issues on both Android and iOS - I also noticed we have two APIs for transactions on our adapter, would it not make sense for us to simplify it into a single one, and make the `runInTransaction` API use the core transaction API?

I also added some info in the existing Expo driver page, as I didn't think the `next` driver deserved a whole new page given its nature as a successor of the previous expo driver.

msfstef:
> Is there anything preventing us from merging this to main? Is it a question of alignment with demo release?

Demo works without this driver as it's using the regular expo one already, so not a blocker for the demo. That being said, it'd be a nice plus to highlight the new driver.

@kevin-dp has spotted a potential issue with the SerialDatabaseAdapter but it shouldn't block this.

I've been told on other PRs that we split documentation updates so I'll have to pick the documentation out of this PR, will do so shortly.

clients/typescript/src/drivers/expo-sqlite-next/database.ts
```diff
@@ -0,0 +1,6 @@
+import { SQLiteDatabase } from 'expo-sqlite/next'
+
+export type Database = Pick<
+  SQLiteDatabase,
+  'getAllAsync' | 'withTransactionAsync' | 'runAsync' | 'databaseName'
```
kevin-dp, line 5: Do we need `withTransactionAsync` ?
It only seemed to be used in the mock.

clients/typescript/src/drivers/expo-sqlite-next/mock.ts
```diff
@@ -0,0 +1,39 @@
+import { Database } from './database'
+import {
+  SQLiteBindParams,
+  SQLiteVariadicBindParams,
+  SQLiteRunResult,
+} from 'expo-sqlite/next'
+
+export class MockDatabase implements Database {
+  constructor(public databaseName: string, public fail?: Error) {}
+  withTransactionAsync(task: () => Promise<void>): Promise<void> {
```
kevin-dp, line 10: don't think we actually need this method

kevin-dp (approved):
Looks good to me.
Only need to really address the comment about the `Promise.all`.
Left some additional minor comments.

theprashant-one:
Any Eta when it is going to release?

PR #869, opened 2024-01-22
fix(cli): Move most of the electric service definition into compose.yaml

alco:
Fixes VAX-1557.

Trying to run the `start` command was failing with an error from Docker Compose on my machine:
```
$ npx electric-sql start --with-postgres
Starting ElectricSQL sync service with PostgreSQL
Docker compose config: {
[...]
}
service ports services.electric.ports.[0] is missing a target port
```

I traced it to a bug in the latest version of Docker Compose which I had installed - https://github.com/docker/compose/issues/11378.

Now, the upstream issue is slated to be resolved in the next release. But to avoid relying on an upstream fix getting applied to people's installations of Docker Compose and to ensure a smooth onboarding experience for our users, I've created this workaround that moves all service-level definitions that include environment variable interpolation from `compose-base.yaml` to `compose.yaml`.

### DIFF
--- a/clients/typescript/src/cli/docker-commands/docker/compose-base.yaml
+++ b/clients/typescript/src/cli/docker-commands/docker/compose-base.yaml
@@ -2,27 +2,7 @@ version: '3.8'
 
 services:
   electric-no-postgres:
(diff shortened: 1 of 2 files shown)


samwillis (approved):
👍

Issue #863, opened 2024-01-21
electric-sql generate stuck while using Supabase


marksteve:
### Logs

```
Generating Electric client...
Service URL: http://localhost:5133
Proxy URL: postgresql://prisma:@localhost:65432/electric
```

> \\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\[!NOTE\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\]
> `ELECTRIC_PROXY` wasn't being respected but I could at least do
>
> ```
> export ELECTRIC_DATABASE_NAME=electric
> export ELECTRIC_PG_PROXY_PASSWORD=$PG_PROXY_PASSWORD
> ```

electricsql/electric:0.9

```
02:35:17.962 pid=<0.3144.0> proxy_session_id=2 [warning] Unsupported SASL mechanism "SCRAM-SHA-256-PLUS"
02:35:18.192 pid=<0.3143.0> proxy_session_id=2 [info] Upstream connection is ready to accept queries
02:35:18.216 pid=<0.3143.0> proxy_session_id=2 [info] Matched prisma introspection query to Elixir.Electric.Postgres.Proxy.Prisma.Query.NamespaceV5_2
02:35:18.218 pid=<0.3143.0> proxy_session_id=2 [info] Matched prisma introspection query to Elixir.Electric.Postgres.Proxy.Prisma.Query.TableListV4_8
02:35:18.220 pid=<0.3143.0> proxy_session_id=2 [info] Matched prisma introspection query to Elixir.Electric.Postgres.Proxy.Prisma.Query.TypeV4_8
02:35:18.222 pid=<0.3143.0> proxy_session_id=2 [info] Matched prisma introspection query to Elixir.Electric.Postgres.Proxy.Prisma.Query.ColumnV4_8
02:35:18.225 pid=<0.3143.0> proxy_session_id=2 [info] Matched prisma introspection query to Elixir.Electric.Postgres.Proxy.Prisma.Query.ForeignKeyV4_8
02:35:18.227 pid=<0.3143.0> proxy_session_id=2 [info] Matched prisma introspection query to Elixir.Electric.Postgres.Proxy.Prisma.Query.IndexV4_8
[...]

PR #872, opened 2024-01-23
POC: Generate a Prisma schema using migration metadata obtained from the sync service

alco:
Fixes VAX-1339, #863.

This is a proof-of-concept pull request that demonstrates the ability to obtain a Prisma schema which is later used in the type-safe client generation process without having Prisma connect to the Migrations proxy.

Overview of the included changes:
* The `SatOpMigrate.Column` message gets a new `is_nullable` field. This information is needed to correctly indicate optional fields in the generated Prisma schema.
* The `SatOpMigrate` message is extended to allow for inclusion of an enum type definition: a `CREATE TYPE ... AS ENUM ...` SQL statement and associated metadata ([relevant commit range](https://github.com/electric-sql/electric/pull/872/files/b446a597076a85752c26e0681dd268d541bc1fe1..649abc9d4644140464d5e39071ee4118f212a94b#diff-29dc52da8f163e16abbad989a77e7ca7b32b4dfec1a509aaf8e49b0d7189bbc0). @icehaunter @magnetised, I would love to get your feedback on this.
* The `introspectDB()` function [is reimplemented](https://github.com/electric-sql/electric/pull/872/files#diff-511687b4341aa319db30c8f1f9a219e6d3c6bc65253cc60ea81e33cc4f9e647eL456-R448) to generate a Prisma schema from the migrations metadata and write it to the temporary `schema.prisma` file.
* The complete implementation of the new approach to generating a Prisma schema can be found in the [`clients/typescript/src/cli/migrations/migration-to-prisma-models.ts`](https://github.com/electric-sql/electric/pull/872/files#diff-d166da73e7a454d2272cb76b8fea71cd6c5adb4b2f8bcbe710c0264a0aaa387f) file.

How to proceed from here:
* I'd like to have both the backend and the frontend teams look at this POC and point out any gotchas or missing functionality
* Once we agree that this is a workable solution, I can open a new pull request that will only include changes to the protocol and the server-side part of the implementation
* Someone on the frontend team can then open a pull request (based on the server one) that will take my client-side changes, clean them up and address all the bits that I've missed or implemented poorly due to my lack of TS expertise.

TODO:
* deduplicate enum definitions in ddl_commands.


icehaunter:
> At this point, merging it would do more harm than good. People will need to update both the server and the client at the same time to keep the client-side CLI working correctly with the new username. There are probably other potential breakages that weren't a concern at the time.

as per @alco, so closing for now

KyleAMathews:

Issue #871, opened 2024-01-22
The electric-sql generate --proxy flag is broken in electric-sql@0.9


KyleAMathews:
It just creates a url from the other possible env variables resulting in something like: `Proxy URL: postgresql://prisma:proxy_password@localhost:65432/garmin-app`

KyleAMathews:
actually, this means it doesn't work anywhere. I was testing locally and it seemed like it was working but then I realized it was because it was connecting to a local running electric process.

samwillis:
Hey @KyleAMathews 

Thanks for the report, yes there's a bug in the generate command. `--proxy` should be overriding the env vars and defaults but isn't. I will have a PR fixing this ready today.

Until then you can use `--pg-proxy-host`, `--pg-proxy-port` etc to override individual parts. To see the full list run `npx electric-sql help generate`.

KyleAMathews:
Thanks!

PR #873, opened 2024-01-23
fix(cli) Split inferred values from default values - Fixes VAX-1569

samwillis:
This splits inferring config values from defaults.

This fixes VAX-1569 (#871), the bug was due to the proxy url being constructed from it's parts, but the parts were not inferred from command line args, only env vars.


kevin-dp (approved):
LGTM

PR #875, opened 2024-01-23
chore (examples): update expo-sqlite dependency in expo example app

kevin-dp:
This PR updates the `expo-sqlite` dependency in the expo example app to be `>= 11.7.0` which is the minimal version that is needed in order for `RETURNING` clauses to work on Android.

### DIFF
--- a/examples/expo/package.json
+++ b/examples/expo/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "electric-sql": "^0.7.0",
     "expo": "~49.0.8",
-    "expo-sqlite": "~11.3.3",
+    "expo-sqlite": ">= 11.7.0",
     "expo-status-bar": "~1.6.0",
     "react": "^18.2.0",
     "react-native": "0.72.6",


alco (approved):
It feels a bit icky to specify a version range that's unbounded from the top in the JS land. `^11.7` should be somewhat safer against future breakage.

kevin-dp:
> It feels a bit icky to specify a version range that's unbounded from the top in the JS land. `^11.7` should be somewhat safer against future breakage.

I must admit it's a bit scary. I've used `>= 11.7.0` because that's also how all the other examples are configured. Would you prefer that i update all dependencies to `^11.7.0`?

alco:
> > It feels a bit icky to specify a version range that's unbounded from the top in the JS land. `^11.7` should be somewhat safer against future breakage.
> 
> I must admit it's a bit scary. I've used `>= 11.7.0` because that's also how all the other examples are configured. Would you prefer that i update all dependencies to `^11.7.0`?

Nah, ignore me if that's how the other examples do it.

PR #876, opened 2024-01-23
chore (docs): update explanation on how to integrate new DB drivers

kevin-dp:
This PR updates the explanation on how to integrate new DB drivers in Electric.
While the old explanation is still valid, i extended this section to explain that the simplest option is to extend one of the two generic database adapters.

### DIFF
--- a/docs/integrations/drivers/other/generic.md
+++ b/docs/integrations/drivers/other/generic.md
@@ -35,6 +35,27 @@ export interface DatabaseAdapter {
 }
 ```
 
+For convenience, we provide two generic database adapters, `SerialDatabaseAdapter` and `BatchDatabaseAdapter`, that implement the parts of this interface that are common to most adapters such that you can implement a simpler interface:
+```tsx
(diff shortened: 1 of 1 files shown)


docs/integrations/drivers/other/generic.md
```diff
@@ -35,6 +35,27 @@ export interface DatabaseAdapter {
 }
 ```
 
+For convenience, we provide two generic database adapters, `SerialDatabaseAdapter` and `BatchDatabaseAdapter`, that implement the parts of this interface that are common to most adapters such that you can implement a simpler interface:
```
thruflo, line 38: Minor suggestion of shorter sentences:

> For convenience, we provide two generic database adapters, `SerialDatabaseAdapter` and `BatchDatabaseAdapter`. These implement the parts of the interface that are common to most adapters. This allows you to implement your own driver adapters using a simpler interface.

thruflo (approved):

PR #878, opened 2024-01-24
fix typo: "cloud-first" instead of "client-first"

RicoTrevisan:
I think you meant to write "cloud-first" since this is the local- vs cloud-first intro.


alco (approved):
Thanks! :heart:

PR #880, opened 2024-01-24
fix (client): clear all subscriptions from the DB in 1 transaction

kevin-dp:
When a client reconnects to Electric and can't resume it resets its local state (i.e. unsubscribes all subscriptions) and starts from scratch. Currently, we unsubscribe subscription per subscription. For each subscription, we delete the contents of the shapes from the DB. However, this can lead to FK violations. e.g.:

- Table User
- Table Post has a FK to user
- Subscription 1 subscribes to user table
- Subscription 2 subscribes to post and user table

When clearing all subscriptions, if we first unsubscribe from subscription 1, we are deleting the contents of the user table and thereby violating any FKs that are in the post table.

This PR avoids this problem by clearing all subscriptions within a single database transaction that defers the foreign keys until the end of the transaction. Also introduced a unit test that reproduces the FK violation and is gone with this fix.

### DIFF
--- a/clients/typescript/src/satellite/shapes/manager.ts
+++ b/clients/typescript/src/satellite/shapes/manager.ts
@@ -25,7 +25,7 @@ export class InMemorySubscriptionsManager
   implements SubscriptionsManager
 {
   private inFlight: SubcriptionShapeRequests = {}
-  private fulfilledSubscriptions: SubcriptionShapeDefinitions = {}
+  protected fulfilledSubscriptions: SubcriptionShapeDefinitions = {}
   private readonly shapeRequestHashmap: Map<string, SubscriptionId> = new Map()
 
   private readonly gcHandler?: GarbageCollectShapeHandler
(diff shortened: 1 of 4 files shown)


clients/typescript/src/satellite/shapes/manager.ts
```diff
@@ -107,27 +107,42 @@ export class InMemorySubscriptionsManager
...
+    if (this.gcHandler) {
+      await this.gcHandler(shapes)
     }
+    // also remove all subscriptions from memory
+    this._gcSubscriptions(ids)
     return ids
   }
 
+  unsubscribeAll(): Promise<string[]> {
+    const ids = Object.keys(this.fulfilledSubscriptions)
```
msfstef, line 142: should this also cancel all in-flight subscription requests?

clients/typescript/src/satellite/shapes/manager.ts
```diff
@@ -107,27 +107,42 @@ export class InMemorySubscriptionsManager
...
+    subIds: SubscriptionId | SubscriptionId[]
+  ): Promise<SubscriptionId[]> {
+    const ids = Array.isArray(subIds) ? subIds : [subIds]
+    const shapes: ShapeDefinition[] = ids.flatMap(
+      (id) => this.shapesForActiveSubscription(id) ?? []
+    )
+
+    // GC all subscriptions in a single DB transaction
+    if (this.gcHandler) {
+      await this.gcHandler(shapes)
```
msfstef, line 134: since there is an async gap here, would it make sense for the fulfilled subscriptions (and if we want to cancel the inflight ones as well) to be removed from memory synchronously and let the gcHandler clean them up immediately after? If inflight are to be removed as well, between calling `unsubscribe`, the gcHandler running, an then the inflight being removed from memory, an inflight request could have been delivered and a shape remains even though it should have been cleaned up (?)

msfstef (approved):
Looks good! Left a couple of comments asking for clarification on what look to me like race conditions (unless I'm misunderstanding the usage of the subscription manager)

clients/typescript/src/satellite/shapes/manager.ts
```diff
@@ -161,3 +176,23 @@ function computeRequestsHash(requests: ShapeRequestOrDefinition[]): string {
 function computeClientDefsHash(requests: ClientShapeDefinition[]): string {
   return hash(requests)
 }
+
+export class MockSubscriptionsManager extends InMemorySubscriptionsManager {
```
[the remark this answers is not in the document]
kevin-dp: It could but the rest of the code base also defines the mock in `src` and not in `test`.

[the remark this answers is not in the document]
msfstef: fair enough - I've noticed that the mock classes are included in the `dist` folder, might be bloating it a bit and not sure if subsequent bundling/tree-shaking by the developers using electric takes care of them

clients/typescript/src/satellite/shapes/manager.ts
kevin-dp: We only call `unsubscribeAll` inside the `_resetClientState` method in the Satellite process after we disconnected from Electric. So inflight subscriptions cannot arrive. However, `unsubscribeAll` is a public method on the subscriptions manager which is also publicly available on the Satellite process, so a user could in theory call that method while being connect to Electric in which case this can occur. So perhaps it's indeed best to move their order just to be sure.

kevin-dp:
Hi @msfstef, thanks for your review. I understand your concern regarding inflight subscriptions. However, unsubscribing also from inflight subscriptions would change the current semantics and is therefore out of scope for this PR. This PR is meant to keep the current semantics but execute all unsubscribe queries in a single transaction.

Nevertheless your point is valid and could be implemented in a separate PR if we agree that the semantics should change to also unsubscribe inflight subscriptions.

msfstef:
No worries, that's why I approved either way, I'm mostly asking to understand the codebase more myself :) Separate PR is definitely more appropriate.
Looks good!

clients/typescript/src/satellite/shapes/manager.ts
balegas:
```ts
    if (shapes) {
       ...
      delete this.inFlight[subId]
      delete this.fulfilledSubscriptions[subId]
```
that line of code is clearly redundant, but that code looks okays.

balegas: > so a user could in theory call that method while being connect to Electric in which case this can occur. So perhaps it's indeed best to move their order just to be sure.

The `SubscriptionManager` is not public in the Satellite interface, it's only at the implementation. It is not expected that anyone would call it directly, even if theoretically possible. On a next iteration of this code, we can make it private.

PR #884, opened 2024-01-24
feat(client): Abstract subscription to query results away from frameworks

msfstef:
I've lifted the idea of subscribing to a live query to the core client level, which simplifies writing and maintaining bindings for the various front-end frameworks (React, Svelte, Solid, Angular, Vue). It also allows for using this pattern outside of a framework for more state-related business logic (e.g. for offline persistent RPC like in the request/response pattern)

I've opted for adding to the `LiveResultContext` a `subscribe` method, that takes a `handler` which receives either a result or an error, and an update timestamp. I've kept the schema backwards compatible with our existing React bindings.

Will add tests to the `createQueryResultSubscribeFunction` function but figured I'd check if you agree with the approach first.


kevin-dp:
@msfstef I absolutely ❤️ this refactoring.
By moving more logic into the DAL it simplifies the implementation of all drivers, brilliant!

msfstef:
> @msfstef I absolutely ❤️ this refactoring. By moving more logic into the DAL it simplifies the implementation of all drivers, brilliant!

Glad you like it! I've added a test suite for the subscription function generator and also mentioned this approach to the client documentation page.

If you think it's best we keep this undocumented for start I can remove it, although I figured it might help with getting people to use it to build stuff on top of the client.

samwillis (changes requested):
Love this, particularly this:

```
const liveQuery = db.issues.liveMany()
const unsubscribe = liveQuery.subscribe(
  (resultUpdate) => * handle updated result *
)
```

We can hopefully extend that in future with more subscription options, such as a `subscribeChanges` that emits the changes that have happed (deletes, inserts, edits)...

Only comment, we need to split the docs into a separate PR, doc changes are immediately published to the website and need to be merged after a release including this.

msfstef:
Put the docs in a separate PR, I'll keep it in mind from now on!

I agree with enhancing it, one more thing that would be great would be if the notifier only triggered updates when relevant rows are updated (and not just the relevant table), which would improve the performance of this API under the hood while also letting us subscribe to e.g. changes to a particular row with `once` semantics which could be powerful.

samwillis (approved):
👍

PR #891, opened 2024-01-25
feat(examples): Tauri SQLite example app

kevin-dp:
This PR adds a driver for Tauri SQLite (fixes VAX-1595).
It also adds an example Tauri application.

To test the example app it needs to run the modified version of our ts-client.
To this end, `npm pack` it and modify the `electric-sql` dependency of the Tauri example app to point to it.

Then follow the steps in the example app's readme but follow this deviation:
**after generating the client** go into the generated client `src/generated/client/index.ts` and add `// @ts-nocheck` at the top of the file. This is needed because there seem to be type errors in the generated client (used not to be the case before, need to investigate as a separate issue). Afterwards, you can continue with the steps from the readme.


examples/tauri/README.md
```diff
@@ -0,0 +1,83 @@
...
+    />
+    <img alt="ElectricSQL logo"
+        src="https://raw.githubusercontent.com/electric-sql/meta/main/identity/ElectricSQL-logo-black.svg"
+    />
+  </picture>
+</a>
+
+# Welcome to your ElectricSQL app!
+
+This is a web application using ElectricSQL in the browser with [wa-sqlite](https://electric-sql.com/docs/integrations/drivers/web/wa-sqlite).
```
samwillis, line 17: Readme instructions wrong need updating to explain this is a Tauri v2 demo. And the correct requirements for Tauri.

samwillis (approved):
Look great. One question is on the example, do we include it? If so we need to update the readme file.

Other than that, let's get it merged.

kevin-dp:
> Look great. One question is on the example, do we include it? If so we need to update the readme file.
> 
> Other than that, let's get it merged.

Right, let's decide on including the example or not during today's dev call 👍

kevin-dp:
@samwillis Did `npm install` in the example app also complain for you?
```
Unsupported URL Type "workspace:": workspace:*
```
I guess this issue has something to do with the packed version of the client i'm using.
However, i can follow all steps using yarn instead of npm.

PR #895, opened 2024-01-26
fix(docs): Address a few issues in the client docs

### LINKED ISSUE #886: npm run start in quickstart is wrong
The [quickstart](https://electric-sql.com/docs/quickstart#setup) says to use `npm run start` to start the app.

This should be `npm run dev` instead, as the start script does not exist in package.json after running `npx create-electric-app@latest my-app`.

alco:
Fixes #886.


samwillis (approved):
👍

PR #898, opened 2024-01-29
chore(electric): Docker-related cleanups

alco:
This PR makes two changes:
* remove `privileged: true` from services defined in several `compose.yaml` files
* use `SIGINT` as the default stop signal for the Electric service. This results in faster shutdown after pressing Ctrl-C or stopping all services with `docker compose stop/down`.


magnetised (approved):
why did we ever have privileged: true? nice.

PR #899, opened 2024-01-29
feat!: shape subscriptions

icehaunter:
Main feature of this commit is the support for arbitrary FK traversal when requesting shape subscription. This makes previous "multiple full table" requests invalid, because when a table is requested without following it's FKs, we cannot guarantee FK consistency of the client's data view.

When requesting a shape, the server will do it's best to fulfill "at least" the requested shape: if there are any outgoing FKs from the tables mentioned in the requested include tree, those will also be traversed and sent to the client.

Another important change is introduction of "additional" transactions: those are non-timestamped transactions that the server sends to the clients when they need to see some data they haven't seen before, but those rows weren't affected directly. They are tied to a "real" transaction that caused this additional data to appear by a reference number, so the client may wait for the full view of the data if it so chooses. Only one-to-many relations are followed with a lag like that,
many-to-one relations will always be present in the transaction itself to ensure FK consistency on the client.

This functionality is supported by a sent rows FK graph that's maintained per client in the WebSocket process. It may be persisted, but should probably have a similar lifetime to WAL cache window, since that's the current boundary for client reconnection.

---
Closes VAX-1260, VAX-593


magnetised (approved):
`magic()` - awesome stuff.

clients/typescript/src/satellite/merge.ts
```diff
@@ -147,6 +154,13 @@ function mergeOpTags(
   local: OplogEntryChanges,
   remote: ShadowEntryChanges
 ): Tag[] {
+  // When the server sends a GONE message, it means we need to delete this row from our side as no further
+  // updates will come through. Server doesn't keep track of seen tags, however, so we make the GONE operation
+  // have a higher priority than anything else.
+
+  // TODO: Does deleting on GONE make sense at all?
```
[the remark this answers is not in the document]
icehaunter: That's addressed already by some chats w/ @balegas, and I'm indeed going to expand upon this in further issues

clients/typescript/src/util/types.ts
```diff
@@ -174,7 +176,7 @@ export type RelationColumn = {
   name: string
   type: string
   isNullable: boolean
-  primaryKey?: boolean
```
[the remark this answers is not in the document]
icehaunter: Unfortunately

components/electric/.tool-versions
```diff
@@ -0,0 +1,2 @@
+erlang 25.3
```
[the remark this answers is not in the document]
icehaunter: Mis-addition. I work with electric folder as VSCode root and missed that it got added

components/electric/lib/electric/postgres/extension/functions.ex
```diff
@@ -11,12 +11,12 @@ defmodule Electric.Postgres.Extension.Functions do
...
 
-  template_dir_path = Application.app_dir(:electric, @template_dir)
-  sql_template_paths = Path.wildcard(template_dir_path <> "/**/*.sql.eex")
+  @template_dir_path Application.app_dir(:electric, @template_dir)
+  sql_template_paths = Path.wildcard(@template_dir_path <> "/**/*.sql.eex")
 
   function_paths =
     for path <- sql_template_paths do
-      relpath = Path.relative_to(path, template_dir_path)
+      relpath = Path.relative_to(path, @template_dir_path)
```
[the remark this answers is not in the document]
alco: No need for `@external_resource`. SQL template files are loaded at runtime, so a content change in any of them doesn't affect module compilation.

components/electric/lib/electric/satellite/protocol/state.ex
```diff
@@ -0,0 +1,85 @@
...
+          auth: nil | Electric.Satellite.Auth.t(),
+          last_msg_time: :erlang.timestamp() | nil | :ping_sent,
+          client_id: String.t() | nil,
+          in_rep: InRep.t(),
+          out_rep: OutRep.t(),
+          auth_provider: Electric.Satellite.Auth.provider(),
+          connector_config: Keyword.t(),
+          subscriptions: map(),
+          subscription_data_fun: fun(),
+          move_in_data_fun: fun(),
```
alco, line 30:
```suggestion
          subscription_data_fun: fun() | nil,
          move_in_data_fun: fun() | nil,
```

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -955,25 +940,57 @@ defmodule Electric.Satellite.Protocol do
...
-    telemetry =
-      Telemetry.start_subscription_span(
-        state.telemetry,
-        id,
-        %{
-          included_tables: length(Enum.uniq(included_tables)),
-          shapes: total_requests
-        },
-        Map.put(common_metadata(state), :shape_hashes, hashes)
+  @spec query_move_in_data(Shapes.subquery_actions(), State.t()) ::
```
alco, line 943:
```suggestion
  @spec query_move_in_data(actions(), State.t()) ::
```

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -955,25 +940,57 @@ defmodule Electric.Satellite.Protocol do
...
+      |> ShapeRequest.prepare_filtering_context()
+      |> Map.put(:user_id, Pathex.get(state, path(:auth / :user_id)))
+
+    # I'm dereferencing these here because calling this in Task implies copying over entire `state` just for two fields.
+    fun = state.move_in_data_fun
+    move_in_ref = state.out_rep.move_in_next_ref
+    # span = state.telemetry.subscription_spans[id]
+
+    Task.start(fn ->
+      # This is `InitialSync.query_subscription_data/2` by default, but can be overridden for tests.
```
alco, line 964: `fun` here actually holds `InitialSync.query_after_move_in/4`.

components/electric/lib/electric/utils/gen_stage.ex
```diff
@@ -0,0 +1,48 @@
+defmodule Electric.Utils.GenStage do
```
alco, line 1: It's quite confusing to see `Utils.GenStage` and `GenStage` mentioned in close proximity. How about changing the module name to `Electric.Utils.GenStageHelpers`?

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -44,31 +47,43 @@ defmodule Electric.Satellite.Serialization do
...
+      state.known_relations
+    }
+  end
+
+  def serialize_move_in_data_as_tx(ref, changes, known_relations) do
+    begin_op = %SatTransOp{op: {:additional_begin, %SatOpAdditionalBegin{ref: ref}}}
+    commit_op = %SatTransOp{op: {:additional_commit, %SatOpAdditionalCommit{ref: ref}}}
+
+    state = %{
+      ops: [commit_op],
```
alco, line 78: It is somewhat hacky to put the `commit_op` here with the expectation that trans-ops will be added to the state inside `Enum.reduce()` calling `serialize_changes()` below, and then top it off with prepending `begin_op`. It threads the knowledge about the exact contenst of the `state.ops` field between here and the implementation of `serialize_change/2`.

Also, compared to, for example, the handling of ops in the `serialize_trans()` function above (compare line 67 to 86), the order of operations is theorder of `changes` reversed. Does the order within a transaction not matter?

components/electric/lib/electric/satellite/ws_server.ex
```diff
@@ -177,7 +189,9 @@ defmodule Electric.Satellite.WebsocketServer do
     origin = Electric.Replication.Connectors.origin(state.connector_config)
     migrations = InitialSync.migrations_since(schema_version, origin, lsn)
 
-    {msgs, state} =
+    # We're ignoring actions here since we've "manufactured" migration events
+    # which by definition aren't shape-dependent, so actions are always empty
+    {msgs, _actions, state} =
```
alco, line 194: Why not assert that `[] = _actions`?

components/electric/lib/electric/satellite/serialization.ex
icehaunter: Not for a "fake" transaction like this. This is just data being sent - maybe it's better just not modelled within a txn message at all

components/electric/lib/electric/satellite/ws_server.ex
icehaunter: %{}, but yeah, good point

components/electric/lib/electric/satellite/protocol/state.ex
icehaunter: no, they need to be set always. We can even `@enforce_keys` on them - state is built as a new object exactly once in `Electric.Satellite.WebsocketServer.init/1`

components/electric/lib/electric/replication/shapes/change_processing.ex
```diff
@@ -0,0 +1,467 @@
...
+      case record_or_id do
+        record when is_map(record) -> id(record, layer.target_table, layer.target_pk)
+        id -> id
+      end
+
+    Graph.add_edge(graph, :root, own_id, label: layer.key)
+  end
+
+  def add_to_graph(graph, %Layer{} = layer, record_or_id, parent_record_or_id)
+      when (is_map(record_or_id) or is_tuple(record_or_id)) and layer.direction != :first_layer and
```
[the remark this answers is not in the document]
icehaunter: That's only partially true. Point here is that if `layer.direction == :first_layer`, the fourth argument MUST be null, otherwise it MUST NOT be null. This is only achievable with these two guards, and I can say it guarded my errors a couple of time while writing tests so I'll insist on keeping that

clients/typescript/src/util/types.ts
```diff
@@ -174,7 +176,7 @@ export type RelationColumn = {
...
+  primaryKey?: number | false
```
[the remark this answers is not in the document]
icehaunter: practically, it forces a stronger check than `!== undefined`, to make sure only positive numbers are relevant. This is a bit hazy though and I can simplify down to `undefined

components/electric/test/electric/satellite/subscriptions_test.exs
```diff
@@ -722,6 +659,381 @@ defmodule Electric.Satellite.SubscriptionsTest do
...
+
+        # Updates to GONE entries are not propagated
+        {:ok, 1} =
+          :epgsql.equery(
+            pg_conn,
+            "UPDATE public.comments SET content = $1 WHERE id = $2",
+            ["New comment 2", comment_2_id]
+          )
+
+        refute_receive {^conn, %SatOpLog{}}, 100
```
alco, line 821: This might be nitpicky but DB queries in tests and their subsequent processing by Electric's logical replication consumer, especially when running on CI, may take a while. A timeout of 1000 to confirm that a message **is not** received would look more convincing here.

alco (approved):
Brilliant! :rocket: 

Thanks for including an extensive test coverage for the new functionality!

It would be helpful to have brief, high-level module docs added to the new modules to help the reader orient themselves within the shape procesing data flow. It's a lot of new code to take in, so I can't claim I was able to wrap my head around it.

components/electric/lib/electric/satellite/protocol/state.ex
alco: Sorry, I know this is a nitpick, it's just they are set to `nil` by default in the `defstruct` definition above. Perhaps Dialyzer has figureed out that they are never `nil` when they are referenced in other code. I wonder why it sometimes misses cases like this but then at some point starts complaining about them when some other, seemingly unrelated code, is modified.

components/electric/lib/electric/replication/shapes/change_processing.ex
```diff
@@ -0,0 +1,467 @@
...
+    |> Enum.reduce(state, fn {event, layer, own_id}, state ->
+      move_out(state, event, layer, own_id)
+    end)
+  end
+
+  def move_in(state, %Changes.UpdatedRecord{} = event, layer, own_id, parent_id \\ nil)
+      when is_struct(layer, Layer) and is_reduction(state) do
+    graph = add_to_graph(reduction(state, :graph), layer, event.record, parent_id)
+
+    process_next_layers(layer, reduction(state, graph: graph), event.record, event: :new)
```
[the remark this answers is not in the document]
icehaunter: Made it nice and pipy

components/electric/lib/electric/replication/shapes/change_processing.ex
```diff
@@ -0,0 +1,467 @@
...
+    parent_id =
+      case parent_record_or_id do
+        record when is_map(record) -> id(record, layer.source_table, layer.source_pk)
+        id -> id
+      end
+
+    Graph.add_edge(graph, parent_id, own_id, label: layer.key)
+  end
+
+  def add_to_graph(graph, %Layer{} = layer, own_id, parent_id)
```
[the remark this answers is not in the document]
icehaunter: Yeah, you're correct, it was a leftover from previous iterataions

components/electric/lib/electric/replication/shapes/change_processing.ex
```diff
@@ -0,0 +1,467 @@
...
+
+    Graph.add_edge(graph, :root, own_id, label: layer.key)
+  end
+
+  def add_to_graph(graph, %Layer{} = layer, record_or_id, parent_record_or_id)
+      when (is_map(record_or_id) or is_tuple(record_or_id)) and layer.direction != :first_layer and
+             parent_record_or_id != nil and
+             (is_map(parent_record_or_id) or is_tuple(parent_record_or_id)) do
+    own_id =
+      case record_or_id do
```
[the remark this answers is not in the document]
icehaunter: I disagree on this, let's leave it at that - it's only 2 places in the code and they don't share their args

components/electric/lib/electric/satellite/protocol/state.ex
icehaunter: because you cannot "not" set nil as default in these struct definitions, but Dialyzer assumes you'll set a real value when you first build it. So if your function return type is `t()` and you forget this assignment it'll complain

icehaunter: There is no "sane" compile-time default to be had here, it's just a value that should never be nil after the record is constructed

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -955,25 +940,57 @@ defmodule Electric.Satellite.Protocol do
...
+      # Please see documentation on that function for context on the next `receive` block.
+      fun.(move_in_ref, actions, context,
+        reply_to: {ref, parent},
+        connection: state.connector_config
+        # telemetry_span: span
       )
+    end)
+
+    receive do
+      {:subscription_insertion_point, ^ref, xmin} ->
```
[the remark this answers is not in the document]
icehaunter: Sure, will rename, but the logic is exactly the same, hence the same name

components/electric/test/electric/satellite/subscriptions_test.exs
icehaunter: No, good point, I definitely agree on a refute

components/electric/lib/electric/replication/shapes/querying.ex
```diff
@@ -0,0 +1,292 @@
...
+    columns = Enum.map_join(table.columns, ", ", &~s|this."#{&1.name}"|)
+    pk_clause = Enum.map_join(layer.target_pk, " AND ", &~s|this."#{&1}" = shadow."#{&1}"|)
+
+    where_clauses =
+      [
+        where_target(layer.where_target),
+        parent_pseudo_join(layer, from),
+        context_filters(table, context, layer)
+      ]
+      |> Enum.reject(&(is_nil(&1) or &1 == ""))
```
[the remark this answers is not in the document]
icehaunter: It's literally 3 element list, stream buildup is more expensive than iterating that 2 times

clients/typescript/src/satellite/client.ts
```diff
@@ -616,6 +701,7 @@ export class SatelliteClient implements Client {
         return { error: startReplicationErrorToSatelliteError(resp.err) }
       } else {
         this.inbound.isReplicating = ReplicationStatus.ACTIVE
+        this.inbound.maxUnackedTxs = resp.unackedWindowSize ?? 30
```
magnetised, line 704: Can we move this magic number to a named constant

components/electric/lib/electric/replication/shapes.ex
```diff
@@ -110,8 +110,16 @@ defmodule Electric.Replication.Shapes do
     end
   end
 
-  @spec merge_actions(subquery_actions(), subquery_actions()) :: subquery_actions()
-  def merge_actions(a1, a2) do
-    Map.merge(a1, a2, fn _, v1, v2 -> v1 ++ v2 end)
+  @type action_context :: {actions :: subquery_actions(), source_tx_ids :: [non_neg_integer()]}
```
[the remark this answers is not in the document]
icehaunter: True but that's too many places to change, I'll do a refactor later, it doesn't feel impactful here

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -0,0 +1,494 @@
...
+
+    results = :ets.select(@additional_data_ets, matchspec)
+    :ets.select_delete(@additional_data_ets, matchspec)
+
+    results
+  end
+
+  defp clear_stored_actions(client_id, txn_ids) do
+    matchspec =
+      for id <- txn_ids, do: {{{client_id, id}, :_}, [], [:"$$"]}
```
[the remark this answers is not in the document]
icehaunter: Great catch!

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -0,0 +1,494 @@
...
+  @type additional_data_txn_row ::
+          {additional_data_txn_key(), graph_diff :: Graph.t(),
+           source_txns :: [non_neg_integer(), ...]}
+  @type additional_data_row :: additional_data_sub_row() | additional_data_txn_row()
+
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  def clear_all_data(client_id) do
```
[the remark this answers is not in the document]
alco: > > needs to be made private
> 
> Why so, given it's explicitly used in `protocol.ex` right under the comment you've corrected? We need to clear all data for a client who connects outside the WAL window, and I don't see a way to make it private here.

I knew I saw it used outside! For some reason, when I checked out your branch on my machine and did a code search in VSCode, it only showed uses in this module. Probably an indexing glitch. Sorry for the noise.

Regarding your other points, I'm not trying to start a debate here but just wanted to clarify what I meant. None of it requires any action in this PR, but I think these arguments still hold:

We _should_ be mindful of concurrent processes acting on shared data structures when we add such data structures in the code. The code written now may live on for years when other features will be added to it and different people will be modifying it. If a lot of logic may break when called by concurrent processes, doesn't that sound like a problem? Elixir gives us tools to delineate parts of the code that need to be serialized (gen servers) vs parts of the code that can be run concurrently (thanks to immutable data).

If a public function is vulnerable to data races when called from concurrent processes, it cannot be called an implementation detail of the module. Disregarding my earlier comment about making `clear_all_data()` private, I would appreciate a comment on it explaining that concurrent access is not a concern for us because each client_id is guaranteed to have a single websocket process for it thanks to the use of ClientManager.


> > reconnection info is truly local for each client
> 
> It is truly local, but it needs to persist reconnects (even disregarding Electric restarts), so it can't be in the WS process state. [...] given the next step of putting all this info into a shared persistent cache (for restarts), I felt like implementing a set of ETS tables is a better option that leaves us with a more explicit API surface for a cache.

Fair point. With my take I was pointing out the fact that a different structuring of data in the code could remove the "concurrency problem" entirely. And keeping client-local state local to the client isn't creating potential performance bottlenecks where an ordered table has to be scanned repeatedly for the same `client_id` while other client IDs are less frequently touched. But that's a concern for a later time.

clients/typescript/test/satellite/process.test.ts
```diff
@@ -1617,6 +1615,88 @@ test('applied shape data will be acted upon correctly', async (t) => {
   }
 })
 
+test.only('additional data will be stored properly', async (t) => {
```
davidmartos96, line 1618: It looks like a `.only` remained in the code :)

components/electric/lib/electric/replication/shapes/querying.ex
```diff
@@ -0,0 +1,292 @@
...
+      Enum.map_intersperse(columns, ", ", &["this.", ?", &1, ?"]),
+      ") IN (",
+      matches,
+      ?)
+    ]
+  end
+
+  defp wrap_values(value_list), do: Enum.map(value_list, &quote_escape/1)
+  defp quote_escape(nil), do: "NULL"
+  defp quote_escape(str), do: [?', :binary.replace(str, "'", "''"), ?']
```
magnetised, line 204:
```suggestion
  defp quote_escape(str), do: [?', :binary.replace(str, "'", "''", [:global]), ?']
```

components/electric/lib/electric/replication/shapes/querying.ex
```diff
@@ -0,0 +1,292 @@
...
+         tags: ShadowTableTransformation.convert_tag_list_pg_to_satellite(tags, origin)
+       }}
+    end)
+  end
+
+  defp context_filters(%Replication.Table{} = table, context, %Layer{} = _layer) do
+    ownership_column = Ownership.id_column_name()
+
+    if context[:user_id] && Enum.any?(table.columns, &(&1.name == ownership_column)) do
+      escaped = :binary.replace(context[:user_id], "'", "''")
```
magnetised, line 287:
```suggestion
      escaped = :binary.replace(context[:user_id], "'", "''", [:global])
```

magnetised (changes requested):
found missing `:global` modifiers on `:binary.replace/3`

clients/typescript/test/satellite/process.test.ts
icehaunter: Thanks!

magnetised (approved):

PR #901, opened 2024-01-29
fix(electric): Support JWTs without a trailing dot in Insecure mode

### LINKED ISSUE #900: JWT token: unsigned tokens are invalid
Hi

Documentation [suggests](https://electric-sql.com/docs/usage/auth/insecure) to use https://token.dev/ service to manually generate unsigned tokens. token.dev generates tokens like `eyJhbGciOiJub25lIiwidHlwIjoiSldUIn0.eyJzdWIiOiIxIn0`. Note that there is only one `.`. It's totally valid unsigned JWT token. But Joken library doesn't support them at all (see the line
https://github.com/joken-elixir/joken/blob/main/lib/joken.ex#L177)

I think it would be great at least to explicitly mention it in documentation and/or make more helpful error messages (right now it's `Client authentication failed: Invalid token`)

Workaround: adding `.` to the end of token helps.

alco:
Fixes #900.

### DIFF
--- a/.changeset/shaggy-planes-bake.md
+++ b/.changeset/shaggy-planes-bake.md
@@ -1,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
+Support JWTs without a trailing dot in the Insecure auth mode. (#900)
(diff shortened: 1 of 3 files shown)


icehaunter (approved):

magnetised (approved):
JWTs, a big bug waiting to happen. ⚡

PR #905, opened 2024-01-30
feat (client): modified shapes to adhere to new protocol format

kevin-dp:
This PR modifies the shapes on the client to adhere to the new protocol format for shapes.
It also introduces support for `where` clauses at the **top-level** of a shape.
Those where clauses support:

```ts
where: {
  name: 'Kevin',
  OR: [
    { age: 28 },
    { age: 29 }
  ],
  NOT: [
    { country: 'France' }
  ]
  // AND is also supported but is redundant
}
```

**BEWARE:** `where` support is experimental and is not yet finished: typings and Zod schemas are too permissive and we don't support `where` clauses in nested `include`. One should not use relation fields in the shape and only check field values (no nested objects).

### DIFF
--- a/clients/typescript/src/client/input/syncInput.ts
+++ b/clients/typescript/src/client/input/syncInput.ts
@@ -1,3 +1,4 @@
-export interface SyncInput<Include> {
+export interface SyncInput<Include, Where> {
   include?: Include
+  where?: Where
 }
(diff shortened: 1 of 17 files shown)


icehaunter:
I merged this in manually locally

PR #906, opened 2024-01-31
Clarify actual limitations on adding not-null constraints

fooware:
@balegas & @thruflo : This is my attempt at clarifying the limitations of adding not-null constrained columns, as per the Discord discussion.

I tried my best to make the text as short as possible without loosing out on details _why_ it isn't really supported.

### DIFF
--- a/docs/usage/data-modelling/constraints.md
+++ b/docs/usage/data-modelling/constraints.md
@@ -55,19 +55,20 @@ CREATE TABLE items (
 Adding a column with a not-null constraint is supported, but **not advisable** until default values are implemented:
 
 ```sql
+  -- Unsafe additive migration, requires the table to be empty
 ALTER TABLE items
-  -- Additive migration, supported
   ADD COLUMN baz TEXT NOT NULL;
 
-   -- Possible substitute for default values
-  SET baz = "fie";
+  -- Safe additive migration, but default values are not yet supported
+ALTER TABLE items
+  ADD COLUMN fie TEXT NOT NULL DEFAULT 'some_value';
 ```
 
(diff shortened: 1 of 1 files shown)


thruflo (approved):
Reads great, thank you!

docs/usage/data-modelling/constraints.md
```diff
@@ -37,39 +37,47 @@ Electric currently does not allow adding a new foreign key column with `ALTER TA
...
-Adding a column with a not-null constraint is supported because it's *additive*:
+Adding a column with a not-null constraint is supported, but **not advisable** until default values are implemented:
 
 ```sql
 ALTER TABLE items
+  -- Additive migration, supported
   ADD COLUMN baz TEXT NOT NULL;
+
+   -- Possible substitute for default values
+  SET baz = "fie";
```
alco, line 63: I'm not sure if this is pseudo-code or was meant as a real example. For a valid SQL, this is missing `UPDATE items ...`. But that wouldn't work either unless constraints are explicitly deferred for the duration of the transaction.

Or, alternatively, it could be a sequence of steps where a `NULL` column is added first, followed by `UPDATE items SET baz = "fie";`, followed by `ALTER TABLE ALTER COLUMN baz SET NOT NULL`. However, this last statement will be rejected by the migrations proxy.

All in all, with the way things are currently, `ALTER TABLE items ALTER COLUMN baz TEXT NOT NULL` would only work if the table is empty.

fooware: Update, no, you are right @alco, it only works if the table is empty. Must have confused it with my other data migration tests. I'll update the PR.

fooware: I have pushed an update to the doc changes now, please have another look at it @alco .

alco: Brilliant! Thank you!

PR #907, opened 2024-02-01
Use absolute paths in linux installation script

magnetised:


balegas:
Closed because I believe this is no longer necessary.

PR #908, opened 2024-02-01
fix (CLI): fix generator engine library issue

kevin-dp:
This PR fixes an issue with the CLI's generate command and fixes missing schemas in the generated client (fixes VAX-1580).
Our typescript client uses Prisma v4 to introspect the Postgres DB and to generate the Prisma and Electric clients.
However, if a user project includes Prisma v5 (e.g. to manage migrations using Prisma), then the generator crashed with the following error:
```
Error: Command failed: node /home/alco/code/electric-sql/electric/examples/nextjs-electric/node_modules/electric-sql/node_modules/prisma/build/index.js generate --schema=".electric_migrations_tmp_oM5W7E/prisma/schema.prisma"
Error: (query-engine-node-api library)
Details: NodeAPIQueryEngineLibrary.dmmf is not a function
[Context: getDmmf]

Prisma CLI Version : 4.8.1

    at ChildProcess.exithandler (node:child_process:422:12)
    at ChildProcess.emit (node:events:526:35)
    at maybeClose (node:internal/child_process:1105:16)
    at ChildProcess._handle.onexit (node:internal/child_process:305:5) {
  code: 1,
  killed: false,
  signal: null,
  cmd: 'node /home/alco/code/tmp/nextjs-electric/node_modules/electric-sql/node_modules/prisma/build/index.js generate --schema=".electric_migrations_tmp_sYnULy/prisma/schema.prisma"'
}
generate command failed: Generator script exited with error code: 1
```

After tracing down the problem i found that the crash happens when generating the Prisma client using Prisma v4 when the user's project has Prisma v5 installed. However, if we use Prisma v5 to generate the Prisma client then there is no problem. We cannot use Prisma v5 for everything because:
- the proxy only works up till Prisma v5.2
- our Electric client generator requires Prisma v4 beause it was forked from a Zod generator which at the time only supported Prisma v4

So this PR solves the problem by splitting the generation of the Prisma client and the Electric client in 2 passes. For the generation of the Prisma client it uses v5 and for the Electric client it uses v4. We could also use v5.2 for introspection and for generating the Prisma client and use v4 only for generating the Electric client.

### Testing

To test this PR, create a fresh Electric project `npx create-electric-app@latest test`.
Then add Prisma v5 and @prisma/client dependencies:
```sh
yarn add @prisma/client
yarn add prisma -D
```

Start Electric and migrate the DB:
```sh
yarn backend:start
# in another tab
yarn db:migrate
```

Now check that you can reproduce the issue:
```sh
yarn client:generate
# this should crash with the error from above
```

Now replace the electric-sql dependency by an npm pack-ed version of the ts-client in this PR.
Then:
```sh
yarn
yarn client:generate
```

That should have worked.
You can now run the app:
```sh
yarn dev
```

### DIFF
--- a/.changeset/afraid-keys-begin.md
+++ b/.changeset/afraid-keys-begin.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Modified the CLI's generate command to fix issues with the generator when user projects include Prisma v5.
--- a/clients/typescript/package.json
+++ b/clients/typescript/package.json
@@ -137,7 +137,7 @@
   },
   "dependencies": {
     "@electric-sql/prisma-generator": "workspace:*",
-    "@prisma/client": "5.2.0",
+    "@prisma/client": "4.8.1",
     "async-mutex": "^0.4.0",
     "base-64": "^1.0.0",
     "better-sqlite3": "^8.4.0",
@@ -163,7 +163,8 @@
     "long": "^5.2.0",
     "object.hasown": "^1.1.2",
     "ohash": "^1.1.2",
-    "prisma": "5.2.0",
+    "prisma": "4.8.1",
+    "prisma-v5": "npm:prisma@5.9.0",
     "prompts": "^2.4.2",
     "protobufjs": "^7.1.1",
     "react-native-uuid": "^2.0.1",
--- a/clients/typescript/src/cli/migrations/migrate.ts
+++ b/clients/typescript/src/cli/migrations/migrate.ts
@@ -26,6 +26,7 @@ import { withConfig } from '../configure/command-with-config'
 // We use the same method to resolve the path to `@electric-sql/prisma-generator`.
 const require = Module.createRequire(import.meta.url)
 const prismaPath = require.resolve('prisma')
+const prismaV5Path = require.resolve('prisma-v5')
 const generatorPath = path.join(
   path.dirname(require.resolve('@electric-sql/prisma-generator')),
   'bin.js'
@@ -250,7 +251,7 @@ async function _generate(opts: Omit<GeneratorOptions, 'watch'>) {
     // Fetch the migrations from Electric endpoint and write them into `tmpFolder`
     await fetchMigrations(migrationEndpoint, migrationsFolder, tmpFolder)
 
-    const prismaSchema = await createPrismaSchema(tmpFolder, opts)
+    const prismaSchema = await createIntrospectionSchema(tmpFolder, opts)
 
     // Introspect the created DB to update the Prisma schema
     await introspectDB(prismaSchema)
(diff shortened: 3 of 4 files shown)


samwillis (approved):
I have tested this, I'm confident that it fixes the issue with no regressions. However, there is the flaky applying of migration on versions of Prisma lower than 4.12 I noted on Discord.

I'm approving, but we should confirm that that is an unrelated bug.

ImSingee:
Does it work with pnpm? After upgrading I got a new zod error

image

ImSingee:
Even 

```
npx create-electric-app@latest test
cd test
# change version of electric-sql to 0.9.2
pnpm install
pnpm add zod
pnpm backend:start
pnpm db:migrate
pnpm client:generate
pnpm build
```

not works

The error is similar to my above



```

> test@0.1.0 build /private/tmp/test2/test
> tsc && vite build

src/generated/client/index.ts:3:38 - error TS6133: 'Relation' is declared but its value is never read.

3 import { type TableSchema, DbSchema, Relation, ElectricClient, type HKT } from 'electric-sql/client/model';
                                       ~~~~~~~~

src/generated/client/index.ts:63:14 - error TS2322: Type 'ZodObject<{ value: ZodOptional<ZodString>; }, "strict", ZodTypeAny, { value?: string | undefined; }, { value?: string | undefined; }>' is not assignable to type 'ZodType<ItemsWhereUniqueInput, ZodTypeDef, ItemsWhereUniqueInput>'.
  Types of property '_type' are incompatible.
    Type '{ value?: string | undefined; }' is not assignable to type 'ItemsWhereUniqueInput'.
      Type '{ value?: string | undefined; }' is not assignable to type '{ value: string; } & { value?: string | undefined; AND?: ItemsWhereInput | ItemsWhereInput[] | undefined; OR?: ItemsWhereInput[] | undefined; NOT?: ItemsWhereInput | ... 1 more ... | undefined; }'.
        Type '{ value?: string | undefined; }' is not assignable to type '{ value: string; }'.
          Types of property 'value' are incompatible.
            Type 'string | undefined' is not assignable to type 'string'.
              Type 'undefined' is not assignable to type 'string'.

63 export const ItemsWhereUniqueInputSchema: z.ZodType<Prisma.ItemsWhereUniqueInput> = z.object({
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/generated/client/index.ts:3
```

djbutler:
I'm getting the same errors as @ImSingee . Here is my schema:

```
CREATE TABLE IF NOT EXISTS images (
  id                  UUID PRIMARY KEY NOT NULL,
  scanId              UUID REFERENCES scans(id),
  ...
};

CREATE TABLE IF NOT EXISTS scans (
  id                  UUID PRIMARY KEY NOT NULL,
  ...
);
``` 

I get lots of type errors of the form:

`Property 'id' is optional in type '{ id?: string; ... }' but required in type '...'`

PR #910, opened 2024-02-01
feat(examples): Linearlite multiple projects

samwillis:
To do:

- [x] UI for new projects (currently its using a Browser prompt input)
- [x] Show/hide the sync button when a project is synced (Requires sync status api)
- [x] Unsync project button (Requires an unsung api)


msfstef:
@samwillis updated the project to be ready for the unsub and status APIs

I also updated the `load-data` script to also use a `projects.json` file and assign each issue to a random project - I know that the issues don't match the project names but it makes it easier to get a good feel of multi-project syncing

samwillis:
> I also updated the `load-data` script to also use a `projects.json` file and assign each issue to a random project - I know that the issues don't match the project names but it makes it easier to get a good feel of multi-project syncing

Sounds good for now. We should make a script that produces a dump from a public GutHub using its api. Then we can have multiple "real" projects in there.

msfstef:
@samwillis @kevin-dp could you have a quick look at this PR so we merge Linearlite multiproject? I want to deploy it with 0.12.1 but I'd rather merge the multiproject work first and update the version as soon as we release

examples/linearlite/.eslintrc.cjs
```diff
@@ -27,5 +27,12 @@ module.exports = {
       { allowConstantExport: true },
     ],
   },
-  ignorePatterns: ['**/*/mock.ts', 'dist', '.eslintrc.cjs', 'src/generated/*'],
+  ignorePatterns: [
+    '**/*/mock.ts',
+    'db/**/*',
+    'src/generated/*',
+    'src/generated/client/**/*',
```
kevin-dp, line 34: is there anything in `src/generated` we want to keep?
If not, we can go with one rule: `src/generated/**/*`

examples/linearlite/db/load_data.js
```diff
@@ -28,6 +40,17 @@ async function makeInsertQuery(db, table, data) {
   `)
 }
 
+async function upsertProject(db, data) {
+  const columns = Object.keys(data)
+  const columnsNames = columns.join(', ')
+  const values = columns.map((column) => data[column])
+  return await db.query(sql`
+    INSERT INTO project (${sql(columnsNames)})
```
kevin-dp, line 48: Would be safest if all queries quote the table and column names because otherwise this won't work if some tables have uppercase letters.

kevin-dp (approved):
Great work, all seems good 👍

examples/linearlite/.eslintrc.cjs
msfstef: whoops, rebase artefact!

PR #913, opened 2024-02-02
fix(cli): Add `ELECTRIC_WRITE_TO_PG_MODE` option and fix default values

msfstef:
Fixes https://github.com/electric-sql/electric/pull/913

I've added `ELECTRIC_WRITE_TO_PG_MODE` as an option to the CLI to be able to put it in `direct_writes` mode.

Changed `DATABASE_USE_IPV6` and `ELECTRIC_USE_IPV6` to be `true` by default as per the docs.

Also changed `DATABASE_REQUIRE_SSL` to be `true`  and `AUTH_MODE` to be `secure` by default as this is the default for Electric as well, but I do understand that we might want them to be `false` and `insecure` for easier setup. Let me know if I should roll those back.

Finally, I've added a small test suite that really just re-confirms that the environment variables are present and some defaults are as expected, mostly to force us to update both the code and the test when touching the CLI. I find this safeguard to help, but happy to remove it.


clients/typescript/src/cli/config-options.ts
```diff
@@ -13,7 +13,7 @@ import { LIB_VERSION } from '../version'
 const minorVersion = LIB_VERSION.split('.').slice(0, 2).join('.')
 
 
-// Name will be prefixed with ELECTRIC_ as environment variables.
+// Name will be prefixed with ELECDATABASE_REQUIRE_SSLTRIC_ as environment variables.
```
alco, line 14: Funky ^-^

clients/typescript/src/cli/config-options.ts
```diff
@@ -236,7 +252,7 @@ export const configOptions : Record<string, AnyConfigOption> = {
     groups: ['electric', 'client', 'proxy'],
   },
   AUTH_MODE: {
-    defaultVal: 'insecure',
+    defaultVal: 'secure',
```
alco, line 255: I've discussed this with Sam before. See https://linear.app/electric-sql/issue/VAX-1590/add-a-mention-to-our-docs-and-perhaps-readme-of-the-starter-app-about.

We want to keep the `insecure` default for simple onboarding and document why this is different from sync service's default. Any other option adds hurdles on the path from zero to a working Electric-powered app and we want to avoid that as much as possible.

clients/typescript/src/cli/config-options.ts
```diff
@@ -151,27 +151,37 @@ export const configOptions : Record<string, AnyConfigOption> = {
 
   // *** Electric options ***
   DATABASE_REQUIRE_SSL: {
-    defaultVal: false,
+    defaultVal: true,
```
alco, line 153: Enabling this by default will make the bundled Postgres unreachable unless we also include a self-signed certificate with it. And if a include a certificate, we'll probably need to update it periodically. Better keep the old default.

alco (commented):
Nice one :heart: 

I've provided some background for why we'd want to keep some of the controversial defaults unchanged.

clients/typescript/src/cli/config-options.ts
msfstef: whoops

clients/typescript/src/cli/config-options.ts
samwillis: If we do this, which we should, I would like to also add a --insecure and -I flag as shortcuts for --auth-mode=insecure.

msfstef: Makes sense (for both this and the SSL option) - added a comment on both of those to point out that they differ from the Electric service's defaults.

msfstef:
Rolled the SSL and AUTH_MODE defaults back and added a comment on their default values to point out they differ from the sync service's defaults - this _might_ be slightly confusing for someone reading the docs but I do agree that it's too much friction for a CLI tool meant to get you set up with Electric.

Slightly related to CLI usability and env vars, and I might raise a separate issue, a thing I experienced was that because we set all env vars with an `ELECTRIC_` prefix for the CLI but any errors coming from those vars are printed by Electric without the prefix it can be slightly confusing, but it's a minor issue.

clients/typescript/src/cli/config-options.ts
```diff
@@ -149,34 +150,62 @@ export const configOptions = {
...
-      connections will only be accepted over IPv6 when this setting is enabled.
+      Set to false to force Electric to only listen on IPv4 interfaces.
+
+      By default, Electric will accept inbound connections over both IPv6 and IPv4
+      when running on Linux. On Windows and some BSD systems inbound connections
+      over IPv4 will not be accepted unless this option is disabled.
+    `,
+    groups: ['electric'],
+  },
+  ELECTRIC_WRITE_TO_PG_MODE: {
```
alco, line 182: We need to add this to the list of environment variables in `clients/typescript/src/cli/docker-commands/docker/compose.yaml` for this setting to take effect when `electric-sql start` is executed.

msfstef: good catch! forgot about it

alco (approved):
:muscle:

PR #915, opened 2024-02-02
fix(electric): Downgrade Erlang from version 26 to 25

alco:
With the latest version of OTP Electric sends an invalid startup packet to Postgres which causes Postgres to close the connection.

We don't know what the underlying issues is so we're reverting to a previous version of OTP for now.


magnetised (approved):
weird issue. seems like the right fix though

PR #918, opened 2024-02-03
Support for op-sqlite

gokul1099:
This PR adds compatibility with [op-sqlite](https://github.com/OP-Engineering/op-sqlite/tree/main) via a new adapter. Tested with react native example app.


clients/typescript/test/drivers/op-sqlite.test.ts
```diff
@@ -0,0 +1,39 @@
+import test from "ava"
+
+import {DatabaseAdapter} from "../../src/drivers/op-sqlite/adapter"
+import {MockDatabase} from "../../src/drivers/op-sqlite/mock"
+
+
+test("database adapted run works",async (t)=>{
```
msfstef, line 7: nit: typo in "adapter"

clients/typescript/src/drivers/op-sqlite/adapter.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Row, SqlValue } from '../../util/types'
+import { SerialDatabaseAdapter as GenericDatabaseAdapter } from '../generic'
+import { Statement } from '../../util'
+import { Database } from './database'
+import { RunResult } from '../../electric/adapter';
+
+
+export class DatabaseAdapter extends GenericDatabaseAdapter {
+  readonly db: Database
+  #rowsAffected = 0
```
msfstef, line 11: You should be able to remove this property as both `execBatch` and `_run` can return rows affected without keeping a running count.

clients/typescript/src/drivers/op-sqlite/adapter.ts
```diff
@@ -0,0 +1,49 @@
...
+      values: (args ?? []) as SqlValue[],
+     }))
+     
+     const result = this.db.executeBatch(set)
+
+     this.#rowsAffected = result?.rowsAffected ?? 0
+     return {rowsAffected : this.#rowsAffected}
+
+  }
+  getRowsModified(): number {
```
msfstef, line 45: Can remove this method as well (not part of the adapter interface and not required to calculate rows affected, see comment above)

clients/typescript/src/drivers/op-sqlite/adapter.ts
```diff
@@ -0,0 +1,49 @@
...
+  readonly db: Database
+  #rowsAffected = 0
+
+  constructor(db: Database) {
+    super()
+    this.db = db
+  }
+
+  async _query(statement: Statement): Promise<Row[]> {
+    const result = this.db.execute(
```
msfstef, line 19: Might be better to use the asynchronous `executeAsync` method to [avoid blocking the UI](https://github.com/OP-Engineering/op-sqlite?tab=readme-ov-file#async-operations).

clients/typescript/src/drivers/op-sqlite/adapter.ts
```diff
@@ -0,0 +1,49 @@
...
+    return {rowsAffected : result.rowsAffected}
+  }
+
+  async execBatch(statements: Statement[]): Promise<RunResult>{
+    const set: any[] = statements.map(({sql,args})=>({
+      statement:sql,
+      values: (args ?? []) as SqlValue[],
+     }))
+     
+     const result = this.db.executeBatch(set)
```
msfstef, line 39: See comment above about async operations, can use `executeBatchAsync` here instead.

clients/typescript/package.json
```diff
@@ -88,6 +90,7 @@
         "./dist/frameworks/react/index.d.ts"
       ],
       "react-native": [
+        "./dist/drivers/op-sqlite/index.d.ts",
```
msfstef, line 93: since it's its own export, might be worth adding the type under its own entry
```json
"opsqlite": [
    "./dist/drivers/op-sqlite/index.d.ts",
] 
```

clients/typescript/package.json
msfstef, line 1: fyi you can run `pnpm prettier . --write` to format everything appropriately
you can use `pnpm check-styleguide` to ensure it passes lint, and can also do `pnpm typecheck` to make sure it compiles with no issues

msfstef (commented):
Looks good, thank you for the PR! Left some comments, most importantly about the use of the synchronous APIs.

clients/typescript/src/drivers/op-sqlite/mock.ts
```diff
@@ -0,0 +1,26 @@
+import { BatchQueryResult, QueryResult } from '@op-engineering/op-sqlite'
+import { DbName } from '../../util/types'
+import { Database } from './database'
+
+export class MockDatabase implements Database {
+  constructor(public dbname: DbName, public fail?: Error) {}
```
msfstef, line 6: nit: `dbname` should be `dbName` for consistency (what `electrify` reads)

clients/typescript/src/drivers/op-sqlite/adapter.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  async execBatch(statements: Statement[]): Promise<RunResult> {
+    const set: any[] = statements.map(({ sql, args }) => ({
+      statement: sql,
+      values: (args ?? []) as SqlValue[],
+    }))
+
+    const result = await this.db.executeBatchAsync(set)
+
+    return { rowsAffected: result?.rowsAffected ?? 0 }
```
msfstef, line 32: can `result` actually be null here? looking at the OPSQLite implementation this should be unnecessary.
Also, `rowsAffected` can be null and that seems odd as well, [the comments seem to reference a failure state](https://github.com/OP-Engineering/op-sqlite/blob/355c555c700e234b0ffea7ab06a82762b5783f6d/src/index.ts#L111) that is not encoded in the result, and this differs from the `QueryResult` interface for `executeAsync`.

If this error state is indeed possible without the operation throwing then falling back to 0 rows affected silently might be an issue, although I suspect it's a deprecated comment and interface?

clients/typescript/src/drivers/op-sqlite/adapter.ts
```diff
@@ -0,0 +1,34 @@
...
+  readonly db: Database
+
+  constructor(db: Database) {
+    super()
+    this.db = db
+  }
+
+  async _query(statement: Statement): Promise<Row[]> {
+    const result = await this.db.executeAsync(statement.sql, statement.args)
+    return result.rows?._array ?? []
```
msfstef, line 17: Similarly to my comment below about the nullable rows affected in `executeBatchAsync`, OPSQLite seems to indicate that `rows` is undefined [only in an error state](https://github.com/OP-Engineering/op-sqlite/blame/355c555c700e234b0ffea7ab06a82762b5783f6d/src/index.ts#L61), although this `status` property doesn't appear anywhere.

Again, just want to confirm that there is not a case where `rows` is undefined because of a silent error and it's just an out of date specification (?)

clients/typescript/src/drivers/op-sqlite/database.ts
```diff
@@ -0,0 +1,9 @@
+import { OPSQLiteConnection } from '@op-engineering/op-sqlite'
+import { DbName } from '../../util'
+
+type OriginalDatabase = OPSQLiteConnection
+
+export interface Database
+  extends Pick<OriginalDatabase, 'executeAsync' | 'executeBatchAsync'> {
+  dbName?: DbName
```
msfstef, line 8: I see this is the case for other drivers as well, but as I understand this the client would need to either alter, extend, or create a wrapper for the respective driver instance to expose the `dbName` before passing it to the `electrify` call, which feels a bit awkward (?)

Tagging @kevin-dp for this discussion: would it make more sense for the `dbName` to be explicitly passed as an argument to the `electrify` call for drivers that don't expose it? It does decouple the db name in the driver initialisation from the electrification which can be bug prone, but I feel other alternatives would be to expose the adapter or provide our own wrapper as well?

Either way since this is the case for other drivers as well you can ignore my comment as it would require a breaking change, just wanted to flag it here.

clients/typescript/src/drivers/op-sqlite/database.ts
```diff
@@ -0,0 +1,9 @@
+import { OPSQLiteConnection } from '@op-engineering/op-sqlite'
+import { DbName } from '../../util'
+
+type OriginalDatabase = OPSQLiteConnection
```
msfstef, line 4: nit: is this necessary to reference the type?

msfstef (approved):
Looks good to me! Left a few more minor comments that have more to do with OPSQLite's interfaces but worth addressing.

@kevin-dp if you can have a quick look as well I think this is good to go!

clients/typescript/package.json
```diff
@@ -221,10 +226,11 @@
     "typeorm": "^0.3.9",
     "typescript": "^5.3.3",
     "wa-sqlite": "rhashimoto/wa-sqlite#semver:^0.9.8",
-    "web-worker": "^1.2.0"
+    "web-worker": "^1.2.0",
+    "@op-engineering/op-sqlite": ">= 2.0.16",
```
kevin-dp, line 230: Is there a reason it needs to be a dev dependency and not a peer dependency like the others?
If it can be a peer dependency then `@capacitor-community/sqlite` can probably also remain a peer dependency. Also it it present in `peerDependenciesMeta` so i'm guessing it was meant to be a peer dependency.

clients/typescript/src/drivers/op-sqlite/adapter.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+  async _run(statement: Statement): Promise<RunResult> {
+    const result = await this.db.executeAsync(statement.sql, statement.args)
+    return { rowsAffected: result.rowsAffected }
+  }
+
+  async execBatch(statements: Statement[]): Promise<RunResult> {
```
kevin-dp, line 24: This `execBatch` method is not used because your adapter extends the `SerialDatabaseAdapter`.
You need to extend the `BatchDatabaseAdapter` instead.

clients/typescript/src/drivers/op-sqlite/adapter.ts
```diff
@@ -0,0 +1,34 @@
...
+    const set: any[] = statements.map(({ sql, args }) => ({
```
kevin-dp, line 25: Can we improve this type? According to op-sqlite docs this should be of type `SQLBatchTuple[]`.

clients/typescript/src/drivers/op-sqlite/database.ts
kevin-dp: Indeed, it would require adding the `dbName` property to the database before passing it to the driver's `electrify` call, which is a bit odd. For databases that don't track the name we could indeed pass it as an additional argument to the driver's `electrify` function.

kevin-dp, line 8: This should not be optional because in the `electrify` function we rely on it being present since we do `const dbName: DbName = db.dbName!`. Once we make it required we can also remove the `!` in `electrify`.

clients/typescript/src/drivers/op-sqlite/mock.ts
kevin-dp, line 6: `fail` is not used, either remove it or modify the unit test to also check the case where it fails.

clients/typescript/src/drivers/op-sqlite/mock.ts
```diff
@@ -0,0 +1,26 @@
...
+      rowsAffected: 1,
+      rows: {
+        _array: [
+          {
+            column1: 'text1',
+            column2: 'text2',
+          },
+        ],
+        length: 1,
+        item: (idx: number) => idx,
```
kevin-dp, line 19: is this function supposed to return the `idx`-th row from the results?
If that is the case, we could implement it like this:
```ts
executeAsync(): Promise<QueryResult> {
    const _array = [
      {
        column1: 'text1',
        column2: 'text2',
      },
    ]
    return Promise.resolve({
      rowsAffected: 1,
      rows: {
        _array,
        length: 1,
        item: (idx: number) => _array[idx],
      },
  })
}
```

kevin-dp (approved):
Thanks for your contribution!
The implement looks good, i left a couple of comments.
After addressing @msfstef's comments and mine this is ready to be merged 👍

clients/typescript/src/drivers/op-sqlite/adapter.ts
gokul1099: @kevin-dp  SQLBatchTuple[] is specific to op-sqlite right? But here the adapter extends from BatchDatabaseAdapter requires statements as Statement[] type. How can I  include SQLBatchTuple [] here?

clients/typescript/package.json
gokul1099: @kevin-dp  will check and move to peerDependencies

clients/typescript/src/drivers/op-sqlite/adapter.ts
gokul1099: @msfstef  Yes, I think it's  a deprecated comment

clients/typescript/src/drivers/op-sqlite/adapter.ts
gokul1099: @msfstef  I think in the case of silent error the rows will be undefined. executeBatchAsync will return a promise, so can we use the failure callback to handle the error case ?

msfstef: We don't need to handle the error as it will be caught by the callers of the adapter (AFAIK there's no error wrapper for these adapters right now), my main concern is that if something does go wrong it _should_ throw and not just silently return an empty list of rows.

If you try running a bogus query with the `executeAsync` method does it return a promise rejection or does it resolve with undefined `rows`?

If it's the former, then I think it's fine to just assert that the rows will never be undefined (and maybe leave a comment about this), so `return result.rows!._array`

If it's the latter, there should be some way to catch/expose the underlying driver error but at the very least we should catch the rows being undefined and throw an error like "There was an error with the query blah blah".

gokul1099: I tried running some incorrect queries, the executeAsync method returns a promise rejection. 

Screenshot 2024-02-19 at 7 48 07 PM

clients/typescript/src/drivers/op-sqlite/adapter.ts
kevin-dp: I meant to replace the `any`on line 25, so `const set: SQLBatchTuple[] = ...`

gokul1099:
@kevin-dp  @msfstef  I have made all the changes as per your review. Please check

clients/typescript/package.json
```diff
@@ -53,6 +54,8 @@
     "./capacitor": "./dist/drivers/capacitor-sqlite/index.js",
     "./expo": "./dist/drivers/expo-sqlite/index.js",
     "./expo-next": "./dist/drivers/expo-sqlite-next/index.js",
+    "./opsqlite":"./dist/drivers/op-sqlite/index.js",
```
msfstef, line 57: duplicate export here

clients/typescript/src/drivers/op-sqlite/database.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export interface Database
+  extends Pick<OriginalDatabase, 'executeAsync' | 'executeBatchAsync'> {
+  dbName: DbName
```
msfstef, line 8: I think having the `dbName` as an extra `electrify` argument rather than an enforced property that we know doesn't exist on the original db connection instance is a better pattern (as discussed [here](https://github.com/electric-sql/electric/pull/918#discussion_r1485921379))

so something like `electrify(db, dbName, dbDescription, ..etc)`

clients/typescript/package.json
```diff
@@ -238,9 +244,11 @@
     "vue": "^3.4.19",
     "vue-tsc": "^1.8.27",
     "wa-sqlite": "rhashimoto/wa-sqlite#semver:^0.9.8",
-    "web-worker": "^1.2.0"
+    "web-worker": "^1.2.0",
+    "@capacitor-community/sqlite": ">= 5.4.1"
```
msfstef, line 248: this should be a peer dependency only

clients/typescript/test/drivers/op-sqlite.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  const sql = [
+    { sql: 'select * from electric;', args: [] },
+    { sql: 'select * from opsqlite', args: [] },
+  ]
+  const result = await adapter.execBatch(sql)
+
+  t.is(result.rowsAffected, 1)
+})
+
+test("database adapter reject promise on failure",async(t)=>{
```
msfstef, line 43: nit: don't forget to run the formatter as well

clients/typescript/src/drivers/op-sqlite/database.ts
msfstef: bump for this nit - we should be able to just have `Pick<OPSQLiteConnection, 'executeAsync' | 'executeBatchAsync'>`

clients/typescript/src/drivers/op-sqlite/database.ts
gokul1099: @msfstef  Should I ignore this for now, since it has to be incorporated for all the drivers, right?

clients/typescript/package.json
```diff
@@ -53,6 +54,7 @@
     "./capacitor": "./dist/drivers/capacitor-sqlite/index.js",
     "./expo": "./dist/drivers/expo-sqlite/index.js",
     "./expo-next": "./dist/drivers/expo-sqlite-next/index.js",
+    "./opsqlite": "./dist/drivers/op-sqlite/index.js",
```
msfstef, line 57: nit: mismatch between `opsqlite` export and `op-sqlite` types - maybe keep `op-sqlite` everywhere for consistency

clients/typescript/src/drivers/op-sqlite/database.ts
msfstef: There's not need to incorporate it for all the drivers, some of them do indeed expose the database name, and each driver's `electrify` call has its own API, so it should be fine to just move this requirement from the database instance to an argument of `electrify`

clients/typescript/package.json
```diff
@@ -251,9 +256,13 @@
     "react-native-sqlite-storage": ">= 6.0.0",
     "typeorm": ">=0.3.0",
     "vue": ">=3.0.0",
-    "wa-sqlite": "rhashimoto/wa-sqlite#semver:^0.9.8"
+    "wa-sqlite": "rhashimoto/wa-sqlite#semver:^0.9.8",
+    "web-worker": "^1.2.0"
```
msfstef, line 260: is there a reason the `web-worker` dependency was moved to `peerDependencies`? I'm unsure about whether it's used at all but having it in `peerDependencies` without an optional flag will make a warning appear on electric installations without it. If it was moved because it is unused I'd suggest placing it back and it can be cleaned up on a separate PR

clients/typescript/src/drivers/op-sqlite/database.ts
gokul1099: @msfstef  For this I just have to remove dbName in database interface right?? And for `electrify` call already we have `db` argument. Should I have to add `dnName` also ??

msfstef: Yep I'd say this becomes:
```ts
export interface Database extends Pick<OriginalDatabase, 'executeAsync' | 'executeBatchAsync'>
```
and the `electrify` call gets an additional argument
```ts
export const electrify = async <T extends Database, DB extends DbSchema<any>>(
  db: T,
  dbName: DbName, // accept it as extra argument
  dbDescription: DB,
  config: ElectricConfig,
  opts?: ElectrifyOptions
): Promise<ElectricClient<DB>> => {
  // no longer need to retrieve it from the db instance
  // const dbName: DbName = db.dbName!
  // ...
```

ideally we should be able to extract it from the DB connection instance but [op-sqlite doesn't seem to make it available](https://github.com/OP-Engineering/op-sqlite/blob/cfea4c3ade66501e56b46342af727b623fbca5a6/src/index.ts#L397-L460).

msfstef:
@gokul1099 PR looking good!

I think you might need to rebase and/or run `pnpm install` to get a consistent lockfile as I see tests are failing because of that.
If you can, please also add a changeset by going into the root dir and doing `pnpm changeset` and following instructions.

balegas:
@msfstef please open an internal follow up issue for us to update our react-native example to use this.

@gokul1099 thanks so much for the work.

.changeset/five-games-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"electric-sql": minor
```
msfstef, line 2: nit: we currently do patch releases for everything as we're still in alpha

PR #921, opened 2024-02-05
feat (client): driver for tauri sqlite

kevin-dp:


clients/typescript/package.json
```diff
@@ -177,6 +181,7 @@
   "devDependencies": {
     "@electric-sql/prisma-generator": "workspace:*",
     "@ikscodes/browser-env": "^1.0.0",
+    "@tauri-apps/plugin-sql": "2.0.0-alpha.5",
```
msfstef, line 184: This is added as both a peer and dev dependency, and not marked as optional - is this intended?

clients/typescript/src/drivers/tauri-sqlite/index.ts
msfstef, line 1: nit: there's no driver mock and test files that we have for other drivers - they don't test much but might be worth adding them at least for the logic of name assignment and valid instantiation?

msfstef (approved):
Looks good!

clients/typescript/package.json
samwillis: I believe it is. All the drivers are required as devDependencies for development and in peerDependencies to show that they can be optional installed in a project when support is required.

clients/typescript/src/drivers/tauri-sqlite/index.ts
samwillis: I don't think it's an issue with this driver for now. We are going to add some e2e tests for the drivers at some point.

samwillis (approved):
LGTM

PR #922, opened 2024-02-06
Migrate squel to kysely

ygarg465:
This PR is just a PoC of migrating squel to kysely for now.


clients/typescript/src/client/model/kyselyBuilder.ts
```diff
@@ -0,0 +1,468 @@
...
+        }
+
+        const whereObject = i.where
+        const identificationFields = this.getFields(whereObject, idRequired)
+
+        if (!this.shapeManager.hasBeenSubscribed(this._tableName))
+            Log.debug('Reading from unsynced table ' + this._tableName)
+
+        const query = db.selectFrom(this._tableName)
+        const addFieldSelectionP = this.addFieldSelection.bind(this, i, selectWhereFields ? identificationFields : [])
```
kevin-dp, line 123: nit: could you add back the comment right before this line to explain what it does:
```
// only select the fields provided in `i.select` and the ones in `i.where`
```

clients/typescript/src/client/model/kyselyBuilder.ts
```diff
@@ -0,0 +1,468 @@
...
+        // which are not columns and thus should not be selected
+        const fields = identificationFields
+            .filter((f) => this._fields.includes(f))
+            .concat(selectedFields)
+            .map((f) => this.castBigIntToText(f))
+
+        return query.select(fields)
+    }
+
+    castBigIntToText(field: string) {
```
kevin-dp, line 197: Here also, put back the comment that clarifies why this is needed:
```
/**
   * Casts a field to TEXT if it is of type BigInt
   * because not all adapters deal well with BigInts
   * (e.g. better-sqlite3 requires BigInt support to be enabled
   *       but then all integers are returned as BigInt...)
   * The DAL will convert the string into a BigInt in the `fromSqlite` function from `../conversions/sqlite.ts`.
   */
```

clients/typescript/src/client/model/kyselyBuilder.ts
```diff
@@ -0,0 +1,468 @@
...
+import Log from "loglevel";
+import flow from "lodash.flow";
+import * as z from "zod";
+
+type AnyFindInput = FindInput<any, any, any, any, any>
+
+const db = new Kysely<any>({
+    dialect: {
+        createAdapter() {
+            return new PostgresAdapter();
```
kevin-dp, line 27: We should use SQLite adapter, introspector and compiler.
The squel version used PG because of a feature of squel that only worked when using the Postgres flavour.
But for Kysely this should not be needed.

ygarg465:
Tests are updated a little bit to match syntax generated by kysely.
1) SQL keywords are in lower case
2) table and column names are in double quotes
3) parameters are tested separately instead of inline

ygarg465:
A problem I am currently facing to complete this is migrating the method makeBooleanFilter.

With sequel library we were just concatenating the statements but this does not work very easily in kysely.

The "OR" statements needs to be add by calling or method which does fit very well in the current style of code written here.

There is another way of using raw SQL which does support parameters automatically however I am unable to implement that.

ygarg465:
Hey @kevin-dp, yes that is the code copied from squel implementation and is not working with Kysely.

The problem is that all the `or` statements need to be added by calling the `or` method of the Kysely's respective object which does not fill well in the current implementation style of the code.

And there is another way to achieve this which is raw SQL, that is I am unable to understand and implement.

kevin-dp:
@ygarg465 Do you understand the changes that are needed to make it work with `or`? I'm fine with changing to a better approach to build these queries (using Kysely). I would avoid dropping down to raw SQL as i don't think that will mix well with the rest of the builder that uses Kysely. Unfortunately, I don't have the capacity to help on the implementation of the builder right now.

ygarg465:
@kevin-dp I have to look at the implementation again for `or` as it's been some time since I have worked with the library.
I will get back to you with a implementation for the same that you can review

KyleAMathews:

PR #928, opened 2024-02-07
chore(electric): Change the default PG_PROXY_PASSWORD to match web-wa-sqlite

alco:
This removes the only hurdle that was preventing the stock code for examples/web-wa-sqlite from working with the locally running sync service started with `iex -S mix`.


magnetised (approved):
👍🏻

PR #929, opened 2024-02-07
fix(electric): Rewrite __validate_table_constraints() without array_agg()

alco:
The old implementation kept producing different orderings between runs:

    1) test table electrification rejects columns with CHECK, UNIQUE or EXCLUDE constraints (Electric.Postgres.ExtensionTest)
       test/electric/postgres/extension_test.exs:519
       Assertion with == failed
       left:  "Cannot electrify t1 [...]:\n  t1\n  \"Ts\"\n  uu"
       right: "Cannot electrify t1 [...]:\n  \"Ts\"\n  t1\n  uu"


alco:
See https://github.com/electric-sql/electric/actions/runs/7813426880/job/21312525938?pr=928 for a real faliure example.

alco:
> Hah, I fixed this in my branch by changing the test assertion but this is probably the right solution!

I don't why, but it flickers. Changing the order in the test either way would still leave the room for failure when the order flips again. That's why I've removed the use of `array_agg` entirely.

PR #930, opened 2024-02-07
chore(cli): Improve CLI options test

davidmartos96:
Minor improvement to the new config options test. With this we can detect when a new option is added and it doesn't get added to the list.


msfstef:
great idea, looks good!

msfstef (approved):

PR #937, opened 2024-02-08
feat(electric): Enable SSL certificate validation for database connections

alco:
Closes VAX-1446.

### DIFF
--- a/.changeset/tender-falcons-speak.md
+++ b/.changeset/tender-falcons-speak.md
@@ -1,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
+Enable SSL certificate validation for database connections when DATABASE_REQUIRE_SSL=true.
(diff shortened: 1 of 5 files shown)


magnetised (commented):
Looks good, but don't we also need to support self-signed/custom cacerts? Some kind of cacert_file setting that would allow for the use of an in-house CA. `:public_key.cacerts_load/1` allows for loading via a file.

alco:
> Looks good, but don't we also need to support self-signed/custom cacerts? Some kind of cacert_file setting that would allow for the use of an in-house CA. `:public_key.cacerts_load/1` allows for loading via a file.

Eventually, yes. But that would be untimely and would require a less short&sweet PR.

For now, if you have a hostname that cannot be validated with the default certificate chain provided by the OS, you can start Electric with `DATABASE_REQUIRE_SSL=false` to avoid cert validation failures.

magnetised (approved):

PR #942, opened 2024-02-13
Change the way electrify() captures DDL statements

alco:
The old implementation incorrectly stores both the CREATE TYPE and CREATE TABLE statements in a single string. This was causing issues with translating those statements into SQLite dialect for the purpose of including them in SatOpMigrate messages.

There's no provision for users who already have electrified enums. We will ask them to reset their DB and start from scratch once this change is released.


magnetised (approved):
Hairy stuff.

PR #944, opened 2024-02-13
fix(cli): Make `--debug` flag for CLI `generate` command always maintain migrations folders

msfstef:
As @alco pointed out, the expected behaviour of the `--debug` flag would be to always retain the temporary migrations folder for inspection and debugging regardless of whether the generation succeeds or not.

I've added a few more tests to cover this behaviour for all 4 paths (with/without debug, failure/success).

I've also added the flag to the documentation for the CLI `generate` command and added more info on what the flag does in the CLI help text.

### DIFF
--- a/.changeset/chatty-lies-impress.md
+++ b/.changeset/chatty-lies-impress.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Make `--debug` flag in CLI `generate` command always retain temporary migrations folders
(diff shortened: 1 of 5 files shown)


alco (approved):
Great! Thanks, @msfstef!

I have taken the doc change out of this PR and moved it into a separate one - https://github.com/electric-sql/electric/pull/945.

There was a lot of noise caused by your editor autoformatting the file. I've cleaned that up. Generally, we make changes to the docs separately from code changes to avoid updating the published docs before a release that includes the documented change is published.

msfstef:
Ah thanks for the info about the docs update approach! And sorry for the formatting, I had auto-format on save enabled for markdown and missed the changes.

PR #945, opened 2024-02-13
chore(docs): Document the updated behaviour of `generate --debug`

alco:
Follow-up to https://github.com/electric-sql/electric/pull/944.


samwillis (approved):
👍

PR #946, opened 2024-02-13
chore(electric): Limit client connections when the sync service's Postgres connection is down

alco:
Fixes VAX-1290.

This change also addresses the problem where a request sent to `/api/status` could time out if the Postgres Manager process was itself blocked on an IO call at the time.

### DIFF
--- a/components/electric/lib/electric/plug/satellite_websocket_plug.ex
+++ b/components/electric/lib/electric/plug/satellite_websocket_plug.ex
@@ -63,7 +63,7 @@ defmodule Electric.Plug.SatelliteWebsocketPlug do
 
       :error ->
         reason = "Missing satellite websocket subprotocol"
-        Logger.debug("Client WebSocket connection failed with reason: #{reason}")
+        Logger.warning("Client WebSocket connection failed with reason: #{reason}")
         {:error, 400, reason}
     end
   end
@@ -75,7 +75,7 @@ defmodule Electric.Plug.SatelliteWebsocketPlug do
       reason =
         "Cannot connect satellite version #{assigns.satellite_vsn}: this server requires #{requirements}"
 
-      Logger.debug("Client WebSocket connection failed with reason: #{reason}")
+      Logger.warning("Client WebSocket connection failed with reason: #{reason}")
       {:error, 400, reason}
     end
   end
--- a/components/electric/lib/electric/plug/satellite_websocket_plug.ex
+++ b/components/electric/lib/electric/plug/satellite_websocket_plug.ex
@@ -26,7 +26,8 @@ defmodule Electric.Plug.SatelliteWebsocketPlug do
   def call(conn, handler_opts) do
     with :ok <- check_if_valid_upgrade(conn),
          {:ok, conn} <- check_if_subprotocol_present(conn),
-         {:ok, conn} <- check_if_vsn_compatible(conn, with: @currently_supported_versions) do
+         {:ok, conn} <- check_if_vsn_compatible(conn, with: @currently_supported_versions),
+         :ok <- check_if_postgres_is_ready() do
       Logger.metadata(
         remote_ip: conn.remote_ip |> :inet.ntoa() |> to_string(),
         instance_id: Electric.instance_id()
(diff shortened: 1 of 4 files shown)


components/electric/lib/electric/plug/satellite_websocket_plug.ex
```diff
@@ -14,19 +16,18 @@ defmodule Electric.Plug.SatelliteWebsocketPlug do
...
       end)
 
   @currently_supported_versions ">= 0.6.0 and <= #{%{Electric.vsn() | pre: []}}"
 
   def call(conn, handler_opts) do
     with :ok <- check_if_valid_upgrade(conn),
          {:ok, conn} <- check_if_subprotocol_present(conn),
-         {:ok, conn} <- check_if_vsn_compatible(conn, with: @currently_supported_versions) do
+         {:ok, conn} <- check_if_vsn_compatible(conn, with: @currently_supported_versions),
+         :ok <- check_if_postgres_is_ready() do
```
icehaunter, line 30: I get the reasoning, but is it a good idea to deny the connection wholesale here? PG might be temporarily unavailable, and we can hold the client meanwhile. If we're merging this, I'd make sure the client has a reconnection behaviour set up correctly, which I can't recall right now

icehaunter: This also may lead to a thundering herd issue at server restart (omitting current issues with restarts)

magnetised: Perhaps we just delay the subscription ready response until pg is up

components/electric/lib/electric/replication/postgres_manager.ex
```diff
@@ -80,34 +86,54 @@ defmodule Electric.Replication.PostgresConnectorMng do
     Logger.metadata(origin: origin)
     Process.flag(:trap_exit, true)
 
-    connector_config = preflight_connector_config(connector_config)
+    # Use an ETS table to store data that are regularly looked up by other processes.
+    :ets.new(ets_table_name(origin), [:protected, :named_table, {:read_concurrency, true}])
```
magnetised, line 90:
```suggestion
    :ets.new(ets_table_name(origin), [:protected, :named_table, read_concurrency: true])
```

magnetised (approved):
👍🏻 Agree with @icehaunter that perhaps just rejecting the connection is too brutal an approach considering that the most likely scenario is that the db will come on line a few seconds later.

alco:
@icehaunter @magnetised Let me share my reasoning behind this change.

Rejecting this change would not by itself solve the thundering herd problem. First, we're not immediately killing live connections here. Second, the Satellite client has exponential backoff behaviour implemented. Third, we still need to be ready to handle thundering herd type of load to avoid getting easily DoS'ed.

Allowing a client to connect and authenticate with the server when we know it can't pull any data from the server or push local changes to it makes the client's state more complicated. Instead of just having a connectivity state that the developer can reasonably treat to mean "sync with the server is on/off", the client would have separate connectivity state and "being able to read-write data" state. If the server ends up being misconfigured after a configuration update or Postgres is down for a prolonged period, the client would keep reporting the connection as being fine, effectively masking the server's unavailability.

Having a more direct relationship between connectivity and "writability" makes debugging problems easier in development and later in prod.

Now, in saying the above, I'm glossing over the fact that, even with this change in place, live client connections won't get dropped when Postgres becomes unavailable. This is a more complicated case to handle. I think we shouldn't drop such connections immediately to smooth out remporary Electric<->Postgres connectivity problems. But there likely needs to be a timeout after which Electric should close client connections if it can't re-establish connectivity with Postgres in time.

icehaunter:
@alco I get your point, however, to expand on the thundering herd problem: yes, Electric in general should be resilient to high load, but I don't see a reason to make it harder for Electric to cope after a restart. We already have states like "being able to read/write data" - we have a handshake in each direction of the protocol replication, if the server never asks for data the client won't just start pushing (and thus won't update the replication state), and the other way around as well. Client code already treats pre-finished-handshake state as not-yet-connected, so I don't find "new client state" argument compelling given we're already explicitly handling it.

Your point about exponential backoff is correct, however it's still exactly that -- exponential. If the server goes down and up, all the clients will initially be trying to reconnect quite a lot repeatedly. Holding an open connection feels cheaper for us than going through denies on loop.

My concern is about Electric fully crashing (although unlikely under Erlang) and restarting, leading to a lot of clients reconnecting, being denied and then immediately reconnecting again

PR #947, opened 2024-02-13
Serialise GRANTS and ASSIGNS to protobuf

magnetised:
closes: VAX-1644, VAX-1647


components/electric/lib/electric/ddlx/command.ex
```diff
@@ -1,31 +1,238 @@
...
+alias Electric.Satellite.SatPerms
+
+defmodule Electric.DDLX.Command do
+  alias Electric.DDLX
+  alias Electric.DDLX.Command.PgSQL
+
+  defstruct [:cmds, :stmt, :tag, tables: []]
+
+  @type t() :: %__MODULE__{
+          cmds: struct(),
```
icehaunter, line 15: Dumb question, but why plural here?

components/electric/test/electric/postgres/proxy/query_analyser_test.exs
```diff
@@ -660,27 +661,28 @@ defmodule Electric.Postgres.Proxy.QueryAnalyserTest do
     test "ELECTRIC...", cxt do
       assert [
                %QueryAnalysis{
-                 action: {:electric, %Electric.DDLX.Command.Enable{}},
+                 action:
+                   {:electric, %Electric.DDLX.Command{cmds: %Electric.DDLX.Command.Enable{}}},
```
icehaunter, line 665: How about an `alias Electric.DDLX.Command` at the top of the file?

components/electric/lib/electric/ddlx/parser/build.ex
```diff
@@ -90,4 +93,56 @@ defmodule Electric.DDLX.Parser.Build do
...
+  def pb_columns([], ddlx) do
+    {:error,
+     %Command.Error{sql: ddlx, line: 1, position: 1, message: "Invalid empty column list"}}
+  end
+
+  def pb_columns(names, _ddlx) do
+    {:ok, %SatPerms.ColumnList{names: names}}
+  end
+
+  def pb_privs(privs) do
```
icehaunter, line 124: Nitpick: let's not safe characters here, because it's not an immediately obvious contraction to me

icehaunter: I'm also curious why are you sure of lower case `"update"` etc here

icehaunter (approved):
Only a couple of nitpicks, LGTM

components/electric/lib/electric/ddlx/command.ex
magnetised: it was a list, then I changed it and couldn't think of a better name.

components/electric/lib/electric/ddlx/parser/build.ex
magnetised: because the values come in from the ddlx parser. actually there's no reason why these come from the parser as strings and have to be translated back to atoms, I can just return atoms from the parser

components/electric/lib/electric/ddlx/command.ex
```diff
@@ -1,31 +1,238 @@
...
+      unassign.role_column,
+      unassign.role_name,
+      unassign.scope
+    ])
+  end
+
+  defp hash(terms) do
+    terms
+    |> Enum.map(&fingerprint/1)
+    |> Enum.intersperse("\n")
```
alco, line 114: Why intersperse with newlines something that is fed into a hashing function in the next step?

alco (approved):
:trophy:

components/electric/lib/electric/ddlx/command.ex
magnetised: Old habits --preventing, IIRC, length extension attacks or some-such by interspersing the terms with a delimiter that can't appear in the terms

components/electric/lib/electric/postgres/extension/migrations/20240212161153_ddlx_commands.ex
```diff
@@ -0,0 +1,44 @@
+defmodule Electric.Postgres.Extension.Migrations.Migration_20240212161153_DDLXCommands do
+  alias Electric.Postgres.Extension
+
+  @behaviour Extension.Migration
+
+  @impl true
+  def version, do: 2024_02_12_16_11_53
```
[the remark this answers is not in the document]
magnetised: The migration system doesn't depend on migrations being in order -- it's just a diff of existing versions - applied versions. Is this request just for clarity, fear of conflicts or something else?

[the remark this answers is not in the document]
magnetised: I'll re-order the migrations list to be in date order, np. I think i mis-read your comment and was thinking you wanted the version changed. my bad.

PR #950, opened 2024-02-14
fix(starter): Bring back the .gitignore template file

alco:
It was unintentionally removed in https://github.com/electric-sql/electric/pull/676.

Fixes VAX-1591.


samwillis (approved):
👍

PR #952, opened 2024-02-14
feat(electric): Validate signing keys at startup

alco:
This PR adds validation for public keys and makes it possible to use base64-encoded symmetric keys.

Fixes VAX-1584:
Validating public keys at startup provides early feedback to the developer about misconfigured auth.

Fixes VAX-1583:
Pasting a key used for token signature verification verbatim into the AUTH_JWT_KEY configuration option is problematic for binary symmetric keys.

Online JWT tools, such as jwt.io and token.dev, take base64-encoded keys by default. So newcomers to Electric get confused when they try to use their base64-encoded key with it and see signature verification errors logged by Electric.

We can detect if the value configured for AUTH_JWT_KEY is base64-encoded since the encoding has a limited set of characters.

### DIFF
--- a/.changeset/metal-bats-worry.md
+++ b/.changeset/metal-bats-worry.md
@@ -1,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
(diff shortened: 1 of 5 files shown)


icehaunter (approved):
Great solution for DX!

PR #956, opened 2024-02-15
Add support for ?dialect=postgresql to /api/migrations

alco:
Closes VAX-1659.

Passing `?dialect=postgresql` to `/api/migrations` will now result in the returned migrations to have the original Postgres DDL.

### DIFF
--- a/components/electric/lib/electric/postgres/replication.ex
+++ b/components/electric/lib/electric/postgres/replication.ex
@@ -69,7 +69,7 @@ defmodule Electric.Postgres.Replication do
         {:ok, [], []}
 
       propagate_ast ->
-        {msg, relations} = build_replication_msg(propagate_ast, schema_version, dialect)
+        {msg, relations} = build_replication_msg(propagate_ast, stmt, schema_version, dialect)
 
         {:ok, [msg], relations}
     end
@@ -89,6 +89,9 @@ defmodule Electric.Postgres.Replication do
         :ALTER_ADD_COLUMN
     end
   end
+
+  defp to_sql(_ast, stmt, Dialect.Postgresql), do: stmt
+  defp to_sql(ast, _stmt, dialect), do: Dialect.to_sql(ast, dialect)
 
   def affected_tables(stmts, dialect \\ @default_dialect) when is_list(stmts) do
     stmts
@@ -112,7 +115,7 @@ defmodule Electric.Postgres.Replication do
     []
   end
 
-  defp build_replication_msg(ast, schema_version, dialect) do
+  defp build_replication_msg(ast, stmt, schema_version, dialect) do
     affected_tables = affected_tables(ast, dialect)
 
     relations = Enum.map(affected_tables, &{&1.schema, &1.name})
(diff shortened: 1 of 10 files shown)


components/electric/lib/electric/postgres/replication.ex
```diff
@@ -168,17 +171,17 @@ defmodule Electric.Postgres.Replication do
...
-      pks: Enum.flat_map(table.constraints, &replication_msg_table_pk(&1, dialect))
+      pks: Enum.flat_map(table.constraints, &replication_msg_table_pk(&1))
     }
   end
 
-  defp replication_msg_table_col(%Proto.Column{} = column, dialect) do
+  defp replication_msg_table_col(%Proto.Column{} = column) do
     %SatOpMigrate.Column{
       name: column.name,
       pg_type: replication_msg_table_col_type(column.type),
```
magnetised, line 183: my original `{sqlite,pg}_type` thing makes no sense. why don't we just have `type` here?

components/electric/test/electric/plug_test.exs
```diff
@@ -149,6 +149,76 @@ defmodule Electric.PlugTest do
...
+                    type: :CREATE_TABLE,
+                    sql: "CREATE TABLE a (id uuid PRIMARY KEY, value text NOT NULL);"
+                  }
+                ],
+                table: %SatOpMigrate.Table{
+                  name: "a",
+                  columns: [
+                    %SatOpMigrate.Column{
+                      name: "id",
+                      sqlite_type: "TEXT",
```
magnetised, line 202: so the `pg_type` bit carries the original type info dropped by the conversion to sqlite's type. hmm. perhaps just dropping the `sqlite_` prefix and having a dialect-defined `type` which is just a string?

magnetised (approved):
I think this is exactly the way to do it. maybe we should sort out that type field in the migration message now though. what is it for? i'm guessing the intention was to make the schema info available to the clients without them having to parse the migration sql, In which case what's the best way to pass that? we could have a generic structure that gave the base type as a string and any modifiers (like length, precision etc) as a map or something? if we knew we were only ever going to support sqlite and pg, then what we have works but that makes me wary.

alco:
@magnetised 

>  we could have a generic structure that gave the base type as a string and any modifiers (like length, precision etc) as a map or something?

We have the `PgColumnType` message that can be extended with more info over time.

From looking at the client code, it doesn't need `sqlite_type` and can only use `pg_type` to determine correct value encoding/decoding based on the original PG schema.

So we can drop `sqlite_type`, rename `pg_type` to `type_info` or just `type` and populate it with PG type info regardless of the chosen dialect.

components/electric/lib/electric/postgres/replication.ex
alco: I believe I left both `sqlite_type` and `pg_type` here to avoid breaking existing code, as this was a POC. I did mark both as deprecated in the protocol definition, so ideally only `type_info` should remain.

But as of https://github.com/electric-sql/electric/pull/1054, Kevin and I encountered some TypeScript typing issues with `type_info`. Long story short, the latest iteration of this changes now lives in https://github.com/electric-sql/electric/pull/1065 where `sqlite_type` is deprecated (can be removed today, just needs some client-side tests for `/api/migrations` updated) and only `pg_type` remain as the source of truth for the column type.

PR #920, opened 2024-02-05
chore (docs): auth state management doc changes

kevin-dp:

PR #958, opened 2024-02-15
chore(docs): Update Expo driver docs to include expo-sqlite/next

msfstef:
Follow up to https://github.com/electric-sql/electric/pull/867.

### DIFF
--- a/docs/integrations/drivers/mobile/expo.md
+++ b/docs/integrations/drivers/mobile/expo.md
@@ -29,15 +29,13 @@ import { electrify } from 'electric-sql/expo'
 // Import your generated database schema.
 import { schema } from './generated/client'
 
-// Define custom configuration if needed
-const config = {
-  url: 'https://example.com:5133'
-}
-
 // Create the expo-sqlite database connection. The first argument
 // is your database name. Changing this will create/use a new
 // local database file.
 const conn = SQLite.openDatabase('electric.db')
+
+// Define custom configuration if needed
+const config = { url: 'https://example.com:5133' }
 
 // Instantiate your electric client.
 const electric = await electrify(conn, schema, config)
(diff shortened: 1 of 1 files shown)


docs/integrations/drivers/mobile/expo.md
```diff
@@ -35,6 +35,8 @@ const config = {
 // Create the expo-sqlite database connection. The first argument
 // is your database name. Changing this will create/use a new
 // local database file.
+//
+// When using expo-sqlite/next, use SQLite.openDatabaseSync('electric.db')
```
alco, line 39: This line does not depend on the `import { schema }` or `const config` lines above. I think it would be more readable to have two code blocks, one for expo-sqlite and one for export-sqlite/next, like so:

expo-sqlite:
```
import * as SQLite from 'expo-sqlite'
import { electrify } from 'electric-sql/expo'

const conn = SQLite.openDatabase('electric.db')
```

expo-sqlite/next:
```
import * as SQLite from 'expo-sqlite/next'
import { electrify } from 'electric-sql/expo-next' 

const conn = SQLite.openDatabaseSync('electric.db')
```

and make it switchable with a tab bar similar to how it's done in the [Quickstart guide](https://electric-sql.com/docs/quickstart#setup).

msfstef: But we still want the electrification step in the usage example right?

I'm thinking:

expo-sqlite
```ts
import * as SQLite from 'expo-sqlite'
import { electrify } from 'electric-sql/expo'

// Import your generated database schema.
import { schema } from './generated/client'

// Create the expo-sqlite database connection. The first argument
// is your database name. Changing this will create/use a new
// local database file.
const conn = SQLite.openDatabase('electric.db')

// Define your config with at least an auth token.
// See Usage -> Authentication for more details.
const config = {
  auth: {
    token: '...'
  }
}

// Instantiate your electric client.
```

expo-sqlite/next
```ts
import * as SQLite from 'expo-sqlite/next'
import { electrify } from 'electric-sql/expo-next' 

// Import your generated database schema.
import { schema } from './generated/client'

// Create the expo-sqlite database connection. The first argument
// is your database name. Changing this will create/use a new
// local database file.
const conn = SQLite.openDatabaseSync('electric.db')

// Define your config with at least an auth token.
// See Usage -> Authentication for more details.
const config = {
  auth: {
    token: '...'
  }
}

// Instantiate your electric client.
```

and just have a toggle between the two? there's redundant information but ultimately you pick the one you want to implement and it will have everything you need, I don't think it's much of a maintenance burden

alco: Sounds good. I thought we would leave the electrification step as is and just put the expo driver initialization in a tab bar above it. But what you're suggesting is better.

There's a conflicting change to the config block in Kevin's PR - https://github.com/electric-sql/electric/pull/920/files#diff-081d6fa57073cf0f59a708ea66afaa438865d72ae3d82a835a1448334f684c17. You can rebase on top of his branch now and avoid resolving conflicts later.

msfstef:
@alco looks good! yeah you're right it was unnecessary duplication - I've made a minor change to await the `findMany` query call.

Made another PR for fixing the other driver docs doing the same: https://github.com/electric-sql/electric/pull/975

PR #963, opened 2024-02-18
feate(starter): Add `--template` option to `create-electric-app` CLI tool

msfstef:
Added option to select a template when using `create-electric-app` - currently the options are React and Vue and more can be added via the `template-config.js` configuration file and adding a valid template in `src/templates.ts`.


icehaunter (approved):

balegas:
DOn't forget to add PR for docs (disregard my message if you're already into this)

msfstef:
> DOn't forget to add PR for docs (disregard my message if you're already into this)

Yep no PRs will be merged without a doc PR ready, but before writing docs I'd rather agree on the APIs introduced first.

For this particular case though, as far as I can tell we don't actually have a reference/API doc for the `create-electric-app` utility other than in the [quickstart section](https://electric-sql.com/docs/quickstart#setup) - I can update that one to reference the `--template` option, but do we maybe want some dedicated documentation for this tool?

PR #967, opened 2024-02-19
chore(docs): Update `create-electric-app` CLI docs to include `--template` parameter

msfstef:
Follow-up to https://github.com/electric-sql/electric/pull/963

Adds reference to `--template` CLI parameter along with the currently available options.


alco (approved):

samwillis (approved):
👍

PR #968, opened 2024-02-19
chore(docs): Add documentation for the Vue bindings

msfstef:
Follow-up to https://github.com/electric-sql/electric/pull/961

Documentation follows similar pattern to the React bindings documentation. Subject to change as original PR will determine the defined API.


samwillis (approved):
👍 great job

PR #978, opened 2024-02-21
chore(starter): Update starter templates to use electric `0.9.3`

msfstef:
Updates the example apps for react and vue to use the new auth APIs


samwillis (approved):
Brilliant!

One thing we may want to do, should we show this pattern I used in Linearlite to stop HMR breaking:

```
let userId = window.sessionStorage.getItem('userId')
  if (!userId) {
    userId = genUUID()
    window.sessionStorage.setItem('userId', userId)
  }
  const authToken = insecureAuthToken({ sub: userId })
```

PR #979, opened 2024-02-21
chore(docs): Remove backwards compatibility claim for `user_id` in token claims

msfstef:
Breaking change to the Auth APIs also removed the backwards compatibility for the token to have a `user_id` as claim instead of `sub.

https://github.com/electric-sql/electric/pull/828/files#diff-6f5e692c874649d2aee74958b2c1369a83d09ec9be84e60330f8eaf5ec4fc023R37-R43

```ts
export function decodeToken(token: string): JWTPayload & { sub: string } {
  const decoded = decodeJwt(token)
  if (typeof decoded.sub === 'undefined') {
    throw new InvalidArgumentError('Token does not contain a sub claim')
  }
  return decoded as JWTPayload & { sub: string }
}
```

We also say "`sub` (formerly `user_id`)" everywhere in the docs - should we still keep that or remove it entirely and use just `sub` going forwards?

### DIFF
--- a/docs/usage/auth/token.md
+++ b/docs/usage/auth/token.md
@@ -11,6 +11,10 @@ Authentication uses JSON Web Tokens, a [standard method](https://tools.ietf.org/
 
 A valid authentication JWT must contain a `sub` claim (formerly `user_id`). This must be a non-empty string, which should correspond to the primary key UUID of the authenticated app user.
 
+:::note
+Since version 0.8.0, the `sub` claim replaces the `user_id` as the standard "subject" designator in JWT tokens. However, Electric still validates tokens that have the `user_id` claim for backwards compatibility with old clients.
+:::
+
 ## Validated claims
 
 If you are using [insecure mode](./insecure.md) any other claims are optional and unvalidated.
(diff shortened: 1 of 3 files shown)


samwillis:
I think the situation now is that the sync service is backwards compatible, but the client isn't.

We should probably drop reference everywhere IMO (or push a patch making the client backwards compatible)

alco:
Dropping support for user_id was not necessary. And it doesn't look like it was intentional, more like an oversight. @kevin-dp hasn't mentioned it as a breaking change in the lead up to the release.

We should bring support for user_Id back in the client.

docs/usage/auth/token.md
```diff
@@ -11,10 +11,6 @@ Authentication uses JSON Web Tokens, a [standard method](https://tools.ietf.org/
 
 A valid authentication JWT must contain a `sub` claim (formerly `user_id`). This must be a non-empty string, which should correspond to the primary key UUID of the authenticated app user.
 
-:::note
```
samwillis, line 14: With #986 let's still do this PR (changed docs to `sub`) but keep this note in the docs.

samwillis (approved):
Approved but with keeping the backwards compatibility note.

docs/usage/auth/token.md
msfstef: I'll revert the removal of this notice, but should I remove the "formerly `user_id`" references?

samwillis: Nah, leave it for now.

Issue #982, opened 2024-02-21
error following quickstart


barbalex:
I tried to install a new quickstart example following https://electric-sql.com/docs/quickstart:

```console
alexandergabriel@MacBook-Air ~ % npx create-electric-app@latest test-app --template react
Port 5133 for Electric is already in use.
prompt: Hit Enter to keep it or enter a different port number:  (5133) 5134
Port 65432 for Electric's proxy is already in use.
prompt: Hit Enter to keep it or enter a different port number:  (65432) 65433
⠋ Creating project structurenode:internal/process/esm_loader:34
      internalBinding('errors').triggerUncaughtException(
                                ^

[Error: ENOENT: no such file or directory, lstat '/Users/alexandergabriel/.npm/_npx/bb5e66687992cffd/node_modules/create-electric-app/template-react'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'lstat',
  path: '/Users/alexandergabriel/.npm/_npx/bb5e66687992cffd/node_modules/create-electric-app/template-react'
}

Node.js v20.11.1
```

I am working an MacOS 14.3.1 (23D60) using node.js v20.11.1

msfstef:
There was an issue with the last published version where the actual templates were not being included - will be releasing a fix ASAP (see https://github.com/electric-sql/electric/pull/983)

PR #983, opened 2024-02-22
fix(starter): Include template folders in `create-electric-app` NPM package

msfstef:
The current version is broken (see https://github.com/electric-sql/electric/issues/982) because the template folders are not being included, this is a hotfix to make it work again.

Will see to adding some tests for this so it doesn't slip through.


examples/starter/package-lock.json
alco, line 1: @msfstef Was this file's inclusion intentional? It hadn't been a part of the starter prior to this PR.

PR #985, opened 2024-02-22
fix(client): Delay incoming messages when subscribing.

davidmartos96:
Related to: https://github.com/electric-sql/electric/issues/426

When porting the latest changes from v0.9.3 to the Dart client we started experiencing a race condition when a new client subscribes to a shape. It looks like the RPC SatSubsResp message comes after SatSubsDataEnd, causing the error "Received subscribe response for unknown subscription (some_id)"

With the PR change this doesn't happen anymore and it's how it's done in startReplication to avoid race conditions

Without delayMessages:
image

With delayMessages:
image

cc @icehaunter 

Would this be the correct approach? Dart and JS have slight differences in the event loop, so sometimes a race condition is easier to catch in some language or in the other

### DIFF
--- a/clients/typescript/src/satellite/client.ts
+++ b/clients/typescript/src/satellite/client.ts
@@ -468,9 +468,13 @@ export class SatelliteClient implements Client {
 
     this.subscriptionsDataCache.subscriptionRequest(request)
 
-    return this.service
-      .subscribe(request)
-      .then(this.handleSubscription.bind(this))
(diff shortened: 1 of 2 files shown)


davidmartos96:
Why is OTP_VERSION variable not resolved in the E2E CI task? 🤔

icehaunter (approved):
Thanks for the PR on this, and sorry for the slow review! Nice job catching it

PR #993, opened 2024-02-22
Manage global and per-user permissions state

magnetised:
- adds immutable global and per-user permissions state
- permissions state changes through ddlx commands and changes to assigns tables
- propagation of permissions changes to client connection (with a no-op on receipt)

additionally:

- replaces pg-based assign-role triggers with pure elixir version
- declares `SchemaLoader` as an implementation of itself and fixes resulting type issue with order of `SchemaLoader.connect/2`
- adds support for `ROW.column` style expressions in where/check clauses in ddlx (and fixes problem with parsing integers)


alco (approved):
Looks like a categorical improvement over the status quo to me!

PR #994, opened 2024-02-22
chore(example): Update Expo demo to 0.9.3

msfstef:
Updated the electric dependency on the expo demo - it requires some polyfills to work (see under `polyfill.ts`) but it does the trick.

We can worry about making the client work without developer-provided polyfills later.


samwillis (approved):
👍

PR #995, opened 2024-02-22
feat(e2e): Improved Lux matchers

davidmartos96:
Please correct me if I'm wrong, as I don't know much about lux. For the Dart port we use the same e2e tests to verify the behavior.
After some of the latest changes in Electric we noticed the 03.12 test started failing. 

After some digging we found that the ?$node matchers after the electrify macro in 03.12 migth be interfering with the new statement that is being sent to the Node CLI (!if (typeof exp === 'undefined') { var exp = undefined }), which wasn't being accounted for.

After this changes, everything started working again. If I'm understanding correctly, `?$node` is not necessary because it's `?\[rpc\] recv: #SatInStartReplicationResp` the one that ensures that the command finished, like in 03.10.


magnetised (approved):
Looks like a great change. Thanks for taking the time to figure this out.

PR #996, opened 2024-02-22
chore(examples): Update React Native demo

msfstef:
Re-initialized the react native demo completely as it was not running and felt beyond repair.

I've verified iOS is running and working.

I've included the android folder although the react-native-sqlite-storage package seems to be having trouble with it still.


samwillis (approved):
👍

PR #997, opened 2024-02-22
chore(client): Fix implementation of `react-native-sqlite-driver`

msfstef:
It was using wrong typings so the `dbName` was actually `dbname` and causing issues.

I've also refactored it to use the serial database adapter to simplify it.


samwillis (approved):
👍

Issue #990, opened 2024-02-22
v0.9.3: Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'


barbalex:
I did the following in [this project](https://github.com/barbalex/ps/tree/electric-sql-0-9-3):

- updated from 0.9.2 to 0.9.3
- re-instated docker
- copied Electric Provider from here: https://electric-sql.com/docs/integrations/frontend/react#electricprovider
- corrected `const isMounted = true` to be mutable: `let isMounted = true`
- cleared site data / application cache
- `npm run dev`

the result is:
<img width="717" alt="Bildschirmfoto 2024-02-22 um 12 25 25" src="https://github.com/electric-sql/electric/assets/669783/08df5216-4401-416d-a3a4-541e3f199970">

Maybe I'm doing something wrong?

barbalex:
The error seems to point to this line:

```js
const conn = await ElectricDatabase.init('electric.db', '')
```

msfstef:
The second argument `locateSqliteDist` is now optional, and the bundler used can find the sqlite distribution without it - you can look at the [sample app](https://github.com/electric-sql/electric/tree/main/examples/web-wa-sqlite) for a working example.

You're right that the docs are outdated, I've opened a PR to fix those https://github.com/electric-sql/electric/pull/992.

barbalex:
Thanks.

Somehow not this provider is returning undefined for db:
```ts
import { makeElectricContext } from 'electric-sql/react'

import { Electric } from './generated/client'

export const { ElectricProvider, useElectric } = makeElectricContext<Electric>()
```

msfstef:
You mean `useElectric()` returns `undefined` even when used within a child of the provider after an Electric client has been instantiated and provided? Could you provide a full repro example? (also, are you getting this issue when running the sample app mentioned above?)

barbalex:
> You mean useElectric() returns undefined even when used within a child of the provider after an Electric client has been instantiated and provided?

yes

> Could you provide a full repro example?

Do you mean a minimized fork of my app?

> are you getting this issue when running the sample app mentioned above?

I am getting it when running https://github.com/barbalex/ps/tree/electric-sql-0-9-3

msfstef:
> I am getting it when running https://github.com/barbalex/ps/tree/electric-sql-0-9-3

Running your repo I can't seem to reproduce what you're mentioning - have you managed to resolve this?

[...]

PR #999, opened 2024-02-22
chore(docs): Make `isMounted` a mutable variable in React docs

msfstef:
Was mentioned in this issue https://github.com/electric-sql/electric/issues/990


samwillis (approved):
👍

PR #1002, opened 2024-02-23
fix(e2e): Use connect/disconnect instead of `connectivityStateChanged` in e2e test.

davidmartos96:
After the recent changes to the client connectiviy, we can use connect and disconnect. The status `available` doesn't exist anymore.


msfstef (approved):
Looks good, thank you!

PR #1003, opened 2024-02-23
chore(ci): Drop the use of GitHub Actions variables

alco:
Their usage makes it impossible to run actions for pull requests from forks.


balegas (approved):
If waiting for an approval, I'll give it :D

PR #1005, opened 2024-02-26
fix(e2e): Improve E2E Satellite client Dockerfile build performance

davidmartos96:
One thing we noted when building the satellite_client Dockerfile to run the E2E tests locally is that any slight change in the Electric client source code was invalidating most of the install and build work in Docker.
After some investigation we found out that latest PNPM version is much more performant in the docker build.

I also took the chance to add some comments to the Dockerfile to better understand the multiple steps.

Here is a comparison between the main branch and this PR. As you can see, the times are much faster, so when debugging E2E tests for the client the DX improves.

### Main branch

* Full docker build (--no-cache): 3m 24s
* After some edit in the Electric client: 2m 56s

### This PR

* Full docker build (--no-cache): 1 min 3 s
* After some edit in the Electric client: 30s

### DIFF
--- a/e2e/satellite_client/Dockerfile
+++ b/e2e/satellite_client/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:18-alpine AS workspace
 
 RUN apk update && apk add git
-RUN corepack enable && corepack prepare pnpm@8.6.0 --activate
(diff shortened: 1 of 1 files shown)


alco (approved):
Thanks!

PR #1009, opened 2024-02-26
fix(client): Minimize Web API dependencies and provide polyfills

msfstef:
In the last update we introduced the `jose` dependency for JWT token singing and decoding, which depends on the `crypto` API which is not available in some environments (e.g. react-native).

I've tried to remove as much web API dependency or at least polyfill them to make integration easier:

- Asynchronously import `jose` dependency for new `secureAuthToken` method which will still fail as it requires `crypto` but we have not documented the method yet so we can sort it out separately @kevin-dp 
- Implement custom `jwtDecode` method based on the [npm package of the same name](https://www.npmjs.com/package/jwt-decode) - reason not to add another dependency is because it polyfills base64 with a different base64 dependency than the one we already use, and it's fairly thin and easy to maintain.
- Use our custom `base64` utility for encoding and decoding everywhere, to avoid using  web APIs `atob` and `btoa` - we were already using a non-web API implementation so this is not an extra burden.
- Provide TextEncoder polyfill globally (i.e. always try to polyfill it) with [`fastestsmallesttextencoderdecoder`](https://www.npmjs.com/package/fastestsmallesttextencoderdecoder) rather than adding it on the driver entry points - there's separate code (like in the /auth utils) that uses text encoding and import order can cause issues - we already had this as a dependency and it's fairly light, I think it's safe to include.
- Remove `uuid` dependency and `setUUIDImpl` method and use the `genUUID` utility everywhere - it falls back to using `Math.random` for rng which does not give very strong uniqueness guarantees so I added a `console.warn` to provide a polyfill.
- [BREAKING] Add [`react-native-get-random-values`](https://www.npmjs.com/package/react-native-get-random-values) peer dependency for expo and react-native drivers so that `genUUID` works properly (see above point).

With these changes, react-native developers only have to include `react-native-get-random-values` as an extra dependency (and we could make that optional as well if we are ok with `Math.random` as our UUID source). We can also add it as a proper dependency and only include it in our react-native driver entry points to reduce friction even more.

### DIFF
--- a/.changeset/olive-kangaroos-chew.md
+++ b/.changeset/olive-kangaroos-chew.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Provide targeted polyfills for web API dependencies for compatibility with other runtimes


clients/typescript/src/drivers/expo-sqlite-next/index.ts
```diff
@@ -3,6 +3,9 @@
 // alternative entrypoint in `./test` to avoid importing this.
 import { DbName } from '../../util/types'
 
+// Provide implementation for crypto.getRandomValues
+import 'react-native-get-random-values'
```
samwillis, line 7: I wander if there is a way to make this optional, that way the user sees you new warning? (almost certainly not)

clients/typescript/src/util/common.ts
```diff
@@ -1,12 +1,10 @@
+// Provide implementation for base64 encoding/decoding
```
samwillis, line 1: This comment confused me, "Provide implementation" is used else where to indicate a polyfill. This is a normal import. I would drop the comment.

samwillis (approved):
Really great refactor! 👍 

As I mentioned earlier, I kind of dislike polyfils in libraries as they can result in a version of a polyfil different from what a developer of an app has decided to use.

For `fastestsmallesttextencoderdecoder` is there a way we can import the encoder from it, but not patch global? then use that, maybe from a `utils` re-export throughout our code?

If not, not a problem, and it may be outside this PR anyway. (I also my be wrong on this)

clients/typescript/src/util/common.ts
msfstef: I see your point - although technically the `base-64` package is unnecessary on the web but we're also always using it and not polyfilling it so I'll change this.

msfstef:
> For `fastestsmallesttextencoderdecoder` is there a way we can import the encoder from it, but not patch global? then use that, maybe from a `utils` re-export throughout our code?

You're right we should avoid global polyfills where possible - I've refactored it to only use the package if there isn't already a text encoder in the `globalThis`, but at least this is not done globally and instead only in our `common` utilities.

clients/typescript/src/auth/decode.ts
```diff
@@ -0,0 +1,54 @@
...
+    decoded = base64.decode(part)
+  } catch (e) {
+    throw new InvalidTokenError(
+      `Invalid token specified: invalid base64 for part #${pos + 1} (${
+        (e as Error).message
+      })`
+    )
+  }
+
+  try {
```
kevin-dp, line 45: Can we unify these two try-catch blocks into one?
We could either differentiate based on the error that is thrown in order to re-throw either the invalid base64 error or the invalid json error. If we can't differentiate the errors, we can use a flag:

```ts
let decodeSuccess = false
try {
  const decoded = base64.decode(part)
  decodeSuccess = true
  return JSON.parse(decoded) as JwtPayload
} catch (e) {
  const problem = decodeSuccess ? 'invalid json' : 'invalid base64'
  throw new InvalidTokenError(
    `Invalid token specified: ${problem} for part #${pos + 1} (${
      (e as Error).message
    })`
  )
}
```

clients/typescript/src/auth/secure/index.ts
```diff
@@ -1,20 +1,21 @@
...
 
-export function secureAuthToken(opts: {
+export async function secureAuthToken(opts: {
   claims: TokenClaims
   iss: string
   key: string
   alg?: string
   exp?: string
 }): Promise<string> {
+  const { SignJWT } = await import('jose')
```
kevin-dp, line 13: Did i understand correctly from your PR description that previously React-Native apps would throw an error because `jose` was imported at the top-level and is not available in RN, hence, we turn it into a dynamic import inside `secureAuthToken` such that it only throws in RN if this function is called?

It's a bit annoying that it makes this function async but on the other hand it is probably no big deal.
Maybe we can avoid it by doing a top-level dynamic import?
Something like:
```ts
const { SignJWT } = (await import('jose')).catch(_e => undefined)
```

Similarly, RN will only throw an error if `secureAuthToken` is called because `SignJWT` will be undefined and thos it can't be called on line 20. I know that top-level await is not supported everywhere so maybe this is a bad idea.

clients/typescript/src/auth/secure/index.ts
```diff
@@ -50,5 +51,7 @@ export function decodeToken(token: string): JWTPayload & { sub: string } {
       'Token does not contain a sub or user_id claim'
     )
   }
-  return decoded as JWTPayload & { sub: string }
+
+  decoded.sub ??= decoded.user_id
```
kevin-dp, line 55: nit: is there a reason we mutate the sub claim instead of `return (decoded.sub ?? decoded.user_id) as ...`
It's no big deal since the decoded token is only used here.

clients/typescript/src/satellite/process.ts
```diff
@@ -1473,7 +1472,7 @@ export class SatelliteProcess implements Satellite {
     let clientId = await this._getMeta(clientIdKey)
 
     if (clientId === '') {
-      clientId = uuid() as Uuid
+      clientId = genUUID() as Uuid
```
kevin-dp, line 1475: It seems a bit weird that we have to cast the result of a function that produces a uuid to the `Uuid` type.
Shouldn't we modify the return type of that function to be `Uuid` such that we don't have to cast it here?

clients/typescript/src/util/common.ts
```diff
@@ -28,6 +23,19 @@ export const base64 = {
     ),
   toBytes: (string: string) =>
     Uint8Array.from(BASE64.decode(string), (c) => c.charCodeAt(0)),
+  encode: (string: string) => base64.fromBytes(textEncoder.encode(string)),
+  decode: (string: string) => textEncoder.decode(base64.toBytes(string)),
+}
+
+export const textEncoder = {
+  encode: (string: string): Uint8Array =>
+    'TextEncoder' in globalThis && !!globalThis.TextEncoder
```
kevin-dp, line 32: `'TextEncoder' in globalThis` and `!!globalThis.TextEncoder` seem redundant, would `!!globalThis.TextEncoder` be enough?

clients/typescript/src/util/random.ts
```diff
@@ -15,6 +15,12 @@ export const genUUID = (): string => {
     globalThis.crypto.getRandomValues(bytes)
   } else {
     // fallback to Math.random, if the Crypto API is completely missing
+    console.warn(
```
kevin-dp, line 18: Not really related to this PR but since we are modifying this file, perhaps we could simplify line 13 to:
```ts
if (!!globalThis.crypto?.getRandomValues) {
```

clients/typescript/test/auth/secure.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    iss: 'test-issuer',
+    key: 'test-key',
+  })
+
+  t.is(typeof token, 'string')
+  t.regex(token, /^eyJh[\w-.]+$/)
+  const decodedToken = jwtDecode(token) as JwtPayload & { type: string }
+  t.is(decodedToken.sub, 'test-user')
+  t.is(decodedToken.iss, 'test-issuer')
+  t.is(decodedToken.type, 'access')
```
kevin-dp, line 18: Where does this `type` claim come from? `auth/decode` doesn't seem to add it and i'm not familiar with this JWT claim.

clients/typescript/test/drivers/expo-next.test.ts
```diff
@@ -1,7 +1,7 @@
 import test from 'ava'
 
 import { MockDatabase } from '../../src/drivers/expo-sqlite-next/mock'
-import { DatabaseAdapter } from '../../src/drivers/expo-sqlite-next'
+import { DatabaseAdapter } from '../../src/drivers/expo-sqlite-next/adapter'
```
kevin-dp, line 4: nit: don't think this change is needed, the index file of `expo-sqlite-next` exports the adapter

clients/typescript/test/drivers/expo.test.ts
```diff
@@ -1,7 +1,7 @@
 import test from 'ava'
 
 import { MockDatabase } from '../../src/drivers/expo-sqlite/mock'
-import { DatabaseAdapter } from '../../src/drivers/expo-sqlite'
+import { DatabaseAdapter } from '../../src/drivers/expo-sqlite/adapter'
```
kevin-dp, line 4: same as the other comment

kevin-dp (approved):
Great work! Left a couple of minor comments and questions. Feel free to merge when you deem it ready

clients/typescript/test/drivers/expo-next.test.ts
msfstef: Importing the top-level index file which includes `react-native-get-random-values` causes issues as it is a peer dependency - unless we include it as a regular or dev dependency (see discussion above with Sam)

clients/typescript/src/util/random.ts
msfstef: esling doesn't like double negations on this so I'll turn it into `if (globalThis.crypto?.getRandomValues) ...` for this and the others like it

kevin-dp: Well it's outside the scope of this PR anyway, so if it is scattered around the place feel free to ignore this comment

clients/typescript/src/auth/secure/index.ts
msfstef: This utility is supposed to return the whole decoded token, so we could avoid mutation by recreating it like
```js
...
return { ...token, sub: decoded.sub ?? decoded.user_id) }
```
Alternatively I can create an additional utility to extract the user ID from the token alone - which I think I might do cause I see we have the "user_id" and "sub" backwards compat code elsewhere as well

clients/typescript/src/auth/secure/index.ts
msfstef: The `sign` API from `jose` is async either way so the function signature is the same - I'm reluctant to do top-level dynamic imports cause it introduces weird race conditions on whether something is defined or not that are hard to reason about and test

PR #1011, opened 2024-02-27
feat(client): Add `useConnectivityState` to Vue bindings

msfstef:
Adds the `useConnectivityState` method to the Vue bindings.

I've abstracted the subscription to the notifier connectivity changes like we had for the live queries (you basically curry the subscription with a handler).

I've also created a `/shared` library for the frameworks, for logic that is not strictly related to the client but can be common across the framework bindings.

(docs PR: https://github.com/electric-sql/electric/pull/1012)


clients/typescript/src/frameworks/vuejs/reactive/useLiveQuery.ts
```diff
@@ -135,7 +135,10 @@ function useLiveQueryWithQueryUpdates<Res>(
     { immediate: true }
   )
 
-  onUnmounted(() => unsubscribeRef.value?.())
+  onUnmounted(() => {
+    unwatch()
```
samwillis, line 139: Why swap to an explicit `unwatch`? I believe the watch should automatically be unwatched when the component unmounts.

Your change here also has prompted a thought. Our `useLiveQuery` will only work inside a components `setup` method or `<script setup>`  as we are using `onUnmounted`. It would be interesting to explore, outside this PR, the possibility of making this `useLiveQuery` work outside components, just in the Vue reactive system. They would also then translate to Alpine.js that also uses the Vue reactive system.

samwillis (approved):
brilliant 👍

clients/typescript/src/frameworks/vuejs/reactive/useLiveQuery.ts
msfstef: You're right that from the moment we're using `onUnmounted` then unwatching is unnecessary.

The reason I added that was the same thought you expressed - I would like for the reactivity to work and also correctly clean itself up outside of a `setup` method.

We could make the live query return an `unwatch` or `unsubscribe` method as well (like the watchers do) but we'd still want to keep the default `onUnmounted` behaviour - I'll remove this change for this PR and we can think about it in a separate one with appropriate tests.

PR #1013, opened 2024-02-27
chore(docs): Replace `yarn` with `npm` in the Expo demo README

msfstef:
We've started using npm for all our examples and including the package-lock.json as well, so doing this for consistency.


samwillis (approved):
👍

Issue #1001, opened 2024-02-23
Generated client in 0.9.x can still contain "string is not assignable to type undefined" errors


fooware:
There have been problems with incompatible Zod and Prisma versions in previous Electric-SQL versions, requiring the developer to use specific versions of Prisma and Zod, which is somewhat complicated, see PR #700.

That PR got rejected due to #727 adding a workaround using casts in the generated client, which fixed our initial problems.

However, as late as Electric 0.9.3 we still have this problem with our generated client with a more "advanced" DB structure, and due to internal changes in the way Electric generates the client, we can no longer override the Zod version using PNPM in a good way, so we are stuck at version 0.8.1 for now.

## Test case
The main branch of the test repo below contains a mostly stock `create-electric-app` using ElectricSQL 0.9.3 and it can not generate a correct client.

### Error
![image](https://github.com/electric-sql/electric/assets/984279/23d6dc00-372d-4b4c-b793-531275675787)
### Test Repo
https://github.com/fooware/electric-testcase-01
### Commit with the DB-structure
https://github.com/fooware/electric-testcase-01/commit/10e5ff75b4fc92852ef04cbff69c167024aac7d1

kevin-dp:
Hi @fooware, i found that these type errors got introduced by our upgrade to Prisma v5 for generating the Prisma client as v5 seems to introduce some breaking changes in the generated types. I could reproduce your type errors on a fresh `create-electric-app` with your schema. Also tested it with a modified Electric client that uses Prisma v4 and the type errors were gone. That confirms the problem i described. We will need to revert to using Prisma v4. Will keep you posted here when that is done. Thanks for reporting this issue.

fooware:
Sounds great @kevin-dp, looking forward to testing the new release when it is out.

atanaskanchev:
Could this be related to this long lived Zod bug? https://github.com/colinhacks/zod/issues/2184

msfstef:
@fooware this issue should be resolved now in 0.9.6 - can you confirm?
If other problems related to the generator come up it's better if they are added as separate issues.

jljorgenson18:
@msfstef I am still seeing errors in 0.9.6 and had to add `@ts-expect-error` comments throughout the generated code

msfstef:
@jljorgenson18 is it still the `string is not assignable to type undefined`  error or something else like `'Relation' is declared but its value is never read.`?

jljorgenson18:
I am receiving two types of errors. 

[...]

PR #1014, opened 2024-02-27
fix (generator): fix type error with `WhereUniqueInput` schemas

kevin-dp:
This PR fixes type errors in the generated client with `WhereUniqueInput` schemas, as reported in: https://discord.com/channels/933657521581858818/1211387122351870072.
The type errors come from the fact that we are now using Prisma v5 for generating the Prisma client but in Prisma v5 the type for `WhereUniqueInput` schemas is now wrapped in an `AtLeast<..., ...>` type which is a breaking type change compared to the generated type in v4. Since our Zod schemas are generated using v4 the types are incompatible. This PR patches the `AtLeast` type such that the resulting type is equivalent to the type in v4.

### DIFF
--- a/.changeset/yellow-knives-itch.md
+++ b/.changeset/yellow-knives-itch.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Fixes type errors with WhereUniqueInputSchema in generated client.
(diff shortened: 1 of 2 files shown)


samwillis (approved):
Love a bit of regex 👍

kevin-dp:
I would hold off merging this for the moment because this does not seem to be the only problem introduced by our upgrade to Prisma v5. There is another open issue which is related to that upgrade: https://github.com/electric-sql/electric/issues/1001. I tested that issue and it works when our generator uses Prisma v4 but not when we use Prisma v5. Again, due to a breaking type change in v5. Perhaps we should revert back to using Prisma v4 also for generating the Prisma client, even though we know that leads to the generator crashing if the user has Prisma v5 installed in their project.

kevin-dp:
Created a PR for reverting to Prisma v4: https://github.com/electric-sql/electric/pull/1016

PR #1016, opened 2024-02-27
fix (generator): revert to Prisma v4 for generating Prisma client

kevin-dp:
Revert to Prisma v4 for generating Prisma client because v5 introduced some breaking type changes.
Also pinned version of Zod to latest one before a bug got introduced.


samwillis (approved):
👍

PR #1019, opened 2024-02-29
chore(client): Clean up React bindings

msfstef:
- Removed unused and undocumented `useRandom` hook which was originally already removed in this PR https://github.com/electric-sql/electric/pull/585 but was added back with another refactor
- Removed dependency on the injected electric client in `useLiveQuery`


samwillis (approved):
👍

PR #1024, opened 2024-03-04
chore(examples): Remove `beer-stars` example

msfstef:
The beer-stars example is significantly outdated, and after some discussions my understanding is that we want to remove it rather than update it.


samwillis (approved):
👍 
@thruflo should confirm.

alco (approved):

PR #1025, opened 2024-03-04
fix(CLI): CLI support for Docker v2.24.6 by removing inheritance from docker-compose files

### LINKED ISSUE #1018: service "electric-with-postgres" can't be used with `extends` as it declare `depends_on`
```bash
PUPPETEER_SKIP_DOWNLOAD=true npx create-electric-app@latest ntcr --template react
cd ntcr && npx electric-sql start --with-postgres

Starting ElectricSQL sync service with PostgreSQL
Docker compose config: {
  SERVICE: 'http://localhost:5133',
...
  LOGICAL_PUBLISHER_HOST: 'electric'                                                                                                                          
}                                                                                                                                                             
service "electric-with-postgres" can't be used with `extends` as it declare `depends_on`                                                                      
Failed to start the Electric backend. Check the output from 'docker compose' above. If the error message mentions a port already being allocated or address be
ing already in use, please change the configuration to an alternative port via the ELECTRIC_HTTP_PORT or ELECTRIC_PG_PROXY_PORT environment variables.

```

Looks related to https://github.com/docker/compose/issues/11544

samwillis:
Fixes #1018

This removes the inheritance from the docker compose files. There are now two mostly duplicated compose files, with the "with-postgres" file have this addition for the electric service:

```
depends_on:
      - postgres
```

Not pretty, but works and is far less likely to break with future versions of Docker.


clients/typescript/src/cli/docker-commands/docker/compose-base.yaml
```diff
@@ -1,12 +0,0 @@
-version: '3.8'
-
-services:
-  electric-no-postgres:
-    init: true
-    stop_signal: SIGINT # use SIGINT as the more speedy alternative to SIGTERM
```
msfstef, line 6: should we replicate the `init` and `stop_signal` configurations on the new compose files?

msfstef (approved):
looks good to me!

clients/typescript/src/cli/docker-commands/docker/compose-base.yaml
samwillis: Good spot!

PR #1026, opened 2024-03-04
fix (client): avoid duplicate rows when including several relations

kevin-dp:
This PR fixes a bug in the DAL that caused rows to be included several times when including multiple relations.
The issue was reported on Discord: https://discord.com/channels/933657521581858818/1210387734234005614

### DIFF
--- a/clients/typescript/src/client/model/table.ts
+++ b/clients/typescript/src/client/model/table.ts
@@ -771,9 +771,6 @@ export class Table<
       return onResult(rows)
     else {
       const relationFields = Object.keys(include)
-      let includedRows: Kind<GetPayload, T>[] = []
-      // TODO: everywhere we use forEachCont we probably don't need continuation passing style!
-      //       so try to remove it there and then rename this one to `forEachCont`
       forEach(
         (relationField: string, cont: () => void) => {
           if (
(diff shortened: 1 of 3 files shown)


clients/typescript/src/client/model/table.ts
```diff
@@ -796,22 +793,20 @@ export class Table<
             relationName
           )
 
+          // `fetchInclude` mutates the `rows` to include the related objects
```
msfstef, line 796: Unless I'm misreading this, it seems that you have to follow all the way through to the `joinObjects` call to find the mutation (uses `Object.assign`), from a chaing of `fetchIncludes` -> `fetchInclude` -> `fetchRelated` -> `joinObjects`.

Is this a performance optimization? otherwise, wouldn't it be clearer and safer if `joinObjects` returned a new row rather than mutate the existing one and we reconstruct the joined rows in `fetchIncludes`?

msfstef (approved):
Nice catch, looks good! I've just left a question about why we use mutation in the first place

clients/typescript/src/client/model/table.ts
kevin-dp: Yes you're right the mutation happens in `joinObjects` at the end of the chain. With the way how it is currently implemented, i'm afraid that implementing this immutably would be require a fair bit of rewriting as all this information would need to be threaded through. Currently, we're essentially extending the object once we have all information about the relation. I tend to avoid mutations where possible but i don't think it's a safety problem here as every read from the database returns a a copy of the rows that were read. So those (mutated) objects are not reused between queries.

The problem with `joinObjects` returning a row instead of mutating is that:
- each call to `joinObjects` would only extend the row with 1 relation
- so we will end up with N "duplicate" rows (where N is the number of included relations) and each of those duplicates only containing their relation
  - for example: `user.findMany({ include: { profile: true, posts: true }})` would return each row twice once extended with the user's profile and once extended with the user's posts. By mutating the object, `joinObjects` ensures that we don't have duplicates and that the row is extended with each relation.
  
 Does that make sense? :-)

Issue #938, opened 2024-02-09
An upsert's update on json field is getting double-stringified


KyleAMathews:
```ts
              db.ingredient_events.upsert({
                create: {
                  id: genUUID(),
                  ingredient_id: ingredient.id,
                  timestamp: new Date(),
                  from_values: { fill_level: ingredient.fill_level },
                  to_values: { fill_level: newFillLevel },
                  user_id,
                },
                update: {
                  to_values: { fill_level: newFillLevel },
                },
                where: {
                  id: recentEvent?.id,
                },
              })
```

E.g. for this code, the to_values is ending up as a string in the db instead of JSON.

KyleAMathews:
Actually, it seems any update has this problem

KyleAMathews:
err... well maybe just the upsert? I'm doing a manual create or delete now and the JSON is fine.

balegas:
cc @kevin

msfstef:
Resolved by https://github.com/electric-sql/electric/pull/1029

PR #1029, opened 2024-03-04
fix(client): Fix `upsert` transforming inputs twice

msfstef:
Fixes https://github.com/electric-sql/electric/issues/938

The `upsert` operation validated _and transformed_ its inputs, but then subsequently called either `create` or `update` which re-transformed the already transformed inputs.

For most types its fine as the transform is idempotent, but for JSON we either need to check if the input is "deserializable" and assume it does not need to be re-serialized, or make sure we don't attempt to transform it more than once.

I might be wrong, but if we use the first approach there are cases where people might insert already serialized JSON values, leading to potential missed serialization (and then to an incorrect or failed deserialization).

Therefore I've made `upsert` only validate its inputs, and relegate the transformation to the `create` or `update` calls that follow.


kevin-dp (approved):
Great catch, thanks for fixing!

PR #1033, opened 2024-03-06
fix (client): fix generic serial database adapter to always release the lock

kevin-dp:
This PR fixes a bug with the serial database adapter that caused the lock not to be released. This could happen when `COMMIT` failed in the `finally` block (e.g. due to a FK constraint violation) which would cause an error to be thrown and execution would never reach the call to `release()`. We fixed this by moving the `COMMIT` inside the `try` block such that any exception thrown by commit is caught. If `COMMIT` fails we will `ROLLBACK` in the catch block and we will always release the lock in the `finally` block.


clients/typescript/src/drivers/generic/adapter.ts
```diff
@@ -131,26 +131,25 @@ export abstract class SerialDatabaseAdapter
...
       }
+      await this._run({ sql: 'COMMIT' })
       return {
         rowsAffected: rowsAffected,
       }
     } catch (error) {
-      await this._run({ sql: 'ROLLBACK' })
-      open = false
+      if (transactionBegan) {
+        await this._run({ sql: 'ROLLBACK' })
```
msfstef, line 149: Have you settled on the expected behaviour of a failed ROLLBACK? Might be worth documenting here with a link to the SQLite spec or a short comment the expectation

clients/typescript/src/drivers/generic/adapter.ts
msfstef, line 1: is it possible to add a unit test that ensures the lock is released in the case of a failed COMMIT?

msfstef (approved):
Nice catch, looks good!

clients/typescript/src/drivers/generic/adapter.ts
kevin-dp: If the rollback fails, i don't think we can fix it as it may require checking the error code and fixing it appropriately (there is a related discussion on [stackoverflow](https://stackoverflow.com/questions/59410490/sqlite-c-interface-what-happens-if-transaction-rollback-fails)). This is something the user would have to do so the user would need to catch potential problems and handle them. However, with our live queries and `runInTransaction` being used internally i don't think the user can always catch them and fix them. As described in the SO thread, for some problems we may retry. But for others, it could require a manual investigation?

It's an interesting problem but i don't think it should block this PR as this problem has always been in our codebase. Something to discuss internally and agree on a solution which we can implement as a separate PR?

kevin-dp: cc @balegas @samwillis @icehaunter

clients/typescript/test/drivers/generic-adapters.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      tx.run(insert)
+      throw new Error('mocked failure')
+    }),
+    {
+      message: 'mocked failure',
+    }
+  )
+
+  // @ts-ignore protected field but we're testing the internal mutex of the adapter
+  t.false(adapter.txMutex.isLocked())
```
msfstef, line 144: nit: might be worth adding a read-only getter on the adapter like `isLocked()` to hide the mutex implementation but still be able to check? feels like a useful API for the adapter to have, at least for debugging and asserting

msfstef (approved):
Looks good, nice catch and nice work with the tests!

clients/typescript/src/drivers/generic/adapter.ts
kevin-dp: Tracked this in Linear: VAX-1724

PR #1034, opened 2024-03-06
fix(client): Fix text encoding polyfill to work cross-platform

msfstef:
- Replaced the [fastestsmallesttextencoderdecoder](https://www.npmjs.com/package/fastestsmallesttextencoderdecoder) polyfill with [text-encoder-lite](https://www.npmjs.com/package/text-encoder-lite) which has modular exports and presents fewer issues with building the client for various runtimes and platforms.
- Replaced use of TextDecoder with custom implementation to ensure polyfill works.
- Removed unuser react-native-uid types dependency


clients/typescript/src/util/common.ts
```diff
@@ -1,7 +1,7 @@
-import BASE64 from 'base-64'
-import { default as TextEncodeDecodeAlternative } from 'fastestsmallesttextencoderdecoder'
-
 import { SatelliteError } from './types'
+import BASE64 from 'base-64'
+// @ts-ignore No type declarations available - check back to add
```
kevin-dp, line 3: This package seems very small, basically 2 methods `encode` and `decode`. Can we write our own type declarations for this package and keep it somewhere in our repo?

kevin-dp (approved):
Looks good to me!
Have tested it on a web app created with latest `create-electric-app` and i get no more errors 🎉

kevin-dp (approved):

PR #1035, opened 2024-03-07
fix (client): fix snapshot query that sets the cleartags

kevin-dp:
This PR fixes a problem with `performSnapshot` which caused a deleted row to re-appear in the following scenario:
TX1: `insert row`
TX2 (causally after TX1): `update row; delete row`

After TX2 the deleted row would re-appear with the value of the update.
This occurs because `performSnapshot` would only set the clearTags on the first operation in each transaction which caused the `delete` not to override the `update`. As discussed with @icehaunter, the solution is to modify `performSnapshot` such that it sets the clearTags of all operations in a TX (except the first) to the tag of that transaction.

This PR applies the above fix and i can confirm that the anomaly is gone. I have added a unit test that checks this specific case. **However, another unit test**, called "origin tx (INSERT) concurrently with local txses (INSERT -> DELETE)" in `process.tags.test.ts` **started failing**. I investigated a bit further but that unit test seems weird to me because on line 564 it says "For this key we receive transaction which was older" and it contains a tag that keeps the row alive (but our own, later tx already deleted that row, see L554). So i'm not sure if it makes sense that the server would ever send us this old transaction. I don't know if it was meant like this or it was meant to send the ack of our original tx in which case the tags should be empty because we deleted that row (L572, and also delete L587).

I have modified the tags in that failing unit test according to what i believe the server would send. @icehaunter if you could take a look at that test to see if the original is indeed wrong and whether my fix is correct.

### DIFF
--- a/.changeset/smooth-seals-design.md
+++ b/.changeset/smooth-seals-design.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Fixes bug that would cause deleted rows to re-appear under specific conditions.
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -973,12 +973,20 @@ export class SatelliteProcess implements Satellite {
       }
 
       // For each first oplog entry per element, set `clearTags` array to previous tags from the shadow table
+      // For the subsequent oplog entries per element, set the `clearTags` array to the current tag for this TX
+      // This is fine because only the first operation of the TX must override previous TXs
+      // and the rest of the operations in the TX just need to be aware of the current TX
+      // because conflict resolution on Electric happens on the level of TXs
       const q2: Statement = {
         sql: `
       UPDATE ${oplog}
-      SET clearTags = updates.tags
+      SET clearTags =
+            CASE WHEN rowid = updates.rowid_of_first_op_in_tx
+                 THEN updates.tags
+                 ELSE ? -- singleton array containing tag of thix TX
+            END
       FROM (
-        SELECT shadow.tags as tags, min(op.rowid) as op_rowid
+        SELECT shadow.tags as tags, min(op.rowid) as rowid_of_first_op_in_tx
         FROM ${shadow} AS shadow
         JOIN ${oplog} as op
           ON op.namespace = shadow.namespace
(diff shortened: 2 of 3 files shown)


balegas:
Two questions:
You mentioned that the issue would only happen if operation when through the server, I'd like to understand the specifics of how the server is handling tags that was causing arm (from our conversation I thought that the issue was just the way we process the local oplog when merging incoming changes).

> it contains a tag that keeps the row alive (but our own, later tx already deleted that row, see L554)
You're saying that the receiving client had already seen that tag and is receiving it again? that would be a violation of the delivery guarantees (and a know pathological case that revives rows incorrectly).

kevin-dp:
> Two questions: You mentioned that the issue would only happen if operation when through the server, I'd like to understand the specifics of how the server is handling tags that was causing arm (from our conversation I thought that the issue was just the way we process the local oplog when merging incoming changes).
> 
> > it contains a tag that keeps the row alive (but our own, later tx already deleted that row, see L554)
> > You're saying that the receiving client had already seen that tag and is receiving it again? that would be a violation of the delivery guarantees (and a know pathological case that revives rows incorrectly).

Consider the example with the 2 TXs from the PR description:

TX1: insert row
TX2 (causally after TX1): update row; delete row

On the client, when `performSnapshot` runs after TX1, it assigns a timestamp to the insert operation in the oplog. So let's say that the tag of TX1 (which is based on this timestamp) is `Tag1`.
Then, TX2 is executed locally and then we perform a snapshot. `performSnapshot` assigns timestamps to both the update and delete from TX2. It also assigns `Tag1` (the tag of TX1) to the `clearTags` of the `update` entry in the oplog. Note that the clearTags of the `delete` entry is the empty array (i.e. no tags). So the problem here is that the `delete` entry does not record the fact that it has seen the `update` entry from its own TX. Then, these oplog entries are converted into the protobuf format (not so important, just need to know that it includes the `clearTags`) and sent to Electric.

Then on Electric, these tags are used to apply CR. Since the `clearTags` for `delete` is missing the tag of the update (i.e. the tag of TX2), it does not get conflict resolved correctly and Electric thinks that the `update` should win over the `delete`. So Electric sends back TX2 to the client and it only contains the update because the delete lost from the update. So the client applies those operations? and the deleted row re-appears.

The explanation about the CR on Electric may not be accurate, i would ask @icehaunter to confirm or provide clarifications where needed.

icehaunter:
Yeah, @kevin-dp, you're spot-on with explanation of what's going on here.

icehaunter (approved):

balegas:
The explanation looks right to me too, what I'd like to recollect and would be informative for the conversation is how CR works on the server. If I remember correctly, the algorithm does the `shadow_tags` \ `clearTags` and doesn't persist `clearTags` for the row, given that they will always be downstreamed with the op that produced them and they are not more required for the server.

PR #1036, opened 2024-03-07
chore(client) Export websocket implementations to enable use in custom drivers

samwillis:
Fixes VAX-1723

Needed for LiveStore


msfstef (approved):
Looks good! Based on the issue description I assume the intention is to allow for people to build custom drivers outside of the ones Electric itself provides?

PR #1038, opened 2024-03-07
feat (client): support for PG on the client

kevin-dp:
This PR adds support for Postgres on the client:
- adds DB drivers for Postgres in NodeJS and in Tauri
- modifies the initial Electric schema (meta tables)
- modifies the queries executed by the `BundleMigrator`
- modifies the triggers to also support PG
- modifies the Satellite process to also support PG
- modifies the CLI to bundle both SQLite and PG migrations
- modifies the DAL to transform JS values to PG when writing and the other way around when reading

**To test** this PR you will need to:
- use a local build of Electric from `alco/vax-1659-postgresql-dialect`
- npm pack the generator
- use the packed generator in the ts-client
- npm pack the ts-client
- use the packed version of ts-client in the example app (and use node-postgres or tauri-postgres driver)

Limitations:
- Floating-point numbers that are out of range cannot be inserted (PG throws an error, in contrast to SQLite which silently converts them to +Inf/-Inf (for too big values) or 0 (for too small values))
- Not sure if PGlite version currently supports infinity values and/or `NaN` for floating-point numbers. Our triggers do some casting of these values, need to test this with some e2e tests or actual tauri application to see if it works or not.

A couple of things to note regarding the implementation:
- Had to change a trigger in both SQLite and PG for storing the JSON of a row's PK:
  - In PG, `jsonb` adds whitespace and re-orders keys. Whereas, SQLite has no whitespace and preserves the ordering of keys. So in PG we use `json` instead of `jsonb` and strip any whitespace but that has the side effect of also removing `null` values. Hence, we also remove `null` values in SQLite.
- Postgres requires migrations coming from Electric to define all FKs as `DEFERRABLE`
- We no longer defer foreign key checks before applying an incoming transaction as PG does not like deferred constraints when trying to apply some DDL. This happens because PG wants to run all trigger events before applying DDL but they are deferred so they are only allowed to run when committing so PG is stuck. So if a TX contains DML and DDL then when it encounters the DDL we get errors like this: `cannot ALTER TABLE "parent" because it has pending trigger events`

Remains to be done (perhaps in a follow-up PR):
- e2e tests for PG client

------------

For reviewers:

- The crux of this PR consists of modifying the TS-client to abstract away dialect-specific SQL queries behind a query builder abstraction:
  - `src/migrators/query-builder/builder.ts` defines an abstract `QueryBuilder` class that every dialect must implement
  - `src/migrators/query-builder/sqliteBuilder.ts` and ``src/migrators/query-builder/pgBuilder.ts` implement the abstract class for SQLite and Postgres respectively
- When electrifying a database, the dialect can (optionally) be provided in the config object. It defaults to `'SQLite'`. Postgres drivers' `electrify` function pass `'Postgres'` dialect through this config object. The DAL checks the dialect in order to apply the right conversions from JS values to SQLite/PG values and back.
- The main changes are in `src/migrators/bundle.ts`, `src/migrators/schema.ts`, and `src/migrators/triggers.ts` to use the dialect-specific query builder object.
- Likewise, the satellite process (`src/satellite/process.ts`) has been changed to use the dialect-specific query builder for constructing its specialised queries.
- Another important change consists of supporting data types such as booleans, dates, json, etc.
  - In SQLite such values were converted to strings and stored as strings (since SQLite has no real support for them).
  - Postgres has support for those data types so we want to leverage that.
  - Again, we abstracted away the data type conversions behind a `Converter` interface (`src/client/conversions/converter.ts`).
    - Then we created 2 specific converters, 1 for SQLite `src/client/conversions/sqlite.ts` and 1 for PG `src/client/conversions/postgres.ts` 
- Similar to the value conversions, we also had to change serialisation/deserialisation of values in the satellite client. Currently it was encoding SQLite values on the wire and decoding values from the wire into SQLite values. We abstracted those encoders/decoders and created 2 instances of each, one for SQLite and one for PG (cf. `src/util/encoders`).


kevin-dp:
Additional TODO:

- Stefanos added BLOB support on main. Will need to also add it here

clients/typescript/src/client/model/builder.ts
```diff
@@ -159,7 +184,12 @@ export class Builder {
     if (!this.shapeManager.hasBeenSubscribed(this._tableName))
       Log.debug('Reading from unsynced table ' + this._tableName)
 
-    const query = squelPostgres.select().from(this._tableName) // specify from which table to select
+    // don't autoquote field names in the selection
+    // because if a field is a BigInt we cast it
+    // and squel would add quotes around the entire cast
+    const query = squelPostgres
+      .select({ autoQuoteFieldNames: false })
```
msfstef, line 191: I understand the issue with bigint but wont this be an issue with field names like "group" that we've seen before? Should `addFieldSelection` do the quoting instead? I see that `castBigIntToText` manually quotes the field but I'm unclear about how the other fields are quoted since it is disabled here.

clients/typescript/src/client/model/builder.ts
```diff
@@ -329,7 +365,7 @@ export function makeFilter(
         prefixFieldsWith
       ),
     ]
-  } else if (isObject(fieldValue)) {
+  } else if (isObject(fieldValue) && !(fieldValue instanceof Date)) {
```
msfstef, line 368: I'm assuming this is where the `isDataObject` (that feels a bit confusing although I see the need now!) would be appropriate?

clients/typescript/src/migrators/triggers.ts
```diff
@@ -278,23 +253,31 @@ export function generateTriggers(tables: Tables): Statement[] {
...
   // serializability into JSON
   const transformIfNeeded = (col: string, targetedCol: string) => {
-    const tpes = colTypes[col]
-    const sqliteType = tpes.sqliteType
-    const pgType = tpes.pgType
+    const colType = colTypes[col]
 
     // cast REALs, INT8s, BIGINTs to TEXT to work around SQLite's `json_object` bug
-    if (sqliteType === 'REAL' || pgType === 'INT8' || pgType === 'BIGINT') {
+    if (
```
msfstef, line 265: nit: another petition here to maybe use a switch statement for readability

clients/typescript/src/satellite/client.ts
```diff
@@ -1373,49 +1425,59 @@ function calculateNumBytes(column_num: number): number {
...
     case PgBasicType.PG_FLOAT4:
     case PgBasicType.PG_FLOAT8:
     case PgBasicType.PG_REAL:
-      return typeDecoder.float(column)
+      return decoder.float(column)
     case PgDateType.PG_TIMETZ:
-      return typeDecoder.timetz(column)
+      return decoder.timetz(column)
     case PgBasicType.PG_BYTEA:
       return column
```
msfstef, line 1446: nit: maybe worth having a decoder for `bytea` and have it be a no-op? an argument could be made in the future for encoding byte arrays for the wire and it'll make it easier to spot

clients/typescript/src/satellite/client.ts
```diff
@@ -1373,49 +1425,59 @@ function calculateNumBytes(column_num: number): number {
...
 ): Uint8Array {
   switch (columnType) {
     case PgBasicType.PG_BOOL:
-      return typeEncoder.bool(columnValue as number)
+      return (encoder.bool as any)(columnValue) // the encoder accepts the number or bool
     case PgDateType.PG_TIMETZ:
-      return typeEncoder.timetz(columnValue as string)
+      return encoder.timetz(columnValue as string)
     case PgBasicType.PG_BYTEA:
       return columnValue as Uint8Array
```
msfstef, line 1470: nit: see argument above about no-op `bytea` encoder/decoder

msfstef (commented):
Amazing work! Left a few minor comments, might do a second pass/review later!

clients/typescript/src/client/model/builder.ts
kevin-dp: `addFieldSelection` returns:
```ts
const fields = identificationFields
  .filter((f) => this._fields.includes(f))
  .concat(selectedFields)
  .map((f) => this.castBigIntToText(f))

return q.fields(fields)
```

If we look at `castBigIntToText` then we see that it quotes all fields:
```ts
if (pgType === PgBasicType.PG_INT8 && this.dialect === 'SQLite') {
  return `cast("${field}" as TEXT) AS "${field}"`
}
return `"${field}"` // <-- we quote it here
```

And all the other functions such add `addOrderBy`, `addDistinct`, etc. also quote all field names.

msfstef: Aah I see, I missed that, thank you for clarifying!

msfstef (approved):
Did a second pass, looking good!

kevin-dp:
> e2e test `03.25_node_satellite_can_resume_replication_after_server_restart.lux` consistently fails on CI but always passes on my machine locally. Not sure why it fails on CI.

When client reconnects, Electric restores some information from the DB we then checked that that information was gone from the DB. However, CI is slow and Electric did not yet restore the information by the time we checked that the information was gone from the DB. Now added an explicit match in the e2e tests to make sure we only check the DB state after that information has been restored by Electric.

clients/typescript/src/satellite/client.ts
```diff
@@ -1373,49 +1425,57 @@ function calculateNumBytes(column_num: number): number {
...
+      return encoder.timetz(columnValue as string)
     case PgBasicType.PG_BYTEA:
-      return columnValue as Uint8Array
+      return encoder.bytea(columnValue as Uint8Array)
+    case PgBasicType.PG_JSON:
+    case PgBasicType.PG_JSONB:
+      return (encoder.json as any)(columnValue)
     default:
-      return typeEncoder.text(columnValue as string)
+      return encoder.text(columnValue as string)
```
msfstef, line 1473: this will accept numbers as well without converting them to strings (but we force cast the type to string) - we should not rely on the encoder being able to handle non-string inputs (see https://github.com/electric-sql/electric/pull/1204/files)

kevin-dp: I don't think we need to address that in this PR, either #1204 gets merged before this one and we will rebase this PR to include it, or this PR gets merged first and yours will then affect the merged code. In both cases, the fix will be present.

icehaunter (commented):
some nitpicks, reviewing other stuff still

clients/typescript/src/satellite/process.ts
```diff
@@ -1435,35 +1451,45 @@ export class SatelliteProcess implements Satellite {
...
-    const triggers = this.opts.triggersTable.toString()
-    if (tablenames.length > 0)
+  _updateTriggerSettings(
+    tables: QualifiedTablename[],
+    flag: 0 | 1
+  ): Statement[] {
+    const triggers = `"${this.opts.triggersTable.namespace}"."${this.opts.triggersTable.tablename}"`
+    const namespacesAndTableNames = tables
+      .map((tbl) => [tbl.namespace, tbl.tablename])
+      .flat()
```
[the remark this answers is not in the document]
kevin-dp: :upside_down_face:

clients/typescript/package.json
```diff
@@ -178,6 +192,7 @@
     "frame-stream": "^3.0.1",
     "get-port": "^7.0.0",
     "jose": "^4.14.4",
+    "kysely": "^0.27.2",
```
icehaunter, line 195: oof, `kysely` + `prisma` + `squel` in direct non-dev deps... we need to do a pass over this

icehaunter: not right now, but soon

kevin-dp: Indeed... used Kysely for internally building some simple queries in a dialect agnostic way. However, those queries are very simple, we could as well not use Kysely and use our own query builder abstraction (iirc the only difference in those queries are the positional parameters, i.e. `?` vs `$i`).

kevin-dp: Had a look again. Kysely is definitely overkill at this point, we're only using it to construct 2 very simple queries.

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -0,0 +1,390 @@
...
+  /**
+   * Checks if the given table exists.
+   */
+  abstract tableExists(tableName: string, namespace?: string): Statement
+
+  /**
+   * Counts tables whose name is included in `tables`.
+   * The count is returned as `countName`.
+   */
+  abstract countTablesIn(countName: string, tables: string[]): Statement
```
icehaunter, line 71: nitpick: tbh not a fan of this name or signature. `countName` is useless, just hardcode that as `result` or `count`, it's literally used once in the code base. Name's also weird, but I'm strugging to come up with a better one. `countTablesIfExist`?

clients/typescript/src/satellite/process.ts
```diff
@@ -968,8 +978,8 @@ export class SatelliteProcess implements Satellite {
     }
 
     try {
-      const oplog = this.opts.oplogTable
-      const shadow = this.opts.shadowTable
+      const oplog = `"${this.opts.oplogTable.namespace}"."${this.opts.oplogTable.tablename}"`
```
icehaunter, line 981: another candidate for `this.opts.oplogTable.toString()` :D

clients/typescript/src/satellite/process.ts
```diff
@@ -1176,15 +1175,21 @@ export class SatelliteProcess implements Satellite {
           tags: encodeTags(entryChanges.tags),
         }
 
+        const qualifiedTableName = QualifiedTablename.parse(tablenameStr)
```
icehaunter, line 1178: Technically should be outside of the inner loop here

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -0,0 +1,390 @@
...
+   * Makes the i-th positional parameter,
+   * e.g. '$3' For Postgres when `i` is 3
+   *      and always '?' for SQLite
+   */
+  abstract makePositionalParam(i: number): string
+
+  /**
+   * Checks if the given table exists.
+   */
+  abstract tableExists(tableName: string, namespace?: string): Statement
```
icehaunter, line 65: nitpick: this function is used in a single place of call, and yet we have an optional argument making for a non-intuitive namespace after table name argument order. Maybe make it non-optional and order as expected? Any place-of-call has access to `instance.defaultNamespace` anyway

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  /**
+   * Returns the given query if the current SQL dialect is PostgreSQL.
+   */
+  abstract pgOnly(query: string): string
+
+  /**
+   * Returns an array containing the given query if the current SQL dialect is PostgreSQL.
+   */
+  abstract pgOnlyQuery(query: string): string[]
```
icehaunter, line 43: unused

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  /**
+   * Returns the given query if the current SQL dialect is SQLite.
+   */
+  abstract sqliteOnly(query: string): string
+
+  /**
+   * Returns an array containing the given query if the current SQL dialect is SQLite.
+   */
+  abstract sqliteOnlyQuery(query: string): string[]
```
icehaunter, line 53: unused

clients/typescript/src/util/relations.ts
```diff
@@ -1,23 +1,25 @@
...
-  opts: SatelliteOpts
+  opts: SatelliteOpts,
+  builder: QueryBuilder
 ): Promise<{ [k: string]: Relation }> {
-  const tableNames = await _getLocalTableNames(adapter, opts)
+  const tableNames = await _getLocalTableNames(adapter, opts, builder)
   const relations: RelationsCache = {}
 
   let id = 0
   const schema = 'public' // TODO
```
icehaunter, line 17: Question: shouldn't this be default schema as well? And used in `getTableInfo` call?

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -0,0 +1,390 @@
...
+   * The information includes all column names, their type,
+   * whether or not they are nullable, and whether they are part of the PK.
+   */
+  abstract getTableInfo(tablename: string): Statement
+
+  /**
+   * Insert a row into a table, ignoring it if it already exists.
+   */
+  abstract insertOrIgnore(
+    table: string,
```
icehaunter, line 108: So definintely for this and all function below, and maybe for 2 above: why are we leaving `schema` as an optional argument at all here, if every place of call fills it? It also occurs to me that most if not all places of call pull both table and schema from a `QualifiedTablename` instance, so why not accept that as the first argument in all of those instead?

clients/typescript/src/config/index.ts
```diff
@@ -44,16 +44,22 @@ export interface ElectricConfig {
   connectionBackOffOptions?: ConnectionBackOffOptions
 }
 
+export type ElectricConfigWithDialect = ElectricConfig & {
+  dialect?: 'SQLite' | 'Postgres' // defaults to SQLite
```
icehaunter, line 48: nitpick: use JSDoc comments in cases like this, helps with hover tooltips


```suggestion
  /** defaults to SQLite */
  dialect?: 'SQLite' | 'Postgres'
```

clients/typescript/src/util/common.ts
```diff
@@ -204,10 +62,5 @@ export function getWaiter(): Waiter {
  * @returns {boolean} whether the `value` is an actual object
  */
 export function isObject(value: any): value is object {
-  return (
-    typeof value === 'object' &&
-    value !== null &&
-    !Array.isArray(value) &&
-    !ArrayBuffer.isView(value)
```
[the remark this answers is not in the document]
kevin-dp: This was not the right place to perform this check, it was moved to the [`isDataObject` function](https://github.com/electric-sql/electric/pull/1038/files/eaf50dd60983069c94e1b541b2de897d2da571e5#diff-fcc94e2387326ffec6132e5b462a3adefd42a29b21e9f00ecb3eec27fb75bea9R27). The reason we have `isObject` and `isDataObject` is because we rely on the [`isFilterObject` function](https://github.com/electric-sql/electric/blob/eaf50dd60983069c94e1b541b2de897d2da571e5/clients/typescript/src/client/conversions/input.ts#L380) to determine whether an input object is a filter (e.g. `{ lt: 5 }`) or a data object (e.g. a date or a UInt8Array or ...). So `isObject` should match any JS object and `isDataObject` should match real input values (data to be stored).

clients/typescript/src/util/tablename.ts
```diff
@@ -14,7 +14,24 @@ export class QualifiedTablename {
   }
 
   toString(): string {
-    return `${this.namespace}.${this.tablename}`
+    // Don't collapse it to "<namespace>.<tablename>" because that can lead to clashes
+    // since both `QualifiedTablename("foo", "bar.baz")` and `QualifiedTablename("foo.bar", "baz")`
+    // would be collapsed to "foo.bar.baz".
```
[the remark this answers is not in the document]
kevin-dp: The quotes in the comment are confusing, the quotes in `"<namespace>.<tablename>"` represent a JS string, the actual namespace and tablename in that string are not quoted. i.e. `this.namespace` is for example `"main"`, note that this is the string "main" and that string does not contain quotes otherwise it would be `"\"main\""`. So in the collapsed version, since there are no quotes around namespace and tablename, we cannot distinguish the two possibilities as per the comment.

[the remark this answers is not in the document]
kevin-dp: Perhaps it would have been more clear if the comment used single quotes in order not to confuse this, so `'<namespace>.<tablename>'` which would lead to 'foo.bar.baz' and we cannot distinguish between the 2 posibilities.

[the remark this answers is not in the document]
kevin-dp: Still, i see your point that if we serialise it already with the quotes we can distinguish both cases and use it as `${table}` in SQL. I will give it a try

clients/typescript/src/satellite/process.ts
```diff
@@ -476,8 +473,19 @@ export class SatelliteProcess implements Satellite {
     lsn: LSN,
     additionalStmts: Statement[] = []
   ) {
+    const namespace = this.builder.defaultNamespace
     const stmts: Statement[] = []
-    stmts.push({ sql: 'PRAGMA defer_foreign_keys = ON' })
+
+    if (this.builder.dialect === 'Postgres') {
```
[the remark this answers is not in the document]
kevin-dp: Yeah, i had it like that originally but didn't like the fact that depending on the dialect it was actually doing something different (deferring vs disabling). But i had it named as `deferFKs` but with the name you propose it is probably fine.

clients/typescript/src/util/tablename.ts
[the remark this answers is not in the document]
kevin-dp: @icehaunter I see you are escaping double quotes in the namespace and tablename. Can double quotes really occur in namespaces and/or tablenames? Quickly tried this `CREATE TABLE "\"weird" (a INTEGER PRIMARY KEY);` but PG is not understanding the escaped quote as it is still waiting for a quote to end this statement. So i don't think it's possible. PG documentation also doesn't mention anything about escaping quotes inside quoted names.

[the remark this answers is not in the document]
icehaunter: I'm sarying the "collapsed" version should use SQL quoting policy of explicitly double-quoting the namespace and the tablename separately, and that properly-escaped version should be the canonical string representation (as it is in SQL)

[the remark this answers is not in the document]
icehaunter: Basically, I'm arguing that the comment "we can't collapse" is countered with "you're collapsing it wrong, it should be SQL standards"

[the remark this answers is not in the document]
kevin-dp: Fixed this as proposes and refactored all instances of `"${table.namespace}"."${table.tablename}"` and modified the query builder methods to take qualified table names instead of taking tablename and namespace separately. You may want to have a look at the way [how qualified table names are parsed](https://github.com/electric-sql/electric/pull/1038/files#diff-0f99aa60443a9a9b65800c5ab13fd3629977a934c14207bfffb06aa942eeb414R22).

clients/typescript/src/util/relations.ts
kevin-dp: The `getTableInfo` call should indeed take the default namespace.
However, the schema in the inferred relation should be `'public'`.
I added a comment explaining why:

```ts
// schema needs to be 'public' because these relations are used
// by the Satellite process and client to replicate changes to Electric
// and merge incoming changes from Electric, and those use the 'public' namespace.
```

clients/typescript/test/migrators/builder.ts
```diff
@@ -0,0 +1,357 @@
...
+    version: '20230613112725_814',
+  }
+}
+
+export type ContextType = {
+  migrationMetaData: ReturnType<typeof makeMigrationMetaData>
+  builder: QueryBuilder
+}
+
+export const bundleTests = (test: TestFn<ContextType>) => {
```
davidmartos96, line 106: I guess this was intended to be named builderTests

clients/typescript/package.json
```diff
@@ -166,6 +179,7 @@
   "dependencies": {
     "@electric-sql/prisma-generator": "workspace:*",
     "@prisma/client": "4.8.1",
+    "@tauri-apps/api": "^1.5.3",
```
icehaunter, line 182: Where is this used? Do we need this as a prod dependency, can't we get it into dev + peers?

clients/typescript/src/util/tablename.ts
```diff
@@ -14,7 +14,26 @@ export class QualifiedTablename {
...
   toString(): string {
-    return `${this.namespace}.${this.tablename}`
+    // Escapes double quotes because names can contain double quotes
+    // e.g. CREATE TABLE "f""oo" (...) creates a table named f"oo
+    return `"${escDoubleQ(this.namespace)}"."${escDoubleQ(this.tablename)}"`
+  }
+
+  static parse(fullyQualifiedName: string): QualifiedTablename {
+    try {
+      const [_, namespace, tablename] = /"(.*)"\."(.*)"/.exec(
```
icehaunter, line 24: Counterexample: `"main"".""".""".""body"` which is a valid SQL value with (in unescaped form) schema `main"."` and table name `"."body`, but won't be parsed as such. Edge case for sure, but also makes invalid strings valid, like `"main""."s".""".""body"`. We also should match string boundaries.

So, to never return to this again, allow only paired double quotes within the quotes. Also identifers can't be empty: 

```suggestion
      const [_, namespace, tablename] = /^"((?:[^"]|"")+)"\."((?:[^"]|"")+)"$/.exec(
```

clients/typescript/src/satellite/process.ts
```diff
@@ -294,24 +293,22 @@ export class SatelliteProcess implements Satellite {
...
     const deleteStmts = tables.map((x) => ({
-      sql: `DELETE FROM ${x.toString()}`,
+      sql: `DELETE FROM ${x}`,
     }))
 
     const stmtsWithTriggers = [
-      // reverts to off on commit/abort
-      { sql: 'PRAGMA defer_foreign_keys = ON' },
-      ...this._disableTriggers(tablenames),
+      { sql: this.builder.deferForeignKeys },
```
icehaunter, line 308: I thought you said this doesn't work on Postgres with non-deferrable constraints?

clients/typescript/src/satellite/process.ts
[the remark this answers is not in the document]
icehaunter: You said in the DMs that you don't like this approach either, but I don't like that Postgres builder implements `deferFKs` which you're now telling me not to use at all. I feel like that's a foot-gun for ourselves next time we add something that requires no FK checks for txn duration. I think builder exposing exactly one field here will force people to think about what's going on. Maybe even more explicit, `deferOrDisableFKsForTxn`

clients/typescript/src/migrators/schema.ts
```diff
@@ -1,30 +1,41 @@
...
+          `CREATE TABLE IF NOT EXISTS ${migrationsTable} (\n  "id" ${builder.AUTOINCREMENT_PK},\n  "version" TEXT NOT NULL UNIQUE,\n  "applied_at" TEXT NOT NULL\n);`,
+          //`-- Initialisation of the metadata table\n`,
+          `INSERT INTO ${metaTable} (key, value) VALUES ('compensations', 1), ('lsn', ''), ('clientId', ''), ('subscriptions', ''), ('seenAdditionalData', '');`,
+          //`-- These are toggles for turning the triggers on and off\n`,
+          `DROP TABLE IF EXISTS ${triggersTable};`,
+          `CREATE TABLE ${triggersTable} ("namespace" TEXT, "tablename" TEXT, "flag" INTEGER, PRIMARY KEY ("namespace", "tablename"));`,
+          //`-- Somewhere to keep dependency tracking information\n`,
+          `CREATE TABLE "${shadowTable.namespace}"."${
+            shadowTable.tablename
+          }" (\n ${builder.pgOnly(
```
icehaunter, line 32:
```suggestion
          `CREATE TABLE "${shadowTable}" (\n ${builder.pgOnly(
```

clients/typescript/src/client/conversions/input.ts
```diff
@@ -21,161 +21,322 @@ type WhereInput = { where?: object }
...
 export function transformFields(
   o: object,
   fields: Fields,
-  transformation: Transformation = Transformation.Js2Sqlite
+  converter: Converter,
+  transformation: Transformation = Transformation.Encode
 ): object {
   // only transform fields that are part of this table and not related fields
   // as those will be transformed later when the query on the related field is processed.
   const fieldsAndValues = Object.entries(keepTableFieldsOnly(o, fields))
```
icehaunter, line 336: nitpick: We go `o: object -> Object.entries -> filter -> Object.fromEntries -> Object.entries -> map -> Object.fromEntries` over this object `o`, which feels hot-pathy to me. I'd rather we do `o -> Object.entries -> filter -> map -> Object.fromEntries`.

Smth like
```typescript
const copied = {...o}

for (const [field, value] of Object.entries(o)) {
  const pgType = fields.get(field)
  // Skip anything that's not an actual column on the table
  if (pgType === undefined) continue

  const transformedValue =
    transformation === Transformation.Encode
      ? converter.encode(value, pgType)
      : converter.decode(value, pgType)

  copied[field] = transformedValue
}

return copied
```

Your `transformField` does `fields.get` and checks if it exists althought this function is the only place of call in entire codebase and you've filtered it out in `keepTableFieldsOnly`. This implementation saves about 3 iterations/transformations (including `Map.keys() -> new Set()` although `Map.has` exists too)

icehaunter: BTW you can make `filterKeys(o: object, keys: Set<string>)` signature to be `filterKeys(o: object, keys: { has: (x: string) => boolean })` and it'll accept map instances just as well

clients/typescript/src/client/model/builder.ts
```diff
@@ -261,7 +286,10 @@ export class Builder {
         )
 
       const squelOrder = order === 'asc' // squel expects 'true' for ascending order, 'false' for descending order
-      return query.order(field, squelOrder)
+      // have to quote the field name ourselves
+      // because Squel does not seem to auto quote
+      // field names in order by statements
+      return query.order(`"${field}"`, squelOrder)
```
icehaunter, line 292: nitpick: this quoting technically doesn't match SQL `.replaceAll('"', '""')`, but I'm not sure how much it currently matters. Worth opening a good-first-issue on GitHub maybe?

clients/typescript/src/client/model/builder.ts
```diff
@@ -377,8 +408,9 @@ export function makeFilter(
       const [filter, handler] = entry
       if (filter in obj) {
         const sql = handler(
-          prefixFieldsWith + fieldName,
-          obj[filter as keyof typeof obj]
+          fieldName,
+          obj[filter as keyof typeof obj],
+          prefixFieldsWith
```
icehaunter, line 413: nitpick: why not do `prefixFieldsWith + '"' + fieldName + '"'` into first argument instead of doing this in every single handler?...

icehaunter (changes requested):
Last batch of comments. Honestly mostly nitpicks, although I really want `package.json` and `deferFKs` addressed before we proceed. Feel free to ignore/address the rest as you see fit.

Very good work overall, speaks to a reasonable code organisation that this change was actually impressively reviewable and contained

clients/typescript/src/drivers/node-postgres/database.ts
```diff
@@ -0,0 +1,135 @@
...
+  rows: Row[]
+  rowsModified: number
+}
+
+export interface Database {
+  name: string
+  exec(statement: Statement): Promise<QueryResult>
+}
+
+export class ElectricDatabase implements Database {
```
samwillis, line 19: @kevin-dp 

Can I just check I'm understanding this correctly, we have a `ElectricDatabase` wrapper for a node-postgres connection which is then passed to electrify rather than passing it the connection directly?

For the embedded postgres this is done automatically with our `createEmbeddedPostgres` wrapper, but for a normal node-postgres a user would have to do this:

```ts
import pg from 'pg'
import { electrify, ElectricDatabase } from 'electric-sql/node-postgres'

// Import your generated database schema.
import { schema } from './generated/client'

// Define custom configuration if needed
const config = {
  url: 'https://example.com:5133'
}

// Create the node-postgres database connection.
const client = new pg.Client({
  // Connection configuration, see:
  // https://node-postgres.com/apis/client
})
await client.connect()
const conn = ElectricDatabase(client, 'name-of-database')

// Instantiate your electric client.
```

Is there a reason the patching of `exec` below can't be done inside the adapter and not this secondary wrapper. It makes initiating the database connection a little complex.

Ideally we could just do:

```ts
const client = new pg.Client({
  // ...
})
await client.connect()
const electric = await electrify(client, schema, config)
```

alco:
@kevin-dp Don't forget to add a changeset.

clients/typescript/src/satellite/process.ts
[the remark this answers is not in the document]
kevin-dp: Applied the proposes change. Also had to change L308 in `_garbageCollectShapeHandler` which was using `deferForeignKeys` but for Postgres that would not work as expected (since that would only defer FKs that were defined as deferrable in the first place). So replaced that by `deferOrDisableFKsForTx`.

clients/typescript/src/migrators/query-builder/pgBuilder.ts
```diff
@@ -0,0 +1,438 @@
...
+  ) {
+    return `CREATE INDEX IF NOT EXISTS ${indexName} ON ${onTable} (${columns
+      .map(quote)
+      .join(', ')})`
+  }
+
+  getLocalTableNames(notIn: string[] = []): Statement {
+    let tables = dedent`
+      SELECT table_name AS name
+        FROM information_schema.tables
```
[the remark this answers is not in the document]
kevin-dp: Great tip! I did a quick and dirty `EXPLAIN ANALYZE` for my original query it took:
```
 Planning Time: 31.000 ms
 Execution Time: 12.000 ms
```

For your query it took:
```
 Planning Time: 15.000 ms
 Execution Time: 13.000 ms
```

clients/typescript/package.json
kevin-dp: It's not being used (anymore?). Removed it.

clients/typescript/src/client/model/builder.ts
kevin-dp: Right, let's make a good-first-issue on GH. We can always pick it up ourselves later in a separate PR.

clients/typescript/src/client/model/builder.ts
kevin-dp: Because some handlers use the field name, e.g. `makeBooleanFilter` uses the field name to index the object.

kevin-dp: Could repurpose the handlers' `prefixFieldsWith` parameter to pass the full field name with prefix and quotes. Such that every handler gets just the name and the full name.

clients/typescript/src/client/model/builder.ts
kevin-dp: Ended up doing it right away.

clients/typescript/src/client/conversions/input.ts
```diff
@@ -193,12 +360,13 @@ export function transformFields(
  * @param transformation Which transformation to execute.
  * @returns The transformed field.
  */
-function transformField(
+export function transformField(
```
icehaunter, line 363: This function is unused, should be deleted

icehaunter (approved):
Thanks for addressing feedback! Final function deletion and LGTM

PR #1039, opened 2024-03-07
chore(client) Remove `db` field from `DatabaseAdapter` and export it to aid in external database adapter creation

samwillis:


kevin-dp (approved):
Great refactoring 👍

PR #1043, opened 2024-03-12
feat(electric): Use two replications slots to allow clients to resume replication after Electric restarts

alco:
Continuing on from https://github.com/electric-sql/electric/pull/1099, this PR introduces the concept of a "persistent replication slot". Now, instead of having an auto-advancing replication slot, Electric can keep around WAL records that contain transactions it has already seen and acknowledged previously. As a result, when Electric restarts, it can repopulate its in-memory cache of transactions and be able to resume clients' replication streams.

New configuration options `ELECTRIC_TXN_CACHE_SIZE` and `ELECTRIC_RESUMABLE_WAL_WINDOW` have been added. The relevant docs are updated in https://github.com/electric-sql/electric/pull/1050.

### DIFF
--- a/components/electric/lib/electric/config.ex
+++ b/components/electric/lib/electric/config.ex
@@ -3,6 +3,12 @@ defmodule Electric.Config do
 
   @spec format_required_config_error(list) :: maybe_string
   def format_required_config_error(potential_errors) when is_list(potential_errors) do
+    format_invalid_config_error(potential_errors, true)
+  end
+
+  @spec format_invalid_config_error(list) :: maybe_string
+  @spec format_invalid_config_error(list, boolean) :: maybe_string
+  def format_invalid_config_error(potential_errors, required? \\ false) do
     errors =
       for {varname, {:error, str}} <- potential_errors do
         "  * #{varname} " <> str
@@ -12,7 +18,7 @@ defmodule Electric.Config do
       Electric.Errors.format_error(
         :conf,
         """
-        The following required configuration options have invalid or missing values:
+        The following#{if required?, do: " required"} configuration options have invalid or missing values:
 
         #{Enum.join(errors, "\n\n")}
         """,
@@ -313,4 +319,63 @@ defmodule Electric.Config do
   defp parse_database(nil, username), do: username
   defp parse_database("/", username), do: username
   defp parse_database("/" <> dbname, _username), do: dbname
+
+  @doc """
+  Parse human-readable memory/storage size string into bytes.
+
+  ## Examples
+
+    iex> parse_human_readable_size("1G", 1)
+    {:ok, #{1024 * 1024 * 1024}}
+
+    iex> parse_human_readable_size("2.23g", 1)
+    {:ok, 2_230_000_000}
+
+    iex> parse_human_readable_size("256M", 1)
+    {:ok, #{256 * 1024 * 1024}}
+
+    iex> parse_human_readable_size("377m", 1)
+    {:ok, 377_000_000}
+
+    iex> parse_human_readable_size("430K", 1)
+    {:ok, #{430 * 1024}}
+
+    iex> parse_human_readable_size("142888k", 1)
+    {:ok, 142_888_000}
+
+    iex> parse_human_readable_size("123456789", 1)
+    {:ok, 123_456_789}
+
+    iex> parse_human_readable_size(nil, 111222333)
+    {:ok, 111_222_333}
+
+    iex> parse_human_readable_size("", 1)
+    {:error, "has invalid value: \\"\\""}
+
+    iex> parse_human_readable_size("foo", 1)
+    {:error, "has invalid value: \\"foo\\""}
+  """
+  @spec parse_human_readable_size(binary | nil, pos_integer) ::
+          {:ok, pos_integer} | {:error, binary}
+
+  def parse_human_readable_size(nil, default), do: {:ok, default}
+
+  def parse_human_readable_size(str, _) do
+    with {num, suffix} <- Float.parse(str),
+         true <- num > 0,
+         suffix = String.trim(suffix),
+         true <- suffix == "" or suffix in ~w[G M K g m k] do
+      {:ok, trunc(num * size_multiplier(suffix))}
+    else
+      _ -> {:error, "has invalid value: #{inspect(str)}"}
+    end
+  end
+
+  defp size_multiplier(""), do: 1
+  defp size_multiplier("k"), do: 1_000
+  defp size_multiplier("K"), do: 1024
+  defp size_multiplier("m"), do: 1_000_000
+  defp size_multiplier("M"), do: 1024 * 1024
+  defp size_multiplier("g"), do: 1_000_000_000
+  defp size_multiplier("G"), do: 1024 * 1024 * 1024
 end
--- a/components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
+++ b/components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
@@ -187,6 +187,20 @@ defmodule Electric.Postgres.CachedWal.EtsBacked do
     #       to be aware of all transaction IDs and LSNs that happen, otherwise flakiness begins. I don't like that,
     #       so we probably want to be able to store a shallower pair than a full transaction object and handle that
     #       appropriately in the consumers. Or something else.
+    #
+    #
+    # 9 Apr 2024. ALCO's UPDATE TO THE ABOVE NOTE FROM ILIA:
+    #
+    # Versions of Postgres before 15 include all transactions in the logical replication
+    # stream, even those that touched tables not included in electric_publication. That is the
+    # source of empty transactions Electric sees on the replication stream. I'm not aware of
+    # other sources of such transactions that have an empty list of changes.
+    #
+    # Since version 15.0, Postgres no longer sends such empty transactions. It stands to reason
+    # that this whole comment blob can be removed for good.
+    #
+    # Here's the relevant change in Postgres' source tree -
+    # https://www.postgresql.org/message-id/E1nZNz3-001zFN-UA%40gemulon.postgresql.org
     |> Stream.each(
       &Logger.debug(
         "Saving transaction #{&1.xid} at #{&1.lsn} with changes #{inspect(&1.changes)}"
--- a/components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
+++ b/components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
@@ -182,7 +182,6 @@ defmodule Electric.Postgres.CachedWal.EtsBacked do
   @spec handle_events([Transaction.t()], term(), state()) :: {:noreply, [], any}
   def handle_events(events, _, state) do
     events
-    |> Stream.each(& &1.ack_fn.())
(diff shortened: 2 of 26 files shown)


components/electric/lib/electric/config.ex
```diff
@@ -313,4 +319,62 @@ defmodule Electric.Config do
   defp parse_database(nil, username), do: username
   defp parse_database("/", username), do: username
   defp parse_database("/" <> dbname, _username), do: dbname
+
+  @spec parse_human_readable_size(binary, pos_integer) :: {:ok, pos_integer} | {:error, binary}
```
icehaunter, line 323:
```suggestion
  @spec parse_human_readable_size(binary | nil, pos_integer) :: {:ok, pos_integer} | {:error, binary}
```

Slightly wrong spec

components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
```diff
@@ -324,9 +360,30 @@ defmodule Electric.Replication.Postgres.LogicalReplicationProducer do
...
     Logger.debug("Acknowledging #{lsn}", origin: origin)
     Client.acknowledge_lsn(conn, lsn)
   end
+
+  # Advance the replication slot to let Postgres discard old WAL records.
+  #
+  # TODO: make sure we're not removing transactions that are about to be requested by a newly
+  # connected client. See VAX-1552.
+  #
+  # TODO(optimization): do not run this after every consumed transaction.
```
icehaunter, line 374: So I have 2 questions here:

1. Why are we running this whenever we see a `Commit`? We have `ack_fn` set to `ack` being the previous function - I believe it makes more sense to advance the slot as we do right now - once stored in memory at least (although I'll agree that it matters a bit less given that we're advancing a secondary slot here).
2. I see the "optimization" note you've left here, but I'm not sure I'm comfortable leaving this as a TODO. Advancing a slot on every transaction is fine-ish, however we're opening a new fully separate connection to PG on every received transaction. This is going to be extremely slow and costly for us, and PG won't be able to keep up with the amount of connections we're opening for one-shot queries. I'd be against even a pooling approach here - I believe we need to just open and hold a stable connection to PG to perform these tasks.

components/electric/lib/electric/replication/postgres/slot_server.ex
```diff
@@ -371,7 +373,7 @@ defmodule Electric.Replication.Postgres.SlotServer do
          %Changes.Transaction{commit_timestamp: ts, changes: changes, origin: origin},
          %State{} = state
        ) do
-    first_lsn = Lsn.increment(state.current_lsn)
+    first_lsn = Lsn.increment(state.current_lsn, @lsn_step)
```
icehaunter, line 376: What's the reason to introduce these steps?

icehaunter (changes requested):
Great work, exactly what I had in mind myself, and additional thanks for cleaning and testing the LSN module. One big question from me

components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
alco: The optimization note wasn't meant to be left in the code for long. I split the code into multiple PRs to help myself draw boundaries between changes to different components and to get to a working version faster to then be able to focus on the edge cases and inefficiencies.

So, point taken. Holding a persistent connection might be the way to go if the current approach of advancing the slot often remains in place. Though I'm hoping to implement more of a batched approach, e.g. advancing every N transactions or in size-on-disk increments, at which point relying on a connection pool may be sufficient. I'll circle back to this discussion when I have more concrete ideas to share but I'm expecting those to be implemented in a follow-up PR in any case.

---

N.B.: I had a connection pool implemented as part of the early implementation work but had to take that out for now just to keep the scope of changes that are already spread over multiple PRs manageable.

icehaunter: With or without batching, holding a single open connection just for slot advancement is fine, since only one process is likely to ever do logical replication stream consumption. I think just spawn a sibling GenServer that holds the advancing connection - to split the responsibility a little - and just `cast` to that server on `ack` - this makes the whole approach easy and doesn't clog up logical replication consumer with ack message handling since `cast`s will be sent to a different process. And if that genserver goes down and we lose a couple casts in the middle - no problem, worst case we'll have slightly more WAL stored at a point in time, but it'll resolve on any next cast. 

I'm not sure how you want to structure the PR merge sequence, but I would insist on adding this functionality in this PR. It's not much code, but the difference in perfomance implications is huge.

alco: The implementation has changed now that this PR is based on https://github.com/electric-sql/electric/pull/1099 which is a fix for https://github.com/electric-sql/electric/issues/1083.

See the new commit - https://github.com/electric-sql/electric/pull/1043/commits/76addc4da4b5a4f5d092dfaace65fc964b99e3d5.

TL;DR: LogicalReplicationProducer now opens a persistent "service connection" during initialization and uses it to advance the main slot with the help of a timer.

components/electric/lib/electric/postgres/lsn.ex
```diff
@@ -1,62 +1,194 @@
...
   end
 
   defimpl String.Chars do
-    alias Electric.Postgres.Lsn
-    def to_string(lsn), do: Lsn.to_string(lsn)
+    def to_string(lsn), do: "#{Electric.Postgres.Lsn.to_iolist(lsn)}"
+  end
+
+  defimpl List.Chars do
+    def to_charlist(lsn), do: ~c'#{Electric.Postgres.Lsn.to_iolist(lsn)}'
```
magnetised, line 192: nitpick: do you need to wrap this with the `~c` sigil?

alco: Mix formatter has been adding those since [Elixir v1.15.0](https://github.com/elixir-lang/elixir/blob/v1.15/CHANGELOG.md#v1150-2023-06-19):
> [Code] Code.format_string!/2 now converts 'charlists' into ~c"charlists" by default

magnetised (approved):
Super

components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
```diff
@@ -334,21 +411,45 @@ defmodule Electric.Replication.Postgres.LogicalReplicationProducer do
...
+    repl_conn = state.repl_conn
     origin = state.origin
 
     %Transaction{
       transaction
       | lsn: end_lsn,
         # Make sure not to pass state.field into ack function, as this
         # will create a copy of the whole state in memory when sending a message
-        ack_fn: fn -> ack(conn, origin, end_lsn) end
+        ack_fn: fn -> ack(repl_conn, origin, end_lsn) end
```
icehaunter, line 422: I think this `ack` should not operate on `repl_conn`. 

`ack_fn` function is going to be called in a different process later down the line - in particular, in WAL cache consumer stage. This means that if this repl connection is closed for any reason, old transcations will call it with a stale value and probably crash - I don't know how `:epgsql.standby_status_update/3` acts when it's passed a nonexistent pid. 

To decouple that a little and make sure that the process calling the `ack_fn` doesn't crash, I'd rather `ack_fn` sent a `GenServer.cast/2` to a process that would actually hold the repl conn and do the acknowledge. If the cast fails because the holding GenServer is temporarily dead -- no problem, main code continues to operate, and worst case we acknowledge on the next magic write anyway. So I'd rather put this "advancing" repl conn into a separate GenServer that can restart and reconnect separately, and just `GenServer.cast` into it - maybe make it a `:gproc` via-tuple using the origin so that we don't even care where the repl process currently lives. 

What do you think?

icehaunter (changes requested):

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
```diff
@@ -182,19 +182,30 @@ defmodule Electric.Postgres.CachedWal.EtsBacked do
...
+    # https://www.postgresql.org/message-id/E1nZNz3-001zFN-UA%40gemulon.postgresql.org
     |> Stream.each(
       &Logger.debug(
         "Saving transaction #{&1.xid} at #{&1.lsn} with changes #{inspect(&1.changes)}"
       )
     )
-    |> Stream.map(fn %Transaction{lsn: lsn} = tx ->
-      {lsn_to_position(lsn), %{tx | ack_fn: nil}}
-    end)
+    |> Stream.map(fn %Transaction{} = tx -> {lsn_to_position(tx.lsn), tx} end)
```
icehaunter, line 208: nitpick: I don't understand why you don't like destructuring assignment of `lsn`, given that it's faster than dynamic dispatch of a `.` operator. Doesn't really matter in this case, just caught my eye.

components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
```diff
@@ -339,37 +374,28 @@ defmodule Electric.Replication.Postgres.LogicalReplicationProducer do
...
-  defp emit_events(state, []) do
-    {:noreply, [], state}
-  end
+  defp dispatch_events(%{demand: demand, queue: queue, queue_len: queue_len} = state) do
+    {to_emit, queue_remaining} = :queue.split(demand, queue)
+    to_emit |> :queue.last() |> ack(state)
 
-  defp emit_events(state, events) do
-    {:noreply, Enum.reverse(events), state}
+    state = %{state | queue: queue_remaining, queue_Len: queue_len - demand, demand: 0}
```
icehaunter, line 397: Typo that should lead to a runtime error, possibly means that this code path is not covered by test (although this is caught by Dialyzer)
```suggestion
    state = %{state | queue: queue_remaining, queue_len: queue_len - demand, demand: 0}
```

icehaunter (approved):
Nice design, I like it! a couple of nitpicks, but nothing more. Let's coordinate a release before merging this though

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
alco: It's not that I don't like it. As a rule of thumb, I use pattern matching in function heads when the same field is used multiple times in the function body. Because we have instances in the code where some fields are pattern-matched while others are accessed via the dot operator, the split usually looks arbitrary.

I believe that BEAM is smart enough to optimize uses of the dot operator if there's pattern-matching/guard in the function head. Can't find the source now, but here's at least a partial confirmation of that https://elixirforum.com/t/way-too-much-detail-about-matching-in-the-head-vs-accessing-maps-in-the-body/49167/2

components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
alco: Yeah, I had to stop working on this abruptly yesterday and pushed what I had ready by then. This condition requires a specific test setup to cover. I think having dialyzer complain about it is good enough.

components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
alco: As we've discussed elsewhere, it is fine to acknowledge transactions inside LogicalReplicationProducer and rely on Elixir's supervision tree to restart things if anything fails. Since we now have a replication slot with a persistent starting point, we can replay already-seen transactions when a new replication connection opens.

PR #1044, opened 2024-03-12
chore(electric): Refactor CachedWal.EtsBacked to use the new WAL config

alco:
Follow-up to https://github.com/electric-sql/electric/pull/1043.

This PR makes the Elixir process run by `EtsBacked` scoped under the connector origin, similar to how other Postgres connector processes behave. It also updates telemetry metrics to use more precise names for what the `EtsBacked` cache actually stores: instead of "cache count" it's "transaction count".

### DIFF
--- a/components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
+++ b/components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
@@ -26,7 +26,7 @@ defmodule Electric.Postgres.CachedWal.EtsBacked do
 
   @typep state :: %{
            notification_requests: %{optional(reference()) => {Api.wal_pos(), pid()}},
-           table: ETS.Set.t(),
+           table: :ets.table(),
            last_seen_wal_pos: Api.wal_pos(),
            current_tx_count: non_neg_integer(),
            wal_window_size: non_neg_integer()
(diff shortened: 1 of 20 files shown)


components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
```diff
@@ -108,15 +116,17 @@ defmodule Electric.Postgres.CachedWal.EtsBacked do
 
   @impl GenStage
   def init(opts) do
-    set = ETS.Set.new!(name: @ets_table_name, ordered: true)
+    origin = Keyword.fetch!(opts, :origin)
+
+    table = :ets.new(ets_table_name(origin), [:named_table, :ordered_set])
     Logger.metadata(component: "CachedWal.EtsBacked")
```
icehaunter, line 122: Wouldn't hurt to add `origin` to the state just in case (& it will make e.g. observer usage a touch easier), and to logging metadata here

components/electric/lib/electric/postgres/cached_wal/ets_backed.ex
```diff
@@ -191,19 +201,18 @@ defmodule Electric.Postgres.CachedWal.EtsBacked do
...
         # All transactions were empty
         {:noreply, [], state}
 
-      {{position, _}, total} ->
+      {position, _} ->
         state =
           state
           |> Map.put(:last_seen_wal_pos, position)
           |> fulfill_notification_requests()
-          |> Map.update!(:current_cache_count, &(&1 + total))
```
icehaunter, line 206: What's the reason to remove this tracking field? It had a nice property of allowing us to touch ETS only when it reached a certain value, and also to keep live track of how the cache is growing without asking ETS for size each time

icehaunter (changes requested):
I like scoping the cache by origin, should maybe make our tests more parallel in the future. I do have a couple of questions to the impl

icehaunter (approved):
Ok, LGTM!

PR #1045, opened 2024-03-12
chore(examples): Update Ionic demo

msfstef:
Updates the ionic demo to use the latest version of Electric.

The capacitor driver has a few issues, the iOS ones have been fixed which is why I'm opening this PR but Android still isn't working and I believe is because `RETURNING` statements are not working properly but I need to dig deeper.


examples/ionic-demo/ios/App/App/Assets.xcassets/AppIcon.appiconset/AppIcon-512@2x.png
samwillis, line 1: We should keep the "electric appointments" app icon.

samwillis (approved):
👍 All good, just the missing icon.

PR #1047, opened 2024-03-12
chore(examples): Update `react-native` example to use `op-sqlite` driver

msfstef:
Replace `react-native-storage-sqlite` with `op-sqlite` driver.

Included a few minor fixes, like excluding config files from the eslint config, using the experimental metro flags for export resolution, and removing unnecessary polyfills.

I've also updated the docs for the react-native driver integration to use op-sqlite.


samwillis (approved):
👍

PR #1048, opened 2024-03-12
Add polyfills section to expo documentation

malte-j:
When using ElectricSQL with react native, the build fails because electric is using some browser APIs that are not available in  react native / expo. To get it to compile, you need to polyfill the missing function calls. This is definitely not the only way to do this, but the only one I got working. This should also work for react native without expo, but I don't have this set up to test it properly right now.

### DIFF
--- a/docs/integrations/drivers/mobile/expo.md
+++ b/docs/integrations/drivers/mobile/expo.md
@@ -18,6 +18,79 @@ npx expo install expo-sqlite
 
 This package includes both the regular and `next` drivers. See the [expo-sqlite docs](https://docs.expo.dev/versions/latest/sdk/sqlite/) or [expo-sqlite/next docs](https://docs.expo.dev/versions/latest/sdk/sqlite-next/) for more information.
 
+## Polyfills
+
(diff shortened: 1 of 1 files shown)


msfstef:
Thank you for updating the docs!

We have pushed a fix for this to avoid requiring people to install polyfills for electric to work, see merged PR https://github.com/electric-sql/electric/pull/1009

The only polyfill that will not be strictly required but included in the drivers and encouraged is the Crypto API `getRandomValues` polyfill.

I'm reluctant to add to the documentation as we should be releasing a new version that will render this obsolete.

malte-j:
Nice to see this beeing fixed! How long until the release that fixes this? We could merge this in the meantime to avoid confusing users that want to try out electric right now.

msfstef:
Hey @malte-j - version 0.9.6 which released last Fri no longer requires all those polyfills, only the `getRandomValues` one and there's [a section in the docs](https://electric-sql.com/docs/integrations/drivers/mobile/expo#polyfills) for it on both the Expo and RN integrations.

Thanks for the contribution and sorry for the bad timing! I will be closing this PR as it no longer applies.

PR #1049, opened 2024-03-12
chore(electric): Create publication on startup

alco:
This helps fix the replication stream from Postgres to Electric if the publication got deleted by accident.

### DIFF
--- a/components/electric/lib/electric/replication/postgres/client.ex
+++ b/components/electric/lib/electric/replication/postgres/client.ex
@@ -5,6 +5,9 @@ defmodule Electric.Replication.Postgres.Client do
   Uses `:epgsql` for it's `start_replication` function. Note that epgsql
   doesn't support connecting via a unix socket.
   """
+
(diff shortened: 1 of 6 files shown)


components/electric/lib/electric/replication/postgres/client.ex
```diff
@@ -177,6 +180,27 @@ defmodule Electric.Replication.Postgres.Client do
...
+    table_list =
+      tables
+      |> Enum.map(&quote_ident/1)
+      |> Enum.join(", ")
+
+    create_publication(conn, name, "FOR TABLE #{table_list}")
+  end
+
+  def create_publication(conn, name, table_spec) when is_binary(table_spec) do
+    case squery(conn, ~s|CREATE PUBLICATION "#{name}" #{table_spec}| |> IO.inspect()) do
```
icehaunter, line 197: Let's remove `IO.inspect()` here

icehaunter (approved):

PR #1051, opened 2024-03-13
feat(electric): Persist shape subscriptions to be able to resume client replication after Electric restarts

alco:
Follow-up to https://github.com/electric-sql/electric/pull/1044.

This is a part of series of PRs starting with https://github.com/electric-sql/electric/pull/1043, finally allowing clients to reconnect and avoid resetting their local state even if the sync service has been restarted.


components/electric/lib/electric/satellite/subscription_manager.ex
```diff
@@ -39,45 +51,119 @@ defmodule Electric.Satellite.SubscriptionManager do
...
+  defp query(state, query) when is_binary(query) do
+    state.connector_config
+    |> Connectors.get_connection_opts()
+    |> Client.with_conn(fn conn -> :epgsql.squery(conn, query) end)
+  end
+
+  defp query(state, query, params) when is_binary(query) and is_list(params) do
+    state.connector_config
+    |> Connectors.get_connection_opts()
+    |> Client.with_conn(fn conn -> :epgsql.equery(conn, query, params) end)
```
icehaunter, line 167: Can't we open a connection and just hold it in the process instead of reconnecting? It will be more important with shapes PR, but wouldn't hurt here either

icehaunter (commented):
This will be so painful to rebase upon :sob:

components/electric/lib/electric/satellite/subscription_manager.ex
alco: This is me writing code while thinking of an imminent addition of a connection pool to our code base :sweat_smile: Good point.

alco:
> This will be so painful to rebase upon 😭

I'd like to lead the rebasing effort on your PR after my series is merged as a while to get more familiar with the refactorings you've made to support shapes.

PR #1052, opened 2024-03-13
fix(client): Use `execute` Capacitor driver API to fix parsing issue

msfstef:
Using the `execute` API rather than the `run` one for when no bind values are provided as it has less pre-processing and solves a parsing issue that `run` has for statements like `BEGIN` and `ROLLBACK`.

Not strictly necessary but provides an immediate solution to a bug that is blocking the use of this driver on Android without requiring changes from the driver side.

See this issue: https://github.com/capacitor-community/sqlite/issues/526

### DIFF
--- a/.changeset/early-schools-taste.md
+++ b/.changeset/early-schools-taste.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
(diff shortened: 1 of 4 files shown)


samwillis (approved):
👍 good catch!

PR #1061, opened 2024-03-14
chore(client) Export base adapter types to make creating external adapter possible.

samwillis:
Change to #1039 which is unreleased.


samwillis:
> looks good! I wonder if we should be exporting the generic database adpaters as well?

Thanks. That's one of the things added in #1036 👍

PR #1062, opened 2024-03-14
chore(examples): Update react-native example to use 0.9.5

msfstef:
Removes packed electric dependency, to be merged as soon as 0.9.5 is released.


samwillis (approved):
👍

PR #1063, opened 2024-03-14
fix(client): Remove `react-native-get-random` values dependency

msfstef:
Transitive depedencies don't install the corresponding pods in projects and thus the polyfill fails unless its included in the root project.

We will fire a warning and add docs explaining that this polyfill ensures better uniquness guarantees for UUID generation.


samwillis (approved):
👍

PR #1067, opened 2024-03-14
fix(recipes): Disable simulating high cpu load by default

msfstef:
For the sake of deploying the backend to a VM, I'd rather not have artificial CPU loads inserted, natural fluctuations are enough to highlight the pattern.


samwillis (approved):
👍

PR #1069, opened 2024-03-14
chore(examples): Update Expo example and remove unnecessary polyfills

msfstef:
Updates the Electric dependency to 0.9.5 which no longer has explicit dependencies on Web APIs to work (or conditional polyfills were provided internally for cases where it did).

Mobile environments such as Expo and React Native thus no longer require polyfills for APIs such as the `TextEncoder`, `base64` encoding/decoding, and the whole `crypto` API.

The `react-native-get-random-values` polyfill is the only one left in the project, which is not strictly necessary but will ensure that UUID generation has the expected uniqueness guarantees (included in [Expo and RN integration docs](https://electric-sql.com/docs/integrations/drivers/mobile/expo#polyfills)).


icehaunter:
Could you add a paragraph in PR description (and thus the eventual commit) as to why those are unnecessary? For those unfamiliar with Expo intricacies :laughing:

msfstef:
@icehaunter fair point - updated the PR description to include more info about the removal of the polyfills.

PR #1075, opened 2024-03-19
chore(examples): Prepare Recipes example for website demo

msfstef:
Some changes to the recipes example in order to include them as live demos on the website.

I've been having some trouble with the Stackblitz web containers failing to Vite build, seemingly inconsistent (sometimes it works sometimes it doesn't) - I'll have to dig a bit deeper for that.

Here's an example of how we would embed the project: [embed url](https://stackblitz.com/github/electric-sql/electric/tree/msfstef/prepare-recipes-for-demo/examples/recipes?embed=1&startScript=demo&file=src%2Flog_viewer%2FLogViewer.tsx&initialpath=log-viewer&title=Log%20Viewer)


KyleAMathews:

PR #1086, opened 2024-03-22
fix(electric): Filter out dropped columns in `electric.lookup_columns()`

### LINKED ISSUE #1085: Electrification of a table with previously dropped columns fails
Trying to electrify a table with at least one column removed via `ALTER TABLE ... DROP COLUMN ...` fails with an error that looks like this:

```
=# ALTER TABLE foo ENABLE ELECTRIC;
ERROR:  syntax error at or near "-"
LINE 24:     "__reordered_........pg.dropped.16........" -,
                                                         ^
QUERY:  CREATE TABLE electric.shadow__public__foo (
[...]
    "__reordered_........pg.dropped.16........" -,
    "__reordered_........pg.dropped.17........" -,
[...]
    "_tag_........pg.dropped.16........" electric.tag,
    "_tag_........pg.dropped.17........" electric.tag,
[...]
)
CONTEXT:  PL/pgSQL function electric.ddlx_make_or_update_shadow_tables(text,text,oid) line 63 at EXECUTE
SQL statement "SELECT electric.ddlx_make_or_update_shadow_tables('CREATE TABLE', _schema, _oid)"
PL/pgSQL function electric.electrify(text,text) line 50 at PERFORM
SQL statement "CALL electric.electrify(table_name)"
PL/pgSQL function electric.enable(text) line 3 at CALL
```

The error is caused by the fact that the `electric.lookup_columns()` SQL function returns information about columns that have been dropped. Postgres keeps records for such columns in the `pg_attribute` system catalog but renames them to `............pg.dropped.N........` and sets their type OID to `0`.

The solution is to filter out such records when building up a column list in `electric.lookup_columns()`.

alco:
Fixes #1085.

### DIFF
--- a/.changeset/silly-seals-crash.md
+++ b/.changeset/silly-seals-crash.md
@@ -1,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
+Fix electrification of tables with previously dropped columns.
--- a/components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers/shadow_table_creation_and_update.sql
+++ b/components/electric/lib/electric/postgres/extension/migrations/20230512000000_conflict_resolution_triggers/shadow_table_creation_and_update.sql
@@ -1,29 +1,3 @@
-
--- This function returns information about columns of the table based on it's oid, skipping any columns starting with an underscore
-CREATE OR REPLACE FUNCTION electric.lookup_columns(search_oid oid, include_hidden_columns boolean DEFAULT false)
-    RETURNS TABLE (col_name name, col_type text, col_not_null boolean, col_default text, col_primary boolean)
-    STABLE PARALLEL SAFE
-    RETURNS NULL ON NULL INPUT
-    ROWS 10
-    LANGUAGE SQL AS $$
-        SELECT
-            attname AS col_name,
-            pg_catalog.format_type(atttypid, atttypmod) AS col_type,
-            attnotnull AS col_not_null,
-            pg_get_expr(adbin, adrelid) AS col_default,
-            indexrelid IS NOT NULL AS col_primary
-        FROM pg_attribute
-        LEFT JOIN pg_attrdef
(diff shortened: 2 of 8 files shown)


icehaunter:
Agree with @magnetised on the e2e test. In particular, I'm worried about not sending "enough" columns on write in a replication stream and how pg copes

alco:
@magnetised @icehaunter  Not sure what you're both getting at. The change in this PR affects the generation of shadow tables. The structure of the created table can be verified in a unit test, either in `Electric.Replication.ElectrificationTest` or in `Electric.Postgres.ExtensionTest`.

icehaunter:
@alco Thanks for the reminder about this PR, I've been looking for this! So the test I'm worried about, is that Electric will see less columns than exist on PG, and it's triggers and structures on the wire down to PG will not include them. I'm worried that this may cause a failed write in triggers or when sending data to PG. I want a super simple e2e test where we electrify  a table with a dropped column and make a conflicting write to this table from Satellite - to make sure that:
1. Although Electric receives a full object with "dropped" columns from PG, it doesn't fail when serializing it for the client
2. A write from Satellite makes it's way to PG successfully via replication stream (I'm less worried about direct writes)
3. Conflict resolution functions don't get funky because a column present on main table is not present on a shadow table

icehaunter (approved):

davidmartos96:
@alco Over in the Dart client we've encountered some flakiness with the e2e test introduced here. I'm not sure if due to the implementation of the fix or the test itself. Most of the time the expected result is obtained though.

The error in particular is that the `_tag_content` column contains the value of the other satellite client id.

Here are the lux logs if you are interested: 
Action run: https://github.com/SkillDevs/electric_dart/actions/runs/8987002252
Artifact: https://github.com/SkillDevs/electric_dart/actions/runs/8987002252/artifacts/1480542095

davidmartos96:
It could be related with the concerns Ilia had. We are using `logical_replication` mode in the tests, if that matters.

PR #1091, opened 2024-03-24
Ensure that the maximum path length is sufficient

joeyates:
As indicated in the code comment, `IDBBatchAtomicVFS`'s default for `mxPathName` is 64.
SQLite allocates memory for the file name length **plus 8 bytes**, so file names longer than 56 bytes cause SQLite `open` commands to fail.

N.B. This change alone is not sufficient to resolve the problem.

There is an underlying problem caused by a bug in wa-sqlite that causes `mxPathName` to be ignored. The fix is in [PR 163](https://github.com/rhashimoto/wa-sqlite/pull/163).


joeyates:
The wa-sqlite PR has been merged

joeyates:
wa-sqlite [release 0.9.13](https://github.com/rhashimoto/wa-sqlite/releases/tag/v0.9.13) includes this fix.

KyleAMathews:

PR #1098, opened 2024-03-26
fix(electric): Add all published tables to the Electric publication when it's recreated

alco:
This is a follow-up to https://github.com/electric-sql/electric/pull/1049 which was an incomplete solution for VAX-1465.

- [x] add changelog
- [x] write a test


icehaunter (approved):
Good job, and great e2e test to save us in the future

PR #1099, opened 2024-03-26
fix(electric): Fix unbounded WAL size growth on disk

### LINKED ISSUE #1083: Replica size / disk usage
Hi 
First of all what an amazing project. 
Currently we are running a sync server on a digital ocean droplet with a managed postgres db on digital ocean. I have two questions.

1. The postgres disk usage is gradually increasing in linear fashion and within a day filled up a 10 gb instance, Im surly doing something wrong, the data is itself is just a few mb. 
Looking at the Insights it seems that the replica size is growing to several gb.
Any ideas or clues that could help me understand why this is happening ?

2. With regards to logical replication and digital ocean managed db it seems like a prerequisite is to have the ability to create a super user, as far as i know most managed db do not allow this. The alternative is direct write. Comparing the two is there significant down side with the later ?

alco:
As explained in https://github.com/electric-sql/electric/issues/1083#issuecomment-2015354293, Electric cannot let Postgres know it can discard obsolete WAL records until it sees a write to an electrified table.

Here we issue such writes on a regular basis even if there are no user writes being performed by the clients or directly in the database.

Fixes #1083.

- [x] add changelog
- [x] add a test that confirms that periodic writes are happening

### DIFF
--- a/components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
+++ b/components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
@@ -45,7 +45,8 @@ defmodule Electric.Replication.Postgres.LogicalReplicationProducer do
               client: nil,
               origin: nil,
               types: %{},
-              span: nil
+              span: nil,
+              magic_write_timer: nil
 
     @type t() :: %__MODULE__{
             repl_conn: pid(),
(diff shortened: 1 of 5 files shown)


components/electric/lib/electric/replication/postgres/logical_replication_producer.ex
```diff
@@ -106,14 +133,18 @@ defmodule Electric.Replication.Postgres.LogicalReplicationProducer do
...
+      state =
+        %State{
+          repl_conn: repl_conn,
+          svc_conn: svc_conn,
+          queue: :queue.new(),
+          publication: publication,
+          origin: origin,
+          span: span
+        }
+        |> schedule_magic_write()
```
icehaunter, line 145: Can you add a commend for why is this needed on startup?

alco: Is this what you mean?

> The replication slot prevents Postgres from discarding old WAL records until Electric sees a write to an electrified table. However, any writes in Postgres cause the WAL to grow, and, moreover, managed Postgres instances from any provider exhibit some background write rate even when the user application is idle. To keep WAL's disk usage in check, we issue a period no-op write that gets delivered to Electric over the replications stream, allowing it to ack the latest WAL position and let Postgres discard older records.

magnetised (approved):
Magic writes.

PR #1101, opened 2024-03-26
feat(electric): Support BLOB/BYTEA columns

msfstef:


components/electric/lib/electric/postgres/types/bytea.ex
```diff
@@ -28,4 +30,45 @@ defmodule Electric.Postgres.Types.Bytea do
   defp decode_hex_char(char) when char in ?0..?9, do: char - ?0
```
msfstef, line 30: @alco I've noticed that we have custom hex decoding logic here - any particular reason not to use elixir's `Base` module, `Base.decode16` and `Base.encode16` methods?

clients/typescript/src/satellite/oplog.ts
```diff
@@ -360,6 +366,13 @@ function deserialiseRow(str: string, rel: Pick<Relation, 'columns'>): Rec {
     if (columnType === 'INT8' || columnType === 'BIGINT') {
       return BigInt(value)
     }
+    if (
+      (columnType === 'BYTEA' || columnType === 'BLOB') &&
+      typeof value === 'string'
+    ) {
+      return hexStringToBlob(value)
```
alco, line 373: Kevin mentioned that this serialization/deserialization dance could be redundant for oplog processing.

components/electric/lib/electric/replication/postgres/writer.ex
```diff
@@ -172,8 +172,15 @@ defmodule Electric.Replication.Postgres.Writer do
   defp encode_value("t", :bool), do: "true"
   defp encode_value("f", :bool), do: "false"
 
-  defp encode_value(bin, :bytea),
-    do: bin |> Electric.Postgres.Types.Bytea.to_postgres_hex() |> quote_string()
+  # Possible to receive both hex encoded and raw byte arrays so
```
alco, line 175: Have you identified a possible origin of a hex-encoded bytea value ending up here?

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -288,6 +288,17 @@ defmodule Electric.Satellite.Serialization do
...
 
+  # Values of type `bytea` are coming over Postgres' logical replication stream encoded using either
+  # a regular hex encoding or Postgres' legacy escape format, see:
+  # https://www.postgresql.org/docs/current/datatype-binary.html
+  #
+  # We "encode" bytea values by actually decoding them into raw byte arrays to send over the wire,
+  # which avoids any additional size from encoding and having to coordinate with receivers on
+  # the encoding format - what they see is what they get.
+  defp encode_column_value(val, :bytea) do
+    Electric.Postgres.Types.Bytea.from_postgres_serialized(val)
```
alco, line 299: This can be simplified because we ensure Postgres is using the hex encoding for the logical replication stream. See `Electric.Replication.Postgres.Client.set_display_settings_for_replication()`.

I had a draft implementation for this in https://github.com/electric-sql/electric/compare/main...alco/vax-823-bytea-type-support#diff-11ee61155c8d413582313d21317149ae3fc66074cfbf4d345f293152f179f40fR274-R280.

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -529,10 +540,18 @@ defmodule Electric.Satellite.Serialization do
...
+  def decode_column_value!(val, type) when type in [:text, :varchar] do
     val
   end
 
+  # Assume bytea columns received from a Satellite client are raw byte arrays and not encoded in
+  # any way (e.g. hex or base64), so "decoding" them involves turning them into hex strings
+  # such that Postgres can ingest them, see:
+  # https://www.postgresql.org/docs/current/datatype-binary.html
+  def decode_column_value!(val, :bytea) do
+    Electric.Postgres.Types.Bytea.to_postgres_serialized(val, :hex)
```
alco, line 552: Calling `to_postgres_hex()` here would be simpler.

alco (approved):
Beautiful!

I've found just a few minor things that could be improved in Elixir code.

components/electric/lib/electric/postgres/types/bytea.ex
alco: No reason. I have pushed a commit that replaces this hand-rolled implementation with functions from the `Base` module.

clients/typescript/src/satellite/oplog.ts
msfstef: We store the serialised rows in JSON format in the oplog table, so the bytearray _has_ to be serialised somehow. Since it needs to be serialisable by the sqlite oplog triggers one of the only options is hex formatting, so we have to handle that on JS side as well.

So we definitely need to serialise at least once, to insert into the oplog, but we could techinically _not_ deserialise the hex encoded byte array and send it over the satellite replication stream as a hex encoded string, but I opted for sending it as a raw byte array for optimal space efficiency as well as for protocol simplification (what you see is what you get, a raw bytearray, no need to know about its encoding)

components/electric/lib/electric/satellite/serialization.ex
msfstef: Ah that's good to know - I'll remove the unnecessary code and references to the escape format

components/electric/lib/electric/replication/postgres/writer.ex
msfstef: I have not and I was thinking that even if we detect if something _looks_ like a postgres hex encoded bytearray it could just be a coincidence that the bytes are arranged that way and they should not in fact be decoded (unlikely but not impossible) so this "detect and decode" approach could end up causing an insidious bug in the future, so I'll investigate more.

A quick glance makes me suspicious of the LSN entries in `Electric.Replication.SatelliteCollectorProducer` that use `to_postgres_hex` explicitly, I'll have to check.

clients/typescript/src/satellite/oplog.ts
alco: Sounds good. Thanks for explaining this!

components/electric/lib/electric/replication/postgres/writer.ex
alco: > A quick glance makes me suspicious of the LSN entries in `Electric.Replication.SatelliteCollectorProducer` that use `to_postgres_hex` explicitly, I'll have to check.

I know this is hella confusing in the code, so let me explain this bit. `SatelliteCollectorProducer` behaves differently based on the write-to-pg-mode. When the mode is `logical_replication`, the default, `SatelliteCollectorProducer` streams client transactions to `Electric.Replication.Postgres.SlotServer.Postgres.SlotServer`. The latter acts as a fake Postgres primary which the real Postgres is replicating from. `SatelliteCollectorProducer.update_acked_client_lsn()` is only invoked in this mode and it doesn't interact with the `Electric.Replication.Postgres.Writer` module at all.

Now, when the write-to-pg-mode is `direct_writes`, `SatelliteCollectorProducer` doesn't invoke `update_acked_client_lsn()` and simply streams client transactions to `Electric.Replication.Postgres.Writer`. The latter builds an SQL statement to update the acknowledged_client_lsns table and so it will only see one type of `lsn`, there's no need to inspect the value to determine it's encoding.

msfstef: @alco turns out it was the LSN but in the `Writer` itself (in fact in the equivalent method of the one in the `SatelliteCollectorProducer`) - I don't love the solution I've made here but since LSN is not serialized in the same way as everything else that is ingested through satellite we need to handle it somehow.

Change is in this commit: https://github.com/electric-sql/electric/pull/1101/commits/ef176e669a82b2fab789a583551b4dc3345d44ce

Let me know if you agree with the approach - at least it simplifies the code such that there's no need for special handling of byte array encoding.

alco: Looks great!

I've pushed a commit that removed the no longer used `postgres_hex_encoded?()` function.

This is good to merge now. :tada:

msfstef:
Corresponding docs PR: https://github.com/electric-sql/electric/pull/1111

Issue #1095, opened 2024-03-26
Same instance of NonTransactionalDB's query() getting called with SELECTs for different tables but this._fields is the same


ondrej:
Hello,

Simple setup here:
* Two tables, `companies` and `people`
* No FK in `companies`
* One FK in `people`: nullable column `company_id`

In React, sync is established:
```
const { synced } = await db.people.sync({
  include: { companies: true },
});
```
And query is made to return a particular person with their company (`company_id` is set):
```
db.people.liveMany({
  where: { id: { equals: 'e1407498-2eb8-4733-8d6f-96403ffc0588' } },
  include: { companies: true },
}),
```

I get the error `SyntaxError: Unexpected token 'T', "Test Company" is not valid JSON`.

Setting some breakpoints in nonTransactionalDB.js, I see NonTransactionalDB getting constructed once with `_fields` set to `people`'s fields. That same instance then has `query()` run twice:
* Once with a SELECT statement for the person with ID e1407....
* Once with a SELECT statement for their company

But because _fields is the same, `transformFields()` eventually fails — in my case because both tables have a column with the same name (`name`) but the column is a JSON in `people` and a simple string in `companies`.

ondrej:
What may be happening. As expected:
1. Two Tables get created for each table, with their own executors created with their respective `_fields`
2. `findMany()` gets called on the `people` table with a `where` and `include` (for `companies`)
3. `findMany()` uses its own executor to call its `_findMany()` with the correct `db` (NonTransactionalDB)
4. A NonTransactionalDB gets created for `people`
5. `_findMany()` gets called with a `people`'s `db` and calls `db.query()` (table, SQL statement, db all match)
6. `db.query()`'s callback is called and we call `fetchIncludes()` with `results` (from `people`), what to include, and `people`'s db
7. `fetchIncludes()` eventually calls `fetchInclude()`, still with `people`'s `db`
8. `fetchInclude()` eventually calls `fetchRelated()`

The problem seems to be around here:
1. `fetchRelated()` looks up the other table (`companies`) and calls its `_findMany()` — but with `people`'s `db`

ondrej:
This would repro in the ElectricSQL LinearLite example as well but it doesn't surface because a) the example never uses `include: ` and b) no 2+ tables have columns with the same name but different type.

For example, if you add:
```
include: {
  issue: true,
}
```
[...]

PR #1107, opened 2024-03-28
fix(client): Fix incorrect field transformation in queries including related tables

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1095

All operations must remain synchronous and use callbacks and continuations due to some driver limitations, which led to the passing around of `DB` instances that had field transformation and validation incorporated for a specific table.

This was not an issue for most cases as either related tables won't have shared column names, or if they do they are likely to also share the type and thus the field transformation worked.

In cases where a related table might have a column with the same name as the source table but a different type, the `DB` instance would try to validate/transform the related table column with the assumption that it is of the type of the source table column, leading to either 1) incorrect transformations (like stringified jsons) or 2) errors in attempts to transform (like strings being parsed as jsons)

I've added a few tests for this case and a workaround by allowing the DB instances to be cloned with a different table schema. I'd prefer a cleaner solution but it might require significant refactoring so I think this will do the trick for now - the bug is serious enough that I think we should take care of it ASAP.

### DIFF
--- a/clients/typescript/test/client/model/table.test.ts
+++ b/clients/typescript/test/client/model/table.test.ts
@@ -89,6 +89,10 @@ const profile2 = {
   userId: 2,
 }
 
+// An invalid JSON that will throw if `JSON.parse()` is called
+// on it - the valid string literal would be "'invalid json"'
+const invalidJson = 'invalid json'
+
 const sortById = <T extends { id: number }>(arr: Array<T>) =>
   arr.sort((a, b) => b.id - a.id)
 
@@ -300,7 +304,7 @@ test.serial(
           create: {
             id: 4013,
             name: 'kevin',
-            meta: 'invalid json',
+            meta: invalidJson,
           },
         },
       },
@@ -317,7 +321,7 @@ test.serial(
       user: {
         id: 4013,
         name: 'kevin',
-        meta: 'invalid json',
+        meta: invalidJson,
       },
     })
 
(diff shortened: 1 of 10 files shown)


clients/typescript/test/client/model/table.test.ts
```diff
@@ -282,6 +288,43 @@ test.serial('create query', async (t) => {
...
+    const res = await profileTable.create({
+      data: {
+        id: 4012,
+        bio: 'test',
+        meta: { bar: 'test' },
+        user: {
+          create: {
+            id: 4013,
+            name: 'kevin',
+            meta: 'invalid json',
```
kevin-dp, line 303: Nit: maybe leave a comment as to why this is invalid json because it is a bit subtle. A string literal is valid json, but in order to represent a string literal in JSON we need to quote it inside or JS string, i.e. `JSON.parse('invalid json')` doesn't work because `invalid json` is not valid JSON but `JSON.parse('"valid json"')` does work because `"valid json"` is just a string literal.

kevin-dp:
Thanks for fixing this!
The related table has an executor instance which internally does create the right kind of DB with the right fields, but since we really need to create a copy of the DB instance at several points in the code (but with the fields of the related table) your solution looks right and simple.

kevin-dp (approved):
Thanks for fixing this!
The related table has an executor instance which internally does create the right kind of DB with the right fields, but since we really need to create a copy of the DB instance at several points in the code (but with the fields of the related table) your solution looks right and simple.

PR #1108, opened 2024-03-28
fix(electric): Sync rows that were in the table prior to electrification

### LINKED ISSUE #1106: Client doesn't receive rows if those rows existed on the table prior to electrification
Setup:
1. Create a simple table (no FK, etc.)
2. Add a row to the table
3. Electrify the table
4. Sync the table on the client (`const shape = await db.{table}.sync(); await shape.synced;`)
5. Query the table on the client (`const { results } = useLiveQuery(db.{table}.liveMany());`)

Expected results: `results` contains the row.

Actual results: `results` doesn't. If you modify the row in any way, it will.

Also in Discord https://discordapp.com/channels/933657521581858818/1221983203812835328

alco:
When a table is electrified, its newly created shadow table is empty. If the user table had any rows prior to electrification, they would remain invisible to Electric due to the JOIN between the user table and its shadow table performed when querying for initial subscription data.

Fixes #1106, VAX-1750.

### DIFF
--- a/components/electric/lib/electric/postgres/schema.ex
+++ b/components/electric/lib/electric/postgres/schema.ex
@@ -1,12 +1,13 @@
 defmodule Electric.Postgres.Schema do
-  alias Electric.Postgres.Schema.Proto
+  import Electric.Postgres.Extension, only: [is_extension_relation: 1]
+  import Electric.Postgres.Schema.Proto, only: [is_unique_constraint: 1]
+
+  alias Electric.Postgres.Extension
   alias Electric.Postgres.Replication
+  alias Electric.Postgres.Schema.Proto
   alias PgQuery, as: Pg
 
   require Logger
-
-  import Electric.Postgres.Extension, only: [is_extension_relation: 1]
-  import Electric.Postgres.Schema.Proto, only: [is_unique_constraint: 1]
 
   @public_schema "public"
   @search_paths [nil, @public_schema]
@@ -509,7 +510,7 @@ defmodule Electric.Postgres.Schema do
         &Map.update!(&1, :name, fn n -> String.slice("__reordered_#{n}", 0..63) end)
       )
 
-    {schema, table_name} = shadow_table_name(main.name.schema, main.name.name)
+    {schema, table_name} = Extension.shadow_of({main.name.schema, main.name.name})
 
     {:ok, oid} = oid_loader.(:table, @schema, table_name)
 
@@ -523,21 +524,5 @@ defmodule Electric.Postgres.Schema do
       constraints: Enum.filter(main.constraints, &match?(%{constraint: {:primary, _}}, &1)),
       indexes: []
     }
-  end
-
-  @doc """
-  Returns the schema and name of the shadow table for the given table.
-  """
-  @spec shadow_table_name(name(), name()) :: namespaced_name()
-  def shadow_table_name(schema, table) do
-    {@schema, "shadow__#{schema}__#{table}"}
-  end
-
-  @doc """
-  Returns the schema and name of the tombstone table for the given table.
-  """
-  @spec tombstone_table_name(name(), name()) :: namespaced_name()
-  def tombstone_table_name(schema, table) do
-    {@schema, "tombstone__#{schema}__#{table}"}
   end
 end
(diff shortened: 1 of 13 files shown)


components/electric/lib/electric/postgres/schema.ex
```diff
@@ -530,9 +530,12 @@ defmodule Electric.Postgres.Schema do
   """
   @spec shadow_table_name(name(), name()) :: namespaced_name()
   def shadow_table_name(schema, table) do
-    {@schema, "shadow__#{schema}__#{table}"}
+    {@schema, "shadow__#{unwrap_quotes(schema)}__#{unwrap_quotes(table)}"}
   end
 
+  defp unwrap_quotes("\"" <> _ = name), do: String.slice(name, 1..-2//1)
```
magnetised, line 536: to be entirely over-defensive about it we should also be replacing any `""` with `"`. Wouldn't `String.trim_trailing\2`be a little more explicit about what we're doing here too?

magnetised (commented):
This probably works for the case you're testing, but I hit a related problem in my permissions e2e test.

https://github.com/electric-sql/electric/pull/993/files#diff-eb1e9904c46985b04740fdb3c781a86be013de57652d796dea7db7e40095ffb6R36

In this test if I swap the order of the inserts and electrification I get an error later in the test when inserting into the project_memberships table caused by logical replication message with `nil` tags. 

Details here: https://linear.app/electric-sql/issue/VAX-1750/inserting-into-a-table-fails-when-electrifying-a-table-with-existing

might be worth adding an e2e test that reproduces this situation

alco:
@magnetised Good catch!

We could try addressing this issue in a general way by creating a view over the shadow table that would return default values for all columns if a shadow row is missing. But then those defaults would need to work in different contexts where shadow rows are queried.

Right now we just have these two places in the code where only `_tags` are queried from the shadow table. So I'm inclined to apply a fix to the problem you've pointed out and, going forward, keep in mind the possibility that a row from a user table might not have a corresponding shadow row.

@icehaunter thoughts?

components/electric/lib/electric/postgres/schema.ex
icehaunter: Technically not equivalent given that a double-quote inside the quoted string will be represented by `""`

icehaunter:
I think addressing it in those places is good enough

components/electric/lib/electric/postgres/schema.ex
alco: I started hating this "defensive" function definition the moment wrote it. I have since changed the approach to use unquoted {<schema>, <table>} tuples and only produce quoted identifiers immediately prior to using them in an SQL string.

components/electric/lib/electric/replication/shapes/querying.ex
```diff
@@ -61,32 +66,40 @@ defmodule Electric.Replication.Shapes.Querying do
           {:ok, [Changes.NewRecord.t()], results(), Graph.t()}
           | {:error, term()}
   defp do_query_layer(conn, %Layer{} = layer, schema_version, origin, context, from) do
-    table =
+    table_info =
       schema_version
       |> SchemaVersion.table!(layer.target_table)
       |> Schema.single_table_info(schema_version.schema)
 
-    columns = Enum.map_join(table.columns, ", ", &~s|this."#{&1.name}"|)
```
magnetised, line 69: we should maybe document a restriction that electrified tables or columns in electrified tables can't contain `"` in their names... seems a) reasonable and b) that it would save us a lot of time defensively escaping quotes everywhere...

components/electric/lib/electric/replication/shapes/querying.ex
```diff
@@ -61,32 +66,40 @@ defmodule Electric.Replication.Shapes.Querying do
...
         parent_pseudo_join(layer, from),
-        context_filters(table, context, layer)
+        context_filters(table_info, context, layer)
       ]
       |> Enum.reject(&(is_nil(&1) or &1 == ""))
       |> Enum.intersperse(" AND ")
 
     where = if where_clauses != [], do: ["WHERE ", where_clauses], else: ""
 
+    user_table = {table_info.schema, table_info.name}
```
magnetised, line 88: isn't this just `layer.target_table`?

magnetised (approved):
Nice

components/electric/lib/electric/replication/shapes/querying.ex
alco: Yeah, great point!

icehaunter (approved):
Very nice job!

PR #1111, opened 2024-03-28
chore(docs): Add BYTEA as supported type and note driver support

msfstef:
Followup to https://github.com/electric-sql/electric/pull/1101

Adds `bytea` as a supported type (without caveats, at least for now :))

Also adds info block with table of driver support for byte arrays.

Deploy preview PR https://github.com/electric-sql/website/pull/119
Deploy preview: https://deploy-preview-119--electric-sql-website.netlify.app/docs/usage/data-modelling/types#supported-data-types

image

### DIFF
--- a/docs/usage/data-modelling/types.md
+++ b/docs/usage/data-modelling/types.md
@@ -48,7 +48,7 @@ We are actively working on building out data type support. If you need a type we
 
 - `boolean`
 - `enum` (any user-defined [enumerated type](https://www.postgresql.org/docs/current/datatype-enum.html))
-- `bytea` (see [supported SQLite drivers](#bytea-driver-support))
+- `bytea` / `blob` (see [supported SQLite drivers](#bytea-driver-support))
 - `jsonb`
 - `uuid`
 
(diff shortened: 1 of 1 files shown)


alco (approved):

msfstef:
I have an open PR to fix the empty blob issue (which only seems to be an issue for empty blobs as bind parameters) on wa-sqlite: https://github.com/rhashimoto/wa-sqlite

thruflo:
@msfstef am I right in understanding that `BYTEA` is the type for a `BLOB`? If so, can I suggest that we make this clear in the docs, i.e.: somehow that we support blobs?

If this unlocks Cozo support, which I believe it does, perhaps we could also add a Cozo example? Or an additional blog post about using Electric's blob support to sync Cozo?

msfstef:
@thruflo yep indeed `BYTEA` is the PG name for byte arrays or blobs. I can make the entry "`bytea` / `blob`" in the docs for clarity.

I'd need a bit more info on how this relates to Cozo (I'm assuming Cozo DB?) and how we're unlocking support for it

PR #1113, opened 2024-03-29
feat: where clause support for GRANTs and ASSIGNs

magnetised:


components/electric/lib/electric/replication/eval/parser.ex
```diff
@@ -19,7 +19,16 @@ defmodule Electric.Replication.Eval.Parser do
...
-    defstruct [:args, :type, :implementation, :name, strict?: true, immutable?: true, location: 0]
+    defstruct [
+      :args,
+      :type,
+      :implementation,
+      :name,
+      strict?: true,
+      immutable?: true,
+      location: 0,
+      cast: nil
```
icehaunter, line 30: Is the only reason for this field to make it available in formatter?

magnetised: > Is the only reason for this field to make it available in formatter?

yes

alco (approved):

PR #1116, opened 2024-04-01
feat(electric): Persist client reconnection info to the database to enable non-disruptive server restarts

alco:
A new E2E test that covers this implementation is added in https://github.com/electric-sql/electric/pull/1135.

* ~[ ] persist `move_in_next_ref` for each client so that it doesn't reset after server restart~
* [x] add changelog

### DIFF
--- a/components/electric/lib/electric/postgres/repo.ex
+++ b/components/electric/lib/electric/postgres/repo.ex
@@ -1,0 +1,38 @@
+defmodule Electric.Postgres.Repo do
+  @moduledoc """
+  Ecto repo for managing a pool of DB connections.
+
+  This repo must be started as part of a `PostgresConnectorSup` supervision tree, configured
+  with the same `connector_config` that its siblings use. Under the hood, it will start a
+  dynamic Ecto repo that can be looked up by the name returned from `name/1`.
+
+  Connections in the pool will use SSL if the Postgres connector is configured with
+  `DATABASE_REQUIRE_SSL=true`. This is different from `epgsql`'s default behaviour where it
+  tries to use SSL first and only falls back to using plain TCP if that fails.
+  """
+
+  use Ecto.Repo, otp_app: :electric, adapter: Ecto.Adapters.Postgres
+
+  alias Electric.Replication.Connectors
+
+  @default_pool_size 10
+
+  def config(connector_config, opts) do
+    origin = Connectors.origin(connector_config)
+    conn_opts = Connectors.get_connection_opts(connector_config)
+
+    [
+      name: name(origin),
+      hostname: conn_opts.host,
+      port: conn_opts.port,
+      username: conn_opts.username,
+      password: conn_opts.password,
+      database: conn_opts.database,
+      ssl: conn_opts.ssl == :required,
+      pool_size: Keyword.get(opts, :pool_size, @default_pool_size),
+      log: false
+    ]
+  end
+
+  def name(origin), do: :"#{inspect(__MODULE__)}:#{origin}"
+end
--- a/components/electric/lib/electric/satellite/client_reconnection_info.ex
+++ b/components/electric/lib/electric/satellite/client_reconnection_info.ex
@@ -243,7 +243,7 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
       {{client_id, :_, :_, :subscription, subscription_id}, :_, :_}
     )
 
-    Client.with_pool(origin, fn ->
+    Client.pooled_transaction(origin, fn ->
       subs_uuid = encode_uuid(subscription_id)
 
       Enum.each(
@@ -263,7 +263,7 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
   "shared" rows and/or migrations.
   """
   def store_initial_checkpoint(origin, client_id, wal_pos, sent_rows_graph) do
-    Client.with_pool(origin, fn ->
+    Client.pooled_transaction(origin, fn ->
       :ok = clear_all_data(client_id)
       store_client_checkpoint(client_id, wal_pos, sent_rows_graph)
     end)
@@ -348,7 +348,7 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
           {new_graph, discarded_acc} =
             advance_by_additional_data(new_graph, client_id, received_data, discarded_acc)
 
-          Client.with_pool(origin, fn ->
+          Client.pooled_transaction(origin, fn ->
             delete_discarded_cache_entries(discarded_acc)
             store_client_checkpoint(client_id, new_wal_pos, new_graph)
             store_client_actions(pending_actions)
@@ -771,7 +771,7 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
 
     origin = Connectors.origin(connector_config)
 
-    Client.with_pool(origin, fn ->
+    Client.checkout_from_pool(origin, fn ->
       restore_checkpoint_cache(checkpoint_table)
       restore_subscriptions_cache(subscriptions_table)
       restore_additional_data_cache(additional_data_table)
--- a/components/electric/lib/electric/satellite/client_reconnection_info.ex
+++ b/components/electric/lib/electric/satellite/client_reconnection_info.ex
@@ -450,10 +450,7 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
         end
 
       discarded_acc =
-        merge_discarded(
-          discarded_acc,
-          {@additional_data_ets, {client_id, {:before_txid, txn.xid}}}
-        )
+        merge_discarded(discarded_acc, {@additional_data_ets, {client_id, {:lte_txid, txn.xid}}})
 
       {graph, pending_actions, count + 1, discarded_acc}
     end)
@@ -551,10 +548,10 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
     end
   end
 
-  defp delete_additional_data({client_id, _xmin, order, _subject, _subscription_id} = key) do
+  defp delete_additional_data({client_id, xmin, order, _subject, _subscription_id} = key) do
     :ets.delete(@additional_data_ets, key)
 
-    {@additional_data_ets, {client_id, {:ord, order}}}
+    {@additional_data_ets, {client_id, {:pk, xmin, order}}}
   end
 
   @insert_subscription_query """
@@ -702,21 +699,25 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
     )
   end
 
-  defp merge_discarded(acc, {@additional_data_ets, {client_id, {:ord, order}}}) do
+  defp merge_discarded(acc, {@additional_data_ets, {client_id, {:pk, xmin, order}}}) do
     Map.update(
       acc,
       @additional_data_ets,
-      {client_id, [order], nil},
-      fn {^client_id, orders, txid} -> {client_id, [order | orders], txid} end
+      {client_id, [xmin], [order], nil},
+      fn {^client_id, xmins, orders, txid} ->
+        {client_id, [xmin | xmins], [order | orders], txid}
+      end
     )
   end
 
-  defp merge_discarded(acc, {@additional_data_ets, {client_id, {:before_txid, new_txid}}}) do
+  defp merge_discarded(acc, {@additional_data_ets, {client_id, {:lte_txid, new_txid}}}) do
     Map.update(
       acc,
       @additional_data_ets,
-      {client_id, [], new_txid},
-      fn {^client_id, orders, txid} -> {client_id, orders, max(new_txid, txid)} end
+      {client_id, [], [], new_txid},
+      fn {^client_id, xmins, orders, txid} ->
+        {client_id, xmins, orders, max(new_txid, txid)}
+      end
     )
   end
 
@@ -731,7 +732,10 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
   DELETE FROM
     #{Extension.client_additional_data_table()}
   WHERE
-    client_id = $1 AND (ord = ANY($2) OR coalesce(min_txid <= $3, false))
+    client_id = $1 AND (
+      (min_txid, ord) = ANY(SELECT * FROM unnest($2::xid8[], $3::bigint[]))
+      OR coalesce(min_txid <= $4, false)
+    )
   """
 
   # Given the accumulator of discarded ETS entries, issue one DELETE statement per table to
@@ -743,8 +747,8 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
       {@actions_ets, {client_id, txids}} ->
         Client.query!(@delete_actions_for_xids_query, [client_id, txids])
 
-      {@additional_data_ets, {client_id, orders, txid}} ->
-        Client.query!(@delete_additional_data_query, [client_id, orders, txid])
+      {@additional_data_ets, {client_id, xmins, orders, txid}} ->
+        Client.query!(@delete_additional_data_query, [client_id, xmins, orders, txid])
     end)
   end
 
--- a/components/electric/lib/electric/satellite/client_reconnection_info.ex
+++ b/components/electric/lib/electric/satellite/client_reconnection_info.ex
@@ -765,9 +765,9 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
       Client.query!("SELECT * FROM #{Extension.client_checkpoints_table()}")
 
     checkpoints =
-      for {client_id, wal_pos, sent_rows_graph} <- rows do
+      Enum.map(rows, fn [client_id, wal_pos, sent_rows_graph] ->
         {client_id, wal_pos, :erlang.binary_to_term(sent_rows_graph)}
-      end
+      end)
 
     :ets.insert(checkpoint_table, checkpoints)
   end
@@ -777,10 +777,10 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
       Client.query!("SELECT * FROM #{Extension.client_shape_subscriptions_table()}")
 
     subscriptions =
-      for {client_id, subscription_id, xmin, pos, shape_requests_bin} <- rows do
-        {{client_id, decode_uuid(subscription_id)}, String.to_integer(xmin),
+      Enum.map(rows, fn [client_id, subscription_id, xmin, pos, shape_requests_bin] ->
+        {{client_id, decode_uuid(subscription_id)}, xmin,
          :erlang.binary_to_term(shape_requests_bin), pos}
-      end
+      end)
 
     :ets.insert(subscriptions_table, subscriptions)
   end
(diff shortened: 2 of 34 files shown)


magnetised (approved):
lovely work as usual. would it make sense to pull the persistence functions out of `ClientReconnectionInfo` and into a separate module, just to keep that module a little cleaner? it's doing a lot atm and separating those concerns would make the algo a bit clearer.

looking forward to that e2e test...

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -471,62 +637,199 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
+  defp restore_additional_data_cache(additional_data_table) do
+    {[
+       "client_id",
+       "min_txid",
+       "ord",
+       "subject",
+       "subscription_id",
+       "graph_diff",
+       "included_txns"
+     ], rows} = Client.query!("SELECT * FROM #{Extension.client_additional_data_table()}")
```
icehaunter, line 797: Why rely at all on standard PG column auto-ordering at all and match instead of just listing them in the `SELECT` clause? Same q for all above

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -471,62 +637,199 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
-     }}
+
+    checkpoint_table = :ets.new(@checkpoint_ets, [:named_table, :public, :set])
+    subscriptions_table = :ets.new(@subscriptions_ets, [:named_table, :public, :ordered_set])
+    additional_data_table = :ets.new(@additional_data_ets, [:named_table, :public, :ordered_set])
+    actions_table = :ets.new(@actions_ets, [:named_table, :public, :set])
+
+    origin = Connectors.origin(connector_config)
+
+    Client.with_pool(origin, fn ->
```
icehaunter, line 753: I like the write-through cache approach!

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -471,62 +637,199 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
     ])
   end
 
+  @insert_subscription_data_query """
+  INSERT INTO
+    #{Extension.client_additional_data_table()}(
```
icehaunter, line 642:
```suggestion
    #{Extension.client_additional_data_table()} (
```

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -427,16 +549,16 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
           # Client has seen this subscription! Use it to advance the graph and delete it as confirmed.
           diff = :ets.lookup_element(@additional_data_ets, key, 2)
           graph = merge_in_graph_diff(graph, diff)
-          :ets.delete(@additional_data_ets, key)
+          :ok = delete_additional_data(origin, key)
           # See if we can do one more
-          advance_by_additional_data(graph, client_id, acknowledged)
+          advance_by_additional_data(graph, client_id, acknowledged, origin)
```
icehaunter, line 554: Optimization nitpick: we may want to agg deletions and execute one query instead of checking out a connection for each encountered element in `delete_additional_data`.

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -324,51 +413,54 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
       |> Enum.reduce(state, fn %Transaction{} = txn, {graph, pending_actions, count} ->
         {graph, pending_actions} =
           client_id
           |> pop_additional_data_before(txn.xid)
           |> Enum.reduce({graph, pending_actions}, fn
-            {:transaction, _, diff, included_txns}, {graph, pending_actions} ->
+            {:transaction, diff, included_txns}, {graph, pending_actions} ->
               graph = merge_in_graph_diff(graph, diff)
-              clear_stored_actions(client_id, included_txns)
+              clear_stored_actions(origin, client_id, included_txns)
```
icehaunter, line 456: Same optimization nitpick as below: checking out a connection per item feels suboptimal against aggregating intended writes

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -446,23 +568,67 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
+      client_id,
+      subscription_id,
+      min_txid,
+      ord,
+      shape_requests
+    )
+  VALUES
+    ($1, $2, $3, $4, $5)
+  ON CONFLICT
+    (client_id, subscription_id)
```
icehaunter, line 609: Why an `ON CONFLICT` clause at all here? This should not be reachable at all under current model, so I'd rather it fail visibly when we do horizontal scaling. I'm not sure just taking a later write here is correct in a general case. This applies to other queries as well

components/electric/test/electric/satellite/subscriptions_test.exs
```diff
@@ -29,12 +29,7 @@ defmodule Electric.Satellite.SubscriptionsTest do
 
       token = Electric.Satellite.Auth.Secure.create_token(user_id)
 
-      on_exit(fn ->
-        drain_pids(active_clients())
-        Electric.Satellite.ClientReconnectionInfo.clear_all_data(client_id)
```
icehaunter, line 34: What's the reason to delete that?

icehaunter (changes requested):
Good work! I've a couple of questions and comments I'd like to address before merging, but I really like the general approach with write-through cache

components/electric/test/electric/satellite/subscriptions_test.exs
alco: ClientReconnectionInfo has been moved from global scope (where it used to be started inside the `Application.start()` callback) to origin-based scope. So it will now shut down together with the whole `PostgresConnectorSup` supervision tree.

The function `active_clients()` above this line relies on the global `Electric.Satellite.ClientManager` but since we only start a websocket listener for the duration of a single test case, there's no need to do explicit clean up of clients, IMO. They will get cleaned up from `ClientManager`'s state via process monitors.

components/electric/lib/electric/satellite/client_reconnection_info.ex
alco: Good catch. This is a leftover from my initial attempt to store additional data for subscriptions in the `client_subscriptions` table. Removed.

components/electric/lib/electric/satellite/client_reconnection_info.ex
alco: Yep, thought about this myself when I was writing it. Just finished an optimization pass on the code to add this, though, admittedly, its readability has arguably suffered a bit as a result.

components/electric/lib/electric/postgres/repo.ex
```diff
@@ -0,0 +1,38 @@
...
+  This repo must be started as part of a `PostgresConnectorSup` supervision tree, configured
+  with the same `connector_config` that its siblings use. Under the hood, it will start a
+  dynamic Ecto repo that can be looked up by the name returned from `name/1`.
+
+  Connections in the pool will use SSL if the Postgres connector is configured with
+  `DATABASE_REQUIRE_SSL=true`. This is different from `epgsql`'s default behaviour where it
+  tries to use SSL first and only falls back to using plain TCP if that fails.
+  """
+
+  use Ecto.Repo, otp_app: :electric, adapter: Ecto.Adapters.Postgres
```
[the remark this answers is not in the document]
alco: * Ecto is a simpler choice because it provides a high-level API on top of Postgrex and DBConnection.
* I would like us to adopt a more structured approach to querying and updating our internal tables. Ecto schemas could be used for this, as opposed to writing plain-text queries as I did in this PR for expedience.
* Our own custom pooler on top of `nimble_pool` would be guaranteed to take up more time to write, plus we'd need to redo some of the work already done in Ecto.
* I'm not sure there are any problems with read queries: the only time client reconnection info (including subscriptions) is fetched from the DB is when the ClientReconnectionInfo process starts. All subsequent lookups are done by ETS.

alco:
@magnetised 
> would it make sense to pull the persistence functions out of `ClientReconnectionInfo` and into a separate module, just to keep that module a little cleaner?

I've done an optimization pass on the code since your review, to make fewer checkouts from the connection pool (and fewer SQL queries overall). As a result, it's become even more difficult to untangle persistence from the rest of the implementation than before. Not that the logic in this module is sensitive to stateful operations performed on ETS tables, so we cannot simply reorder those. As for DB queries, I've done a bit of grouping for those, such that they are now less coupled to ETS operations but still very much coupled to the overall module implementation.

components/electric/lib/electric/postgres/repo.ex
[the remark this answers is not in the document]
alco: * The current implementation of `Electric.Postgres.Repo` can be tweaked a bit to make the repo name depend on another parameter in addition to `origin`. That will make it possible to have multiple disjoint pools at the same time.
* I think fetching client's last acked lsn can use the same pool as restoring reconnection info cache from the DB. Both are client-related concerns.
* I agree that a different pool should be used for direct writes because of a different DB access patterns in that case.
* Loading up the ETS cache from DB on client connections sounds nice! I'll refrain from adding it here but will create an issue for it.

I've had somewhat similar thoughts about the in-memory wal window cache: as an example, on startup Electric could fetch last acked lsn for all clients and leave out the lowest quartile. The minimum lsn of the remaining values could be used as the earliest lsn that we keep cached in memory (provided we can still read it from the replication slot). But again, it's not clear to me if it's worth it or whether filling up the whole available cache space is actually a better approach in practice.

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -260,41 +344,106 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
+          )
+
+          if map_size(pending_actions) > 0 do
+            :ets.insert(@actions_ets, Enum.to_list(pending_actions))
+          end
+
+          {new_graph, discarded_acc} =
+            advance_by_additional_data(new_graph, client_id, received_data, discarded_acc)
+
+          Client.with_pool(origin, fn ->
```
icehaunter, line 382: nitpick: this should probably be `with_transaction` or something like that. I'd rather these writes be atomic

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -204,8 +240,15 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
     :ets.match_delete(@subscriptions_ets, {{client_id, :_}, :_, :_, :_})
-    :ets.match_delete(@additional_data_ets, {{client_id, :_, :_, :_, :_}, :_})
+    :ets.match_delete(@additional_data_ets, {{client_id, :_, :_, :_, :_}, :_, :_})
     :ets.delete(@checkpoint_ets, client_id)
+
+    Client.query!(@delete_subscriptions_query, [client_id])
+    Client.query!(@delete_checkpoint_query, [client_id])
+    Client.query!(@delete_actions_query, [client_id])
+    Client.query!(@delete_additional_data_for_client_query, [client_id])
+
```
icehaunter, line 250: Nitpick: wrap in txn and same connection checkout?

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -260,41 +344,106 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
+            :ets.insert(@actions_ets, Enum.to_list(pending_actions))
+          end
+
+          {new_graph, discarded_acc} =
+            advance_by_additional_data(new_graph, client_id, received_data, discarded_acc)
+
+          Client.with_pool(origin, fn ->
+            delete_discarded_cache_entries(discarded_acc)
+            store_client_checkpoint(client_id, new_wal_pos, new_graph)
+            store_client_actions(pending_actions)
```
icehaunter, line 385: nitpick: I'd appreciate a `!` in this function name, because my mind went immediately to the question of error handling in `with_pool/2` return value, but this uses a `query!` which will fail loudly instead - which I agree with, just want a pointer that this is what happens

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -446,23 +614,56 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
-          graph
+          {graph, discarded_acc}
         end
     end
   end
 
+  defp delete_additional_data({client_id, _xmin, order, _subject, _subscription_id} = key) do
+    :ets.delete(@additional_data_ets, key)
+
+    {@additional_data_ets, {client_id, {:ord, order}}}
```
icehaunter, line 635: I see you're using order here as a unique key, but that's incorrect. Order was meant to be a ternary item to sort additional data insertions that have the same `xmin`. Primary key on that table should be `(client_id, min_txid, ord)`, and the deletion here should use that tuple too. Original logic relies on full key deletion because we care about last element of the tuple as the fully disambiguating element, but `(client_id, min_txid, ord)` tuple is also enough. This also addresses your "persist `move_in_next_ref`" concern, as `min_txid` is guaranteed to be advancing.

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -324,77 +473,97 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
       # while additional data diffs are stored as soon as they were received, and
       # acknowledged additional data is removed while advancing. So we can just
       # delete all additional data here and return all the actions.
-      :ets.match_delete(@additional_data_ets, {{client_id, :_, :_, :_, :_}, :_})
+      :ets.match_delete(@additional_data_ets, {{client_id, :_, :_, :_, :_}, :_, :_})
+
+      origin = Keyword.fetch!(opts, :origin)
+      Client.pooled_query!(origin, @delete_additional_data_for_client_query, [client_id])
```
icehaunter, line 479: nitpick: can we wrap this with `advance_checkpoint` call higher up with a `checkout` or a `transaction` call? That should work with nested `checkout` call within `advance_checkpoint` if I remember `Ecto.Repo` docs correctly.

icehaunter (changes requested):
Thank you for addressing my questions! I've a couple more nitpicks, and one important question in my opinion

components/electric/lib/electric/postgres/repo.ex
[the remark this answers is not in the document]
icehaunter: I'd maybe argue for per-client cache filling earlier rather than later, because it gives us a better resiliency pattern: when a query in this module fails (which it definitely can with timeouts, as our demos have shown), WS connection will close, but ETS writes have already been done, which leaves the cache and DB in an inconsistent state. If we cleared the ETS part of the cache for this client when error in this module occurs, disconnected the client, and refilled from PG on reconnection, that would give us more consistent error handling per client instead of all-or-nothing we have right now.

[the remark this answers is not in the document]
alco: Good point!

Still, I wouldn't put that into this PR as this will likely delay its merging by one or more days. There's always more nuance to adding a feature than seems at first.

To address the inconsistency you've described, we can reorder ETS and DB writes such that DB writes happen first. If one fails, the match ETS tables will remain unchanged. It's not perfect but can be addressed in this PR before it's merged.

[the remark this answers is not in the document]
icehaunter: Sure, I'm not against a separate PR, just arguing for not delaying it because it can be very important for correctness. I also don't think reordering ETS writes will be super easy, given we do them in a loop while aggregating DB writes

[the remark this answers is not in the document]
icehaunter: My personal gut feeling is that per-client loading is less work than reordering and rewriting parts of the algrithm, but you're welcome to take it however you want

components/electric/lib/electric/satellite/client_reconnection_info.ex
alco: I have moved this part into the `advance_checkpoint()` function that already has a `pooled_transaction` block.

components/electric/lib/electric/satellite/client_reconnection_info.ex
alco: Thanks, fixed.

components/electric/lib/electric/satellite/client_reconnection_info.ex
alco: Would you like to see all functions that call `Client.query!()`  inside to also have the ! in their name?

components/electric/lib/electric/satellite/client_reconnection_info.ex
alco: Done.

components/electric/lib/electric/satellite/client_reconnection_info.ex
alco: Good point. I've moved the functions that work with a repo into a separate module named `Electric.Postgres.Repo.Client`, updated their names, and added one that executes the anonymous function in a transaction.

components/electric/lib/electric/satellite/client_reconnection_info.ex
icehaunter: Yes, preferrably, just to have a reminder of how error handling is done here

alco: If I do it transitively, almost every function in this module would get a `!` in its name.  I'm thinking of only applying this naming change to public functions, so that uses of this module's API in other modules communicate their error behaviour at a glance, while for private functions inside the module it can be seen from their implementation.

alco: Just noting here that I've pushed a new commit that adds `!` to public function names only.

icehaunter (approved):
Great work, and thanks for addressing my concerns!
I've one more nitpick, but feel free to ignore if you don't agree.
I'll be waiting for the PR on per-client cache filling :D

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -539,7 +540,9 @@ defmodule Electric.Satellite.Protocol do
       end
     else
       # Once the client is outside the WAL window, we are assuming the client will reset its local state, so we will too.
-      ClientReconnectionInfo.clear_all_data(state.client_id)
+      Client.pooled_transaction(state.origin, fn ->
+        ClientReconnectionInfo.clear_all_data(state.client_id)
+      end)
```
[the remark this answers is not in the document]
icehaunter: Especially when all other function calls hide this inside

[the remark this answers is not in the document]
icehaunter: Awesome, thanks!

Issue #1114, opened 2024-03-29
npm run build fails after project generation


agdolla:
Hello!

After I created a new project and tried to build it, I get two errors:

`npx create-electric-app@latest my-app --template react`
`npm run build`

> my-app@0.1.0 build
> tsc && vite build

src/generated/client/index.ts:3:38 - error TS6133: 'Relation' is declared but its value is never read.

3 import { type TableSchema, DbSchema, Relation, ElectricClient, type HKT } from 'electric-sql/client/model';
                                       ~~~~~~~~

src/generated/client/index.ts:298:11 - error TS2430: Interface 'ItemsGetPayload' incorrectly extends interface 'HKT'.
  Types of property 'type' are incompatible.
    Type 'ItemsGetPayload<this["_A"]>' is not assignable to type 'Record<string, any> | undefined'.
      Type '"Please either choose `select` or `include`" | Items | { [x: string]: never; [x: number]: never; }' is not assignable to type 'Record<string, any> | undefined'.
        Type 'string' is not assignable to type 'Record<string, any>'.

298 interface ItemsGetPayload extends HKT {
              ~~~~~~~~~~~~~~~

Found 2 errors in the same file, starting at: src/generated/client/index.ts:3

msfstef:
Thank you for reporting the error!

The first error about `Relation` being declared but never read is going to be fixed by this PR https://github.com/electric-sql/electric/pull/1117 hopefully part of the next release - in the meantime you can either remove it yourself or add a @ts-ignore annotation for it

The second error seems to occur with Typescript 5.4 - reverting to typescript version 5.3.3 will no longer result in this error. I will be looking into this, but fixing TS to 5.3.3 in the meantime should allow you to compile the client.

agdolla:
@msfstef  ok thanks I try with TS 5.3.3

Issue #925, opened 2024-02-06
Cannot work with pnpm


ImSingee:
After upgrading to 0.9.2, can't use with pnpm any more.

Reproduce:

```
npx create-electric-app@latest test
cd test
# change version of electric-sql to 0.9.2
pnpm install
pnpm add zod
pnpm backend:start
pnpm db:migrate
pnpm client:generate
pnpm build
```

Error:

![image](https://uploads.linear.app/0467b297-8a55-4fc0-82da-624730a09e44/7d40c0af-3bf1-441d-9863-4b4f9d9d7587/ca23560a-4d26-4f22-bd0a-69abea3d4367?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzA0NjdiMjk3LThhNTUtNGZjMC04MmRhLTYyNDczMGEwOWU0NC83ZDQwYzBhZi0zYmYxLTQ0MWQtOTg2My00YjRmOWQ5ZDc1ODcvY2EyMzU2MGEtNGQyNi00ZjIyLWJkMGEtNjlhYmVhM2Q0MzY3IiwiaWF0IjoxNzA3MjMyOTY2LCJleHAiOjMzMjc3NzkyOTY2fQ.2v_q3y8MWQBalh3G5RYJgQ-u59D8TdXPBXJSY1bMfG8 "=1724x")

```

> test@0.1.0 build /private/tmp/test2/test
> tsc && vite build

src/generated/client/index.ts:3:38 - error TS6133: 'Relation' is declared but its value is never read.

3 import { type TableSchema, DbSchema, Relation, ElectricClient, type HKT } from 'electric-sql/client/model';
                                       ~~~~~~~~

src/generated/client/index.ts:63:14 - error TS2322: Type 'ZodObject<{ value: ZodOptional<ZodString>; }, "strict", ZodTypeAny, { value?: string | undefined; }, { value?: string | undefined; }>' is not assignable to type 'ZodType<ItemsWhereUniqueInput, ZodTypeDef, ItemsWhereUniqueInput>'.
  Types of property '_type' are incompatible.
    Type '{ value?: string | undefined; }' is not assignable to type 'ItemsWhereUniqueInput'.
[...]

PR #1117, opened 2024-04-01
fix(client, generator): Fix TS issues with generated client and add tests

msfstef:
Relevant issues: https://github.com/electric-sql/electric/issues/1001 https://github.com/electric-sql/electric/issues/925 https://github.com/electric-sql/electric/issues/1114

Only one thing is fixed in this PR, the `Relation` is no longer imported if the data model has no relations in order to avoid `unused imports` TS errors.

Main changes are moving all of the generator tests into their own test directory rather than inside the source code, for consistency with the TS client, ease of use, and improved reusability of test utilities.

I've also added a sort of integration test for the CLI generation step, which is crucial to getting a working, compileable TS client as it modifies some of the Prisma generated typings. This test generates the client given a Prisma schema and performs all appropriate modifications to it, and then attempts to compile it with TS and specified compiler options.

Also found that with TS version >=5.4 we have [stricter conditional type checking](https://devblogs.microsoft.com/typescript/announcing-typescript-5-4/#notable-behavioral-changes) which breaks clients generated with Prisma versions earlier than 4.16.0 (see issue https://github.com/electric-sql/electric/issues/1114), however 4.16.0 introduces breaking changes with the "extensions" feature, so I've added a hacky workaround for it:
https://github.com/electric-sql/electric/pull/1117/files#diff-fb0a50df4fc86f09b81a210162984d7822276fe19c8ae6496cf5ffe7d09f5adb

We can add more tests that would test various compiler options, making it easier to avoid regressions when users report them with more complicated schemas and TS compiler options.

Finally, I've added a Caution banner in our Usage section in the docs about requiring TS `strict` mode, since Zod has that as a requirement as well and its the reason why we see some inconsistencies in bug reports.


clients/typescript/test/cli/migrations/migrate.generation.test.ts
```diff
@@ -0,0 +1,180 @@
...
+  const migrationsPath = `${generatedClientPath}/migrations.ts`
+  fs.writeFileSync(schemaFilePath, inlinePrismaSchema)
+  // clean up the generated client if present
+  fs.rmSync(generatedClientPath, { recursive: true, force: true })
+  await _testing.generateClient(schemaFilePath, generatedClientPath)
+  await fs.writeFileSync(migrationsPath, 'export default []')
+  return generatedClientPath
+}
+
+test.before(() => {
```
kevin-dp, line 161: I was expecting a `test.afterAll` that cleans up the created temp directory which contains the test Prisma schema, etc.

generator/test/functions/writeSingleFileImportStatements.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    expect(
+      dmmf.datamodel.models.every((model) => !model.hasRelationFields)
+    ).toBe(true)
+
+    const writer = new TestFileWriter()
+    const fileString = writer.createFileString((writer) =>
+      writeSingleFileImportStatements(dmmf, writer)
+    )
+
+    console.log(fileString)
```
kevin-dp, line 38: This log statement is probably a leftover from testing?

clients/typescript/package.json
```diff
@@ -152,6 +152,7 @@
     "build": "shx rm -rf dist && npm run build:copy-docker && concurrently \"tsup\" \"tsc -p tsconfig.build.json\" && node scripts/fix-imports.js",
     "build:copy-docker": "shx mkdir -p ./dist/cli/docker-commands/docker && shx cp -r ./src/cli/docker-commands/docker ./dist/cli/docker-commands",
     "test": "ava",
+    "generate-test-client": "npx tsx ./test/client/generateTestClient.ts",
```
kevin-dp, line 155: Not sure why we expose this in our `package.json`?
Is this called manually by the developer before testing?
I was expecting some test to invoke `npx tsx ./test/client/generateTestClient.ts` directly.

kevin-dp (approved):
Great work!
Left some minor comments here and there.

clients/typescript/package.json
kevin-dp: Oh i understand now 👍 
I'm fine keeping it like this at the moment as it's basically an automated version of what we already did manually.

kevin-dp (approved):

PR #1119, opened 2024-04-02
fix(electric): Fix a data race that may occur when a client is sufficiently behind the latest Postgres state

alco:
This PR addresses the following edge case:
1. a client connects to the server
2. the server confirms that the client is in the cached WAL window
3. a garbage collection cycle occurs and the cache drops old transactions as a result
4. the client starts replicating from the server but never receives the transactions it was supposed to get

This leads to the client having an inconsistent state with the server.

The way this is addressed is by implementing a reservation system inside the cache to prevent it from dropping reserved transactions even if they are already outside of the cached WAL window size.


alco:
Moved to a different branch - https://github.com/electric-sql/electric/pull/1199.

Issue #1097, opened 2024-03-26
null in a nullable foreign key column results in Zod validation error


ondrej:
Hello,

Simple setup here:

* Two tables, `companies` and `people`
* No FK in `companies`
* One FK in people: nullable column `company_id`

A query is made to return a particular person with their company (however, that person's `company_id` is not set):
```
db.people.liveMany({
  where: { id: { equals: 'some-uuid' } },
  include: { companies: true },
}),
```

Expected result is that the `people` result row's `companies` field will be null — after all, the person isn't associated with a company.

Actual result is a Zod error: `Expected string, received null`. But even without Zod validation, `_findMany()` would have run `db.query()` with `i` set to `'{"where":{"id":{"in":[null]}}}'` — wouldn't have expected a query at all in this case.

msfstef:
Thank you for another great catch! It seems there's various issues with nullable foreign keys in the generated client, I'll be taking a look and increasing test coverage for them - that will include the fix of the `_findMany` call looking for null relations that you've correctly pointed out!

PR #1121, opened 2024-04-03
fix(client): Fix attempts to query for null FKs when dealing with nullable FK models

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1097

The queries were attempting to retrieve relations for null foreign keys and failing the subsequent schema validation - the suggested fix is to filter out `undefined` or `null` foreign keys when retrieving relations.

I've added two failing tests for `findMany` and `create` queries that are fixed through by this PR.


kevin-dp (approved):
Great, thanks for fixing this!

PR #1122, opened 2024-04-03
feat: support `IN` and `NOT IN` in where clauses

icehaunter:


components/electric/lib/electric/replication/eval/parser.ex
```diff
@@ -269,6 +274,35 @@ defmodule Electric.Replication.Eval.Parser do
     end
   end
 
+  defp handle_in(%PgQuery.A_Expr{name: [name]} = expr, refs, env) do
+    name = unwrap_node_string(name)
+
+    # It can only be a list here because
```
alco, line 280: Because...?

components/electric/test/electric/replication/eval/parser_test.exs
```diff
@@ -260,5 +260,38 @@ defmodule Electric.Replication.Eval.ParserTest do
 
       assert %Const{value: true, type: :bool} = result
     end
+
+    test "should work with IN clauses" do
+      env = Env.new()
+
+      assert {:ok, %Expr{eval: result}} =
+               Parser.parse_and_validate_expression(~S|1 IN (1, 2, 3)|, %{["test"] => :int4}, env)
```
alco, line 268: Unused column reference: `%{["test"] => :int4}`.

components/electric/test/electric/replication/eval/parser_test.exs
```diff
@@ -260,5 +260,38 @@ defmodule Electric.Replication.Eval.ParserTest do
...
+      env = Env.new()
+
+      assert {:ok, %Expr{eval: result}} =
+               Parser.parse_and_validate_expression(
+                 ~S|test NOT IN ('hello', 'world')|,
+                 %{["test"] => :text},
+                 env
+               )
+
+      assert %Func{type: :bool, name: "or"} = result
```
alco, line 283: I'm struggling to make sense of this assertion. The test name implies that "parse_and_validate_expression/3 should work with NOT IN clauses". But what does working mean in this case? Is the goal to verify that `NOT IN` does not cause a syntax error?

In the previous test `1 IN (1, 2, 3)` is evaluated to a boolean constant, that's more a straightforward test assertion to work with.

alco (approved):

components/electric/test/electric/replication/eval/parser_test.exs
icehaunter: Converted down to boolean

PR #1123, opened 2024-04-03
feat: support string where clauses and most of Prisma api

icehaunter:


clients/typescript/src/client/model/table.ts
```diff
@@ -1645,47 +1645,42 @@ export function liveRawQuery(
...
-  const clauses = [fieldSqlClause, andSqlClause, orSqlClause, notSqlClause]
-  return clauses.filter((clause) => clause !== '').join(' AND ')
+function makeSqlWhereClause(where: string | Record<string, any>): string {
+  if (typeof where === 'string') return where
+
+  const statements = Object.entries(where)
+    .flatMap(([key, value]) => makeFilter(value, key, 'this.'))
+    .map(interpolateArgs)
+
+  if (statements.length < 2) return statements.join('')
```
kevin-dp, line 1655: Could you document this special case where the length is `< 2`.
In theory, there's nothing to join if there is at most 1 element in the array.
I'm guessing you're "abusing" `join` here to convert the element to a string.
Let's leave a comment explaining this or let's perhaps do this more explicitly: `String(statements[0] ?? '')`.

clients/typescript/src/client/model/table.ts
```diff
@@ -1645,47 +1645,42 @@ export function liveRawQuery(
...
+  args,
+}: {
+  sql: string
+  args?: unknown[]
+}): string {
+  if (args === undefined) return sql
+
+  let matchPos = 0
+  const argsLength = args.length
+  return sql.replaceAll(/(?<!\w)\?/g, (match) =>
```
kevin-dp, line 1670: can we leave a small comment here explaining what this does

kevin-dp (approved):
Looks good!
Left some very minor comments.

PR #1124, opened 2024-04-04
feat(electric): Resume client's replication stream from the on-disk WAL window

alco:
This PR makes it possible to resume the replication stream for clients whose position in the replication stream is outside of the in-memory cached WAL window. This is done by streaming transactions directly from the replication slot. Assuming the slot's start position goes back further than the in-memory cache can fit, this effectively extends the resumable WAL window that the sync service can use before having to ask clients to reset their local state.

This implementation builds on the primitives introduced in https://github.com/electric-sql/electric/pull/1119.

### DIFF
--- a/components/electric/lib/electric/postgres/cached_wal/api.ex
+++ b/components/electric/lib/electric/postgres/cached_wal/api.ex
@@ -24,13 +24,15 @@ defmodule Electric.Postgres.CachedWal.Api do
           cache_memory_total: non_neg_integer()
         }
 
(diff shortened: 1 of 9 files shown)


alco:
Closing this since it has become evident that the total size of electric transactions contained in the on-disk WAL backlog is unlikely to exceed the size of `ELECTRIC_TXN_CACHE_SIZE`. So by replaying the full WAL backlog at startup and storing all transactions in the in-memory cache, Electric is already doing the most it can do for catching up clients tha are lagging behind.

PR #1129, opened 2024-04-05
docs: updates for shapes

icehaunter:


docs/usage/data-access/shapes.md
```diff
@@ -7,19 +7,17 @@ sidebar_position: 30
 
 Shapes are the core primitive for controlling sync in the ElectricSQL system.
 
-Local apps establish shape subscriptions. This syncs data from the cloud onto the local device using the [Satellite replication protocol](../../api/satellite.md), into the local embedded SQLite database. Once the initial data has synced, [queries](./queries.md) can run against it.
-
-The [Electric sync service](../installation/service.md) maintains shape subscriptions and streams any new data and data changes onto the local device. In this way, local devices can sync a sub-set of a larger database for interactive offline use.
+Local apps ask the server for a specific set of related data that gets synced to the local device. The central Postgres instance will always have too much data to fit on any one device, so shapes allow us to sync only a required subset of data onto the device. There is a balance between local data availability and storage usage on the device that is unique to every application, and shapes allow you to balance those properties while maintaining required guarantees.
```
KyleAMathews, line 10:
```suggestion
Local apps ask the server for a specific set of related data that gets synced to the local device. The central Postgres instance will often have too much data to fit on any one device, so shapes allow us to sync only a required subset of data onto the device. There is a balance between local data availability and storage usage on the device that is unique to every application, and shapes allow you to balance those properties while maintaining required guarantees.
```

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,151 @@ For example, this `sync` call causes a project and all its issues, their comment
...
+      },
+    },
+  },
+});
 ```
 
 ### Filter clauses
 
-Shape query clauses support:
+You can filter data at the top of the shape, or when including data over a one-to-many relation using a `where` clause on the request object. Shape `where` clauses may be either an object, or a string. They may only reference the columns on the filtered table itself. It's important that many-to-one relations cannot be filtered that way (e.g. you cannot do `comments: { include: { author: { where { id: true }}}}`), because if a
```
samwillis, line 96:
```suggestion
You can filter data at the top of the shape, or when including data over a one-to-many relation using a `where` clause on the request object. Shape `where` clauses may be either an object, or a string. They may only reference the columns on the filtered table itself.

It's important not note that many-to-one relations cannot be filtered that way (e.g. you cannot do `comments: { include: { author: { where { id: true }}}}`), because if a
```

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,151 @@ For example, this `sync` call causes a project and all its issues, their comment
...
+target row would be filtered out, that would break FK consistency on the client.
+
+#### Object `where` clause
+
+:::caution Work in progress
+TypeScript type of the object accepted in `.sync` call may be wider than what is actually supported.
```
samwillis, line 102:
```suggestion
The TypeScript type of the object accepted in `.sync` call may be wider than what is actually supported.
```

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,151 @@ For example, this `sync` call causes a project and all its issues, their comment
...
 See the <DocPageLink path="api/clients/typescript" /> docs for more details.
 
-### Permissions
+#### String `where` clause
+
+:::note Work in progress
+String `where` clause is currently only supported as top-level filtering. This will be fixed in a future release.
+:::
+
+Because the where clause will be applied on Postgres and on the ElectricSQL server, you can drop down to a string representation that's close to PostgreSQL `WHERE` clause on a regular query. You can use the usual PostgreSQL syntax, although not all functions are supported. Columns of the table you're filtering can be referenced using `this.` prefix:
```
samwillis, line 145:
```suggestion
As the where clause will be applied on Postgres and on the ElectricSQL server, you can drop down to a string representation that's close to that of a SQL `WHERE` clause on a regular query. You can use the usual PostgreSQL syntax, although not all functions are supported. The columns of the table you're filtering can be referenced using `this.` prefix:
```

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,151 @@ For example, this `sync` call causes a project and all its issues, their comment
...
+
+```ts
+await db.projects.sync({
+  where: "this.status IN ('active', 'pending')",
+});
+```
+
+This allows for more flexibility (i.e. not entire supported PostgreSQL subset can currently be expressed via a Prisma-like API), but you have to be more careful with interpolating your own values in the string in a format that PostgreSQL would accept for the expected type.
```
samwillis, line 153:
```suggestion
This allows for more flexibility as the Prisma-like does not support the all possible expressions supported by the Electric sync engine. However, you need to be more careful with interpolating your own values in the string to ensure that PostgreSQL would accept for the expected type.
```

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,151 @@ For example, this `sync` call causes a project and all its issues, their comment
...
+
+```ts
+await db.projects.sync({
+  where: "lower(this.status) ~~ 'pending%'",
+});
+```
+
+Full current list of supported Postgres functions is easiest found [in code](https://github.com/electric-sql/electric/blob/main/components/electric/lib/electric/replication/eval/env/known_functions.ex), but here's a gist:
```
samwillis, line 161: I don't like this, we should never be pointing people to source code to see a list of supported features. Even if it's verbose it should be here.

True, it's a maintain burden to keep the docs up to date, but the friction of link to source is too high.

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,151 @@ For example, this `sync` call causes a project and all its issues, their comment
...
+- String functions: concatenation, `LIKE`/`ILIKE`
 
-As a result, the data that syncs onto a users' local device for a given shape may be a **partial result** and will often be different for different users.
+:::note Work in progress
+More basic type filtering support is being added, as well as more functions over those types.
+:::
 
-See <DocPageLink path="usage/data-modelling/permissions" /> for more information.
+:::warning Limitation
+Current filtering implementation does not support non-deterministic functions. You cannot have `now()` PostgreSQL function in the where clause, because there's currently no way for us to garbage-collect rows that fall out of e.g. `this.updated_at > now()` filter without an update to the row itself.
```
samwillis, line 172:
```suggestion
The current filtering implementation does not support non-deterministic functions. For example it's not possible to use the PostgreSQL `now()` function in the where clause.
```

We don't need to try and explain why it's not supported at this point.

docs/usage/data-access/shapes.md
```diff
@@ -155,82 +201,110 @@ For example, this is OK:
...
 
 For many applications you can simply define the data you want to sync up-front, for example, at app load time and then just code against the local database once the data has synced in. For others, you can craft more dynamic partial replication, for instance, syncing data in as the user navigates through different routes or parts of the app.
 
+#### Updating shapes
+
+:::danger Potential foot-gun in development
+We're working to fix this limitation
+:::
+
+Once a subscription is established, it remains statefully in SQLite database even when you change the code. For example, doing `db.projects.sync({ where: { id: 1 }})`, starting the application, then changing the code to `db.projects.sync({ where: { id: 2 }})` will result in **2 subscriptions** established, with both projects synced to the device. We're working on lifting this limitation.
```
samwillis, line 280:
```suggestion
Once a subscription is established, it remains statefully in the local SQLite database even when you change the code. For example, doing `db.projects.sync({ where: { id: 1 }})`, starting the application, then changing the code to `db.projects.sync({ where: { id: 2 }})` will result in **2 subscriptions** established, with both projects synced to the device. We're working on lifting this limitation.
```

docs/usage/data-access/shapes.md
```diff
@@ -155,82 +201,110 @@ For example, this is OK:
...
+
+```ts
+await db.projects.sync({
+  where: { status: "active" },
+  include: { issues: true },
+});
+```
+
+If a project were to change it's status to `active`, the client would be eligible to see it, but also would be interested in the issues of this newly visible project. Due to consistency considerations, those additional rows that come from following a one-to-many relation will show up on the device slightly later than the project row itself. It's important to keep this in mind when designing the UI.
+
```
samwillis, line 294:
```suggestion
#### Move-in lag

Due to consistency considerations, when additional rows move into a shape as a result of following a one-to-many relation these will show up on the device slightly later than the parent row itself. It's important to keep this in mind when designing the UI.

For example, with this this shape:

```ts
await db.projects.sync({
  where: { status: "active" },
  include: { issues: true },
});
```

If a project were to have its status changed to `active`, the client would now be eligible to see it, along with its issues. However, the project will be synced to the local database before the issues.
```

File: docs/usage/data-access/shapes.md [hunk 10]
```diff
@@ -155,82 +201,110 @@ For example, this is OK:
...
+:::note Work in progress
+We expect to add client-side hooks notifying of such move-in events so that UI has more information to act upon.
+:::
+
+## Limitations and issues
+
+Shape-based sync is under active development, and we're aware of some issues with it. We're working on fixing the bugs and lifting limitations as we go.
+
+- `.sync` method has a wider type signature in TypeScript than what's really supported. In particular, `limit`, `sort` and other keywords under `include` should not be there.
+- `DELETE` of the top row on the client without having synced all the children may not result in a `DELETE` on the server and the row will be restored
```
samwillis, line 304: I think this is referring to cascades, I think it needs a clearer explanation of what the issue is.

docs/usage/data-access/shapes.md
```diff
@@ -155,82 +201,110 @@ For example, this is OK:
...
+- Recursive and mutually recursive tables are not supported at all for now. A FK loop will prevent the shape subscription from being established.
```
samwillis, line 305:
```suggestion
- Recursive and mutually recursive tables are not supported at all for now. A foreign key loop will prevent the shape subscription from being established.
```

docs/usage/data-access/shapes.md
icehaunter: I get that, but also that list will be so large later on... Probably deserves it's own page.

icehaunter: I'm not sure still it's worth it to do right now, and how to do that exactly. I'd rather not duplicate PG docs here either...

docs/usage/data-access/shapes.md
icehaunter: willdo

docs/usage/data-access/shapes.md
```diff
@@ -7,19 +7,17 @@ sidebar_position: 30
 
 Shapes are the core primitive for controlling sync in the ElectricSQL system.
 
-Local apps establish shape subscriptions. This syncs data from the cloud onto the local device using the [Satellite replication protocol](../../api/satellite.md), into the local embedded SQLite database. Once the initial data has synced, [queries](./queries.md) can run against it.
-
-The [Electric sync service](../installation/service.md) maintains shape subscriptions and streams any new data and data changes onto the local device. In this way, local devices can sync a sub-set of a larger database for interactive offline use.
+Local apps ask the server for a specific set of related data that gets synced to the local device. The central Postgres instance will often have too much data to fit on any one device, so shapes allow us to sync only a required subset of data onto the device. There is a balance between local data availability and storage usage on the device that is unique to every application, and shapes allow you to balance those properties while maintaining required guarantees.
```
balegas, line 10: - Local apps and local device seem a little redundant.
- I preferred more the language of syncing a subset of the larger database than saying the Postgrew will 'often' have much more data than fits on one device. Shapes are there for getting only the data the user needs into the device.

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,153 @@ For example, this `sync` call causes a project and all its issues, their comment
...
+        },
+      },
+    },
+  },
+});
 ```
 
 Once the data has synced onto the local device, it's kept in sync using a **shape subscription**. This monitors the replication stream and syncs any new data, updates and deletions onto the local device for as long as the shape's [subscription and retention semantics](#subscription-and-retention-semantics) define.
 
+### Foreign key and query consistency
```
balegas, line 47: nit: we don't need to talk following FKs to explain we fetch all data that is necessary for maintaining integrity. I'm also okay if you prefer expressing this way.

- I'd move this down and call it "Maintaining [Foreign Key] Integrity". 
- Maybe we can talk oversynching pattern

docs/usage/data-access/shapes.md
```diff
@@ -155,82 +203,111 @@ For example, this is OK:
...
-}
+  return <h1>{results.length}</h1>;
+};
 ```
 
 For many applications you can simply define the data you want to sync up-front, for example, at app load time and then just code against the local database once the data has synced in. For others, you can craft more dynamic partial replication, for instance, syncing data in as the user navigates through different routes or parts of the app.
 
+#### Updating shapes
+
+:::danger Potential foot-gun in development
```
balegas, line 278: can we not use foot-gun in public docs :)

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,153 @@ For example, this `sync` call causes a project and all its issues, their comment
...
-### Permissions
+#### String `where` clause
 
-When a user subscribes to a shape, they will only see data that they have permission to read. Any data that belongs to a shape that a user does not have permission to read is dropped from the replication stream.
+:::note Work in progress
+String `where` clause is currently only supported as top-level filtering. This will be fixed in a future release.
+:::
 
-As a result, the data that syncs onto a users' local device for a given shape may be a **partial result** and will often be different for different users.
+As the where clause will be applied on Postgres and on the ElectricSQL server, you can drop down to a string representation that's close to that of a SQL `WHERE` clause on a regular query. You can use the usual PostgreSQL syntax, although not all functions are supported. The columns of the table you're filtering can be referenced using `this.` prefix:
```
balegas, line 147: > As the where clause will be applied on Postgres and on the ElectricSQL server, you can drop down to a string representation that's close to that of a SQL `WHERE` clause on a regular query
Could just say that we can do raw where clauses.

balegas (approved):
Great pass.

docs/usage/data-access/shapes.md
thruflo: Yup, I would request reverting the deletion here. This high level context is important for building the mental model of the developer. I think the added para could happily be added after as additional context.

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,153 @@ For example, this `sync` call causes a project and all its issues, their comment
 ```js
 await db.projects.sync({
   where: {
-    id: 'abcd'
+    id: "abcd",
```
thruflo, line 29: Just on the syntax changes across these code samples:

1. single quotes are fine, no need to change to double quotes
2. trailing commas and semicolons -- we don't use these in the docs or the code samples

I'm not sure if there's a difference in editor settings but certainly when I was working on this the linting config was compatible with single quotes and the absence of commas and semicolons.

docs/usage/data-access/shapes.md
thruflo: Yes, it's great that we're clear about this but I would move this section down to the limitations section. I think we should be explicit that this means that data you may not want to end up on the client will end up on the client. I.e.: project belongs to a user, so the user gets synced, even if the user has a password field ...

Is this behaviour explicit or implicit? I.e.: if the relation is required in the shape, do we force the shape to define it, so you have to specify in the include tree? Or do we sync by magic in the background? It strikes me that explicit would be clearer, as it at least forces the developer to be aware of the behaviour.

docs/usage/data-access/shapes.md
```diff
@@ -28,103 +26,153 @@ For example, this `sync` call causes a project and all its issues, their comment
...
+      },
+    },
+  },
+});
 ```
 
 ### Filter clauses
 
-Shape query clauses support:
+You can filter data at the top of the shape, or when including data over a one-to-many relation using a `where` clause on the request object. Shape `where` clauses may be either an object, or a string. They may only reference the columns on the filtered table itself.
```
thruflo, line 96: I think this would really benefit from a code sample showing use of a where clause in an include object.

I'm not sure in this sentence what the `request` object is.

docs/usage/data-access/shapes.md
thruflo: I think it's fine myself. It's clear, there is a a foot gun here. Foot guns exist.

docs/usage/data-access/shapes.md
```diff
@@ -155,82 +203,111 @@ For example, this is OK:
...
+  include: { issues: true },
+});
+```
+
+If a project were to have its status changed to `active`, the client would now be eligible to see it, along with its issues. However, the project will be synced to the local database before the issues.
+:::note Work in progress
+We expect to add client-side hooks notifying of such move-in events so that UI has more information to act upon.
+:::
+
+## Limitations and issues
```
thruflo, line 302: I suggest moving the updating shapes and move-in lag sections into limitations. I also think an explicit "you can't currently delete a subscription" section would be valuable.

thruflo (changes requested):
Great stuff 👍 left some comments.

thruflo:
Another note on this: the shapes section in the roadmap limitations page needs to be updated. It still states that shapes are whole table only.

docs/usage/data-access/shapes.md
icehaunter: Yeah, autoformatter got me at some point

docs/reference/roadmap.md
```diff
@@ -40,16 +40,18 @@ The DDLX rules for permissions, roles, validation or local SQLite commands docum
...
+There is one temporary feature to limit data that goes on the local device: set an `electric_user_id` field on your table. If you do, then rows will only be synced if the value of that column matches the value of the authenticated user_id provided in your [auth token](../usage/auth/index.md).
+
+If you're OK with just filtering the available data, you should use [filter clauses on shapes](../usage/data-access/shapes.md#filter-clauses).
 
 This is a very temporary workaround and will be removed soon!
 :::
 
+### Shapes
+
+[Shape-based sync](../usage/data-access/shapes.md) using the [`sync()` function](../api/clients/typescript.md#sync) is currently supported with some [limitations](../usage/data-access/shapes.md#limitations-and-issues). You can ask for specific tables and rows, filter them, and follow relations. Biggest current limitation is that you cannot unsubscribe from a shape.
```
thruflo, line 53: `s/Biggest/The biggest`.

docs/usage/data-access/shapes.md
```diff
@@ -223,14 +271,58 @@ const MyComponent = () => {
...
 }
 ```
 
 For many applications you can simply define the data you want to sync up-front, for example, at app load time and then just code against the local database once the data has synced in. For others, you can craft more dynamic partial replication, for instance, syncing data in as the user navigates through different routes or parts of the app.
 
+## Limitations and issues
+
+Shape-based sync is under active development, and we're aware of some issues with it. We're working on fixing the bugs and lifting limitations as we go.
+
+- `.sync` method has a wider type signature in TypeScript than what's really supported. In particular, `limit`, `sort` and other keywords under `include` should not be there.
```
thruflo, line 284: Could hyperlink the sync method to that section of the client api doc.

docs/usage/data-access/shapes.md
```diff
@@ -223,14 +271,58 @@ const MyComponent = () => {
...
+- `DELETE` of the top row on the client without having synced all the children may not result in a `DELETE` on the server and the row will be restored
+- Recursive and mutually recursive tables are not supported at all for now. A foreign key loop will prevent the shape subscription from being established.
+- Shape unsubscribe is not available, which means any `.sync` method (in development in particular) is going to be statefully persisted regardless of code changes.
```
thruflo, line 287: "any `.sync` method" <- I think we should refer to a shape subscription, i.e.: any shape subscription established by calling `sync()` is going to be statefully ...

thruflo (approved):
Thanks -- left a couple more minor suggestions.

PR #1131, opened 2024-04-05
chore: add a more universal way to define comparison classes

icehaunter:


icehaunter:
Accidentally merged with another branch

PR #1140, opened 2024-04-10
fix: improve PG timeout handling on client connection

icehaunter:


magnetised (approved):
never wrong to add more tuples

PR #1141, opened 2024-04-10
fix(client_tests): Run the migrate.generation test as serial

davidmartos96:
There have been some instances of the CI for the client tests failing in the new migrate.generation.test.ts file.
https://github.com/electric-sql/electric/actions/runs/8630646197
https://github.com/electric-sql/electric/actions/runs/8615651764

I believe this is because there are multiple prisma generates running at the same time, which can cause issues: 
https://github.com/prisma/prisma/issues/19124#issuecomment-1753187096

This is a simple attempt to fix it, and easy to revert if it's not entirely the root of the problem.

cc @msfstef

### DIFF
--- a/clients/typescript/test/cli/migrations/migrate.generation.test.ts
+++ b/clients/typescript/test/cli/migrations/migrate.generation.test.ts
@@ -169,7 +169,10 @@ test.before(() => {
 
 // This test runs the prisma generator under the hood, which can
 // cause issues when running concurrently, so we run them with serial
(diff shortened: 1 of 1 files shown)


msfstef:
Hey - thanks for the fix!

Initially I was running them serially, then made them parallel with different folder names and I was not getting any errors but I should've suspected that the parallel executions might still cause issues.

I noticed it's failing the formatting checks so if you can just run prettier on the modified file this is good to go!

msfstef (approved):

PR #1142, opened 2024-04-10
feat(starter): Pull all examples into starter template CLI

msfstef:
Refactored the starter template to make ti more easily extensible and added `expo` and `react-native` as template options.

I'd like to also add our Ionic and Tauri examples but I'd rather do that in a separate PR and when @samwillis is back from holiday to sort out some details (like making the Ionic Electric Appointments app into a regular add/remove items example)

As part of this PR I've also refactored most examples to follow the same pattern both in terms of project structure, UI, and READMEs.

Next step would also be to add tests for the CLI and leverage it for e2e tests of our integrations as well.


examples/starter/README.md
```diff
@@ -20,7 +20,7 @@ You can optionally pass the following arguments to the `create-electric-app` com
 
 | Argument              | Value                   | Default   | Description
 |-----------------------|-------------------------|-----------|--------------
-| `--template`          | `'react' \| 'vue'`      | `'react'` | Starter template to use
+| `--template`          | `'react' \| 'vue' | 'expo' | 'react-native'`      | `'react'` | Starter template to use
```
kevin-dp, line 23: The pipe between react and vue is escaped but the other pipes you added are not.

PR #1143, opened 2024-04-10
fix(client): Don't notify of actual change if there are no changes

davidmartos96:
Just a minor fix. There could be instances where `actuallyChanged` is called with an empty array.


davidmartos96:
@msfstef Done!

msfstef (approved):

msfstef:
@davidmartos96 awesome! sorry for this but could you also add a patch changeset for this using `pnpm changeset` at the root of the project? We want to do a release today and would love to include this PR

davidmartos96:
@msfstef True, this might be interesting to include in the changelog as a fix for undesired renders/events

PR #1146, opened 2024-04-11
Updated and clarified JWT documentation

fooware:
- Added a note that AUTH_JWT_NAMESPACE does not affect the validation of the standard claims, like the `aud` check you enable with AUTH_JWT_AUD.
- ~~Fixed a typo that said `sub` was a custom claim.~~
- Updated a reference to `user_id` that did not use the newer `sub`.
- Fixed some spelling errors


fooware:
@alco , this is a simple doc change and has been open for a while now. Feel free to reject it if you don't like it, but otherwise should probably just be merged?

docs/usage/auth/secure.md
```diff
@@ -108,7 +108,7 @@ await electric.connect(token)
 
 ## How the server validates auth tokens
 
-The sync service running in secure auth mode expects the standard `iat` and `exp` claims, as well as the custom `sub` claim (formerly `user_id`) to be included in the token.
+The sync service running in secure auth mode expects the standard `iat`, `exp` and `sub` (formerly a custom `user_id`) claims to be included in the token.
```
alco, line 111: I would actually keep the old version of this paragraph for now.

There's another instance in this document, in the `AUTH_JWT_NAMESPACE` section below, where we say
> Currently, only the `sub` custom claim (formerly `user_id`) is required.

Given our incorrect treatment of `sub` as a custom claim (i.e. it is affected by the `AUTH_JWT_NAMESPACE`  setting contrary to the claim that you added in that section for clarity), I would opt for maintaining internal consistency within this doc and referring to it as a custom claim in all places.

We should then address the incorrect treatment of it in a separate PR and adjust the docs appropriately once that's done.

alco (commented):
Thanks for the typo fixes and the suggested content changes!

docs/usage/auth/secure.md
fooware:
```suggestion
The sync service running in secure auth mode expects the standard `iat` and `exp` claims, as well as the custom `sub` claim (formerly `user_id`) to be included in the token.
```

Sure, reverting.

alco (approved):

alco:
Thank you!

PR #1154, opened 2024-04-12
feat(client): Log the origin of the notified changes

davidmartos96:
I think it would be convenient to log the origin of the changes now that the changed tables are logged too.
In the process we can also remove the INFO `notify changes` log, which was superfluous because an `actuallyChanged` call is always executed after it.


msfstef (approved):
Good idea, thank you for the contribution!

Issue #616, opened 2023-11-02
Use json_each for bulk data load


gorbak25:
Instead of making batches of a dozen inserts at a time do all inserts in 1 query using a json table passed to json_each. In postgres the same can be done using unnest but sqlite doesn't support arrays :( 
https://github.com/electric-sql/electric/blob/02d2b6c69b4d1cd78b49bd38b3128f3949f3fc52/clients/typescript/src/satellite/process.ts#L458

kevin-dp:
Hey @gorbak25, thanks for your input, this could be a good solution to the problem!
We will need to test it and benchmark it a bit but we will definitely give it a try 👍

PR #1157, opened 2024-04-15
feat(client): Use `json_each` for batch insert of subscription data

msfstef:
Addresses https://github.com/electric-sql/electric/issues/616

I've benchmarked the difference for our `_applySubscriptionData` method as a whole, below you can see a small bar chart for results using better_sqlite3 in Node where the difference between the previous and new method is within +-2% of each other across all ranges (and that is inserting simple child and parent records).

image

I've also benchmarked it using wa-sqlite with our canonical items example, and at 50-100k items the initial sync with the new `json_each` approach was consistently ~10% faster. For small numbers of records the JSON serialization and deserialization might be a small overhead but with larger numbers of records I suspect we gain speed because there's fewer back and forths with sqlite.

Overall my main argument for opting for this method over the previous one is that it requires fewer magic numbers and `maxSqlParameters` configurations and is relatively easy to understand and relies more on SQLite than it does on client code.

### DIFF
--- a/.changeset/fair-ravens-chew.md
+++ b/.changeset/fair-ravens-chew.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Use `json_each` for bulk loading initial synced data
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -63,7 +63,7 @@ import {
 import { Mutex } from 'async-mutex'
 import Log from 'loglevel'
 import { generateTableTriggers } from '../migrators/triggers'
-import { prepareInsertBatchedStatements } from '../util/statements'
+import { prepareInsertJsonBatchedStatement } from '../util/statements'
 import { mergeEntries } from './merge'
 import { SubscriptionsManager, getAllTablesForShape } from './shapes'
 import { InMemorySubscriptionsManager } from './shapes/manager'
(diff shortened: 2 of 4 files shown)


msfstef:
Investigated failing e2e - since blobs cannot be trivially serialized and it would require knowledge of _what_ is being serialized and how in order to properly deserialize the appeal of the simplicity of this approach is lost to the need to introduce these kinds of checks and conversions. Similar thing for special JS number values (Infinity etc).

Delegating to the SQLite wrapper's bind parameter implementation is a safer choice.

It would still be possible to use this approach for the shadow table as it is simple and only has strings and number by definition but I'm not sure if having two methods to do batch inserts that have similar performance has any benefit, it only complicates the codebase.

If you agree @kevin-dp we can close the source issue?

samwillis (commented):
As this touches the same code as #1038 which introduces a query builder for the snapshot queries it should not be merged until after that as PG on the client is a priority right now. I've not looked at how this PR would need to change to be compatible with it, or if we can have alternative methods for both SQLite and Postgres.

It would be interesting to benchmark the same JSON based method on top of Postgres and PGlite too.

msfstef:
@samwillis if you look at my [comment above](https://github.com/electric-sql/electric/pull/1157#issuecomment-2057036107) inserting data using `json_each` requires first serializing our input and then deserializing it within the context of SQLite (or Postgres soon).

Given the minimal performance gains from using this approach (at least from my preliminary benchmarking), once we need to start handling special cases for serialization and deserialization that also depend on the SQLite version and json functions, the "simplification" benefit is kind of lost. Sticking with SQLite parameter binding I think is much safer and better to deal with the limit on number of parameters to bind.

If you agree I will close this PR and the referenced issue as well.

kevin-dp:
Hi Stefanos, i agree with you on closing this for now. It looks appealing at first, but if we have to start special casing, the simplicity benefit is lost. If we would introduce this purely for performance reasons, then - as mentioned by Sam - we will also want to benchmark it on PG and PGlite.

Issue #666, opened 2023-11-15
Expose the constructor for ElectricDatabase


gorbak25:
I've moved sqlite off the main thread using comlink for the ui in quickmix.io to be snappy during heavy db operations. I've then started playing around with the VFS and settled for AccessHandlePoolVFS. The problem is that the constructor for ElectricDatabase is private and i need to ignore the typechecker for that line of code.

Here's how i initialize the db:
```js
async function initDb(dbName: string) {
  if (_electricDb != null) {
    return;
  }

  // Initialize SQLite
  const SQLiteModule = await SQLiteESMFactory()

  // Build API objects for the module
  const sqlite3 = SQLite.Factory(SQLiteModule)

  const vfs = new AccessHandlePoolVFS(`/${dbName}`)
  await vfs.isReady;

  // Register a Virtual File System with the SQLite runtime
  sqlite3.vfs_register(vfs, true)

  // Open the DB connection
  // see: https://rhashimoto.github.io/wa-sqlite/docs/interfaces/SQLiteAPI.html#open_v2
  const db = await sqlite3.open_v2(
    dbName,
    SQLite.SQLITE_OPEN_CREATE | SQLite.SQLITE_OPEN_READWRITE)

  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
  // @ts-ignore
  _electricDb = new ElectricDatabase(dbName, sqlite3, db);
}
```
I would love to avoid using ` // @ts-ignore`

kevin-dp:
Hi @gorbak25, i'll create a PR to expose the constructor and will keep you informed 👍

kevin-dp:
In addition, we could also extend `init` with an optional argument to provide the VFS to use.

balegas:
I was talking with @samwillis today about this. One issue is that some of the VFSs don't work well across tabs -- like the `AccessHandlePoolVFS`.

An intermediate solution, would be to add a worker-based driver that is pre-configured with a tab-safe VFS (if there is any that is really so). This would help with the issues when applying queries take time, blocking the UI, but is still safe for the developer.

Exposing the constructor needs to come with a large warning banner and so far we have refrained from doing that. Exposing the constructor and having a sensible default would also be okay.

msfstef:
Closing as we are focusing on making our web support work with multiple tabs and generally making it as consistent as possible - we are now exporting our driver implementation interface so if there's a unique need you can implement a custom driver.

See https://github.com/electric-sql/electric/pull/1158 for discussion.

PR #1158, opened 2024-04-15
feat(client): Allow choosing custom `wa-sqlite` VFS

msfstef:
Addresses https://github.com/electric-sql/electric/issues/666 (issue of the beast 🤘)

Adds a secondary static initialiser to `ElectricDatabase` to provide custom VFS - we can skip documenting it or add a mention to it to the `wa-sqlite` driver docs, but it addresses the above issue.

We can add a note about not guaranteeing compatibility on that initialiser, but If we don't want to allow this at all then we should probably close the issue?

### DIFF
--- a/.changeset/short-mugs-doubt.md
+++ b/.changeset/short-mugs-doubt.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Allow choosing a VFS for `wa-sqlite` driver integration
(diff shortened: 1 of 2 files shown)


samwillis:
I'm not sure how I feel about this at this point, it's a sensible change but part of the current design work is around how we make Electric work with OPFS and multiple tabs. This change doesn't enable OPFS as wa-sqlite needs to be in a web worker, it only really enables swapping to an alternative IndexedDB or an in-memory VFS.

msfstef:
@samwillis I agree, what we provide should work out of the box and now that we've exposed custom driver implementations people that are willing to experiment with more specific configurations are free to re-implement their own driver anyway.

If you agree I can close/reject this PR and close the reference issue?

samwillis:
@msfstef Agreed, close for now

Issue #911, opened 2024-02-01
electric-sql generate `PROXY_URL` always the same


digoburigo:
When trying the tutorial for manually installing `ElectricSQL` when I try to use `npx electric-sql generate` it's always pointing to this URL `postgresql://prisma:********@localhost:65432/electric-app`.

Tried to use the `--proxy` flag on the CLI and setting the `ELECTRIC_PROXY` env and none of them work.

The only one thing that I was able to change was the database name that is linked to the `package.json` `name` conifiguration.

Is there a bug with this or maybe I'm missing something. Can someone help me with this?

using watch mode

```bash
npx electric-sql generate -w -p postgresql://postgres:proxy_password@localhost:65432/postgres
Generating Electric client...
Service URL: http://localhost:5133
Proxy URL: postgresql://prisma:********@localhost:65432/electric-app
```

kevin-dp:
Hi @digoburigo, i have encountered a similar issue and flagged it up. There is a workaround that works on my machine which is to define the environment variables in a file called `.env` in the root folder of your project (i.e. in the same directory as where your `package.json` resides). And then calling `npx electric-sql generate -w` should take into account the environment variables.

samwillis:
@digoburigo, I've just been looking into this. There are two things happening in the copied command and output (also copied below)

```
npx electric-sql generate -w -p postgresql://postgres:proxy_password@localhost:65432/postgres 
Generating Electric client...
Service URL: http://localhost:5133
Proxy URL: postgresql://prisma:********@localhost:65432/electric-app
```

1. The username is hard coded to `prisma` in the generator, as it puts the proxy it into a special introspection mode. So this is correct.

2. The database name is incorrect, and from some initial testing does seem to not be carried through correctly.

Were those the two things you noticed?

I have tested changing the hostname and port, they are correctly handled.

I need to check with @magnetised, but I believe the database name is ignored by the proxy as electric is only connecting to a single database. With those two things, I think the generator command should have worked as copied.

The bug that @kevin-dp found is related to the `start` command with the `--image` or `ELECTRIC_IMAGE` env var and I believe is unrelated.

kevin-dp:
@samwillis indeed, a quick test confirms that the host and port are correctly fetched from the argument and it always uses the project's name as the database name but as you said, i believe that the proxy ignores the database name anyway.

alco:
>  the proxy ignores the database name anyway
I can confirm that.

digoburigo:
[...]

PR #1159, opened 2024-04-15
fix(cli): Fix CLI `PROXY` option to correctly infer DB name and not print introspection URL

msfstef:
Fixes https://github.com/electric-sql/electric/issues/911

As per the title, we fallback to the proxy URL to infer the database name if the database URL is not specified, and we print the proper proxy url as the proxy url rather than the Prisma introspection one to avoid confusing users.

### DIFF
--- a/clients/typescript/src/cli/migrations/migrate.ts
+++ b/clients/typescript/src/cli/migrations/migrate.ts
@@ -88,7 +88,7 @@ export async function generate(options: GeneratorOptions) {
       }
     }
     console.log('Service URL: ' + opts.config.SERVICE)
-    console.log('Proxy URL: ' + stripPasswordFromUrl(opts.config['PROXY']))
+    console.log('Proxy URL: ' + stripPasswordFromUrl(opts.config.PROXY))
     // Generate the client
     if (opts.watch) {
       watchMigrations(opts)
--- a/clients/typescript/src/cli/migrations/migrate.ts
+++ b/clients/typescript/src/cli/migrations/migrate.ts
@@ -88,9 +88,7 @@ export async function generate(options: GeneratorOptions) {
       }
     }
     console.log('Service URL: ' + opts.config.SERVICE)
-    console.log(
-      'Proxy URL: ' + stripPasswordFromUrl(buildProxyUrl(opts.config))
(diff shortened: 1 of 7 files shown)


clients/typescript/src/cli/migrations/migrate.ts
```diff
@@ -89,8 +89,7 @@ export async function generate(options: GeneratorOptions) {
     }
     console.log('Service URL: ' + opts.config.SERVICE)
     console.log(
-      'Proxy URL: ' +
-        stripPasswordFromUrl(buildProxyUrlForIntrospection(opts.config))
+      'Proxy URL: ' + stripPasswordFromUrl(buildProxyUrl(opts.config))
```
samwillis, line 92: should this not just be `'Proxy URL: ' + opts.config.PROXY` as that's the fully built proxy url?

Then I'm not sure we need the  `buildProxyUrl`

msfstef: yep you're right

msfstef:
@samwillis also fixed the `ELECTRIC_` vars not working as CLI options:
https://github.com/electric-sql/electric/pull/1159/commits/8f6bf175fd965ba29fc7297029eba3675964c7f3

samwillis (approved):
👍

alco:
> @samwillis also fixed the `ELECTRIC_` vars not working as CLI options: [8f6bf17](https://github.com/electric-sql/electric/commit/8f6bf175fd965ba29fc7297029eba3675964c7f3)

Does it correctly handle options such as `ELECTRIC_USE_IPV6`?

msfstef:
@alco this fix was for arguments passed to the CLI directly, at least the [documented ones](https://electric-sql.com/docs/api/cli#options-2) which don't include `ELECTRIC_USE_IPV6`.

We document setting `ELECTRIC_USE_IPV6` as an environment variable which does indeed work, although I plan on adding more unit and integration tests to the CLI - I'm sure I'll capture other issues doing that, but if you're already aware of some please do let me know.

alco:
@msfstef Oh, sorry, my bad. Wasn't paying enough attention.

PR #1160, opened 2024-04-16
fix(electric): make sure migration statements preserve order

icehaunter:


components/electric/lib/electric/replication/shapes/change_processing/reduction.ex
```diff
@@ -47,14 +47,17 @@ defmodule Electric.Replication.Shapes.ChangeProcessing.Reduction do
...
           {Graph.t(), [Changes.change() | Changes.Gone.t()], subquery_actions()}
   def unwrap(reduction(graph: graph, operations: ops, gone_nodes: gone, actions: actions) = state) do
-    changes =
-      reduction(state, :passthrough_operations) ++
-        Enum.map(gone, fn {relation, pk} -> %Changes.Gone{relation: relation, pk: pk} end) ++
-        Enum.map(ops, &unwrap_operation/1)
+    operations = Enum.map(ops, &unwrap_operation/1)
+    ops_and_gone = Enum.reduce(gone, operations, &[build_gone_msg(&1) | &2])
+
+    # Side-effect here is that migration statements within a transaction are hoisted to be first changes.
```
magnetised, line 53: yikes. that feels risky. is there no way to preserve the order of the statements? under the current additive migrations only constraint i can't see it causing any problems at the moment but if it's possible to get the order right now it makes sense to do it.

magnetised (approved):
if its possible to completely preserve a txn's statement order i think we should do it. if you think that fails a cost-benefit analysis atm then it should be ok for now.

components/electric/lib/electric/replication/shapes/change_processing/reduction.ex
icehaunter: It was the same case last time you reviewed this code, I just added a comment now. I believe mixed DML/DDL transactions are an edge case in general, and I don't want to address it right now.

As you can see in the spec, ops is a map, not a list, so we lose operation ordering right now. I can work later to preserve order better

PR #1162, opened 2024-04-16
chore(e2e): Web example e2e testing

msfstef:
Added basic e2e tests for the web examples using [Playwright](https://playwright.dev/) and added a CI workflow with a matrix for the web projects. The idea is that the same workflow will also handle the mobile and desktop projects as well.

There is a certain level of duplication for these tests but I think it's better to keep them isolated per project/example as each one might have it's own specific requirements - the alternative would be to have yet another project called something like "examples-e2e" with the e2e logic (which _should_ be shared among all canonical examples) and then just checkout and run the projects and test against them.

We can sort out those details as we go but I think it's worth adding _something_ that's testing them right now.

I've taken care to remove the e2e logic from the starter templates in the `create-electric-app` CLI and used the opportunity to fix an issue with configuring ports and add more e2e tests for it (including interactivity e2e test).


.github/workflows/examples_e2e.yml
```diff
@@ -0,0 +1,43 @@
...
+        run: |
+          npm run backend:up &&
+          npm run db:migrate &&
+          npm run client:generate
+      - name: Start dev server in background
+        working-directory: ${{ matrix.example_dir }}
+        run: npm run dev < /dev/null &
+      - name: Run e2e tests
+        working-directory: ${{ matrix.example_dir }}
+        run: npm run e2e
```
icehaunter, line 43: Slightly more consise

```suggestion
    defaults:
      run:
        working-directory: ${{ matrix.example_dir }}
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-node@v4
        with:
          node-version: lts/*
      - name: Install dependencies
        run: npm ci
      - name: Start backend, run migrations, generate client
        run: |
          npm run backend:up &&
          npm run db:migrate &&
          npm run client:generate
      - name: Start dev server in background
        run: npm run dev < /dev/null &
      - name: Run e2e tests
        run: npm run e2e
```

msfstef: Ah thank you, I was looking for that config!

kevin-dp (approved):
Really amazing job you did here!

I'm just wondering about the code duplication of the playwright tests.
From what i could see, the test logic is exactly the same for both web-wa-sqlite and web-wa-sqlite-vuejs which is normal since we want these 2 example apps to look and do the same.

Could we extract the test logic into a common file and have the 2 examples call those tests? Similarly to how i did in the PG on the client PR (a file containing common test logic that exports a `test` function which is then called by PG and SQLite versions passing specific context if needed)

msfstef:
@kevin-dp I hear your point and I think I like the refactored approach more.

I moved everything into a new project under `examples/_testing` (underscored to make it clear to anyone looking at examples that it's only there out of convenience) that handles all the testing code.

Since these tests are meant to be e2e and therefore don't need to know any implementation detail, only the expected user-facing behaviour, I think it's fine to keep them separate. This should help maximise reusability as well.

kevin-dp (approved):

Issue #868, opened 2024-01-22
Listener warning


KyleAMathews:
I have a view with a lot of small components that all have a small query listening for updates to their specific object — which should be a supported use case I'd assume? Or if there is an actual limit you want to warn against, set it to that. But I'm assuming the eventemitter is just set to the default (11 I think).

<img width="685" alt="Screen Shot 2024-01-22 at 1 05 49 PM" src="https://github.com/electric-sql/electric/assets/71047/66774cd7-0c8e-408e-9afa-2cc84ee0ec38">

msfstef:
Looking at the notifier code under `src/notifiers/event.ts`, we have the following:

```ts
// Global singleton that all event notifiers use by default. Emitting an event
// on this object will notify all subscribers in the same thread. Cross thread
// notifications use the `./bridge` notifiers.
const globalEmitter = new EventEmitter()

// Increase the maximum number of listeners because multiple components
// use this same emitter instance.
globalEmitter.setMaxListeners(250)
```

This limit was increased to 250 by @thruflo int https://github.com/electric-sql/electric/pull/377 - since this is a global emitter I find it hard to think of any reasonable way to adjust this limit for a large application while still retaining the ability to detect if listeners are being accumulated or not appropriately removed. Maybe removing this warning and checking if many listeners are being added in the same "place", or having them be identified somehow could help construct a better warning.

For your particular case, would a "parent" component with view-only children work as well or would it complicate the code too much?

KyleAMathews:
Yeah I'm sure I could refactor — I think ideally though the subscriptions are efficient enough it's not necessary.

KyleAMathews:
I haven't noticed any slowdowns with the app & memory is fine — though all these listeners are to objects that are rarely changed. So what exactly is the upper limit is pretty arbitrary depending on what's updating in your app.

msfstef:
@samwillis what's your take on this? I think it makes sense to remove the warning altogether as we explicitly use a global emitter - if we want to catch leaks we can implement our own tests or mechanisms from within our `EventNotifier` to catch them.

PR #1163, opened 2024-04-16
fix(client): Remove global emitter for `EventNotifier` and remove max listener warning

msfstef:
Addresses https://github.com/electric-sql/electric/issues/868

Initially I just removed the max listener limit as the way we use it any limit we set is going to be arbitrary and just an annoyance to real use cases (like Kyle's).

Afterwards I figured it would also be pretty simple to remove the global singleton emitter pattern altogether as it wasn't quite clear why it was there, since we only ever instantiate one `EventNotifier` when we electrify and that electric instance is meant to be shared across the application instance.

Unless we have a particular reason for using a singleton I think removing it is less likely to cause issues in the future (and help with the work on being able to completely "clean up" electric resources)

### DIFF
--- a/.changeset/grumpy-years-cheer.md
+++ b/.changeset/grumpy-years-cheer.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Remove global `EventEmitter` and remove max listener warning.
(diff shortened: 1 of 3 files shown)


samwillis (approved):
Looks good to me.
I suspect the global one bates back to when there as muti-threading/tab support in the old Vaxine based Electric.

PR #1164, opened 2024-04-17
Linearlite with PGlite

samwillis:
Enables using Linearlite with PGlite or wa-sqlite by setting the `ELECTRIC_CLIENT_DB` env var to either `wa-sqlite` or `pglite`. This can be done in the `.env` file or any other standard way. It will be embedded into the build so a build will be for one or the other.

To be merged after #1038 with the `electric-sql` rep either pointing at a canary build or a published version.

### DIFF
--- a/examples/linearlite/package.json
+++ b/examples/linearlite/package.json
@@ -30,6 +30,7 @@
     "vite": "vite"
   },
   "dependencies": {
+    "@electric-sql/pglite": "^0.1.4",
     "@firefox-devtools/react-contextmenu": "^5.1.0",
     "@headlessui/react": "^1.7.17",
     "@svgr/plugin-jsx": "^8.1.0",
@@ -46,7 +47,7 @@
     "animate.css": "^4.1.1",
     "classnames": "^2.3.2",
     "dayjs": "^1.11.9",
-    "electric-sql": "^0.9.3",
+    "electric-sql": "file:../../clients/typescript/electric-sql-0.10.1.tgz",
     "fractional-indexing": "^3.2.0",
     "jsonwebtoken": "^9.0.2",
     "lodash.debounce": "^4.0.8",
@@ -62,7 +63,8 @@
     "tiptap-markdown": "^0.8.2",
     "uuid": "^9.0.0",
     "vite-plugin-svgr": "^3.2.0",
-    "wa-sqlite": "github:rhashimoto/wa-sqlite#semver:^0.9.8"
+    "wa-sqlite": "github:rhashimoto/wa-sqlite#semver:^0.9.8",
+    "zod": "^3.22.4"
   },
   "devDependencies": {
     "@databases/pg": "^5.4.1",
--- a/examples/linearlite/package.json
+++ b/examples/linearlite/package.json
@@ -47,7 +47,7 @@
     "animate.css": "^4.1.1",
     "classnames": "^2.3.2",
     "dayjs": "^1.11.9",
-    "electric-sql": "file:../../clients/typescript/electric-sql-0.10.1.tgz",
+    "electric-sql": "^0.10.2-canary.450a65b",
     "fractional-indexing": "^3.2.0",
     "jsonwebtoken": "^9.0.2",
     "lodash.debounce": "^4.0.8",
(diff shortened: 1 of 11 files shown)


examples/linearlite/package.json
```diff
@@ -46,7 +47,7 @@
     "animate.css": "^4.1.1",
     "classnames": "^2.3.2",
     "dayjs": "^1.11.9",
-    "electric-sql": "^0.9.3",
+    "electric-sql": "file:../../clients/typescript/electric-sql-0.10.1.tgz",
```
AntonOfTheWoods, line 50: This is a "special" 0.10.1, right?

samwillis: Yes, i's a local built of the client from this branch. We're nearly there, only a couple of things left to fix.

samwillis: Very very close....

examples/linearlite/src/electric.tsx
```diff
@@ -29,32 +30,70 @@ export const DEBUG = debugParam ? debugParam === 'true' : DEV_MODE || DEBUG_ENV
...
+  return {
+    electric,
+    conn,
+    config,
+  }
+}
+
+export const initElectric = async () => {
+  const electricUrl = ELECTRIC_SERVICE ?? 'ws://localhost:5133'
+  const tabId = uniqueTabId().tabId.slice(0, 8)
```
msfstef, line 75: I've noticed that `uniqueTabId` is not guaranteed to give you a unique id per tab - is there a reason we're not calling this at top level and using it as a global like the discriminator and the lib version? (react strict-mode causes multiple renders and thus sometimes multiple dbs to initialize per tab)

examples/linearlite/src/components/TopFilter.tsx
```diff
@@ -36,12 +36,18 @@ export default function ({
 
   const filteredIssuesCount = issues.length
 
-  const handleSearchInner = debounce((query: string) => {
+  // const handleSearchInner = debounce((query: string) => {
```
msfstef, line 39: nit: can remove commented out code? or if it's there fore debugging maybe add a comment (debounce to avoid crashes when using SQLite or something?)

examples/linearlite/package.json
```diff
@@ -62,7 +63,8 @@
     "tiptap-markdown": "^0.8.2",
     "uuid": "^9.0.0",
     "vite-plugin-svgr": "^3.2.0",
-    "wa-sqlite": "github:rhashimoto/wa-sqlite#semver:^0.9.8"
+    "wa-sqlite": "github:rhashimoto/wa-sqlite#semver:^0.9.8",
+    "zod": "^3.22.4"
```
msfstef, line 67: we require a specific version of zod (3.21.1) because of some limitations with our generated client - is there a specific reason this is added as a dependency?

examples/linearlite/src/electric.tsx
```diff
@@ -18,6 +17,8 @@ const DEV_MODE = import.meta.env.DEV
 const ELECTRIC_SERVICE =
   import.meta.env.ELECTRIC_SERVICE || import.meta.env.ELECTRIC_URL
 const DEBUG_ENV = import.meta.env.DEBUG
+const CLIENT_DB: 'wa-sqlite' | 'pglite' =
```
msfstef, line 20: will we eventually have an in-app toggle to switch between the two? if not, which version will we be deploying to the website?

msfstef (approved):
Looks great! Left some minor comments, main thing is the added zod dependency violates the requirement for the 3.21.1 version

examples/linearlite/package.json
samwillis: The version of Zod our client requires has issues with none v4 UUIDs. We use a none standard UUID for the imported dataset and it throws with Zod 3.21.1 (there is a bug in that versions UUID validation). Version 3.22.4 fixes the issue so that the app runs at the expense of the broken types in the client 🤷 

We need to fix the generator

examples/linearlite/src/electric.tsx
samwillis: I am expecting us to probably deploy both versions on separate URLs so that they can both be owned in the same browser at once in separate tabs.

This pattern will also work for if we add a Tauri or Capacitor wrapper around the app to specify the driver.

PR #1165, opened 2024-04-17
Integrate permissions with shapes

magnetised:
Also includes fixes for:

- Support for booleans encoded as `"true"` and `"false"` coming from the shapes associated data triggers


components/electric/lib/electric/postgres/schema/fk_graph.ex
```diff
@@ -10,40 +10,82 @@ defmodule Electric.Postgres.Schema.FkGraph do
...
-        {{name.schema, name.name}, {fk.pk_table.schema, fk.pk_table.name}, fk.fk_cols}
+    fks =
+      tables
+      |> Enum.flat_map(fn %Proto.Table{constraints: constraints, name: name} ->
+        constraints
+        |> Stream.filter(&match?(%{constraint: {:foreign, _}}, &1))
+        |> Enum.map(fn %{constraint: {:foreign, fk}} ->
+          {{name.schema, name.name}, {fk.pk_table.schema, fk.pk_table.name}, fk.fk_cols}
+        end)
       end)
```
alco, line 49: Merely a code style suggestion. I believe this is simpler to parse when reading the code:

```suggestion
      for %Proto.Table{constraints: constraints, name: name} <- tables,
          %{constraint: {:foreign, fk}} <- constraints do
        {{name.schema, name.name}, {fk.pk_table.schema, fk.pk_table.name}, fk.fk_cols}
      end
```

components/electric/lib/electric/postgres/schema/fk_graph.ex
```diff
@@ -10,40 +10,82 @@ defmodule Electric.Postgres.Schema.FkGraph do
...
+    pks =
+      tables
+      |> Stream.flat_map(fn %Proto.Table{constraints: constraints, name: name} ->
+        constraints
+        |> Stream.filter(&match?(%{constraint: {:primary, _}}, &1))
+        |> Enum.map(fn %{constraint: {:primary, pk}} ->
+          {{name.schema, name.name}, pk.keys}
+        end)
+      end)
+      |> Map.new()
```
alco, line 60:
```suggestion
      for %Proto.Table{constraints: constraints, name: name} <- tables,
          %{constraint: {:primary, pk}} <- constraints,
          into: %{} do
        {{name.schema, name.name}, pk.keys}
      end
```

components/electric/lib/electric/postgres/schema/fk_graph.ex
```diff
@@ -64,30 +123,137 @@ defmodule Electric.Postgres.Schema.FkGraph do
...
+  defp label_path([relation | path], label) do
+    [{label, relation} | label_path(path, label)]
+  end
+
+  @doc """
+  Get the foreign key path information between the `root` table and the given `relation`.
+
+  Each entry in the path is either
+  - `{:many_to_one, {from_table, foreign_key_columns}, {to_table, primary_key_columns}}`, or
+  - `{:one_to_many, {from_table, primary_key_columns}, {to_table, foreign_key_columns}}`
```
alco, line 206: This fails to mention `{:one_to_one, ...}` that is returned by the first clause of this function. Function spec does not mention it either.

alco: Hm, it appears this function is never called.

components/electric/lib/electric/satellite/permissions/read.ex
```diff
@@ -1,62 +1,204 @@
...
+  @spec filter_read_with_context(
+          perms(),
+          graph() | nil,
+          graph(),
+          change_list(),
+          xid(),
+          mapper_fun()
+        ) ::
+          {[term()], [term()], moves()}
+  defp filter_read_with_context(perms, old_graph, graph, changes, xid, mapper_fun) do
```
alco, line 98: I'd suggest renaming this to `filter_changes_with_context` to make the separation of concerns clearer: i.e. there's `filter_read_with_context()` that calls `filter_changes_with_context()` and `filter_referenced_records[_with_context` in its implementation.

Also, it's not so clear what the word "context" is referring to.

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -1147,12 +1226,23 @@ defmodule Electric.Satellite.Protocol do
     end
   end
 
-  defp serialize_unknown_relations(unknown_relations) do
+  defp shape_request_context(state) do
+    state.subscriptions
+    |> Map.values()
+    |> List.flatten()
+    |> ShapeRequest.prepare_filtering_context()
```
alco, line 1233: This function always returns an empty map. Is this a placeholder for future work?

components/electric/lib/electric/postgres/schema/fk_graph.ex
magnetised: Hmm, yeah. A left-over api call - replaced with the stuff in `Permissions.Structure` and `routes/3`. I'll remove it.

components/electric/lib/electric/satellite/permissions/structure.ex
```diff
@@ -0,0 +1,334 @@
...
+
+  def id(structure, %Changes.DeletedRecord{} = change) do
+    id(structure, change.relation, change.old_record)
+  end
+
+  @doc """
+  Returns the primary key value for the given record.
+  """
+  @spec id(t(), relation(), Graph.record()) :: Graph.id()
+  def id(%__MODULE__{pks: pks}, relation, record) do
```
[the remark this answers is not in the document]
magnetised: good idea

components/electric/lib/electric/satellite/permissions/structure.ex
```diff
@@ -0,0 +1,334 @@
...
+  def foreign_keys(%__MODULE__{} = structure, root, relation) do
+    case path(structure, root, relation) do
+      [^relation, parent | _] ->
+        [{parent, fk_cols(structure, parent, relation)}]
+
+      _ ->
+        []
+    end
+
+    # structure
```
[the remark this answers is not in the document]
magnetised: oops

components/electric/lib/electric/satellite/protocol.ex
magnetised: Think so. It's @icehaunter's shape stuff that I left -- he may have left it for me to use but if so I didn't

components/electric/lib/electric/satellite/permissions/read.ex
magnetised: have given these functions more externally relevant names

components/electric/lib/electric/postgres/schema/fk_graph.ex
magnetised: i never reach for comprehensions, but you have a point

components/electric/lib/electric/satellite/permissions/structure.ex
```diff
@@ -0,0 +1,334 @@
...
+    # └----<----┘
+    #
+    # we setup the permissions based on R:
+    #
+    # ELECTRIC ASSIGN 'owner' TO R.user_id;
+    # ELECTRIC GRANT ALL ON A TO (R, 'owner')
+    # ELECTRIC GRANT ALL ON B TO (R, 'owner')
+    #
+    # naively we can get a path from B to R just by following the foreign key on B, skipping A
+    # completely, but that does not match the permissions design, which is based on the path
```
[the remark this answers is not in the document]
magnetised: yeah. this is kinda hard to justify except its easy to break your permissions system with a simple "comment owned by author" structure, although I wonder if some of our example/test schemas are a bit odd TBH. I've added a comment re "replace this with something better"

components/electric/lib/electric/satellite/permissions/graph.ex
```diff
@@ -138,107 +69,127 @@ defmodule Electric.Satellite.Permissions.Graph do
...
-    Enum.reduce(changes, impl, &apply_change(&2, roots, &1))
+  @spec transaction_context(impl(), Structure.t(), [Changes.change()]) :: impl()
+  def transaction_context({_module, _state} = impl, structure, changes, referenced_records \\ []) do
+    referenced_records
+    |> Enum.flat_map(fn {_relation, records} ->
+      Enum.map(records, fn {_pk, %{relation: relation, record: record}} ->
+        %Changes.NewRecord{relation: relation, record: record}
+      end)
+    end)
+    |> Enum.concat(changes)
```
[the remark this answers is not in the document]
magnetised: thanks. good points

components/electric/lib/electric/satellite/protocol/state.ex
```diff
@@ -100,4 +107,11 @@ defmodule Electric.Satellite.Protocol.State do
       when status in [:active, :paused, :suspended] do
     %{state | out_rep: %OutRep{out | status: status}}
   end
+
+  @spec user_id(t()) :: Electric.Satellite.Auth.user_id() | nil
+  def user_id(%__MODULE__{auth: %{user_id: user_id}}), do: user_id
+  def user_id(_state), do: nil
+
+  @spec permissions_version(t()) :: pos_integer() | nil
+  def permissions_version(%__MODULE__{}), do: nil
```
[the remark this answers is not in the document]
magnetised: placeholder.... I will add an implementation because at some point we'll will need it

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -1026,13 +1099,28 @@ defmodule Electric.Satellite.Protocol do
...
     tx
     |> maybe_strip_migration_ddl(state.out_rep.last_migration_xid_at_initial_sync)
-    |> advance_graph_by_tx(graph, current_shapes(state), state.auth.user_id)
+    |> apply_permissions_and_shapes(graph, current_shapes(state), state.permissions)
   end
 
-  defp advance_graph_by_tx(tx, graph, shapes, user_id) do
+  defp apply_permissions_and_shapes(tx, graph, shapes, permissions) do
+    {filtered_tx, _rejected_changes, moves_out} =
+      Permissions.filter_read(permissions, Electric.Replication.ScopeGraph.impl(graph), tx)
```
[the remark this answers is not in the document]
magnetised: this is the pr where the rubber hits the road, so it's here that we go from an isolated set of code paths to integration where the permissions finally affect the running system. the feature flag to make this optional is in the following pr, #1206

components/electric/lib/electric/replication/eval/env/known_functions.ex
```diff
@@ -51,8 +51,18 @@ defmodule Electric.Replication.Eval.Env.KnownFunctions do
   defcompare "timestamp", using: &NaiveDateTime.compare/2
   defcompare "timestamptz", using: &DateTime.compare/2
 
-  defpostgres "bool = bool -> bool", delegate: &Kernel.==/2
-  defpostgres "bool <> bool -> bool", delegate: &Kernel.!=/2
+  defpostgres "bool = bool -> bool" do
+    def bool_equal(l, r) do
+      Casting.cast_bool(l) == Casting.cast_bool(r)
```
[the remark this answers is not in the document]
magnetised: yeah, i was expecting this to be "wrong". the test case that this addresses is added here: https://github.com/electric-sql/electric/pull/1165/files#diff-d6b353c07d17e4137e6db23528b5fb966f9e4760bd9469e783596f58917c4f8bR72

happy to take this out (along with the new functions in Casting) if you want to address this elsewhere

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -625,31 +630,79 @@ defmodule Electric.Satellite.Protocol do
...
+        {[change], state}
+    end)
+  end
+
+  defp update_permissions(state, attrs) do
+    Map.update!(state, :permissions, &Permissions.update!(&1, attrs))
+  end
+
+  defp after_permissions_change(state) do
+    # TODO(magnetised): updated permissions must be applied to the shapes
```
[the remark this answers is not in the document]
magnetised: this is the hook for applying perms changes to the srg etc, yes. at the moment in #1206, it throws the new reset db message.

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -625,31 +630,79 @@ defmodule Electric.Satellite.Protocol do
...
-            Logger.debug(fn -> "Global permissions updated for connection" end)
+  defp apply_migrations({changes, state}) do
+    Enum.flat_map_reduce(changes, state, fn
+      %Changes.Migration{} = migration, state ->
+        state =
+          update_permissions(%{state | schema_version: migration.version},
+            schema: migration.schema
+          )
+
+        {[migration], after_permissions_change(state)}
```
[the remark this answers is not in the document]
magnetised: oh if you insist...

components/electric/lib/electric/replication/shapes/querying.ex
```diff
@@ -292,13 +302,25 @@ defmodule Electric.Replication.Shapes.Querying do
     end)
   end
 
-  defp context_filters(%Replication.Table{} = table, context, %Layer{} = _layer) do
-    ownership_column = Ownership.id_column_name()
+  defp apply_permissions(changes, graph, filtering_context) do
+    case filtering_context do
+      %{perms: perms, xid: xid} ->
+        {accepted_changes, rejected_changes} =
+          Permissions.Read.filter_shape_data(perms, graph, changes, xid)
```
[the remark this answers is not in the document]
magnetised: since at the moment the permissions scopes rely on foreign key relations, and the shapes queries follow and fill those fk relations the graph here is complete for permissions resolution (atm at least).

at the point where permissions scopes rely on non-fk info, e.g. in a USING clause on the grants, then we'll either need that information to be passed to the shape query or will need to make cross-graph lookups.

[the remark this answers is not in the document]
magnetised: the code is tested in `test/electric/satellite/subscriptions/permissions_read_filter_test.exs` along with the rest of the integration.

components/electric/lib/electric/replication/eval/env/known_functions.ex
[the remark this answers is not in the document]
icehaunter: The test case you provided is wrong. `Runner.execute` expects cast Elixir values, so it should be `=> true`, not `=> "t"`. If you want to use `=> "t"` in the object, then you should write the expression as 
```elixir
~S|new.valid::bool = true|
|> parse_and_validate_expression!(%{["new", "valid"] => :text})
|> Runner.execute(%{["new", "valid"] => "t"})
```

If you're passing an unparsed, uncast object to `execute`, then let's change the call location to do casting. We have an explicit function for that in [`Runner.record_to_ref_values/3`](https://github.com/electric-sql/electric/blob/main/components/electric/lib/electric/replication/eval/runner.ex#L11) which calls `Env.parse_const/3`. If you don't need `this` prefixes for keys, let's rewrite that as a more generic function

components/electric/lib/electric/satellite/permissions/eval.ex
```diff
@@ -134,47 +134,55 @@ defmodule Electric.Satellite.Permissions.Eval do
...
+    {:ok, result} = execute_expr(expr_cxt, expr, values)
     result
   end
 
-  defp execute_expr(expr_cxt, op, values) do
-    expr = Map.fetch!(expr_cxt.expr, op)
+  defp values(expr, record, prefix) do
+    {:ok, record} = Runner.record_to_ref_values(expr.used_refs, record)
+    Map.new(record, &prefix_value(&1, prefix))
+  end
```
icehaunter, line 180: `Runner.execute/2` fetches values from the provided values map just based on the full key, no "additional" matching. So if the expression contains a reference to a column, it **has** to be structured as `new.column` instead of just `column`. You're validating it exactly like that by building only prefixed `table_refs` in this module -- which is great, as it avoids ambiguity, but if that's the case, the single-column clause just can't happen - neither here in `prefix_value`, nor in `Runner.record_to_ref_values/2`.

components/electric/lib/electric/satellite/permissions/read.ex
```diff
@@ -66,19 +256,104 @@ defmodule Electric.Satellite.Permissions.Read do
...
+        )
+      end)
+
+    {Map.new(accepted_changes), rejected_changes ++ rejected_dependent_changes}
+  end
+
+  @spec shape_data_mapper(shape_data_row()) :: Changes.change()
+  defp shape_data_mapper({_key, {change, _}}), do: change
+
+  @spec ensure_graph_impl(Elixir.Graph.t() | Graph.impl()) :: Graph.impl()
```
icehaunter, line 351: That's the `Elixir.Graph.t()` type?...

components/electric/lib/electric/satellite/permissions/read.ex
```diff
@@ -1,62 +1,252 @@
...
+  In both these cases we remove the offending row from the list of visible changes but also add a
+  move-out message so the client can be informed of the loss of access to data.
+  """
+
   alias Electric.Replication.Changes
   alias Electric.Satellite.Permissions
+  alias Electric.Satellite.Permissions.Eval
   alias Electric.Satellite.Permissions.MoveOut
   alias Electric.Satellite.Permissions.Role
   alias Electric.Satellite.Permissions.Graph
```
icehaunter, line 30: nitpick: Can we maybe alias as a `PermGraph` to avoid shadowing the `Graph` library leading to confusion while reading code?

components/electric/lib/electric/satellite/permissions/read.ex
```diff
@@ -1,62 +1,252 @@
...
+
+  # @spec filter_changes_with_context(
+  #         perms(),
+  #         graph() | nil,
+  #         graph(),
+  #         change_list(),
+  #         xid(),
+  #         mapper_fun()
+  #       ) ::
+  #         {[term()], [term()], moves()}
```
icehaunter, line 145:
```suggestion
```

icehaunter (approved):
Looks good with some nitpicks and a question I think warrants addressing, but that's still pretty small. I'm a bit worried about red elixir tests tho...

components/electric/lib/electric/replication/eval/runner.ex
```diff
@@ -7,15 +7,30 @@ defmodule Electric.Replication.Eval.Runner do
...
   def record_to_ref_values(used_refs, record, env \\ Env.new()) do
     used_refs
     # Keep only used refs that are pointing to current table
-    |> Enum.filter(&match?({["this", _], _}, &1))
-    |> Enum.reduce_while({:ok, %{}}, fn {["this", key] = path, type}, {:ok, acc} ->
+    |> Enum.filter(&is_row_reference?/1)
+    |> Enum.reduce_while({:ok, %{}}, fn {path, type}, {:ok, acc} ->
+      key =
+        case path do
+          [key] -> key
```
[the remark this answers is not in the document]
magnetised: I think it's useful to support prefix-less columns out of the scope of permissions, to avoid requiring every column reference to be prefixed with e.g. `this` in where clauses etc. the prefix-less version feels very natural, it's exactly how you'd do it in e.g. a select where you only need to prefix to avoid ambiguity. 

so there may be some redundancy here, but I think it's worth it

components/electric/lib/electric/satellite/permissions/read.ex
magnetised: oops.

components/electric/lib/electric/satellite/permissions/read.ex
magnetised: Not sure what you're asking here. I'm planning a bit of a rename before merging, partly because this Permissions.Graph choice was a bad one in hindsight, partly to just simplify things.

components/electric/lib/electric/satellite/permissions/read.ex
magnetised: You have a point. I've been managing this name conflict but seeing that it'll be much more confusing in 6 months time. See above re renaming.

magnetised:
> I'm a bit worried about red elixir tests tho...

The next pr in the stack fixes the test failures. Haven't bothered to port those fixes down the stack...

components/electric/lib/electric/satellite/permissions/eval.ex
magnetised: You can get non-prefixed columns if the where/if clause is written without them. I've added an explicit prefix-less test that fails if I remove the code you're talking about.

PR #1168, opened 2024-04-18
feat(electric): Try decoding the JWT key using base64 only when the config option is set

alco:
This reverts [the change](https://github.com/electric-sql/electric/commit/0deba4d79de61a31aa19515d055a2a977a8e1b4e#diff-20b7a1cb738afa77ff1e885426e33a8c92faa3cc84aa9f62fd2f67b217e6f020R102-R113) introduced in the 0.9.3 release where the configured signing key would get automatically decoded if it looked like a valid base64-encoded string. That change has caused unintended problems to arise for users of Supabase Auth since its secret keys are base64-encoded but are meant to be used as is.

This PR introduces a separate configuration option `AUTH_JWT_KEY_IS_BASE64_ENCODED`. Corresponding doc update - https://github.com/electric-sql/electric/pull/1169.

### DIFF
--- a/components/electric/test/electric/config_test.exs
+++ b/components/electric/test/electric/config_test.exs
@@ -53,6 +53,13 @@ defmodule Electric.ConfigTest do
 
       assert {{Electric.Satellite.Auth.Insecure, %{joken_config: %{}, namespace: "ns"}}, []} =
                validate_auth_config("insecure", namespace: {"AUTH_JWT_NAMESPACE", "ns"})
+
+      # Settings without values are ignored
+      assert {{Electric.Satellite.Auth.Insecure, %{joken_config: %{}, namespace: "ns"}}, []} =
+               validate_auth_config("insecure",
+                 namespace: {"AUTH_JWT_NAMESPACE", "ns"},
+                 key: {"AUTH_JWT_KEY", nil}
+               )
     end
 
     test "validates secure mode" do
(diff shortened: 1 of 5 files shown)


components/electric/test/electric/config_test.exs
```diff
@@ -72,6 +79,7 @@ defmodule Electric.ConfigTest do
                validate_auth_config("secure",
                  alg: {"AUTH_JWT_ALG", "HS256"},
                  key: {"AUTH_JWT_KEY", String.duplicate(".", 32)},
+                 key_is_base64_encoded: {"AUTH_JWT_KEY_IS_BASE64_ENCODED", false},
```
magnetised, line 82: we could just leave this explicit setting of this flag out to test the default behaviour, or add an explicit test for the default behaviour

magnetised (approved):
eventually we can just reduce the entire app to env vars. 🥇

alco:
> eventually we can just reduce the entire app to env vars. 🥇

We could already benefit from using a hierarchical configuration language like TOML. It'll become more pressing as we add more configuration options to Electric over time.

components/electric/test/electric/config_test.exs
alco: The default behaviour is to accept a string of sufficient length regardless of its encoding. This is already verified in the first assertion in this test, to the extent that the config with `key_is_base64_encoded` omitted from it is deemed valid.

The actual key decoding behaviour is tested in `Auth.SecureTest`.

PR #1172, opened 2024-04-19
fix(electric): Override display settings for any new DB connections

alco:
We already set display settings for the logical replication connection that Electric opens to Postgres, in order to ensure deterministic date-time, bytea, and float formatting. However, we don't do that for ad-hoc or pooled connections which are opened when needed, such as during the initial shape sync and when fetching additional shape data. As a result, trying to serialize initial or additional data that contains timestamps or bytea values formatted using configuration Electric doesn't expect leads to errors.

In this PR, we ensure that the same display settings get set for all DB connections Electric opens.

### DIFF
--- a/.changeset/tender-garlics-fetch.md
+++ b/.changeset/tender-garlics-fetch.md
@@ -1,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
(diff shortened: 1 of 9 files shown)


magnetised (approved):
looks great.

PR #1174, opened 2024-04-19
chore(e2e): Group e2e tests using directories

alco:


alco:
When a shell command is executed from a Lux script, the current directory is set to the `dirname` of the script file's location. So when I moved script files one directory level down, they could no longer issue `make` commands due to the `Makefile` now residing in a different directory.

I haven't found a configuration setting that could fix this globally for all tests, so I'm abandonding this effort.

PR #1175, opened 2024-04-19
Add Angular bindings

alexonaci:
Very much Work in Progress

- Added the bindings under client/typescript/frameworks with own package.json and tsconfig.json to comply with the Angular Package Format: https://angular.io/guide/angular-package-format
- Excluded the angular sources from tsup build because they require a different compilation - using the angular compiler and ng-packagr
- Added example project under examples/web-wa-sqlite-angular
- The goal is to have signal-based bindings for the electricClient and to have maybe a similar structure with the TanStack Query Angular adapter: https://tanstack.com/query/v5/docs/framework/angular/overview, https://github.com/TanStack/query/tree/main/packages/angular-query-experimental/src - although is not quite idiomatic for the Angular ecosystem (yet) to have custom inject functions (like `injectLiveQuery` or `injectConnectivityState`) they seem to fit well in the current structure and I think they would improve the developer experience. But please let me know what you think


KyleAMathews:

alexonaci:
thank you for the info, this looks very interesting. Unfortunately I didn't manage to progress more with this PR in the last months.. The new structure seems very clean and will definitely be more easy to add new client libraries, can't wait to dive into it

PR #1179, opened 2024-04-22
fix(client): Don't leave a snapshot running when stopping the Satellite process

davidmartos96:
This race condition was discovered on https://github.com/SkillDevs/electric_dart/issues/21
A unit test is running, and it can happen that electric and the database connection are closed, while a `performSnapshot` call is scheduled/already running. An error is thrown when the snapshot tries to operate with the closed database.

The easy fix would be to "wait" for the snapshot mutix after cancelling the throttle, to ensure that it has ended to gracefully stop.

Another thing discovered is that `stop` was not returning a Promise, just calling `_stop` under the hood.

### DIFF
--- a/.changeset/chilly-masks-prove.md
+++ b/.changeset/chilly-masks-prove.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Don't leave a snapshot running when stopping the Satellite process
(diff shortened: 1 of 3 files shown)


kevin-dp (approved):
Hi @davidmartos96, this is a great catch and a great fix!
Thank you for your contribution 🥇

PR #1180, opened 2024-04-22
chore(website): Update website demos

msfstef:
This updates the website demos to use electric 0.10.x

I've also added a bundler and lint/formatting to the project so it can be ran from the main electric repo without the website as well - obviously they will not be 100% in sync but it at least gives a way to run them independently (and I can write some basic e2e tests for them as well)

Main change is getting the client API up to speed (e.g. use the `connect` API).

Merging this _should_ fail the website build as it will require updating the website repo as well https://github.com/electric-sql/website/pull/124


alco:
Looks good overall!

I tried building and running this locally, outside of the website repo.

Had to uncomment dummy variables after seeing complaints about them in the browser log. On the screenshot, you can see it working but there are still two errors in the console.

Screenshot from 2024-04-22 15-35-20

---

Another weird but non-critical issue I saw was that, for some reason, the `yarn db:migrate` command gets stuck for ~45 seconds regardless of whether there are any migrations to be applied. The first time I ran it, I saw this in the terminal:
```
$ yarn db:migrate
yarn run v1.22.21
$ npx electric-sql with-config "npx pg-migrations apply --database {{ELECTRIC_PROXY}} --directory ./db/migrations"
Applying 1689167402-initial_data_model.sql
Applied 1689167402-initial_data_model.sql
1 migrations applied
Done in 50.91s.
```
and this in Electric's log
```
12:27:00.468 pid=<0.3730.0> proxy_session_id=3 [warning] Not validating user "postgres"
12:27:00.471 pid=<0.3730.0> proxy_session_id=3 [info] Initialising injector in capture mode %Electric.Postgres.Proxy.Injector.Electric{}
12:27:00.506 pid=<0.3730.0> proxy_session_id=3 [info] Upstream connection is ready to accept queries
12:27:49.395 pid=<0.3419.0> pg_producer=postgres_1 [info] Migrating version <nil> -> 1689167402
12:27:49.395 pid=<0.3419.0> pg_producer=postgres_1 [info] Applying migration 1689167402: "CREATE TABLE demos (\n    id text NOT NULL,\n    name text NOT NULL,\n    inserted_at text NOT NULL,\n    updated_at text NOT NULL,\n    electric_user_id text NOT NULL,\n    CONSTRAINT demos_pkey PRIMARY KEY (id)\n);\n\n\n"
12:27:49.395 pid=<0.3419.0> pg_producer=postgres_1 [info] CREATE TABLE demos (...)
```
Notice the time gap between `[info] Upstream connection is ready to accept queries` and the line that follows it.

The next time I executed `yarn db:migrate` again, I saw this:
```
$ yarn db:migrate
yarn run v1.22.21
$ npx electric-sql with-config "npx pg-migrations apply --database {{ELECTRIC_PROXY}} --directory ./db/migrations"
No migrations required
Done in 50.99s.
```
and only this new output in Electric log:
```
12:31:08.408 pid=<0.3739.0> proxy_session_id=7 [warning] Not validating user "postgres"
12:31:08.411 pid=<0.3739.0> proxy_session_id=7 [info] Initialising injector in capture mode %Electric.Postgres.Proxy.Injector.Electric{}
12:31:08.416 pid=<0.3739.0> proxy_session_id=7 [info] Upstream connection is ready to accept queries
```

I'm really not sure what's up with this but am not considering this a blocker for the PR.

msfstef:
@alco thank you for reviewing and testing it out!

re: dummy variables - indeed that's why for now I've just [added a note in the README](https://github.com/electric-sql/electric/pull/1180/files#diff-a9c3ec887d8ccf59f8b3c62e3db5b43d69b94755bd30adf728693849e9e26bd6R21) about them - I really didn't want to use webpack, I'll find a better workaround in a subsequent PR for easier building :)

re: other errors - the rest of the backend for the intro demos is in the website repo, I could move it into the monorepo as well and use some one-off secret keys etc for testing. Because it is not deployed you'll see some failures when it tries to reach the API.

re: migration delay - I'm experiencing it as well and I'm not experiencing something similar on other projects, I can triage it a bit more separately from this PR

alco:
> re: other errors - the rest of the backend for the intro demos is in the website repo, I could move it into the monorepo as well and use some one-off secret keys etc for testing. Because it is not deployed you'll see some failures when it tries to reach the API.

Let's not go that far. I've added a note to the readme warning people about this (pushed a new commit to your branch).
 
> re: migration delay - I'm experiencing it as well and I'm not experiencing something similar on other projects, I can triage it a bit more separately from this PR

It's only worth it investigating if we run migrations as part of the deployment step in prod and the delay appears in that environment as well.

alco:
BTW, @magnetised might have a guess as to where to look for the source of the delay.

msfstef:
> It's only worth it investigating if we run migrations as part of the deployment step in prod and the delay appears in that environment as well.

Migrations are done manually not as an automated deployment step, so not an immediate issue for this.

Will merge this so I can start the website redeployment.

PR #1183, opened 2024-04-23
fix (client): don't delay initial snapshot to next tick

kevin-dp:
This PR modifies the Satellite process' `start` method in order not to delay the initial snapshot to the next tick because that could cause the snapshot to be ran after the process was closed, e.g. if someone starts the process and immediately closes it. I also added a corresponding unit test for this.


balegas (approved):
My legacy of learn while coding is being eroded :D. Nice catch.

PR #1186, opened 2024-04-23
Add new migration message

magnetised:
Which encapsulates a migration into a single unit with a version, updated schema version and satellite migration ops, as opposed to multiple ddl_command inserts.


components/electric/lib/electric/replication/changes.ex
```diff
@@ -26,11 +26,15 @@ defmodule Electric.Replication.Changes do
   @type tag() :: String.t()
   @type pk() :: [String.t(), ...]
 
-  @type change() ::
+  @type db_change() ::
```
alco, line 29: `data_change` seems more apt to me. Migrations can also be thought of as "DB changes".

e2e/tests/02.03_partial_replication_based_on_user_id.lux
```diff
@@ -26,12 +26,14 @@
     ?$psql
 
 [shell electric]
-    # We expect to send the transaction to both satellites
+    # migration is sent for both clients
+    ?client_id=client_1_1 .+ user_id=1 \[debug\] trans:(.*)%Electric.Replication.Changes.Migration
```
alco, line 30: Is the order between this and the following log entry deterministic? I would think either client could take precedence, at random.

```suggestion
    ?+client_id=client_1_1 .+ user_id=1 \[debug\] trans:(.*)%Electric.Replication.Changes.Migration
```

components/electric/lib/electric/postgres/migration.ex
```diff
@@ -0,0 +1,223 @@
...
+  end
+
+  @doc false
+  @spec to_ops([String.t()], SchemaLoader.Version.t()) ::
+          {Changes.Migration.Ops.t(), [Electric.Postgres.relation()]}
+  def to_ops(stmts, schema_version) do
+    stmts
+    |> Enum.reduce({%Changes.Migration.Ops{}, MapSet.new()}, fn stmt, {ops, relations} ->
+      Changes.Migration.dialects()
+      |> Enum.reduce(ops, fn {key, dialect}, ops ->
```
alco, line 58: I find it odd to produce ops in both dialects. How does this mesh with the `ddl` field of `Changes.Migration` which is using one dialect? The approach we've played with so far is to let the client specify which dialect it wants to use at connection time and use that for the lifetime of the connection.

magnetised: the idea is to produce both versions here then sending the appropriate dialect to the client is just a matter of picking the right version based on the client's config. it moves the dialect conversion out of the client connection process so we do it once per migration, not once per migration per client.

the ddl field is the original ddl of the migration, nothing to do with what's sent to the clients.

alco: >  it moves the dialect conversion out of the client connection process so we do it once per migration, not once per migration per client.

Great point. Didn't think of that myself.

alco (approved):

PR #1187, opened 2024-04-23
chore (client): add Zod and Prisma as (optional) peer dependencies

kevin-dp:
Every Electric app needs Zod in order to use the generated client. However, in recent versions of Zod something changed that breaks the types in the generated client (cf. https://github.com/electric-sql/electric/pull/700). Therefore, Electric apps should always use Zod version 3.21.1. Added this as a peer dependency such that npm complains if the app installs a different version. Similarly, added an optional peer dependency for Prisma v4 such that it complains if the app installs another version of Prisma.

### DIFF
--- a/clients/typescript/package.json
+++ b/clients/typescript/package.json
@@ -258,12 +258,14 @@
     "@op-engineering/op-sqlite": ">= 2.0.16",
     "@tauri-apps/plugin-sql": "2.0.0-alpha.5",
     "expo-sqlite": ">= 13.0.0",
+    "prisma": "^4.0.0",
     "react": ">= 16.8.0",
     "react-dom": ">= 16.8.0",
     "react-native": ">= 0.68.0",
     "typeorm": ">=0.3.0",
     "vue": ">=3.0.0",
-    "wa-sqlite": "rhashimoto/wa-sqlite#semver:^0.9.8"
+    "wa-sqlite": "rhashimoto/wa-sqlite#semver:^0.9.8",
+    "zod": "3.21.1"
   },
   "peerDependenciesMeta": {
     "@op-engineering/op-sqlite": {
@@ -276,6 +278,9 @@
       "optional": true
     },
     "expo-sqlite": {
+      "optional": true
+    },
+    "prisma": {
       "optional": true
     },
     "react": {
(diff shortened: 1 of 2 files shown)


samwillis (changes requested):
We can pin Prisma, but not Zod. There are valid reasons to use a different version of Zod, one being that at the moment the pinned version of Zod does not handle UUIDv7. So it's up to the user to decide if they want to trade type errors for UUIDv7 (and other UUID format) support.

kevin-dp:
> We can pin Prisma, but not Zod. There are valid reasons to use a different version of Zod, one being that at the moment the pinned version of Zod does not handle UUIDv7. So it's up to the user to decide if they want to trade type errors for UUIDv7 (and other UUID format) support.

@samwillis I think most projects will want to use the version of Zod required by Electric, and only a small number of projects will want to use another version of Zod, e.g. for UUIDv7 support. Therefore, i like pinning Zod as this informs developpers automatically if they have a wrong version of Zod in their project. For the occasional case where one really needs another version of Zod, they can use `npm install --force` to force their version of Zod, right?

samwillis (approved):
@kevin-dp agreed, lest merge this and revisit in future.

clients/typescript/package.json
```diff
@@ -258,12 +258,14 @@
     "@op-engineering/op-sqlite": ">= 2.0.16",
     "@tauri-apps/plugin-sql": "2.0.0-alpha.5",
     "expo-sqlite": ">= 13.0.0",
+    "prisma": "^4.0.0",
```
msfstef, line 261: I think this should be 4.8.1, as I'm pretty sure 4.12 or something like that doesn't work and already breaks the generation.

msfstef (approved):

msfstef:
Looks good and agree with adding them as peerDeps - although I'm pretty sure Prisma also needs to be _exactly_ 4.8.1 because I remember trying to update it to the latest v4 version when I looked at the generation and there were already issues with minor version updates.

I've also opened a PR to add some other missing optional flags for existing peer deps which are causing issues in projects:
https://github.com/electric-sql/electric/pull/1205

PR #1188, opened 2024-04-23
chore(electric): Load and clear client reconnection info per client

alco:
This is a follow-up to https://github.com/electric-sql/electric/pull/1116, specifically addressing [Ilia's request](https://github.com/electric-sql/electric/pull/1116#discussion_r1568646066).

Client reconnection info is now reloaded from the database at client connection time. This way we can ensure consistency between the data cached in ETS and data stored in the database.

### DIFF
--- a/components/electric/lib/electric/satellite/client_reconnection_info.ex
+++ b/components/electric/lib/electric/satellite/client_reconnection_info.ex
@@ -198,18 +198,6 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
            source_txns :: [non_neg_integer(), ...]}
   @type additional_data_row :: additional_data_sub_row() | additional_data_txn_row()
 
-  # How long to wait before loading client reconnection info from the database.
-  #
-  # A new debounce timer starts when a client asks to load its reconnection info and there's no
-  # timer currently running. Any other clients that ask to load their info within
-  # the debounce duration will be added to the batch query which is executed when the timer
-  # fires.
-  @debounce_duration 1_000
-
-  if Mix.env() == :test do
-    @debounce_duration 1
-  end
-
   def start_link(connector_config) do
     origin = Connectors.origin(connector_config)
     GenServer.start_link(__MODULE__, connector_config, name: name(origin))
@@ -281,6 +269,12 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
     end)
 
     :ok
+  rescue
+    exception ->
+      # Clear in-memory cache for the client to force its reloading from the database when the
+      # client reconnects.
+      clear_all_ets_data(client_id)
+      reraise exception, __STACKTRACE__
   end
 
   @doc """
@@ -290,7 +284,7 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
   def store_initial_checkpoint!(origin, client_id, wal_pos, sent_rows_graph) do
     Client.pooled_transaction(origin, fn ->
       :ok = clear_all_data!(origin, client_id)
-      store_client_checkpoint(client_id, wal_pos, sent_rows_graph)
+      :ok = store_client_checkpoint(client_id, wal_pos, sent_rows_graph)
     end)
   end
 
@@ -307,9 +301,12 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
   """
 
   defp store_client_checkpoint(client_id, wal_pos, sent_rows_graph) do
-    :ets.insert(@checkpoint_ets, {client_id, wal_pos, sent_rows_graph})
     sent_rows_graph_bin = :erlang.term_to_binary(sent_rows_graph)
     Client.query!(@upsert_checkpoint_query, [client_id, wal_pos, sent_rows_graph_bin])
+
+    :ets.insert(@checkpoint_ets, {client_id, wal_pos, sent_rows_graph})
+
+    :ok
   end
 
   @doc """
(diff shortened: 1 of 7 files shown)


components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -55,6 +55,11 @@ defmodule Electric.Satellite.Protocol do
     with {:ok, auth} <- Electric.Satellite.Auth.validate_token(token, state.auth_provider),
          true <- Electric.safe_reg(reg_name, 1000),
          :ok <- ClientManager.register_client(client_id, reg_name, state.origin) do
+      # Client reconnection info must be loaded before this WebsocketServer process can proceed
+      # to ANY other actions but we don't want to delay the auth response. Hence the async
+      # loading.
+      send(self(), :load_client_reconnection_info)
+
```
icehaunter, line 62: This `send` kinda doesn't guarantee that: the `:subscribe` message from `SatelliteCollectorConsumer` stage (started from `ClientManager.register_client/3` call chain) hasn't arrived already. After the `SatAuth` RPC call only the `SatInStartReplicationReq` call is allowed to come from the client (let's put aside replication data coming from the client to the server for a moment). Handler for that call is essentially the first place to use the cache -- why not fill it there? Technically the client may send `SatAuthReq` and `SatInStartReplicationReq` in the same frame (ours doesn't, but, you know) - it would be perfectly valid order of operations if `SatAuthReq` passes successfully and we move on to handle replication request with a new, authorized state. I'd rather load the cache in the `SatInStartReplicationReq` handler, because the cache is wrong at the time of loading, until you advance it to the point of client reconnection LSN

icehaunter (changes requested):
I like the approach, thanks for addressing these concerns! I do have two discussion points that I think warrant solving before merge

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -771,74 +804,143 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
+    # `GenServer.call()` to return.
+    #
+    # The purpose of this is to avoid issuing a separate SQL query per client. We'll need to
+    # benchmark this to tweak the duration for optimal batching.
+    state =
+      state
+      |> Map.update!(:debounced_client_ids, &[client_id | &1])
+      |> Map.update!(:waiting_callers, &[from | &1])
+      |> Map.update!(:client_monitors, &Map.put(&1, monitor, client_id))
+      |> debounce()
```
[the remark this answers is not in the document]
alco: > a possible optimization here to not load cache from DB at all if it's already in memory, and clear it only if one of the writing functions in this module failed

I like this approach. Can't remember why I decided not to pursue it but your comment has reaffirmed that it is a viable option: load data for all clients at startup, then reload on a per-client basis if any of that client's writes to replication info fails.

To be honest, 1 second delay does not seem long to me, given the fact that the load is happening in background. So if a client is on a spotty network, loading of its reconnection info might already be in progress by the time it sends a startReplication() message, provided we start loading the data early on after the client has connected.

It's all about trade-offs. We'll need to keep an eye on these data going forward to be able to make an informed decision with regards to whether it's OK to keep it in main memory for all clients (if graphs don't grow too large) or whether it's better to load and unload it to only maintain in-memory cache for connected clients.

components/electric/lib/electric/satellite/protocol.ex
alco: > Handler for that call is essentially the first place to use the cache -- why not fill it there?
Agreed.

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -504,7 +504,11 @@ defmodule Electric.Satellite.Protocol do
        ) do
     # This particular client is connecting for the first time, so it needs to perform
     # the initial sync before we start streaming any changes to it.
-    #
+
+    # Make sure we don't have any leftover cached info for this client in case its local data
+    # was cleared without the server knowing about it.
+    ClientReconnectionInfo.clear_all_data!(state.origin, state.client_id)
```
icehaunter, line 510: Why do we need to add this here, given `:perform_initial_sync_and_subscribe` handler will call `ClientReconnectionInfo.store_initial_checkpoint!/4` that will clear all this data atomically anyway?

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -1130,11 +1135,31 @@ defmodule Electric.Satellite.Protocol do
...
   defp restore_client_state(subscription_ids, observed_txn_data, lsn, %State{} = state) do
+    :ok = ClientReconnectionInfo.restore_cache_for_client(state.origin, state.client_id)
+    Logger.debug("Successfully loaded client reconnection info")
+
     with {:ok, state} <- restore_subscriptions(subscription_ids, state) do
       restore_graph(lsn, observed_txn_data, state)
     end
   end
 
+  defp restore_subscriptions(subscription_ids, %State{} = state) do
```
icehaunter, line 1146: nitpicky question: what's with function reordering here? Makes it harder to compare diffs

icehaunter (approved):
LGTM with one nitpick to remove a piece of code

components/electric/lib/electric/satellite/protocol.ex
alco: I simply moved this function definition up because it is called first in `restore_client_state()`, and `restore_graph()` is called after it.

components/electric/lib/electric/satellite/protocol.ex
alco: It's a fair point. I added this call because it looked like a logical thing to do for this module if you don't take into account that `ClientReconnectionInfo.store_initial_checkpoint!()` calls it internally.

I know that one is required to know what happens in all three modules--WebsocketServer, Protocol, ClientReconnectionInfo--in order to be able to make sense of the data flow for any given client, but I would still argue that having this call here does more good than harm since it's handling an edge case. At some point some implementation detail may change elsewhere that could affect the part of logic defined here, in Protocol. Better be prepared for that than sorry.

PR #1189, opened 2024-04-23
Fix docker instructions in auth docs

neonfuz:
The instructions in the docs under auth for running in docker incorrectly use the image name `electric-sql/electric` rather than the actual image name `electricsql/electric`. This PR fixes this.


msfstef (approved):
Good catch, and thank you for submitting a fix!

PR #1190, opened 2024-04-24
fix(client): handle null values in BigInt columns

kevin-dp:
This PR fixes a bug where the Satellite process is trying to parse `null` values into a `BigInt`, as reported [on Discord](https://discord.com/channels/933657521581858818/1229458801799729233/1229458801799729233). I've extended a unit test to check BigInt values and also extended the e2e test for integers to check this corner case. The e2e test would reproduce the bug and confirms that it is solved by this fix.

### DIFF
--- a/clients/typescript/src/util/types.ts
+++ b/clients/typescript/src/util/types.ts
@@ -181,7 +181,7 @@ export function isDataChange(change: Change): change is DataChange {
 }
 
 export type Record = {
-  [key: string]: string | number | Uint8Array | undefined | null
+  [key: string]: string | number | BigInt | Uint8Array | undefined | null
 }
 
 export type Replication<TransactionType> = {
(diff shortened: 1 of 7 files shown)


clients/typescript/src/util/types.ts
```diff
@@ -181,7 +181,7 @@ export function isDataChange(change: Change): change is DataChange {
 }
 
 export type Record = {
-  [key: string]: string | number | Uint8Array | undefined | null
+  [key: string]: string | number | bigint | Uint8Array | undefined | null
```
msfstef, line 184: is this necessary? I thought `Record` only contained potential serialised values ready to be fed to/returned by SQLite adapters, and bigint is always going to be serialized to a string, no?

msfstef (approved):
Good catch and nice work!

icehaunter (approved):

clients/typescript/src/util/types.ts
kevin-dp: You're totally right! That is not possible with SQLite, only once we add PG but this is beyond the scope of this PR. Reverted these changes.

PR #1192, opened 2024-04-24
fix(client): Reintroduce global emitter

msfstef:
See https://github.com/electric-sql/electric/pull/1191 for description of issues that arose from removing the global emitter.

I'm opening this PR in the case we decide that we want to reintroduce it as it ensures that all instantiated electric clients will share notifications between them.

### DIFF
--- a/.changeset/grumpy-years-cheer.md
+++ b/.changeset/grumpy-years-cheer.md
@@ -2,4 +2,4 @@
 "electric-sql": patch
 ---
 
-Remove global `EventEmitter` and remove max listener warning.
+Remove max listener warning on `EventNotifier`'s event emitter.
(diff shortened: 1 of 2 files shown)


samwillis:
I would lean towards this revert rather than a larger change at this time. When we add multi-tab support the event miter architecture is going to change.

msfstef:
@samwillis I think I agree since we plan on overhauling it soon then might as well keep this global - ultimately the issues arise from the fact that we use a global registry as well, the handling of multiple electrifications and disposing of resources needs a bit more care.

kevin-dp:
These are my 2 cents after a discusion on this topic with Stefanos. I don't see why we want to initialize a DB connection and electrify it inside a UI component, every time that component is mounted/unmounted the DB has to be opened/closed. Setting up a DB connection doesn't seem like a UI concern to me so I would rather move it outside of the component and pass it through React Context into my components.

kevin-dp (approved):

PR #1194, opened 2024-04-24
feat: support `IS`/`IS NOT` operators in where clauses

icehaunter:


components/electric/test/electric/replication/eval/parser_test.exs
```diff
@@ -368,5 +368,41 @@ defmodule Electric.Replication.Eval.ParserTest do
...
+      assert {:ok, %Expr{eval: result}} =
+               Parser.parse_and_validate_expression(~S|'true' IS TRUE|, %{}, env)
+
+      assert %Const{value: true, type: :bool} = result
+
+      assert {:ok, %Expr{eval: result}} =
+               Parser.parse_and_validate_expression(~S|false IS NOT FALSE|, %{}, env)
+
+      assert %Const{value: false, type: :bool} = result
+
```
alco, line 403: It seems fitting to add a test involving `null` as the argument to demonstrate that it's not affected by the type checking done for non-null values.
```suggestion

      assert {:ok, %Expr{eval: result}} =
               Parser.parse_and_validate_expression(~S|null IS NOT FALSE|, %{}, env)

      assert %Const{value: true, type: :bool} = result
      
      assert {:ok, %Expr{eval: result}} =
               Parser.parse_and_validate_expression(~S|null IS TRUE|, %{}, env)

      assert %Const{value: false, type: :bool} = result
```

.changeset/violet-kiwis-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
+shapes: support `IN`/`NOT IN` operators in where clauses
```
alco, line 5:
```suggestion
shapes: support `IS NULL`, `IS NOT NULL`, `IS TRUE/FALSE` and `IS NOT TRUE/FALSE` operators in where clauses
```

alco (approved):
Brilliant!

PR #1196, opened 2024-04-24
fix: make sure PKs in any positions work with compensations

icehaunter:
Original bug was that values from PKs were put in different columns. Reason for that bug is sneaky: I used a `%ROWTYPE` variable to store results of a `SELECT`, but `SELECT` selected PKs of the table first, and then all other columns. I assumed that `SELECT ... INTO` would fill the variable based on the column names, given that they match column names on the `%ROWTYPE` variable, but instead it used a column order to fill the fields of the variable, so PK values got into first columns of the table, and all columns up to PK were "moved" one over.

The solution here is to use a generic `RECORD` variable instead, because we're referring to columns by name later, and the way `SELECT` is built, it guarantees that every column of the original table is present.

### DIFF
--- a/e2e/tests/03.26_node_pk_position_does_not_matter_for_compensations.lux
+++ b/e2e/tests/03.26_node_pk_position_does_not_matter_for_compensations.lux
@@ -75,8 +75,8 @@
 
 [shell pg_1]
     # and expect the previously deleted row to be present
+    [invoke wait-for "SELECT id, content FROM public.items;" "00000000-0000-0000-0000-000000000001 \| hello world" 10 $psql]
     !\x
-    [invoke wait-for "SELECT content FROM public.items;" "content \| hello world" 10 $psql]
     [invoke wait-for "SELECT COUNT(*) FROM public.other_items;" "count \| 1" 10 $psql]
 
 [shell satellite_1]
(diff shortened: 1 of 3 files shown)


e2e/tests/03.26_node_pk_position_does_not_matter_for_compensations.lux
```diff
@@ -0,0 +1,88 @@
...
+    ?$node
+
+
+[shell satellite_1]
+    # Reconnect the client, expecting no errors to show up
+    !db.connect()
+    ??[proto] send: #SatOpLog
+    ??[proto] recv: #SatOpLog
+
+[shell pg_1]
```
alco, line 76: These assertions at the end of the test do not thoroughly check that data in other columns is not written over. Would you say that it's redundant to also check the content of the `id` column here and in the satellite shell below?

icehaunter: Good idea, I'll add that

fooware:
The result of the bug was that when you added rows in table A that points to rows in table B, then Electric would corrupt the related rows in table B by overwriting the first column with the value of the primary key. Details and screenshots are available in [this discord thread.](https://discord.com/channels/933657521581858818/1232013525031977052/1232013525031977052)

magnetised (approved):
👍🏻

fooware:
I can confirm that this patch removes the `electric_client_id` column corruption we had when that was the first column in our tables. Tested using latest canary server + client. Thanks for the prompt fix @icehaunter. ❤️

PR #1197, opened 2024-04-24
fix: filter the server compensations on `electric_user_id`

icehaunter:


fooware:
I can confirm that this patch removes the tenant isolation leak we had when running the latest canary server + client. Thanks for the prompt fix @icehaunter. ❤️

PR #1200, opened 2024-04-25
fix(starter): Avoid using `npx` in scripts

msfstef:
Removes `npx` from the example scripts as npx [will fetch a remote package](https://docs.npmjs.com/cli/v8/commands/npx) by default if a local one is not found.

There was an issue raised by someone who was running `npx pg-migrations blah blah` without the appropriate @databases/pg-migrations package installed which led to `npx` downloading the wrong `pg-migrations` package leading to confusing red herring errors.

This way it ensures that scripts require the dependencies to be installed to run, and that they run with the specified versions.


alco:
Interesting!

From my recent tests I came out with the conclusion that `npx` wouldn't fetch anything from the Internet silently. See https://discord.com/channels/933657521581858818/933657523049885698/1231877570312540170.

On the other hand, I also didn't know the use of  `npx` in `package.json` scripts was redundant. LGTM!

msfstef:
@alco you're right about it not fetching anything _silently_, but it might still confuse users running it expecting it to work and accepting any packages `npx` suggests downloading

See this discussion that made me think that it's probably better to avoid it altogether:
https://discord.com/channels/933657521581858818/1232938401309659166

PR #1201, opened 2024-04-25
feat(examples): Adding E2E coverage for website intro demos

msfstef:
Adds playwright e2e suite for testing the intro demos, similar to the website e2e's but the duplication is necessary due to 1) the submodule approach and 2) the different build settings and deployments (the website tests e2e agains the actual deployed netlify build)


kevin-dp (approved):
All good 👍

PR #1202, opened 2024-04-26
chore(e2e): Move e2e test to 03 group

davidmartos96:
Should this test belong to the 03 group?
For the Dart client we are using all the 03.x tests, but we noticed this is the only test that imports the satellite node macros and it's not in the 03 group.

cc @alco


davidmartos96:
@alco It's a minor thing. I started the discussion mainly for convenience over in the Dart client port, where we use this command to run the e2e. We use the same 03 group Lux logic with a real Dart client.

image

alco:
Alright, sounds good. But please use the `03.26` as the prefix in the new name since we have renamed a few e2e tests in the upcoming https://github.com/electric-sql/electric/pull/1038 PR. We can merge this after that PR has been merged.

alco:
Nevermind!

alco (approved):

PR #1204, opened 2024-04-29
fix(client): Fix text encoding polyfill being fed numbers

msfstef:
Addressing issue [raised in discord](https://discord.com/channels/933657521581858818/1231997358133477477)

The `TextEncoderLite` polyfill for the TextEncoder does not coerce its input to a string, and we force cast numbers with `as string` in the encoding stage, resulting in the input being encoded as an empty bytearray.

This does mean that the polyfill does not have the _exact_ same behaviour as the original (which it should), but the immediate solution is to respect types and not force cast to a string when it isn't.

### DIFF
--- a/clients/typescript/src/satellite/client.ts
+++ b/clients/typescript/src/satellite/client.ts
@@ -1410,6 +1410,9 @@ function serializeColumnData(
     case PgBasicType.PG_BYTEA:
       return columnValue as Uint8Array
     default:
+      if (typeof columnValue === 'number') {
+        return typeEncoder.text('' + columnValue)
+      }
       return typeEncoder.text(columnValue as string)
   }
 }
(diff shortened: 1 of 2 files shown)


clients/typescript/src/satellite/client.ts
```diff
@@ -1410,7 +1410,9 @@ function serializeColumnData(
     case PgBasicType.PG_BYTEA:
       return columnValue as Uint8Array
     default:
-      return typeEncoder.text(columnValue as string)
+      return typeEncoder.text(
+        typeof columnValue === 'string' ? columnValue : '' + columnValue
```
kevin-dp, line 1414: We don't need the ternary since `'' + columnValue` won't have any impact if the value is already a string. I'd also prefer a more explicit conversion: `String(columnValue)`.

kevin-dp (approved):

clients/typescript/src/satellite/client.ts
msfstef: my thought was that if it is already a string we can avoid either the cost of concatenation or of re-instantiation, but that might be unnecessary optimization

re `String(columnValue)` - agreed I'll switch to that

PR #1205, opened 2024-04-29
chore(client): Add missing optional flag for peer dpeendencies

msfstef:
Adds peer dependency optional flags that are missing (also reordered to match peer dep order).

Relevant PR: https://github.com/electric-sql/electric/pull/1187


samwillis (approved):
👍

PR #1207, opened 2024-04-29
chore(starter): Test CLI more thoroughly for dep installation failure

msfstef:
We had an issue with some peer deps not being marked as optional that started causing the npm install to fail (see https://github.com/electric-sql/electric/pull/1205).

Despite that because the starter cli does not return a non zero exit code if it fails at the dep installation phase, some tests were passing - I've modified them to check explicitly that the dep installation error is not shown.

Until we upgrade the electric-sql package with the optional peer dep flag, I've added the `--legacy-peer-deps` to `npm install` calls to work around this issue. I will roll this back immediately after release after confirming that the individual examples work correctly as well.

### DIFF
--- a/.changeset/lazy-rocks-switch.md
+++ b/.changeset/lazy-rocks-switch.md
@@ -1,0 +1,5 @@
+---
+"create-electric-app": patch
+---
+
+Install with `--legacy-peer-deps` to circumvent temporary issue with `electric-sql` package.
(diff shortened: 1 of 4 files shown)


kevin-dp (approved):
Good catch, i had ran into this issue too.
I agree with the workaround for now which we can rollback when publishing the next release that contains the fix for the optional peer dependencies.

msfstef:
FWIW turns out the e2e started failing because React released a new minor version (18.3.0) and suddenly the peer deps started not matching, whereas before they matched but coincidentally.

PR #1208, opened 2024-04-30
fix(client): Fix duplicate rollback issue in adapter

msfstef:
Fixes:
1) Duplicate rollbacks in interactive transactions, as a failing statement within a transaction would first trigger a rollback from within the `Transaction` object and then trigger a second one from the adapter's `transaction` API (see [discord chat](https://discord.com/channels/933657521581858818/1224163932630159461))
2) Interleaving of statements before a rollback occurs, as a failing statement within an interactive transaction would release the adapter's lock _before_ running `ROLLBACK` and thus lead to interleaving of statements run.

I've added failing tests for both cases and subsequently a fix for both.

I've kept multiple lock `release` calls to maintain the performance optimization of calling `release` _immediately_ after a commit is successful to avoid another event loop cycle blocking other adapter tasks, but if we believe that is a minimal improvement we could just keep the `release` call in the `finally` clause for clarity.

### DIFF
--- a/.changeset/strange-rabbits-compete.md
+++ b/.changeset/strange-rabbits-compete.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
+Fix duplicate `ROLLBACK`s when using interactive transactions through the adapter's `transaction` API.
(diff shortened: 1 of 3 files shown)


kevin-dp (approved):
Great work! 🥇

PR #1210, opened 2024-04-30
Change the default value of DATABASE_USE_IPV6 to "false"

alco:
The motivation for the original decision to set DATABASE_USE_IPV6 to true by default was enabling seamless interop with Fly Postgres which only supports IPv6 connections on Fly's private network.

However, at the time when I made the change, I was ignorant about the real state of IPv6 availability at different levels of Internet traffic routing. My vision of "try IPv6 and fall back to IPv4 if that fails" hasn't panned out due to the simple fact that a successful DNS query for an IPv6 address does not imply the ability to open a TCP connection to that address.

Given our recommendation to use Docker as a runtime environment for Electric sync service both during development and when deploying to production, using IPv6 by default adds a hurdle on the learning path, forcing developers to manually configure their Docker daemon with IPv6 networking support. Even then, it doesn't always resolve the problem, for example, if the developer's home/work ISP does cannot route IPv6 traffic.

Given all of the above, it makes sense for Electric to be more conservative and stick to IPv4 by default when using IPv6 may result in a broken service due to no fault of the developer.

Corresponding doc update - https://github.com/electric-sql/electric/pull/1211.


magnetised (approved):
excellent pr description. thanks!

PR #1211, opened 2024-04-30
chore(docs): Update the documentation of DATABASE_USE_IPv6

alco:
Doc update for https://github.com/electric-sql/electric/pull/1210.


magnetised (commented):
feels a bit overly wordy. something as simple as "set to true to force electric to connect to the database over IPv6..." would be clearer IMHO.

alco:
@magnetised Thanks, that's good critique. I've made another pass.

magnetised (approved):
that's perfect.

PR #1213, opened 2024-04-30
Protocol message to reset client state

magnetised:


icehaunter (approved):

clients/typescript/src/satellite/process.ts
```diff
@@ -1415,6 +1417,15 @@ export class SatelliteProcess implements Satellite {
     ])
   }
 
+  async _handleCommand(cmd: SatClientCommand) {
+    if (cmd.resetDatabase) {
+      await this._resetClientState({
+        keepSubscribedShapes: true,
+      })
+      this._subscribePreviousShapeRequests()
```
alco, line 1425: Just an FYI: this working is contingent on the server behaviour. If the server does not remove client subscriptions from the ClientReconnectionInfo and from the WebsocketServer process' state, it will respond with errors to client's attempt to subscribe using a known subscription ID.

alco (approved):

clients/typescript/src/satellite/process.ts
magnetised: Thanks @alco, helpful reminder

kevin-dp:
I'm missing some context here but from @magnetised's explanation this morning, i'm guessing this message is meant to reset the client state when the client's permissions change, right? This is needed such that both the client and server are "in sync", i.e. they both restart from the "empty state" for this client (which i'm assuming is an enabling simplification). While this will also erase data that a client no longer has permission to see, this is not really it's primary purpose since we can't enforce that on malicious clients, right?

clients/typescript/src/satellite/client.ts
```diff
@@ -451,6 +456,14 @@ export class SatelliteClient implements Client {
     this.emitter.removeListener('relation', callback)
   }
 
+  subscribeToCommands(callback: CommandCallback): void {
+    this.emitter.on('command', callback)
+  }
+
+  unsubscribeToCommands(callback: CommandCallback): void {
```
kevin-dp, line 463: nit: we unsubscribe _from_ something

clients/typescript/test/satellite/process.test.ts
```diff
@@ -2262,3 +2267,39 @@ test("don't leave a snapshot running when stopping", async (t) => {
...
+        reason: SatClientCommand_ResetDatabase_Reason.PERMISSIONS_CHANGE,
+      }),
+    })
+  )
+
+  const results = await adapter.query({
+    sql: 'SELECT * FROM main.parent',
+  })
+
+  t.deepEqual(results, [])
```
kevin-dp, line 2304: Should we also check that the subscribed shapes are kept?

kevin-dp (approved):
Great work! I only left some minor comments.

clients/typescript/src/satellite/client.ts
magnetised: agreed - I was just being consistent with the current unsubscribe function names e.g. `unsubscribeToAdditionalData`, `unsubscribeToTransactions`, `unsubscribeToRelations`

clients/typescript/test/satellite/process.test.ts
magnetised: sure. rather than crowbar it in here though, I'll add it in a later pr and ping you.

PR #1218, opened 2024-05-02
Issue client reset message when permissions change

magnetised:
Rather than attempt to reset the server connection's state after we've asked the client to reset itself, we just close the connection. The client will then re-connect automatically and, since it's had its state reset, the server will treat it as a first connection (including dropping all client reconnection state).

Since this is intended as an enabling simplification to maintain consistency between client and server in the simplest way possible, pending a more efficient, less brutal but much more complicated solution, I've tried to keep the changes to the protocol as small as possible. 

Hence the use of `throw` -- it's either that or restructure all the message handling on the server side to be more of a `reduce_while` style.

### DIFF
--- a/clients/typescript/test/satellite/process.test.ts
+++ b/clients/typescript/test/satellite/process.test.ts
@@ -2312,6 +2312,8 @@ test('SatClientCommand.ResetDatabase clears all data', async (t) => {
   const { synced } = await satellite.subscribe([shapeDef])
   await synced
   await satellite._performSnapshot()
+  const subscriptionCount =
+    satellite.subscriptions.getFulfilledSubscriptions().length
 
   await client.commandCb!(
     SatClientCommand.fromPartial({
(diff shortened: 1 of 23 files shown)


clients/typescript/test/satellite/process.test.ts
magnetised, line 1: @kevin-dp here's the additional assertion you requested...

icehaunter (approved):

e2e/tests/03.03_node_satellite_sends_and_recieves_data.lux
```diff
@@ -4,15 +4,18 @@
...
-[invoke setup_client 2 "electric_1" 5133]
-
 [shell proxy_1]
     [invoke migrate_items_table 20230504114018]
 
+[invoke setup_client 1 "electric_1" 5133]
+[invoke setup_client 2 "electric_1" 5133]
+
 [shell satellite_1]
+    [timeout 60]
```
alco, line 14: Why is such a large timeout needed here specifically?

alco (approved):

e2e/tests/03.03_node_satellite_sends_and_recieves_data.lux
magnetised: oh, i should remove that -- good spot

PR #1220, opened 2024-05-02
chore(examples): web-pglite example use canary version of electric

samwillis:


msfstef:
I'm assuming we can close this in favour of https://github.com/electric-sql/electric/pull/1236?

PR #1228, opened 2024-05-06
feat(electric): Increase the ping heartbeat to 20 seconds from 5 seconds

davidmartos96:
After some discussion with @kevin-dp we are under the assumption that 5 seconds of ping timeout is quite low. Other popular websocket libraries like https://socket.io use a ping interval of 25 seconds and a ping timeout of 20 seconds. The timeout default changed from 5 seconds to 20 seconds over the years.

On top of this change, we noticed that the scheduler can sometimes undershoot. That is, if we schedule after 5000 ms, we could then obtain a `last_msg_diff` of 4999.7 ms, which wouldn't match the `>` operator. Because of that, a small margin error of 500 microseconds is added.
That would prevent obtaining uneven ping durations -> 20s, 20s, 40s, 20s, 40s,... (That's because if it doesn't match, we will schedule for another 20 seconds)

### DIFF
--- a/.changeset/poor-deers-marry.md
+++ b/.changeset/poor-deers-marry.md
@@ -1,0 +1,5 @@
+---
+"@core/electric": patch
+---
+
+Increase the ping heartbeat to 20 seconds from 5 seconds.
--- a/components/electric/lib/electric/satellite/ws_server.ex
+++ b/components/electric/lib/electric/satellite/ws_server.ex
@@ -43,8 +43,10 @@ defmodule Electric.Satellite.WebsocketServer do
   alias Electric.Satellite.Protocol.InRep
   alias Electric.Satellite.Protocol.Telemetry
 
-  # in milliseconds
-  @ping_interval 5_000
+  # Time interval at which the server will ping the client, in milliseconds
+  # This will also be the time that the server will wait for a response to the ping
+  # before disconnecting the client
+  @ping_interval 20_000
 
   def reg_name(name) do
     Electric.name(__MODULE__, name)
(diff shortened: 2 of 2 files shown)


icehaunter (approved):

davidmartos96:
@alco If I log the diff this is what I get. Most ones will be a bit over the target duration, but it can get a bit short sometimes.
But as you say, it might be something else.

image
 

image

alco:
@davidmartos96 @icehaunter [Here's an alternative approach](https://github.com/electric-sql/electric/commit/0144f4e92a9c74ee363cbb8fa2d9c2928d24e508) that, instead of adding a margin of error to timer firings, keeps a separate timestamp for sent pings. Every time a ping interval elapses, we check if any messages have arrived from the client after the previous ping had been sent. This way it doesn't matter whether the timer fires slightly earlier or slightly later.

davidmartos96:
@alco That looks easier, thanks! Should I pull those changes in, or can you commit on this branch?

A bit on topic with the pings, we found an odd behavior with the inactive websocket disconnection. I reported it the other day on Discord, but I don't know if you guys have it tracked on Linear.
https://discord.com/channels/933657521581858818/1240698235262337103/1240698237992833065

alco:
@icehaunter any comments?

icehaunter:
@alco no comments, I prefer your approach, GJ to both of you

alco:
@davidmartos96 I have pulled your changes into my branch and opened a new PR based on that - https://github.com/electric-sql/electric/pull/1334. Thanks for the contribution!

alco:
> A bit on topic with the pings, we found an odd behavior with the inactive websocket disconnection. I reported it the other day on Discord, but I don't know if you guys have it tracked on Linear. https://discord.com/channels/933657521581858818/1240698235262337103/1240698237992833065

I have filed this as a GH issue here - https://github.com/electric-sql/electric/issues/1335.

PR #1232, opened 2024-05-07
fix (client): pass correct dialect spelling to migrations endpoint

kevin-dp:
This PR fixes the CLI to pass the right dialect argument to the migrations endpoint and also error if the status code indicates a failure.


samwillis (approved):
👍 perfect

PR #1240, opened 2024-05-08
feat!: server-driven unsubscribes

icehaunter:
1. Adds new boundary messages to the protocol allowing the server to send a GONE batch - a set of messages for the client to know that those rows will not receive further updates after an unsubscribe call.
2. Server is the source of truth on which rows the client should have
3. Only while-online unsubscribes are possible
4. Client may reconnect at any point after sending the unsubscribe call and the server will adjust seen rows correctly and send a GONE batch if we don't believe the client had seen that -- see tests in `subscriptions_test.exs`.

Missing is a public API for unsubscribes, that's in the next PR.


clients/typescript/src/satellite/client.ts
```diff
@@ -473,6 +488,16 @@ export class SatelliteClient implements Client {
...
   }
 
+  subscribeToGoneBatch(callback: GoneBatchCallback) {
+    this.emitter.on('goneBatch', async (lsn, ids, changes, ack) => {
+      await callback(lsn, ids, changes)
+      ack()
+    })
+  }
+  unsubscribeToGoneBatch(_callback: GoneBatchCallback) {
+    // TODO: real removeListener implementation, because the old one for txns doesn't work
```
kevin-dp, line 498: Why can't we implement it as `this.emitter.removeListener('goneBatch', callbackFn)`?
(assuming we store the callback functions such that we can reference them here)

clients/typescript/src/satellite/client.ts
```diff
@@ -911,6 +936,8 @@ export class SatelliteClient implements Client {
   private handleTransaction(message: SatOpLog) {
     if (!this.subscriptionsDataCache.isDelivering()) {
       this.processOpLogMessage(message)
+    } else if (this.inbound.receivingUnsubsBatch) {
```
kevin-dp, line 939: Are we sure that `this.subscriptionsDataCache.isDelivering` and `this.inbound.receivingUnsubsBatch` can never be true at the same time?
Otherwise we could be getting an unsubs message that would pass the first if test and get processed by `processOpLogMessage`.
I have the impression that instead of testing for `this.inbound.receivingUnsubsBatch` we may want to track this in the `subscriptionsDataCache` or in the subscriptions manager.

clients/typescript/src/satellite/client.ts
```diff
@@ -963,6 +990,35 @@ export class SatelliteClient implements Client {
...
+  }
+
+  private handleUnsubsDataEnd(_msg: SatUnsubsDataEnd): void {
+    if (!this.inbound.receivingUnsubsBatch)
+      throw new SatelliteError(
+        SatelliteErrorCode.PROTOCOL_VIOLATION,
+        'Received a `SatUnsubsDataEnd` message but not the begin message'
+      )
+
+    const subscriptionIds = [...this.inbound.receivingUnsubsBatch]
```
kevin-dp, line 1005: Are we making a copy here? If yes, let's make this explicit through a comment.
Not sure why we are making a copy though since on L1018 we're setting `this.inbound.receivingUnsubsBatch = []` so doesn't seem to matter if we would mutate the original array.

clients/typescript/src/satellite/client.ts
```diff
@@ -963,6 +990,35 @@ export class SatelliteClient implements Client {
...
+
+    this.emitter.enqueueEmit(
+      'goneBatch',
+      this.inbound.last_lsn!,
+      subscriptionIds,
+      [...this.inbound.goneBatch],
```
kevin-dp, line 1011: same, not sure why we need to copy it

clients/typescript/src/satellite/shapes/manager.ts
```diff
@@ -117,15 +117,13 @@ export class InMemorySubscriptionsManager
...
-  /**
-   * Unsubscribes from one or more subscriptions.
-   * @param subId A subscription ID or an array of subscription IDs.
-   */
-  async unsubscribe(
-    subIds: SubscriptionId | SubscriptionId[]
-  ): Promise<SubscriptionId[]> {
-    const ids = Array.isArray(subIds) ? subIds : [subIds]
-    const shapes: ShapeDefinition[] = ids.flatMap(
+  unsubscribe(subIds: SubscriptionId[]): void {
```
kevin-dp, line 120: Where is `unsubscribe` used? I only found the version with GC being used.

clients/typescript/src/util/proto.ts
```diff
@@ -127,6 +127,8 @@ const msgtypetuples: MappingTuples = {
   SatRpcRequest: [21, Pb.SatRpcRequest],
   SatRpcResponse: [22, Pb.SatRpcResponse],
   SatOpLogAck: [23, Pb.SatOpLogAck],
+  SatUnsubsDataBegin: [24, Pb.SatUnsubsDataBegin],
```
kevin-dp, line 130: Both have number 24, is that normal? Just to sanity check as i'm not sure what these represent

clients/typescript/src/util/types.ts
```diff
@@ -202,9 +209,12 @@ export interface InboundReplication extends Replication<ServerTransaction> {
   additionalData: AdditionalData[]
   unseenAdditionalDataRefs: Set<string>
   incomplete?: 'transaction' | 'additionalData'
+  goneBatch: DataGone[]
+  receivingUnsubsBatch: false | string[]
```
kevin-dp, line 213: not particularly a fan of using this variable to do 2 things: 1) as a flag indicating whether we are not currently receiving unsubscription messages, and 2) to store an array of unsubs messages. Feels like an option type would be a cleaner abstraction here but yeah... we're doing JS, this is poor man's option type ;D

clients/typescript/test/satellite/process.ts
```diff
@@ -1918,6 +1921,47 @@ export const processTests = (test: TestFn<ContextType>) => {
     t.deepEqual(results, [])
   })
 
+  test('GONE batch is applied as DELETEs', async (t) => {
+    const { client, satellite, adapter } = t.context
+    const { runMigrations, authState, token } = t.context
```
kevin-dp, line 1926: let's move this into a single destructuring assignment

kevin-dp (commented):
Overall LGTM, left some comments here and there but nothing major.

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -346,4 +346,68 @@ export abstract class QueryBuilder {
...
+        .slice(processed, processed + currentDeleteCount)
+        .flatMap((record) => columns.map((col) => record[col] as SqlValue))
+
+      processed += currentDeleteCount
+      stmts.push({ sql, args })
+    }
+    return stmts
+  }
+
+  public makeQT(tablename: string): QualifiedTablename {
```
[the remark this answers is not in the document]
icehaunter: Counter nit: I don't like verbosity of `QualifiedTablename` class name everywhere, and would maybe shorten it to QT or QTN across the codebase :D

clients/typescript/src/satellite/client.ts
icehaunter: Because we rewrite the provided cb function to call an ack at the end, see above, so calling off with original callback will do nothing, but it was written like that before. I removed the buggy implementation some time ago, but didn't put anything on instead because we actually dont use the unsubscribe functions

clients/typescript/src/satellite/client.ts
icehaunter: Yes, server impl will never interleave batches of any kind

clients/typescript/src/util/types.ts
icehaunter: Me too, but we don't have options and I don't want to manage 3 fields instead

clients/typescript/src/util/proto.ts
icehaunter: Good catch

clients/typescript/src/satellite/shapes/manager.ts
icehaunter: https://github.com/electric-sql/electric/blob/ilia/vax-1289-unsubscribe-handling/clients/typescript/src/satellite/process.ts#L460

clients/typescript/src/satellite/client.ts
icehaunter: Our interaction with protocol is a FSM, but current code spreads that fact over 2 files already. I think this is reasonable enough because interleaving is not intended by the protocol

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -346,4 +346,68 @@ export abstract class QueryBuilder {
...
+    records: T[],
+    maxParameters: number,
+    suffixSql = ''
+  ): Statement[] {
+    const stmts: Statement[] = []
+    const columnCount = columns.length
+    const recordCount = records.length
+    let processed = 0
+    let positionalParam = 1
+    const pos = (i: number) => `${this.makePositionalParam(i)}`
```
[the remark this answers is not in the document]
icehaunter: This is a copy of your own code above! :D

components/electric/lib/electric/satellite/protobuf.ex
```diff
@@ -32,7 +34,9 @@ defmodule Electric.Satellite.Protobuf do
     SatShapeDataEnd => 18,
     SatRpcRequest => 21,
     SatRpcResponse => 22,
-    SatOpLogAck => 23
+    SatOpLogAck => 23,
+    SatUnsubsDataBegin => 24,
+    SatUnsubsDataEnd => 25
```
magnetised, line 39: see @kevin-dp's point re duplicated 24s

magnetised (changes requested):
excellent work. you do need to fix those protobuf id conflicts tho.

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -346,4 +346,68 @@ export abstract class QueryBuilder {
...
+        { length: columnCount },
+        (_, i) => `"${columns[i] as string}" = ${pos(positionalParam++)}`
+      ).join(' AND ')})`
+
+    // Largest number below maxSqlParamers that evenly divides by column count,
+    // divided by columnCount, giving the amount of rows we can insert at once
+    const batchMaxSize =
+      (maxParameters - (maxParameters % columnCount)) / columnCount
+    while (processed < recordCount) {
+      positionalParam = 1 // start counting parameters from 1 again
```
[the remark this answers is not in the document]
icehaunter: Yes but we''d be recreating the closure for a function. I feel like `= 1` for this very bounded block of code is easy enough to read while being more performant

clients/typescript/src/satellite/client.ts
icehaunter: It matters 10 lines down in the ack callback function. It will close over `this` and not over value of `this.inbound.receivingUnsubsBatch`, meaning that at the point when the callback will be called it will use already-cleared value of that fields instead of a full list.

clients/typescript/src/satellite/client.ts
icehaunter: here we don't

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -198,6 +200,10 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
            source_txns :: [non_neg_integer(), ...]}
   @type additional_data_row :: additional_data_sub_row() | additional_data_txn_row()
 
+  @unsubscribe_points_ets :unsubscribe_points
+  @type unsubscribe_points_row ::
+          {{client_id(), sub_id :: String.t(), wal_pos :: non_neg_integer()}}
```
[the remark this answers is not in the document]
icehaunter: I really don't want to. I structured the rest of the file like this (and you [had approved](https://github.com/electric-sql/electric/pull/899/files#diff-a62b6f57d365fae49cdd32774a3cc1c3f209357b5522b62467ddf2fc1705c86fR163-R189) it before) because for me it's easier to see name of the ETS table paired immediately with the row structure in the type.

components/electric/lib/electric/satellite/client_reconnection_info.ex
```diff
@@ -444,23 +441,92 @@ defmodule Electric.Satellite.ClientReconnectionInfo do
...
 
-      {:ok, new_graph, actions}
+      # We're also clearing the unsubscription points that are in the future
+      # but the subscription continued on reconnection
+      opts
+      |> Keyword.fetch!(:including_subscriptions)
+      |> Enum.map(&{{{client_id, &1, :_}}, [], [true]})
+      |> then(&:ets.select_delete(@unsubscribe_points_ets, &1))
+
+      # FIXME: this needs to clear all that from the future, which is now done not in this function!!!!!!!
```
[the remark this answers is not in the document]
icehaunter: Actually forgor :skull: 
I'll fix that in this PR

components/electric/lib/electric/postgres/extension/migrations/20240501000000_client_reconnection_unsub_points.ex
```diff
@@ -0,0 +1,25 @@
...
+  def version, do: 2024_05_01_00_00_00
+
+  @impl true
+  def up(_schema) do
+    [
+      """
+      CREATE TABLE #{Extension.client_unsub_points_table()} (
+        client_id VARCHAR(64) NOT NULL,
+        subscription_id UUID NOT NULL,
+        wal_pos BIGINT NOT NULL,
```
alco, line 16: I remember you raised an issue with this `BIGINT` type. It's not too late to change it here. Making the same change in the older migration can be done in a separate PR (I can handle that).

components/electric/lib/electric/replication/shapes/shape_request.ex
```diff
@@ -143,9 +143,14 @@ defmodule Electric.Replication.Shapes.ShapeRequest do
         origin,
         context
       ) do
-    # We're converting these records to a list of keys to query next layers on
-    curr_records =
-      Enum.map(moved_in_records, fn {id, record} -> {id, %Changes.NewRecord{record: record}} end)
+    # We're converting these records to a list of keys to query next layers on and building a fake-rooted graph.
+    # We're "rooting" the top layer of records so that `SentRowsGraph` functions know where to start traversal.
+    # It's important to get of that fake root later.
```
alco, line 148: I'm not sure what this says. "Get off that fake root"? At what point later?

magnetised (approved):
fantastic stuff

alco (approved):
:trophy:

PR #1241, opened 2024-05-08
fix(electric): Correct handling of multi-statement migration sql

magnetised:


samwillis (commented):
I've tested this in combination with #1242 and #1244 and it working. I'm not qualified for a code review though.

icehaunter (approved):
LGTM but let's maybe move VAX-1828 to triage or otherwise highlight it to @balegas while we're in the planning stage?

PR #1244, opened 2024-05-08
fix(client) Change _applyNonDeleteOperation to use pk column for creating the `ON CONFLICT` statement

samwillis:
Previous version prior to PG support didn't reference columns in the `ON CONFLICT`:https://github.com/electric-sql/electric/blame/c8e698148d4082d2c3fc1f081d410434f228b01b/clients/typescript/src/satellite/process.ts#L1631

I suspect this bug was inherited from the Tauri Linearlite demo as it only had PKs as "id"

### DIFF
--- a/.changeset/spicy-items-poke.md
+++ b/.changeset/spicy-items-poke.md
@@ -1,0 +1,5 @@
+---
+"electric-sql": patch
+---
+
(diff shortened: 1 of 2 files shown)


msfstef (approved):
Nice catch, the arbitrary `id` col does stick out!

PR #1247, opened 2024-05-09
fix(client): asyncEventEmitter to not silence unhandled exceptions raised in event handlers

samwillis:
Currently if there is an unhandled exception in an event handler of asyncEventEmitter they are silenced and don't make it to a global error handler or console. This changes it to re-throw them async using `queueMicrotask` so that they happen outside of the promise awaited by `allSettled`.

### DIFF
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -386,7 +386,7 @@ export class SatelliteProcess implements Satellite {
     this.disconnect()
 
     if (shutdown) {
-      this.client.shutdown()
+      await this.client.shutdown()
     }
   }
 
@@ -503,24 +503,29 @@ export class SatelliteProcess implements Satellite {
       subsData.subscriptionId
     )
 
-    await this._applySubscriptionData(
+    const applied = await this._applySubscriptionData(
       subsData.data,
       subsData.lsn,
       [],
       subsData.subscriptionId
     )
 
-    const toBeUnsubbed = afterApply()
-    if (toBeUnsubbed.length > 0) await this.unsubscribeIds(toBeUnsubbed)
+    if (applied) {
+      const toBeUnsubbed = afterApply()
+      if (toBeUnsubbed.length > 0) await this.unsubscribeIds(toBeUnsubbed)
+    }
   }
 
-  /** Insert incoming subscription data into the database. */
+  /**
+   * Insert incoming subscription data into the database.
+   * Returns flag indicating whether application was successful or not.
+   */
   private async _applySubscriptionData(
     changes: InitialDataChange[],
     lsn: LSN,
     additionalStmts: Statement[] = [],
     subscriptionId?: string
-  ) {
+  ): Promise<boolean> {
     const namespace = this.builder.defaultNamespace
     const stmts: Statement[] = []
 
@@ -662,6 +667,7 @@ export class SatelliteProcess implements Satellite {
         })
       })
       this.notifier.actuallyChanged(this.dbName, notificationChanges, 'initial')
+      return true
     } catch (e) {
       this._handleSubscriptionError(
         new SatelliteError(
@@ -670,6 +676,7 @@ export class SatelliteProcess implements Satellite {
         ),
         subscriptionId
       )
+      return false
     }
   }
 
--- a/clients/typescript/src/util/asyncEventEmitter.ts
+++ b/clients/typescript/src/util/asyncEventEmitter.ts
@@ -161,11 +161,11 @@ export class AsyncEventEmitter<Events extends EventMap> {
         // re-throw any rejected promises such that the global error
         // handler can catch them and log them, otherwise they would
         // be suppressed and bugs may go unnoticed
-        settledPromises
-          .filter((prom) => prom.status === 'rejected')
-          .forEach((rejectedProm) => {
+        settledPromises.forEach((rejectedProm) => {
+          if (rejectedProm.status === 'rejected') {
             throw rejectedProm.reason
-          })
+          }
+        })
       })
     } else {
       // signal that the queue is no longer being processed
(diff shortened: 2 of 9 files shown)


msfstef (approved):
I feel like it would be better to allow listeners to specify error handling as well and forward the error handling to them but this is definitely better than suppressing the error entirely!

Could you add a test for this as well? we can mock the global error handler in the test to assert it catches this

samwillis:
@msfstef 
> it would be better to allow listeners to specify error handling as well

I think a listener should handle it's own errors in the usual way, they can wrap their whole handler in a `try` if needed.

msfstef:
@samwillis definitely possible to leave it like that, although making the listener accept an error handler and wrapping it internally feels cleaner and forces the subscriber to handle errors (i.e. not a choice)

clients/typescript/src/satellite/process.ts
```diff
@@ -503,24 +503,29 @@ export class SatelliteProcess implements Satellite {
...
+   * Insert incoming subscription data into the database.
+   * Returns flag indicating whether application was successful or not.
+   */
   private async _applySubscriptionData(
     changes: InitialDataChange[],
     lsn: LSN,
     additionalStmts: Statement[] = [],
     subscriptionId?: string
-  ) {
+  ): Promise<boolean> {
```
msfstef, line 528: really ugly fix but this is a symptom of how we've mixed and matched the architecture here - if `_applySubscriptionData` fails, it still resolves but fires `_handleSubscriptionError`. However the error handler resets the state of the `subscriptionManager`, so the `afterApply` call in lines 513-516 still fires because `_applySubscriptionData` resolves without an issue but because the state of the subscription manager has been reset it attempts to resolve promises that are no longer present.

I think satellite probably needs a bit of rearchitecting as we're seeing various issues w.r.t. timing and async.

samwillis:
@msfstef Looks good to me, at least for now. I agree we need to re-arch some of this.

There is a failing test, not sure if that's the flaky one or a new one?

msfstef:
@samwillis the e2e was not flaky, the async event emitter change revealed a minor regression we had during the PG in the client integration where bytearrays were not being deserialized to pure `Uint8Array`s - not a significant issue, I doubt any real users would encounter it, but it was making better-sqlite3 throw for the empty bytearray edge case.

clients/typescript/src/util/asyncEventEmitter.ts
```diff
@@ -148,15 +146,29 @@ export class AsyncEventEmitter<Events extends EventMap> {
...
-        .allSettled(listenerProms)
-        .then(() => this.processQueue()) // only process the next event when all listeners have finished
+      const listenerProms = ls.map(async (listener) => {
+        try {
+          await listener(...args)
+        } catch (e) {
+          // If a listener throws an error, we re-throw it asynchronously so that the queue can continue
+          // to be processed, this ensures that the exception isn't swallowed by allSettled below.
+          // It will likely be caught by a global error handler, or be logged to the console.
+          queueMicrotask(() => {
```
kevin-dp, line 156: Is `queueMicrotask` available in all environments?
The JS docs say:
> The queueMicrotask() method, which is exposed on the [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window) or [Worker](https://developer.mozilla.org/en-US/docs/Web/API/Worker) interface

So they are talking about browser environments.
Looked it up for node and node seems to support it as well.
Not sure about native environments such as react-native?

Perhaps we don't need this `queueMicrotask` trick since `allSettled` returns an array of promise statuses (fulfilled or rejected). So we could loop over the `processingProm` array, aggregate all errors into one, and then throw the aggregated error. Would that work?

kevin-dp (approved):
Left one comment wrt `queueMicrotask()` function, for the rest LGTM.

clients/typescript/src/util/asyncEventEmitter.ts
msfstef: I've merged the PR to avoid having it linger for too long as I think it's an important one to have in main, but if there's any disagreements on this last change let me know.

clients/typescript/src/util/asyncEventEmitter.ts
```diff
@@ -148,15 +146,27 @@ export class AsyncEventEmitter<Events extends EventMap> {
...
+      // only process the next event when all listeners have finished
+      this.processing.then((settledPromises) => {
+        this.processQueue()
+
+        // re-throw any rejected promises such that the global error
+        // handler can catch them and log them, otherwise they would
+        // be suppressed and bugs may go unnoticed
+        settledPromises.forEach((rejectedProm) => {
+          if (rejectedProm.status === 'rejected') {
+            throw rejectedProm.reason
```
davidmartos96, line 166: @msfstef One question. If there were 2 failing promises, wouldn't this only throw one of them? The `throw` would be stopping the `forEach`.

samwillis: Hey @msfstef 

just catching up on this (little time yesterday). I considered what @kevin-dp suggested but went with queueMicrotask as it would immediately re-throw the error and not wait for all to complete (which could be never)

msfstef: @samwillis you're right about the possibility of the promises never resolving and thus never throwing - I think then we all agree to revert the throwing to the listener execution, only detail is whether we want to use `queueMicrotask` or perhaps `setTimeout` if we have compatibility concerns.

From what I'm seeing I doubt we'll have issues with `queueMicrotask` so we might as well go with that and address any compatibility issues if and when they come up.

msfstef: I've opened a small PR to revert this to the original error throwing: https://github.com/electric-sql/electric/pull/1312

PR #1249, opened 2024-05-09
fix: disable automatic fallback from SSL to non-SSL DB connection

icehaunter:
`Postgrex` connector we're using now doesn't support fallback behavior like this, so we're disabling it for `epgsql` connections as well. This commit also fixes application of the default value for requiring SSL, which was erroneously "false" before. This means that if your DB doesn't support SSL, you need to explicitly specify it via `DATABASE_REQUIRE_SSL=false` env variable or `?sslmode=disable` in the connection string.

### DIFF
--- a/components/electric/config/runtime.exs
+++ b/components/electric/config/runtime.exs
@@ -220,22 +220,20 @@ connector_config =
     require_ssl? =
       case {require_ssl_config, conn_config[:sslmode]} do
         {nil, :require} -> true
-        {nil, _} -> false
         {nil, nil} -> default_database_require_ssl
+        {nil, _} -> false
         {true, _} -> true
         {false, _} -> false
       end
 
-    # When require_ssl?=true, epgsql will try to connect using SSL and fail if the server does not accept encrypted
-    # connections.
-    #
-    # When require_ssl?=false, epgsql will try to connect using SSL first, then fallback to an unencrypted connection
-    # if that fails.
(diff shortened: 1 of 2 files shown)


components/electric/config/runtime.exs
```diff
@@ -220,22 +220,20 @@ connector_config =
     require_ssl? =
       case {require_ssl_config, conn_config[:sslmode]} do
         {nil, :require} -> true
-        {nil, _} -> false
         {nil, nil} -> default_database_require_ssl
+        {nil, _} -> false
```
icehaunter, line 224: @alco match order made it so `{nil, nil}` case never matched before this change

samwillis:
@icehaunter 
Does this require a docs change and calling out in the release notes as it changes the default. Whats the impact on existing setups?

icehaunter:
@samwillis 
This impacts any setups where the server does not support SSL - `docker-compose`, same-cluster k8s, or local dev. Probably does warrant a callout in release notes, yeah. Without this change any setups without an explicit flag but with DBs expecting SSL are broken until it's set. With this change setups without an explicit flag but with DBs not supporting SSL are broken, but broken better with explicit error on startup instead of a hidden one later on.

We can also change the default to NOT use ssl - so that we break setups expecting SSL but with a better explicit error on startup stating "hey set this flag probably"

samwillis (approved):
Approved, but don't merge until there is a docs PR please.

PR #1251, opened 2024-05-09
fix(client): Better throttle snapshot stop strategy 

davidmartos96:
We've encountered a race condition that can occurr if, after cancelling the throttle when stopping the process, some other piece of code calls throttlesnapshot. That's because cancelling the throttle doesn't prevent from scheduling new ones.

The proposed solution is to assign undefined to the function, so places that would normally trigger a snapshot, like the poll interval, would just do nothing until properly closed.

On top of this, we found that `setClientListeners` is only called when instantiating the process, but it's cleaned up when calling `stop` with the `disconnect` function. This makes it so that stopping and then starting the process won't have some of the client listeners properly set up. The proposed solution is to just move the client listeners intialization to the `start` function.

I'm not sure how to test this. It would probably need some way to mock client events after a stop and a start.

### DIFF
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -130,7 +130,7 @@ export class SatelliteProcess implements Satellite {
 
   _pollingInterval?: any
   _unsubscribeFromPotentialDataChanges?: UnsubscribeFunction
-  _throttledSnapshot?: ThrottleFunction
+  _throttledSnapshot: ThrottleFunction
 
   _lsn?: LSN
 
@@ -177,8 +177,14 @@ export class SatelliteProcess implements Satellite {
     this.subscriptions = new InMemorySubscriptionsManager(
       this._garbageCollectShapeHandler.bind(this)
     )
-    this._throttledSnapshot = this._createThrottledSnapshotFun()
-
+    this._throttledSnapshot = throttle(
+      this._mutexSnapshot.bind(this),
+      opts.minSnapshotWindow,
+      {
+        leading: true,
+        trailing: true,
+      }
+    )
     this.subscriptionNotifiers = {}
 
     this.subscriptionIdGenerator = () => genUUID()
@@ -204,9 +210,6 @@ export class SatelliteProcess implements Satellite {
       await this.logDatabaseVersion()
     }
 
-    if (!this._throttledSnapshot) {
-      this._throttledSnapshot = this._createThrottledSnapshotFun()
-    }
     this.setClientListeners()
 
     await this.migrator.up()
@@ -245,18 +248,16 @@ export class SatelliteProcess implements Satellite {
 
     // Request a snapshot whenever the data in our database potentially changes.
     this._unsubscribeFromPotentialDataChanges =
-      this.notifier.subscribeToPotentialDataChanges(() =>
-        this._throttledSnapshot?.()
-      )
+      this.notifier.subscribeToPotentialDataChanges(this._throttledSnapshot)
 
     // Start polling to request a snapshot every `pollingInterval` ms.
     this._pollingInterval = setInterval(
-      () => this._throttledSnapshot?.(),
+      this._throttledSnapshot,
       this.opts.pollingInterval
     )
 
     // Starting now!
-    await this._throttledSnapshot?.()
+    await this._throttledSnapshot()
 
     // Need to reload primary keys after schema migration
     this.relations = await this._getLocalRelations()
@@ -274,17 +275,6 @@ export class SatelliteProcess implements Satellite {
     if (subscriptionsState) {
       this.subscriptions.setState(subscriptionsState)
     }
-  }
-
-  private _createThrottledSnapshotFun(): ThrottleFunction {
-    return throttle(
-      this._mutexSnapshot.bind(this),
-      this.opts.minSnapshotWindow,
-      {
-        leading: true,
-        trailing: true,
-      }
-    )
   }
 
   private async logDatabaseVersion(): Promise<void> {
@@ -329,7 +319,7 @@ export class SatelliteProcess implements Satellite {
     this.client.subscribeToRelations(this._updateRelations.bind(this))
     this.client.subscribeToTransactions(this._applyTransaction.bind(this))
     this.client.subscribeToAdditionalData(this._applyAdditionalData.bind(this))
-    this.client.subscribeToOutboundStarted(() => this._throttledSnapshot?.())
+    this.client.subscribeToOutboundStarted(this._throttledSnapshot.bind(this))
 
     this.client.subscribeToSubscriptionEvents(
       this._handleSubscriptionData.bind(this),
@@ -344,8 +334,7 @@ export class SatelliteProcess implements Satellite {
 
   private async _stop(shutdown?: boolean): Promise<void> {
     // Stop snapshotting and polling for changes.
-    this._throttledSnapshot?.cancel()
-    this._throttledSnapshot = undefined
+    this._throttledSnapshot.cancel()
 
     // Ensure that no snapshot is left running in the background
     // by acquiring the mutex and releasing it immediately.
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -132,7 +132,6 @@ export class SatelliteProcess implements Satellite {
   _pollingInterval?: any
   _unsubscribeFromPotentialDataChanges?: UnsubscribeFunction
   _throttledSnapshot: ThrottleFunction
-  _startProcessPromise?: Promise<void>
 
   _lsn?: LSN
 
@@ -210,12 +209,6 @@ export class SatelliteProcess implements Satellite {
   }
 
   async start(authConfig?: AuthConfig): Promise<void> {
-    this._startProcessPromise = this._startProcess(authConfig)
-    await this._startProcessPromise
-    this._startProcessPromise = undefined
-  }
-
-  private async _startProcess(authConfig?: AuthConfig): Promise<void> {
     if (this.opts.debug) {
       await this.logDatabaseVersion()
     }
@@ -380,13 +373,6 @@ export class SatelliteProcess implements Satellite {
   }
 
   private async _stop(shutdown?: boolean): Promise<void> {
-    // Wait for the start promise to finish first
-    // This would only happen if we stop the process too soon and the process
-    // hasn't finished starting yet
-    if (this._startProcessPromise) {
-      await Promise.allSettled([this._startProcessPromise])
-    }
-
     // Stop snapshot polling
     if (this._pollingInterval !== undefined) {
       clearInterval(this._pollingInterval)
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -411,10 +411,13 @@ export class SatelliteProcess implements Satellite {
     this._removeClientListeners?.()
     this._removeClientListeners = undefined
 
-    // Cancel the snapshot throttle and wait for any ongoing snapshot to finish
+    // Cancel the snapshot throttle
     this._throttledSnapshot.cancel()
+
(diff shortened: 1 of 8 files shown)


msfstef:
Thanks for spotting this issue and submitting a fix!

Although unsetting the throttled snapshot resolves the issue I am a bit reluctant about using that as a solution cause ultimately if it's getting called after termination we have some more underlying timing issue which should probably be solved instead.

The way I understand it, the throttled snapshot is called in four places:
1) when the process is started
2) on a polling interval
3) as a response to the notifier "potential data changes" subscription
4) as a response to the client "outbound started" subscription


I think one main issue right now is that in `_stop` we first wait for the lock to be acquired before clearing the polling interval, notifier subscriptions, and client subscriptions, so until any current snapshot is done running and releases the lock all the schedulers can still schedule snapshots (or do any other work they are scheduled to do, which is why I think it's better to fix the order of operations).

To solve 2) and 3), I think moving the lock acquisition for the sake of waiting on any current snapshots after clearing the interval and notifier subscriptions should work:
https://github.com/electric-sql/electric/blob/d2df5fe2c97d3f21c361d8e5703818547418159a/clients/typescript/src/satellite/process.ts#L339-L342

To solve 4) I think there needs to be an additional, separate call to unsubscribe from client listeners (since `client.shutdown` is not necessarily called and thus does not necessarily clear the scheduler for snapshots), in a similar way with the additional fix you've made for calling `setClientListeners` in the `start` method.

The 1) case is only an issue if `stop` is called before `start` finishes executing, which is really an edge case, and there are a few ways we could handle it but I think it can be split out as a separate fix - I suspect the issue you're running into stems from the scheduling of calls as you've said.

Let me know if this sounds reasonable to you, the solution you've proposed should work but we'd rather solve the underlying issues first as they could cause problems elsewhere!

P.S. indeed a test would be great for catching this, doing a timing mock for the 2), a notifier event mock for 3), or a client event mock for 4) I believe should capture this error somehow

davidmartos96:
@msfstef Thank you for reviewing!
You are right that moving the "wait for snapshot" code to the end would solve the scheduling issues.

While doing the test I noticed that I wasn't able to properly trigger it. I then noticed that the problem is that the clean function in the context doesn't actually close the underlying db connection, so even if the db files are removed, the connection still works, preventing seeing the issue of snapshoting when the db is closed.

I used the same mechanism for postgres and attach the stop function to the sqlite tests context. Making this change already manifested problem number 1 in some of the tests, so I included my proposed solution for that here as well.
I opted for a "startProcess" promise which would need to be awaited before stopping.

davidmartos96:
It looks like some pglite tests can now fail because the db connection is actually closed when tearing down each test.

For instance this simple test, because the connection promise is not awaited, when doing `startReplication` the db is closed, so it will fail when reading the meta table. Should we update those tests to actually wait for the connection to finish? Or should this be fixed somehow in the source code? Maybe if the process is stopped, don't propagate the errors in the `initializing` Promise

```js
test('can use sub in JWT', async (t) => {
    const { satellite, authState } = t.context

    await t.notThrowsAsync(async () => {
      await startSatellite(
        satellite,
        authState,
        insecureAuthToken({ sub: 'test-userB' })
      )
    })
  })
```

msfstef:
@davidmartos96 thank you for updating the tests and everything!

You're right, there are some tests that are not waiting for the connection promise to fulfil and that causes issues. From my testing it seems the following tests in`process.ts` require this kind of fix:
```ts
await t.notThrowsAsync(async () => {
    const { connectionPromise } = await startSatellite(
      satellite,
      authState,
      insecureAuthToken({ user_id: 'test-userA' })
    )
    await connectionPromise
})
```

- `set persistent client id`
- `can use sub in JWT`
- `require user_id or sub in JWT`
- `cannot update user id` (needs to await connection after failure assertion)


There seems to be one additional test for postgres failing, might require the following fix in `drivers/node-postgres/database.ts`:
```ts
  let stopPromise: Promise<void>

  // We use the database directory as the name
  // because it uniquely identifies the DB
  return {
    db,
    stop: () => {
      if (stopPromise) return stopPromise
      stopPromise = pg.stop()
      return stopPromise
    },
  }
```

With those fixes in place it seems tests are good! I will separately review the updated changes

clients/typescript/src/satellite/process.ts
```diff
@@ -208,10 +210,18 @@ export class SatelliteProcess implements Satellite {
   }
 
   async start(authConfig?: AuthConfig): Promise<void> {
+    this._startProcessPromise = this._startProcess(authConfig)
```
msfstef, line 213: I had discussed this with @kevin-dp that we wanted to avoid this pattern of chaining starts and stops - that being said if it fixes existing issues in tests etc then we can keep this in and improve it later. This still leaves room for "starts" not waiting for "stop" to finish, and multiple starts not waiting for the previous one to finish, so there's still annoying edge cases when it comes to user-driven starts and stops.

clients/typescript/src/satellite/process.ts
```diff
@@ -362,13 +404,30 @@ export class SatelliteProcess implements Satellite {
...
+    this._throttledSnapshot.cancel()
+
+    await this._ensureNoActiveSnapshot()
+
     if (shutdown) {
       this.client.shutdown()
     }
   }
 
+  // Make sure no snapshot is running after we stop the process, otherwise we might be trying to
```
msfstef, line 422: nit: I think a more appropriate name for this is `_waitForActiveSnapshots()` as it basically enters a queue waiting for the lock to be acquired from previous snapshots that have it - also this top level comment might be better suited to where we actually call this inside `_stop` and moving the inner comment as this method's description

clients/typescript/src/satellite/process.ts
```diff
@@ -362,13 +404,30 @@ export class SatelliteProcess implements Satellite {
       }
     })
 
+    this._removeClientListeners?.()
+    this._removeClientListeners = undefined
+
     this.disconnect()
 
+    // Stop snapshotting and polling for changes.
```
msfstef, line 412: nit: comment refers to "stop polling for changes" but that's done on line 386 when the polling interval is cleared - might be worth updating the comments appropriately

clients/typescript/src/satellite/process.ts
```diff
@@ -362,13 +404,30 @@ export class SatelliteProcess implements Satellite {
       }
     })
 
+    this._removeClientListeners?.()
+    this._removeClientListeners = undefined
+
     this.disconnect()
```
msfstef, line 410: I think it makes more sense to call `disconnect` right before the client shutdown in lines 417-419 - the overall order looks like:

1. wait for satellite to finish starting
2. remove any listeners/intervals that are scheduling snapshots
3. clear any throttled snapshots
4. wait for any snapshots that were already running to complete
5. NOW disconnect and shut down, only once any active snapshots have already finished running (as they may require sending data over the wire so disconnecting before they are done would be an issue)

clients/typescript/src/satellite/process.ts
```diff
@@ -314,17 +324,50 @@ export class SatelliteProcess implements Satellite {
...
 
-  setClientListeners(): void {
-    this.client.subscribeToError(this._handleClientError.bind(this))
-    this.client.subscribeToRelations(this._updateRelations.bind(this))
-    this.client.subscribeToTransactions(this._applyTransaction.bind(this))
-    this.client.subscribeToAdditionalData(this._applyAdditionalData.bind(this))
-    this.client.subscribeToOutboundStarted(this._throttledSnapshot.bind(this))
+  // Returns a function that removes the listeners
+  setClientListeners(): () => void {
+    if (this._removeClientListeners) {
```
msfstef, line 329: It feels slightly odd that `setClientListeners` is aware of `_removeClientListeners` but it does not set it internally - perhaps every call to `setClientListeners` sets the `_removeClientListeners` internally to make sure that any subsequent changes to this take that into account?

Either that or the `_removeClientListeners` call is placed outside right before a call to `setClientListeners` (e.g. line 222)

clients/typescript/src/satellite/process.ts
```diff
@@ -333,13 +376,12 @@ export class SatelliteProcess implements Satellite {
...
-    const releaseMutex = await this.snapshotMutex.acquire()
-    releaseMutex()
+    // Wait for the start promise to finish first
+    // This would only happen if we stop the process too soon and the process
+    // hasn't finished starting yet
+    if (this._startProcessPromise) {
+      await Promise.allSettled([this._startProcessPromise])
+    }
 
     if (this._pollingInterval !== undefined) {
```
msfstef, line 386: nit: no need for an undefined check, might as well just clear it (regardless of what it is) - additionally, might be worth clearing it right before setting it in `_startProcess` as well for good measure to make sure we don't end up with any orphaned intervals

msfstef (changes requested):
Looks good! Left some comments about ordering and some nits, if you could add a changeset for this as well that would be great! (`pnpm changeset` in the root directory, should be a patch change)

clients/typescript/src/satellite/process.ts
davidmartos96: @msfstef  What do you mean here?
Something like this? Does this really solve any edge case?

```ts
this._startProcessPromise = undefined
this._startProcessPromise = this._startProcess(authConfig)
```

clients/typescript/src/satellite/process.ts
davidmartos96: Now that some tests have been updated by awaiting the connection, it looks liek this might not be needed anymore. So if you think this is unwanted we can remove it until another alternative is implemented that doesn't involve this chaining.

clients/typescript/src/satellite/process.ts
davidmartos96: True, I first put it before because it was implicitly clearing the client listeners, but now that there is a explicit clear for those listeners, it makes sense to do it at the end.

clients/typescript/src/satellite/process.ts
msfstef: sorry I meant for the polling interval:
```ts
clearInterval(this._pollingInterval)
```
without the check - and I meant clearing it right before setting it in `_startProcess`

clients/typescript/src/satellite/process.ts
msfstef: If it's not too much trouble yeah let's remove the chaining as we want to handle it some other way - for now we assume it is the user's responsibility to await their starts before stopping

clients/typescript/src/satellite/process.ts
```diff
@@ -314,17 +325,53 @@ export class SatelliteProcess implements Satellite {
     await this.adapter.runInTransaction(...stmtsWithTriggers)
   }
 
+  // Returns a function that removes the listeners
```
msfstef, line 321: nit: this no longer returns a function that removes listeners

davidmartos96:
@msfstef Tests seem to be ok now, thank you for the comments!

msfstef (approved):
Looks good, thank you for your contribution! This was a good catch!

clients/typescript/src/satellite/process.ts
davidmartos96: Oops, good catch

clients/typescript/src/satellite/process.ts
```diff
@@ -333,14 +374,7 @@ export class SatelliteProcess implements Satellite {
...
-    // Stop snapshotting and polling for changes.
-    this._throttledSnapshot.cancel()
-
-    // Ensure that no snapshot is left running in the background
-    // by acquiring the mutex and releasing it immediately.
-    const releaseMutex = await this.snapshotMutex.acquire()
-    releaseMutex()
-
+    // Stop snapshot polling
     if (this._pollingInterval !== undefined) {
```
kevin-dp, line 378: As noted by @msfstef, we don't have to do this if test here. We can directly do `clearInterval(this._pollingInterval)`.

kevin-dp (approved):
Great catch and solution, awesome! 💯
Left some minor comments.

davidmartos96:
@kevin-dp Done!

PR #1255, opened 2024-05-13
fix: set `Postgrex` SSL mode based on the connection by `:epgsql`

icehaunter:
`Postgrex` connector doesn't support a behaviour equivalent to `sslmode=prefer` like `:epgsql` does, which we made use of when `DATABASE_REQUIRE_SSL=false`. To emulate support for that, after we make the first connection with `:epgsql`, we update the config to either force SSL or disable SSL based on how `:epgsql` connected.

### DIFF
--- a/components/electric/lib/electric/replication/postgres_manager.ex
+++ b/components/electric/lib/electric/replication/postgres_manager.ex
@@ -270,6 +270,9 @@ defmodule Electric.Replication.PostgresConnectorMng do
           "Successfully initialized origin #{origin} at extension version #{List.last(versions)}"
         )
 
+        # We're inspecting the state of `epgsql_sock` GenServer to figure out if the established
+        # connection is using SSL or not. This is needed to configure `Electric.Postgres.Repo`
+        # later on. There is no current public API on `:epgsql` that allows us to get this info otherwise.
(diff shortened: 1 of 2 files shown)


components/electric/lib/electric/replication/postgres_manager.ex
```diff
@@ -266,7 +270,7 @@ defmodule Electric.Replication.PostgresConnectorMng do
           "Successfully initialized origin #{origin} at extension version #{List.last(versions)}"
         )
 
-        :ok
+        {:ok, elem(:sys.get_state(conn), 1) == :ssl}
```
magnetised, line 273: is there no other way!!???! 😱  Is the final ssl state a union of the initial config plus whatever happens during the connection setup? in which case still yikes.

icehaunter: Yeah, so we need to configure `Postgrex` with either `true` or `false` for ssl, and it doesn't do "try ssl and fall back to plain". `:epgsql` does, but once it's established the connection it doesn't expose an API to inspect if the connection is in SSL or TCP mode. But the state has that info :upside_down_face: 
We can make a PR to epgsql for an API for this later

PR #1229, opened 2024-05-06
chore(e2e): Reduce flake in e2e tests by waiting for migrations to complete

alco:

PR #1256, opened 2024-05-13
chore(e2e): Fix flaky E2E test

magnetised:


e2e/common.mk
```diff
@@ -115,7 +115,7 @@ docker-attach-%:
 	docker compose -f ${DOCKER_COMPOSE_FILE} exec $* bash
 
 docker-prisma:
-	DOCKER_COMPOSE_FILE=${DOCKER_COMPOSE_FILE} ../prisma_example/run.sh ${ARGS}
+	DOCKER_COMPOSE_FILE=${DOCKER_COMPOSE_FILE} ../prisma_example/run.sh $(ARGS)
```
alco, line 118: While we're at it, this make target should be moved into `e2e/tests/Makefile`. There's nothing "common" about it since it relies on the fact that the current working directory is `e2e/<x>`, that's the only way this relative path reference can work.

e2e/prisma_example/run.sh
```diff
@@ -2,4 +2,11 @@
 
 set -ex
 
-docker compose -f ${DOCKER_COMPOSE_FILE} run "$@"
+pwd
```
alco, line 5: Leftover debug info?

e2e/tests/04.02_prisma_introspection_generates_correct_schema.lux
```diff
@@ -103,8 +103,8 @@
     !mkdir -p ../prisma_example/prisma/
     ?SH-PROMPT:
     !cp ../prisma_example/schema.prisma ../prisma_example/prisma/
-    [timeout 60]
     ?SH-PROMPT:
+    [timeout 60]
```
alco, line 107: This timeout should be reset to the default value afterwards, see how I did it in https://github.com/electric-sql/electric/pull/1229/files#diff-389d7656991862b0cfa6e4db62db0b32f07b6f5730756cf34c16a09e2a894f12

e2e/prisma_example/run.sh
magnetised: hey, yeah, i'm still working on this -- the tests are passing but I'm not convinced the flakiness has been fixed.

alco (approved):
Brilliant!

PR #1258, opened 2024-05-14
chore (e2e): generate fresh client and fix typings

kevin-dp:
Generates a new client for the e2e test client and fixes the type issues.


e2e/satellite_client/src/client.ts
```diff
@@ -35,11 +38,11 @@ async function makePgDatabase(): Promise<Database> {
   return client
 }
 
-export const make_db = async (name: string): Promise<any> => {
+export const make_db = async (name: string): Promise<DB> => {
   dbName = name
   console.log("DIALECT: " + process.env.DIALECT)
   if (process.env.DIALECT === 'Postgres') {
-    electrify = electrifyPg
+    electrify = <Schema extends DbSchema<any>>(db: DB, dbDescription: Schema, config: ElectricConfig) => electrifyPg(db as Database, dbDescription, config)
```
adelgado, line 45: How come we need this `as` as well as the one on line 24?

kevin-dp: We need it because here we are calling `electrifyPg` which expects a `Database` instance but `db` is of type `Database | BetterSqliteDatabase` so TS would complain. However, since we know `DIALECT` was configured with `Postgres` we know it will be a PG database, so we typecast it to `Database`. L24 is similar but for SQLite.

The root of this problem comes from the fact that we are declaring the `electrify` variable which will contain either the `electrify` function for SQLite or for PG. But those expect a different type of DB. That's why i created a single function signature which uses the `DB` union type: `<Schema extends DbSchema<any>>(db: DB, dbDescription: Schema, config: ElectricConfig) => Promise<ElectricClient<Schema>>`. Ideally i would like that `electrify` is of type `typeof electrifySqlite | typeof electrifyPg` but that doesn't seem to work, here's a minimal snippet showing that it does not work like that:
```ts
function fun(
  f: ((a: number) => boolean) | ((a: string) => boolean),
  a: number | string
): boolean {
  return f(a) // type error
}
```
So apparently, TS doesn't seem to like these kind of overloaded function signatures

e2e/satellite_client/src/client.ts
```diff
@@ -325,7 +330,7 @@ export const get_json = async (electric: Electric, id: string) => {
     },
     select: {
       id: true,
-      js: true,
+      //js: true,
```
msfstef, line 333: nit: any reason not to remove this (and the `_js` argument for `write_json` below and any references to that argument in e2es)? I see we've removed it from the schema anyway

e2e/satellite_client/src/client.ts
```diff
@@ -35,11 +38,11 @@ async function makePgDatabase(): Promise<Database> {
   return client
 }
 
-export const make_db = async (name: string): Promise<any> => {
+export const make_db = async (name: string): Promise<DB> => {
   dbName = name
   console.log("DIALECT: " + process.env.DIALECT)
   if (process.env.DIALECT === 'Postgres') {
```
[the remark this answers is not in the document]
kevin-dp: I removed the mutable builder variable. Are you happy with having `make_db` build a SQLite/PG DB based on the env var? Or you want to move this also into `electrify_db`? I think creating the DB and electrifying are 2 separate concerns and thus best implemented as separate functions.

e2e/satellite_client/src/client.ts
kevin-dp: I had left it there such that we can uncomment when we support JSON on Electric. But these changes are minimal so might as well remove it for clarity.

e2e/satellite_client/src/client.ts
```diff
@@ -60,20 +58,25 @@ export const electrify_db = async (
...
     schema.pgMigrations = migrations
   } else {
     schema.migrations = migrations
   }
-  const result = await electrify(db, schema, config)
+  
+  const electric = process.env.DIALECT === 'Postgres'
+    ? await electrifyPg(db as PgDatabase, schema, config)
+    : await electrifySqlite(db as BetterSqliteDatabase, schema, config)
```
adelgado, line 70: Perhaps we could substitute the cast for a type guard so the logic would be encoded explicitly? Something like:

```typescript
  function isPostgresDb(dialect: string | undefined, db: DB): db is PgDatabase {
    return dialect === 'Postgres'
  }

  const electric = isPostgresDb(process.env.DIALECT, db)
    ? await electrifyPg(db, schema, config)
    : await electrifySqlite(db, schema, config)
```

e2e/satellite_client/src/client.ts
```diff
@@ -60,20 +62,25 @@ export const electrify_db = async (
...
-  const result = await electrify(db, schema, config)
+  
+  const electric = isPostgresDb(process.env.DIALECT, db)
+    ? await electrifyPg(db, schema, config)
+    : await electrifySqlite(db, schema, config)
+  
   const token = await mockSecureAuthToken(exp)
 
-  result.notifier.subscribeToConnectivityStateChanges((x: any) => console.log(`Connectivity state changed: ${x.connectivityState.status}`))
+  electric.notifier.subscribeToConnectivityStateChanges((x: any) => console.log(`Connectivity state changed: ${x.connectivityState.status}`))
```
adelgado, line 78: Here we don't need to annotate `x` as being `any`, it's unnecessarily widening the type. If you just keep it without an annotation it'll correctly be inferred to `ConnectivityStateChangeNotification` which is what the callback provides.

Screenshot 2024-05-16 at 12 25 44

e2e/satellite_client/src/client.ts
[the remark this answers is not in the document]
msfstef: I think they can remain separate - and make_db can have a switch statement on the dialect fro creating the db (with a default clause throwing an error for good measure, no reason to assume sqlite is default)

since we assume the env var is static having both `make_db` and `electrify_db` depend on it separately is better than a mutable top level var I think

e2e/satellite_client/src/client.ts
kevin-dp: This is solved based on the change proposed by @msfstef

e2e/satellite_client/src/client.ts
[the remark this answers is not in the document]
kevin-dp: Had to do some improvements because the dialect is only provided when running Postgres. If not provided, we default to SQLite. I introduced a `getDialect` that based on the different possibilities returns the chosen dialect. Then i could use a switch statement as suggested.

PR #1260, opened 2024-05-14
fix(toolbar): Suppress warning for invalid annotation

msfstef:
Rollup removes invalid legacy annotations - they are warnings not errors but they fail the build so suppressing them should work for now.


samwillis (approved):
LGTM

PR #1261, opened 2024-05-14
fix(electric): Fix the order of case clauses when parsing database SSL config

alco:
@icehaunter spotted this in https://github.com/electric-sql/electric/pull/1249#discussion_r1595304797.

I have reordered the case clauses by specificity and added comments for readability.


icehaunter:
This is a breaking change because it switches the default from `false` to `true` on `DATABASE_REQUIRE_SSL`. This matches the docs though. This is specified in the issue VAX-1839. So coordinate with @samwillis and @balegas on this

alco:
When the default value of `DATABASE_REQUIRE_SSL` was set to `true`  in https://github.com/electric-sql/electric/pull/767 (2023-12-22), it was implemented correctly. The later change that accidentally flipped the default (https://github.com/electric-sql/electric/pull/848) was made on 2024-01-18.

I would argue this is a bug fix, more than a breaking change, but still worth mentioning in the Release Notes.

alco:
Corrections:
* the original enforcement of SSL connections by default was released in [v0.9.0](https://github.com/electric-sql/electric/releases/tag/electric-sql%400.9.0) on 2024-01-18.
* the "fix" that added support for `sslmode` and accidentally flipped the default value was released in [v0.9.1](https://github.com/electric-sql/electric/releases/tag/%40core%2Felectric%400.9.1) on 2024-01-24.

samwillis:
I agree it should be should to true.

icehaunter (approved):

alco:
I have prepared [a draft of release notes](https://deploy-preview-136--electric-sql-website.netlify.app/docs/reference/release_notes#2024-xx-xx---patch-release) if we release this change as part of a patch release.

msfstef:
Seems like everyone agrees on this change (I see valter acked on Linear too) so I think we can get this merged? (I've had deploy issues/confusion as well because of this so making sure we get it merged!)

PR #1262, opened 2024-05-14
feat(toolbar): Support Postgres dialect in toolbar

msfstef:
Add support for postgres dialect - modified the test suite to check for both postgres and sqlite dialects.

Abstracted the sql related code to support both dialects to a different module.

The schema hover card for PG uses a table with the column specification rather than the plain SQL since it doesn't expose it like SQLite, that makes them not behave the same but I think it's fine for now and for the purpose of the toolbar.

### DIFF
--- a/components/toolbar/src/api/statements.ts
+++ b/components/toolbar/src/api/statements.ts
@@ -117,7 +117,6 @@ const getTableColumns = async (
     SELECT
       column_name as name,
       data_type as type,
-      column_default as dflt_value,
       (is_nullable = 'NO') as notnull
     FROM information_schema.columns
     WHERE table_name = '${tableName}';`,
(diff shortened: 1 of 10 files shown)


components/toolbar/src/api/statements.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (
+      err &&
+      !err.message.includes('relation "sqlite_master" does not exist')
+    ) {
+      throw err
+    }
+  }
+  try {
+    await adapter.query({
+      sql: `SELECT 1 FROM information_schema.tables;`,
```
samwillis, line 56: `information_schema` is in the SQL standard, not a PG specific thing, and so SQLite could adopt it in future, would be better to use `pg_catalog.pg_class`.

```suggestion
      sql: `SELECT 1 FROM pg_catalog.pg_class;`,
```

samwillis (approved):
other than one comment LGTM

PR #1267, opened 2024-05-15
chore(ci): use latest setup-protoc action

magnetised:
Pin the version of protoc used to verify the code is in sync with the protobuf definitions and add to .tool-versions to make development easier


protocol/satellite.proto
```diff
@@ -658,3 +658,6 @@ message SatShapeDataBegin {
 // End delimiter for the initial shape data
 message SatShapeDataEnd {
 }
+
+// update to run ci tasks
+// touch: 2024-05-15
```
alco, line 663: What does this mean exactly? Does the file need to be updated or only this line to trigger the CI task? Should the date be updated by hand? It's not clear!

magnetised: blimey. yes, i'll make that clearer... 😄

alco (approved):
:heart:

PR #1268, opened 2024-05-15
chore(toolbar): Remove sourcemap files from published package

msfstef:
Removing sourcemap files from published package after a discussion with @balegas - will reduce the package size on NPM significantly, if anyone wants to debug the toolbar they can do that on the project directly.


balegas (approved):
Trusting that this works

PR #1269, opened 2024-05-15
chore(toolbar): Convert toolbar to card

msfstef:
This sets up some groundwork for light mode - unfortunately light mode also requires doing some codemirror styling as well as fixing the data grid library's theming which seems to be a source of constant issues.


samwillis (approved):
LGTM

PR #1273, opened 2024-05-16
fix(client): Make the throttle skip a snapshot if one is already running

davidmartos96:
This is a behavior we've seen when creating the test for the last PR (#1251). If you add a log to the mock snapshot function, you will see that when stopping, the process waits for a large amount of snapshots to finish before stopping completely.
This is because when polling, they are added to a queue due to the snapshot mutex. 

Is this intentional? Or should the throttle just skip the call if the lock is taken?

cc @kevin-dp @msfstef

### DIFF
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -56,7 +56,7 @@ import {
   toTransactions,
 } from './oplog'
 
-import { Mutex } from 'async-mutex'
+import { Mutex, MutexInterface } from 'async-mutex'
 import Log from 'loglevel'
 import { generateTableTriggers } from '../migrators/triggers'
 import { mergeEntries } from './merge'
@@ -87,9 +87,9 @@ export type ShapeSubscription = {
   synced: Promise<void>
 }
 
-type ThrottleFunction = {
+type ThrottleFunction<T> = {
   cancel: () => void
-  (): Promise<Date> | undefined
+  (): Promise<T> | undefined
 }
 
 type MetaEntries = {
@@ -131,7 +131,12 @@ export class SatelliteProcess implements Satellite {
 
   _pollingInterval?: any
   _unsubscribeFromPotentialDataChanges?: UnsubscribeFunction
-  _throttledSnapshot: ThrottleFunction
+
+  /**
+   * Throttle the snapshot function to avoid calling it too often.
+   * It returns the time of the snaphot or undefined if the snapshot was skipped.
+   */
+  _throttledSnapshot: ThrottleFunction<Date | undefined>
 
   _lsn?: LSN
 
@@ -196,23 +201,51 @@ export class SatelliteProcess implements Satellite {
     this._connectRetryHandler = connectRetryHandler
   }
 
-  _onSnapshotThrottleTick() {
-    if (this.snapshotMutex.isLocked()) {
-      return
-    }
-    return this._mutexSnapshot()
+  _onSnapshotThrottleTick(): Promise<Date | undefined> {
+    return this._mutexSnapshot({ queueUpIfRunning: false })
   }
 
   /**
    * Perform a snapshot while taking out a mutex to avoid concurrent calls.
+   * @param queueUpIfRunning If false and a snapshot is already running, the snapshot will be skipped.
+   * @returns The timestamp of the snapshot or undefined if the snapshot was skipped.
    */
-  async _mutexSnapshot() {
-    const release = await this.snapshotMutex.acquire()
+  async _mutexSnapshot(
+    { queueUpIfRunning } = { queueUpIfRunning: true }
+  ): Promise<Date | undefined> {
+    let release: MutexInterface.Releaser
+
+    if (!queueUpIfRunning) {
+      const maybeRelease = await this.acquireIfFree(this.snapshotMutex)
+      if (!maybeRelease) {
+        // Skip the snapshot
+        return
+      }
+      release = maybeRelease
+    } else {
+      release = await this.snapshotMutex.acquire()
+    }
+
     try {
       return await this._performSnapshot()
     } finally {
       release()
     }
+  }
+
+  /**
+   * Only acquire the mutex if it's not already locked.
+   */
+  private async acquireIfFree(
+    mutex: Mutex
+  ): Promise<MutexInterface.Releaser | null> {
+    // WARNING: We can do this check right before the mutex acquire because there is no async
+    // call until then, so the event loop won't interleave other code.
+    // More context here: https://github.com/electric-sql/electric/pull/1273
+    if (mutex.isLocked()) {
+      return null
+    }
+    return await mutex.acquire()
   }
 
   async start(authConfig?: AuthConfig): Promise<void> {
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -181,7 +181,7 @@ export class SatelliteProcess implements Satellite {
       this._garbageCollectShapeHandler.bind(this)
     )
     this._throttledSnapshot = throttle(
-      this._mutexSnapshot.bind(this),
+      this._onSnapshotThrottleTick.bind(this),
       opts.minSnapshotWindow,
       {
         leading: true,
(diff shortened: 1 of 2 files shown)


kevin-dp:
This seems unintentional and i'm certainly in favor of skipping the periodic snapshot if a snapshot is already in progress.
cc @balegas

balegas:
It sounds good to skip... in the correct conditions :)... 

but I think this code might miss changes: by the time you check the lock is taken, the mutex area is already beyond the change-capturing phase, so you could miss the last new changes that were produced that triggered the snapshot you're evaluating. 

Did I miss something? I haven't looked deeply.

balegas:
Looking at this again. If polling misses a tick because of a race, maybe it's okay, but do you know why you're observing a large amount of snapshots being queued? Is the polling time not configured to be large enough? or is the apply function taking the mutex too agressively?

msfstef:
I agree with @balegas the current change makes it possible for explicit calls to `_throttledSnapshot` to not result in a snapshot - I think it makes sense to add this check for whether the lock is busy in the polling interval alone since missing a polling tick is the same as just having a slightly longer poll interval which is arbitrary either way.

When `_throttledSnapshot` is called explicitly, in the `start` method and as a result of `_handleClientOutboundStarted`, it should be guaranteed that it is queued IMO

balegas:
thanks for clarifying. In the case of a snpashot taking longer than the polling period they would get queued, but they wouldn't run after calling stop because the throttled function is cancelled. Is the behaviour really that they are running?

clients/typescript/src/satellite/process.ts
```diff
@@ -196,6 +196,13 @@ export class SatelliteProcess implements Satellite {
     this._connectRetryHandler = connectRetryHandler
   }
 
+  _onSnapshotThrottleTick() {
+    if (this.snapshotMutex.isLocked()) {
```
kevin-dp, line 200: Looking back at this piece of code, i'm a bit worried about the gap between this check and the point where `await this.snapshotMutex.acquire()` is called in `_mutexSnapshot`. Could another snapshot be started in that gap?
I guess not since the code until we reach the call is synchronous so due to the nature of the JS event loop that can't happen, but i'm not 100% sure, someone should confirm.

kevin-dp: In languages where the mutex could be acquired by someone else in the gap we would definitely need an atomic action that checks whether it's locked and if not acquires it immediately. Even if this can't happen here because of details about how the JS event loop works, i think it is very brittle and can easily be broken later so might be better using an atomic "acquire mutex if free" method. Looking at `async-mutex` i don't think they provide such a method. In which case, i think it is best to have the check immediately followed by acquiring the lock on the next line with a big disclaimer in comment saying why this is needed. Even better would be that we implement a function `acquireIfFree` which does exactly that. This reduces the chances of breaking it later by accidentally introducing an async call during the gap.

kevin-dp: > Yeah, only one piece of JS is executed at any given time, and since there is a straight line without `await` calls from here to he acquire, we should never release the event loop before calling that

Thanks for confirming. Still, i would rather not implicitly rely on this remaining the case, and make sure that it cannot happen by moving that to a dedicated function as described in my last comment. Open to any suggestions/comments/critique on this.

davidmartos96: I've updated the code with the acquireIfFree utility function and a description of why we can do the if check right before.

Which places should use the option `queueUpIfRunning = false` other than the polling interval ?

clients/typescript/src/satellite/process.ts
```diff
@@ -196,18 +203,55 @@ export class SatelliteProcess implements Satellite {
     this._connectRetryHandler = connectRetryHandler
   }
 
+  _onSnapshotThrottleTick(
+    { skipIfRunning } = { skipIfRunning: false }
```
kevin-dp, line 207: shouldn't the default be `true` here?

davidmartos96: Whatever you guys prefer. I used "dont skip as a default" and opt in to the skip with the polling interval.
We can instead opt out in the places you consider.

davidmartos96: You mentioned the other day there could be places where we don't want to miss the snapshot

kevin-dp: I would argue that for periodic snapshots the default of skipping if a snapshot is already in progress is fine. Let's see what @balegas thinks.

davidmartos96: @kevin-dp I agree too to the skip by default on periodic snapshots (polling interval), but the snippet you point out is for the throttled snapshot function, which is used across the process.ts file and tests.
The polling already skips here: https://github.com/davidmartos96/electric/blob/b06b941b9c8f701e34e3f24effaffc0c393d5520/clients/typescript/src/satellite/process.ts#L332

davidmartos96: Maybe _onSnapshotThrottleTick is not the best name because of the "Tick". I wasn't sure how to call it

balegas (changes requested):
I'd like us to give a step back

The idea of the long polling function was to pick changes from anyone that was interacting with the database directly.

Looking at the code, I see that potentialDataChangeHandler is making use of throttledSnapshot. That is wrong, because this code should take a snapshot regardless of that because new changes were just applied. Any debouncing is introducing delay in capturing changes from the database.

Also, I don't think the queuing up of snapshots by the throttling function is the behaviour we intend (I'm aware I did that code :D). If a snapshot takes longer than the polling period, we don't want to queue multiple requests, we just want to call a next one immediately and get back into schedule.

wrt this PR, I think the idea of skipping the long polling call if a snapshot is already ongoing is valid, but I think we need to fix the parts above first and (that will lead to simpler code for the skipping part).

@davidmartos96, I understand that we're stepping up the complexity of this PR. So, I'm happy we take over. At the same time, I'm happy we continue discussing with you if you're motivated to do the change.

clients/typescript/src/satellite/process.ts
```diff
@@ -196,18 +203,55 @@ export class SatelliteProcess implements Satellite {
...
       return await this._performSnapshot()
     } finally {
       release()
     }
   }
 
+  /**
+   * Only acquire the mutex if it's not already locked.
+   */
+  private async acquireIfFree(
```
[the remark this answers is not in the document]
kevin-dp: Indeed, i would be highly in favor of adding this method as it captures the intent and doesn't make it possible for someone to accidentally slip in an async call in the gap between the check and the acquire.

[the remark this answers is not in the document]
msfstef: fwiw I think this would be clearer if rather than being marked as async it is left as synchronous an returns `Promise.resolve(null)` if the mutex is locked - this way there's no ambiguity with regards to any async gaps introduced by making the method itself async.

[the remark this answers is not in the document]
kevin-dp: If you put a big fat comment on top the function signature to explain that it should always remain sync, than this could do. But i think having a dedicated `acquireIfFree` method that acts as an atomic operation is cleaner, or is this just me?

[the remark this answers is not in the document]
davidmartos96: @msfstef @kevin-dp  I updated the signature to this so that it is sync. Is Promise.resolve(null) necessary? Wouldn't that be the same as the previous code? `async fun` + `return null/undefined`
New signature: `private acquireIfFree(mutex: Mutex): Promise<MutexInterface.Releaser> | null` 

@balegas Given there are other things that were discovered, I think it would be easier for communication purposes if you guys take it over then. Feel free to reuse this PR or the unit test provided.

davidmartos96:
Another way to think about this issue would be to always skip the throttled snapshot if the mutex is taken and flag the process with a boolean property if someone requested a snapshot while one was running.
Then, before the mutex is released, run the snapshot again if needed. This would work as long as the caller of the throttled snapshot doesn't need the output result, which looking at the source code, no one does.
So the worst case scenario would be waiting an extra snapshot instead of an arbitrary queue from the mutex.

balegas:
yeah, that makes sense for me too. The polling a bit of a corner case, so I prefer to have simpler code even if we lose a tick in some situations (against my original comment).

KyleAMathews:

PR #1274, opened 2024-05-16
feat(toolbar, client): Add support for new shape sync API on toolbar

msfstef:
- Exposes `SyncStatus` type as well as a method for retrieving all subs and their status.
- Implements new method in the toolbar


samwillis (approved):
LGTM 👍

PR #1276, opened 2024-05-16
feat(client): Shape subscription status notifier

msfstef:
The usual notifier pattern but for shape subscription status

Added callback to shape manager to internalize notifying logic entirely to the shape manager rather than having arbitrary calls within the satellite process

We could add additionall utility to subscribe to particular subscription with a given key but I think for now just having this general notifier method will cover use cases that need it, we don't need to expose/document it.


kevin-dp (approved):
LGTM 👍

PR #1281, opened 2024-05-20
feat (client): flag to disable FKs on incoming TXs in SQLite

kevin-dp:
This PR introduces an additional flag that can be enabled when electrifying a SQLite database. When the flag is enabled, the Satellite process disables FK checks on incoming transactions. This flag is useful for performance reasons as skipping FK checks greatly speeds up initial data sync when there is a lot of data.

Note that disabling FK checks can't be done transactionally so we have to disable FKs, apply the TX, and then re-enable FKs. Therefore, it is important that no transactions are ran concurrently (this is ensured by the mutex that our drivers hold).
We assume that Electric exlusively owns the DB connection and thus that nobody executes a concurrent TX directly on the DB connection.

### DIFF
--- a/clients/typescript/src/config/index.ts
+++ b/clients/typescript/src/config/index.ts
@@ -66,7 +66,7 @@ export type HydratedConfig = {
   debug: boolean
   connectionBackOffOptions: ConnectionBackOffOptions
   namespace: string
-  disableFKs: boolean
+  disableFKs: boolean | undefined
 }
 
 export type InternalElectricConfig = {
@@ -88,7 +88,6 @@ export const hydrateConfig = (
   const auth = config.auth ?? {}
 
   const debug = config.debug ?? false
-  const disableFKs = config.disableFKs ?? false
   const url = new URL(config.url ?? 'http://localhost:5133')
 
   const isSecureProtocol = url.protocol === 'https:' || url.protocol === 'wss:'
@@ -134,6 +133,6 @@ export const hydrateConfig = (
     debug,
     connectionBackOffOptions,
     namespace: defaultNamespace,
-    disableFKs,
+    disableFKs: config.disableFKs,
   }
 }
--- a/clients/typescript/src/config/index.ts
+++ b/clients/typescript/src/config/index.ts
@@ -43,10 +43,10 @@ export interface ElectricConfig {
    */
   connectionBackOffOptions?: ConnectionBackOffOptions
   /**
-   * Whether to disable FK checks when applying incoming (i.e. remote) transactions to the local SQLite database.
+   * Whether to disable FK checks when applying downstream (i.e. incoming) transactions to the local SQLite database.
    * When using Postgres, this is the default behavior and can't be changed.
    */
-  disableFKs?: boolean
+  disableForeignKeysDownstream?: boolean
 }
 
 export type ElectricConfigWithDialect = ElectricConfig & {
@@ -133,6 +133,6 @@ export const hydrateConfig = (
     debug,
     connectionBackOffOptions,
     namespace: defaultNamespace,
-    disableFKs: config.disableFKs,
+    disableFKs: config.disableForeignKeysDownstream,
   }
 }
--- a/clients/typescript/src/config/index.ts
+++ b/clients/typescript/src/config/index.ts
@@ -42,6 +42,8 @@ export interface ElectricConfig {
    * Optional backoff options for connecting with Electric
    */
   connectionBackOffOptions?: ConnectionBackOffOptions
+  /** Whether to disable FK checks when applying incoming (i.e. remote) transactions. */
+  disableFKs?: boolean
 }
 
 export type ElectricConfigWithDialect = ElectricConfig & {
@@ -61,6 +63,7 @@ export type HydratedConfig = {
   debug: boolean
   connectionBackOffOptions: ConnectionBackOffOptions
   namespace: string
+  disableFKs: boolean
 }
 
 export type InternalElectricConfig = {
@@ -73,6 +76,7 @@ export type InternalElectricConfig = {
   }
   debug?: boolean
   connectionBackOffOptions?: ConnectionBackOffOptions
+  disableFKs?: boolean
 }
 
 export const hydrateConfig = (
@@ -81,6 +85,7 @@ export const hydrateConfig = (
   const auth = config.auth ?? {}
 
   const debug = config.debug ?? false
+  const disableFKs = config.disableFKs ?? false
   const url = new URL(config.url ?? 'http://localhost:5133')
 
   const isSecureProtocol = url.protocol === 'https:' || url.protocol === 'wss:'
@@ -126,5 +131,6 @@ export const hydrateConfig = (
     debug,
     connectionBackOffOptions,
     namespace: defaultNamespace,
+    disableFKs,
   }
 }
--- a/clients/typescript/src/config/index.ts
+++ b/clients/typescript/src/config/index.ts
@@ -42,7 +42,10 @@ export interface ElectricConfig {
    * Optional backoff options for connecting with Electric
    */
   connectionBackOffOptions?: ConnectionBackOffOptions
-  /** Whether to disable FK checks when applying incoming (i.e. remote) transactions. */
+  /**
+   * Whether to disable FK checks when applying incoming (i.e. remote) transactions to the local SQLite database.
+   * When using Postgres, this is the default behavior and can't be changed.
+   */
   disableFKs?: boolean
 }
 
--- a/clients/typescript/src/drivers/generic/adapter.ts
+++ b/clients/typescript/src/drivers/generic/adapter.ts
@@ -112,7 +112,6 @@ abstract class DatabaseAdapter
       query: this._query.bind(this),
       transaction: this._transaction.bind(this),
       runInTransaction: this._runInTransaction.bind(this),
-      group: this._group.bind(this),
     }
     return await f(adapter)
   }
--- a/clients/typescript/src/drivers/generic/adapter.ts
+++ b/clients/typescript/src/drivers/generic/adapter.ts
@@ -102,9 +102,9 @@ abstract class DatabaseAdapter
     })
   }
 
-  async _group(
-    f: (adapter: UncoordinatedDatabaseAdapter) => Promise<void> | void
-  ): Promise<void> {
+  async _group<T>(
+    f: (adapter: UncoordinatedDatabaseAdapter) => Promise<T> | T
+  ): Promise<T> {
     // We create an adapter that does not go through the mutex
     // when used by the function`f`, since we already take the mutex here
     const adapter = {
@@ -117,9 +117,9 @@ abstract class DatabaseAdapter
     return await f(adapter)
   }
 
-  group(
-    f: (adapter: UncoordinatedDatabaseAdapter) => Promise<void> | void
-  ): Promise<void> {
+  group<T>(
+    f: (adapter: UncoordinatedDatabaseAdapter) => Promise<T> | T
+  ): Promise<T> {
     return this.txMutex.runExclusive(() => {
       return this._group(f)
     })
--- a/clients/typescript/src/drivers/generic/adapter.ts
+++ b/clients/typescript/src/drivers/generic/adapter.ts
@@ -102,7 +102,7 @@ abstract class DatabaseAdapter
     })
   }
 
-  async _group<T>(
+  async _runExclusively<T>(
     f: (adapter: UncoordinatedDatabaseAdapter) => Promise<T> | T
   ): Promise<T> {
     // We create an adapter that does not go through the mutex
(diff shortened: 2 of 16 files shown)


samwillis (commented):
@kevin-dp I think its possible for a non-satellite transaction to happen between dissabling FK checks and the actual satellite transaction, and again possible afterwards, before they are turned on again.

Our adapter mutex is at the transaction level only. Do we not need another mutex, a connection level one, that satellite can hold during what is in effect multiple transactions?

davidmartos96:
@kevin-dp Do you have some numbers about this performance improvement?

I'm wondering if using  `builder.deferOrDisableFKsForTx` inside the adapter transaction would be equivalent, same as it's currently done in `_applySubscriptionData`. It would solve the concurrency issue.

kevin-dp:
> @kevin-dp I think its possible for a non-satellite transaction to happen between dissabling FK checks and the actual satellite transaction, and again possible afterwards, before they are turned on again.
> 
> Our adapter mutex is at the transaction level only. Do we not need another mutex, a connection level one, that satellite can hold during what is in effect multiple transactions?

You're right, concurrent TXs could occur between the DAL and the Satellite process which could get badly interleaved.
Here's a summary of a discussion with @samwillis about how our drivers currently use mutexes:

- drivers that are sync don't need the mutex in order to implement the interactive transactions safely (better-sqlite3)
- our transaction APIs use the callback style so that in sync code there is no promise/await and so no mutex needed for sync drivers
- async drivers (wa-sqlite, capacitor, tauri, and others) need the mutex for the transactions

All async drivers are built on top of our generic drivers.
The generic driver coordinates all methods `run`/`query`/`runInTransaction`/`transaction` through a mutex.

What we are trying to achieve here is to group `PRAGMA foreign_keys = OFF`, a remote transaction, and `PRAGMA foreign_keys = ON` inside a SQLite transaction. However, SQLite does not allow the `foreign_keys` pragma to occur in a transaction. Hence, to guarantee isolation with other TXs we need to manually coordinate through a mutex. I propose to introduce a `group(f: (uncoordinatedAdapter: Adapter) => void)` method in the adapter interface which can be used to conceptually group several statements that are isolated from other transactions by coordinating through a mutex.

So that the Satellite process could do:
```ts
adapter.group((a) => {
  a.run({ sql: `PRAGMA foreign_keys = OFF` })
  ...
  a.run({ sql: `PRAGMA foreign_keys = ON` })
})
```

We can extend the generic database adapter with the `group` method which starts by acquiring the lock, then calls the provided function (passing along a version of the adapter that does coordinate through the lock), and then releases the lock.

kevin-dp:
> @kevin-dp Do you have some numbers about this performance improvement?

@samwillis did some initial experiments with disabling FKs and has some numbers about the speedup it can provide.

> I'm wondering if using `builder.deferOrDisableFKsForTx` inside the adapter transaction would be equivalent, same as it's currently done in `_applySubscriptionData`. It would solve the concurrency issue.

Do you mean to defer/disable FKs inside every TX and thus calling that inside the adapter's `transaction` method?
The idea is to support a flag that developers can enable to speedup e.g. initial sync of large amounts of data by disabling FKs. If we just defer the foreign key checks to the end of the TX we will not get the speedup we're after.

davidmartos96:
> Do you mean to defer/disable FKs inside every TX and thus calling that inside the adapter's transaction method?

Yes, that's what I meant. I suggested the defer because I thought that a possible performance degradation could be SQLite doing foreign key checks for every INSERT/UPDATE in the Satellite transaction and just deferring all checks to the end would be faster.

kevin-dp:
I added an implementation of `group` on the database adapters and used it in the Satellite process to disable FKs for certain transactions when the `disableFKs` flag is set. There is one weird case which is when we received a remote migration, this calls into the migrator's `applyIfNotAlready` method which had to be changed to also disable the FKs when running SQLite and when the flag is set.

To be honest, this PR adds quite some additional complexity around the drivers for this optimization. I don't think i like it.

samwillis:
@davidmartos96 

The performance issue that this PR is addressing is related to large initial syncs. What we have found is that under some circumstances the time to apply the transaction increases exponentially.

With the linearlite example,  1100 issues + their comments takes under a second (to fully load the app), and has scaled to that point almost linearly. At 1200 it's up to something like 5-10 seconds, and at 1300 it's at over 30. I forget the exact numbers (currently sitting in an airplane awaiting takeoff).

It's directly related to the order of the inserts and if the FK is pointing to a row inserted before or after its own row. I'm suspicious it affects the WASM SQLite specifically, but need to test. I think it's related to doing a full scan for each FK reference, and it may be that there is memory allocations that cause the slow down with WASM.

Deferring FK checks does not solve the problem unfortunately, only turning them off does. However, there is another potential fix, to topologically sort the inserts based on FK, this almost entirely eliminates the slow down, but unfortunately won't solve the problem for circular references.

I'm not sure yet which solution we will go for, maybe both.

samwillis:
@kevin-dp agreed, it's looking quite complex. Maybe we should take a look at the topological sort of the inserts before we make a decision?

davidmartos96:
@samwillis Thanks for the input!

We would be interested to try to reproduce it in the Dart client. But first we would like to reproduce it with the official client. In the linearlite example we've added this. It creates 1500 issues and 1 comment per issue (using [faker](https://www.npmjs.com/package/@faker-js/faker) to get random values).
After inserting everything and replicating to Postgres, if we reset the local database, the browser console says it synced in around 1.5s. Higher values don't seem to make much difference in the sync time either. We are using the wa-sqlite driver in Linearlite.
What's noticeable is that higher values (3000-5000) will leave the issues list blank for a while, even after the inserts where persisted, but that could be something else, as the sync time stays low. 

```tsx
     <button
        onClick={async () => {
          const total = 1500

          for (let i = 0; i < total; i++) {
            console.log(i, 'out of', total)

            const title = faker.company.name()

            const date = new Date()
            const issue_id = uuidv4()
            db.issue.create({
              data: {
                id: issue_id,
                title: title,
                username: 'testuser',
                priority: Priority.NONE,
                status: Status.BACKLOG,
                description: faker.lorem.paragraphs(3),
                modified: date,
                created: date,
                kanbanorder: generateKeyBetween(null, null),
              },
            })

            db.comment.create({
              data: {
                id: uuidv4(),
                issue_id: issue_id,
                body: faker.lorem.paragraphs(1),
                created_at: new Date(),
                username: 'testuser',
              },
            })
          }
        }}
      >
        INSERT RANDOM
      </button>
```

clients/typescript/src/util/transactions.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @returns
+ */
+export async function runInTransaction(
+  adapter: DatabaseAdapter,
+  disableFKs: boolean,
+  ...stmts: Statement[]
+): Promise<RunResult> {
+  return adapter.group(async (uncoordinatedAdapter) => {
+    let enableFKs = false
+    if (disableFKs) {
```
msfstef, line 20: if `disableFKs` is explicitly set to `false`, it should also check if foreign keys are disabled in order to enable them for the transaction and re-enable them after (flag could also be undefined/null to skip this check altogether)

also, would be good to cover these cases with a couple of tests (disable FKs, enable FKs, use existing behaviour)

clients/typescript/src/migrators/index.ts
```diff
@@ -30,7 +30,10 @@ export function makeStmtMigration(migration: Migration): StmtMigration {
 export interface Migrator {
   up(): Promise<number>
   apply(migration: StmtMigration): Promise<void>
-  applyIfNotAlready(migration: StmtMigration): Promise<boolean>
+  applyIfNotAlready(
+    migration: StmtMigration,
+    disableFKs: boolean
```
msfstef, line 35: nit: I personally prefer boolean flags to be part of a configuration object (such that it's easier to tell what is being enabled/disabled when reading it) but it's a preference so feel free to ignore this

clients/typescript/src/drivers/generic/adapter.ts
```diff
@@ -98,6 +96,35 @@ abstract class DatabaseAdapter
...
+    f: (adapter: UncoordinatedDatabaseAdapter) => Promise<T> | T
+  ): Promise<T> {
+    // We create an adapter that does not go through the mutex
+    // when used by the function`f`, since we already take the mutex here
+    const adapter = {
+      run: this._run.bind(this),
+      query: this._query.bind(this),
+      transaction: this._transaction.bind(this),
+      runInTransaction: this._runInTransaction.bind(this),
+      group: this._group.bind(this),
```
msfstef, line 115: is there any reason for `group` to be part of the uncoordinated adapter interface? Since it's uncoordinated `group` doesn't really do anything and could be misleading

clients/typescript/src/drivers/generic/adapter.ts
```diff
@@ -64,22 +58,26 @@ abstract class DatabaseAdapter
...
+      // something went wrong
+      // let's roll back and rethrow
+      return this._run({ sql: 'ROLLBACK' }).then(() => {
+        throw err
       })
-      .finally(release)
+    })
+  }
+
+  async transaction<T>(
```
msfstef, line 75: no suggestion, just wanted to say that I definitely prefer how this looks now :)

clients/typescript/src/config/index.ts
```diff
@@ -42,6 +42,11 @@ export interface ElectricConfig {
    * Optional backoff options for connecting with Electric
    */
   connectionBackOffOptions?: ConnectionBackOffOptions
+  /**
+   * Whether to disable FK checks when applying incoming (i.e. remote) transactions to the local SQLite database.
+   * When using Postgres, this is the default behavior and can't be changed.
+   */
+  disableFKs?: boolean
```
msfstef, line 49: nit: I'm a fan of descriptive config options but I've noticed there might be disagreements on that (e.g. `disableForeignKeyChecksOnReplication` or something like that instead of `disableFKs`) so feel free to ignore me - although we could still keep an abbreviated form inside the code but expose a verbose option for API clarity

msfstef (commented):
Looks good, nice work! I like the refactoring of the `transaction` adapter API and the `group` one is quite simple so I think it works well, left a couple of comments w.r.t. tests and conventions

clients/typescript/src/config/index.ts
kevin-dp: Yes could be `disableForeignKeysDownstream`

clients/typescript/src/util/transactions.ts
kevin-dp: Done, could you please have a look at https://github.com/electric-sql/electric/pull/1281/commits/a0fc10464bf89f80495d081198123879ecc661f4

clients/typescript/src/migrators/index.ts
kevin-dp: i agree with putting configuration options in an object if there are several options that can be configured.
However, if there is only one option, wrapping it in an object seems overkill.

clients/typescript/src/util/transactions.ts
```diff
@@ -3,37 +3,49 @@ import { DatabaseAdapter, RunResult } from '../electric'
...
-    if (enableFKs) {
-      // re-enable FKs
-      await uncoordinatedAdapter.run({ sql: 'PRAGMA foreign_keys = ON;' })
+
+    try {
+      // apply the statements in a TX
+      const res = await uncoordinatedAdapter.runInTransaction(...stmts)
+      return res
+    } finally {
+      // TX throwed, need to restore the pragma
```
msfstef, line 42: nit: finally will run regardless of whether TX throws, comment might be slightly confusing

samwillis:
@kevin-dp If we are going ahead with this could we maybe name the new grouping method `exclusive` rather than `group`?

clients/typescript/src/migrators/bundle.ts
```diff
@@ -133,7 +133,7 @@ export abstract class BundleMigratorBase implements Migrator {
 
   async apply(
     { statements, version }: StmtMigration,
-    disableFKs = false
+    disableFKs?: boolean
```
msfstef, line 136: I'm assuming the inconsistency between `boolean | undefined` and `?: boolean` is because it requires `?` to make the argument optional? If so ignore this comment, just wanted to make sure the `null` case is checked somewhere if we're going from `?` to `type | undefined`

kevin-dp:
> @kevin-dp If we are going ahead with this could we maybe name the new grouping method `exclusive` rather than `group`?

Not a fan of `exclusive` it doesn't rely describe what it does.
What about `groupExclusively` ?

clients/typescript/src/config/index.ts
kevin-dp: I renamed the public facing flag and kept the shorter name internally.

msfstef (approved):

clients/typescript/src/migrators/bundle.ts
kevin-dp: yep, this is subtle. Basically, here i'm allowing the argument to be omitted so you don't have to call it with `undefined`. Inside the body, the `disableFKs` variable will be a `boolean | undefined`. On L144 we then call `runInTransaction` which requires the flag to be provided but it can be the value `undefined`.

kevin-dp:
Agreed with @samwillis and @msfstef to rename `group` to `runExclusively`.

PR #1286, opened 2024-05-21
fix(toolbar): Use shape sub notifier in toolbar

msfstef:
Remove the `setInterval` call used to monitor shapes in the toolbar and use the new notifier method instead.


kevin-dp (approved):
LGTM!

PR #1287, opened 2024-05-21
fix(toolbar): Remount toolbar rather than block it

msfstef:
Instead of guarding against remounting the toolbar, clean it up by removing the whole shadow dom root and remount it.

Main modification required was to deep clone the style from the template rather than just import it.

Alternative fix to https://github.com/electric-sql/electric/pull/1280

### DIFF
--- a/.changeset/polite-donkeys-reflect.md
+++ b/.changeset/polite-donkeys-reflect.md
@@ -2,4 +2,4 @@
 "@electric-sql/debug-toolbar": patch
 ---
 
-Guard against adding toolbar to DOM twice
+Allow toolbar to remount properly on consecutive `addToolbar` calls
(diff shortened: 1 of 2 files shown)


thruflo:
Thanks for this — if it works fine, I think this is a better approach.

msfstef:
@thruflo @samwillis ping to get this either merged or closed, just needs a decision.

I'm happy to make it remount the toolbar, gives it a nice "reset" based idempotency rather than a "check" based one (every `addToolbar` will clear any existing toolbar and add one, always) at the cost of potentially having a jarring toolbar reset while being used, but I'm not particularly invested in either choice.

samwillis:
@msfstef lets merge this as it, if users complain about UX then we come back to it.

samwillis (approved):
👍

PR #1288, opened 2024-05-21
fix(client_test): Improve flakyness of shape lifecycle test 

davidmartos96:
It looks like the new test `notifies for shape lifecycle` can fail with the postgres driver. When expecting 1 notifications, 2 were sent instead. This seems to be because the subscription data is emitted before the `Process.subscribe` ends (there is an `await setMeta` that can take a bit longer than the 1 ms delay the mock uses, when using the Postgres driver. 

I don't know if there is better way to solve this.

cc @msfstef

### DIFF
--- a/clients/typescript/src/satellite/mock.ts
+++ b/clients/typescript/src/satellite/mock.ts
@@ -329,9 +329,7 @@ export class MockSatelliteClient
         setTimeout(resolveProm, 1)
       } else {
         // Otherwise, we resolve the promise before delivering the subscription.
-        // Give enough time for the `Process.subscribe` promise to fully resolve before
-        // delivering the subscription.
-        if (!this.doSkipNextEmit) setTimeout(emit, 50)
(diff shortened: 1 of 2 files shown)


msfstef (commented):
I'm happy with the `notifies for shape lifecycle` test to also sync the shape fully and _then_ assert that there are 2 notifications and check them individually (essentially making sure the order is correct) to avoid timing issues altogether, since there's always a chance that the order messes up because of unpredictable IO.

davidmartos96:
@msfstef That can work too, good point. I've updated it

msfstef (approved):

msfstef:
Thanks for the fix! Should've foreseen that there would be timing issues :)

davidmartos96:
> Thanks for the fix! Should've foreseen that there would be timing issues :)

No problem! A bit of a hidden issue though, since the CI doesn't run the node postgres driver. Our SQLite tests also suffice with the 1 ms delay.

PR #1289, opened 2024-05-21
fix(client): Replace not widely supported regex lookbehind

msfstef:
Issue was first encountered with the website not working properly on iOS 16.2 Safari (see https://github.com/electric-sql/website/issues/134)

Turns out that regex lookbehinds [are not that widely supported](https://caniuse.com/js-regexp-lookbehind), e.g. only supported on iOS since 16.4, which is very recent.

I've replaced it with an uglier, but more supported version - fairly well tested and performance isn't really an issue for this operation, if anyone has a better approach I'm happy to switch it. I'm not quite sure why we're avoiding matching question marks preceded by a word character, as this still will match question marks within a string (e.g. "what's up ?").

@icehaunter tagging as it was your implementation and I want to make sure it doesn't introduce unwanted behaviour.


thruflo (commented):
Looks great to me, thanks for picking this up.

One question: can this be a prompt to run cross browser testing in CI?

msfstef:
@thruflo we have cross-browser testing for our web examples, although they use a fixed version of electric, as well as for the website.

We could either add some extra e2e tests inside the client itself for cross browser testing of the current build or rewire the web example e2e to use the workspace version. Probably the former to avoid running into compatibility issues with breaking changes.

That being said, this issue wouldn't have been caught unless we were testing older versions of browsers as well :/

icehaunter (approved):
Great work!

PR #1290, opened 2024-05-22
fix(electric): Handle connector errors during startup

alco:
Previously, we would ignore any non-raising connector startup failures, failing to even log them.

With this change we make sure all connectors are running and log a fatal error when that's not the case.

Side note: there's an ongoing work in https://github.com/electric-sql/electric/pull/1226 on adding an e2e test to verify the correct output of such errors.


msfstef (approved):
Looks good! Might be worth adding a changeset?

PR #1299, opened 2024-05-24
fix(cli): Use an IPV6-enabled network in CLI's compose file

### LINKED ISSUE #863: electric-sql generate stuck while using Supabase

alco:
This PR changes two things:

1. Removes one of the compose files, reducing duplication. Now we can no longer get into a situation where the sync service definition is updated in one file but not in the other one.
2. Adds an IPv6-enabled network definition to the compose file and makes the electric service use that network.

### Background 

Beyond requiring support in the OS, the network router and the ISP, there are two more problems with using IPv6 for database connections:

* it's not automatically enabled in Docker
* an explicit network configuration needs to be added to the compose file when the sync service is started by `docker compose`

The first problem is relevant to those users who utilize `docker run` to start the sync service. The second problem currently makes our CLI unusable with IPv6 because the compose file bundled with it has no provisions for IPv6 networking. This PR address the latter.

### Testing

I have tested connectivity from the sync service to a Supabase database over IPv6 and to a DigitalOcean Managed PostgreSQL over IPv4. Here's what I've learned:

* Once IPv6 support has been enabled in the Docker daemon, it seems to persist even after the configuration file is removed and the daemon is restarted. For example, with IPv6 disabled in the Docker daemon (such that `docker run ... curlimages/curl -6 https://ifconfig.co/ip` fails to connect to the remote host) it is still possible to define an IPv6-enabled network in a compose file and the services started by `docker compose` will be able to connect to remote hosts over IPv6.
* The compose file may include a minimalistic network definition that only includes the `enabled_ipv6: true` line. But for this to work, the Docker daemon must be configured with an IPv6 default pool. Otherwise, the network definition in the compose file has to include a subnet definition.
* Alternatively, the network definition in the compose file may be marked with `external: true` to instruct `docker compose` to use a user-created Docker network instead of creating a new one.
* The CIDR prefix configured in Docker daemon's default pool(s) and the subnet in the compose file do not need to match.
* An IPv6-enabled network also supports IPv4 connectivity.

In summary, including a default IPv6-enabled network definition in our packaged compose file should work regardless of how users have their Docker daemon configured.

There is a potential for conflict, though. If there's already a Docker network on the machine that has an overlapping subnet with the one we define in our compose file, `docker compose up` will fail. To minimize the chances of conflicts, I have chosen a subnet with randomly generated groups that has the size of only 4 address. To make it fully conflict-proof, the compose file has three environment variables that can be used to either redefine the subnet or instruct `docker compose` to use a different, user-created Docker network by name.

Fixes #863.


clients/typescript/src/cli/docker-commands/docker/compose.yaml
```diff
@@ -7,6 +18,24 @@ configs:
...
+      config:
+        # Subnet definition isn't required if the Docker daemon has a default address pool for
+        # IPv6 addresses configured. However, since that's not the case for Docker
+        # out-of-the-box and since we want to free our users from additional manual
+        # configuration when possible, we include a default subnet definition here that should
+        # work with any Docker daemon configuration.
+        #
+        # The fd00:: prefix is part of the address space reserved for Unique Local Addresses,
+        # i.e. private networks. This is analogous to 192.168.x.x in the IPv4 land.
+        - subnet: fd00::/112
```
alco, line 37: NOTE: this subnet is too wide currently. We need to pick an alternative one that includes a handful of addresses (less than 10) and that is unlikely to conflict with other networks created by the same Docker daemon.

clients/typescript/src/cli/docker-commands/docker/compose.yaml
```diff
@@ -7,6 +18,41 @@ configs:
...
+networks:
+  # An IPv6-enabled network is necessary for the sync service to connect to
+  # databases that are only reachable at IPv6 addresses. IPv4 can also be used as usual with
+  # this network.
+  ip6net:
+    enable_ipv6: true
+    # These two options provide an escape hatch that allows users to define their own Docker
+    # network using `docker network create` and make the services defined in this compose file
+    # use that.
+    external: "${DOCKER_NETWORK_USE_EXTERNAL:-false}"
```
alco, line 30: I need some guidance on where to expose this env var and the other two below to allows users to set them such that they are correctly passed to CLI's invocations of `docker compose`.

clients/typescript/src/cli/docker-commands/docker/compose.yaml
```diff
@@ -7,6 +18,41 @@ configs:
...
+networks:
+  # An IPv6-enabled network is necessary for the sync service to connect to
+  # databases that are only reachable at IPv6 addresses. IPv4 can also be used as usual with
+  # this network.
+  ip6net:
+    enable_ipv6: true
+    # These two options provide an escape hatch that allows users to define their own Docker
+    # network using `docker network create` and make the services defined in this compose file
+    # use that.
+    external: '${DOCKER_NETWORK_USE_EXTERNAL:-false}'
```
alco, line 30: I need some guidance on where to expose this env var and the other two below to allows users to set them such that they are correctly passed to CLI's invocations of `docker compose`.

msfstef: I wonder if this particular env var is necessary? If a user defines a network name I'd assume their intent is already to attach to an external one.

I think it makes sense to try and make this work with defaults as much as possible, and pointing to a guide on setting up electric as part of a docker compose file for more complex cases. I'm suggesting this to keep the CLI from proxying too much Docker configuration as I feel its purpose is to offer a quick way to bootstrap electric with as little work as possible.

Perhaps we exclude the possibility of an external network, designate a unique enough network name (e.g. `electric-cli-network` or `electric-cli-{APP_NAME}-network` to allow for multiple networks using the cli), and only potentialy expose the `DOCKER_NETWORK_SUBNET` with a short note on how to resolve conflicts?

Even the latter I wonder if it would be simpler to have a list of random subnets and catch the known `Pool overlaps` error and retry with a few which should cover the vast majority of cases (and if conflicts persist perhaps it's a unique enough case where a custom docker compose config might be necessary anyway)

what do you think?

msfstef (commented):
Looks great, and nice work with reducing the config to a single file, much easier to maintain!

I've left a comment with regard to configuring the network, as my opinion is that we should aim to make it work by default as much as possible and offload any additional requirements to the user via a guide on setting up a custom docker compose, e.g. by adding a CLI command `electric-sql eject` that copies the docker compose file to the dev's project for them to configure any way they want.

clients/typescript/src/cli/docker-commands/docker/compose.yaml
alco: Thanks for the input, Stefanos!

I want the default config to work for the majority of users but there's always unknown unknowns we just can't anticipate. Cycling through a few subnets seems like too much magic for our own good: it makes the setup different for different users and it's difficult to future-proof completely since error messages outside of our control (e.g. Docker) may change at any point.

Adding the ability to export a docker compose file for manual editing does sounds like a good escape hatch, but it has its downsides too. For one, the CLI needs to get a slightly more complicated API to work with an external compose file. Secondly, some users may prefer to keep all config in one .env file instead of editing a compose file.

To minimize the number of env variables we expose related to IPv6 networking, I'd prefer to leave the ability to specify a custom Docker network over specifying a subnet. Network names that Docker Compose creates are already prefixed by project name, so uniqueness is not a concern here. Or, put differently, the same uniqueness concern is shared by the network name, volume name, and service names. We haven't seen issues with those, so let's not try dwell on this point.

If our network definition clashes with another Docker network on the system, then the most flexible route to resolving the conflict is letting the user define a separate network and then instruct the CLI to use that. Now that I think of it, it would also be useful to provide a way to use the host networking mode for those users who have IPv6 working in their OS -- this will obviate the need for them to mess with Docker IPv6 config.

To sum up, I would rewrite the current network definition in such a way so as to expose only one environment variable: `DOCKER_USE_EXTERNAL_NETWORK="host"|<name of custom network>`. Service definitions should also get the `network_mode` field, conditional on the same environment variable.

msfstef: I see your point, and I had forgotten about compose's prefixing so that should already cover us!

I like the idea of collapsing this configuration to a `DOCKER_USE_EXTERNAL_NETWORK` env var - anyone interested in a custom configuration can take the rest of the responsibility of setting up the network entirely on them and we just provide an easy interface for plugging into that.

As for the `eject` option the idea would be that once you `eject` you no longer rely on the CLI for starting/stopping etc, you take the responsibility of running the docker containers yourself (perhaps with a small README or pointer to our docs for custom/manual deploys).

alco:
I've tested the update implementation in the following scenarios, running a starter app on an Ubuntu DO droplet:
- connect to Supabase using host networking by setting `ELECTRIC_DOCKER_NETWORK_USE_EXTERNAL=host`
- connect to Supabase using a Compose-created network (default behaviour)
- create a standalone Docker network with `docker network create --ipv6 --subnet 2001:db8::/64 ip6net` and connect to Supabase using the setting `ELECTRIC_DOCKER_NETWORK_USE_EXTERNAL=ip6net`
- connect to a Crunchybridge database over IPv4, with `DATABASE_USE_IPV6` set to either `true` or `false`

components/cli/src/docker-commands/docker-utils.ts
```diff
@@ -2,40 +2,75 @@ import { spawn } from 'child_process'
...
-)
+function composeFilePath(filename: string) {
+  return path.join(
+    path.dirname(fileURLToPath(import.meta.url)),
+    'docker',
+    filename
+  )
+}
+
+function useExternalDockerNetwork(opt: string): string {
```
msfstef, line 13: nit: we can use a more descriptive/strict type signature and argument
```ts
function useExternalDockerNetwork(networkName?: string) : 'true' | 'false' {
    return (networkName && networkName.length > 0) ? 'true' : 'false'
}
```
or even return a boolean and let the calling function cast it to an appropriate string (?) I'd avoid the `toString` call cause its behaviour never feels obvious to me

components/cli/src/docker-commands/docker-utils.ts
```diff
@@ -2,40 +2,75 @@ import { spawn } from 'child_process'
...
+}
 
-const composeFileWithPostgres = path.join(
-  path.dirname(fileURLToPath(import.meta.url)),
-  'docker',
-  'compose-with-postgres.yaml'
-)
+// Derive network name from the current working directory, matching Docker Compose's default
+// naming.
+function deriveNetworkName(opt: string): string {
```
msfstef, line 20: nit: type `deriveNetworkName(networkName?: string) : string` or `networkName: string | undefined` or similar if we want to keep argument non-optional

msfstef (changes requested):
Nice work! Looks good and love the new way of configuring the network!

~~Testing it I seem to be having issues with flags and defaults - the `--with-postgres` flag doesn't seem to be working at all no matter how I set it. I saw you raised a seemingly related issue https://github.com/electric-sql/electric/issues/1404, are you seeing the same thing or have I perhaps messed something up in my local build?~~ EDIT: it was a local build issue, runs fine when I make sure I'm running the right CLI build!

Also, we might want to switch the `ELECTRIC_WRITE_TO_PG_MODE` to `direct_writes` default rather than `logical_replication` since we're shifting to that, cause when the `--with-postgres` flag was not detected the error I got was the `LOGICAL_PUBLISHER_HOST` not being set which confused me initially

components/cli/src/docker-commands/docker-utils.ts
alco: But returning a boolean here means using `toString()` later to cast it. I'm not sure what is not obvious about the behaviour of `toString`.

components/cli/src/docker-commands/docker/compose.ip6net.yaml
```diff
@@ -0,0 +1,46 @@
...
+        # The fd00:: prefix is part of the address space reserved for Unique Local Addresses,
+        # i.e. private networks. This is analogous to 192.168.x.x in the IPv4 land.
+        #
+        # There is a possibility that this subnet overlaps with another network configured
+        # separately for the same Docker daemon. That situation would result in `docker compose
+        # up` failing with the error message
+        #
+        #     Error response from daemon: Pool overlaps with other one on this address space
+        #
+        # Users can specify a different subnet via the DOCKER_NETWORK_SUBNET variable to
```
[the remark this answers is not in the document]
alco: Good catch! Updating these.

alco:
> Also, we might want to switch the `ELECTRIC_WRITE_TO_PG_MODE` default to `direct_writes` rather than `logical_replication` since we're shifting to that soon (see #1342 and we've already updated examples and docs to use direct writes), cause when the `--with-postgres` flag was not detected the error I got was the `LOGICAL_PUBLISHER_HOST` not being set which can be confusing (as opposed to the "check your DATABASE_URL" error in `direct_writes` mode) - feel free to ignore this for this PR though

This needs to be addressed separately.

components/cli/src/docker-commands/docker-utils.ts
msfstef: mostly that in JS I'm never sure what the `toString` implementation will result in (e.g. `'t'` and `'f'`) and if we specifically need `'true'` or `'false'` strings having a `flag ? 'true' : 'false'` feels more explicit. Granted for the boolean case it's fairly standard so feel free to ignore me, I just always start from a "don't trust JS to have standard behaviour" standpoint :P

msfstef:
@alco fwiw I think this is ready to merge, if you're happy with it too!

PR #1303, opened 2024-05-30
feat(linearlite): Linearlite with permissions

msfstef:
Addition of read-only permissions to Linearlite project based on https://github.com/electric-sql/electric/pull/768


examples/linearlite/db/migrations/01-create_tables.sql
```diff
@@ -1,42 +1,146 @@
...
+  TO ("comment", 'owner');  
+
+
+
 -- ⚡
 -- Electrify the tables
-ALTER TABLE project ENABLE ELECTRIC;
-ALTER TABLE issue ENABLE ELECTRIC;
-ALTER TABLE comment ENABLE ELECTRIC;
+ALTER TABLE "profile" ENABLE ELECTRIC;
```
magnetised, line 143: I think you should swap the electrification and permissions order. In the future I think we should add checks on things like `ASSIGN` statements to validate that the referenced columns actually exist, and we can only do that if the table has been electrified.

msfstef: I was wondering what the "correct" order should be - do you think perhaps we should be rejecting a migration doing any `ASSIGN` to an unelectrified table with an explanatory error message already to build "good habits"? If electrification exposes info that permissions needs it makes sense to enforce this as soon as possible IMO

KyleAMathews:

PR #1304, opened 2024-05-30
feat(client): Add ELECTRIC_FEATURES option to CLI for ease of use

msfstef:
As https://github.com/electric-sql/electric/pull/768 adds the read-only permissions as an experimental feature enabled via the `ELECTRIC_FEATURES` environment variable, I found it useful to have it in the CLI as well.

Will open respective docs PR as well if people agree with adding this env var to the CLI.


samwillis (approved):
Agreed, all env vars from the server should be configurable with the cli.

samwillis:
@msfstef Whatever makes sense to you. Obv can merge until his is.

msfstef:
Will merge directly to main, we already have this `ELECTRIC_FEATURES` option in code so might as well decouple it

PR #1313, opened 2024-06-03
chore(client: Fix lint issues in TS client (test files etc)

msfstef:
I've fixed the linting issues in the various test files and edited the eslint config and npm script such that both source and test files are checked in CI to avoid getting in this situation again.

I've also disabled the `@typescript-eslint/no-inferrable-types` rule as I find it useful to always have explicit typings rather than rely on type inferral.


samwillis (approved):
👍 LGTM

I'm happy to disable `@typescript-eslint/no-inferrable-types`, but should check with @kevin-dp

kevin-dp:
Yep, i agree with disabling `@typescript-eslint/no-inferrable-types`, i have encountered instances of this rule that forced me to remove the type even though i'd prefer having them explicitly.

clients/typescript/src/util/debug/index.ts
```diff
@@ -14,8 +14,8 @@ export const setLogLevel = (level: LogLevelDesc = 'TRACE') => {
 /** True if current running environment is NodeJS */
 export const isNode =
   typeof process !== 'undefined' &&
-  process.versions != null &&
```
kevin-dp, line 17: Good catch, `null == undefined` is `true`. Nasty!! :smirk:

kevin-dp (approved):
Great, this is a good improvement to ensure the quality of our test code.

clients/typescript/src/util/debug/index.ts
msfstef: the fact that the `eqeqeq` rule is not a lint error by default is insane!

PR #1315, opened 2024-06-03
fix(electric): Fix temporary replication slot going over character limit

msfstef:
Addresses [this issue](https://linear.app/electric-sql/issue/VAX-1888/temporary-replication-slot-name-too-long-crashes-electric)

This is a quick fix to avoid failures because of long names but the longer term solution is to ensure the replication slot names are smaller in general, see issue description for suggestions.


alco (approved):
:heart:

PR #1316, opened 2024-06-03
feat(cli): Use canary Docker image when using canary CLI client

msfstef:
This has been discussed but when using the canary version of the client it's currently not clear that you should also manually set the `ELECTRIC_IMAGE` to the canary tag as well - the proposed solution is to make the CLI automatically detect that and use the appropriate image.

The `LIB_VERSION` generated by the changesets does not update for the canary releases so I had to use the `package.json` directly, but since this is for the CLI only (which we are also splitting out of the client) that should be fine, it's not meant to be browser-runnable code.

This is mostly a quality of life improvement for us developing electric as the current requirement to remember to point to canary releases for both the sync service and client is slightly annoying.


clients/typescript/src/cli/util/version.ts
```diff
@@ -0,0 +1,8 @@
+import { version } from '../../../package.json'
```
samwillis, line 1: I'm nervous of relying on importing a JSON file, this is the cli and it shouldn't be a problem but it's safer to open the file and parse it, unlikely to break on some platforms. This is how we get the app name in the CLI: https://github.com/electric-sql/electric/blob/27ac2a149e2acf2d0293fbedfa09fba1ac0fbd46/clients/typescript/src/cli/util/parse.ts#L8-L10

samwillis (approved):
Other than the one comment all good 👍

PR #1317, opened 2024-06-03
chore (cli): extract cli to separate package

kevin-dp:
This PR extracts the CLI out of the `electric-sql` package and into its own `@electric-sql/cli` package.

TODO:
- We may want to add `.js` extension to every import to avoid having to run the `fix-imports.js` script.


components/cli/src/client/generateTestClient.ts
```diff
@@ -77,7 +77,18 @@ model Dummy {
 const thisDir = path.dirname(fileURLToPath(import.meta.url))
 const prismaSchemaDir = path.join(thisDir, 'prisma')
 const prismaSchemaPath = path.join(prismaSchemaDir, 'schema.prisma')
-const generatedClientDir = path.join(thisDir, 'generated')
+const generatedClientDir = path.join(
```
msfstef, line 80: nit: might be worth reading `process.argv` for a path of a prisma schema and for the output path of the generated cilent to make this more useable (does this currently work? I see it's being called from the ts-client but it's a bit odd)

we can also make it an executable of the CLI as it's generally a useful tool to be able to generate a test client from a prisma schema alone.

components/cli/.eslintrc.cjs
```diff
@@ -0,0 +1,23 @@
+module.exports = {
```
msfstef, line 1: nit: I wonder if, now that we're splitting things out, we should have common base eslint and prettier configs that we extend rather than duplicating them - e.g. I've added the `eqeqeq` triple equality rule in another PR, I think it's important enough that it should propagate to all projects and not rely on us remembering to include it in the other configs

components/cli/scripts/fix-imports.js
```diff
@@ -0,0 +1,117 @@
+import * as fs from 'fs'
```
msfstef, line 1: nit: perhaps we also want to factor this out as a common utility between packages (although really I'd like us to find a way to not need this altogether haha) - maybe as an unpublished pnpm package?

components/cli/tsconfig.json
```diff
@@ -0,0 +1,34 @@
+{
```
msfstef, line 1: nit: same as for eslintrc and prettier, we might want to have a shared base one for our projects that we extend

msfstef (approved):
Nice work, looks good! Left some minor comments mostly relating to having some way to share configs (eslint, prettier, tsconfig) and utilities (like `fix-imports.js`) which can be addressed in a separate PR

components/cli/scripts/fix-imports.js
samwillis: This landed in #659 and was also discussed in the abandoned #642

#642 would have goine with the "use .js imports everywhere route". Personally I would go that way with any new package (it's what I've done with PGlite) but there was push back and we landed on this somewhat ugly hack.

components/cli/src/client/generateTestClient.ts
kevin-dp: So, generating a client is a feature of the CLI.
Our electric-sql typescript package needs to generate a test client, but we don't want to make our electric-sql package dependent on the cli package because that would introduce a cyclic dependency. So the only solution i saw was to call this file from the electric-sql package. It's a bit hacky but essentially we can't move this file to electric-sql package without introducing the cyclic dependency.

I agree we should make this script more useful by being able to customize path of prisma schema and output path, but i think that's best done in a separate, follow-up PR.

components/cli/.eslintrc.cjs
```diff
@@ -1,23 +1,2 @@
...
-    '@typescript-eslint/ban-ts-comment': [
-      'error',
-      { 'ts-ignore': 'allow-with-description' },
-    ],
-    'no-constant-condition': ['error', { checkLoops: false }],
-  },
-  ignorePatterns: ['**/*/mock.ts'],
-}
+const config = require('../../common/.eslintrc.cjs')
+module.exports = config
```
msfstef, line 2: nit: I'm pretty sure you can use a regular config and do
```
extends: ['../../common/.eslintrc.cjs']
```

msfstef: Ah, ignore me then!

components/cli/scripts/fix-imports.js
kevin-dp: Agree with Sam, i would prefer the use .js imports route. I might update all imports in the CLI package after rebasing and including the changes made to the CLI in concurrent PRs.

PR #1318, opened 2024-06-03
chore(examples): Update examples to latest electric

msfstef:
I've updated most examples (left Ionic one as we might deprecate it, the checkout as it will also require a deploy, and linearlite cause it has multiproject, multiuser PRs open that use updated electric)

I've manually tested these against the canary as well but wanted this PR to be in a mergeable state.


samwillis (approved):
All looks great, thanks!

PR #1320, opened 2024-06-04
chore(docs): Add `ELECTRIC_FEATURES` environment variable to CLI

msfstef:
To be merged once https://github.com/electric-sql/electric/pull/1304 lands in main.

### DIFF
--- a/docs/api/_ELECTRIC_FEATURES.md
+++ b/docs/api/_ELECTRIC_FEATURES.md
@@ -1,0 +1,1 @@
+Flags to enable or disable features. Configurable via a PATH-like syntax, where flag settings are separated by colons.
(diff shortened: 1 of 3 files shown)


samwillis:
@msfstef @magnetised is there a docs PR for the new `ELECTRIC_FEATURES` env var? These cli docs for the env var should link to the docs for the server env var and how it can be used.

msfstef:
@samwillis you're right I don't see this in the docs anywhere but it is documented in code:
https://github.com/electric-sql/electric/blob/25523d96c0e73f3c80838b00d8b3764bed472615/components/electric/lib/electric/features.ex#L11-L18

samwillis:
@msfstef Maybe this PR can add both parts of the docs?

KyleAMathews:

PR #1323, opened 2024-06-04
fix(cli): Redact passwords in docker compose config CLI printout

msfstef:
Redacts any value marked as confidential in the configuration options from the config.

I've also removed the version from the docker compose files as it is deprecated and shows warnings any time we use them - I'd also collapse them in one as @alco did but it can wait until https://github.com/electric-sql/electric/pull/1299 is merged

### DIFF
--- a/clients/typescript/src/cli/config-options.ts
+++ b/clients/typescript/src/cli/config-options.ts
@@ -142,6 +142,7 @@ export const configOptions: Record<string, any> = {
     valueType: String,
     inferVal: (options: ConfigMap) => inferDbUrlPart('password', options),
     defaultVal: 'db_password',
+    confidential: true,
     groups: ['database'],
   },
   DATABASE_NAME: {
@@ -249,6 +250,7 @@ export const configOptions: Record<string, any> = {
       Password to use when connecting to the Postgres proxy via psql or any other 
       Postgres client.
     `,
+    confidential: true,
     groups: ['electric', 'client', 'proxy'],
   },
   AUTH_MODE: {
--- a/clients/typescript/src/cli/config.ts
+++ b/clients/typescript/src/cli/config.ts
@@ -16,6 +16,7 @@ export interface AnyConfigOption {
     | boolean
     | ((options: ConfigMap) => string | number | boolean)
   constructedDefault?: string
+  confidential?: true
   groups?: Readonly<string[]>
 }
 
@@ -203,15 +204,16 @@ function redactConfigValue(config: Config, stringToRedact: string): Config {
  * Redacts sensitive information like secrets and passwords from the
  * config and returns a separate, redacted version.
  *
- * Redaction is done based on keys that contain "PASSWORD" or "SECRET".
+ * Redaction is done based on the `confidential` property of the
+ * configuration option.
  */
 export function redactConfigSecrets(config: Config): Config {
-  const passKeys = Object.keys(config).filter(
-    (k) => k.includes('PASSWORD') || k.includes('SECRET')
+  const keysToRedact = Object.keys(config).filter(
+    (k) => configOptions[k].confidential
   )
   let redactedConfig = { ...config }
-  for (const passKey of passKeys) {
-    redactedConfig = redactConfigValue(redactedConfig, config[passKey])
+  for (const keyToRedact of keysToRedact) {
+    redactedConfig = redactConfigValue(redactedConfig, config[keyToRedact])
   }
   return redactedConfig
 }
--- a/clients/typescript/src/cli/config.ts
+++ b/clients/typescript/src/cli/config.ts
@@ -183,6 +183,39 @@ export function envFromConfig(config: Config) {
   )
 }
 
+/**
+ * Redacts the given `stringToRedact` from the `config`, returning
+ * a new config with any mention of `stringToRedact` redacted.
+ */
+function redactConfigValue(config: Config, stringToRedact: string): Config {
+  const redactedConfig = { ...config }
+  Object.entries(redactedConfig).forEach(([key, value]) => {
+    if (typeof value === 'string' && value.includes(stringToRedact)) {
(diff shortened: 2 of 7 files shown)


clients/typescript/src/cli/docker-commands/docker/compose-with-postgres.yaml
```diff
@@ -1,5 +1,3 @@
-version: '3.8'
```
alco, line 1: The warning was added in Docker Compose v2.25.0 released on 2024-03-15. I'm not sure if earlier versions of Compose require this line but it was definitely the case in the early days of Compose.

I'd err on the side of keeping our compose files compatible with older versions of Compose that a lot people may still be using, due to running an LTS OS, for example.

clients/typescript/src/cli/config.ts
```diff
@@ -183,6 +183,39 @@ export function envFromConfig(config: Config) {
...
+  return redactedConfig
+}
+
+/**
+ * Redacts sensitive information like secrets and passwords from the
+ * config and returns a separate, redacted version.
+ *
+ * Redaction is done based on keys that contain "PASSWORD" or "SECRET".
+ */
+export function redactConfigSecrets(config: Config): Config {
```
alco, line 210: Correct me if I'm reading this wrong but this implementation seems to be doing a lot of redundant work:
- iterate over all config keys to find the ones containing the words 'PASSWORD' and 'SECRET'
- clone the whole config object
- for each secret key
  * clone the whole config object again
  * iterate over all config entries, looking for string matches agains the value of the secret key for the current loop iteration
  * using `split()` and `join()` instead of `replaceAll()`

I would suggest using one of the following approaches instead:
1. Write a custom print function for the config. So instead of passing it to `console.log`, we'd invoke a custom function, say, `dumpConfig(config)`. Inside that function, we'd have a loop over config entries and would replace sensitive strings prior to printing them. This obviates the need to copy the config object.
2. Construct a list of sensitive values, not sensitive keys. Then clone the config object once and loop over its entries, replacing sensitive values with `replaceAll()`. This removes the need to do the `config[passKey]` lookup and reduces the number of times the config object is cloned.

clients/typescript/src/cli/config-options.ts
```diff
@@ -142,6 +142,7 @@ export const configOptions: Record<string, any> = {
     valueType: String,
     inferVal: (options: ConfigMap) => inferDbUrlPart('password', options),
     defaultVal: 'db_password',
+    confidential: true,
```
alco, line 145: "sensitive" or "secret" seems more idiomatic to me.

alco (changes requested):

clients/typescript/src/cli/config.ts
msfstef: Thank you for the suggestions!

- The function comment is wrong, it's not finding keys with `PASSWORD` or `SECRET` that was my first draft and I forgot to update the docstring whoops - the values to redact are found through filtering the `configOptions` dict to find keys marked as `confidential`
- I clone the config multiple times to favour immutability, since performance for this use case is not particularly relevant (config should never get as large so as for this to matter!)
- I forgot about `replaceAll`! The `split` and `join` solution looked ugly so thanks for catching that :)

Since it's finding the values to redact through the config options it's already constructing a list of sensitive keys matching to the sensitive values, the lookups need to happen once for each key anyway whether I construct the list of values at the start or during the redaction process.

Overall if the config has N entries and M sensitive values we go through NxM checks, and we _could_ save a few clones but I really don't see the need to do that for the sake of performance here?

I'll construct the list of values at the start, use `replaceAll` and make `redactConfigValue` mutate the provided config object in place, such that now there is a single config cloning and NxM checks for redaction. I'm reluctant to do a `dumpConfig` print function, we can introduce one for pretty printing it later, but I prefer the redaction to be independent from the printing (for easier testability as well)

clients/typescript/src/cli/config-options.ts
msfstef: I was scratching my head trying to decide what word to use haha - I'll go with "secret"

clients/typescript/src/cli/docker-commands/docker/compose-with-postgres.yaml
msfstef: According to [this article](https://nickjanetakis.com/blog/docker-tip-51-which-docker-compose-api-version-should-you-use) not defining a version was considered "legacy" between 2016-2020, but since late 2020 it conforms to the [spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element-obsolete) which has the version as obsolete (see [release notes](https://docs.docker.com/compose/release-notes/#1270) for spec matching).

I'm torn because I agree that it's better to err on the side of backwards compatibility but with the new version of compose (which most people will use) you will be getting deprecation warnings when running Electric which is not a great look/experience.

alco: Right, I have found [this commit](https://github.com/compose-spec/compose-spec/commit/d9e2265dddc1d67b442138eafe03f6e75552f575#diff-bc6661da34ecae62fbe724bb93fd69b91a7f81143f2683a81163231de7e3b545) which changed the description of `version` from required to optional in the spec. It's from 2020. So let's indeed remove it from our compose file.

clients/typescript/src/cli/config.ts
alco: > Overall if the config has N entries and M sensitive values we go through NxM checks, and we could save a few clones but I really don't see the need to do that for the sake of performance here?

I agree with you, performance is not a concern here. I just found the algorithm confusing due to the seemingly superfluous config cloning. It's probably a case of when looking at the same problem from a different perspective, I got different ideas from yours.

alco (approved):
Looks great!

PR #1329, opened 2024-06-05
chore(electric): revert removal of electric_user_id filtering "hack"

magnetised:


alco:
@magnetised It would be nice to learn about the rationale for this change.

components/electric/lib/electric/features.ex
```diff
@@ -90,6 +90,14 @@ defmodule Electric.Features do
     :ok
   end
 
+  def list(name \\ __MODULE__) do
+    name
+    |> :ets.match(:"$1")
+    |> Enum.into(%{}, fn [f] -> f end)
+    |> Map.merge(Process.get({__MODULE__, name}, %{}))
+    |> Enum.to_list()
```
alco, line 98: If the intent here is to get all table entries, I find `:ets.tab2list()` to be more descriptive. And the temporary conversion to map is not strictly necessary here:

```suggestion
    features = :ets.tab2list(name)
    proc_overrides = Process.get({__MODULE__, name}, %{})
    
    Enum.reduce(proc_overrides, features, fn {key, val}, features ->
      Keyword.put(features, key, val)
    end)
```

components/electric/lib/electric/replication/changes/ownership.ex
```diff
@@ -0,0 +1,47 @@
+defmodule Electric.Replication.Changes.Ownership do
+  alias Electric.Replication.Changes
+  alias Electric.Satellite.Auth
+
+  @user_id_column "electric_user_id"
+
+  @spec id_column_name() :: Changes.db_identifier()
+  def id_column_name, do: @user_id_column
```
alco, line 8: I'd use the same name for the function here, i.e. `user_id_column`. I had to look this implementation up to learn what `id_column_name()` meant when it was referenced in another module.

components/electric/lib/electric/replication/changes/ownership.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  @doc """
+  Given a change, checks if belongs to the given user_id.
+
+  For a change to belong to a user with id `user_id` the row must satisfy one of the conditions:
+
+  1. The table being modified has an `#{@user_id_column}` and the value of this column either
```
alco, line 15:
```suggestion
  1. The table being modified has an `#{@user_id_column}` column and the value of this column either
```

components/electric/lib/electric/replication/changes/ownership.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  def change_belongs_to_user?(%Changes.Migration{}, _user_id, _owner_column) do
+    true
+  end
+
+  @empty [nil, ""]
+
+  @spec validate_record(Changes.record(), Auth.user_id(), Changes.db_identifier()) :: boolean()
+  defp validate_record(record, user_id, owner_column) do
+    case record do
```
alco, line 40: It's a matter of taste, I guess, but `case Map.get(record, owner_column) do` would yield less repetition below.

alco (approved):

components/electric/lib/electric/replication/changes/ownership.ex
magnetised: This ownership file is just being un-deleted, and has a limited lifespan, so though I appreciate your feedback, I'm just going to leave as-is.

components/electric/lib/electric/features.ex
icehaunter: Or you could do a list conversion which is also super simple

```suggestion
    overrides =
      {__MODULE__, name}
      |> Process.get(%{})
      |> Map.to_list()

    name
    |> :ets.tab2list()
    |> Keyword.merge(overrides)
```

magnetised: Thanks for the reminder re tab2list. I could re-implement this as you say but I personally don't see any great advantage to your alternative versions considering the test-only role of this function.

PR #1330, opened 2024-06-05
chore (docs): document disableForeignKeysDownstream flag

kevin-dp:


davidmartos96:
I feel like it could be slightly reworded. With the current wording to me it's not clear whether on postgres foreign keys are disabled or enabled during replication.

docs/api/clients/typescript.md
```diff
@@ -135,6 +135,11 @@ const config: ElectricConfig = {
   Configuration of the backoff strategy used when trying to reconnect to the Electric sync service after a failed
   connection attempt.
 
+- `disableForeignKeysDownstream?: boolean`
+
+  Can be set to true to disable FK checks when applying downstream (i.e. incoming) transactions to the local SQLite database.
+  This is the default behavior when using Postgres and can't be changed.
```
alco, line 141: Does it mean "when using a Postgres driver"? Just saying Postgres seems ambiguous.

davidmartos96:
@kevin-dp Yes, that makes it clearer. Thanks!

alco (approved):
:+1:

PR #1331, opened 2024-06-06
docs: unsubscribe, sync status, and keyed sync

icehaunter:


docs/api/clients/typescript.md
```diff
@@ -202,10 +212,87 @@ await sync2;
...
+
+Every Electric client contains a `sync` property that allows control over the sync process and status.
+
+```ts
+const { db, sync } = await electrify()
+```
+
+### `sync.unsubscribe` method
+
+Using the key returned from the `.sync()` call you can cancel a previously established subscription using a `.unsubscribe(keys)` method on the top-level `sync` object:
```
alco, line 249:
```suggestion
Using the key returned from the `.sync()` call you can cancel a previously established subscription using the `.unsubscribe(keys)` method on the top-level `sync` object:
```

docs/api/clients/typescript.md
```diff
@@ -202,10 +212,87 @@ await sync2;
...
+
+```ts
+const { db, sync } = await electrify()
+const { key, synced } = await db.projects.sync()
+
+sync.unsubscribe([key])
+```
+
+Unsubscribing will cause any rows on the device that were part of this subscription (but not any others) to be removed. Returns a promise that resolves as soon as the server accepts the unsubscribe.
```
alco, line 258: I would move this last sentence to just above the code snippet. It feels lost here, after the remark on the behaviour of unsubscribing.

docs/api/clients/typescript.md
```diff
@@ -69,6 +70,15 @@ type Statement = {
...
   args?: BindParams;
 };
+
+interface SyncManager {
+  unsubscribe(keys: string[]): Promise<void>
+  syncStatus(key: string): SyncStatus
+}
+
+type SyncStatus =
+  | undefined
```
alco, line 80: This may just be my ignorance of TypeScript's type system but it seems more appropriate to me to make this `undefined` part of the return type of the `syncStatus()` method, while the `SyncStatus` type itself becomes simpler and non-nullable.

docs/api/clients/typescript.md
```diff
@@ -202,10 +212,87 @@ await sync2;
...
+```
+
+Unsubscribing will cause any rows on the device that were part of this subscription (but not any others) to be removed. Returns a promise that resolves as soon as the server accepts the unsubscribe.
+
+### `sync.syncStatus` method
+
+Using the key returned from the `.sync()` call you can check the status of a subscription. It returns either `undefined` if that key is not known, or an object with a status of a subscription.
+
+```ts
+export type SyncStatus =
```
alco, line 265: Is this a different type from `SyncStatus` added at the top of this diff? Kinda confusing.

docs/api/clients/typescript.md
```diff
@@ -202,10 +212,87 @@ await sync2;
...
+  | undefined
```
alco, line 266: Same remark here as I left on the other `SyncStatus` type above. Why not make the type of `sync.syncStatus` read `SyncStatus | undefined`, so that the `SyncStatus` type itself becomes non-nullable, i.e. easier to reason about?

alco (approved):
:muscle:

docs/api/clients/typescript.md
icehaunter: I don't want to do that because `SyncStatus` explicitly lists what you need to deal with as valid result options. I can pull `undefined` out, but it feels more descriptive to have this in the type as one of the valid values, not "or nothing" tucked on at the return of a particular function

docs/api/clients/typescript.md
icehaunter: unified

PR #1332, opened 2024-06-06
chore (drivers): extract drivers to a separate package

kevin-dp:


msfstef (approved):
Looks great, nice work!

My only concern is with the extraction of some utilities, like the `QualifiedTablename` class, `parseTableNames`, and `parseDate`, which end up being core to the client as well but they belong to the drivers package, despite them only being used in tests or not used at all in the whole of the driver package.

I wonder if, since they are so few as well, we might benefit from keeping them in the client and re-implementing thinner abstractions for the drivers only?

PR #1337, opened 2024-06-06
chore(electric): Bundle the latest tzdata database into sync service's Docker image

alco:
Since our Docker-based distribution of the sync service is stateless, every time a Docker container is started, it downloads the same tzdata database from a remote host.

To avoid doing that, we bundle the latest database file with the release.

### DIFF
--- a/components/electric/Dockerfile
+++ b/components/electric/Dockerfile
@@ -24,6 +24,7 @@ COPY mix.* /app/
 RUN mix deps.get
 RUN mix deps.compile
 
+RUN mix run --no-compile --no-start -e "Application.ensure_all_started(:tzdata); Tzdata.ReleaseUpdater.poll_for_update()"
 COPY config/*runtime.exs /app/config/
 COPY lib /app/lib/
 COPY priv /app/priv


icehaunter (approved):
Good idea, but couldn't we pull it in as a docker step instead of putting it into git?

alco:
@icehaunter That would require running Elixir during Docker build and waiting for tzdata to do its network lookup.

Ideally, a new version of tzdata that includes the updated database is published on Hex. That was the original promise by the author but, as of today, the library hasn't been updated for almost 3 years. Forking it would also be suboptimal because fetching Mix dependecies from GitHub is way slower than from Hex.

So this works well as a quick fix for us. I'll ping the maintainer of tzdata separately regarding publishing a new version.

icehaunter:
@alco We're already running Elixir during docker build in the form of `mix compile`. It's actually as easy as

```diff
diff --git a/components/electric/Dockerfile b/components/electric/Dockerfile
index 148498eb2..d1dc40352 100644
--- a/components/electric/Dockerfile
+++ b/components/electric/Dockerfile
@@ -24,9 +24,10 @@ COPY mix.* /app/
 RUN mix deps.get
 RUN mix deps.compile
 
+COPY priv /app/priv
+RUN mix run --no-compile --no-start -e "Application.ensure_all_started(:hackney); Application.ensure_all_started(:tzdata); Tzdata.DataBuilder.load_and_save_table()"
 COPY config/*runtime.exs /app/config/
 COPY lib /app/lib/
-COPY priv /app/priv
 COPY src /app/src/
 
 ARG ELECTRIC_VERSION=local
 ```
 
 and saves us any hassle updating it going forward

icehaunter:
While on the other hand adding this to Git makes all checkouts in GH actions slower by 2mb for no  reason

alco:
@icehaunter I have modified your diff slightly to call the higher level `poll_for_update()` function.

PR #1339, opened 2024-06-06
Check permissions for incoming operations

magnetised:
As part of the move to read-only permissions this adds in permissions checks for writes from the clients. With this in place and a later pr to disable GRANTing write permissions, we can enable read-only permissions without needing to do the full implementation of scoped write permissions (which we need to figure out). If you can't grant write perms then we don't need to resolve them.

Note for reviewers: The permissions feature flag naming ceases to make sense with this pr, but that's all sorted in the next one, #1347

### DIFF
--- a/components/electric/lib/electric/satellite/permissions/write.ex
+++ b/components/electric/lib/electric/satellite/permissions/write.ex
@@ -1,0 +1,59 @@
+defmodule Electric.Satellite.Permissions.Write do
+  alias Electric.Replication.Changes
+  alias Electric.Satellite.Permissions
+  alias Electric.Satellite.Protocol
+
+  defmodule Error do
+    defstruct [:reason, :tx]
+
+    @type t() :: %__MODULE__{reason: String.t(), tx: Changes.Transaction.t()}
+
+    def error_response(%__MODULE__{} = error) do
+      %Electric.Satellite.SatErrorResp{
+        error_type: :PERMISSION_DENIED,
+        lsn: error.tx.lsn,
+        message: error.reason
+      }
+    end
+
+    defimpl String.Chars do
+      def to_string(error) do
+        "Changes in transaction [LSN #{error.tx.lsn || "?"}] " <>
+          "were refused for the following reason: " <>
+          error.reason
+      end
+    end
+  end
+
+  def validate(txns, %Protocol.State{} = state) when is_list(txns) do
+    if Permissions.filter_reads_enabled?() do
+      %{permissions: permissions, out_rep: %{sent_rows_graph: srg}} = state
+      graph_impl = Electric.Replication.ScopeGraph.impl(srg)
+      validate_txns(txns, permissions, graph_impl, [])
+    else
+      {:ok, state.permissions}
+    end
+  end
+
+  def error_response(%{message: message, tx: tx}) do
+    %Electric.Satellite.SatErrorResp{
+      error_type: :PERMISSION_DENIED,
+      lsn: tx.lsn,
+      message: message
+    }
+  end
+
+  defp validate_txns([], permissions, _graph, _acc) do
+    {:ok, permissions}
+  end
+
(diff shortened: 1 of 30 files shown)


components/electric/lib/electric/satellite/permissions/write.ex
```diff
@@ -0,0 +1,59 @@
...
+  def error_response(%{message: message, tx: tx}) do
+    %Electric.Satellite.SatErrorResp{
+      error_type: :PERMISSION_DENIED,
+      lsn: tx.lsn,
+      message: message
+    }
+  end
+
+  defp validate_txns([], permissions, _graph, acc) do
+    {:ok, permissions, Enum.reverse(acc)}
```
alco, line 47: What's the point of returning the list of transactions to the caller of `validate()` in this successful case? It looks like the same exact list as the one originally passed in? Am I missing something?

components/electric/lib/electric/satellite/protocol.ex
```diff
@@ -467,11 +498,12 @@ defmodule Electric.Satellite.Protocol do
...
 
-          case WriteValidation.validate_transactions!(complete, {SchemaCache, state.origin}) do
-            {:ok, accepted} ->
-              {nil, send_transactions(accepted, incomplete, state)}
-
-            {:error, accepted, error, trailing} ->
+          with {:ok, accepted} <-
+                 WriteValidation.validate_transactions!(complete, {SchemaCache, state.origin}),
+               {:ok, permissions, accepted} <- Permissions.Write.validate(accepted, state) do
+            {nil, send_transactions(accepted, incomplete, %{state | permissions: permissions})}
```
alco, line 504: Are filtering some transactions out between the calls to `validate_transactions!()` and `validate()`? Otherwise, I see no reason to thread `accepted` through these function calls. What happens to the discarded transactions?

Side note: `validate_transactions!()` doesn't need the exclamation mark in its name.

components/electric/test/electric/satellite/ws_write_permissions_test.exs
```diff
@@ -0,0 +1,514 @@
...
+    start_link_supervised!({Bandit, port: port, plug: plug})
+
+    server_id = Electric.instance_id()
+
+    start_link_supervised!({Electric.Satellite.ClientReconnectionInfo, connector_config})
+
+    %{port: port, server_id: server_id, origin: origin}
+  end
+
+  setup_with_mocks([
```
alco, line 103: Ugh, do we really need another module, for which 90% of the sync service is mocked out? It's tedious to maintain and is, well, not representative of running the real thing. More than half of this module is setup boilerplate.

Perhaps this could be written as an E2E test where we have control over both the client and the server actions?

components/electric/lib/electric/satellite/permissions/write.ex
magnetised: I was just keeping the structure similar to the write validations result, but I guess for no good reason.

components/electric/test/electric/satellite/ws_write_permissions_test.exs
magnetised: yeah, get that. I refactored this test and its sibling `ws_server_test` to at least move the (mostly shared) boiler plate into a helper module. Not ideal but TBH at this moment we have bigger fish to fry IMHO.

Also, though I hate the mocking as much as you do, I do think that we should be unit testing this stuff, rather than e2e testing.

components/electric/lib/electric/satellite/protocol.ex
magnetised: I'm going to remove the return of the txns from the Permissions.Write.validate function, as you're right, there's no need. But I will keep the use of `accepted` because I think the validate_transactions call should be able to return a filtered list of txns.

alco (approved):

PR #1340, opened 2024-06-07
fix: psql utility

fnacarellidev:
As reported to @samwillis through electrify discord I discovered a problem with `electric-sql psql` utility that, if you didn't start the proxy with `electric-sql start --with-postgres` the `electric-sql psql` utility would fail to connect to the database. This PR adds a fallback that forces runs the connection using the host machine, still need to check if the host machine has `psql` installed to perform the connection, but I tested like this and it worked just fine. Before actually making that check I want to see the opinion of the owners of the project about this solution.


KyleAMathews:

PR #1347, opened 2024-06-10
Disable granting write permissions

magnetised:
- [x] documentation

Adds a feature flag `permissions` which turns on perms for both reads and writes (on the server), and another separate flag which enables granting write permissions.

So enabling `permissions` turns moves electric into a read-only mode using the permissions expressed via the DDLX

### DIFF
--- a/components/electric/config/runtime.exs
+++ b/components/electric/config/runtime.exs
@@ -190,18 +190,15 @@ config :electric,
 # override these using the `ELECTRIC_FEATURES` environment variable, e.g.
 # to add a flag enabling `ELECTRIC GRANT` use:
 #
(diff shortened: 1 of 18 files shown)


alco (approved):
Looks good!

I've just noticed that this branch, as well as the very first one in the PR chain, has documentation changes. Those need to be taken out into a separate branch to avoid updating our online docs when this PR chain is merged but not yet released.

magnetised:
> Looks good!
> 
> I've just noticed that this branch, as well as the very first one in the PR chain, has documentation changes. Those need to be taken out into a separate branch to avoid updating our online docs when this PR chain is merged but not yet released.

Oh bugger. ok. I'll add that to the todos

PR #1352, opened 2024-06-12
Support macOS Big Sur and above

robacourt:
Currently on the latest macOS the tests fail with `eaddrinuse`

This is due to the port being blocked:
- "On MacOS Big Sur and later, port 55555 ... is blocked" - [source](https://docs.solace.com/Software-Broker/SW-Broker-Set-Up/Containers/Set-Up-Docker-Container-macOS.htm#:~:text=On%20MacOS%20Big%20Sur%20and,%22port%20in%20use%22%20error)
- See also: https://forums.developer.apple.com/forums/thread/671197

This PR changes the port number used in the test


msfstef (approved):
Looks good!

PR #1353, opened 2024-06-12
VAX-1934 : Improve error when tables can't be Electrified due to FK constraint boundaries

magnetised:


e2e/tests/04.06_electrification_with_unsatisfied_fks.lux
```diff
@@ -0,0 +1,31 @@
...
+            id TEXT PRIMARY KEY,
+            content TEXT NOT NULL,
+            item_id UUID REFERENCES public.items (id)
+        );
+
+        ALTER TABLE public.other_items ENABLE ELECTRIC;
+    ROLLBACK;
+    """
+    ??ERROR:  Cannot electrify table "public"."other_items" because it has a FOREIGN KEY reference to an un-electrified table: "public"."items"
+    ?$psql
```
alco, line 26: Nit: these two lines are superfluous. We've checked the error message already, nothing else needs to be done for this test, apart from teardown

e2e/tests/04.06_electrification_with_unsatisfied_fks.lux
```diff
@@ -0,0 +1,31 @@
+[doc Electrifying a table with fks to unelectrified tables returns a helpful error]
+
+[include _shared.luxinc]
+[include _satellite_macros.luxinc]
```
alco, line 4: No satellite client is used here, so this `include` is unneccessary.

alco (approved):
🤘

e2e/tests/04.06_electrification_with_unsatisfied_fks.lux
magnetised: maybe superfluous but also very explicit, I'd prefer to assert that we're getting the error rather than just that we're not getting the success

PR #1355, opened 2024-06-13
feat (client): extract sync API and make DAL optional

kevin-dp:
This PR extracts the sync method out of the DAL by introducing a `subscribe` method on the `ElectricClient.sync` object. To compute the shape requested by the user we need a description of the database schema, namely:
- Table name, and for each table:
  - columns (column name + PG type)
  - outgoing and incoming relations

The outgoing relations are the FKs in the table that reference a column in (another) table.
The incoming relations are FKs defined in (another) table that reference a column of this table.
The incoming relations are required in order to be able to traverse FKs in the reverse direction, e.g. this is needed when syncing the following shape because there is no FK from `issues` to `comments`, only from `comments` to `issues`:
```json
{
  "table": "issues",
  "include": {
    "comments": true
  }
}
```

The database description used to be generated by our generator and included in the generated Electric client. To truly extract the sync API i had to make it independent of the DAL and the generated client. Therefore, i extended the CLI to use the metadata that is provided with the migrations to generate the DB description and bundle that into the app.

Note:
- I modified our TS satellite client in the e2e tests such that every DAL query has raw equivalent and to dynamically dispatch to the right version depending whether we are running with or without the DAL. I introduced a matrix in CI such that every Satellite e2e test is ran once with the DAL and once without the DAL.

Questions:
- The modified CLI changes the behavior of `electric-sql generate` as by default it will no longer generate the DAL. If you want to generate the DAL, you need to run it with the `--with-dal` flag as in `electric-sql generate --with-dal`. Is thi acceptable, or we want to keep generating the DAL by default and have a `--no-dal` flag instead?
- I modified the format of the table columns in the generated DB description, it was previously a `Map<TableName, PgType>` and is now represented as a `Record<TableName, PgType>`. While this is not strictly necessary, this simplifies the generation of the DB description. Downside is that clients generated with the old generator won't work with newer versions of the ts-client and will require re-generating the DAL with the new ts-client. Is this fine or should i change it back to using a `Map<TableName, PgType>` ?


samwillis:
>The modified CLI changes the behavior of electric-sql generate as by default it will no longer generate the DAL. If you want to generate the DAL, you need to run it with the --with-dal flag as in electric-sql generate --with-dal. Is thi acceptable, or we want to keep generating the DAL by default and have a --no-dal flag instead?

I would be inclined to support both CLI args and have it default to`--with-dal` for now, we can then easily merge this PR asap without having to coordinate messaging. It's then a flip of the default at some point in the future.

>I modified the format of the table columns in the generated DB description, it was previously a Map<TableName, PgType> and is now represented as a Record<TableName, PgType>. While this is not strictly necessary, this simplifies the generation of the DB description. Downside is that clients generated with the old generator won't work with newer versions of the ts-client and will require re-generating the DAL with the new ts-client. Is this fine or should i change it back to using a Map<TableName, PgType> ?

I much prefer `Record<TableName, PgType>` 👍

clients/typescript/src/cli/util/serialize.ts
```diff
@@ -0,0 +1,42 @@
+import { dedent } from 'ts-dedent'
+import { DbSchema } from '../../client/util/relations'
+
+export function serializeDbDescription(dbDescription: DbSchema) {
```
msfstef, line 4: nit: might be worth mentioning in a comment that this serialization makes use of the `Relation` class/data model. Alternatively if this could be serialized to a pure object/json and then parsed into this form at the entrypoint where the db description is read that could be more portable?

clients/typescript/src/client/model/client.ts
```diff
@@ -154,30 +194,44 @@ export class ElectricClient<
...
+          dialect
+        )
+      }
 
-    // Now inform each table about all tables
-    Object.keys(dal).forEach((tableName) => {
-      dal[tableName].setTables(new Map(Object.entries(dal)))
-    })
+      // Create all tables
+      dal = Object.fromEntries(
```
msfstef, line 224: nit: why not make `dal` a const and then populate it with a `forEach` loop here? Not super important but since the object is mutable regardless I find it better to block reassignment since we don't intend to have it happen anyway

clients/typescript/src/client/model/client.ts
```diff
@@ -154,30 +194,44 @@ export class ElectricClient<
...
+        Object.keys(tables).map((tableName) => {
+          return [tableName, createTable(tableName)]
+        })
+      ) as ClientTables<DB>
+
+      // Now inform each table about all tables
+      Object.keys(dal).forEach((tableName) => {
+        dal[tableName].setTables(new Map(Object.entries(dal)))
```
msfstef, line 232: nit: generate the `new Map(Object.entries(dal))` once outside of the loop

clients/typescript/src/client/model/client.ts
```diff
@@ -191,10 +245,12 @@ export class ElectricClient<
     return new ElectricClient(
```
msfstef, line 245: nit: not for this PR but I think we should really move to object arguments rather than positional haha

clients/typescript/src/client/model/sync.ts
```diff
@@ -0,0 +1,96 @@
...
+    typeof tableName !== 'string'
+  ) {
+    throw new Error(
+      'Cannot sync the requested shape. Table name must be a non-empty string'
+    )
+  }
+
+  // Remove the table name from the input
+  const input: ShapeInput = i
+  delete input.table
```
msfstef, line 35: do we really need to do this? this mutates the input object provided by the user which feels a bit odd, could we not just keep passing it along and cast it as a `ShapeInput`, since the `table` property will just be ignored

clients/typescript/src/client/model/table.ts
```diff
@@ -381,6 +436,10 @@ export class Table<
     )
   }
 
+  /**
+   * @deprecated The DAL is deprecated and will be removed in a future release.
```
msfstef, line 440: nit for @samwillis - even if we keep `--with-dal` as the default, would we still want to show the deprecation warnings on the DAL?

clients/typescript/src/client/model/sync.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  // Recursively go over the included fields
+  const include = i.include ?? {}
+  const where = i.where ?? ''
+  const includedFields = Object.keys(include)
+  const includedTables = includedFields.map((field: string): Rel => {
+    // Fetch the table that is included
+    const relatedTableName = dbSchema.getRelatedTable(tableName, field)
+    const fkk = dbSchema.getForeignKey(tableName, field)
```
msfstef, line 60: nit: assuming this is a typo of `fk` but I might be missing something

clients/typescript/src/client/model/sync.ts
```diff
@@ -0,0 +1,96 @@
...
+  // Remove the table name from the input
+  const input: ShapeInput = i
+  delete input.table
+
+  // Compute the shape from the user input
+  const shape = computeShape(dbDescription, tableName, i)
+  return shapeManager.subscribe([shape], key)
+}
+
+export function computeShape(
```
msfstef, line 42: I remember we had discussed about code duplication for this and the Table's `computeShape` method, but could the Table's method not just use this method instead and pass along its own `tableName` and `dbDescription` that way we keep the `computeShape` code covered with the same tests, since right now this is not fully covered.

msfstef (commented):
Amazing work 🔥 ! Left some comments for clarifications

I agree with @samwillis on keeping the `--with-dal` to `true` by default, although we need to decide if we want to show the deprecation warnings already in that case. When we're happy with making a breaking change with a release we just switch the default.

I'm also a fan of using the `Record<TableName, PgType>`, but it does mean that this update would require a regeneration of the client and we still haven't made it clear if that constitutes a breaking change or not.

clients/typescript/src/cli/util/serialize.ts
kevin-dp: Yep will clarify that in comments. The reason it's serialised like this is because it is bundled into the app and that file will then be imported. So there is no deserialisation happening so we need to make sure that the serialised version is instantiating the class.

clients/typescript/src/client/model/client.ts
kevin-dp: This code will be removed in a follow-up PR.

kevin-dp:
@samwillis @msfstef I addressed most of your comments. The main change is the flip of the flag default. I kept the `--with-dal` flag but it defaults to `true`. If you don't want to generate the DAL, then you pass `--with-dal false`.

clients/typescript/src/cli/migrations/command-generate.ts
```diff
@@ -66,14 +67,27 @@ export function makeGenerateCommand(): Command {
...
+      `
+    )
+
     .action(async (opts: GenerateCommandArgs) => {
-      const { watch, withMigrations, debug, ...restOpts } = opts
+      const { watch, withMigrations, debug, withDal, ...restOpts } = opts
       const config = getConfig(restOpts)
 
+      const withDalDefault = withDal ?? 'true' // defaults to true
+      const withDalOpt = withDalDefault === 'true' // convert to boolean
```
msfstef, line 84: nit: not sure how the CLI tool we're using works exactly but why is the option not a boolean directly? Also, populating `withDalDefault` with the provided options feels a bit counterintuitive, would it not be better to do
```ts
const { ..., withDal = true, ... } = opts
genOpts: {  ..., withDal, ... }
```

can the user specify `--with-dal` with no argument, would that work?

msfstef (approved):
Looks good, ready to merge! I do think we need to confirm that by merging this, although we are keeping the same current behaviour by default, we will be showing deprecation warnings everywhere.

clients/typescript/src/cli/migrations/command-generate.ts
kevin-dp: The command line args tool provides the value passed to the option as a string. So we would get `"true"` or `"false"`. So first if it is not provided at all, we default to `"true"` and then we turn the value into a boolean by checking if that string `=== "true"`.

> can the user specify --with-dal with no argument, would that work?
No the command line args tool does not allow this because the docs for that option specify that an argument is required.

msfstef: but if the user provides anything else like "whatever", that would fail the `=== 'true'` and act as `--with-dal false` right? Just checking that this is the intended behaviour, or would the CLI tool reject any option other than `"true"` and `"false"`?

kevin-dp: Right, fixed it like this:
```ts
const withDalOpt = withDalDefault !== 'false'
```

clients/typescript/src/client/model/table.ts
kevin-dp: Moved deprecation warnings to a separate PR

PR #1358, opened 2024-06-13
docs: switch config examples to use direct writes mode.

thruflo:
Recommending setting a `LOGICAL_PUBLISHER_HOST` complicates the examples compared with just setting `ELECTRIC_WRITE_TO_PG_MODE=direct_writes` because this way you don't need to worry about having your docker electric addressable from Postgres.

### DIFF
--- a/docs/usage/installation/service.md
+++ b/docs/usage/installation/service.md
@@ -14,12 +14,12 @@ You can run the [pre-packaged Docker images](#images) published on Docker Hub, o
 The Electric sync service is configured using environment variables. The three required variables are:
 
 - `DATABASE_URL` in the format of a Postgres [Connection URI](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS)
-- `LOGICAL_PUBLISHER_HOST` that the sync service is running on (must be accessible from the Postgres instance to establish an inbound replication subscription)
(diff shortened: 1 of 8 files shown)


docs/usage/installation/service.md
```diff
@@ -14,12 +14,12 @@ You can run the [pre-packaged Docker images](#images) published on Docker Hub, o
 The Electric sync service is configured using environment variables. The three required variables are:
 
 - `DATABASE_URL` in the format of a Postgres [Connection URI](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS)
-- `LOGICAL_PUBLISHER_HOST` that the sync service is running on (must be accessible from the Postgres instance to establish an inbound replication subscription)
+- either `LOGICAL_PUBLISHER_HOST` that the sync service is running on (must be accessible from the Postgres instance to establish an inbound replication subscription) or set `ELECTRIC_WRITE_TO_PG_MODE=direct_writes`
```
msfstef, line 17: nit: maybe reverse order if we want to emphasize direct writes

msfstef (approved):
Looks good! There's a few more mentions that might be worth changing:

- In `architecture.md` under Streaming into Electric the logical replication is mentioned as the way this happens, we probably want to switch that or at least mention direct writes as the primary method
- In `running.md` under `docs/examples/notes` we describe the logical publisher host as the method for connecting electric to postgres

alco:
FYI we're planning to remove the `logical_replication` mode but with no committed timeline yet - https://github.com/electric-sql/electric/pull/1343.

PR #1363, opened 2024-06-13
chore(client): Performance improvements on initial sync downstream path

msfstef:
I profiled the downstream path for initial syncs, excluding SQLite performance as that will be driver and storage dependent and I plan to focus on that separately, and implemented some changes on the most significant bottlenecks for that pathway.

The tests were done using 10k issues and 50k comments on Linearlite (although now benchmarking suite @icehaunter made is available so I will be using that), and performance improvements were checked in isolation with JS perf tests as well (on Chrome/V8 but I avoided any super niche optimizations)

In memory performance is improved for this path, excluding the execution of the SQL statements the 10k issues and 50k comments go through `_applySubscriptionData` in under 200ms on my machine.

One very significant overhead is the decoding, first of the Satellite messages in the generated library which I don't think we can do much about, but then also to JS variables with our decoders that spend a lot of time and also cause lots of GC pauses. There are ways we could optimise this path but I was reluctant to spend time on that, ultimately a lot of the deserialization is pointless in the sense that it is re-serialized for inserting to permanent storage, but if we want to keep things modular and pluggable with e.g. JS SQLite drivers this might be necessary.

Some individual commits contain messages with a brief description of the performance bits, mostly I tried to reuse objects to reduce GC pauses and that seemed to give the most benefit.

Measured performance with 10k issues and 50k comments of `_handleSubscriptionData`:
on main: `265ms ± 36ms`
after improvements: `135ms ± 9ms`


alco:
> There are ways we could optimise this path but I was reluctant to spend time on that

I expect that we'll see a lot of such loose threads of optimisation. No good jumping on each of them immediately, but we should keep track of those someplace to be able to evaluate later which potential optimisation would have the biggest impact.

clients/typescript/src/satellite/oplog.ts
```diff
@@ -393,11 +393,12 @@ export function extractPK(c: DataChange) {
...
 
   return primaryKeyToStr(
-    Object.fromEntries(
-      c.relation.columns
-        .filter((c) => c.primaryKey)
-        .map((col) => [col.name, columnValues[col.name]!])
-    )
+    c.relation.columns
+      .filter((c) => c.primaryKey)
+      .reduce((primaryKeyRec, col) => {
```
alco, line 398: If we go from `fromEntries` to `reduce` because the latter is faster, what can be said about using a regular for loop then? I bet writing
```
for (...) {
  if (c.primaryKey) {
    primaryKeyRec[col.name] = columnValues[col.name]!
  }
}
```
would make it faster still.

But we should take care and avoid focusing too much on micro-optimisation. There's always a single-digit-percent improvement available, but those also face the risk of
- making the code more complex and therefore prone to hinding bugs
- getting lot in a refactoring/higher-level logic rewrite

alco (approved):
🤘🤘

clients/typescript/src/satellite/oplog.ts
msfstef: yep I try to avoid turning array methods to loops as I think it hurts readability - and also most js engines tend to try and optimize them regardless

the main part where I wanted to focus on optimization was operations that are ran per-row on large initial syncs and that can be optimized in isolation of the rest of the higher level logic/pipeline

clients/typescript/src/satellite/process.ts
```diff
@@ -561,26 +562,37 @@ export class SatelliteProcess implements Satellite {
...
+        fullTableNameLookup.set(op.relation.table, tableName)
+        groupedChange = {
           relation: op.relation,
+          primaryKeyColNames: op.relation.columns
+            .filter((col) => col.primaryKey)
+            .map((col) => col.name),
           records: [op.record],
-          table: qt,
-        })
+          table: new QualifiedTablename(namespace, op.relation.table),
```
kevin-dp, line 578: we can reuse `qt` here

kevin-dp (approved):
LGTM!

clients/typescript/src/satellite/oplog.ts
icehaunter: @alco `for` and `reduce` are same speed in JS on V8. It's actrually really hard to reason about performance on JS because of JIT optimisations V8 is doing

PR #1370, opened 2024-06-16
feat(electric): Internal API for distributed tracing using OpenTelemetry

alco:
This PR adds OpenTelemetry libraries and covers parts of the initial shape sync process with tracing spans.

The E2E setup is modified to run an [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/) with each test for collecting both client- and server-originating traces and exporting them to Honeycomb.io.

https://github.com/electric-sql/electric/pull/1371 is a follow-up that adds OT support in the client and also implements passing of the tracing context in client-originating RPC requests.


davidmartos96:
This looks useful! 
A question regarding the implementation is if it could cause issues running PRs from external contributors.
The HONEYCOMB secret wouldn't be available, so a solution could be simply running the e2e tests without the OpenTelemetry in those situations.

alco:
@davidmartos96 We shall see. I think e2e tests will run fine but nothing will be sent to Honeycomb in such cases.

components/electric/lib/electric/telemetry/open_telemetry.ex
```diff
@@ -0,0 +1,139 @@
+defmodule Electric.Telemetry.OpenTelemetry do
+  @moduledoc """
```
magnetised, line 2: nice.

magnetised (approved):
Don't like the extra complexity in say protocol.ex, which is already extremely messy, but ok, guess we have to live with it.

PR #1371, opened 2024-06-16
feat(satellite,electric): Implement distributed tracing using OpenTelemetry in the client

alco:
This is a follow-up to https://github.com/electric-sql/electric/pull/1370.

Besides adding OT to the client, here we also implement the passing of client's tracing context to the server in RPC calls across the network boundary. The GH Actions workflows that run E2E tests are updated to enable the export of client traces to Honeycomb.io where each test gets its own `ci.test_file` attribute and a bunch of other CI-related attributes.


clients/typescript/src/satellite/process.ts
```diff
@@ -358,6 +368,7 @@ export class SatelliteProcess implements Satellite {
   }
 
   private async _stop(shutdown?: boolean): Promise<void> {
+    const span = startSpan('satellite.process.stop')
```
kevin-dp, line 371: What would happen if `_stop` is called twice and they run concurrently?
i.e. the 2nd invocation would call `startSpan` again with the same label, what happens in that case?
Does it create a separate span that is somehow distinct from the first, or does it aggregate them, or does it ignore the 2nd call, or it throws an error, or something else? Tried looking it up but couldn't immediately find an answer in the docs.

clients/typescript/package.json
```diff
@@ -178,6 +178,13 @@
   },
   "dependencies": {
     "@electric-sql/prisma-generator": "workspace:*",
+    "@opentelemetry/api": "^1.9.0",
```
kevin-dp, line 181: Btw, found this in the docs about dependencies:
> If you’re instrumenting a library, only install the OpenTelemetry API package for your language. Your library will not emit telemetry on its own. It will only emit telemetry when it is part of an app that uses the OpenTelemetry SDK. For more on instrumenting libraries, see [Libraries](https://opentelemetry.io/docs/concepts/instrumentation/libraries/).

Are we somehow different that we do need all the dependencies? Or could we move them into optional peer dependencies?

clients/typescript/src/util/telemetry/spanUtilities.ts
```diff
@@ -0,0 +1,102 @@
...
+      kind: isClientSpan ? SpanKind.CLIENT : SpanKind.INTERNAL,
+      attributes,
+    },
+    parentSpan && trace.setSpan(context.active(), parentSpan)
+  )
+  return span
+}
+
+/**
+ * Sets the span status to `ERROR` and optinally records the provided error
```
kevin-dp, line 34: typo in "optinally"

clients/typescript/src/util/telemetry/spanUtilities.ts
```diff
@@ -0,0 +1,102 @@
...
+ * ass an exception
```
kevin-dp, line 35: type in "ass"

clients/typescript/src/util/telemetry/spanUtilities.ts
```diff
@@ -0,0 +1,102 @@
...
+      functionToTrace(span)
+    )
+
+    // if result is a promise, chain span actions to it
+    if (result instanceof Promise) {
+      result
+        .then((_) => span.setStatus({ code: SpanStatusCode.OK }))
+        .catch((err) => recordSpanError(span, err))
+        .finally(() => span.end())
+      return result
```
kevin-dp, line 74: Just leaving a comment here to make sure that we are on purpose returning the original promise and not the chained promise.

kevin-dp (commented):
I don't really like the additional complexity added by the telemetry calls.
For example, the shape `sync` method which used to be 5 LoC is now >30 LoC.
Also, we know that we also want to add some improved error logging, which might add yet other code unrelated to the core logic.

Looks like the ideal use case for some aspected-oriented programming but haven't seen good AoP solutions for JS. Essentially, i want to move the telemetry logic outside of the core logic here. Perhaps the chain-of-responsibility pattern could solve this problem but we would have to see if and how that fits our implementation. If we're fine with only tracing public API methods, we can perhaps go for the chain-of-responsibility pattern.

alco:
@msfstef Can you help me out here with addressing Kevin's feedback?

I see there are merge conflicts but haven't rebased so as not to mess up code comments from Kevin.

clients/typescript/src/satellite/process.ts
msfstef: Each span has a unique ID, so multiple concurrent calls will just generate multiple spans with the same label that are measured concurrently.

clients/typescript/src/util/telemetry/spanUtilities.ts
msfstef: yes cause IMO the caller handles the errors independently and doesn't need to be chained to the span ending - the only potential issue I guess would be if for some reason anything chained to the original promise after this call is scheduled to occur before the `span.end` chained call - but since this chain is established first I would imagine it's also first in the event loop queue

clients/typescript/package.json
msfstef: That's a good point - @alco what do you think about making the client use just the API dependency and then moving the initialization and everything else into the satellite e2e client? Since all our telemetry is going to be within the context of e2e and dedicated tests that might be a better way to go about it

alco: I think OT export should be a part of our library. The motivation is for a customer who wants to debug some specific performance issue to be able to set a flag + endpoint URL and start collecting traces, as opposed to having to add OT initialization code into their app.

If we ship our library together with the SDK dependency, it should be compatible with both scenarios:
1. customer just wants to set a flag to start exporting traces
2. customer has their own OT SDK integration and wants to make our library's traces part of its exports

msfstef: In that case I've changed the initialization slightly to potentially support externally initialized providers as well (see https://github.com/electric-sql/electric/pull/1371/commits/4d2ca2593955e551594e87e92f5ea21223cbbf30)

KyleAMathews:

PR #1374, opened 2024-06-17
fix (client): fix immutable fields in setReplicationTransform

kevin-dp:
This PR modifies the immutable fields that are checked in `setReplicationTransform` such that it checks for the outgoing FK columns and the PK columns that are being pointed at. Also adds unit tests to check that these are computed correctly and an error is thrown if we try to modify them.


msfstef (approved):
Nice and concise, great fix! 🚀

PR #1375, opened 2024-06-17
fix(electric): Add python psycopg3 support

magnetised:
psycopg sends begin, commit and rollback statements using the extended, parse, bind, execute, protocol, which is... different

### DIFF
--- a/components/electric/lib/electric/postgres/proxy/injector/electric.ex
+++ b/components/electric/lib/electric/postgres/proxy/injector/electric.ex
@@ -259,6 +259,21 @@ defmodule Electric.Postgres.Proxy.Injector.Electric do
 
               {[Operation.Wait.new(msgs, state, signal), stack], {electric, state}}
 
+            # psycopg sends its txn commands using the extended protocol, annoyingly
+            # it uses a [parse, describe, bind, execute, sync] message block, so all we
+            # need to do is pass that on and mark the connection as in a transaction
+            %{action: {:tx, action}} = _analysis when action in [:begin, :rollback, :commit] ->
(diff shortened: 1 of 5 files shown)


components/electric/lib/electric/postgres/proxy/injector/electric.ex
```diff
@@ -259,6 +259,21 @@ defmodule Electric.Postgres.Proxy.Injector.Electric do
...
               {[Operation.Wait.new(msgs, state, signal), stack], {electric, state}}
 
+            # psycopg sends its txn commands using the extended protocol, annoyingly
+            # it uses a [parse, describe, bind, execute, sync] message block, so all we
+            # need to do is pass that on and mark the connection as in a transaction
+            %{action: {:tx, action}} = _analysis when action in [:begin, :rollback, :commit] ->
+              state = State.transaction(state, action)
+
+              op =
+                if Enum.any?(msgs, &is_struct(&1, M.Execute)) do
```
alco, line 269: Pardon my ignorance but can we be reasonably sure that `msgs` includes all messages mentioned in the comment above at once, i.e. without getting broken up into multiple lists during injestion inside the proxy?

magnetised: Yes, we can be reasonably sure that if all the messages are present they'll be in a single list. But we can't be sure that what we're seeing is definitely from psycopg, so I've added a case to handle two sets of messages, [parse, describe], [bind, execute], which is the usual sequence. it's not a well tested path admittedly, but I wanted to nod to it.

alco (approved):

PR #1376, opened 2024-06-17
feat (client): renamed disableFKs flag and modified default to disable FKs on SQLite

kevin-dp:
Decided to go for an enumeration as this is more explicit than the `boolean | undefined`:
```ts
enum ForeignKeyChecks {
  enabled = 'enabled',
  disabled = 'disabled',
  inherit = 'inherit',
}
```


msfstef (approved):
Nice, much nicer now!

Just needs a changeset and an accompanying doc update PR

Issue #1301, opened 2024-05-25
Unable to install electric-sql on MacOS & Node@22


WailRoth:
`Computer specs`
```
OS: MacOS 14.2.1 (23C71)
Node: 22.0
```

`error`
```
./src/util/macros.lzz:150:35: error: no matching member function for call to 'SetAccessor'
│         recv->InstanceTemplate()->SetAccessor(
│         ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
│ /Users/XXX/Library/Caches/node-gyp/22.0.0/include/node/v8-template.h:1049:8: note: candidate function not viable: no known conversion from 'v8::AccessControl' to 'v8::PropertyAttribute' for 5th argument
│   void SetAccessor(
│        ^
│ /Users/xxx/Library/Caches/node-gyp/22.0.0/include/node/v8-template.h:1055:8: note: candidate function not viable: no known conversion from 'v8::AccessorGetterCallback' (aka 'void (*)(Local<v8::String>, cons…
│   void SetAccessor(
│        ^
│ ./src/objects/database.lzz:180:21: warning: variable 'status' set but not used [-Wunused-but-set-variable]
│                 int status = sqlite3_db_config(db_handle, SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION, 1, NULL);
│                     ^
│ 1 warning and 1 error generated.
│ make: *** [Release/obj.target/better_sqlite3/src/better_sqlite3.o] Error 1
```

So has you can see, (or not with the error), it tries to run the `better-sql3` installation script,  `node_modules/.pnpm/better-sqlite3@8.7.0/node_modules/better-sqlite3`

from https://github.com/electric-sql/electric/blob/main/pnpm-lock.yaml
```
better-sqlite3:
 specifier: ^8.4.0
 version: 8.7.0
```

But, this version is not supported with node@22

WailRoth:
For those who want to downgrade to node@20 on macos:
https://apple.stackexchange.com/questions/171530/how-do-i-downgrade-node-or-install-a-specific-previous-version-using-homebrew

msfstef:
It seems like the better-sqlite3 version currently used does not support Node v22, I've opened a PR to upgrade the dependency (which comes at the cost of dropping Node v16 support, but that's at its EOL) https://github.com/electric-sql/electric/pull/1378

Thank you for raising the issue!

PR #1378, opened 2024-06-17
chore(client): Drop Node 16 support and add Node 22 support

msfstef:
Node 16 has been [dead for a while](https://nodejs.org/en/blog/announcements/nodejs16-eol) and we are not testing more modern Node LTS versions.

Most prominently, the better-sqlite3 we were using in the client does not support Node v22 as raised by @robacourt  and others (https://github.com/electric-sql/electric/issues/1301)

I've updated the CI workflows to use Node v22 and the matrices to test v18 v20 and v22


clients/typescript/src/drivers/better-sqlite3/mock.ts
```diff
@@ -20,7 +20,10 @@ export class MockDatabase implements Database {
     return this
   }
 
-  prepare<T extends BindParams = []>(_sql: string): Statement<T> {
+  // @ts-ignore annoying typings for mock
```
msfstef, line 23: I'm unsure whether this is a TS issue or I'm failing to conform to the types, but the typings are so obtuse that I refuse to spend more time on this mock

alco (approved):
:+1:

clients/typescript/src/drivers/better-sqlite3/mock.ts
```diff
@@ -42,13 +45,13 @@ export class MockDatabase implements Database {
     }
 
     // Valid only for mocking since we don't expect to need to mock full interface
-    return mockStatement as Statement<T>
+    return mockStatement as Statement<T, Result>
```
kevin-dp, line 48: This type error can easily be "avoided", we know that this is a wrong typecast that is only valid for mocking, so we need to make it an unsafe typecast:
```ts
return mockStatement as unknown as Statement<T, Result>
```

clients/typescript/src/drivers/better-sqlite3/mock.ts
kevin-dp: I don't really like this but the typings are indeed insane. I spent some time on it trying to figure it out but i didn't succeed. Would need more time to figure it out...
Not sure whether we should spend the time or not.

clients/typescript/test/drivers/better-sqlite3.test.ts
```diff
@@ -9,6 +9,7 @@ import { UncoordinatedDatabaseAdapter } from '../../src/electric'
 
 test('database adapter run works', async (t) => {
   const db = new MockDatabase('test.db')
+  //@ts-ignore issue with typechecking
```
kevin-dp, line 12: I don't think it is acceptable that we have to `ts-ignore` every instantiation of the DB adapter with a mock database.
We can apply a dirty fix that solves this problem by modifying the `MockDatabase`: instead of defining the `prepare` method we can define a `prepare` constant and just do an unsafe typecast of the arrow function to the expected type.

kevin-dp (changes requested):

kevin-dp (approved):
Much better!
Good to go for me.

Issue #1365, opened 2024-06-13
After shape.synced resolves, there is some delay before data available in local DB


Gobbo89:
Hello, I was talking with @balegas about this potential issue a few weeks ago.  
I've noticed this since electric-sql v0.9.x.  
I personally do not think it's the intended behavior, it looks more like a bug or at least something that could be improved.

As the tile says, it looks like after `shape.synced` resolves, there still is some delay before the data is actually available in the local DB.  
This happens with an **already initialized client DB.** While the client is not running, you add some data in PG. Then you start the client.

The following code is just a shortened version of the problem, it might not be enough to reproduce.
```javascript
const shape = await db.items.sync()
await shape.synced

const syncedItems = await db.items.findMany()
console.log('Synced items', syncedItems) // Gives me []

// However, if I wait an arbitrary amount of time...
setTimeout(async () => {
  const syncedItems = await db.items.findMany()
  console.log(`Synced items (timeout)`, syncedItems) // Gives me ['...', (...)]
}, 15)
```

Because it might be a bit tricky to understand the conditions under which this occurs, I have created [a repo to reproduce this consistently](https://github.com/Gobbo89/electric-shape-synced), along with some log screenshots.

Thank you

Gobbo89:
Hey @msfstef thanks for the clarification.
&nbsp;

> This means that shape.synced immediately resolves, your query returns the data already available in the database

What lead me to think otherwise, is [this part of the docs](https://electric-sql.com/docs/usage/data-access/shapes#promise-workflow:~:text=the%20second%20synced%20promise%20resolves%20when%20the%20data%20in%20the%20shape%20has%20fully%20synced%20onto%20the%20local%20device):
`the second synced promise resolves when the data in the shape has fully synced onto the local device`
That is indeed true only if we are talking about **the very first sync** between a client and the sync service, but it's very misleading in the other cases, since the synced promise resolves immediately, like you said.

Put that together with point no. 1 stating _"the first sync() promise resolves when the shape subscription has been confirmed by the server"_ and it sounds like the second promises **always** resolves after the server data (which we are aware of, after no.1) is loaded in the DB on the local device.    

&nbsp;
> Would a name like syncedOnce perhaps clarify the observed behaviour?

I don't know. You risk to go down the rabbit hole of trying and make variable / function names most descriptive as possible, when in this case it's more complicated than that, and you can put that logic in just a name.  
At the end of the day, in some cases that promises does resolve when the data is synced, just not always.  
[...]

PR #1380, opened 2024-06-18
docs: clarify when `await shape.synced` resolves.

thruflo:
Addresses some of the suggestions in #1365.


docs/usage/data-access/shapes.md
```diff
@@ -197,6 +197,10 @@ await shape.synced
 
 If the shape subscription is invalid, the first promise will be rejected. If the data load fails for some reason, the second promise will be rejected.
 
+:::note
+If you establish a shape subscription that has already synced its initial data, awaiting `shape.synced` will always resolve immediately. I.e.: imagine that you re-run the code above on subsequent page loads. Awaiting `shape.synced` a second time will only ensure that the initial shape load is complete. It does not ensure that the replication stream has fully caught up.
```
msfstef, line 201: nit: perhaps we should avoid using phrases like "fully caught up" - perhaps a plain "caught up" might still do

```suggestion
If you establish a shape subscription that has already synced its initial data, awaiting `shape.synced` will always resolve immediately. I.e.: imagine that you re-run the code above on subsequent page loads. Awaiting `shape.synced` a second time will only ensure that the initial shape load is complete. It does not ensure that the replication stream has caught up to the central Postgres' more recent state.
```

msfstef (approved):
Looks good!

PR #1391, opened 2024-06-20
Add process labels

robacourt:
Adding labels to processes to make using the Erlang Observer and various debugging tasks easier


icehaunter (approved):
Great job!

PR #1392, opened 2024-06-20
feat (client): drop the DAL and the generator

kevin-dp:
This PR drops the DAL and the generator from the client.
DAL unit tests have been removed.
The remaining unit tests and e2e tests have been modified accordingly by replacing DAL queries by raw queries.

TODO:

- Update website (docs)


clients/typescript/src/util/types.ts
```diff
@@ -19,7 +19,14 @@ export type Query = string
...
 export type RowId = number
-export type SqlValue = string | number | null | Uint8Array | bigint
+export type SqlValue =
+  | boolean
+  | string
+  | number
+  | Uint8Array
+  | undefined
+  | null
+  | bigint
```
msfstef, line 29: super nitty nit: perhaps having `bigint` under `number` looks better?

msfstef (approved):
Looks good! Nice to see so many things being removed!

Should this PR also remove Zod as a dependency? Or are we still using it for schema validation with the minimal DB schema provided? I'm wondering if we are using it minimally if it might be worth doing a custom implementation of a thin schema validator for rows just for the sake of removing the problematic dependency.

kevin-dp:
> Looks good! Nice to see so many things being removed!
> 
> Should this PR also remove Zod as a dependency? Or are we still using it for schema validation with the minimal DB schema provided? I'm wondering if we are using it minimally if it might be worth doing a custom implementation of a thin schema validator for rows just for the sake of removing the problematic dependency.

We could make a follow-up PR that removes the Zod dependency altogether and replaces the usages of Zod by some custom validation code.

KyleAMathews:

PR #1394, opened 2024-06-20
fix(client): Fix batch deletes causing call stack overflow

msfstef:
Addresses [VAX-1985](https://linear.app/electric-sql/issue/VAX-1985/large-unsubscribes-lead-client-with-wa-sqlite-driver-to-a-stack)

Apparently the method we were using, where we chained `OR` and `AND` clauses in the `WHERE` clause, led to:
1. SQLite call stack overflow (at least with `wa-sqlite`) at around 6k `OR` clauses, far below the max parameter limit
2. Sub-optimal performance in both SQLite and PGlite

I've converted the batching to use `IN` statements for both single-column queries and multi-column ones, like with composite primary keys. The performance is significantly improved and has room for more improvement:

##### Shadow Table 30k deletes
- ~4700ms pglite
- ~500ms wa-sqlite (arbitrary small batching to avoid overflow)

##### Comment 9k deletes
- ~700ms pglite
- ~2000ms wa-sqlite (max 6k batches)

##### Shadow Table 30k deletes
- ~100ms pglite
- ~450ms wa-sqlite

##### Comment 9k deletes
- ~40ms pglite
- ~700ms wa-sqlite

While the shadow table deletes take similar time for wa-sqlite (~500ms), the shadow table uses a triple composite key of 3 string columns. There's room for significant optimization there, concatenating the parameters and deleting based on the concatenated PK columns reduced the time to execute from ~500ms to ~50ms - pglite also had a very small improvement but hard to measure.

I think it's possible to avoid having a composite primary key for the shadow table since we know it's 3 string columns and they can be collapsed into one if that proves to be a significant optimization, but that is outside of the scope of this fix, just raising it here as a potential optimization.

NOTE: this improvement is the result of the same investigation that yielded https://github.com/electric-sql/electric/pull/1389, both together result in a useable linearlite with >2k issues otherwise we run into timeouts and call stack overflows.

### DIFF
--- a/clients/typescript/src/migrators/query-builder/builder.ts
+++ b/clients/typescript/src/migrators/query-builder/builder.ts
@@ -321,7 +321,7 @@ export abstract class QueryBuilder {
    */
   prepareInsertBatchedStatements<T extends Row>(
     baseSql: string,
-    columns: Extract<keyof T, string>[],
+    columns: Array<keyof T>,
     records: Row[],
     maxParameters: number,
     suffixSql = ''
@@ -395,7 +395,7 @@ export abstract class QueryBuilder {
    */
   public prepareDeleteBatchedStatements<T extends Row>(
     baseSql: string,
-    columns: Extract<keyof T, string>[],
+    columns: string[],
     records: T[],
     maxParameters: number,
     suffixSql = ''
--- a/clients/typescript/src/migrators/query-builder/builder.ts
+++ b/clients/typescript/src/migrators/query-builder/builder.ts
@@ -290,6 +290,14 @@ export abstract class QueryBuilder {
   /**
    * Generates IN clause for a WHERE statement, checking that the given
    * columns have a value present in the given tupleArgs array
+   *
+   * The `args` array must be an array of values to compare each column to.
+   *
+   * If using a single column, then it can be a 1-dimensional array of
+   * values to check against.
+   *
+   * If using multiple columns, then it needs to be a 2-dimensional array where
+   * each entry is a row of values that the columns need to conform to
    */
   protected abstract createInClause(
     columns: string[],
--- a/clients/typescript/src/migrators/query-builder/builder.ts
+++ b/clients/typescript/src/migrators/query-builder/builder.ts
@@ -1,5 +1,5 @@
 import { ForeignKey } from '../triggers'
-import { QualifiedTablename, SqlValue, Statement } from '../../util'
+import { QualifiedTablename, Row, SqlValue, Statement } from '../../util'
 
 export type Dialect = 'SQLite' | 'Postgres'
 export abstract class QueryBuilder {
@@ -311,10 +311,10 @@ export abstract class QueryBuilder {
    * @param suffixSql optional SQL string to append to each insert statement
    * @returns array of statements ready to be executed by the adapter
    */
-  prepareInsertBatchedStatements(
+  prepareInsertBatchedStatements<T extends Row>(
     baseSql: string,
-    columns: string[],
-    records: Record<string, SqlValue>[],
+    columns: Extract<keyof T, string>[],
+    records: Row[],
     maxParameters: number,
     suffixSql = ''
   ): Statement[] {
@@ -385,10 +385,10 @@ export abstract class QueryBuilder {
    * @param suffixSql optional SQL string to append to each insert statement
    * @returns array of statements ready to be executed by the adapter
    */
-  public prepareDeleteBatchedStatements(
+  public prepareDeleteBatchedStatements<T extends Row>(
     baseSql: string,
-    columns: string[],
-    records: Record<string, SqlValue>[],
+    columns: Extract<keyof T, string>[],
+    records: T[],
     maxParameters: number,
     suffixSql = ''
   ): Statement[] {
--- a/clients/typescript/src/migrators/query-builder/builder.ts
+++ b/clients/typescript/src/migrators/query-builder/builder.ts
@@ -288,6 +288,15 @@ export abstract class QueryBuilder {
   ): string
 
   /**
+   * Generates IN clause for a WHERE statement, checking that the given
+   * columns have a value present in the given tupleArgs array
+   */
+  protected abstract createInClause(
+    columns: string[],
+    args: (string | string[])[]
+  ): string
+
+  /**
    * Prepare multiple batched insert statements for an array of records.
    *
    * Since SQLite only supports a limited amount of positional `?` parameters,
@@ -376,22 +385,19 @@ export abstract class QueryBuilder {
    * @param suffixSql optional SQL string to append to each insert statement
    * @returns array of statements ready to be executed by the adapter
    */
-  public prepareDeleteBatchedStatements<T extends object>(
+  public prepareDeleteBatchedStatements(
     baseSql: string,
-    columns: Array<keyof T>,
-    records: T[],
+    columns: string[],
+    records: Record<string, SqlValue>[],
     maxParameters: number,
     suffixSql = ''
   ): Statement[] {
     const stmts: Statement[] = []
     const columnCount = columns.length
     const recordCount = records.length
+    const isSingleColumnQuery = columnCount === 1
     // Amount of rows we can delete at once
     const batchMaxSize = Math.floor(maxParameters / columnCount)
-
-    // keep a temporary join array for joining strings, to avoid
-    // the overhead of generating a new array every time
-    const tempColumnComparisonJoinArr = Array.from({ length: columnCount })
 
     let processed = 0
     let prevDeleteCount = -1
@@ -403,17 +409,20 @@ export abstract class QueryBuilder {
       // of `batchMaxSize` - ideally we can externalize this cache since for a
       // given adapter this is _always_ going to be the same
       if (currentDeleteCount !== prevDeleteCount) {
-        deletePattern = Array.from(
-          { length: currentDeleteCount },
-          (_, recordIdx) => {
-            for (let i = 0; i < columnCount; i++) {
-              tempColumnComparisonJoinArr[i] = `"${
-                columns[i] as string
-              }" = ${this.makePositionalParam(recordIdx * columnCount + i + 1)}`
-            }
-            return ` (${tempColumnComparisonJoinArr.join(' AND ')})`
-          }
-        ).join(' OR')
+        deletePattern =
+          ' ' +
+          this.createInClause(
+            columns,
+            Array.from({ length: currentDeleteCount }, (_, recordIdx) =>
+              isSingleColumnQuery
+                ? this.makePositionalParam(recordIdx + 1)
+                : Array.from({ length: columnCount }, (_, colIdx) =>
+                    this.makePositionalParam(
+                      recordIdx * columnCount + colIdx + 1
+                    )
+                  )
+            )
+          )
       }
       let sql = baseSql + deletePattern
 
--- a/clients/typescript/src/migrators/query-builder/builder.ts
+++ b/clients/typescript/src/migrators/query-builder/builder.ts
@@ -289,7 +289,7 @@ export abstract class QueryBuilder {
 
   /**
    * Generates IN clause for a WHERE statement, checking that the given
-   * columns have a value present in the given tupleArgs array
+   * columns have a value present in the given args array
    *
    * The `args` array must be an array of values to compare each column to.
    *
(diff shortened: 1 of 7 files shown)


clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -376,23 +385,20 @@ export abstract class QueryBuilder {
    * @param suffixSql optional SQL string to append to each insert statement
    * @returns array of statements ready to be executed by the adapter
    */
-  public prepareDeleteBatchedStatements<T extends object>(
+  public prepareDeleteBatchedStatements(
     baseSql: string,
-    columns: Array<keyof T>,
-    records: T[],
+    columns: string[],
```
kevin-dp, line 390: I'd argue to keep the type parameter `T` because that way we can be sure that the provided columns are actual keys of the provided records.

kevin-dp: If we want we can do `T extends Row`

clients/typescript/src/migrators/query-builder/pgBuilder.ts
```diff
@@ -438,6 +438,20 @@ class PgBuilder extends QueryBuilder {
   makePositionalParam(i: number): string {
     return this.paramSign + i
   }
+
+  protected createInClause(
+    columns: string[],
+    args: string[] | string[][]
+  ): string {
+    const useTuples = typeof args[0] === 'object'
```
kevin-dp, line 446: Could you add a comment here explaining what's going on when the `args` array is 2-dimensional?
In which case would you pass a 2-dimensional array? of args?

clients/typescript/src/migrators/query-builder/sqliteBuilder.ts
```diff
@@ -312,6 +312,18 @@ class SqliteBuilder extends QueryBuilder {
   makePositionalParam(_i: number): string {
     return this.paramSign
   }
+
+  protected createInClause(
+    columns: string[],
+    args: (string | string[])[]
+  ): string {
+    const useTuples = typeof args[0] === 'object'
+    return `(${columns.map((c) => `"${c}"`).join(`, `)}) IN (${
```
kevin-dp, line 321: can map `quote` as you did in the PG query builder.

clients/typescript/test/migrators/builder.ts
```diff
@@ -428,19 +428,53 @@ export const builderTests = (test: TestFn<ContextType>) => {
         ? ['?', '?', '?', '?']
         : ['$1', '$2', '$3', '$4']
 
+    const withValues = builder.dialect !== 'SQLite'
+
     t.deepEqual(stmts, [
       {
-        sql: `DELETE FROM test WHERE ("a" = ${posArgs[0]} AND "b" = ${posArgs[1]}) OR ("a" = ${posArgs[2]} AND "b" = ${posArgs[3]})`,
+        sql: `DELETE FROM test WHERE ("a", "b") IN (${
```
kevin-dp, line 435: would be cleaner to modify the constant directly:

```ts
const withValues = builder.dialect !== 'SQLite' ? ' VALUES ' : ''
```

And then here and in the strings below:
```ts
`DELETE FROM test WHERE ("a", "b") IN (${withValues}(...))`
```

clients/typescript/test/migrators/builder.ts
```diff
@@ -428,19 +428,53 @@ export const builderTests = (test: TestFn<ContextType>) => {
...
+    ]
+    const stmts = builder.prepareDeleteBatchedStatements(
+      'DELETE FROM test WHERE',
+      ['a'],
+      data,
+      5 // at most 5 `?`s in one SQL statement, so we should see the split
+    )
+
+    const posArgs: string[] =
+      builder.dialect === 'SQLite'
```
kevin-dp, line 465: This is not the way to do it. This won't work if tomorrow we add another dialect.
The query builder abstraction has a `makePostionalParam` method.
So you can use `builder.makePositionalParam(i)` instead.
Also don't think this `posArgs` array is needed here. Can just inline it in the SQL statement.
Please also fix this in the rest of these tests.

clients/typescript/test/migrators/builder.ts
kevin-dp: Note that the query builder abstraction has a method `abstract pgOnly(query: string): string` that you can use for doing exactly this.

```ts
`DELETE FROM test WHERE ("a", "b") IN (${builder.pgOnly(" VALUES ")}(...))`
```

This comment applies to most tests you introduced here.

kevin-dp (changes requested):
I left some comments, mainly simplifications for the unit tests.
Also don't forget to add a changeset.

clients/typescript/src/migrators/query-builder/pgBuilder.ts
msfstef: I will add an explanatory comment on the abstract method's docstring to cover all cases - `args` is 2-dimensional if we're comparing  multiple columns against a list of rows basically.

clients/typescript/src/satellite/process.ts
```diff
@@ -1564,9 +1564,10 @@ export class SatelliteProcess implements Satellite {
 
     // Batch-delete shadow entries
     const stmts = this.builder.prepareDeleteBatchedStatements(
-      `DELETE FROM ${this.opts.shadowTable} WHERE `,
+      `DELETE FROM ${this.opts.shadowTable} WHERE`,
       ['namespace', 'tablename', 'primaryKey'],
-      fakeOplogEntries,
+      // force casting as only the three string entries will be used
+      fakeOplogEntries as unknown as Record<string, string>[],
```
icehaunter, line 1570: I'm not a fan of a forced cast here - we lose any benefit of a generic with a `keyof`. There is a better way to fix this by specifying `OplogEntry` to be a `type` instead of an interface.

```diff
diff --git a/clients/typescript/src/satellite/oplog.ts b/clients/typescript/src/satellite/oplog.ts
index bc748c02d..d88991fec 100644
--- a/clients/typescript/src/satellite/oplog.ts
+++ b/clients/typescript/src/satellite/oplog.ts
@@ -24,7 +24,7 @@ export type Tag = string
 export type ShadowKey = string
 
 // Oplog table schema.
-export interface OplogEntry {
+export type OplogEntry = {
   namespace: string
   tablename: string
   primaryKey: string // json object
diff --git a/clients/typescript/src/satellite/process.ts b/clients/typescript/src/satellite/process.ts
index 64523c32b..7b66cbde1 100644
--- a/clients/typescript/src/satellite/process.ts
+++ b/clients/typescript/src/satellite/process.ts
@@ -40,6 +40,7 @@ import {
   ReplicatedRowTransformer,
   DataGone,
   ServerTransaction,
+  Row,
 } from '../util/types'
 import { SatelliteOpts } from './config'
 import { Client, Satellite } from './index'
@@ -1566,8 +1567,7 @@ export class SatelliteProcess implements Satellite {
     const stmts = this.builder.prepareDeleteBatchedStatements(
       `DELETE FROM ${this.opts.shadowTable} WHERE`,
       ['namespace', 'tablename', 'primaryKey'],
-      // force casting as only the three string entries will be used
-      fakeOplogEntries as unknown as Record<string, string>[],
+      fakeOplogEntries,
       this.maxSqlParameters
     )
 
@@ -1589,7 +1589,7 @@ export class SatelliteProcess implements Satellite {
         ...this.builder.prepareDeleteBatchedStatements(
           `DELETE FROM ${fqtn} WHERE`,
           pkCols,
-          gone.map((x) => x.oldRecord) as Record<string, SqlValue>[],
+          gone.map((x) => x.oldRecord as Row),
           this.maxSqlParameters
         )
       )
```

icehaunter (changes requested):
small nitpicks, but GJ otherwise!

clients/typescript/test/migrators/builder.ts
```diff
@@ -361,18 +361,22 @@ export const builderTests = (test: TestFn<ContextType>) => {
...
+            )}, ${builder.makePositionalParam(2 * idx + 2)})`
+        ).join(', ')}`,
         args: [1, 2, 3, 4],
       },
       {
-        sql: `INSERT INTO test (a, b) VALUES (${posArgs[0]}, ${posArgs[1]})`,
+        sql: `INSERT INTO test (a, b) VALUES (${Array.from(
+          { length: 2 },
+          (_, idx) => builder.makePositionalParam(idx + 1)
+        ).join(', ')})`,
```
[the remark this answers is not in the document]
msfstef: I see your point (which is why I had it more hardcoded before) as this also introduces a dependency to `makePositionalParam`, but it does make the test more adaptable and ultimately what is being tested (the correct grouping of the positional parameters in an `IN` clause) is still covered so I think I'll leave it as is

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -302,10 +319,10 @@ export abstract class QueryBuilder {
    * @param suffixSql optional SQL string to append to each insert statement
    * @returns array of statements ready to be executed by the adapter
    */
-  prepareInsertBatchedStatements(
+  prepareInsertBatchedStatements<T extends Row>(
     baseSql: string,
-    columns: string[],
-    records: Record<string, SqlValue>[],
+    columns: Extract<keyof T, string>[],
```
kevin-dp, line 324: Extracting only the keys that are strings is neat but it seems unnecessary since `T extends Row` and `Row` is defined as `type Row = { [key: string]: SqlValue }`. But if you think it's best to have it for robustness in case we extend the definition of `Row` later, i'm fine with that too.

kevin-dp: What i meant with my previous comment was to define this function like this:
```ts
prepareInsertBatchedStatements<T extends Row>(
  baseSql: string,
  columns: Array<keyof T>,
  records: T[],
  ...
)
```

This makes sure that the strings in `columns` correspond to the keys of the rows passed in `records`.

clients/typescript/src/migrators/query-builder/sqliteBuilder.ts
```diff
@@ -312,6 +314,18 @@ class SqliteBuilder extends QueryBuilder {
   makePositionalParam(_i: number): string {
     return this.paramSign
   }
+
+  protected createInClause(
+    columns: string[],
+    args: (string | string[])[]
```
kevin-dp, line 320: This type is different from the type of `args` in `createInClause` in `pgBuilder.ts`. Is this intentional?

clients/typescript/test/migrators/builder.ts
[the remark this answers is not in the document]
kevin-dp: Indeed, we could have a separate unit test to ensure correct working of `makePositionalParam`.

clients/typescript/src/migrators/query-builder/builder.ts
```diff
@@ -287,6 +287,15 @@ export abstract class QueryBuilder {
     shadowTable: QualifiedTablename
   ): string
 
+  /**
+   * Generates IN clause for a WHERE statement, checking that the given
+   * columns have a value present in the given tupleArgs array
```
kevin-dp, line 292: i don't see a "tupleArgs array" on an "args" array.
So the `args` is basically either a row or a list of rows right?
And we check that the columns array matches one of those rows.

kevin-dp (approved):
Great job, left a few minor comments.

clients/typescript/src/migrators/query-builder/builder.ts
msfstef: whoops I'll fix the docstring!

yep the args are either a list of values or a list of rows, depending on whether you're inserting data to a single column or multiple columns

kevin-dp: Could you add this phrase to the docstring please? Found this more comprehensible :)

clients/typescript/src/migrators/query-builder/builder.ts
msfstef: Although this works for `prepareInsertBatchedStatements` it does not work for `prepareDeleteBatchedStatements` because I want the columns to be `string`, and TS casts `keyof Record<string, any>` to also include `number` and `symbol`.

For the delete batching I'll enforce the columns to be strings - the only reason this is not the case in the insert batching is because the insert batching adds the column names in the `sqlBase` already.

clients/typescript/src/migrators/query-builder/builder.ts
msfstef: The docstring is expanded below to describe the case for single and multiple columns as well

clients/typescript/test/migrators/builder.ts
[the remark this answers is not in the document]
msfstef: added a simple test for `makePositionalParam` so this should now be covered

PR #1397, opened 2024-06-24
feat (client): make converter API public

kevin-dp:
This PR extends the converter API with `encodeRow(s)`/`decodeRow(s)` methods and exports the API.
This API is a nice to have when dealing with raw queries and having to correctly serialise/deserialise rows to/from the database.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,6 +1,6 @@
 elixir 1.17.0-otp-27
 erlang 27.0
-nodejs 20.2.0
+nodejs 21.4.0
 pnpm 9.4.0
 # keep synchronised with version pinned in .github/workflows/satellite_proto.yml
 protoc 26.1
--- a/clients/typescript/src/client/conversions/converter.ts
+++ b/clients/typescript/src/client/conversions/converter.ts
@@ -1,3 +1,5 @@
+import { Row } from '../../util'
+import { AnyTableSchema } from '../model'
 import { PgType } from './types'
 
 export interface Converter {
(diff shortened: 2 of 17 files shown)


.tool-versions
```diff
@@ -1,6 +1,6 @@
 elixir 1.17.0-otp-27
 erlang 27.0
-nodejs 20.2.0
+nodejs 21.4.0
```
msfstef, line 3: might need to update the CI workflows to use this version as well in the test matrix, build, etc

clients/typescript/src/client/conversions/converter.ts
```diff
@@ -26,3 +64,34 @@ export interface Converter {
 export function isDataObject(v: unknown): boolean {
   return v instanceof Date || typeof v === 'bigint' || ArrayBuffer.isView(v)
 }
+
+export function mapRow<
+  T extends Record<string, unknown> = Record<string, unknown>
```
msfstef, line 69: nit: might be worth extracting `Record<string, unknown>` to a type alias like `RowRecord` or something as it's so common

clients/typescript/src/client/conversions/converter.ts
```diff
@@ -26,3 +64,34 @@ export interface Converter {
...
+>(
+  row: Record<string, unknown>,
+  tableSchema: Pick<AnyTableSchema, 'fields'>,
+  f: (v: any, pgType: PgType) => any
```
msfstef, line 73: nit: more explicit name like `convert`? `f` feels a little bit _too_ terse haha

clients/typescript/src/client/conversions/converter.ts
```diff
@@ -26,3 +64,34 @@ export interface Converter {
...
+
+  return mappedRow
+}
+
+export function mapRows<
+  T extends Record<string, unknown> = Record<string, unknown>
+>(
+  rows: Array<Record<string, unknown>>,
+  tableSchema: Pick<AnyTableSchema, 'fields'>,
+  f: (v: any, pgType: PgType) => any
```
msfstef, line 94: nit: same as above

clients/typescript/src/client/conversions/postgres.ts
```diff
@@ -11,7 +12,7 @@ import { PgBasicType, PgDateType, PgType } from './types'
  * Currently, no conversions are needed for the data types we support.
  */
 
-function toPostgres(v: any, pgType: PgType): any {
+export function toPostgres(v: any, pgType: PgType): any {
```
msfstef, line 15: nit: why are these exported, they are exported via the converter interface no? (same for SQLite)

clients/typescript/src/client/model/table.ts
```diff
@@ -221,6 +225,10 @@ export class Table<
    * As such, one can compose these methods arbitrarily and then run them inside a single transaction.
    */
 
+  /**
```
msfstef, line 228: nit: might need more rebasing? haha

msfstef (approved):
Looks good, nice work! Left a few minor comments

KyleAMathews:

PR #1398, opened 2024-06-24
chore (client): deprecation warnings for DAL

kevin-dp:
This PR adds deprecation warnings on all DAL methods.


msfstef (approved):

kevin-dp:
Plan is to only merge this into main for the release where we want the deprecation warnings to occur.

PR #1406, opened 2024-06-27
Server to client error messages

robacourt:
https://linear.app/electric-sql/issue/VAX-1983/improve-error-reporting-from-server-to-client

This PR add the ability of the client to show friendly error messages when the error has originated on the server.

For example:

```
an error occurred in satellite: INVALID_REQUEST Unexpected value for int2 column: 32769
```

At this stage I've not tried to hide any noise such as the stack trace in these scenarios. So the user in fact sees all of this:
```
an error occurred in satellite: INVALID_REQUEST Unexpected value for int2 column: 32769
Client can't connect with the server after a fatal error. This can happen due to divergence between local client and server. Use developer tools to clear the local database, or delete the database file. We're working on tools to allow recovering the state of the local database.
/Users/rob/src/electric-sql/electric/clients/typescript/dist/satellite/error.js:35
  return new SatelliteError(
         ^


SatelliteError: Fatal error: Unexpected value for int2 column: 32769. Check log for more information
    at wrapFatalError (/Users/rob/src/electric-sql/electric/clients/typescript/dist/satellite/error.js:35:10)
    at SatelliteProcess._handleOrThrowClientError (/Users/rob/src/electric-sql/electric/clients/typescript/dist/satellite/process.js:542:13)
    at SatelliteProcess._handleClientError (/Users/rob/src/electric-sql/electric/clients/typescript/dist/satellite/process.js:525:10)
    at <anonymous> (/Users/rob/src/electric-sql/electric/clients/typescript/dist/util/asyncEventEmitter.js:103:17)
    at Array.map (<anonymous>)
    at AsyncEventEmitter.processQueue (/Users/rob/src/electric-sql/electric/clients/typescript/dist/util/asyncEventEmitter.js:101:32)
    at AsyncEventEmitter.enqueueEmit (/Users/rob/src/electric-sql/electric/clients/typescript/dist/util/asyncEventEmitter.js:129:12)
    at SatelliteClient.handleError (/Users/rob/src/electric-sql/electric/clients/typescript/dist/satellite/client.js:725:18)
    at Object.SatErrorResp (/Users/rob/src/electric-sql/electric/clients/typescript/dist/satellite/client.js:98:37)
    at SatelliteClient.handleIncoming (/Users/rob/src/electric-sql/electric/clients/typescript/dist/satellite/client.js:801:47) {
  code: 12
}
```

I've also added friendly error message generation on the server for data validation errors as an example of how you would implement the friendly error messages. Once the permissions system has been added, rejected writes could also generate a friendly error message in a similar way.

The client will now also see error messages that were already being set in the SatErrorResp such as:
- Postgres is unavailable
- Acknowledged unknown txn
- Connection to central database failed, cannot continue the replication because of possible consistency issues

### DIFF
--- a/components/electric/lib/electric/satellite/serialization.ex
+++ b/components/electric/lib/electric/satellite/serialization.ex
@@ -611,7 +611,7 @@ defmodule Electric.Satellite.Serialization do
         val
 
       _ ->
-        raise DataValidationError, message: "Non integer value given for #{type} column: #{val}"
+        raise DataValidationError, message: "Non-integer value given for #{type} column: #{val}"
     end
   end
 
--- a/components/electric/lib/electric/satellite/serialization.ex
+++ b/components/electric/lib/electric/satellite/serialization.ex
@@ -746,28 +746,13 @@ defmodule Electric.Satellite.Serialization do
 
   defp assert_valid_month!(month) when month in 1..12, do: :ok
 
-  defp assert_valid_month!(month),
-    do: raise(DataValidationError, message: "Month out of range: #{month}")
-
   defp assert_valid_day!(day) when day in 1..31, do: :ok
-
-  defp assert_valid_day!(day),
-    do: raise(DataValidationError, message: "Day out of range: #{day}")
 
   defp assert_valid_hours!(hours) when hours in 0..23, do: :ok
 
-  defp assert_valid_hours!(hours),
-    do: raise(DataValidationError, message: "Hours out of range: #{hours}")
-
   defp assert_valid_minutes!(minutes) when minutes in 0..59, do: :ok
 
-  defp assert_valid_minutes!(minutes),
-    do: raise(DataValidationError, message: "Minutes out of range: #{minutes}")
-
   defp assert_valid_seconds!(seconds) when seconds in 0..59, do: :ok
-
-  defp assert_valid_seconds!(seconds),
-    do: raise(DataValidationError, message: "Seconds out of range: #{seconds}")
 
   defp assert_valid_fractional_seconds!(""), do: :ok
 
@@ -776,9 +761,6 @@ defmodule Electric.Satellite.Serialization do
     _ = String.to_integer(fs_str)
     :ok
   end
-
-  defp assert_valid_fractional_seconds!(val),
-    do: raise(DataValidationError, message: "Fractional seconds out of range: #{val}")
 
   defp assert_float_in_range!(_num, :float8), do: :ok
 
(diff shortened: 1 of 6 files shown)


components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -582,6 +583,8 @@ defmodule Electric.Satellite.Serialization do
     _ = Date.from_iso8601!(val)
 
     val
+  rescue
+    _ -> raise DataValidationError, message: "Unexpected value for :date column: #{inspect(val)}"
```
alco, line 587: This ignored the specific error message from any of the function calls above and substitutes that with this generic message. That's not a problem, IMO, but then we don't need those extra catch-all clauses you added below, such as `Month our of range` and others.

Also, this needs to be a `reraise` to preserve the original stacktrace:
```suggestion
    _ -> reraise DataValidationError, [message: "Unexpected value for :date column: #{inspect(val)}"], __STACKTRACE__
```

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -592,16 +595,25 @@ defmodule Electric.Satellite.Serialization do
...
-    |> String.to_integer()
-    |> assert_valid_integer!(type)
+    case Integer.parse(val) do
+      {int, ""} ->
+        assert_valid_integer!(int, type)
+        val
 
-    val
+      _ ->
+        raise DataValidationError, message: "Non integer value given for #{type} column: #{val}"
```
alco, line 604:
```suggestion
        raise DataValidationError, message: "Non-integer value given for #{type} column: #{val}"
```

components/electric/lib/electric/satellite/serialization.ex
```diff
@@ -621,36 +633,68 @@ defmodule Electric.Satellite.Serialization do
     _ = Time.from_iso8601!(val)
 
     val
+  rescue
+    _ -> raise DataValidationError, message: "Unexpected value for :time column: #{inspect(val)}"
```
alco, line 637: The same considerations as for the other use of `rescue` above apply here 

```suggestion
    _ -> reraise DataValidationError, [message: "Unexpected value for :time column: #{inspect(val)}"], __STACKTRACE_
```

alco (changes requested):

kevin-dp:
Minor comment but i think the error that is displayed on the client should make very clear that this is a server-side error, otherwise it will confuse people and they will think there is a problem with the client.

components/electric/lib/electric/satellite/serialization.ex
robacourt: Great points. I've left the raise in `assert_valid_year!/1` since that is used without being rescued in `decode_column_value!/2`

robacourt:
> Minor comment but i think the error that is displayed on the client should make very clear that this is a server-side error, otherwise it will confuse people and they will think there is a problem with the client.

The error message is now:

`an error occurred in satellite: INVALID_REQUEST Server error: Unexpected value for int2 column: 32769`

Hope that's clean enough @kevin-dp ?

PR #1411, opened 2024-06-28
fix (toolbar): modify drivers dependency to be a workspace dependency

kevin-dp:


alco (approved):
LGTM with one caveat:

I have updated pnpm to version 9 on CI, and the pnpm-lockfile has been updated accordingly in `main`. Judging from the diff, you still have pnpm v6. Would you mind upgrading and reruning `pnpm i` to fix the lockfile diff?

PR #1415, opened 2024-07-01
chore(linearlite): Fix linearlite deps

msfstef:
I've updated the linearlite dependency as it was failing to build and also modified the toolbar workspace dependencies to allow for the closest patch version rather than the exact one.

We might want to do something similar for the CLI and Drivers packages but since they are more tightly coupled we can keep them as `workspace:*` rather than `workspace:~` - once we move to v1 we might want `worksapce:^`

### DIFF
--- a/.changeset/forty-tables-study.md
+++ b/.changeset/forty-tables-study.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/debug-toolbar": patch
+---
+
+Use looser peer dependencies for electric packages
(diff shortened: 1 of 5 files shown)


kevin-dp (approved):
LGTM.
`"electric-sql": "workspace:~"` dependency may be slightly dangerous as we are in alpha and may introduce a breaking change on the client with only a patch bump. Need to remember to specify the dependency more specifically when that happens.

PR #1481, opened 2024-08-06
Fix: rename action header -> operation to match PG

### LINKED ISSUE #1437: Standardize on "operation" vs. "action"
My prototype of the HTTP API used "action" as the key for messages whether they contained an insert or update or delete. But this isn't what is used in Postgres so we should change that to "operation".

KyleAMathews:
Fixes #1437


examples/nextjs-example/tsconfig.tsbuildinfo
alco, line 1: This looks like a generated file. Do we need it in the Git repo?

robacourt:
Sorry Kyle, I've merged #1485 and it conflict with this. You'll need to rebase and there's a few mentions of `action` in my PR

examples/nextjs-example/tsconfig.tsbuildinfo
KyleAMathews: hmm no, I'll remove that in a separate PR as it came in with the nextjs one.

KyleAMathews: fix: https://github.com/electric-sql/electric/pull/1491

Issue #1436, opened 2024-08-05
Cleanup new website a bit


KyleAMathews:
Comment out links to:
- Writing clients (we'll get this eventually but it's not critical right now)
- Connectors (they don't exist yet so let's not worry about documenting them atm)

Fill out "Shapes" page — there's good language on the old website we can copy.

KyleAMathews:
This issue is pretty vague — @samwillis want to create a new issue with your ideas to improve the typescript client page?

PR #1484, opened 2024-08-06
chore: cleanup / improve docs site

KyleAMathews:
Partially address #1436


docs/api/integrations/react.md
```diff
@@ -0,0 +1,39 @@
...
+
+Example usage in a component.
+```tsx
+import { useShape } from "@electric-sql/react"
+
+export default function MyComponent () {
+  const { isUpToDate, data as fooData } = useShape({
+    shape: { table: `foo` },
+    baseUrl: "http://my-api.com/",
+  })
```
samwillis, line 17:
```suggestion
export default function MyComponent() {
  const { isUpToDate, data: fooData } = useShape({
    url: `http://localhost:3000/v1/shape/foo`,
  })
```

I think from reading the source `useShape` inherits the option of `ShapeStream` so it requires the full url to the shape.

Somewhat off topic:

Looking at this example I'm still not 100% sold on the change we made from `baseUrl` and `table` to just `url`, particularly in this context. I would be much nicer if the user was able to set the `baseUrl` centrally, then have a declarative shape where they want it.

samwillis (approved):

docs/api/integrations/react.md
KyleAMathews: I dunno — it's not hard to e.g. create a function which let's you generate the URL. People will need something like that anyways as they'll have different URLs in dev/staging/prod. I've been creating a `shape-definitions.ts` file for my playing around as you almost always end up wanting the same shape in multiple components. This is handy too as you can let components pass in an argument for a selector if they want to filter the shape array down e.g. https://github.com/KyleAMathews/vite-remix-electric-sql-starter/blob/01b75fb3111d4feb704ca7a009256caa1a8d4ed6/app/routes/contacts.%24id.tsx#L85-L94

PR #1485, opened 2024-08-07
feat(sync-service): Encode JSON encode on write rather than read

robacourt:
JSON encoding on write rather than read to reduce memory footprint. For example the memory use of an initial sync of a 35MB/200k row table can be reduced from 50MB to 25MB on the first initial sync (which includes the encoding and the writing to storage) and to 6MB on subsequent initial syncs (where we just read from storage and there's no encoding).

This change also allows some simplifications, so I have included them as refactoring in this PR:
- Logic to do with the structure and how to create Log Items has been consolidated to a new module `LogItems`
- The `prepared_change` intermediate state has been removed. The transformation `Changes > list(prepared_change) > LogItems` now is simply `Changes > LogItems` which makes the code more readable ( `prepared_change` was a 5 tuple rather than a nicely labeled map) and easier to reason about (since there's one less data structure to worry about)
- The snapshot row intermediate state has a few less references, ideally I'd like to get encapsulate it into a single module, but for now it's mainly `Shapes.Querying` that creates it and `LogItems.from_snapshot_row/4` that reads it.
-  Some logic duplicated in InMemoryStorage and CubDbStorage has been consolidated.

I've rebased this PR into 2 commits, the functional change and the refactoring, to aid with reviewing.

### DIFF
--- a/packages/sync-service/lib/electric/log_items.ex
+++ b/packages/sync-service/lib/electric/log_items.ex
@@ -1,0 +1,121 @@
+defmodule Electric.LogItems do
+  alias Electric.Replication.Changes
+  alias Electric.Replication.LogOffset
+  alias Electric.Shapes.Shape
+  alias Electric.Utils
+
+  @moduledoc """
+  Defines the structure and how to create the items in the log that the electric client reads.
+
+  The log_item() data structure is a map for ease of consumption in the Elixir code, 
+  however when JSON encoded it's the format that the electric client accepts.  
+  """
+
+  @type log_item :: %{
+          key: String.t(),
+          value: map(),
+          headers: map(),
+          offset: LogOffset.t()
+        }
+
+  @spec from_change(
+          Changes.data_change(),
+          txid :: non_neg_integer() | nil,
+          pk_cols :: [String.t()]
+        ) :: [log_item(), ...]
+  def from_change(%Changes.NewRecord{} = change, txid, _) do
+    [
+      %{
+        key: change.key,
+        value: change.record,
+        headers: %{action: :insert, txid: txid, relation: Tuple.to_list(change.relation)},
+        offset: change.log_offset
+      }
+    ]
+  end
+
+  def from_change(%Changes.DeletedRecord{} = change, txid, pk_cols) do
+    [
+      %{
+        key: change.key,
+        value: take_pks_or_all(change.old_record, pk_cols),
+        headers: %{action: :delete, txid: txid, relation: Tuple.to_list(change.relation)},
+        offset: change.log_offset
+      }
+    ]
+  end
+
+  # `old_key` is nil when it's unchanged. This is not possible when there is no PK defined.
+  def from_change(%Changes.UpdatedRecord{old_key: nil} = change, txid, pk_cols) do
+    [
+      %{
+        key: change.key,
+        value: Map.take(change.record, Enum.concat(pk_cols, change.changed_columns)),
+        headers: %{action: :update, txid: txid, relation: Tuple.to_list(change.relation)},
+        offset: change.log_offset
+      }
+    ]
+  end
+
+  def from_change(%Changes.UpdatedRecord{} = change, txid, pk_cols) do
+    [
+      %{
+        key: change.old_key,
+        value: take_pks_or_all(change.old_record, pk_cols),
+        headers: %{
+          action: :delete,
+          txid: txid,
+          relation: Tuple.to_list(change.relation),
+          key_change_to: change.key
+        },
+        offset: change.log_offset
+      },
+      %{
+        key: change.key,
+        value: change.record,
+        headers: %{
+          action: :insert,
+          txid: txid,
+          relation: Tuple.to_list(change.relation),
+          key_change_from: change.old_key
+        },
+        offset: LogOffset.increment(change.log_offset)
+      }
+    ]
+  end
+
+  defp take_pks_or_all(record, []), do: record
+  defp take_pks_or_all(record, pks), do: Map.take(record, pks)
+
+  @spec from_snapshot_row_stream(
+          row_stream :: Stream.t(list()),
+          offset :: LogOffset.t(),
+          shape :: Shape.t(),
+          query_info :: %Postgrex.Query{}
+        ) :: log_item()
+  def from_snapshot_row_stream(row_stream, offset, shape, query_info) do
+    Stream.map(row_stream, &from_snapshot_row(&1, offset, shape, query_info))
+  end
+
+  defp from_snapshot_row(row, offset, shape, query_info) do
+    value = value(row, query_info)
+
+    key = Changes.build_key(shape.root_table, value, Shape.pk(shape))
+
+    %{
+      key: key,
+      value: value,
+      headers: %{action: :insert},
+      offset: offset
+    }
+  end
+
+  defp value(row, %Postgrex.Query{columns: columns, result_types: types}) do
+    [columns, types, row]
+    |> Enum.zip_with(fn
+      [col, Postgrex.Extensions.UUID, val] -> {col, Utils.encode_uuid(val)}
+      [col, _, val] -> {col, to_string(val)}
+    end)
+    |> Map.new()
+  end
+end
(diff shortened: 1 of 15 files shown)


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -10,7 +10,7 @@ defmodule Electric.Plug.ServeShapePlug do
   @before_all_offset LogOffset.before_all()
 
   # Control messages
-  @up_to_date [%{headers: %{control: "up-to-date"}}]
+  @up_to_date [Jason.encode!(%{headers: %{control: "up-to-date"}})]
   @must_refetch [%{headers: %{control: "must-refetch"}}]
```
alco, line 14: I think we should either encode both here, or leave a comment explaining why only one of the control headers is encoded to JSON right here.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -10,7 +10,7 @@ defmodule Electric.Plug.ServeShapePlug do
   @before_all_offset LogOffset.before_all()
 
   # Control messages
-  @up_to_date [%{headers: %{control: "up-to-date"}}]
+  @up_to_date [Jason.encode!(%{headers: %{control: "up-to-date"}})]
```
alco, line 13: I'd call this `@up_to_date_json` for readability.

packages/sync-service/test/electric/shape_cache/storage_implementations_test.exs
```diff
@@ -22,6 +22,7 @@ defmodule Electric.ShapeCache.StorageImplimentationsTest do
     }
   }
   @snapshot_offset LogOffset.first()
+  @snapshot_offset_encoded @snapshot_offset |> LogOffset.to_iolist() |> :erlang.iolist_to_binary()
```
alco, line 25: This can be written as `to_string(@snapshot_offset)`.

packages/sync-service/test/electric/shape_cache/storage_implementations_test.exs
```diff
@@ -33,7 +34,8 @@ defmodule Electric.ShapeCache.StorageImplimentationsTest do
     [<<2::128>>, "row2"]
   ]
 
-  for module <- [InMemoryStorage, CubDbStorage] do
+  for module <- [CubDbStorage] do
+    # for module <- [InMemoryStorage, CubDbStorage] do
```
alco, line 38: Why is this no longer testing `InMemoryStorage`?

packages/sync-service/test/electric/shape_cache/storage_implementations_test.exs
```diff
@@ -122,18 +124,18 @@ defmodule Electric.ShapeCache.StorageImplimentationsTest do
...
-                   value: %{"id" => "00000000-0000-0000-0000-000000000002", "title" => "row2"},
-                   key: ~S|"public"."the-table"/"00000000-0000-0000-0000-000000000002"|,
-                   headers: %{action: :insert}
+                   "offset" => @snapshot_offset_encoded,
+                   "value" => %{"id" => "00000000-0000-0000-0000-000000000002", "title" => "row2"},
+                   "key" => ~S|"public"."the-table"/"00000000-0000-0000-0000-000000000002"|,
+                   "headers" => %{"action" => "insert"}
                  }
-               ] = Enum.to_list(stream)
+               ] = Enum.map(stream, &Jason.decode!/1)
```
alco, line 138: Nit: could have used `Jason/decode!(&1, keys: :atoms)` here to avoid rewriting the maps in the assertion above.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -308,7 +308,14 @@ defmodule Electric.Plug.ServeShapePlug do
...
       hold_until_change(conn, shape_id)
     else
-      send_resp(conn, 200, Jason.encode_to_iodata!(log ++ @up_to_date))
+      send_resp(
+        conn,
+        200,
+        [log, @up_to_date]
+        |> Stream.concat()
+        |> to_json_stream()
+        |> Enum.to_list()
```
alco, line 317: I realize it's the wrong PR for such a comment, but this realization of a stream into a list looks out of place now. It feels like we should use `send_stream()` in this case too. Perhaps in a follow-up PR?

alco (commented):
I have reviewed the first commit.

Will take a separate look at the refactoring and get back to you.

packages/sync-service/lib/electric/log_items.ex
```diff
@@ -0,0 +1,121 @@
...
+          Changes.data_change(),
+          txid :: non_neg_integer() | nil,
+          pk_cols :: [String.t()]
+        ) :: [log_item(), ...]
+  def from_change(%Changes.NewRecord{} = change, txid, _) do
+    [
+      %{
+        key: change.key,
+        value: change.record,
+        headers: %{action: :insert, txid: txid, relation: Tuple.to_list(change.relation)},
```
alco, line 31: Heads up! There's a concurrent PR from Kyle where action is renamed to operation - https://github.com/electric-sql/electric/pull/1481

packages/sync-service/lib/electric/log_items.ex
```diff
@@ -0,0 +1,121 @@
...
+  alias Electric.Replication.Changes
+  alias Electric.Replication.LogOffset
+  alias Electric.Shapes.Shape
+  alias Electric.Utils
+
+  @moduledoc """
+  Defines the structure and how to create the items in the log that the electric client reads.
+
+  The log_item() data structure is a map for ease of consumption in the Elixir code, 
+  however when JSON encoded it's the format that the electric client accepts.
```
alco, line 11: I'm not seeing any JSON encoding in this module. This comment can either be dropped or needs more context.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
robacourt: Good idea. Done.

packages/sync-service/test/electric/shape_cache/storage_implementations_test.exs
robacourt: Lovely, thanks. Done.

packages/sync-service/test/electric/shape_cache/storage_implementations_test.exs
robacourt: Oops! Good spot. Done.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
alco: Sorry, I should've clarified that I was suggesting this in case you choose to only keep this attributed JSON-encoded. Since you've made both control messages JSON-encoded now, all is good.

packages/sync-service/test/electric/shape_cache/storage_implementations_test.exs
robacourt: Oh, nice idea. Thanks. Done.

packages/sync-service/lib/electric/log_items.ex
robacourt: Hm. I'm trying to convey that this module defines the JSON structure even though it doesn't do the JSON encoding. What if I changed `when JSON encoded` to `when JSON encoded (not done in this module)` would that help?

packages/sync-service/lib/electric/log_items.ex
robacourt: Noted. Thanks!

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
robacourt: Yeah, I agree. It's an easy enough change so I've added it to this PR.

packages/sync-service/lib/electric/log_items.ex
robacourt: Done

alco: I think I understand what you meant now, after reading your comment.

What about this version:
>  The log_item() data structure is a map for ease of consumption in the Elixir code, its structure matches the one used for JSON-encoded client responses in the HTTP API.

PR #1486, opened 2024-08-07
fix: Completely disable ReplicationClient's auto-reconnection logic

alco:
Some changes to make some types of test failure less likely.

Recent example - https://github.com/electric-sql/electric-next/actions/runs/10256059321/job/28374360808?pr=254. These happen occasionally but are impossible to reproduce deliberately. I suspect the issue is related to the use of persistent slot and auto-reconnection logic that results Postgres delivering a logical message with a lower WAL offset than the one in `ReplicationClient`'s state.


robacourt (approved):
Yeah, seems like the right move. Good work!

Issue #1472, opened 2024-07-12
Add initial observability


balegas:
Add tracing and metrics to the system with OpenTelemetry and setup a backend for collecting data.

Document how to run Electric w/ observability.

PR #1487, opened 2024-08-07
Tracing and host metrics with Opentelemetry

alco:
- [x] use otel collector to collect and export metric
- [x] add new docker-compose for telemetry services
- [x] ~~Do something with statsd metrics~~ collect and export metrics with openTelemetry collector
- [x] Add other exporters for local use
- [x] Learn what was done in our prev observability [PR](https://github.com/electric-sql/electric/commit/febb710e97cc90ccb4cc76e505235989c7e73c96), which seems a lot more comprehensive
- [x] Need to document how to setup
- [x] Need to replace host metric collector with prometheus-based one
- [x] Discuss with team issue with opentelemetry_bandit

Closes https://github.com/electric-sql/electric/issues/1472.


balegas:
Deciding not to add the bandit plugin atm. Follow up PR with better usage of spans when we feel its necesasry.

alco:
@balegas I've just finished testing this. Great work, the setup is working with no issues out of the box, I only had to make a few adjustments.

For some reason I don't see any traces in Honeycomb, only metrics scraped by Prometheus. I have verified that Electric exports trace data to the collector via the `debug` exporter.

Let's merge this and investigate further once we try collecting trace data for a real use case.

robacourt (approved):
Great work!

balegas:
Cool. Thanks guys.
@alco, what has happened to me was that sometimes I would not cleanup the local shapes cache and then the span was never triggered because it is only on the path of creating a snapshot for a shape.

PR #1489, opened 2024-08-07
fix(sync-server): Fix error that occurs when `/shape` response stream is closed before it is complete

### LINKED ISSUE #1440: Error syncing tables from certain size
Commit: 5522305209a46a2ce7516a19676e5a4a700014ad

How I reproduce it

```bash
docker compose -f ${SYNC_DIR}/dev/docker-compose.yml up -d
rm -fr shapes && iex -S mix

# init linearlite state
npm run db:migrate && ISSUES_TO_LOAD=1000 npm run db:load-data

curl -I -X GET http://localhost:3000/v1/shape/issue\?offset\=-1
```

```
13:46:49.899 [error] ** (Plug.Conn.WrapperError) ** (CaseClauseError) no case clause matching: {:error, "closed"}
    (electric 0.2.4) lib/electric/plug/serve_shape_plug.ex:332: anonymous fn/2 in Electric.Plug.ServeShapePlug.send_stream/3
    (elixir 1.17.2) lib/stream.ex:1879: anonymous fn/3 in Enumerable.Stream.reduce/3
    (elixir 1.17.2) lib/stream.ex:273: anonymous fn/4 in Stream.chunk_while_fun/2
    (elixir 1.17.2) lib/stream.ex:1079: Stream.do_transform_each/3
    (elixir 1.17.2) lib/enum.ex:4858: Enumerable.List.reduce/3
    (elixir 1.17.2) lib/stream.ex:1027: Stream.do_transform_inner_list/7
    (elixir 1.17.2) lib/stream.ex:1052: Stream.do_transform_inner_enum/7
    (elixir 1.17.2) lib/stream.ex:1891: Enumerable.Stream.do_each/4
    (elixir 1.17.2) lib/enum.ex:2585: Enum.reduce_while/3
    (electric 0.2.4) lib/electric/plug/serve_shape_plug.ex:1: Electric.Plug.ServeShapePlug.plug_builder_call/2
    (electric 0.2.4) deps/plug/lib/plug/router.ex:246: anonymous fn/4 in Electric.Plug.Router.dispatch/2
    (telemetry 1.2.1) /Users/vbalegas/workspace/electric-next/packages/sync-service/deps/telemetry/src/telemetry.erl:321: :telemetry.span/3
    (electric 0.2.4) deps/plug/lib/plug/router.ex:242: Electric.Plug.Router.dispatch/2
    (electric 0.2.4) lib/electric/plug/router.ex:1: Electric.Plug.Router.plug_builder_call/2
    (bandit 1.5.5) lib/bandit/pipeline.ex:124: Bandit.Pipeline.call_plug!/2
    (bandit 1.5.5) lib/bandit/pipeline.ex:36: Bandit.Pipeline.run/4
    (bandit 1.5.5) lib/bandit/http1/handler.ex:12: Bandit.HTTP1.Handler.handle_data/3
    (bandit 1.5.5) lib/bandit/delegating_handler.ex:18: Bandit.DelegatingHandler.handle_data/3
    (bandit 1.5.5) /Users/vbalegas/workspace/electric-next/packages/sync-service/deps/thousand_island/lib/thousand_island/handler.ex:411: Bandit.DelegatingHandler.handle_continue/2
    (stdlib 6.0.1) gen_server.erl:2163: :gen_server.try_handle_continue/3
    (stdlib 6.0.1) gen_server.erl:2072: :gen_server.loop/7
    (stdlib 6.0.1) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
```

I still get data. Not sure if all, but response looks okay.

```
HTTP/1.1 200 OK
transfer-encoding: chunked
date: Fri, 02 Aug 2024 12:50:26 GMT
cache-control: max-age=1, stale-while-revalidate=3
x-request-id: F-fqU3O1Y71-fa8AAAAi
server: ElectricSQL/0.2.4
access-control-allow-origin: *
access-control-expose-headers: *
access-control-allow-methods: GET, POST, OPTIONS
content-type: application/json; charset=utf-8
x-electric-shape-id: 44524560-1722602809857
x-electric-chunk-last-offset: 0_0
x-electric-schema: {"created":{"type":"timestamptz"},"description":{"type":"text"},"id":{"type":"uuid","pk_index":0},"kanbanorder":{"type":"text"},"modified":{"type":"timestamptz"},"priority":{"type":"text"},"status":{"type":"text"},"title":{"type":"text"},"username":{"type":"text"}}
etag: 44524560-1722602809857:-1:0_0
```

robacourt:
This closes #1440

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -330,12 +330,6 @@ defmodule Electric.Plug.ServeShapePlug do
         {:ok, conn} ->
           {:cont, conn}
 
-        # The documented way to handle closed connections
-        # See https://hexdocs.pm/plug/Plug.Conn.html#chunk/2
-        {:error, :closed} ->
-          {:halt, conn}
-
-        # This also occurs, for example when `curl --head` is used to call the endpoint
         {:error, "closed"} ->
           {:halt, conn}
       end
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -332,6 +332,10 @@ defmodule Electric.Plug.ServeShapePlug do
 
         {:error, "closed"} ->
           {:halt, conn}
+
+        {:error, reason} ->
+          Logger.error("Error while streaming response: #{inspect(reason)}")
+          {:halt, conn}
       end
     end)
   end
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -330,7 +330,13 @@ defmodule Electric.Plug.ServeShapePlug do
         {:ok, conn} ->
           {:cont, conn}
 
+        # The documented way to handle closed connections
+        # See https://hexdocs.pm/plug/Plug.Conn.html#chunk/2
         {:error, :closed} ->
+          {:halt, conn}
+
+        # This also occurs, for example when `curl --head` is used to call the endpoint
+        {:error, "closed"} ->
           {:halt, conn}
       end
     end)
--- a/.changeset/fifty-frogs-warn.md
+++ b/.changeset/fifty-frogs-warn.md
@@ -1,0 +1,6 @@
+---
+"@core/sync-service": patch
+---
+
+Fix error that occurs when a `/shape` response stream is closed before it is complete,
+for example when `curl --head` is used to call the endpoint.


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -330,8 +330,14 @@ defmodule Electric.Plug.ServeShapePlug do
...
           {:cont, conn}
 
+        # The documented way to handle closed connections
+        # See https://hexdocs.pm/plug/Plug.Conn.html#chunk/2
         {:error, :closed} ->
           {:halt, conn}
+
+        # This also occurs, for example when `curl --head` is used to call the endpoint
+        {:error, "closed"} ->
+          {:halt, conn}
```
robacourt, line 340: I'm not sure how to test this. The Plug testing framework doesn't seem to have the ability to test this scenario. It's not really an integration test as our client doesn't elicit this behaviour at present. I could write a whole new type of test but it doesn't seem worth it for this one scenario.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -330,8 +330,14 @@ defmodule Electric.Plug.ServeShapePlug do
         {:ok, conn} ->
           {:cont, conn}
 
+        # The documented way to handle closed connections
+        # See https://hexdocs.pm/plug/Plug.Conn.html#chunk/2
```
alco, line 334: Technically, the documented return type is `{:error, term}`, so Plug does not restrict what can be returned from `chunk/2`. I believe that the `{:error, :closed}` tuple in the example code there is either valid for the Cowboy adapter (which was the first one used with Plug historically) or is simply an illustrative example.

Regardless, we should be more liberal in the types of error reasons we accept here. Let's just match against any term and log it as an error.

For the curious, since we're using Bandit instead of Cowboy, here's the underlying implemenation of `chunk` that gets called - https://github.com/mtrudel/bandit/blob/6dd82d9625ddce43685f91f8ab5a3150e1fd6a61/lib/bandit/adapter.ex#L190-L205.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
alco: From my perspective, an integration tests does not necessarily involve multiple components. I've written the first integration test under `integration-tests/` to verify the behaviour of the server which couldn't be verified inside it: the fact that it halts execution when it encounters a specific error.

But let's not have an unwritten test block this PR from getting merged. Manual testing is good for now.

alco (changes requested):

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
robacourt: Ah, make sense! I might get rid of the `:closed` case then.

As for handling other cases, what's wrong with let it crash?

alco: I believe https://github.com/electric-sql/electric/issues/1440 was raised as an issue to begin with because just letting it crash resulted in a hard-to-read error that seemed to point to a bug in the server.

To fix that, we can match on `{:error, reason}` and log the reason in a way that won't confuse us when we see it in the logs.

robacourt: Ok, good point. It would make for an easier to read error message. I'll update it.

alco: Ah, I think you meant that if there's an error with a reason other than `"closed`", we should let it crash to see if it's pointing at an actual bug?

Fair enough. Though I would love to have some explanation for this `"closed"` reason we're handling in the code. Could be a URL to this issue.

PR #1508, opened 2024-08-10
Yjs Electric provider example

balegas:
A basic codemirror example with Yjs using Electric for sync.

I have removed a lot of stuff I've done in this example to just have the bare minimum. The provider shall work with persistence backend and broadcasting, but right now I removed all that code to keep the ElectricProvider to the bare minimum. I'll add back support for that when I publish the provider.

TODO
- [ ] Readme
- [ ] Review and publish ElectricProvider
- [x] Better App
- [x] Deployment: Provide the pooled url; fix issues with serverless deployment or setup a NextJs server


examples/nextjs-yjs-codemirror/README.md
```diff
@@ -0,0 +1,22 @@
+# Basic Remix example
```
KyleAMathews, line 1: README needs updated

examples/nextjs-yjs-codemirror/app/y-electric.js
```diff
@@ -0,0 +1,409 @@
...
+ */
+const setupShapeStream = (provider) => {
+  if (provider.shouldConnect && provider.operationsStream === null) {
+    provider.connecting = true
+    provider.connected = false
+    provider.synced = false
+
+    provider.operationsStream = new ShapeStream({
+      url: provider.operationsUrl,
+      signal: new AbortController().signal,
```
KyleAMathews, line 97: signals are optional. There's no need to pass one in unless you want the option to abort the stream at some point. Since you're just creating it here and passing it in, you can't do that so might as well remove it.

KyleAMathews (commented):
nice work!

balegas:
**Experimental compaction of YDoc**
I've implemented compaction on the server by creating a new Ydoc, truncating the table and storing the new Ydoc. The truncation causes a shape rotation that will essentially clear the shape and start fresh from the Ydoc checkpoint. 

This obviously isn't a complete implementation as the TRUNCATE also deletes all other Ydocs that we might be storing :D. The idea was to demonstrate the use of shape rotation to clear the log of operations. We should think if we want to expose the delete shape endpoint on the backend or not. The proxy could disallow any requests coming from the user.

examples/yjs-provider/app/y-electric.ts
```diff
@@ -0,0 +1,336 @@
...
+        process.on(`exit`, () => this.exitHandler!())
+      }
+    }
+
+    if (options.connect) {
+      this.connect()
+    }
+  }
+
+  private get operationsUrl() {
```
kevin-dp, line 119: No need to be getters. We never change the `serverUrl` so these urls can be defined as private fields.

kevin-dp (approved):
Amazing job!
Left some comments, nothing major.

examples/yjs-provider/app/y-electric.ts
```diff
@@ -0,0 +1,336 @@
...
+        throw e
+      }
+
+      const handleSyncMessage = (messages: Message<OperationMessage>[]) => {
+        messages.forEach((message) => {
+          if (isChangeMessage(message) && message.value.op) {
+            const decoder = message.value.op
+            const encoder = encoding.createEncoder()
+            encoding.writeVarUint(encoder, messageSync)
+            const syncMessageType = syncProtocol.readSyncMessage(
```
KyleAMathews, line 235:
```suggestion
            syncProtocol.readSyncMessage(
```

examples/yjs-provider/app/electric-editor.tsx
```diff
@@ -0,0 +1,104 @@
...
+      extensions: [
+        keymap.of([...yUndoManagerKeymap]),
+        basicSetup,
+        javascript(),
+        EditorView.lineWrapping,
+        yCollab(ytext, awareness),
+      ],
+    })
+
+    const view = new EditorView({ state, parent: editor.current ?? undefined })
```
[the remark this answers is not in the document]
balegas: yes. it puts the editor under some element in the DOM

examples/yjs-provider/app/electric-editor.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  useEffect(() => {
+    if (typeof window === `undefined`) {
+      return
+    }
+
+    if (typeof window !== `undefined` && !network) {
+      awareness = new Awareness(ydoc)
+      network = new ElectricProvider(
```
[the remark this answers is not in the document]
balegas: Can't remember the story behind this, but made the suggested changes and it works

examples/yjs-provider/app/y-electric.ts
```diff
@@ -0,0 +1,336 @@
...
+  sync: (state: boolean) => void
+  synced: (state: boolean) => void
+  status: (status: {
+    status: `connecting` | `connected` | `disconnected`
+  }) => void
+  // eslint-disable-next-line quotes
+  "connection-close": () => void
+}
+
+const messageSync = 0
```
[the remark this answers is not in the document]
balegas: Might leave this one here because that is the message code for the protocol

examples/yjs-provider/app/y-electric.ts
balegas: A lot has changed :)

examples/yjs-provider/app/api/operation/route.ts
```diff
@@ -0,0 +1,62 @@
...
+      await saveAwarenessOperation(room, op, clientId, pool)
+    }
+
+    return NextResponse.json({})
+  } catch (e) {
+    const resp = e instanceof Error ? e.message : e
+    return NextResponse.json(resp, { status: 400 })
+  } finally {
+    if (connected) {
+      pool.end()
```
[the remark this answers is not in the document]
balegas: The code was a mix of things I tried to make work in the past. I've changed this code a bit now. Thanks for highlighting this.

Issue #1507, opened 2024-08-09
Cannot convert null to bigint (and presumably other types)


KyleAMathews:
The parser tries to convert a null to bigint. We probably need to load into the schema if a column can be null or not (it'd be useful for creating types e.g. Typescript types.

```
TypeError: Cannot convert null to a BigInt
    at BigInt (<anonymous>)
    at parseBigInt (parser.ts:11:40)
    at MessageParser.parseRow (parser.ts:128:12)
    at parser.ts:95:27
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (parser.ts:94:26)
    at JSON.parse (<anonymous>)
    at MessageParser.parse (parser.ts:87:17)
    at ShapeStream.<anonymous> (client.ts:267:40)
    at Generator.next (<anonymous>)
```

KyleAMathews:
Fixed by #1509

PR #1509, opened 2024-08-12
feat: include nullability in schema and parse null values on client

kevin-dp:
This PR addresses https://github.com/electric-sql/electric/issues/1507.
The sync service is modified to include a `not_null: true` field on columns that have a `NOT NULL` constraint. For nullable columns we do not include this information as it is assumed to be nullable by default (consistent with Postgres).

The TypeScript client parses Database NULL values into JavaScript `null` only if the column is nullable.

### DIFF
--- a/.changeset/loud-roses-fetch.md
+++ b/.changeset/loud-roses-fetch.md
@@ -1,0 +1,6 @@
+---
+"@electric-sql/client": patch
+"@core/sync-service": patch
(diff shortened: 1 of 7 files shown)


KyleAMathews (approved):
:shipit: nice!

icehaunter (approved):
LGTM

Issue #1506, opened 2024-08-08
Store information about Postgres' timeline and replication slot's WAL offset outside of Postgres


alco:
Postgres' [point-in-time-recovery](https://www.postgresql.org/docs/current/continuous-archiving.html) capability allows it to be restored from backup to a state that precedes the state it had prior to the recovery. This creates a potentially catastrophic problem for Electric because if it simply continued streaming updates from whatever state Postgres had gotten restored to, all of its clients would end up in diverging from the server's state.

The replication slot's offset may be manually advanced forward, or the slot may be removed while Electric is not connected to Postgres.

Another situation where it will be useful for Electric to verify that the database it is connecting to is the same one it connected to the last time is when Electric is restarted using a different `DATABASE_URL`.

Electric should store some minimal amount of information in its on-disk shape cache to be able to verify Postgres' timeline and replication slot offset every time it opens a new replication connection.

kevin-dp:
Electric could store the current timeline ID which we can get with:
```sql
SELECT substring(pg_walfile_name(pg_current_wal_lsn()),1,8);
```

Then on every Postgres restart, we need to fetch the timeline ID and compare it with the one we have stored. If they are different, we're in a different timeline meaning a PITR happened. Now, a blunt solution ("worse is better") would be to rotate all shapes, a more sophisticated but also more complicated solution would be to restore all shapes to the LSN on which this timeline is based. The advanced solution requires finding the last LSN of our timeline on which this new timeline is based. This can be found in the history file of the new timeline (it should contain an entry for Electric's timeline and the LSN on which the new timeline is based, see part "3. The History File" in this [blog post](https://www.highgo.ca/2021/11/01/the-postgresql-timeline-concept/)).

PR #1511, opened 2024-08-12
tests: add a test of idempotency of same txn storage

### LINKED ISSUE #1505: Write a test that verifies the idempotency of processing the same transaction more than once
The situation when Electric receives the same transaction more than once via the replication stream can arise in the following scenario. When the database cluster shuts down, all of Electric's connections to the database will close and Electric will go into the reconnection-with-backoff loop. Once Postgres comes back up, Electric will be able to resume its logical replication stream from the same point at which it stopped streaming previously thanks to the persistent replication slot it had created.

If Electric has processed a transaction (by persisting it to the shape log) but hasn't acknowledged its WAL offset to Postgres because the connection was closed, the next time Postgres starts, Electric will read the already seen transaction from the logical stream once again.

icehaunter:
Closes #1505

### DIFF
--- a/packages/sync-service/test/electric/replication/shape_log_collector_test.exs
+++ b/packages/sync-service/test/electric/replication/shape_log_collector_test.exs
@@ -361,7 +361,6 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
 
       assert [^op1, ^op2] =
                Storage.get_log_stream(shape_id, LogOffset.before_all(), storage)
-               |> Enum.to_list()
                |> Enum.map(&:json.decode/1)
     end
   end
--- a/packages/sync-service/test/electric/replication/shape_log_collector_test.exs
+++ b/packages/sync-service/test/electric/replication/shape_log_collector_test.exs
@@ -2,6 +2,8 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
   use ExUnit.Case, async: true
   import Mox
 
+  alias Electric.ShapeCache
+  alias Electric.ShapeCache.Storage
   alias Support.StubInspector
   alias Electric.Replication.Eval.Parser
   alias Electric.Shapes.Shape
@@ -20,23 +22,6 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
 
   setup :verify_on_exit!
 
-  setup do
-    # Start a test Registry
-    registry_name = Module.concat(__MODULE__, Registry)
-    start_link_supervised!({Registry, keys: :duplicate, name: registry_name})
-
-    # Start the ShapeLogCollector process
-    opts = [
-      name: :test_shape_log_storage,
-      registry: registry_name,
-      shape_cache: {MockShapeCache, []},
-      inspector: {MockInspector, []}
-    ]
-
-    {:ok, pid} = start_supervised({ShapeLogCollector, opts})
-    %{server: pid, registry: registry_name}
-  end
-
   @shape Shape.new!("public.test_table",
            inspector: StubInspector.new([%{name: "id", type: "int8", pk_position: 0}])
          )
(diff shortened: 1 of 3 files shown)


balegas:
@icehaunter see https://github.com/electric-sql/electric/issues/1505#issuecomment-2283495707

icehaunter:
@balegas This adds the aforementioned test. As to whether we should avoid writing at all in case of duplication, I think we can loop back to that question later because under current implementations it's not necessary. If our storage engine becomes actually append-log, then it might become an important idea

balegas:
Do you mind leaving a comment in the code noting that the way we write to storage actually implies idempotency and why it is required?

I suppose this is remains true for an in-memory approach

packages/sync-service/test/electric/replication/shape_log_collector_test.exs
```diff
@@ -288,4 +290,79 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
...
+      assert [op1, op2] =
+               Storage.get_log_stream(shape_id, LogOffset.before_all(), storage)
+               |> Enum.map(&:json.decode/1)
+
+      # If we encounter & store the same transaction, log stream should be stable
+      assert :ok = ShapeLogCollector.store_transaction(txn, server)
+
+      assert [^op1, ^op2] =
+               Storage.get_log_stream(shape_id, LogOffset.before_all(), storage)
+               |> Enum.to_list()
```
alco, line 364:
```suggestion
```

packages/sync-service/test/electric/shape_cache/storage_implementations_test.exs
```diff
@@ -199,6 +199,35 @@ defmodule Electric.ShapeCache.StorageImplimentationsTest do
...
+          storage.get_log_stream(@shape_id, LogOffset.first(), LogOffset.last(), opts)
+          |> Enum.map(&:json.decode/1)
+
+        :ok = storage.append_to_log!(@shape_id, log_items, opts)
+
+        log2 =
+          storage.get_log_stream(@shape_id, LogOffset.first(), LogOffset.last(), opts)
+          |> Enum.map(&:json.decode/1)
+
+        assert log1 == log2
```
alco, line 229: Nit: No need to decode from JSON for this assertion.

alco (approved):
The code looks good but I have some thoughts on whether it addresses the issue.

Right now the replication stream is only responsible for saving incoming transactions to log storage. But we've already had discussions about streaming transactions in parallel with persisting them to log storage in some cases. As the system evolves, at some point we may replace the current `transaction_received` callback passed to `ReplicationClient` with another function potentially from another module, side-stepping the test that only verifies the behaviour of ShapeLogCollector.

I think we can potentially address the duplicate transaction processing problem at the source, i.e. by having information inside ReplicationClient about which lsn it processed last. It's just that implementing this approach requires some design work first.

packages/sync-service/test/electric/shape_cache/storage_implementations_test.exs
icehaunter: True, but the comparison error will be better

balegas:
I see this being merged and the original issue being closed, but I feel that the PR misses the point of the issue.
Ideally the test should cover the any implementation rather than a specific one, but if it doesn't we need to document the assumption, because this implicit property is required for the correctness of the system. I think that is helpful that the test covers any changes that people might do to the current implementation, but I feel that documentation is needed for someone that is making a new storage implementation. 

I don't think we need to do the engineering ahead of switching implementations and we might actually touch that code when we do https://github.com/electric-sql/electric/issues/1506. So, I'm happy addressing the issue at the source there. However, I strongly think that we need to document in code the implicit property.

alco:
I have filed a new issue specifically for addressing the storage of replication state for the purpose of detecting repeat transactions without involving the shape storage subsystem - https://github.com/electric-sql/electric/issues/1515.

icehaunter:
@balegas I've added a storage-level test explicitly for this reason, so that it's less tied to the shape log collector implementation. If someone adds an append-only storage, I would assume they would look at the failing test and question the assumption explicitly stated there. Do you feel like that's not enough?

Issue #1444, opened 2024-07-31
Sync Service breaks on upgrade if format of the storage changes


robacourt:
For example electric-sql/electric-next#182 changed the CubDB storage format (by [reordering the terms in the tuple](https://github.com/electric-sql/electric-next/pull/182/files#diff-3fea7cf7976e09e017504f11f3688d6f5092aeabd377d257158d3787cb45c76dR205))

So if you synced a shape with a previous version then upgraded to that version, if you try to sync the shape again you get a 500 due to the service trying to read the old format but thinking it's the new format.

Perhaps the files should be versioned and wiped if the version changes.

robacourt:
Fixed in #1517, version changes will now wipe the shapes. Perhaps in the future we could upgrade the data as @alco suggested as an optimisation, but we are already safe.

Issue #1438, opened 2024-08-02
Timeout waiting for snapshot


balegas:
Had to throw enough (large) initial query requests to trigger it.
In my case 10 init requests for issue table with 100000 issues.

```
14:15:31.813 [error] GenServer #PID<0.593.0> terminating
** (stop) exited in: GenServer.call(Electric.ShapeCache, {:wait_for_snapshot, "50159385-1722604526811"}, 5000)
    ** (EXIT) time out
    (elixir 1.17.2) lib/gen_server.ex:1128: GenServer.call/3
    (electric 0.2.4) lib/electric/shapes.ex:16: Electric.Shapes.get_snapshot/2
    (electric 0.2.4) lib/electric/plug/serve_shape_plug.ex:275: Electric.Plug.ServeShapePlug.serve_log_or_snapshot/2
    (electric 0.2.4) lib/electric/plug/serve_shape_plug.ex:1: Electric.Plug.ServeShapePlug.plug_builder_call/2
    (electric 0.2.4) deps/plug/lib/plug/router.ex:246: anonymous fn/4 in Electric.Plug.Router.dispatch/2
    (telemetry 1.2.1) /Users/vbalegas/workspace/electric-next/packages/sync-service/deps/telemetry/src/telemetry.erl:321: :telemetry.span/3
    (electric 0.2.4) deps/plug/lib/plug/router.ex:242: Electric.Plug.Router.dispatch/2
    (electric 0.2.4) lib/electric/plug/router.ex:1: Electric.Plug.Router.plug_builder_call/2
    (bandit 1.5.5) lib/bandit/pipeline.ex:124: Bandit.Pipeline.call_plug!/2
    (bandit 1.5.5) lib/bandit/pipeline.ex:36: Bandit.Pipeline.run/4
    (bandit 1.5.5) lib/bandit/http1/handler.ex:12: Bandit.HTTP1.Handler.handle_data/3
    (bandit 1.5.5) lib/bandit/delegating_handler.ex:18: Bandit.DelegatingHandler.handle_data/3
    (bandit 1.5.5) /Users/vbalegas/workspace/electric-next/packages/sync-service/deps/thousand_island/lib/thousand_island/handler.ex:411: Bandit.DelegatingHandler.handle_continue/2
    (stdlib 6.0.1) gen_server.erl:2163: :gen_server.try_handle_continue/3
    (stdlib 6.0.1) gen_server.erl:2072: :gen_server.loop/7
[...]

Issue #1499, opened 2024-08-08
Tune PG connection pool to serve initial syncs


balegas:
![image](https://github.com/user-attachments/assets/e9e7f067-4cf5-4083-bbcd-6bb26e3df373)

This picture shows the max number of concurrent initial syncs per shape size. The takeaway is that when shape is large enough, we can't handle the next request because there are no more connections available in the Pg connection pool, because they are all busy.

We're going to tune the initial pool size to be better fit with the number of requests the server can handle.

balegas:
@icehaunter can you leave here the same picture as above with the new default pool size configuration?
Let's measure the improvement to decide on further action

icehaunter:
![image](https://github.com/user-attachments/assets/3862452a-3e44-451e-9995-a4960626a0ff)
This is (incomplete) run after #1503

balegas:
It still falls over relatively soon, is it now falling over because of the snapshot timeout?

alco:
@icehaunter That's not pool-related, it's a query running for longer than the default query timeout.

alco:
@magnetised

`ownership_timeout` is specific to the [`DBConnection.Ownership`](https://hexdocs.pm/db_connection/DBConnection.Ownership.html) pool which one can use in place of the default [`DBConnection.ConnectionPool`](https://hexdocs.pm/db_connection/DBConnection.ConnectionPool.html).

`timeout` is an option you pass to [`Postgrex.query()`](https://hexdocs.pm/postgrex/Postgrex.html#query/4-options). It is set to `15000` by default. The reason the error message in the log originates in `DBConnection` is because `Postgrex` forwards the `timeout` option to `DBConnection` which then starts a timer that determines how long a given query can be holding a connection for.

PR #1517, opened 2024-08-13
feat(sync-service): Streaming the response to the client without waiting for the snapshot to finish

robacourt:
Previously shapes with one million rows or more (170MB+) would timeout waiting for the snapshot to be created. Now the snapshot is streamed from the database into storage while simultaneously being streamed from storage to the client. This means the first packets of the response can be sent without waiting for the snapshot to finish. This means we now support tables with over a million rows. Ilia is currently benchmarking this branch to see what the new limits are. 

This PR addresses #1438 and #1444

I think there are quite a few simplifications that could happen off the back of this change, but I have kept the refactoring to a minimum in this PR and will instead address the simplifications in separate PRs.

### DIFF
--- a/packages/sync-service/lib/electric/concurrent_stream.ex
+++ b/packages/sync-service/lib/electric/concurrent_stream.ex
@@ -1,6 +1,18 @@
 defmodule Electric.ConcurrentStream do
   @default_poll_time 10
 
+  @doc """
+  Allows concurrent reading while writing of a stream.
+  There can be mutiple reading processes however there must be only one writing process.
+
+  The writing process must append an end marker to the end of the stream when it has finished
+  to signal to the reading processes that the stream has ended.
+
+  If a read process runs out of data to read before the end marker has been written
+  it waits the `poll_time_in_ms` for more data to be written, then resumes the stream
+  with the `stream_fun`.
+  """
+
   def stream_to_end(opts) do
     excluded_start_key = Keyword.fetch!(opts, :excluded_start_key)
     end_marker_key = Keyword.fetch!(opts, :end_marker_key)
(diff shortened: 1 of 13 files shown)


packages/sync-service/lib/electric/shape_cache/cub_db_storage.ex
```diff
@@ -1,9 +1,14 @@
 defmodule Electric.ShapeCache.CubDbStorage do
+  alias Electric.ConcurrentStream
   alias Electric.LogItems
   alias Electric.Replication.LogOffset
   alias Electric.Telemetry.OpenTelemetry
   @behaviour Electric.ShapeCache.Storage
 
+  # If the storage format changes, increase `@version` to prevent
+  # the incompatable older versions being read
+  @version 1
```
robacourt, line 10: The storage format changes as part of this PR (end markers are stored) and so I had to resolve #1444 . This is the most basic implementation, older versions are not read and the clients will need to re-sync. A better mechanism that upgrades the stored data can be introduced later.

packages/sync-service/lib/electric/concurrent_stream.ex
```diff
@@ -0,0 +1,51 @@
...
+  Allows concurrent reading while writing of a stream.
+  There can be mutiple reading processes however there must be only one writing process.
+
+  The writing process must append an end marker to the end of the stream when it has finished
+  to signal to the reading processes that the stream has ended.
+
+  If a read process runs out of data to read before the end marker has been written
+  it waits the `poll_time_in_ms` for more data to be written, then resumes the stream
+  with the `stream_fun`.
+  """
```
robacourt, line 14: This how the concurrent streaming works

icehaunter:
time-to-headers really shows the power of this PR:
Old (up to 800k rows):
image

New (up to 3M rows):
image

Mind you that for 3M test I'm not reading the body, so we're not hitting the timeouts mentioned in #1499  which we are definitely hitting when reading the body out

icehaunter (approved):
LGTM! It would be good to maybe move from polling to message fanout, but we've discussed that on a call

Issue #1452, opened 2024-07-23
Disparity in OpenAPI spec and TS types for "offset"


samwillis:
Open api spec shows the offset in the format "number/number", the types and the repose from the sync service have "number_number".

PR #1528, opened 2024-08-15
docs: fix offset formatting in OpenAPI spec

kevin-dp:
Some examples in the OpenAPI spec were still using the old `number/number` format for log offsets instead of `number_number`. This fixes those occurences as per https://github.com/electric-sql/electric/issues/1452.


KyleAMathews (approved):
:shipit:

PR #1532, opened 2024-08-16
feat: move JSON serialization at snapshot creation to PostgreSQL query

icehaunter:
This moves us building up JSON from Elixir to Postgres Query, which should be faster.

### DIFF
--- a/packages/sync-service/lib/electric/log_items.ex
+++ b/packages/sync-service/lib/electric/log_items.ex
@@ -1,13 +1,11 @@
 defmodule Electric.LogItems do
   alias Electric.Replication.Changes
   alias Electric.Replication.LogOffset
-  alias Electric.Shapes.Shape
-  alias Electric.Utils
 
   @moduledoc """
   Defines the structure and how to create the items in the log that the electric client reads.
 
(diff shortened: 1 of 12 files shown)


PerplexityCoding:
Hello, thanks for this feature. This is an interesting project.

It also fixed the issue where every NULL column were returning an empty string : arrays and jsonb. As mention on https://github.com/electric-sql/electric/issues/1536

Before, using my database it was making jsonb  / arrays throws errors during parsing.

Now just pgArrayParser is throwing an error because the value of array in null.
I don't think parsing null row is necessary (?).

With that i can now received my data from ShapeStream.

Here is the patch
```diff
--- a/packages/typescript-client/src/parser.ts	(revision aaee4e7ab54b90b31b7753300629effbc2cc88c8)
+++ b/packages/typescript-client/src/parser.ts	(date 1724249417943)
@@ -92,7 +92,9 @@
         // Parse the row values
         const row = value as Record<string, Value>
         Object.keys(row).forEach((key) => {
-          row[key] = this.parseRow(key, row[key] as string, schema)
+          if (row[key] != null) {
+            row[key] = this.parseRow(key, row[key] as string, schema)
+          }
         })
       }
       return value
```

PR #1533, opened 2024-08-16
feat: introduce CubDB+ storage

icehaunter:
This introduces a new storage backend, that's build as a plain file for snapshots, and CubDB for any metadata and ongoing logs. The benefit here is that we don't have to close the file while processing the results coming from the PostgreSQL query, unlike with batched CubDB writes. The result is that we're writing the snapshot roughly at the same rate as PostgreSQL generates query results.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/mixed_disk_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/mixed_disk_storage.ex
@@ -1,0 +1,217 @@
+defmodule Electric.ShapeCache.MixedDiskStorage do
+  alias Electric.Telemetry.OpenTelemetry
+  alias Electric.Replication.LogOffset
+  @behaviour Electric.ShapeCache.Storage
+
+  @version 1
+  @version_key :version
+
+  def shared_opts(opts) do
+    storage_dir = Access.get(opts, :storage_dir, "./shapes")
+    db = Access.get(opts, :db, :mixed_shape_db)
+
+    {:ok, %{storage_dir: storage_dir, db: db, version: @version}}
+  end
+
+  def child_spec(opts) do
+    %{
+      id: __MODULE__,
+      start: {__MODULE__, :start_link, [opts]},
+      type: :worker,
+      restart: :permanent
+    }
+  end
+
+  def start_link(opts) do
+    File.mkdir_p(opts.storage_dir)
+    File.mkdir_p(Path.join([opts.storage_dir, "cubdb"]))
+    File.mkdir_p(Path.join([opts.storage_dir, "shapes"]))
+    CubDB.start_link(data_dir: Path.join([opts.storage_dir, "cubdb"]), name: opts.db)
+  end
+
+  def add_shape(shape_id, shape, opts) do
+    CubDB.put(opts.db, shape_key(shape_id), shape)
+  end
+
+  def set_snapshot_xmin(shape_id, xmin, opts) do
+    CubDB.put(opts.db, xmin_key(shape_id), xmin)
+  end
+
+  defp snapshot_xmin(shape_id, opts) do
+    CubDB.get(opts.db, xmin_key(shape_id))
+  end
+
+  def list_shapes(opts) do
+    opts.db
+    |> CubDB.select(min_key: shapes_start(), max_key: shapes_end())
+    |> Enum.map(fn {{:shapes, shape_id}, shape} ->
+      %{
+        shape_id: shape_id,
+        shape: shape,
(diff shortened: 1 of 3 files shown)


packages/sync-service/lib/electric/shape_cache/mixed_disk_storage.ex
```diff
@@ -0,0 +1,217 @@
...
+       )}
+    else
+      raise "Snapshot no longer available"
+    end
+  end
+
+  defp open_snapshot_file(shape_id, opts) do
+    case File.open(shape_snapshot_path(shape_id, opts), [:read, :raw, read_ahead: 1024]) do
+      {:ok, file} -> file
+      {:error, :enoent} -> open_snapshot_file(shape_id, opts)
```
msfstef, line 149: any chance this could get into an infinite loop (`open_snapshot_file` called on file that actually does not exist and is not being created)? Would it benefit from a timeout or retry limit?

packages/sync-service/lib/electric/shape_cache/mixed_disk_storage.ex
```diff
@@ -0,0 +1,217 @@
...
+    [
+      snapshot_meta_key(shape_id),
+      shape_key(shape_id),
+      xmin_key(shape_id),
+      snapshot_started_key(shape_id)
+    ]
+    |> Enum.concat(keys_from_range(log_start(shape_id), log_end(shape_id), opts))
+    |> then(&CubDB.delete_multi(opts.db, &1))
+
+    File.rm_rf(shape_snapshot_path(shape_id, opts))
```
msfstef, line 191: if for some reason the metadata is deleted but the file was not, would we have a cleanup process? or we could delete the file first, then the metadata (so we can use it as an index for cleanup)?

packages/sync-service/lib/electric/shape_cache/mixed_disk_storage.ex
```diff
@@ -0,0 +1,217 @@
...
+      {LogOffset.first(),
+       Stream.resource(
+         fn -> open_snapshot_file(shape_id, opts) end,
+         fn file ->
+           case IO.binread(file, :line) do
+             {:error, reason} ->
+               raise IO.StreamError, reason: reason
+
+             :eof ->
+               Process.sleep(10)
```
msfstef, line 128: How come we `Process.sleep`? I assume with the assumption that the 4 bytes or more data should get written after the wait and actually terminate, but is there any case where this just keeps on going and going and the 4 bytes never get written?

Maybe this keeps track of the last line read and "times out" if it doesn't read any new lines after some time?

msfstef (approved):
Looks good, nice work!

Catching up on sync service work sorry if I'm addressing things that have already been discussed!

packages/sync-service/lib/electric/shape_cache/mixed_disk_storage.ex
icehaunter: This is actually a `tail -f` comparable behavior, where we're reading up until EOF, then sleeping for 10 ms waiting for more data to be written to the end of file. But if we're always reading until EOF, we need a stop condition different from "just reading until EOF" which I picked to be ASCII end-of-transmission (`0x04`) byte. So this is an infinite loop (until this function returns the `{:halt, file}` tuple reading from file until end of transmission byte.

packages/sync-service/lib/electric/shape_cache/mixed_disk_storage.ex
icehaunter: Yeah, not a bad idea

packages/sync-service/lib/electric/shape_cache/mixed_disk_storage.ex
msfstef: right - but `tail -f` waits for an interrupt whereas this would wait for the 4 end-of-transmission bytes - should we not guard against the case where those bytes are never written (for whatever reason)?

packages/sync-service/lib/electric/shape_cache/mixed_disk_storage.ex
msfstef: could create a tmp placeholder like `{snapshot_file}.deleted` first, then delete metadata, then delete both the file and tmp placeholder - and cleanup would just involve looking at placeholders and deleting both them and corresponding snapshots if they exist?

KyleAMathews:
This is obsolete now right?

icehaunter:
Superseded by #1604

PR #1539, opened 2024-08-18
Update README.md to use '@electric-sql/client'

stolinski:
Was giving this a whirl today and noticed this small issue.


KyleAMathews (approved):
Oops, thanks!

Issue #1516, opened 2024-08-12
Investigate the possibility of reconciling shape logs with Postgres' state after a point-in-time recovery


alco:
This involves at least two things:
1. Electric should be able to start streaming using an older timeline, then switch to the current one
2. Electric should identify transactions that are in the future for the current timeline, discard them from storage and tell clients to do the same 

From https://www.postgresql.org/docs/current/protocol-replication.html#PROTOCOL-REPLICATION-START-REPLICATION:
> If the client requests a timeline that's not the latest but is part of the history of the server, the server will stream all the WAL on that timeline starting from the requested start point up to the point where the server switched to another timeline. If the client requests streaming at exactly the end of an old timeline, the server skips COPY mode entirely.

balegas:
It sounds hard to get right, but actually the shape log can turn time travelling into INSERT, DELETE statements.

PR #1541, opened 2024-08-19
feat: clean shapes on PITR

kevin-dp:
Postgres supports Point In Time Recovery (PITR) whereby the DB is recovered to a certain point in time. If someone performs a PITR, the current state of Electric does not match the state on PG.

Postgres has the concept of timelines. When performing a PITR, the timeline ID changes. Thus, in this PR we modify Electric to keep track of the Postgres timeline ID. Every time the connection to PG is restarted, we check the Postgres timeline ID. If it is different from the one Electric knows, then we clean all shapes and update the timeline ID in our cache.

Questions:
- `clean_up_all_shapes` currently fetches all active shapes then calls `clean_up_shape` in a loop for each active shape. Is there a better way to do this with ets? i.e. looking for an erlang pattern that deletes a shape if it's ID is in a list of IDs. Didn't find in the docs if this is possible.


balegas:
Closes #1516

kevin-dp:
Fixes https://github.com/electric-sql/electric/issues/1506.

@balegas this PR does not seem to address #1516 as that one is about handling timelines in a smarter way, whereas this PR implements a blunt approach that cleans all shapes when a PITR is detected.

balegas:
We only want to do the blunt approach for the time being. Would the PR be up to review in that case?

kevin-dp:
> We only want to do the blunt approach for the time being. Would the PR be up to review in that case?

Unfortunately, not yet. I'm waiting for Garry's work on the revised storage to land before adding persistent storage of timeline information to this PR.

packages/sync-service/test/test_helper.exs
```diff
@@ -1 +1,3 @@
+Mox.defmock(Electric.ShapeCacheMock, for: Electric.ShapeCacheBehaviour)
```
magnetised, line 1: I added a `Support.Mocks` module to avoid duplicate mock defs in a previous PR - they were spread all over the place and were causing sporadic errors on compilation. Could you put this there pls

packages/sync-service/lib/electric/timeline.ex
```diff
@@ -0,0 +1,54 @@
...
+    cache = Keyword.fetch!(opts, :timeline_cache)
+    electric_timeline = TimelineCache.get_timeline(cache)
+    handle(pg_timeline, electric_timeline, opts)
+  end
+
+  # Handles the different cases of timeline comparison
+  @spec handle(timeline(), timeline(), keyword()) :: :ok
+  defp handle(nil, _, opts) do
+    Logger.warning("Unknown Postgres timeline; rotating shapes.")
+    Shapes.clean_all_shapes(opts)
```
magnetised, line 28: this is going to conflict so badly with my open pr... 😭

magnetised (changes requested):
I think this would be simpler as a single module if possible, but critically pls see my comments re writing to the kv.

packages/sync-service/lib/electric/timeline_cache.ex
```diff
@@ -0,0 +1,84 @@
...
+        nil
+
+      error ->
+        Logger.warning("Failed to load timeline ID from persistent storage: #{error}")
+        nil
+    end
+  end
+
+  defp load_timeline_id(timeline_id, kv) do
+    PersistentKV.set(kv, @timeline_key, timeline_id)
```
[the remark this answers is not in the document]
kevin-dp: Right, that's what i wanted to ask you this morning but then it seemed to work by just writing an integer. The KV store should definitely have the same behavior across all implementations to avoid such corner cases :)

Re "load", yes naming is not ideal. It basically stores if provided and otherwise loads from storage.

packages/sync-service/lib/electric/timeline.ex
```diff
@@ -0,0 +1,81 @@
...
+        Logger.warning("Failed to load timeline ID from persistent storage: #{error}")
+        nil
+    end
+  end
+
+  defp store_timeline(timeline_id, opts) do
+    kv_backend = Keyword.fetch!(opts, :persistent_kv)
+    # defaults to Jason encoder
+    kv = PersistentKV.Serialized.new!(backend: kv_backend)
+    PersistentKV.set(kv, @timeline_key, timeline_id)
```
magnetised, line 78:
```suggestion
    :ok = PersistentKV.set(kv, @timeline_key, timeline_id)
```

packages/sync-service/lib/electric/timeline.ex
```diff
@@ -0,0 +1,81 @@
...
+        nil
+
+      error ->
+        Logger.warning("Failed to load timeline ID from persistent storage: #{error}")
+        nil
+    end
+  end
+
+  defp store_timeline(timeline_id, opts) do
+    kv_backend = Keyword.fetch!(opts, :persistent_kv)
```
magnetised, line 75: could we dry this up a touch. the duplication of creating the kv instance here and in `load_timeline/1` makes it easy for the two versions to drift out of sync.

packages/sync-service/lib/electric/timeline.ex
```diff
@@ -0,0 +1,81 @@
...
+  def check(pg_timeline, opts) do
+    electric_timeline = load_timeline(opts)
+    handle(pg_timeline, electric_timeline, opts)
+  end
+
+  # Handles the different cases of timeline comparison
+  @spec handle(timeline(), timeline(), keyword()) :: :ok
+  defp handle(nil, _, opts) do
+    Logger.warning("Unknown Postgres timeline; rotating shapes.")
+    Shapes.clean_all_shapes(opts)
```
magnetised, line 30: if we're passing the shape cache in the opts here, why the indirection via `Shapes` - we could just directly use the shape cache value here and call `shape_cache.clear_all_shapes`.

packages/sync-service/lib/electric/timeline.ex
```diff
@@ -0,0 +1,81 @@
...
+  @doc """
+  Checks the provided `pg_timeline` against Electric's timeline.
+  Normally, Postgres and Electric are on the same timeline and nothing must be done.
+  If the timelines differ, that indicates that a Point In Time Recovery (PITR) has occurred and all shapes must be cleaned.
+  If we fail to fetch timeline information, we also clean all shapes for safety as we can't be sure that Postgres and Electric are on the same timeline.
+  """
+  @spec check(timeline(), keyword()) :: :ok
+  def check(pg_timeline, opts) do
+    electric_timeline = load_timeline(opts)
+    handle(pg_timeline, electric_timeline, opts)
```
magnetised, line 23: `handle` would be better called `verify` or `validate` to be clear about what it's doing.

packages/sync-service/test/electric/timeline_test.exs
```diff
@@ -0,0 +1,63 @@
...
+
+    test "returns nil when no timeline ID is available", %{kv: kv} do
+      assert Timeline.load_timeline(persistent_kv: kv) == nil
+    end
+  end
+
+  describe "check/2" do
+    setup context do
+      timeline = context[:electric_timeline]
+      kv = Electric.PersistentKV.Memory.new!()
```
magnetised, line 21: could we add a very simple read/write check with a real kv backend. I'd feel better knowing that we're going to read what we write...

packages/sync-service/lib/electric/timeline.ex
```diff
@@ -0,0 +1,81 @@
...
+
+  # Handles the different cases of timeline comparison
+  @spec handle(timeline(), timeline(), keyword()) :: :ok
+  defp handle(nil, _, opts) do
+    Logger.warning("Unknown Postgres timeline; rotating shapes.")
+    Shapes.clean_all_shapes(opts)
+    store_timeline(nil, opts)
+  end
+
+  defp handle(pg_timeline_id, electric_timeline_id, _opts)
```
magnetised, line 34: bit of elixir technique, if you want to match on two values being identical, you just have to use the same name twice, that way the match only happens if the two values are the same, so in this case you could replace this function + when clause with 

`defp handle(timeline_id, timeline_id, _opts) do`

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -350,14 +364,21 @@ defmodule Electric.ShapeCache do
...
-    if state.shape_status.existing_shape(state.persistent_state, shape_id) do
-      true
-    else
-      false
+  defp clean_up_all_shapes(state) do
+    shape_ids =
+      state.persistent_state |> state.shape_status.list_shapes() |> Enum.map(&elem(&1, 0))
+
+    # FIXME: is there a better way to do this?
+    #        ideally we'd want an `:ets.select_delete` where the pattern matches any `shape_id` that is in the `shape_ids` list
```
magnetised, line 372: do we need this comment? we need to delete the shape information from the kv store, so we either need to iterate them here or have a dedicated function in ShapeStatus that deletes everything in a single ets call plus a single kv write.

But I think the iteration is fine considering the (rare) circumstances, the ShapeStatus behaviour is already complex enough

packages/sync-service/lib/electric/shapes.ex
```diff
@@ -73,7 +73,16 @@ defmodule Electric.Shapes do
...
+  end
 
+  @doc """
+  Clean up all data (meta data and shape log + snapshot) associated with all shapes
+  """
+  @spec clean_all_shapes(keyword()) :: :ok
+  def clean_all_shapes(opts \\ []) do
+    {shape_cache, opts} = Access.get(opts, :shape_cache, {ShapeCache, []})
+    server = Access.get(opts, :server, shape_cache)
+    shape_cache.clean_all_shapes(server)
```
magnetised, line 85: i don't think we need this function, but looking at the implementation we shouldn't be requiring that the `opts` has a `server` key, that's assuming a specific implementation of the shape cache -- the call should be `shape_cache.clean_all_shapes(opts)`

magnetised (changes requested):
Am I being annoying? Sorry. I think this feature could take up less space -- it's pretty self-contained.

packages/sync-service/lib/electric/timeline.ex
kevin-dp: I renamed it `verify_timeline`.

packages/sync-service/lib/electric/shapes.ex
kevin-dp: Fixed the implementation.
Couldn't we say the same thing about `clean_shape`?
All it does it call `shape_cache.clean_shape`.
I added this one to be consistent with the other ones + we can call this function anywhere we need it instead of having to every time fetch the shape cache out of the ops and then call it (and also more maintainable if later on we change the opt name for instance).

packages/sync-service/test/electric/timeline_test.exs
kevin-dp: My reasoning was that any storage implementation implements the `PersistentKV` abstraction and so they should all be equivalent.
I will change this to use the fylesystem storage backend but ultimately, don't we want to change the in-memory storage backend to behave the same as the filesystem one (i.e. only read and write strings) ?

kevin-dp:
@magnetised addressed most of your feedback. Few things that i didn't address yet, left some comments there asking for clarifications :-)

packages/sync-service/lib/electric/shapes.ex
kevin-dp: Moved it to the Timeline module as per your request 👍

magnetised (approved):
Looks great. I've added some very nit-picky comments. How hard would it be to do an end-to-end test of this?

kevin-dp:
@magnetised Doing an e2e test of this would definitely be a separate PR. I've done it once locally on my machine but it is very finicky as you need to do a PG dump, and need to configure PG to make backups to a certain folder and then restart in a mode that restores a certain backup, etc.

packages/sync-service/lib/electric/shapes.ex
```diff
@@ -88,8 +88,6 @@ defmodule Electric.Shapes do
     for shape_id <- shape_ids do
       shape_cache.clean_shape(shape_id, opts)
     end
-
```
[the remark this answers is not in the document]
kevin-dp: good catch!

Issue #1535, opened 2024-08-16
Restart the Shapes affected by a schema change


KyleAMathews:
I.e. a new column is added to a table

samwillis:
Two thoughts:

1. It would be nice if additions such as nullable columns, with a default of null, didn't require a full shape restart.
2. With ability to select subset of columns #1530 it would be nice if the shape wasn't reset if the change didn't effect the selected columns.

Obviously the short path is restart on all/any DDL changes, but with a persisted client side store it would be great if we could avoid that.

PR #1544, opened 2024-08-20
feat: clean shapes affected by relation changes

kevin-dp:
This PR addresses https://github.com/electric-sql/electric/issues/1535 by detecting relation changes and cleaning up the shapes affected by the relation change. Note that Postgres' logical replication stream propagates relation changes lazily. So we won't detect the migration until somebody actually inserts/updates/deletes a row from a migrated table.

This PR is affected by a bug that causes schema information not to be deleted when a shape is cleaned. Because of that, when testing this PR you will notice that if we add a column the shape is cleaned but then if we fetch the shape again the new column is not present in the resulting rows. This should be fixed when the schema information bug is fixed.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/replication_client/collector.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client/collector.ex
@@ -14,7 +14,8 @@ defmodule Electric.Postgres.ReplicationClient.Collector do
     NewRecord,
     UpdatedRecord,
     DeletedRecord,
-    TruncatedRelation
+    TruncatedRelation,
+    RelationChange
   }
 
   defstruct transaction: nil, tx_op_index: nil, relations: %{}
@@ -51,11 +52,22 @@ defmodule Electric.Postgres.ReplicationClient.Collector do
   def handle_message(%LR.Type{}, state), do: state
 
   def handle_message(%LR.Relation{id: id} = rel, %__MODULE__{} = state) do
-    if Map.get(state.relations, id, rel) != rel do
-      Logger.warning("Schema for the table #{rel.namespace}.#{rel.name} had changed")
-    end
+    new_state = Map.update!(state, :relations, &Map.put(&1, rel.id, rel))
+    old_rel = Map.get(state.relations, id, rel)
 
-    Map.update!(state, :relations, &Map.put(&1, rel.id, rel))
+    if old_rel != rel do
+      Logger.warning("Schema for the table #{rel.namespace}.#{rel.name} had changed")
+
+      {
+        %RelationChange{
+          old_relation: relation_to_tuple(old_rel),
+          new_relation: relation_to_tuple(rel)
+        },
+        new_state
+      }
+    else
+      new_state
+    end
   end
 
   def handle_message(%LR.Insert{} = msg, %__MODULE__{} = state) do
@@ -190,6 +202,14 @@ defmodule Electric.Postgres.ReplicationClient.Collector do
     |> Map.new(fn {%{name: column_name}, value} ->
       {column_name, value_fun.(column_name, value)}
     end)
+  end
+
+  defp column_to_tuple(column) do
+    {column.name, column.type_oid}
+  end
+
+  defp relation_to_tuple(%LR.Relation{namespace: schema, name: table, columns: cols}) do
+    {schema, table, Enum.map(cols, &column_to_tuple/1)}
   end
 
   defp column_value(_column_name, value), do: value
--- a/packages/sync-service/lib/electric/postgres/replication_client/collector.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client/collector.ex
@@ -15,7 +15,8 @@ defmodule Electric.Postgres.ReplicationClient.Collector do
     UpdatedRecord,
     DeletedRecord,
     TruncatedRelation,
-    RelationChange
+    Relation,
+    Column
   }
 
   defstruct transaction: nil, tx_op_index: nil, relations: %{}
@@ -51,23 +52,21 @@ defmodule Electric.Postgres.ReplicationClient.Collector do
   def handle_message(%LR.Origin{} = _msg, state), do: state
   def handle_message(%LR.Type{}, state), do: state
 
-  def handle_message(%LR.Relation{id: id} = rel, %__MODULE__{} = state) do
+  def handle_message(
+        %LR.Relation{id: id, namespace: ns, name: name, columns: cols} = rel,
+        %__MODULE__{} = state
+      ) do
     new_state = Map.update!(state, :relations, &Map.put(&1, rel.id, rel))
-    old_rel = Map.get(state.relations, id, rel)
 
-    if old_rel != rel do
-      Logger.warning("Schema for the table #{rel.namespace}.#{rel.name} had changed")
-
-      {
-        %RelationChange{
-          old_relation: relation_to_tuple(old_rel),
-          new_relation: relation_to_tuple(rel)
-        },
-        new_state
-      }
-    else
+    {
+      %Relation{
+        id: id,
+        schema: ns,
+        table: name,
+        columns: Enum.map(cols, fn col -> %Column{name: col.name, type_oid: col.type_oid} end)
+      },
       new_state
-    end
+    }
   end
 
   def handle_message(%LR.Insert{} = msg, %__MODULE__{} = state) do
@@ -202,14 +201,6 @@ defmodule Electric.Postgres.ReplicationClient.Collector do
     |> Map.new(fn {%{name: column_name}, value} ->
       {column_name, value_fun.(column_name, value)}
     end)
-  end
-
-  defp column_to_tuple(column) do
-    {column.name, column.type_oid}
-  end
-
-  defp relation_to_tuple(%LR.Relation{namespace: schema, name: table, columns: cols}) do
-    {schema, table, Enum.map(cols, &column_to_tuple/1)}
   end
 
   defp column_value(_column_name, value), do: value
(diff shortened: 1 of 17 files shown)


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -101,4 +121,30 @@ defmodule Electric.Replication.ShapeLogCollector do
           do: send(pid, {ref, :new_changes, latest_log_offset})
     end)
   end
+
+  defp is_affected_by_relation_change?(
```
KyleAMathews, line 125: I'm curious how'll this scale for tons of shapes but that's a problem we can worry about layer if/when it presents itself

KyleAMathews (approved):
Looks good to me! :shipit:

packages/sync-service/lib/electric/postgres/replication_client/collector.ex
```diff
@@ -51,11 +52,22 @@ defmodule Electric.Postgres.ReplicationClient.Collector do
...
+    old_rel = Map.get(state.relations, id, rel)
+
+    if old_rel != rel do
       Logger.warning("Schema for the table #{rel.namespace}.#{rel.name} had changed")
-    end
 
-    Map.update!(state, :relations, &Map.put(&1, rel.id, rel))
+      {
+        %RelationChange{
+          old_relation: relation_to_tuple(old_rel),
```
alco, line 63: `old_rel` will be `nil` when it's the first time we're seeing this relation in the current replication session. We cannot simply compare `old_rel != rel` and then try converting `nil` into a tuple.

alco (approved):
This is a good partial solution for the issue but it is not complete.

Every time Electric opens a new replication connection to Postgres, it will have an empty relation cache. So in a situation where Electric was running for some time, then got shut down, then a table's schema was changed in Postgres, then Electric was launched again, it would receive a relation message while having an empty relation cache, so it wouldn't detect the schema change.

The proper solution for detecting schema changes accross restarts would be to pass all relation messages to the shape log collector and have it load the previous relation schema from its shape metadata store to compare against the new relation schema received in the replication stream.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
kevin-dp: Yes, currently it's linear to the number of shapes as we're filtering them out one-by-one. We could improve performance by keeping an index of shapes where the shapes are indexed by the matching criteria, i.e. currently that would be a dictionary that maps qualified table names to the shapes that are defined over those tables. That warrants a PR of its own though.

packages/sync-service/lib/electric/postgres/replication_client/collector.ex
kevin-dp: We're using the new relation as a default value for `old_rel`:
```ex
old_rel = Map.get(state.relations, id, rel)
```

So, if the relation is not yet known, than `old_rel = rel` and we go into the else branch.

alco: Oh, I missed that. I would prefer `old_rel` to be `nil` for readability. The fact that the variable may be referring the new relation contradicts its name, making the code more confusing.

kevin-dp:
> This is a good partial solution for the issue but it is not complete.
> 
> Every time Electric opens a new replication connection to Postgres, it will have an empty relation cache. So in a situation where Electric was running for some time, then got shut down, then a table's schema was changed in Postgres, then Electric was launched again, it would receive a relation message while having an empty relation cache, so it wouldn't detect the schema change.
> 
> The proper solution for detecting schema changes accross restarts would be to pass all relation messages to the shape log collector and have it load the previous relation schema from its shape metadata store to compare against the new relation schema received in the replication stream.

Good catch!
I pushed a commit that addresses this problem by storing relation information in the meta table and having the shape log collector fetch the relation from the shape cache and comparing the new relation against the stored one.

Note that we are now storing relations in ETS but we are never actually cleaning those up. We would want to delete a relation from ETS when the corresponding table is dropped from the DB but i don't think Postgres' logical replication stream informs us when a table is dropped? @alco @icehaunter

packages/sync-service/lib/electric/postgres/replication_client/collector.ex
kevin-dp: Sure, i introduced persistent storage of the relations and it now works as per your suggestion.

packages/sync-service/test/electric/postgres/replication_client/collector_test.exs
```diff
@@ -58,21 +58,22 @@ defmodule Electric.Postgres.ReplicationClient.CollectorTest do
...
-       %{collector: collector} do
-    new_relation = %{
-      @relation
-      | columns: [%LR.Relation.Column{name: "id", flags: [:key], type_oid: 20, type_modifier: -1}]
-    }
-
-    log = capture_log(fn -> Collector.handle_message(new_relation, collector) end)
-    assert log =~ "Schema for the table public.users had changed"
-  end
+  # TODO: move this test to the shape log collector tests
```
alco, line 66: Forgot to remove this comment?

kevin-dp:
@alco ugh.. right i thought ETS was persistent, my bad.

packages/sync-service/lib/electric/postgres/replication_client/collector.ex
```diff
@@ -51,23 +52,21 @@ defmodule Electric.Postgres.ReplicationClient.Collector do
...
-        new_state
-      }
-    else
+
+    {
+      %Relation{
+        id: id,
+        schema: ns,
+        table: name,
+        columns: Enum.map(cols, fn col -> %Column{name: col.name, type_oid: col.type_oid} end)
```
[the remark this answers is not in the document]
kevin-dp: That doesn't work as `Electric.Postgres.LogicalReplication.Messages.Relation.Column` doesn't implement the `Enumerable` protocol. Should we make it implement it?

packages/sync-service/test/electric/replication/shape_log_collector_test.exs
```diff
@@ -315,7 +315,44 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
...
+        schema: "public",
+        table: "test_table",
+        columns: []
+      }
+
+      MockShapeCache
+      |> expect(:get_relation, 1, fn ^relation_id, _ -> rel end)
+      |> allow(self(), server)
+
+      assert :ok = ShapeLogCollector.handle_relation_msg(rel, server)
```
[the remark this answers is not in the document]
kevin-dp: Indeed, i'll make it explicit:
```ex
|> expect(:clean_shape, 0, fn _, _ -> :ok end)
```

packages/sync-service/lib/electric/postgres/replication_client/collector.ex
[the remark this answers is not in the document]
alco: I don't see why that protocol would be needed here:
```elixir
iex(1)> defmodule Column do
...(1)> defstruct [:name, :type_oid]
...(1)> end
{:module, Column,
 <<70, 79, 82, 49, 0, 0, 8, 24, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 249,
   0, 0, 0, 22, 13, 69, 108, 105, 120, 105, 114, 46, 67, 111, 108, 117, 109,
   110, 8, 95, 95, 105, 110, 102, 111, 95, 95, ...>>,
 %Column{name: nil, type_oid: nil}}

iex(2)> map = %{name: "email", type_oid: 1355, some_other_field: "doesn't matter"}
%{name: "email", type_oid: 1355, some_other_field: "doesn't matter"}

iex(3)> struct(Column, map)
%Column{name: "email", type_oid: 1355}

iex(4)> cols = [map, %{name: "name", type_oid: 33}]
[
  %{name: "email", type_oid: 1355, some_other_field: "doesn't matter"},
  %{name: "name", type_oid: 33}
]

iex(5)> Enum.map(cols, &struct(Column, &1))
[%Column{name: "email", type_oid: 1355}, %Column{name: "name", type_oid: 33}]
```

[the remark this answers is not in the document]
alco: Oh, I see what you're saying now. My code works because I pass maps as the 2nd argument to `struct()`, but in the case of `LR.Relation{columns: cols}`, `cols` is a list of _structs_, not maps:
```
iex(11)> term = %Column{name: "email", type_oid: 1355}
%Column{name: "email", type_oid: 1355}

iex(12)> struct(Column, term)
** (Protocol.UndefinedError) protocol Enumerable not implemented for %Column{name: "email", type_oid: 1355} of type Column (a struct)
```

This can be worked around by rewriting my original suggestion like so:
```
columns: Enum.map(cols, &struct(Column, Map.from_struct(&1))
```
but at this point it doesn't provide any readability benefit on top of what you've already wrote. So disregard this whole suggestion.

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -424,7 +422,11 @@ defmodule Electric.ShapeCache do
...
-    Storage.get_relations(state.storage)
-    |> Stream.each(&store_relation_ets(&1, state))
+    state.storage
+    |> Storage.get_relations()
+    |> Enum.each(&store_relation_ets(&1, state))
   end
 end
+
+# TODO:
+# - Write tests to check that shape cache recovers relations on restart
```
alco, line 432: This is already tested in the new ShapeCache test, no?

alco (approved):
:muscle:

packages/sync-service/lib/electric/shape_cache.ex
kevin-dp: Yes, this is a leftover comment. Good catch, will remove it.

gruckion:
Sorry, why are we not using event triggers? 

https://github.com/electric-sql/pglite/pull/266

Official PG doesn’t support event triggers in single user mode. But this PR changes that. So why are we not fixing this correctly with an event sourcing solution?

note though Supabase doesn’t allow provide superuser access.

https://zero.rocicorp.dev/docs/connecting-to-postgres

thruflo:
Do you mean detecting DDL changes in the source Postgres using event triggers? Because yes, event triggers are not compatible with many Postgres hosting platforms.

Issue #1469, opened 2024-07-15
Detect race condition when creating a new shape


balegas:
When creating a new shape we start tracking LR stream for changes that match that shape to reconcile it with the result of initial query for the shape. 

While the chances are small, it is possible that between starting a new transaction and querying the ```xmin``` for that transaction, some transactions that have a ```xid``` greater than ```xmin``` might have already been processed (and dropped) at the server.

The solution discussed is to have a fixed-size buffer of transactions in Electric to be able to reconcile the shape state.

icehaunter:
This is addessed by #1551

msfstef:
Closing this as it has been addressed

PR #1551, opened 2024-08-21
Move to process- and storage-per-shape

### LINKED ISSUE #1522: Shape maintenance is single-process
The data flow is postgres replication -> replication consumer -> shape log collector -> shape cache.

Within the shape log collector, each available shape is updated via a for loop over the list of active shapes. The data for each shape is written to a shared storage instance.

So for large number of shapes, the throughput is limited by this single-process update of shapes. It also means that per-shape state is harder to manage.

Instead of this single process and storage per replication stream we should have a process and storage instance per shape, removing contention and distributing the load of updating the shapes across the CPUs.

The downside is increased copying of transaction data between threads, but the benefits should easily outweigh this.

magnetised:
Fixes #1522, #1469


balegas:
We should be able to run some benchmarks against this PR to validate the assumptions. It all sounds quite sane, therefore it will be nice to see the speed up.

Let’s check with the team how to take those numbers.

packages/sync-service/test/support/mocks.ex
```diff
@@ -0,0 +1,5 @@
+defmodule Support.Mock do
```
magnetised, line 1: I've moved these mocks to a shared module because I was getting sporadic errors caused by multiple test modules trying to generate the same mocks simultaneously.

packages/sync-service/lib/electric/shape_cache/storage.ex
```diff
@@ -88,22 +108,36 @@ defmodule Electric.ShapeCache.Storage do
           latest_offset: non_neg_integer(),
           snapshot_xmin: non_neg_integer()
         ]
-  def list_shapes({mod, opts}), do: apply(mod, :list_shapes, [opts])
+  # TODO: remove this, or assert that opts are per-shape?
```
kevin-dp, line 109: Is this TODO still relevant?

packages/sync-service/lib/electric/shape_cache/storage.ex
```diff
@@ -116,43 +150,56 @@ defmodule Electric.ShapeCache.Storage do
...
-
-  @doc "Get all stored relations"
-  @spec get_relations(storage()) :: Enumerable.t(Relation.t())
-  def get_relations({mod, opts}), do: mod.get_relations(opts)
+  def has_log_entry?(shape_id, offset, {mod, opts}) do
+    shape_opts = mod.for_shape(shape_id, opts)
+    mod.has_log_entry?(shape_id, offset, shape_opts)
+  end
+
+  # @doc "Store a relation containing information about the schema of a table"
```
kevin-dp, line 181: Can remove these commented functions

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -148,6 +163,91 @@ defmodule Electric.Shapes.Shape do
...
+    shape_matches?(shape, old_schema, old_table)
+  end
+
+  def is_affected_by_relation_change?(shape, %Changes.RelationChange{
+        new_relation: %Changes.Relation{schema: schema, table: table}
+      }) do
+    shape_matches?(shape, schema, table)
+  end
+
+  # TODO: test this machinery of cleaning shapes on any migration
```
kevin-dp, line 186: Can remove this comment. We decided to just go with blunt clean on any migration approach.

kevin-dp (approved):
I went through the code changes (skipped most of the changes in the tests though).
Changes look good to me, the `GenStage` refactoring made it sometimes a bit hard to understand what's going on from the diffs as a lot of code has moved.

packages/sync-service/lib/electric/persistent_kv/filesystem.ex
```diff
@@ -0,0 +1,56 @@
...
+
+      case File.read(path) do
+        {:ok, data} -> {:ok, data}
+        {:error, :enoent} -> {:error, :not_found}
+        error -> error
+      end
+    end
+
+    def set(fs, key, value) when is_binary(key) do
+      with file_path = mkdir(fs, key),
```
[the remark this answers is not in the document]
magnetised: The keys - at least ATM - are not from user or app data, they're created by us so there's no particular need for input sanitising.

packages/sync-service/lib/electric/persistent_kv.ex
```diff
@@ -0,0 +1,4 @@
+defprotocol Electric.PersistentKV do
```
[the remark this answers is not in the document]
magnetised: I personally prefer protocols for this kind of interface as a protocol is basically implementation pointer plus (fixed) configuration. Our current impl is basically re-implementing this via behaviours. My rule of thumb about protocols vs behaviours is if the api mutates the config then use behaviours, if the config is static, use protocols, but as I say, this is just my personal choice. If this is going to confuse things I can re-write.

packages/sync-service/lib/electric/persistent_kv/memory.ex
```diff
@@ -0,0 +1,50 @@
...
+  def contents(s) do
+    s.pid
+    |> Agent.get(& &1)
+    |> Enum.sort_by(&elem(&1, 0))
+  end
+
+  defimpl Electric.PersistentKV do
+    def set(memory, key, value) do
+      Agent.update(memory.pid, fn data ->
+        notify(memory, {:set, key, value})
```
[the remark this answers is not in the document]
magnetised: > Or will an Agent always finish the current function set before get is executed because they are an abstraction on top of Erlang processes and thus this is not possible?

This. Plus this notify is a test-only thing (the whole implementation is a test-only thing)

packages/sync-service/lib/electric/postgres/inspector.ex
```diff
@@ -5,9 +5,13 @@ defmodule Electric.Postgres.Inspector do
   @type column_info :: %{
           name: String.t(),
           type: String.t(),
+          type_mod: integer() | nil,
```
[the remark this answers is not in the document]
magnetised: Ah, my bad. I was just updating the types based on usage in the code. There are also a bunch of refs to `type_modifier` - is this the same thing?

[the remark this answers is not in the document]
magnetised: Yeah, the `DirectInspector` returns both the `type_id` tuple and this `type_mod` so the type spec is accurate, at least in terms of the values coming out of the pg query

packages/sync-service/lib/electric/shape_cache/storage.ex
magnetised: Mostly...

packages/sync-service/lib/electric/postgres/inspector.ex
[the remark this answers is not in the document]
kevin-dp: Yeah, `type_modifier` is on the replication stream. Should be the same thing, it just originates from a different place (replication stream vs explicit DB query by the inspector).

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -0,0 +1,205 @@
...
+
+  def init(config) do
+    %{log_producer: producer, storage: storage} = config
+
+    Logger.metadata(shape_id: config.shape_id)
+
+    :ok = ShapeCache.Storage.initialise(storage)
+
+    # TODO: replace with a more specific call to get the current position
+    :ok = ShapeCache.Storage.add_shape(config.shape_id, config.shape, storage)
```
icehaunter, line 49: I don't understand the point of this or `list_shapes` call, if the storage is per-shape as it is right now. Can you explain this bit please?

icehaunter (approved):
I like the changes, and what they give us in terms of potential scalability. I think this warrants a rewrite of the `Storage` api to better reflect a per-shape approach, wdyt?

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: Yeah, this is temporary. Didn't want to totally re-write the storage api in this pr. I've created #1563 which documents the requirements. This `list_shapes` call should be replaced with some "get the current position" function which gets the snapshot xmin and offset for the shape.

magnetised:
> I like the changes, and what they give us in terms of potential scalability. I think this warrants a rewrite of the `Storage` api to better reflect a per-shape approach, wdyt?

already planned: #1563

icehaunter:
So this PR doesn't bring that much performance benefits on it's own:
1. Snapshot creation is already done in separate `Task` process in `main`, so no benefits here
2. `CubDB` is used in a same way, so same overhead costs for open-close-rename files on each
3 `CubDb` transaction is NOT the answer, it turned out. I did a local test, and what it does is force you to accumulate changes in memory AND THEN write them in bulk, which is not what we want for large shapes.

It *should* bring perf benefits when we're fanning out a txn over 1k+ existing shapes, but I don't have a bench for that ready yet.

icehaunter (changes requested):
Given the new process arch, a bug or failure-to-write to a log a particular shape becomes unrecoverable, right?
Current arch crashes the writer process, which crashes the replication client, which will continue from an unacknowledged txn, which will get re-written to all shapes, which is safe because we have a test that a write to the same offset in storage is idempotent
New arch crashes the writer process, which does nothing to the producer, and when it's back up it continues consuming transactions not knowing that some of those were missed.
I think we need to know how to address that before merging

magnetised:
Requirement for handling of crashed consumers documented here: https://github.com/electric-sql/electric/issues/1585

icehaunter (approved):
OK to merge, not OK to release until #1585 is addressed

Issue #1552, opened 2024-08-21
infinite syncing loop


stratospark:
I'm getting this issue where the frontend repeatedly tries to sync data after getting 409 errors. It seems as if the offset being requested is incorrect.

I've created a simple reproduction where modifying two tables within a transaction creates this behavior.

https://github.com/stratospark/electricsql-broken-sync-repro

alco:
Hey @stratospark :wave: 

Brilliant work on the report and the reproduction repo! One suggestion: use the `electricsql/electric` Docker Hub image if you want to run the latest published release or `electricsql/electric:canary` to run the latest build from the `main` branch. `electricsql/electric-next` is no longer updated.

I can confirm that the behaviour you've described is reproducible using the latest `canary` image of `electricsql/electric`.

kevin-dp:
![Screenshot 2024-08-22 at 10 27 19](https://github.com/user-attachments/assets/6812f51b-2847-44c5-b03b-ba3bdd34f9ca)

Looking at the network activity i notice that upon receiving a 409, the response contains a `location` header for the new location we need to hit and also the last offset and the shape ID. Looking and subsequent requests, it seems that the client is hitting the new location but also still retrying the old one.

I believe it is due to this part of the code in the JS client:
```ts
let response!: Response
try {
  const maybeResponse = await this.fetchWithBackoff(fetchUrl)
  if (maybeResponse) response = maybeResponse
  else break
} catch (e) {
  if (!(e instanceof FetchError)) throw e // should never happen
  if (e.status == 409) {
    // Upon receiving a 409, we should start from scratch
    // with the newly provided shape ID
    const newShapeId = e.headers[`x-electric-shape-id`]
    this.reset(newShapeId)
    this.publish(e.json as Message[])
    continue
  } else if (e.status >= 400 && e.status < 500) {
    // Notify subscribers
    this.sendErrorToUpToDateSubscribers(e)
    this.sendErrorToSubscribers(e)
    // 400 errors are not actionable without additional user input, so we're thr
    throw e
  }
}
const { headers, status } = response
const shapeId = headers.get(`X-Electric-Shape-Id`)
if (shapeId) {
  this.shapeId = shapeId
}
const lastOffset = headers.get(`X-Electric-Chunk-Last-Offset`)
if (lastOffset) {
  this.lastOffset = lastOffset as Offset
}
```

[...]

PR #1555, opened 2024-08-22
fix: don't check for exact log offset existence

kevin-dp:
Fixes https://github.com/electric-sql/electric/issues/1552.
This PR addresses a bug that caused an infinite loop of failing requests.
The root cause for the bug is that we were always using the offset of the last change in a transaction but a shape may not be affected by all changes in the transaction (in which case the real latest offset would be the offset of the latest relevant change for that shape in the transaction). Because of that, when the shape plug checks if we have a log entry for the given offset, it may not exist and return a 409.

As a result of not finding an entry for the exact offset, Electric responds with a 409. The client thus comes back with an initial sync request (offset -1) and then Electric sends the latest offset it knows about which is again the commit offset. And thus we do a request with that offset and get again a 409. Then we do again an initial sync request and we are in this loop forever.

We should note that although the commit offset of the transaction may be bigger than the offset of the real latest change, it is still valid to use the commit offset because Postgres linearizes concurrent transactions. So we can be sure that the next relevant changes will happen in a subsequent transaction which will have offsets that are bigger than this offset (because the next transaction's commit LSN will be bigger than this transaction's commit LSN which is used as the transaction ID which is part of the offset of the changes in the transaction (offset = `{ <tx_id>, <op_index> }`).

We solve the problem by no longer looking for an entry for the exact log offset. We now only check that the `shape_id` exists. If the `shape_id` exists then we know that we have the full log for that shape (because we don't do any compaction at the moment). And thus we are able to resume from any point in the shape (doesn't need to be an exact point as we do range queries).

### DIFF
--- a/.changeset/wicked-bottles-attack.md
+++ b/.changeset/wicked-bottles-attack.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Don't search for exact log entry with provided offset. Fixes a bug that caused an infinite loop of initial syncs followed by 409s.
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -179,14 +179,14 @@ defmodule Electric.Plug.ServeShapePlug do
     conn
   end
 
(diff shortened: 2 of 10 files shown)


magnetised (approved):
Lovely, thanks

PR #1561, opened 2024-08-26
fix: refactor out JSON serialization

### LINKED ISSUE #1536: JSONB fields that are null are sent as empty string.
To reproduce it, create a table with a JSONB field and insert some data but leave the JSONB field to be null.
Request the shape for table, and you will see the JSONB field being sent as an empty string, but it should be null.

This causes the sync to pglite to fail because, parsing an empty string throws.

Here's a [link](https://discord.com/channels/933657521581858818/933657523049885698/1274436467309154306) discussion related to this issue in Discord.

icehaunter:
Closes #1536, pulls out the JSON-ification responsibility out of storage and into querying for easier iteration over that responsibility.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
@@ -1,6 +1,6 @@
 defmodule Electric.ShapeCache.InMemoryStorage do
+  alias Electric.Shapes.Querying
   alias Electric.ConcurrentStream
-  alias Electric.LogItems
   alias Electric.Replication.LogOffset
   alias Electric.Replication.Changes.Relation
   alias Electric.Telemetry.OpenTelemetry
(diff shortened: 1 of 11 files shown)


packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
```diff
@@ -101,25 +101,21 @@ defmodule Electric.ShapeCache.InMemoryStorage do
...
         ) :: :ok
-  def make_new_snapshot!(shape_id, shape, query_info, data_stream, opts) do
+  def make_new_snapshot!(shape_id, data_stream, opts) do
     OpenTelemetry.with_span("storage.make_new_snapshot", [storage_impl: "in_memory"], fn ->
       ets_table = opts.snapshot_ets_table
 
       data_stream
-      |> LogItems.from_snapshot_row_stream(@snapshot_offset, shape, query_info)
       |> Stream.with_index(1)
-      |> Stream.map(fn {log_item, index} ->
```
magnetised, line 120: should we not make the storage api simpler, to remove this duplication?

magnetised (approved):
Nice.

packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
icehaunter: Storage API is very simple after this change, it's fed a list of JSON objects. Each storage can decide how to chunk or store data on it's own (e.g. direct-to-file wouldn't do chunking at all)

Issue #1550, opened 2024-08-21
Shape meta data not cleaned up properly on delete


kevin-dp:
When we delete a shape we are forgetting to clean up some meta data somewhere because the old schema information re-appears if we later create a new shape for the same table.

To reproduce this problem, first create a table:
```sql
CREATE TABLE foo (c1 INT PRIMARY KEY); INSERT INTO foo VALUES (1), (2);
```

Now, fetch the shape (i'm using HTTPie, can be done with curl too):
```bash
HTTP GET "localhost:3000/v1/shape/foo" offset==-1
```

Ok, we get the shape with the 2 rows. This is fine.
Now drop the table:
```sql
DROP TABLE foo;
```

Now, tell Electric to delete the shape:
```bash
HTTP DELETE "localhost:3000/v1/shape/foo"
```

Ok, the shape is deleted.
Let's recreate the table but with an extra column:
```sql
CREATE TABLE foo (c1 INT PRIMARY KEY, c2 TEXT);
INSERT INTO foo VALUES (9, 'bar');
```

Let's fetch that table:
```
HTTP GET "localhost:3000/v1/shape/foo" offset==-1
````

Now, the returned data is wrong as the rows only have the c1 column and not the c2 column (also the schema only contains c1).

kevin-dp:
This seems to be the reason for the bug:
- when an HTTP request arrives, we create the shape
- `Shape.new` load the column info via the inspector: `Inspector.load_column_info`
- The first time, the inspector loads this information from PG and stores it in ETS
- All subsequent `load_column_info` calls fetch this information from ETS

Thus, we must ensure that the ETS entry is removed when a table is altered.

PR #1562, opened 2024-08-26
fix: clean cached column information on delete

kevin-dp:
Fixes https://github.com/electric-sql/electric/issues/1550.
This PR extends the shape log collector to also clean the cached column information from ETS when a relation changes. This ensures that when a table is migrated, we don't re-use the old table information from ETS but instead load it from Postgres and re-populate the cache with the new column info.

**There's one corner case that this PR does not address:**
- Create table, insert data
- Sync a shape containing that table
- Drop the table
- Delete the shape
- Recreate the table but with a different schema
- Insert some data into the new table
- Sync a shape containing the newly recreated table

In the above scenario, when syncing the newly recreated table, we get the new data but in the old schema (so we only get the columns that also existed in the old schema). This is because Postgres logical replication stream does not inform us when a table is dropped.

As a result, we can't detect that a table was dropped and thus don't know that we need to clean the cached column information. It's only when we get a Relation message that we know this. But Postgres only sends a Relation message the first time the data in the table changes and **we're subscribed to that table in the replication stream** (and that's only after syncing the shape).

So, the data that was inserted before we synced the table in the last step, does not lead to a Relation message. Only if we insert data into the table after that sync step, will Postgres send a Relation message that will make us clean the cached column information. But note that at that point the row that was inserted in the previous step is already stored in storage in the format of the old schema.

### DIFF
--- a/.changeset/soft-gifts-watch.md
+++ b/.changeset/soft-gifts-watch.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Clean cached column info on relation changes.
--- a/packages/sync-service/lib/electric/application.ex
+++ b/packages/sync-service/lib/electric/application.ex
@@ -28,7 +28,7 @@ defmodule Electric.Application do
 
       inspector =
         {Electric.Postgres.Inspector.EtsInspector,
-          server: Electric.Postgres.Inspector.EtsInspector}
+         server: Electric.Postgres.Inspector.EtsInspector}
 
       shape_cache =
         {Electric.ShapeCache,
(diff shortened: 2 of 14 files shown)


magnetised (approved):
You've done it again... I have a lot of rebasing in my future... 🤪 

Looks good, though I'm not sure your relation change tests belong under the shape log collector. I would test that the log collector puts relations into the GenStage pipeline, then put tests for what's actually done with that message into the tests for the component that's acting on it, in this case it's the shape cache

Bit academic though as my current pr moves the relations out of the gen stage pipeline completely, in favour of a direct call to the shape cache.

kevin-dp:
@magnetised i'm feeling your pain :D spent the day rebasing this one on top of storage changes

kevin-dp:
@magnetised Thanks for your review. As discussed in DM, i won't move the unit tests as you have an upcoming PR that moves the code that handles relation changes out of the genstage. So the unit tests can be moved together with that change.

Issue #1455, opened 2024-07-19
HTTP proxy authorization / permissions example


thruflo:
The example is a one page app showing a list of projects (or whatever). There are two roles, admin and user. The admin can see all projects and the user can see only ones that they own.

The shape calls should proxy through the API similar to the next.js or remix examples and set the where clause for the `user` role so it filters down projects.

At the top of the page, you can click to become the admin or one of 3 users. That'll set a query string with the role of the user and then refresh the page so the app re-fetches the right shape.

Use a custom `fetchClient` to set custom auth headers.

PR #1566, opened 2024-08-26
Feat: add Auth example & guide

KyleAMathews:
Fixes https://github.com/electric-sql/electric/issues/1455

https://github.com/user-attachments/assets/eab62c23-513c-4ed8-a6fa-249b761f8667


docs/guides/authentication-and-authorization.md
```diff
@@ -0,0 +1,109 @@
...
+
+In the app, you pass a `fetchWrapper` to the Electric client which adds your `Authorization` header when
+Electric requests shape data.
+
+## Sample code
+
+In the client:
+
+```tsx
+const fetchWrapper = async (...args: Parameters<typeof fetch>) => {
```
balegas, line 27: Love this. This is a real neat pattern.

examples/auth/app/shape-proxy/[...table]/route.ts
Comment (balegas) [inline examples/auth/app/shape-proxy/[...table]/route.ts:1]: One thing I don't like a lot in these examples is that the we proxy the internal shape API, where a more realistic example IMO would be to translate the request path into the shape url internally. 

I wonder if we should adopt that pattern globally --- I get that this is a complexity that the example is not trying to cover.

balegas (approved):
Vey cool. Thanks for working on this.

docs/guides/authentication-and-authorization.md
KyleAMathews: copied it from @msfstef actually 😎

examples/auth/app/shape-proxy/[...table]/route.ts
Comment (KyleAMathews) [reply examples/auth/app/shape-proxy/[...table]/route.ts:1]: Yeah... but I think a lot of people will just use the same shape API — it is a pretty reasonable design. People could convert e.g. `/api/users` -> `/v1/shapes/users` but `/api/shapes/users` would make sense to a lot of people and clearly differentiate between shape requests and "normal" requests.

Comment (balegas) [reply examples/auth/app/shape-proxy/[...table]/route.ts:1]: yes and no? It leaks database information. I think many people probably want to convert it to the _domain_ of your API endpoints.

No strong opinion about specific action for this PR, but something people might ask about

kevin-dp (approved):
Code looks good.
My only remark is that the example is not representative of real-world authentication and authorization. It only teaches to set the authorization header. I think basic authentication would be better as it is a real authentication scheme and remains simple enough for an example app. I can give it a try later today if you want.

kevin-dp:
I went ahead and extended this example with basic auth.
Opened a PR for it: https://github.com/electric-sql/electric/pull/1567

examples/auth/app/page.tsx
```diff
@@ -0,0 +1,147 @@
...
+        <ul>
+          <li style={{ display: `inline` }}>
+            <a
+              href=""
+              onClick={(e) => {
+                e.preventDefault()
+                window.location.search = ``
+              }}
+              className={
+                window.location.search === `` ? `active-link` : `white-link`
```
kevin-dp, line 59: Accessing `window` here doesn't play nicely with SSR.
I get an error on the NextJS server because it complains that `window` is not defined.
Instead, the `useSearchParams` hook works perfectly:
```tsx
import { useSearchParams } from 'next/navigation'

export default function Home() {
  const searchParams = useSearchParams()
  ...
  className={
    searchParams.get(`username`) === null
      ? `active-link`
      : `white-link`
  }
  
  ...
  
  className={
    searchParams.get(`username`) === `Alice`
      ? `active-link`
      : `white-link`
  }
}
```

examples/auth/app/page.tsx
```diff
@@ -0,0 +1,147 @@
...
+          </li>
+          {` `}|{` `}
+          <li style={{ display: `inline` }}>
+            <a
+              href="?user=1"
+              onClick={(e) => {
+                e.preventDefault()
+                window.location.search = `?org_id=1`
+              }}
+              className={
```
kevin-dp, line 73: This `className` logic is repeated for every user.
We can hide it behind this function:
```ts
const classFor = (user: string | null) => {
  return searchParams.get(`username`) === user ? `active-link` : `white-link`
}

...
className={classFor(null)}
...
className={classFor(`Alice`)}
...
```

examples/auth/app/page.tsx
```diff
@@ -0,0 +1,147 @@
...
+  [key: number]: User[]
+}
+
+const fetchWrapper = async (...args: Parameters<typeof fetch>) => {
+  const queryParams = new URLSearchParams(window.location.search)
+  const org_id = queryParams.get(`org_id`)
+  const modifiedArgs = [...args]
+  if (org_id) {
+    const headers = new Headers((modifiedArgs[1] as RequestInit)?.headers || {})
+    headers.set(`Authorization`, org_id)
```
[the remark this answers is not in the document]
KyleAMathews: Just to reiterate my comment in https://github.com/electric-sql/electric/pull/1567#issuecomment-2312465343 — the goal here is to teach how an API can authenticate and authorize who sees shape and the exact shape that returns. People can go elsewhere for realistic examples of authentication.

kevin-dp:
@KyleAMathews i don't really get the added value of this example app. I was expecting this auth example app to be a starting point that people could clone and build their own app that features authentication. But it's not really doing any form of authentication.. it's only passing an organization ID to the server which is never actually checked. I get that this may be more intended as a blue print of how you can do authentication, but with the way how it's implemented now, the blueprint is not ideal imo. If the goal is to be a blue print of how one would do actual auth, then i would move all the application-dependent auth stuff into a separate function such that the pattern is very clear and that the user can just take the boilerplate code, fill in the auth function with their own implementation, and off they go. And if you want something runnable, you could provide a dummy implementation of that function that uses the org ID as you did.

KyleAMathews:
> I was expecting this auth example app to be a starting point that people could clone and build their own app that features authentication

Examples are just for reading and learning from. Starters are for starting actual projects off. The point of this again isn't to teach anyone about auth (that's why outside of the scope of Electric) but how to "auth" Shape requests within whatever authentication setup they're doing. Examples are only real within the very narrow scope they're trying to teach and then ignore or fake everything else.

Issue #1441, opened 2024-08-02
Implement support for the missing comparison predicates in WHERE clauses


alco:
Just for completeness' sake and since it's not that much work, we can support all comparison predicates described in [Section 9.2](https://www.postgresql.org/docs/current/functions-comparison.html#FUNCTIONS-COMPARISON-PRED-TABLE) of the official Postgre docs. 

- [x] `x IS DISTINCT FROM y`
- [x] `x IS NULL`
- [x] `x IS NOT NULL`
- [x] `x IS TRUE`
- [x] `x IS NOT TRUE`
- [x] `x IS FALSE`
- [x] `x IS NOT FALSE`
- [x] `x IS NOT DISTINCT FROM y` - https://github.com/electric-sql/archived-electric-next/pull/242

#1569:
- [x] `x BETWEEN y AND z`
- [x] `x NOT BETWEEN y AND z`
- [x] `z BETWEEN SYMMETRIC y AND z`
- [x] `z NOT BETWEEN SYMMETRIC y AND z`
- [x] `x IS UNKNOWN`
- [x] `x IS NOT UNKNOWN`

PR #1569, opened 2024-08-27
feat(sync-service): Support missing comparison predicates

robacourt:
Support for the missing comparison predicates in WHERE clauses, as mentioned in #1441:
-  x BETWEEN y AND z
-  x NOT BETWEEN y AND z
-  z BETWEEN SYMMETRIC y AND z
-  z NOT BETWEEN SYMMETRIC y AND z
-  x IS UNKNOWN
-  x IS NOT UNKNOWN

### DIFF
--- a/packages/sync-service/lib/electric/replication/eval/env/explicit_casts.ex
+++ b/packages/sync-service/lib/electric/replication/eval/env/explicit_casts.ex
@@ -6,54 +6,52 @@ defmodule Electric.Replication.Eval.Env.ExplicitCasts do
 
   ## List of explicit casts
 
-  | source    | target         | function name   |
-  | --------- | -------------- | ----------------|
-  | bool      | int4           | bool_to_int4    |
-  | char      | int4           |                 |
-  | int8      | bit            |                 |
-  | int4      | bit            |                 |
-  | int4      | bool           | int4_to_bool    |
-  | int4      | char           |                 |
-  | text      | xml            |                 |
-  | lseg      | point          |                 |
-  | box       | point          |                 |
-  | box       | lseg           |                 |
-  | box       | circle         |                 |
-  | polygon   | box            |                 |
-  | polygon   | circle         |                 |
-  | polygon   | point          |                 |
-  | circle    | box            |                 |
-  | circle    | point          |                 |
-  | circle    | polygon        |                 |
-  | bpchar    | xml            |                 |
-  | varchar   | xml            |                 |
-  | bit       | int4           |                 |
-  | bit       | int8           |                 |
-  | jsonb     | bool           |                 |
-  | jsonb     | int8           |                 |
-  | jsonb     | int2           |                 |
-  | jsonb     | int4           |                 |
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/replication/eval/env/explicit_casts.ex
```diff
@@ -6,49 +6,51 @@ defmodule Electric.Replication.Eval.Env.ExplicitCasts do
...
+  | jsonb     | float8         |                 |
+  | jsonb     | numeric        |                 |
+  | int4range | int4multirange |                 |
+  | numrange  | nummultirange  |                 |
+  | tsrange   | tsmultirange   |                 |
+  | tstzrange | tstzmultirange |                 |
+  | daterange | datemultirange |                 |
+  | int8range | int8multirange |                 |
+  | xid8      | xid            |                 |
+  | unknown   | bool           | unknown_to_bool |
```
icehaunter, line 46: Why do we need that explicit cast? I believe that an explicit `::bool` cast shouldn't be special-cased here in this list, but instead should always work for a `NULL` value, as it always tries to work for text values

packages/sync-service/lib/electric/replication/eval/parser.ex
```diff
@@ -365,26 +372,68 @@ defmodule Electric.Replication.Eval.Parser do
...
+         comparisons = [left_comparison, right_comparison],
+         {:ok, comparisons} <- Utils.map_while_ok(comparisons, &maybe_reduce/1),
+         {:ok, reduced} <-
+           build_bool_chain(%{name: "and", impl: &Kernel.and/2}, comparisons, expr.location) do
+      {:ok, reduced}
     end
   end
 
-  defp build_or_chain([head | tail], location) do
+  defp between_sym(expr, left_bound, right_bound, refs, env) do
```
icehaunter, line 421: Can you add a comment here about suboptimality of recalculating a subtree for both branches? We don't have to address it today, just for the futute

icehaunter (changes requested):
LGTM, with one request about explicit casts

packages/sync-service/lib/electric/replication/eval/parser.ex
robacourt: Out of curiosity, how would it be optimised? I thought of doing a seeing whether the left or right bounds were greater before doing the between but I'm didn't think that would be any quicker apart from in specific cases such as left bound and right bound being constants.

packages/sync-service/lib/electric/replication/eval/env/explicit_casts.ex
robacourt: Something like this? https://github.com/electric-sql/electric/pull/1569/commits/d18be4af952bd884ee8c20284e3e424532b43e0e

packages/sync-service/lib/electric/replication/eval/parser.ex
robacourt: Done.

Issue #1453, opened 2024-07-23
Message type doesn't allow discriminating between a ControlMessage and ChangeMessage at the type level.


samwillis:
The code below is unable to discriminate that the message is a `ChangeMessage`:

```ts
if (message.headers.action === "insert") {
  // message is still ControlMessage | ChangeMessage
}
```

It would be nice if there was a way to discriminate them without having to manually use `as`

kevin-dp:
It does not seem possible to discriminate on the value of a key that may or may not be present. It's only possible to discriminate on a key that is known to be present but that has different values in different types.

For example:
```ts
export type ControlMessage = {
  tag: 'Control'
  headers: Header
}

export type ChangeMessage<T> = {
  tag: 'Change'
  key: string
  value: T
  headers: Header & { operation: `insert` | `update` | `delete` }
  offset: Offset
}

// Define the type for a record
export type Message<T extends Value = { [key: string]: Value }> =
  | ControlMessage
  | ChangeMessage<T>

const message: Message = undefined as unknown as Message

if (message.tag === 'Change') {
    message.value // `message` is typed as `ChangeMessage`
} else {
    message.headers // `message` is typed as `ControlMessage`
}
```

However, we don't have this `tag` property on which we can discriminate.
So let's change the `Header` interface for a second:
```ts
interface Header {
  //[key: string]: Value
  foo: number
}
```

Now let's try to discriminate as you did:
```ts
if (message.headers.operation === "insert")
```

uh oh... type error:
```
Property 'operation' does not exist on type 'Header | (Header & { operation: "insert" | "update" | "delete"; })'.
```

Right... we can't be sure that the `operation` property exists on that union type. So let's try to narrow the type using the `in` operator:
```ts
if ('operation' in message.headers) {
    message.value // error 
}
```

Unfortunately, narrowing using the `in` operator on a nested property (`message.headers`) does not seem to work in TypeScript.

So, why didn't TypeScript throw an error in your original example `message.headers.operation === "insert"` ?
Because `Header` is defined as:
```ts
interface Header {
  [key: string]: Value
}
```
So, any key that is a string is accepted, and thus `headers.operation` typechecks but it will not narrow as you were expecting since a `ControlMessage` could also have an `operation` property in its headers.

**If you need to narrow it down, you can use the `in` operator on a top-level property:**
```ts
if ('key' in message) {
  message.value
} else {
  message.headers
}
```

kevin-dp:
[...]

PR #1572, opened 2024-08-27
feat: `Message` type guard helpers

msfstef:
Addresses https://github.com/electric-sql/electric/issues/1453

Adds the type guards to check whether a message is a `ControlMessage` or a `ChangeMessage`, documented, tested, and exported.

I've also honed in on the message type definitions a bit more as I had experienced an uncaught bug due to the generality of the `Headers` type.

### DIFF
--- a/.changeset/pretty-donuts-rule.md
+++ b/.changeset/pretty-donuts-rule.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
+Add `Message` type guard helpers `isChangeMessage` and `isControlMessage`.
--- a/examples/nextjs-example/app/match-stream.ts
+++ b/examples/nextjs-example/app/match-stream.ts
@@ -1,4 +1,8 @@
-import { ShapeStream, ChangeMessage } from "@electric-sql/client"
+import {
+  ShapeStream,
+  ChangeMessage,
+  isChangeMessage,
+} from "@electric-sql/client"
 
 export async function matchStream<T>({
   stream,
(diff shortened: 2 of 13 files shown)


KyleAMathews (approved):
great! These will be very helpful.

Could you check the examples for where these would be helpful as well? Anywhere using ShapeStream. E.g. the Redis example needs it https://github.com/electric-sql/electric/blob/b272d3f5917eb6ddb5102508b3040278626097f9/examples/redis-client/src/index.ts#L24

kevin-dp (approved):
Great!
Is there a reason we have [packages/typescript-client/test/helpers.test.ts](https://github.com/electric-sql/electric/pull/1572/files#diff-7340150745c56dc68f1a9f37daf29f3aa09a6e9fc8ceb52b306c302fe1997c48) and [packages/typescript-client/test/helpers.test-d.ts](https://github.com/electric-sql/electric/pull/1572/files#diff-638834217697333014d94e85da619b0c8ab448289cdbb6fc77e956177ac9e703) ? These are executing different tests, can we merge them into one file?

msfstef:
@KyleAMathews will update codebase in this same PR to use these helpers

@kevin-dp the `test-d.ts` files are "type tests" (see [vitest docs](https://vitest.dev/guide/testing-types)) which essentially are making sure the type system is behaving as intended. In order for the test to test types rather than test code (and separately have `tsc` for the typecheck of the test files), I like having the separate "type tests" that ensure our type system makes sense explicitly.

PR #1577, opened 2024-08-27
chore: fix auth example's caching

KyleAMathews:
The API calls for each user were cached so this PR adds a query-string per org-id so the browser caching works correctly.

### DIFF
--- a/examples/auth/app/page.tsx
+++ b/examples/auth/app/page.tsx
@@ -53,10 +53,7 @@ export default function Home() {
 
   const classFor = (org_id: string | null) => {
     const orgSearchParam = searchParams.get(`org_id`)
-    return orgSearchParam === org_id ||
-      orgSearchParam?.includes(org_id as string)
-      ? `active-link`
-      : `white-link`
+    return orgSearchParam === org_id ? `active-link` : `white-link`
   }
 
   return (
--- a/examples/auth/app/page.tsx
+++ b/examples/auth/app/page.tsx
@@ -30,8 +30,13 @@ const fetchWrapper = async (...args: Parameters<typeof fetch>) => {
 
 const usersShape = (): ShapeStreamOptions => {
   if (typeof window !== `undefined`) {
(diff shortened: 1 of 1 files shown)


examples/auth/app/page.tsx
```diff
@@ -30,8 +30,13 @@ const fetchWrapper = async (...args: Parameters<typeof fetch>) => {
 
 const usersShape = (): ShapeStreamOptions => {
   if (typeof window !== `undefined`) {
+    const queryParams = new URLSearchParams(window.location.search)
+    const org_id = queryParams.get(`org_id`)
     return {
-      url: new URL(`/shape-proxy/users`, window.location.origin).href,
+      url: new URL(
+        `/shape-proxy/users?org_id=${org_id}`,
```
kevin-dp, line 37: what if there is no org_id in the query params?

KyleAMathews: then it's just null

kevin-dp (approved):
Looks good to me!
Just left 1 question wrt org_id param being undefined and the behavior that follows from that.

examples/auth/app/page.tsx
```diff
@@ -46,8 +51,12 @@ export default function Home() {
   const searchParams = useSearchParams()
   const { data: users, isError, error } = useShape(usersShape())
 
-  const classFor = (user: string | null) => {
-    return searchParams.get(`username`) === user ? `active-link` : `white-link`
+  const classFor = (org_id: string | null) => {
+    const orgSearchParam = searchParams.get(`org_id`)
+    return orgSearchParam === org_id ||
+      orgSearchParam?.includes(org_id as string)
```
kevin-dp, line 57: Why do we need 2 checks? i understand `orgSearchParam === org_id` but i don't see why we need `      orgSearchParam?.includes(org_id as string)`

KyleAMathews: I think an earlier version of the logic needed it? But checked and didn't now.

PR #1589, opened 2024-08-28
fix(sync-service): Fix a crash caused by incorrect conversion of an UPDATE into an INSERT or a DELETE depending on whether it is a shape move-in or a shape move-out case

### LINKED ISSUE #1588: Sync service crashes when updated row matches where clause of the shape
### Discussed in https://github.com/electric-sql/electric/discussions/1559

<div type='discussions-op-text'>

<sup>Originally posted by **jo16oh** August 23, 2024</sup>
Hi all 👋
I've been playing around with electric-next for a while and encountered error described in the title. 

### How to reproduce
sync service version : 0.3.3
1. Create table
```sql
CREATE TABLE foo (
  id SERIAL PRIMARY KEY,
  name VARCHAR(255),
  value FLOAT
);

INSERT INTO foo (name, value) VALUES
  ('Alice', 3.14),
  ('Bob', 2.71),
  ('Charlie', -1.618),
  ('David', 1.414),
  ('Eve', 0);
```

2. Establish connection
```js
	url: `http://localhost:3000/v1/shape/foo`,
	where: "name = 'Bob'",
});
```

3. Update existing row
```sql
UPDATE foo SET name = 'Bob' WHERE name = 'Alice';
```

4. Then it crashes!</div>

alco:
Fixes #1588.

### DIFF
--- a/.changeset/kind-crabs-study.md
+++ b/.changeset/kind-crabs-study.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix a crash caused by incorrect conversion of an UPDATE into an INSERT or a DELETE depending on whether it is a shape move-in or a shape move-out case.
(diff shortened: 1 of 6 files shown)


robacourt (approved):
Nice!

icehaunter (approved):

PR #1594, opened 2024-08-29
Resume crashed consumers at a consistent point

### LINKED ISSUE #1585: Crashing Shape consumer should resume at a consistent point
With the old single-process model, if a write to storage failed with some error it would crash the log collector process, which would in turn crash the replication producer process. On restarting the replication stream would resume from the last unacked point and, due to idempotent storage apis, all the shapes would resume at a consistent point.

Under the new multi-process model,  a consumer crashing due to some underlying storage issue will be restarted but the replication stream would have been progressing without it and the crashing shape would resume at some random point in the replication stream.

A crashing consumer should resume at a consistent point. A potential solution is to replicate the previous behaviour, a crashing consumer crashes everything, including the replication producer, and when everything restarts it will do so as before, with postgres resuming the replication stream at the last un-acked point and all clients receiving all transactions.

magnetised:
Fixes #1585

Firstly this moves the replication client, shape cache, consumer supervisor and log  collector under a supervisor with strategy :one_for_all.               
                                                                       
Then a crashing client triggers the `handle_cancel/3` callback in the log  collector which terminates and causes the entire supervision tree to  restart.

The log collector uses a custom `GenStage.Dispatcher` implementation that enables us to make the  `store_transaction/2` callback to be synchronous: it will only return when all consumers have processed/stored the transaction. This works because demand is only signalled to the producer when all dispatchers who wanted a given event have processed it so we can use the `GenStage.demand/2` callback to reply to the replication client calling into the `ShapeLogCollector` process.

As a side effect, this does make the connection and error handling logic in `ConnectionManager` much more fiddly, which isn't ideal. I'm going to work on a restructure to clean that up with @alco's input.

### DIFF
--- a/packages/sync-service/lib/electric/application.ex
+++ b/packages/sync-service/lib/electric/application.ex
@@ -3,6 +3,11 @@ defmodule Electric.Application do
 
   @process_registry_name Electric.Registry.Processes
 
+  @spec process_name(atom()) :: {:via, atom(), atom()}
+  def process_name(module) when is_atom(module) do
+    {:via, Registry, {@process_registry_name, module}}
+  end
+
   @spec process_name(atom(), term()) :: {:via, atom(), {atom(), term()}}
   def process_name(module, id) when is_atom(module) do
     {:via, Registry, {@process_registry_name, {module, id}}}
@@ -42,8 +47,26 @@ defmodule Electric.Application do
 
       core_processes = [
         {Registry,
-         name: @process_registry_name, keys: :unique, partitions: System.schedulers_online()},
-        {Electric.ShapeCache.ShapeSupervisor, []}
+         name: @process_registry_name, keys: :unique, partitions: System.schedulers_online()}
+      ]
+
+      connection_manager_opts = [
+        connection_opts: Application.fetch_env!(:electric, :connection_opts),
+        replication_opts: [
+          publication_name: publication_name,
+          try_creating_publication?: true,
+          slot_name: slot_name,
+          transaction_received: {Electric.Replication.ShapeLogCollector, :store_transaction, []},
+          relation_received:
+            {Electric.ShapeCache, :handle_relation_msg, [[server: Electric.ShapeCache]]}
+        ],
+        pool_opts: [
+          name: Electric.DbPool,
+          pool_size: Application.fetch_env!(:electric, :db_pool_size),
+          types: PgInterop.Postgrex.Types
+        ],
+        log_collector: {Electric.Replication.ShapeLogCollector, inspector: inspector},
+        shape_cache: shape_cache
       ]
 
       per_env_processes =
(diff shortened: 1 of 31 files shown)


packages/sync-service/lib/electric/application.ex
```diff
@@ -40,8 +45,27 @@ defmodule Electric.Application do
 
       core_processes = [
         {Registry,
-         name: @process_registry_name, keys: :unique, partitions: System.schedulers_online()},
-        {Electric.ShapeCache.ShapeSupervisor, []}
+         name: @process_registry_name, keys: :unique, partitions: System.schedulers_online()}
+      ]
+
+      connection_manager_opts = [
```
magnetised, line 51: because of the way we've structured the postgres connection initialisation, the connection manager is responsible for starting the shape consumer pipeline via `Shapes.Supervisor`. This isn't great, but to restructure we'd have to re-write the connection configuration probing, which feels out of scope

kevin-dp (commented):
Looks good to me!
The GenStage refactoring is above my paygrade though so an Elixir expert should review that. Also, `integration.test.ts` is said to have changed but there are only whitespace differences, this seems to be unintentional.

magnetised:
> Also, `integration.test.ts` is said to have changed but there are only whitespace differences, this seems to be unintentional

Actually that was intentional -- poking around to get the tests to pass and cleaned the ws in passing

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -109,6 +121,12 @@ defmodule Electric.Shapes.Consumer do
     {:noreply, [], %{state | buffer: buffer ++ txns}}
   end
 
+  # Want to be careful with this case as we don't want to ask for more
+  # transactions unless we got one.
+  def handle_events([], _from, state) do
+    {:noreply, [], state}
+  end
```
icehaunter, line 128: I don't understand this. First, is this case even possible? Is `handle_events` ever called with an empty list? Second, wouldn't this put us in a position where we've not asked for more on a fullfilled demand, and thus we'll never recieve more?

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -30,26 +32,81 @@ defmodule Electric.Replication.ShapeLogCollector do
     GenStage.call(server, {:new_txn, txn})
```
icehaunter, line 32: I don't see from the code how, as per PR description, 

> Shape consumers are configured with :manual demand. Along with the BroadcastDispatcher this enables us to make the store_transaction/2 callback to be synchronous: it will only return when all consumers have processed/stored the transaction.

Since this is essentially the core reason for this PR, can you elaborate here and in a comment in code as to how this is acheived?

icehaunter (changes requested):
This is a bit more complicated than I expected, but I've wrapped my head around the logic here apart from one core question. I'll be happy to approve once that's documented. Also got a couple of questions/nitpicks

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: With the `BroadcastDispatcher` handle events is called with an empty list if the selector function has filtered them all out. This guard was an attempt to manage consumer demand in that case, so we don't end up asking for more demand if we haven't received any txns. But that didn't work -- the broadcast dispatcher automatically requests demand for a consumer that has filtered out events -- so this guard made the unit tests pass but the e2e tests fail... Ultimately this was the reason for 1) dropping the shape cache from the genstage pipeline then 2) implementing a dispatcher that works how we need. 

I'll remove this as it's no longer needed

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: I'll explain in a comment

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -186,109 +197,98 @@ defmodule Electric.ShapeCache do
...
 
     recover_shapes(state)
 
-    {:consumer, state,
-     subscribe_to: [{opts.log_producer, selector: &is_struct(&1, Changes.Relation)}]}
+    {:ok, state, {:continue, :start_log_producer}}
+  end
+
+  @impl GenServer
+  def handle_continue(:start_log_producer, state) do
```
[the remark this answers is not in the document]
magnetised: I used a continue to make sure that the current process was subscribed before the producer started emitting events. I'll work on the naming though....

icehaunter (approved):
Great job!

PR #1596, opened 2024-08-29
Next.js SSR example

balegas:
Items example using SSR.

Load the entire shape on the server and send it down to the client, so that the client gets the initial state of the shape on first response and can resume replication from the tip of the shape stream.

First time with nextjs, I still don't know how some things work. I suppose this is good forn an example. I'm happy to receive feedback from nextjs experts :D


examples/nextjs-ssr-example/app/page.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+import { getShapeData } from "./shape"
+import Home from "./Home"
+import { unstable_noStore as noStore } from "next/cache"
+
+const Page = async () => {
+  noStore()
```
KyleAMathews, line 7: why (again) no server rendering?

KyleAMathews (commented):
Left a few comments.

Don't you want to use either https://nextjs.org/docs/pages/building-your-application/data-fetching/get-server-side-props or https://nextjs.org/docs/pages/building-your-application/data-fetching/get-static-props here? Right now I don't see that any data is being cached during server rendering.

EDIT: Oh I see... this is RSC stuff

examples/nextjs-ssr-example/app/Home.tsx
```diff
@@ -0,0 +1,148 @@
...
+      state.concat([{ id: newId }]).forEach((item) => {
+        itemsMap.set(item.id, { ...itemsMap.get(item.id), ...item })
+      })
+      return Array.from(itemsMap.values())
+    }
+
+    return []
+  })
+
+  // Can't render entries on the server because order of
```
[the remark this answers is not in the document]
balegas: If we want to provide an API to render shapes on the server/client, we need to des/serialise the payload into json deterministically. Right now, we use Maps, which don't ensure any order of key iteration.

I thought of ordering the elements. If we're doing the example and moving on, we should just educate what's happening on the example, otherwise we could fix this ourselves and people could use useShape out-of-the box.

Note that we need to patch react-hook anyways for this to work, so we might also solve the ser/deserialization ordering issues

packages/typescript-client/src/client.ts
```diff
@@ -383,9 +383,16 @@ export class ShapeStream {
...
 
     // eslint-disable-next-line no-constant-condition -- we're retrying with a lag until we get a non-500 response or the abort signal is triggered
-    while (true) {
+    while (true && (!signal || !signal.aborted)) {
       try {
-        const result = await this.fetchClient(url.toString(), { signal })
+        // The request are being cached and I can't disable it,
+        // so I'm adding a random query parameter to avoid it.
+        // This needs to be fixed, but all stuff I tried didn't work.
+        url.searchParams.set(`_`, Math.random().toString())
```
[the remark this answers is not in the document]
balegas: I tried that with no great luck. I spent an evening figuring out what was going on. 
When I'm back I'll give another look and raise an issue with them.

examples/nextjs-ssr-example/app/page.tsx
balegas: This turns static rendering on the server off. Was the best working solution I've found.

examples/nextjs-ssr-example/app/Home.tsx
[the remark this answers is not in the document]
KyleAMathews: Shapes in general don't have any order guarantees so we don't need to worry about serialization and deserialization messing with things.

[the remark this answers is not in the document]
balegas: You do, because hydration on the client fails if elements are rendered in a different order from the server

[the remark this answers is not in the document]
KyleAMathews: Ok sure. While browser engines don't guarantee orders of maps they all seem to serialize and deserialize the same.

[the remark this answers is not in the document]
balegas: Doing server rendering of items now.
Ordering elements based on creation time, since serialization order is not guaranteed to hold

packages/typescript-client/src/client.ts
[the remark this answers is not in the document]
balegas: Moved this elsewhere. The problem still persists.

examples/nextjs-ssr-example/app/page.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react"
+import { getSerializedShape } from "@electric-sql/react"
+import Home from "./Home"
+import { unstable_noStore as noStore } from "next/cache"
+
+// Hack to avoid caching behavior in Next.js
+// should work just with {cache: `no-store`}
+const fetchClient: typeof fetch = (...args) => {
```
balegas, line 8: Still need to do this hack, despite nextjs [documentation](https://nextjs.org/docs/app/building-your-application/caching). Maybe someone has any idea

examples/nextjs-ssr-example/app/page.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const itemShape = {
+  url: new URL(`http://localhost:3000/v1/shape/items`).href,
+  fetchClient,
+}
+
+const Page = async () => {
+  noStore()
+
+  return <Home shapes={{ items: getSerializedShape(itemShape) }} />
```
balegas, line 24: When I came back to this example, my naive intention was to use the shapeCache in the React hook to be able to preload a shape and then in the client component, I would just load the shape.

I didn't find any way of passing server-side data to a client component without going through the component properties, which is not transparent. I think this code is quite clean, but I'm not sure if it's possible to do best.

I didn't want to investigate using external storage for that, since I considered that would be out of scope of the example.

thruflo: I wonder how useful this pattern is.

Generally, we can sync data that is available for low latency / edge SSR.

Pushing that data to the client as serialised JSON in the html source seems … a bit old school and bloats the page that you’re trying to accelerate.

With default SSR you have to make sure the data is exactly the same. Unless we can somehow work around that.

I wonder if it’s better to have the server render and then have the client take over once the data is synced into the client. But only if we can make the pattern work?

balegas: The client receives the rendered page (with the items) with the response.
That JSON object is used to hydrate the shape when the component renders in the client, which will allow it to resume shape sync at the position that the server was, instead of starting from scratch.

thruflo:
Great stuff @samwillis :)

On 2, I wonder if there's a way of supporting `order_by` and `limit` of the shape results? An arbitrary query is powerful but means the user needs to format the results in the same way as the shape data. This will become harder as and when we support nested associations. So is there an approach where the user could order/limit/(filter?) the shape when convenient and then drop down to a function when that's not expressive enough?

examples/nextjs-ssr-example/app/page.tsx
balegas: That did the trick! thanks

examples/nextjs-ssr-example/app/ssr-shapes-provider.ts
```diff
@@ -0,0 +1,43 @@
...
+    const shapeOptions = {
+      ...serverOptions,
+      url: clientUrl,
+      offset: data.offset,
+      shapeId: data.shapeId,
+    }
+
+    const shapeData = new Map(Object.entries(data.data ?? new Map()))
+    /* eslint-disable react-hooks/rules-of-hooks */
+    useShape({ ...shapeOptions, shapeData })
```
balegas, line 39: We don't need to call useShape, can refresh the cache directly

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -24,7 +31,14 @@ export async function preloadShape<T extends Row<unknown> = Row>(
 }
 
 export function sortedOptionsHash<T>(options: ShapeStreamOptions<T>): string {
-  return JSON.stringify(options, Object.keys(options).sort())
```
balegas, line 27: I needed this, but I think the sorted shapes can't be for all options of the shape, otherwise we can have multiple shapes depending on specific options for the call.

I think we need `headers` because of authorization

examples/nextjs-ssr-example/app/page.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const serverOptions = {
+  url: new URL(`http://localhost:3000/v1/shape/items`).href,
+}
+
+const Page = async () => {
+  const data = getSerializedShape(serverOptions)
+
+  return (
+    <SSRShapesInitializer serializedShapes={[{ data, serverOptions }]}>
```
balegas, line 14: I think this removes the problem of initializing shapes data  by passing the data on useShape and is more of an explicit pattern than injection

packages/typescript-client/src/shape.ts
```diff
@@ -42,14 +42,15 @@ export type ShapeChangedCallback<T extends Row<unknown> = Row> = (
...
 
-  readonly #data: ShapeData<T> = new Map()
+  readonly #data: ShapeData<T>
   readonly #subscribers = new Map<number, ShapeChangedCallback<T>>()
 
   #hasNotifiedSubscribersUpToDate: boolean = false
   #error: FetchError | false = false
 
-  constructor(stream: ShapeStreamInterface<T>) {
+  constructor(stream: ShapeStreamInterface<T>, shapeData?: ShapeData) {
```
balegas, line 51: I can remove this now (by having a specific method to hydrate shape caches)

examples/nextjs-ssr-example/app/Home.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { Row, ShapeStream, ShapeStreamOptions } from "@electric-sql/client"
+import { useOptimistic } from "react"
+
+const ELECTRIC_URL = process.env.ELECTRIC_URL || `http://localhost:3000`
+
+const parser = {
+  timestamptz: (date: string) => new Date(date).getTime(),
+}
+
+const shapeOptions: () => ShapeStreamOptions<Row> = () => {
```
balegas, line 16: I think this is still not great, we should have a single place to declare shape options, however the urls are different from _outer_ shape options. However, it is not a very bad pattern either thinking this is the shape for the component and is decoupled from the _wrapping_ we did for sending server data

balegas (commented):
@samwillis inspired on your example and [Nextjs docs](https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#using-context-providers), I gave another try. The idea here is to have a context around the component that takes the server-sent data from props and initializes the shapes on the client --- totally transparent on the client component!

 Left some comments

balegas:
> Render a subset of the data on the server,

@samwillis, @thruflo my initial idea was to call the database directly on the server. This is a powerful pattern to show, but because of the problems with ordering and paging, I thought it would be better to build on Shapes on the server too and don't go into those problems. Note that there is a benefit with this that we build on caching again and don't put load on the database.

A naive way of rendering the subset of data on the server is to just drop it :)

balegas:
@samwillis , I did a DX refactoring addressing your concerns with it being simple to specify the shapes across server and client. Have a look at it or ping me to see together.

I also want to replace the base nextjs example with this one (cc @KyleAMathews)

.github/workflows/deploy_examples.yml
```diff
@@ -62,7 +62,7 @@ jobs:
 
 
       - name: Deploy NextJs example
-        working-directory: examples/nextjs-example
+        working-directory: examples/nextjs-ssr-example
```
thruflo, line 65: Just a naming nit: there's no need to suffix with `-example`, that's communicated by the parent folder name, so I would suggest `examples/nextjs-ssr`.

examples/nextjs-ssr-example/README.md
```diff
@@ -0,0 +1,22 @@
+# Next.js SSR example
```
thruflo, line 1: Generally, the README for an example should ideally communicate a bit about what it's trying to show you, i.e.: what is the example, why is it interesting, what can you achieve with the pattern it demonstrates, etc. Plus perhaps some signposts into key parts of the code.

examples/nextjs-ssr-example/public/logo.svg
thruflo, line 1: We're purple nowadays/

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -4,12 +4,19 @@ import {
...
+  Offset,
 } from '@electric-sql/client'
 import React from 'react'
 import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/with-selector.js'
 
 type UnknownShape = Shape<Row<unknown>>
 type UnknownShapeStream = ShapeStream<Row<unknown>>
+export type SerializedShape = {
+  offset: Offset
+  shapeHandle: string | undefined
```
thruflo, line 17: Another very minor nit but seeings how this is in the react-hooks library not just the example, I've been trying to move us towards just using `handle` rather than `shapeHandle`.

Below when you getSerialisedShape you can also just read it from `shape.handle` rather than `shapeStream.shapeHandle`.

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -161,3 +180,22 @@ export function useShape<
...
   return useShapeData()
 }
+export function getSerializedShape(
+  options: ShapeStreamOptions
+): SerializedShape {
+  const shapeStream = getShapeStream(options)
+  const shape = getShape(shapeStream)
+  return {
+    shapeHandle: shapeStream.shapeHandle,
+    offset: shapeStream.offset,
```
thruflo, line 190: Does `offset` work? Is the property not `lastOffset`?

packages/typescript-client/src/client.ts
```diff
@@ -245,6 +245,10 @@ export class ShapeStream<T extends Row<unknown> = Row>
     return this.#shapeHandle
   }
 
+  get offset() {
```
thruflo, line 248: I think this is duplicated now -- we already expose lastOffset as a public getter.

examples/nextjs-ssr-example/README.md
balegas: Will work on that when we decide to publish

packages/react-hooks/src/react-hooks.tsx
balegas: yep, i started this a while back, things like the offset were not public at the time. Will cleanup.

gruckion:
I’ve read over the comments on the front page. Not dug into the files changed yet.

I believe this approach should be using hydration and suspense.



https://react.dev/reference/react/Suspense

examples/nextjs-ssr-example/app/Home.tsx
```diff
@@ -0,0 +1,120 @@
+"use client"
```
gruckion, line 1: In nextJS the file naming conventions are lower kebab case. 

https://nextjs.org/docs/app/api-reference/file-conventions

https://x.com/kentcdodds/status/1361536960876085248?s=46

examples/nextjs-ssr-example/app/Home.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+import { v4 as uuidv4 } from "uuid"
```
gruckion, line 3: Avoid external dependencies. You do not need one for a cryptographically secure uuid anymore.

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

https://x.com/benlesh/status/1857451931427352654?s=46

examples/nextjs-ssr-example/app/Home.tsx
```diff
@@ -0,0 +1,120 @@
...
+  // Post to backend to delete everything
+  const fetchPromise = fetch(`/api/items`, {
+    method: `DELETE`,
+  })
+
+  return await Promise.all([findUpdatePromise, fetchPromise])
+}
+
+export default function Home() {
+  const { data: items } = useShape(getProxiedOptions(options)) as unknown as {
```
gruckion, line 60: As unknown as any? Can generics not be used here?

examples/nextjs-ssr-example/app/Home.tsx
```diff
@@ -0,0 +1,120 @@
...
+      >
+        <input type="hidden" name="new-id" value={uuidv4()} />
+        <button type="submit" className="button" name="intent" value="add">
+          Add
+        </button>
+        <button type="submit" className="button" name="intent" value="clear">
+          Clear
+        </button>
+      </form>
+      {optimisticItems
```
gruckion, line 111: Avoid using sort and map directly in the render response. Use a cost above with useMemo.

Also do not use index as the key.

https://react.dev/learn/rendering-lists#why-does-react-need-keys

examples/nextjs-ssr-example/app/shape-proxy/[...table]/route.ts
```diff
@@ -0,0 +1,44 @@
+export async function GET(request: Request) {
+  const url = new URL(request.url)
+  const originUrl = new URL(
```
Comment (gruckion) [inline examples/nextjs-ssr-example/app/shape-proxy/[...table]/route.ts:3]: Cleaner to just throw if the environment variable is missing or use an escape guard.

examples/nextjs-ssr-example/app/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+  return `${window?.location.origin}/shape-proxy/v1/shape`
+}
+
+export function getProxiedOptions(
+  options: SerializedShapeOptions
+): ShapeStreamOptions {
+  // ensure shape is not syncing on the server
+  const serverOptions: Partial<ShapeStreamOptions> = {}
+  if (typeof window === `undefined`) {
+    const controller = new AbortController()
```
gruckion, line 23: Using fetch and AbortControllers adds a lot of bloat. Maybe use useQuery which has solved this problems already and provides a beautiful API.

Tanstack react query

gruckion (changes requested):
Nice work 👍

balegas:
We're probably not going to work on this anymore

balegas:
It does support SSR. You can come up with your own tailored SSR implementation.
#2232 is a follow up of this PR that has more convenient abstractions.

We're happy to help someone finish this work.

PR #1601, opened 2024-09-01
fix: useShape re-renders on where state change

tyzoo:
The `useShape` hook is behaving strangely when switching between different `where` conditions. I noticed that the `const latestShapeData = useRef(parseShapeData(shape))` was not updating when the underlying `where` parameters changed, while the other values within the useShape hook like `shape`, `shapeStream`, etc. were updating correctly. I propose adding the `shapeHash` in the `UseShapeResult`, as well as as a second argument to `parseShapeData`, so when the `useShape` is rendered we can determine wether we need to reset the `UseShapeResult`/`parseShapeData` reference.

nice work on this project all!

### DIFF
--- a/.changeset/seven-forks-happen.md
+++ b/.changeset/seven-forks-happen.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/react": patch
+---
+
+add test that useShape re-renders on where state change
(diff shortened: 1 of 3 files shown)


Makisuo:
Nice just wanted to start working on it and saw you already did, good stuff!

KyleAMathews (approved):
Thanks for the bug report and working on a fix @tyzoo and @Makisuo !

PR #1604, opened 2024-09-02
feat!: add mixed storage

icehaunter:
This introduces a new storage backend, that's build as a plain file for snapshots, and CubDB for any metadata and ongoing logs. The benefit here is that we don't have to close the file while processing the results coming from the PostgreSQL query, unlike with batched CubDB writes. The result is that we're writing the snapshot roughly at the same rate as PostgreSQL generates query results.

This storage greatly improves initial snapshot performance (about triples the speed of snapshot creation), as well as RAM usage of Electric (benchmark made with 15s connection checkout timeout, hence the cutoff):
image

### DIFF
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -72,6 +72,7 @@ cache_stale_age = env!("CACHE_STALE_AGE", :integer, 60 * 5)
 statsd_host = env!("STATSD_HOST", :string?, nil)
 
 cubdb_file_path = env!("CUBDB_FILE_PATH", :string, "./shapes")
+mixed_file_path = env!("MIXED_STORAGE_DIR", :string, "./shapes_mixed")
 persistent_state_path = env!("PERSISTENT_STATE_FILE_PATH", :string, "./state")
 
 persistent_kv =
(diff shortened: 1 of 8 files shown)


packages/sync-service/config/runtime.exs
```diff
@@ -72,6 +72,7 @@ cache_stale_age = env!("CACHE_STALE_AGE", :integer, 60 * 5)
 statsd_host = env!("STATSD_HOST", :string?, nil)
 
 cubdb_file_path = env!("CUBDB_FILE_PATH", :string, "./shapes")
+mixed_file_path = env!("MIXED_STORAGE_DIR", :string, "./shapes_mixed")
```
robacourt, line 75: Please could we have all our dirs we write to under one dir? I already lost a load of time not realising you had to delete the `state` dir as well as the `shapes` dir. Maybe we could put all 3 dirs in a `metadata` dir so you can just delete the metadata dir if you want to start afresh?

robacourt (approved):
Neat!

robacourt:
What's the advantage of keeping the non-mixed cub-db storage? It might be better to delete it and remove the cost of maintaining it

packages/sync-service/config/runtime.exs
icehaunter: I'll make that a separate PR

PR #1605, opened 2024-09-02
feat(sync-service): Remove snapshot timeout

robacourt:


packages/sync-service/lib/electric/shapes/snapshotter.ex
```diff
@@ -71,31 +71,31 @@ defmodule Electric.Shapes.Snapshotter do
...
-        # that way it has the relation, but it is still missing the pk_cols
-        Storage.make_new_snapshot!(shape_id, stream, storage)
-      end)
-    end)
+          # could pass the shape and then make_new_snapshot! can pass it to row_to_snapshot_item
+          # that way it has the relation, but it is still missing the pk_cols
+          Storage.make_new_snapshot!(shape_id, stream, storage)
+        end)
+      end,
+      timeout: :infinity
```
robacourt, line 98: This line is the only change `timeout: :infinity`. The rest is formatting

robacourt, line 98: This line is the only change: `timeout: :infinity`. The rest is formatting

magnetised (approved):
Love it. In my version, I also ended up adding `ownership_timeout: :infinity` because I think the pool started claiming back its connections even if the query wasn't timing out. Have you seen that or was I just unlucky?

robacourt:
> Love it. In my version, I also ended up adding `ownership_timeout: :infinity` because I think the pool started claiming back its connections even if the query wasn't timing out. Have you seen that or was I just unlucky?

I tested it as is and it worked fine. Also, @alco commented in #1499 that "ownership_timeout is specific to the [DBConnection.Ownership](https://hexdocs.pm/db_connection/DBConnection.Ownership.html) pool which one can use in place of the default [DBConnection.ConnectionPool](https://hexdocs.pm/db_connection/DBConnection.ConnectionPool.html)." and I believe we're not using `Ownership` so it shouldn't make a difference.

magnetised:
> I tested it as is and it worked fine. Also, @alco commented in #1499 that "ownership_timeout is specific to the [DBConnection.Ownership](https://hexdocs.pm/db_connection/DBConnection.Ownership.html) pool which one can use in place of the default [DBConnection.ConnectionPool](https://hexdocs.pm/db_connection/DBConnection.ConnectionPool.html)." and I believe we're not using `Ownership` so it shouldn't make a difference.

Gotcha. Should have read that comment... 😊

Issue #1581, opened 2024-08-27
Return larger shape logs in chunks


KyleAMathews:
It's not a good idea to return huge shape logs in one request as:
- many HTTP proxies won't cache large shapes (e.g. 20mb+)
- clients can't process the shape log until they finish downloading it — e.g. if a client is inserting data into pglite, it'd be faster if they could parallelize downloading/parsing data with inserting rows into pglite. The same is true of any integration.

Ideal chunk size is around ~10mb.

The API was designed for this as clients continue requesting data until they the server reports they're at the end of the log so there'd be no breaking changes required for this.

msfstef:
@kevin-dp and I are workshopping this - will keep the thread updated

Issue #1534, opened 2024-08-16
example: 100mb of analytics data in pglite 


KyleAMathews:
narrative:
- loads relatively fast
- client-side interactivity is very fast
- works offline
- loads very fast with cached data/code

msfstef:
Would this be like a table/graph frontend? How would the story go other than "useShape to sync data in, the rest is all SQL/Postgres in the client"?

KyleAMathews:
Yeah some sort of table graph frontend where the user can change some options or whatever so triggers a new client-side query/render. The goal is to show a) 100mb is actually a reasonable amount of data to load and b) client-side interactivity is still really fast.

But the UI itself is whatever strikes your fancy.

msfstef:
@KyleAMathews as previously discussed, should we morph this into a more "live streaming HackerNews data using Electric and PGlite?" rather than a plain one-time analytics workload?

KyleAMathews:
I like the idea of something live streaming but the main point here is to show that the sync engine & pglite can easily load large amounts of data. Live streaming is nice but not the point.

A HN clone doesn't necessarily show that but it could e.g. if you include some charts about posts / month over the past 10 years or whatever that shows off an aggregation query. So yeah an aggregation query + full-text search could be fun (though perhaps creating the full-text index would be excessively slow?).

What do you think?

msfstef:
Hmm if the main point is still the loading of large amounts of data then I think it's better to stick with the thing I've already been building (large fixed dataset) and do something with that - probably no point in reworking it for a different data source. We can think of different examples to highlight the live streaming analytics use case.

KyleAMathews:
👍 yeah, good to keep examples focused

balegas:
It's easy to find large dataset from HN and GH which would allow to do some nice demos with loads of data that interests the community. The problem with these dataset is that they are all a bit old and therefore there is a gap between the collected data and the data that we would load live.

If we're looking into synching large amounts of data I'd still look into one very large dataset. In the order of GB (or the max the browser allows).

As discussed, what's special is the ability to also get live data for your analytics, I'll create a separate issue for that.

balegas:
https://github.com/electric-sql/electric/issues/1745

KyleAMathews:
This is done enough for now — https://linearlite.examples.electric-sql.com/ loads 150mb of data into pglite so good demo

PR #1606, opened 2024-09-02
feat: Log chunking

msfstef:
Addresses https://github.com/electric-sql/electric/issues/1581

Will outline work done tomorrow, need to add some more tests as well - also I have deviated from the original design due to [this thread](https://electric-sql.slab.com/posts/rfc-chunking-the-log-2zee9qad#thread-xqaece2t).

I have externalised the chunk index from within the log itself but have kept it in storage - this might have a non-trivial impact on performance due to the additional IO involved in 1) storing the checkpoints and 2) reading the checkpoints on every request.

The benefits of externalising the index is that it makes it queriable, so we can still immediately know the end offset of a chunk before actually streaming it, which allows parallelisation of requests etc.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
@@ -37,6 +37,9 @@ defmodule Electric.ShapeCache.InMemoryStorage do
   def start_link(compiled_opts) do
     if is_nil(compiled_opts.shape_id), do: raise("cannot start an un-attached storage instance")
 
+    {chunking_module, chunking_opts} = Access.fetch!(compiled_opts, :log_chunking)
+    apply(chunking_module, :start_link, [chunking_opts])
+
     Agent.start_link(
       fn ->
         %{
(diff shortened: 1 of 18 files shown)


KyleAMathews:
It could be neat to build https://github.com/electric-sql/electric/issues/1534 as part of this PR to check what the browser experience is like.

packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
```diff
@@ -14,12 +16,21 @@ defmodule Electric.ShapeCache.InMemoryStorage do
     snapshot_ets_table_name = Access.get(opts, :snapshot_ets_table, :snapshot_ets_table)
     log_ets_table_name = Access.get(opts, :log_ets_table, :log_ets_table)
 
+    chunk_checkpoint_ets_table_name =
```
robacourt, line 19: There shouldn't be an ETS table now, right?

packages/sync-service/test/electric/shape_cache/log_chunker.exs
```diff
@@ -0,0 +1,49 @@
...
+      assert {:threshold_exceeded, 0} =
+               LogChunker.add_to_chunk(chunk_bytes, 2 * chunk_byte_size, threshold)
+    end
+
+    test "should ignore zero length bytestrings", _ do
+      threshold = 10
+      just_below_threshold = threshold - 1
+      # large_string = String.duplicate("a", threshold)
+
+      # assert :ok = LogChunker.add_to_chunk(large_string, threshold)
```
robacourt, line 28: Why's this commented out?

robacourt (changes requested):
I think you've missed the InMemoryStorage, it still has an ETS table. But otherwise it looks great! Good job! 🚀

packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
robacourt: Ahh, of course. Make sense.

robacourt (approved):
Approved! Nice work @msfstef !

PR #1610, opened 2024-09-03
chore (react-hooks): improve type of selector argument in useShape hook

kevin-dp:
The default value for the `selector` argument used to be typed as `never` which is wrong. This PR improves the type of the default value.

### DIFF
--- a/.changeset/lemon-eggs-care.md
+++ b/.changeset/lemon-eggs-care.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/react": patch
+---
+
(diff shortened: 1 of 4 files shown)


msfstef (approved):
Nice - might be worth adding a type test if you want!

kevin-dp:
> Nice - might be worth adding a type test if you want!

Not sure what you want to type test. This is type casting the default argument to the expected type. It's typing custom `selector` functions provided by users.

msfstef (approved):
Nice!

Issue #1597, opened 2024-08-30
Typescript Client throws on nullable arrays


TWith2Sugars:
[This line here](https://github.com/electric-sql/electric/blob/97a9f5e6a1d2459bd0deca97eb8284ebe2c12d7f/packages/typescript-client/src/parser.ts#L39)

When trying to parse this payload:

```json
[
    {
        "offset": "0_0",
        "value": {
            "attributes": null,
            "created_by_id": "01bf5e02-814e-4d80-b533-d7e271dce504",
            "created_on": "2024-08-25 10:49:47.055251",
            "deleted_by_id": null,
            "deleted_on": null,
            "description": "Dev Production Unit Description 1",
            "labels": null,
            "modified_by_id": "01bf5e02-814e-4d80-b533-d7e271dce504",
            "modified_on": "2024-08-25 10:49:47.055251",
            "name": "Production Unit 1",
            "production_unit_id": "726d2f05-df6e-4be6-b754-27d46845c9d5",
            "project_id": "00000000-0000-0000-0000-000000000000"
        },
        "key": "\"public\".\"production_units\"/\"726d2f05-df6e-4be6-b754-27d46845c9d5\"",
        "headers": {
            "relation": [
                "public",
                "production_units"
            ],
            "operation": "insert"
        }
    },
    {
        "offset": "0_0",
        "value": {
            "attributes": null,
            "created_by_id": "e17290b7-c7b7-4444-bf22-86414f144185",
            "created_on": "2024-08-25 10:49:47.055251",
            "deleted_by_id": null,
            "deleted_on": null,
            "description": "Dev Production Unit Description 2",
            "labels": null,
            "modified_by_id": "e17290b7-c7b7-4444-bf22-86414f144185",
            "modified_on": "2024-08-25 10:49:47.055251",
            "name": "Production Unit 2",
            "production_unit_id": "d8fc4695-8b6b-4f48-a665-6bd7572767b2",
            "project_id": "00000000-0000-0000-0000-000000000000"
        },
        "key": "\"public\".\"production_units\"/\"d8fc4695-8b6b-4f48-a665-6bd7572767b2\"",
        "headers": {
            "relation": [
                "public",
                "production_units"
            ],
            "operation": "insert"
        }
    },
    {
        "headers": {
            "control": "up-to-date"
        }
    }
]
```

```
HTTP/1.1 200 OK
X-Powered-By: Express
Access-Control-Allow-Origin: http://localhost:5173
Vary: Origin
[...]

PR #1611, opened 2024-09-03
fix: Nullable arrays not being parsed as `null`

msfstef:
Addresses https://github.com/electric-sql/electric/issues/1597

I also made the types include the possibility of `null` being a value that needs to be parsed and changed the parsers accordingly.

I don't love that currently if a field is not nullable and `null` is received it tries to force it to a field (like `null` to `0` if it's a number), and this logic definitely does not work on arrays marked as `NOT NULL` that are `null`. Are we against throwing for those cases?

### DIFF
--- a/packages/typescript-client/src/parser.ts
+++ b/packages/typescript-client/src/parser.ts
@@ -3,8 +3,12 @@ import { ColumnInfo, Message, Schema, Value } from './types'
 type NullToken = null | `NULL`
 type Token = Exclude<string, NullToken>
 type NullableToken = Token | NullToken
-export type ParseFunction<T extends Token | NullableToken = Token> = (
-  value: T,
+export type ParseFunction = (
+  value: Token,
+  additionalInfo?: Omit<ColumnInfo, `type` | `dims`>
+) => Value
+type NullableParseFunction = (
+  value: NullableToken,
   additionalInfo?: Omit<ColumnInfo, `type` | `dims`>
 ) => Value
 export type Parser = { [key: string]: ParseFunction }
@@ -136,7 +140,7 @@ function makeNullableParser(
   parser: ParseFunction,
   columnInfo: ColumnInfo,
   columnName?: string
-): ParseFunction<NullableToken> {
+): NullableParseFunction {
   const isNullable = !(columnInfo.not_null ?? false)
   // The sync service contains `null` value for a column whose value is NULL
   // but if the column value is an array that contains a NULL value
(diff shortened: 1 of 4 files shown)


kevin-dp:
> if a field is not nullable and null is received

This feels like it would be a bug in Electric.
Is this possible? How can this happen?

msfstef:
@kevin-dp it should not be possible, which is why I'm wondering if we should be throwing instead of forcing it to a non-null value

kevin-dp:
> @kevin-dp it should not be possible, which is why I'm wondering if we should be throwing instead of forcing it to a non-null value

Yep, i'd much rather throw an error saying we got an unexpected null value from Electric for a non-nullable column rather than forcing it into the right type which is just very likely to lead to problems or unexpected behavior somewhere else.

kevin-dp (changes requested):
I've added some comments and requested some changes because i don't think we need to handle `null` values everywhere as is done currently. We should be able to limit handling of `null` values to only the `makeNullableParser` function.

packages/typescript-client/src/parser.ts
```diff
@@ -1,7 +1,10 @@
 import { ColumnInfo, Message, Schema, Value } from './types'
 
-export type ParseFunction = (
-  value: string,
+type NullToken = null | `NULL`
+type Token = Exclude<string, NullToken>
+type NullableToken = Token | NullToken
+export type ParseFunction<T extends Token | NullableToken = Token> = (
```
kevin-dp, line 6: `Token | NullableToken` is redundant since `Token` is already part of `NullableToken`, so should just be `NullableToken`.

kevin-dp: I'm not a fan of using generics for representing two versions of this function type, it complicates the understanding a bit.
imo it's easier to understand if we define them as two separate types:
```ts
export type ParseFunction = (value: Token, additionalInfo?: Omit<ColumnInfo, `type` | `dims`>) => Value
export type NullableParseFunction = (value: NullableToken, additionalInfo?: Omit<ColumnInfo, `type` | `dims`>) => Value
```

kevin-dp (commented):
Good work!
I like how you improved it now but we should add a unit test that checks that it now correctly parses `null` values on nullable array columns and perhaps a second test that checks that the parser complains if you try to parse a null value on a non-nullable column.

kevin-dp (approved):

PR #1612, opened 2024-09-03
chore (docs): fix React integration snippet

kevin-dp:
The snippet on the React integration page was renaming `data` to `fooData` but then using `data` which does not exist. Removed the renaming.


KyleAMathews (approved):

thruflo (approved):
👍

Issue #1466, opened 2024-07-16
Let users set the type of data in `ShapeStream`, `Shape`, `useShape`


KyleAMathews:
Handwritten types for now but ideally we can generate a types file for them in the future from PG's schema.

Context: https://github.com/electric-sql/electric-next/pull/22#discussion_r1678987845 cc @kevin-dp

PR #1613, opened 2024-09-03
feat: Allow client type templating

msfstef:
Addresses https://github.com/electric-sql/electric/issues/1466

Work on this was already done for `useShape` by @kevin-dp - this extends it to all of the client APIs.


KyleAMathews (approved):
Looks great! Excited to have this. Perhaps add typing to an example or two to demo this? Also is there anywhere in the docs we could/should mention this?

packages/typescript-client/src/types.ts
```diff
@@ -25,7 +27,7 @@ export type ChangeMessage<T> = {
 }
 
 // Define the type for a record
-export type Message<T extends Value = { [key: string]: Value }> =
+export type Message<T extends Row = { [key: string]: Value }> =
```
kevin-dp, line 30: Can simplify to `export type Message <T extends Row = Row>`

packages/typescript-client/src/types.ts
```diff
@@ -25,7 +27,7 @@ export type ChangeMessage<T> = {
...
   | ControlMessage
   | ChangeMessage<T>
```
kevin-dp, line 32: We may also want to add the subclass requirement on `export type ChangeMessage<T extends Row>`

packages/typescript-client/src/types.ts
```diff
@@ -104,7 +106,7 @@ export type ColumnInfo =
 
 export type Schema = { [key: string]: ColumnInfo }
 
-export type TypedMessages<T extends Value = { [key: string]: Value }> = {
+export type TypedMessages<T extends Row = { [key: string]: Value }> = {
```
kevin-dp, line 109: `T extends Row = Row`

packages/typescript-client/src/helpers.ts
```diff
@@ -17,7 +17,7 @@ import { ChangeMessage, ControlMessage, Message, Value } from './types'
  * }
  * ```
  */
-export function isChangeMessage<T extends Value = { [key: string]: Value }>(
+export function isChangeMessage<T extends Row = { [key: string]: Value }>(
```
kevin-dp, line 20: `T extends Row = Row`

packages/typescript-client/src/helpers.ts
```diff
@@ -40,7 +40,7 @@ export function isChangeMessage<T extends Value = { [key: string]: Value }>(
  * }
  * ```
  */
-export function isControlMessage<T extends Value = { [key: string]: Value }>(
+export function isControlMessage<T extends Row = { [key: string]: Value }>(
```
kevin-dp, line 43: `T extends Row = Row`

packages/typescript-client/test/integration.test.ts
```diff
@@ -637,7 +637,7 @@ describe(`HTTP Sync`, () => {
       return response
     }
 
-    const issueStream = new ShapeStream({
+    const issueStream = new ShapeStream<IssueRow>({
```
kevin-dp, line 640: Since we added the `IssueRow` type argument here, i think we can remove it from L651 as `forEachMessage` should be able to infer it from the `issueStream` argument that is passed to it.

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -40,13 +42,15 @@ export function getShapeStream(options: ShapeStreamOptions): ShapeStream {
   }
 }
 
-export function getShape(shapeStream: ShapeStream): Shape {
+export function getShape<T extends Row = Row>(
```
kevin-dp, line 45: I don't think we want to provide a default type argument here because the type argument can be inferred from the `shapeStream` argument.
So we want that `getShape(myShapeStream)` uses the type from `myShapeStream` but because we define the default type argument to be `Row` it would use `Row` (i think).

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -55,33 +59,38 @@ export function getShape(shapeStream: ShapeStream): Shape {
...
+  error: Shape<T>[`error`]
   isError: boolean
   /**
    * Has the ShapeStream caught up with the replication log from Postgres.
    */
   isUpToDate: boolean
 }
 
-function shapeSubscribe(shape: Shape, callback: () => void) {
+function shapeSubscribe<T extends Row = Row>(
```
kevin-dp, line 81: Same here, i would avoid the default type argument as it can be inferred from the `shape` argument.

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -55,33 +59,38 @@ export function getShape(shapeStream: ShapeStream): Shape {
...
+  callback: () => void
+) {
   const unsubscribe = shape.subscribe(callback)
   return () => {
     unsubscribe()
   }
 }
 
-function parseShapeData(shape: Shape): UseShapeResult {
+function parseShapeData<T extends Row = Row>(
```
kevin-dp, line 91: same about default type arg

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -95,16 +104,20 @@ function identity<T>(arg: T): T {
...
-interface UseShapeOptions<Selection> extends ShapeStreamOptions {
-  selector?: (value: UseShapeResult) => Selection
+interface UseShapeOptions<Selection, SourceData extends Row = Row>
+  extends ShapeStreamOptions {
+  selector?: (value: UseShapeResult<SourceData>) => Selection
 }
 
-export function useShape<Selection = UseShapeResult>({
+export function useShape<
+  Selection = UseShapeResult<Row>,
```
kevin-dp, line 113: same here about default type args

kevin-dp (approved):
Nice type improvements!
Left some minor comments, mainly the same comment but added it everywhere it needs that change.

packages/react-hooks/src/react-hooks.tsx
msfstef: I've added a type test for this scenario and the `Shape` infers the type from the `ShapeStream` - even if it has a default, since the argument is a `ShapeStream<T>` it will use that to infer.

kevin-dp: That's nice, so inferred type takes preference over default type, perfect!

packages/react-hooks/src/react-hooks.tsx
msfstef: I'm going to switch these two around as I want the `Selection` to be `UseShapeResult<SourceData>` by default - but I also want the `SourceData` to be defaulted to `Row` cause `useShape` does not have any parameters that will allow it to infer the type, it needs to be explicitly typed.

packages/react-hooks/src/react-hooks.tsx
kevin-dp: Still, it's always going to infer the type argument from the actual argument, so there's no need to provide a default, is there?

msfstef: That's fair - I'll remove it

packages/react-hooks/src/react-hooks.tsx
msfstef: (ended up not switching them around cause it feels more ergonomic to be able to _only_ specify the selection type)

msfstef:
@kevin-dp  addressed all your comments!

@KyleAMathews added typings to all examples (missed one reference which usees Dates, I'll have to think about that one separately as we currently don't parse dates into JS dates) - I've also added a type template in the TS client docs which I think should be enough?

docs/api/clients/typescript.md
```diff
@@ -32,7 +32,7 @@ The client exports a `ShapeStream` class for getting updates to shapes on a row-
 import { ShapeStream } from '@electric-sql/client'
 
 // Passes subscribers rows as they're inserted, updated, or deleted
-const stream = new ShapeStream({
+const stream = new ShapeStream<RowType>({
```
KyleAMathews, line 35: I wonder if this will be clear enough to people? They might think it's missing a type import and not realize it's something they need to write themselves. Perhaps just leave it out for now and the examples are good enough documentation for now?

examples/nextjs-example/app/page.tsx
```diff
@@ -58,7 +58,7 @@ async function clearItems() {
 }
 
 export default function Home() {
-  const { data: items } = useShape(itemShape()) as unknown as { data: Item[] }
+  const { data: items } = useShape<Item>(itemShape())
```
KyleAMathews, line 61: SOOO GOOOD!

KyleAMathews:
> added typings to all examples (missed one reference which usees Dates, I'll have to think about that one separately as we currently don't parse dates into JS dates)

Add a custom parse for the date?

docs/api/clients/typescript.md
kevin-dp: JavaScript devs don't need to worry about this. TypeScript devs should be familiar with generics, so shouldn't be a problem?

kevin-dp: I agree with @KyleAMathews here that it isn't clear what `RowType` is/should be.
Perhaps we can give an example or explain it in the text.

msfstef:
@KyleAMathews re: dates - might pick it up in separate PR

PR #1616, opened 2024-09-03
feat: move row-to-JSON serialization for initial shape data to PG

icehaunter:
Improves performance heavily by moving row-to-JSON initial snapshot logic into PG. Currently utilizes hand-crafted JSON structure, but I'll make another test later if we can use a cleaner JSON building approach. 

Currently, if used with "mixed storage pr", it gives good improvements:
image

### DIFF
--- a/.changeset/ninety-snakes-rest.md
+++ b/.changeset/ninety-snakes-rest.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Move row-to-JSON serialization for initial shape data from Electric to PG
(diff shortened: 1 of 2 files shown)


robacourt (approved):
Nice! I like how you've refactored it since I last it, much easier to read.

balegas:
Did we measure the performance penalty on PG? It's okay to move the cost to PG, but does it make orders of magnitude slower? Does it produce a memory usage spike? Does it fall of for large queries?

KyleAMathews:
I suspect it's fine as postgres is already manipulating query results to return them so it's just a bit more fiddling with the data.

Issue #1468, opened 2024-07-15
Expose from client when it's not up-to-date with server


KyleAMathews:
This could be due to temporary network/server issues or if the client is coming back from being offline and hasn't yet downloaded & applied all the updates.

- [ ] `isUpToDate` should be set to false
- [ ] expose variable with the number of ms since the last successful sync

kevin-dp:
Nice idea to expose a variable that is the time elapsed since last sync!

Regarding `isUpToDate`, i'm not sure about setting it to false.
Internally, it's used to determine whether to use live mode or not. If it turns out we're not up to date, we will get the part of the log we are missing and will then go back into live mode on the latest offset. So internally, i don't think there's a need to put `isUpToDate` to false.

However, we also expose `isUpToDate` externally, which may lead users to think they are up to date while it may be that they were up to date 1 hour ago and that we failed to sync since then. But it's hard to tell when `isUpToDate` should become false... because there's no way to know for sure other than contacting Electric.

The point i'm trying to make is that if we pessimistically put `isUpToDate` to false because the last N fetch requests failed, that doesn't necessarily mean we are out of date, we wil only know when we manage to reach Electric again. So maybe, we want an internal `_isUpToDate` that we keep as it is now, and have an external `isUpToDate` that is true if we managed to sync recently, e.g. if `timeSinceLastSync < UP_TO_DATE_THRESHOLD`. One could also argue that exposing the time elapsed since last sync is enough for this purpose and clients can decide for themselves whether they are up to date or not.

msfstef:
I like the idea of an `UP_TO_DATE_THRESHOLD`, perhaps a configurable one with a sensible default?

KyleAMathews:
> But it's hard to tell when isUpToDate should become false...

Wouldn't it just be anytime there's a fetch error? https://github.com/electric-sql/electric/blob/b322b957a1010c6450175001b5519b9567558fbc/packages/typescript-client/src/client.ts#L390-L412

But yeah, I see you're point. Maybe `isUpToDate` is the wrong variable name for what I'm thinking. Basically what I want is what we had in the old electric where you could see if you were connected to Electric or not. There's not an exact analogy to http long-polling as we don't have a persistent connection but `isConnected` could be true once we've successfully fetched data and get switched to false as soon as there's any fetch errors.

> UP_TO_DATE_THRESHOLD

NO OPTIONS! 😛

kevin-dp:
@KyleAMathews i see!
[...]

PR #1625, opened 2024-09-04
feat (client): expose time elapsed since last sync

kevin-dp:
This PR fixes https://github.com/electric-sql/electric/issues/1468 as it exposes a `lastSynced` getter on the `ShapeStream` and `Shape` classes which is the number of milliseconds since last sync with the server.

### DIFF
--- a/.changeset/brown-turtles-listen.md
+++ b/.changeset/brown-turtles-listen.md
@@ -1,5 +1,0 @@
----
-"@electric-sql/client": patch
----
-
-Expose a `lastSynced` field on the `ShapeStream` and `Shape` classes which is the time elapsed since the last sync with Electric (in milliseconds).
--- a/.changeset/brown-turtles-listen.md
+++ b/.changeset/brown-turtles-listen.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
+Expose a `lastSynced` field on the `ShapeStream` and `Shape` classes which is the time elapsed since the last sync with Electric (in milliseconds).
--- a/.changeset/brown-turtles-listen.md
+++ b/.changeset/brown-turtles-listen.md
@@ -1,5 +1,6 @@
 ---
-"@electric-sql/client": patch
+"@electric-sql/client": minor
 ---
 
 Expose a `lastSyncedAt` field on the `ShapeStream` and `Shape` classes which is the time elapsed since the last sync with Electric (in milliseconds).
+Remove the `isUpToDate` field on the `Shape` class.
--- a/.changeset/brown-turtles-listen.md
+++ b/.changeset/brown-turtles-listen.md
@@ -2,4 +2,4 @@
 "@electric-sql/client": patch
 ---
 
-Expose a `lastSynced` field on the `ShapeStream` and `Shape` classes which is the time elapsed since the last sync with Electric (in milliseconds).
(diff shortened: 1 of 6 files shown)


icehaunter (approved):
LGTM, but add a changeset pls

kevin-dp:
@icehaunter Thanks for the quick review!

Btw, this PR does not fully address the linked issue because @KyleAMathews was proposing to also put `isUpToDate` to false. But i'm uncertain about that and left [some comments](https://github.com/electric-sql/electric/issues/1468#issuecomment-2328477067) on the issue.

msfstef (approved):
Looks good! I see you've left a comment regarding the `isUpToDate` so will continue from there!

packages/typescript-client/src/client.ts
```diff
@@ -390,9 +414,12 @@ export class ShapeStream<T extends Row = Row> {
     while (true) {
       try {
         const result = await this.fetchClient(url.toString(), { signal })
-        if (result.ok) return result
-        else throw await FetchError.fromResponse(result, url.toString())
+        if (result.ok) {
+          this.connected = true
```
KyleAMathews, line 418: we also need to check if the user is subscribed — if not then they're never "connected"

KyleAMathews (changes requested):

kevin-dp:
@KyleAMathews i re-introduced the `isUpToDate` getter on the `Shape` class because it is being used by the react hooks which expose an `isUpToDate` field on the shape data: https://github.com/electric-sql/electric/blob/main/packages/react-hooks/src/react-hooks.tsx#L89

I'm not sure what we want to do with that. It seems that no example is currently using that `isUpToDate` field on the data and the name is misleading so i would drop it altogether. wdyt?

packages/typescript-client/src/client.ts
kevin-dp: Why do we want to tie them together?
Connectivity status is not affected by subscriptions imo.
Note the `try.. finally` block i introduced around the `while` loop which ensures that when we stop fetching the shape, the `connected` flag is put to false. So if a user creates a shapestream with `subscribe: false` the flag will only be true for a very brief period of time, between L418 and the `finally` block.

KyleAMathews: `connected` means you're subscribed and currently getting updates. It just doesn't mean anything if you're doing a one-time fetch. I guess it doesn't really matter though as someone wouldn't use `connected` anywhere in the case they're doing a one-off fetch.

KyleAMathews:
>  It seems that no example is currently using that isUpToDate field on the data and the name is misleading so i would drop it altogether. wdyt?

Yeah... it's really just an internal state. For consumers of the Shape, the shape is *always* up-to-date. If we're getting new data, we don't tell them about it until we've caught up with the Log. So not exposing it makes sense to me.

packages/typescript-client/src/client.ts
kevin-dp: Ok, i added the extra check.

.changeset/brown-turtles-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@electric-sql/client": patch
```
KyleAMathews, line 2: Technically this is a breaking change now so let's make this a minor (and update it to tell about the field we're removing).

KyleAMathews (approved):
Great dx improvement, thanks!

PR #1626, opened 2024-09-04
chore: Externalize log chunking from storage to `ShapeConsumer`

msfstef:
Based on feedback from @icehaunter - moving the chunking to the `ShapeConsumer` to reduce duplication and limit the storage layer's responsibility (as well as splitting IO from processing/preparing).

The JSON serializing and chunking of the log is now handled by the `ShapeConsumer` (see `prepare_log_entries` method), so storage only needs to distinguish between actual log items and chunk boundaries. That's perhaps the main coupling point/interface change that is noteworthy.

I've also moved the `chunk_byte_threshold` config outside of storage as it is no longer its responsibility, which meant threading it through a bunch of places. I'm happy to change that if there's a better pattern to do that.

### DIFF
--- a/.changeset/lemon-countries-own.md
+++ b/.changeset/lemon-countries-own.md
@@ -1,5 +1,0 @@
----
-"@core/sync-service": patch
----
-
-Move log chunking to `ShapeConsumer` to reduce the storage layer's responsibility.
--- a/.changeset/lemon-countries-own.md
+++ b/.changeset/lemon-countries-own.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Move log chunking to `ShapeConsumer` to reduce the storage layer's responsibility.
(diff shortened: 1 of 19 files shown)


icehaunter (approved):
LGTM, nice job! I think it reads better without essentially same map_reduce in each storage impl

.changeset/lemon-countries-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Move log chunking to `ShapeConsumer` to reduce the storage layer's responsibility.
```
robacourt, line 5: This is a refactoring, right? I was under the impression refactorings (i.e. ones that don't change functionality/performace) don't need a changeset

robacourt (approved):
Oh yeah, that's much nicer. Great work @msfstef and good call @icehaunter

.changeset/lemon-countries-own.md
msfstef: I wasn't sure what our policy is technically if we haven't released the log chunking yet then this is still under the log chunking changeset I suppose so I can remove it.

Issue #1519, opened 2024-08-13
`ShapeStream` should persist the Shape Log in the client for faster loading of shapes (and offline loading)


KyleAMathews:
The Typescript `ShapeStream` class is responsible for reading the Shape Log from the Electric server and feeding it to other code that'd like to do something with the stream e.g. the `Shape` class which creates an in-memory materialized view of the Shape as a JS Map.

Shape Logs are cached in Electric (and in an http cache proxy (e.g. nginx/cdn/etc) if you're using one) so they're generally cheap to load. But still, if shapes are larger than a few mbs, persisting shape logs to disk would speed up loading — especially for people on limited mobile networks.

## The implementation for this would be:

On Initializing:
1. `ShapeStream` checks if there's a persisted log. If so it loads that, the `offset`, and `shapeId` for the log.
2. Then it makes a fetch to the server like normal to catchup with the shape log on the server

On receiving new shape log messages: 
1. `ShapeStream` would emit these to subscribers and then append them to the persisted shape log.

Compaction — Shape Logs fill up eventually with a lot of redundant information (e.g. 100s of updates to a single row). So periodically compacting the log is necessary to keep reading the log fast. Compaction is basically reading through the log from the start and merging together repeated operations on a single row. When a row is deleted, we can remove all log messages. When compaction is finished, the log is left with one `insert` operation / row in the shape.

## Considerations
1. benchmark this — sometimes client storage can be slow, particularly on mobile. Sometimes reading from the network is faster than reading from a local cache. We'd need to benchmark carefully (on desktop and mobile) the perf of persisting vs. not. Perhaps it is always faster in which case persisting can be the default. Or perhaps it's always faster on desktop so could be enabled there. Regardless, benchmarking will be a key part of the design process.
2. what storage option to use? What form should we store them in? Indexeddb & opfs are the two options. Indexeddb is more widely available than opfs so should be the option unless opfs is vastly faster.

samwillis:
I would argue that this functionality should be a layer above ShapeStream. ShapeStream is a nice thin abstraction of the protocol, and perfect for feeding into other stores. If that store has persistence, ShapeStream doesn't need any itself.

Maybe a PersistedShapeStream? It should be composable with a store implementation you could pass it (IndexedDB, OPFS, Node FS)?

KyleAMathews:
Ooo yes! I've been uneasy about throwing this into ShapeStream and persistence is a natural layer to compose above it.

```tsx
[...]

PR #1630, opened 2024-09-04
feat: `PersistedShapeStream` with generic storage interface

msfstef:
Addresses https://github.com/electric-sql/electric/issues/1519

Followed by https://github.com/electric-sql/electric/issues/1519 for more refactoring and improvements, split into two PRs for reviewability

Introduces a `PersistedShapeStream` - follows the newly defined `ShapeStreamInterface` so it can be used interchangeably, but requires:
1. A `storage` option to be provided, implementing the `ShapeStreamStorage` interface, which is a key-value store oriented interface
2. To not provide an `offset` and `shapeId` as the persisted shape stream cannot resume from an arbitrary offset and shapeId, it uses the persisted ones (although I suppose we could make it wipe the underlying store if an `offset` and `shpaeId` are provided to simulate the exact same behaviour)

What it does is it subscribes to the underlying `ShapeStream` and compacts entries to the provided store, very similar to what `Shape` does. When someone subscribes to the `PersistedShapeStream`, it looks into the store and returns the compacted entries as `insert`s and resumes from there.

One tricky bit with a bit of a hacky solution is the persistence of the `offset` and `shapeId`, needed to actually resume the stream. We _could_ store them in a separate `metadata` store entry, but that might require either 1) custom key prefixes on the store to separate metadata and 2) deal with inconsistent state.

For example, if metadata is stored separately from the operations and the store has no transactional guarantees (which the interface assumes), an operation might be stored with offset `10_3`, and the metadata fails to get stored because the app shuts down - so upon resuming the `10_3` operation is fed to the subscriber but then the stream is resumed from the previous latest offset that we managed to persist (e.g. `10_2`), duplicating the change.

The workaround is that the latest stored `offset` can be retrieved by reading into the stored entries and comparing their offsets (assuming the offset is not opaque here, but we can do that by adding a timestamp otherwise) and using the most recent one - has an overhead of needing to read the whole store to resume but is foolproof and can also do consistency checks to decide if it actually _can_ resume.

The `shapeId` I've stored alongside the entries, but really only in one entry stored, such that when we iterate over the data to get the latest offset we will eventually find the `shapeId` as well, but without needing to store it fully in every entry. Depending on the actual storage implementation this approach might be effective or slightly annoying. My main concern was to 1) avoid dealing with consistency with non-transactional stores and 2) not introduce separate storage APIs that people need to conform to.

If someone wants to use a transactional storage engine, they might as well go to `PGlite` with the `sync` plugin - I believe the use case for this is simpler key-value stores like `localStorage` etc.

### DIFF
--- a/.changeset/friendly-emus-invite.md
+++ b/.changeset/friendly-emus-invite.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
+Introduce `PersistedShapeStream` API for local caching of the shape stream.
--- a/packages/sync-service/.gitignore
+++ b/packages/sync-service/.gitignore
@@ -28,5 +28,6 @@ electric-*.tar
 # The shape database, created when the Sync Service runs
 /shapes/
 /state/
+/persistent/
 
 .env.dev.local
--- a/packages/typescript-client/package.json
+++ b/packages/typescript-client/package.json
@@ -12,6 +12,11 @@
       "types": "./dist/index.d.ts",
       "import": "./dist/index.mjs",
       "default": "./dist/cjs/index.cjs"
+    },
+    "./persist": {
+      "types": "./dist/persist.d.ts",
+      "import": "./dist/persist.mjs",
+      "default": "./dist/cjs/persist.cjs"
     }
   },
   "files": [
--- a/packages/typescript-client/src/async-or.ts
+++ b/packages/typescript-client/src/async-or.ts
@@ -1,0 +1,44 @@
+import { PromiseOr } from './types'
+
+export function isPromise<T>(promise: PromiseOr<T>): promise is Promise<T> {
+  return (
+    !!promise &&
+    typeof promise === `object` &&
+    `then` in promise &&
+    typeof promise.then === `function`
+  )
+}
+
+export function asyncOrCall<T>(
+  item: PromiseOr<T>,
+  callback: (item: T) => void,
+  onError?: (error: unknown) => void
+): PromiseOr<void> {
+  if (!isPromise(item)) {
+    try {
+      return callback(item)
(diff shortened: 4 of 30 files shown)


kevin-dp (changes requested):
This PR introduces many changes which makes it hard to review and merge.

I'm mainly concerned with the way how the code changed to handle synchronous and asynchronous callbacks differently. The message processor used to handle every callback as an asynchronous callback (also synchronous ones). As a result, the code was clear and concise. However, this PR handles sync callbacks differently (mainly to avoid JS pushing synchronous continuations on the microtask queue). As a result, code is written in a continuation passing style everywhere which makes it harder to understand.

I had a brief discussion with @icehaunter which shared the concern and would also stick with `await` for now. If performance is a concern here, we should benchmark a sync storage with the old implementation (that always awaits) and with the optimized implementation that doesn't await.

KyleAMathews:
> an operation might be stored with offset 10_3, and the metadata fails to get stored because the app shuts down - so upon resuming the 10_3 operation is fed to the subscriber but then the stream is resumed from the previous latest offset that we managed to persist (e.g. 10_2), duplicating the change.

We could also filter out the extra operations to avoid the duplicates.

That seems better so that on loading we can immediately send the sync call and not have to wait to load all the old data first.

KyleAMathews:
How about we put this on the back burner for some time? Being able to persist the shape log will be a nice feature at some point but it's not particularly urgent until we get a lot more users and some that definitely need this. Most users won't need this and I don't think it's worth a lot of time to benchmark and fine-tune this — especially if it requires a big refactor.

KyleAMathews:
This is pretty out-of-date w/ main & not a priority so I'll close it for now.

PR #1633, opened 2024-09-05
Fix sporadic tests erroring with :already_started

robacourt:
Some tests would sporadically fail `:already_started` due to multiple tests using the same name for the ShapeCache process. This PR ensures each test has it's own unique name.

### DIFF
--- a/packages/sync-service/test/electric/shape_cache_test.exs
+++ b/packages/sync-service/test/electric/shape_cache_test.exs
@@ -953,12 +953,9 @@ defmodule Electric.ShapeCacheTest do
     ]
 
     setup(ctx) do
-      shape_cache_server = __MODULE__.ShapeCache
-
       ctx =
         with_shape_cache(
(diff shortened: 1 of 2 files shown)


packages/sync-service/test/electric/shape_cache_test.exs
```diff
@@ -953,12 +953,9 @@ defmodule Electric.ShapeCacheTest do
     ]
 
     setup(ctx) do
-      shape_cache_server = __MODULE__.ShapeCache
```
robacourt, line 956: @magnetised it looks like you added this bit. Is there a reason why the name of the ShapeCache has to be specified here rather than using a unique name per test?

magnetised: hmm. I remember adding that, but it doesn't seem to be needed anymore does it. I think I needed to know the name elsewhere but no idea why now.

magnetised (approved):
Sorry!

PR #1634, opened 2024-09-05
chore: Client refactoring

msfstef:
As part of the work done for https://github.com/electric-sql/electric/pull/1630 I had to introduce a `ShapeStreamInterface`, and generally make the client slightly more modular in its utilities.

I've split out a bunch of things used and added unit tests for most (found some issues such as the `fetchWithBackoff` never terminating).

I've also refactored the classes to use the [proper private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties) in modern JS.


KyleAMathews:
> I've also refactored the classes to use the [proper private fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties) in modern JS.

These still aren't fully supported — but I assume they transpile fine to normal class fields?

packages/typescript-client/src/fetch.ts
```diff
@@ -0,0 +1,69 @@
...
+): typeof fetch {
+  const { initialDelay, maxDelay, multiplier, debug = false } = backoffOptions
+  return async (...args: Parameters<typeof fetch>): Promise<Response> => {
+    const url = args[0]
+    const options = args[1]
+
+    let delay = initialDelay
+    let attempt = 0
+
+    // eslint-disable-next-line no-constant-condition -- we're retrying with a lag until we get a non-500 response or the abort signal is triggered
```
KyleAMathews, line 34: > with a lag

What lag?

KyleAMathews (commented):
The scheme for modules looks great — it's hard to know what's changed though – could you leave comments on places where you made actual changes vs refactoring?

packages/typescript-client/src/fetch.ts
msfstef: it uses the delay, which it awaits in line 56

KyleAMathews: oh ok — the language is a bit confusing — the lag is only if there's an error.

KyleAMathews, line 34:
```suggestion
    // eslint-disable-next-line no-constant-condition — we re-fetch the shape log continuously until we get a non-ok response. For recoverable errors, we retry the fetch with exponential backoff. Users can pass in an AbortController to abort the fetching an any point.
```

msfstef:
@KyleAMathews yep the private fields get transpiled, although [there are some arguments](https://news.ycombinator.com/item?id=35747480) over whether this new private field spec is good...

The main benefit is that it is native JS encapsulation, private methods are no longer just private by convention/contract but are actually not useable. I'm happy to revert that change and go back to TS privates but I figured we could use the latest conventions

packages/typescript-client/src/constants.ts
msfstef, line 1: Moved HTTP API constants to one place - easier to change and less likely to have typos

packages/typescript-client/src/error.ts
msfstef, line 1: Moved `FetchError` here to have the various errors in one place

packages/typescript-client/src/fetch.ts
msfstef, line 1: Changed `fetchWIthBackoff` from a private method on `ShapeStream` to a fetch wrapper generator, tested separately. Moved it in its own file along with its options, and also fixed a bug where it would never actually stop retrying once it reached the max delay (unless that was our intended purpose? it feels weird to do that though)

packages/typescript-client/src/queue.ts
msfstef, line 1: We had a `MessageProcessor` class in our `client.ts` initially, which was created and typed specifically for messages although it was a very generic class, so I moved it here and built an `AsyncOrProcessingQueue` which I use in the persistence as well.

packages/typescript-client/src/shape.ts
msfstef, line 1: Moved the `Shape` implementation to its own file to have `ShapeStream`'s impl as lean as possible file-wise, and updated the private fields to use `#`

msfstef (commented):
Left comments on what changes I made, broadly. Rest of the files are added and refactored tests @KyleAMathews

packages/typescript-client/src/fetch.ts
KyleAMathews: >unless that was our intended purpose? 

Yeah it is intended that we keep trying ~forever — if you go offline, you want the ShapeStream to resume when you come back online even if that's hours/days later. There's not a good way to check if you're online or not other than just to keep trying.

msfstef: Hmm ok then let me remove the test asserting that and restore that behaviour

KyleAMathews: We could e.g. try leveraging https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine though the logic gets fussy e.g. localhost would work even if the machine isn't connected to wifi. Also `true` isn't guaranteed to actually make anything work. Fetching is simple and works.

I'm not sure what you mean by "nasty leaks". Fetches are cheap. But yeah, happy to discuss other options in a separate PR/issue.

KyleAMathews (approved):
I don't have time for a through review today but at a high-level it looks good to go — especially as there's no behavioral changes now.

This is getting merged into another PR so we can make more tweaks there if necessary.

packages/typescript-client/src/fetch.ts
msfstef: I think we probably want to avoid web exclusive APIs (at least `fetch` and others are easily polyfilled), and that one in particular I've worked before and ended up implementing my own "try to fetch google.com to determine if you're offline" as it's always hard to tell the exact state of the connection.

PR #1635, opened 2024-09-05
Clean up storage API

### LINKED ISSUE #1563: Clean up storage API
With the per-process storage implementation in #1551 the storage behaviour callbacks no longer need the `shape_id` (since each storage implementation only holds a single shape). 

The API is instead 

``` elixir
shape_opts = Storage.for_shape(shape_id, shared_opts)
# e.g.
Storage.make_new_snapshot!(stream, shape_opts)
```

I've re-written the current API to use the 2-step process in case there are call sites that aren't properly using a per-shape storage config, but this is a temporary solution.

The storage implementations should check that they're being called with a correctly configured backend and crash if not. That should weed out problems.

Also `Storage.list_shapes/1` is redundant. It should be replaced with some call that allows for getting the current snapshot xmin and offset (which is what I've used it for in `Shapes.Consumer` as a temporary measure).

magnetised:
Fixes #1563

With the new single-shape storage implementation, we can remove the `shape_id` param from almost all the Storage behaviour calls and also remove it from the actual db keys.

Some of the callbacks are also unecessary, such as `list_shapes` and `add_shape`. Instead we need a simple way to get a shapes current offset and snapshot xmin

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/file_storage.ex
@@ -1,36 +1,40 @@
 defmodule Electric.ShapeCache.FileStorage do
   alias Electric.Telemetry.OpenTelemetry
   alias Electric.Replication.LogOffset
-  @behaviour Electric.ShapeCache.Storage
+  alias __MODULE__, as: FS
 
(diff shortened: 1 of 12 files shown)


robacourt (approved):
Nice tidy up!

PR #1642, opened 2024-09-07
fix: pnpm workspace for website & broken link

KyleAMathews:


thruflo (approved):
Thanks — had hoped to have benchmarks page for go live. Let’s add next week.

We can also look at renaming DDN to Cloud and having signup as CTA.

KyleAMathews:
Yeah makes sense. I like DDN but it's pretty enterprise-y sounding so perhaps better for much later.

Issue #1646, opened 2024-09-08
support OPTIONS http requests


KyleAMathews:
Electric should respond to the OPTIONS method with which http methods we support (GET, OPTIONS, HEAD).

This is necessary to support browser preflight requests for CORS https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

PR #1649, opened 2024-09-09
feat (sync-service): support OPTIONS requests

kevin-dp:
This PR fixes https://github.com/electric-sql/electric/issues/1646 by adding support for HTTP OPTIONS requests.

### DIFF
--- a/.changeset/cold-wolves-cheer.md
+++ b/.changeset/cold-wolves-cheer.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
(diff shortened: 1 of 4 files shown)


robacourt (approved):
Nice!

kevin-dp:
@robacourt Current implementation accepts all origins by explicitly allowing the provided origin (when provided as an `origin` header in the request). I chose to implement it like that because `access-control-allow-origin: *` only works for requests without credentials. So this looked to be a more robust way of accepting any origin.

PR #1653, opened 2024-09-09
(chore): Fix sporadic tests by isolating test processes

robacourt:
We have various tests sporadically failing due to multiple tests using the same processes.

One example error is `:already_started` for the ShapeConsumer: two tests would simultaneously start the ShapeConsumer for the same shape_id. This is possible since the shape_id is based on time so if the tests generate a shape_id for the same table and where clause in the same millisecond they generate the same shape_id.

This PR isolates processes used in each test from processes used in other tests. This is done by introducing an `electric_instance_id` which is an atom based on the test name in the tests or `:default` in prod and dev.


packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -362,7 +361,11 @@ defmodule Electric.ShapeCache do
   end
 
   defp clean_up_shape(state, shape_id) do
-    Electric.Shapes.ConsumerSupervisor.stop_shape_consumer(state.consumer_supervisor, shape_id)
+    Electric.Shapes.ConsumerSupervisor.stop_shape_consumer(
+      state.electric_instance_id,
```
magnetised, line 365: Minor point. I get why you've put the instance id first, to be consistent with it everywhere, but the other funcs in this module take the supervisor's name as the first arg. Feels like that consistency is more important?

magnetised (approved):
Appreciate what a slog that must have been, and also nicely done finding the problem. Would it not be simpler to just ensure that shape ids are unique? 

We could have a single id generator process that would guarantee that shape ids never repeated and remove a lot of configuration noise (and pretty sure it wouldn't be a bottleneck). A simple `"#{hash(shape)}-#{time()}-#{counter()}"` id would solve the underlying problem and be much less invasive. Could even partition it across cores if we're worried about perf, so `"#{hash(shape)}-#{partition_id()}-#{time()}-#{counter()}"`. 

The amount of boilerplate just goes up monotonically (😉 ) and we'll be paying the cost with every PR.

Am approving with reservations.

robacourt:
> Appreciate what a slog that must have been, and also nicely done finding the problem. Would it not be simpler to just ensure that shape ids are unique?
> 
> We could have a single id generator process that would guarantee that shape ids never repeated and remove a lot of configuration noise (and pretty sure it wouldn't be a bottleneck). A simple `"#{hash(shape)}-#{time()}-#{counter()}"` id would solve the underlying problem and be much less invasive. Could even partition it across cores if we're worried about perf, so `"#{hash(shape)}-#{partition_id()}-#{time()}-#{counter()}"`.
> 
> The amount of boilerplate just goes up monotonically (😉 ) and we'll be paying the cost with every PR.
> 
> Am approving with reservations.

In some ways that is simpler @magnetised , but it doesn't achieve the same thing. Both solutions are introducing test isolation, but this PR is isolating as much as possible - proactive test isolation, and what you're suggesting just isolates in the areas where we have seen problems - reactive test isolation.

The problem with reacting to sporadic tests is that often you don't notice them when you write them, they get merged to main and then the team has flakey tests making development harder (was that my change that broke the test? having to run CI multiple times etc.) and then someone has to fix the test which is hard because it doesn't happen every time and you can't limit the cause to a small change since you don't know when it was introduced, you have to consider the whole codebase.

Proactive isolation on the other hand is simple, you isolate everything, singletons are not allowed, the supervision tree has an instance ID passed around in it. You don't have to think much. It becomes a simple to implement pattern. You prevent a huge class of sporadic tests from ever happening in the first place. Having the instance ID already in the supervision tree means that if you have a test that say rips down a process to ensure everything recovers nicely (which would certainly need to be isolated) it's easy to isolate, the instance ID is already there.

@icehaunter had already introduced proactive test isolation, but where you name the processes based on the test name. I looked at doing that for these processes but you ended up having lots of variables to pass round the code. The instance ID means you just have one thing to pass round, and also allows us to test large subsections of the supervision tree because you can easily isolate large sections of the supervision tree.

packages/sync-service/lib/electric/shape_cache.ex
robacourt: Good spot. I'll change that. Thanks

robacourt: Done

PR #1654, opened 2024-09-09
web: tidy up blog listings.

thruflo:


website/.vitepress/theme/custom.css
```diff
@@ -48,7 +48,7 @@
   font-family: OpenSauceOne;
   font-style: italic;
   font-weight: 400;
-  src: url('https://electric-sql.com/assets/fonts/OpenSauceOne-Italic-cc2c08f0979b8ebe8465993e6a1f8434.woff2') format('woff2');
```
KyleAMathews, line 51: normally you'd import a font right? That way it works locally when deving.

website/public/img/blog/electric-next/header.jpg
KyleAMathews, line 1: change the file name?

website/.vitepress/theme/custom.css
thruflo: I've switched to a local URL, so it will work in dev. You can import but this way loads in parallel rather than part of the bundle, I believe?

website/public/img/blog/electric-next/header.jpg
thruflo: For cache busting reasons?

website/.vitepress/theme/custom.css
KyleAMathews: yeah — "importing" a font means it'll just be added to the HTML head automatically but with the hashed url.

website/public/img/blog/electric-next/header.jpg
KyleAMathews: no just `electric-next` -> `electric`

thruflo: Ah, that is the slug of the blog post, as in https://electric-sql.com/blog/2024/07/17/electric-next because it's the "announcing electric next" post.

KyleAMathews: Oh haha nvm

PR #1655, opened 2024-09-09
website: more invalid markup fixes.

thruflo:


samwillis (approved):
All good except for https://deploy-preview-1655--electric-next.netlify.app/blog/2023/12/15/secure-transactions-with-local-first which seems to still have problems.

PR #1656, opened 2024-09-09
feat: add cache max-age of 5 seconds to ?live requests so request collapsing works

KyleAMathews:
We need a short max-age cache on ?live responses so http proxies will collapse long-polling requests.

The time is a bit arbitrary but two considerations:
- it's shorter than 20 seconds (our long-polling timeout) — which is necessary as otherwise clients would just poll the cached response over and over until the cache expired.
- it's long enough to ensure the vast majority of clients all request within the same five second window. Live clients all get responses at the same time so all request again at the same time. So even accounting for world-wide spread of clients, five seconds should collect pretty much everyone.

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -251,10 +251,11 @@ defmodule Electric.Plug.ServeShapePlug do
 
   defp put_resp_cache_headers(%Conn{} = conn, _) do
     if conn.assigns.live do
-      conn
-      |> put_resp_header("cache-control", "no-store, no-cache, must-revalidate, max-age=0")
-      |> put_resp_header("pragma", "no-cache")
-      |> put_resp_header("expires", "0")
+      put_resp_header(
+        conn,
+        "cache-control",
+        "max-age=5, stale-while-revalidate=5"
+      )
     else
       put_resp_header(
         conn,


icehaunter (approved):

thruflo:
I get the live clients connecting at the same time. Because they block on data and then reconnect after it arrives.

> There's a very slight chance that a new message could be returned within the 5 second timeout and then someone with ?live gets a cached response.

Trying to wrap my head around the scenario here. New data arrives within 5 seconds, the live request for one (or many) clients returns. A subsequent request gets that response, even if between the initial response and their request additional data has also arrived?

So if you miss being in the collapse gang and data is arriving in sequence you could end up with 5 seconds latency (for weird / hard to debug reasons from a front end POV)?

samwillis:
If I understand correctly the 5s TTL starts when the response is returned (either due to a message or the timeout), and so if it was possible to have a 1ms TTL that would also allow all connections within the 20s long poll to be collapsed.

what's the shortest possible TTL that triggers request collapsing?

this is very cool!

KyleAMathews:
> So if you miss being in the collapse gang and data is arriving in sequence you could end up with 5 seconds latency

No because you'd get a cache hit for the initial response and then the client would immediately join the *new* gang at the new offset and get collapsed into that request. It's possible that they missed several responses within a five second TTL window but that'd just mean they get individual messages for a series of cached http responses so very fast so not the worst thing in the world (in a pretty unlikely edge case).

In general, this is following our normal scheme where we trade off the possibility of needing to do a few http requests in sequence so that we can have longer caches & increase cache hit ratio.

KyleAMathews:
> If I understand correctly the 5s TTL starts when the response is returned (either due to a message or the timeout), and so if it was possible to have a 1ms TTL that would also allow all connections within the 20s long poll to be collapsed.

I'm a bit fuzzy on this but I *think* Fastly, et al. have heuristics around request collapsing so if they've seen similar URLs that have very short caches, they might stop pretty quickly collapsing requests. So e.g. a 1s cache might lead to later requests not being collapsed. Maybe. We could probably do more testing to figure out exactly how this works but it doesn't seem to matter too much as long as it satisfies the two considerations I outlined.

thruflo:
> No because you'd get a cache hit for the initial response and then the client would immediately join the new gang at the new offset

I see -- so the idea is that the live response is triggered when there is data, which means it always contains a new offset, so the client then makes a new request with the new offset.

Just to sanity check in case it's a bug that breaks this assumption, do we only set a cache header on a non-empty live request? What happens in the event of a 20s timeout response?

KyleAMathews:
> Just to sanity check in case it's a bug that breaks this assumption, do we only set a cache header on a non-empty live request? What happens in the event of a 20s timeout response?

We are but... it's already expired by then 😆 so it doesn't have any effect (other than to tell the CDN that they should keep collapsing requests to the origin)

thruflo:
> We are but... it's already expired by then 😆

Trying to twist my brain around that statement. Also reading https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#max-age and https://www.fastly.com/documentation/reference/http/http-headers/Age

Is it that the response has an age added because the CDN monitored the time it took to generate the response? My naive mental model is that the 20s timeout response is our business, generates a response with a max-age and the cdn/client gets that with an age of 0. Thus caches it for 5 seconds.

(On a separate note, I trust you that it's not but if an empty live response were to be cached it would also trigger zero delay polling between the client and the CDN.)

KyleAMathews:
yeah my mental model is that the max-age starts at the time the request is received at the CDN not when the response finishes. Which makes sense as the origin might take 500ms to transmit the data — but the staleness ticking clock starts *as soon* as the origin starts transmitting the data not when it happens to finish.

Also nothing broke dramatically i.e. clients started furiously looping getting the same cached response. I'll be doing more testing today so we'll see.

Issue #1448, opened 2024-07-24
Electric should be resilient to disconnecting to Postgres 


KyleAMathews:
Whether it's Postgres restarting or network connectivity issues.

- [ ] Write a test to verify Electric's graceful handling of Postgres connection closures.

balegas:
@alco shall we create the test and close this?

PR #1657, opened 2024-09-10
chore: Add an integration test that verifies Electric's ability to reconnect to the DB

alco:
Closes https://github.com/electric-sql/electric/issues/1448.


icehaunter (approved):
Great test to add!

PR #1668, opened 2024-09-11
Fix null value parsing

dustin-H:
Hi there,

first of all, I really like your project! Amazing work! 🚀 
Thank you!

While integrating it I found a bug:
When a shape includes a column, that is nullable and currently null or set to null, the typescript client is crashing:

Screenshot 2024-09-11 at 07 20 16

The code is checking if the `typeof value` is `object` and is then trying to iterate over it's keys.
However, `typeof null` is also `object`. This is why it crashes here, as it cannot iterate over null.
This PR fixes https://github.com/electric-sql/electric/issues/1673.

A simple check for `null` fixes this as proposed in this PR.

Could you please merge it? Or is there anything I've overseen?
Thanks!

### DIFF
--- a/packages/typescript-client/test/parser.test.ts
+++ b/packages/typescript-client/test/parser.test.ts
@@ -214,11 +214,11 @@ describe(`Message parser`, () => {
 
   it(`should parse null value on column named value`, () => {
     const schema = {
-      a: { type: `text`, dims: 1 },
+      value: { type: `text`, dims: 1 },
     }
 
-    expect(parser.parse(`[ { "value": null } ]`, schema)).toEqual([
-      { value: null },
+    expect(parser.parse(`[ { "value": {"value": null} } ]`, schema)).toEqual([
+      { value: { value: null } },
     ])
   })
 })
--- a/packages/typescript-client/test/parser.test.ts
+++ b/packages/typescript-client/test/parser.test.ts
@@ -160,8 +160,16 @@ describe(`Message parser`, () => {
   const parser = new MessageParser()
 
   it(`should parse null values`, () => {
-    const messages = `[ { "value": { "a": null } } ]`
-    const expectedParsedMessages = [{ value: { a: null } }]
+    const messages = `[ { "value": { "a": null } }, { "value": { "value": null } } ]`
+    const expectedParsedMessages = [
+      { value: { a: null } },
+      { value: { value: null } },
+    ]
+
+    const schema = (tpe: string, dims: number | undefined) => ({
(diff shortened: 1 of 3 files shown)


KyleAMathews:
Hey thanks for reporting and fixing the issue! 

Could you also add a test for the issue? Lemme know if you need help running the tests. The contribution docs are immature still.

dustin-H:
@KyleAMathews Thanks for your quick reply.

I just wrote a test, made the check more precise and added a comment to explain why the check for null is required.

If you run the test without my change, it will fail with the error posted above.

packages/typescript-client/test/parser.test.ts
```diff
@@ -211,4 +211,14 @@ describe(`Message parser`, () => {
       parser.parse(`[ { "value": { "a": "{1,2,NULL,4,5}" } } ]`, schema)
     ).toEqual([{ value: { a: [1, 2, null, 4, 5] } }])
   })
+
+  it(`should parse null value on column named value`, () => {
+    const schema = {
+      a: { type: `text`, dims: 1 },
+    }
+
+    expect(parser.parse(`[ { "value": null } ]`, schema)).toEqual([
```
kevin-dp, line 220: I believe the message in this test `[ { "value": null } ]` is wrong.
The structure of a message is described by [this type](https://github.com/electric-sql/electric/blob/main/packages/typescript-client/src/types.ts#L22-L27):

```ts
export type ChangeMessage<T extends Row = Row> = {
  key: string
  value: T
  headers: Header & { operation: `insert` | `update` | `delete` }
  offset: Offset
}

export type Row = { [key: string]: Value }
```

Essentially, the value of the `value` key must always be an object, namely the row in the table. That row has a key for each column and it's the value that is associated to that column that can be `null`. So, the message for the example you wanted to test would be `[ { "value": { "a": null } } ]`. If you look at [line 163](https://github.com/electric-sql/electric/blob/main/packages/typescript-client/test/parser.test.ts#L163), that's exactly what we test.

dustin-H:
Hi @kevin-dp,

you are right. I accidentally created the test with an invalid message, which led to the same result as the correct one.

I created the issue you requested: https://github.com/electric-sql/electric/issues/1673

I also fixed the test to match the message I mean.

kevin-dp:
Fixes https://github.com/electric-sql/electric/issues/1673.

kevin-dp (approved):
Thank you for contributing the test case we were missing for columns named "value" and providing the fix 💯

PR #1672, opened 2024-09-11
chore (sync-service): run filter logic for PG version 15 in unit tests

kevin-dp:
This modifies the unit tests to use the publication filter logic for PG 15 as that is more advanced than the logic for version 14 and below.

### DIFF
--- a/.changeset/rare-cooks-marry.md
+++ b/.changeset/rare-cooks-marry.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Run filter logic for PG version 15 in unit tests.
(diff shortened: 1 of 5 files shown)


alco (commented):
This works fine locally, but fails on CI when Postgres 14 is used. To fix that CI job, the hard-coded version needs to be replaced by a setting looked up at runtime.

I think it'd be best to add an environment variable `POSTGRES_MAJOR_VERSION` for both CI jobs and look it up in `config/runtime.exs` inside a conditional block:
```
if config_env() == :test do
  config :electric, major_pg_version_for_tests: env!("POSTGRES_MAJOR_VERSION", :integer, 15)
end
```

Then in the test code you can fetch the setting via `Application.fetch_env!(:electric, :major_pg_version_for_tests)`.

alco (approved):
:+1:

Issue #1663, opened 2024-09-10
Validate that the shape_id matches the rest of the URL structure


KyleAMathews:
This isn't an issue when using the official clients but when hand-constructing shape URLs or when writing a new one, it's possible to write a URL where the various parts don't match each other.

We validate already that the shape_id is the current shape_id for the url structure by returning a 409 if it's not.

We also need to validate that the shape_id is for the shape that the URL describes i.e. if a shape is defined as `v1/shape/foo` but then the shape_id for that shape def is applied to `v1/shape/foo?where='foo=true'` then we should return a 400.

Follow-up issue to https://github.com/electric-sql/electric/issues/1659

PR #1678, opened 2024-09-12
feat (sync-service): validate shape ID matches shape definition from URL

kevin-dp:
This PR fixes https://github.com/electric-sql/electric/issues/1663.

The implemented semantics are the following.
- When shape id exists:
  - and is the currently active shape for this shape definition:
    - serve it normally
  - and is not the currently active shape for that shape definition:
    - return 400 (because there must be a mismatch between shape ID and shape def, otherwise this would be the active shape for that shape definition)
- When shape id does not (or no longer) exist:
  - if there is an active shape for this shape definition:
    - return 409 that redirects to the active shape
  - if there is no active shape for this shape definition:
    - create the shape
    - return a 409 with a redirect to the newly created shape

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -12,7 +12,6 @@ defmodule Electric.Plug.ServeShapePlug do
   # Control messages
   @up_to_date [Jason.encode!(%{headers: %{control: "up-to-date"}})]
   @must_refetch Jason.encode!([%{headers: %{control: "must-refetch"}}])
-  @shape_mismatch "The provided shape ID does not match the shape definition."
 
   defmodule Params do
     use Ecto.Schema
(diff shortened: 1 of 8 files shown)


kevin-dp:
This approach has a rather annoying problem.
Previously, when someone would request a shape ID that does not exist, we would create the shape and return a 409 with a redirect to the newly created shape.

In this PR, if someone comes in with a shape ID that does not exist, there are 2 possibilities:
1. There exists another shape with the same definition, we return a 409 with a redirect to that shape
2. There is no shape for this shape definition. We return a 400 bad request since the shape ID cannot possibly match the shape definition (otherwise we would have found it our have been in case 1).

The second case is problematic when combined with caching.
For example, imagine user 1 creates a shape with id 1.
Then, user 2 deletes that shape.
Then, user 1 tries to go live, the shape does not exist and there is no shape for this definition, so we return 400.
As a result, user 1 goes back to initial sync with offset -1 but will get served a cached response so will get shape id 1, will try to go live again and will again get a 400.

kevin-dp:
Thinking further about it. Case 2 seems wrong. As demonstrated by the delete example, a user can end up in case 2 because the shape got deleted. So if the shape ID does not exist and there is no other shape that matches the shape definition, we should create the shape and return a 409 that redirects to the newly created shape.

Only in the case where someone requests a shape ID that exists but does not match the shape def, do we return a 400.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -146,61 +145,103 @@ defmodule Electric.Plug.ServeShapePlug do
...
+    handle_shape_info(conn, shape_info)
   end
 
-  defp schema(shape) do
-    shape.table_info
-    |> Map.fetch!(shape.root_table)
-    |> Map.fetch!(:columns)
-    |> Schema.from_column_info()
-    |> Jason.encode!()
+  # No shape_id is provided so we can get the existing one for this shape
```
KyleAMathews, line 152: should we validate that no `shape_id` also corresponds with `offset=-1`? A client has to start at offset -1 and would then have a shape_id so shouldn't ever be in a situation where they request a non -1 offset w/o a shape_id.

KyleAMathews (approved):
Looks great! Logic seems crisp and correct. :shipit:

robacourt (approved):
👍

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
kevin-dp: We are checking that here: https://github.com/electric-sql/electric/blob/a525374159852df7efa794d81468fb783b2f7de1/packages/sync-service/lib/electric/plug/serve_shape_plug.ex#L78

PR #1682, opened 2024-09-12
docs: update local development guide.

thruflo:
https://deploy-preview-1682--electric-next.netlify.app/docs/guides/installation
https://deploy-preview-1682--electric-next.netlify.app/docs/guides/troubleshooting


website/docs/guides/local-development.md
```diff
@@ -45,4 +112,25 @@ npx @radically-straightforward/caddy run --config - --adapter caddyfile <<EOF
 EOF
 ```
 
-Now change your shape URLs to use port 3001 instead of port 3000 and you're done!
+Now change your shape URLs to use port `3001` instead of port 3000 and everything will run much faster 🚀
+
+### Resetting server state
```
KyleAMathews, line 117: these latter bits aren't really that Electric-specific? It's more "how do I run postgres for dev" stuff. I think keeping our docs simple and focused on Electric-specific stuff would be good for now until we have doc resources to expand our focus a bit.

website/docs/guides/local-development.md
```diff
@@ -1,41 +1,108 @@
...
+
+```sh
+docker run \
+    -e "DATABASE_URL=postgresql://..." \
+    -p 3000:3000 \
+    -t \
+    electricsql/electric:latest
+```
+
+### Using Elixir
```
KyleAMathews, line 43: Shouldn't this be in an Elixir-specific guide? 99% of our users won't be using Elixir so this would be somewhat confusing to them.

KyleAMathews (changes requested):
I think this doc can be pretty bare — the quickstarts & examples are better guides for getting started — I'd mostly point people there & then have this be more an FAQ for problems like the http 1.1 one

website/docs/guides/local-development.md
thruflo: My experience is this type of issue, where you get stuck with state in Electric of Postgres is exactly the kind of issue you get in dev and that’s hard to understand and debug out of unless you understand the system.

You wipe your local db and docker down and up. Why the hell is there data in my shape that isn’t in my DB?

How the hell do I properly reset this thing?

The WAL one is more a general gotcha but whilst we’re listing them … maybe it belongs in an “operating successfully” page. But we don’t have one of those yet.

website/docs/guides/local-development.md
thruflo: Yes, it’s more of a niche thing. But this guide isn’t on a hot onboarding path and some devs may want to know how the thing actually runs.

website/docs/guides/local-development.md
KyleAMathews: yeah sure — if we make this a grab bag type page, maybe make it more explicitly FAQ style where each section is a question? My experience is that people read docs assuming that each part is of equal importance. Making each section a separate question limits that anxiety that they need to understand everything.

thruflo:
I think we need a guide page about running the software. A quickstart is a short happy path, not details on how to install and run.

For me, the content here is more of a “Running Electric” reference page. It doesn’t have to be high up the IA but it’s a reference page for how to run the thing successfully.

thruflo:
@KyleAMathews refactored this into separate Installation and Troubleshooting guides. The former allowed me to simplify the Quickstart a touch. The latter could evolve into a more structured FAQ page in time.

KyleAMathews (approved):
Love it!

PR #1691, opened 2024-09-13
(chore) Fix some sporadic tests

robacourt:


packages/sync-service/test/electric/shapes/consumer_test.exs
```diff
@@ -101,8 +102,8 @@ defmodule Electric.Shapes.ConsumerTest do
         })
 
       {:ok, producer} =
-        Electric.Replication.ShapeLogCollector.start_link(
-          name: __MODULE__.ShapeLogCollector,
+        ShapeLogCollector.start_link(
+          electric_instance_id: ctx.electric_instance_id,
```
robacourt, line 106: This fixes the sporadic test that fails with 

```
** (MatchError) no match of right hand side value: {:error, {:already_started, #PID<0.2293.0>}}
     stacktrace:
       test/electric/shapes/consumer_test.exs:103: Electric.Shapes.ConsumerTest.__ex_unit_setup_0_2/1
       Electric.Shapes.ConsumerTest.__ex_unit_describe_0/1
  ```
As seen [here](https://github.com/electric-sql/electric/actions/runs/10854991043/job/30126613288?pr=1691)
  
Elixir doesn't guarantee that the processes from the previous test are terminated before starting the next test, so named processes must be named uniquely for the test. I've achieved here with an `electric_instance_id`, carrying on the pattern introduced in #1653 which means it not only fixes this test but makes it harder to introduce the same issue in tests written in the future.

File: packages/sync-service/lib/electric/replication/shape_log_collector.ex [hunk 2]
```diff
@@ -26,22 +22,26 @@ defmodule Electric.Replication.ShapeLogCollector do
 
   def start_link(opts) do
     with {:ok, opts} <- NimbleOptions.validate(opts, @schema) do
-      GenStage.start_link(__MODULE__, Map.new(opts), name: opts[:name])
+      GenStage.start_link(__MODULE__, Map.new(opts), name: name(opts[:electric_instance_id]))
     end
   end
 
+  def name(electric_instance_id) do
+    :"shape_log_collector #{electric_instance_id}"
```
magnetised, line 30: Any particular reason not to use the `Electric.Application.process_name` scheme?

magnetised (approved):
Nicely done. Think we should be consistent with the process names, not sure if you have some philosophical objection to using the separate registry.

packages/sync-service/test/electric/shapes/consumer_test.exs
alco: As an aside, a point of clarification from ExUnit docs (something I've clarfieid for myself just now and would like to share it here):

> [start_supervised/2](https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#start_supervised/2) is used to start processes under a supervisor. The supervisor is linked to the current test process. The supervisor as well as all child processes are guaranteed to terminate before any [on_exit/2](https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#on_exit/2) callback runs.

and this useful bit from the documentation for `start_supervised/2`:

> The started process is not linked to the test process and a crash will not necessarily fail the test. To start and link a process to guarantee that any crash would also fail the test use [start_link_supervised!/2](https://hexdocs.pm/ex_unit/ExUnit.Callbacks.html#start_link_supervised!/2).

alco (approved):
Superb!

packages/sync-service/test/electric/shapes/consumer_test.exs
robacourt: Thank you so much @alco !  I was pretty surprised when I realised that ExUnit didn't guarantee that the processes from the previous test are terminated before starting the next test, but now it makes sense. That's super useful. I'm pretty sure I can simplify some of the tests I've written (outside of this PR) with this.

I still quite like the `electric_instance_id` pattern, because as mentioned above it makes it harder to introduce the same issue in tests written in the future, and as a general pattern, as mentioned in #1653 it "means you just have one thing to pass round, and also allows us to test large subsections of the supervision tree because you can easily isolate large sections of the supervision tree" so I might keep the PR as is for now.

PR #1694, opened 2024-09-13
fix: fix package.json names that were duplicated

beorn:
Duplicated package names causes problems with various monorepo tools.  The package names now match the folder name.


KyleAMathews (approved):
Nice fix! Thanks!

Issue #1689, opened 2024-09-13
Electric doesn't clean shapes after connecting to a new postgres with a different schema


KyleAMathews:
steps to reproduce:
1. Follow the quickstart and add the table `foo` with a few rows of data
2. Fetch the foo table shape
3. Restart Postgres as an empty db
4. Electric reconnects
5. Try fetching the foo shape and see that the old data is returned instead of the table not found error

[DIscord link](https://discord.com/channels/933657521581858818/1284172917059616970/1284177755935739925)

KyleAMathews:
lirbank (discord user name) also tried inserting new rows into the new database (with identical table) and those weren't added to the shape. So perhaps we're also missing adding the right publication when we connect to the new db? cc @kevin-dp

lirbank:
@lirbank on GH too 😃

thruflo:
I've slightly addressed this here https://electric-sql.com/docs/guides/troubleshooting#clear-data-mdash-how-do-i-clear-the-server-state -- I wonder (a) if we want to add some functionality to wipe shapes when PG changes or (b) whether we want to be clearer about the behaviour and if so where and how to document.

One thing we've done before for telemetry is write a unique identifier into the PG instance. Perhaps one approach here could be when Electric starts it can get-or-create a uuid and then use it internally as part of the shape lookup.

lirbank:
I just did this:

1. Fired up the quickstart docker compose file
2. Created todo table, added 200,000 rows
3. Run the curl command, shows the 200,000 rows ✅
4. Dropped the todo table
5. Run the curl command, shows the 200,000 rows ❌
6. Create the todo table again, added 200,000 rows again but with different "task" names so it is clearly different from the original data set
5. Run the curl command, still shows the original 200,000 rows ❌

KyleAMathews:
ok so very broken 😬

lirbank:
Here is another:

1. Remove all docker containers and start fresh
2. Create the todo table
3. Insert some rows
4. Run `DELETE FROM todo;`
5. Now `curl -i 'http://localhost:3000/v1/shape/todo?offset=-1` still return the stale data
6. `SELECT COUNT(*) FROM todo;` return 0

Docker image is now canary:

`electricsql/electric:canary`
`sha256:5d0db221daf0f74d6dd2bbba034612d626e15905f3a784273b8e27c75c5e81cd`

lirbank:
[...]

PR #1703, opened 2024-09-16
feat (sync-service)!: clean all shapes when switching to a different DB

kevin-dp:
Fixes https://github.com/electric-sql/electric/issues/1689.
This PR uses Postgres' `system_identifier` to detect when Electric is connected to a different Postgres database than before. When that is the case, we log a warning and clean all shapes. Cleaning all shapes is consistent with how we handle Point In Time Recoveries (where we also clean all shapes).

Note: this is a breaking change because we modify the structure of the timeline we store persistently. We used to just store the timeline ID, but now we store a list `[pg_id, timeline_id]` so we can't read out old persisted timeline information as we are expecting the new structure.

### DIFF
--- a/.changeset/thick-experts-call.md
+++ b/.changeset/thick-experts-call.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Detect when Electric is connected to a different Postgres DB than before and clean all shapes.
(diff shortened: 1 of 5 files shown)


robacourt (approved):
Nice work!

Issue #1667, opened 2024-09-10
Chunking algorithm is creating very small chunks


KyleAMathews:
I have a shape that probably now has several 1000 operations in it. For some reason the chunking algorithm is returning an initial chunk with 87 messages than over a hundred now different chunks, each with 37 messages in it.

The total data transferred is still only 150 kb so I'd expect that there'd still be only one chunk not hundreds. The number of chunks is creating a noticeable delay now in starting the app.

<img width="1758" alt="Screenshot 2024-09-10 at 1 16 58 PM" src="https://github.com/user-attachments/assets/1149e38b-c43a-481b-86ea-364f3d6a3bd1">

KyleAMathews:
Another in-the-wild report of this https://discord.com/channels/933657521581858818/1280199098833698889/1284213285666689146

KyleAMathews:
Another one https://discord.com/channels/933657521581858818/1284719450544803954/1284723039686098998

msfstef:
Closing as https://github.com/electric-sql/electric/pull/1704 should address this - it was chunking at 10kB rather than 10MB

PR #1704, opened 2024-09-16
fix: Adjust chunk size to 10 MB

msfstef:
Addresses https://github.com/electric-sql/electric/issues/1667

We were using  10kB chunks rather than 10MB chunks - whoops!

This is still the uncompressed threshold, if gzip is added the chunks should be _significantly_ smaller, but we need some work on estimating that (generally speaking for json it could be 90-95% size reduction, but it's hard to tell if some rows contain large items that are not easily compressable)

I've kept the dev environment chunk size to 10kB such that integration tests can work with small chunks and still be fast (i.e. we don't have to load PG with hundreds of MB of data to test chunking)

### DIFF
--- a/packages/sync-service/.env.dev
+++ b/packages/sync-service/.env.dev
@@ -2,3 +2,4 @@ DATABASE_URL=postgresql://postgres:password@localhost:54321/electric?sslmode=dis
 ENABLE_INTEGRATION_TESTING=true
 CACHE_MAX_AGE=1
 CACHE_STALE_AGE=3
+LOG_CHUNK_BYTES_THRESHOLD=10000
--- a/packages/sync-service/.env.dev
+++ b/packages/sync-service/.env.dev
@@ -2,4 +2,5 @@ DATABASE_URL=postgresql://postgres:password@localhost:54321/electric?sslmode=dis
 ENABLE_INTEGRATION_TESTING=true
 CACHE_MAX_AGE=1
 CACHE_STALE_AGE=3
+# using a small chunk size of 10kB for dev to speed up tests
 LOG_CHUNK_BYTES_THRESHOLD=10000
--- a/packages/sync-service/lib/electric/shape_cache/log_chunker.ex
+++ b/packages/sync-service/lib/electric/shape_cache/log_chunker.ex
@@ -1,5 +1,5 @@
 defmodule Electric.ShapeCache.LogChunker do
-  @default_threshold 10_000
+  @default_threshold 10_000_000
 
   @doc """
   Add bytes to the current chunk of a given shape - if the chunk exceeds the specified
--- a/packages/sync-service/lib/electric/shape_cache/log_chunker.ex
+++ b/packages/sync-service/lib/electric/shape_cache/log_chunker.ex
@@ -1,5 +1,5 @@
 defmodule Electric.ShapeCache.LogChunker do
-  @default_threshold 10_000_000
+  @default_threshold 10 * 1024 * 1024
 
   @doc """
   Add bytes to the current chunk of a given shape - if the chunk exceeds the specified
--- a/packages/sync-service/lib/electric/shape_cache/log_chunker.ex
+++ b/packages/sync-service/lib/electric/shape_cache/log_chunker.ex
@@ -1,4 +1,6 @@
 defmodule Electric.ShapeCache.LogChunker do
+  # Default chunk size of 10 MB to ensure caches accept them
+  # see: https://github.com/electric-sql/electric/issues/1581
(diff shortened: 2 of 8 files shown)


packages/sync-service/.env.dev
```diff
@@ -2,3 +2,4 @@ DATABASE_URL=postgresql://postgres:password@localhost:54321/electric?sslmode=dis
 ENABLE_INTEGRATION_TESTING=true
 CACHE_MAX_AGE=1
 CACHE_STALE_AGE=3
+LOG_CHUNK_BYTES_THRESHOLD=10000
```
kevin-dp, line 5: Is it normal that this is still 10 000 while you seem to have replace all occurences of 10 000 by 10M ?

msfstef: Ah yes forgot to add to PR description - I kept the dev setting to a small chunk so we can do integration tests of chunking without having to deal with very large data and slow tests down (similar to how we have very small cache ages for the tests)

packages/sync-service/lib/electric/shape_cache/log_chunker.ex
```diff
@@ -1,5 +1,5 @@
 defmodule Electric.ShapeCache.LogChunker do
-  @default_threshold 10_000
+  @default_threshold 10_000_000
```
robacourt, line 2: 10,000,000 bytes is an odd amount, no? 9.5MB? Doesn't matter much but it's the sort of thing that makes me wonder why and think about when there's nothing to think about 😄 I would probably do:
```
@default_threshold 10 * 1024 * 1024
```

packages/sync-service/test/electric/replication/shape_log_collector_test.exs
```diff
@@ -55,7 +55,7 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
       shape_cache_opts =
         [
           storage: {Mock.Storage, []},
-          chunk_bytes_threshold: 10_000,
+          chunk_bytes_threshold: Electric.ShapeCache.LogChunker.default_chunk_size_threshold(),
```
robacourt, line 58: Yeah, great to remove the duplication of the numbers. Thanks!

robacourt (approved):
Lovely! Just left a little suggestion

packages/sync-service/lib/electric/shape_cache/log_chunker.ex
msfstef: good point! ultimately the threshold is super loose, chunks can contain quite a bit more than the specified number and gzipping will change that significantly as well - but good to have actually meaningful numbers!

thruflo:
Just a suggestion observing the discussion of the intention behind the numbers in this thread. Perhaps it’s worth adding comments into the source that explain what the number represents and why it is the value it is. “10,000 bytes, this is a small chunk value in dev because reasons …”

packages/sync-service/lib/electric/shape_cache/log_chunker.ex
```diff
@@ -1,5 +1,5 @@
 defmodule Electric.ShapeCache.LogChunker do
-  @default_threshold 10_000
+  @default_threshold 10 * 1024 * 1024
```
alco, line 2: Is this default value even used anywhere? I see some test code that explicitly passes the threshold argument. Yet in the regular code a value that's originally read from the Application env is used and that one is set to 10_000 in config/runtime.exs.

alco: Oops, sorry, I made the mistake of looking at a different version of the code that I have checked out on my machine as opposed to the diff here. Nevermind!

Anyway, it'd be nice to have a brief explanation for the 10MB as a comment here. Maybe a URL to an issue explaining the reasoning behind it?

msfstef: The default value is returned by `Electric.ShapeCache.LogChunker.default_chunk_size_threshold()` which now both tests and the config in `runtime.exs` use

The only case where `10_000` is used is when `.env.dev` is used where I set it to that for faster integration tests

KyleAMathews (approved):
10mb uncompressed seems fine for now. We can keep fine-tuning this number with more testing in the future! Thanks for the fix!

packages/sync-service/lib/electric/shape_cache/log_chunker.ex
msfstef: re: explanation - added single-sentence explanation and a link to the original chunking issue! I think that should cover it, thanks for bringing that up

PR #1705, opened 2024-09-16
fix: Clean shape data when individual shape is cleaned up

msfstef:
Cleans up the underlying data for the shape, not only the metadata.

I'm not entirely sure if this is the best approach or if there's a cleaner way @robacourt ?

(also fixed an integration test that was causing issues)

### DIFF
--- a/.changeset/wicked-beds-play.md
+++ b/.changeset/wicked-beds-play.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Clean up underlying shape data when cleaning up shape.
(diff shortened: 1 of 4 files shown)


kevin-dp (changes requested):
Can we add a unit to shape cache that tests that the storage is properly cleaned up when the shape is cleaned?

robacourt (approved):
Looks good to me! 

This may have introduces a sporadic test though: https://github.com/electric-sql/electric/actions/runs/10883204574/job/30195829206?pr=1705 I've been looking a lot at sporadic tests recently, fixed all the ones I could find, and didn't spot this one. That's no guarantee that it's anything to do with your changes though 😄 I can help with diagnosing it though if you like

msfstef:
yep sorry adding tests now wanted to check if you agree with the approach first! the `ShapeCache` tests need to mock out the storage now

robacourt (approved):
Nice!

kevin-dp (approved):

PR #1708, opened 2024-09-16
Add configuration to publish hex package

magnetised:
Claiming the `electric` package name for future fun.

- Copied the repo `LICENSE` file to the root of the Elixir app so it's present when installing as a dependency
- Removed the `installation` section from the README.md to avoid problems syncing the version
- Added a workflow step to publish the hex package alongside the NPM ones
- Fixed minor type issue (there's no such type as `Stream.t()`)
- Note I had to publish `pg_query_ex` to Hex as well, because you can't publish hex packages with Github-based dependencies https://github.com/electric-sql/pg_query_ex/pull/4


alco (approved):
:rocket:

PR #1711, opened 2024-09-16
chore: Refactor client

msfstef:
Extracted out refactoring from https://github.com/electric-sql/electric/pull/1630

~~Had to also add a `subscribeSync` method on `ShapeStream` otherwise we cannot guarantee that `Shape` is up to date when the stream is up to date, since the processing is done asynchronously. We have three options:~~
~~1.  We re-introduce the `PromiseOr` abstractions in order for the single `subscribe` call to handle both sync and async~~
~~2. We introduce an explicit `subcsribeSync` like I have done here, which provides strong guarantees on up-to-date synchronicity~~
~~3. We change `Shape` to maintain it's own "up to date" logic as it is not guaranteed that when the stream is up to date that the `Shape` will have materialized its local state fully, since processing is async~~

Reintroduced thin version of sync + async handling in the queue to maintain the current API and intended behaviour.

I've discussed with @kevin-dp and we think that actually making the processing callbacks create backpressure on the stream rather than the stream collecting the responses non-stop might be a better API and allow the developers to choose the pace at which the stream will collect data through the choice of processing callback.  Will open a new PR for this, so we can merge this one with the refactor.


packages/typescript-client/src/client.ts
```diff
@@ -57,86 +56,28 @@ export interface ShapeStreamOptions {
...
+    callback: (messages: Message<T>[]) => MaybePromise<void>,
+    onError?: (error: FetchError | Error) => void
+  ): void
+  subscribeSync(
+    callback: (messages: Message<T>[]) => void,
+    onError?: (error: FetchError | Error) => void
+  ): void
+  unsubscribeAllUpToDateSubscribers(): void
+  unsubscribeAll(): void
+  subscribeOnceToUpToDate(
```
kevin-dp, line 70: API-wise i'm wondering why there are 2 versions of `subscribe` but only 1 version of `subscribeOnceToUpToDate`.

packages/typescript-client/src/client.ts
```diff
@@ -169,214 +110,249 @@ export class FetchError extends Error {
...
   /** True during initial fetch. False afterwise.  */
   isLoading(): boolean {
     return !this.isUpToDate
   }
 
-  private notifyUpToDateSubscribers() {
-    this.upToDateSubscribers.forEach(([callback]) => {
+  #publish(messages: Message<T>[]) {
+    this.#subscribers.forEach(([subscriber, _]) => {
+      subscriber.process(messages)
```
kevin-dp, line 338: An alternative to having `subscribe` and `subscribeSync` would be to only have `subscribe` and replace this `forEach` by a regular for loop in which we `await subscriber.process(messages)`.

packages/typescript-client/src/client.ts
```diff
@@ -385,248 +361,35 @@ export class ShapeStream<T extends Row = Row> {
...
-        throw e
-      }
-    )
-  }
-
-  lastSynced(): number {
-    return this.stream.lastSynced()
+function validateOptions(
+  options: Partial<ShapeStreamOptions>
+): options is ShapeStreamOptions {
```
kevin-dp, line 375: This function's return type is a bit contradictory with its implementation.
This function behaves as a validation function, which throws if the input is not valid.
If the function returns normally, the input is valid.

However, the return type indicates that this function is used to check if the input if of type `ShapeStreamOptions` but then always returns true (because if it is not valid it will throw).
So i would argue that either it should not throw and return true or false, or, it should throw when invalid and return void when valid.
Since we never use the return value of this function, i would opt for the 2nd option.

packages/typescript-client/src/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const SHAPE_ID_HEADER = `X-Electric-Shape-Id`
```
kevin-dp, line 1: HTTP headers are case-insensitive and are converted to lowercase in both directions.
Therefore, i would make all these constants lowercase to avoid subtle bugs, e.g. when checking if a certain header is present (although if we use the right method for this it probably is smart enough to convert to lowercase).

packages/typescript-client/src/constants.ts
```diff
@@ -0,0 +1,7 @@
...
+export const CHUNK_LAST_OFFSET_HEADER = `X-Electric-Chunk-Last-Offset`
```
kevin-dp, line 2: This is beyond the scope of this PR, but noting here that the convention of prefixing custom headers with X- has been deprecated and advised against since 2012 (https://datatracker.ietf.org/doc/html/rfc6648#page-4).

packages/typescript-client/src/fetch.ts
```diff
@@ -0,0 +1,73 @@
+import { FetchError, FetchBackoffAbortError } from './error'
+
+export interface BackoffOptions {
```
kevin-dp, line 3: Let's add docstrings to the options to clarify the unit. I believe `initialDelay` and `maxDelay` are in milliseconds.

packages/typescript-client/src/offset.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export function compareOffset(offsetA: Offset, offsetB: Offset): 1 | 0 | -1 {
+  const [oAx, oAy] = splitOffset(offsetA)
+  const [oBx, oBy] = splitOffset(offsetB)
+  if (oAx > oBx) return 1
+  if (oAx < oBx) return -1
+  if (oAy > oBy) return 1
+  if (oAy < oBy) return -1
+  return 0
+}
```
kevin-dp, line 21: Comparator functions in JS are not limited to returning 1, 0, or -1.
In general, they should return negative number if A < B, zero when equal, and a positive number when A > B.
This is very convenient as it allows comparison to be implemented as `A - B`.
For this example, that means we can simplify it to:

```suggestion
export function compareOffset(offsetA: Offset, offsetB: Offset): number {
  const [a1, a2] = splitOffset(offsetA)
  const [b1, b2] = splitOffset(offsetB)
  return a1 - b1 || a2 - b2
}
```

packages/typescript-client/test/offset.test.ts
```diff
@@ -0,0 +1,36 @@
+import { describe, it, expect } from 'vitest'
+import { compareOffset } from '../src/offset'
+
+describe(`compareOffset`, () => {
+  it(`should return 1 when the first part of offsetA is greater than offsetB`, () => {
```
kevin-dp, line 5: Considering the refactor of `compareOffset` we could modify these tests to check whether the number is negative or positive, or, perhaps simpler, we could write a single test that passes an array of unordered offsets to JS' sort function and check that the result is correctly sorted.

.changeset/dull-hornets-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
+Add `subscribeSync` method to ensure `Shape` up-to-date guarantees.
```
kevin-dp, line 5: Probably also want to add a sentence about the refactorings.

kevin-dp (approved):
Good work!
I left some comments but nothing major.
Perhaps, let's seek a third opinion (@KyleAMathews ) about the `subscribeSync` method vs not having that method and awaiting the callbacks instead (see my comment about that on L338 in `client.ts`).

packages/typescript-client/src/client.ts
```diff
@@ -57,86 +56,28 @@ export interface ShapeStreamOptions {
...
-    )
-    this.name = `FetchError`
-    this.status = status
-    this.text = text
-    this.json = json
-    this.headers = headers
-  }
+export interface ShapeStreamInterface<T extends Row = Row> {
+  subscribe(
+    callback: (messages: Message<T>[]) => MaybePromise<void>,
```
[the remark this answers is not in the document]
msfstef: I kept it as is for backwards compatibility - but also there's no reason not to permit sync callbacks in this method, it just doesn't have sync guarantees, so you can mix sync and async callbacks if you want

packages/typescript-client/src/client.ts
msfstef: I guess because the `subscribeOnceUpToDate` callback does not provide anything to process, and does not have any backpressure mechanism or underlying queue or anything like that, so it doesn't really matter

packages/typescript-client/src/client.ts
msfstef: I'm not sure how this would change any of the current behaviour? The underlying queue still processes them in order and awaits each processing call.

The issue is that if you want to reuse the state of the stream as the state of your data consumer, that is only possible if all of the processing happens synchronously, otherwise the stream consumer and stream will be out of sync with each other and the stream consumer needs to maintain its independent flags for when it is up to date etc.

packages/typescript-client/src/constants.ts
msfstef: Yes the header APIs are case insensitive but I'll convert them all to lower case either way for consistency

packages/typescript-client/src/offset.ts
msfstef: whoops sorry this is no longer used anywhere so I'm going to remove the whole module

packages/typescript-client/src/constants.ts
msfstef: If/when we use your proposed ones we can switch these - we'll probably keep both initially for backwards compatibility if we do anyway

thruflo:
One observation is that the Typescript client is a canonical implementation for other clients to reference. Looking at the source in `client.ts`, I didn't see any commenting of this `subscribe()` vs `subscribeSync()` distinction.

If we have these two methods, then, both in the source and the website docs (api -> clients -> typescript) it would be good to be clear about why there are two methods, what they do, when to use one over the other *and* whether this is a common pattern for other clients to follow.

packages/typescript-client/src/client.ts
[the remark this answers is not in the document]
msfstef: Resolving this as I removed the `subscribeSync` method

packages/typescript-client/src/client.ts
msfstef: Will be addressing this and with a new PR

msfstef:
@thruflo the `subscribeSync` was a tentative solution, I've rolled it back in this PR and the API remains exactly the same but handles sync/async more correctly - I'll open a new PR with a suggested behaviour change that we've discussed with @kevin-dp that would still not change the API itself but would give meaning to the type of callback you pass to `subscribe`; i.e. async callbacks would actually be awaited by the stream and provide backpressure, rather than what we do now where all the stream is collected in memory regardless of the speed of processing of the subscribers

thruflo:
Cool 👍 

I guess the "let's make sure we document clearly" point still applies to the alternative solution.

kevin-dp (approved):

PR #1716, opened 2024-09-16
fix: remove deprecated "version" attribute from docker compose files

beorn:
Seems safe to remove according to warning:
```console
% docker compose -f .support/docker-compose.yml up
WARN[0000] .../.support/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```

### DIFF
--- a/.support/docker-compose.yml
+++ b/.support/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.3"
 name: "electric_example-${PROJECT_NAME:-default}"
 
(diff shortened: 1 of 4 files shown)


KyleAMathews:
We've left it in them as it is still required on older versions of Docker. I'm sure we can remove it at some point but it's pretty harmless otherwise to keep supporting older versions.

Thanks for helping improve Electric!

Issue #1712, opened 2024-09-16
Bug: Wrong type cast leads to crash when starting


dustin-H:
At this line https://github.com/electric-sql/electric/blob/7765d50cbe57c9990245802366caee8bcd35d6ff/packages/sync-service/lib/electric/connection_manager.ex#L409 electric is reading the pg_setting `server_version` and casting it to a number.

However this leads to crashes as many postgres as a service providers are setting this value to e.g. `16.4 (Debian 16.4-1.pgdg1123+1)`.

Also it's quite wrong to expect a number as it is defined as string by postgres:

https://www.postgresql.org/docs/current/runtime-config-preset.html#GUC-SERVER-VERSION

KyleAMathews:
ah good catch! Thanks for the detailed bug report.

kevin-dp:
Hi @dustin-H,

Thanks for the report!
I was expecting the version number string to always be in the format `x.y` but as you explain this may not always be the case. As per your link to the documentation, the `server_version_num` setting seems more appropriate as it is guaranteed to be an integer (although stored as text). Thus, it seems safe to read `server_version_num` and cast it to an integer.

PR #1721, opened 2024-09-17
fix (sync-service): query PG version number using PG's `server_version_num` setting

kevin-dp:
This PR fixes https://github.com/electric-sql/electric/issues/1712.
We were previously reading out `server_version` which is a string and were casting it to an integer in order to get PG's major version number. However, casting to integer does not work when the `server_version` is set to e.g. `16.4 (Debian 16.4-1.pgdg1123+1)` which some providers do. This PR changes this such that we no longer cast the string but instead read the version number from the `server_version_num` setting which is an integer (cf. https://pgpedia.info/s/server_version_num.html and https://pgpedia.info/s/server_version_num.html).

### DIFF
--- a/.changeset/four-chicken-crash.md
+++ b/.changeset/four-chicken-crash.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix how Electric reads PG's version number.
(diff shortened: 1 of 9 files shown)


robacourt (approved):
Perfect ⭐

PR #1722, opened 2024-09-17
Remove password from state inspection

### LINKED ISSUE #1718: ConnectionManager stores password in clear text in its state
The password needs to be obfuscated such that when ConnectionManager crashes and its state is logged, the password is not visible in the log output.

magnetised:
Fixes #1718

### DIFF
--- a/.changeset/famous-actors-tell.md
+++ b/.changeset/famous-actors-tell.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Remove password from ConnectionManager logs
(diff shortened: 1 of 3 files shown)


alco:
@magnetised I have a different take on the same issue - https://github.com/electric-sql/electric/pull/1772. IMO, it's better to keep the password obfuscated in memory, makes it more resilient to future code changes where we rename the connection manager module or move the connection logic into a different one. It also works in cases where the Inspect protocol is not used to print the process state, such as the case with `:io.format('~p', [...])`, for example.

PR #1724, opened 2024-09-17
feat (client): expose time at which client last synced with Electric sync service

kevin-dp:
This PR extends the client to expose a `lastSyncedAt` field which indicates the last time at which the client synced with Electric.


robacourt (approved):
Great! Nice work, Kev. Matt at Trigger.dev will be interested in this as it was him that raised #1687

KyleAMathews (approved):
lovely!

PR #1726, opened 2024-09-17
Defer timeout handling to storage backends

### LINKED ISSUE #1725: Append transaction times out with a moderate number of shapes
This is a quick fix while we work out a better strategy long term.

Currently each tx is written to each active shape in lock-step: new txns are only requested from the replication stream once all shapes have written the last one to disk.

We currently have a problem with timeouts because the write function is done via a `GenServer.call/3` with a default timeout of 5s. So if there are a 1000 shapes, each taking ~5ms to write, then this call times out.

This is trivial to fix, instead of having an overall timeout for writing to all shapes, we move the timeout handling to the storage, so each individual shape will crash if its write takes too long, and the overall write call just sets a timeout of `:infinity` and waits.

magnetised:
Fixes #1725


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -37,8 +37,15 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+  #
+  # This `call/3` has a timeout of `:infinity` because timeouts are
+  # handled at the storage layer, that is this function doesn't
+  # assume any aggregate max time for the shape consumers to actually commit
+  # the new tx to disk, instead the storage backend is responsible for
+  # determinining how long a write should reasonably take and if that fails
+  # it should raise.
   def store_transaction(%Transaction{} = txn, server) do
-    GenStage.call(server, {:new_txn, txn})
+    GenStage.call(server, {:new_txn, txn}, :infinity)
```
KyleAMathews, line 48: does this also mean that we're no longer waiting to request the next transaction?

magnetised: No, unfortunately not. This is just a v. quick fix to make the current version more reliable while we work on that

KyleAMathews: cool 👍 step-by-step

robacourt (approved):
Yup, good quick fix 👍

PR #1729, opened 2024-09-17
chore (docs): fix open brace in TypeScript client docs

kevin-dp:
Fixes a broken open brace in the TypeScript client docs (it was using the unicode code for an open brace instead of open brace itself).


KyleAMathews (approved):
:shipit

robacourt (approved):

PR #1730, opened 2024-09-17
chore: Fix hex.pm release task

magnetised:


alco (approved):

magnetised:
@alco my publish task was in the wrong place, it would have run after every merge to main, which is wrong. have instead integrated into the `pnpm ci:publish` task used by the changeset tool

alco (approved):

PR #1731, opened 2024-09-17
feat: `ShapeStream` backpressure from subscriber processing callbacks

msfstef:
As discussed with @kevin-dp, the `ShapeStream` currently runs ahead to the end of the stream regardless of what its subscribers do with the messages.

This means that a stream will always accumulate all the messages from the network in memory regardless of the speed of processing of its subscribers, with no options for the developers to create backpressure and limit the rate at which the stream is read. This also leads to the side effect of the `ShapeStream` reaching an `isUpToDate` state that is misleading, since there is no guarantee that its subscribers will also be up to date.

The proposed solution is to get rid of all queuing mechanisms (less code, yay!) and make the `ShapeStream` wait for all subscribers to process the message batch before moving forward.

This gives the option to the developer to regulate backpressure - they can still let the stream run ahead as it does now by providing a synchronous callback that schedules the message processing e.g. in a queue that they maintain. But they can also put a subscriber in front that asynchronously stores data to disk, and thus the stream can run a little slower to avoid memory buildup.

Basically via this API we allow for developers to flexibly decide how fast the stream will try to retrieve data from the network in a relatively intuitive way.

Of course this should be more clearly documented both in the code and the docs, but I want to have people's opinion on whether this is something we want or note before doing so.


packages/typescript-client/src/client.ts
```diff
@@ -319,10 +318,12 @@ export class ShapeStream<T extends Row = Row>
     return !this.isUpToDate
   }
 
-  #publish(messages: Message<T>[]) {
-    this.#subscribers.forEach(([subscriber, _]) => {
-      subscriber.process(messages)
-    })
+  async #publish(messages: Message<T>[]): Promise<void> {
+    await Promise.allSettled(
```
samwillis, line 322: We may want this to be `Promise.all`. If a callback throws we probably want to stop the stream rather than continue blindly pushing data into a callback that may be broken?

```suggestion
    await Promise.all(
```

samwillis (commented):
Taking a look as I use this for the PGlite sync. One comment above.

packages/typescript-client/src/client.ts
msfstef: I'm trying to replicate the original behaviour which swallowed errors - the subscribe provides a callback whose errors the subscriber needs to handle themselves, the stream might have multiple subscribers

KyleAMathews: Yeah pglite can stop the stream by aborting it directly in case there's a pglite error

samwillis: `allSettled` silently eats all errors, doesn't even show them in the console, and they don't bubble up to a global error handler such as `window.addEventListener('unhandledrejection',  () => {})`.

In PGlite we used `queueMicrotask` to lift the callback out of the current control flow. So something like:

```ts
  async #publish(messages: Message<T>[]): Promise<void> {
    await Promise.all(
      Array.from(this.#subscribers.values()).map(([callback, __]) =>
        queueMicrotask(() => callback(messages))
      )
    )
  }
```

samwillis: Although that now doesn't block while processing the callback

We can log the error, but again that doesn't bubble them to a global error handler:

```ts
  async #publish(messages: Message<T>[]): Promise<void> {
    await Promise.all(
      Array.from(this.#subscribers.values()).map(([callback, __]) =>
        try {
          callback(messages)
        } catch (error) {
          console.error(error)
        }
      )
    )
  }
```

samwillis: Final note, in the old electric we had code like this using `Promise.allSettled` for a callback and it silently eats errors in our CI causing a bug that was super hard to track down.

msfstef: With the `subscribe` call clients provide also an `onError` callback, but that's meant to provide actual stream errors, like network errors etc, not errors bubbled up from the callback itself, so I'm ruling out using that.

Other than that using `Promise.all` with a try catch block seems reasonable to me - errors should at least pop up in the console, I agree with that.

kevin-dp: So, would we support an optional `onError` callback to be passed to `subscribe`?
If it's provided we do `await callback().catch(onError)` and if no error callback is provided we re-throw or at least log the error.

kevin-dp (approved):
Looks good to me, love the simplicity!
Let's sort out the behavior we want if one of the callbacks throws before we merge this.

KyleAMathews (changes requested):
I like this idea — especially that we wait to tell users that the stream is up-to-date until all subscribers have finished processing data — that'll be really important to keep e.g. pglite consistent with other consumers of data e.g. people might have useShape & pglite going in the same app.

I don't like though that we now wait to start fetching until the previous messages are consumed. Backpressure is used to moderate the buffer — but this eliminates the buffer altogether and will slow down processing large shapes with many chunks. I think reasonable back pressure is we don't get more than one fetch ahead of processing. Keeping some buffer data in memory is fine.

packages/typescript-client/src/client.ts
msfstef: Subscribe already has an `onError` callback but it surfaces errors not related to the processing callback itself - if the developer wants to handle the callback errors they can wrap it in a try...catch statement and decide if they want to stop the subscription to the stream etc

I think we should just log the errors to console via `Promise.all` as I think that was the issue with `Promise.allSettled`

samwillis: I think the requirements should be that error in a callback is both logged to the console, but also makes it to any global error handler (which naturally gets logged to the console).

We don't need to handle that error, or pass it to an error handler. The user can internally wrap the callback with a try.

What we must not do is silently eat errors, and ideally let them bubble to the global handler.

KyleAMathews:
Sure they can maintain their own buffer. But my point is that it's also a good idea for our client to maintain a buffer as well of one fetched batch of messages.

KyleAMathews:
sounds good 👍 

I'm not suggesting by any means we do another refactor right now but something to look at and ponder is if we should pull in https://effect.website/ at some point — managing all these flows and error states, etc. might get easier.

packages/typescript-client/src/client.ts
msfstef: doing `Promise.all`with `try...catch` and bubbling error up with `queueMicrotask(() => { throw err })` - resolving this

packages/typescript-client/src/client.ts
```diff
@@ -319,10 +318,18 @@ export class ShapeStream<T extends Row = Row>
...
-      subscriber.process(messages)
-    })
+  async #publish(messages: Message<T>[]): Promise<void> {
+    await Promise.all(
+      Array.from(this.#subscribers.values()).map(async ([callback, __]) => {
+        try {
+          await callback(messages)
+        } catch (err) {
+          queueMicrotask(() => {
+            throw err
```
KyleAMathews, line 328: how would a user catch this error?

samwillis: With a global window.onerror or unhandledpromisererejection.
Key thing is that it will show in console and get logged by tools such as Sentry.

KyleAMathews: Yeah I guess that works. This is just an escape hatch. Generally you'd catch and handle the errors in the actual subscriber code.

samwillis: Yep exactly. It's to stop them from being silently eaten.

Issue #1706, opened 2024-09-16
Document limitations of Postgres Logical Replication


KyleAMathews:
E.g. like Debezium does here: https://debezium.io/documentation/reference/stable/connectors/postgresql.html

>
> The connector relies on and reflects the PostgreSQL logical decoding feature, which has the following limitations:
>

> Logical decoding does not support DDL changes. This means that the connector is unable to report DDL change events back to consumers.
> 
> Logical decoding replication slots are supported on only primary servers. When there is a cluster of PostgreSQL servers, the connector can run on only the active primary server. It cannot run on hot or warm standby replicas. If the primary server fails or is demoted, the connector stops. After the primary server has recovered, you can restart the connector. If a different PostgreSQL server has been promoted to primary, adjust the connector configuration before restarting the connector.
> 
> Because logical decoding replication slots publish changes during commit — and not post commit — undesirable side-effects can occur. There are two main scenarios when clients can observe inconsistent states. First, publishing uncommitted changes when the master dies before replication completes. Second, publishing changes that cannot be read (i.e., read-after-write consistency) temporarily because they are being replicated. For example, an EmbeddedEngine consumer receives a notification of a row that was created but it cannot be read by a transaction.
> 
> Additionally, the pgoutput logical decoding output plug-in does not capture values for generated columns, resulting in missing data for these columns in the connector’s output.

PR #1732, opened 2024-09-17
docs: document shape deletion on table drop

kevin-dp:
Extend the shape docs to explain that shapes are not deleted automatically when tables are dropped.
Fixes https://github.com/electric-sql/electric/issues/1706.

### DIFF
--- a/website/docs/guides/shapes.md
+++ b/website/docs/guides/shapes.md
@@ -173,3 +173,8 @@ You can upvote and discuss adding support for mutable shapes here:
 ... add links to benchmarks here ...
 
 -->
+
+### Dropping tables
+
+When dropping a table from Postgres you need to *manually* delete all shapes that are defined on that table.
+Electric doesn't automatically delete the shape because Postgres does not stream DDL statements (such as `DROP TABLE`) on the logical replication stream that Electric uses to detect changes.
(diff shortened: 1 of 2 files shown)


website/docs/guides/shapes.md
```diff
@@ -173,3 +173,8 @@ You can upvote and discuss adding support for mutable shapes here:
 ... add links to benchmarks here ...
 
 -->
+
+### Dropping tables
+
+When dropping a table from Postgres you need to *manually* delete all shapes that are defined on that table.
```
KyleAMathews, line 179: I agree we don't need to handle this immediately but it'd be nice to eventually automatically handle it. Perhaps create an issue with our different ideas and link to it from here?

Also this doesn't quite get at the problem of dropping a table. The problem (as @lirbank experienced) is that old shapes still have data from the rows in the table. It's more "drop to truncate and then add new data" where people will see problems. If they just drop the table and then don't recreate it, there'll be stale shapes but presumably people won't care as they won't try to use them again.

kevin-dp: I've improved the explanation. Can you have a look if it's better now? :)

KyleAMathews (approved):
Looks great!

Issue #1664, opened 2024-09-10
Add initial spans for opentelemetry


KyleAMathews:
We have the initial setup of opentelemetry done but let's now add the following basic spans & attributes (not exhaustive, whoever does this feel free to add more that seem right):

- shape http request
  - attributes:
    - `url`
    - query parameters
    - and other http info e.g. requester, status code that we return, body size, etc. etc.
- sql query when creating snapshot
  - attributes:
    - shape_id
    - the full query & parts of it broken up e.g. the table, select clause, where clause, etc.

KyleAMathews:
A few notes from testing just now:

We'll want `is_cached`, `is_live`, and `offset` attributes added to shape handling spans.

For replication log spans, we need spans created for the code that updates the shape & sends it to any subscribers. That way we can see the tree of a replication log coming to Electric & then all subsequent work to update shapes & subscribers.

KyleAMathews:
I also think we should just have one shape for all shape requests — a simpler name would be nice too — e.g. `ShapeGet`. We could then have `ShapeDelete` (or `shape_get` and `shape_delete`).

KyleAMathews:
We need an attribute to distinguish between requests that go into long polling or just immediately respond. `is_live` isn't sufficient as if a user is behind the tip of the log they get an immediate response. Perhaps a `is_long_polling` attribute.

KyleAMathews:
We need attributes about the shape — `table`, `shape_def` which includes the table & where clause and any other shape definition.

thruflo:
`is_immediate_response`?

KyleAMathews:
> is_immediate_response?

yeah that works too — true for normal responses and false for long polling

PR #1736, opened 2024-09-17
chore(sync-service): Expand the set of OT span attributes assigned in ServeShapePlug

alco:
Part of #1664.

@KyleAMathews WIP so far:
Screenshot from 2024-09-17 17-33-51
Screenshot from 2024-09-17 17-34-08


KyleAMathews:
Looks great!

alco:
@KyleAMathews Just pushed new changes that largely implement the ideas described in #1664.

The name of the root span for shape requests is now `ShapeGet`:
Screenshot from 2024-09-24 04-05-56

Grouping by `shape_req.is_cached`:
Screenshot from 2024-09-24 04-05-24

Serving a largish initial snapshot or 2000 linearlite issues:
Screenshot from 2024-09-24 04-11-48


Bits that are still no there:
- [ ] propagating tracing context on the write path, all the way from Postgres to individual shape logs
- [ ] recording exceptions in ShapeGet processing as errors in the OT span (can be done by using `Plug.ErrorHandler`)
- [ ] estimating shape size and assigning it as an attribute to tracing spans. This is doable by keeping track of how much data has been streamed to the client and then setting that as a dynamic attr on the span.

KyleAMathews:
- is `ShapeGet.load_shape_data` when we run the actual SQL query? It'd be nice to have it clear what's the query. Also adding the query as an attribute would be great too.
- `ShapeGet` is camel case while all the sub-spans are snake case — perhaps just do `shape_get` as well?

Nice progress!

alco:
> * is `ShapeGet.load_shape_data` when we run the actual SQL query? It'd be nice to have it clear what's the query. Also adding the query as an attribute would be great too.

You're probably referring to `load_shape_info`. This span covers the "get_or_create_shape_id" function, so it's going to be cached data after the shape is first created. It makes sense to only trace the shape creation bit. I'll try adding the SQL query to that.
 
> * `ShapeGet` is camel case while all the sub-spans are snake case — perhaps just do `shape_get` as well?

Naming convention can be anything we like here. I kinda like the way the root span's name stands out. Also makes it easy to see when a non-root span shows up as a root one in Honeycomb because of incorrect context tracking in the code.

So I would make all root spans use CamelCase. But we could use some other convention to keep everything snake case, e.g. `_shape_get` or `R_shape_get` for the root span.

KyleAMathews:
> I kinda like the way the root span's name stands out

Hmm yeah interesting point. I don't feel that strongly about specifics of the conventions — just that we're consistent.

robacourt (approved):
Great work!

Issue #1737, opened 2024-09-19
Rename HTTP headers without the deprecated X- prefix


kevin-dp:
The sync service uses custom headers starting with `x-`, however, this convention has been deprecated since 2012 and is actually advised against doing that (cf. https://datatracker.ietf.org/doc/html/rfc6648#page-4). 

> 3.  Recommendations for Creators of New Parameters
> 
>    Creators of new parameters to be used in the context of application
>    protocols:
>    1.  SHOULD assume that all parameters they create might become standardized, public, commonly deployed, or usable across multiple implementations.
>    2.  SHOULD employ meaningful parameter names that they have reason to believe are currently unused.
>    3.  SHOULD NOT prefix their parameter names with "X-" or similar constructs.
> 
>    Note: If the relevant parameter name space has conventions about associating parameter names with those who create them, a parameter  name could incorporate the organization's name or primary domain name (see Appendix B for examples).

We already follow their advise of incorporating our organisation name, so we should rename headers like `x-electric-shape-id` to `electric-shape-id`.

PR #1739, opened 2024-09-23
chore!: remove x- prefix from custom HTTP headers

kevin-dp:
This PR removes the x- prefix from Electric's custom headers, as essentially the x- prefix has been deprecated for a while. **This is a breaking change** as it changes the names of the headers. Fixes https://github.com/electric-sql/electric/issues/1737.


KyleAMathews (approved):
Thanks for finding this!

PR #1742, opened 2024-09-23
Add a SQLite backed storage implementation

magnetised:
Alternative to the hybrid file + cubdb storage implementation.

Currently implemented with both snapshots and log written to sqlite, but we could always go back to the hybrid model if this is an issue.

To improve test reliability I upped some waits in the typescript client test. This is not because of slowness in the sqlite implementation afaict because the same tests were failing with the old file storage backend.

Naive benchmarking puts this at ~10-100x faster to append to the log than the cubdb version.


packages/sync-service/config/runtime.exs
```diff
@@ -127,6 +127,9 @@ chunk_bytes_threshold =
     Electric.ShapeCache.LogChunker.default_chunk_size_threshold()
   )
 
+default_file_storage_impl = Electric.ShapeCache.FileStorage
+# default_file_storage_impl = Electric.ShapeCache.SQLiteStorage
```
magnetised, line 131: was hesitant to make this the default but then did make sqlite the default for the tests, so maybe should have just gone for it...

magnetised: in fact maybe I should add an option to the env... on it

magnetised:
> hey, this looks nice! Would love to see the performance of this with our benchmarks. Particularly interested in knowing how it handles lots of shapes and lots of writes

It's basically a dupe of the cubdb storage, but with sqlite as the db and replacing the pure file-dump snapshot with a table in sqlite. Each shape will have its own db so this is basically the same characteristics as for cubdb but with faster writes (~10-100 microseconds for a log append as opposed to ~4ms)

>     * Do you handle log compaction, or have an idea on how we'd implement it?

No, haven't touched it, haven't thought about it. We could do it with say replacing within a range in the db. with sqlite we have the advantage of having multiple tables within the file, so there may be some kind of dupe table, compact, rename or overwrite table dance within isolated snapshots. we could use multiple concurrent tables and have some pointer? dunno, lots of options though
 
>     * The same questions for chunking. Is that just a range query :D?

I've replicated the cub db version, with chunk log offsets in a separate table and then the same log stream offset stuff.

packages/typescript-client/test/integration.test.ts
```diff
@@ -548,7 +548,7 @@ describe(`HTTP Sync`, () => {
       { title: `foo8` }
     )
     // Wait for server to get all the messages.
-    await sleep(40)
+    await sleep(100)
```
msfstef, line 551:
```suggestion
    await vi.waitFor(async () => {
      const res = await fetch(
        `${BASE_URL}/v1/shape/${issuesTableUrl}?offset=-1`
      )
      const body = (await res.json()) as Message[]
      expect(body.length).greaterThan(1)
    })
```

We use this approach elsewhere, probably better than fiddling with timeouts

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
```diff
@@ -0,0 +1,519 @@
...
+  end
+
+  def for_shape(shape_id, storage_opts, opts) do
+    %{base_path: base_path, electric_instance_id: electric_instance_id} =
+      storage_opts
+
+    path = mkdir!([base_path, safe_instance_id(electric_instance_id), shape_id, "data.sqlite3"])
+
+    mode =
+      if Keyword.get(opts, :readonly, false) do
```
msfstef, line 54: Why not make this configuration part of the `storage_opts` and instead introduce an additional `opts` arg? The `storage_opts` are already different per storage implementation and we don't currently have a per shape options requirement?

packages/sync-service/lib/electric/replication/log_offset.ex
```diff
@@ -128,7 +131,18 @@ defmodule Electric.Replication.LogOffset do
       :lt
   """
   @spec last() :: t
-  def last(), do: %LogOffset{tx_offset: 0xFFFFFFFFFFFFFFFF, op_offset: :infinity}
+  def last(), do: %LogOffset{tx_offset: @last_tx, op_offset: @last_op}
+
+  @doc """
+  Tests to see if this is the `last/0` offset.
+  """
+  @spec last?(t()) :: boolean()
```
msfstef, line 139: nit: a doctest for good measure?

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
```diff
@@ -0,0 +1,519 @@
...
+      {:ok, row}
+    else
+      {:error, _reason} = error -> error
+      {:read, :done} -> :empty
+    end
+  end
+
+  defp update_row(%S{conn: conn}, query, params) do
+    with {:ok, statement} <- Sqlite3.prepare(conn, query),
+         :ok <- Exqlite.Sqlite3.bind(conn, statement, params),
```
msfstef, line 480:
```suggestion
         :ok <- Sqlite3.bind(conn, statement, params),
```

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
```diff
@@ -0,0 +1,519 @@
...
+    else
+      {:error, _reason} = error -> error
+      {:read, :done} -> :empty
+    end
+  end
+
+  defp read_row(%S{conn: conn}, query, params) do
+    with {:ok, stmt} <- Sqlite3.prepare(conn, query),
+         :ok = Sqlite3.bind(conn, stmt, params),
+         {:read, {:row, row}} <- {:read, Sqlite3.step(conn, stmt)},
```
msfstef, line 468: Why not this?
```suggestion
         {:row, row} <- Sqlite3.step(conn, stmt),
```

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
```diff
@@ -0,0 +1,519 @@
...
+  end
+
+  defp mkdir!(parts) when is_list(parts) do
+    path = Path.join(parts)
+    dir = Path.dirname(path)
+    File.mkdir_p!(dir)
+    path
+  end
+
+  def select(%S{conn: conn}, query) do
```
msfstef, line 500: nit: this is public and unused (?)

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
```diff
@@ -0,0 +1,519 @@
...
+    else
+      raise "Snapshot no longer available"
+    end
+  end
+
+  @impl Electric.ShapeCache.Storage
+  def append_to_log!(log_items, %S{} = storage) do
+    %{conn: conn, stmts: %{append_stmt: append_stmt, chunk_stmt: chunk_stmt}} = storage
+
+    Enum.reduce(log_items, {append_stmt, chunk_stmt}, fn
```
msfstef, line 351: should this all be done in a transaction? also, could it potentially be more efficient to split boundaries and log entries and write them sequentially (although I'm assuming that if it is done transactionally writes are optimized)?

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
```diff
@@ -0,0 +1,519 @@
...
+        LogOffset.new(off_tx, off_op)
+
+      :empty ->
+        nil
+    end
+  end
+
+  @impl Electric.ShapeCache.Storage
+  def cleanup!(%S{conn: conn}) do
+    stmts = [
```
msfstef, line 439: should this all be done transactionally?

msfstef (approved):
Looks good! Left a few questions here and there, one point about potentially wanting transactional writes for some things?

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
magnetised: I wanted to be absolutely sure that `else` condition below was definitely catching the case where we're reading from an empty db. Might be overkill but I wanted an explicit condition

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
magnetised: I'll leave some docs or something. It's just a way to query sqlite as a stream, otherwise there's no builtin way to get a list of results for a query and that will be annoying when debugging etc

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
msfstef: aah yeah that makes sense thank you - might be worth adding a short comment here with a reference to the 1 writer multiple readers that make this necessary

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
magnetised: TX around this is a good idea. I started with txns but removed them to allow for partial reads from a snapshot but you're right.

No idea about splitting the two writes.

kevin-dp:
@balegas 

>     * Do you handle log compaction, or have an idea on how we'd implement it?

@marc-shapiro, @preguica, and I had a discussion regarding the format of the shape log and log compaction during last last week's on-site. Currently, we store the shape as a sequence of changes, where every change has an offset. After compaction, every PK will have a single entry in the log with the offset of the latest write (insert/update/delete) to that PK. The format where the shape log is stored as a sequence of changes seems good for storing in files on disk.

However, we can have an alternative representation for the shape log, that is more appropriate for storing in a DB. We could have a table per shape that for every PK that is part of the shape tracks its latest value and the "time"/"offset" at which that value was written. When a row is updated or deleted, we update its value and the offset in that shape's table. When a user requests a shape, starting at a given offset, we query the table for all rows with an offset > the provided offset.

The advantage of this approach is that we only ever store the compacted log (as for updates/deletes we effectively update the entry for that row in the DB). Another advantage is that we don't have to implement and maintain our own file-based storage system which is probably quite complex if we want to properly optimize that.

robacourt (approved):
Looks good to me. Nice work! I'm just in the process of running some benchmarks for it

packages/typescript-client/test/integration.test.ts
magnetised: I think this is a good idea but maybe not for this PR.

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
```diff
@@ -0,0 +1,539 @@
...
+
+        {:ok, stmt} = Sqlite3.prepare(conn, query)
+        :ok = Sqlite3.bind(conn, stmt, params)
+
+        {:stream, conn, stmt}
+      end,
+      fn
+        {:stream, conn, stmt} ->
+          case Sqlite3.multi_step(conn, stmt, @snapshot_read_row_count) do
+            {:rows, rows} -> {List.flatten(rows), {:stream, conn, stmt}}
```
kevin-dp, line 424: Wouldn't it be safer if we explicitly order the rows based on their offset?
This only works because the rows are inserted in the order of their offsets.
Also, i find it a bit strange that we are inserting a row per offset, just for the purpose of being able to find the end offset of a chunk.
Can't we save space by keeping only 1 row per `offset_tx` and always updating the `offset_op` to the largest `offset_op` we have seen for that `offset_tx` ?

kevin-dp (changes requested):
Great job, i like moving the storage to SQLite!
I left some comments in the code as well as a comment on this PR wrt the format of the shape log we store. While the shape log was a good format for storing e.g. in a file, i think it is not a good format for storing in a relational database. Instead, we could modify the format to be more like a dictionary (cf. my other comment). Ultimately, we would have to benchmark that change to see which format performs best.

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
msfstef: @kevin-dp that would mean that you cannot chunk transactions, which is definitely something we want to be able to do?

kevin-dp: Not sure why not? This function's name suggests that it returns the last offset in the chunk, but i don't see any chunking related logic here. It seems to just return the next offset that is in the log.

packages/sync-service/lib/electric/shape_cache/file_storage.ex
```diff
@@ -27,11 +27,17 @@ defmodule Electric.ShapeCache.FileStorage do
   end
 
   @impl Electric.ShapeCache.Storage
-  def for_shape(shape_id, %FS{shape_id: shape_id} = opts) do
-    opts
+  def for_shape(shape_id, storage, opts \\ [])
```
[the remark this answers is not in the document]
magnetised: @kevin-dp this pattern, where you define the function head but no body, allows you to define defaults for args. you need this when you're pattern matching on the args in other function heads.

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
magnetised: > Wouldn't it be safer if we explicitly order the rows based on their offset?

The query does that, 

```elixir
"""                             
SELECT data                     
  FROM #{@log_table} #{where}   
  ORDER BY offset_tx, offset_op 
"""
```

magnetised: @kevin-dp this is just a copy of the implementation for cubdb backed by sqlite. there may well be optimisations we can do on it to make it even faster but let's just figure out if the approach is faster for the same impl before we start optimising the details.

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
```diff
@@ -0,0 +1,539 @@
...
+  @get_log_stream_range_query log_stream_query.("""
+                              WHERE ((offset_tx > ?1) OR (offset_tx = ?2 AND offset_op > ?3))
+                                AND ((offset_tx < ?4) OR (offset_tx = ?5 AND offset_op <= ?6))
+                              """)
+
+  @snapshot_read_row_count 100
+
+  @impl Electric.ShapeCache.Storage
+  def get_log_stream(offset, max_offset, %S{conn: conn}) do
+    %LogOffset{tx_offset: tx_offset_min, op_offset: op_offset_min} = offset
```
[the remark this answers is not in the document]
magnetised: > But then numbering them makes it very confusing, would be less confusing if we use ? for all of them.

I disagree. I think numbering them makes it all very explicit and has the benefit of counting the params for us. 

However, i did try removing the duplication of the params. I didn't bother before, despite thinking about it, because, well, I got lazy as it seemed a very minor thing and I had other fish to fry. But it works if you have two `?1` params in the query. 

Will push an update

packages/sync-service/lib/electric/replication/log_offset.ex
```diff
@@ -128,7 +131,18 @@ defmodule Electric.Replication.LogOffset do
...
+  def last?(%LogOffset{tx_offset: @last_tx, op_offset: @last_op}), do: true
```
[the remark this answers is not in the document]
magnetised: with that approach you put a little more stress on the GC as you're creating a new struct that will need to be garbage collected later. with my fiddly pattern matching version there's no garbage. i worry about the garbage.

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
msfstef: @kevin-dp the `chunk_offset_table`, if that's what you're referring to, doesn't add an entry for every row, it adds an entry for every "chunk checkpoint" in order to be able to determine the chunk boundaries, so it is quite sparse.

I might be getting confused with what you're referring to, we can sort it out on a call.

packages/sync-service/lib/electric/replication/log_offset.ex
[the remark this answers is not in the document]
kevin-dp: isn't Elixir smart enough to recognize that this is in fact a constant and that only one `last()` struct should ever exist?

packages/sync-service/lib/electric/shape_cache/sqlite_storage.ex
kevin-dp: @magnetised right, my bad. Had missed the order by clause.

balegas:
We've decided not to move on with this approach while we have other performance gains to work on besides the speed of the storage backend. I'm closing the pull request.

Issue #1447, opened 2024-07-25
Only send up-to-date control header for responses with short cache timeouts


KyleAMathews:
Right now I believe we're sending an `up-to-date` control message at the end of every log (correct me if wrong).

This is problematic as the default cache timeout is 60 seconds. So someone could be as much as 59 seconds stale while thinking they're live.

A potential pattern for solving this:
1. -1 offset gets long cache value to optimize cache hits for bulk of data e.g. even 1hr.
2. then vary the length of the cache timeout based on how old the offset on the request is. E.g. if the offset is 30 minutes out of date, set the cache to 30s. If the offset is < 10m then set the cache to 1s.
3. return `up-to-date` if the offset is < 10m or if there's nothing newer than the offset.

This would mean for most initial syncs, we'd do 2 (and occasionally 3) requests to get to `up-to-date`. Since they're all from the CDN, it'll be pretty fast. And even with a 1s cache timeout, the CDN will still handle 99% of the load as with `stale-while-revalidate`, the CDN is the only one going back to the origin to get new data.

msfstef:
This raises a more general issue - the same chunk of the log might contain an `up-to-date` message or it might not, but it gets assigned the same `etag` and will be treated as the same chunk of data cache-wise.

One easy workaround that moves the responsibility to the client would be to include a timestamp in the `up-to-date` control message - that way even if a cached chunk is retrieved by the client they examine the timestamp on the `up-to-date` message, which was assigned by the original server response, and decide whether they want to treat it as out of date (client-configurable if we want). I think this might be good to have regardless.

Other than that I do think that we might want a layer for determining cache control headers for chunks served depending on various factors (is it a "full" chunk? is it an up-to-date chunk? etc) I'm happy to pick this up.

KyleAMathews:
@msfstef and I chatted a bit about this in private chat — the basic idea is to figure out an algorithm that balances freshness and cache hits. Reduce the server work to update caches in http proxies to a bare minimum. Also that clients that don't want to hit the live endpoint have a guarantee around how fresh the cache is.

PR #1743, opened 2024-09-23
feat: Chunk prefetching in `ShapeStream`

msfstef:
Followup to https://github.com/electric-sql/electric/pull/1731

Implemented a fetch wrapper that does not wait for the chunk bodies to be streamed to the client and processed by the subscribers, and instead prefetches a number of chunks from the network as soon as the available headers to determine the next chunk are available.

I also had to implement another response header on our api called `electric-chunk-up-to-date`, similarly to how we use `electric-chunk-last-offset`, which allows us to know all the necessary chunk metadata without having to read the actual body. If we do not know if a chunk is an `up-to-date` chunk, we don't know whether we should request the next one, and whether its caching policy should be adjusted on the CDN.

I've set the default for prefetching to 2 chunks, but we could increase this to the most commen [maximum browser connections to the same domain](https://www.geeksforgeeks.org/what-are-max-parallel-http-connections-in-a-browser/) in order to maximise parallelization.

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -274,6 +274,8 @@ defmodule Electric.Plug.ServeShapePlug do
     if LogOffset.compare(chunk_end_offset, last_offset) == :lt do
       conn
       |> assign(:up_to_date, [])
+      # header might have been added on first pass but no longer valid
+      # if listening to live changes and an incomplete chunk is formed
       |> delete_resp_header("electric-chunk-up-to-date")
     else
       conn
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -128,6 +128,7 @@ defmodule Electric.Plug.ServeShapePlug do
   # asked for last offset
   plug :listen_for_new_changes
   plug :determine_log_chunk_offset
+  plug :determine_up_to_date
   plug :generate_etag
   plug :validate_and_put_etag
   plug :put_resp_cache_headers
@@ -264,6 +265,23 @@ defmodule Electric.Plug.ServeShapePlug do
     |> put_resp_header("electric-chunk-last-offset", "#{chunk_end_offset}")
   end
 
+  defp determine_up_to_date(
+         %Conn{
+           assigns: %{chunk_end_offset: chunk_end_offset, last_offset: last_offset}
+         } = conn,
+         _
+       ) do
+    if LogOffset.compare(chunk_end_offset, last_offset) == :lt do
+      conn
+      |> assign(:up_to_date, [])
+      |> delete_resp_header("electric-chunk-up-to-date")
+    else
+      conn
+      |> assign(:up_to_date, [@up_to_date])
+      |> put_resp_header("electric-chunk-up-to-date", "true")
+    end
+  end
+
   defp generate_etag(%Conn{} = conn, _) do
     %{
       offset: offset,
@@ -328,7 +346,8 @@ defmodule Electric.Plug.ServeShapePlug do
            assigns: %{
              offset: @before_all_offset,
              chunk_end_offset: chunk_end_offset,
-             active_shape_id: shape_id
+             active_shape_id: shape_id,
+             up_to_date: maybe_up_to_date
            }
          } = conn,
          _
@@ -346,7 +365,7 @@ defmodule Electric.Plug.ServeShapePlug do
                   up_to: chunk_end_offset
                 )
 
-              [snapshot, log, maybe_up_to_date(conn)]
+              [snapshot, log, maybe_up_to_date]
               |> Stream.concat()
               |> to_json_stream()
               |> Stream.chunk_every(500)
@@ -372,7 +391,8 @@ defmodule Electric.Plug.ServeShapePlug do
            assigns: %{
              offset: offset,
              chunk_end_offset: chunk_end_offset,
-             active_shape_id: shape_id
+             active_shape_id: shape_id,
+             up_to_date: maybe_up_to_date
            }
          } = conn,
          _
@@ -390,7 +410,7 @@ defmodule Electric.Plug.ServeShapePlug do
         if Enum.take(log, 1) == [] and conn.assigns.live do
           hold_until_change(conn, shape_id)
         else
-          [log, maybe_up_to_date(conn)]
+          [log, maybe_up_to_date]
           |> Stream.concat()
           |> to_json_stream()
           |> Stream.chunk_every(500)
@@ -437,16 +457,6 @@ defmodule Electric.Plug.ServeShapePlug do
     end)
   end
 
-  defp maybe_up_to_date(%Conn{
-         assigns: %{chunk_end_offset: chunk_end_offset, last_offset: last_offset}
-       }) do
-    if LogOffset.compare(chunk_end_offset, last_offset) == :lt do
-      []
-    else
-      [@up_to_date]
-    end
-  end
-
   defp listen_for_new_changes(%Conn{} = conn, _) when not conn.assigns.live, do: conn
 
   defp listen_for_new_changes(%Conn{assigns: assigns} = conn, _) do
@@ -478,6 +488,7 @@ defmodule Electric.Plug.ServeShapePlug do
         |> assign(:chunk_end_offset, latest_log_offset)
         # update last offset header
         |> put_resp_header("electric-chunk-last-offset", "#{latest_log_offset}")
+        |> determine_up_to_date([])
         |> serve_log_or_snapshot([])
 
       {^ref, :shape_rotation} ->
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -280,7 +280,7 @@ defmodule Electric.Plug.ServeShapePlug do
     else
       conn
       |> assign(:up_to_date, [@up_to_date])
-      |> put_resp_header("electric-chunk-up-to-date", "true")
+      |> put_resp_header("electric-chunk-up-to-date", "")
     end
   end
 
--- a/packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
+++ b/packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
@@ -269,6 +269,8 @@ defmodule Electric.Plug.ServeShapePlugTest do
       assert Plug.Conn.get_resp_header(conn, "electric-chunk-last-offset") == [
                "#{next_next_offset}"
              ]
+
+      assert Plug.Conn.get_resp_header(conn, "electric-chunk-up-to-date") == []
     end
 
     test "returns 304 Not Modified when If-None-Match matches ETag" do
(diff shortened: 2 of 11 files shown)


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -264,6 +265,23 @@ defmodule Electric.Plug.ServeShapePlug do
...
+  defp determine_up_to_date(
+         %Conn{
+           assigns: %{chunk_end_offset: chunk_end_offset, last_offset: last_offset}
+         } = conn,
+         _
+       ) do
+    if LogOffset.compare(chunk_end_offset, last_offset) == :lt do
+      conn
+      |> assign(:up_to_date, [])
+      |> delete_resp_header("electric-chunk-up-to-date")
```
kevin-dp, line 277: `delete_resp_header` is only used in live queries, right? Perhaps worth briefly documenting this.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -264,6 +265,23 @@ defmodule Electric.Plug.ServeShapePlug do
...
+    else
+      conn
+      |> assign(:up_to_date, [@up_to_date])
+      |> put_resp_header("electric-chunk-up-to-date", "true")
```
kevin-dp, line 283: do we need to put this to `true`? Can we just not assign a value to it or if not possible just the empty string?
The presence of this header serves as a flag so removing the value or making it smaller minimizes what needs to be sent over the network.

packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
```diff
@@ -358,6 +360,7 @@ defmodule Electric.Plug.ServeShapePlugTest do
              ]
 
       assert Plug.Conn.get_resp_header(conn, "electric-chunk-last-offset") == [next_offset_str]
+      assert Plug.Conn.get_resp_header(conn, "electric-chunk-up-to-date") == ["true"]
```
kevin-dp, line 363: We should also test that `electric-chunk-up-to-date` header is correctly removed when we are in live mode and a TX comes in which is split into several chunks (only the last chunk should have the header).

packages/typescript-client/src/client.ts
```diff
@@ -141,18 +142,24 @@ export class ShapeStream<T extends Row = Row>
...
+
+    const fetchWithBackoffClient = createFetchWithBackoff(baseFetchClient, {
+      ...(options.backoffOptions ?? BackoffDefaults),
+      onFailedAttempt: () => {
+        this.#connected = false
+        options.backoffOptions?.onFailedAttempt?.()
+      },
+    })
+
+    const fetchWithPrefetchAndBAckoffClient = createFetchWithChunkBuffer(
```
kevin-dp, line 157: Typo in this variable's name. The `A` in `BAckoff` shouldn't be capitalized.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
msfstef: Added a comment

packages/typescript-client/src/client.ts
kevin-dp: This variable is only used once so there's no real need to assign it here, can immediately assign it to `this.#fetchClient`.

packages/typescript-client/src/fetch.ts
```diff
@@ -77,3 +85,120 @@ export function createFetchWithBackoff(
...
+    if (!nextUrl || prefetchMap.has(nextUrl)) return result
+
+    // prefetch next response and return current one using the prefetch
+    // client to allow for subsequent responses to be prefetched
+    try {
+      const prefetchPromise = prefetchClient(nextUrl, args[1])
+      prefetchPromise.catch(() => prefetchMap.delete(nextUrl))
+      prefetchMap.set(nextUrl, prefetchPromise)
+    } catch (_) {
+      // ignore prefetch errors
```
kevin-dp, line 156: aren't the errors already caught by `prefetchPromise.catch(...)`?

packages/typescript-client/src/fetch.ts
```diff
@@ -77,3 +85,120 @@ export function createFetchWithBackoff(
...
+  const prefetchEntryClient = (
+    ...args: Parameters<typeof fetch>
+  ): Promise<Response> => {
+    const url = args[0].toString()
+
+    // if already prefetched, serve that and delete from map
+    const prefetchRes = prefetchMap.get(url)
+    if (prefetchRes) {
+      prefetchMap.delete(url)
+      return prefetchRes
```
kevin-dp, line 171: Since we serve a prefetched result, don't we need to prefetch the next one now?

packages/typescript-client/src/fetch.ts
```diff
@@ -77,3 +85,120 @@ export function createFetchWithBackoff(
...
+    }
+
+    // otherwise clear current prefetched responses (and abort active requests)
```
kevin-dp, line 174: let's clarify here in the comment why we need to clear the prefetched responses.
I'm guessing because what is being requested does not match what has been prefetched so something's off?

kevin-dp (changes requested):
Nice work!
I left some comments, nothing major.

packages/typescript-client/src/fetch.ts
msfstef: The async errors from the `prefetchPromise` are caught by the `.catch` clause, but any sync errors caused before the promise is even returned are caught by the `try...catch`

packages/typescript-client/src/fetch.ts
msfstef: Aaah very good point - I'll add a test for this and implement the right thing

packages/typescript-client/src/fetch.ts
msfstef: Yeah since we're using a map and also rely on the map size for the max prefetched responses, if any request comes through this wrapper that is not on the "chain" created by the current map we want to wipe the map clean and start again.

Obviously this wrapper is not super generic and is meant to be used expicitly with `ShapeStream` but I think having this escape hatch is safer - an example would be shape rotations

Will add a comment explaining this further

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
msfstef: We don't _need_ to, but it does enhance the semantics at the cost of 4 bytes - by this argument we could say that we should name headers `electric-ecutd` to save a few bytes on request header keys I think

kevin-dp: It also can make users believe it could be `false`, which is not the case.

msfstef: Right, but it indicates that it is a flag - which means that if it's missing it's false. This is similar to our `live` query parameter where we set `live=true`, although we could technically just do `live=` and it would still work

packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
msfstef: Hmm at the moment live requests will never actually serve an incomplete chunk as we set the chunk last offset to serve to the latest available, received with a notification. Adding a test for a case that does not occur is a bit too verbose imo, and removing the safety net feels a bit unnecessary?

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
msfstef: Either way I removed the `"true"` value as it is unnecessary so I don't have a strong opinion on it either way

packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
kevin-dp: Ok, just a bit weird that we have a code path that cannot occur and that we cannot test either

kevin-dp (approved):

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -264,6 +265,25 @@ defmodule Electric.Plug.ServeShapePlug do
...
 
+  defp determine_up_to_date(
+         %Conn{
+           assigns: %{chunk_end_offset: chunk_end_offset, last_offset: last_offset}
+         } = conn,
+         _
+       ) do
+    if LogOffset.compare(chunk_end_offset, last_offset) == :lt do
+      conn
+      |> assign(:up_to_date, [])
```
KyleAMathews, line 276: Also we should reduce the ttl for caching per https://github.com/electric-sql/electric/issues/1447

packages/typescript-client/src/fetch.ts
```diff
@@ -77,3 +85,187 @@ export function createFetchWithBackoff(
...
+
+  let prefetchQueue: PrefetchQueue
+
+  const prefetchClient = async (...args: Parameters<typeof fetchClient>) => {
+    const url = args[0].toString()
+
+    // try to consume from the prefetch queue first, and if request is
+    // not present abort the prefetch queue as it must no longer be valid
+    const prefetchedRequest = prefetchQueue?.consume(...args)
+    if (prefetchedRequest) return prefetchedRequest
```
KyleAMathews, line 120: could we use `{`s? That's the normal pattern so I my eye always gets confused for a bit when I see braceless code.

KyleAMathews (commented):
This looks good in general. Our little simple client is growing up 😆 

Have you tested this manually on a site that does have a bunch of chunks? The tests look pretty through but seeing it live would be good to double-check everything is wired up correctly.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
msfstef: Yep that's on the chunking todo (determine ttl based on chunk type)

KyleAMathews: cool 👍

packages/typescript-client/src/fetch.ts
msfstef: Sure - we should probably add a lint rule, I won't do that for this PR, but I'll edit this

msfstef:
https://github.com/user-attachments/assets/bb67103d-4856-4726-ae58-20d715419d5b


@KyleAMathews  seems to be working fine with linearlite! You can see how with this prefetching I've made it so that the next request is fired off before the previous one has streamed all of its body yet (using the metadata in the headers) - this is >100k issues on linearlite initial sync

packages/typescript-client/src/fetch.ts
```diff
@@ -77,3 +85,190 @@ export function createFetchWithBackoff(
...
+    // only prefetch when queue is not full
+    if (this.#prefetchQueue.size >= this.#maxPrefetchedRequests) return
+
+    // initialize aborter per request, to avoid aborting consumed requests that
+    // are still streaming their bodies to the consumer
+    const aborter = new AbortController()
+
+    try {
+      const request =
+        this.#prefetchQueue.get(url)?.[0] ??
```
kevin-dp, line 209: Under which circumstances can it be that the URL is already in the prefetch queue?

kevin-dp (approved):
Definitely better!

KyleAMathews (approved):
:shipit: 

Looks great @msfstef!

packages/typescript-client/src/fetch.ts
msfstef: Removed it as it _should_ be unnecessary, although it is harmless and a good safety measure

PR #1755, opened 2024-09-24
feat(sync-service): Support listening on IPv6 network interfaces

### LINKED ISSUE #1751: Support Electric listening on IPv6
From @alco on Discord: https://discord.com/channels/933657521581858818/933657523049885698/1288142038428029102

> Electric only listens on IPv4 interfaces. In the old Electric we had a configuration option called ELECTRIC_USE_IPV6 that configured Bandit to listen on IPv6 interfaces. But we don't currently have that in the new Electric.

alco:
Closes #1751.

Electric can be configured to listen on the `:::` IPv6 address, as opposed to the default `0.0.0.0` IPv4 one, by configuring it with `LISTEN_ON_IPV6=true`.

### DIFF
--- a/.changeset/kind-turkeys-reply.md
+++ b/.changeset/kind-turkeys-reply.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Add LISTEN_ON_IPV6=true|false configuration option to support IPv6 network interfaces.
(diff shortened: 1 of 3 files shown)


robacourt (approved):
Perfect ⭐

Issue #1744, opened 2024-09-23
Move WHERE clause evaluation to single process


balegas:
We're doing shape where clause filtering inside the process that owns the shape log. While processes are cheap in Elixir, we're copying a lot of memory around for each shape and launching async work that ends up being discarded because a single write operation only touches to a few shapes.

We think this is adding a significant compute overhead to the server and want to test an alternative solution where shape where clauses are evaluated in a single process and write operations are sent only to the shape owners that will actually write to the shape log.

- [ ] Try [this](https://github.com/electric-sql/electric/issues/1744#issuecomment-2368163910) suggestion or open a follow-up issue

marc-shapiro:
May I suggest going one step further, and maintaining a single shape log for all subscribers.  You get performance advantages: sequential writes and less bursty I/O.

balegas:
That's a radical :D. I think shape interleaving is worth exploring since the CDN will cache what you want after a first call. Would also be useful for parametrised shape that are not updated frequently.

It is worth investigating, but let's do the pragmatic steps first.

KyleAMathews:
Hmm yeah that is a cool idea. Scanning a file and filtering out rows by shape id would be really fast as long as there wasn't too many shapes interwoven.

PR #1761, opened 2024-09-26
feat(sync-service): Move WHERE clause evaluation to single process

robacourt:
This is the simplest change to meet the criteria of #1744 .

It's faster and uses less memory:
Screenshot 2024-10-10 at 13 48 21

Than our current version:
Screenshot 2024-10-10 at 13 48 38

Because of the reduction in memory use it can handle bigger transactions at 100k shapes:
Screenshot 2024-10-10 at 13 36 12

vs our la
Screenshot 2024-10-10 at 13 35 59
test version:


KyleAMathews (approved):
Seems like a good first step 👍

balegas:
Yep. Sounds good.

PR #1763, opened 2024-09-27
chore: add plausible to website

samwillis:


KyleAMathews (approved):
I assume it doesn't need any ids or auth?

samwillis:
@KyleAMathews  nope, just this. It's the same, except for the `data-domain` as on pglite.dev.

thruflo:
Note that plausible generally underreports traffic significantly and is 30-40% of the traffic reported on Netlify.

KyleAMathews:
All js trackers do as so many people have blocker extensions installed.

Though also Netlify tends to overreport too as it i believe doesn't filter out bot traffic very well.

The best way to do analytics is to include the code in the first party js as then we only don't see data from people who block all js

Issue #1714, opened 2024-09-16
Add a `OTEL_DEBUG` env variable for enabling debug mode


KyleAMathews:
Right now we use an `OTEL_EXPORT` variable for switching between debug mode & exporting to an opentelemetry service — https://github.com/electric-sql/electric/blob/5cd4a9595f86199ebc6ea1039152025de75859cb/packages/sync-service/config/runtime.exs#L35-L70

Instead we should add an `OTEL_DEBUG` which enable debug mode and otherwise export if the other otel configuration is correct.

balegas:
@kevin-dp, my interpretation is that `OTEL_EXPORT` from `OTEL_DEBUG`.
Where you can have the two behaviours commutatively like:
- `OTEL_DEBUG=1` prints data to stdout
- `OTEL_EXPORT=otlp` exports data

I think in this case we don't care about the specific value of `OTEL_EXPORT`, just that it is set

alco:
@balegas Why `OTEL_DEBUG=1` and not `OTEL_DEBUG=true`? We've been using boolean values for configuration flags so far.

Even though it's not spelled out in the issue description, I remember @KyleAMathews saying that OTEL_EXPORT is unnecessary, setting OTLP_ENDPOINT by itself should be enough to enable the export, while OTEL_DEBUG is a complementary option that can enable debug logging to stdout independently from the OTLP export config.

KyleAMathews:
yeah, it's not an either or — `OTEL_DEBUG` overrides `OTEL_EXPORT` when set — it's like any other `DEBUG=true` mode. 

Actually now that I think about it — we'd still want to send traces to an external services even when we're debugging locally. So `OTEL_DEBUG` just adds output to the terminal.

Agree that `OTEL_DEBUG=true` makes more sense. That's how most env variable settings work afaik.

PR #1765, opened 2024-09-30
feat (sync-service): add OTEL_DEBUG environment variable

kevin-dp:
This PR fixes https://github.com/electric-sql/electric/issues/1714 by introducing a separate `OTEL_DEBUG` env var to enable printing the spans to stdout while also being able to export the spans using `OTEL_EXPORT`.

### DIFF
--- a/.changeset/mean-swans-act.md
+++ b/.changeset/mean-swans-act.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Add the OTEL_DEBUG environment variable which is a flag to print Open Telemetry spans to stdout.
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -29,44 +29,44 @@ config :opentelemetry,
   resource_detectors: [:otel_resource_env_var, :otel_resource_app_env],
   resource: %{service: %{name: service_name, version: version}, instance: %{id: instance_id}}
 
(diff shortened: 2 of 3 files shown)


alco:
@kevin-dp I've made some changes to make OTEL_DEBUG and OTLP export compatible with each other. Tested my changes against both an OpenTelemetry Collector and Honeycomb, ensuring that batch export works together with debug output.

In a follow-up PR, we should also remove the `OTEL_EXPORT` env var from the docs and explain that setting `OTLP_ENDPOINT` alone is enough to enable OT export.

alco (approved):

balegas:
> In a follow-up PR, we should also remove the OTEL_EXPORT env var from the docs and explain that setting OTLP_ENDPOINT alone is enough to enable OT export.

We should change docs accordingly in this PR

alco:
> > In a follow-up PR, we should also remove the OTEL_EXPORT env var from the docs and explain that setting OTLP_ENDPOINT alone is enough to enable OT export.
> 
> We should change docs accordingly in this PR

But we still have the same lag between doc changes and code changes that we had with old Electric, where the new code only becomes available to users after a release is published.

balegas:
right, then we should have the other PR ready before code changes are merged :)

alco:
@balegas https://github.com/electric-sql/electric/pull/1795

kevin-dp:
@alco can i merge this?

KyleAMathews:
I'm just gonna go ahead and merge this 👍

PR #1772, opened 2024-10-01
Obfuscate database password to avoid accidentally leaking it in logs

### LINKED ISSUE #1718: ConnectionManager stores password in clear text in its state

alco:
Fix #1718.

Obfuscation is implemented by wrapping the password in a zero-arity function as early as possible, so that the password is never stored in cleartext in any of our processes' state. The deobfuscation happens right before passing connection options to Postgrex. This creates an explicit point of hand-off where the responsibility for keeping the password from getting leaked is transfered from our code to Postgrex.

### DIFF
--- a/.changeset/odd-ants-smell.md
+++ b/.changeset/odd-ants-smell.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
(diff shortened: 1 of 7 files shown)


msfstef (approved):
I like this better!

Issue #1769, opened 2024-09-30
Extend the documentation on shape WHERE clauses


alco:
The current documentation is here - https://electric-sql.com/docs/guides/shapes#where-clause. It doesn't go into detail on which types, operators and functions can or cannot be used.

We should extend it with more specifics and possibly a call-to-action for when people find missing functionality that they need. 

The old website had a bit more elaborate treatment of the subject - https://legacy.electric-sql.com/docs/usage/data-access/shapes#string-where-clause.

PR #1773, opened 2024-10-01
docs: where clause limitations with feature request CTA.

thruflo:
Partly addresses https://github.com/electric-sql/electric/issues/1769 although it would be much better to add more technical details.

https://deploy-preview-1773--electric-next.netlify.app/docs/guides/shapes#where-clause


alco:
@thruflo given the small subset of SQL we support in shape where clauses, I believe it is more helpful to talk about what _is_ supported, first and foremost. I have pushed a commit to that effect to your branch.

alco (approved):

KyleAMathews (approved):
:shipit:

Issue #1764, opened 2024-09-29
CamelCased Tables cause "table name does not match expected format"


AndryHTC:
Example:
`public.Test`

If I request the shape:
`http://localhost:3000/v1/shape/Test` -> `{
	"root_table": [
		"table name does not match expected format"
	]
}`

`http://localhost:3000/v1/shape/test` -> `{
	"root_table": [
		"table not found"
	]
}`

But if I alter the name of the table to lowercase `public.test`, it works.

**It should accept any table format as shape.**

kevin-dp:
Hi @AndryHTC,

Thanks for reporting this. Current schema and table name validation is quite strict as it is based on this regex:
```
^((?<schema>([a-z_][a-zA-Z0-9_]*|"(""|[^"])+"))\.)?(?<table>([a-z_][a-zA-Z0-9_]*|"(""|[^"])+"))$
```

As you can see from the regex, we require the first character to be a lowercase letter, an underscore, or a quote. I remember implementing it like that based on this paragraph from the [PostgreSQL documentation](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS):

> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). Note that dollar signs are not allowed in identifiers according to the letter of the SQL standard, so their use might render applications less portable.

I interpreted this as, names cannot start with uppercase letters. But, after testing it this morning, table names starting with an uppercase letter seem perfectly valid. For now, you can work around this issue by quoting the table name: "Test".

We will need to modify the validation to:
- allow names starting with uppercase letters
- allow $ sign in names
- Allow diacritical marks and non-Latin letters

Note that these restrictions only apply for non-quoted names, when they are quoted, names can contain any character. This can always be achieved by quoting the table when requesting the shape, but for convenience we should perhaps also allow these characters without quoting.

AndryHTC:
@kevin-dp great, it works with quotes so I can continue with the implementation.

What I don't understand is why you're so strict with the table format 😄

kevin-dp:
Indeed, we should relax the validation and just follow Postgres semantics by making unquoted names case insensitive.

kevin-dp:
Here are 2 candidate regexes to validate table names.
[...]

PR #1776, opened 2024-10-01
fix (sync-service): let Postgres handle user-provided root table name

kevin-dp:
This PR fixes https://github.com/electric-sql/electric/issues/1764 according to the approach that was decided by @KyleAMathews, namely to adhere to Postgres' way of handling identifiers. Therefore, this PR modifies Electric such that it passes the user-provided root table to Postgres which then returns the actual relation (i.e. schema and table name). Which we then use throughout the rest of the code as before.

We extended the ETS inspector with a write-through cache that caches the relation. For any incoming request, it tries to fetch the relation for that table from the cache. If not found, it loads it from PG and stores it in the cache. Note that many table name spellings can refer to the same relation, e.g. `users`, `USERS`, `public.users`, etc. all refer to the same relation `{public, users}`. For each different spelling, we will load the relation from PG and store it in the cache. Subsequent requests then load it from cache. We also cache the inverse (mapping relation to the different table names) in order to efficiently clean up.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
+++ b/packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
@@ -20,8 +20,8 @@ defmodule Electric.Postgres.Inspector.DirectInspector do
       {:ok, result} ->
         # We expect exactly one row because the query didn't fail
         # so the relation exists since we could cast it to a regclass
-        info = Enum.at(result.rows, 0)
-        {:ok, {Enum.at(info, 0), Enum.at(info, 1)}}
+        [[schema, table]] = result.rows
+        {:ok, {schema, table}}
 
       {:error, err} ->
         {:error, Exception.message(err)}
--- a/packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
+++ b/packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
@@ -2,6 +2,35 @@ defmodule Electric.Postgres.Inspector.DirectInspector do
   @behaviour Electric.Postgres.Inspector
 
   @doc """
+  Returns the PG relation from the table name.
+  """
+  def load_relation(table, conn) do
+    # The extra cast from $1 to text is needed because of Postgrex' OID type encoding
+    # see: https://github.com/elixir-ecto/postgrex#oid-type-encoding
+    query = """
+    SELECT nspname, relname
+    FROM pg_class
+    JOIN pg_namespace ON relnamespace = pg_namespace.oid
+    WHERE
+      relkind = 'r' AND
+      pg_class.oid = $1::text::regclass
+    """
+
+    case Postgrex.query(conn, query, [table]) do
+      {:ok, result} ->
+        # We expect exactly one row because the query didn't fail
+        # so the relation exists since we could cast it to a regclass
+        info = Enum.at(result.rows, 0)
+        {:ok, {Enum.at(info, 0), Enum.at(info, 1)}}
+
+      {:error, err} ->
+        {:error, Exception.message(err)}
+    end
+  end
+
+  def clean_relation(_, _), do: true
+
+  @doc """
   Load table information (refs) from the database
   """
   def load_column_info({namespace, tbl}, conn) do
@@ -40,27 +69,5 @@ defmodule Electric.Postgres.Inspector.DirectInspector do
 
   def clean_column_info(_, _), do: true
 
-  def get_namespace_and_tablename(table, conn) do
-    # The extra cast from $1 to text is needed because of Postgrex' OID type encoding
-    # see: https://github.com/elixir-ecto/postgrex#oid-type-encoding
-    query = """
-    SELECT nspname, relname
-    FROM pg_class
-    JOIN pg_namespace ON relnamespace = pg_namespace.oid
-    WHERE
-      relkind = 'r' AND
-      pg_class.oid = $1::text::regclass
-    """
-
-    case Postgrex.query(conn, query, [table]) do
-      {:ok, result} ->
-        # We expect exactly one row because the query didn't fail
-        # so the relation exists since we could cast it to a regclass
-        info = Enum.at(result.rows, 0)
-        {Enum.at(info, 0), Enum.at(info, 1)}
-
-      {:error, err} ->
-        {:error, Exception.message(err)}
-    end
-  end
+  def clean(_, _), do: true
 end
--- a/packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
+++ b/packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
@@ -29,6 +29,7 @@ defmodule Electric.Postgres.Inspector.DirectInspector do
     result = Postgrex.query!(conn, query, [tbl, namespace])
 
     if Enum.empty?(result.rows) do
+      # Fixme: this is not necessarily true. The table might exist but have no columns.
       :table_not_found
     else
       columns = Enum.map(result.columns, &String.to_atom/1)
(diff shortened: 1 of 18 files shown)


packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
```diff
@@ -38,4 +39,28 @@ defmodule Electric.Postgres.Inspector.DirectInspector do
   end
 
   def clean_column_info(_, _), do: true
+
+  def get_namespace_and_tablename(table, conn) do
```
KyleAMathews, line 43: I love going back to postgres as the final arbitrator on matching a name. I'm not sure though if there's caching here though? This looks like it gets called a lot which could add significant latency.

kevin-dp: Yeah, we could modify the ETS inspector to cache the result like it does for the column info.

packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -75,7 +75,11 @@ defmodule Electric.Plug.RouterTest do
 
       assert %{status: 400} = conn
 
-      assert %{"root_table" => ["table not found"]} = Jason.decode!(conn.resp_body)
+      assert %{
+               "root_table" => [
+                 ~s|Table "nonexistent" does not exist. If the table name contains capitals or special characters you must quote it.|
```
KyleAMathews, line 80: nice error message!

KyleAMathews (approved):
I gave a brief look through (gotta run to airport in a sec) and it looks great.

One thing we talked about yesterday also is including the sql statement in errors we send back — I'll make a separate issue for that.

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
```diff
@@ -36,20 +65,63 @@ defmodule Electric.Postgres.Inspector.EtsInspector do
...
+    clean_column_info(relation, opts_or_state)
+    clean_relation(relation, opts_or_state)
+  end
+
   ## Internal API
 
   @impl GenServer
   def init(opts) do
     pg_info_table = :ets.new(opts.pg_info_table, [:named_table, :public, :set])
+    pg_relation_table = :ets.new(opts.pg_relation_table, [:named_table, :public, :bag])
```
msfstef, line 79: why not store everything under the same ETS table?

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
```diff
@@ -36,20 +65,63 @@ defmodule Electric.Postgres.Inspector.EtsInspector do
     :ets.delete(ets_table, {table, :columns})
   end
 
+  @impl Electric.Postgres.Inspector
+  def clean(relation, opts_or_state) do
+    clean_column_info(relation, opts_or_state)
```
msfstef, line 70: this could be done with a single `:ets.select_delete` statement - would we ever need to clean only columns or only relations? If not we could just keep a single `clean` API

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
```diff
@@ -36,20 +65,63 @@ defmodule Electric.Postgres.Inspector.EtsInspector do
...
+            # - Forward: user-provided table name -> PG relation (many-to-one)
+            #     e.g. `~s|users|` -> `{"public", "users"}`
+            #          `~s|USERS|` -> `{"public", "users"}`
+            # - Backward: and PG relation -> user-provided table names (one-to-many)
+            #     e.g. `{"public", "users"}` -> `[~s|users|, ~s|USERS|]`
+            #
+            # The forward direction allows for efficient lookup (based on user-provided table name)
+            # the backward direction allows for efficient cleanup (based on PG relation)
+            :ets.insert(state.pg_info_table, {{table, :table_to_relation}, relation})
+            :ets.insert(state.pg_relation_table, {{relation, :relation_to_table}, table})
```
msfstef, line 116: we can use the `pg_info_table` for both and insert them with a single `ets` call

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -261,7 +261,7 @@ defmodule Electric.ShapeCache do
       # because the table name may have changed in the new relation
       # if there is no old relation, we use the new one
       rel = old_rel || relation
-      inspector.clean_column_info({rel.schema, rel.table}, inspector_opts)
+      inspector.clean({rel.schema, rel.table}, inspector_opts)
```
msfstef, line 264: this is the only place `clean_column_info` is ever called, so why not remove all clean APIs and just keep this `clean` API for the whole relation?

msfstef (changes requested):

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
kevin-dp: Because the tables have different semantics. The `pg_info_table` is a set whereas the `pg_relation_table` is a bag because each key (i.e. relation) has several values (namely the user provided identifiers). See comment on L110-111.

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
kevin-dp: we can't, see my comment about sets vs bags before.

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
msfstef: Ah I see - yep you're right, but I see that we only use this for the `tables_from_ets` API, but this PR doesn't use that anywhere? Why would we want to reverse search for the tables?

kevin-dp: For cleanup, when a certain relation is deleted we need to clean up the cache so we need to know all the names that were cached for that relation.

msfstef: We could include the relation in the key of the `pg_info_table` entry like `{table, relation, :table_to_relation}` and then do a `:ets.select_delete` with a match spec like `{table, :_, :table_to_relation}` - no?

kevin-dp: @msfstef Conceptually i used two indexes for performance reasons. A forward index from user-provided identifier to the relation name, and a backward index from relation name to a list of identifiers. The backward index is not strictly needed as we could do a delete from the forward index of all keys where the value is the relation. But that's an expensive operation. I'm not familiar with ETS performance characteristics, but it seems to me that what you're suggesting would be more expensive as it requires searching for such keys, while in the approach with 2 indexes it's a simple lookup (no search needed).

msfstef: Yep was thinking the same w.r.t. to performance - my only thought was that if it is an ordered set, then looking up based on a key that starts with `table` should be quick to find the match, especially given that the size of this would be small.

I understand the reasoning for the two indices, my instinct was just to keep it as a single one and change it if performance is an issue but since it's internal happy to resolve this

kevin-dp: Yeah, well constant vs log time complexity 😄

packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
```diff
@@ -1,6 +1,35 @@
...
+      relkind = 'r' AND
+      pg_class.oid = $1::text::regclass
+    """
+
+    case Postgrex.query(conn, query, [table]) do
+      {:ok, result} ->
+        # We expect exactly one row because the query didn't fail
+        # so the relation exists since we could cast it to a regclass
+        info = Enum.at(result.rows, 0)
+        {:ok, {Enum.at(info, 0), Enum.at(info, 1)}}
```
robacourt, line 24:
```suggestion
        [[schema, table]] = result.rows
        {:ok, {schema, table}}
```

robacourt (approved):
Looks good to me! I just gave one nit-picky suggestion

msfstef (approved):

PR #1784, opened 2024-10-02
fix(sync-service): Query Postgres server version at the start of connection setup

alco:
We were querying it after starting the replication client before, which was too late, because by that time the shape recovery process might have already reached `Electric.Postgres.Configuration.configure_tables_for_replication!()` where the version is used to branch the control flow. This was causing flake in integration tests due to the version not always being there and the unsupported query being issue against Postgres version 14.

In short, this PR resolves the flakiness in integration tests, they should be consistently green from now on. 

The choice of augmenting the `LockConnection` module with version querying and a crude state machine might seem dubious but the end result is simpler this way, compared to adding version querying to the replication client, for example.

### DIFF
--- a/.changeset/dry-buses-fail.md
+++ b/.changeset/dry-buses-fail.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Query Postgres server version as early as possible so that it is available throughout the whole connection initialization process.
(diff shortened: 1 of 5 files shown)


robacourt (approved):
Nice one!

msfstef (approved):
Looks good! We might want to rename `LockConnection` to something like `PrepareDatabaseConnection` or something more generic like that, as it seems to be evolving to all steps required before any other interaction with the db

kevin-dp (approved):

alco:
@msfstef good point! I'll consider that when working on #1792.

Issue #1785, opened 2024-10-03
Slow to respond concurrent requests for new shapes


msfstef:
When multiple concurrent requests for new shapes come in, requests and responses are all routed through the `ShapeCache` causing delays and flooding of the process mailbox. Requests can instead be routed to individual shape worker processes to handle their needs, like waiting for snapshots and requesting a log etc.

The goal is to move stuff off of the critical path of the requests and enable quick responses, even if the responses are 429s because of e.g. pool connection resource limitations.

msfstef:
Closing this as it has been addressed

Issue #1770, opened 2024-09-30
Queue requests and return 429s when the server gets overloaded


KyleAMathews:
Right now we eagerly try to fulfill all shape requests.

Too many requests can crash the server or postgres or trigger timeouts.

We should handle this more gracefully.

First we should queue requests and limit concurrency on fulfilling them.

We should perhaps have one queue for already created shapes vs. ones we're initing as they have different performance characteristics (reading the cached logs is very cheap vs. creating a new shape).

If the queue gets too long, we can start returning requests at the end of the queue with a 429 http code (Too Many Requests). [Clients would need to special-case 429](https://github.com/electric-sql/electric/blob/7f4456529f10b39be453fd206d0ea348233eaf8f/packages/typescript-client/src/fetch.ts#L65-L68) as a retry-able error.

We'll need to figure out heuristics for "too long" of queue — which could be a combination of hard limits + soft limits based on postgres & electric loads i.e. if postgres is slow responding to queries, we could drop concurrency and queue length temporarily.

ericallam:
We're going to be adding rate-limiting at our "proxy" endpoint and returning 429s as well, which sets `x-ratelimit-limit`, `x-ratelimit-remaining`, and `x-ratelimit-reset` headers. Having the client retry by respecting those headers (instead of a simple exponential backoff) would be great.

For reference, the `x-ratelimit-reset` is a unix timestamp in ms.

PR #1787, opened 2024-10-03
feat: Move operations from `ShapeCache` to `Shape.Consumer`

msfstef:
Addresses https://github.com/electric-sql/electric/issues/1785 and partially addresses https://github.com/electric-sql/electric/issues/1770

Moves a lot of the operations that went through `ShapeCache` directly into the `Shape.Consumer`, so that requests can be replied to directly from the shape consumers rather than flooding the `ShapeCache` with casts that take a while to reach the requesters.

I've tried to keep changes to a minimum in order to do this incrementally and keep these PRs easily reviewable - the `ShapeStatus` still persists data on every call, the relations and truncates still go through `ShapeCache` rather than individual shapes, etc

I've also caught the `DBConnection.ConnectionError`s for queue timeouts and converted them to 429 errors.
We need to also handle `GenServer.call` timeouts as sometimes the PG query might not fail but take longer than the default 5 seconds for the GenServer call.

NOTE: I have not updated any tests yet as I first want to ensure people agree with the approach

PERFORMANCE CHECK:
- On my local machine, using in memory stores, running 1000 concurrent new shape connections consistently took ~20sec with these changes, compared to the ~33sec on main, so a ~30% improvement.
- I was also able to succesfully run 10k concurrent connections with this, although it took ~10min to serve, but on main I wasn't able to succsefully run it (@robacourt I think that was the case for you too?) - at least we know it does not get into an unrecoverable state.


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -643,9 +648,8 @@ defmodule Electric.Plug.ServeShapePlug do
 
   @impl Plug.ErrorHandler
   def handle_errors(conn, error) do
-    OpenTelemetry.record_exception(error.reason, error.stack)
-
     error_str = Exception.format(error.kind, error.reason)
+    OpenTelemetry.record_exception(error_str, error.stack)
```
msfstef, line 652: This was throwing because `GenServer.call` timeout errors have a tuple as the reason which OpenTelemetry does not handle - converting to a string first seems to be the safest option (?)

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -293,9 +296,12 @@ defmodule Electric.ShapeCache do
         {:reply, :started, [], state}
 
       true ->
-        Logger.debug("Starting a wait on the snapshot #{shape_id} for #{inspect(from)}}")
+        GenServer.cast(
```
msfstef, line 299: We could avoid going through the shape cache and directly go to a shape consumer, but for the sake of incrementally refactoring this I kept it like this as I think there are more concurrency considerations in the other case

packages/sync-service/lib/electric/shapes/consumer/supervisor.ex
```diff
@@ -12,6 +12,8 @@ defmodule Electric.Shapes.Consumer.Supervisor do
             log_producer: [type: @genserver_name_schema, required: true],
             shape_cache: [type: :mod_arg, required: true],
             registry: [type: :atom, required: true],
+            persistent_state: [type: :any, required: true],
+            shape_status: [type: :atom, required: true],
```
msfstef, line 16: I'd rather pass something like a partially applied "shape_status" module with the persistent state passed into it, but I'm not familiar with how to do this so would appreciate any tips (also `ShapeStatus` has wrong types, and some functions accept a state map and others accept a single table name)

packages/sync-service/lib/electric/shapes/consumer_supervisor.ex
```diff
@@ -27,13 +27,16 @@ defmodule Electric.Shapes.ConsumerSupervisor do
...
     case GenServer.whereis(Consumer.Supervisor.name(electric_instance_id, shape_id)) do
       nil ->
         {:error, "no consumer for shape id #{inspect(shape_id)}"}
 
       pid when is_pid(pid) ->
-        DynamicSupervisor.terminate_child(name, pid)
+        Consumer.Supervisor.clean_and_stop(%{
+          electric_instance_id: electric_instance_id,
+          shape_id: shape_id
+        })
```
msfstef, line 39: I was unsure if we need to also have a fallback here to forcefully do `DynamicSupervisor.stop(name, pid)` - I figured if the Supervisor is alive it should be able to handle its own shutdown gracefully

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
alco: https://github.com/electric-sql/electric/pull/1788 fixes this.

packages/sync-service/lib/electric/shapes/consumer_supervisor.ex
alco: I'm not entirely sure what you're referring to here. The code feels weird in the sense that the passed `name` argument is not used but a name is constructed for the consumer process. It's not your changes, it was weird before them.

It's not important to dwell on this right now, IMO. We can clear up process lifecycles at another incremental step of the overall move from ShapeCache to individual shape process trees.

packages/sync-service/lib/electric/shapes/consumer/supervisor.ex
alco: It seems that various functions from the provided `shape_status` module are called, with varying arguments. If you imply that all of ShapeStatus' functions should take a state argument, they already do. Some of the functions simply have another clause that takes a more specific argument. Those can be converted into differently named private functions if they themselves aren't called from the outside.

alco (commented):
Looking great so far :+1:

packages/sync-service/lib/electric/shapes/consumer_supervisor.ex
msfstef: I think the idea was that we would terminate processes via the dynamic supervisor, hence the name being passed - but for some reason we check if the consumer is alive (rather than handling the error from the `DynamicSupervisor`)

I've changed things around such that processes clean up after themselves, so an external termination would only be a fallback

Either way, I agree that we should do these in steps to keep things reviewable!

msfstef:
@balegas  PR should be ready, @robacourt has been kind enough to kick off a benchmark run for this PR so we can see results on Monday and hopefully merge (?) I expect an improvement in concurrent shape creation and no regressions on other benchmarks - let's see

balegas:
Yeah looking forward to that! If the benchmarks show anything odd, we'll be happy to have ran the benchmarks before :).

robacourt (approved):
Great work! This shows a massive speed improvement for concurrent shape creation while the other benchmarks remain unchanged:
Screenshot 2024-10-08 at 20 06 01

PR #1791, opened 2024-10-03
feat (client): extensible value type for use with custom parsers

kevin-dp:
This PR modifies the TypeScript types of our client to enable users to extend the `Value` type. This is needed when using custom parsers.

For example, the parse below used to throw a type error because `Date` is not a `Value` (where `Value` are the base SQL values):

```ts
const parser: Parser = {
  timestampz: (date: string) => {
    return new Date(date)
  },
}
```

With this PR, we can now write this parser as:
```ts
const parser: Parser<Date> = {
  timestampz: (date: string) => {
    return new Date(date)
  },
}
```

And an example usage of this parser with the `ShapeStream`:
```ts
type CustomRow = {
  foo: number
  bar: boolean
  baz: string
  ts: Date
}

const shapeStream = new ShapeStream<CustomRow>({
  url: `...`,
  parser: {
    timestampz: (date: string) => {
      return new Date(date)
    },
  },
})
```

Note that this would throw an error as `Date` is not a type that occurs in the row:
```ts
type CustomRow = {
  foo: number
  bar: boolean
  baz: string
}

const shapeStream = new ShapeStream<CustomRow>({
  url: `...`,
  parser: {
    // type error here
    timestampz: (date: string) => {
      return new Date(date)
    },
  },
})
```


balegas:
Thanks @kevin-dp, that will solve my problem.

packages/typescript-client/src/types.ts
```diff
@@ -1,13 +1,21 @@
...
   | boolean
   | bigint
   | null
-  | Value[]
-  | { [key: string]: Value }
+  | Extensions
+  | Value<Extensions>[]
+  | { [key: string]: Value<Extensions> }
+
+export type Row<Extensions = never> = Record<string, Value<Extensions>>
```
msfstef, line 15: why is the default `never` instead of `unknown`? It seems that you've modified the code everywhere to specify `Row<unknown>` if it's not specified, I feel like making it the default probably makes sense (rather than forcing people to explicitly type it)?

msfstef (approved):
Looks good, very useful type functionality!

I don't quite understand why we don't use `unknown` as a default which requires explicitly typing it everywhere? I feel like this PR could be reduced to 2 file changes with an `unknown` default, am I missing something?

kevin-dp:
@msfstef `never` acts as the neutral element on union types, `unknown` acts as the absorbing element on union types:

```ts
type Value = string | boolean | unknown
const a: Value = null // is accepted because`Test` is basically `unknown`

type Value2 = string | boolean | never
const b: Value2 = null // type error because `Test = string | boolean`
```

So we want `never` as the default because that means that if the user does not provide anything then it is correctly typed as just the base SQL types. If we would use `unknown` then the entire `Value` type would be reduced to `unknown` which we don't want.

msfstef:
@kevin-dp ah now I see, and the default is still `Row<never>` for all APIs so the behaviour is still the same - thanks for the clarification, good to merge!

PR #1793, opened 2024-10-03
fix(redis example): update for partial updates

### LINKED ISSUE #1789: Update Redis example to handle `update` operations properly
At the time it was written, an `update` operation included the entire updated row so you could just write it directly to Redis. We subsequently changed that to only include the actual changed columns so now you need to get the key, merge in the update, and then write it back again.

KyleAMathews:
Fixes #1789


examples/redis-sync/src/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    // Begin a Redis transaction
+    //
+    // FIXME The Redis docs suggest only sending 10k commands at a time
+    // to avoid excess memory usage buffering commands.
+    const pipeline = client.multi()
+
+    // Loop through each message and make writes to the Redis hash for action messages
+    messages.forEach((message) => {
+      if (!isChangeMessage(message)) return
+      console.log(`message`, message)
```
samwillis, line 48: Left over console.log from debugging?

examples/redis-sync/src/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  const updateKeyScriptSha1 = await client.SCRIPT_LOAD(script)
+
+  const itemsStream = new ShapeStream({
+    url: `http://localhost:3000/v1/shape/items`,
+  })
+  itemsStream.subscribe(async (messages: Message[]) => {
+    // Begin a Redis transaction
+    //
+    // FIXME The Redis docs suggest only sending 10k commands at a time
+    // to avoid excess memory usage buffering commands.
```
samwillis, line 42: Worth creating an issue for?

samwillis (approved):
Looks good, just the stray `console.log`

Love a bit of Lua with Redis!

examples/redis-sync/src/index.ts
KyleAMathews: Oh I left it there on purpose actually as when playing with it it's nice to see the message log get outputted. For a production version yeah you wouldn't want it but I think it's nice for an example.

examples/redis-sync/src/index.ts
KyleAMathews: Eh no. It's a very low priority for an example. It's more a note for someone taking this into production. Perhaps the first to use this at scale can update it.

Issue #1771, opened 2024-09-30
Shape API tweaks


KyleAMathews:
A few things have come up:

## Proposal: move table from its current place as the last path segment to just another (required) query argument.

The root table for a shape is currently in the API path e.g. `v1/shape/${table_name}`. This is a required parameter and my original logic to do it this way was basically that it's easy to write when testing out simple shape per table shapes. E.g. `curl http://localhost:3000/v1/shape/foo`.

But it does get a bit weird as people might need to specify the schema the table is part of. Also there's an odd chance that in the future, we'll support alternative backends that don't have "tables". Also we might want to have some sort of multi-table syntax. Regardless, moving the table out of the path gives us more flexibility.

So that foo shape curl above would look like:

`curl http://localhost:3000/v1/shape?table=foo`

## Proposal: change `shape_id` to `shape_handle`

The problem with `id` is it strongly suggests that it's a permanent identifier for the particular shape. Whereas it's just pointing to a particular ephemeral instantiation of the shape. A particular shape instantiation be removed at any moment so we need a term that conveys the idea of a unique identifier that you can use to access or manipulate a specific instance of the shape log.

`handle` is often used for this sort of thing.

`token` is another term I considered but it has too much whiff of session-y or authorization related usage.

balegas:
> curl http://localhost:3000/v1/shape?table=foo
I think this will age a lot better as we add more complex shape definitions

PR #1796, opened 2024-10-04
chore: Refactor shape_id -> shape_handle for clearer name

KyleAMathews:
Addresses one of the renames proposed in https://github.com/electric-sql/electric/issues/1771

PR #1798, opened 2024-10-04
chore: Update table path param to be a query param

KyleAMathews:
Partially addresses https://github.com/electric-sql/electric/issues/1771


KyleAMathews:
Great thanks! I'll give this a through go over later & double-check docs, examples, etc. and then merge it in

KyleAMathews:
Yeah `root_table` is useful in Elixir to be really clear about what table it is but since only one table is allowed in a shape request, just calling it "table" everywhere else is cleaner.

KyleAMathews:
Ok, this is ready to go — all examples are tested & gave another look over docs.

Per https://github.com/electric-sql/electric/pull/1796#pullrequestreview-2388639969 we'll just either need to decide on the order of merging.

KyleAMathews:
Not sure where those typescript errors are from — not reproducing locally.

thruflo:
So everywhere we have table in the URL we need to switch to table as a param? Can you shout when this is merged so we can update parallel PRs like deployment docs and Elixir client.

KyleAMathews:
The docs are already updated as part of this PR — we can just release it right away when we merge.

icehaunter:
Superseded by #1900

PR #1800, opened 2024-10-06
Format quickstart output

timneedham:
Just a little docs thing...
The JSON output in the quickstart example isn't formatted as stated/expected.


KyleAMathews (approved):
Oops, thanks for the fix!

Issue #1802, opened 2024-10-07
Persistence of unified shape metadata slows system down with large number of shapes


msfstef:
As discovered by @robacourt, currently shape metadata is saved via the `PersistentKV` module that serializes all shape metadata together and saves it to a single file - and does that on the critical path of shape creation. When reaching very large amounts of shapes this slows the system down significantly.

One suggestion as an initial improvement on this is to serialize and store individual shape metadata independently of all other shapes, so there's no scaling performance hit with serialization or contention on a single file.

PR #1805, opened 2024-10-07
feat: Serialise and persist shape metadata individually

msfstef:
Addresses https://github.com/electric-sql/electric/issues/1802

Now the serialisation cost and perhaps other resource contention should not be as high since each shape and relation is serialised individually and stored in a separate file.

I doubt this is the optimal solution, but it should remove the scaling cost of adding more shapes and make that more predictable.

Main concern I suppose is that each shape requires creating a new file, and that operation is in the critical request path for creating a shape.

Ultimately I think it might be better to store metadata as part of the storage interface, and recover shapes similarly to how I do here (listing the directory where shapes are stored and iterating over that).

NOTE: this is a breaking change as electric will no longer be able to resume from its previous `PersistentKV` format - not sure if we want to address that as a minor patch or do anything about it

### DIFF
--- a/.changeset/warm-cycles-live.md
+++ b/.changeset/warm-cycles-live.md
@@ -1,5 +1,5 @@
 ---
-"@core/sync-service": patch
+"@core/sync-service": minor
 ---
 
-Make `PersistentKV` store metadata in individual files on a per-shape basis.
+[BREAKING] Make `PersistentKV` store metadata in individual files on a per-shape basis.
--- a/.changeset/warm-cycles-live.md
+++ b/.changeset/warm-cycles-live.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Make `PersistentKV` store metadata in individual files on a per-shape basis.
--- a/packages/sync-service/lib/electric/persistent_kv.ex
+++ b/packages/sync-service/lib/electric/persistent_kv.ex
@@ -1,4 +1,6 @@
 defprotocol Electric.PersistentKV do
   def get(kv, key)
+  def get_all(kv)
   def set(kv, key, value)
+  def delete(kv, key)
 end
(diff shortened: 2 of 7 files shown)


KyleAMathews (approved):
Nice! Yeah this should be dramatically faster. We'll see in the traces if this causes any other issues.

Minor patch is good. Numbers are cheap so if it's a break change, let's just do a minor.

SQLite would be nice for stuff like this in some future iteration (well actually pglite when we can use it) as it's cheaper still to read/write to it + we can do analytics queries which could be handy.

msfstef:
@KyleAMathews I actually think we can do away with persisting the shape status altogether - it makes more sense for storage to also persist the `Shape` definition and then `ShapeStatus` can just recover its in memory state by using a method on `Storage` that will look at the stored files.

I think that would simplify things quite a bit and remove the need for separate places where we persist things - we already persist shapes with their data, better to keep it all in one place!

msfstef:
I've left [a comment on a related PR](https://github.com/electric-sql/electric/pull/1776/files#r1790207173) that affects this as well.

If we go to Postgres to validate relations we might as well retrieve the OID of the relation and store it as part of the `Shape` definition (right now we only store the qualified table name).

If we do that, then the `Shape` definition alone contains everything we need to know in order to 1) restore the shape and 2) detect if we should clean it based on relation changes, removing the need for all relation metadata storing, significantly simplifying this.

KyleAMathews:
> I actually think we can do away with persisting the shape status altogether

> If we do that, then the Shape definition alone contains everything we need to know in order to 1) restore the shape and 2) detect if we should clean it based on relation changes, removing the need for all relation metadata storing, significantly simplifying this.

Oh ok! I didn't realize what this was tracking — then yeah, if we're tracking to just eliminating this altogether, let's do that 😆

msfstef:
More radical approach in this PR: https://github.com/electric-sql/electric/pull/1808

msfstef:
Closing in favour of https://github.com/electric-sql/electric/pull/1808

PR #1807, opened 2024-10-07
Add parameter to API to send entire row on update

magnetised:
With `replica=full` update messages include the full row, not just
the changed columns.

Phoenix.LiveView Stream integration requires the full row to perform an
update, not just the changed values. This adds a parameter to the shape
definition that turns off the column diffing for updates.

Since we were just dropping unchanged values this is a minor change to
the wiring. Most of the work here is adding the configuration.

### DIFF
--- a/.changeset/sharp-moons-provide.md
+++ b/.changeset/sharp-moons-provide.md
@@ -1,0 +1,6 @@
+---
+"@electric-sql/client": patch
+"@core/sync-service": patch
+---
+
+Add `replica` parameter to change the behaviour for updates to include the full row, not just the modified columns
--- a/packages/typescript-client/test/client.test.ts
+++ b/packages/typescript-client/test/client.test.ts
@@ -3,6 +3,7 @@ import { v4 as uuidv4 } from 'uuid'
 import { setTimeout as sleep } from 'node:timers/promises'
 import { testWithIssuesTable as it } from './support/test-context'
 import { ShapeStream, Shape, FetchError } from '../src'
+import { Message, Row, ChangeMessage } from '../src/types'
 
 const BASE_URL = inject(`baseUrl`)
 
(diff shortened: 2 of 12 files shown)


packages/typescript-client/test/client.test.ts
```diff
@@ -329,4 +330,48 @@ describe(`Shape`, () => {
...
+    issuesTableKey,
+    issuesTableUrl,
+    clearIssuesShape,
+    aborter,
+  }) => {
+    const [id] = await insertIssues({ title: `first title` })
+
+    const shapeStream = new ShapeStream({
+      url: `${BASE_URL}/v1/shape/${issuesTableUrl}`,
+      sendDeltas: false,
```
balegas, line 346: Should be the default to send only the deltas as it is more efficient?
I don't like much the name either. I see it more as `row: full | modified`

balegas (commented):
I think this is useful, and we're seeing use cases for it already. I'll not approve, since I haven't looked at the Elixir.

- [ ] don't forget to create a patch for the docs and get approval for it so that we can release docs immediately after.

packages/typescript-client/test/client.test.ts
magnetised: > Should be the default to send only the deltas as it is more efficient?

Is the default.

balegas: Oh sure. This is when you need to set it :D with the non default value

.changeset/sharp-moons-provide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@electric-sql/client": patch
+"@core/sync-service": patch
+---
+
+Add a parameter to change the behaviour for updates to include the full row, not just the modified columns
```
KyleAMathews, line 6: Add the name of the parameter

KyleAMathews (commented):
Why the change from `send_delta`? That was clearer in my mind as "update_mode" seems like I could refer to any number of things. Also I prefer true/false settings when possible.

.changeset/sharp-moons-provide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@electric-sql/client": patch
+"@core/sync-service": patch
+---
+
+Add `replica` parameter to change the behaviour for updates to include the full row, not just the modified columns
```
robacourt, line 6: `replica: full` seems pretty confusing with me. Kyle's suggestion of `send_delta: false` seems clearer, or even better `send_full_row: true`.

If we are sticking to `replica: full` the options should be in the changeset (e.g. `full` and `default`)

robacourt (approved):
Code looks good. I think we should make the parameter name less confusing though.

.changeset/sharp-moons-provide.md
magnetised: We had the naming discussion, and decided to stay close to the pg nomenclature. Will add the options to the changeset, thanks

PR #1808, opened 2024-10-07
feat: Restoring shapes only through the `Storage` interface

msfstef:
More radical alternative to https://github.com/electric-sql/electric/pull/1805

Addresses https://github.com/electric-sql/electric/issues/1802

Moves the shape definition to the `Storage` interface.

I've started with just making `ShapeStatus` use the `Storage` for recovering shapes - but if we also store OIDs of the relation we can also do away with storing relations and get rid of any persistence in `ShapeStatus`

I believe this is mergeable if we want to move at least shape definitions to the storage and think about relations in a separate PR

### DIFF
--- a/.changeset/weak-chairs-type.md
+++ b/.changeset/weak-chairs-type.md
@@ -2,4 +2,4 @@
 "@core/sync-service": minor
 ---
 
-[BREAKING] Store shape definitions along with shape data and use that to restore them instead of persisted cached metadata.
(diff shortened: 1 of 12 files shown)


.changeset/weak-chairs-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": minor
+---
+
+[BREAKING] Store shape definitions along with shape data and use that to restore them instead of persisted cached metadata.
```
KyleAMathews, line 5: A note about the benefits too would be nice e.g. speeds this up by removing this bottleneck.

PR #1809, opened 2024-10-08
add one alternative

capaj:


thruflo:
Hey, thanks for this. Zero is awesome and definitely a key alternative. However, I think if we're going to flesh out this page we should add more options.

The previous version of the page was fairly comprehensive: https://legacy.electric-sql.com/docs/reference/alternatives

We want to adapt this to reflect our re-focusing on sync so I don't want to copy over wholesale. But as I say I do want to present an overview of a range of options rather than just one.

Issue #1748, opened 2024-09-24
Add ability to pass headers to the ShapeStream client


ericallam:
We're going to be proxying through our backend to ElectricSQL and would like to be able to add an authorization header on shape requests without having to pass in a custom fetch implementation that injects headers.

KyleAMathews:
Nice idea — this is far and away the most common reason for needing a custom fetch implementation so this would be good API sugar.

PR #1810, opened 2024-10-08
feat: Add `headers` utility option to `ShapeStream` API

msfstef:
Implements feature described in https://github.com/electric-sql/electric/issues/1748

Attaches headers to all requests - figured we'd go with the simple route of static dictionary of headers to add, we can always enhance it with a function dependent on the state of the stream and even potentially make it an async function that waits for the headers to be provided before making the next request - but KISS

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -53,6 +53,13 @@ export interface ShapeStreamOptions<T = never> {
    */
   shapeId?: string
   backoffOptions?: BackoffOptions
+
+  /**
+   * HTTP headers to attach to requests made by the client.
+   * Can be used for adding authentication headers.
+   */
+  headers?: { [key: string]: string }
+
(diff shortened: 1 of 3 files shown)


KyleAMathews (approved):
Love it thanks!

packages/typescript-client/src/client.ts
```diff
@@ -53,6 +53,13 @@ export interface ShapeStreamOptions<T = never> {
    */
   shapeId?: string
   backoffOptions?: BackoffOptions
+
+  /**
+   * HTTP headers to attach to requests made by the client.
+   * Can be used for adding authentication headers.
+   */
+  headers?: { [key: string]: string }
```
kevin-dp, line 61: Can be simplified to `Record<string, string>`.

kevin-dp (approved):
Nice!

PR #1818, opened 2024-10-09
feat: Relation OID as part of `Shape` definition

msfstef:
Follow-up to https://github.com/electric-sql/electric/pull/1808 (and puts the nail in the coffin of https://github.com/electric-sql/electric/issues/1802)

Since we now go to PG for table/schema validation we also grab the relation OID and make that part of the `Shape` definition. I think that's important as a table that gets deleted and recreated would not be the same as what the shape is describing.

Since the shape will now have the relation OID as part of its definition, which also gets persisted, we can track relation changes in the shape consumer itself and determine whether it needs to shut down and clean itself up.

This allows us to remove all persistence from `ShapeStatus`, and remove any replication stream listening logic from the `ShapeCache` which now only plays an orchestration and request routing role.

I've also done some additional cleanup, removing some unused APIs - keeping only the code we use for now to make it easier to read and modify.


robacourt (approved):
Superb! Great work

PR #1819, opened 2024-10-09
jobs: clarify the basis and co-funding of PGlite role.

thruflo:
https://deploy-preview-1819--electric-next.netlify.app/about/jobs/pglite-engineer


samwillis (approved):
👍

Issue #1820, opened 2024-10-09
The integration testing framework does not clear persistent state after tests


alco:
It also shares the persistent state of tests with that of the regular dev environment of the sync-service by running it via `iex -S mix` in `packages/sync-service` as the working directory.

We need to separate the dev state from the test state and make sure the latter gets cleaned up and does not allow interference between separate test runs.

PR #1823, opened 2024-10-09
chore: Fix integration test storage isolation

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1820

We specify the storage directory as being within the integration test folder, and after every test the storage is deleted.


alco (approved):
Awesome! :heart:

Issue #1804, opened 2024-10-07
API parameter to filter columns shape


balegas:
This issue is concerned with extending the Shapes API with an option for filtering the columns that are stored into shape.

A use shape call should look something like this:
```
useShape({
  url: `...`,
  table: `foo`,
  columns: [`id`, `title`, `created_at`]
})

```
and the url should have an argument that is a list of comma-separated columns names that will be included in this shape. It's up to discussion if we want to call the argument `select`, `columns` or `filter`. It's definitely does not have the semantics of SELECT, by the amount of things we can't do with the filter See #1676 -- so we deliberately don't want people to use it as the part of SELECT SQL query.

Some considerations:
- Naming convention should mimic what we do in WHERE clauses (see [#1764](https://github.com/electric-sql/electric/issues/1764#issuecomment-2388956155))
- Configure replication filter with the union of filtered columns (see https://github.com/electric-sql/electric/issues/1774#issuecomment-2386506510)

balegas:
- @robacourt has done some work on the initial snapshot. Sync with him to handover/discuss work
- Note the ongoing work to move table name param into an argument  https://github.com/electric-sql/electric/pull/1798/files

Issue #1814, opened 2024-10-09
Implement PG identifier parsing algorithm


kevin-dp:
As proposed by @alco in https://github.com/electric-sql/electric/issues/1764#issuecomment-2391485961:

> I would argue against going to Postgres for name lookup. Replicating Postgres' identifier parsing is much simpler because it can be implemented as a pure function. Postgres is open source and it's not like its definition of what is an identifier is going to change any time soon: no DB queries or caching required.
> 
> I have to admit that Postgres' definition of the `identifier` token (as specified in its [lexer code](https://github.com/postgres/postgres/blob/259a0a99fe3d45dcf624788c1724d9989f3382dc/src/backend/parser/scan.l#L346-L349)) is a bit cryptic: it explicitly mentions the ranges `A-Z`, `a-z` and byte values in the range `\200-\377` (those are octal numbers, in decimal they are 128 to 255), but I think the intent there is to allow any valid UTF-8 code point beyond the 7-bit ASCII Plane, so not only letters. For example, this works in psql:
> ```
> =# create table 🪷(id text);
> CREATE TABLE
> =# \d
>         List of relations
>  Schema │ Name │ Type  │  Owner   
> ────────┼──────┼───────┼──────────
>  public │ foo¡ │ table │ postgres
>  public │ Δ    │ table │ postgres
>  public │ 🪷   │ table │ postgres
> ```
> 
> There is, apparently, a separate validation to determine that the given byte sequence is valid UTF-8 encoding (or any other encoding that the database server is configured to use) because not all sequences of bytes are valid identifiers.
> 
> Regardless, the algorithm for lower-casing identifiers is very small and straightforward. Here's its full implementation -
> https://github.com/postgres/postgres/blob/259a0a99fe3d45dcf624788c1724d9989f3382dc/src/backend/parser/scansup.c#L55-L73
> 
> So it is conceivable that we can implement our own validation of user-provided table identifiers by checking they are valid UTF-8 strings and don't contain certain ASCII characters. Then we can normalize the identifier by implement Postgres' downcasing algorithm to obtain a string that we can use to fetch table info from Postgres and as a cache key for storing table info in memory.

msfstef:
@alco @kevin-dp the postgres algorithm depends on the server encoding as well - should we grab the encoding in the metadata acquisition stage, or just default to UTF8 for start?

msfstef:
[...]

PR #1829, opened 2024-10-10
feat: Add `columns` API query parameter to filter table columns

msfstef:
Implements https://github.com/electric-sql/electric/issues/1804

### Overview
- I've picked `columns` as the query parameter name, as `select` might confuse people and get compared to SQL `SELECT`s, and `filter` is too close to what the `where` parameter does.
- The columns are comma separated column names, parsed at the validation stage
- If the parameter is specified, it *has to include* the primary key columns
  - We could make it such that the PK is always included, but it feels like it might be confusing - explicit is better?
  - For tables without a primary key, we treat _all_ columns as primary keys - we might want to revisit that w.r.t. to this feature
- Columns are validated also by ensuring that all specified column names match the ones in the schema
- The selected columns are stored in the `Shape` definition as an array of column names
  - We had the choice of potentially filtering the `table_info` data based on the selection of columns instead of storing them separately, but this might cause issues with relation change handling and generally I prefer the definition of the shape to contain all information related to it (i.e. the schema at the time of creation + what column filters where applied, rather than compacting the two)
- I modified `Shape` related APIs for converting changes etc to also apply the column filtering - this is all we need to ensure it works correctly for replication stream log entries
- In the `Snapshotter`, I modified a `get_column_names` method to also apply the filtering if present, which takes care of the snapshot log entries.

### Other changes
- I've changed `Shapes.new` to return the errors along with the field they are associated with (`:root_table` or `:where` or `:columns`) in order to return correct validation errors, since all of the stronger validation occrus at `cast_root_table` when the shape is created and PG is used but it really validates more than just `root_table`.
- I've updated the client to accept a `columns` argument which is a list of strings
- I've updated the OpenAPI spec to include the `columns` parameter

### Things to consider
- How do we want to handle column names with special names (quoted)? In my opinion the client needs to type them exactly as they are on postgres, otherwise they get a 400 validation error back telling them which columns are invalid, so it should be fairly easy to fix.
- Replication publication filtering of columns is available, but [updating it might cause errors](https://www.postgresql.org/docs/15/logical-replication-col-lists.html#LOGICAL-REPLICATION-COL-LIST-COMBINING) according to the PG docs, so I'm not sure if that's something we want to consider

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -101,8 +101,14 @@ defmodule Electric.Plug.ServeShapePlug do
 
     def cast_columns(%Ecto.Changeset{} = changeset) do
       case fetch_field!(changeset, :columns) do
-        nil -> changeset
-        columns -> put_change(changeset, :columns, String.split(columns, ","))
+        nil ->
+          changeset
+
+        columns ->
+          case Electric.Plug.Utils.parse_columns_param(columns) do
+            {:ok, parsed_cols} -> put_change(changeset, :columns, parsed_cols)
+            {:error, reason} -> add_error(changeset, :columns, reason)
+          end
       end
     end
 
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -53,6 +53,7 @@ defmodule Electric.Plug.ServeShapePlug do
       field(:shape_id, :string)
       field(:live, :boolean, default: false)
       field(:where, :string)
+      field(:columns, :string)
       field(:shape_definition, :string)
     end
 
@@ -63,6 +64,7 @@ defmodule Electric.Plug.ServeShapePlug do
       )
       |> validate_required([:root_table, :offset])
       |> cast_offset()
+      |> cast_columns()
       |> validate_shape_id_with_offset()
       |> validate_live_with_offset()
       |> cast_root_table(opts)
@@ -95,6 +97,15 @@ defmodule Electric.Plug.ServeShapePlug do
       end
     end
 
+    def cast_columns(%Ecto.Changeset{valid?: false} = changeset), do: changeset
+
+    def cast_columns(%Ecto.Changeset{} = changeset) do
+      case fetch_field!(changeset, :columns) do
+        nil -> changeset
+        columns -> put_change(changeset, :columns, String.split(columns, ","))
+      end
+    end
+
     def validate_shape_id_with_offset(%Ecto.Changeset{valid?: false} = changeset), do: changeset
 
     def validate_shape_id_with_offset(%Ecto.Changeset{} = changeset) do
@@ -122,18 +133,19 @@ defmodule Electric.Plug.ServeShapePlug do
     def cast_root_table(%Ecto.Changeset{} = changeset, opts) do
       table = fetch_change!(changeset, :root_table)
       where = fetch_field!(changeset, :where)
+      columns = get_change(changeset, :columns, nil)
 
-      case Shapes.Shape.new(table, opts ++ [where: where]) do
+      case Shapes.Shape.new(table, opts ++ [where: where, columns: columns]) do
         {:ok, result} ->
           put_change(changeset, :shape_definition, result)
 
-        {:error, reasons} ->
+        {:error, {field, reasons}} ->
           Enum.reduce(List.wrap(reasons), changeset, fn
             {message, keys}, changeset ->
-              add_error(changeset, :root_table, message, keys)
+              add_error(changeset, field, message, keys)
 
             message, changeset when is_binary(message) ->
-              add_error(changeset, :root_table, message)
+              add_error(changeset, field, message)
           end)
       end
     end
--- a/packages/sync-service/lib/electric/plug/utils.ex
+++ b/packages/sync-service/lib/electric/plug/utils.ex
@@ -17,6 +17,8 @@ defmodule Electric.Plug.Utils do
       {:ok, ["id"]}
       iex> Electric.Plug.Utils.parse_columns_param("beta,alpha")
       {:ok, ["alpha", "beta"]}
+      iex> Electric.Plug.Utils.parse_columns_param(~S|"PoT@To",PoTaTo|)
+      {:ok, ["PoT@To", "potato"]}
       iex> Electric.Plug.Utils.parse_columns_param(~S|"PoTaTo,sunday",foo|)
       {:ok, ["PoTaTo,sunday", "foo"]}
       iex> Electric.Plug.Utils.parse_columns_param(~S|\"fo\"\"o\",bar|)
@@ -34,7 +36,7 @@ defmodule Electric.Plug.Utils do
     # Split by commas that are not inside quotes
     |> String.split(~r/,(?=(?:[^"]*"[^"]*")*[^"]*$)/)
     |> Enum.reduce_while([], fn column, acc ->
-      casted_column = remove_surrounding_quotes(column)
+      casted_column = maybe_cast_quoted_identifier(column)
 
       cond do
         contains_unescaped_quote?(casted_column) ->
@@ -61,9 +63,11 @@ defmodule Electric.Plug.Utils do
     Regex.match?(~r/(?<!")"(?!")/, string)
   end
 
-  defp remove_surrounding_quotes(string) do
-    string
-    |> String.replace(~r/^"(.*)"$/, "\\1")
+  defp maybe_cast_quoted_identifier(string) do
+    if Regex.match?(~r/^"(.*)"$/, string),
+      do: String.replace(string, ~r/^"(.*)"$/, "\\1"),
+      # if identifier is not quoted, downcase it like Postgres would
+      else: Electric.Postgres.Identifiers.downcase(string)
   end
 
   defp unescape_quotes(string) do
--- a/packages/sync-service/lib/electric/plug/utils.ex
+++ b/packages/sync-service/lib/electric/plug/utils.ex
@@ -11,6 +11,8 @@ defmodule Electric.Plug.Utils do
   ## Examples
       iex> Electric.Plug.Utils.parse_columns_param("")
       {:error, "Must specify at least one column"}
+      iex> Electric.Plug.Utils.parse_columns_param("foo,")
+      {:error, "Invalid empty column provided"}
       iex> Electric.Plug.Utils.parse_columns_param("id")
       {:ok, ["id"]}
       iex> Electric.Plug.Utils.parse_columns_param("beta,alpha")
@@ -23,6 +25,10 @@ defmodule Electric.Plug.Utils do
       {:error, ~S|Invalid column, unmatched quote: "id|}
   """
   @spec parse_columns_param(binary()) :: {:ok, [String.t(), ...]} | {:error, term()}
+  def parse_columns_param("") do
+    {:error, "Must specify at least one column"}
+  end
+
   def parse_columns_param(columns) when is_binary(columns) do
     columns
     # Split by commas that are not inside quotes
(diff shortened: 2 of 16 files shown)


robacourt (approved):
Neat!

balegas:
I agree with your choices. The only thing is that we can avoid the errors for PKs, but I prefer erroring.

> How do we want to handle column names with special names (quoted)? In my opinion the client needs to type them exactly as they are on postgres, otherwise they get a 400 validation error back telling them which columns are invalid, so it should be fairly easy to fix.

Please make sure it is same logic as table names, check with @kevin-dp 

> Replication publication filtering of columns is available

I think this is an important thing to have by the reasons we've discussed. Since we have a single subscriber and we know all the columns electric needs, maybe PG behaves. Please open a separate ticket to not block merging this one, but I think is one thing we want to validate it's working.

kevin-dp (changes requested):
Some issues here:
- PG allows any character in column names if you quote them, e.g. `CREATE TABLE foo ("a,column""with-weird name" INT);`. Currently, the parser will misinterpret the comma as a separator. So we should allow column names to be quoted (is consistent with table naming parsing)
- The parser is case sensitive by default. This is inconsistent with schema and table names which are case insensitive unless you quote them.

This would be a good moment to implement https://github.com/electric-sql/electric/issues/1814 such that we have a parse function that we can use for parsing schema identifiers, table identifiers, and column identifiers.

packages/sync-service/lib/electric/shapes/querying.ex
```diff
@@ -57,13 +64,18 @@ defmodule Electric.Shapes.Querying do
...
 
-  defp get_column_names(table_info, root_table) do
+  defp get_column_names(table_info, root_table, nil) do
     table_info
     |> Map.fetch!(root_table)
     |> Map.fetch!(:columns)
     |> Enum.map(& &1.name)
   end
 
+  defp get_column_names(table_info, root_table, selected_columns) do
```
kevin-dp, line 74: Why do you need to filter the `selected_columns` out of the column names you get from table info?
The column names are just the `selected_columns` argument itself.
I understand that this function will only return the column names in `selected_columns` that really exist in the table, but should that case be allowed?
I think we should throw a validation error in the shape plug if the user selects some columns that don't actually exist.
Then, here we can rely on the `selected_columns` and no need to get them out of the table info.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -77,6 +79,15 @@ defmodule Electric.Plug.ServeShapePlug do
       end
     end
 
+    def cast_columns(%Ecto.Changeset{valid?: false} = changeset), do: changeset
+
+    def cast_columns(%Ecto.Changeset{} = changeset) do
+      case fetch_field!(changeset, :columns) do
+        nil -> changeset
+        columns -> put_change(changeset, :columns, String.split(columns, ","))
```
kevin-dp, line 87: We need additional validation to check that if the `columns` argument is provided, at least one valid column is specified.
For example, one could pass `columns=""` which would result in `[""] = String.split("", ",")` but the empty string is not a valid column name.
Probably this will be fixed with our own implementation of the PG identifier parser?

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -68,27 +71,64 @@ defmodule Electric.Shapes.Shape do
...
+    case Parser.parse_and_validate_expression(where, info) do
+      {:ok, expr} -> {:ok, expr}
+      {:error, reason} -> {:error, {:where, reason}}
+    end
+  end
+
+  @spec validate_selected_columns([Inspector.column_info()], [String.t()], [String.t()] | nil) ::
+          {:ok, [String.t(), ...]} | {:error, [String.t()]}
+  defp validate_selected_columns(_column_info, _pk_cols, nil) do
+    {:ok, nil}
```
kevin-dp, line 101: The returned tuple does not satisfy the return type in the spec.

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -68,27 +71,64 @@ defmodule Electric.Shapes.Shape do
...
+  end
+
+  defp validate_selected_columns(column_info, pk_cols, columns_to_select) do
+    missing_pk_cols = pk_cols -- columns_to_select
+    invalid_cols = columns_to_select -- Enum.map(column_info, & &1.name)
+
+    cond do
+      missing_pk_cols != [] ->
```
kevin-dp, line 109: If someone provides a select with missing PK columns AND also invalid columns they will only get the missing PK error. Would be useful for them to also get the invalid columns error.

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -165,8 +210,23 @@ defmodule Electric.Shapes.Shape do
       {false, true} -> [Changes.convert_update(change, to: :new_record)]
       {false, false} -> []
     end
+    |> Enum.map(&filter_change_columns(selected_columns, &1))
```
kevin-dp, line 213: nit: i don't like the case expression being piped into `Enum.map` i would assign its result to a variable and pipe the variable into it.

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -165,8 +210,23 @@ defmodule Electric.Shapes.Shape do
...
+    |> Enum.filter(
+      &case &1 do
```
kevin-dp, line 215: nit: i would remove this lambda and introduce a function `filtered_columns_changed` which pattern matches directly against its argument:
```ex
defp filtered_columns_changed(%Changes.UpdatedRecord{old_record: record, record: record}), do: false
defp filtered_columns_changed(_), do: true
```

packages/typescript-client/src/client.ts
```diff
@@ -193,6 +201,8 @@ export class ShapeStream<T extends Row<unknown> = Row>
       ) {
         const fetchUrl = new URL(url)
         if (where) fetchUrl.searchParams.set(WHERE_QUERY_PARAM, where)
+        if (columns)
```
kevin-dp, line 204: Perhaps also check that it isn't the empty array.

website/electric-api.yaml
```diff
@@ -108,11 +108,25 @@ paths:
...
+
+        - name: columns
+          in: query
+          schema:
+            type: string
+          description: |-
+            Optional list of columns to include in the rows from the `root_table`.
+
+            They should always include the primary key columns, and should be formed
+            as a comma separated list of column names exactly as they are on the database schema.
```
kevin-dp, line 125: i think it should be "as they are **in** the database schema"

kevin-dp (changes requested):
Well done!
I left some comments mostly asking for minor changes.

packages/sync-service/lib/electric/shapes/shape.ex
msfstef: I thought about it but eh, they will get the second error once they fix the first, IMO it's overkill

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
msfstef: This now gets validated in the PG identifier parser where zero-length identifiers are not allowed

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -0,0 +1,50 @@
+defmodule Electric.Plug.Utils do
```
kevin-dp, line 1: Let's perhaps not introduce several utility files but keep everything in the outer `utils.ex` file?

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -0,0 +1,50 @@
...
+    |> String.split(~r/,(?=(?:[^"]*"[^"]*")*[^"]*$)/)
+    |> Enum.reduce_while([], fn column, acc ->
+      case Electric.Postgres.Identifiers.parse(column) do
+        {:ok, casted_column} -> {:cont, [casted_column | acc]}
+        {:error, reason} -> {:halt, {:error, reason}}
+      end
+    end)
+    |> then(fn result ->
+      case result do
+        # sort to keep selected columns identical
```
kevin-dp, line 43: nit: sorting this list seems to be out of scope for this abstraction. This seems to be a concern of the plug rather than this parsing function

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -0,0 +1,50 @@
...
+      {:error, "Invalid zero-length delimited identifier"}
+      iex> Electric.Plug.Utils.parse_columns_param("id")
+      {:ok, ["id"]}
+      iex> Electric.Plug.Utils.parse_columns_param("beta,alpha")
+      {:ok, ["alpha", "beta"]}
+      iex> Electric.Plug.Utils.parse_columns_param(~S|"PoT@To",PoTaTo|)
+      {:ok, ["PoT@To", "potato"]}
+      iex> Electric.Plug.Utils.parse_columns_param(~S|"PoTaTo,sunday",foo|)
+      {:ok, ["PoTaTo,sunday", "foo"]}
+      iex> Electric.Plug.Utils.parse_columns_param(~S|\"fo\"\"o\",bar|)
```
kevin-dp, line 24: Why are the quotes escaped?
You're using a sigil so it should just be `~S|"fo""o",bar|` or am i missing something?

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -0,0 +1,50 @@
...
+      {:ok, ["bar", ~S|fo"o|]}
+      iex> Electric.Plug.Utils.parse_columns_param(~S|"id,"name"|)
+      {:error, ~S|Invalid unquoted identifier contains special characters: "id|}
```
kevin-dp, line 27: why is this saying "unquoted identifier" ? The sigil does contain quotes.
The problem should be that the quote inside the quote is not escaped (with a double quote).

packages/sync-service/lib/electric/postgres/identifiers.ex
```diff
@@ -0,0 +1,127 @@
...
+      else: {:ok, downcase(ident, truncate, single_byte_encoding)}
+  end
+
+  defp contains_unescaped_quote?(string) do
+    Regex.match?(~r/(?<!")"(?!")/, string)
+  end
+
+  defp unescape_quotes(string) do
+    string
+    |> String.replace(~r/""/, "\"")
```
kevin-dp, line 64: nit: `"\""` is a bit confusing, `~s|"|` is easier to read

packages/sync-service/lib/electric/postgres/identifiers.ex
```diff
@@ -0,0 +1,127 @@
...
+
+    downcased_ident
+  end
+
+  def downcase(ident, true, single_byte_encoding) do
+    downcased_ident = downcase(ident, false, single_byte_encoding)
+
+    truncated_ident =
+      if String.length(ident) >= @namedatalen do
+        result = String.slice(downcased_ident, 0, @namedatalen)
```
kevin-dp, line 107: can remove this intermediate variable and just return `String.slice(...)`

kevin-dp (approved):
Brilliant!
Left a few very minor comments.
Ready to merge for my part 💯

packages/sync-service/lib/electric/plug/utils.ex
msfstef: I feel like this is not a general utility like the other ones and is more meant to be part of the Plug code - we already have a very long serve shae plug module which defines submodules, the idea is that parsing logic could move to this.

packages/sync-service/lib/electric/plug/utils.ex
msfstef: It is important that the order does not matter as they are stored within the shape definition which generates a hash. You are right that this is probably not the right place to sort it, and a `MapSet` would be better here but it requires a significant refactor - I could sort it right before initializing the Shape though

packages/sync-service/lib/electric/plug/utils.ex
msfstef: The provided column was `"id` - so I'm treating this as an unquoted identifier that has included a special character `"`

PR #1830, opened 2024-10-10
fix: Handle 429s in the client

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1770

With https://github.com/electric-sql/electric/pull/1787 we've managed to return 429s whenever there's too many concurrent shape creations that cause the database connection pool to be exhausted.

This PR just ensures that the client does indeed retry on 429s - for now just with our regular exponential backoff, as there is no standard for retry headers to respect.

P.S. additional changes to the openapi spec done by my formatter 👀 I can roll them back if you think they are worse than before

### DIFF
--- a/packages/typescript-client/src/fetch.ts
+++ b/packages/typescript-client/src/fetch.ts
@@ -8,6 +8,10 @@ import {
 } from './constants'
 import { FetchError, FetchBackoffAbortError } from './error'
 
+// Some specific 4xx and 5xx HTTP status codes that we definitely
+// want to retry
+const HTTP_RETRY_STATUS_CODES = [429, 500]
+
 export interface BackoffOptions {
   /**
    * Initial delay before retrying in milliseconds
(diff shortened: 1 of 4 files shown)


packages/typescript-client/src/fetch.ts
```diff
@@ -8,6 +8,10 @@ import {
 } from './constants'
 import { FetchError, FetchBackoffAbortError } from './error'
 
+// Some specific 4xx and 5xx HTTP status codes that we definitely
+// want to retry
+const HTTP_RETRY_STATUS_CODES = [429, 500]
```
KyleAMathews, line 13: Why 500 here as it's already covered in the check below?

KyleAMathews (approved):

packages/typescript-client/src/fetch.ts
msfstef: It's already covered so this is unnecessary - but figured since I'm explicitly specifying retry codes here, plain 500s are definitely ones we want to retry on. I'll remove it since realistically all 5xxs are codes we want to retry on anyway.

PR #1834, opened 2024-10-10
Fix: get long_poll_timeout in seconds not milliesconds

KyleAMathews:


msfstef (approved):
might be worth adding a test for the cursor value!

KyleAMathews:
That would be good... I had Gemini make one for me 😎 — we'll see if it works

PR #1841, opened 2024-10-13
fix(sync-service): Start replication client earlier in the startup process

### LINKED ISSUE #1817: Sync service's supervision tree design leads to invalid startup sequences
We have the root supervision tree in the `Application` module but it's really bare bones. The process trees reponsible for maintaining individual shapes are started at a later point in time, on demand. 

The current design and split of different sub-supervisors between files makes the initialization sequence too opaque. As a consequence, we have a failure state caused by a `Snapshotter` process trying to access the database before the DB connection pool has been started. See, for example,
* https://github.com/electric-sql/electric/actions/runs/11250979969/job/31281239838
* https://github.com/electric-sql/electric/actions/runs/11251255328/job/31285173176

To fix this, we need to reorganize the code such that the start sequence for the core processes, such as the replication client and the DB connection pool, is clearly visible in the code and clearly happens before shape-specific supervision trees are started.

alco:
PR chain: #1842 ← **#​1841** ← #1845 ← #1846 ← #1862

This PR restructures the supervision tree design, bringing the management of replication client's lifetime back under Connection Manager. It also fixes a race between the shapes supervisor and the replication client startup that becomes apparent when Postgres is shut down and later restarted again.

Fixes #1817.


packages/sync-service/test/electric/shapes/consumer_test.exs
```diff
@@ -768,6 +768,7 @@ defmodule Electric.Shapes.ConsumerTest do
       %{slot_name: "electric_shapes_consumertest_replication_stream"}
     end
 
+    @tag :skip
```
msfstef, line 771: nit: is this getting fixed in a subsequent PR?

packages/sync-service/lib/electric/connection_manager.ex
```diff
@@ -228,29 +214,24 @@ defmodule Electric.ConnectionManager do
...
 
-  def handle_info({:DOWN, _ref, :process, pid, reason}, %{replication_client_pid: pid} = state) do
-    halt_if_fatal_error!(reason)
-
-    # The replication client will be restarted automatically by the
-    # Electric.Shapes.Supervisor so we can just carry on here.
-    {:noreply, %{state | replication_client_pid: nil}}
+  def handle_info({:DOWN, _ref, :process, pid, reason}, %{shape_log_collector_pid: pid} = state) do
+    Logger.warning("ShapeLogCollector down: #{inspect(reason)}")
+    Electric.Postgres.ReplicationClient.stop(state.replication_client_pid)
```
msfstef, line 233: would this trigger an `:EXIT` which subsequently lets the connection manager restart itself?

msfstef (approved):
Love to see that integration test passing! 🚀

packages/sync-service/test/electric/shapes/consumer_test.exs
alco: Since your review, I have rewritten the commit history in this PR and improved process lifecycle handling in ConnectionManager.

This unit test has been replaced by a new integration test.

packages/sync-service/lib/electric/connection_manager.ex
alco: Exactly.

There are, in fact, multiple possibilities here. If we encounter a storage failure while writing a transaction from the replication stream to shape consumers, replication client will exit together with the ShapeLogCollector process exiting. However, I keep this explicit `stop()` call here to cover scenarios where shape log collector exists at a point that doesn't itself cause replication client to exit as well.

I've updated this part of the code with documentation comments and better error handling.

magnetised (approved):
Very nice. Feels very consistent, though that connection manager state machine still gives me the heebie jeebies.

Issue #1591, opened 2024-08-28
Add multi-database support


KyleAMathews:
I.e. each Electric server can be serving arbitrary number of Postgres databases (though we'll probably want to have a default limit).

## Background

There are many use cases for multi-tenancy. A self-hosted Electric server might have a staging and prod database that it fronts. A company might have multiple production databases they sync from. Electric Cloud will need this to efficiently host large numbers of customers.

## Proposal

Right now we start the server with a `DATABASE_URL` environment variable to set the backend. This will still be supported but we'll create an admin API for CRUDing backing databases.

Each database needs a Postgres connection string and an id. If an ID is not included, a UUID will be generated for the id. E.g. the original `DATABASE_URL` would have a UUID generated for it.

### Managing databases

You can start the database with an `ADMIN_TOKEN` to authenticate admin API calls. If one is provided, API calls must include it in the `Authorization` header.

To create an instance:

```
curl -X POST http://localhost:3000/admin/database \
-H "Content-Type: application/json" \
-d '{
  "UUID": "...",
  "DATABASE_URL": "postgresql://username:password@host:port/database"
}' \
-H "Accept: application/json"
```

UUID again is optional.

Can we complete setting up the new backend db within the span of the request? Ideally we just hold the connection until it's setup. If there's a timeout, the caller should be able to GET the instance to see the status (or follow the shape for database updates).

To read an instance:
```
curl -X GET "http://localhost:3000/admin/database/[UUID]"
```

To delete an instance:
```
curl -X DELETE "http://localhost:3000/admin/database/[UUID]"
```

Can we complete cleaning up the deleted backend db within the span of the request?

### Shape API change

Right now the path is `/v1/shape/${baseTable}` with query parameters.

With multi-tenancy, it'll be `/v1/shape/${baseTable}?database_id=${id}`. So this is a non-breaking change. To further make it non-breaking, if an Electric server only has one database instance, then `database_id` isn't required (though allowed).

We'll eventually want per-tenant controls e.g. max shapes, max shape log size, etc. that'll probably come along with https://github.com/electric-sql/electric/issues/1529

### Logs & telemetry
We'll need to add the database_id to the start of logs and as attributes on spans. This would also be a good time to add a JSON log option for easier parsing.

### Internal state shape
Not required for this issue but eventuallyl we'll want to expose admin shapes for monitoring various parts of the internal state. That way you can build monitoring/admin tools on top of standard Electric primitives and see live updates on what's happening.

[...]

PR #1844, opened 2024-10-14
feat (sync-service): multi tenancy support

kevin-dp:
This PR implements https://github.com/electric-sql/electric/issues/1591.

TODO:

- [ ] rebase on top of main
- [ ] make unit tests for multi tenancy
- [ ] add a diagram explaining the new architecture
- [ ] update the open API spec


packages/sync-service/lib/electric/plug/create_database_plug.ex
```diff
@@ -78,7 +78,7 @@ defmodule Electric.Plug.AddDatabasePlug do
       :ok = TenantManager.create_tenant(tenant_id, connection_opts, conn.assigns.config)
 
       conn
-      |> send_resp(200, tenant_id)
+      |> send_resp(200, Jason.encode_to_iodata!(tenant_id))
```
alco, line 81: Isn't tenant_id just a string? Why does it need to be encoded into JSON?

kevin-dp: The response should be JSON. `tenant_id` is indeed a string but if we just send it like that it won't be valid JSON because it will be missing the quotes around the string. So i chose to encode it with Jason but we could manually surround it with quotes as well but need to be careful if the string contains quotes itself. So figured it's easier to just use Jason for this.

PR #1845, opened 2024-10-14
chore(sync-service): Move queries that query Postgres for a bunch of info into ReplicationClient's connection setup phase

alco:
PR chain: #1842 ← #1841 ← **#​1845** ← #1846 ← #1862

Basically, cleaning up ad-hoc queries by performing them in the dedicated connection setup phase, prior to switching the replication connection into logical message streaming mode.

PR #1842, opened 2024-10-13
Alco/application configuration

alco:

PR #1846, opened 2024-10-14
chore(sync-service): Decouple Timeline from ShapeCache

alco:
PR chain: #1842 ← #1841 ← #1845 ← **#​1846** ← #1862

See individual commits to get a better idea of what this PR achieves.


msfstef (approved):
Nice! I think we can actually get rid of `persistent_kv` altogether and make `Timeline` write directly to a single file (it was used for `ShapeStatus` before but I've removed that, so it is now only used by `Timeline`) - we can do that in a separate PR

alco:
Merged into `main` manually as a chain of commits, one per PR.
https://github.com/electric-sql/electric/pull/1867

PR #1850, opened 2024-10-14
Alco/acquire lock in replication client

### LINKED ISSUE #1792: Electric no longer performs automatic fallback to disabling SSL for database connections
Prior to #1762, Electric tried to connect to the database using SSL first, then fall back to using an unencrypted connection unless the DATABASE_URL config had the `sslmode` query parameter in it.

#1762 made changes to the DB connection setup and, as an accidental side effect, Electric no longer performs the automatic fallback. I believe this wouldn't have happened if the fallback logic hadn't been moved from `ConnectionManager` to `ReplicationClient` in #1594.

To fix this regression, we need to move the fallback logic back to `ConnectionManager` and add an integration test to preclude future regressions.

alco:
PR chain: #1842 ← #1841 ← #1845 ← #1846 ← **#​1850**

Fixes #1792.


packages/sync-service/lib/electric/postgres/replication_client/connection_setup.ex
```diff
@@ -23,28 +23,45 @@ defmodule Electric.Postgres.ReplicationClient.ConnectionSetup do
...
 
   ###
 
+  # TODO: add description of the locking design from LockConnection
+  defp acquire_lock_query(state) do
+    query = "SELECT pg_advisory_lock(hashtext('#{state.slot_name}'))"
+    {:query, query, state}
+  end
+
+  defp acquire_lock_result([%Postgrex.Result{columns: ["pg_advisory_lock"]}], state) do
```
msfstef, line 59: would we ever need to handle a `Postgrex.Error` from the advisory lock query? I imagine disconnections etc are handled by the replication client and connection manager, but I don't know if this query can ever fail

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -93,7 +92,8 @@ defmodule Electric.Postgres.ReplicationClient do
     # the connection error. Without this, we may observe undesirable restarts in tests between
     # one test process exiting and the next one starting.
     connection_opts =
-      [auto_reconnect: false] ++ Electric.Utils.deobfuscate_password(connection_opts)
+      [auto_reconnect: false, sync_connect: false] ++
```
msfstef, line 95: aha! is the `sync_connect: false` what allows us to not freeze the `ConnectionManager` while the `ReplicationClient` is going through the steps, or is this unrelated?

packages/sync-service/test/electric/postgres/lock_connection_test.exs
```diff
@@ -6,6 +6,7 @@ defmodule Electric.Postgres.LockConnectionTest do
   alias Electric.Postgres.LockConnection
 
   @moduletag :capture_log
+  @moduletag :skip
```
msfstef, line 9: I assume you intend to "move" the lock tests to the replication client tests later?

msfstef (approved):
Nice and clean! 🚀 🔒

packages/sync-service/lib/electric/postgres/replication_client.ex
alco: Yeah, this is the crucial trick. `Postgrex.ReplicationConnection` by default has `sync_connect` set to `true` and that results in it being stuck in the initialisation phase until the connection process switches to the streaming mode. In other words, calling `Postgrex.ReplicationConnection.start_link()` blocks the calling process until the connection process starts streaming.

Setting `sync_connect` to `false` makes the `ReplicationConnection` behaviour behave more like a regular gen server: with this setting `Postgrex.ReplicationConnection.start_link()` returns right after the newly started connection process returns from executing the `init()` callback.

alco:
After working a bit more on connection process life cycle management in #1841 I've decided to scratch this change. Even though moving the advisory lock acquisition to the replication connection would help us save one idle PG connection, it would rob us of the ability to restart the replication client independently of the other connections, namely, the lock connection and the DB pool.

If we merged this PR, restarting the replication client process (which we do to ensure consistent storaged of streamed transactions by shape consumers) would always necessitate restarting the whole `Electric.Connection.Supervisor` tree.

packages/sync-service/lib/electric/postgres/replication_client.ex
msfstef: the missing piece of the puzzle... nice!

Issue #1837, opened 2024-10-11
Fix race condition for live-polling cursors


KyleAMathews:
Follow-up to https://github.com/electric-sql/electric/pull/1826 after some more testing

There's still a slight chance that a live request is cached too long as the first long-polling request might get the same cursor as the second given networking inconsistencies, etc.

The fix is that when creating the cursor, the server should check what cursor was used by the current request and if the two cursors are the same, it should append something to ensure the cursor busts the caches.

PR #1856, opened 2024-10-15
Fix: prevent the server from returning the same cursor as it received

KyleAMathews:
Fix https://github.com/electric-sql/electric/issues/1837

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -36,6 +36,18 @@ defmodule Electric.Plug.ServeShapePlug do
           diff_in_seconds = DateTime.diff(now, @oct9th2024, :second)
           next_interval = ceil(diff_in_seconds / long_poll_timeout_sec) * long_poll_timeout_sec
 
+          # Check if the generated cursor is the same as what's passed in.
+          cursor = conn.query_params["cursor"]
+
+          next_interval =
+            if cursor && "#{next_interval}" == cursor do
+              # Generate a random integer between 0 and 99999
+              random_integer = :rand.uniform(100_000)
+              next_interval + random_integer
+            else
+              next_interval
+            end
+
           next_interval
       end
     end
--- a/packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
+++ b/packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
@@ -71,7 +71,10 @@ defmodule Electric.Plug.ServeShapePlugTest do
     test "seconds_since_oct9th_2024_next_interval" do
       # Mock the conn struct with assigns
       # 20 seconds
-      conn = %Plug.Conn{assigns: %{config: %{long_poll_timeout: 20000}}}
+      conn = %Plug.Conn{
+        assigns: %{config: %{long_poll_timeout: 20000}},
+        query_params: %{"cursor" => nil}
+      }
 
       # Calculate the expected next interval
       now = DateTime.utc_now()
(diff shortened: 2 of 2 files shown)


packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
```diff
@@ -71,7 +71,10 @@ defmodule Electric.Plug.ServeShapePlugTest do
     test "seconds_since_oct9th_2024_next_interval" do
```
msfstef, line 71: nit: these tests would be better placed in a separate suite since they are only testing `seconds_since_oct9th_2024_next_interval` - the shape test should check end-to-end a request being returned by the plug

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -36,6 +36,18 @@ defmodule Electric.Plug.ServeShapePlug do
           diff_in_seconds = DateTime.diff(now, @oct9th2024, :second)
           next_interval = ceil(diff_in_seconds / long_poll_timeout_sec) * long_poll_timeout_sec
 
+          # Check if the generated cursor is the same as what's passed in.
+          cursor = conn.query_params["cursor"]
+
+          next_interval =
+            if cursor && "#{next_interval}" == cursor do
+              # Generate a random integer between 0 and 99999
+              random_integer = :rand.uniform(100_000)
```
msfstef, line 45: I don't get it - if we're doing something random here why is this a responsibility of the server? The only reason to create a cache buster on the server is to coordinate clients, otherwise the clients can bust the cache themselves - this behaviour does not guarantee any collapsing between the random cache busting requests.

If this approach is not yielding the desired result perhaps we should re-examine it more closely? Perhaps we need to use a slightly smaller time frame than `long_poll_timeout_sec` to avoid edge cases? Perhaps the client can keep track of the frequency of live requests and prevent this behaviour rather than making it the responsibility of the server?

msfstef (changes requested):

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
msfstef: But I still don't see why this resolution is a sync-service concern? Ultimately the "user issue" is that the client might start doing tons of live requests one after another because it is getting a cached response - perhaps clients should bust their cache if they get a 204 within too short a time frame instead?

Either way my main concern is that this is the server arbitrarily busting the cache after we implemented a solution for the problem that would coordinate clients, so perhaps the previous solution is not working as expected.

Do we have any measurement of how often it ends up in this random cache busting clause? Is it an actual fallback or do more than half of the requests end up there?

KyleAMathews: The race condition isn't common and that's too reproduce but both @samwillis and I saw it. 

The point of doing it on the server is that it maintains request collapsing by moving all clients connected to the same http proxy to the same subsequent request. It's harder to coordinate client generated cache busting vs on the server. 

Also there'll be maybe client implementations and the logic for just plopping on the returned cursor to each new request is easy to implement as it's the same idea as the offset.

msfstef: Ah I see yeah the clients will still be coordinated by the random request since they all collapse to the same generated one - was missing that piece of info!

msfstef (approved):
Looks good!

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
KyleAMathews: Yeah, it's all an annoyingly elaborate workaround for Cloudflare. Perhaps some other http proxy has this issue but we only discovered it with Cloudflare.

Issue #1838, opened 2024-10-11
Requests to the beginning of the shape log (offset=-1) should have a really long cache


KyleAMathews:
Perhaps 1 month for the max-age and 3 months more for SWR

`cache-control: public, max-age=2629746, stale-while-revalidate=7889238`

It's dramatically faster if the browser pulls shapes out of their own cache and almost as fast to read from the CDN and the slowest by far to read off the Electric cache.

We can do this as shape logs are immutable and append-only. The worse case scenario is that a shape is requested when it has no rows and then grows a lot — but even this isn't that bad as a client can make multiple requests in a row fairly quickly and would still get the bulk of the data in the 2nd request.

Also in the worst-case scenario, we could add a way for Electric to purge shape paths in caches if they're badly out-of-date.

Related to https://github.com/electric-sql/electric/issues/1447

Also related to improving chunking as a stable initial chunk would also mean that caches don't need to expire.

KyleAMathews:
Ok thinking about this more — we essentially want the user's private cache to keep the segment of the shape log ~forever. Browsers store the parsed JSON so loading that from cache is very fast. E.g. 1mb of cached JSON on disk loads on my laptop in around ~20-40ms. Far faster than a browser could ever load from the CDN.

And even if it ends up that for a long-lived shape that a browser has cached offsets 1-10k and then 10k-57k and then 57k-123k or whatever, all of those will load in ~100ms. So very fast to get the sequence. So it's better it does that vs. going back to reload 1-123k.

So we want max-age to be set to a very high number so browsers cache JSON segments.

But we *don't* want shared HTTP caches to cache for that long. Because when a client is loading a log from scratch, it is far better for them if they load complete logs vs. loading one and then going back and grabbing another one, etc. (though we do some nice optimizations here still e.g. immediately start fetching the next segment from the offset in the headers).

So we want to use `s-maxage` for this as that controls the TTL of the cache in shared caches like CDNs. We could set max-age for private caches in clients to 3 months, s-maxage for shared caches to 4 hours, and then stale-while-revalidate also to 3 months so the shared cache will still quickly respond with old but rarely fetched shape logs to serve quickly but will still re-validate them frequently against Electric so shapes that change frequently are mostly up-to-date on the shared cache.

KyleAMathews:
[...]

PR #1857, opened 2024-10-15
Set a much longer max-age for -1 offsets and don't return 

KyleAMathews:
Fix for https://github.com/electric-sql/electric/issues/1838

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -384,39 +384,38 @@ defmodule Electric.Plug.ServeShapePlug do
     end
   end
 
-  defp put_resp_cache_headers(
-         %Conn{assigns: %{offset: offset, config: config, live: live}} = conn,
-         _
-       ) do
-    # If the offset is -1, set a 1 week max-age, 1 hour s-maxage (shared cache) and 1 month stale-while-revalidate
-    # We want private caches to cache the initial offset for a long time but for shared caches to frequently revalidate
-    # so they're serving a fairly fresh copy of the initials shape log.
-    if offset == @before_all_offset do
+  # If the offset is -1, set a 1 week max-age, 1 hour s-maxage (shared cache) and 1 month stale-while-revalidate
+  # We want private caches to cache the initial offset for a long time but for shared caches to frequently revalidate
+  # so they're serving a fairly fresh copy of the initials shape log.
+  defp put_resp_cache_headers(%Conn{assigns: %{offset: @before_all_offset}} = conn, _),
+    do:
       conn
       |> put_resp_header(
         "cache-control",
         "public, max-age=604800, s-maxage=3600, stale-while-revalidate=2629746"
       )
-    else
-      if live do
-        conn
-        |> put_resp_header(
-          "cache-control",
-          "public, max-age=5, stale-while-revalidate=5"
-        )
-        |> put_resp_header(
-          "electric-next-cursor",
-          TimeUtils.seconds_since_oct9th_2024_next_interval(conn) |> Integer.to_string()
-        )
-      else
-        put_resp_header(
-          conn,
-          "cache-control",
-          "public, max-age=#{config[:max_age]}, stale-while-revalidate=#{config[:stale_age]}"
-        )
-      end
-    end
-  end
+
+  # For live requests we want shorrt cache lifetimes and to update the live cursor
+  defp put_resp_cache_headers(%Conn{assigns: %{live: true}} = conn, _),
+    do:
+      conn
+      |> put_resp_header(
+        "cache-control",
+        "public, max-age=5, stale-while-revalidate=5"
+      )
+      |> put_resp_header(
+        "electric-next-cursor",
+        TimeUtils.seconds_since_oct9th_2024_next_interval(conn) |> Integer.to_string()
+      )
+
+  # For all other requests use the configured cache lifetimes
+  defp put_resp_cache_headers(%Conn{assigns: %{config: config, live: false}} = conn, _),
+    do:
+      conn
+      |> put_resp_header(
+        "cache-control",
+        "public, max-age=#{config[:max_age]}, stale-while-revalidate=#{config[:stale_age]}"
+      )
 
   def cors(conn, _opts) do
     conn
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -330,7 +330,6 @@ defmodule Electric.Plug.ServeShapePlug do
          } = conn,
          _
        ) do
-
     # The log can't be up to date if the last_offset is not the actual end.
     # Also if client is requesting the start of the log, we don't set `up-to-date`
     # here either as we want to set a long max-age on the cache-control.
@@ -390,7 +389,7 @@ defmodule Electric.Plug.ServeShapePlug do
       conn
       |> put_resp_header(
         "cache-control",
-        "public, max-age=604800, s-maxage=3600 stale-while-revalidate=2629746"
+        "public, max-age=604800, s-maxage=3600, stale-while-revalidate=2629746"
       )
     else
       if live do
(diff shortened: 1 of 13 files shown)


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -419,7 +429,7 @@ defmodule Electric.Plug.ServeShapePlug do
             up_to: chunk_end_offset
           )
 
-        [snapshot, log, maybe_up_to_date]
```
msfstef, line 422: why remove this? this is breaking a bunch of tests and I can't easily see why we would remove it?

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -366,22 +366,33 @@ defmodule Electric.Plug.ServeShapePlug do
   end
 
   defp put_resp_cache_headers(%Conn{assigns: %{config: config, live: live}} = conn, _) do
-    if live do
+    # If the offset is -1, set a 1 week max-age, 1 hour s-maxage (shared cache) and 1 month stale-while-revalidate
+    # We want private caches to cache the initial offset for a long time but for shared caches to frequently revalidate
+    # so they're serving a fairly fresh copy of the initials shape log.
+    if conn.query_params["offset"] == "#{@before_all_offset}" do
```
msfstef, line 372: this would be better and more idiomatically handled with multiple pattern matched clauses on `put_resp_cache_headers`

I think since we want different cache times for different responses this might warrant a small but separate module that also has defined aliases for the cache ages we want to use

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -366,22 +366,33 @@ defmodule Electric.Plug.ServeShapePlug do
...
+        )
+        |> put_resp_header(
+          "electric-next-cursor",
+          TimeUtils.seconds_since_oct9th_2024_next_interval(conn) |> Integer.to_string()
+        )
+      else
+        put_resp_header(
+          conn,
+          "cache-control",
+          "public, max-age=#{config[:max_age]}, stale-while-revalidate=#{config[:stale_age]}"
```
msfstef, line 393: we are currently configuring a max age and stale age, but the way this is evolving the configuration is only for non-initial, non-live requests

perhaps it would be better to encapsulate this business logic in its own module and also perhaps it would be good to parametrize all cache times on these configurations?

msfstef (changes requested):
I think at this point it warrants encapsulating the caching logic in a small module where we have better visibility and control over the various parameters, and also add a couple of tests on these

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
KyleAMathews: You can't have a long cache for a shape segment which also includes an up-to-date message.

KyleAMathews: See the discussion on the attached issue

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
KyleAMathews: A refactor sounds like a good follow up issue but it's beyond my elixir capacity and I don't think it should block this change.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
KyleAMathews: The only reason there's config for Max age and SWR is for tests. These aren't values that people need to tweak. So not opposed to a refactor but that's a separate PR.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
msfstef: Then it makes more sense to pass the completed response through a dedicated piece of caching logic and check if the response contains an up to date and set the cache time appropriately? If we chunk the initial data as well you would only ever get a -1 request with an up to date if the initial request is small enough

Removing this here will break a ton of things, and it's not removing the header that also marks if something is up to date either

KyleAMathews: What does this break?

The idea here is this is a quick intermediate step on the way to a more comprehensive solution

See https://github.com/electric-sql/electric/issues/1838#issuecomment-2409013618 and https://github.com/electric-sql/electric/issues/1447

msfstef: You can look at the test suites failing on CI, but generally the API so far has been assumed to include an up-to-date message when you receive all the latest data so I suspect a ton of tests will rely on that.

Also to make this have the correct behaviour it's better to put this logic in `determine_up_to_date` in line 314 rather than in the serving logic, so the headers are also correctly updated

KyleAMathews: > the API so far has been assumed to include an up-to-date message when you receive all the latest data so I suspect a ton of tests will rely on that.

Sure yeah I'll fix any other tests affected. Just to be clear though, the previous logic was wrong — when fetching an immutable log, there's two concerns we're balancing — 1) we want as much as the log to be cached both on end clients and on intermediate http caches and 2) we want end clients to have strong confidence they're up-to-date with the log. You can't really combine the two into one request.

If you put `up-to-date` in the initial request, then you have to have a short max-age as otherwise, clients will think they have the latest data when they don't. Also CDNs are currently dropping their cache as well constantly so too many requests are going back to the server.

So the solution is to do two requests. The first is heavily cached so mostly gets resolves either in the client itself or on the CDN so is consistently very fast — and the second is a cheap almost-no-data-returned request that gets quickly resolved by the CDN or end electric server.

In my testing, even larger 4-5 MB requests load from the cache in < 100ms and then the 2nd "am I up to date?" request returns also in < 100ms for < 200ms total. For the current logic, because of the short max-age, for many visits to an app, both the client and CDN cache have expired which means a 5-6s slow download from Electric.

KyleAMathews: > Also to make this have the correct behaviour it's better to put this logic in determine_up_to_date in line 314 rather than in the serving logic, so the headers are also correctly updated

Ah thanks, I'll try that.

KyleAMathews:
@msfstef ok something is goofy here that I can't figure out. The elixir tests were pretty easy to modify as they do raw http calls but the typescript client, especially the prefetching is having a thing. I'm seeing `getNextChunkUrl` get called twice and two fetches get called concurrently.

Can we grab some time tomorrow to figure this out?

KyleAMathews:
Perhaps the typescript client doesn't like an empty initial response? All my manual testing things worked great but that was always with shapes that had rows in it.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -377,23 +384,37 @@ defmodule Electric.Plug.ServeShapePlug do
...
-  defp put_resp_cache_headers(%Conn{assigns: %{config: config, live: live}} = conn, _) do
-    if live do
+  defp put_resp_cache_headers(
+         %Conn{assigns: %{offset: offset, config: config, live: live}} = conn,
+         _
+       ) do
+    # If the offset is -1, set a 1 week max-age, 1 hour s-maxage (shared cache) and 1 month stale-while-revalidate
+    # We want private caches to cache the initial offset for a long time but for shared caches to frequently revalidate
+    # so they're serving a fairly fresh copy of the initials shape log.
+    if offset == @before_all_offset do
```
msfstef, line 394: I think that `put_resp_cache_headers` should be the last in the plug pipeline, and instead on relying on `offset` to decide this it should rely on whether `conn.assigns.up_to_date` is present, since semantically that is really what we want.

The "don't send up to date for offset -1" logic is in the "determine up to date" call, here we should be checking if it is up to date.

msfstef (approved):
I've updated the tests for the client, there were some issues as now you always have 2 initial requests to get to up to date and also the cache tests were failing as initial requests are no longer using the configured max age and stale age.

robacourt (approved):

KyleAMathews:
Thanks @msfstef !

PR #1859, opened 2024-10-16
fix: Parse and validate `REPLICATION_STREAM_ID`

msfstef:
While setting up another project, I noticed that replication slots _cannot_ contain special characters, so parsing and validating the provided id early will prevent unnecessary headaches.


magnetised (approved):
👍🏻

PR #1861, opened 2024-10-16
docs: wip on benchmarks reference page.

thruflo:


website/docs/reference/benchmarks.md
```diff
@@ -0,0 +1,144 @@
...
+Electric is designed to be simple, scalable and low-cost to run. This means that it should handle a high number of concurrent users, with high read and write workloads, corresponding to high throughput of data from Postgres out to many clients, through various shape topologies, for a variety of workloads.
+
+Our goals are to support millions of concurrent users, hundreds of thousands of shapes and high read and write throughput from a single Electric instance, with minimal impact on the source Postgres. With low, stable and predictable compute and memory usage. In order to work towards this, we use benchmarks internally to measure and inform work to improve performance.
+
+This page lists a selection of these benchmarks, purely to give you an indication of:
+
+1. the types of benchmarks that the Electric team is running
+2. the performance levels that the Electric team are seeing for their internal benchmark runs
+
+> [!Warning] Benchmarks are always highly **workload**, **version** and **hardware** dependent.
```
samwillis, line 34: Did you intend there to be bold formatting? It renders the asterisks.

website/docs/reference/benchmarks.md
```diff
@@ -0,0 +1,144 @@
...
+import WriteFanoutMemory from '/static/img/benchmarks/write-fanout-memory.png?url'
+</script>
+
+# Benchmarks
+
+We run benchmarks to measure and help improve the Electric's performance.
+
+## Context
+
+We run benchmarks for both the [Electric sync engine](/product/sync) directly, and the [Electric Cloud](/product/cloud), which runs the sync engine behind a CDN.
```
samwillis, line 21: "run(s)" twice in one sentence, maybe "hosts"?

```suggestion
We run benchmarks for both the [Electric sync engine](/product/sync) directly, and the [Electric Cloud](/product/cloud), which hosts the sync engine behind a CDN.
```

website/docs/reference/benchmarks.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Running benchmarks yourself
+
+We are in the process of open sourcing our [electric-sql/benchmarking-fleet](https://github.com/electric-sql/benchmarking-fleet) repo. When public, you will be able to follow the instructions there to run the benchmarks yourself.
+
+### Continuous integration
+
+We are working to set up benchmarks to run on every release (patch, minor and major). When this is done, we will document how to see the release benchmarks and how to track improvements and/or regression in performance.
+
+## Benchmarks
```
samwillis, line 47: The heading levels seem off, we get an extra dividing line here as the two headings are adjacent and the same level.

Maybe we need to drop the level of all headings below by one.

samwillis (approved):
I think this is a really good start, and worth posting as is. But a little more analysis, explaining what the metrics mean in the context of an end user may be helpful.

However, this is mostly to show we're taking this seriously, not about saying use us because of these numbers, so it doesn't need to be overly analysed.

PR #1868, opened 2024-10-17
feat (sync-service): multi tenancy support

alco:
_Supersedes #1844. I kept that PR up to consult its diff until all the issues have been resolved here._

This PR implements #1591.

TODO:

- [x] rebase on top of main
- [ ] review commented out code and either remove it or make sure the place it had been moved to has been updated for multitenancy
- [ ] Implement a two-tier ETS storage for column info and relation data. In this PR we replace two named tables with anonymous ones, necessitating a roundtrip to a single gen server just to fetch the table handle. This gen server becomes a bottleneck. We should either create another static ETS table to lookup individual ETS table handles in or store all column info and relation info in two static (named) tables, ensuring they don't trump each other via proper key design.
- [ ] make unit tests for multi tenancy
- [ ] add a diagram explaining the new architecture
- [ ] update the open API spec

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -9,6 +9,7 @@ defmodule Electric.Plug.ServeShapePlug do
 
   alias Electric.Shapes
   alias Electric.Schema
+  alias Electric.TenantManager
   alias Electric.Replication.LogOffset
   alias Electric.Telemetry.OpenTelemetry
   alias Plug.Conn
@@ -63,6 +64,7 @@ defmodule Electric.Plug.ServeShapePlug do
       field(:root_table, :string)
       field(:offset, :string)
       field(:shape_id, :string)
+      field(:database_id, :string)
       field(:live, :boolean, default: false)
       field(:where, :string)
       field(:columns, :string)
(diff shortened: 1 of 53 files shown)


alco:
@kevin-dp This is where this PR is currently at after I have rebased it on top of `main`:
- Elixir unit tests all pass
- lux-based integration tests fail. We need to create a test tenant for integration tests
- TS-based integration tests fail for the same reason: missing test tenant
- there's a block of commented-out test code I haven't yet resolved. Those are tests that were moved to other modules after you had started the original multi-tenancy PR. We just need to make sure that those moved tests haven't been lost in the rebasing process.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -192,6 +196,51 @@ defmodule Electric.Plug.ServeShapePlug do
   # end_telemetry_span needs to always be the last plug here.
   plug :end_telemetry_span
 
+  defp validate_tenant_id(%Conn{} = conn, _) do
```
KyleAMathews, line 199: tenant_id needs to be a UUID so we need to validate that here.

alco:
I've replaced the test tenant with a configurable default tenant that works in any environment. All tests are green now.

The code still needs a good cleanup and deduplication pass. I think we can avoid trickling `electric_instance_id` and `tenant_id` to so many module by passing a partially-applied `process_name()` function under the `name` option instead.

PR #1873, opened 2024-10-21
chore: Validate table names locally first

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1814

We still go to PG to validate the table as we need its OID, so this is not strictly necessary, but since we have validation logic we might as well validate things locally to avoid going to PG for clearly invalid table identifiers?

This also fixes the accepted characters by PG for identifiers [based on the docs](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) as it was excluding some unicode chars and dollar signs before.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/identifiers.ex
+++ b/packages/sync-service/lib/electric/postgres/identifiers.ex
@@ -2,6 +2,30 @@ defmodule Electric.Postgres.Identifiers do
   @namedatalen 63
   @ascii_downcase ?a - ?A
 
+  defmodule StringSplitter do
+    @moduledoc """
+    Utility module for splitting strings on a schema delimiter
+    """
+    def split_outside_quotes(string) do
+      split_outside_quotes(string, "", [], false)
+    end
+
+    # Return the accumulated parts
+    defp split_outside_quotes("", current, acc, _in_quotes), do: acc ++ [current]
+
+    # If we hit a period and we're not in quotes, split here
+    defp split_outside_quotes(<<".", rest::binary>>, current, acc, false),
+      do: split_outside_quotes(rest, "", acc ++ [current], false)
+
+    # Toggle the in_quotes flag when encountering a quote
+    defp split_outside_quotes(<<"\"", rest::binary>>, current, acc, in_quotes),
+      do: split_outside_quotes(rest, current <> "\"", acc, !in_quotes)
+
+    # Continue processing, accumulating characters
+    defp split_outside_quotes(<<char, rest::binary>>, current, acc, in_quotes),
+      do: split_outside_quotes(rest, current <> <<char>>, acc, in_quotes)
+  end
+
   @doc """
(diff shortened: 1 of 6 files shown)


balegas:
That is a cool little thing. Nice!

robacourt (approved):

packages/sync-service/lib/electric/postgres/identifiers.ex
```diff
@@ -65,7 +98,52 @@ defmodule Electric.Postgres.Identifiers do
...
+      [table] ->
+        case parse(table) do
+          {:ok, parsed} -> {:ok, {"public", parsed}}
+          {:error, reason} -> {:error, reason}
+        end
+
+      [_, _] = idents ->
+        Enum.reduce_while(idents, {:ok, {}}, fn ident, {:ok, schema} ->
+          case parse(ident) do
+            {:ok, parsed} -> {:cont, {:ok, schema |> Tuple.append(parsed)}}
```
kevin-dp, line 139: Do we need `Tuple.append` or can we write `{schema, parsed}` ?

packages/sync-service/lib/electric/postgres/identifiers.ex
```diff
@@ -2,6 +2,30 @@ defmodule Electric.Postgres.Identifiers do
   @namedatalen 63
   @ascii_downcase ?a - ?A
 
+  defmodule StringSplitter do
+    @moduledoc """
+    Utility module for splitting strings on a schema delimiter
+    """
+    def split_outside_quotes(string) do
```
kevin-dp, line 9: Can we put a typespec on this public function

packages/sync-service/lib/electric/postgres/identifiers.ex
```diff
@@ -65,7 +98,52 @@ defmodule Electric.Postgres.Identifiers do
...
+  def parse_relation(ident) do
+    case StringSplitter.split_outside_quotes(ident) do
+      [table] ->
+        case parse(table) do
+          {:ok, parsed} -> {:ok, {"public", parsed}}
+          {:error, reason} -> {:error, reason}
+        end
+
+      [_, _] = idents ->
+        Enum.reduce_while(idents, {:ok, {}}, fn ident, {:ok, schema} ->
```
kevin-dp, line 137: Do we need a `reduce_while` if there are only 2 identifiers?
Could we use a with statement instead?

```ex
[schema, tbl] = idents ->
  with {:ok, schema} <- parse(schema)
       {:ok, table} <- parse(tbl)
  do
    {:ok, {schema, table}}
  else
    err -> err
```

kevin-dp (approved):
Nice work!
I left a few minor comments.

packages/sync-service/lib/electric/postgres/identifiers.ex
msfstef: Yeah I think this is a clearer approach - will do this

Issue #1875, opened 2024-10-21
Typescript client tests are particularly flaky


msfstef:
On almost all PRs we get flakes/failed TS tests, we should address whatever the underlying issue is in order for the tests to work consistently.

PR #1877, opened 2024-10-21
chore: Fix flaky TS tests

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1875

Attempting to reduce flakiness - still getting some errors I can't quite pinpoint every now and then but seems like an improvement


KyleAMathews (approved):
Nice! Thanks for looking into these 🙏

Issue #1876, opened 2024-10-21
Fix infinite loop when the typescript client requests a non-existing table


KyleAMathews:
The server returns a 400 and this line then restarts the fetching https://github.com/electric-sql/electric/blob/371dc560083f131d10477b7c101cc30a1adcaa08/packages/typescript-client/src/client.ts#L235

It seems 404 is the proper response code for the server and would fix this issue.

thruflo:
Are there any other cases where a `400` could be returned. Do we also want to fix infinite loop for `400` generally?

KyleAMathews:
I'm not sure why we special-case the 400 there — 409 is the code we use to tell the client to refetch. In general 4xx codes are a user error so the client should just throw ann error and quit.

But yeah, to your point, we need to do two things — return 404 from the server for non-existing tables _and_ don't special case the 400 status code.

balegas:
We used a 400 initially on the server to not specialize the reasons. I guess things have evolved.

msfstef:
>We should change the client to separate resource creation (i.e. constructor) from connection initialisation (start) to make it easier to catch non-actionable errors.

I agree, I think we should open a new issue about this in particular.

I've opened https://github.com/electric-sql/electric/pull/1880 that should fix this immediate issue of infinite loops.

As an aside, I don't think we should be returning 404 errors since as far as we're concerned any request that defines a shape that is not valid (inexistent or invalid root table, inexistent or invalid column names, mismatched shape id and definition) are badly formed requests, and thus 400s.

If we want to split out errors (e.g. missing table from PG is 404, invalid table name is 400), it would be perhaps more explicit and informative but the handling from our side would be exactly the same, so it would just be a matter of improving API semantics.

PR #1880, opened 2024-10-22
fix: Handle 400 errors as unrecoverable

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1876

The issue was that we were sending `must-refetch` messages with some of our 400s, although they were unrecoverable and re-fetching would trigger an infinite loop.

I've made the server send 400 errors for the cases where the `shape_id` and `shape_definition` do not match, with an explanatory message, and I've made the client terminate when receiving such an error.

As a temporary measure, the error is not thrown into the ether but stored in the `ShapeStream` (accessible via `stream.error`).

We have had some discussions with @balegas about making errors more "handleable" by changing our approach to the `ShapeStream` instantiation. I think this warrants opening a new issue and corresponding PR to discuss more thoroughly.

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -22,6 +22,11 @@ defmodule Electric.Plug.ServeShapePlug do
   # Control messages
   @up_to_date [Jason.encode!(%{headers: %{control: "up-to-date"}})]
   @must_refetch Jason.encode!([%{headers: %{control: "must-refetch"}}])
+  @shape_definition_mismatch Jason.encode!(%{
+                               message:
+                                 "The specified shape definition and ID do not match. " <>
+                                   "Please ensure the shape definition is correct or omit the shape ID from the request to obtain a new one."
+                             })
 
   defmodule TimeUtils do
     @oct9th2024 DateTime.from_naive!(~N[2024-10-09 00:00:00], "Etc/UTC")
@@ -236,7 +241,7 @@ defmodule Electric.Plug.ServeShapePlug do
       # thus the shape ID does not match the shape definition
       # and we return a 400 bad request status code
       conn
-      |> send_resp(400, @must_refetch)
+      |> send_resp(400, @shape_definition_mismatch)
       |> halt()
     else
       # The shape ID does not exist or no longer exists
@@ -269,10 +274,8 @@ defmodule Electric.Plug.ServeShapePlug do
     if Shapes.has_shape?(config, shape_id) do
       # The shape with the provided ID exists but does not match the shape definition
       # otherwise we would have found it and it would have matched the previous function clause
-      IO.puts("400 - SHAPE ID NOT FOUND")
-
       conn
-      |> send_resp(400, @must_refetch)
+      |> send_resp(400, @shape_definition_mismatch)
       |> halt()
     else
       # The requested shape_id is not found, returns 409 along with a location redirect for clients to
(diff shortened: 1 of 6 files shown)


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -236,7 +241,7 @@ defmodule Electric.Plug.ServeShapePlug do
       # thus the shape ID does not match the shape definition
       # and we return a 400 bad request status code
       conn
-      |> send_resp(400, @must_refetch)
+      |> send_resp(400, @shape_definition_mismatch)
```
KyleAMathews, line 244: Why isn't this a 409? If the shape definition is valid perhaps they just have a really old shape_id.

KyleAMathews: Where people were hitting this problem (myself included) is when we'd request an invalid shape — e.g. the table didn't exist.

msfstef: This is the case where the `shape_id` is valid and exists (so there is an active shape that matches it), but the provided shape definition does not match that. That is a bad request and cannot be resolved with a `must-refetch`.

You need to:
- remove the `shape_id` so we start serving you the shape that matches the shape definition provided, *OR*
- provide the correct shape definition that matches this active `shape_id` that you are providing

KyleAMathews: Hmm no — when a shape is deleted on the server (for whatever reason), the next request for that shape would re-create the shape with a new shape_handle — then the next person to request the shape would have a mis-match between their shape_handle and the shape definition because their handle is pointing to a shape log that no longer exists. Which isn't a problem as a new shape now exists so they just need to refetch it.

This is the main use case for must-refetch.

400 is for a bad shape definition e.g. we can't parse the where clause or whatever.

KyleAMathews: (re: `shape_handle` https://github.com/electric-sql/electric/issues/1771)

msfstef: Yes my point is that this particular case is the case when the shape for the `shape_id` still exists, but you are hitting it with the wrong shape definition

KyleAMathews: Oh ok — yeah this shouldn't happen unless the client has a bug.

Did you check the "request a shape where the table doesn't exist"?

KyleAMathews: So nice seeing all those green checkboxes now!

msfstef: (see the integration test I added to the client)

when there is a mismatch, there are two cases:
1. the `shape_id` no longer exists, so we return a 409 with the new shape handle for the given definition
2. the `shape_id` exists and matches a shape definition that is not the provided one in the URL, so we return a 400

msfstef: > Did you check the "request a shape where the table doesn't exist"?

This case will return a 400 at the shape validation stage - I didn't add an integration test for this but as far as the client is concerned right now any 400 will terminate the stream so we should be fine

KyleAMathews (approved):
:shipit:

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
KyleAMathews: ok great!

kevin-dp (approved):

PR #1886, opened 2024-10-24
feat (sync service): multi tenancy

### LINKED ISSUE #1591: Add multi-database support

kevin-dp:
_Closes #1868, closes #1844. I kept that PR up to consult its diff until all the issues have been resolved here._

This PR fixes #1591.

### Architectural overview

multi-tenant-electric

TODO:

- [x] rebase on top of main
- [x] review commented out code and either remove it or make sure the place it had been moved to has been updated for multitenancy
- [x] add a `database_id` query param to the health endpoint to request the health of a given tenant
- [x] extract `validate_tenant_id`, `load_tenant` and `assign_tenant` which currently are duplicated in the create shape plug and delete shape plug.
- [x] Implement a two-tier ETS storage for column info and relation data. In this PR we replace two named tables with anonymous ones, necessitating a roundtrip to a single gen server just to fetch the table handle. This gen server becomes a bottleneck. We should either create another static ETS table to lookup individual ETS table handles in or store all column info and relation info in two static (named) tables, ensuring they don't trump each other via proper key design.
- [x] unit tests for multi tenancy
- [x] unit tests for add DB plug
- [x] unit tests for delete DB plug
- [x] e2e tests for multi tenancy
  - [x] debug flaky cache tests
- [x] add a diagram explaining the new architecture
- [x] update the open API spec (for add DB endpoint and remove DB endpoint)
- [x] persist tenant information and restore it on startup
- [ ] e2e test for persistence of tenants
- [x] Generate a random database id (i.e. tenant id) if you only provide a database_url env var


icehaunter (dismissed):

packages/sync-service/lib/electric/plug/add_database_plug.ex
```diff
@@ -0,0 +1,246 @@
...
+        |> halt()
+
+      {:error, error} ->
+        conn
+        |> send_resp(500, Jason.encode_to_iodata!(error))
+        |> halt()
+    end
+  end
+
+  def cors(conn, _opts) do
```
msfstef, line 132: if you can please reuse the cors logic at the top-level router, since it ensures that we return the right set of allowed methods per path

icehaunter (approved):

icehaunter:
Benchmark this

PR #1889, opened 2024-10-25
Simplify configuration of the electric app

magnetised:
When electric is included as a dependency then the configuration done in our `config/runtime.exs` is not used, and the application needs to set electric up itself.

Because we've made every configuration option required in the code (relying on runtime.exs) this is onerous.

But, because every option except the db connection actually has sensible defaults, we can simplify this and reduce the required configuration settings down to the db connection stuff.

The only downside to this is that now we have default values for the parameters in two locations, runtime.exs and with every `Application.get_env/3` call.

I took the opportunity to re-factor the configuration stuff and move it out of the application.ex module.

## TODO

- [ ] document `config :electric, :storage_dir, "/my/path"`

### DIFF
--- a/.changeset/shiny-ties-relate.md
+++ b/.changeset/shiny-ties-relate.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Simplify configuration by giving all options a default. Now the only required setting is the db connection parameters.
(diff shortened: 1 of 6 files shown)


magnetised:
NOTE: I'm pausing this until the multi-tenant work is complete as the two have dangerous amounts of overlap

PR #1892, opened 2024-10-25
docs: Writing your own client guide.

thruflo:
See https://deploy-preview-1892--electric-next.netlify.app/docs/guides/writing-your-own-client

The URL format and headers have been updated to match the latest in https://github.com/electric-sql/electric/pull/1900

### DIFF
--- a/website/docs/guides/writing-your-own-client.md
+++ b/website/docs/guides/writing-your-own-client.md
@@ -86,7 +86,7 @@ In live mode, if the server doesn't have any new data, it will hold open your re
 
 Set `live=true` to switch Electric into live mode. Make sure your request timeout is higher than the server timeout (which defaults to `20s`)
 
-If the previous response contains an `electric-cursor` header, then also set the `cursor` parameter to its value. (This is an extra cache-busting parameter used to normalise [request-coalescing](/docs/api/http#coalescing-live-requests) behaviour across different CDNs).
+If the previous response contains an `electric-cursor` header, then also set the `cursor` parameter to its value. (This is an extra cache-busting parameter used to normalise [request-collapsing](/docs/api/http#collapsing-live-requests) behaviour across different CDNs).
 
 For example:
 
--- a/website/docs/guides/writing-your-own-client.md
+++ b/website/docs/guides/writing-your-own-client.md
@@ -38,10 +38,10 @@ In the initial sync phase, you make a series of requests to get Shape data, incr
 
 #### Construct your shape URL
 
-Encode a [shape definition](/docs/guides/shapes#defining-shapes) into a `GET /v1/shape` URL. See the <a href="/openapi.html#/paths/~1v1~1shape~1%7Broot_table%7D/get">specification for the URL structure here</a>. For example, a Shape that contains all of the rows in the `items` table has the path:
+Encode a [shape definition](/docs/guides/shapes#defining-shapes) into a `GET /v1/shape` URL. See the <a href="/openapi.html#/paths/~1v1~1shape~1%7Broot_table%7D/get">specification for the URL structure here</a>. For example, a Shape that contains all of the rows in the `items` table would be requested with:
 
 ```http
-GET /v1/shape/items
+GET /v1/shape?table=items
 ```
 
 #### Make the initial `offset=-1` request
@@ -49,12 +49,12 @@ GET /v1/shape/items
 The first request to a shape should set the `offset` parameter to `-1`. This indicates to Electric that you want to consume all of the data from the beginning of the [Shape log](/docs/api/http#shape-log). For example, you might make a request to:
 
 ```http
-GET /v1/shape/items?offset=-1
+GET /v1/shape?table=items?offset=-1
 ```
 
 The body of the response will contain a JSON array of messages. The headers of the response will contain two pieces of important information:
 
-- `electric-shape-handle` an ephemeral identifier to an existing shape log
+- `electric-handle` an ephemeral identifier to an existing shape log
 - `electric-offset` the offset value for your next request
 
 If the last message in the response body contains an `up-to-date` control message:
@@ -70,7 +70,7 @@ Then the response will also contain an:
 Either of which indicate that you can [process the messages](#materialise-the-shape-log) and switch into [live mode](#live-mode). Otherwise, you should continue to accumulate messages by making additional requests to the same URL, with the new shape handle and offset. For example:
 
 ```http
-GET /v1/shape/items?shape_handle=38083685-1729874417404&offset=0_0
+GET /v1/shape?table=items&handle=38083685-1729874417404&offset=0_0
 ```
 
 In this way, you keep making GET requests with increasing offsets until you load all the data that the server is aware of, at which point you get the `up-to-date` message.
@@ -91,7 +91,7 @@ If the previous response contains an `electric-cursor` header, then also set the
 For example:
 
 ```http
-GET /v1/shape/items?live=true&shape_handle=38083685-1729874417404&offset=27344208_0&cursor=1674440
+GET /v1/shape?table=items&handle=38083685-1729874417404&offset=27344208_0&cursor=1674440&live=true
 ```
 
 #### Keep polling
@@ -120,18 +120,18 @@ Shape log messages are either control messages or logical `insert`, `update` or 
 ```ts
 switch (message.headers.operation) {
   case `insert`:
-    this.#data.set(message.key, message.value)
+    data.set(message.key, message.value)
 
     break
   case `update`:
-    this.#data.set(message.key, {
-      ...this.#data.get(message.key)!,
+    data.set(message.key, {
+      ...data.get(message.key)!,
       ...message.value,
     })
 
     break
   case `delete`:
-    this.#data.delete(message.key)
+    data.delete(message.key)
 
     break
 }
@@ -256,7 +256,8 @@ class Shape(object):
         # Fetch the response.
         response = requests.get(url)
 
-        # A real client should errors and backoff, reconnect, etc.
+        # This is a happy path example, so we just log error codes.
+        # A real client should handle errors, backoff, reconnect, etc.
         if response.status_code > 204:
             raise Exception("Error: {}".format(response.status_code))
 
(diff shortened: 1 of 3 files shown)


website/docs/guides/writing-your-own-client.md
```diff
@@ -1,7 +1,399 @@
...
+How you choose to provide this kind of API is very language dependent. You could support registering callbacks (like `shape.subscribe`) and then call these whenever you've finished materialising your shape, or you could some kind of broadcast mechanism.
+
+## Examples
+
+Let's walk through the process of implementing a client in a real programming language.
+
+### Brainfuck
+
+```shell
+++++++++[>++++++++++>++++++++++++++>+++++++++++++++>++++>+++++++>+++++<<<<<<-]>-.>--.--.>+.>.<<--.+++++.----.--.+++++.-------.>>.>+++.>+.
```
samwillis, line 186: Love this

website/docs/guides/writing-your-own-client.md
```diff
@@ -1,7 +1,399 @@
...
+```python
+import requests
+from urllib.parse import urlencode
+
+class Shape(object):
+    """Syncs a shape log and materialises it into a `data` dict."""
+
+    def __init__(self, base_url='http://localhost:3000', offset=-1, handle=None, table=None, where=None):
+        if table is None:
+            raise "Must provide a table"
```
samwillis, line 219:
```suggestion
            raise ValueError("Must provide a table")
```

website/docs/guides/writing-your-own-client.md
```diff
@@ -1,7 +1,399 @@
...
+                self.live = True
+                self.process_messages()
+
+        # Set the shape handle, offset and optionally cursor for
+        # the next request from the response headers.
+        self.handle = response.headers['electric-handle']
+        self.offset = response.headers['electric-offset']
+
+        if 'electric-cursor' in response.headers:
+            self.cursor = r.headers['electric-cursor']
```
samwillis, line 280:
```suggestion
            self.cursor = response.headers['electric-cursor']
```

website/docs/guides/writing-your-own-client.md
```diff
@@ -1,7 +1,399 @@
...
+```shell
+# Optionally in a virtualenv:
+# virtualenv .venv
+# source .venv/bin/activate
+python -m pip install requests
+```
+
+Now let's write our `Shape` client, saving the following in `client.py`:
+
+```python
```
samwillis, line 210: This is great!

We could also offer an implementation (as a tab on the code block?) using asyncio (untested, but think its mostly correct):

```sh
python -m pip install aiohttp
```

```python
import aiohttp
import asyncio
from urllib.parse import urlencode

class Shape:
    """Async syncs a shape log and materializes it into a `data` dict."""

    def __init__(self, base_url='http://localhost:3000', offset=-1, handle=None, table=None, where=None):
        if table is None:
            raise ValueError("Must provide a table")

        # Request state used to build the URL.
        self.base_url = base_url
        self.cursor = None
        self.handle = handle
        self.live = False
        self.offset = offset
        self.table = table
        self.where = where

        # Materialiased data.
        self.data = {}

        # Accumulated messages (waiting for an `up-to-date` to apply).
        self.messages = []

        # Registered callbacks to notify when the data changes.
        self.subscribers = []

        # Task management for async syncing.
        self._sync_task = None  # Background task for the sync loop.
        self._stop_event = asyncio.Event()  # Event used to stop the sync loop.

    def subscribe(self, callback):
        """Register a callback function to be called whenever data changes."""
        
        self.subscribers.append(callback)

    async def start_sync(self):
        """Begin syncing in the background, without blocking."""
        # Only start if no existing sync task is running.
        if self._sync_task is None or self._sync_task.done():
            # Clear the stop event and create a new sync task.
            self._stop_event.clear()
            self._sync_task = asyncio.create_task(self._run_sync())

    async def stop_sync(self):
        """Stop the background sync task gracefully."""
        if self._sync_task and not self._sync_task.done():
            # Set the stop event to signal the sync task to exit.
            self._stop_event.set()
            await self._sync_task  # Wait for the task to complete.

    async def _run_sync(self):
        """Internal method to run the sync loop until a stop signal is received."""
        while not self._stop_event.is_set():  # Continue until stop is requested.
            await self.request()  # Perform a sync request.
            await asyncio.sleep(1)  # Add delay to avoid rapid polling.

    async def request(self):
        """Make an asynchronous request to `GET /v1/shape` and process the response."""
        
        # Build the request URL with current parameters.
        url = self.build_url()

        async with aiohttp.ClientSession() as session:

            # Fetch the response.
            async with session.get(url) as response:

                # This is a happy path example, so we just log error codes.
                # A real client should handle errors, backoff, reconnect, etc.
                if response.status > 204:
                    raise Exception(f"Error: {response.status}")

                # If the response is 200 then we may have new data to process.
                if response.status == 200:
                    self.messages.append(await response.json())  # Queue the new data.

                    # If we're up-to-date, switch into live mode and process
                    # the accumulated messages.
                    if 'electric-up-to-date' in response.headers:
                        self.live = True
                        await self.process_messages()

                # Set the shape handle, offset and optionally cursor for
                # the next request from the response headers.
                self.handle = response.headers.get('electric-handle')
                self.offset = response.headers.get('electric-offset')

                if 'electric-cursor' in response.headers:
                    self.cursor = response.headers['electric-cursor']

    # Snip... the rest is the same
```

website/docs/guides/writing-your-own-client.md
```diff
@@ -1,7 +1,399 @@
...
+            params['where'] = self.where
+
+        return "{}/v1/shape?{}".format(self.base_url, urlencode(params))
+```
+
+Now let's create a test file to test running the client. Save the following in `client.test.py`:
+
+```python
+import multiprocessing
+import unittest
```
samwillis, line 367: A asyncio version (untested):

```py
import asyncio
import unittest

from client import Shape

class TestClient(unittest.IsolatedAsyncioTestCase):
    async def test_shape_sync(self):
        shape = Shape(table='items')
        data_received = []
        event = asyncio.Event()

        def callback(data):
            data_received.append(data)
            event.set()

        shape.subscribe(callback)
        await shape.start_sync()
        await asyncio.wait_for(event.wait(), timeout=5)
        self.assertEqual(type(data_received[0]), dict)
        await shape.stop_sync()

if __name__ == '__main__':
    unittest.main()
```

samwillis (approved):
This is really great!! I think it provides an even better overview of how sync works than the HTTP API docs page.

Python client is good, and provides such a simple overview of how to use the api. We could offer an asyncio version as that seems to be where Python has moved with all this stuff, and removes the need for the blocking `sync()` call and multiple processing. (If only gevent had been picked over asyncio...)

website/docs/guides/writing-your-own-client.md
thruflo: You and your fancy modern code ;)

I do think it's best to keep async/thread/process stuff out of the client. It just makes it much less readable and for me this is more intended to be read than run.

Perhaps we can use this as the basis for an actual Python client?

website/docs/guides/writing-your-own-client.md
thruflo: Will this work with the non asyncio client? I've tried it locally and for me it just runs and hangs forever, so I suspect not?

website/docs/guides/writing-your-own-client.md
samwillis: Yep, good call!

We probably ~~want~~ need to have both versions of the client.

website/docs/guides/writing-your-own-client.md
samwillis: Yep, this relies on the asyinio loop, I believe `unittest.IsolatedAsyncioTestCase` will start it.

Let's leave it out and use this in the actual client.

website/docs/guides/writing-your-own-client.md
```diff
@@ -1,7 +1,399 @@
...
+In live mode, if the server doesn't have any new data, it will hold open your request until either a timeout or new data arrives. This allows you to implement long polling, where you keep the request open, and reconnect immediately when new data arrives.
+
+> [!Tip] Why not websockets?!
+> Consuming data over HTTP allows us to [leverage CDNs](/docs/api/http#caching), simplifies observability and allows you to implement auth (and other capabilities) [using HTTP proxies](/docs/guides/auth#recommended-pattern).
+
+#### Add `live` and `cursor` parameters
+
+Set `live=true` to switch Electric into live mode. Make sure your request timeout is higher than the server timeout (which defaults to `20s`)
+
+If the previous response contains an `electric-cursor` header, then also set the `cursor` parameter to its value. (This is an extra cache-busting parameter used to normalise [request-coalescing](/docs/api/http#coalescing-live-requests) behaviour across different CDNs).
```
KyleAMathews, line 89: I've only seen the term "request collapsing" btw

website/docs/guides/writing-your-own-client.md
```diff
@@ -1,7 +1,399 @@
...
+}
+```
+
+How you choose to provide this kind of API is very language dependent. You could support registering callbacks (like `shape.subscribe`) and then call these whenever you've finished materialising your shape, or you could use some kind of broadcast mechanism.
+
+## Examples
+
+Let's walk through the process of implementing a client in a real programming language.
+
+### Brainfuck
```
KyleAMathews, line 183: hmm this is a bit esoteric / off-beat of a joke? I looked it up and it's "hello world"...? I'd guess this will confuse more people than it'll amuse so I'd take it out.

KyleAMathews (changes requested):
Great stuff! This will be super helpful for all the engineers wanting to know how what's going on under the hood plus of course those wanting to build a new client in some other language.

website/docs/guides/writing-your-own-client.md
thruflo: Ok, I'll standardise on that.

website/docs/guides/writing-your-own-client.md
thruflo: I think the upsides of the humour outweigh the downside of potential confusion. This is at the bottom of the writing your own client guide. Not a location likely to be reached by the faint of heart.

KyleAMathews (approved):
:shipit:

Issue #1854, opened 2024-10-15
Add way to cleanup replication slots


KyleAMathews:
When shutting down Electric, sometimes we'll want to cleanup replication slots in Postgres. E.g. a PR environment where Electric is spun up for some tests and then shut down again.

Perhaps this is an environment variable? `CLEANUP_REPLICATION_SLOTS_ON_SHUTDOWN`

alco:
If this is a separate operational mode from the regular one, it's best to use an env variable that will instruct Electric to create a temporary slot (as opposed to the regular persistent one). Then Postgres will automatically remove the slot when the replication connection closes.

PR #1896, opened 2024-10-29
Allow configuration of temporary replication slots

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1854

For test environments it is convenient to have the option to use temporary replication slots such that no subsequent cleanup work is required.


robacourt (approved):
Nice!

KyleAMathews (commented):
💥

KyleAMathews:
We should also document this somewhere cc @thruflo

Issue #1888, opened 2024-10-24
Shape should expose an array not a map


KyleAMathews:
The Map we use inside the Shape is just for internal bookkeeping and isn't ever useful to end users.

@samwillis suggests we expose `shape.rows` that uses a getter to get the values off the map. `useShape` should use this as well to return an array.

We can keep `shape.value` but document `shape.rows` as the main way to use this.

PR #1897, opened 2024-10-29
feat: `rows` and `rowsSync` API on `Shape`

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1888

I've added both a `rows` and `rowsSync` API to match what we already have for `value` and `valueSync`.

The react hooks already expose a `data` field that _is_ the rows (it was doing the conversion there), so I don't think we need to change that API.

### DIFF
--- a/.changeset/friendly-toes-check.md
+++ b/.changeset/friendly-toes-check.md
@@ -1,0 +1,6 @@
+---
+"@electric-sql/client": patch
+"@electric-sql/react": patch
+---
+
+Implement `rows` and `rowsSync` getters on `Shape` interface for easier data access.
--- a/packages/react-hooks/src/react-hooks.tsx
+++ b/packages/react-hooks/src/react-hooks.tsx
@@ -96,7 +96,7 @@ function parseShapeData<T extends Row<unknown>>(
   shape: Shape<T>
 ): UseShapeResult<T> {
   return {
-    data: [...shape.valueSync.values()],
+    data: shape.rowsSync,
     isLoading: shape.isLoading(),
     lastSyncedAt: shape.lastSyncedAt(),
     isError: shape.error !== false,
--- a/packages/typescript-client/src/shape.ts
+++ b/packages/typescript-client/src/shape.ts
@@ -69,6 +69,14 @@ export class Shape<T extends Row<unknown> = Row> {
     return this.#stream.isUpToDate
   }
 
+  get rows(): Promise<T[]> {
+    return this.value.then((v) => Array.from(v.values()))
+  }
+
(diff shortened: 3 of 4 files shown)


samwillis (approved):
Looks good.

One comment would be that we may want to change more of the tests to use `.rows` and `.rowsSync` as they are the user facing APIs.

msfstef:
> we may want to change more of the tests to use .rows and .rowsSync

I agree but that was my point about the `subscribe` API which currently returns the Map version of the data. If we want to push users to the `rows` version then we should also change the `subscribe` API otherwise it might be a bit confusing.

I'm happy to open a new PR with the full breaking change refactor.

samwillis:
Ah yes. I would argue we should make this braking change now. But maybe this so that its more extendable in future:

```ts
export type ShapeData<T extends Row<unknown> = Row> = Map<string, T>
export type ShapeChangedCallback<T extends Row<unknown> = Row> = (data: { // <-- new wrapper object
  value: ShapeData<T> // <-- can still get the "value"
  rows: T[]  //  <-- now have the rows
}) => void
```

so that it can used with:
```ts
shape.subscribe(({ rows }) => {
  // Do stuff...
});
```

We can then add additional metadata as further props to the callback in future.

@KyleAMathews what do you think?

msfstef:
> Because if so “sync” is a bit overloaded/ambiguous in our context.

I do think that the combination of `sync` being the suffix, that it is a getter, and of the common JS convention for synchronous API equivalents is enough to differentiate it from our general data sync concept.

That being said, the `rowsSync` and `valueSync` are not synchronous wrappers of the promise version - they access the current materialized shape regardless of its "up to date" status. Therefore they are "unsafe" in the sense that you might see transactionally incomplete data. In that sense I would agree that it might make sense to use a different term, like `rowsUnsafe` or something

thruflo:
Yeah, maybe `currentRows` / `currentValue`?

msfstef:
Ok so based on this feedback I'm applying @samwillis suggestion (that is still a breaking change but keeps both the `Map` and rows version), and taking this opportunity for a breaking change to rename `valueSync` (and the new rows one) to `currentValue` and `currentRows`.

KyleAMathews:
I like your api suggestion @samwillis 👍 

And yeah, we're making a number of breaking changes now — so toss this on the pile. This is a great time to fine-tune the API to a huge extent as we have enough breadth and depth of experience to make better judgements on stuff like this.

msfstef:
should be ready to merge now - updated the changeset to reflect the breaking change

KyleAMathews (approved):
neat!

Issue #1690, opened 2024-09-13
Delete shapes from the react integrations cache if it's aborted 


KyleAMathews:
When a new shape is created with an abortcontroller, listen for aborts so we can remove the shape from the cache. Right now if you abort a shape, you can't restart it as the cache will just return the aborted shape.

PR #1898, opened 2024-10-29
fix: Clear react caches when stream is errored

msfstef:
Fixes https://github.com/electric-sql/electric/issues/1690

### DIFF
--- a/packages/react-hooks/src/react-hooks.tsx
+++ b/packages/react-hooks/src/react-hooks.tsx
@@ -35,13 +35,13 @@ export function getShapeStream<T extends Row<unknown>>(
   // If the stream is already cached, return it if valid
   if (streamCache.has(shapeHash)) {
     const stream = streamCache.get(shapeHash)! as ShapeStream<T>
-    // if stream is cached but errored/aborted, remove it and related shapes
-    if (stream.error !== undefined || stream.options.signal?.aborted) {
-      streamCache.delete(shapeHash)
-      shapeCache.delete(stream)
-    } else {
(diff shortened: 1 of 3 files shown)


packages/react-hooks/src/react-hooks.tsx
```diff
@@ -32,35 +32,49 @@ export function getShapeStream<T extends Row<unknown>>(
...
+    if (stream.error !== undefined || stream.options.signal?.aborted) {
+      streamCache.delete(shapeHash)
+      shapeCache.delete(stream)
+    } else {
+      return stream
+    }
+  }
 
-    streamCache.set(shapeHash, newShapeStream)
+  const newShapeStream = new ShapeStream<T>(options)
```
KyleAMathews, line 47: perhaps move this logic inside the if error clause?

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -32,35 +32,49 @@ export function getShapeStream<T extends Row<unknown>>(
...
+    ) {
+      streamCache.delete(sortedOptionsHash(shapeStream.options))
+      shapeCache.delete(shapeStream)
+    } else {
+      return shapeCache.get(shapeStream)! as Shape<T>
+    }
+  }
 
-    shapeCache.set(shapeStream, newShape)
+  const newShape = new Shape<T>(shapeStream)
```
KyleAMathews, line 72: same

KyleAMathews (approved):
Thanks!

PR #1900, opened 2024-10-29
feat!: move table to query param & rename existing headers and query params

### LINKED ISSUE #1771: Shape API tweaks

thruflo:
Included renames:
- `electric-chunk-last-offset` -> `electric-offset`
- `electric-next-cursor` -> `electric-cursor`
- `electric-shape-id` -> `electric-handle`
- `?shape_id` -> `?handle`

Added query parameter: `?table` instead of `/:table` in the path

Closes #1771, #1796, #1798


integration-tests/tests/crash-recovery.lux
```diff
@@ -41,8 +41,8 @@
 [shell client]
   # strip ANSI codes from response for easier matching
   !curl -v -X GET http://localhost:3000/v1/shape/items?offset=-1
-  ?electric-shape-id: ([\d-]+)
-  [local shape_id=$1]
+  ?electric-shape-handle: ([\d-]+)
```
msfstef, line 44:
```suggestion
  ?electric-handle: ([\d-]+)
```

thruflo: Yeah, what is that `.lux` file? I left it alone in case it was some kind of artefact.

msfstef: We have a set of integration tests using lux - this one tests crash recovery

thruflo:
Btw, I don't understand why the typecheck tests are failing. `pnpm typecheck` runs fine for me locally in both the `react-hooks` and `typescript-client` packages:

```
➜  react-hooks git:(thruflo/header-naming) ✗ pnpm run typecheck

> @electric-sql/react@0.4.7 typecheck /Users/thruflo/Development/electric-sql/electric/packages/react-hooks
> tsc -p tsconfig.json

➜  react-hooks git:(thruflo/header-naming) ✗ cd ..
➜  packages git:(thruflo/header-naming) ✗ cd typescript-client
➜  typescript-client git:(thruflo/header-naming) ✗ pnpm run typecheck

> @electric-sql/client@0.6.5 typecheck /Users/thruflo/Development/electric-sql/electric/packages/typescript-client
> tsc -p tsconfig.json
```

msfstef:
@thruflo the `react-hooks` package depends on the typescript client, so you need to first `pnpm buiild` the typescript client and then run `pnpm typecheck` on the react hooks package

thruflo:
Yup, I've `pnpm -r build`ed from the repo root locally.

I'm not really the right person to revise the way we run our CI tests :)

msfstef:
@thruflo I was also seeing the same issue - typecheck passed locally but failed on CI, and the lines referenced on CI for the errors did not match the actual test file, so something odd happening there...

There's a conflict now so tests aren't running though.

thruflo:
@msfstef fixed the conflict.

msfstef:
@thruflo types fixed 💪 🚀

icehaunter (approved):

PR #1905, opened 2024-10-29
Ack the WAL Keep-Alives

robacourt:


msfstef:
Linking to a [relevant article](https://www.artie.com/blogs/preventing-wal-growth-on-postgres-db-running-on-aws-rds) - in AWS RDS there are heartbeats written to the DB every 5 minutes that make the WAL grow very quickly.

I _think_ it makes sense to flush on keep alive messages in order to avoid this happening and this shouldn't mess with our logic in any way. A second set of eyes more familiar with PG would be good (@icehaunter?), but I this might resolve the issue.

We could be smarter about when we ack the WAL in the keep alive if there are performance implications on doing it every time, but I suspect this is already better than letting it grow indefinitely.

msfstef (approved):

balegas:
Our past experience: https://github.com/electric-sql/electric/issues/1083

> I think it makes sense to flush on keep alive messages in order to avoid this happening and this shouldn't mess with our logic in any way.

 Can't see why not. Let's run this on RDS tomorrow and see what happens.

icehaunter (approved):

Issue #1774, opened 2024-10-01
Clean up publication filters when shapes are removed


msfstef:
Currently the replication publication gets altered when a shape is created (adds a table and potentially a row filter) but no cleanup occurs - so the publication can potentially grow to include everything, even between restarts and deploys, and slow down the system over time.

My suggestion to address this is to change the `Electric.Postgres.Configuration` to alter the publication based on all active shapes rather than based on each individual one, in that case every call will update the entire publication and its filters as necessary and resuming/cleaning can be a matter of calling this every time a shape is deleted and once upon starting (with recovered shapes or no shapes).

KyleAMathews:
> can be a matter of calling this every time a shape is deleted and once upon starting (with recovered shapes or no shapes).

How much work is it to generate this call? What if 1000s of new shapes are being created every second?

msfstef:
The cleanups in general can be scheduled, but it is important to note that not cleaning them up also means that we're processing data through the replication stream that we don't need, which also takes up resources - delaying a few requests in order to free up resources is probably still a better option, but we'd need to benchmark that. Cleaning it up at _any_ point is probably the first step.

We could also avoid calling it whenever a shape is deleted and only call it whenever a shape is created like we do now, and the creation call will clean up what is not used (and perhaps trigger periodic cleanups as well rather than on every shape deletion).

For reference we're already doing at least 2 roundtrips with Postgres whenever a shape is created in order to adjust the publication filter and that occurs on the critical path, so I doubt the cleaning up will be a bottleneck.

balegas:
It's a bit of an orthogonal point to this issue, but the need to re-think how we do filtering.

We cannot set replication filter per shape. We need to be more selective on the WHERE clauses. It will never work to pass filters like `WHERE id = x`, `WHERE id = y`  (filtering by id) as the number of shapes grow. 

If we think at a macro-level, one electric instance needs to be able to handle a certain amount of rows, no matter how many filters you put into the replication slot. If we get to the point that the max tp of electric in not high enough, we need to do shape matching more efficiently.

- I'd like to evaluate what are the performance gains just by filtering columns that we don't need, which would be a very simple filter to maintain.
- There might be other easy filters we might be able to do

I'll create an issue for that

KyleAMathews:
Columns + tables seems sustainable.

balegas:
[...]

PR #1908, opened 2024-10-31
fix: address deadlock with concurrent snapshot creation

icehaunter:
This commit addresses 2 issues:

First, the deadlock occurred when trying to add where clauses to existing tables - we executed `ALTER TABLE` that took out an exclusive access lock over one table definition, then executed `SET TABLES` on a publication that requested shared access lock over all tables mentioned in the publication which includes tables that are already exclusively locked by another concurrent shape creation. This caused a deadlock. This was addressed by switching the order of alter table and alter publication.

Second, there was a race condition when 2 new shapes tried updating the publication where clause and because it's done in a read-then-update way, not as a single statement, the read might have missed the concurrent update and the resulting state was wrong. This was addressed by a lock.

### DIFF
--- a/.changeset/tall-chairs-punch.md
+++ b/.changeset/tall-chairs-punch.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix a possible deadlock issue when creating or updating multiple where-claused shapes that occured while updating the Postgres publication (only on PG 15+). Fix a possible race condition between reading the existing publication and writing the updated version.
--- a/packages/sync-service/lib/electric/postgres/configuration.ex
+++ b/packages/sync-service/lib/electric/postgres/configuration.ex
@@ -7,7 +7,6 @@ defmodule Electric.Postgres.Configuration do
   alias Electric.Shapes.Shape
 
   @type filter() :: String.t() | nil
-  @type maybe_filter() :: filter() | :relation_not_found
   @type filters() :: %{Electric.relation() => filter()}
 
   @pg_15 150_000
@@ -42,8 +41,6 @@ defmodule Electric.Postgres.Configuration do
   defp configure_tables_for_replication_internal!(pool, relations, pg_version, publication_name)
        when pg_version < @pg_15 do
     Postgrex.transaction(pool, fn conn ->
-      set_replica_identity!(conn, relations)
-
       for {relation, _} <- relations,
           table = Utils.relation_to_sql(relation),
           publication = Utils.quote_name(publication_name) do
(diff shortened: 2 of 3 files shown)


packages/sync-service/test/electric/postgres/configuration_test.exs
```diff
@@ -218,6 +229,67 @@ defmodule Electric.Postgres.ConfigurationTest do
...
+        Task.async(fn ->
+          Postgrex.transaction(conn, fn conn ->
+            Configuration.configure_tables_for_replication!(
+              conn,
+              [{{"public", "other_other_table"}, "(value ILIKE '2%')"}],
+              get_pg_version,
+              publication
+            )
+
+            :ok
```
msfstef, line 260: do we need some way to ensure the transaction doesn't end so that they overlap here? are we relying on the tasks being started concurrently for the transactions to consistently cause a deadlock?

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -69,37 +66,37 @@ defmodule Electric.Postgres.Configuration do
...
     end)
   end
 
   defp configure_tables_for_replication_internal!(pool, relations, _pg_version, publication_name) do
     Postgrex.transaction(pool, fn conn ->
+      # We're using advisory locks to prevent race conditions when multiple
+      # processes try to read-then-update the publication configuration. We're not using `SELECT FOR UPDATE`
+      # because it doesn't read the value that was updated by other transaction holding the lock. This lock
+      # is thus acquired before reading the existing configuration, so the first read sees the latest value.
+      Postgrex.query!(conn, "SELECT pg_advisory_xact_lock($1)", [:erlang.phash2(publication_name)])
```
msfstef, line 80: nice - this way we are essentially serialising the updates to the publication via the PG advisory lock right? If so for the future I do think that this would be better served by a process that handles this perhaps (to address https://github.com/electric-sql/electric/issues/1774 as well)

.changeset/tall-chairs-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix a possible deadlock issue when creating or updating multiple where-claused shapes that occured while updating the Postgres publication (only on PG 15+). Fix a possible race condition between reading the existing publication and writing the updated version.
```
msfstef, line 5: super nit: split the two fixes to two lines for readability?

msfstef (approved):
Tricky problem and nice fix!

packages/sync-service/test/electric/postgres/configuration_test.exs
icehaunter: tasks started concurrently is good enough, hits every time without the fix

packages/sync-service/lib/electric/postgres/configuration.ex
icehaunter: Yup, it does exacly that

.changeset/tall-chairs-punch.md
icehaunter: It'll look weirder in the full changelog :P

PR #1836, opened 2024-10-11
docs: deployment guide, with supporting config and integrations.

thruflo:
Fleshing out deployment guide. To support it, added a sync service config page in the api docs section and specific integrations pages for various deployment platforms.

Quite a lot to review here. Check out the:

- [Sync service config page at /docs/api/config](https://deploy-preview-1836--electric-next.netlify.app/docs/api/config)
- [Updated Typescript client page at /docs/api/clients/typescript](https://deploy-preview-1836--electric-next.netlify.app/docs/api/clients/typescript)
- [Deployment page at /docs/guides/deployment](https://deploy-preview-1836--electric-next.netlify.app/docs/guides/deployment)
- Individual integration pages in the "Integrations" section of the docs nav, e.g.: [AWS](https://deploy-preview-1836--electric-next.netlify.app/docs/integrations/aws), [React](https://deploy-preview-1836--electric-next.netlify.app/docs/integrations/react) and [Supabase](https://deploy-preview-1836--electric-next.netlify.app/docs/integrations/supabase)

Note that:

- Netlify has a one-click-deploy form that uses https://github.com/electric-sql/standalone-basic-example
- [Render](https://deploy-preview-1836--electric-next.netlify.app/docs/integrations/supabase) has a [Blueprint repo](https://github.com/electric-sql/render-blueprint) that also uses the same standalone-basic-example

As @KyleAMathews commented on Discord, these might be better to include in the monorepo and update using a GitHub action. Perhaps that can be a subsequent Issues / PR.

The integrations also include a [Phoenix page](https://deploy-preview-1836--electric-next.netlify.app/docs/integrations/phoenix) which is based on merging #1833 and publishing an updated release for https://hexdocs.pm/electric_phoenix/0.1.0-dev-2/Electric.Phoenix.html

Note that the deployment docs all assume a single `DATABASE_URL`, i.e.: this is pre multi-tenant.

PR #1917, opened 2024-11-04
Wrap more steps in snapshot with telemetry spans

magnetised:
Plus:

- set the db connection pool timeout to `:infinity` for snapshot creation (large datasets can timeout when writing the snapshot otherwise)
- set `LOG_OTP_REPORTS=true` to turn on SASL reporting at runtime
- add some more fine-grained log messages
- include some tests that validate our handling of nulls when evaluating array comparisons in where clauses

### DIFF
--- a/.changeset/chatty-adults-heal.md
+++ b/.changeset/chatty-adults-heal.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Add tracing of snapshot creation and more logging of postgres connection status. Prevent connection timeouts when writing snapshot data. Add `LOG_OTP_REPORTS` environment variable to enable OTP SASL reporting at runtime.
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -241,8 +241,13 @@ defmodule Electric.Postgres.ReplicationClient do
         # receives more demand.
         # The timeout for any call here is important. Different storage
         # backends will require different timeouts and the timeout will need to
-        # accomodate varying number of shape consumers. The default of 5_000 ms
-        # should work for our file-based storage backends, for now.
+        # accomodate varying number of shape consumers.
+        #
+        # The current solution is to set timeout: :infinity for the call that
+        # sends the txn message to the consumers and waits for them all to
+        # write to storage, but crash individual consumers if the write takes
+        # too long. So it doesn't matter how many consumers we have but an
+        # individual storage write can timeout the entire batch.
         OpenTelemetry.with_span(
           "pg_txn.replication_client.transaction_received",
           [num_changes: length(txn.changes), num_relations: MapSet.size(txn.affected_relations)],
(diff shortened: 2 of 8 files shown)


.changeset/chatty-adults-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Add tracing of snapshot creation and more logging of postgres connection status. Prevent connection timeouts when writing snapshot data. Add `LOG_OTP_REPORTS` environment variable to enable OTP SASL reporting at runtime.
```
balegas, line 5: @thruflo one to add to your PR

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -241,8 +241,13 @@ defmodule Electric.Postgres.ReplicationClient do
         # receives more demand.
         # The timeout for any call here is important. Different storage
         # backends will require different timeouts and the timeout will need to
-        # accomodate varying number of shape consumers. The default of 5_000 ms
-        # should work for our file-based storage backends, for now.
+        # accomodate varying number of shape consumers.
+        #
+        # The current solution is to set timeout: :infinity for the call that
```
balegas, line 246: It makes me a bit nervous to set an infinite timeout as a consumer could hang forever. I rather have a generous timeout than hiding any potential pathology for the cases that shouldn't take so long.

I understand that this could still prevent us from doing veeeeery large shapes. Happy for you to leave it like that if you have strong opinion,

balegas (approved):
I'm happy to approve, but not judging the Elixir code.

packages/sync-service/lib/electric/postgres/replication_client.ex
magnetised: No, a consumer can't hang forever

> but crash individual consumers if the write takes too long

the storage backend has a timeout so a write cannot just hang indefinitely

.changeset/chatty-adults-heal.md
thruflo: https://github.com/electric-sql/electric/pull/1836/commits/311fdfb78b554cbfc5a9289559f354cc388618c0

Really would like to get the doc PRs merged.

robacourt (approved):
Nice. Thanks for the helpful comments too

packages/sync-service/lib/electric/postgres/replication_client/connection_setup.ex
```diff
@@ -41,12 +41,15 @@ defmodule Electric.Postgres.ReplicationClient.ConnectionSetup do
   # streaming mode.
   @spec start_streaming(state) :: callback_return
   def start_streaming(%{step: :ready_to_stream} = state) do
+    Logger.debug("ReplicationClient step: start_streaming")
```
KyleAMathews, line 44: Is there some way we can also send these logs to open telemetry as well? I don't hate logs by any means but debugging is a lot easier IME in a observability tool.

Not a blocker for this PR.

balegas (approved):

packages/sync-service/lib/electric/postgres/replication_client/connection_setup.ex
balegas: Found this discussion about the state of it  https://elixirforum.com/t/what-is-recommended-way-to-send-logs-to-otel-collector/63779

Issue #1927, opened 2024-11-05
Consistently Namespace our Environment Variables with ELECTRIC_


KyleAMathews:
Most system software does this e.g. Kubernetes, Redis, MySQL, Postgres, etc. as if you run multiple bits of software in the same environment than it's easy to get namespace clashes.

thruflo:
Yup, I think `ELECTRIC_` reads much better.

PR #1928, opened 2024-11-05
chore: rename env variables so consitently prefixed with ELECTRIC_

KyleAMathews:
Fix https://github.com/electric-sql/electric/issues/1927

### DIFF
--- a/.changeset/big-shoes-care.md
+++ b/.changeset/big-shoes-care.md
@@ -2,4 +2,4 @@
 "@core/sync-service": patch
 ---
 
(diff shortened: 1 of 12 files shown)


thruflo:
I think it would be useful to do this on top of the new deployment guide, given that has a new page documenting the env var config.

KyleAMathews:
Is that going in right now too? This is another breaking change so I want it to go out with all the others (ideally today)

KyleAMathews:
Happy to pr the updates separately to the doc page

KyleAMathews:
ok merging and then I'll open a PR to the docs PR with the same updates

PR #1915, opened 2024-11-01
examples: add public deployment of LinearLite

KyleAMathews:
Depends on https://github.com/electric-sql/electric/pull/1886

Uses Electric Cloud & sst.dev to deploy Linearlite.

TODO:
- [ ] Add github action to deploy the site

PR #1929, opened 2024-11-05
examples: add public deployment of LinearLite 

balegas:
Took changes from https://github.com/electric-sql/electric/pull/1915 on top of main


KyleAMathews (approved):
:shipit

.github/workflows/deploy_linearlite.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: actions/cache@v4
+        with:
+          path: .sst
+          key: sst-cache-main-${{ runner.os }}
+          restore-keys: |
+            sst-cache-main-${{ runner.os }}
+
+      - name: Initialize SST environment
+        run: |
+          sudo apt install -y tini
```
KyleAMathews, line 41: the tini stuff was an experiment I did to see if it'd help sst gracefully shut down (i.e. save it's state correctly when github actions kills it). It didn't seem to work actually so what is needed is concurrency controls so only one action can run at a time & running ones don't get killed.

balegas: yeah. I am trying to figure this out. I thought it could be something I've done locally.

KyleAMathews: haha yeah — i've spent a lot of time sorting through this error — you just gotta click around AWS' console looking for the resource(s) mentioned and delete them manually. 

Then at the top of file you want:

```
concurrency:
  group: prod-deploy-group
```

.github/workflows/deploy_examples.yml
```diff
@@ -0,0 +1,53 @@
+name: Deploy Examples Production
+
+on:
+    push:
+      branches: ['main']
+    pull_request:
```
balegas, line 6: Create a pr deployment when one of the examples is modified. Because we need to teardown the PR stack and it breaks with Neon, maybe we comment this out for now?

examples/linearlite/sst.config.ts
```diff
@@ -0,0 +1,117 @@
...
+      branchId: project.defaultBranchId,
+    }
+
+    const db = new neon.Database(`linearlite`, {
+      ...base,
+      ownerName: `neondb_owner`,
+    })
+
+    const databaseUri = getNeonDbUri(project, db)
+    try {
```
balegas, line 34: Do we need to be defensive? I actually haven't tested the exception path because everything is working now.

balegas (commented):
@KyleAMathews I think we're finally done! Ready for review

.github/workflows/deploy_examples.yml
KyleAMathews: yeah good call — too much overhead right now — we can run a site locally for a PR if we want to test it.

examples/linearlite/sst.config.ts
KyleAMathews: yeah I'd say so until we have our own Provider that handles errors correctly

.github/workflows/teardown_examples_pr_stack.yml
```diff
@@ -0,0 +1,62 @@
...
+          key: sst-cache-${{ github.event.number }}-${{ runner.os }}
+          restore-keys: |
+            sst-cache-${{ runner.os }}
+
+      - name: Remove Linearlite
+        working-directory: examples/linearlite
+        run: |
+            export PR_NUMBER=${{ github.event.number }}
+            echo "Removing stage pr-$PR_NUMBER"
+            pnpm sst remove --stage "pr-$PR_NUMBER"
```
KyleAMathews, line 55: use $DEPLOY_ENV here to be consistent with other workflow?

examples/nextjs-example/sst.config.ts
```diff
@@ -0,0 +1,115 @@
...
+  const passwordOutput = neon.getBranchRolePasswordOutput({
+    projectId: project.id,
+    branchId: project.defaultBranchId,
+    roleName: db.ownerName,
+  })
+
+  return $interpolate`postgresql://${passwordOutput.roleName}:${passwordOutput.password}@${project.databaseHost}/${db.name}?sslmode=require`
+}
+
+async function addDatabaseToElectric(
```
KyleAMathews, line 94: fine for now but in a future PR moving all these common helper functions into a common lib would be good

KyleAMathews (approved):
Looking great! Excited to see examples go live!

Another nice feature in the future would be posting comments w/ URLs to the sites — there's a good example here: https://github.com/taylorfsteele/TSS/blob/920d33974685280743fce243043099f3d45a4261/.github/workflows/pull-request.yml#L125C2-L168C14

examples/nextjs-example/sst.config.ts
balegas: We should be using hono client!

PR #1931, opened 2024-11-05
docs: fixup deployment docs

thruflo:


website/docs/integrations/cloudflare.md
```diff
@@ -54,7 +54,7 @@ export default {
     }
 
     const url = new URL(request.url)
-    const shapeUrl = `${ELECTRIC_URL}${url.pathname}`
+    const shapeUrl = `${ELECTRIC_URL}${url.pathname}${url.search}`
```
KyleAMathews, line 57: this also needs to copy over query params e.g. https://github.com/electric-sql/electric/blob/dadb9fb72b68f348326fc2bf807a2d0fae49a757/examples/auth/app/shape-proxy/route.ts#L6-L8

thruflo: Yup, that's what url.search does:

```js
> new URL('http://foo.com/baz?foo=bar')
URL {
  href: 'http://foo.com/baz?foo=bar',
  origin: 'http://foo.com',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'foo.com',
  hostname: 'foo.com',
  port: '',
  pathname: '/baz',
  search: '?foo=bar',
  searchParams: URLSearchParams { 'foo' => 'bar' },
  hash: ''
}
```

KyleAMathews: ah right :-D

PR #1932, opened 2024-11-05
fix(@electric-sql/client): The table option isn't required as a team using a proxy API might set the table there

KyleAMathews:


thruflo (commented):
Note that the options type is duplicated here https://electric-sql.com/docs/api/clients/typescript#options

(Would be nice to include this from the source. I couldn’t see a way to do this easily. Perhaps we could script it at build time).

KyleAMathews:
Ah right — thanks!

kevin-dp (approved):

PR #1934, opened 2024-11-05
fix: also change DATABASE_ID to ELECTRIC_DATABASE_ID

KyleAMathews:
Missed this when working on https://github.com/electric-sql/electric/issues/1927

### DIFF
--- a/.github/workflows/ts_test.yml
+++ b/.github/workflows/ts_test.yml
@@ -71,7 +71,7 @@ jobs:
       run:
         working-directory: ${{ matrix.package_dir }}
     env:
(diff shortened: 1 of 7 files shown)


thruflo (approved):
I observe that we don't have this env var documented yet. Which is fine as we don't have any multi-tenancy documentation yet.

Do the potential re-factoring changes that @magnetised and @kevin-dp are discussing impact anything around the API for multi-tenancy? Or are we able to document the usage when we want to?

KyleAMathews:
We can merge this now & they can then change it / add docs when they get to it.

PR #1939, opened 2024-11-06
chore: e2e test for multi tenancy

kevin-dp:
This PR adds an e2e test for multi tenancy.
It checks:
- adding new tenants
- per-tenant health check
- initial sync works
- live queries work and does not interfere between tenants
- tenants are persisted and reloaded when Electric restarts
- tenants can be deleted


robacourt (approved):
I'm not familiar with lux, but it looks good to me.

PR #1953, opened 2024-11-07
chore (sync service): refactor tenant manager to store tenant info in ETS table

kevin-dp:
This PR applies the refactoring proposed by @icehaunter which consists of storing the tenant information in a protected ETS table. Any process can read the tenant information from the ETS table.

#### Concurrent reads
On every shape request the tenant information is loaded from the ETS table.
Concurrent shape requests can read this information from the ETS table concurrently.

#### Serialised writes
Writes to the ETS table are less frequent (only when adding or removing tenants) and are still serialised through the tenant manager's genserver and thus are not prone to race conditions.

### DIFF
--- a/.changeset/funny-spoons-trade.md
+++ b/.changeset/funny-spoons-trade.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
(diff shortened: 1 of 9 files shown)


robacourt (approved):
Lovely!

Issue #1950, opened 2024-11-06
Throw error if a shape response doesn't include necessary headers


KyleAMathews:
Two people on discord today reported they hit an infinite loop where a shape request with offset=-1 came back with a 200 status but was missing the `electric-offset` header.

Both had proxies that for one reason or another had stripped the header(s).

Electric doesn't work without its headers so we should give a more helpful clue to people writing middleware on what is wrong.

The error message should look something like:

```
The response for the shape request to ${url} didn't include the following required headers:
- electric-offset
- etc

This is often due to a proxy not allowing all headers to be returned. See this doc page for more information: https://electric-url.com/something
```

thruflo:
That link at the end of the error message could go to a section on this in the troubleshooting guide.

KyleAMathews:
Yeah that'd be perfect

KyleAMathews:
Perhaps even simpler is we just check that we never make a request to the same URL twice. As that covers a lot of problems.

thruflo:
Yup, avoiding the scenarios where we hammer the sever with requests in a loop would be good.

There are cases where we do want to call the same URL more than once — eg when trying to sync a table that doesn’t exist yet, we get an error response and keep trying the same URL until the table is created. So there we maybe want more of a backoff.

PR #1957, opened 2024-11-07
chore (ts-client): validate required Electric headers on response

kevin-dp:
Fixes https://github.com/electric-sql/electric/issues/1950

Questions:
- Are the headers only required on 2XX responses or also on other responses?
   They are currently only being checked on 2XX responses.
- Are we checking *all* required headers?
   Currently we check `electric-offset`, `electric-handle`, and `electric-schema`

### DIFF
--- a/packages/typescript-client/src/error.ts
+++ b/packages/typescript-client/src/error.ts
@@ -48,3 +48,14 @@ export class FetchBackoffAbortError extends Error {
     super(`Fetch with backoff aborted`)
   }
 }
+
+export class MissingHeadersError extends Error {
+  constructor(url: string, missingHeaders: Array<string>) {
+    let msg = `The response for the shape request to ${url} didn't include the following required headers:\n`
+    missingHeaders.forEach((h) => {
+      msg += `- ${h}\n`
+    })
+    msg += `\nThis is often due to a proxy not allowing all headers to be returned.`
+    super(msg)
+  }
+}
--- a/packages/typescript-client/src/error.ts
+++ b/packages/typescript-client/src/error.ts
@@ -55,7 +55,7 @@ export class MissingHeadersError extends Error {
     missingHeaders.forEach((h) => {
       msg += `- ${h}\n`
     })
-    msg += `\nThis is often due to a proxy not allowing all headers to be returned.`
+    msg += `\nThis is often due to a proxy not setting CORS correctly so that all Electric headers can be read by the client.`
     super(msg)
   }
 }
--- a/packages/typescript-client/src/fetch.ts
+++ b/packages/typescript-client/src/fetch.ts
@@ -175,7 +175,8 @@ export function createFetchWithResponseHeadersCheck(
       addMissingHeaders(requiredElectricResponseHeaders)
 
       const input = args[0]
-      const url = new URL(input.toString())
+      const urlString = input.toString()
+      const url = new URL(urlString)
       if (url.searchParams.has(LIVE_QUERY_PARAM, `true`)) {
         addMissingHeaders(requiredLiveResponseHeaders)
       }
@@ -188,7 +189,7 @@ export function createFetchWithResponseHeadersCheck(
       }
 
       if (missingHeaders.length > 0) {
-        throw new MissingHeadersError(args[0].toString(), missingHeaders)
+        throw new MissingHeadersError(urlString, missingHeaders)
       }
     }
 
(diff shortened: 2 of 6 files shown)


msfstef (approved):
Looks good, I like the reuse of the fetch wrapper pattern!

I think possibly only 409 responses would also require an `electric-handle` to ensure it is possible to switch to the "right" shape, but I think doing just 200s is good enough to solve the issue of potentially misconfigured proxies

thruflo:
Can we also validate an `electric-cursor` header when we receive a `204` in response to a live request?

Not having the cursor param can result in re-requesting to the same URL.

thruflo:
Plus can we please write a section in the https://electric-sql.com/docs/guides/troubleshooting guide about this and include a link to it in the error message?

Basically we would want to explain that if headers are missing then your proxy needs to make sure it includes them, maybe with a code sample showing how (Kyle has some proxy code that clones the headers).

I can write this if you'd like.

packages/typescript-client/src/error.ts
```diff
@@ -48,3 +48,14 @@ export class FetchBackoffAbortError extends Error {
...
   }
 }
+
+export class MissingHeadersError extends Error {
+  constructor(url: string, missingHeaders: Array<string>) {
+    let msg = `The response for the shape request to ${url} didn't include the following required headers:\n`
+    missingHeaders.forEach((h) => {
+      msg += `- ${h}\n`
+    })
+    msg += `\nThis is often due to a proxy not allowing all headers to be returned.`
```
KyleAMathews, line 58:
```suggestion
    msg += `\nThis is often due to a proxy not setting CORS correctly so that all Electric headers can be read by the client.`
```

packages/typescript-client/src/fetch.ts
```diff
@@ -146,6 +150,33 @@ export function createFetchWithChunkBuffer(
   return prefetchClient
 }
 
+export const requiredElectricResponseHeaders = [
+  `electric-offset`,
+  `electric-handle`,
+  `electric-schema`,
```
KyleAMathews, line 156: `electric-cursor` & `electric-up-to-date` are also required.

KyleAMathews (changes requested):
What did y'all think also about the idea of also tracking which URLs we request and check that we don't request the same URL twice (within a given session)? 

This would catch any other issue we haven't thought of.

The only tricky thing here I can think of is we'd need to wipe the Set on shape rotation.

KyleAMathews:
Checking just 200s for headers seems fine as it covers all the headers.

packages/typescript-client/src/fetch.ts
kevin-dp: Our API spec says that the cursor is optional on the response, is that wrong?

thruflo: As [above](https://github.com/electric-sql/electric/pull/1957#issuecomment-2461998099) I believe we only return a cursor when returning a 204 to a live request.

msfstef: We actually include `electric-cursor` for all live requests, 200 or 204

KyleAMathews: Yeah ok it's a bit more tricky to check for these but it's still really bad if they're not available so it it's hard to check that'd be good. But with docs on what headers CORS allows through, probably just checking the minimal always sent set is enough

KyleAMathews: The point is the dev would see an error while developing and fix things.

kevin-dp: So, the docs are correct and `electric-cursor` is optional as it is only provided in live mode. I could modify the check such that it checks if the original URL contains the `live` query param and only then check for the presence of `electric-cursor`.

@KyleAMathews Regarding `electric-up-to-date`, it is only present on the last chunk we will get from the sync service, so we don't want to check this on every message?

kevin-dp:
@KyleAMathews Can you have another review at this PR to see if you approve the changes?
There are some required headers that are checked on all 2xx responses. There are also some headers that are required only in live mode and some only in non-live mode. I also check those on 2xx responses.

I will open a separate docs PR before merging this one.

KyleAMathews (approved):
:shipit: thanks!

kevin-dp:
For reference, this is the accompanying docs PR: https://github.com/electric-sql/electric/pull/1976

PR #1958, opened 2024-11-07
swift-client: initial sketch.

thruflo:


icehaunter:
@thruflo is this still relevant/has priority? WE've added a bunch of changes & features since this, so we can regenerate this from scratch with Claude/ChatGPT when actually needed, so I think we can close this

icehaunter:
@balegas said we can close this

PR #1959, opened 2024-11-07
rename: `up-to-date` to `frontier`.

thruflo:


samwillis:
I'd like to hold merging this until we chat a little about the difference between a differential dataflow frontier and what our up-to-date message does. While very similar, it doesn't do everything that the d2 version needs. If we are going to use d2 for stream processing we should just check that we don't wand an additional message that meets the semantics required.

key difference is we send out up-to-data at the end of the LSN/"timestamp", with D2 they are at the beginning of them.

thruflo:
I’m not particularly convinced that this PR is improvement. I think the idea that the up-to-date message indicates the data was up to date when the server generated it is valid/understandable and the term “up-to-date” is more readable and easier to generally talk about, when not in distributed systems edge case mode.

So perhaps we close this?

PR #1963, opened 2024-11-11
examples: add gatekeeper-auth example.

thruflo:
See https://github.com/electric-sql/electric/tree/thruflo/gatekeeper-auth-example/examples/gatekeeper-auth -- the example folder and subfolders all have READMEs.

I haven't yet updated the auth guide. I suspect it's best to merge this and update the guide in a separate PR.


examples/gatekeeper-auth/api/lib/api_web/router.ex
```diff
@@ -0,0 +1,45 @@
...
+    pipe_through :api
+
+    # The gatekeeper endpoint at `POST /gatekeeper/:table` authenticates the user,
+    # authorises the shape access, generates a shape-scoped auth token and returns
+    # this along with other config that the Electric client can use to stream the
+    # shape directly from Electric.
+    scope "/gatekeeper" do
+      pipe_through :gatekeeper
+
+      post "/:table", Electric.Phoenix.Gateway.Plug, client: &Authenticator.client/0
```
thruflo, line 31: @magnetised the `client` syntax / concept isn't the most obvious here. It's great that we can technically pass in an `Authenticator` via the client option. However the reason to be using a "client" in this context isn't super obvious. I'd much rather write e.g.:

```elixir
post "/:table", Electric.Phoenix.Gateway.Plug, authenticator: {MyAuthenticator, []}
```

Or even just the function that generates the auth headers:

```elixir
post "/:table", Electric.Phoenix.Gateway.Plug, auth_headers: &MyAuthenticator.authenticate_shape/2
```

Following the thought, that `Authenticator.authenticate_shape/2` function gets the `shape` and a `config` options keyword list. It doesn't get the `conn`. Which, when generating an auth token, might be useful context, for example to put a `user_id` in the token, etc.

examples/gatekeeper-auth/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+
+## How it works
+
+There are two steps to the gatekeeper pattern:
+
+1. first a client posts authentication credentials to a gatekeeper endpoint to generate an auth token
+2. the client then makes requests to Electric via an authorising proxy that validates the auth token against the shape request
+
+The auth token can be *shape-scoped* (i.e.: can sign a specific shape definition). This allows the proxy to authorise a shape request by comparing the shape signed into the token with the [shape defined in the request parameters](https://electric-sql.com/docs/quickstart#http-api). This allows you to:
```
kevin-dp, line 22: Nit: i found the explanation of signing the shape in the token a bit confusing. I think it would be more accurate and clear to use the "claim" terminology. This is what i propose:

> 
```suggestion
The auth token can be *shape-scoped* (i.e.: include a claim containing the shape definition). This allows the proxy to authorise a shape request by comparing the shape claim from the token with the [shape defined in the request parameters](https://electric-sql.com/docs/quickstart#http-api). This allows you to:
```

examples/gatekeeper-auth/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+<a href="./docs/img/gatekeeper-flow.jpg" target="_blank">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="./docs/img/gatekeeper-flow.dark.png" />
+    <source media="(prefers-color-scheme: light)" srcset="./docs/img/gatekeeper-flow.png" />
+    <img alt="Gatekeeper flow diagramme" src="./docs/img/gatekeeper-flow.jpg"
+        width="100%"
+    />
+  </picture>
+</a>
```
kevin-dp, line 42: Nit: in this diagram it looks like Electric replies directly to the client but actually it replies to the proxy which forwards it to the client. Perhaps we want to make this explicit in this diagram as it may otherwise cause confusion with the troubleshooting guide which will explain that the proxy must not strip the electric headers on the response.

EDIT: just noted that the explanation explain this. I'll let you decide whether the diagram needs to show this more explicitly or not.

examples/gatekeeper-auth/README.md
```diff
@@ -0,0 +1,334 @@
...
+      "http://localhost:4000/proxy/v1/shape?table=items&offset=-1&where=true"
+...
+< HTTP/1.1 403 Forbidden
+...
+```
+
+Note that we got an empty response when successfully proxied through to Electric above because there are no `items` in the database. If you like, you can create some, e.g. using `psql`:
+
+```console
+$ psql "postgresql://postgres:password@localhost:54321/electric?sslmode=disable"
```
kevin-dp, line 151: We're using `psql` here but it's not listed in the prerequisites.

examples/gatekeeper-auth/README.md
```diff
@@ -0,0 +1,334 @@
...
+        "public",
+        "items"
+      ]
+    },
+    "offset": "0_0"
+  }
+]
+```
+
+So far we've shown things working with Electric's lower-level [HTTP API](https://electric-sql.com/docs/api/http). You can also setup the [higher level clients](https://electric-sql.com/docs/api/clients/typescript) to use an auth token. See the [auth guide](https://electric-sql.com/docs/guides/auth) for more details.
```
kevin-dp, line 199: i believe "higher level" should be hyphenated in "higher-level clients" because it is a multi-word adjective before a noun.

examples/gatekeeper-auth/README.md
```diff
@@ -0,0 +1,334 @@
...
+```shell
+export AUTH_TOKEN="<token>"
+```
+
+First let's make a `GET` request to the proxy endpoint *without* the auth token. It will be rejected with a `403` status:
+
+```console
+$ curl -sv "http://localhost:4000/proxy/v1/shape?table=items&offset=-1"
+...
+< HTTP/1.1 403 Forbidden
```
kevin-dp, line 124: Shouldn't the proxy return a 401 as in the Caddy example?

examples/gatekeeper-auth/README.md
```diff
@@ -0,0 +1,335 @@
...
+2. with the [API as gatekeeper and Caddy as the proxy](#2-caddy-as-proxy)
+3. with the [API as gatekeeper and an edge function as the proxy](#3-edge-function-as-proxy)
+
+It makes sense to run through these in order.
+
+### Pre-reqs
+
+You need [Docker Compose](https://docs.docker.com/compose/) and [curl](https://curl.se). We also (optionally) use [`psql`](https://www.postgresql.org/docs/current/app-psql.html) and pipe into [`jq`](https://jqlang.github.io/jq/) for JSON formatting.
+
+The instructions below all use the same [`./docker-compose.yaml`](./docker-compose.yaml) file in the root of this repo. With a different set of services and environment variables.
```
magnetised, line 72: the dockerfile is not at the root of the repo, so this is confusing considering the hyperlink points to the right place

what about "the dockerfile in this folder"

examples/gatekeeper-auth/README.md
```diff
@@ -0,0 +1,335 @@
...
+
+In a new terminal, make a `POST` request to the gatekeeper endpoint:
+
+```console
+$ curl -sX POST "http://localhost:4000/gatekeeper/items" | jq
+{
+  "headers": {
+    "Authorization": "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJKb2tlbiIsImV4cCI6MTczMTUwMjM2OSwiaWF0IjoxNzMxNDk1MTY5LCJpc3MiOiJKb2tlbiIsImp0aSI6IjMwM28zYmx0czN2aHYydXNiazAwMDJrMiIsIm5iZiI6MTczMTQ5NTE2OSwic2hhcGUiOnsibmFtZXNwYWNlIjpudWxsLCJ0YWJsZSI6Iml0ZW1zIiwid2hlcmUiOm51bGwsImNvbHVtbnMiOm51bGx9fQ.8UZehIWk1EDQ3dJ4ggCBNkx9vGudfrD9appqs8r6zRI"
+  },
+  "url": "http://localhost:3000/proxy/v1/shape",
```
magnetised, line 103: you have the port of the `/proxy` endpoint as 3000 here, but below you're making requests to `:4000/proxy/...`. 

For consistency the params returned from the gatekeeper auth request should work as config for the client, so I think the `url` here should be `http://localhost:4000/proxy/v1/shape`

and in fact in the next para you say the url is `:4000` so this is a typo

examples/gatekeeper-auth/README.md
```diff
@@ -0,0 +1,335 @@
...
+
+```console
+$ curl -sv --header "Authorization: Bearer ${AUTH_TOKEN}" \
+      "http://localhost:8080/v1/shape?table=items&offset=-1&where=true"
+...
+< HTTP/1.1 403 Forbidden
+...
+```
+
+Take a look at the [`./caddy/Caddyfile`](./caddy/Caddyfile) for more details.
```
magnetised, line 265: Impressive Caddyfile foo 🥇

magnetised (approved):
Excellent

examples/gatekeeper-auth/api/lib/api_web/router.ex
magnetised: good points.

PR #1965, opened 2024-11-11
Drop replication slot on tenant delete

### LINKED ISSUE #1924: DELETE database_id shall clean up replication slot
When we DELETE a database we tear down the processes related with it, but still leave the replication slot. Given this is an action to delete the database, it means the user is no longer interested in shapes sources by it and should delete any associated resources: storage and replication slot.

Leaving replication slots hanging is a problem for users and the WAL file will keep growing unexpectedly. We should also take care of teardown for single-tenant electric #1923.

robacourt:
Fixes #1924

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -112,6 +112,23 @@ defmodule Electric.Connection.Manager do
     GenServer.call(server, :get_status)
   end
 
+  def await_active(server) do
+    case get_status(server) do
+      :active ->
+        :ok
+
+      _ ->
+        Process.sleep(10)
+        await_active(server)
+    end
+  end
+
+  def drop_replication_slot(server) do
+    # Ensure that the connection pool is available
+    await_active(server)
+    GenServer.call(server, :drop_replication_slot)
+  end
+
   def exclusive_connection_lock_acquired(server) do
     GenServer.cast(server, :exclusive_connection_lock_acquired)
   end
@@ -184,6 +201,13 @@ defmodule Electric.Connection.Manager do
       end
 
     {:reply, status, state}
+  end
+
+  def handle_call(:drop_replication_slot, _from, %{pool_pid: pool_pid} = state)
+      when pool_pid != nil do
+    publication_name = Keyword.fetch!(state.replication_opts, :publication_name)
+    Postgrex.query!(pool_pid, "DROP PUBLICATION #{publication_name}", [])
+    {:reply, :ok, state}
   end
 
   @impl true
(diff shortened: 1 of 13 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -186,6 +203,13 @@ defmodule Electric.Connection.Manager do
     {:reply, status, state}
   end
 
+  def handle_call(:drop_replication_slot, _from, %{pool_pid: pool_pid} = state)
+      when pool_pid != nil do
+    publication_name = Keyword.fetch!(state.replication_opts, :publication_name)
+    Postgrex.query!(pool_pid, "DROP PUBLICATION #{publication_name}", [])
```
robacourt, line 209: There isn't a way to switch the `Postgrex.ReplicationConnection` from stream mode back into query mode to run this query without waiting for something to come in on the stream. WAL Keep alives will eventually come but this is rather slow. It seemed better to just run a query directly on the pool.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -112,6 +112,23 @@ defmodule Electric.Connection.Manager do
...
+        Process.sleep(10)
+        await_active(server)
+    end
+  end
+
+  def drop_replication_slot(server) do
+    # Ensure that the connection pool is available
+    await_active(server)
+    GenServer.call(server, :drop_replication_slot)
+  end
```
robacourt, line 130: I don't particularly like this polling wait. The connection manager could drop it immediately inside the call if the connection pool has already been set up, or if it hasn't, register a request for the slot to be deleted after `:start_connection_pool` has completed.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -186,6 +191,20 @@ defmodule Electric.Connection.Manager do
...
+    {:reply, :ok, state}
+  end
+
+  def handle_call(:drop_replication_slot, from, state) do
+    {:noreply, %{state | drop_slot_requesters: [from | state.drop_slot_requesters]}}
+  end
+
+  defp drop_publication(state) do
+    publication_name = Keyword.fetch!(state.replication_opts, :publication_name)
+    Postgrex.query!(state.pool_pid, "DROP PUBLICATION #{publication_name}", [])
```
robacourt, line 205: There isn't a way to switch the Postgrex.ReplicationConnection from stream mode back into query mode to run this query without waiting for something to come in on the stream. WAL Keep-alives will eventually come but this is rather slow. It seemed better to just run a query directly on the pool.

packages/sync-service/lib/electric/tenant_tables.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Electric.TenantTables do
```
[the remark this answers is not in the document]
robacourt: Done

PR #1966, opened 2024-11-12
chore (sync-service): unique publication and slot names per tenant

kevin-dp:
This PR modifies the publication and slot names such that they contain the hash of the tenant ID.
Since tenant IDs are expected to be unique across all Electric instances this will result in unique publication and slot names. This makes it possible for several Electric instances to connect to the same PG database.

### DIFF
--- a/.changeset/fluffy-eagles-fly.md
+++ b/.changeset/fluffy-eagles-fly.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Make publication and slot names unique per tenant.
--- a/packages/sync-service/lib/electric/tenant_manager.ex
+++ b/packages/sync-service/lib/electric/tenant_manager.ex
@@ -195,17 +195,13 @@ defmodule Electric.TenantManager do
     # Note that this assumes that tenant IDs are unique across all Electric instances.
     tenant_hash = tenant_id |> :erlang.phash2() |> Integer.to_string()
 
-    updated_app_config =
+    per_tenant_app_config =
       app_config
       |> update_in([:replication_opts, :publication_name], &(&1 <> "_" <> tenant_hash))
       |> update_in([:replication_opts, :slot_name], &(&1 <> "_" <> tenant_hash))
 
-    # tenant_publication_name = app_config.replication_opts.publication_name <> "_" <> tenant_hash
-    # tenant_slot_name = app_config.replication_opts.slot_name <> "_" <> tenant_hash
-    # updated_replication_opts = %{app_config.replication_opts | publication_name: tenant_publication_name, slot_name: tenant_slot_name}
-
     start_tenant_opts = [
-      app_config: updated_app_config,
+      app_config: per_tenant_app_config,
       electric_instance_id: electric_instance_id,
       tenant_id: tenant_id,
       connection_opts: connection_opts,
--- a/packages/sync-service/lib/electric/tenant_manager.ex
+++ b/packages/sync-service/lib/electric/tenant_manager.ex
@@ -187,8 +187,25 @@ defmodule Electric.TenantManager do
           connection_opts: connection_opts
         ]
 
+    # We deliberately don't include the electric instance ID
+    # in the publication and slot name because
+    # those are shared between all Electric instances connected to the same PG DB
+    # and there may be cases where we want one Electric to take over the publication
+    # and slot for a certain tenant of another Electric instance, e.g. during a rolling deploy.
+    # Note that this assumes that tenant IDs are unique across all Electric instances.
+    tenant_hash = tenant_id |> :erlang.phash2() |> Integer.to_string()
+
+    updated_app_config =
+      app_config
+      |> update_in([:replication_opts, :publication_name], &(&1 <> "_" <> tenant_hash))
+      |> update_in([:replication_opts, :slot_name], &(&1 <> "_" <> tenant_hash))
+
+    # tenant_publication_name = app_config.replication_opts.publication_name <> "_" <> tenant_hash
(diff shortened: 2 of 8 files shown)


packages/sync-service/lib/electric/tenant_manager.ex
```diff
@@ -187,8 +187,25 @@ defmodule Electric.TenantManager do
...
+      |> update_in([:replication_opts, :publication_name], &(&1 <> "_" <> tenant_hash))
+      |> update_in([:replication_opts, :slot_name], &(&1 <> "_" <> tenant_hash))
+
+    # tenant_publication_name = app_config.replication_opts.publication_name <> "_" <> tenant_hash
+    # tenant_slot_name = app_config.replication_opts.slot_name <> "_" <> tenant_hash
+    # updated_replication_opts = %{app_config.replication_opts | publication_name: tenant_publication_name, slot_name: tenant_slot_name}
+
     start_tenant_opts = [
-      app_config: app_config,
+      app_config: updated_app_config,
```
robacourt, line 208: The app config is currently fixed for all tenants, you're making it tenant specific which changes it's semantics and makes it a bit confusing in my opinion. I would instead rename `publication_name` to `publication_prefix` and do the tenant ID hash append elsewhere.

packages/sync-service/lib/electric/tenant_manager.ex
```diff
@@ -187,8 +187,25 @@ defmodule Electric.TenantManager do
...
+    # and slot for a certain tenant of another Electric instance, e.g. during a rolling deploy.
+    # Note that this assumes that tenant IDs are unique across all Electric instances.
+    tenant_hash = tenant_id |> :erlang.phash2() |> Integer.to_string()
+
+    updated_app_config =
+      app_config
+      |> update_in([:replication_opts, :publication_name], &(&1 <> "_" <> tenant_hash))
+      |> update_in([:replication_opts, :slot_name], &(&1 <> "_" <> tenant_hash))
+
+    # tenant_publication_name = app_config.replication_opts.publication_name <> "_" <> tenant_hash
```
msfstef, line 203: nit: remove commented out code?

robacourt (changes requested):

packages/sync-service/lib/electric/tenant_manager.ex
```diff
@@ -187,8 +187,25 @@ defmodule Electric.TenantManager do
...
 
+    # We deliberately don't include the electric instance ID
+    # in the publication and slot name because
+    # those are shared between all Electric instances connected to the same PG DB
+    # and there may be cases where we want one Electric to take over the publication
+    # and slot for a certain tenant of another Electric instance, e.g. during a rolling deploy.
+    # Note that this assumes that tenant IDs are unique across all Electric instances.
+    tenant_hash = tenant_id |> :erlang.phash2() |> Integer.to_string()
+
+    updated_app_config =
```
msfstef, line 198:
```suggestion
    per_tenant_app_config =
```
maybe more semantically rich than "updated"?

packages/sync-service/lib/electric/tenant_manager.ex
robacourt: To give an example of the confusion, `Application.Configuration` is implemented as a `persistent_term` so that any piece of code can grab the config, but if it does this it will get the wrong `publication_name`

.changeset/fluffy-eagles-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Make publication and slot names unique per tenant.
```
msfstef, line 5: If we decide to add the hash for the default tenant (i.e. "non-multi-tenancy mode") then this is a breaking change that would benefit from a little bit more explantation in the changeset or elsewhere. We should advise users that the new deploy will "clear" their cache and they will need to clean up their old replication slot.

We need to make the decision as to whether we want to not include this hash for the cases where `tenant_id === default_tenant`. I think it makes sense, especially because currently if `ELECTRIC_TENANT_ID` is not specified (which is the most likely scenario for non-multi-tenancy) we auto generate a UUID which means that on every electric boot the default tenant id is going to be different and would create a different replication slot.

packages/sync-service/lib/electric/tenant_manager.ex
kevin-dp: Doing the tenant ID hash append elsewhere is error prone. I used to do it in `tenant/supervisor.ex` on L64 and 66 but the app config is also used in other places where the append would also have to happen, resulting in code duplication and easy to forget one which would break it. The most robust approach is to turn the app config in a per tenant config such that it contains the right publication and slot names everywhere it is passed. Perhaps let's rename it to `per_tenant_app_config` as per @msfstef's suggestion.

.changeset/fluffy-eagles-fly.md
kevin-dp: Yes, there are pros and cons for both approaches (always using the hash or only in multi tenant mode). Let's discuss this with the wider team and make a decision.

packages/sync-service/lib/electric/tenant_manager.ex
robacourt: Well `TenantConfig` would be more concise 😄 If you want to go that route, then you'll need to change the semantics of the module and probably get of the rid of the persistant_term. It sounds like this is worth leaving until the multi-tenant code has been separated from the core sync-service code.

kevin-dp:
Status update: this PR has been put on hold until we extract multi tenancy out of the core Electric library.

kevin-dp:
Closing as this is addressed by https://github.com/electric-sql/electric/pull/1991

PR #1968, opened 2024-11-12
Drop all databases on `DELETE /v1/admin/databases`

### LINKED ISSUE #1923: Terminate electric admin endpoint
We need a termination endpoint for Electric to cleanup storage and remove any existing replication slot. 

Right now if someone wants to stop using Electric, they need to manually remove the replication slot from Postgres, otherwise WAL  will keep growing.

robacourt:
Fixes #1923

### DIFF
--- a/packages/sync-service/test/electric/plug/remove_all_databases_plug_test.exs
+++ b/packages/sync-service/test/electric/plug/remove_all_databases_plug_test.exs
@@ -1,0 +1,66 @@
+defmodule Electric.Plug.RemoveAllDatabasesPlugTest do
+  use ExUnit.Case, async: true
+  import Plug.Conn
+
+  alias Electric.Plug.RemoveAllDatabasesPlug
+
+  import Support.ComponentSetup
+  import Support.DbSetup
+  import Support.TestUtils, only: [with_electric_instance_id: 1]
+
+  alias Support.Mock
+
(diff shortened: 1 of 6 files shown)


packages/sync-service/test/electric/plug/remove_all_databases_plug_test.exs
```diff
@@ -0,0 +1,66 @@
...
+      ]
+
+      conn =
+        Plug.Test.conn("DELETE", "/")
+        |> assign(:config, config)
+        |> RemoveAllDatabasesPlug.call([])
+
+      assert conn.status == 200
+
+      assert Electric.Tenant.Persistence.load_tenants!(
```
kevin-dp, line 57: Not sure if loading from persistence is the right thing to check here.
I'd check that `Electric.TenantManager.get_only_tenant(config)` returns `{:error, :not_found}`

kevin-dp (approved):
Overall LGTM, left some minor comments.

robacourt:
Closing as the purpose of this PR is to allow is to allow users of electric (not the cloud) to clear all replication slots so they can permanently stop electric. But since the multi-tenancy separation there is no concept of "all databases" now in electric, there is only one database, so all we need is an endpoint to delete the one database, which I'll do on a separate PR.

PR #1971, opened 2024-11-13
Drop replication slot on database delete

robacourt:
Following on from #1965 this drops the replication slot as well as the publication.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -60,7 +60,8 @@ defmodule Electric.Connection.Manager do
       # ID used for process labeling and sibling discovery
       :stack_id,
       :tweaks,
-      drop_slot_requesters: []
+      awaiting_active: [],
+      drop_slot_requested: false
     ]
   end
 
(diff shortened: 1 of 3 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -577,4 +582,26 @@ defmodule Electric.Connection.Manager do
 
     log_collector_pid
   end
+
+  defp drop_publication(%{pool_pid: pool} = state) do
+    publication_name = Keyword.fetch!(state.replication_opts, :publication_name)
+    slot_name = Keyword.fetch!(state.replication_opts, :slot_name)
+    slot_temporary? = Keyword.fetch!(state.replication_opts, :slot_temporary?)
+
+    if slot_temporary? == false do
```
kevin-dp, line 591: let's use negation `if !slot_temporary? do ...`

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -577,4 +582,26 @@ defmodule Electric.Connection.Manager do
...
+      execute_logging_any_error(pool, "SELECT pg_drop_replication_slot('#{slot_name}');")
+    end
+
+    execute_logging_any_error(pool, "DROP PUBLICATION #{publication_name}")
+  end
+
+  defp execute_logging_any_error(pool, query) do
```
kevin-dp, line 598: nit: i find this name a bit confusing, what about `execute_with_error_logging` or `execute_and_log_errors` ?

packages/sync-service/test/electric/plug/remove_database_plug_test.exs
```diff
@@ -67,6 +67,14 @@ defmodule Electric.Plug.RemoveDatabasePlugTest do
       assert conn.status == 200
       assert Jason.decode!(conn.resp_body) == ctx.tenant_id
 
+      # Ensure the replication slot has been dropped
```
kevin-dp, line 70: Should we also check the possibility where the slot is not dropped because `slot_temporary?` was set to `true`?

kevin-dp (approved):
The changes look good to me! 👍 
Left a couple of minor comments.

PR #1972, opened 2024-11-13
chore: more renamings from shape id -> shape handle

KyleAMathews:


msfstef (approved):
Good from my side but @icehaunter and @kevin-dp asked to avoid too many changes while they do the multi-tenancy extraction.
This is just changes to comments and strings so I suspect it should be good to merge but good to check with them.

PR #1976, opened 2024-11-14
docs: document missing headers problem

kevin-dp:
This PR documents the problem when Electric client complains about missing headers and how to solve it.
It was requested in https://github.com/electric-sql/electric/pull/1957 and that PR will need to be updated such that the error message refers to the section we add to the troubleshooting guide in this PR. Once this PR has been approved and the title of the subsection won't change we can add the link in the other PR.

TODO:
- Potentially add a code snippet showing proxy configuration. Although it seems that Caddy forwards the headers automatically unless you explicitly remove headers. So not sure if a snippet is needed, @KyleAMathews ?

### DIFF
--- a/website/docs/guides/troubleshooting.md
+++ b/website/docs/guides/troubleshooting.md
@@ -71,7 +71,7 @@ docker compose up
 
 Electric creates a durable replication slot in Postgres to prevent data loss during downtime. 
 
-During normal execution, Electric consumes the WAL file and keeps advancing `confirmed_flush_lsn`. However, if Electric is disconnected, the WAL file accumulates the changes that haven't be delivered to Electric.
+During normal execution, Electric consumes the WAL file and keeps advancing `confirmed_flush_lsn`. However, if Electric is disconnected, the WAL file accumulates the changes that haven't been delivered to Electric.
 
 ##### Solution &mdash; Remove replication slot after Electric is gone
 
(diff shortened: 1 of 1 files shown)


website/docs/guides/troubleshooting.md
```diff
@@ -72,3 +72,13 @@ If you don't consume this publication, the WAL can fill up and your Postgres dat
 ##### Solution &mdash; run Electric
 
 The simplest way to avoid this is to make sure you're running the Electric sync service against Postgres. This will consume the publication and allow the WAL to be released.
+
+### Missing headers &mdash; why is the client complaining about missing headers?
+
+When Electric responds to shape requests it includes headers that are required by the client to follow the shape log.
+It is common to run Electric behind a proxy to authenticate users and authorise shape requests.
+However, the proxy might not keep the response headers in which case the client may complain about missing headers.
```
KyleAMathews, line 80: There are two things here:

- the main thing here that the proxy must copy the response headers from the server to the response to the user.

Beyond that, Electric already exposes the right CORS headers to make things work — the main problem we've seen if people overriding `Access-Control-Expose-Headers` when they respond to a prefllight request with `OPTIONS` method — if they set `Access-Control-Expose-Headers` there, then that seems to take precedence over what Electric returns.

thruflo (approved):

Issue #1983, opened 2024-11-15
Audit errors in Typescript class


KyleAMathews:
The typescript client seems to be swallowing errors at times e.g. those thrown in the while loop — which makes it very difficult to debug at times.

Best practice for errors in js classes are:

- always throw, never swallow errors
- have named error classes so people can match against these

Issue #1982, opened 2024-11-15
Best pattern for handling 401/403 token expiry?


thruflo:
I've written a little client for the new gatekeeper auth pattern as per https://github.com/electric-sql/electric/pull/1963

Basically I need to handle token expiry and when a token expires, get a new one and then reconnect. I'm doing this with the `Shape` and `ShapeStream`: https://github.com/electric-sql/electric/blob/thruflo/auth-guide/examples/gatekeeper-auth/client/index.ts

I have this working, but AFAICS it requires having access to the `stream` instance to process messages to manually keep track of the last offset. I wonder:

- could the ShapeStream provide public read access to `lastOffset`
- what about code that's using `useShape` or doesn't want to be handling a shape stream?

I'm thinking that anywhere we support providing auth headers, we should have an easy way of handling a FetchError and resetting them to reconnect with new auth (or whatever).

What do you think?

KyleAMathews:
Getting a fresh token had come up before and one idea is that for headers/params could take an async function vs. a straight value — then that function would ensure that the token is still valid on each request.

thruflo:
Yup, that’s good.

There’s also the control flow issue that atm, the ShapeStream just stops on a 4xx status. So I believe a Shape or a useShape with a `headers` option just isn’t going to try to reconnect.

It might be nice to be able to explicitly configure a function that handles 4xx errors where the return value determines whether to retry with new request config or not?

Or is this what the fetchClient option is already able to support?

KyleAMathews:
yeah there's no way to automatically restart a shapestream if it errors. We have a problem that in general there's no way to easily listen for errors (and optionally do something about them).

thruflo:
If we want to "restart" by creating a new stream, then this will help: https://github.com/electric-sql/electric/pull/1989

KyleAMathews:
@balegas @kevin-dp @msfstef and I chatted about this a bit earlier (amongst other topics). We were discussing adding a global `onError` handler to streams for https://github.com/electric-sql/electric/pull/1985 — as part of that, people could handle certain types of errors by returning an update to some of the stream options. Then internally the stream could retry with the new options.

So e.g.

```ts
new ShapeStream({
  url,
  params: {
    token,
  },
  onError: async (error) => {
    if (error instanceof FetchError & error.status === 401) {
      const token = await getToken()
      return { params: { token }}
    }
  },
})
```

[...]

PR #1985, opened 2024-11-16
fix: always throw errors & always use named Error classes

KyleAMathews:
Fixes https://github.com/electric-sql/electric/issues/1983

### DIFF
--- a/.changeset/fair-pants-pretend.md
+++ b/.changeset/fair-pants-pretend.md
@@ -2,4 +2,15 @@
 "@electric-sql/client": patch
 ---
 
-always throw errors & always use named Error classes
+refactor: improve error handling with new error classes & stream control
+
+- Add `autoStart` option to ShapeStream constructor enabling two error handling patterns:
+  1. try/catch with `autoStart: false`
+  2. error handler in subscribe callback
+- Add new error classes:
+  - MissingShapeUrlError
+  - InvalidSignalError
+  - MissingShapeHandleError
+  - ReservedParamError
+  - ParserNullValueError
+  - ShapeStreamAlreadyRunningError
--- a/.changeset/fair-pants-pretend.md
+++ b/.changeset/fair-pants-pretend.md
@@ -4,6 +4,7 @@
 
 refactor: improve error handling with new error classes & stream control
 
+- Add `onError` handler to ShapeStream for centralized error handling
 - Add `autoStart` option to ShapeStream constructor enabling two error handling patterns:
   1. try/catch with `autoStart: false`
   2. error handler in subscribe callback
@@ -14,3 +15,4 @@ refactor: improve error handling with new error classes & stream control
   - ReservedParamError
   - ParserNullValueError
   - ShapeStreamAlreadyRunningError
+- Improve error propagation through ShapeStream lifecycle
--- a/.changeset/fair-pants-pretend.md
+++ b/.changeset/fair-pants-pretend.md
@@ -5,9 +5,6 @@
 refactor: improve error handling with new error classes & stream control
 
 - Add `onError` handler to ShapeStream for centralized error handling
-- Add `autoStart` option to ShapeStream constructor enabling two error handling patterns:
-  1. try/catch with `autoStart: false`
-  2. error handler in subscribe callback
 - Add new error classes:
   - MissingShapeUrlError
   - InvalidSignalError
--- a/.changeset/fair-pants-pretend.md
+++ b/.changeset/fair-pants-pretend.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
+always throw errors & always use named Error classes
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -168,12 +168,7 @@ export interface ShapeStreamInterface<T extends Row<unknown> = Row> {
     callback: (messages: Message<T>[]) => MaybePromise<void>,
     onError?: (error: FetchError | Error) => void
   ): void
-  unsubscribeAllUpToDateSubscribers(): void
   unsubscribeAll(): void
-  subscribeOnceToUpToDate(
-    callback: () => MaybePromise<void>,
-    error: (err: FetchError | Error) => void
-  ): () => void
 
   isLoading(): boolean
   lastSyncedAt(): number | undefined
@@ -237,10 +232,6 @@ export class ShapeStream<T extends Row<unknown> = Row>
       (messages: Message<T>[]) => MaybePromise<void>,
       ((error: Error) => void) | undefined,
     ]
-  >()
-  readonly #upToDateSubscribers = new Map<
-    number,
-    [() => void, (error: FetchError | Error) => void]
   >()
 
   #lastOffset: Offset
@@ -380,7 +371,6 @@ export class ShapeStream<T extends Row<unknown> = Row>
             continue
           } else if (e.status >= 400 && e.status < 500) {
             // Notify subscribers
-            this.#sendErrorToUpToDateSubscribers(e)
             this.#sendErrorToSubscribers(e)
 
             // 400 errors are not actionable without additional user input,
@@ -422,7 +412,6 @@ export class ShapeStream<T extends Row<unknown> = Row>
 
         // Update isUpToDate
         if (batch.length > 0) {
-          const prevUpToDate = this.#isUpToDate
           const lastMessage = batch[batch.length - 1]
           if (isUpToDateMessage(lastMessage)) {
             this.#lastSyncedAt = Date.now()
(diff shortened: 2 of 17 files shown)


KyleAMathews:
I'm skeptical now too of our passing errors to the subscribe functions. We have no way of recovering from the errors and also try/catch is the normal way of seeing errors. I'll remove those in another pr unless people disagree.

KyleAMathews:
It'd be nice to have an observable style interface as well but we're not using that anywhere so in practice right now we just lose the errors. Observables are the exception so people should opt into that interface.

KyleAMathews:
Also if do observables we need the third parameter for the stream complete callback.

KyleAMathews:
I need to update a few tests to catch errors that are thrown.

KyleAMathews:
Hmm actually these are all 404s — which suggests they're related to the multi-tenant work?

packages/typescript-client/src/error.ts
```diff
@@ -48,3 +48,42 @@ export class FetchBackoffAbortError extends Error {
     super(`Fetch with backoff aborted`)
   }
 }
+
+export class MissingShapeUrlError extends Error {
+  constructor() {
+    super(`Invalid shape options: missing required url parameter`)
+    this.name = `MissingShapeUrlError`
```
kevin-dp, line 55: Should we also overwrite `this.name` in `FetchBackoffAbortError`?

kevin-dp:
> Hmm actually these are all 404s — which suggests they're related to the multi-tenant work?

If they are multi tenant related they should be gone once we extract multi tenancy to a separate app.
Planning to finish that today, once it's in main we could rebase to see if the errors persist.

thruflo (commented):
This looks much clearer and more explicit and great to have the specific errors documented.

One observation is that this leads to two places where there is similar error handling: one is on instantiation and one is in the subscribe handler. I see the types of error handling may be different for the two locations -- I.e.: validate params on instantiate, handle FetchError ongoing. So I think this makes sense. But just to float the alternative, it could perhaps be possible to have a single try catch around a `stream.run()` function.

I know you're questioning the subscribe handler in the comments here. I do think it would be helpful to have a definitive pattern for how to handle 4xx errors from downstream Shape / useShape style functions.

In reality, there are circumstances where 4xx errors are transient or fixed on retry. For example, where they will be fixed by time passing or a new auth token, etc. When you have a component with a useShape binding in a local app, the sync binding needs to be absolutely bulletproof.

So I think we are going to have to expose a "handle FetchErrror when running" callback and I think we should have the result of that callback be able to determine whether the client continues with a backoff algorithm and also to potentially change the shape options. This could then be the mechanism to "restart" the shape stream, because actually if the return value is `[true, {...updated_options...}]` (or whatever) then the stream actually never stops.

KyleAMathews:
> But just to float the alternative, it could perhaps be possible to have a single try catch around a stream.run() function.

That's the Elixir client :-P for JS, when you start a new ShapeStream, it immediately starts running.

I hear you about needing to make transient 401 errors easier to handle — you can handle it with a try/catch which then restarts the stream — the problem with this of course is then you have to restart the fetching every time and it's not particularly ergonomic. 

Exposing a way to customize error handling makes sense too — it'd be easy to mess up the shape stream that way but the server would protect against that by forcing the client to refetch.

These are separate problems from what this PR is trying to fix so let's continue the discussion in #1982

packages/typescript-client/src/error.ts
KyleAMathews: Oh right, yeah missed that one.

thruflo:
Cool, let's get this merged.

N.b.:

> when you start a new ShapeStream, it immediately starts running.

When I read this I immediately thought: if you don't immediately pass the stream to a shape, you may miss messages from the stream that are fired before you register subscribers. Presumably this is an "add it when people ask for it" feature but an `autoRun: false` option + `stream.run()` combo may be useful at some point.

KyleAMathews:
Ok I hadn't wrapped my head around this when I banged this out Friday evening — but yeah — calling `run()` from the constructor does mean that a try/catch around the `new ShapeStream` call won't catch any errors thrown while the stream is running...

So `autoRun` false would work or... just disable autoRun and make `.run()` an explicit step as then error catching is neatly split between constructing the class and running the class.

balegas:
> So autoRun false would work or... just disable autoRun and make .run() an explicit step as then error catching is neatly split between constructing the class and running the class.

I'm a strong advocate for separating creating and running the stream. There are errors you want to catch early because those are fatal errors and it's worthless to start the stream. In general any side effects from a constructor are hard to work with because you cannot await.

(cc @msfstef )

KyleAMathews:
@balegas yeah I added an autoRun option but we could just make the two step process the default. This is a breaking change but now's the time to do that.

KyleAMathews:
Let's reconvene tomorrow to chat about this

KyleAMathews:
@kevin-dp I incorporated your changes but commented out the retry logic as that'll need tests & docs and this PR is getting way too long in the tooth as it is — so you or someone else can pick that up in a subsequent PR.

I'm out of time to work on this today — if someone wants to figure out why the react-hooks tests are all not working, that'd be great.

.changeset/fair-pants-pretend.md
```diff
@@ -0,0 +1,18 @@
+---
+"@electric-sql/client": patch
+---
+
+refactor: improve error handling with new error classes & stream control
+
+- Add `onError` handler to ShapeStream for centralized error handling
+- Add `autoStart` option to ShapeStream constructor enabling two error handling patterns:
```
msfstef, line 8: this comment about `autoStart` (and the list below) should be removed from the changeset

packages/typescript-client/src/client.ts
```diff
@@ -63,6 +67,12 @@ type ReservedParamKeys =
 
 type ParamsRecord = Omit<Record<string, string>, ReservedParamKeys>
 
+// type RetryOpts = {
```
msfstef, line 70: nit: remove commented out code?

packages/typescript-client/src/client.ts
```diff
@@ -420,14 +440,37 @@ export class ShapeStream<T extends Row<unknown> = Row>
       }
     } catch (err) {
       this.#error = err
+      if (this.#onError) {
+        this.#onError(err as Error)
+        // TODO add tests and documentation for this.
```
msfstef, line 445: nit: remove until we actually implement - not that much code to keep? feel free to ignore me though if you want to keep it in and/or we are implementing this asap

packages/typescript-client/src/client.ts
```diff
@@ -420,14 +440,37 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+      throw this.#error
     } finally {
       this.#connected = false
     }
   }
 
   subscribe(
     callback: (messages: Message<T>[]) => MaybePromise<void>,
-    onError?: (error: FetchError | Error) => void
+    onError: (error: FetchError | Error) => void = () => {}
```
msfstef, line 473: do we need the `FetchError` type? is `Error` not enough? (and if not, why is the `FetchError` not an instance of `Error`?)

packages/typescript-client/src/client.ts
```diff
@@ -444,11 +487,11 @@ export class ShapeStream<T extends Row<unknown> = Row>
 
   subscribeOnceToUpToDate(
     callback: () => MaybePromise<void>,
-    error: (err: FetchError | Error) => void
+    onError: (error: FetchError | Error) => void = () => {}
```
msfstef, line 490: same as above - why `FetchError` type?

msfstef (approved):
Nice, I like the named errors! I've left some requests for changes (especially removing the `autoStart` mention in the changeset)

packages/typescript-client/src/client.ts
kevin-dp: I've implemented this already here: https://github.com/electric-sql/electric/pull/2011
I'm a bit reluctant to removing it here as it will create conflicts with that PR.

packages/typescript-client/src/client.ts
kevin-dp: Indeed, i noticed the same here. `FetchError` is an instance of `Error` so i'll just remove keep `Error` here.

kevin-dp:
I cherry picked 12fd091c4fde2972be0d691953dc53586bfc6d5a because it was conflicting with that PR which is in main and rebasing is a pain because of the enormous amounts of commits in the branch + some of those commits introduce the explicit start feature and then removes it in a commit later down the line which causes a lot of conflicts.

KyleAMathews:
Thanks for finishing this up @kevin-dp! I'm very much looking forward to always seeing nice clear errors now 😆

PR #1987, opened 2024-11-18
examples: Linearlite load initial data

balegas:
A small one to load initial data in Linearlite example


KyleAMathews (approved):
:shipit:

PR #1989, opened 2024-11-18
client: expose shape and stream `lastOffset` via public getter.

thruflo:


balegas (approved):
I think there is no harm is exposing it.. I've been building on this on the examples.

PR #1991, opened 2024-11-19
chore!: restructure the app to be usable as a library

icehaunter:
1. Splits out all the multitenancy logic into the separate app
2. Makes Electric function as a library so that a replication stack can be spun up by a application using Electric as a library
3. Renames parts of the supervision tree to be clearer & adds some code documentation


robacourt (approved):
That's awesome. Not only has it split out the multi-tenancy part, it looks neater and more maintainable now. Great work!

PR #1998, opened 2024-11-19
elixir-client: Revert change to way base_url is used

magnetised:
Originally the /v1/shape path was appended  onto the `base_url` path, but #1994 changed that to ignore any path elements in the base_url.

This change un-does that and reverts back to appending the /v1/shape path onto the base_url.

(no changeset included as the one in #1994 covers this)


msfstef (approved):
whoops!

Issue #1914, opened 2024-11-01
Look for tracecontext headers and add them to http traces


KyleAMathews:
Incoming http requests might be sent by a system using opentelemetry. These systems can set [tracecontext headers](https://www.w3.org/TR/trace-context/) — Electric's traces can then set these as parents. The idea being then that you can see the connection between external systems & Electric wherever all the traces are collected.

PR #2000, opened 2024-11-19
feat: use traceparent header from incoming shape requests as parent

KyleAMathews:
Fix https://github.com/electric-sql/electric/issues/1914

### DIFF
--- a/packages/sync-service/lib/electric/plug/trace_context_plug.ex
+++ b/packages/sync-service/lib/electric/plug/trace_context_plug.ex
@@ -1,0 +1,66 @@
+defmodule Electric.Plug.TraceContextPlug do
+  @moduledoc """
+  A plug that extracts trace context from incoming HTTP headers and sets it as the parent span.
+  """
+  @behaviour Plug
+
+  require Logger
+
+  def init(opts), do: opts
+
+  def call(conn, _opts) do
+    case extract_trace_context(conn) do
+      {:ok, span_ctx} ->
+        Logger.info("TraceContextPlug: Successfully extracted trace context")
+        :otel_tracer.set_current_span(span_ctx)
+        conn
+      :error ->
+        Logger.info("TraceContextPlug: No trace context found")
+        conn
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/plug/trace_context_plug.ex
```diff
@@ -0,0 +1,58 @@
...
+  # Note: The key is passed first, carrier second by the propagator
+  defp header_getter(key, carrier) when is_list(carrier) do
+    case Enum.find(carrier, fn {k, _v} -> String.downcase(k) == String.downcase(key) end) do
+      {_key, value} -> value
+      nil -> []
+    end
+  end
+
+  # Fallback clause for when carrier is not a list
+  defp header_getter(_key, _carrier), do: []
```
icehaunter, line 57: I believe this can be way simpler and not use `:otel_propagator_trace_context.extract` which is marked as private API and the link you provided advises against using it directly. If we use "proper" API, although it's really badly documented imho, it's down to 5 lines of code bar comments

```suggestion
  def call(%Plug.Conn{req_headers: headers} = conn, _opts) do
    # Extract function expects a list of headers as tuples and knows
    # the expected header keys, so we don't have to prefilter.
    ctx = :otel_propagator_text_map.extract_to(:otel_ctx.new(), headers)

    # Get the span context from the extracted context
    case :otel_tracer.current_span_ctx(ctx) do
      :undefined ->
        # No parent, continue as-is
        conn

      span_ctx ->
        # Parent found, set as current span
        :otel_tracer.set_current_span(span_ctx)
        conn
    end
  end
```

icehaunter (changes requested):

packages/sync-service/lib/electric/plug/trace_context_plug.ex
KyleAMathews: ah neat! Leme try it locally.

KyleAMathews:
@icehaunter still is working 🫡

icehaunter (approved):

PR #2002, opened 2024-11-20
chore: fixed website builds

icehaunter:
Previous matching was generating wrong blog structure (paths were split on arbitrary `-`), which led to "dead link" complaints. Also the `.split` of the path was wrong because preceding slash is not part of the file, but maybe that's incosistent between mac & linux so I replaced it with a definitely crossplatform call

### DIFF
--- a/website/.vitepress/config.mts
+++ b/website/.vitepress/config.mts
@@ -54,7 +54,8 @@ export default defineConfig({
     ]
   },
   rewrites: {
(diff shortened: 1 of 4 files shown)


thruflo:
I had already fixed this in https://github.com/electric-sql/electric/pull/1988 -- waiting on a review for a while now.

It was worth waiting for me to input.

PR #2009, opened 2024-11-20
feat (sync-service): Prevent shape consumer errors from affecting other shapes

### LINKED ISSUE #1925: Remove a shape if anything goes wrong with it
The supervision tree for the shape log processes is set to restart when there is an exception in one of the processes.

In #1911, one shape process crashed while trying to query Postgres (apparently), which lead entire supervision tree to restart in a loop, always failing at the same point. It's not clear the root cause of the issue, but Electric got stuck in a restart loop and with no external visibility about the occurrence.

In this task we're going to change the exception handling to remove the shape when it throws an error. The idea is that this might increase the chance of Electric to recover from bugs/transient issues related with the shape.  If later someone asks for the shape again either we do the same thing or maybe the issue is gone. 

These events should be published through Telemetry, so that developers can setup alerts to be informed about the occurrences.

robacourt:
Fixes #1925 . Errors that occur while consuming the replication stream that are to do with a specific shape, cause that shape's consumer to remove the shape and shut down, leaving the other shapes unaffected.

Errors can occur:
1. In the selector which happens on a process common to all shapes. 
2. On the consumer process.

This PR addresses both types of error. Previous to this PR, these errors would cause the sync service to get into an infinite crash loop and would stop responding to HTTP requests.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -116,7 +116,7 @@ defmodule Electric.Shapes.Consumer do
 
     # TODO: ensure cleanup occurs after snapshot is done/failed/interrupted to avoid
     # any race conditions and leftover data
-    state = cleanup(state)
+    cleanup(state)
     {:stop, :normal, :ok, state}
   end
 
@@ -160,7 +160,7 @@ defmodule Electric.Shapes.Consumer do
         )
 
     state = reply_to_snapshot_waiters({:error, error}, state)
-    state = cleanup(state)
+    cleanup(state)
     {:stop, :normal, state}
   end
 
@@ -183,7 +183,7 @@ defmodule Electric.Shapes.Consumer do
           "Error in Shapes.Consumer.handle_events: #{inspect(error)}\n#{Exception.format(:error, error)}"
         )
 
-        state = cleanup(state)
+        cleanup(state)
         {:stop, :normal, state}
     end
   end
@@ -207,7 +207,7 @@ defmodule Electric.Shapes.Consumer do
         state
       )
 
-    state = cleanup(state)
+    cleanup(state)
     {:stop, :normal, state}
   end
 
@@ -350,7 +350,6 @@ defmodule Electric.Shapes.Consumer do
     %{shape_status: {shape_status, shape_status_state}} = state
     shape_status.remove_shape(shape_status_state, state.shape_handle)
     ShapeCache.Storage.cleanup!(state.storage)
-    state
   end
 
   defp reply_to_snapshot_waiters(_reply, %{awaiting_snapshot_start: []} = state) do
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -174,9 +174,23 @@ defmodule Electric.Shapes.Consumer do
     reply_to_snapshot_waiters({:error, "Shape terminated before snapshot was ready"}, state)
   end
 
+  def handle_events(events, from, state) do
+    try do
+      do_handle_events(events, from, state)
+    rescue
+      error ->
+        Logger.error(
+          "Error in Shapes.Consumer.handle_events: #{inspect(error)}\n#{Exception.format(:error, error)}"
+        )
+
+        state = cleanup(state)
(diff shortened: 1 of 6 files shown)


packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -168,8 +177,19 @@ defmodule Electric.Shapes.Consumer do
...
+
+    if is_error?(reason) do
+      cleanup(state)
+    end
+
+    state
+  end
+
+  defp is_error?(reason) do
+    reason not in [:normal, :shutdown, {:shutdown, :truncate}, :killed]
```
icehaunter, line 192: why is `{:shutdown, :truncate}` special-cased here? I think we should just handler `{:shutdown, _}` as "normal" exit code as per OTP convention, no?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -168,8 +177,19 @@ defmodule Electric.Shapes.Consumer do
...
   end
 
-  def terminate(_reason, state) do
-    reply_to_snapshot_waiters({:error, "Shape terminated before snapshot was ready"}, state)
+  def terminate(reason, state) do
+    state =
+      reply_to_snapshot_waiters({:error, "Shape terminated before snapshot was ready"}, state)
+
+    if is_error?(reason) do
+      cleanup(state)
```
icehaunter, line 185: I wonder, is there a world where this call is slow? Calling this inside the `terminate` callback implies a time limit on this operation before the process is hard-killed, and that's if it's not hard-killed in the first place. 

Elixir docs say:

> [terminate/2](https://hexdocs.pm/elixir/1.16.1/GenServer.html#c:terminate/2) is useful for cleanup that requires access to the [GenServer](https://hexdocs.pm/elixir/1.16.1/GenServer.html#content)'s state. However, it is not guaranteed that [terminate/2](https://hexdocs.pm/elixir/1.16.1/GenServer.html#c:terminate/2) is called when a [GenServer](https://hexdocs.pm/elixir/1.16.1/GenServer.html#content) exits. Therefore, important cleanup should be done using process links and/or monitors. A monitoring process will receive the same exit reason that would be passed to [terminate/2](https://hexdocs.pm/elixir/1.16.1/GenServer.html#c:terminate/2).

so I guess my question is: is this good enough, or should we have a monitoring process that actually does the cleanup to guarantee it being done? With the outlook that it's possible that our storage may not be on local disk at some point

icehaunter (approved):
Good work, and good tests, but I did have 2 concerns

packages/sync-service/lib/electric/shapes/consumer.ex
robacourt: Good idea. Done.

packages/sync-service/lib/electric/shapes/consumer.ex
robacourt: It's a good point. I'm wondering if this is good enough for now, because:
1. The first and most important part of cleanup is deleting the shape_status which should be quick. With this deleted the other shapes can carry on happily.
2. The second part of cleanup is deleting the shape from storage, which may be slow, but if this doesn't happen it'll just use more disk space than it needs to, which seems fine especially as...
3. Cleanup on terminate will only occur if we have a bug

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -91,13 +91,22 @@ defmodule Electric.Shapes.Consumer do
      subscribe_to: [{producer, [max_demand: 1, selector: &selector(&1, config.shape)]}]}
   end
 
-  defp selector(%Transaction{changes: changes}, shape) do
+  defp selector(event, shape) do
+    process_event?(event, shape)
+  rescue
+    # Swallow errors here to avoid crashing the ShapeLogCollector.
+    # Return `true` so the event is processed, which will then error
+    # for the same reason and cleanup the shape.
```
alco, line 99: Could you elaborate on how this part happens? Does "error for the same reason" assume that similar code is executed when the event is processed by the consumer as in this `process_event?()` function? If that's the case, can we make it a single private function that is called in both places? Or at least explain it in a code comment to avoid the confusion I've just found myself in?

PR #2011, opened 2024-11-20
feat: enable error handler to recover from failed HTTP requests

kevin-dp:
This PR modifies the `onError` handler that is supported by the shape stream such that it can recover from certain HTTP request errors. For example, if the HTTP request fails with a 401 because of an expired token, the handler could fetch a new token, and try again by returning the modified HTTP headers. The shapestream will then retry.

- [x] Should we make the onError handler async such that it can for instance fetch the new token asynchronously?

**Accompanying docs PR: https://github.com/electric-sql/electric/pull/2018**

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -66,7 +66,13 @@ type ReservedParamKeys =
 
 type ParamsRecord = Omit<Record<string, string>, ReservedParamKeys>
 
-type ShapeStreamErrorHandler = (error: Error) => void
+type RetryOpts = {
+  params?: ParamsRecord
+  headers?: Record<string, string>
+}
+type ShapeStreamErrorHandler = (
+  error: Error
+) => void | RetryOpts | Promise<void | RetryOpts>
 
 /**
  * Options for constructing a ShapeStream.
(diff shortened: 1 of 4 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -66,7 +66,13 @@ type ReservedParamKeys =
...
 type ParamsRecord = Omit<Record<string, string>, ReservedParamKeys>
 
-type ShapeStreamErrorHandler = (error: Error) => void
+type RetryOpts = {
+  params?: ParamsRecord
+  headers?: Record<string, string>
+}
+type ShapeStreamErrorHandler = (
+  error: Error
+) => void | Promise<void> | RetryOpts | Promise<RetryOpts>
```
msfstef, line 75:
```suggestion
) => void | RetryOpts | Promise<void | RetryOpts>
```
nit, I assume it would accept a function returning e.g. `Promise<void>` - if not ignore me

packages/typescript-client/src/client.ts
```diff
@@ -438,12 +444,26 @@ export class ShapeStream<T extends Row<unknown> = Row>
     } catch (err) {
       this.#error = err
       if (this.#onError) {
-        this.#onError(err as Error)
+        const retryOpts = await this.#onError(err as Error)
+        if (typeof retryOpts === `object`) {
+          this.#reset()
```
msfstef, line 449: I think as part of this it would be good to perhaps extract the fetch client construction into a private utility, and `this.#reset` also reassigns the fetch client - it should work without doing that but since some of the layers have internal state (like the prefetching) it would be good for the reset call to discard that as well I think and create a new fetch client

msfstef (approved):
Nice!

packages/typescript-client/src/client.ts
kevin-dp: I agree that would be good to do. The only problem i see is that we are not in control of the optional fetch client that users can pass. So we can get into weird situations where we reset the state of our own internal fetch client wrappers but not the user-provided fetch client.

Are we fine with that? Or do we want to modify the signature such that the user provides a *function* that constructs a fresh fetch client? That way, we can reset also the user's fetch client.

kevin-dp: After discussing this in DM we came to the following conclusion and thus this is not needed:

> if a request comes in "out of order" then the cache (in our fetch client) is invalidated and started again. So upon restarting the existing cache will be thrown out anyway so no need to reset it.

PR #2014, opened 2024-11-21
feat: allow http1 protocol on Electric client

icehaunter:
This allows electric client to be used in non-https environments, such as local development, without additional configuration


magnetised (approved):
Yeah. I agonized about that http2 only option, seems that it was a bad idea.

PR #2015, opened 2024-11-21
feat: accept URI structs as endpoint/base_url options

icehaunter:


magnetised (changes requested):
It works, but can we add a test to prove that it works pls. There's a suite of `new/1` tests in client_test.exs

magnetised (approved):
Souper

Issue #1996, opened 2024-11-19
Remove `ShapeStream.subscribeOnceToUpToDate()`


KyleAMathews:
It's not a concern of ShapeStream so we can simplify our API surface.

PR #2017, opened 2024-11-21
feat! (ts client): Remove subscribeOnceToUpToDate from shape stream

kevin-dp:
Fixes https://github.com/electric-sql/electric/issues/1996.

We remove `subscribeOnceToUpToDate` in favor of subscribing to the stream and "waiting" for the first `up-to-date` control message. **This is a breaking change.**


kevin-dp:
> Needs a [BREAKING] changeset!

Good catch, i created it but forgot to push it

PR #2018, opened 2024-11-21
docs: document `onError` option

kevin-dp:
This PR documents the `onError` option added to the `ShapeStream` options in https://github.com/electric-sql/electric/pull/2011


thruflo:
We should update at least the gatekeeper-auth client code: https://github.com/electric-sql/electric/blob/main/examples/gatekeeper-auth/client/index.ts

Are there any other places we have example code using this client that should be updated too?

KyleAMathews (approved):
Looks great!

KyleAMathews:
@thruflo I'm not sure if most people will care to explicitly handle errors — generally just letting them bubble up is fine as people will see them in their console during dev & error trackers will see them in prod.

thruflo:
> I'm not sure if most people will care to explicitly handle errors

Fair enough, we don't need to highlight this everywhere.

I think the auth token renewal is a common case where (unless we do something else like auto call a function to retry with new headers) it is useful to handle the fetch error. My point about the gatekeeper client example is that's exactly what that code is trying to illustrate -- how to handle token renewal.

KyleAMathews:
@thruflo oh for sure — I just meant elsewhere — this addition was entirely to handle that use case 😆

Issue #1922, opened 2024-11-04
Don't allow Electric to create shapes if it's not connected to a replication slot


KyleAMathews:
Just saw this — my Electric wasn't connecting as it was conflicting with a stale publication slot entry — once I cleared it, Electric did connect but the shapes that were created before now had stale info as they'd missed events.

PR #2019, opened 2024-11-21
feat: Add global stack event registry and block requests before ready

msfstef:
PR by @icehaunter and me - makes the `StackSupervisor` accept a stack event registry that it uses to dispatch status events about the state of the stack.

This was preliminary work for multitenancy, and also fixes https://github.com/electric-sql/electric/issues/1922 since now we hold connections when the stack is not ready, and release them when we receive a "ready event" or time them out with a 503 - avoids crashing the ETS inspector which was trying to use a DB connection from an uninitialised pool.


packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
```diff
@@ -688,4 +698,41 @@ defmodule Electric.Plug.ServeShapePlugTest do
...
+  describe "stack not ready" do
+    setup :with_stack_id_from_test
+
+    test "returns 503", ctx do
+      conn =
+        ctx
+        |> conn(:get, %{"table" => "public.users"}, "?offset=-1&replica=full")
+        |> call_serve_shape_plug(ctx)
+
+      assert conn.status == 503
```
kevin-dp, line 711: How come that this returns a 503? 
From what i can tell the serve shape plug holds the connection until the stack is ready or until the timeout. The timeout is set to 100ms in this test. Is it just that the stack takes longer than 100ms to set up? That would be very brittle as running it on a faster machine this test may fail. Therefore, i'm probably missing something and there must be another reason why this is guaranteed to return a 503.

kevin-dp (approved):
Nice work, nothing to remark here 🥇

packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
msfstef: For the "stack to be ready" in this test we need to manually dispatch an event (see "waits until stack ready and proceeds" test below) - so this won't flake as the stack will never be ready, and the 100ms is just an arbitrary short timeout

kevin-dp: I see. Normally, the supervisor sends that message but here the supervisor has actually been overridden in the process registry (L86) and therefore will never send that message and that's why in these tests they will always timeout unless we manually send that message. Correct?

msfstef: These are plug tests so more like unit tests, the rest of the stack is all mocked out and we use Mox for expectations etc - normally the stack ready message would come from one part of the supervision tree but we don't really have one in here.

But other than that yeah, ultimately because the stack doesn't exist, the plug is just waiting to receive a message from the specified registry which we have to manually dispatch in the absence of the rest of the stack

robacourt (approved):

Issue #1997, opened 2024-11-19
Move `stream.subscribe()` to be part of the constructor


KyleAMathews:
e.g. `new ShapeStream({ subscribe: () => {} })`

The problem with `.subscribe()` is then people get the sense that they can subscribe whenever and they'll get the full stream where if you delay calling .subscribe then you'll miss the early messages. So to avoid this footgun and in general, keep construction of streams fixed & declarative, let's move subscriptions to the constructor.

msfstef:
I think we want this to be a declarative list of subscribers, soo something like:

```ts
type ShapeStreamSubscriber = {
  onMessages: (messages: Message[]) => void | Promise<void>,
  onError: (err: unknown) => void
}

subscribers: ShapeStreamSubscriber[] | ShapeStreamSubscriber
```

thruflo:
Is there not a slight chicken and egg thing of needing to instantiate a Shape to have the subscribers to instantiate the ShapeStream which is needed to instantiate the Shape ...?

samwillis:
The alternative is an explicit .start() method on a shape stream. It solves two issues:

- allowing multiple or lazy subscribing to a stream (passing it to a Shape)
- catching exceptions during the initial connection

.start() would return a promise that resolves/fails depending on initial connection - such as auth failing. This is imposible to do with the sync constructor (the alternative is an async static .create() but that doesn't solve lazy subscription)

After calling .start() we can lock the subscriptions and throw on any further .subscribe() calls. (Although there is an argument for late subscription for power users and edge cases...)

msfstef:
> Is there not a slight chicken and egg thing of needing to instantiate a Shape to have the subscribers to instantiate the ShapeStream which is needed to instantiate the Shape ...?

This is part of the problem in how we currently use `ShapeStream` - as soon as we instantiate it it starts streaming, which means that if you "feed it" to a `Shape` asynchronously it will have missed messages and have inconsistent state. Passing an ongoing stream to a shape to materialize it is odd, since the materialization requires the shape to have access to the stream from the start (and that is not guaranteed when being passed an instantiated stream).

The way to do it with the declarative approach would be that `Shape` actually accepts `Omit<ShapeStreamOptions, 'subscribers'>` rather than a `ShapeStream` instance, and it internally creates a shape stream with itself as a subscriber.

There is also another way, using a sort of `ShapeStreamBuilder`, see below:

> The alternative is an explicit .start() method on a shape stream.

[...]

PR #2028, opened 2024-11-21
examples: update gatekeeper auth client to use `onError`.

thruflo:
This is what I requested in https://github.com/electric-sql/electric/pull/2018#issuecomment-2491085763

The client now uses the new onError to handle 401/403 and reconnect with a new auth token.

One observation: with the onError being defined before the shape is instantiated, I'm not sure what the best control flow is to handle a `405` must refetch. Certainly it's hard to handle in the onError callback. I guess this is what's being discussed in https://github.com/electric-sql/electric/issues/1997 etc.

### DIFF
--- a/examples/gatekeeper-auth/client/index.ts
+++ b/examples/gatekeeper-auth/client/index.ts
@@ -1,54 +1,42 @@
-import { Shape, ShapeStream } from '@electric-sql/client'
+import { FetchError, Shape, ShapeStream } from '@electric-sql/client'
 
 const API_URL = process.env.API_URL || "http://localhost:4000"
 
-interface Definition {
-  table: string,
-  where?: string,
-  columns?: string
-}
-
 /*
- * Fetch the shape options and start syncing. When new data is recieved,
- * log the number of rows. When an auth token expires, reconnect.
+ * Makes a request to the gatekeeper endpoint to fetch a config object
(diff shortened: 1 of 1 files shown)


examples/gatekeeper-auth/client/index.ts
```diff
@@ -1,54 +1,42 @@
...
-      const status = shape.error.status
-      console.warn('fetch error: ', status)
+  const resp = await fetch(url, {method: "POST"})
+  return await resp.json()
+}
 
+// Stream the shape through the proxy, using the url and auth headers
+// provided by the gatekeeper.
+const config = await fetchConfig()
+const stream = new ShapeStream({...config, onError: async (error) => {
```
KyleAMathews, line 20: prettify?

KyleAMathews:
409 must-fetch responses never would reach `onError` — https://github.com/electric-sql/electric/blob/e9813969dccb679ab763773205f59a3fe7166bdc/packages/typescript-client/src/client.ts#L371-L377

thruflo:
So the stream handles it and the shape gets the instruction in a control message?

Makes sense, cool.

KyleAMathews:
> So the stream handles it and the shape gets the instruction in a control message?

yeah exactly — the shape just resets its materialized view and recreates it — subscribers would never notice

PR #2039, opened 2024-11-25
docs: add unexpected 409 section to troubleshooting guide.

thruflo:


website/docs/guides/troubleshooting.md
```diff
@@ -65,6 +65,31 @@ You can then bring a fresh backend up from scratch:
...
+conflict reading Location
+
+url: http://localhost:3000/v1/shape?table=projects&offset=0_0&handle=51930383-1732543076951
+sec: 0.003023737 seconds
+status: 200
+```
+
+This indicates that your client library or proxy layer is caching requests to Electric and responding to them without actually hitting Electric for the correct response.
+
+##### Solution &mdash; clear your cache
```
KyleAMathews, line 89: also worth noting that this problem will resolve itself as client/proxy caches empty — it's not actually breaking anything. The main downside is an extra round-trip to the server to get told to start over.

KyleAMathews: Also could link to https://electric-sql.com/docs/api/http#control-messages & the must-refetch bit

paulharter (approved):
yup

PR #2042, opened 2024-11-26
fix(elixir-client): Should terminate on 400s

msfstef:
Ignore the branch name - 400s are terminal errors indicating a malformed request and should not trigger restarts.


magnetised (approved):
Super

Issue #2031, opened 2024-11-22
Track replication lag in otel metrics and log warning when it gets too high


KyleAMathews:
We really need visibility into this.

balegas:
Why not do both? I think it is useful to know the amount of bytes that are pending in the WAL and what is the latency for electric to write a transaction into the log

KyleAMathews:
Yeah both sound great. I didn't know we could get the actual time diff so didn't mention it but that'd be great to have as well.

Yeah warnings from your otel collector is a lot more flexible — people can already set warnings from postgres data directly. We can potentially add warnings directly in the Electric logs as well but a good sequencing of work is to first gather the data and then later decide exactly how to communicate it.

PR #2043, opened 2024-11-26
chore (sync service): report replication lag in bytes

kevin-dp:
Part of https://github.com/electric-sql/electric/issues/2031.
Exports the replication lag in bytes as a metric to Prometheus.


icehaunter (approved):

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -423,4 +425,21 @@ defmodule Electric.Shapes.Consumer do
...
+
+  defp report_replication_lag(%Transaction{commit_timestamp: commit_timestamp}) do
+    # Compute time elapsed since commit
+    # since we are comparing PG's clock with our own
+    # there may be a slight skew so we make sure not to report negative lag.
+    # Since the lag is only useful when it becomes significant, a slight skew doesn't matter.
+    now = DateTime.utc_now()
+    lag = Kernel.max(0, DateTime.diff(now, commit_timestamp, :millisecond))
+
+    OpenTelemetry.with_span(
```
KyleAMathews, line 437: should this just be an attribute on the handle transaction span?

Issue #2041, opened 2024-11-26
Order params in the query strings for the clients we maintain


balegas:
Cloudflare (and potentially other CDNs) do not threat query string independently of the param order, which will result in cache misses if different clients are building the query strings with the params in different order.

We should ensure that our client set URL query parameters in a consistent order, e.g. by ordering them.

- [ ] add a note about this in our dcos

msfstef:
Opened https://github.com/electric-sql/electric/pull/2044 to address this - not sure what the note in the docs should be and where?

PR #2044, opened 2024-11-26
feat(clients): Always order query params

msfstef:
Closes https://github.com/electric-sql/electric/issues/2041

To be fair, by virtue of the clients having a fixed process by which the parameters are appended and updated they were already stable, but an explicit ordering step ensures that changes in the order of those operations don't break caching between versions, and is generally more explicit.

### DIFF
--- a/.changeset/selfish-rings-mate.md
+++ b/.changeset/selfish-rings-mate.md
@@ -1,0 +1,6 @@
+---
+"@electric-sql/client": patch
+"@core/elixir-client": patch
+---
+
+Always use sorted query parameters in official clients for stable URLs.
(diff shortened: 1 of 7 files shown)


.changeset/selfish-rings-mate.md
```diff
@@ -0,0 +1,6 @@
+---
+"@electric-sql/client": patch
+"@core/elixir-client": patch
+---
+
+Always use sorted query parameters in official clients for stable URLs.
```
KyleAMathews, line 6:
```suggestion
Always use sorted query parameters in official clients to ensure Shape URLs are cached consistently.
```

kevin-dp (approved):
LGTM!
Left a few minor improvements.

KyleAMathews (approved):
:shipit:

PR #2045, opened 2024-11-26
examples: always start clean and update running instructions.

thruflo:
I've been working with the examples quite a lot and I've commonly experienced stale database and disk state from running `docker compose up` in the foreground followed by `ctl+c` to stop and also from forgetting to run `pnpm backend:down`.

So ... as a systemic solution, I've updated the `example-backend:up` script in the root package.json to always run `example-backend:down` first. This means that examples always start with a clean slate and no weird / stuck table and or disk state.

I also then did a pass through the example READMEs, explaining this and re-formatting the instructions a touch.

### DIFF
--- a/examples/basic-example/README.md
+++ b/examples/basic-example/README.md
@@ -1,22 +1,46 @@
 # Basic example
 
+This is an example of a basic ElectricSQL app using React. The Electric-specific code is in [`./src/Example.tsx`](./src/Example.tsx).
+
 ## Setup
 
-1. Make sure you've installed all dependencies for the monorepo and built packages
+This example is part of the [ElectricSQL monorepo](../..) and is designed to be built and run as part of the [pnpm workspace](https://pnpm.io/workspaces) defined in [`../../pnpm-workspace.yaml`](../../pnpm-workspace.yaml).
 
-From the root directory:
+Navigate to the root directory of the monorepo, e.g.:
 
-- `pnpm i`
-- `pnpm run -r build`
+```shell
+cd ../../
+```
 
-2. Start the docker containers
+Install and build all of the workspace packages and examples:
 
-`pnpm run backend:up`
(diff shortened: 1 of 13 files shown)


thruflo:
The problem occurs when re-running the same example as much as across examples. I’d rather have the example work without weird stale data than support persistent data across restarts. Bear in mind it does print what’s happening to stdout.

Is there a solution that can support the best of both worlds?

alco:
@thruflo 

There shouldn't be any problems stopping and restarting an example. The database remains as a source of truth and the shape log should eventually get synced up to it once the replication connection is reestablished and catches up to the latest LSN.

The stale data problems you've experienced are likely due to the fact that our compose file in `.support/docker-compose.yml` doesn't define a persistent volume for the sync service.

Can you provide a quick series of steps I could try to reproduce the problem with any of the examples? I could then try fixing it by updating our compose config.

thruflo:
Ok, brilliant, that would be great 👍 

Minimal repro:

1. pnpm backend:up in one example -> forget to bring down -> run in another example folder
2. somehow backed:up or manually docker compose up and ctrl+c to close foregrounded or somehow stop the containers another way (using docker CLI or dashboard) -> restart and replication no longer works

I am also slightly reporting behaviour I’ve seen with the gatekeeper-auth docker compose. This is different from the standard examples one (for reasons that will be clear from the README). Testing stuff manually in local dev, I kept coming up the situation where I had a psql session that I would reconnect after ctlr+c and docker compose up and would write to PG without anything coming through electric.

So that’s another way to reproduce weirdness that would be good to fix across all compose workflows.

alco:
@thruflo I've cracked my brain against this but haven't come up with a better working solution. So please go ahead and merge this. Nice work!

alco (approved):

PR #2047, opened 2024-11-26
chore (sync service): add replication lag to current span

kevin-dp:
This PR addresses [Kyle's comment](https://github.com/electric-sql/electric/pull/2043#pullrequestreview-2461985569) and adds the replication lag to the current span instead of creating a new span.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -434,12 +434,6 @@ defmodule Electric.Shapes.Consumer do
     now = DateTime.utc_now()
     lag = Kernel.max(0, DateTime.diff(now, commit_timestamp, :millisecond))
 
-    OpenTelemetry.with_span(
-      "shape_write.consumer.do_handle_txn.report_replication_lag",
-      [lag: lag],
-      fn ->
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -434,12 +434,6 @@ defmodule Electric.Shapes.Consumer do
...
     lag = Kernel.max(0, DateTime.diff(now, commit_timestamp, :millisecond))
 
-    OpenTelemetry.with_span(
-      "shape_write.consumer.do_handle_txn.report_replication_lag",
-      [lag: lag],
-      fn ->
-        lag
-      end
-    )
+    OpenTelemetry.add_span_attributes(lag: lag)
```
KyleAMathews, line 437:
```suggestion
    OpenTelemetry.add_span_attributes(replication_lag: lag)
```

Attributes should be semantically self-contained as it gets confusing if we use `lag` somewhere else as then both would show up when you're searching in honeycomb.

KyleAMathews (changes requested):

KyleAMathews (approved):
:shipit:

PR #2049, opened 2024-11-26
fix: Log error without falling over for failed recovery

msfstef:
We have seen some issues related to https://github.com/lucaong/cubdb/issues/71 come up during recovery, and it currently causes issues with the whole electric.

Initially my thought was to delete the shape entirely if it fails, but perhaps as an initial workaround we might just want to log the error and continue recovery without the problematic shape?

### DIFF
--- a/.changeset/warm-brooms-approve.md
+++ b/.changeset/warm-brooms-approve.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Log error but don't fall over if failing to initialise recovered shape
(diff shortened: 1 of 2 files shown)


KyleAMathews:
If we don't "recover" the shape — what happens if someone requests it?

msfstef:
@KyleAMathews as far as electric is concerned, the shape does not exist - it just leaves the data on disk rather than cleaning it up (and thus will try to recover it on a subsequent deploy again)

So effectively it's abandoning a shape but leaving it on disk, either for manual cleanup or if we figure out the issue and it's not actually corrupted data it can be restored.

robacourt (approved):

KyleAMathews:
@msfstef cool then leaving it seems fine — presumably we'll either figure a fix or just move off cubdb making the problem moot).

PR #2051, opened 2024-11-26
fix: Clean up shape folders and remove unrecoverable shapes

msfstef:
Followup to https://github.com/electric-sql/electric/pull/2049

We are currently leaving folders behind when deleting shapes, which do take up a little bit of space and eventually could be a bit of an issue. This PR takes care of removing everything.

Additionally, when a shape fails to recover by not being able to start the consumer at all (see CubDB issue), the shape data is removed using a new method called `unsafe_cleanup!` which simply `rm -rf`s the shape directory. Not great to use while the shape is active as it can cause various bugs, but great for cleaning up.

### DIFF
--- a/.changeset/hungry-apples-sell.md
+++ b/.changeset/hungry-apples-sell.md
@@ -1,0 +1,6 @@
+---
+"@core/sync-service": patch
+---
+
+Clean up directories when removing shapes.
+Remove corrupted shapes from store when recovery fails.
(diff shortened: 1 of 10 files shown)


robacourt (approved):
Lovely

magnetised (approved):
Good stuff, though disconcerting lack of tests..

msfstef:
@magnetised it wouldn't be called `unsafe` if it was tested 🤔  (added test https://github.com/electric-sql/electric/pull/2051/commits/313729d9816ba1c1e43e20266eb154ab8b712a1f)

PR #2054, opened 2024-11-27
chore (sync service): probablistic sampling of transaction spans

kevin-dp:
Part of https://github.com/electric-sql/electric/issues/2032.

We noticed that most spans are descendants of the `pg_txn.replication_client.process_x_log_data` root span.
Therefore, we decided to only sample a portion of those spans.

This PR introduces a custom sampler that works as follows:
- Samples a configurable ratio of `pg_txn.replication_client.process_x_log_data` root spans
- Samples all other root spans
- Child spans are sampled if their parent is sampled

### Problem

We would like to sample all errors.
To do this we need to make the sampling decision at the end of the span when we have all attributes and events because errors are recorded using an "exception" event, this is known as tail sampling. However, the Erlang opentelemetry library only seems to support head sampling:
> Sampling is performed at span creation time by the Sampler configured on the Tracer

cf. https://docs.honeycomb.io/manage-data-volume/sample/techniques/ if you're not familiar with head vs tail sampling.

EDIT: solving this problem may require using HoneyComb's "Refinery" mechanism. So Electric would sample all traces and we would setup Refinery with custom tail sampling logic. However, this requires extra infrastructure to set up.

### DIFF
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -86,7 +86,18 @@ otel_sampling_ratio = env!("ELECTRIC_OTEL_SAMPLING_RATIO", :float, 0.01)
 
 config :opentelemetry,
   processors: [otel_batch_processor, otel_simple_processor] |> Enum.reject(&is_nil/1),
-  sampler: {Electric.Telemetry.Sampler, %{ratio: otel_sampling_ratio}}
+  # sampler: {Electric.Telemetry.Sampler, %{ratio: otel_sampling_ratio}}
+  # Sample root spans based on our custom sampler
+  # and inherit sampling decision from remote parents
+  sampler:
+    {:parent_based,
+     %{
+       root: {Electric.Telemetry.Sampler, %{ratio: otel_sampling_ratio}},
+       remote_parent_sampled: :always_on,
+       remote_parent_not_sampled: :always_off,
+       local_parent_sampled: :always_on,
+       local_parent_not_sampled: :always_off
+     }}
 
 database_url = env!("DATABASE_URL", :string!)
 
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -82,8 +82,11 @@ otel_simple_processor =
     {:otel_simple_processor, %{exporter: {:otel_exporter_stdout, []}}}
   end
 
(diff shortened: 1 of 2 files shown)


balegas:
Too bad that erlang otel doesn't support it yet. Do they have plans to support it?

kevin-dp:
Comments from a discussion with @alco:

> I'm afraid we cannot add Refinery as another runtime component because then we'd be forcing its setup and maintenance on our customers. Sampling is performed in :otel_span_utils:start_span() in the opentelemetry-erlang library. There's also the end_span() function to go with it. It will take a bit of hacking but I think we can make tail sampling work there. There's a tailsampling process for the OT Collector. So, for now at least, we could require the use of an OT collector in case one wants to have tail sampling. But out of the box, if you're running Electric without an OT collector, you'd only get head sampling.

> Yeah, I believe we should be able to do the following:
write a sampler that returns the "RECORD_ONLY" constant for samples that should be "dropped"
modify our record_exception() function such that it overwrites trace_flags on the current span's context and marks it as sampled regardless of the sampling decision made when the span was started This is based on the observation that opentelemetry-erlang only checks trace_flags when a span is ended to decide whether it needs to be exported or dropped. In addition to the above, we'll likely need to also walk up the span hierarchy to update trace_flags on parent spans until we reach the root. In any case, my point is that we'd better avoid this work now if it's not absolutely critical. In other words, require the use of an OT collector in the middle between Electric and Honeycomb if we want advanced sampling that always exports errors.

alco:
To add to the PR description: seeing lots of spans rooted in `pg_txn.replication_client.process_x_log_data` means the underlying database is experiencing a high write rate. It's perfectly fine to sample those spans since we don't have any error recording logic in them.

We only record errors for the `shape_get.plug.serve_snapshot` and `shape_get.plug.serve_shape_log` spans right now. We can use conditional sampling to only sample the spans that cover the replication stream and exclude shape serving.

kevin-dp:
> To add to the PR description: seeing lots of spans rooted in `pg_txn.replication_client.process_x_log_data` means the underlying database is experiencing a high write rate. It's perfectly fine to sample those spans since we don't have any error recording logic in them.
> 
> We only record errors for the `shape_get.plug.serve_snapshot` and `shape_get.plug.serve_shape_log` spans right now. We can use conditional sampling to only sample the spans that cover the replication stream and exclude shape serving.

I'm going to merge this PR since this is already what this PR does 👍

alco (approved):

PR #2055, opened 2024-11-27
fix (example): fix database ID option in linearlite example

kevin-dp:


msfstef (approved):

examples/linearlite/src/shapes.ts
```diff
@@ -4,8 +4,8 @@ import { baseUrl, databaseId, token } from './electric'
 export const issueShape: ShapeStreamOptions = {
   url: `${baseUrl}/v1/shape/`,
   table: `issue`,
-  databaseId,
   params: {
     token,
+    databaseId,
```
KyleAMathews, line 9:
```suggestion
    database_id: databaseId,
```

PR #2056, opened 2024-11-27
Rename "replication lag" to "retained WAL size"

alco:
The thing being measured here is the size of the WAL that Electric's replication slot retains, so I think the new name describes it better.

This change also restructures the code a bit to make the dependency on the DB connection pool crystal clear.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -146,8 +146,8 @@ defmodule Electric.Connection.Manager do
     GenServer.cast(server, {:pg_info_looked_up, pg_info})
   end
 
-  def query_replication_lag(server) do
-    GenServer.call(server, :query_replication_lag)
+  def report_retained_wal_size(server) do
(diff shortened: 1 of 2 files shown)


examples/linearlite/src/shapes.ts
```diff
@@ -4,8 +4,8 @@ import { baseUrl, databaseId, token } from './electric'
 export const issueShape: ShapeStreamOptions = {
   url: `${baseUrl}/v1/shape/`,
   table: `issue`,
-  databaseId,
   params: {
+    databaseId,
```
kevin-dp, line 8: I have an open PR for fixing this: https://github.com/electric-sql/electric/pull/2055
It should actually be `database_id: databaseId`.
Let's not introduce it here and merge my PR instead. I did not yet merge it because deployment failed yesterday, but it's unrelated to the PR itself so we can merge now.

alco: I saw your PR and have since removed this change from mine.

robacourt (approved):

PR #2057, opened 2024-11-27
feat: add more metrics and expose them as usage telemetry

icehaunter:


robacourt (approved):
Nice! LGTM

website/docs/reference/telemetry.md
```diff
@@ -7,7 +7,7 @@ outline: deep
 
 # Telemetry
 
-Electric provides telemetry data — such as traces, logs, and metrics — for real-time system monitoring. 
+Electric provides telemetry data — such as traces, logs, and metrics — for real-time system monitoring. Self-hosted Electric instances also send anonymous usage data to electric-sql.com to help us understand how the software is being used - see [Anonymous Usage Data](#anonymous-usage-data) for more details.
```
thruflo, line 10: Suggest:

```md
Self-hosted Electric instances are also configured by default to also send aggregated, anonymous usage data to ElectricSQL to help us understand how our software is being used. You can opt-out of this reporting by setting an environment variable. See the [Anonymous usage data](#anonymous-usage-data) section below for more details.
```

website/docs/reference/telemetry.md
```diff
@@ -56,3 +56,9 @@ Set `ELECTRIC_HNY_DATASET` and `ELECTRIC_HNY_API_KEY` environment variables in a
 ```shell
 docker compose -f docker-compose-otel.yml up
 ```
+
+## Anonymous Usage Data
```
thruflo, line 60: Capitalisation: `Anonymous usage data`

website/docs/reference/telemetry.md
```diff
@@ -56,3 +56,9 @@ Set `ELECTRIC_HNY_DATASET` and `ELECTRIC_HNY_API_KEY` environment variables in a
...
+
+Electric instances are configured to send anonymized usage data to checkpoint.electric-sql.com to help us understand how the software is being used. This data includes the disk usage by the shape cache, CPU/memory information about the running Electric instance, Postgres version, number of shapes, amount of distinct shape requests, and numerical information about processed transactions: byte size and amount of operations. No information about transaction content is sent. Aggregated statistics are sent every 30 minutes.
```
thruflo, line 62: ... configured by default ...

thruflo: Suggest:

```md
... Absolutely no information about transaction contents is sent. I.e.: absolutely none of your data that you're using Electric to replicate is captured in the telemetry information or shared with the Electric checkpoint service.

Aggregated statistics are sent every 30 minutes.
```

website/docs/reference/telemetry.md
```diff
@@ -56,3 +56,9 @@ Set `ELECTRIC_HNY_DATASET` and `ELECTRIC_HNY_API_KEY` environment variables in a
...
+
+To disable anonymous usage data, set the `ELECTRIC_CALL_HOME_TELEMETRY` environment variable to `false`. We encourage everyone to keep this enabled so we can get a better understanding of how Electric is used.
```
thruflo, line 64: I assume we need a prefix to the env var because we have other types of telemetry too? If not then maybe `ELECTRIC_TELEMETRY_ENABLED`, if so, perhaps `ELECTRIC_USAGE_TELEMETRY`?

I believe we previously had a string value in case we wanted to introduce verbosity levels in future. Boolean is fine if simpler for now.

We should add this env var config to here https://electric-sql.com/docs/api/config#telemetry and perhaps we could also make the checkpoint service url configurable by env var? Looking at that section, it's also fairly confusing what the different types of telemetry are and do. And the OTEL is enabled by providing a URL, rather than an enable flag.

Perhaps we can tighten this up? For example, should we just have this new type of usage reporting as the "telemetry" and then somehow re-classify the other more debugging stuff under "debugging"? My goal is to be very clear to a user about what data is being sent home. Looking at usage, OTEL, HNY, PROMETHEUS and STATSD kind of freaks me out a bit.

icehaunter: Sending home is one thing, but honeycomb and prometheus aren't necessarily ours, or even debuggy, it may just be part of the observability setup that the customer has. Let's think about it, but i think keeping usage report telemetry explicitly separate is good for visibility into what's sent where

icehaunter: We can remove statsd from docs, it's used only for benchmarking Stat collection

icehaunter: I've changed the variable name to `ELECTRIC_USAGE_REPORTING`, and added it to config docs under a separate header "Usage reporting", keeping the "Telemetry" header intact. Can you take another look?

thruflo (approved):
Docs changes look great to me, thanks 👍

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -115,18 +115,6 @@ defmodule Electric.Plug.Utils do
     |> Map.merge(Map.new(conn.resp_headers, fn {k, v} -> {"http.response.header.#{k}", v} end))
   end
 
-  defp client_ip(%Plug.Conn{remote_ip: remote_ip} = conn) do
```
alco, line 118: What's the reason for removing this?

PR #2058, opened 2024-11-27
chore (sync service): remove obsolete application config

kevin-dp:
This PR removes the application configuration which is a remnant from before and is no longer used inside core Electric.


.changeset/brown-lies-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Remove unused app config.
```
robacourt, line 5: This is not customer affecting in any way so I wouldn't include a changeset

robacourt (approved):
🧹

PR #2061, opened 2024-11-28
Include cache headers in 304 responses

### LINKED ISSUE #2059: Extend max-age in cache-control headers for 304 responses
Small fix from discussion in https://github.com/electric-sql/electric/issues/2050 & specifically https://github.com/electric-sql/electric/issues/2050#issuecomment-2501592766

304 responses currently respond with `cache-control: max-age=0, private, must-revalidate` — this means that the browser will then have to keep re-validating the resources on every load of the app — but if the shape log request hasn't changed then it's probably still good for as long as the original max-age.

So for `offset=-1` we can set a max-age of 1 week and for other offsets set the default `max_age` from the config.

alco:
Fixes #2059

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -166,9 +166,9 @@ defmodule Electric.Plug.ServeShapePlug do
   plug :listen_for_new_changes
   plug :determine_log_chunk_offset
   plug :determine_up_to_date
+  plug :put_resp_cache_headers
   plug :generate_etag
   plug :validate_and_put_etag
-  plug :put_resp_cache_headers
   plug :serve_log_or_snapshot
 
   # end_telemetry_span needs to always be the last plug here.
@@ -351,7 +351,7 @@ defmodule Electric.Plug.ServeShapePlug do
       get_req_header(conn, "if-none-match")
       |> Enum.flat_map(&String.split(&1, ","))
       |> Enum.map(&String.trim/1)
-      |> Enum.map(&String.trim(&1, ~S|"|))
+      |> Enum.map(&String.trim(&1, <<?">>))
 
     cond do
       conn.assigns.etag in if_none_match ->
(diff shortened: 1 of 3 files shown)


robacourt (approved):

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -351,7 +351,7 @@ defmodule Electric.Plug.ServeShapePlug do
       get_req_header(conn, "if-none-match")
       |> Enum.flat_map(&String.split(&1, ","))
       |> Enum.map(&String.trim/1)
-      |> Enum.map(&String.trim(&1, ~S|"|))
+      |> Enum.map(&String.trim(&1, <<?">>))
```
KyleAMathews, line 354: What's this do?

KyleAMathews (approved):
Nice! So it was just a bug from the ordering of the imports?

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
alco: It's doing the same thing as before but is easier to read to my Elixir eyes.

`<<...>>` is a container syntax for binaries in Elixir. By default, every element is treated as a byte:
```
<<80, 79, 80>> == "POP"
```

The `?...` syntax is a parser feature the replaces itself with the ASCII code of the character, so:
```
?P == 80
?O == 79
?" == 34
```

The `<<...>>` syntax is used almost every time when you need to parse a string while doing different things for different classes of characters. See https://github.com/electric-sql/electric/blob/main/packages/sync-service/lib/electric/postgres/identifiers.ex for some examples.

I much prefer using it over ad-hoc delimiters with the `~S` sigil because it's so common in Elixir code already. TO be fair, we also have some occurrences of `"\""` in the code, it's yet another way to write the same string containing a single double-quote character.

alco:
> Nice! So it was just a bug from the ordering of the imports?

I've changed the order of plug applications, which is simply changing the order of function application as each plug is a function. Now, we first set the response headers on the `conn` struct, then perform the Etag check which may return early in case the Etag in the request header matches.

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
KyleAMathews: Ah neat!

PR #2068, opened 2024-11-28
Add ability to disable colored logs; rename logging-related env vars

alco:
Related documentation update - https://github.com/electric-sql/electric/pull/2067.

### DIFF
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -8,7 +8,7 @@ if config_env() in [:dev, :test] do
 end
 
 log_level_config =
-  env!("LOG_LEVEL", :string, "info")
+  env!("ELECTRIC_LOG_LEVEL", :string, "info")
   |> Electric.ConfigParser.parse_log_level()
 
 case log_level_config do
@@ -19,7 +19,7 @@ case log_level_config do
     raise message
 end
 
-if !env!("LOG_COLORS", :boolean, true) do
+if !env!("ELECTRIC_LOG_COLORS", :boolean, true) do
   config :logger, :default_formatter, colors: [enabled: false]
 end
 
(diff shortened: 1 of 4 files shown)


packages/sync-service/config/runtime.exs
```diff
@@ -19,14 +19,14 @@ case log_level_config do
...
-if !env!("LOG_COLORS", :boolean, true) do
+if !env!("ELECTRIC_LOGGING_COLORS", :boolean, true) do
   config :logger, :default_formatter, colors: [enabled: false]
 end
 
 # Enable this to get **very noisy** but useful messages from BEAM about
 # processes being started, stopped and crashes.
 # https://www.erlang.org/doc/apps/sasl/error_logging#sasl-reports
-sasl? = env!("ELECTRIC_LOG_OTP_REPORTS", :boolean, false)
+sasl? = env!("ELECTRIC_LOGGING_OTP_REPORTS", :boolean, false)
```
robacourt, line 29: Log is a verb here (Shall we log OTP reports? true or false), but logging cannot be a verb, so I don't this makes sense. Personally I'd keep all the logging things saying "LOG" and only use the word log for logging, at least for customer facing stuff.

alco: I'm thinking about it as if it were `ELECTRIC_LOGGING_OTP_REPORTS(ENABLED)`, in the same vein as with `ELECTRIC_LOGGING_COLORS(ENABLED)`.

Even stick to `LOG` doesn't fully solve the inconsistency because what's currently called `LOG_LEVEL` is using that word as a noun.

I wanted to use `ELECTRIC_LOGGING` as a namespace and separate it from the rest of the name with two underscores, as in `ELECTRIC_LOGGING__OTP_REPORTS`. But going this far might require retouching more than these env vars, which might be an undesirable scope right now? I don't know.

robacourt: Sure, personally I'd stick with log, but I don't have a strong opinion. The main thing is not using the word log for shapes.

balegas:
Hei, please modify the SST config to use this option

alco:
> Hei, please modify the SST config to use this option

Electric's config is not used when it is started in multi-tenancy mode. So no changes are needed to cloud's SST config. Unless you were referring to the two `sst.config.ts` files we have in two example apps in this repo?

robacourt (approved):

PR #2076, opened 2024-12-02
feat: optimise where clause filtering

### LINKED ISSUE #2040: improvement: better disjoint where clause evaluation
Right now, when Electric knows about N shapes with where clauses, we're quite naive about figuring out which shapes does a given transaction fall into. For example, if we have N shapes over the same table with where clauses in the pattern of  `WHERE id = $1`.

We can improve the performance and reduce the CPU usage if we're smarter about it. For the example above, we can infer that if one of the where clauses is true, all the others are false and we don't need to evaluate them. Moreover, for `=` check, we can precompute a lookup structure to know in one lookup which where clauses of the set _can_ evaluate to `true` at all. If where clauses have `WHERE ID = $1 AND .... form`, then we can cut away definitely-false where clauses and execute only M << N checks instead.

The precomute can be done on a relatively rare event of shape creation (rare in relation to transaction processing).

robacourt:
Fixes #2040 and is estimated to improve Trigger.dev's replication stream processing by 400x

### DIFF
--- a/packages/sync-service/lib/electric/shapes/filter.ex
+++ b/packages/sync-service/lib/electric/shapes/filter.ex
@@ -36,7 +36,7 @@ defmodule Electric.Shapes.Filter do
         |> Enum.map(fn {table_name, table} ->
           {table_name, Table.remove_shape(table, handle)}
         end)
-        |> Enum.reject(fn {_table, table} -> map_size(table.indexes) == 0 end)
+        |> Enum.reject(fn {_table, table} -> Table.empty?(table) end)
         |> Map.new()
     }
   end
--- a/packages/sync-service/lib/electric/shapes/filter.ex
+++ b/packages/sync-service/lib/electric/shapes/filter.ex
@@ -109,7 +109,7 @@ defmodule Electric.Shapes.Filter do
 
   defp other_shapes_affected(%{other_shapes: shapes}, record) do
     for %{handle: handle, shape: shape} <- shapes,
-        Shape.record_in_shape?(shape.where, record),
+        Shape.record_in_shape?(shape, record),
         into: MapSet.new() do
       handle
     end
--- a/packages/sync-service/lib/electric/shapes/filter.ex
+++ b/packages/sync-service/lib/electric/shapes/filter.ex
@@ -12,29 +12,32 @@ defmodule Electric.Shapes.Filter do
   defstruct tables: %{}
 
   def new(shapes), do: shapes |> Map.to_list() |> new(empty())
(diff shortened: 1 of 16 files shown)


balegas:
Exciting!

packages/sync-service/lib/electric/shapes/filter.ex
```diff
@@ -50,51 +51,72 @@ defmodule Electric.Shapes.Filter do
     }
   end
 
-  def affected_shapes(%Filter{} = filter, %Relation{} = relation) do
+  def affected_shapes(%Filter{} = filter, change) do
+    shapes_affected_by_change(filter, change)
+  rescue
```
alco, line 77: Why is this needed? What kinds of errors do we expect to see here?

packages/sync-service/test/electric/shapes/filter_test.exs
```diff
@@ -134,6 +134,59 @@ defmodule Electric.Shapes.FilterTest do
...
+        |> Enum.reduce(Filter.empty(), fn i, filter ->
+          Filter.add_shape(filter, i, Shape.new!("t1", where: "id = #{i}", inspector: @inspector))
+        end)
+
+      reductions =
+        reductions(fn ->
+          assert Filter.affected_shapes(filter, change("t1", %{"id" => "7"})) == MapSet.new([7])
+        end)
+
+      assert reductions < 500
```
alco, line 150: Would be nice to see this specific number explain. What is the significance of it in this test? What should we do if this assertion fails?

alco (approved):
:muscle:

packages/sync-service/lib/electric/shapes/filter.ex
robacourt: Super safety 😄 My reasoning is that if filtering crashes, all the shapes crash, so we need to be more careful than if this was shape specific code. There's a much more targeted `rescue` in `Table` that deals with record parsing errors which is much more likely to be hit. This one is more of a catch-all just to be extra safe. 

Types of error is could catch:
1. `Shape.is_affected_by_relation_change?` exceptions
2. change types the filter is not expecting are passed to `affected_shapes/2` 

These hopefully will never happen, but if they do, we'd rather the filter didn't crash.

packages/sync-service/test/electric/shapes/filter_test.exs
robacourt: Good point. Let me write a comment

icehaunter (approved):
Great job, exactly what I had in mind

Issue #2079, opened 2024-12-02
Move non-protocol options to the `params` sub-key


KyleAMathews:
For `ShapeStreamOptions` — https://github.com/electric-sql/electric/blob/1f2a5ff673107c675fa269265930067364d60503/packages/typescript-client/src/client.ts#L77-L161

In the future, Electric the protocol will support multiple sources, not just PG (not ETA but that's the general plan).

So to future-proof the client, we should remove from the top-level options that are specific to PG so that the client isn't pg-specific.

Options to move:
- table
- where
- columns
- replica

We can still type/document these — perhaps in the future we can make the param types source-specific so when you're creating a shape stream for pg vs something else there's perhaps separate imports. But that can come later — just distinguishing between protocol options like `offset` and `handle` vs. source options is my goal now.

PR #2081, opened 2024-12-02
Move non-protocol options like table & where to the params sub-key

KyleAMathews:
Fix https://github.com/electric-sql/electric/issues/2079

Electric's TypeScript client is currently tightly coupled to PostgreSQL-specific options in its `ShapeStreamOptions` interface. As Electric plans to support multiple data sources in the future, we need to separate protocol-level options from source-specific options.

## Changes

1. Created a new `PostgresParams` type to define PostgreSQL-specific parameters:
   - `table`: The root table for the shape
   - `where`: Where clauses for the shape
   - `columns`: Columns to include in the shape
   - `replica`: Whether to send full or partial row updates
2. Moved PostgreSQL-specific options from the top-level `ShapeStreamOptions` interface to the `params` sub-key
3. Updated `ParamsRecord` type to include PostgreSQL parameters
4. Updated the `ShapeStream` class to handle parameters from the `params` object
5. Updated documentation to reflect the changes

## Migration Example

Before:
```typescript
  table: 'users',
  where: 'id > 100',
  columns: ['id', 'name'],
  replica: 'full'
})
```

After:
```typescript
  params: {
    table: 'users',
    where: 'id > 100',
    columns: ['id', 'name'],
    replica: 'full'
  }
})
```

### DIFF
--- a/.changeset/tiny-socks-drive.md
+++ b/.changeset/tiny-socks-drive.md
@@ -1,5 +1,6 @@
 ---
 "@electric-sql/client": minor
+"@electric-sql/react": minor
 ---
 
 [BREAKING]: Move non-protocol options like table & where to the params sub-key
--- a/.changeset/tiny-socks-drive.md
+++ b/.changeset/tiny-socks-drive.md
@@ -46,17 +46,3 @@ const stream = new ShapeStream({
   }
 })
 ```
-
-## Benefits
-
-1. **Future-proofing**: The client is now better prepared for supporting multiple data sources beyond PostgreSQL
-2. **Clean separation**: Protocol-level options (`offset`, `handle`) are now clearly separated from source-specific options
-3. **Extensibility**: Makes it easier to add support for new data sources in the future
-4. **Type safety**: Source-specific options are now properly typed through the `PostgresShapeParams` type
-
-## Implementation Notes
-
-- The changes are backwards compatible as the existing code paths are maintained
-- Protocol-level options like `offset` and `handle` remain at the top level
-- Documentation has been updated to reflect the new parameter structure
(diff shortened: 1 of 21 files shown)


KyleAMathews:
On extending types — we have this issue with Electric Cloud where `token` and `database_id` params are required when calling the API.

balegas (approved):
I always though of what you're calling `PostgresParams` as the `ShapeDefinition`. I understand the reasoning that `where` clauses are specific to SQL/relational databases and not the shape, and that other shapes will have different models, but we also don't know what the API is going to look like two steps down the road.

I also think that except for `replica` all the parameters belong to the class of `SQLParams`, where `MySQL`, etc would also fit. I think this can be addresses with the hierarchy of types and only handled when necessary, but if we're trying to communicate support for multiple databases in the future I think this would be a nice thing to model.

I'm sharing my two cents, but I am happy to accept the changes as they are

thruflo:
I agree that the motivation of supporting non PG sources is a long way off and I would argue we should have an explicit way of defining a shape, via a shape definition.

I think the move to having what is currently the shape definition under a params key is harmless but I’m not sure the motivation behind it holds.

samwillis:
My 2 cents... I think this is good. 'params' works at the url params level, and is compatible with any proxy. If we one day add a shape def in json, it's like a "shape" param and would be under options.params.shape. And maybe a shortcut to it via options.shape.

Maybe we let users add any key+value to params to add to the url?

I do think we need to start thinking what a more expressive definition of a shape is, speculatively, to guide what this api should look like. Include, joins, remapping, even aggregates, are things we may want to do in future. It's worth while before we evolve this api much further.

👍

PR #2083, opened 2024-12-03
chore (sync service): store OTEL attributes per stack in a persistent term

kevin-dp:
This PR is an alternative to https://github.com/electric-sql/electric/pull/2077.
Instead of injecting the OTEL attributes and passing them around everywhere, we store each stack's OTEL attributes in a dedicated persistent term. When creating a span we lookup the OTEL attributes for that stack and add them to the span.


robacourt (changes requested):
Great work! Alas it doesn't reduce the number or core code changes as much as I hoped because of having to pass the stack_id around. But it is less changes and it is `stack_id` which is useful for other things as opposed to something open telemetry specific. So I much prefer this. Thanks. I've just added one thing I think should change.

robacourt (approved):
Excellent! Very minor point: I'd remove the redundant word "telemetry" from OpenTelemetry.set_stack_telemetry_span_attrs

Issue #1885, opened 2024-10-23
Separate `DATABASE_URL`s for replication and query connections.


thruflo:
I've been doing a pass on deployment docs, including verifying the steps to connect to managed Postgres providers. In a few cases (Supabase, GCP, etc.) they provide a direct IP address for logical replication and a connection-pooled IP address for interactive access.

We currently accept a single `DATABASE_URL`. Because this is used for logical replication, it must be the direct IP address. Which works but is inefficient. Because we only need to special case the logical replication connection. Our interactive queries could perfectly well go through the user's normal DB pooler, reducing our impact on their database. 

I propose that we adjust our config to support a single `DATABASE_URL` as the current default but also support providing a pair of `REPLICATION_DATABASE_URL` and `INTERACTIVE_DATABASE_URL` or whatever the best naming is.

Plus we would need to extend this pattern for multi-tenancy, so that apps can be configured with a pair of values, instead of one URL per database ID.

alco:
Another way to think about this is keeping `DATABASE_URL` as the primary one, which can be configured standalone just like right now, but adding an optional config named `DATABASE_POOL_URL` or `READ_DATABASE_URL`. Setting the latter will cause Electric to use it for general queries.

alco:
In terms of implementing this separation, it seems quite straightforward: add database connection options to `replication_opts` that are passed to `Electric.Connection.Manager` and make the manager use those for opening a replication connection. Boom, we're done.

Well, almost. We'll also need to handle the fallback from SSL to noSSL for one more set of connection options. Just another argument for adding that fallback logic to the Postgrex proper.

msfstef:
Picking this up as it has become quite necessary with the deploying of multiple examples and sharing of postgres instances.

PR #2085, opened 2024-12-03
Add an optional DATABASE_POOL_URL config

alco:
This implements #1885.

When configured, the lock connection and the database pool will connect to this pooled URL.

By default, when `DATABASE_POOL_URL` is not set, all connections are open using the direct database connection URL configured with `DATABASE_URL`.

Things to add before merging:
- [ ] Changeset
- [ ] Docs
- [ ] A test to check that all DB connections aside from the replication one are using the pool URL

### DIFF
--- a/.changeset/stupid-weeks-look.md
+++ b/.changeset/stupid-weeks-look.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Restore the automatic fallback to unencrypted database connections when SSL isn't available.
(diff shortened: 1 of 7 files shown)


icehaunter (approved):
LGTM, but can we have a sibling PR for the cloud because otherwise after this we're breaking the new contract of StackSupervisor and NimbleOptions will complain there

balegas:
Seems we had some miscommunication and did this again in a different PR https://github.com/electric-sql/electric/pull/2438
@alco, @msfstef  is anything from this PR that we should pick up, or shall we close it

msfstef:
whoops somehow I missed this! However I used @alco 's comments on the issue as a guide to implement my own PR and looking at the changes here it seems it's almost identical so I think we can close this

PR #2092, opened 2024-12-03
fix: only clear the React stream/shape cache when the previous stream/shape was aborted

KyleAMathews:
The current behavior also clears them when there's an error which leads to an infinite loop with a bad stream being recreated over and over.


samwillis (approved):
LGTM (and solves what I've witnessed!)

KyleAMathews:
> and solves what I've witnessed!

great!

PR #2095, opened 2024-12-04
chore (sync service): improve nimble spec for telemetry span attributes

kevin-dp:
This PR improves the nimble spec for validating span attributes.

### DIFF
--- a/packages/sync-service/lib/electric/stack_supervisor.ex
+++ b/packages/sync-service/lib/electric/stack_supervisor.ex
@@ -116,12 +116,8 @@ defmodule Electric.StackSupervisor do
                  ]
                )
 
-  def validate(opts) do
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/stack_supervisor.ex
```diff
@@ -80,13 +80,48 @@ defmodule Electric.StackSupervisor do
...
+                               {:list, {:or, [:atom, :string, :integer, :float, :boolean]}},
+                               :map
+                             ]}
+                          ]}}
+                      ]},
                    required: false
                  ]
                )
 
+  def validate(opts) do
```
icehaunter, line 119: I don't see a reason to extract this as a public function tbh

icehaunter (changes requested):
small nitpick

icehaunter (approved):

PR #2096, opened 2024-12-04
fix(elixir-client): link request & monitor so client never stalls

icehaunter:


robacourt (approved):

packages/elixir-client/lib/electric/client/fetch/monitor.ex
```diff
@@ -37,6 +37,10 @@ defmodule Electric.Client.Fetch.Monitor do
   def wait(ref) do
     receive do
       {:response, ^ref, response} -> response
+    after
+      60_000 ->
```
magnetised, line 41: I worry about a global timeout because the connection may be receiving a large chunk of data successfully, but slowly. I think with the process linking we should handle any errors and I'd prefer timeouts to be handled by the HTTP client, which is aware of what's happening on the connection (e.g. the difference between a slow download and a broken, unresponsive, connection)

magnetised (changes requested):
I'd have gone with a monitor but this works. Have concerns around the global timeout though.

packages/elixir-client/lib/electric/client/fetch/monitor.ex
icehaunter: Ok, sounds good. We do know that the response process died, so that should be alright

magnetised (approved):
Lovely

PR #2098, opened 2024-12-04
docs: update shape definitions to use new `params` key.

thruflo:
Missed a few places where we need to update the switch to put the shape definition in the `params` key from https://github.com/electric-sql/electric/pull/2081


KyleAMathews (approved):
:shipit:

PR #2100, opened 2024-12-05
feat: configure logging to be more explicit

icehaunter:
Adds stack id & other useful metadata to all processes we spawn, and configures logger to actually print it

### DIFF
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -7,21 +7,14 @@ if config_env() in [:dev, :test] do
   source!([".env.#{config_env()}", ".env.#{config_env()}.local", System.get_env()])
 end
 
-log_level_config =
-  env!("ELECTRIC_LOGGING_LEVEL", :string, "info")
-  |> Electric.ConfigParser.parse_log_level()
+config :logger,
+  level: env!("ELECTRIC_LOG_LEVEL", &Electric.ConfigParser.parse_log_level!/1, :info)
 
-case log_level_config do
-  {:ok, log_level} ->
-    config :logger, level: log_level
-
-  {:error, message} ->
-    raise message
-end
-
-if !env!("ELECTRIC_LOG_COLORS", :boolean, true) do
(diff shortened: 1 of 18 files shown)


packages/sync-service/config/runtime.exs
```diff
@@ -7,21 +7,14 @@ if config_env() in [:dev, :test] do
...
-
-case log_level_config do
-  {:ok, log_level} ->
-    config :logger, level: log_level
-
-  {:error, message} ->
-    raise message
-end
+config :logger,
+  level: env!("ELECTRIC_LOGGING_LEVEL", &Electric.ConfigParser.parse_log_level!/1, :info)
```
alco, line 11: Wow, I didn't notice this name ended up in main. I already renamed it to `ELECTRIC_LOG_LEVEL` but then it must have gotten lost when you merged the PR before I had the chance to push the updated naming.

I'll fix these names in a new PR.

packages/sync-service/config/runtime.exs
```diff
@@ -7,21 +7,14 @@ if config_env() in [:dev, :test] do
...
 
-if !env!("ELECTRIC_LOG_COLORS", :boolean, true) do
-  config :logger, :default_formatter, colors: [enabled: false]
-end
+config :logger, :default_formatter,
+  # Doubled line breaks serve as long message boundaries
+  format: "\n$time $metadata[$level] $message\n",
+  metadata: [:pid, :shape_handle, :request_id],
+  colors: [enabled: env!("ELECTRIC_LOG_COLORS", :boolean!, true)]
```
alco, line 17: This is great, I already hit a problem with the previous version of this where I couldn't force the use of colors in the terminal when piping output to `tee`.

alco (approved):

kevin-dp (approved):
Perfect!

packages/sync-service/config/runtime.exs
icehaunter: I'll fix it in this PR

PR #2103, opened 2024-12-05
NextJs Example: fixes for serverless deployment

balegas:
Some fixes and updated SST


examples/nextjs-example/app/api/items/route.ts
```diff
@@ -1,17 +1,32 @@
...
   const body = await request.json()
-  const result = await db.query(
-    `INSERT INTO items (id)
-    VALUES ($1) RETURNING id;`,
-    [body.uuid]
-  )
-  return NextResponse.json({ id: result.rows[0].id })
+
+  try {
+    await client.connect()
```
kevin-dp, line 12: If you use a pool you can do `pool.query` instead of this pattern of having to connect, query, and then disconnect.
Same for Line 26.

thruflo (approved):

examples/nextjs-example/app/api/items/route.ts
balegas: yep. applying these changes throughout open PRs as I go through them. thanks you for raising

KyleAMathews (approved):
:shipit:

PR #2105, opened 2024-12-06
website: mark products with status label.

thruflo:
Minor one to merge when the BETA release is being made.


alco (approved):
Nice!

I would add more flare, at least to the "Private alpha" and "Alpha" labels, perhaps coloring the former in red and the latter in amber. The current style makes it easy to miss the label during a cursory look.

PR #2109, opened 2024-12-06
Deployments script for Todos app

balegas:
- [ ] add to GH action
- [ ] clean docker image a bit more

Will not add to GH action because we're going the spend all replication slots with all demos. Would be nice to generate an image with all demos and share a single database. I don't think this is very hard.


examples/todo-app/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM node:lts-alpine AS base
+
+# Stage 1: Install dependencies
+FROM base AS deps
```
kevin-dp, line 4: Why do you need this `FROM` ?
Wouldn't it be equivalent to do `FROM node:lts-alpine AS deps` on line 1?

kevin-dp: Unless on L22 you would do `FROM base AS prod`

examples/todo-app/server.js
```diff
@@ -4,15 +4,9 @@ import cors from "cors"
...
-  host: `localhost`,
-  port: 54321,
-  password: `password`,
-  user: `postgres`,
-  database: `electric`,
-})
-client.connect()
+const { Pool } = pg
+const pool = new Pool({connectionString: process.env.DATABASE_URL})
+await pool.connect()
```
kevin-dp, line 9: Remove this line.
`connect` acquires a client from the pool but you're never actually using it.
So you're acquiring a client that is never released.

examples/todo-app/src/routes/index.tsx
```diff
@@ -19,10 +19,12 @@ type ToDo = {
 
 export default function Index() {
   const { data: todos } = useShape<ToDo>({
-    url: `http://localhost:3000/v1/shape`,
+    url: `${import.meta.env.VITE_ELECTRIC_URL}/v1/shape/`,
```
kevin-dp, line 22: Using the `URL` constructor to contruct a URL from a path and a base URL would be more robust since here `VITE_ELECTRIC_URL` may e.g. not contain a slash at the end. Same comment applies for line 47, 70, and 91.

examples/todo-app/sst.config.ts
```diff
@@ -0,0 +1,169 @@
...
+    const project = neon.getProjectOutput({ id: process.env.NEON_PROJECT_ID! })
+    const base = {
+      projectId: project.id,
+      branchId: project.defaultBranchId,
+    }
+
+    const db = new neon.Database(`todo-app-example`, {
+      ...base,
+      name:
+        $app.stage === `Production`
```
kevin-dp, line 31: Perhaps we can remove the ternary if we name it as follows:
```
name: `todo-app-${$app.stage.toLowerCase()}`
```

examples/todo-app/sst.config.ts
```diff
@@ -0,0 +1,169 @@
...
+          ? `todo-app-production`
+          : `todo-app-${$app.stage}`,
+      ownerName: `neondb_owner`,
+    })
+
+    const databaseUri = getNeonDbUri(project, db, false)
+    // const databasePooledUri = getNeonDbUri(project, db, true)
```
kevin-dp, line 38: let's remove this commented line

examples/todo-app/sst.config.ts
```diff
@@ -0,0 +1,169 @@
...
+    try {
```
kevin-dp, line 39: why do we start the `try` block only here?
I'm assuming any of the previous steps may fail as well, therefore, i would start the `try` block at the very beginning of this `run` function.

examples/todo-app/sst.config.ts
```diff
@@ -0,0 +1,169 @@
...
+      const vpc = new sst.aws.Vpc(`todo-app-${$app.stage}-vpc`)
+      const cluster = new sst.aws.Cluster(`todo-app-${$app.stage}-cluster`, {
+        vpc,
+      })
+
+      const service = cluster.addService(`todo-app-service-${$app.stage}`, {
+        loadBalancer: {
+          ports: [{ listen: "443/https", forward: "3010/http" }],
+          domain: {
+            name: `todo-app-server-${$app.stage === `production` ? `` : `-stage-${$app.stage}`}.electric-sql.com`,
```
kevin-dp, line 55: On L31 we checked for `$app.stage === 'Production'`, here we check if it is `'production'`.
One of the two is wrongly capitalised?

balegas (commented):
Thanks For the comment Kevin

KyleAMathews (approved):

PR #2110, opened 2024-12-06
Update data-sync.md to add "no stale data" as key benefit

KyleAMathews:


website/use-cases/data-sync.md
```diff
@@ -12,6 +12,7 @@ solves: "state transfer"
 benefits:
   - Simplifies your code
   - No loading spinners
+  - No stale data
```
thruflo, line 15: All the use cases show two benefit points. I'd suggest dropping "No loading spinners"?

KyleAMathews: oh ok sure then — stale data is the root of the problem — spinners are the symptom. Also to be very nit-picky, we'd still have loading spinners on the initial load so it'd be more we have "fewer spinners" which doesn't sound as good.

Issue #2107, opened 2024-12-06
@electric-sql/client masquerading as ESM when using node16 moduleResolution


ericallam:
Even though `@electric-sql/client` does publish a CJS file, I'm not able to import the library without using a dynamic import when using the node16 moduleResolution setting, in a CJS context. This is because TypeScript detects the types of the import is an ESM module, since you are only 
publishing the ESM module types. See these results from ATTW:

https://arethetypeswrong.github.io/?p=%40electric-sql%2Fclient%400.9.0

![CleanShot 2024-12-06 at 10 41 27](https://github.com/user-attachments/assets/0f3320e8-b58d-4295-82a4-d7a0c92d3128)

PR #2112, opened 2024-12-06
Fix node 16 cjs import

KyleAMathews:
Fixes https://github.com/electric-sql/electric/issues/2107

From running `npm pack` locally on PR:

Screenshot 2024-12-06 at 9 24 58 AM

The change is making exports explicit for esm or cjs:
```json
"exports": {
  ".": {
    "import": {
      "types": "./dist/index.d.ts",
      "default": "./dist/index.mjs"
    },
    "require": {
      "types": "./dist/cjs/index.d.cts",
      "default": "./dist/cjs/index.cjs"
    }
  }
}
```

And generating separate type declarations.


packages/typescript-client/package.json
```diff
@@ -49,22 +26,50 @@
...
     "vitest": "^2.0.2"
   },
+  "type": "module",
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.mjs"
+      },
```
KyleAMathews, line 36: diff is noisier than ideal as I ran https://www.npmjs.com/package/fixpack on the package.json

Issue #2072, opened 2024-11-29
useShape sync broken on Expo react native - CORS config error and missing headers warning


federico-hv:
Hi, first of all thanks for this amazing piece of tech. I love Postgres and I'm really excited about the future of Electric. I was recently trying the new version and I deployed the service to a Digital Ocean Droplet with Docker as recommended on the docs https://electric-sql.com/docs/integrations/digital-ocean. Worth noting I had to set up an IPV6 network to connect from within to Supabase (maybe this could be a good addition to the docs). Now, the service seems to be working fine on the initial sync and I get all the headers right while testing with curl but I'm having one issue with the `useShape` throwing an warning on react-native when I make updates to the database and not retrieving the latest data (only react-native on web seems to work fine) . Also, I know the hook is only for read-only (that's what I'm testing it for while making updates directly to the DB). I'm attaching a screenshot of the error, it seems that there are missing headers but this is only happening on react native (Expo) not on web. I checked the service with curl but it already has the headers and seems ok, however I read some people are using services on top (proxies) like hono.js and they had to use a specific [CORS config](https://github.com/electric-sql/electric/issues/1951) for these headers to be preserved. I tried adding an express proxy on top but the issue is still there. Any ideas on what could be causing this issue? Just wondering if I missed any configuration? 

![Screenshot 2024-11-29 at 2 47 27 PM](https://github.com/user-attachments/assets/8e943467-6851-4a44-b415-db6c33310e00)

<img width="969" alt="Screenshot 2024-11-29 at 3 31 31 PM" src="https://github.com/user-attachments/assets/8b7bc8ef-5a72-4533-90ee-76d8d2102981">

KyleAMathews:
Hmm so the schema isn't returned on live quests (when request includes `?live`) — so this error shouldn't happen as the client only looks for `electric-cursor` not `electric-schema` — https://github.com/electric-sql/electric/blob/32d4b5085a3c54a53216e37ad230a2fb79fdbb1f/packages/typescript-client/src/fetch.ts#L184-L189

Do live updates come through? I'm not very familiar with why Expo wouldn't be working here. Perhaps the fetch implementation is different somehow?

federico-hv:
[...]

PR #2116, opened 2024-12-07
refactor(ts client): remove url object usage in fetch.ts

federico-hv:
This PR removes the use of URL object from the typscript client to avoid compatibility issues with react-native
https://github.com/electric-sql/electric/issues/2072 and adds a generic check with urlite


KyleAMathews:
Closing as discussed in the issue https://github.com/electric-sql/electric/issues/2072#issuecomment-2524844376

PR #2117, opened 2024-12-07
examples: add encryption example.

thruflo:
https://github.com/electric-sql/electric/pull/2117/files#diff-2af99c046c43f919c4cfe72af4c3870c1ca943f076d92e7cb5c39a882c5b8d19


examples/encryption/src/Example.tsx
```diff
@@ -0,0 +1,169 @@
+import base64 from 'base64-js'
```
samwillis, line 1: As this is a web/browser example you don't need a separate dependancy for base64, the browser [`btoa`](https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa) and [`atob`](https://developer.mozilla.org/en-US/docs/Web/API/Window/atob) apis should do, I think. Particularly as we are using the browser crypto apis.

examples/encryption/src/Example.tsx
```diff
@@ -0,0 +1,169 @@
...
+import React, { useEffect, useState } from 'react'
+import { v4 as uuidv4 } from 'uuid'
```
samwillis, line 3: Similar to above, we can use [`crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) in the browser.

examples/encryption/src/Example.tsx
```diff
@@ -0,0 +1,169 @@
...
+}
+
+const API_URL = import.meta.env.API_URL || 'http://localhost:3001'
+const ELECTRIC_URL = import.meta.env.ELECTRIC_URL ?? 'http://localhost:3000'
+
+// For this example, we hardcode a deterministic key that works across page loads.
+// In a real app, you would implement a key management strategy. Electric is great
+// at syncing keys between users :)
+const rawKey = new Uint8Array(16)
+const key = await crypto.subtle.importKey('raw', rawKey, 'AES-GCM', true, [
```
samwillis, line 25: Good choice of `AES-GCM` 👍

samwillis (approved):
This is great! Anyone that goes this route is going to be experienced in encryption, and so a basic example like this is perfect to show the concept without having to go too far in building out a full system. We don't want to show patterns for key exchange or generation.

examples/encryption/src/Example.tsx
thruflo: I tried atob and btoa first but they just weirdly didn’t handle the array buffer encoding correctly. There’s https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64 but it’s not supported yet, so I either needed a polyfill or a dep.

samwillis: Ah, sure! Web APIs :man_shrugging:

PR #2119, opened 2024-12-08
feat: introduce initial snapshot chunking

icehaunter:
This PR changes the way we write snapshots to disk, the storage API to request them, and some logic around offsets.

1. The `Storage.get_snapshot/1` method is gone. It's now unified under `Storage.get_log_stream/3`, with some caveats.
2. The offsets within the snapshot rows are not updated, so now latest item on the sent log may have mismatching `offset` field to the `electric-offset` header of the same response. All our clients respect the header, and that's the only correct behaviour going forward. There is a decision to remove `offset` from our items entirely, but that's a separate PR.
3. While we're making the snapshot, we cannot know how many chunks are going to be there until we make them. Thus we allow clients to request any chunks up to `LogOffset.new(0, :infinity)`, and consider all offsets at txn 0 to be virtual pointers.
4. The `get_log_stream/3` for the snapshot section ignores the contract a little bit - you cannot request the stream to be up to an arbitrary point, it's always up to the chunk boundary, even when specified otherwise. Since we've moved to "pure" chunking behaviour, it should be enshrined in the `Storage` interface so that reading functions can act optimally and honestly. I'm hoping to address that in a separate PR
5. If the instance is started with the old snapshot already created, we update the naming and metadata to avoid recreating shapes if we can avoid it.

### DIFF
--- a/packages/elixir-client/lib/electric/client/offset.ex
+++ b/packages/elixir-client/lib/electric/client/offset.ex
@@ -67,6 +67,9 @@ defmodule Electric.Client.Offset do
       iex> from_string("1378734_3")
       {:ok, %#{__MODULE__}{tx: 1378734, op: 3}}
 
+      iex> from_string("0_inf")
+      {:ok, %#{__MODULE__}{tx: 0, op: :infinity}}
+
       iex> from_string("not a real offset")
       {:error, "has invalid format"}
 
--- a/packages/elixir-client/lib/electric/client/offset.ex
+++ b/packages/elixir-client/lib/electric/client/offset.ex
@@ -78,13 +78,16 @@ defmodule Electric.Client.Offset do
     else
       with [tx_offset_str, op_offset_str] <- :binary.split(str, "_"),
            {tx_offset, ""} <- Integer.parse(tx_offset_str),
-           {op_offset, ""} <- Integer.parse(op_offset_str) do
+           {op_offset, ""} <- parse_int_or_inf(op_offset_str) do
         {:ok, %__MODULE__{tx: tx_offset, op: op_offset}}
       else
         _ -> {:error, "has invalid format"}
       end
     end
   end
+
+  defp parse_int_or_inf("inf"), do: {:infinity, ""}
+  defp parse_int_or_inf(int), do: Integer.parse(int)
 
   @doc """
   Create a new #{__MODULE__} struct from the given LSN and operation
@@ -115,7 +118,7 @@ defmodule Electric.Client.Offset do
   end
 
   def to_string(%__MODULE__{tx: tx, op: op}) do
-    "#{Integer.to_string(tx)}_#{Integer.to_string(op)}"
+    "#{Integer.to_string(tx)}_#{if op == :infinity, do: "inf", else: Integer.to_string(op)}"
   end
 
   @spec to_tuple(t()) :: {tx_offset(), op_offset()}
--- a/packages/react-hooks/test/react-hooks.test.tsx
+++ b/packages/react-hooks/test/react-hooks.test.tsx
@@ -3,7 +3,7 @@ import { describe, expect, inject, it as bareIt } from 'vitest'
 import { setTimeout as sleep } from 'node:timers/promises'
 import { testWithIssuesTable as it } from './support/test-context'
 import { useShape, sortedOptionsHash, UseShapeResult } from '../src/react-hooks'
-import { Shape, Message } from '@electric-sql/client'
+import { Shape, ShapeStream } from '@electric-sql/client'
 
 const BASE_URL = inject(`baseUrl`)
 
@@ -351,14 +351,20 @@ describe(`useShape`, () => {
     unmount()
 
     // Add another row to shape
-    const [newId] = await insertIssues({ title: `other row` })
+    const [_] = await insertIssues({ title: `other row` })
+
+    const parallelWaiterStream = new ShapeStream({
+      url: `${BASE_URL}/v1/shape`,
+      params: {
+        table: issuesTableUrl,
+      },
+      signal: aborter.signal,
+      subscribe: true,
+    })
+
     // And wait until it's definitely seen
     await waitFor(async () => {
-      const res = await fetch(
-        `${BASE_URL}/v1/shape?table=${issuesTableUrl}&offset=-1`
-      )
-      const body = (await res.json()) as Message[]
-      expect(body).toMatchObject([{}, { value: { id: newId } }])
+      return parallelWaiterStream.isUpToDate || (await sleep(50))
     })
 
     await sleep(50)
--- a/packages/sync-service/lib/electric/replication/log_offset.ex
+++ b/packages/sync-service/lib/electric/replication/log_offset.ex
@@ -47,7 +47,8 @@ defmodule Electric.Replication.LogOffset do
       ** (FunctionClauseError) no function clause matching in Electric.Replication.LogOffset.new/2
   """
   def new(tx_offset, op_offset)
-      when is_integer(tx_offset) and tx_offset >= 0 and is_integer(op_offset) and op_offset >= 0 do
+      when is_integer(tx_offset) and tx_offset >= 0 and is_integer(op_offset) and op_offset >= 0
+      when is_integer(tx_offset) and tx_offset >= 0 and op_offset == :infinity do
     %LogOffset{tx_offset: tx_offset, op_offset: op_offset}
   end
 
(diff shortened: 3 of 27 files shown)


icehaunter:
@msfstef can you please take another look at `.skip`-ed tests in the TS client that deal with headers and confirm that this me disabling them is correct? The general idea is "well we're serving only full chunks when serving snapshot, and up to file boundary, so switching to 'live' log is a separate HTTP request, hence e-tags should no longer mutate because response no longer mutates.

packages/elixir-client/lib/electric/client/offset.ex
```diff
@@ -78,14 +78,17 @@ defmodule Electric.Client.Offset do
     else
       with [tx_offset_str, op_offset_str] <- :binary.split(str, "_"),
            {tx_offset, ""} <- Integer.parse(tx_offset_str),
-           {op_offset, ""} <- Integer.parse(op_offset_str) do
+           {op_offset, ""} <- parse_int_or_inf(op_offset_str) do
```
msfstef, line 81: nit: doctest with `0_inf` or `{num}_inf` for good measure?

packages/react-hooks/test/react-hooks.test.tsx
```diff
@@ -351,14 +351,20 @@ describe(`useShape`, () => {
...
+    })
+
     // And wait until it's definitely seen
     await waitFor(async () => {
-      const res = await fetch(
-        `${BASE_URL}/v1/shape?table=${issuesTableUrl}&offset=-1`
-      )
-      const body = (await res.json()) as Message[]
-      expect(body).toMatchObject([{}, { value: { id: newId } }])
+      return parallelWaiterStream.isUpToDate || (await sleep(50))
```
msfstef, line 367: nit: the `waitFor` utility has `waitForOptions` as a second argument that accepts an `interval` parameter for how long to wait between tries

packages/sync-service/lib/electric/replication/log_offset.ex
```diff
@@ -130,6 +135,12 @@ defmodule Electric.Replication.LogOffset do
   @spec last() :: t
   def last(), do: %LogOffset{tx_offset: 0xFFFFFFFFFFFFFFFF, op_offset: :infinity}
 
+  @doc """
+  The last possible offset for the "virtual" part of the log - i.e. snapshots.
+  """
+  @spec last_before_real_offsets() :: t()
+  def last_before_real_offsets(), do: %LogOffset{tx_offset: 0, op_offset: :infinity}
```
msfstef, line 142: nit: not sure about the name, since we have an `is_snapshot_offset` guard and thus are exposing that aspect in the offset design perhaps this should also be `last_snapshot_offset`? The `last_before_real_offsets` requires the reader to know that we use virtual offsets _for_ snapshots, I think using `last_snapshot_offset` will make it easier to read

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -94,7 +94,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
     {hash, shape_handle} = Shape.generate_id(shape)
     # fresh snapshots always start with a zero offset - only once they
     # are folded into the log do we have non-zero offsets
-    offset = LogOffset.first()
+    offset = %LogOffset{LogOffset.last() | tx_offset: 0}
```
msfstef, line 97: why do we not want to start with `0_0` here, since as far as we know we have no data? also, isn't this equivalent to `last_before_real_offsets` or whatever we end up naming it? (edit: I read your comment in `FileStorage`, so I get why, but this is still `last_before_real_offsets` I think)

packages/sync-service/mix.exs
```diff
@@ -115,7 +115,9 @@ defmodule Electric.MixProject do
   defp aliases() do
     [
       start_dev: "cmd --cd dev docker compose up -d",
-      stop_dev: "cmd --cd dev docker compose down -v"
+      stop_dev: "cmd --cd dev docker compose down -v",
+      clean_persistant: "cmd rm -rf perisistent",
```
msfstef, line 119:
```suggestion
      clean_persistent: "cmd rm -rf perisistent",
```

packages/sync-service/mix.exs
```diff
@@ -115,7 +115,9 @@ defmodule Electric.MixProject do
...
+      reset: "do clean_persistant, stop_dev, start_dev"
```
msfstef, line 120:
```suggestion
      reset: "do clean_persistent, stop_dev, start_dev"
```

packages/typescript-client/test/integration.test.ts
```diff
@@ -308,11 +308,8 @@ describe(`HTTP Sync`, () => {
       )
 
       await vi.waitFor(async () => {
-        const res = await fetch(
-          `${BASE_URL}/v1/shape?table=${tableUrl}&offset=-1`
-        )
-        const body = (await res.json()) as Message[]
-        expect(body.length).greaterThan(1)
+        await sleep(100)
```
msfstef, line 311: nit: see `interval` option in `waitFor` (same for below)

packages/typescript-client/test/integration.test.ts
```diff
@@ -563,7 +557,9 @@ describe(`HTTP Sync`, () => {
     expect(catchupOpsCount).toBe(9)
   })
 
-  it(`should return correct caching headers`, async ({
+  // This test doesn't work anymore because server sends initial snapshot as a complete
+  // stable chunk, so e-tag is the same between requests.
+  it.skip(`should return correct caching headers`, async ({
```
msfstef, line 562: @icehaunter for these tests we want to be looking at an incomplete chunk instead of just a `-1` request (which used to collate the snapshot with the log) - so perhaps we do all the same checks but look at the next request rather than the first one, so we see the chunk starting with `0_inf` changing (same for the test below)

packages/sync-service/lib/electric/shape_cache/file_storage.ex
```diff
@@ -228,77 +257,65 @@ defmodule Electric.ShapeCache.FileStorage do
...
+      end,
+      fn {chunk_num, file} ->
+        if is_nil(file) and chunk_num == 0 do
+          # Special case if the source stream has ended before we started writing any chunks - we need to create the empty file for the first chunk.
+          {[chunk_num], {chunk_num, open_snapshot_chunk_to_write(opts, chunk_num)}}
+        else
+          {[], {chunk_num, file}}
+        end
+      end,
+      fn {_chunk_num, file} ->
```
msfstef, line 297: I might be misunderstanding, but how is this clause ever reached? will the one above not match everything before this? Or perhaps should the one above have the `is_nil(file) and chunk_num == 0` as a guard?

packages/sync-service/lib/electric/shape_cache/file_storage.ex
```diff
@@ -328,8 +436,65 @@ defmodule Electric.ShapeCache.FileStorage do
...
+
+    cond do
+      File.exists?(path, [:raw]) ->
+        stream_snapshot_chunk!(opts, chunk_number)
+
+      CubDB.has_key?(opts.db, @snapshot_meta_key) ->
+        []
+
+      true ->
+        Process.sleep(50)
```
msfstef, line 465: nit: alias the `50` to reuse between this and `total_wait_time + 50` for consistency?

msfstef (approved):
Very nice work! Love the ripping out of snapshot logic in the serve plug!

Left some comments and questions - as for the TS tests checking etags, we can modify them to be looking at the first "incomplete" actual log chunk rather than looking at `-1` (which used to collate the transaction log)

packages/sync-service/lib/electric/shape_cache/file_storage.ex
icehaunter: Empty enum (i.e. empty table for the shape) will not trigger the reduction function, but will trigger end-of-source-stream function here

msfstef: Ugh I missed the commas and got confused - checked docs for `Stream.transform/5` now

packages/elixir-client/lib/electric/client/offset.ex
icehaunter: Done

packages/react-hooks/test/react-hooks.test.tsx
icehaunter: Done

packages/sync-service/lib/electric/replication/log_offset.ex
icehaunter: I'm not sure we want to expose snapshot logic to offsets just yet, so I renamed to `is_virtual_offset`

packages/sync-service/lib/electric/shape_cache/shape_status.ex
icehaunter: Yeah, missed this when I introduced a helper

packages/sync-service/lib/electric/shape_cache/file_storage.ex
icehaunter: it's 3 lines apart, I think it's OK

PR #2122, opened 2024-12-09
chore (sync service): add logger metadata as Sentry tags

kevin-dp:
This PR adds logger metadata as Sentry tags such that errors in Sentry can be searched/filtered on this metadata.


robacourt (approved):
👍

PR #2123, opened 2024-12-09
chore: Linearlite+pglite example

samwillis:
Stacked on the rename of the existing linearlite example in order to maintain its history when merged #2132


examples/linearlite/server.ts
```diff
@@ -0,0 +1,101 @@
...
+  const content = await c.req.json()
+  let parsedChanges: ChangeSet
+  try {
+    parsedChanges = changeSetSchema.parse(content)
+    // Any additional validation of the changes can be done here!
+  } catch (error) {
+    console.error(error)
+    return c.json({ error: 'Invalid changes' }, 400)
+  }
+  const changeResponse = await applyChanges(parsedChanges)
```
kevin-dp, line 42: If `applyChanges` fails it will throw an error and we won't actually answer the HTTP request.
We should probably catch the error here and return an error response.

examples/linearlite/server.ts
```diff
@@ -0,0 +1,101 @@
...
+// Start the server
+const port = 3001
+console.log(`Server is running on port ${port}`)
+
+serve({
+  fetch: app.fetch,
+  port,
+})
+
+async function applyChanges(changes: ChangeSet): Promise<{ success: boolean }> {
```
kevin-dp, line 55: Why is this function declared to return a `{ success: boolean }` if we only ever return `{ success: true }`.
Either it should return `Promise<void>` and reject the promise when not successful, or it should return a boolean which is false on failure instead of throwing.

examples/linearlite/src/App.tsx
```diff
@@ -0,0 +1,203 @@
...
+      live,
+      sync: electricSync(),
+    },
+  })
+}
+
+const pgPromise = createPGlite()
+
+let syncStarted = false
+pgPromise.then(async (pg) => {
```
kevin-dp, line 48: No need for defining an async arrow function if we're not using `await`

thruflo (approved):
Approved from my side.

PR #2124, opened 2024-12-09
feat: Scalability chart widget for website

msfstef:
A graph widget for scaling metrics using Chart.js - currently using dummy data for development and iteration.

You can build the website in this branch to check the result, attaching a screenshot for ease of use:

image

### DIFF
--- a/website/blog/posts/2024-12-09-electric-beta-release.md
+++ b/website/blog/posts/2024-12-09-electric-beta-release.md
@@ -89,7 +89,9 @@ So many real-time sync systems demo well but break under real load.
 
 Electric has been [engineered from the ground up](/docs/api/http) to handle high-throughput workloads, like [Trigger.dev](https://trigger.dev/launchweek/0/realtime), with low latency and flat resource use. You can stream real-time data to **millions of concurrent users** from a single commodity Postgres:
 
-<ScalabilityChart />
+<figure>
+  <ScalabilityChart />
+</figure>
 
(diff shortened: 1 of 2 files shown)


thruflo:
Looking great, love it.

Some minor feedback:

- [ ] the write workloads are `1 req/s`, `8 req/s` and `16 req/s`. I guess with the middle as the default.
- [ ] capitalisation on the dropdown is inconsistent, suggest using `Sync service memory` format on the RHS list
- [ ] stick a `<figure />` element around the chart in the blog post flow, it'll give it some vspacing
- [ ] drop "coming soon" label from cloud section of benchmarking page
- [ ] ideally do a copy pass on that cloud section i.e.: instead of:

> Cloud benchmarks test the performance and scalability of Electric when running behind a CDN.

Maybe:

> Electric is designed to run behind a CDN, using the [CDN's request collapsing](/docs/api/http#collapsing-live-requests) capability to scale out to millions of concurrent users. The graph below shows the latency and compute resource of a single Electric server using this technique to handle millions of concurrent users.
>
> <figure> ... graph ... </figure>
>
> You can see the benchmarking fleet deployed to achieve this test in [<garry's folder location>](#).

thruflo:
Also, at least 1m clients!

msfstef:
> Also, at least 1m clients! Let's drop in the real data :)
Real data now available - although the graphs are looking very noisy...

>How dynamic are your axis labels? Because the data will vary across metric and workloads.
Units are adjusted per metric type, and they all are set to start at 0 on the y-axis and ticks are automatically set. If there's something specific you have in mind do tell!

PR #2128, opened 2024-12-09
Some styling fixups for the todo app

msfstef:
Some style fixups for the TODO app as requested. I didn't keep the actual Electric colors cause we are using shadcn themed components so I just went with a similar color. I also hardcoded dark mode as it looks better. Good enough for this IMO but let me know.

Screenshot 2024-12-09 at 18 23 42
image


examples/todo-app/src/main.tsx
```diff
@@ -2,10 +2,11 @@ import React from "react"
 import ReactDOM from "react-dom/client"
 import { createBrowserRouter, RouterProvider } from "react-router-dom"
 import ErrorPage from "./error-page"
-import "@radix-ui/themes/styles.css"
-import "../public/typography.css"
```
KyleAMathews, line 6: why did you delete this?

msfstef: typography CSS was conflicting with radix theme CSS and components were messed up (no borders etc)

KyleAMathews: Ok then we'll need to cleanup the plugin generating it as well

https://github.com/electric-sql/electric/blob/dae3b0d90da48ab4fb420a64f45be908168c4458/examples/todo-app/vite.config.ts#L10

thruflo:
Could you swap the background and green for the background and our brand pink/purple as per the write-patterns example:

image

thruflo (approved):

PR #2129, opened 2024-12-09
Write patterns deployment

balegas:
Deployment and dockers files for hosting Write patterns example.

I did some changes to load env vars and add db ids, etc.

There seems to be some problem with pattern 4, but I'm not sure if is just me.


balegas:
This should generally be working, but I wasn't able to test it. Opening PR for visibility and for someone to take over if necessary.

examples/write-patterns/sst.config.ts
```diff
@@ -0,0 +1,181 @@
...
+    const databaseUri = getNeonDbUri(project, db, false)
+    try {
+      databaseUri.apply(applyMigrations)
+
+      const electricInfo = databaseUri.apply((uri) =>
+        addDatabaseToElectric(uri)
+      )
+
+      const vpc = new sst.aws.Vpc(`write-patterns-${$app.stage}-vpc`)
+      const cluster = new sst.aws.Cluster(
```
KyleAMathews, line 47: let's reuse the our dev cluster from cloud

KyleAMathews (changes requested):

examples/write-patterns/sst.config.ts
balegas: Alright. Well have to do that though all examples

KyleAMathews: that'd be good — they're slow to setup & fairly expensive

examples/write-patterns/patterns/1-online-writes/index.tsx
```diff
@@ -19,6 +20,8 @@ export default function OnlineWrites() {
     url: `${ELECTRIC_URL}/v1/shape`,
     params: {
       table: 'todos',
+      database_id: import.meta.env.VITE_ELECTRIC_DATABASE_ID,
```
thruflo, line 23: The challenge here is that this complicates the example code just for our deployment purposes.

One request is please can you follow the pattern of declaring a `const `DATABASE_URL ...` next to `ELECTRIC_URL` above and providing a default value there. It keeps the actual shape definition cleaner.

balegas: I moved the params to the top but is not great. I suggest you change it to your taste.

In general is a bit of a pain that we need to set Cloud params explicitly. One possible way to hide is to pass Cloud in the base url, but I'm not sure if the code picks it up properly

examples/write-patterns/patterns/1-online-writes/index.tsx
```diff
@@ -3,7 +3,17 @@ import { v4 as uuidv4 } from 'uuid'
 import { useShape } from '@electric-sql/react'
 import api from '../../shared/app/client'
 
-const ELECTRIC_URL = import.meta.env.ELECTRIC_URL || 'http://localhost:3000'
+const ELECTRIC_URL =
```
thruflo, line 6: Maybe we could wrap this up into:

```ts
const { electricUrl, envParams } from '../../shared/app/config'
```

thruflo: I.e.: across all the patterns, given that it's a bunch of noise at the top of the code samples that are all displayed on the page in the Writes guide, etc.

thruflo (changes requested):

KyleAMathews (approved):
:shipit:

PR #2131, opened 2024-12-09
fix building @electric-sql/react

KyleAMathews:


balegas (approved):
Approving hopping this doesn’t break any of the examples.

PR #2133, opened 2024-12-09
Fix write-patterns schema.

thruflo:


thruflo:
This is fixed in #2129.

PR #2134, opened 2024-12-09
Simplify scalability chart

KyleAMathews:


thruflo (approved):
Looks great 👍 

One thing is that it would be nice to still display what the write workload is. Even if we pick one, it’s still useful to know that this is with a realistic write workload going on.

The other thing that we lose is the CPU. Perhaps we can just plot that on a separate graph on the actual benchmarks page. But for a summary for both blog post and benchmarks page to start with this is great to go ahead with.

KyleAMathews:
oh right yeah — the 960 writes / minute is important context — I'll add that back in. It'd be interesting too to list how many bytes / second that is at the 1 million client mark e.g. if the transactions are 5kb (fairly small updates) that's 80 GB / second we'd be pushing out.

KyleAMathews:
@magnetised do you know how big the transactions or shape updates are more-or-less?

PR #2150, opened 2024-12-11
chore (examples): deployment of proxy auth example

kevin-dp:
This PR configures the proxy-auth example with SST for deploying to the cloud.


KyleAMathews (approved):
Thanks!

PR #2151, opened 2024-12-11
elixir-client: ensure monitor + request is atomic

magnetised:
It's possible for a process to start a request monitor process but then exit before the actual http request process is started, leaving the monitor process stuck (but still registered under the request id).

In this case if a later process tries to make the same request it will find the stuck monitor and register itself with it but no response will arrive (since no request is running).

This moves the making of the request into the monitor process so it will always definitely happen.

I've also added extra process monitoring so that if the monitor crashes the caller will know about it.


packages/elixir-client/lib/electric/client/fetch/monitor.ex
```diff
@@ -11,32 +11,52 @@ defmodule Electric.Client.Fetch.Monitor do
...
-  def wait(ref) do
+  def wait({caller_monitor_ref, monitor_caller_ref}) do
     receive do
-      {:response, ^ref, response} -> response
+      {:response, ^monitor_caller_ref, response} ->
+        Process.demonitor(caller_monitor_ref, [:flush])
+        response
+
+      {:DOWN, ^caller_monitor_ref, :process, _pid, reason} ->
+        raise RuntimeError,
```
msfstef, line 58: Can this/should this be an `Electric.Client.Error`? We look for these in particular in some clauses elsewhere and it feels reasonable to expect this

msfstef (approved):
nice catch, looks good, needs a changeset!

Issue #2147, opened 2024-12-11
Exit prerelease mode for changeset


KyleAMathews:
It turns out we don't want this as it doesn't use the latest tag for npm. It's only if you're doing prereleases alongside normal releases.

PR #2152, opened 2024-12-11
Explicitly set releases to use latest tag

KyleAMathews:
Fix https://github.com/electric-sql/electric/issues/2147

https://github.com/changesets/changesets/blob/main/docs/prereleases.md


KyleAMathews:
No :-D not until now anyways — where I learned we do need to be in prerelease mode as otherwise a patch version will just remove the beta and go to a `1.0.0` release. We can get around the `beta` tag problem by just explicitly specifying that the `latest` tag is to be used when publishing.

icehaunter (approved):

Issue #1831, opened 2024-10-10
Filter replication publication columns


msfstef:
Following https://github.com/electric-sql/electric/issues/1804 and corresponding PR https://github.com/electric-sql/electric/pull/1829

We want to alter the publication to only send over the selected/required columns over the replication stream, which would be the selected columns via the `column` parameter, and potentially also any columns referenced in the `where` clause if specified.

While PG does support partial replication of a table (at least for PG15 and above I think), there are [some warnings in the docs](https://www.postgresql.org/docs/15/logical-replication-col-lists.html#LOGICAL-REPLICATION-COL-LIST-COMBINING):

> it is [...] possible to get into [a problematic] situation by adding or altering column lists on the publication side after a subscription has been created.
>This means changing the column lists of tables on publications that are already subscribed could lead to errors being thrown on the subscriber side.

This implies that even with a single publication and subscription, altering the column lists of tables might lead to errors - this needs to be investigated.

balegas:
After discussing, we think the warning does not affect us, since we're impersonating the protocl and we only have one subscriber for one publication anyways.

msfstef:
Short update on issue that I ran into with this:

It seems that we cannot use `REPLICA IDENTITY` as `FULL`, which always sends the old column values of updated rows, if we have column filtering on the publication (see [docs](https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-REPLICA-IDENTITY))

The error that I get if I do is:
```
** (Postgrex.Error) ERROR 42P10 (invalid_column_reference) cannot update table "wide_table"
Column list used by the publication does not cover the replica identity.
```

balegas:
We have discussed down publication where clause filters as we've seen that it is actually slower than doing the filtering in Elixir (...maybe not in all cases, we'd have to check), but if we do this than we'd be good for that issue.

Can you describe the other issue you mentioned in our call?

msfstef:
The other issue with filtering columns, which requires setting `REPLICA IDENTITY DEFAULT`, is the following:

Suppose you have a table `foo (id, value1, value2, value3)`
1. Create a Shape A that selects columns `id,value1`
2. Create a Shape B that selects columns `id,value2`
3. At this point the publication would only replicate `id,value1,value2`, so `REPLICA IDENTITY` is set to `DEFAULT` and thus update statements only carry the old values of the primary key cols (i.e. old `id` values)
4. An update is made to a row with `id=1` such that only `value2` is affected
[...]

PR #2154, opened 2024-12-11
feat(sync-service): Clean up publication filters

msfstef:
Closes https://github.com/electric-sql/electric/issues/1774

This work started to introduce column filters (see https://github.com/electric-sql/electric/issues/1831) but ended up on a road block because of us using `REPLICA IDENTITY FULL` - however the work also takes care of cleaning up filters.

- Introduced singular process for updating publication - we were locking on it before anyway, might as well linearise it ourselves.
- Process maintains reference counted structure for the filters per relation, including where clauses and filtered columns, in order to produce correct overall filters per relation
- Update to the publication is debounced to allow batching together many shape creations
- Every update does a complete rewrite of the publication filters so they are maintained clean - but also introduced a `remove_shape` call so that if electric remains with no shapes it should also have no subscriptions to tables.

## TODOs
- [ ] Write tests for `PublicationManager`
- [ ] Write procedure for recovering in-memory state from `shape_status.list_shapes` in `recover_shapes`
- [ ] Split where clauses at top-level `AND`s to improve filter optimality (suggested be @icehaunter )


msfstef:
Regarding https://github.com/electric-sql/electric/issues/1831 - we need to figure out _when_ we need `REPLICA IDENTITY FULL` and make it so that it is only set when necessary (and column filters are removed in that case)

Once we determine if we can afford not to use `REPLICA IDENTITY FULL` at all times then we should be able to do column filtering easily with these changes.

msfstef:
benchmark this

robacourt (approved):
Nice work 👍

msfstef:
benchmark this

packages/sync-service/lib/electric/replication/eval/expr.ex
```diff
@@ -15,4 +15,26 @@ defmodule Electric.Replication.Eval.Expr do
...
+      iex> used_refs = %{["id"] => :int8, ["created_at"] => :timestamp}
+      iex> current_table_refs(%Expr{query: "id = 1", used_refs: used_refs})
+      ["created_at", "id"]
+
+      iex> used_refs = %{["id"] => :int8, ["potato", "created_at"] => :timestamp}
+      iex> current_table_refs(%Expr{query: "id = 1", used_refs: used_refs, returns: :int8})
+      ["id"]
+  """
+  @spec current_table_refs(t()) :: [String.t()]
+  def current_table_refs(%__MODULE__{used_refs: used_refs}) do
```
icehaunter, line 34: nitpick: (nerd voice) well technically if there were any joins an unqualified table name would not be enough to disambiguate columns, so the proper function name is `unqualified_refs`

msfstef:
benchmark this

msfstef:
benchmark this

PR #2155, opened 2024-12-12
chore: make sure .tool-versions matches package.json

icehaunter:
We always should use `.tool-versions` as the authoritative source across the toolchain


msfstef (approved):
Nice, might be worth removing the version specification from `package.json`? unless it serves a functional purpose, I don't know

icehaunter:
@msfstef I don't know why @KyleAMathews added it in the first place, so we can remove in a separate PR

kevin-dp (approved):

PR #2156, opened 2024-12-12
fix: truncates no longer crash replication

icehaunter:
We had an issue where a truncate caused the `Consumer` process to `GenServer.call` `ShapeCache`, which in turn `GenServer.call`-ed the `Consumer` and then both crashed with a timeout and weren't cleaned up correctly. This addresses that and makes the shape process after the Truncate clean up and stop gracefully, test included.


msfstef (approved):
Looks good, nice catch!

PR #2157, opened 2024-12-12
Fix elixir-client race condition

magnetised:
Fixes https://electricsql-04.sentry.io/issues/14068898/?project=4508410462404688&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=2


icehaunter (approved):
Great find

PR #2161, opened 2024-12-12
feat: Convert `waiting` and `starting` status to 202

msfstef:
Before we were returning 503s for anything other than `active` status - for some health check services (like AWS Fargate) you can only rely on the status code, and technically Electric in `waiting` and `active` status is ready to accept requests, just not service them.

I've changed it so that:
- `waiting` and `starting` return 202, requests are accepted and held but failed after a timeout if not activated
- `active` returns 200
- `stopping` returns 503, we haven't implemented it yet anyway

Not sure if we should consider this a breaking change and how we handle releases at the moment @KyleAMathews .

### DIFF
--- a/.changeset/lucky-cats-teach.md
+++ b/.changeset/lucky-cats-teach.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Return `202` for `waiting` and `starting` health status - accepts requests but will fail to service them.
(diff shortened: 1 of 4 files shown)


KyleAMathews (approved):
Seems more like a bug fix than a breaking change so :shipit:

msfstef:
Added a [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck) to the dockerfile as well, so that we leverage Docker's "healthy" status too

PR #2163, opened 2024-12-12
Add cpu and memory usage info

magnetised:
and tidy up telemetry config a little and make the call-home url configurable

### DIFF
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -1,14 +1,15 @@
 import Config
 import Dotenvy
 
+alias Electric.ConfigParser
+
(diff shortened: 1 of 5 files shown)


alco (approved):
Nice work! :heart: 

I have refactored the configuration a bit more and added a touch of humanity on top. Curious to hear your thoughts.

alco:
Oh, and by the way, I have added the builtin telemetry_poller metrics to the list of periodic measurements in #2167. We're not currently collecting those anywhere even though we have the metric definitions for them.

magnetised:
@alco you went for it huh. Great stuff - thanks

PR #2164, opened 2024-12-12
fix: Remove `--start-interval` as it is a relatively recent docker feature

msfstef:
It means it will be slightly slower to turn `healthy` but that's ok


KyleAMathews (approved):
:shipit:

PR #2165, opened 2024-12-13
Set OTEL span status to error when an exception is recorded for it

### LINKED ISSUE #2027: For opentelemetry spans — we need to set the status to error in addition to recording an exception
If there was an actual fatal error, just recording an exception doesn't mark the span as having errored. It's important to do that as otel highlight errored spans whereas exceptions aren't as prominent.

alco:
Closes #2027.


KyleAMathews (approved):
Thanks!

PR #2167, opened 2024-12-14
Add missing metric gathering to the telemetry poller config

alco:
Follow-up to #2163.

### DIFF
--- a/packages/sync-service/lib/electric/telemetry.ex
+++ b/packages/sync-service/lib/electric/telemetry.ex
@@ -196,12 +196,16 @@ defmodule Electric.Telemetry do
     [
       last_value("system.cpu.core_count"),
       last_value("system.cpu.utilization.total"),
+      last_value("electric.postgres.replication.wal_size", unit: :byte),
       last_value("electric.storage.used", unit: {:byte, :kilobyte}),
       last_value("electric.shapes.total_shapes.count"),
       last_value("vm.memory.total", unit: :byte),
       last_value("vm.memory.processes_used", unit: :byte),
       last_value("vm.memory.binary", unit: :byte),
       last_value("vm.memory.ets", unit: :byte),
+      last_value("vm.system_counts.process_count"),
+      last_value("vm.system_counts.atom_count"),
+      last_value("vm.system_counts.port_count"),
       last_value("vm.total_run_queue_lengths.total"),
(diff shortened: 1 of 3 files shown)


packages/sync-service/lib/electric/telemetry.ex
```diff
@@ -204,7 +207,9 @@ defmodule Electric.Telemetry do
     stack_id = Keyword.fetch!(opts, :stack_id)
 
     [
-      # A module, function and arguments to be invoked periodically.
+      :memory,
```
icehaunter, line 210: Why do we need that? the memory stat collection already works (for sure it does bc it's what I use in benchmarks)

icehaunter: so do RQL for that matter, and I'm assuming system counts too. Adding `last_value` to the prometheus is good though

alco: @icehaunter 

I've double-checked the source code of `telemetry_poller` and it definitely doesn't try to inject its default measurements into an explicitly provided list. It only measures what we give to it.

We can collect memory and other default measurements because the `telemetry_poller` OTP application starts its own supervisor and a default poller process. Looks like this is unintended and we should disable it by passing `runtime: false` to the dep in `mix.exs`? In that case, the addition of `:memory` and other default measurements in this PR should remain.

icehaunter: What do you mean by "unintended"? It's in the [docs](https://hexdocs.pm/telemetry_poller/telemetry_poller.html) for the `telemetry_poller`:

> We will discuss each measurement in detail. Also note that the telemetry_poller application ships with a built-in poller that measures memory, total_run_queue_lengths and system_counts. This takes the VM measurement out of the way so your application can focus on what is specific to its behaviour.

icehaunter: I disagree given it's a default, documented, and expected behaviour and pattern, and I added the poller knowing that those metrics will be available, and a separate process is trivially cheap. But if you insist, go ahead.

alco: I insist on having all periodic measurements defined in a single place, so I went ahead and disabled the default telemetry poller and added a comment explaining that to our Telemetry module. Thanks for the discussion!

icehaunter (approved):

PR #2169, opened 2024-12-14
linearlite: make transaction batching configurable when loading seed data

alco:
This allows for easier benchmarking of transaction processing in Electric.


samwillis (approved):
LGTM 👍

PR #2173, opened 2024-12-16
Simplify configuration of the electric app

magnetised:
Refactor app config to make most values optional.

Except for connection_opts, all the config settings have defaults. When embedding electric inside another elixir app, our runtime.exs won't be sourced so the developer would be forced to unnecessarily set all these configuration values.

Moves the default values from runtime.exs to the point of usage so we don't have defaults in two places.

### DIFF
--- a/packages/sync-service/README.md
+++ b/packages/sync-service/README.md
@@ -68,10 +68,10 @@ reuse that configuration if you want:
 
 Or if you're getting your db connection from an environment variable, then you
 can use
-[`Electric.ConfigParser.parse_postgresql_uri!/1`](https://hexdocs.pm/electric/Electric.ConfigParser.html#parse_postgresql_uri!/1):
+[`Electric.Config.parse_postgresql_uri!/1`](https://hexdocs.pm/electric/Electric.Config.html#parse_postgresql_uri!/1):
 
     # config/*.exs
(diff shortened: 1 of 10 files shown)


magnetised:
@alco while you were writing that comment I was busily addressing some of your concerns in parallel, motivated by those same concerns. I've moved the defaults to a single place and reference them within the docs so there's less chance of the two drifting.

If we're to make embedding electric straightforward, we can't rely on `Application.fetch_env/2` because that requires the developer setting values for all the config, and there's a lot, which a) makes embedding electric more difficult and error-prone and b) means that if we evolve our defaults for whatever reason embedded instances won't get them, they'll be stuck on whatever value the developer copy-pasted when bringing electric in.

In `runtime.exs` we could use those defaults but I don't see why -- the defaults will be applied if they're not set and it just makes runtime.exs more noisy. 

Remember that if we're to make electric easy to embed we can't rely on our runtime.exs to do any of the work. I think it would actually be better to remove it from this repo so that we stop relying on it except we need it in our docker builds.

alco:
@magnetised Much better. Though still a bit quirky due to the convoluted "look up OS env -> set OTP env if missing -> replace the `nil` from OTP env with a default" lookup route.

One step closer to the ideal, so kudos!

alco (approved):

PR #2176, opened 2024-12-17
chore: Add pg17 to test matrix

msfstef:
Had to trim the unique test db names, for some reason Postgrex was failing otherwise with pg17 (used short hashes to improve uniqueness guarantee)

### DIFF
--- a/packages/sync-service/test/support/db_setup.ex
+++ b/packages/sync-service/test/support/db_setup.ex
@@ -13,7 +13,14 @@ defmodule Support.DbSetup do
     {:ok, utility_pool} = start_db_pool(base_config)
     Process.unlink(utility_pool)
 
-    db_name = to_string(ctx.test)
+    full_db_name = to_string(ctx.test)
+
+    db_name_hash =
+      full_db_name |> :erlang.phash2(9999) |> to_string() |> String.pad_leading(4, "0")
+
+    # Truncate the database name to 63 characters, use hash to guarantee uniqueness
+    db_name = "#{db_name_hash} ~ #{String.slice(full_db_name, 0..54)}"
+
     escaped_db_name = :binary.replace(db_name, ~s'"', ~s'""', [:global])
 
     Postgrex.query!(utility_pool, "DROP DATABASE IF EXISTS \"#{escaped_db_name}\"", [])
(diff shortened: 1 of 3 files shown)


packages/sync-service/test/support/db_setup.ex
```diff
@@ -13,7 +13,14 @@ defmodule Support.DbSetup do
     {:ok, utility_pool} = start_db_pool(base_config)
     Process.unlink(utility_pool)
 
-    db_name = to_string(ctx.test)
+    full_db_name = to_string(ctx.test)
+
+    db_name_hash =
+      full_db_name |> :erlang.phash2(9999) |> to_string() |> String.pad_leading(4, "0")
```
robacourt, line 19: 10k range doesn't seem very big, we have ~500 tests already so that's a 1 in 20 chance of a collision. Ok, they don't all have their own database, but still. Perhaps a higher range and base64 encode to keep the string shorter?

robacourt: Actually, chances of collision are higher than that, it's 1/10k + 2/10k + .. + 499/10k ... but then minus multiple collisions.... anyway a bit of spreadhsheeting came up with a 99.99969% chance of collision!

robacourt (changes requested):
Nice. Please reduce hash collisions though!

packages/sync-service/test/support/db_setup.ex
msfstef: The hashes can collide, but I also include part of the test name, and not all tests use a PG db - but I can increase the part that is taken up by the hash

I avoided encoding the test name to make it relatively recognizable

robacourt: And I know you are appending part of the name of the test, but the test names can be similar and we only need 15 tests to make a 1% chance of a collision

packages/sync-service/test/support/db_setup.ex
```diff
@@ -13,7 +13,17 @@ defmodule Support.DbSetup do
...
     Process.unlink(utility_pool)
 
-    db_name = to_string(ctx.test)
+    full_db_name = to_string(ctx.test)
+
+    db_name_hash =
+      full_db_name
+      |> :erlang.phash2(64 ** 5)
+      |> :binary.encode_unsigned()
+      |> Base.encode64()
```
robacourt, line 22:
```suggestion
      |> Base.encode64()
      |> String.replace_trailing("==","")
```

robacourt (approved):
Nice!

PR #2177, opened 2024-12-17
add params map to client config

### LINKED ISSUE #2175: Support additional query params in Elixir client
Currently, the Elixir client only supports a number of predefined options which are often passed through env vars such as `ELECTRIC_URL`, `ELECTRIC_DATABASE_ID`, etc. However, it does not currently support `ELECTRIC_TOKEN` which is required when deploying apps to the cloud. I would suggest to add a generic mechanism as in the TypeScript client which allows users to pass arbitrary parameters to be added as query parameters:

```ex
Client.new(base_url: "...", params: %{"token" => "value"})
```

magnetised:
remove explicit database_id setting from client

Fixes #2175


kevin-dp (approved):
LGTM and seems to work fine! 👍

PR #2180, opened 2024-12-18
fix(electric,elixir-client): Write static version string to mix.exs prior to publishing

### LINKED ISSUE #2179: enhancement/bug: electric version falls back to 0.0.0 when used as hex package
When adding electric to an existing elixir project its version will be set to 0.0.0, due to the functions(load from env or from package.json which is not included in hex package) in the mix.exs.

This can lead to the following:
```text
Unchecked dependencies for environment dev:
* electric (Hex package)
  the dependency does not match the requirement "~> 1.0.0-beta.1", got "0.0.0"
** (Mix) Can't continue due to errors on dependencies
```

one can maybe provide the version again via an env var - but this is unintuitive and also repetitive 😅

alco:
Fixes #2179


icehaunter:
I'm not against this approach, but in `elixir-client` it's solved just by including `package.json` in [published filies](https://github.com/electric-sql/electric/blob/main/packages/elixir-client/mix.exs#L80). Maybe it's easier to just do that here too?

magnetised:
I solved this for the client but forgot about the top-level package. Including the package.json in the package files is a much simpler and more robust fix -- I deliberately avoided doing any source-code munging because it scares me.

if we just add `files:` to the `package()` output and include `package.json` there it works perfectly:

```elixir
  defp package do
    [
      licenses: ["Apache-2.0"],
      links: %{
        "Electric SQL" => "https://electric-sql.com",
        "Github" => @github_repo
      },
      files: ~w(lib .formatter.exs mix.exs README.md LICENSE package.json)
    ]
  end
```

magnetised:
@alco happy to finish this since I feel responsible ... 😊

balegas:
@magnetised can you finish this one?

icehaunter (approved):

PR #2182, opened 2024-12-18
(website) Add throughput benchmarks

robacourt:


website/docs/reference/benchmarks.md
```diff
@@ -163,6 +170,51 @@ In this benchmark there are a varying number of shapes with just one client subs
...
+
+<figure>
+  <a :href="ReplicationThroughputOptimised">
+    <img :src="ReplicationThroughputOptimised"
+        alt="Benchmark measuring how many writes per second Electric can process"
+    />
+  </a>
+</figure>
+
+This benchmark measures how long each write takes to process with a varying number of shapes. Each shape in this benchmark
```
thruflo, line 183: I think this description needs a clearer statement of what an optimised where clause is vs a non-optimised clause. You could even call this out to an info panel e.g.:

```markdown
> [!Tip] Optimised where clauses
> ... describe here ...
```

From the current description I read (towards the end of the para) that I can think of it as "analogous to having an index on the field" but really I would like to know what the term means when you're first using it.

thruflo (changes requested):
Nice work, thanks!

Also, I'm loving the new dark theme for the images :)

website/docs/reference/benchmarks.md
robacourt: Good idea. Done.

PR #2185, opened 2024-12-18
(website) Update shape docs to include details of where clause optimisation

robacourt:
Following on from #2182 on @thruflo 's suggestion, this PR add's where clause optimisation documentation to the main docs and links to it from various places. This includes what types of where clauses are optimised.


thruflo:
@robacourt if you don't mind I'm just going to do a quick copy pass on this. Just to ask:

> For optimised where clauses, we can process 4000 - 6000 row changes per second, regardless of how many shapes there are. For non-optimised where clauses, we can process 140k row changes per second, per shape. For more details see the [benchmarks](/docs/reference/benchmarks#_7-write-throughput-with-optimised-where-clauses).

Is this correct? Because it's a bit confusing to me. The copy refers to millions of where clauses at once but then we can only handle 4000 - 6000 row changes per second. Whereas non-optimised we can do 140k?

Ideally I'd like to express the total or per shape throughput so that it's higher for optimised where clauses.

PR #2186, opened 2024-12-19
chore (website): add deployed examples to demos on website

kevin-dp:
This adds all of the deployed examples as demos on the website.


thruflo (changes requested):
Great, thanks for doing this 👍

Just one thing is that I notice you’ve added demo: true and homepage: true to all of them.

The idea is that demo: true distinguishes interesting visual demos from technical examples. Just because it’s deployed doesn’t mean it should be a “demo”.

And for the homepage we should only have our three most interesting demos featured.

kevin-dp:
Ok, wasn't sure about that. Will fix that

PR #2187, opened 2024-12-19
Fix PR benchmark images

robacourt:


robacourt:
benchmark this

KyleAMathews:
heh, those benchmarks are proving we need to do another release haha. Which I'll get on today.

PR #2188, opened 2024-12-19
feat(ts-client): Start stream after subscribe

msfstef:
Addresses part of https://github.com/electric-sql/electric/issues/1997

Had to change some tests around - obviously this is not a full move towards the desired behaviour, e.g. we might even want for the stream to stop when it has _no_ subscribers, but I think this provides some beneficial properties to the stream (e.g. if subscribing to it after async gap no messages lost)

If we think this just complicates the client and that if we want to address the issue we should do it in one comprehensive go, I'm happy to close this as well

### DIFF
--- a/.changeset/mean-humans-clean.md
+++ b/.changeset/mean-humans-clean.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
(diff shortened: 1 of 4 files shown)


KyleAMathews (approved):
Looks good!

Incremental changes ftw

PR #2189, opened 2024-12-19
Shut down the connection supervisor in case of an unrecoverable replication error

### LINKED ISSUE #2178: Don't stop the system when we exceed slot size
https://github.com/electric-sql/electric/commit/077c6090c923f4734bdfc0b32d4b201b1b768c9e#diff-83f5e4e15850784d6499417614cf3a15ce6b4ee41a5b1122d2e6a7ef2ca04c08R235

We currently use `System.exit/1` which stops the Erlang node, but we should instead limit this shutdown to a particular tenant

alco:
The StackSupervisor will shut itself down as a consequence.

In a multi-tenant setup, this effectively shuts down a single tenant while allowing the other tenants to keep chugging along.

Closes #2178.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -649,6 +649,12 @@ defmodule Electric.Connection.Manager do
          } = error,
          state
        ) do
+    Electric.StackSupervisor.dispatch_stack_event(
+      state.stack_events_registry,
+      state.stack_id,
+      {:database_slot_invalidated, %{error: error}}
+    )
+
     # Perform supervisor shutdown in a task to avoid a circular dependency where the manager
     # process is waiting for the supervisor to shut down its children, one of which is the
     # manager process itself.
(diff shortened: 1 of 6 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -573,23 +577,25 @@ defmodule Electric.Connection.Manager do
...
-         } = error
+         } = error,
+         state
        ) do
-    System.stop(1)
-    exit(error)
+    # Perform supervisor shutdown in a task to avoid a circular dependency where the manager
+    # process is waiting for the supervisor to shut down its children, one of which is the
+    # manager process itself.
+    Task.start(Electric.Connection.Supervisor, :shutdown, [state.stack_id, error])
```
icehaunter, line 593: Great job, but let's `dispatch_stack_event` here too, so that the control plane can be aware of what's happened and why this is down

icehaunter (changes requested):

icehaunter (approved):

PR #2197, opened 2024-12-19
Add partitioned tables support

### LINKED ISSUE #2118: Support shapes on partitioned tables
When trying to use electric on a table that is partitioned:

```sql
create table runs (
  id bigint,
  task_id bigint not null,
  payload text,
  status text,
  created_at timestamp with time zone default now(),
  primary key (id, created_at)
)
partition by
  range (created_at);

create table runs_2023 partition of runs for
values
from
  ('2023-01-01') to ('2024-01-01');

create table runs_2024 partition of runs for
values
from
  ('2024-01-01') to ('2025-01-01');
```

If you try and subscribe to a shape using the `run` table, the following error occurs:

```
2024-12-07 21:10:43 21:10:43.713 [info] Query String: offset=-1&table=public.runs
2024-12-07 21:10:43 
2024-12-07 21:10:43 21:10:43.714 [error] GenServer {:"Elixir.Electric.ProcessRegistry:single_stack", {Electric.Postgres.Inspector.EtsInspector, nil}} terminating
2024-12-07 21:10:43 ** (MatchError) no match of right hand side value: []
2024-12-07 21:10:43     (electric 0.9.4) lib/electric/postgres/inspector/direct_inspector.ex:23: Electric.Postgres.Inspector.DirectInspector.load_relation/2
2024-12-07 21:10:43     (electric 0.9.4) lib/electric/postgres/inspector/ets_inspector.ex:114: Electric.Postgres.Inspector.EtsInspector.handle_call/3
2024-12-07 21:10:43     (stdlib 6.0.1) gen_server.erl:2209: :gen_server.try_handle_call/4
2024-12-07 21:10:43     (stdlib 6.0.1) gen_server.erl:2238: :gen_server.handle_msg/6
2024-12-07 21:10:43     (stdlib 6.0.1) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
2024-12-07 21:10:43 Process Label: {:ets_inspector, "single_stack"}
2024-12-07 21:10:43 Last message (from #PID<0.2976.0>): {:load_relation, "public.run"}
2024-12-07 21:10:43 
2024-12-07 21:10:43 21:10:43.717 [error] ** (Plug.Conn.WrapperError) ** (ArgumentError) cannot fetch key "handle" from conn.query_params because they were not fetched. Call Plug.Conn.fetch_query_params/2, either as a plug or directly, to fetch it
2024-12-07 21:10:43     (plug 1.16.1) lib/plug/conn/unfetched.ex:35: Plug.Conn.Unfetched.raise_unfetched/3
2024-12-07 21:10:43     (elixir 1.17.2) lib/access.ex:322: Access.get/3
2024-12-07 21:10:43     (electric 0.9.4) lib/electric/plug/serve_shape_plug.ex:654: Electric.Plug.ServeShapePlug.end_telemetry_span/2
2024-12-07 21:10:43     (electric 0.9.4) lib/electric/plug/serve_shape_plug.ex:689: Electric.Plug.ServeShapePlug.handle_errors/2
2024-12-07 21:10:43     (plug 1.16.1) lib/plug/error_handler.ex:113: Plug.ErrorHandler.__catch__/6
2024-12-07 21:10:43     (electric 0.9.4) deps/plug/lib/plug/router.ex:246: anonymous fn/4 in Electric.Plug.Router.dispatch/2
2024-12-07 21:10:43     (telemetry 1.2.1) /app/deps/telemetry/src/telemetry.erl:321: :telemetry.span/3
2024-12-07 21:10:43     (electric 0.9.4) deps/plug/lib/plug/router.ex:242: Electric.Plug.Router.dispatch/2
2024-12-07 21:10:43     (electric 0.9.4) lib/electric/plug/router.ex:1: Electric.Plug.Router.plug_builder_call/2
2024-12-07 21:10:43     (bandit 1.5.5) lib/bandit/pipeline.ex:124: Bandit.Pipeline.call_plug!/2
2024-12-07 21:10:43     (bandit 1.5.5) lib/bandit/pipeline.ex:36: Bandit.Pipeline.run/4
2024-12-07 21:10:43     (bandit 1.5.5) lib/bandit/http1/handler.ex:12: Bandit.HTTP1.Handler.handle_data/3
2024-12-07 21:10:43     (bandit 1.5.5) lib/bandit/delegating_handler.ex:18: Bandit.DelegatingHandler.handle_data/3
2024-12-07 21:10:43     (bandit 1.5.5) /app/deps/thousand_island/lib/thousand_island/handler.ex:411: Bandit.DelegatingHandler.handle_continue/2
2024-12-07 21:10:43     (stdlib 6.0.1) gen_server.erl:2163: :gen_server.try_handle_continue/3
2024-12-07 21:10:43     (stdlib 6.0.1) gen_server.erl:2072: :gen_server.loop/7
2024-12-07 21:10:43     (stdlib 6.0.1) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
```

Since queries against the `runs` table (`SELECT * FROM runs`) works across partitions, it seems like it should be possible for electric to perform the same "magic".

Currently, electric does work if the partition table name is used instead of the parent table (e.g.  `runs_2024`), but then unfortunately shapes cannot really travel across partitions. This could be a real issue if partitions are small (e.g. every minute).

magnetised:
Allows for subscribing to partitions of a partitioned table (which you could already do) but also subscribing to the root, partitioned, table and receiving all updates across all partitions.

Fixes #2118 

Major changes:

- Add support for partitioned table types in the various inspector queries
- Include partition hierarchy information in relation information from pg and in inspector relation info
- Force reload of table information when receiving a new relation that is a partition
- Add partition information to `Filter` so that it can correctly pass on a change on a partition table to a shape on the partition root.
- For a shape subscribed to the partition root, changes coming in on partitions will be re-written so that their relation matches the partition root. The change `key`s are not being re-written - I don't think this is an issue as the keys are guaranteed to be unique and that's enough.
- Relation messages notifying of a new partition table are forwarded onto root-partition consumers so the shape can be updated with knowledge of the new partition (this does not cause the shape to be deleted)
- I've added a full-stack test to verify this stuff so that we're not making assumptions about pg's handling of partitions

Minor:

- support querying the inspector based on the {schema, name} relation tuple, as well as the table name string because I needed to be able to look up table info based
- Filter instances now have an inspector instance so they can do table info lookups to get partition information
- The filter functions used by the dispatcher now return `{filter, shape_ids}` so that the filter's state can be updated
- I've tweaked StubInspector so we can return richer info from the load_relation call
- Added missing serialisation of a shape's replica setting

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -58,7 +58,8 @@ defmodule Electric.Replication.ShapeLogCollector do
     state = Map.merge(opts, %{producer: nil, subscriptions: {0, MapSet.new()}})
     # start in demand: :accumulate mode so that the ShapeCache is able to start
     # all active consumers before we start sending transactions
-    {:producer, state, dispatcher: Electric.Shapes.Dispatcher, demand: opts.demand}
+    {:producer, state,
+     dispatcher: {Electric.Shapes.Dispatcher, inspector: state.inspector}, demand: opts.demand}
   end
 
   def handle_subscribe(:consumer, _opts, from, state) do
@@ -148,12 +149,41 @@ defmodule Electric.Replication.ShapeLogCollector do
 
     OpenTelemetry.add_span_attributes("rel.is_dropped": true)
 
+    reload_partitioned_table(rel, state)
+
     {:reply, :ok, [], state}
   end
 
   defp handle_relation(rel, from, state) do
     OpenTelemetry.add_span_attributes("rel.is_dropped": false)
+    reload_partitioned_table(rel, state)
     {:noreply, [rel], %{state | producer: from}}
+  end
+
+  defp reload_partitioned_table(rel, state) do
+    case Inspector.load_relation(rel, state.inspector) do
+      {:ok, %{parent: nil}} ->
+        :ok
+
+      {:ok, %{parent: {_, _} = parent}} ->
+        # probably a new partition for an existing partitioned table
+        # so force a reload of the relation info
+
+        # TODO: we should probabaly have a way to clean the inspector cache
+        # just based on the relation, there's a chance that this results in
+        # a query to pg just to then drop the info
+        with {:ok, info} <- Inspector.load_relation(parent, state.inspector) do
+          Inspector.clean(info, state.inspector)
+        end
+
+      {:ok, _} ->
+        # probably a malformed value from a test inspector
+        :ok
+
+      {:error, _} ->
+        # just ignore errors here, they're unlikely anyway
+        :ok
+    end
   end
 
   defp remove_subscription(from, %{subscriptions: {count, set}} = state) do
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -177,11 +177,8 @@ defmodule Electric.Replication.ShapeLogCollector do
         end
 
       {:ok, _} ->
-        # probably a malformed value from a test inspector
-        :ok
-
-      {:error, _} ->
-        # just ignore errors here, they're unlikely anyway
+        # probably a malformed value from a mock test inspector that isn't
+        # returning the full inspector result with a `:parent` key
         :ok
     end
   end
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -149,38 +149,12 @@ defmodule Electric.Replication.ShapeLogCollector do
 
     OpenTelemetry.add_span_attributes("rel.is_dropped": true)
 
-    reload_partitioned_table(rel, state)
-
     {:reply, :ok, [], state}
   end
 
   defp handle_relation(rel, from, state) do
     OpenTelemetry.add_span_attributes("rel.is_dropped": false)
-    reload_partitioned_table(rel, state)
     {:noreply, [rel], %{state | producer: from}}
-  end
-
-  defp reload_partitioned_table(rel, state) do
-    case Inspector.load_relation(rel, state.inspector) do
-      {:ok, %{parent: nil}} ->
-        :ok
-
-      {:ok, %{parent: {_, _} = parent}} ->
-        # probably a new partition for an existing partitioned table
-        # so force a reload of the relation info
-
-        # TODO: we should probabaly have a way to clean the inspector cache
-        # just based on the relation, there's a chance that this results in
-        # a query to pg just to then drop the info
-        with {:ok, info} <- Inspector.load_relation(parent, state.inspector) do
-          Inspector.clean(info, state.inspector)
-        end
-
-      {:ok, _} ->
-        # probably a malformed value from a mock test inspector that isn't
-        # returning the full inspector result with a `:parent` key
-        :ok
-    end
   end
 
   defp remove_subscription(from, %{subscriptions: {count, set}} = state) do
--- a/packages/sync-service/lib/electric/shapes/filter.ex
+++ b/packages/sync-service/lib/electric/shapes/filter.ex
@@ -10,6 +10,7 @@ defmodule Electric.Shapes.Filter do
   the table specific logic to the `Filter.Table` module.
   """
 
+  alias Electric.Postgres.Inspector
   alias Electric.Replication.Changes
   alias Electric.Replication.Changes.DeletedRecord
   alias Electric.Replication.Changes.NewRecord
@@ -20,15 +21,20 @@ defmodule Electric.Shapes.Filter do
   alias Electric.Shapes.Filter
   alias Electric.Shapes.Filter.Table
   alias Electric.Shapes.Shape
+
   require Logger
 
-  defstruct tables: %{}
+  @enforce_keys [:inspector]
+  defstruct [:inspector, tables: %{}, partitions: %{}, partition_ownership: %{}]
 
   @type t :: %Filter{}
   @type shape_id :: any()
 
-  @spec new() :: Filter.t()
-  def new, do: %Filter{}
+  @spec new(keyword()) :: Filter.t()
+  def new(opts) do
+    {:ok, inspector} = Keyword.fetch(opts, :inspector)
+    %Filter{inspector: inspector}
+  end
 
   @doc """
   Add a shape for the filter to track.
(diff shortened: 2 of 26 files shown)


magnetised:
Think i have a more up-to-date version of this on my laptop which I seem to have forgotten to push or have over-written with my old work-machine version. @robacourt please hold off review until I can recover the situation

magnetised:
@robacourt situation recovered. review away. great start to the year!

balegas:
Nice work! We should document the two possible ways to subscribe to partitioned tables.

magnetised:
> Nice work! We should document the two possible ways to subscribe to partitioned tables.

@balegas I've added basic docs highlighting the behaviour of shapes on partitioned tables - pls review

robacourt (changes requested):
Nice work! I've left a few comments.

I've also been thinking there might be an alternative implementation, whereby the replication steam is processed (by a PartitionTransformer for want of a better name) before it gets to the filter, and if it sees a change for a partition it inserts a change into the stream for the logical table. That way, all the partition logic can go in the PartitionTransformer. Shape, Filter, Consumer etc. would remain unchanged. Sure it will create some changes that might not be used, but that seems a good tradeoff for the simplicity. Is there a reason why this is not possible? Or undesirable?

website/docs/guides/shapes.md
```diff
@@ -56,6 +56,36 @@ This is the root table of the shape. All shapes must specify a table and it must
 
 The value can be just a tablename like `projects`, or can be a qualified tablename prefixed by the database schema using a `.` delimiter, such as `foo.projects`. If you don't provide a schema prefix, then the table is assumed to be in the `public.` schema.
 
+#### Partitioned Tables
```
balegas, line 59: 👌

magnetised:
> I've also been thinking there might be an alternative implementation, whereby the replication steam is processed (by a PartitionTransformer for want of a better name) before it gets to the filter, and if it sees a change for a partition it inserts a change into the stream for the logical table. That way, all the partition logic can go in the PartitionTransformer. Shape, Filter, Consumer etc. would remain unchanged. Sure it will create some changes that might not be used, but that seems a good tradeoff for the simplicity. Is there a reason why this is not possible? Or undesirable?

@robacourt that's a good idea and does remove undesirable complexity from the Filter. it does add a coordination point where we have to add information about partitions to the partition "transformer" on shape creation. currently this is nicely encapsulated by the Shapes.Dispatcher `subscribe/2` callback which updates the filter with partition information.

so though I like the encapsulation you suggest, I think it ends up as more complex as shape information has to be threaded into more places and fundamentally we end up with more moving parts.

I will look at the integration with the filter to see if I can encapsulate the partition handling a bit more.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -148,14 +149,43 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+        # TODO: we should probabaly have a way to clean the inspector cache
+        # just based on the relation, there's a chance that this results in
+        # a query to pg just to then drop the info
+        with {:ok, info} <- Inspector.load_relation(parent, state.inspector) do
+          Inspector.clean(info, state.inspector)
+        end
+
+      {:ok, _} ->
+        # probably a malformed value from a test inspector
+        :ok
```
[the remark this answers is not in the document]
magnetised: some of our mocks don't return the full table info, including a parent value. rather than fix all our tests and add the burden of adding all these extra keys to all mock results in perpetuity i just added a clause that handles an implicit `parent: nil` clause in the inspector result.

robacourt:
> update: thinking a bit more, I could maybe model the partition handling in the filter itself using this transformer approach.

I can see the appeal of putting the logic in the filter as it needs a similar interface to the Filter. But it would be nice to keep the Filter ignorant of partitions. One possibility would be for the Dispatcher to call both, as in:
```elixir
filter = Filter.add_shape(state.filter, from, shape)
partition_transformer = PartitionTransformer.register_table(state.transformer, shape.root_table)
```
then later
```elixir
{partition_transformer, events} = PartitionTransformer.process_events([event])
...
Filter.affected_shapes(state.filter, event)
```

Ideally then the only thing that knows about partitions would be the PartitionTransformer.

magnetised:
> Ideally then the only thing that knows about partitions would be the PartitionTransformer.

@robacourt  you're absolutely right! have pushed a version that works exactly as you propose, except I dropped the `Transformer` suffix.

this version is much simpler as we don't have manage relations as well as changes.

pls have a look. pr is not ready to merge as i have some serious issues with truncation to work out

packages/sync-service/lib/electric/shapes/partitions.ex
```diff
@@ -0,0 +1,199 @@
...
+
+  @doc """
+  Update the partition information table with the given shape.
+
+  If the shape is  defined on a partitioned table (not a partition of that
+  table) then this will expand the mapping function to add a change to the
+  partition root for every change to a partition of that root.
+  """
+  @spec add_shape(t(), shape_id(), Electric.Shapes.Shape.t()) :: t()
+  def add_shape(%__MODULE__{} = transformer, shape_id, shape) do
```
robacourt, line 39: I'm guessing it's called `transformer` as the module was once called `PartitionTransformer`, it's perhaps a little confusing to call it that now without any other context. `partitions` might be a bit ambiguous, so perhaps it should be called `state` or `partitions_state`?

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -148,14 +149,40 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+        # a query to pg just to then drop the info
+        with {:ok, info} <- Inspector.load_relation(parent, state.inspector) do
+          Inspector.clean(info, state.inspector)
+        end
+
+      {:ok, _} ->
+        # probably a malformed value from a mock test inspector that isn't
+        # returning the full inspector result with a `:parent` key
+        :ok
+    end
```
robacourt, line 183: Could this go in `Partitions` too?

robacourt (approved):
Awesome. Great work, Gary! I've just left a couple of nit-picks

packages/sync-service/lib/electric/shapes/filter.ex
```diff
@@ -20,15 +19,18 @@ defmodule Electric.Shapes.Filter do
...
 
   defstruct tables: %{}
 
   @type t :: %Filter{}
   @type shape_id :: any()
 
-  @spec new() :: Filter.t()
-  def new, do: %Filter{}
+  @spec new(keyword()) :: Filter.t()
+  def new(_opts \\ []) do
```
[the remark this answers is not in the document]
magnetised: I wanted to keep the `new` signature the same in the Filter and the Partitions modules. We may not need the opts now but I don't see the harm is passing them around.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: good point!

PR #2201, opened 2024-12-20
docs: add a short guide on security.

thruflo:
Covering data access and encryption:
https://deploy-preview-2201--electric-next.netlify.app/docs/guides/security


website/docs/guides/security.md
```diff
@@ -0,0 +1,103 @@
...
+For example, you can achieve end-to-end encryption by:
+
+- *encrypting* data before it leaves the client
+- *decrypting* data when it comes off the replication stream into the client
+
+You can see an example of this in the [encryption example](/demos/encryption):
+
+<<< @../../examples/encryption/src/Example.tsx{tsx}
+
+### Key management
```
KyleAMathews, line 80: I don't have a lot of experience with key distribution but isn't storing the key (unencrypted) alongside the encrypted data more-or-less make it basically the same as if it wasn't encrypted? Since anyone with access to the db can now decrypt whatever data they want?

website/docs/guides/security.md
```diff
@@ -0,0 +1,103 @@
...
+  </a>
+</figure>
+
+This API is [public by default](#public-by-default). It should be secured in production using [network security](#network-security) and/or an [authorization proxy](#authorization).
+
+### Public by default
+
+Electric connects to Postgres as a normal [database user](https://www.postgresql.org/docs/current/user-manag.html). It then exposes access to **any&nbsp;data** that its database user can access in Postgres to **any&nbsp;client** that can connect to the Electric HTTP API.
+
+If you don't want to expose public access to the contents of your database, you **must** secure access to the Electric HTTP API.
```
KyleAMathews, line 44:
```suggestion
You generally do _not_ want to expose public access to the contents of your database, so you **must** secure access to the Electric HTTP API.
```

website/docs/guides/deployment.md
```diff
@@ -51,6 +57,13 @@ You also often want to proxy requests to Electric through your API, or other pro
 
 Note also that, when running Electric behind a CDN, you may want your proxy in front of the CDN. This is where primitives like [edge functions](/docs/integrations/supabase#sync-into-edge-function) and [edge workers](/docs/integrations/cloudflare#workers) can be very useful.
 
+### Securing data access
+
+By default, Electric exposes public access to the contents of your database. If you don't want to expose the contents of your database, you need to [lock down access](/docs/guides/security#secure-data-access) to the Electric HTTP API.
```
KyleAMathews, line 62:
```suggestion
By default, Electric exposes public access to the contents of your database. You generally don't want to expose the contents of your database, so you need to [lock down access](/docs/guides/security#secure-data-access) to the Electric HTTP API.
```

KyleAMathews (changes requested):
I suggest we make it clear that not locking down access is a very bad thing :-D 

I'm also skeptical of the value of the key distribution example? It doesn't seem particularly useful in real-world e2e encrypted apps? Unless I'm missing something.

website/docs/guides/security.md
thruflo: You could store keys in a different database and just sync the right ones. I do think that the point of Electric being good to sync keys is valid: it's just like the distributed config case. You need to share the right keys with the right users. What better mechanism?

KyleAMathews: 🤷 like I said I dunno — maybe if the key is encrypted before sharing so it's just a backup/cloud syncing mechanism.

thruflo: Have jigged the description a bit.

thruflo: Ultimately for e2e encryption you need to get keys into the client.

KyleAMathews: you can have client-generated keys right?

KyleAMathews (approved):
:shipit:

website/docs/guides/security.md
thruflo: Yup, but if you're sharing the data, you have to share keys.

Issue #1921, opened 2024-11-04
Canary builds should provide a stable tag


magnetised:
Currently we build a "canary" image `electricsql/electric:canary` off of `main`. Because the tag `electricsql/electric:canary` is not stable, if we ask someone to pull an image based off of main, we have no idea which actual build they end up with.

We should tag canary images as `electricsql/electric:canary` and `electricsql/electric:canary-$COMMIT_SHA` so that `electricsql/electric:canary` always points to the most up-to-date build, but the same build is also available under an immutable tag.

robacourt:
It's stable if you use the SHA (currently it's [sha256:644c25271c09e663fed4439c5c68fbe8e57e5dcc0d67ea92df32a40cdb11b12a](https://hub.docker.com/layers/electricsql/electric/canary/images/sha256-8a5ecb644006dd87b566cb61e4390220e6cdeeb415c16230c34412c8903b4405?context=explore) )

But currently there's no record of what version of main is what SHA, and github doesn't display a history of the SHAs.

alco:
The old Electric used to log its version when starting which also helped identify the commit that's being run.

PR #2202, opened 2024-12-20
For every canary image build, create a tag with the commit sha

alco:
Addresses #1921.

### DIFF
--- a/.buildkite/docker-image.yml
+++ b/.buildkite/docker-image.yml
@@ -14,9 +14,11 @@ steps:
       - export ELECTRIC_IMAGE_NAME="${DOCKERHUB_REPO}/${IMAGE_NAME}"
       - cd ./packages/sync-service
       - export ELECTRIC_VERSION=$(git describe --abbrev=7 --tags --always --first-parent --match '@core/sync-service@*' | sed -En 's|^@core/sync-service@||p')
+      - export SHORT_COMMIT_SHA=$(git rev-parse --short HEAD)
       - docker buildx build --platform linux/arm64/v8,linux/amd64 --push
         --build-arg ELECTRIC_VERSION=$${ELECTRIC_VERSION}
         -t $${ELECTRIC_IMAGE_NAME}:canary
+        -t $${ELECTRIC_IMAGE_NAME}:canary-$${SHORT_COMMIT_SHA}
         .
   - wait
   - label: "🚀 Build & publish the release image to Docker Hub"
--- a/packages/sync-service/lib/electric/application.ex
+++ b/packages/sync-service/lib/electric/application.ex
@@ -1,7 +1,7 @@
 defmodule Electric.Application do
   use Application
 
-  require Config
+  require Logger
 
   @doc """
   This callback starts the entire application, but is configured to run only when
@@ -81,6 +81,9 @@ defmodule Electric.Application do
         ],
         prometheus_endpoint(Electric.Config.get_env(:prometheus_port))
       ])
+
+    app_vsn = Application.spec(:electric, :vsn)
+    Logger.info("Starting ElectricSQL #{app_vsn}")
 
     Supervisor.start_link(children, strategy: :one_for_one, name: Electric.Supervisor)
   end


icehaunter (approved):
I like the idea, but I don't like having a spammed list of tags on DockerHub. A bit sad that we can't have both

magnetised (commented):
looks good. perhaps to avoid spamming the main image with random tags, @icehaunter, we could push the canary images as `electricsql/electric-canary:latest` and `electricsql/electric-canary:$sha` as well as keeping `electricsql/electric:canary` for backwards compatibility.

i stand by the requirement: it's really important to send an image tag and know that it's not going to change between sending it and it being pulled  by the user.

you can just share shas but it provides no documentation to us or the user which version they're running - having the commit sha in the image at least allows for some introspection.

alco:
@magnetised 
> i stand by the requirement: it's really important to send an image tag and know that it's not going to change between sending it and it being pulled by the user.

What use cases require such a canary tag to be sent to a user?

Upon taking another look at this change, I wouldn't want to fill up our Docker Hub repo with lots of tags. We may hit an internal limit there or cause the UI to slow down, etc. Canary builds are transient anyway.

If the main purpose is being able to tell which version of Electric is being run, that is solved by logging the version in the log output. I can leave that part of the PR.

When sharing a specific image tag is important on a case-by-case basis, it's always possible to manully push a locally available image to your personal Docker Hub repo, specifying whatever tag you choose, and then ask the user to pull that.

magnetised:
> > i stand by the requirement: it's really important to send an image tag and know that it's not going to change between sending it and it being pulled by the user.
> 
> What use cases require such a canary tag to be sent to a user?

I literally had to do this when investigating a customer problem. i couldn't send them the :canary tag because that was a moving target and some conversation like "what version does it print when it starts..." doesn't seem very helpful when trying to investigate, probably complex, issues in a customer environment

> it's always possible to manully push a locally available image to your personal Docker Hub repo, specifying whatever tag you choose, and then ask the user to pull that.

yes you could do that but why make things difficult for ourselves.

I don't honestly see this as such controversial issue. It just makes a bunch of sense to me to be able to say - "this image" to someone and know that weird things aren't going to happen. I think the idea of using elecric/canary as the name solves the issue with overloading the main image with a bunch of noise and if we hit limits in docker hub then we can address those.

KyleAMathews:
👍 to named releases for each commit. This is quite common in NPM land & it can be very handy. Granted Docker images are heavier than npm packages so we might run into limits in Docker Hub but we could e.g. just start deleting old ones at that point.

I agree the [UI on Docker Hub](https://hub.docker.com/r/electricsql/electric/tags) isn't really designed for this but... that might not be a problem? We have lists of releases elsewhere... but also a new electric/canary neatly solves the issue.

msfstef:
Bumping this @magnetised @alco as it would be convenient for ourselves in cloud as well.

alco:
I have created a new repo on Docker Hub named `electric-canary`. The change in this PR makes so every commit on `main` is pushed to `electricsql/electric:canary`, `electricsql/electric-canary:latest` and `electricsql/electric-canary:<abbreviated SHA>`.

I've updated the terminology in the `.buildkite/docker-image.yml` file to match the real-world usage:
- an **organization** is like a user account on Docker Hub that can host any number of **repositories** under its personal namespace
- a **repository** is a collection of images that all share the same base name
- an **image tag** is an arbitrary identifier which allows for distinguishing one image from the other
- an **image** name is obtained by combining `<organization>/<repository>:<image tag>`

What we in spoken language call "images" and "tags" are actually "repos" and "images" in Docker Hub's terms. "Pushing a tag" actually means creating a new image inside the chosen repository. All repositories whose names start with the same `<prefix>/` belong to the same organization named after that `<prefix>`.

msfstef (approved):
Nice work! I like that we are keeping the original canary release on top of adding properly addressable stable ones

KyleAMathews (approved):
:shipit:

PR #2208, opened 2024-12-20
fix bash example client to work with latest server

KyleAMathews:
https://github.com/user-attachments/assets/e91888ac-f659-44ab-9a28-350055ed787d


alco (approved):
:+1:

Issue #2224, opened 2025-01-13
Sync issues when using more than 2 conditions in where clause


godiagonal:
There seems to be a limitation to the number of conditions allowed in where clauses. I can't find anything in the docs that indicate that this is expected behavior.

When my where clause contains one or two conditions the syncing works as expected.

```tsx
const shape = useShape({
  url: 'http://localhost/v1/shape',
  params: {
    table: 'activity',
    // Works as expected
    where: `team_id = '${teamId}' AND deleted_at IS NULL`
  },
});
```

But when I add a third condition (which is inevitable for my use case) the shape becomes "stale". It doesn't update even though the underlying table data is updated.

```tsx
const shape = useShape({
  url: 'http://localhost/v1/shape',
  params: {
    table: 'activity',
    // Does not work
    where: `team_id = '${teamId}' AND deleted_at IS NULL and archived = false`
  },
});
```

I've tried arranging conditions in different order and combinations but concluded that it's the number of conditions (>= 3) that result in sync issues. Also tried using `replica: 'full'` but it didn't make a difference.

Is this a known limitation or a bug?

robacourt:
It is a bug. I've just reproduced it. I'm working on a fix now

robacourt:
While you're waiting for the bug fix, a workaround is to use brackets:

```javascript
where: `team_id = '${teamId}' AND (deleted_at IS NULL and archived = false)`
```

robacourt:
This is now fixed in version 1.0.0-beta.4

godiagonal:
Thanks for the quick fix @robacourt!

PR #2225, opened 2025-01-13
fix(sync-service): Fix bug with logical operators with 3 or more arguments

robacourt:
Fix for #2224


robacourt:
> Doesn't the test have the braces so doesn't have that property?

Oops, I was just checking the workaround worked but accidentally checked it in 🤦

Sorted now @icehaunter

icehaunter (approved):

Issue #2219, opened 2025-01-07
Add SSR support


KyleAMathews:
The basic pattern is that when server rendering, the resolved shapes' data is serialized to the HTML as JSON and then hydrated into shapes again when the client is setup.

We can modify `getShape` in the client so that when it creates a new shape, it first looks to the HTML for serialized shape data.

MarioSimou:
What's the required change for this feature? 

Reading the code of [@electric-sql/react](https://github.com/electric-sql/electric/blob/main/packages/react-hooks/src/react-hooks.tsx) package, it seems that we need to extend the [UseShapeOptions](https://github.com/electric-sql/electric/blob/main/packages/react-hooks/src/react-hooks.tsx#L157C4-L157C19) options to acccept a `shape` option. With that way we can [preload a shape](https://github.com/electric-sql/electric/blob/main/packages/react-hooks/src/react-hooks.tsx#L17) on the server, pass it down to the client and use it within the [useShape](https://github.com/electric-sql/electric/blob/main/packages/react-hooks/src/react-hooks.tsx#L151) hook.

I am available to work on this issue personally if necessary.

balegas:
I've made some experiments with the pattern in the [past](https://github.com/electric-sql/electric/pull/1596/files). We can definitely use the `useShape` hook to load a shape that was passed from the server. The challenge, as Kyle points out, is that the shape data needs to be serialised in the HTML when passing the data to the client. I've ended up with this [double wrapping](https://github.com/electric-sql/electric/pull/1596/files#diff-879fd624c118c48cc679ffdb50e3e5ec99c3219da6be13f1cdcb2951356af7bcR13) pattern to hydrate the shape from the serialized JSON, but I don't think the DX is great and I wanted to revise the server part to be more serverless-friendly. This is how far I got.

It would be awesome if you want to work on a better SSR support for Electric.

MarioSimou:
Thank you for sharing that. 

I'm thinking of experimenting over the weekend to see how much I can achieve.

PR #2232, opened 2025-01-17
Add SSR Support for React Hooks

KyleAMathews:
This PR adds server-side rendering (SSR) support to the React hooks.

Fix https://github.com/electric-sql/electric/issues/2219

With this PR, non-RSC SSR Just Works™️ by adding a new `<ElectricScripts />` component to your layout.

Supporting RSC proved more complicated than I wanted to handle with this PR — https://github.com/electric-sql/electric/tree/rsc is where I got to with it. I liked the hydration pattern in tanstack/react-query (https://tanstack.com/query/latest/docs/framework/react/guides/advanced-ssr) and that's what I was trying to implement but got bogged down in build errors, etc. so decided to focus on getting this in first as it works great in Remix & with Next pre-app-router. The tricky part about RSC is you don't have any global context anymore as each component renders by itself. So hydration into useShape on the server & client is different. Anyways, something to dig into another time.

Here's a detailed breakdown of the implementation:

### 1. Preloading Data on Server
```typescript
// Remix loader
export const loader: LoaderFunction = async () => {
  return await preloadShape(itemShape())
}
```

### 2. Server Rendering
The server component renders as normal with `useShape` using the preloaded data.

### 3. Serialization to HTML
The SSR state is serialized into the HTML using a script tag with a new `<ElectricScripts />` component.

Similar to frameworks like React Query and SWR, we fetch data during the server render phase. However, unique to Electric, we also need to preserve metadata like the shape's offset and handle for proper client-side rehydration.

### 4. Client Hydration
On the client, before any components render:
1. The script tag content is parsed and stored in memory
2. When `useShape` is called:
   - Checks for matching SSR data
   - If found, initializes the shape with that data before starting the stream

```typescript
function getShape(stream: ShapeStream, optionsHash: string) {
  const ssrData = getSSRState().shapes[optionsHash]
  if (ssrData) {
    // Initialize with SSR data before starting stream
    shape.initializeWithSSRData(new Map(ssrData.rows))
  }
  return shape
}
```

### Metadata Preservation
Beyond just the data, we preserve:
- offset: So client knows where to resume streaming
- handle: For authentication with the streaming endpoint
- lastSyncedAt: For cache freshness tracking


samwillis:
I like this strategy a lot, it's almost identical to what I experimented with when chatting to Valter a few months ago: https://github.com/samwillis/ssr-shape-poc/blob/main/src/app/shapes.tsx (tiny POC with no actual shapes)

KyleAMathews:
I want to make another go sometime at looking at RSC — it is fairly different and worth a serious think through as it might change a bit how everything is ordered.

MarioSimou:
Feel free to check this [draft](https://github.com/electric-sql/electric/pull/2237) -  I did some changes on the `ElectricProvider` and added support for `RSC`.

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -4,23 +4,66 @@ import {
...
+}
+
+hydrateSSRState()
+
+export async function preloadShape<T extends Row<unknown>>(
   options: ShapeStreamOptions<GetExtensions<T>>
 ): Promise<Shape<T>> {
-  const shapeStream = getShapeStream<T>(options)
-  const shape = getShape<T>(shapeStream)
-  await shape.rows
```
ebramanti, line 22: This should not be removed, without awaiting `shape.rows` you run into a race condition when the shape stream has not yet subscribed and initializes the rows to empty with an offset of `-1`.

packages/typescript-client/src/shape.ts
```diff
@@ -142,6 +142,24 @@ export class Shape<T extends Row<unknown> = Row> {
...
+    this.#data.clear()
+
+    // Copy entries from SSR data
+    for (const [key, value] of data.entries()) {
+      this.#data.set(key, value)
+    }
+
+    // Mark as up to date if we have data
+    if (data.size > 0) {
+      this.#hasNotifiedSubscribersUpToDate = true
```
ebramanti, line 159: One other change needs to be considered: how `isLoading` is handled. Right now, if a `useShape` is hydrated from the server, it will not update the `isUpToDate` value of the underlying `ShapeStream`, which is what is currently used to determine the `isLoading` value of `useShape`. To handle this, I have done something I think is a bit hacky:
```ts
  /** True during initial fetch. False afterwise.  */
  isLoading() {
    if (this.#data.size > 0) {
      return false;
    }
    return this.stream.isLoading();
  }
```
This matches up to the same check here, and handles the case where there may already be data set but there will be no received events from the newly created subscription since the data has not changed.

ebramanti (commented):
Left a couple comments from my time trying to integrate this with Tanstack Start. Very new to your codebase, so apologies if my feedback is a bit confusing 🙇🏻

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -148,17 +199,97 @@ interface UseShapeOptions<SourceData extends Row<unknown>, Selection>
...
+// Function to serialize SSR state for ElectricScripts
+function serializeSSRState(usedShapes: Set<string>): string {
+  // eslint-disable-next-line
+  const shapes: { [key: string]: SSRShapeData<any> } = {}
+
+  // Only get shapes that were used in this render
+  for (const optionsHash of usedShapes) {
+    try {
+      // Parse the options from the hash
+      const options = JSON.parse(optionsHash)
```
ebramanti, line 230: Found another edge case where this does not work. When passing a `signal` it gets serialized to an `{}` and then throws an error server side. This recovers on the client but there may need to be more granular handling of these settings rather than using JSON.

ebramanti:
@KyleAMathews Any update on this? Would be great to get this merged in with all the changes the client has undergone

KyleAMathews:
Not yet, we've been focused on getting electric cloud launched. Not firm when we'll get back to this. If you'd like to PR updates for the latest client and your fixes that'd be great!

icehaunter:
This is very stale and is superseded by work in TanStack DB (as per @samwillis)

PR #2234, opened 2025-01-21
Rename database_id to source_id in examples deployment

magnetised:
Also fix other problems with example deployments:

- new admin api endpoint
- nextjs example renamed
- add support for pooled connections


msfstef (approved):
I think this doesn't cover everything - for example the `todo-app` seems to still be using a `useShape` with a `database_id` - was this a full search and replace? Also I noticed that some examples are actually under @KyleAMathews 's repo so we can't fix those

msfstef (approved):

Issue #2229, opened 2025-01-14
Memory usage steadily increases without recovery in Electric instance


sullyo:
We're experiencing steadily increasing memory usage in our Electric instance that doesn't recover until restart.

Current behavior:
- Memory usage shows continuous growth over time
- No automatic recovery of unused memory
- Requires manual restarts to reset memory levels

Environment:
- Electric deployed on render via container
- 2gb memory, 1 cpu

Theory:
- Could be related to total requests. Saw memory exceed limit with more requests

Reproduction steps:
1. Run Electric instance
2. Monitor memory usage over time
3. Observe steady increase in memory consumption
4. Eventually requires restart to recover

Current workaround:
Manual restart of the instance clears the memory back to normal levels.

![Image](https://github.com/user-attachments/assets/29aa430b-216c-4b78-a764-388be67e2bd2)

![Image](https://github.com/user-attachments/assets/d83d9314-3408-4afc-b874-5d53f42bb037)

![Image](https://github.com/user-attachments/assets/2faaeb11-4fa0-44cb-9c64-ca0c39e91d7a)

balegas:
Hi, we've seen this happen before. We think it is related with how Elixir manages processes memory. We're gonna work on a fix ASAP.

robacourt:
Please could you try something for us? We think garbage collection is causing the issue so please could you try running it manually:

On the machine with the electric docker image running run
```
docker exec -it IMAGE_NAME bin/entrypoint remote
```
(Use `docker ps` to get the IMAGE_NAME)

This will get you into an elixir shell, where you can run:

````
:erlang.garbage_collect()
```

It's worth running this 5 times to get the memory use right down.

Let us know if this helps, if it does we can automate the garbage collection to solve your issue.

robacourt:
Another thing to try is to let electric use all 2GB of your container's RAM and see what happens. If our hypothesis is correct then hitting the container limit will cause a garbage collection and so electric should flatline at 2GB memory use and continue to operate normally.

If you'd like it to flatline at a lower value, the container's memory can be restricted with the `--memory=1g` docker run flag or with your cloud provider's settings.

sullyo:
```
iex(electric@srv-ct3p9mqj1k6c73eb68qg-768b695c96-xrz46)1> :erlang.garbage_collect()
true
iex(electric@srv-ct3p9mqj1k6c73eb68qg-768b695c96-xrz46)2> :erlang.garbage_collect()
true
[...]

PR #2240, opened 2025-01-22
feat(sync-service): Reduce memory use by hibernating idle shape processes

robacourt:
Potential fix for #2229

Electric's memory footprint can get very large due to shape processes holding on to memory even when idle. This PR hibernates idle shape processes after 30 seconds of inactivity which garbage collects unused memory.

This first graph shows 10MB changes coming in for 100 shapes (2 shapes per second). You can see this taking up 2000MB of memory which is not released even though the memory is no longer needed (it's been written to disk).
Screenshot 2025-01-22 at 12 06 41
Not only is the memory not released, but the memory is amplified. In this example it's amplified by 2x since 2000MB is twice the 1000MB needed (10MB * 100 shapes). This is due to a binary copy that happens when CubDB persists data to file. Amplification can increase further do to having a binary version of the change as well as an Elixir term version of the change which is used for where clause filtering. 

This second graph is with `ELECTRIC_SHAPE_HIBERNATE_AFTER` set to `50ms`. A value this short is unlikely to be used in production but allows the effect to be seen quickly for the purposes of this benchmark.
Screenshot 2025-01-22 at 12 09 38
As you can see, memory use flatlines at 120MB with hibernation rather than blowing up to 2000MB without hibernation.


magnetised (approved):
Beautifully simple change after a ton of research. Thanks!

Just FYI we'll be unable to publish electric to hex while we keep a github-based dependency, so here's hoping they accept that patch.

robacourt:
> Just FYI we'll be unable to publish electric to hex while we keep a github-based dependency, so here's hoping they accept that patch.

Good point. I don't think they're going to be quick so I guess I'll have to publish the fork to hex.

PR #2241, opened 2025-01-22
docs: Fix caddy reverse proxy docs

samwillis:
It appears that you now (at least on my machine) need to specify the `localhost` host name in the caddy file to get it to properly enable tls.


KyleAMathews (approved):
:shipit:

PR #2243, opened 2025-01-22
chore: Update example deployment code to use custom Neon provisioning

msfstef:
Uses custom Neon deployment code that can clean up after itself.

Should use new production cloud.


examples/yjs/sst.config.ts
```diff
@@ -172,3 +153,96 @@ async function addDatabaseToElectric(
 
   return await result.json()
 }
+
+function getNeonConnectionString({
```
KyleAMathews, line 157: time to pull these out into a util file?

magnetised (approved):

PR #2245, opened 2025-01-22
chore: Simpler example stage DB naming

msfstef:
Previous was wrong and redundant, as the stage specified in CI is `production` not `Production`


KyleAMathews (approved):
:shipit:

PR #2248, opened 2025-01-23
chore (examples): Add authentication to the examples

kevin-dp:
Since the cloud API will require authentication we need to authenticate the requests to add the example DBs.
I modified the TODO app to authenticate the request. We need to do the same for the other examples.

**Make sure that the `ELECTRIC_ADMIN_API` secret in GH points to the right path! The path should now include `/api` at the end.**

Deployment of examples in CI relies on 2 new secrets `ELECTRIC_ADMIN_API_TOKEN_ID` and `ELECTRIC_ADMIN_API_TOKEN_SECRET` which are automatically written to this repository when deploying the cloud into production.


.github/workflows/deploy_all_examples.yml
```diff
@@ -61,10 +61,14 @@ jobs:
         working-directory: ./examples/linearlite-read-only
         run: pnpm sst deploy --stage production
 
-      - name: Deploy Linearlite
-        working-directory: ./examples/linearlite
-        run: pnpm sst deploy --stage production
+      # - name: Deploy Linearlite
```
balegas, line 64: We're keeping this one out because it needs more work.

KyleAMathews (approved):
great work!

PR #2250, opened 2025-01-23
fix race condition when using mock backend

magnetised:
rather than complex logic around subscriber vs response ordering (which was fallible), simply don't launch the request until we have at least one subscriber

@msfstef this was causing flaky tests in the cloud repo


msfstef (approved):
Looks good!

PR #2258, opened 2025-01-27
(chore): Fix sporadically failing test due to capture_log

robacourt:
This PR fixes the sporadic test that can be seen failing here:
https://github.com/electric-sql/electric/actions/runs/12934429277/job/36075427601

I've also added the cause to slab so that we can hopefully prevent this kind of issue in the future:
https://electric-sql.slab.com/posts/sporadically-failing-elixir-tests-pfyia5on?shr=vrDh-k20QJMXaygzaPfMu4J4


magnetised (approved):
Good catch. Have been staring at failures with matching the log message and wondering WTF for a while now.

msfstef:
oof nice catch

PR #2265, opened 2025-01-29
fix: Do not await for responses when recovering pub filters

msfstef:
The `recover_shape` and `refresh_publication` are in the `init` path of `ShapeCache`, and there is no reason why we would be waiting for responses.

The `recover_shape` has no expected failure as it simply updates a map with appropriate counters.

The `refresh_publication` will schedule a publication update, but we don't have to wait for it to finish to consider `ShapeCache` ready, as the shape consumers are already up and running and any new shapes will trigger another publication update after this one.

These changes are based off of Sentry errors noticed from the cloud.

I've also set the publication update debounce time to 0ms, so now it simply waits until end of current process message queue. This addresses an unnecessary delay in `prepare_table` calls (@robacourt), since even when there is no contention it would still debounce for 50ms. I think that simply reinserting the update at the end of the message queue should still handle things pretty well but we can see the benchmarks.

### DIFF
--- a/.changeset/healthy-seas-bathe.md
+++ b/.changeset/healthy-seas-bathe.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Do not await for responses while recovering publication filters.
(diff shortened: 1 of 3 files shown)


msfstef:
benchmark this

robacourt (approved):
Nice work! I've just left one comment

PR #2267, opened 2025-01-29
fix: invalid json while reading a concurrently written snapshot

### LINKED ISSUE #2259: Invalid JSON occasionally returned
There's an extra comma added after the colon on created_at:

`"value":{"id":"sr_YeHm9KcTu4V9KtVV","sheet_id":"sheet_eXe8pt7SLAC1ymXK","created_at":,"2025-01-27 `

@sullyo reported it [on discord](https://discord.com/channels/933657521581858818/1332466843788775546) & says it only seems to happen on their prod instance & only on 100mb+ shapes. And even then, it's inconsistent as he tried loading a shape 5 times and it succeeded 2 times (valid JSON) and failed 3 times with invalid JSON.

- production pg version is 15.1.1.41 local pg version is 16 — _he can only reproduce problem in production_
- it never happens on small shapes — dividing line seems to be ~3k rows (1 mb of JSON)
- this only happens when creating the shape — if the shape is created successfully then it'll work thereafter

So it seems like a PG 15 (and perhaps below) bug around how we're generating JSON in pg.

icehaunter:
Closes #2259

### DIFF
--- a/.changeset/modern-toes-drop.md
+++ b/.changeset/modern-toes-drop.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: ensure correct JSON forming when reading concurrently written snapshot
--- a/.changeset/modern-toes-drop.md
+++ b/.changeset/modern-toes-drop.md
@@ -2,4 +2,4 @@
 "@core/sync-service": patch
 ---
 
-fix: ensure correct JSON forming when reading concurrently written snapshot
+fix: ensure correct JSON formating when reading concurrently written snapshot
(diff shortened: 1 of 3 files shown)


robacourt (approved):
🚀

robacourt:
benchmark this

magnetised (approved):
Nice. Thanks

.changeset/modern-toes-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: ensure correct JSON forming when reading concurrently written snapshot
```
KyleAMathews, line 5:
```suggestion
fix: ensure correct JSON formating when reading concurrently written snapshot
```

robacourt: It's not so much the formatting, it whether it's valid JSON or not.

```suggestion
fix: fix bug that produced invalid JSON when reading concurrently written snapshot
```

KyleAMathews:
Are the benchmarks still comparing against 0.9.4?

icehaunter:
@KyleAMathews yes, I'll trigger a run for the latest release, further comparisons will be against that

KyleAMathews:
Ah ok, you have to manually move forward the base 👍

PR #2271, opened 2025-01-30
Use `unsafe_cleanup!/1` to delete data after consumer has stopped

magnetised:
When cleaning up a shape we call `cleanup!` and (in the case of FileStorage) attempt to remove the data directory while the CubDb process is still running.

This fails (`File.rm_rf/1` returns `{:error, :eexist, _}`) because the files are still in use. 

Instead of removing the directory on `cleanup!` while the process(es) are still running, instead teardown the consumer supervision tree and use `unsafe_cleanup!/1` do remove the directories. 

`unsafe_cleanup!` is already documented as something that should be runnable without any active processes, so this is an ideal use.

Fixes problems with https://github.com/electric-sql/stratovolt/pull/243

### DIFF
--- a/.changeset/seven-files-thank.md
+++ b/.changeset/seven-files-thank.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Use `Storage.unsafe_cleanup!/1` to delete data after a shape has been removed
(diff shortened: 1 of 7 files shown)


msfstef (approved):
Nice! Fixes a sentry error I wanted to chase up on as well, two birds with one stone - more unsafe cleanups is the way!

PR #2273, opened 2025-01-30
add abstraction to manage shape request-response behaviour

magnetised:
Used by https://github.com/electric-sql/electric/pull/2297


packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -0,0 +1,471 @@
...
+  """
+  @spec validate(t(), %{(atom() | binary()) => term()}) ::
+          {:ok, Request.t()} | {:error, Response.t()}
+  def validate(%Api{} = api, params) when is_configured(api) do
+    with :ok <- hold_until_stack_ready(api) do
+      with {:ok, request} <- validate_params(api, params),
+           {:ok, request} <- load_shape_info(request) do
+        {:ok, seek(request)}
+      end
+    end
```
icehaunter, line 85: Why nested `with` clauses with no `else`?

magnetised: good point!

packages/sync-service/lib/electric/shapes/api/response.ex
```diff
@@ -0,0 +1,164 @@
...
+
+    struct(__MODULE__, opts)
+  end
+
+  def error(%Api.Request{} = request, message, args) do
+    opts =
+      args
+      |> Keyword.put_new(:status, 400)
+      |> Keyword.put(:body, Api.encode_message(request, message))
+      |> Keyword.put(:shape_definition, get_in(request.params.shape_definition))
```
icehaunter, line 60: Why do we need `get_in` here?

packages/sync-service/lib/electric/shapes/api/response.ex
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  def shape_definition_mismatch(request) do
+    error(request, @shape_definition_mismatch)
+  end
+
+  def error(api_or_request, message, args \\ [])
+
+  @spec error(Api.t(), term(), keyword()) :: t()
+  def error(%Api{} = api, message, args) do
```
icehaunter, line 46: I looked through the codebase, and we have:
```
error(..., %{message: "stack not ready"}, ...)
error(..., %{status: "DELETE not allowed"}, ...)
error(..., %{status: "DELETE not allowed"}, ...)
error(..., [%{headers: %{control: "must-refetch"}}], ...)
error(..., %{offset: ["out of bounds for this shape"]}, ...)
error(..., %{error: "unable to retrieve"}, ...)
```

In think we could use this opportunity to normalize the messages? maybe allow passing a binary to this function and wrap it in `%{message: }` inside this function so that the caller doesn't have to care about error presentation?

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -0,0 +1,470 @@
...
+        {:current_stacktrace, [_ | stacktrace]} = Process.info(self(), :current_stacktrace)
+
+        %{
+          request
+          | response:
+              Response.error(
+                request,
+                %{
+                  error:
+                    "Unable retrieve shape log:\n#{Exception.format(:error, error, stacktrace)}"
```
icehaunter, line 359: I really don't like the idea of leaking our exception logs via the API responses. Feels very JAVA circa 2007. Can we do `Logger.error` with these contents but leave them out of the response?

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -0,0 +1,470 @@
...
+  @doc """
+  Return shape log data.
+  """
+  def serve_shape_log(%Request{} = request) do
+    validate_serve_usage!(request)
+
+    with_span(request, "shape_get.plug.serve_shape_log", fn ->
+      request
+      |> do_serve_shape_log()
+      |> then(fn %{response: response} -> response end)
```
icehaunter, line 310: Since all you're doing here is unpacking the response from the `%Request{}` struct, is there a reason to have a complicated "update the response on the original request struct to add an error" logic inside the `do_serve_shape_log`? You could just return the response directly and simplify multiple code paths

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -0,0 +1,472 @@
...
+       )
+       when is_nil(shape_handle) or shape_handle == active_shape_handle do
+    # We found a shape that matches the shape definition
+    # and the shape has the same ID as the shape handle provided by the user
+    {:ok,
+     Map.update!(
+       %{request | handle: active_shape_handle, last_offset: last_offset},
+       :response,
+       &%{&1 | handle: active_shape_handle}
+     )}
```
icehaunter, line 191: I don't like this update-in-an-update syntax - could we add a helper method to `Request` like `set_handle` that would take care of setting it in both the request and response?

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -0,0 +1,472 @@
...
+      last_offset: last_offset,
+      chunk_end_offset: chunk_end_offset,
+      params: %{offset: offset}
+    } = request
+
+    # The log can't be up to date if the last_offset is not the actual end.
+    # Also if client is requesting the start of the log, we don't set `up-to-date`
+    # here either as we want to set a long max-age on the cache-control.
+    if LogOffset.compare(chunk_end_offset, last_offset) == :lt || offset == @before_all_offset do
+      Map.update!(request, :response, &%{&1 | up_to_date: false})
```
icehaunter, line 295: We're doing a bit of those over this file - can we add a `Request.update_response` method?

icehaunter (changes requested):
This a great change. I think this encapsulation is worth it and makes parts of the flow much cleaner. I think some nested structure updates are a bit hard to follow at times, but that's OK. I have some nitpicks, with only blocking request being the one about exposing stacktraces

packages/sync-service/lib/electric/shapes/api/request.ex
```diff
@@ -0,0 +1,29 @@
...
+  @type t() :: %__MODULE__{
+          chunk_end_offset: nil | LogOffset.t(),
+          handle: nil | shape_handle(),
+          last_offset: nil | LogOffset.t(),
+          new_changes_ref: nil | reference(),
+          new_changes_pid: nil | pid(),
+          api: Api.t(),
+          params: Api.Params.t(),
+          response: Api.Response.t(),
+          valid: boolean()
```
icehaunter, line 27: I don't see this flag being used anywhere for what it's used in e.g. Ecto: piping that ignores subsequent steps once changeset is not valid. I'd think that pattern would be amazing to use, but currently I don't think this flag is ever read

magnetised: your're right, it's unused. I did have a plug-like flow which would have needed this, emulating the version in the original plug, but then I put (almost) all the error handling into the validate function.

packages/sync-service/lib/electric/shapes/api.ex
magnetised: Love it

packages/sync-service/lib/electric/shapes/api.ex
magnetised: good point.

packages/sync-service/lib/electric/shapes/api/response.ex
magnetised: we don't!

packages/sync-service/lib/electric/shapes/api.ex
magnetised: seems excessive considering we only do this once. it's a bit cleaner with your Request.update_response/2 suggestion anyway.

packages/sync-service/lib/electric/shapes/api/response.ex
magnetised: yeah, didn't want to change anything about the output so as not to confuse things, but I agree. there's no reason to be so complicated about our error messages.

normalizing...

magnetised: except the must-refetch message, which all the clients are currently looking for as-is

PR #2279, opened 2025-02-02
fix: don't show column details if they weren't selected

### LINKED ISSUE #2253: Only return schemas in the `electric-schema` header for the selected columns
E.g. if the table has columns a,b,c and the request only wants column a, the `electric-schema` should also only include a.

---

### Discussed in https://github.com/electric-sql/electric/discussions/2252

<div type='discussions-op-text'>

<sup>Originally posted by **gartz** January 24, 2025</sup>
First of all, I want to say I really appreciate the concepts behind Electric-SQL—it's an amazing project!

While learning about the shapes for proxy/gatekeeping, I noticed the recommendation to filter out sensitive or backend-only columns to ensure they aren't exposed to the client. However, I observed that the `electric-schema` header still includes columns that are not part of the shape.

## Example

For instance, a `users` table might include sensitive columns such as `salt` and `password_hash`, which shouldn't be synced with the client. However, columns like `username` and `email` should be exposed. Despite excluding sensitive columns from the shape, they still appear in the `electric-schema` header.

## Question

What is the best approach to address this issue? Could the documentation provide more clarity on this scenario? Some potential solutions I've considered are:

1. **Creating a View Table**: Define a database view that includes only the columns the client should access. Use this view in the shape definition instead of the original table.
2. **Manipulating the `electric-schema` JSON**: Modify the schema header to include only client-safe columns before propagating it.

Could you provide guidance on the recommended approach or suggest a best practice? Additionally, if a specific method is preferred, it would be helpful to include it in the documentation for others encountering this scenario.

Thank you for your time and for building such an exciting tool!
</div>

icehaunter:
Fixes #2253

### DIFF
--- a/.changeset/eighty-masks-battle.md
+++ b/.changeset/eighty-masks-battle.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: don't expose information about columns that weren't selected
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -307,7 +307,7 @@ defmodule Electric.Plug.ServeShapePlug do
     shape.table_info
     |> Map.fetch!(shape.root_table)
     |> Map.fetch!(:columns)
-    |> Schema.from_column_info()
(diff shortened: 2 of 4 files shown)


robacourt (approved):
Neat

PR #2288, opened 2025-02-03
feat(sync-service): Connect shape consumer spans to replication traces and add OTEL metrics

### LINKED ISSUE #2269: Add the number of shapes to metrics
Perhaps total shapes + shapes that have an active subscriber (defined as got a request in < 60 seconds perhaps)?

robacourt:
This PR improves observability by:
- Connecting shape consumer spans to replication traces
- Adding `electric.shapes.total_shapes.count` metric to OTEL metrics (Closes #2269 although I have not added "shapes that have an active subscriber" as this information is already available in the traces)
- Adding `electric.storage.used` metric to OTEL metrics
- Adding CPU % utilisation per core telemetry and exposing them to the OTEL metrics. CPU metrics are already available, but % utilisation per core seems a simple one to comprehend and in a form our customers will understand.

### DIFF
--- a/packages/sync-service/lib/electric/telemetry.ex
+++ b/packages/sync-service/lib/electric/telemetry.ex
@@ -210,6 +210,8 @@ defmodule Electric.Telemetry do
 
   defp otel_metrics() do
     [
+      last_value("system.cpu.core_count"),
+      last_value("system.cpu.utilization.total"),
       last_value("electric.storage.used", unit: {:byte, :kilobyte}),
       last_value("electric.shapes.total_shapes.count"),
       last_value("system.load_percent.avg1"),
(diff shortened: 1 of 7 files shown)


packages/sync-service/lib/electric/telemetry.ex
```diff
@@ -210,6 +210,10 @@ defmodule Electric.Telemetry do
 
   defp otel_metrics() do
     [
+      last_value("system.cpu.core_count"),
+      last_value("system.cpu.utilization.total"),
+      last_value("electric.storage.used", unit: {:byte, :kilobyte}),
+      last_value("electric.shapes.total_shapes.count"),
```
icehaunter, line 216: Let's move them up to prometheus metrics. Generally, anything not-distribution should live there to expose well

icehaunter (approved):
One nitpick, lgtm otherwise

packages/sync-service/lib/electric/telemetry.ex
robacourt: Ok. I'll address that in a separate PR as I'm out and about and wouldn't mind getting a version to trigger.dev

PR #2293, opened 2025-02-04
fix: Passthrough errors to avoid infinite loop when falling back to IPv4

msfstef:
Fixes issue introduced in https://github.com/electric-sql/electric/pull/2283

The passthrough would not work and if an error came through that wasn't `:nxdomain` it would enter an infinite loop (whoops).

I've added the case of an error where network is unreachable that @samwillis run into, assuming it's because IPv6 is not enabled on the machine, but am reluctant on disabling IPv6 on _all_ `DBConnection` errors or tcp errors.

We definitely do not want to make IPv6 enabled by default until we sort this out.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -528,7 +528,7 @@ defmodule Electric.Connection.Manager do
   end
 
   defp handle_connection_error(
-         %DBConnection.ConnectionError{message: message, severity: :error} = error,
+         %DBConnection.ConnectionError{message: message, severity: :error},
          %State{connection_opts: connection_opts, ipv6_enabled: true} = state,
          mode
        ) do
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -550,7 +550,7 @@ defmodule Electric.Connection.Manager do
       step = current_connection_step(state)
       handle_continue(step, state)
     else
-      handle_connection_error(error, state, mode)
+      handle_connection_error({:passthrough, message}, state, mode)
     end
   end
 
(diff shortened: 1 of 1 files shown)


samwillis (approved):
👍 it worked for me 😃

icehaunter (approved):

alco:
We used to have IPv6 by default in the old Electric. Then I had to disable that after people had been reporting issues. I don't remember the exact ones but the PR description [here](https://github.com/electric-sql/electric/pull/1210) says that a domain name may have an AAAA record and the machine Electric is running on may have a working IPv6 stack but the remote server might still refuse connection attempts. In that case, Electric will keep retrying the connection in a loop and won't try to fall back to IPv4.

I'm not suggesting we change the current behaviour in Electric, but I'm pointing out that doing any decisions with regards to switching between IPv4 and IPv6 (in any direction) inside Electric cannot be made to work reliably in all environments.

The `DATABASE_USE_IPV6` setting was supposed to give the user full control of which IP network stack to use with no hidden surprises. Now that the fallback behaviour has been introduced, it's become less obvious whether Electric should or shouldn't try to switch to IPv6 when it cannot connect using an IPv6 address multiple times in a row.

msfstef:
@alco absolutely agree, which is why ideally we get an answer to the issues referenced in https://github.com/electric-sql/electric/pull/2283 to properly address this. I've tried to keep this as tight a fallback as possible for missing AAAA records and lack of IPv6 networking, which I believe covers a significant portion of IPv6 configuration issues. And this is still only happening if `DATABASE_USE_IPV6` is set.

There's the option of reverting this fallback altogether and keeping the strict behaviour, but it will require some extra work on the cloud configuration side first - the main reason this naive fallback was added was to unblock things there.

We also have the option to have more than a boolean flag and have `ipv4` `ipv6` `adaptable` enum config, and document that we try our best to connect with whatever we find available in the `adaptable` use case - users who want Electric to fail if it doesn't use the desired protocol can set what they prefer but having a plug and play configuration as the default would ease adoption I think, even if it's not always the desired approach.

PR #2298, opened 2025-02-04
feat: change `txid` header to be `txids` 

### LINKED ISSUE #2287: Change `txid` header to be a list
Currenly, on the wire have a header on each (live) operation of `txid`, marking the original transaction ID. This is useful in plenty of cases, one of which is that writes to PG can be tracked back to being observed.

Proposed change is that we replace `txid` with `txids` and list-wrap the currently singular item. The logic that `txids` header is only present when we're not in snapshot/compaction territory persists.

icehaunter:
Closes #2287

### DIFF
--- a/.changeset/sixty-queens-agree.md
+++ b/.changeset/sixty-queens-agree.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+feat: replace `txid` with `txids` header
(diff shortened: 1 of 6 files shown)


msfstef (approved):
Nice

samwillis (approved):
LGTM, also love it!

Do we need to update the protocol definition file?

PR #2300, opened 2025-02-04
Move electric_phoenix package into monorepo

magnetised:
plus fix electric client versioning and update to phoenix liveview 1.x


packages/electric-phoenix/mix.exs
```diff
@@ -47,15 +47,25 @@ defmodule Electric.Phoenix.MixProject do
...
-      {:electric_client, path: "../../../electric/packages/elixir-client/", only: [:test]}
+      {:mox, "~> 1.1", only: [:test]},
+      {:electric, path: "../sync-service/", env: :dev, only: [:test], override: true},
+      {:electric_client, path: "../elixir-client/", env: :dev, only: [:test], override: true}
+    ]
+  end
+
+  defp deps_for_env(:dev) do
+    [
+      {:electric, path: "../sync-service/", env: :dev, only: [:dev, :test], override: true},
```
thruflo, line 60: This "workspace" approach is a bit tricky. Because presumably when `:electric_phoenix` is included as a dependency by someone's app, that app can be run in `:dev` mode.

magnetised: hey - yeah. i'm in "trail of destruction" mode. just trying stuff out and getting it working. no plans to include this in the final release.

Issue #2291, opened 2025-02-04
Add reconnect strategy for connecting to the database


balegas:
Electric stays in a constant-time loop trying to connect to Postgres, we want to have a reconnect strategy so that we can do exponential-backoff.

Additionally, we want to report the number of connect attempts.

msfstef:
Currently it is already doing a backoff but with a maximum retry timeout of 10 seconds - do we want to increase that to, say, a minute? Ideally Electric should reconnect to a database after some outing as soon as possible so I'm reluctant to increase the timeout too much.

I think reporting the reconnection attempts and total time spent reconnecting to the stack events allows for some flexibility on strategies for handling the stack.

KyleAMathews:
I think trying for a little while makes sense but after a few minutes, just quitting also makes sense to me. Locally I'd be fine with a program just quitting with an error saying it couldn't connect and in some sort of deployed Docker context, that'd also work.

msfstef:
Hmm the idea is that anyone can use the `database_connection_failed` stack event (to which I'm adding some `total_retry_time` info) to determine whether to terminate the service and after how long. I think there's two questions to answer:

1. Should we terminate Electric after failing to connect or reconnect for some while? If so, after how long and should it be configurable?
  1a. This is why I prefer the stack event approach, the configurability is built in, and we don't have to think too hard about what it means to "shut down" the system outside of the context it is running in
2. How large should the maximum backoff time be? Whether we end up waiting for 2 minutes or 2 hours, should retries happen at a max interval of 10 seconds or 10 minutes?

KyleAMathews:
Yeah, if someone is embedding Electric into another Elixir/Erlang app then yeah, leaving it open ended makes sense. But that'll be a vanishingly small % of people.

10 minutes seems like a reasonable amount to try. I'd keep it trying every 10s during that.

balegas:
The decision to expose a number of retries/total time spent trying to reconnect was to be make electric non-opinionated about closing the connection. Stop retrying would make sense too, but we don't have to commit to that specific behaviour.

I think the reconnect attempt interval is co-related with the period of time we want to retry. For Cloud, what sounds good to me is to keep a small interval and a small period. If a user is looking through the dashboard and Electric does not connect immediately, the user would potentially address the issue and click the button again. So, no need for a long period IMO, but I don't have any preference. 10 seconds for 10 min are 60 events only :).

(I assumed it was not exponential backoff because 10 sec sounded too quick to me :P... )

[...]

PR #2301, opened 2025-02-04
fix: Add more information to reconnection attempt events

msfstef:
Closes https://github.com/electric-sql/electric/issues/2291

- Adds total time spent reconnecting to the `database_connection_failed` stack event, useful for the event handler to determine what to do with a failing stack without keeping track of that itself (stateless handler)
- Fixes resetting the backoff after a successful reconnection

@balegas we were already doing exponential backoff, but from 1 second up to 10 seconds with an factor of 2 - given that this is also what handles reconnecting in case a database is temporarily unavailable, do we actually want to change the maximum backoff time?

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -48,7 +48,7 @@ defmodule Electric.Connection.Manager do
       # PID of the shape log collector
       :shape_log_collector_pid,
       # Backoff term used for reconnection with exponential back-off
-      :backoff,
+      :connection_backoff,
       # Flag indicating whether the lock on the replication has been acquired
       :pg_lock_acquired,
       # PostgreSQL server version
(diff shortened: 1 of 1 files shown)


KyleAMathews (commented):
I think 10s is fine as if someone fixes an issue, we want it to quickly reconnect. But to reiterate my comment from issue, I think we should set a max reconnect period before quiting.

balegas:
> I think we should set a max reconnect period before quitting.

I left [this](https://github.com/electric-sql/electric/issues/2291#issuecomment-2635341988) comment. We discussed that during the call and made the decision. I don't have any strong arguments, so I'm happy with what the team thought was best. We can always do that later.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -185,7 +185,7 @@ defmodule Electric.Connection.Manager do
         timeline_opts: timeline_opts,
         shape_cache_opts: shape_cache_opts,
         pg_lock_acquired: false,
-        backoff: {:backoff.init(1000, 10_000), nil},
+        connection_backoff: {:backoff.init(1000, 10_000), nil, nil},
```
robacourt, line 188: Changing this tuple into a map would improve readability.

robacourt (changes requested):
Looks good to me! Just added one comment

packages/sync-service/lib/electric/connection/manager.ex
msfstef: done

icehaunter (approved):

robacourt (approved):
Nice work!

Issue #1709, opened 2024-09-16
Add support for ENUM types in shape's WHERE clause


alco:
- [ ] Allow casting any enum type to a character type: `text`, `char(N)`, `varchar(N)`.
  This will allow WHERE clauses to have equality, non-equality, list inclusion, nullability checks via an explicit cast. For example, `status::text = 'draft'` or `status::text IN ('draft', 'in_review')`.

- [ ] Implicitly cast character literals to the appropriate enum type.
  This will allow shapes to have WHERE clauses such as `status IN ('draft', 'published')` or `status > 'draft'`.
  
  Note that `status::text > 'draft'` is different from `status > 'draft'::status`. The result depends on the order of enum values in the type definition:

  ```sql
  =# create type status as enum ('published', 'draft');
  CREATE TYPE

  =# create table example(id bigserial primary key, status status);
  CREATE TABLE

  =# insert into example(id, status) VALUES (1, 'draft'), (2, 'published');
  INSERT 0 2

  =# select * from example where status::text > 'draft';
   id │  status   
  ────┼───────────
    2 │ published
  (1 row)

  =# select * from example where status > 'draft'::status;
   id │ status 
  ────┼────────
  (0 rows)

  =# select * from example where status < 'draft'::status;
   id │  status   
  ────┼───────────
    2 │ published
  (1 row)
  ```

msfstef:
Issue we've run into working on this with @icehaunter is that row filters on the replication publication don't allow for user-defined types, see [docs](https://www.postgresql.org/docs/17/sql-createpublication.html):

>A row filter expression (i.e., the WHERE clause) must contain only columns that are covered by the REPLICA IDENTITY, in order for UPDATE and DELETE operations to be published. For publication of INSERT operations, any column may be used in the WHERE expression. The row filter allows simple expressions that don't have user-defined functions, user-defined operators, user-defined types, user-defined collations, non-immutable built-in functions, or references to system columns.

markpsiano:
bump, is this on the roadmap? seems like postgres wouldn't be able to use indexes on enum type columns when casting to text

also, the current error message "invalid syntax for type enum" is misleading. can it be more explicit and just say enums aren't supported at all, and to cast them to text? I spent a bit of time trying to figure out what was wrong with my syntax which is valid postgres syntax when it's just an electricsql limitation

PR #2304, opened 2025-02-05
feat: Handle casting enums to `text` in where clauses

msfstef:
Addresses part of https://github.com/electric-sql/electric/issues/1709

- Adds a `type_kind` property on column information, indicating whether a type is e.g. `:enum` or `:domain`
- Allow casting enums to text in where clauses, e.g. `foo::text = 'bar'`
- We cannot do row filtering with custom data types (see [PG docs](https://www.postgresql.org/docs/17/sql-createpublication.html#:~:text=The%20row%20filter%20allows%20simple%20expressions%20that%20don%27t%20have%20user%2Ddefined%20functions%2C%20user%2Ddefined%20operators%2C%20user%2Ddefined%20types%2C%20user%2Ddefined%20collations%2C%20non%2Dimmutable%20built%2Din%20functions%2C%20or%20references%20to%20system%20columns.)) so we treat where clauses that use them as "wide" clauses, like `1=1`.

I've updated the docs with a warning about the row filtering (not sure who should review docs, maybe @balegas ?) - see [deploy preview](https://deploy-preview-2304--electric-next.netlify.app/docs/guides/shapes#where-clause).

Co-authored-by: Ilia Borovitinov <ilia@electric-sql.com>


msfstef:
>  I think the documentation could say that explicitly and then list this warning and other conditions that might cause the same expansion.


@balegas I think it makes sense to keep the warning under the where clauses, as this is a warning about using custom data types in where clauses. As for the columns, we currently replicate all columns always because of the limitation with the replica identity, see https://github.com/electric-sql/electric/issues/1831

samwillis:
Rather than a warning about filtering on custom types. Could we frame it under the optimisations?

So by default we receive everything for a table, but because were awesome we can optimise this where possible (i.e. with pg built in types), and *pre-filter* at the replication level from Postgres reducing the number of rows sent over the wire?

msfstef:
> Rather than a warning about filtering on custom types. Could we frame it under the optimisations?

I could keep it as a warning, but preface the warning with "Electric will use row filtering to limit the amount of data being replicated from Postgres based on active shapes, _but_" - how does that sound?

samwillis:
@msfstef I honestly don't think it needs to be a warning, just listed under the optimisations.

People will assume that we replicate everything, they won't know that filtering at the replication stream is possible, and so it's a nice surprise when we say that there is this optimisation we can do.

packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
```diff
@@ -109,10 +110,28 @@ defmodule Electric.Postgres.Inspector.DirectInspector do
...
   end
 
+  @spec map_type_kind(String.t()) :: Electric.Postgres.Inspector.type_kind()
+  defp map_type_kind("b"), do: :base
+  defp map_type_kind("c"), do: :composite
+  defp map_type_kind("d"), do: :domain
+  defp map_type_kind("e"), do: :enum
+  defp map_type_kind("p"), do: :pseudo
+  defp map_type_kind("r"), do: :range
+  defp map_type_kind("m"), do: :multirange
```
robacourt, line 134: Nitpick but I'd prefer `parse_type_kind` or even just `type_kind` rather than `map_type_kind` which has me wondering what the map type is

robacourt (approved):
Lovely! I just left a tiny nitpick

PR #2305, opened 2025-02-05
chore: Minor `Connection.Manager` cleanup and flake reduction

msfstef:


icehaunter (approved):
Nice refactorings, love it! Please don't merge this until Garry's #2273 is in because I think there are conflicts that should go in first

robacourt:
> One thing that really worked for me that I hadn't tried before, is to use a tool like `stress` (e.g. `stress -c 10`) to max out the resources used by my machine while running tests to simulate a lower resource system that the tests on CI run on. I managed to consistently reproduce the consumer test flakes that way and debug them.

Great tip, thanks! I've added it here: https://electric-sql.slab.com/posts/sporadically-failing-elixir-tests-pfyia5on

PR #2308, opened 2025-02-06
Include wx and observer in the extra_application list

alco:
Without it, in recent versions of Elixir, evaluating `:observer.start` doesn't work because the `:observer` module cannot be found.

### DIFF
--- a/packages/sync-service/mix.exs
+++ b/packages/sync-service/mix.exs
@@ -58,7 +58,14 @@ defmodule Electric.MixProject do
 
   def application do
     [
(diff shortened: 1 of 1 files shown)


icehaunter:
I don't think we need to bloat our releases with `wx` for this to work: https://elixirforum.com/t/observer-in-iex-s-mix-error-undefinedfunctionerror-function-wx-object-start-3-is-undefined/58544/6

image

alco:
Ah, good to know. Thanks Ilia!

PR #2310, opened 2025-02-09
fix: make sure `\r\n` doesn't break the compaction

icehaunter:


msfstef:
Linking https://github.com/electric-sql/electric/pull/2350 here as they attempt to solve the same issue

alco:
Incorporated these changes into https://github.com/electric-sql/electric/pull/2350.

Issue #2315, opened 2025-02-10
React peer deps warning


thruflo:
When I install @electric-sql/react in a project with the latest stable React version, I get a peer deps warning:

```
➜  assets git:(main) ✗ npm install --save @electric-sql/react
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR!
npm ERR! While resolving: @electric-sql/react@1.0.0-beta.3
npm ERR! Found: react@19.0.0
npm ERR! node_modules/react
npm ERR!   peer react@"^19.0.0" from react-dom@19.0.0
npm ERR!   node_modules/react-dom
npm ERR!     peerOptional react-dom@">=18" from react-router@7.1.5
npm ERR!     node_modules/react-router
npm ERR!       react-router@"^7.1.5" from the root project
npm ERR!     react-dom@"^19.0.0" from the root project
npm ERR!   peer react@">=18" from react-router@7.1.5
npm ERR!   node_modules/react-router
npm ERR!     react-router@"^7.1.5" from the root project
npm ERR!   2 more (valtio, the root project)
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peerOptional react@"^18.3.1" from @electric-sql/react@1.0.0-beta.3
npm ERR! node_modules/@electric-sql/react
npm ERR!   @electric-sql/react@"*" from the root project
npm ERR!
npm ERR! Conflicting peer dependency: react@18.3.1
npm ERR! node_modules/react
npm ERR!   peerOptional react@"^18.3.1" from @electric-sql/react@1.0.0-beta.3
npm ERR!   node_modules/@electric-sql/react
npm ERR!     @electric-sql/react@"*" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

It’s pretty scary. I have force / override to workaround it. This is quite a big barrier on our primary adoption path.

Can we change / loosen the version compatibility defined in the react-hooks package? It would be nice to remove this warning.

PR #2317, opened 2025-02-10
fix: Looser `react` peer dependency versions

msfstef:
Closes https://github.com/electric-sql/electric/issues/2315

Looser requirement for the peer dependency to avoid the reported warnings with v19 react. We ideally want to be testing it with multiple major react versions though to make sure we actually support them.


samwillis (approved):
LGTM 👍

PR #2318, opened 2025-02-10
More determinism in the invalidated-replication-slot.lux test

alco:
Monitor the stack supervisor process to guarantee its eventual shutdown. Previously, it might not have shut down in time before the liveness check performed in the integration test.


msfstef (approved):
Very nice! Thank you for sorting this out

PR #2320, opened 2025-02-10
fix: Rewrite xid comparison to use proper modulo-2^32 arithmetic 

### LINKED ISSUE #2312: The comparison between 64-bit non-wrapping transaction ID and 32-bit wrapping transaction ID is implemented incorrectly
The function `compare()` and the guard `xid_lt_xid8` implemented in the `Electric.Postgres.Xid` module have a logical flaw in them: the "direct comparison" of two 32-bit values ignores the "~2 billion values preceding it and ~2 billion values following it` rule.

Also, the implementation relies on checking multiple conditional values and is quite hard to read.

A correct and simpler implementation would be a direct translation from Postgres' [`TransactionIdPrecedes()`](https://github.com/postgres/postgres/blob/3d17d7d7fb7a4603b48acb275b5a416f110db464/src/backend/access/transam/transam.c#L277-L293) function which implements a modulo-2^32 comparison. In Elixir it can be emulated using arithmetic within binary string expressions.

alco:
The logic implemented in https://github.com/electric-sql/electric/pull/2217 was wrong. This PR fixes the comparison by porting the logic from Postgres' C implementation to Elixir's bitstring syntax.

Fixes #2312.


packages/sync-service/lib/electric/postgres/xid.ex
```diff
@@ -1,66 +1,98 @@
...
+  iex> compare(#{@uint32_half_max + 1}, 1)
+  :lt
+
+  iex> compare(#{@uint32_half_max}, #{@uint32_max})
+  :lt
+
+  iex> compare(#{@uint32_half_max - 1}, #{@uint32_max})
+  :lt
+
+  iex> compare(#{@uint32_half_max - 2}, #{@uint32_max})
```
msfstef, line 69: might be nice to have a couple of doctests of 64-bit and 32-bit mixed comparisons as well

packages/sync-service/lib/electric/postgres/xid.ex
```diff
@@ -1,66 +1,98 @@
...
+
+  # If one of the numbers is a 32-bit unsigned integer, we compare the two numbers using
+  # modulo-2^32 arithmetic.
+  def compare(xid_l, xid_r) when (uint32?(xid_l) or uint32?(xid_r)) and xid_l > 0 and xid_r > 0 do
+    # This produces equivalent results to the following C code:
+    #
+    #     uint32 xid_l, xid_r;
+    #     int32 signed_diff = (int32)(xid_l - xid_r);
+    #
+    <<signed_diff::signed-32>> = <<xid_l - xid_r::unsigned-32>>
```
msfstef, line 89: this is not a performance sensitive operation, but are there any gains to be made by specifying the endianness to be native for this? I noticed Elixir defaults to big endian.

msfstef (approved):
Very nice, this solution makes sense!

packages/sync-service/lib/electric/postgres/xid.ex
alco: Endianness only matters for the purposes of transferring numbers as sequences of bytes between environments. The actual arithmetic is performed using the CPU's native endianness, we have no influence over that in Elixir.

Here it's only important to have the same endianness on both sides of the equals sign.

msfstef: right - my thought was more around whether the matching performed by Elixir in a given representation might be faster if the representation matches the native endianness

>Here it's only important to have the same endianness on both sides of the equals sign.


yeah correctness is not an issue

alco: It's not something I've been worried about or seen in any codebase. There might be some gain to be had here but it would be in the realm of micro-optimisation. For our purposes, for 99.99% of a shape's lifetime this comparison is not going to happen because of the filtering performed in `Electric.Shapes.Consumer` by matching on snapshot's `filter_txns?` key (added in the follow-up PR).

msfstef: true, more of a question of personal curiosity 😄

packages/sync-service/lib/electric/postgres/xid.ex
alco: Added a hanful. Thanks for the suggestion.

PR #2321, opened 2025-02-10
feat(sync-service): Add optional where clause filtering tracing

robacourt:
To diagnose slow where clause filtering this PR introduces 3 new optional filter spans:
- `filter.filter_using_indexes` - how long index based filtering takes. Attributes: table and index count
- `filter.filter_other_shapes` - how long non-index based filtering takes. Attributes: table and shape count
- `filter.index.filter_matched_shapes` - how long filtering the matched shapes takes after the change has matched an index. Attributes: field, matched_shape_count

In order to do this, this PR also introduces:
- Optional spans (see the `OptionalSpans` module)
- Open Telemetry Baggage propagation between processes. This reduces noise in the code by not requiring `stack_id` to be passed around to everywhere we do tracing.
- Changed the shapes collection in the index from a list to a map, this allows us to efficiently get the number of shapes without having to transverse the list

### DIFF
--- a/packages/sync-service/lib/electric/telemetry/optional_spans.ex
+++ b/packages/sync-service/lib/electric/telemetry/optional_spans.ex
@@ -1,0 +1,4 @@
+defmodule Electric.Telemetry.OptionalSpans do
+  def include?("filter." <> _), do: Application.get_env(:electric, :profile_where_clauses?)
+  def include?(_), do: true
+end
--- a/packages/sync-service/.env.dev
+++ b/packages/sync-service/.env.dev
@@ -7,3 +7,6 @@ ELECTRIC_CACHE_STALE_AGE=3
 ELECTRIC_SHAPE_CHUNK_BYTES_THRESHOLD=10000
 # configuring a second database for multi-tenancy integration testing
 OTHER_DATABASE_URL=postgresql://postgres:password@localhost:54322/electric?sslmode=disable
+ELECTRIC_PROFILE_WHERE_CLAUSES=false
+ELECTRIC_OTEL_SAMPLING_RATIO=1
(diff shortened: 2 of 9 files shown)


KyleAMathews:
Do you have some traces in Honeycomb we could look at?

robacourt:
> Do you have some traces in Honeycomb we could look at?

Here's a simple one:
Screenshot 2025-02-11 at 08 42 37

KyleAMathews:
Any idea what's happening in the first half of the span?

alco:
> Any idea what's happening in the first half of the span?

There's message decoding at the beginning there. It used to be in a span but then, as part of the initiative to lower the number of spans, it got converted into an attribute on the parent span.

packages/sync-service/lib/electric/telemetry/optional_spans.ex
```diff
@@ -0,0 +1,4 @@
+defmodule Electric.Telemetry.OptionalSpans do
+  def include?("filter." <> _), do: Application.get_env(:electric, :profile_where_clauses?)
```
alco, line 2: I like the idea but find it hard to parse the call `OptionalSpans.include?()`. Changing the function name to `sample?` seems more intuitive. Or reverse its action and call it `skip?`.

I think my problem with `include?` is that it doesn't fit into the regular OT terminology so well.

alco (approved):
Beautiful!

I've only got one nitpick.

packages/sync-service/lib/electric/telemetry/optional_spans.ex
robacourt: Ok. I could do with merging this PR so I will address this in a separate PR

Issue #2323, opened 2025-02-11
The @core/sync-service@1.0.0-beta.13 release fails to start with default configuration


alco:
Trying to deploy the release on Fly as described [in the docs](https://electric-sql.com/docs/integrations/fly#deploy-electric) results in the following startup failure:

```
2025-02-11T10:58:55.108 runner[6e82e53b396d78] fra [info] Machine started in 1.168s

2025-02-11T10:58:55.514 app[6e82e53b396d78] fra [info] 2025/02/11 10:58:55 INFO SSH listening listen_address=[fdaa:3:606e:a7b:8a:ae35:6f2:2]:22 dns_server=[fdaa::3]:53

2025-02-11T10:58:55.941 health[6e82e53b396d78] fra [error] Health check on port 3000 has failed. Your app is not responding properly. Services exposed on ports [80, 443] will have intermittent failures until the health check passes.

2025-02-11T10:58:58.891 app[6e82e53b396d78] fra [info] =INFO REPORT==== 11-Feb-2025::10:58:58.890781 ===

2025-02-11T10:58:58.891 app[6e82e53b396d78] fra [info] Loading 140 CA(s) from otp store

2025-02-11T10:58:59.355 app[6e82e53b396d78] fra [info]

2025-02-11T10:58:59.355 app[6e82e53b396d78] fra [info] 10:58:59.355 pid=<0.2737.0> [warning] Database connection in lock_connection mode failed: ssl connect: closed

2025-02-11T10:58:59.356 app[6e82e53b396d78] fra [info]

2025-02-11T10:58:59.356 app[6e82e53b396d78] fra [info] 10:58:59.355 pid=<0.2737.0> [warning] Reconnecting in 2000ms

2025-02-11T10:58:59.361 app[6e82e53b396d78] fra [info] {exit,terminating,[{application_controller,call,2,[{file,"application_controller.erl"},{line,511}]},{application,enqueue_or_start,6,[{file,"application.erl"},{line,380}]},{application,ensure_all_started,3,[{file,"application.erl"},{line,359}]},{elixir,start_cli,0,[{file,"src/elixir.erl"},{line,195}]},{init,start_it,1,[]},{init,start_em,1,[]},{init,do_boot,3,[]}]}

2025-02-11T10:58:59.361 app[6e82e53b396d78] fra [info] Runtime terminating during boot (terminating)

[...]

PR #2324, opened 2025-02-11
Bring back the inclusion of telemetry_poller in releases

alco:
The `runtime: false` option in mix.exs was the wrong one to use: it excluded the dependency from release packaging, resulting in the release startup failure when `ELECTRIC_USAGE_REPORTING=true`.

The correct way to make telemetry_poller's default reporting disabled is to configure it as such.

Fixes https://github.com/electric-sql/electric/issues/2323.


msfstef (approved):
Nice catch!

PR #2327, opened 2025-02-11
feat: AbortController insertion on React.md

rgenge:
This changes will add AbortController option to clean up shapes subscriptions to avoid conflicts.


KyleAMathews (approved):
Thanks!

PR #2328, opened 2025-02-12
Fix connection error handling

alco:
The "passthrough" trick didn't work as planned since the error handling code has been moved around a bit after it got introduced.

To prevent a recursive call loop, we just need to replace the recursive call with different function call.


msfstef (approved):
Ha, clearly the better way which I avoided cause I didn't want to come up with a new name 🤭 probably needs a changeset too!

alco:
No need for a changeset because this hasn't caused any misbehaviour so far. This change basically makes future changes to error handling in this module less error prone.

PR #2329, opened 2025-02-12
Fix the fallback to no SSL for Fly Postgres

alco:


msfstef (approved):
Might be nice to move some of this special error detection (and the ones for IPv6/4) to its own module (we do some similar logic over on Stratovolt)

msfstef:
@alco should we merge this or is this superseeded by the async initialization?

msfstef:
We're seeing this issue pop up in cloud as well so I think it'd be good if we can merge this @alco

alco:
Rebased the original changes in this PR on top of `main`. Tested manually against Fly Postgres.

Issue #2311, opened 2025-02-10
Fix filtering of transactions already included in a shape's initial snapshot


alco:
### Background

When a shape is first created, Electric does two key things:

1. Start buffering transactions coming in from the PG replication stream that will be included in the shape.
2. Query the database for a snapshot of the shape's data. 

The 1st step has to be there so that Electric doesn't miss any transactions committed after the snapshot query starts. However, it means that some transactions received on the replication stream might already be included in the snapshot's data.

To avoid storing the same transaction both as part of the snapshot and in the shape log, we check every incoming transaction's `xid` against snapshot's `xmix` (obtained at the time the snapshot was taken by calling [`pg_current_snapshot()`](https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-PG-SNAPSHOT)) and skip those transactions that had committed by the time the snapshot started.

### Problem

The current logic has two flaws that need addressing:

1. The comparison between snapshot's `xmix` (64-bit non-wrapping transaction ID) and incoming transaction's `xid` (32-bit wrapping transaction ID) is implemented incorrectly.
2. Not enough information is stored on the snapshot to correctly filter all transactions. In addition to `xmix`, both `xmax` and `xip_list` (as returned by the `pg_current_snapshot()` mentioned above) also need to be stored. See the corresponding sub-issue for a detailed explanation.

PR #2331, opened 2025-02-12
Fix transaction filtering by taking into account snapshot's xmax and xip_list

### LINKED ISSUE #2313: Store the complete PG snapshot info on shape snapshots to filter all repeat transactions
@samwillis has had a brilliant insight:

We currently filter transactions incoming on the replication stream based on the snapshot's `xmin`. We need to (also) use `xmax` and `xip_list` to filter correctly. All three values are returned by PG's [`pg_current_snapshot()`](https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-PG-SNAPSHOT) function.

xmin is the lowest active transaction ID, there can be txids > xmin that have committed and so would appear in the snapshot. This, with our buffering of the replocation stream before fetfhing the snapshot, results in a condition where we can potentually have duplicate inserts between the snapshot and the shape log.

Image

The current filtering logic should be corrected by an additional clause: any transaction coming from the replication stream for which `xmin < xid and xid < xmax and xid not in xip_list` needs to be skipped. In pseudocode form:

```
# Transaction already part of the snapshot because it had committed before
# we started taking the snapshot.
defp handle_txn(%Transaction{xid: xid}, %{snapshot_xmin: xmin})
  when xid < xmin,
  do: <skip txn>

# Transaction commited sometime between xmin and the moment when we started
# taking the snapshot.
defp handle_txn(%Transaction{xid: xid}, %{snapshot_xmax: xmax, snapshot_xip_list: xip_list})
  when xid < xmax and xid not in xip_list,
  do: <skip txn>

# Transaction was either active at the time of taking the snapshot:
#   xid == xmin or xid in xip_list
# OR
# it committed after the snapshot had completed:
#   xid >= xmax
defp handle_txn(%Transaction{xid: xid}, _),
  do: <add txn to the log>
```

Once we have stored in the log the first transaction whose `xid` ≥ `xmax`, we can mark the snapshot such that the comparisons can be skipped for all future transactions. At this point, we are sure that any new transaction arriving on the replication stream is not part of the snapshot.

Skipping further comparisons will absolve us of the need to worry about the number of wraparounds of 32-bit transaction IDs.

alco:
Electric now stores the full pg snapshot info in the shape snapshot. Transaction filtering logic is then improved to skip additional transactions that are already included in the snapshot.

This change is backwards-compatible with existing shapes.

Fixes #2313.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -130,16 +130,22 @@ defmodule Electric.Shapes.Consumer do
   end
 
   def handle_cast(
-        {:pg_snapshot_known, shape_handle,
-         %{xmin: xmin, xmax: xmax, xip_list: xip_list} = pg_snapshot},
+        {:pg_snapshot_known, shape_handle, {xmin, xmax, xip_list}},
         %{shape_handle: shape_handle} = state
       ) do
     Logger.debug(
       "Snapshot known for shape_handle: #{shape_handle} xmin: #{xmin}, xmax: #{xmax}, xip_list: #{Enum.join(xip_list, ",")}"
     )
 
-    pg_snapshot = Map.put(pg_snapshot, :filter_txns?, true)
-    state = set_pg_snapshot(pg_snapshot, state)
+    state =
+      %{
+        xmin: xmin,
+        xmax: xmax,
+        xip_list: xip_list,
+        filter_txns?: true
+      }
+      |> set_pg_snapshot(state)
+
     handle_txns(state.buffer, %{state | buffer: []})
   end
 
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -262,6 +262,14 @@ defmodule Electric.Shapes.Consumer do
          %Transaction{xid: xid} = txn,
          %{pg_snapshot: %{xmin: xmin, xmax: xmax, xip_list: xip_list}} = state
        ) do
+    # xmin is the lowest active transaction ID, there can be txids > xmin that have
+    # committed and so would already be included in the shape's data snapshot.
+    # For this reason we store the full pg_snapshot and compare the incoming xid not only
+    # against xmin but also against xip_list, the list of transactions active at the time of
+    # taking the original data snapshot.
+    #
+    # See Postgres docs for details on the pg_snapshot fields:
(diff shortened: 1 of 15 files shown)


packages/sync-service/lib/pg_interop/postgrex/extensions/pg_snapshot.ex
```diff
@@ -0,0 +1,19 @@
...
+  use Postgrex.BinaryExtension, send: "pg_snapshot_send"
+  import Postgrex.BinaryUtils, warn: false
+
+  def encode(_state) do
+    quote location: :keep do
+      _ -> raise DBConnection.EncodeError, "encoding of type pg_snapshot not implemented"
+    end
+  end
+
+  def decode(_) do
```
msfstef, line 11: should we have some basic tests for this?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -241,20 +254,53 @@ defmodule Electric.Shapes.Consumer do
...
-        [xid: txn.xid, total_num_changes: txn.num_changes] ++
-          shape_attrs(state.shape_handle, state.shape)
+  defp handle_txn(txn, %{pg_snapshot: %{filter_txns?: false}} = state) do
+    handle_txn_in_span(txn, state)
+  end
 
-      OpenTelemetry.with_span("shape_write.consumer.handle_txn", ot_attrs, state.stack_id, fn ->
-        do_handle_txn(txn, state)
-      end)
+  defp handle_txn(
```
msfstef, line 261: very nice comments! should we potentially leave a link to [PG docs](https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-PG-SNAPSHOT-PARTS) about this as well? and do our tests cover some or all of these cases?

msfstef (approved):
Nice work!

packages/sync-service/lib/pg_interop/postgrex/extensions/pg_snapshot.ex
alco: Yes! I've added one for encode and one for decode.

packages/sync-service/lib/electric/shapes/consumer.ex
alco: Not right now. Tricky to test, will think of something. Filed as https://github.com/electric-sql/electric/issues/2311?issue=electric-sql%7Celectric%7C2348

msfstef (approved):

PR #2337, opened 2025-02-13
chore(sync-service): Separate stack telemetry from application telemetry

robacourt:
Our cloud version of electric has multiple stacks, one per tenant/database. Since we want metrics from each tenant, this PR separates the stack/tenant specific telemetry from the application/system telemetry. For cloud electric we will then have 1 application metric message and multiple stack metric messages being sent each time we poll.


msfstef (approved):
Haven't gone over the specifics but the splitting makes sense - LGTM! (tests are not running though)

PR #2339, opened 2025-02-15
Fix version format in README.md

nocksock:
The format for the version couldn't resolve:

> Because your app depends on electric >= 1.0.0-beta-1 which doesn't match any versions, version solving failed.
> ** (Mix) Hex dependency resolution failed


alco:
Thanks!

The latest beta version available is `1.0.0-beta.15`. Would you mind updating the version in README to that?

nocksock:
@alco updated to `1.0.0-beta.18`

alco (approved):

PR #2350, opened 2025-02-18
Fix file corruption in Electric.Utils.external_merge_sort()

alco:
When the input file is small enough to fit into a single chunk. The problem is that `File.stream!`, by default, looks for line break characters and normalizes any `\r\n` into `\n` when reading from the file. So the data it writes into the target file is different from the one it read from chunk_0 if it so happens that the latter has the sequence of bytes `<<13, 10>>` in it.

### DIFF
--- a/packages/sync-service/lib/electric/utils.ex
+++ b/packages/sync-service/lib/electric/utils.ex
@@ -532,9 +532,7 @@ defmodule Electric.Utils do
   def merge_sorted_files(paths, target_path, reader, sorter \\ &<=/2)
 
   def merge_sorted_files([path], target_path, _reader, _sorter) do
-    File.stream!(path)
-    |> Stream.into(File.stream!(target_path))
-    |> Stream.run()
+    File.copy!(path, target_path)
   end
 
   def merge_sorted_files(paths, target_path, reader, sorter) do
--- a/.changeset/fuzzy-dingos-doubt.md
+++ b/.changeset/fuzzy-dingos-doubt.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: Fix file corruption when doing external sort during compaction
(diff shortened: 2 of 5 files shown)


packages/sync-service/lib/electric/utils.ex
```diff
@@ -532,9 +532,7 @@ defmodule Electric.Utils do
   def merge_sorted_files(paths, target_path, reader, sorter \\ &<=/2)
 
   def merge_sorted_files([path], target_path, _reader, _sorter) do
-    File.stream!(path)
```
msfstef, line 535: could we not set the `:line_or_bytes` argument to `:bytes` to avoid this conversion of `\r\n`? (from [the docs](https://hexdocs.pm/elixir/1.18.1/File.html#stream!/3))

I'm assuming a stream was being used here to minimise memory usage, but since we use this for a single, small chunk I imagine `File.copy` might be a better fit anyway

msfstef (approved):
Looks good!

packages/sync-service/lib/electric/utils.ex
msfstef: In fact I think @icehaunter already has an open PR to do exactly that https://github.com/electric-sql/electric/pull/2310/files

msfstef: Might be worth incorporating this fix into that PR for the sake of keeping the extra tests as well though

alco: Yeah, I also found Ilia's PR. I've cherry-picked his changes here.

alco: By the way, looking at the implementation of Erlang's `:file.copy`, it reads data from the input file in chunks of 65536 bytes, so the memory usage is minimal.

It could be the reason behind why @icehaunter chose not to use it in his PR, but to me it seems better to rely on a high-level API implemented as part of Erlang in this case.

msfstef: Yep that's what I had read, and assumed the 2048 byte chunking was to keep a low memory overhead for streaming - but I agree that 64kb is not much, we could potentially do some benchmarking?

PR #2356, opened 2025-02-19
feat(experimental): MultiShapeStream and TransactionalMultiShapeStream APIs

samwillis:
This adds two new APIs to the @electric-sql/experimental package:

## `MultiShapeStream`

A multi-shape stream is a stream that can subscribe to multiple shapes. It ensures that all shapes will receive at least an `up-to-date` message from Electric within the `checkForUpdatesAfter` interval.

The emitted messages have an additional `shape` prop that is the name of the shape provided in the shapes object when constructing the `MultiShapeStream`.

```ts
const multiShapeStream = new MultiShapeStream({
  shapes: {
    shape1: {
      url: 'http://localhost:3000/v1/shape1',
    },
    shape2: {
      url: 'http://localhost:3000/v1/shape2',
    },
  },
})

multiShapeStream.subscribe((msgs) => {
  console.log(msgs)
})

// or with ShapeStream instances
const multiShapeStream = new MultiShapeStream({
  shapes: {
    shape1: new ShapeStream({ url: 'http://localhost:3000/v1/shape1' }),
    shape2: new ShapeStream({ url: 'http://localhost:3000/v1/shape2' }),
  },
})
```

## `TransactionalMultiShapeStream`

A transactional multi-shape stream is a multi-shape stream that emits the messages in transactional batches.
It uses the `lsn` metadata to infer transaction boundaries, and the `op_position` metadata to sort the messages within a transaction.

Note that you will only receive change massages on the main subscription (in transactional batches), and not any control massages.

```ts
const transactionalMultiShapeStream = new TransactionalMultiShapeStream({
  shapes: {
    shape1: {
      url: 'http://localhost:3000/v1/shape1',
    },
    shape2: {
      url: 'http://localhost:3000/v1/shape2',
    },
  },
})

transactionalMultiShapeStream.subscribe((msgs) => {
  console.log(msgs)
})

// or with ShapeStream instances
const transactionalMultiShapeStream = new TransactionalMultiShapeStream({
  shapes: {
    shape1: new ShapeStream({ url: 'http://localhost:3000/v1/shape1' }),
    shape2: new ShapeStream({ url: 'http://localhost:3000/v1/shape2' }),
  },
})
```

### DIFF
--- a/packages/experimental/src/multi-shape-stream.ts
+++ b/packages/experimental/src/multi-shape-stream.ts
@@ -25,7 +25,7 @@ interface MultiShapeStreamOptions<
       | ShapeStream<TShapeRows[K]>
   }
   start?: boolean
-  checkForUpdatesAfter?: number // milliseconds
+  checkForUpdatesAfterMs?: number // milliseconds
 }
 
 interface MultiShapeChangeMessage<
@@ -58,7 +58,7 @@ export interface MultiShapeStreamInterface<
   },
 > {
   shapes: { [K in keyof TShapeRows]: ShapeStream<TShapeRows[K]> }
-  checkForUpdatesAfter?: number
+  checkForUpdatesAfterMs?: number
 
   subscribe(
     callback: (
@@ -79,7 +79,7 @@ export interface MultiShapeStreamInterface<
 /**
  * A multi-shape stream is a stream that can subscribe to multiple shapes.
  * It ensures that all shapes will receive at least an `up-to-date` message from
- * Electric within the `checkForUpdatesAfter` interval.
+ * Electric within the `checkForUpdatesAfterMs` interval.
  *
  * @constructor
  * @param {MultiShapeStreamOptions} options - configure the multi-shape stream
@@ -118,7 +118,7 @@ export class MultiShapeStream<
 {
   #shapes: { [K in keyof TShapeRows]: ShapeStream<TShapeRows[K]> }
   #started = false
-  checkForUpdatesAfter?: number
+  checkForUpdatesAfterMs?: number
 
   #checkForUpdatesTimeout?: ReturnType<typeof setTimeout> | undefined
 
@@ -139,10 +139,10 @@ export class MultiShapeStream<
   constructor(options: MultiShapeStreamOptions<TShapeRows>) {
     const {
       start = true, // By default we start the multi-shape stream
-      checkForUpdatesAfter = 100, // Force a check for updates after 100ms
+      checkForUpdatesAfterMs = 100, // Force a check for updates after 100ms
       shapes,
     } = options
-    this.checkForUpdatesAfter = checkForUpdatesAfter
+    this.checkForUpdatesAfterMs = checkForUpdatesAfterMs
     this.#shapes = Object.fromEntries(
       Object.entries(shapes).map(([key, shape]) => [
         key,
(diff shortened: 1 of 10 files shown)


KyleAMathews (approved):
Looks like a great initial stab at this! Excited to see some demos built out with this!

packages/experimental/src/multi-shape-stream.ts
```diff
@@ -0,0 +1,415 @@
...
+ * const transactionalMultiShapeStream = new TransactionalMultiShapeStream({
+ *   shapes: {
+ *     shape1: new ShapeStream({ url: 'http://localhost:3000/v1/shape1' }),
+ *     shape2: new ShapeStream({ url: 'http://localhost:3000/v1/shape2' }),
+ *   },
+ * })
+ * ```
+ */
+
+export class TransactionalMultiShapeStream<
```
balegas, line 330: I think  to avoid confusion in the future, this should be documented clearly that transactional MultiShapeStream is experimental:
- builds on experimental headers
- doesn't work with compaction

balegas (approved):
I know you're working on other things, but capturing what we've said this morning. `op_position` allows us to build execution ordering across shapes and that is very strong, but that  metadata is not stable in the protocol.

Transaction execution ordering can only be built when you have the raw event log. It can be used to maintain local-store ref-intergrity, but integrity is an unsolved problem with maintaining sync-into-db integrity, since the problem of operation reordering exists for initial snapshot/compacted log.

If we say that local cache needs to have FK disabled to be able to handle initial sync, then the transactional boundaries are less important because we can always apply a change set.

packages/experimental/test/multi-shape-stream.test.ts
```diff
@@ -0,0 +1,510 @@
...
+          url: `${BASE_URL}/v1/shape`,
+          params: {
+            table: issuesTableUrl,
+          },
+        },
+      },
+    })
+
+    const subFn = vi.fn((_) => void 0)
+    const unsubscribeFn = multiShapeStream.subscribe(subFn)
```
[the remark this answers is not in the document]
samwillis: I don't think its new semantics, its in the main shape stream client - I copied it to match:

https://github.com/electric-sql/electric/blob/b5e3303fe63fc44f52b7f1a48216eda8036c5ff8/packages/typescript-client/src/client.ts#L592-L594

I agree that unsubscribing from a shape stream is a tine edge case.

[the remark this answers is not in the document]
KyleAMathews: ah ok never mind — I misread this as you were calling unsubscribe dirctly on the stream.

packages/experimental/src/multi-shape-stream.ts
samwillis: Agreed, it's in the experimental package, that makes it somewhat clear.

Note think that I will be using the `TransactionalMultiShapeStream` for pglite sync, and any sqlite based demos (node, Cloudflare D1). So there is also the argument that having it in "experimental" may actually be wrong if it's front and centre in much of our demos and content.

samwillis:
@balegas 

> op_position allows us to build execution ordering across shapes and that is very strong, but that metadata is not stable in the protocol.

I'm tempted to strip that bit out of the implementation, it gives very little.

packages/experimental/src/multi-shape-stream.ts
KyleAMathews: It's fine starting with experimental packages. We'll just need a process to migrate them to official packages at some point

PR #2359, opened 2025-02-19
feat(sync-service): index for array contains

robacourt:
This PR seeks to address trigger.dev's WAL lag issues by adding two optimisations to where clause filtering:
- Allow multiple conditions in a where clause to be optimised (not just one)
- Optimise where clauses that have a condition in the form 'array_field @> array_const'

## Allow multiple conditions in a where clause to be optimised (not just one)

This feature alone should halve the processing time for trigger.dev as they have many where clauses in the form `WHERE runtimeEnvironmentId = ? AND batchId = ?` where if a change matched the `runtimeEnvironmentId` it would then have to iterate through the `batchId`s. Now it doesn't - the `batchId` condition is indexed as well.

## Optimise where clauses that have a condition in the form 'array_field @> array_const'

Trigger.dev's other problematic where clauses are in the form `WHERE runtimeEnvironmentId = ? AND runTags @> ?` and this feature optimises the `@>` operation. The algorithm for this optimisation can be seen in the [InclusionIndex module](https://github.com/electric-sql/electric/pull/2359/files#diff-44c9f3ec68658f9ea6690a7608b99ad0463d7dcc8f1c95e9a2617e1d3d30a3d2).

Comparing the performance of this index against the current method shows quite a dramatic difference:
Screenshot 2025-02-19 at 19 42 59

Here you can see it performs well with 100k shapes (which for trigger.dev would be 100k shapes with the same `runtimeEnvironmentId`, in other words 100k shapes per user)
Screenshot 2025-02-19 at 19 43 47

Both of the charts above are based on a shape array size of 3 and a change array size of 10 which is typical for trigger.dev.

The algorithm seems to scale roughly linearly with change array size:
Screenshot 2025-02-24 at 16 12 34

The algorithm seems to scale well with shape array size for a fixed change array size (10):
Screenshot 2025-02-24 at 16 14 35

### DIFF
--- a/packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
+++ b/packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
@@ -11,7 +11,6 @@ defmodule Electric.Shapes.Filter.Indexes.EqualityIndex do
   alias Electric.Shapes.Filter.Index
   alias Electric.Shapes.Filter.Indexes.EqualityIndex
   alias Electric.Shapes.Filter.WhereCondition
-  alias Electric.Telemetry.OpenTelemetry
   require Logger
 
   defstruct [:type, :values]
@@ -55,7 +54,6 @@ defmodule Electric.Shapes.Filter.Indexes.EqualityIndex do
           MapSet.new()
 
         condition ->
-          OpenTelemetry.add_span_attributes(field: field)
           WhereCondition.affected_shapes(condition, record)
       end
     end
--- a/packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
+++ b/packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
@@ -1,0 +1,83 @@
+defmodule Electric.Shapes.Filter.Indexes.EqualityIndex do
+  @moduledoc """
+  Efficiently finds shapes that are affected by a change when the shape's where clause has `field = const` in it.
+
+  The index maps the values to the shapes that have that value as it's const in the `field = const` condition.
+
+  Rather than directly adding shapes, shapes are added to a `%WhereCondition{}` which represents can contain multiple
+  shapes and allows for further optimisations of other conditions in the shape's where clause.
+  """
+  alias Electric.Replication.Eval.Env
+  alias Electric.Shapes.Filter.Index
+  alias Electric.Shapes.Filter.Indexes.EqualityIndex
+  alias Electric.Shapes.Filter.WhereCondition
+  alias Electric.Telemetry.OpenTelemetry
+  require Logger
+
+  defstruct [:type, :values]
+
+  def new(type), do: %EqualityIndex{type: type, values: %{}}
+
+  defimpl Index.Protocol, for: EqualityIndex do
+    def empty?(%EqualityIndex{values: values}), do: values == %{}
+
+    def add_shape(%EqualityIndex{} = index, value, {shape_id, shape}, and_where) do
+      %{
+        index
+        | values:
+            Map.update(
+              index.values,
+              value,
+              WhereCondition.add_shape(WhereCondition.new(), {shape_id, shape}, and_where),
+              fn condition ->
+                WhereCondition.add_shape(condition, {shape_id, shape}, and_where)
+              end
+            )
+      }
+    end
+
+    def remove_shape(%EqualityIndex{} = index, shape_id) do
+      %{
+        index
+        | values:
+            index.values
+            |> Map.new(fn {value, condition} ->
+              {value, WhereCondition.remove_shape(condition, shape_id)}
+            end)
+            |> Enum.reject(fn {_table, condition} -> WhereCondition.empty?(condition) end)
+            |> Map.new()
+      }
+    end
+
+    def affected_shapes(%EqualityIndex{values: values, type: type}, field, record) do
+      case Map.get(values, value_from_record(record, field, type)) do
+        nil ->
+          MapSet.new()
+
+        condition ->
+          OpenTelemetry.add_span_attributes(field: field)
+          WhereCondition.affected_shapes(condition, record)
+      end
+    end
+
+    @env Env.new()
+    defp value_from_record(record, field, type) do
+      case Env.parse_const(@env, record[field], type) do
+        {:ok, value} ->
+          value
+
+        :error ->
+          raise RuntimeError,
+            message: "Could not parse value for field #{inspect(field)} of type #{inspect(type)}"
+      end
+    end
+
+    def all_shapes(%EqualityIndex{values: values}) do
+      for {_value, condition} <- values,
+          {shape_id, shape} <- WhereCondition.all_shapes(condition),
+          into: %{} do
+        {shape_id, shape}
+      end
+    end
+  end
+end
--- a/packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
+++ b/packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
@@ -21,31 +21,20 @@ defmodule Electric.Shapes.Filter.Indexes.EqualityIndex do
     def empty?(%EqualityIndex{values: values}), do: values == %{}
 
     def add_shape(%EqualityIndex{} = index, value, {shape_id, shape}, and_where) do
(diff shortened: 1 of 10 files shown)


packages/sync-service/lib/electric/shapes/filter/where_condition.ex
```diff
@@ -158,9 +162,16 @@ defmodule Electric.Shapes.Filter.WhereCondition do
...
+    Map.merge(
+      for {_field, index} <- indexes,
+          {shape_id, shape} <- Index.all_shapes(index),
+          into: %{} do
+        {shape_id, shape}
+      end,
+      for {shape_id, %{shape: shape}} <- other_shapes,
+          into: %{} do
+        {shape_id, shape}
+      end
```
alco, line 189: In this case, just calling `Map.new/2` seems more idiomatic:

```elixir
Map.new(other_shapes, fn {shape_id, %{shape: shape}} -> {shape_id, shape} end)
```

packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
```diff
@@ -0,0 +1,83 @@
...
+  alias Electric.Shapes.Filter.Indexes.EqualityIndex
+  alias Electric.Shapes.Filter.WhereCondition
+  alias Electric.Telemetry.OpenTelemetry
+  require Logger
+
+  defstruct [:type, :values]
+
+  def new(type), do: %EqualityIndex{type: type, values: %{}}
+
+  defimpl Index.Protocol, for: EqualityIndex do
```
alco, line 20: It's unusual to see `defimpl` inside a module definition. We usually define the struct module and then the protocols it implements at the same top level.

packages/sync-service/lib/electric/shapes/filter/indexes/inclusion_index.ex
```diff
@@ -0,0 +1,206 @@
...
+            message: "Could not parse value for field #{inspect(field)} of type #{inspect(type)}"
+      end
+    end
+
+    def all_shapes(%InclusionIndex{}) do
+      # for {_value, condition} <- values,
+      #     {shape_id, shape} <- WhereCondition.all_shapes(condition),
+      #     into: %{} do
+      #   {shape_id, shape}
+      # end
```
alco, line 195: TODO?

packages/sync-service/test/electric/shapes/filter_test.exs
```diff
@@ -358,6 +398,47 @@ defmodule Electric.Shapes.FilterTest do
       assert reductions < @max_reductions
     end
 
+    test "where clause in the form `array_field @> const_array` is optimised" do
```
alco, line 401: Maybe randomly choose between "@>" and "<@" to test both?

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
```diff
@@ -80,6 +80,22 @@ defmodule Electric.Shapes.Filter.WhereCondition do
...
+  defp optimise_where(%Func{
+         name: ~s("@>"),
+         args: [%Ref{path: [field], type: type}, %Const{value: value}]
+       })
+       when is_list(value) do
+    %{operation: "@>", field: field, type: type, value: value, and_where: nil}
+  end
+
+  defp optimise_where(%Func{
+         name: ~s("<@"),
```
alco, line 92: This is the only place where this mirror operator is mentioned. It should also be mentioned in the moduledoc and it needs to be added to the dispatch performed by `Index.new/1` to be considered at all, isn't it?

alco (approved):
Great job!

icehaunter (approved):
Very good job! :fire: My only nitpick would be nested update formatting, but I've relayed that to you in DMs, feel free to address or not

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
```diff
@@ -158,9 +162,16 @@ defmodule Electric.Shapes.Filter.WhereCondition do
...
-    for {_field, index} <- indexes, {shape_id, shape} <- Index.all_shapes(index), into: %{} do
-      {shape_id, shape}
-    end
-    |> Map.merge(other_shapes)
+    Map.merge(
+      for {_field, index} <- indexes,
+          {shape_id, shape} <- Index.all_shapes(index),
+          into: %{} do
+        {shape_id, shape}
+      end,
```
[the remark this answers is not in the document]
icehaunter: I think `for`-expressions are more optimal than streams on memory because they're unwrapped into a single function instead of a chained function calls with aggregates, but streams allow you piecemeal buildup of the pipeline while `for` is defined declaratively

packages/sync-service/lib/electric/shapes/filter/indexes/inclusion_index.ex
robacourt: Yes, I'm not sure how I didn't spot that. And there's clearly not a test for that either!

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
robacourt: We optimise `field @> const` but not `field <@ const` at the moment. 

This function head is for `const <@ field` which is equivalent to `field @> const` and so gets mapped to `"@>"` which is why we don't need a separate `Index.new/1` dispatch.

packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
robacourt: I agree it's more typical to see `defimpl` at the top level, but in this case it does have some disadvantages:

- You end up with `defimpl Electric.Shapes.Filter.Index.Protocol, for: Electric.Shapes.Filter.Indexes.EqualityIndex do` which is a bit of a mouthful. `defimpl Index.Protocol, for: EqualityIndex do` is easier to read. Sure I could add top level aliases but that's not normal either.
- Where does the moduledoc sit? If it's in the module, it's describing an algorithm that's not in the module. If it's in the defimpl then it can no longer be a moduledoc.

I can understand `defimpl` being at the top level if it's an extra thing the data structure can do, but in the case it's the _only_ thing the data structure can do, and nesting it in the module seems semantically sane.

What do you think?

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
robacourt: Ok, so you're seeing that I love comprehensions 😃

If the statement was on it's own I would agree with you. But in the context of the function it's in a merge with another comprehension, and personally I find the whole function easier to parse (nicer to look at?) as two similar comprehensions than a comprehension and a Map.new.

I do love comprehensions though. There are times when they are demonstrably more expressive and more efficient, so I use them a lot. I'm so used to the syntax, I end up using them at times when there's no performance benefit and expressiveness is more subjective.

packages/sync-service/lib/electric/shapes/filter/indexes/inclusion_index.ex
robacourt: Done: https://github.com/electric-sql/electric/pull/2359/commits/e5bdf525ed9466c7e9e90476320071cd393b8c47

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
alco: Ah, I see.

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
alco: Fine fine, comprehensions in both places it is.

packages/sync-service/test/electric/shapes/filter_test.exs
alco: Clarifying that I mean randomly choosing between `array_field @> const_array` and `const_array <@ array_field`, just for the sanity checking purpose.

robacourt: Great idea! Done: https://github.com/electric-sql/electric/pull/2359/commits/765c282522d6ca9202eb5ce9430fcce3abb0c45f

Issue #2364, opened 2025-02-23
Safari backoff failure - It stops fetching data.


gartz:
I've found a weird bug in the Safari, it stops to do the backoff due to the `while(true)` if the user switch to another app and go back to the safari.

What it will happen, is that if the debugger is open, it works as expected, but once the debugger is closed, switching apps for a while will cause Safari to kill the loop with no errors or any kind of notification to the app.

When back to the app, everything but the backoff loop will work, therefore no new fetch will be called, making the data in the mobile safari stale. Until the user refresh, or close/open the tab or pwa.

I know they had something to prevent timeout and interval abuse, but looks like if that triggers an infinite loop, the same will also be killed.

KyleAMathews:
Hmm that's bad. This is mobile safari? It looks like we might need to switch away from the while loop then

samwillis:
I've drafted a fix for this https://github.com/electric-sql/electric/pull/2366, but it will need to be tested on an actual iOS device to validate.

PR #2366, opened 2025-02-24
fix(typescript-client): Replace the `while(true)` in createFetchWithBackoff with a run closure + setTimeout

samwillis:
This replaces the `while (true)` with an explicit call to a `run` closure. It *should* fix the issue seeing in #2364 but will **probably need testing on an actual iOS device** to try and replicate.

iOS Safari tabs goes into a kind of hibernated mode when they are in the background. This throttles `setTimeout`s, and appears to cancel the `while (true)` task. By replacing the `while (true)` with an explicit call to a `run` closure in a `setTimeout` I'm hopeful we just see the next tick of the recursive timeout be throttled, paused and then restarted when the tab come back to the foreground.

### DIFF
--- a/packages/typescript-client/src/fetch.ts
+++ b/packages/typescript-client/src/fetch.ts
@@ -54,13 +54,11 @@ export function createFetchWithBackoff(
     let delay = initialDelay
     let attempt = 0
 
-    /* eslint-disable no-constant-condition -- we re-fetch the shape log
-     * continuously until we get a non-ok response. For recoverable errors,
-     * we retry the fetch with exponential backoff. Users can pass in an
-     * AbortController to abort the fetching an any point.
-     * */
-    while (true) {
-      /* eslint-enable no-constant-condition */
+    /* We re-fetch the shape log continuously until we get a non-ok response.
+     * For recoverable errors, we retry the fetch with exponential backoff.
+     * Users can pass in an AbortController to abort the fetching at any point.
+     */
+    const run = async (): Promise<Response> => {
       try {
         const result = await fetchClient(...args)
         if (result.ok) return result
@@ -79,19 +77,27 @@ export function createFetchWithBackoff(
           throw e
         } else {
           // Exponentially backoff on errors.
-          // Wait for the current delay duration
-          await new Promise((resolve) => setTimeout(resolve, delay))
-
+          // Capture current delay before increasing it
+          const currentDelay = delay
           // Increase the delay for the next attempt
           delay = Math.min(delay * multiplier, maxDelay)
 
           if (debug) {
             attempt++
-            console.log(`Retry attempt #${attempt} after ${delay}ms`)
+            console.log(`Retry attempt #${attempt} after ${currentDelay}ms`)
           }
+
+          // Return a new promise that will resolve with the next run after the delay
+          return new Promise((resolve, reject) => {
+            setTimeout(() => {
+              run().then(resolve).catch(reject)
+            }, currentDelay)
+          })
         }
       }
     }
+
+    return run()
   }
 }
 
--- a/.changeset/six-apes-push.md
+++ b/.changeset/six-apes-push.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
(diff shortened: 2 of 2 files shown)


KyleAMathews:
@samwillis thanks for picking this up! Perhaps just push up little app somewhere for testing? I can borrow my wife's iphone (or son's tablet haha) and other's can test as well.

KyleAMathews:
Next step is to reproduce original behavior on IOS & that this PR fixes it.

kevin-dp:
> Next step is to reproduce original behavior on IOS & that this PR fixes it.

I've been able to reproduce the original behavior with Safari on my iPhone.
Sometimes i have to try multiple times for this to happen but when it happens only reloading the page works.

packages/typescript-client/src/fetch.ts
```diff
@@ -79,19 +77,27 @@ export function createFetchWithBackoff(
...
+          return new Promise((resolve, reject) => {
+            setTimeout(() => {
+              run().then(resolve).catch(reject)
+            }, currentDelay)
+          })
         }
       }
     }
+
+    return run()
```
kevin-dp, line 100: Since we don't `await` anything inside the async function that is returned on L50 (only within the nested `run` function) we can remove the `async` keyword on L50.

kevin-dp (approved):
The fix looks good to me. I'll try to test this on iOS before we merge this.

kevin-dp:
Last week I was able to reproduce the problem with our cloud dashboard.
Reproducing it in dev is a bit harder because:
- the page is served from localhost or a LAN ip address
- the page is served over HTTP instead of HTTPS

Because of this, Safari seems to always reload the page when you go back to the tab.
Using ngrok we can reliably reproduce the bug as follows:
- Run ngrok to expose both `http://localhost:3000` (local PG) and `http://localhost:5173` (local dev server)
- Run `pnpm backend:up`
- Run `VITE_ELECTRIC_URL="https://<your subdomain for PG>.ngrok.app" pnpm vite`
- In Safari on iOS visit `https://<your subdomain for dev server>.ngrok.app`
- Lock your iPhone
- Perform some update
- Unlock your iPhone, the app should now be in the old state and not react to updates

kevin-dp:
I also tested this PR with Safari on iOS and the problem persists.
I think another problem we're having is the while loop in `client.ts`' `#start` method:
```ts
while (
  (!this.options.signal?.aborted && !this.#isUpToDate) ||
  this.options.subscribe) {
  ...
}
```

When `this.options.subscribe` is `true` this is essentially another infinite loop that may get cancelled altogether.

I think a good solution would be to introduce the possibility to pause the stream. Then, let apps pause and resume the stream when to app moves to the background or back into the foreground. That way the stream is explicitly resumed and should work independent of browser-specific quirks about loop/timeout/interval suspension/cancellation. We can have utilities such that developers pass the window object to the stream and it automatically adds listeners to pause and resume the stream.

KyleAMathews:
> the stream is explicitly resumed and should work independent of browser-specific quirks about loop/timeout/interval suspension/cancellation

I'm not sure I understand what you mean here? We don't want developers to have to manually pause/resume the stream if that's what you're saying. It should be automatic like it's (supposed) to be now.

kevin-dp:
> > the stream is explicitly resumed and should work independent of browser-specific quirks about loop/timeout/interval suspension/cancellation
> 
> I'm not sure I understand what you mean here? We don't want developers to have to manually pause/resume the stream if that's what you're saying. It should be automatic like it's (supposed) to be now.

Yeah, they could pass the window object to `useShape` and then we can automate this

kevin-dp:
> Can we check for the window object and grab it directly?

Sure! Will just need to check how to make TS happy 😛

KyleAMathews:
Trust in Claude :-D

It would be optional though yeah as when the client runs in a server environment, it doesn't run the risk of being cut off.

kevin-dp (changes requested):
I opened a separate PR https://github.com/electric-sql/electric/pull/2435 that solves the problem.

@samwillis Could you review my PR and test it please?

kevin-dp:
I'm closing this one as we merged my fix in https://github.com/electric-sql/electric/pull/2435

Issue #2360, opened 2025-02-20
Replicate entire table if publication row filtering is not available


msfstef:
Some where clauses might not be supported by the publication row filtering on PG - we could fall back to replicating the entire table while those where clauses are present in active shapes rather than rejecting them outright. Otherwise we can make the row filtering configurable.

See [this Discord thread](https://discord.com/channels/933657521581858818/1341967559758581921) for the discussion that raised this, and we've already seen issues with enums and other custom data types.

Issue #1778, opened 2024-10-01
Evaluate the performance impact of replication filtering


balegas:
We're setting replication slot filtering based on shapes that are requested on Electric to prevent evaluating rows that won't match any shape. This approach doesn't seem very sustainable as the number of shapes grow. See also https://github.com/electric-sql/electric/issues/1774#issuecomment-2386506510. 

We should evaluate the impact of this strategy and compare it with a filtering approach that only filters out columns (and tables) that aren't used by any shape. 

We will decide on further steps once we get some numbers.

Issue #1547, opened 2024-08-21
Garbage Collect Shapes


balegas:
Diamonds are forever, but shapes shouldn't.

Once a shape is created it keeps being updated for eternity, however it might not being consumed anymore because the subscribers are gone, because the origin Postgres no longer exists, etc.

We should have a garbage collection policy in place that ensures that the server stops tracking shapes that are no longer necessary.

Also this is necessary for shedding server load when:
- memory is high
- disk is getting full
- replication lag is growing (server can't process logical replication log fast enough due to too many shapes)

kevin-dp:
What if this would be the responsibility of the application?
The application is the one that requested the shape in the first place, when it no longer needs a shape, it can send a DELETE request. I'm saying this because we can try to be clever and GC a shape when there are no more subscribers, but an application may unsubscribe from a shape (knowing that it will re-subscribe later) but then when it comes back later it needs to refetch it from scratch because the sync service tried to be clever.

balegas:
I was looking at is more as of a maintenance thing. You'd have a long period before cleaning up a shape. Could also be driven by disk-usage.

KyleAMathews:
Yeah +1 to GC vs explicit maintenance by clients as that's more work to implement. It'd also be hard for an app to know when a shape should be deleted as apps are by definition distributed apps so collecting the state across all the parts of the system on what shapes are in use or not would be quite hard.

KyleAMathews:
A simple model for GC is LRU — track which shapes have gotten requests and when. Then when the server deletes shapes, it starts with the least recently used shapes.

Another model to add on top of this is to also track when a GCed shape is recreated and then score that shape def higher to have it be less likely to be deleted. This would help keep shapes that are e.g. visited once a week vs. shapes that are created and used for a brief period then never again.

balegas:
Anyone taking this task, let's do a bit of a design session before jumping into implementation (can be a discussion in the issue).

balegas:
Let's wait until knowing more about memory usage from shapes (cc @robacourt) before picking this issue.

We want Electric to be able to have lot's of shapes before GCing shapes

msfstef:
@balegas @robacourt should we close this issue and open new ones for specific improvements?

balegas:
yep, lets open again when we work more on it

PR #2367, opened 2025-02-24
feat: Fallback to include whole table in replication if where clauses unsupported

msfstef:
Fixes https://github.com/electric-sql/electric/issues/2360

This will make the running Electric fallback to replicating whole tables if it receives any shapes with unsupported where clauses (e.g. enums, varchar with `IN` checks, user-defined data types in general, and who knows what else).

There is no "recovery" mechanism to return to row-filtering as the Postgres error does not allow for an easy way to check which where clause caused the issue - once we go to relation-only filtering we stay there, like we would if an active shape had no where clause or if we were in PG14.

Ideally we would detect where clauses that are unsupported at the relation filter processing level, so we can fine tune that, but until then this fallback makes sure that Electric works even if an unsupported where clause is provided.

As discussed in [this Discord thread](https://discord.com/channels/933657521581858818/1341967559758581921), we could also have a configuration flag and better errors to avoid this sort of radical fallback, but we opted for an "always works" approach here. IIRC some benchmarking had shown that our filtering is fast enough that the PG level filtering might not be as important anyway, although the limiting of transmitted data is definitely nice (despite several issues we have with not being able to limit columns replicated etc).

This is related with https://github.com/electric-sql/electric/issues/1778 , and I'm also referencing https://github.com/electric-sql/electric/issues/1831 as we had encountered many limitations to row filtering which has led to this proposed change.

We can definitely improve this by detecting unsupported where clauses, checking filter diffs to know what caused the issue and reverting back after, periodically attempting to revert back to row-filtering, and an array of different approaches, but this allows all where clauses to be accepted and Electric to adjust accordingly.

### DIFF
--- a/packages/sync-service/lib/electric/replication/publication_manager.ex
+++ b/packages/sync-service/lib/electric/replication/publication_manager.ex
@@ -16,6 +16,7 @@ defmodule Electric.Replication.PublicationManager do
     :relation_filter_counters,
     :prepared_relation_filters,
     :committed_relation_filters,
+    :row_filtering_enabled,
     :update_debounce_timeout,
     :scheduled_updated_ref,
     :retries,
@@ -30,6 +31,7 @@ defmodule Electric.Replication.PublicationManager do
            relation_filter_counters: %{Electric.relation() => map()},
            prepared_relation_filters: %{Electric.relation() => __MODULE__.RelationFilter.t()},
            committed_relation_filters: %{Electric.relation() => __MODULE__.RelationFilter.t()},
+           row_filtering_enabled: boolean(),
            update_debounce_timeout: timeout(),
            scheduled_updated_ref: nil | reference(),
            waiters: list(GenServer.from()),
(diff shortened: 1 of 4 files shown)


KyleAMathews:
Love this. Make it right first (valid shape requests always work) and then make it fast as necessary (e.g. periodically retry to optimize publication slots). The later could be looked at with #1547 as when we GC is a good point to also consider other optimizations.

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -255,9 +295,28 @@ defmodule Electric.Replication.PublicationManager do
       publication_name
     )
 
-    :ok
+    {:ok, state}
   rescue
```
icehaunter, line 299: nitpick: You can move the pattern match into the `rescue` clauses to avoid 2 indentation levels

icehaunter (approved):
Great work & clean implementation

packages/sync-service/lib/electric/replication/publication_manager.ex
msfstef: that's what I tried initially but it wasn't letting me - my understanding is that I can only match the error type in a `rescue` but not a full pattern match (?)

msfstef: I found that a combination of `catch` and `rescue` can do the trick - not sure if that's idiomatic
```elixir
catch
    _kind, %Postgrex.Error{postgres: %{code: :feature_not_supported}} ->
      blah blah
  rescue
    err ->
      {:error, err}
```

PR #2371, opened 2025-02-25
Fix data corruption when concatenating multiple input files into a single output file

alco:
Follow-up to #2350.

### DIFF
--- a/packages/sync-service/lib/electric/utils.ex
+++ b/packages/sync-service/lib/electric/utils.ex
@@ -558,8 +558,7 @@ defmodule Electric.Utils do
       chunk
       |> Enum.sort(sorter)
       |> Stream.map(fn {_, value} -> value end)
-      |> Stream.into(File.stream!(chunk_path))
-      |> Stream.run()
+      |> write_stream_to_file!(chunk_path)
 
       chunk_path
     end)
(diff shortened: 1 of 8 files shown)


packages/sync-service/lib/electric/utils.ex
```diff
@@ -563,13 +574,16 @@ defmodule Electric.Utils do
...
-    |> Stream.concat()
-    |> Stream.into(File.stream!(into))
-    |> Stream.run()
+  @doc """
+  `File.copy!()` uses `:file.copy()` under the hood. The latter reads the contents from the source
+  file and writes them into the destination file in chunks of 65536 bytes.
+  """
+
+  @spec concat_files([Path.t()], Path.t()) :: :ok
+  def concat_files(source_paths, dst_path, modes \\ [:write]) do
```
icehaunter, line 583: This is less efficient without an explicit `:raw` mode, resulting in a separate process and message passing within the `File.copy` call. `File.stream` does that by default. Lets add that here too!

icehaunter (changes requested):

packages/sync-service/lib/electric/utils.ex
alco: Done. Rebased on top of your branch that fixes the file merging and updated the file modes.

alco:
@icehaunter Fair enough, I'll remove those.

I have some ideas about the necessity of having utility functions for working with files but they're not fully formed and not critical for our codebase in its current state. Thanks for pushing back on those particular changes.

And thanks for the review!

icehaunter (approved):

PR #2375, opened 2025-02-25
sync-service: make configuration easier and more accessible 

magnetised:
- remove the requirement to obfuscate db connection passwords in your config. in e.g. dev mode it's common to just have your db creds in config/dev.exs where you don't have access to the functions in electric. so rather than raise in this common scenario, just make the obfuscate_password function idempotent and apply it every time
- make the library mode the default
- add functions to application to get a proper configuration with the possibility of overrides
- make long_poll_timeout configurable because, why not? everything else is....

### DIFF
--- a/.changeset/brown-plants-count.md
+++ b/.changeset/brown-plants-count.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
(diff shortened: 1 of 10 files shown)


robacourt (approved):
LGTM

alco (approved):

PR #2376, opened 2025-02-25
sync-service: improve shapes api encapsulation

magnetised:
- Api.predefined_shape/2 allows for configuring an endpoint with a pre-defined shape
- All http logic is now encapsulated by the api, including not-modified handling and cache-control headers


robacourt (approved):
LGTM

PR #2378, opened 2025-02-25
sync-service: improve relation_to_sql quoting logic

magnetised:
- quote postgres reserved words
- quote names that start with a number


robacourt (approved):
Nice!

PR #2381, opened 2025-02-26
feat: add old values to `replica=full`

icehaunter:


KyleAMathews:
Minimizing data over the wire is generally a good idea. So only including overwritten values & not including old values for deletes seems like a good idea.

If we're going to add a new top-level key to messages, we'll need to make sure modify all the relevant documentation & typings.

samwillis:
> not including old values for deletes seems like a good idea

To clarify, for D2 we need the old values on deletes.

To take a step back to user requirements, I see three users:

1. Only need the changes since the last message, they are covered by the default behaviour
2. Want the full record state on each update message. Makes syncing into things like Redis or other KV stores easer. They don't need the previous value on message
3. People doing any sort of incremental compute on the stream, and they want to minimise cached state (D2). They need the previous value on both the update and delete messages.

Our current `replica=full` sits half way between user 2 and 3. This new `replica=complete` services user 3.

The quest is:

- do we want one option that covers users 2 + 3, and make it minimise data on the wire by only including the changes values in `old_value`?
- or do we have two options, once that is optimal for each of user 2 and 3? For this we can remove the value prop on the delete message of the current `replica=full`.

thruflo:
I think replica=full will the old changed values on update solves both 2 and 3 with minimal overhead. Avoids adding the additional mode.

thruflo:
Inserts don't have an old. Deletes already get the whole old. Replica=full already serves the whole new, just need the old changed. Either tweaking `full` to provide this, or adding `complete` as a mode both makes sense, with the dedupe of values in either case.

samwillis:
Yep, I'm inclined to have just the `replica=full` option (and not introduce a third), with the `old_values` where it only contains the old version of columns that have changed.

So to reconstruct the full old value its `msg.value + msg.old_value`.

@KyleAMathews what do you think?

samwillis:
@icehaunter I've pushed updates to the client. It now parses `old_value` as it did for `value`, and I've changed the type of `old_value` to `Partial<T>`.

I have a separate PR for updated docs stacked on this: https://github.com/electric-sql/electric/pull/2385

Was there any more you needed to do on the backend for this, any more tests?

If not is it ready for a review?

icehaunter:
@samwillis we need to test compaction with it, and update elixir client. Other than that, nothing

robacourt (approved):

Issue #2372, opened 2025-02-25
Slow shape transaction processing causes `ReplicationClient` to lose connection


msfstef:
The processing of transactions is done from the `ReplicationClient` via a `GenServer.call` to the `ShapeLogCollector`, which will fan out the transaction for shape consumers to process and then return.

https://github.com/electric-sql/electric/blob/4ab04c4446fb2b777235960113084c23e85baa4f/packages/sync-service/lib/electric/postgres/replication_client.ex#L284-L293

If any of the shape consumers is slow to process the transaction, e.g. because of IO, then during that time the `ReplicationClient` is blocked on this call and cannot reply to PG's keep-alive messages.

This can lead to cases where, even though the processing was slow, it successfully returns, but at that point an attempt to reply on the connection happens and an `ssl send: closed` or similar connection error occurs, crashing the `ReplicationClient` and leading to a restart and re-processing of the same transaction once the connection is re-established.

This seems to happen with SSL connections, and my hypothesis is that because keep-alives are not being replied to, the connection has closed by the time we actually try to acknowledge the transaction that _we have_ processed.

You can reproduce this with an SSL enabled DB and artificial slowdown of storage processing, like a `Process.sleep(10000)` in the consumer process. The SSL connection seems to die within 5-10 seconds, and setting timeouts didn't seem to change this.

Thankfully the processing of transactions is/should be idempotent as they are indexed on their LSN/offset, although I don't know how that plays with compaction.

My suggestion is to make the transaction processing asynchronous, such that we can reply to keep-alives while the transaction is being processed, while deferring/ignoring subsequent transactions until the previous one is done processing. Not sure if the latter part is easily done.

PR #2382, opened 2025-02-26
fix: Drop transactions if already processed

msfstef:
Fixes https://github.com/electric-sql/electric/issues/2372

In the above issue, the replication connection might drop after a slow processing of a transaction, leading to a replication connection restart and thus a replay of the transaction that has already been processed.

Rather than trying to maintain the connection alive, which is hard to do without processing async, and if we do then we cannot throttle/apply backpressure to the replication, we can just make sure we discard actually processed transactions.

Currently we handle these replayed transactions by idempotently writing the transactions, although I'm unsure if that holds under compaction as well. Instead, since we are tracking the last processed LSN within the `ShapeLogCollector`, I think we should discard transactions that we have already seen wholesale.

By doing this, I think we can close https://github.com/electric-sql/electric/issues/2372 because the replication connection restarting is a low overhead and without potential negative consequences because of replayed transactions.

Changes:
- Use the `State` struct for all constructions of the state in `ReplicationClient`
- Update the `applied_wal` to the most recently applied transaction commit LSN, rather than the replication message `wal_end` which from what I understand (and see in testing) can be farther ahead.
- Add clause to `update_applied_wal` for cases where we already have updated it to a higher LSN (see [error](https://electricsql-04.sentry.io/issues/25080523/?environment=production&project=4508410462404688&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=24h&stream_index=0))
- Only advance the replication slot on heartbeats if we are not actively collecting a transaction - I don't think we actually get interleaved heartbeats between transactions being replicated, but felt like this safety guard ensures we don't accidentally advance the slot for transactions we haven't actually processed.
- Unify `last_processed_lsn` accessing via a `PersistentReplicationState` module
- Reset the `last_processed_lsn` in case of a timeline change


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -218,9 +225,17 @@ defmodule Electric.Replication.ShapeLogCollector do
...
         state.persistent_kv,
         "#{state.stack_id}:last_processed_lsn",
-        state.last_seen_lsn
+        Postgres.Lsn.to_integer(state.last_seen_lsn)
       )
 
     state
   end
+
+  defp get_last_processed_lsn(%{persistent_kv: persistent_kv, stack_id: stack_id} = _opts) do
```
alco, line 234: On main, we have another function that reads `last_processed_lsn` from PersistentKV - `get_global_last_seen_lsn`. It'd be nice to unify them to avoid spreading the internal knowledge about PKV's key structure across the codebase.

alco: On timeline change we nuke all shapes, so all persistent state related to maintaining shape integrity must also be cleared.

msfstef: @alco https://github.com/electric-sql/electric/pull/2382/commits/5abbc926b938bb157650979e830d14450a863473 unifies these calls into a single abstraction and resets it on timeline change

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -312,16 +312,24 @@ defmodule Electric.Connection.Manager do
     case start_connection_pool(state.connection_opts, state.pool_opts) do
       {:ok, pool_pid} ->
         state = mark_connection_succeeded(state)
-        # Checking the timeline continuity to see if we need to purge all shapes persisted so far.
-        check_result =
+        # Checking the timeline continuity to see if we need to purge all shapes persisted so far
+        # and rest any replication related persistent state
```
alco, line 316: You mean "reset"?

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -312,16 +312,24 @@ defmodule Electric.Connection.Manager do
...
+        timeine_changed =
```
alco, line 317: Typo and it's nice to add `?` to names of boolean variables, i.e. `timeline_changed?`.

packages/sync-service/lib/electric/replication/persistent_replication_data.ex
```diff
@@ -0,0 +1,50 @@
+defmodule Electric.Replication.PersistentReplicationState do
```
alco, line 1: File name and the module name do not match.

alco (approved):

packages/sync-service/lib/electric/replication/persistent_replication_data.ex
msfstef: whoops refactored module name and forgot

robacourt (approved):

PR #2387, opened 2025-02-27
Improve external api of elixir client

magnetised:
- improve external api of elixir client
- encapsulate options and delete more fully under shapes api
- remove hard-coding of stack events registry name
- fix problems with external config api and expose http endpoint services
- expose default replica mode as a function


robacourt (approved):
Nice!

PR #2401, opened 2025-03-03
fix: make blog post publication date update on navigation

ldirer:
I noticed that the publication dates on the blog posts were not updated correctly on navigation.

The date loads correctly on the first page, but is not updated on navigation.

**Reproduction**:

1. Go to https://electric-sql.com/blog/2022/05/03/introducing-rich-crdts
See this:
image

2. Navigate to the first blog post in the list, see this:
image


KyleAMathews (approved):
Thanks!

Issue #2362, opened 2025-02-21
Secure by default


KyleAMathews:
This is another layer of security for protecting access to the Electric API when it's deployed on the internet. Not all hosting platforms make it easy to lock down access so we should have our own way of securing access.

samwillis:
Should this be a required param - secure by default - with an opt in `INSECURE=true` that can be used in development or when the user knows they are secure because of their network configuration?

balegas:
prod instance should prevent any requests going through if you don't setup a secret

balegas:
I was chatting with @kevin-dp and @samwillis to clarify how to implement this.

Context: Electric should always be running behind a proxy because it exposes the database. The idea is that by default Electric requires passing an API_SECET such that only requests that present that secret will be able accepted by the server. A developer has to setup that key in any of the clients that are authorized to make requests --- the proxy.

If we follow this implementation route, the server and any authorized client needs to be updated at the same time in the uncommon scenario of changing the secret. This sounds like an 80/20 solution, but we need to consider that this comes with downtime and operational complexity, since a user might have multiple running apps.

The best approach to me seems to do this using a app_key generation endpoint, such that proxy client with a non-expired/revoked key would not lose access to the server even if the secret changes.

kevin-dp:
> Context: Electric should always be running behind a proxy because it exposes the database. The idea is that by default Electric requires passing an API_SECET such that only requests that present that secret will be able accepted by the server. A developer has to setup that key in any of the clients that are authorized to make requests --- the proxy.

With this secret users will still need to run a proxy because it is unsafe to send the requests with the secret from the client. So the client needs to send a request to the proxy which then sends a request to Electric with the Secret included. Since the user is already running a proxy, it seems that in most cases you would just restrict the Electric server to only accept connections from that proxy server (e.g. via some custom VPC or whatever). Now, i hear that some platforms do not allow securing it this way but how common is that? Which cloud platforms don't support this? Is that a real concern for us right now, do we have any clients that we want to onboard that are in this situation?

> The best approach to me seems to do this using a app_key generation endpoint, such that proxy client with a non-expired/revoked key would not lose access to the server even if the secret changes.

[...]

PR #2404, opened 2025-03-05
feat!: Secure mode by default

kevin-dp:
Fixes [secure by default issue](https://github.com/electric-sql/electric/issues/2362).

Accompanying docs PR: https://github.com/electric-sql/electric/pull/2405

Introduces the `ELECTRIC_SECRET` env var that should be used to set an API secret that needs to be provided with all shape requests, unless we are running in insecure mode which can be set with an env var `ELECTRIC_INSECURE=true`.

**This is a breaking change because:**
- Electric must be started with one of the two env vars
- in secure mode, API calls must provide the secret on all shape requests

### DIFF
--- a/integration-tests/tests/secure-mode.lux
+++ b/integration-tests/tests/secure-mode.lux
@@ -50,12 +50,12 @@
   !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1"
   ??HTTP/1.1 401 Unauthorized
 
-  # rejected is secret is wrong
-  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1" --header "electric-secret: wrong-secret"
+  # rejected if secret is wrong
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&api_secret=wrong-secret"
   ??HTTP/1.1 401 Unauthorized
 
   # accepted if secret is provided and correct
-  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1" --header "electric-secret: test-secret"
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&api_secret=test-secret"
   ??HTTP/1.1 200 OK
 
 [cleanup]
--- a/integration-tests/tests/secure-mode.lux
+++ b/integration-tests/tests/secure-mode.lux
@@ -1,0 +1,62 @@
+[doc Verify that the sync service rejects shape requests without a secret when running in secure mode]
+
+[include _macros.luxinc]
+
+[global pg_container_name=secure-mode__pg]
+
+###
+
+## Start a new Postgres cluster
+[invoke setup_pg "" ""]
+
+## Add some data
+[invoke start_psql]
+[shell psql]
+  """!
+  CREATE TABLE items (
+    id UUID PRIMARY KEY,
+    val TEXT
+  );
+  """
+  ??CREATE TABLE
+
+  """!
+  INSERT INTO
+    items (id, val)
+  SELECT
+    gen_random_uuid(),
+    '#' || generate_series || ' test val'
+  FROM
+    generate_series(1, 10);
+  """
+  ??INSERT 0 10
+
+## Start the sync service.
+[invoke setup_secure_electric "test-secret"]
+
+[shell electric]
+  ??[info] Starting replication from postgres
+
+[shell client]
+  !curl -v -X GET "http://localhost:3000/v1/health"
+  # health endpoint is not protected with the secret
+  ??HTTP/1.1 200 OK
+
+  # options request is not protected with the secret
+  !curl -v -X OPTIONS "http://localhost:3000/v1/shape?table=items&offset=-1"
+  ??HTTP/1.1 204 No Content
+
+  # rejected if secret is missing
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1"
+  ??HTTP/1.1 401 Unauthorized
+
+  # rejected is secret is wrong
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1" --header "electric-secret: wrong-secret"
+  ??HTTP/1.1 401 Unauthorized
+
+  # accepted if secret is provided and correct
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1" --header "electric-secret: test-secret"
+  ??HTTP/1.1 200 OK
+
+[cleanup]
+  [invoke teardown]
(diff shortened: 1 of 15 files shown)


robacourt (approved):
Nicely done. Nice clear error/warning messages too 👍

integration-tests/tests/secure-mode.lux
```diff
@@ -0,0 +1,62 @@
...
+  # options request is not protected with the secret
+  !curl -v -X OPTIONS "http://localhost:3000/v1/shape?table=items&offset=-1"
+  ??HTTP/1.1 204 No Content
+
+  # rejected if secret is missing
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1"
+  ??HTTP/1.1 401 Unauthorized
+
+  # rejected is secret is wrong
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1" --header "electric-secret: wrong-secret"
```
KyleAMathews, line 54: Let's make `electric-secret` a query parameter to keep it in line with cloud's `source_secret`.

KyleAMathews (changes requested):
Thanks for picking this up! Looks great

kevin-dp:
TODO: generate a changeset

integration-tests/tests/secure-mode.lux
KyleAMathews: It's to avoid preflight requests which slow things down a lot https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

It's not a big deal tbh either way as your client code isn't the one generally directly making the request to the electric server so the secret isn't shared.

I did want to make the change as in the future, there'll be a way for people to generate tokens specific to individual shapes e.g. for a non-logged in site where a lot of people are subscribed to the same data. We want to avoid the preflights in that scenario.

packages/sync-service/lib/electric/config.ex
```diff
@@ -374,4 +374,23 @@ defmodule Electric.Config do
...
     end
   end
+
+  def validate_security_config!(secret, insecure) do
+    cond do
+      insecure && secret != nil ->
+        raise "Cannot set both ELECTRIC_SECRET and ELECTRIC_INSECURE=true"
+
+      !insecure && secret == nil ->
+        raise "Must set ELECTRIC_SECRET unless ELECTRIC_INSECURE=true. Setting ELECTRIC_INSECURE=true risks exposing your database, only use insecure mode if you know what you are doing."
```
KyleAMathews, line 384:
```suggestion
        raise "You must set ELECTRIC_SECRET unless ELECTRIC_INSECURE=true. Setting ELECTRIC_INSECURE=true risks exposing your database, only use insecure mode in development or you've otherwise secured the Electric API"
```

packages/sync-service/lib/electric/config.ex
```diff
@@ -374,4 +374,23 @@ defmodule Electric.Config do
...
+
+      true ->
+        if insecure do
+          Logger.warning(
+            "Running in insecure mode, this risks exposing your database, only use insecure mode if you know what you are doing."
```
KyleAMathews, line 389:
```suggestion
            "Electric is running in insecure mode — this risks exposing your database — only use insecure mode in development or if you've otherwise secured the Electric API."
```

packages/sync-service/lib/electric/config.ex
```diff
@@ -374,4 +374,23 @@ defmodule Electric.Config do
       {:error, message} -> raise Dotenvy.Error, message: message
     end
   end
+
+  def validate_security_config!(secret, insecure) do
+    cond do
+      insecure && secret != nil ->
+        raise "Cannot set both ELECTRIC_SECRET and ELECTRIC_INSECURE=true"
```
KyleAMathews, line 381:
```suggestion
        raise "You cannot set both ELECTRIC_SECRET and ELECTRIC_INSECURE=true"
```

KyleAMathews (commented):
Some copy suggestions

kevin-dp:
@KyleAMathews I moved the secret to a query parameter named `api_secret`. I used an underscore to be consistent with naming of `source_secret` in cloud. Let me know if you would prefer `api-secret` or another name.

kevin-dp:
In this PR i also added a `secret` option to the `ShapeStream`.
Not sure if that's a good idea because this secret should generally not be sent from the client code but from a secure proxy. Having it on the `ShapeStream` client may lead people to believe it's fine to use it. Shall i remove it?

KyleAMathews:
> @KyleAMathews I moved the secret to a query parameter named api_secret. I used an underscore to be consistent with naming of source_secret in cloud. Let me know if you would prefer api-secret or another name.

`api_secret` is great 👍

KyleAMathews:
> In this PR i also added a secret option to the ShapeStream.
Not sure if that's a good idea because this secret should generally not be sent from the client code but from a secure proxy. Having it on the ShapeStream client may lead people to believe it's fine to use it. Shall i remove it?

Yes :-D we've stripped out all special options in ShapeStream after we added the `headers` & `params` options — people can just add `api_secret` to the params object if they're querying the server directly.

KyleAMathews:
Can you post a screenshot of the logs when it boots up in insecure and secure mode? Since it's pretty critical people see this, we might want to put a box around the text.

kevin-dp:
> > In this PR i also added a secret option to the ShapeStream.
> > Not sure if that's a good idea because this secret should generally not be sent from the client code but from a secure proxy. Having it on the ShapeStream client may lead people to believe it's fine to use it. Shall i remove it?
> 
> Yes :-D we've stripped out all special options in ShapeStream after we added the `headers` & `params` options — people can just add `api_secret` to the params object if they're querying the server directly.

Ok, removing it!

kevin-dp:
> Can you post a screenshot of the logs when it boots up in insecure and secure mode? Since it's pretty critical people see this, we might want to put a box around the text.

Secure mode:
Screenshot 2025-03-06 at 15 30 51

Insecure mode:
Screenshot 2025-03-06 at 15 31 23

KyleAMathews:
Hmm yeah that's probably fine actually 👍

packages/sync-service/lib/electric/config.ex
```diff
@@ -374,4 +374,23 @@ defmodule Electric.Config do
...
     end
   end
+
+  def validate_security_config!(secret, insecure) do
+    cond do
+      insecure && secret != nil ->
+        raise "You cannot set both ELECTRIC_SECRET and ELECTRIC_INSECURE=true"
+
+      !insecure && secret == nil ->
+        raise "You must set ELECTRIC_SECRET unless ELECTRIC_INSECURE=true. Setting ELECTRIC_INSECURE=true risks exposing your database, only use insecure mode in development or you've otherwise secured the Electric API"
```
KyleAMathews, line 384: Another small copy tweak:

```suggestion
        raise "You must set ELECTRIC_SECRET or set ELECTRIC_INSECURE=true. Setting ELECTRIC_INSECURE=true risks exposing your database, only use insecure mode in development or you've otherwise secured the Electric API"
```

balegas:
Let's make this optional now and only enforce it for v1.

It would be nice to also print a warning message like the above saying that the default behavior is going to change for v1.

PR #2405, opened 2025-03-05
docs: Document secure and insecure modes

kevin-dp:
Docs update for https://github.com/electric-sql/electric/pull/2404

I think we should also extend other parts of the documentation to explain that this secret must be passed when making shape requests. Otherwise, people will follow the docs but it won't work. So either the docs show usage in INSECURE mode which doesn't seem ideal, or we add the secret where needed. But i'm a bit afraid of polluting code snippets/docs with a secret. Open for suggestions on the best way to communicate this in the docs.

### DIFF
--- a/website/docs/api/config.md
+++ b/website/docs/api/config.md
@@ -92,6 +92,7 @@ Suffix for the logical replication publication and slot name.
 
 <EnvVarConfig
     name="ELECTRIC_SECRET"
+    required={true}
     example="1U6ItbhoQb4kGUU5wXBLbxvNf">
 
 Secret for shape requests to the [HTTP API](/docs/api/http). This is required unless `ELECTRIC_INSECURE` is set to `true`.
--- a/website/docs/api/config.md
+++ b/website/docs/api/config.md
@@ -88,6 +88,31 @@ Suffix for the logical replication publication and slot name.
 
 ## Electric
 
+### ELECTRIC_SECRET
+
+<EnvVarConfig
+    name="ELECTRIC_SECRET"
+    example="1U6ItbhoQb4kGUU5wXBLbxvNf">
+
+Secret for shape requests to the [HTTP API](/docs/api/http).
+
+</EnvVarConfig>
+
+### ELECTRIC_INSECURE
+
+<EnvVarConfig
+    name="ELECTRIC_INSECURE"
+    defaultValue="false"
+    example="true">
+
+When set to `true`, runs Electric in insecure mode and does not require an `ELECTRIC_SECRET`.
+Use with caution.
+API requests are unprotected and may risk exposing your database.
+Good for development environments.
+If used in production, make sure to lock down access to Electric.
+
+</EnvVarConfig>
+
 ### ELECTRIC_INSTANCE_ID
 
 <EnvVarConfig
@@ -118,7 +143,7 @@ Name of the electric service. Used as a resource identifier and namespace.
     example="true">
 
 Expose some unsafe operations that faciliate integration testing.
-Do not enable this production.
+Do not enable this in production.
 
 </EnvVarConfig>
 
--- a/website/docs/api/config.md
+++ b/website/docs/api/config.md
@@ -96,7 +96,7 @@ Suffix for the logical replication publication and slot name.
     example="1U6ItbhoQb4kGUU5wXBLbxvNf">
 
 Secret for shape requests to the [HTTP API](/docs/api/http). This is required unless `ELECTRIC_INSECURE` is set to `true`.
-By default, the Electric API is public and all shape requests must include the secret.
+By default, the Electric API is public and authorises all shape requests against this secret.
 More details are available in the [security guide](/docs/guides/security).
 
 </EnvVarConfig>
--- a/website/docs/api/config.md
+++ b/website/docs/api/config.md
@@ -92,6 +92,7 @@ Suffix for the logical replication publication and slot name.
 
 <EnvVarConfig
     name="ELECTRIC_SECRET"
+    optional="true"
     example="1U6ItbhoQb4kGUU5wXBLbxvNf">
 
 Secret for shape requests to the [HTTP API](/docs/api/http).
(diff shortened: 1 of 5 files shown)


robacourt (approved):
Nice

KyleAMathews:
The quickstart needs updated

kevin-dp:
I was updating the quickstart to use secure mode but when introducing the TS client i had to set the secret.
But actually we don't want our users to use it this way as setting the secret on the client is insecure.
But we also don't want to complicate the quickstart with a proxy.
Shall we add a warning box saying this is unsafe and linking to the proxy page?
Or should the quickstart run in insecure mode and mention that for prod it should run in secure mode with a proxy?

website/docs/quickstart.md
```diff
@@ -31,7 +31,7 @@ Then download and run this [docker-compose.yaml](https://github.com/electric-sql
 
 ```sh
 curl -O https://electric-sql.com/docker-compose.yaml
-docker compose up
+ELECTRIC_SECRET=mySecret docker compose up
```
KyleAMathews, line 34: I'd suggest actually starting the dev server in insecure mode — when developing locally, you want the lowest friction possible & a secret just complicates everything IMO as you have to pass it around everywhere. Which is fine and good for prod but not fun or helpful when you're just playing around learning how Electric works.

website/docs/api/config.md
```diff
@@ -88,6 +88,32 @@ Suffix for the logical replication publication and slot name.
...
 ## Electric
 
+### ELECTRIC_SECRET
+
+<EnvVarConfig
+    name="ELECTRIC_SECRET"
+    optional="true"
+    example="1U6ItbhoQb4kGUU5wXBLbxvNf">
+
+Secret for shape requests to the [HTTP API](/docs/api/http).
```
thruflo, line 98: We definitely need a bit more context here. I think we should add a section to the [Security guide](https://electric-sql.com/docs/guides/security) explaining the secret -- what it is, what it's for, where and how you should set it. Then this can link to that section.

thruflo: Also, this var should be described as required unless you set insecure mode using the other var.

website/docs/api/config.md
```diff
@@ -88,6 +88,32 @@ Suffix for the logical replication publication and slot name.
...
+
+</EnvVarConfig>
+
+### ELECTRIC_INSECURE
+
+<EnvVarConfig
```
thruflo, line 104: Let's also add a link in this env var description to the section of the security guide.

website/docs/quickstart.md
```diff
@@ -31,11 +31,14 @@ Then download and run this [docker-compose.yaml](https://github.com/electric-sql
 
 ```sh
 curl -O https://electric-sql.com/docker-compose.yaml
-docker compose up
+ELECTRIC_INSECURE=true docker compose up
```
thruflo, line 34: Let's set the env var in the docker compose. Add a comment in the compose that indicates it's not good for prod but it's fine to have in the QuickStart compose. No one is deploying that to prod without looking at it.

This then gets rid of the env var noise in the code samples.

website/docs/quickstart.md
```diff
@@ -31,11 +31,14 @@ Then download and run this [docker-compose.yaml](https://github.com/electric-sql
...
 ```
 
 You can now start using Electric!
 
+> [!Warning] Do not use insecure mode in production
```
thruflo, line 39: I think this is too noisy here. Let's drop it.

website/public/docker-compose.yaml
```diff
@@ -23,6 +23,7 @@ services:
     image: electricsql/electric
     environment:
       DATABASE_URL: postgresql://postgres:password@postgres:5432/electric?sslmode=disable
+      ELECTRIC_SECRET: ${ELECTRIC_SECRET}
```
thruflo, line 26: See comment about setting insecure true here.

thruflo (changes requested):
Generally, I think we need to update the documentation in the security guide. Then link in to that.

Plus we should look at our docker compose files for examples, and the deployment guides and see how we update them. There's loads of places we have docker compose files and a number of the deployment guides have blueprints etc. And there's the AWS terraform and some external repos for DO, Render etc.

I think the task is to read through all of those and update them. Otherwise this update breaks everything.

Ideally we can be smart about avoiding `ELECTRIC_INSECURE` noise everywhere. Dev example can be defaulted to insecure. Production / deployment examples can show `ELECTRIC_SECRET=...`.

thruflo:
I think examples can be defaulted to insecure and we can just set the env var in the docker compose.

kevin-dp:
> I think examples can be defaulted to insecure and we can just set the env var in the (shared!) docker compose.

Perfect. I've set the `ELECTRIC_INSECURE: true` env var in the shared compose file in the implementation PR https://github.com/electric-sql/electric/pull/2404. It's fine to do it there because the env var will essentially be ignored by the current Electric image until we publish the newest version which includes secure and insecure modes.

website/docs/guides/security.md
```diff
@@ -62,6 +62,12 @@ Electric is designed to run behind an [authorizing proxy](/docs/guides/auth#requ
 
 This is the primary method for securing data access to clients and apps and is documented in detail, with examples, in the [Auth guide](/docs/guides/auth).
 
+### API token
+
+Access to Electric can be secured with an [API token](/docs/api/config#electric-secret). This is a secret string that can be set when starting Electric and will be used to authenticate requests to the Electric HTTP API. When an API token is set, Electric will require all requests to include the API token. 
+
+The token should *not* be sent from the client as it will be exposed in the HTTP requests. Instead, it should be included by the authorizing proxy when proxying requests to Electric.
```
balegas, line 69: I think `authorizing proxy` should link to https://electric-sql.com/docs/guides/auth#requests-can-be-proxied or to https://electric-sql.com/docs/guides/security#authorization just below

balegas (approved):
Thanks for handling docs. Looks good. Left a comment

KyleAMathews (approved):
:shipit:

thruflo (approved):
Perfect

Issue #2361, opened 2025-02-20
Update openapi docs for the server returning a 409


KyleAMathews:
It's outdated — https://github.com/electric-sql/electric/blob/8a4b0d59aeaa32cace9bc41cd671b75f477a82c5/website/electric-api.yaml#L374

You need to return a messages array in the body with `must-refetch` and the handle header isn't needed. Perhaps other stuff is wrong.

The client also should throw an error if `must-refetch` isn't included in the server response.

msfstef:
I've opened a PR to update the OpenAPI spec - we need to include the shape handle header as it also acts as a consistent cache buster for resyncing and I think is an important part of the spec.

PR #2406, opened 2025-03-05
fix: Correct OpenAPI spec for 409

msfstef:
Closes https://github.com/electric-sql/electric/issues/2361

Also makes the `location` header in 409 responses carry over the full shape query as the original shape.


alco (approved):

KyleAMathews (approved):
:shipit

Issue #2336, opened 2025-02-13
useShape can miss shape updates between the shape is created and subscribed to leading to loading state forever


hedblad:
When the Electric SQL server connection is established quicker than the React subscription is set up, the shape returned by useShape will be stuck in loading until the shape is modified. 

It is tricky to provide a proper reliable reproduction case. We run Postgres and Electric in Docker Compose and can only reproduce this in Chrome under some quite specific circumstances, since this is a race condition and React is involved. So I will reason about the code instead.

Here are lines 163-180 of packages/react-hooks/src/react-hooks.tsx:
```typescript
  const useShapeData = React.useMemo(() => {
    let latestShapeData = parseShapeData(shape)
    const getSnapshot = () => latestShapeData
    const subscribe = (onStoreChange: () => void) =>
      shapeSubscribe(shape, () => {
        latestShapeData = parseShapeData(shape)
        onStoreChange()
      })

    return () => {
      return useSyncExternalStoreWithSelector(
        subscribe,
        getSnapshot,
        getSnapshot,
        selector
      )
    }
  }, [shape, selector])
```

If `shape` loads between the initial `parseShapeData` and `subscribe` is called by `React`, this change will not be picked up by React and `shape` will be stuck in the loading state forever (or at least until the data changes).

Or to put it in different terms: things might change between setting up the shape and subscribing to it, which means we won't pick up the shape data or loading state. Since we don't have control over when react actually calls the function subscribing to the stream, it would make sense to always update the state when the subscription is created to ensure consistency.

Changing the implementation to the following fixes the issue:
```typescript
const useShapeData = useMemo(() => {
    let latestShapeData = parseShapeData(shape)
    const getSnapshot = () => latestShapeData
    const subscribe = (onStoreChange: () => void) => {
      const updateShapeData = () => {
        latestShapeData = parseShapeData(shape)
        onStoreChange()
      }
      // Trigger an update immediately since we might have missed changes 
      // between setting up the ShapeStream and subscribing to it
      updateShapeData()
      return shapeSubscribe(shape, updateShapeData)
    }
    return () => {
      return useSyncExternalStoreWithSelector(
        subscribe,
        getSnapshot,
        getSnapshot,
        selector
      )
    }
  }, [shape, selector]);
```

A similar issue in TanStack/query is described in this issue: https://github.com/TanStack/query/issues/5443 and fixed with this PR: https://github.com/TanStack/query/pull/5474

[...]

PR #2408, opened 2025-03-05
fix: Avoid `Shape` race conditions in updating state

msfstef:
Closes https://github.com/electric-sql/electric/issues/2336

- Changes `Shape` implementation to have a simpler state-machine-like notification mode
- Removes arbitrary timeouts that reduced flakes, used `waitFor` and `waitUntil` Vitest utilities where necessary
- Fixed up some react hooks tests
- Updated the react hook implementation to potentially update the shape data result upon subscribing, to cover cases where there might have been an async gap between the initial snapshot and subscribing

I couldn't write a reproduction case for the issue mentioned in the issue, but given the issue has been observed and solved in similar ways by other libraries as well I think it is still worth making this change.

### DIFF
--- a/packages/react-hooks/src/react-hooks.tsx
+++ b/packages/react-hooks/src/react-hooks.tsx
@@ -139,6 +139,21 @@ function parseShapeData<T extends Row<unknown>>(
   }
 }
 
+function shapeResultChanged<T extends Row<unknown>>(
+  oldRes: UseShapeResult<T> | undefined,
+  newRes: UseShapeResult<T>
+): boolean {
+  return (
+    !oldRes ||
+    oldRes.isLoading !== newRes.isLoading ||
+    oldRes.lastSyncedAt !== newRes.lastSyncedAt ||
+    oldRes.isError !== newRes.isError ||
+    oldRes.error !== newRes.error ||
+    oldRes.shape.lastOffset !== newRes.shape.lastOffset ||
+    oldRes.shape.handle !== newRes.shape.handle
+  )
+}
+
 function identity<T>(arg: T): T {
   return arg
 }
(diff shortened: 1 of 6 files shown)


packages/typescript-client/src/shape.ts
```diff
@@ -172,28 +168,25 @@ export class Shape<T extends Row<unknown> = Row> {
...
-    // FIXME this would be cleaner with a simple state machine.
-    if (newlyUpToDate || (isUpToDate && dataMayHaveChanged)) {
-      this.#hasNotifiedSubscribersUpToDate = true
-      this.#notify()
-    }
+    if (shouldNotify) this.#notify()
+  }
+
+  #updateShapeStatus(status: ShapeStatus): boolean {
+    console.log(`update`, this.#status, status)
```
KyleAMathews, line 186: left over log

KyleAMathews (approved):
nice cleanups!

packages/react-hooks/src/react-hooks.tsx
```diff
@@ -139,6 +139,21 @@ function parseShapeData<T extends Row<unknown>>(
   }
 }
 
+function shapeResultChanged<T extends Row<unknown>>(
+  oldRes: UseShapeResult<T> | undefined,
+  newRes: UseShapeResult<T>
+): boolean {
+  return (
```
kevin-dp, line 146: This looks brittle as it will need to be updated with every change to the interface of `UseShapeResult`.
I would argue for using a generic deep object comparison utility function.

kevin-dp (approved):
Great work!

packages/react-hooks/src/react-hooks.tsx
KyleAMathews: Yeah the check itself is probably optional also.

PR #2409, opened 2025-03-05
Generate and store installation_id to maintain metric continuity between restarts

alco:
An integration test to verify the persistence and a matching change in cloud are coming up as follow-up PRs.


icehaunter (approved):

balegas:
Thanks, this will be very useful

Issue #2334, opened 2025-02-12
Possible memory leak when using `chainAborter`


samwillis:
We have a `chainAborter` helper to make chaining request aborters. However there may be a memory leak in how it works as it doesn't clean up the callbacks on the parent aborter after a request.

A user may pass a signal to ShapeStream, and this could then accumulate the abort callback for each time a request is made using `chainAborter`.

https://github.com/electric-sql/electric/blob/29d73f1efbf810b2ab1631d8654dd23d676a25f8/packages/typescript-client/src/fetch.ts#L324-L335

https://github.com/electric-sql/electric/blob/29d73f1efbf810b2ab1631d8654dd23d676a25f8/packages/typescript-client/src/fetch.ts#L269

PR #2410, opened 2025-03-05
fix: Clean up chained abort signal listener

msfstef:
Closes https://github.com/electric-sql/electric/issues/2334

Listener gets cleaned up if:
- Parent signal is aborted (because of the `once: true` setting)
- Chained abort controller calls `abort` (because of the `signal: aborter.signal` setting)
- The `cleanup` callback is used, which we do at the end of the request that uses this chaining

This should take care of any potential leaks due to listeners.

### DIFF
--- a/packages/typescript-client/src/fetch.ts
+++ b/packages/typescript-client/src/fetch.ts
@@ -264,10 +264,8 @@ class PrefetchQueue {
     const aborter = new AbortController()
 
     try {
-      const request = this.#fetchClient(url, {
-        ...(args[1] ?? {}),
-        signal: chainAborter(aborter, args[1]?.signal),
-      })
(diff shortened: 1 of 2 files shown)


packages/typescript-client/src/fetch.ts
```diff
@@ -286,6 +284,7 @@ class PrefetchQueue {
           return this.#prefetch(nextUrl, args[1])
         })
         .catch(() => {})
+        .finally(cleanup)
```
KyleAMathews, line 287: nice, so the idea is always try to cleanup and if there's no signal it just returns. Yeah, that's simpler.

KyleAMathews (approved):
:shipit:

Issue #2289, opened 2025-02-03
List of subscribers to shape grows linearly


magnetised:
To reproduce:

- Subscribed a single elixir client to a shape and started inserting rows in the db
- On write to db, the message `Notifying ~n clients about new changes` appears in the log, with `n` increasing constantly, so the first write n would be 1, the second 2, etc etc.
- When debugging, the actual pids receiving the update notification are all the same. so the list of listeners to a shape was a list e.g. `[ #PID<0.780.0>,  #PID<0.780.0>,  #PID<0.780.0>]`

My theory is that the pids aren't being killed because the client connection pool keeps reusing the same connection (plus maybe http2?)

This is obviously wrong/stupid. 

With time that list will grow large and may start affecting the latency plus memory usage. 

It also could have weird affects if the same connection pool is used to subscribe to two distinct shapes - every active connection would get send the notification (though not all would act on it) and this would start to accumulate in the connection's message buffer. 

We should make sure that only one instance of a pid is registered as a listener for each shape or find some way to register a request rather than a connection.

balegas:
Leaving a note here to sync with @robacourt about this, since he has been investigating memory leaks

PR #2412, opened 2025-03-05
fix: Avoid long-lived listeners for changes due to connection process reuse

msfstef:
Closes https://github.com/electric-sql/electric/issues/2289

It seems that connections are being handled by a process pool of sorts, which means that they are never removed from the registry.

>Within a Bandit server, an HTTP/1 connection is modeled as a single process. This process is tied to the lifecycle of the underlying TCP connection; **in the case of an HTTP client which makes use of HTTP's keep-alive feature** to make multiple requests on the same connection, **all of these requests will be serviced by this same process**.

This is from [Bandit's docs](https://github.com/mtrudel/bandit/blob/main/lib/bandit/http1/README.md) and I've verified it by reproducing @magnetised 's situation described in the issue.

I initially used an approach where an ephemeral process/`Task` is the one registering for notifications and forwards them, so that it gets automatically cleaned up by the registry, but I ended up using `Registry.unregister` - it just gets a bit more complicated to ensure that we unregister in all code paths for the request, since it might exit early, and the registration occurs on the `validate` step of the API, and the cleanup on the separate `serve_shape_log` step.

So either we spawn an ephemeral process and let it get cleaned up automatically, or we have to ensure the cleanup is called despite it being split across API calls. Perhaps we would need a `cleanup` API method?

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -361,18 +361,7 @@ defmodule Electric.Shapes.Api do
     if LogOffset.compare(offset, last_offset) != :lt or
          last_offset == LogOffset.last_before_real_offsets() do
       ref = make_ref()
-      parent = self()
-
-      # Register for notifications using a separate task as the calling process
-      # might be reused across mutiple connections
-      Task.start_link(fn ->
-        Registry.register(registry, handle, ref)
-
-        receive do
-          msg when elem(msg, 0) == ref -> send(parent, msg)
-        end
-      end)
-
+      Registry.register(registry, handle, ref)
       Logger.debug("Client #{inspect(self())} is registered for changes to #{handle}")
 
       %{request | new_changes_pid: self(), new_changes_ref: ref}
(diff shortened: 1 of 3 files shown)


packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -361,7 +361,17 @@ defmodule Electric.Shapes.Api do
...
+
+      # Register for notifications using a separate task as the calling process
+      # might be reused across mutiple connections
+      Task.start_link(fn ->
+        Registry.register(registry, handle, ref)
+
+        receive do
+          msg when elem(msg, 0) == ref -> send(parent, msg)
+        end
+      end)
```
robacourt, line 374: I am highly sceptical that this is the right path to take. I find it hard to believe that the pids are reused in Plug/bandit, that would cause a whole host of issues. And even if they are, unregister has to be a cleaner route than a ephemeral process.

msfstef: >Within a Bandit server, an HTTP/1 connection is modeled as a single process. This process is tied to the lifecycle of the underlying TCP connection; **in the case of an HTTP client which makes use of HTTP's keep-alive feature** to make multiple requests on the same connection, **all of these requests will be serviced by this same process**.

This is from [Bandit's docs](https://github.com/mtrudel/bandit/blob/main/lib/bandit/http1/README.md) and I've verified it by reproducing @magnetised 's situation described in the issue.

My issue with unregistering is that my understanding is the only way to do this is like:

```elixir
Registry.update_value(registry, handle, fn refs ->
  List.delete(refs, ref)
end)
```

Which I was unsure if it is more or less efficient than spawning a process and letting the registry remove entries when the process dies, as I don't know what kinds of structures it maintains internally to do that.

robacourt (approved):
Looks good! Thanks for the chat

PR #2421, opened 2025-03-06
feat: allow parameters in shape definitions

### LINKED ISSUE #2295: Reduce the risk of parameter injection
Currently it is the users responsibility to sanitise any parameters that they use to construct a `where` clause for a shape. Unfortunately this opens up a possible route to something like a SQL injection that could cause information leakage.

Say a user is constructing a shape url such as:

```
shapeUrl = `${base}/v1/shape?table=source&where="id='${sourceId}'"`
```

If they fail to validate the `sourceId`, which could be user provided, someone could inject something like `x' OR id LIKE '%` resulting in this URL:

```
/v1/shape?table=source&where="id='x' OR id LIKE '%'"
```

This is very similar to a sql injection that can result in information leakage.

A solution to this, which I think we should consider something of a priority, is to have parameterised where clauses. Possible api:

```
shapeUrl = `${base}/v1/shape?table=source&where="id=$1"&param[1]=${sourceId}`
```

When used with a url api that correctly escapes query params, such as the js URL object, you can do this:

```
const url = new URL('https://example.com/v1/shape')
url.searchParams.set(`table`, `source`)
url.searchParams.set(`where`, `id = $1`)
url.searchParams.set(`params[1]`, `123`)
```

Anything set as a param will be correctly url escaped, and we can then ensure that the parameters are safely handled on the Electric side.

icehaunter:
Adds support for parameters in shape definitions on the server. API now allows where clauses to have `$1` placeholders, which are replaced by corresponding `params[1]=value` query parameters. 

Query parameters are not used when either executing the initial query or adding the clause to the publication filter, instead the value is replaced in the original query after type-checking the usage. For example, in `$1 > 2` clause the `$1` is treated as unknown, so the overload resolver picks the `int4 > int4` operator, and checks if the value passed as param 1 can be cast to desired type. If it can, then we don't do de-serialization, and instead just do input escaping with an explicit typecast, as Postgres is efficient enough to only cast const values once. So the original query will be replaced with (for `params[1]=3` query string) `'3'::int4 > 2`, where interpolation is correctly escaped to avoid SQL injection. 

One side-effect of this is that interpolation is done via rebuilding the query from protobuf via PostgreSQL C Code, which normalizes the casing and whitespace.

Closes #2295


packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -164,6 +164,34 @@ defmodule Electric.Plug.RouterTest do
...
+      # Unfortunately, returns rows
+      refute Jason.decode!(conn.resp_body) == []
+
+      # Requesting with param makes it work
+      conn =
+        conn("GET", "/v1/shape?table=items&offset=-1", %{
+          where: "value = $1",
+          params: %{1 => value}
+        })
+        |> Router.call(opts)
```
samwillis, line 189: As a non-elixir dev, I assume `params` is an array, and the url encoder knows that it need to encode them as `params[0]`, `params[1]` etc?

Postgres placeholders are 1 indexed, and I think the api we want is that `$1` maps to `params[1]`. Does this test that?

samwillis: Also I note the PR description mentions `param[1]=value`, is it that or `params[1]`?

icehaunter: Actually no, params here is a map, and under the hood its `params[1]=whatever`. It validates a map, so an array notation of `params[]=whatever` is considered invalid. It also validates presence, all arguments used, and sequential parameter usage (i.e. can't have $3 without $2)

icehaunter: Trust the test, not the words :D
It's `params`

samwillis: I'm happy with `params[n]`. Maybe @KyleAMathews has any final thoughts?

Obv not qualified to review the code.

icehaunter: It's tested down the stack, but the HTTP request would look smth like (but with URL encoding obv) `?where=$1 < value OR value < $2 AND content ILIKE $3&param[1]=10&param[2]=20&param[3]=hello%`

KyleAMathews: Ok cool 👍 yeah this seems good.

Where are we going to update the docs? This PR or a follow-up one?

icehaunter: I'll make a follow-up

robacourt (approved):
Nice!

PR #2422, opened 2025-03-06
chore: Fix flaky TS tests

msfstef:
Replaces unnecessary `sleep` calls intended to wait for transactions to get processed and go through the sync service with a `waitForIssues` test method, which starts a stream and waits for a given number of changes to go through.

I've added a manual `Process.sleep(200)` in the `process_x_log_data` call where we process the replication stream to simulate slow replication and render all the arbitrary `sleep` calls invalid, and then went through the failing tests one by one and made changes such that they work regardless of these delays that occur in CI.


KyleAMathews (approved):
Nice! No need for timeouts when we got stream processing chops 💥

icehaunter (approved):

PR #2423, opened 2025-03-06
Fix incorrect LSN comparisons

msfstef:
Fixes transactions getting dropped incorrectly because of invalid LSN comparisons, introduced in https://github.com/electric-sql/electric/pull/2382

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -5,6 +5,8 @@ defmodule Electric.Replication.ShapeLogCollector do
   """
   use GenStage
 
+  require Electric.Postgres.Lsn
+  alias Electric.Postgres.Lsn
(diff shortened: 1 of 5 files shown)


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -151,7 +153,7 @@ defmodule Electric.Replication.ShapeLogCollector do
 
   # If we've already processed a transaction, then drop it without processing
   defp handle_transaction(txn, _from, %{last_seen_lsn: last_seen_lsn} = state)
-       when txn.lsn <= last_seen_lsn do
+       when not Electric.Postgres.Lsn.is_larger(txn.lsn, last_seen_lsn) do
```
robacourt, line 156:
```suggestion
       when not Lsn.is_larger(txn.lsn, last_seen_lsn) do
```

robacourt (approved):
Good spot!

PR #2426, opened 2025-03-06
Fix LSN comparison

robacourt:
Following on from #2423 , this fixes another comparison.

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -200,7 +200,7 @@ defmodule Electric.Replication.ShapeLogCollector do
     OpenTelemetry.add_span_attributes("txn.is_dropped": true)
 
     state =
-      if Lsn.is_larger(state.last_seen_lsn, txn.lsn) do
+      if Lsn.is_larger(txn.lsn, state.last_seen_lsn) do
         :ok =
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -200,7 +200,7 @@ defmodule Electric.Replication.ShapeLogCollector do
     OpenTelemetry.add_span_attributes("txn.is_dropped": true)
 
     state =
-      if txn.lsn > state.last_seen_lsn do
+      if Lsn.is_larger(state.last_seen_lsn, txn.lsn) do
```
msfstef, line 203: this is wrong no? it should be the other way around I think

msfstef (changes requested):

packages/sync-service/lib/electric/replication/shape_log_collector.ex
robacourt:
```suggestion
      if Lsn.is_larger(txn.lsn, state.last_seen_lsn) do
```

robacourt: Good spot! Changed @msfstef

msfstef (approved):
Looks good, missed that one! In hindsight, we probably are better off converting the Lsn to an integer early on and do these comparisons, but we can do that after fixing this

PR #2429, opened 2025-03-07
fix: improve file merging performance

### LINKED ISSUE #2370: Using `Utils.merge_sorted_streams()` for sorting file streams results in redundant open/close calls
THe implementation of [`Utils.merge_sorted_streams()`](https://github.com/electric-sql/electric/blob/c4440726a563393a6c33da5780e6323df80ca6cc/packages/sync-service/lib/electric/utils.ex#L421-L451) does not benefit from the fact that its input is a list of streams because it walks each stream explicitly one element at a time, calling `Stream.drop(stream, 1)` on each one. When each stream is a file, it gets even worse because each call to `Enum.take(stream, 1)` and `Stream.drop(stream, 1)` results in the file opening and closing per every stream element.

A separate function specifically for merging files with sorted records will have less redundancy by opening each file just once or, in the case the number of files that need to be simultanously open exceeds the limit of file descriptors, reading multiple records from every file into memory for each open/close pair of syscalls.

icehaunter:
Closes #2370

### DIFF
--- a/.changeset/silent-parrots-smile.md
+++ b/.changeset/silent-parrots-smile.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: fixes file merging during compaction, which was very suboptimal due to a bug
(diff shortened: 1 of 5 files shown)


alco (approved):
:+1: 

Great! I'm rebasing my changes to file operations on top of this.

PR #2434, opened 2025-03-10
Assign installation_id to the OTEL resource

alco:
Follow-up to #2409. This change makes the installation_id available in OTEL traces/logs/metrics.


packages/sync-service/config/runtime.exs
```diff
@@ -50,7 +50,11 @@ instance_id = env!("ELECTRIC_INSTANCE_ID", :string, Electric.Utils.uuid4())
 version = Electric.version()
 
 config :opentelemetry,
-  resource_detectors: [:otel_resource_env_var, :otel_resource_app_env],
+  resource_detectors: [
+    :otel_resource_env_var,
+    :otel_resource_app_env,
+    Electric.Telemetry.OpenTelemetry.ResourceDetector
+  ],
   resource: %{service: %{name: service_name, version: version}, instance: %{id: instance_id}}
```
icehaunter, line 58: So `:opentelemetry` lib doesn't support sending either metrics or logs to OTel endpoints, which is why I wrote our own library for that purpose. I didn't think to respect these resource detectors TBH, but we do need a way to add this field there. For metrics, it can be done at https://github.com/electric-sql/electric/blob/4897fe8acda275b9fbfb7f673b99a5b11db09738/packages/sync-service/lib/electric/telemetry/stack_telemetry.ex#L71 without additional issues

alco: Thanks for pointing this out! I've added installation_id to the metric resource, it also requires [this change](https://github.com/electric-sql/elixir-otel-metric-exporter/pull/2) to be published to work correctly.

alco: @icehaunter PR updated.

Does ELectric already use the OTEL log exporter? It seems to have its own configuration for resource attributes.

icehaunter: @alco OTel log exporter in our library (`OtelMetricExporter.LogHandler`) is used only in the cloud installation currently because it wasn't discussed for self-hosting, so we're just keeping metrics for now

icehaunter (approved):

PR #2435, opened 2025-03-11
fix: Pause and resume shapestream on visibility changes

kevin-dp:
https://github.com/electric-sql/electric/issues/2364 reported an issue with Safari on iOS which caused the application to stop syncing changes. The reason for this behavior is that Safari aggressively throttles/suspends/cancels timers and loops to save battery. Therefore, the loop in `ShapeStream.#start` was sometimes stopped and never resumed.

I tried fixing this with timers (`setTimeout`/`setInterval`) but that did not solve the issue. Instead, i introduced `pause` and `resume` methods on the `ShapeStream`. When the TS client is running in the browser, the `ShapeStream` now uses the [visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to pause the stream when the app goes in the background and to resume it when it comes back into the foreground.

I tested this fix using the React example with Safari on iOS.
To test it for yourself, this is the setup that was reproducing the issue and works with this branch:
- Run ngrok to expose both `http://localhost:3000` (local PG) and `http://localhost:5173` (local dev server)
- Run `pnpm backend:up`
- Run `VITE_ELECTRIC_URL="https://<your subdomain for PG>.ngrok.app" pnpm vite`
- In Safari on iOS visit `https://<your subdomain for dev server>.ngrok.app`
- Lock your iPhone
- Perform some update
- Unlock your iPhone, the app should now be in the old state and not react to updates

## Gotchas

In order to grab the global `document`variable (when in browser environments), i had to add the `DOM` and `DOM.Iterable` target libraries to the TS config. However, because of that the type of the `fetchClient` option for the `ShapeStream` changed since `typeof fetch` was `string | URL` and is now `string | URL | Request`. This is a breaking change since user-provided fetch functions will need to be widened to also support a `Request` input argument.

Either we are fine with this or we remove those target libraries and instead manually typecast the global `document` variable.

## Note to reviewers

Changes to `ShapeStream.#start` are small but i extracted the while loop into an if test that returns fast:
```ts
while (
  (!this.options.signal?.aborted && !this.#isUpToDate) ||
  this.options.subscribe
) {
  <perform request>
}
```
Modified this to:
```ts
if (
  !this.options.subscribe &&
  (this.options.signal?.aborted || this.#isUpToDate)
) {
  return
}

<perform request>
```

Because of this change, the indentation of the code in `<perform request>` changed and Git fails to detect this so it marks the original code as deleted and the indented code as added.


packages/typescript-client/src/client.ts
```diff
@@ -358,6 +361,8 @@ export class ShapeStream<T extends Row<unknown> = Row>
     this.#fetchClient = createFetchWithResponseHeadersCheck(
       createFetchWithChunkBuffer(fetchWithBackoffClient)
     )
+
+    this.#subscribeToVisibilityChanges()
```
samwillis, line 365: It may be worth providing an option to opt out of this behaviour, but not essential.

packages/typescript-client/src/client.ts
```diff
@@ -680,6 +709,30 @@ export class ShapeStream<T extends Row<unknown> = Row>
     })
   }
 
+  #subscribeToVisibilityChanges() {
+    if (
+      typeof document === `object` &&
+      typeof document.hidden === `boolean` &&
+      typeof document.addEventListener === `function`
```
samwillis, line 716: Mostly note to self... this only runs on the main thread, so running in a web worker or Node is unaffected.

packages/typescript-client/src/client.ts
```diff
@@ -583,6 +416,198 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+      this.#requestAbortController?.abort(PAUSE_POLLING)
+    }
+  }
+
+  resume() {
+    if (this.#started && this.#paused) {
+      this.#paused = false
+      this.#start()
+    }
+  }
```
samwillis, line 609: Is it worth placing the stream into non-live mode after a resume?

May play better with caches after a long pause?

samwillis (approved):
I've had a quick skim. I think this is a great change!

I like the introduction of pause + resume as public apis.

I have no issue with the breaking change of the fetch interface. It's more correct.

msfstef:
@kevin-dp there's also the (slightly ugly) possibility of detecting when a pause happened because of visibility, and only calling resume if the pause was because of that.

I like the idea of a public pause and resume API, but that's why I'd like this visibility toggling to be mostly external to `ShapeStream` since it should make use of just the public API (and potentially keep track of some extra state as mentioned above inside a closure or something)?

packages/typescript-client/src/client.ts
msfstef: agree - we should assume we're off of live mode if we pause

kevin-dp:
> @kevin-dp there's also the (slightly ugly) possibility of detecting when a pause happened because of visibility, and only calling resume if the pause was because of that.
> 
> I like the idea of a public pause and resume API, but that's why I'd like this visibility toggling to be mostly external to `ShapeStream` since it should make use of just the public API (and potentially keep track of some extra state as mentioned above inside a closure or something)?

Yes, i see a few solutions to keep it public:
1. Introduce a `BrowserShapeStream` subclass that disables `pause`/`resume` and does it automatically
2. Introduce a shape stream option whether it should automate `pause`/`resume` on visibility changes. Would default to being enabled and can optionally be disabled if someone wants to do it manually in a browser environment.

msfstef:
I think the `BrowserShapeStream` might be a bit overkill practically speaking (although I like it conceptually haha) - having that as an option makes more sense

packages/typescript-client/src/client.ts
kevin-dp: This would only be beneficial for the case where there's new data, right?
I don't think we can assume that there will be new data just because we went into a "long" pause.
Some shapes may experience frequent changes while others don't. I would stick to the original behavior until we have proper evidence of this. Wdyt?

packages/typescript-client/src/client.ts
```diff
@@ -583,6 +416,198 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+    }
+
+    this.#tickPromiseResolver?.()
+    return this.#requestShape()
+  }
+
+  pause() {
+    if (this.#started && !this.#paused) {
+      this.#paused = true
+      this.#requestAbortController?.abort(PAUSE_POLLING)
```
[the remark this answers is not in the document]
kevin-dp: cc @samwillis

packages/typescript-client/src/client.ts
msfstef: There's properties like `isUpToDate` and `isConnected` that once the stream is paused are not guaranteed - I think we should be defensive about those and set them to false (which, for the up to date one, would also switch to non-live requests)

Someone who has paused the stream should not check the `isUpToDate` property and see `true` IMO

packages/typescript-client/src/client.ts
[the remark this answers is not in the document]
samwillis: I think that makes sense to me. the public `.pause` then returns a promise that resolves when we actually have paused (all messages from current request have been sent).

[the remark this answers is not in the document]
kevin-dp: It's impossible to receive a new visibility event before the next tick, right?
Just want to be sure about that otherwise it would be possible that it moves back into focus and that we set `this.#paused = true` after it has already moved back into focus.

samwillis:
To channel @KyleAMathews I think we want to have an api that "just works" with no supprising behaviour. I would argue for the default to be to listen to the visibility api so that uses get a good behaviour by default.

Having a flag that logs the pause was triggered internally to the shape stream due to visibility change, and so can be auto resumed, seems ok to me.

packages/typescript-client/src/client.ts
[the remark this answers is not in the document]
msfstef: you're right - that would then require us to chain `resume`s to the pause promises and it becomes promise chaining hell...

packages/typescript-client/src/client.ts
kevin-dp: @msfstef Normally when the shape stream is paused the `#requestShape` function will return.
So then, the `finally` block on L422 will be reached and `this.#connected` will be set to `false`.
This preserves the original behavior, it didn't seem to set the `isUpToDate` to false though.
Not sure if we want to introduce this here, although it seems logical to set `isUpToDate` to false there or to change the `isUpToDate` getter to check `this.#connected && this.#isUpToDate`.

kevin-dp: Looking further at the code, i think `#upToDate` is only meant to be a flag that is set when we finish the initial sync. So it is set to `true` after initial sync and never put back to false (except on reset).

KyleAMathews:
Just as a general rule, let's not add new APIs w/o multiple (ideally 3+) unique use cases that are demanding it. Once they exist, public APIs are really hard to get rid of so unless we default to no, over the years, more and more APIs will will result.

You can see lots of older projects that didn't have discipline around saying no to APIs and they're littered with overlapping, often partly broken APIs.

KyleAMathews:
> I tried fixing this with timers (setTimeout/setInterval) but that did not solve the issue

I'm very surprised by this — I didn't see a description here or on the other PR about what issue you ran into? I can't find any documentation or reports that IOS Safari would permanently disable a `setTimeout` if a tab is hidden.

Why did you shift to this implementation?

kevin-dp:
> > I tried fixing this with timers (setTimeout/setInterval) but that did not solve the issue
> 
> I'm very surprised by this — I didn't see a description here or on the other PR about what issue you ran into? I can't find any documentation or reports that IOS Safari would permanently disable a `setTimeout` if a tab is hidden.
> 
> Why did you shift to this implementation?

The fix by @samwillis wasn't working. I figured there was an additional (infinite) while loop in the `#start` method. I tried replacing it by a `setTimeout`/`setInterval` but none were working. According to Claude, Safari on iOS is very aggressive on throttling/suspending/cancelling timers on background applications and might not execute a timer/interval. That seems to be confirmed by my tests. Hence, the shift to pause/resume which conceptually seems the right thing to do.

We could make the pause/resume API private for now if we worry about introducing non-essential public APIs.

KyleAMathews:
Using the visibility API also means that even non-ios browsers will immediately pause the stream if you switch tabs? Is that a behavior we want?

Since we're just fixing an IOS behavior — how about we don't ever pause the stream but just check to see if we're actually running? Perhaps then we could just keep using the while loop and this would be a simple check and restart if it was stopped type change?

KyleAMathews:
> How do we detect that it was stopped if the function's execution can be stopped at any point?

Perhaps mark if a fetch is ongoing & when the last fetch was started? Then we know it's stopped if `notFetching && lastFetchStart > 1s`

kevin-dp:
> > How do we detect that it was stopped if the function's execution can be stopped at any point?
> 
> Perhaps mark if a fetch is ongoing & when the last fetch was started? Then we know it's stopped if `notFetching && lastFetchStart > 1s`

I don't think this is a reliable way to detect that execution stopped. The app may be interrupted after setting the flag and the fetch request may never be initiated or it could be cancelled mid flight because the user locks the phone. In which case the state will be wrongly marked as fetching. Also, live requests are long polling so they may run for 20 sec. Which means we cannot do `lastFetchStart > 1s` but would have to do `lastFetchStart > 20s` which is not suitable because that means we would only detect that it stopped after 20sec.

kevin-dp:
> We can find a reliable way to see if a fetch is going. We don't want to auto pause when leaving a tab as that isn't how it works now and it'd be surprising and often undesirable behavior. We just want to check if things are working when the page becomes visible again.

I'm open to any suggestions. I can't think of any reliable way to detect that the fetch loop stopped apart from keeping track of the timestamp of the last fetch but this also isn't reliable because there will be a delay of 20sec and checking for time elapsed >20 sec isn't reliable either since code execution may be paused and then resumed so it could be delayed by pause time + 20 sec in which case we will restart the loop and have 2 loops running.

I wouldn't categorise this as an iOS bug. It happens on iOS because iOS is very aggressive for saving battery but this is perfectly legitimate behavior that any other environments and/or browsers may also adopt.

KyleAMathews:
Talking to Claude about this more. There's no way to detect if the fetch is paused but what we can do is something like this:

```javascript
function fetchWithTimeout(url, options = {}, timeoutMs = 10000) {
  return new Promise((resolve, reject) => {
    // Set up timeout
    const timeoutId = setTimeout(() => {
      reject(new Error('Request timed out - possibly paused by browser'));
    }, timeoutMs);
    
    // Start fetch
    fetch(url, options)
      .then(response => {
        clearTimeout(timeoutId);
        resolve(response);
      })
      .catch(error => {
        clearTimeout(timeoutId);
        reject(error);
      });
  });
}
```

So when the page gets unpacked, the timeout callback is triggered which stops the fetch and then a new one is started.

kevin-dp:
> Talking to Claude about this more. There's no way to detect if the fetch is paused but what we can do is something like this:
> 
> ```js
> function fetchWithTimeout(url, options = {}, timeoutMs = 10000) {
>   return new Promise((resolve, reject) => {
>     // Set up timeout
>     const timeoutId = setTimeout(() => {
>       reject(new Error('Request timed out - possibly paused by browser'));
>     }, timeoutMs);
>     
>     // Start fetch
>     fetch(url, options)
>       .then(response => {
>         clearTimeout(timeoutId);
>         resolve(response);
>       })
>       .catch(error => {
>         clearTimeout(timeoutId);
>         reject(error);
>       });
>   });
> }
> ```
> 
> So when the page gets unpacked, the timeout callback is triggered which stops the fetch and then a new one is started.

I had tried to turn the while loop into a loop of `setTimeout` and also tried with `setInterval`, in both cases it still didn't work on Safari on iOS. It seemed like those timers were cancelled altogether. Would be good if someone can double check this to confirm that it wasn't something else that caused it not to work.

KyleAMathews:
No yeah you're right it seems, `setTimeouts` are also just silently paused.

How about this?

```javascript
function fetchWithCustomTimeout(url, timeoutMs, options = {}) {
  // Store fetch start time
  const startTime = Date.now();
  let isFetchCompleted = false;
  
  // Create visibility change handler
  const visibilityHandler = () => {
    if (document.visibilityState === 'visible' && !isFetchCompleted) {
      const elapsedTime = Date.now() - startTime;
      
      if (elapsedTime > timeoutMs) {
        // Timeout exceeded, abort the fetch
        controller.abort();
        console.log(`Fetch aborted after page return: ${elapsedTime}ms elapsed`);
      }
    }
  };
  
  // Add visibility change listener
  document.addEventListener('visibilitychange', visibilityHandler);
  
  // Create AbortController for the fetch
  const controller = new AbortController();
  const signal = controller.signal;
  
  // Start the fetch
  const fetchPromise = fetch(url, { 
    ...options, 
    signal 
  })
  .then(response => {
    isFetchCompleted = true;
    document.removeEventListener('visibilitychange', visibilityHandler);
    return response;
  })
  .catch(error => {
    isFetchCompleted = true;
    document.removeEventListener('visibilitychange', visibilityHandler);
    throw error;
  });
  
  return fetchPromise;
}

// Usage
fetchWithCustomTimeout('https://example.com/api/data', 30000)
  .then(response => response.json())
  .then(data => console.log(data))
  .catch(error => console.error('Fetch error:', error));
```

kevin-dp:
> No yeah you're right it seems, `setTimeouts` are also just silently paused.
> 
> How about this?
> 
> ```js
> function fetchWithCustomTimeout(url, timeoutMs, options = {}) {
>   // Store fetch start time
>   const startTime = Date.now();
>   let isFetchCompleted = false;
>   
>   // Create visibility change handler
>   const visibilityHandler = () => {
>     if (document.visibilityState === 'visible' && !isFetchCompleted) {
>       const elapsedTime = Date.now() - startTime;
>       
>       if (elapsedTime > timeoutMs) {
>         // Timeout exceeded, abort the fetch
>         controller.abort();
>         console.log(`Fetch aborted after page return: ${elapsedTime}ms elapsed`);
>       }
>     }
>   };
>   
>   // Add visibility change listener
>   document.addEventListener('visibilitychange', visibilityHandler);
>   
>   // Create AbortController for the fetch
>   const controller = new AbortController();
>   const signal = controller.signal;
>   
>   // Start the fetch
>   const fetchPromise = fetch(url, { 
>     ...options, 
>     signal 
>   })
>   .then(response => {
>     isFetchCompleted = true;
>     document.removeEventListener('visibilitychange', visibilityHandler);
>     return response;
>   })
>   .catch(error => {
>     isFetchCompleted = true;
>     document.removeEventListener('visibilitychange', visibilityHandler);
>     throw error;
>   });
>   
>   return fetchPromise;
> }
> 
> // Usage
> fetchWithCustomTimeout('https://example.com/api/data', 30000)
>   .then(response => response.json())
>   .then(data => console.log(data))
>   .catch(error => console.error('Fetch error:', error));
> ```

How would you ensure that we keep fetching in an infinite loop?
Would you modify the `fetchWithCustomTimeout` function to call itself when the fetch is aborted because of a timeout?

I need to think about it more and i would ultimately have to try it but i few concerns i have is that this is going to loop forever, and at some point the app may go to the background at which point iOS may decide to interrupt the loop, if it would interrupt it before the `document.addEventListener` then we won't be notified when the app is visible again. I was tempted to think that moving `document.addEventListener` to the beginning of the function would solve the issue but it does not, it could also interrupt the loop right after the recursive call and before reaching the first statement of the function. Would actually need to test this to be sure that this is true.

samwillis:
> Using the visibility API also means that even non-ios browsers will immediately pause the stream if you switch tabs? Is that a behavior we want?

My thinking here is that we will need to special case iOS. So on iOS we aggressively pause and resume the shape on visibility change, but not on other platforms?

It's notoriously hard to work around these things on iOS, and the visibility api was specifically designed for this use case.

I would normally argue against special casing for a platform as it brings testing complexity and behaviour changes... but this is I think the exception.

balegas:
This PR is open for a long time, can we try reaching a conclusion and finish it?

KyleAMathews:
I merged in origin/main.

I also ran Deep Research on this issue and it agreed that using the visibility API is the most robust way to handle this issue. It also suggested as an extra safety measure, we could add a timeout — thinking about it more a bit removed from our previous discussions, the page visibility API seems fine honestly. If we can find some way of triggering a paused fetch call that visibility doesn't handle then we can add the timeout.

KyleAMathews (approved):
Looks great @kevin-dp !

PR #2436, opened 2025-03-11
Fix license name in package.json(s)

masutaka:
The license management tool, [License Finder](https://github.com/pivotal/LicenseFinder), fails to detect the license, so could you please include the correct license name in package.json?


alco (approved):
Thanks!

PR #2438, opened 2025-03-11
feat: Implement separate, potentially pooled `ELECTRIC_QUERY_DATABASE_URL`

msfstef:
Closes https://github.com/electric-sql/electric/issues/1885

- Introduce separate env var `ELECTRIC_QUERY_DATABASE_URL`
- Add another `connection_opts` field to `replication_opts`, so there are two of them
- Fallback to using the same ones, but still keep them separate and maintain them separately
  - Main disadvantage of this approach is that the fallbacks for SSL and IPv6 might need to happen twice if a separate connection string is not being used - this can be improved by only populating the `replication_opts`'s `connection_opts` right before starting the replication client with the single `connection_opts` available if a separate one was not provided.
- Make all tests (e.g. router tests) use a `pgBouncer` connection string
  - Had to ensure connections are terminated explicitly at the end of the test to avoid hanging connections from the pooler making the tests really slow.

## TODO
- [x] Write integration test with `pgBouncer` as well that simply syncs a shape with a pooled connection URL, although the unit tests already cover that
- [ ] Avoid duplicate SSL and IPv6 fallbacks when separate query connection string is not provided

### DIFF
--- a/.changeset/flat-hornets-ring.md
+++ b/.changeset/flat-hornets-ring.md
@@ -1,0 +1,6 @@
+---
+"@core/sync-service": patch
+"@electric-sql/docs": patch
+---
+
+Implement `ELECTRIC_QUERY_DATABASE_URL` optional env var to perform queries with separate, potentially pooled connection string.
--- a/.github/workflows/elixir_tests.yml
+++ b/.github/workflows/elixir_tests.yml
@@ -41,14 +41,17 @@ jobs:
           - 54321:5432
 
       pgbouncer:
-        image: edoburu/pgbouncer:latest
+        image: bitnami/pgbouncer:latest
+        env:
+          PGBOUNCER_AUTH_TYPE: trust
+          PGBOUNCER_DATABASE: "*"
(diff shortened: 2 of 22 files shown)


msfstef:
ugh - I would need to also start `pgBouncer` in CI for the tests - perhaps we should keep it only as an integration test?

msfstef:
I've decided against trying to avoid a duplicate fallback as this change separates the replication and query connection options - either way I feel this kind of fallback logic should be at the Postgrex layer and not living alongside all the rest of our business logic. We could potentially keep the replication connection opts unpopulated and only populate them after the lock connection succeeds in order to avoid double fallbacks, but it makes the code harder than it already is to follow IMO.


Also, it seems that either adding the pooler or the forced termination of connections at the end of the test has introduced some flakes, I wonder if those are already existing flakes that are being manifested by these changes or if they are truly new flakes because of them, but it's interesting to investigate either way.

icehaunter (approved):
LGTM, nice change!

alco (approved):
Brilliant!

Issue #2440, opened 2025-03-11
Improve tests related with transaction handling


balegas:
After the issues with https://github.com/electric-sql/electric/pull/2382 we decided to add some new tests:
- Add a regression test to prevent the issue from happening again
- Add some unit tests to test comparison for LSN ranges that stress edge conditions
- Have an integration test that tests scenarios where Postgres has ran for a while

Check with @robacourt for _light-fuzzing_ and with @alco for prior work on simulating non-empty Postgres state.

balegas:
We'd also like to avoid structural comparison as it can add silent errors to our code

PR #2442, opened 2025-03-12
chore: Improve transaction handling tests

msfstef:
Closes https://github.com/electric-sql/electric/issues/2440

- Made the `ShapeLogCollector` test slightly more fuzzy, randomly tests a few LSNs starting from 0 and generating successive values in the 64-bit range.
- I've made our tests and integration tests randomly reset the WAL to a random point in the 64-bit range using `pg_resetwal` and a custom `initdb` script - this doesn't apply to CI unit tests as it's a bit tricky to add things to that service, but it should help with adding a bit more fuzz to our elixir tests when ran locally.
- I've introduced a (slightly long) integration test that ensures that replication works with 3 databases, starting at WAL `00000000/00000000`, one at `7FFFFFFF/whatever`, and one at `FFFFFFFF/whatever`, so that it spans the replication across the WAL range. It also has the benefit of just testing the basic replication, since we didn't really have such a straightforward integration test.
- Added test case that compares range of LSN values from start to the 64-bit end - bit of an overkill but just following the spirit of this task

Fun fact this is what Postgres looks like when it reaches the end of the available LSN range:
image
It definitely wasn't pretty - it just kept panicking and restarting.

I attempted to convert the `Lsn` struct to a struct whose structural comparison still makes sense, and interestingly enough I got this warning:

```bash
warning: comparison with structs found:
    left > right
given types:
    %Electric.Postgres.Lsn{offset: integer(), segment: integer()} > %Electric.Postgres.Lsn{
      offset: integer(),
      segment: integer()
    }
where "left" (context ExUnit.Assertions) was given the type:
    # type: %Electric.Postgres.Lsn{offset: integer(), segment: integer()}
    # from: test/electric/postgres/lsn_test.exs:18
    left = %Electric.Postgres.Lsn{offset: 0, segment: 5}
where "right" (context ExUnit.Assertions) was given the type:
    # type: %Electric.Postgres.Lsn{offset: integer(), segment: integer()}
    # from: test/electric/postgres/lsn_test.exs:18
    right = %Electric.Postgres.Lsn{offset: 0, segment: 6}
Comparison operators (>, <, >=, <=, min, and max) perform structural and not semantic comparison. Comparing with a struct won't give meaningful results. Structs that can be compared typically define a compare/2 function within their modules that can be used for semantic comparison.

typing violation found at:
│
18 │     assert %Lsn{segment: 5} > %Lsn{segment: 6}
│                             ~
│
└─ test/electric/postgres/lsn_test.exs:18:29: Electric.Postgres.LsnTest."test LSN implement `List.Chars` protocol"/1
```

So it seems that Elixir does indeed warn when you do structural comparisons, as they don't make sense @balegas @robacourt . I suspect that we just have to wait for the type system to evolve a bit more, as I tried adding Dialyzer types to see if I would get this warning on the relevant code and it compiled just fine.

Alternatively I can refactor the Lsn to be a tuple, which has stable, predictable comparisons, and we can access the `segment` and `offset` parts with explicit accessors, and generate them with a `.new` method. I think this makes sense but I'll defer to more experienced Elixir devs.


packages/sync-service/test/electric/replication/shape_log_collector_test.exs
```diff
@@ -162,34 +160,52 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
...
-
-      assert :ok = ShapeLogCollector.store_transaction(txn2, ctx.server)
-      assert :ok = ShapeLogCollector.store_transaction(txn3, ctx.server)
-      Support.TransactionConsumer.refute_consume(ctx.consumers)
+      change = %Changes.NewRecord{
+        relation: {"public", "test_table"},
+        record: %{"id" => "2", "name" => "foo"}
+      }
+
+      Stream.unfold({0, 1, 0, 1, 0}, fn {runs, xid, prev_xid, lsn_int, prev_lsn_int} ->
```
robacourt, line 168: Stream.unfold seems an unusual way to do this. Perhaps it could be:
```
1..@num_comparisons
|> Enum.reduce({1, 0, 1, 0}, fn _, {xid, prev_xid, lsn_int, prev_lsn_int} ->
```

robacourt (approved):
Great work! I just left one nitpick

packages/sync-service/test/electric/replication/shape_log_collector_test.exs
```diff
@@ -162,34 +160,52 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
...
+        if runs == @num_comparisons do
+          nil
+        else
+          # advance xid and lsn randomly along their potential values to simulate
+          # trnasactions coming in at different points in the DBs lifetime
+          xid = xid + (:rand.uniform(2 ** 32 - xid) - 1)
```
alco: Huh, I missed the introduction of the power operator in Elixir 1.13!

packages/sync-service/dev/reset_wal.sh
alco, line 1: It's worrying that this file duplicates the one in `../../../integrations-tests/scripts` and you've already had to make identical changes to these two files while working on this PR.

Is this script here only to be available for `dev/docker-compose.yml`? Doesn't specifying a relative path such as `../../../integration-tests/scripts/reset-wal.sh` in docker-compose.yml work? Alternatively, we could put the script file at the top level of the repo and create symlinks in `integrations-tests/scripts` and `packages/sync-service/dev`.

alco (approved):
:muscle:

packages/sync-service/dev/reset_wal.sh
msfstef: I was worried of introducing more "shared" top level folders, and of linking things between integration tests and the sync service since they are not on the same "level" in our current directory structure. I could put it under the top level `.support` folder we have, it seems like it could be fit for this purpose

packages/sync-service/test/electric/replication/shape_log_collector_test.exs
msfstef: found it out testing in `iex` haha my JS brain kicked in

packages/sync-service/dev/reset_wal.sh
alco: We have a top-level `scripts` dir. I'd prefer for this script to be placed there and not duplicated in multiple subdirectories.

msfstef: moved it there and created symlinks

PR #2445, opened 2025-03-12
website: flesh out updated messaging flow.

thruflo:
Updated homepage in progress.


samwillis:
I think the header/hero feels a little empty, but I know we're waiting on the updated imagery from the agency.

there is a problem with the layout of the "industry insiders" section. On desktop Chris is larger then the rest, and on mobile it all goes a little mad.

the discord bit under "Fully open source" also get messed up on mobile.
 
On the whole, great 👍

alco:
I love both the new layout and the content. Good job on that!

One minor thing that's giving me a bit of vertigo is the amount of buttons that keep following me wherever I scroll the page.

Like in this case, for example, all 3 buttons can be removed with no loss of functionality:
image

Perhaps some of the buttons could be turned into links. Maybe it's just me, but it does give me anxiety as I scroll down since normally I like to click on navigation buttons and links to open new tabs in background as I scroll down the landing page.

thruflo:
@alco I hear you, although repeating CTAs is kind of a thing.

Let's address when we actually review the website design. This is still a DIY hack.

KyleAMathews (approved):
:shipit:

PR #2446, opened 2025-03-13
Fix LSN encoding in responses and handling in JS

msfstef:
Follow-up to https://github.com/electric-sql/electric/pull/2442

Randomizing the WAL start point in tests has highlighted problems with how we handle LSNs in Javascript.

Javascript can only do max 53-bit number representations, after which it will start doing nonsense without telling you (gotta love it), but LSNs span the entire 64-bit range.

In order to fix that, we encode the `lsn` and `global_last_seen_lsn` as strings in the responses rather than integers, so that Javascript can either compare them as strings or convert them to `BigInt`s, both would yield valid comparisons.

JSON itself does not really care, and other languages are able to parse large integers correctly, but for the sake of portability in the current landscape I think it's better to keep it encoded only once and as a string, rather than e.g. twice as `lsn` and `lsn_str` - feels like overkill.

I've kept the `op_position` as a number, because it would be really odd for a transaction to have `> 2^32` operations in it - maybe if the need arises or if we want consistency between `lsn` and `op_position` I can convert those as well.

For the `MultiShapeStream`, I decided to parse the LSNs into `BigInt`s even though we don't really do any calculations with them and we could just compare them as strings, but I felt like this is more explicit, more easily understood while reading, and less error prone in the future, and the performance cost given that this is a primarily network-heavy utility I think is negligible, but I'm happy to change that if anyone feels strongly about this.

### DIFF
--- a/.changeset/wet-melons-cry.md
+++ b/.changeset/wet-melons-cry.md
@@ -1,0 +1,6 @@
+---
+"@electric-sql/experimental": patch
+"@core/sync-service": patch
+---
+
+Encode LSN as string in JSON responses for correct handling of large values (>53 bits) in Javascript.
--- a/packages/experimental/src/bigint-utils.ts
+++ b/packages/experimental/src/bigint-utils.ts
@@ -1,0 +1,11 @@
+export function bigIntMax(...args: Array<bigint | number>): bigint {
+  return BigInt(args.reduce((m, e) => (e > m ? e : m)))
+}
+
+export function bigIntMin(...args: Array<bigint | number>): bigint {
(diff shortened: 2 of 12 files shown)


samwillis:
Note @balegas is of the option we may want to drop `op_position` as it implies we can maintain order of operations, when we don't. May be worth doing that in this PR? - question for @balegas

samwillis (approved):
client side LGTM 👍

icehaunter (approved):

msfstef:
Rebased to main as these changes can be independent of the PR where I discovered the issue - merging away! @samwillis regarding the `op_offset`, we can definitely do that in a separate PR if we want, no need to accumulate these changes

Issue #2447, opened 2025-03-13
`global_last_seen_lsn` can be ahead of messages in the next request


samwillis:
There appears to be a race condition in how we set `global_last_seen_lsn` on the `up-to-date` control message. It is possible for it to be ahead of / the same as messages lsn in the next request.

See https://discord.com/channels/933657521581858818/959442635972563005/1349407230558076938

This was uncovered in the PGlite sync plugin tests: 

Logs from the tests:

- fetch 4: I get only an up-to-date message with global_last_seen_lsn: 34918728
- fetch 5: I get a data message (with lsn=34918728), and an up-to-date with global_last_seen_lsn: 34918728 again

So fetch 4 sends the up-to-date with a global_last_seen_lsn thats ahead of its own stream.

```
stdout | test-e2e/sync-e2e.test.ts > sync-e2e > resumes sync after unsubscribe
>> resuming from subscription state null
>> fetch 1 http://localhost:3000/v1/shape?offset=-1&table=todo { signal: AbortSignal { aborted: false }, headers: {} }

stdout | test-e2e/sync-e2e.test.ts > sync-e2e > resumes sync after unsubscribe
>> fetch res 1 200 OK Headers {
  'transfer-encoding': 'chunked',
  date: 'Wed, 12 Mar 2025 16:14:15 GMT',
  'cache-control': 'public, max-age=604800, s-maxage=3600, stale-while-revalidate=2629746',
  'x-request-id': 'GCwaSMKROGQL6dMAAPRh',
  server: 'ElectricSQL/1.0.0-beta.23',
  'access-control-allow-origin': '*',
  'access-control-expose-headers': '*',
  'access-control-allow-methods': 'GET, HEAD, DELETE, OPTIONS',
  'content-type': 'application/json; charset=utf-8',
  etag: '"29212139-1741796055687:-1:0_0"',
  'electric-handle': '29212139-1741796055687',
  'electric-schema': '{"done":{"type":"bool"},"id":{"type":"int4","not_null":true,"pk_index":0},"task":{"type":"text"}}',
  'electric-offset': '0_0'
}
>> fetch 2 http://localhost:3000/v1/shape?handle=29212139-1741796055687&offset=0_0&table=todo { signal: AbortSignal { aborted: false }, headers: {} }

stdout | test-e2e/sync-e2e.test.ts > sync-e2e > resumes sync after unsubscribe
>> received message {
  key: '"public"."todo"/"1"',
[...]

PR #2450, opened 2025-03-13
fix: Fix race that causes incorrect `global_last_seen_lsn` delivery on up-to-date, non-live repsonses

msfstef:
Fixes https://github.com/electric-sql/electric/issues/2447

## Problem
- We were determining the `global_last_seen_lsn` at the end of the request processing pipeline
- The `last_offset` available for the shape was determined at the start, and that is what we use to decide if the response is up-to-date or not
- In between reading `last_offset` at the start (and determining response was up-to-date), and reading `global_last_seen_lsn` at the end, a change might have come in, increasing the global seen LSN and adding a change to the current shape's log, which won't be seen by the response as it will read up to and including the `last_offset` which was determined at the start.

## Testing
- I first tried to reproduce the case with integration tests (`test/electric/plug/router.exs`) and adding arbitrary `Process.sleep` calls until I could reproduce the situation - but since this was a race this was a bit hard to add as a proper test
- I moved to `test/electric/shapes/api_test.exs` where I could use our mocks to simulate the global LSN moving right after the `get_chunk_end_log_offset` call is made - this test could now consistently detect this simulated race. Not as good as an integration test but fully reproducible, controlled, and fast to run.

## Solution
- Moved the `global_last_seen_lsn` retrieval to right before we determine the actual log `last_offset`, we prefer potentially being veeery slightly far behind the actual global last seen LSN than having the above race occur.
- Made this step into a `determine_global_last_seen_lsn` method to match our existing structure, and placed it in all appropriate places
  - Right _before_ we determine the chunk end offset when validating the request
  - Right before we re-determine the chunk end offset during a live request when we receive new changes
  - Right before we return an empty live response after a long poll timeout to ensure that we get the latest global seen LSN


samwillis (approved):
Works for me on a local build with https://github.com/electric-sql/pglite/pull/566 🥳 

I'm unqualified to review the code.

alco (approved):

PR #2454, opened 2025-03-14
feat: remove full table info from shape definitions to lower mem

icehaunter:
## Reasons

We have a process setup where we have 2 layers of supervisors which store entire child spec of their children as state, plus the `ShapeLogCollector` that the shape consumers register with, plus the consumers themselves -- all these processes happen to have shapes in their state. _Arguably, we can inject shapes into consumers bypassing supervisors, but that's a different PR._ These shapes, however, happened to know entire schema of the table they were defined against, which was duplicated a lot. This caused a lot of unnecessary memory usage when the table schema was sufficiently large.

So, in numbers, for a 75 column table, without any transactions going through (just shape creation)
Baseline, no shapes: 71kB
- Current, 300 shapes
	- Total (for shape-related processes): 64.1MB
	- Total overhead in shared processes: 28MB 
	- Total overhead in shared per shape: 93kB
	- Per-shape process total: 35.9MB
	- Per-shape 119kB
- With this PR, 300 shapes
	- Total (for shape-related processes): 21MB
	- Total overhead in shared processes: 7.3MB
	- Total overhead in shared per shape: 24kB
	- Per-shape process total: 13MB
	- Per-shape 45kB

So for a sufficiently large table, this PR drops used memory a lot, 60% in this particular case. Further optimizations of state are possible and we'll do them, but this is the lowest hanging fruit.

## Changes

Shapes now don't store their table's entire schema. That's OK when processing transactions or querying the DB, but that's not enough to invalidate the shape when the schema changes, and it can change while Electric is down as well.

To address this, now the burden of figuring out whether a schema-breaking change has happened from the incoming relations now lies upon `ShapeLogCollector`. It tracks all seen relation messages (which should be relatively rare unless PG is constantly disconnecting), and marks the columns that have changed within this relation as compared to a previously seen version. It also saves these last-seen relations when all shapes have reacted to it and invalidated themselves.

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -6,6 +6,7 @@ defmodule Electric.Replication.ShapeLogCollector do
   use GenStage
 
   require Electric.Postgres.Lsn
+  alias Electric.Replication.ShapeLogCollector.AffectedColumns
   alias Electric.Postgres.Lsn
   alias Electric.Replication.PersistentReplicationState
   alias Electric.Postgres.Inspector
@@ -67,6 +68,11 @@ defmodule Electric.Replication.ShapeLogCollector do
       persistent_kv: opts.persistent_kv
     ]
 
+    {:ok, tracker_state} =
+      persistent_replication_data_opts
+      |> PersistentReplicationState.get_tracked_relations()
+      |> AffectedColumns.init()
+
     state =
       Map.merge(opts, %{
         producer: nil,
@@ -74,8 +80,7 @@ defmodule Electric.Replication.ShapeLogCollector do
         persistent_replication_data_opts: persistent_replication_data_opts,
         last_seen_lsn:
           PersistentReplicationState.get_last_processed_lsn(persistent_replication_data_opts),
-        tracked_relations:
-          PersistentReplicationState.get_tracked_relations(persistent_replication_data_opts)
+        tracked_relations: tracker_state
       })
 
     # start in demand: :accumulate mode so that the ShapeCache is able to start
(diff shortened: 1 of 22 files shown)


KyleAMathews:
Amazing! 2/3 memory reduction! 😍

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -181,19 +189,109 @@ defmodule Electric.Replication.ShapeLogCollector do
...
-      "Dropping relation message for #{inspect(rel.schema)}.#{inspect(rel.table)}: no active consumers"
-    end)
+  defp handle_relation(rel, from, state) do
+    OpenTelemetry.add_span_attributes("rel.is_dropped": false)
 
-    OpenTelemetry.add_span_attributes("rel.is_dropped": true)
+    {updated_rel, updated_state} = track_relation(rel, state)
 
-    {:reply, :ok, [], state}
+    case updated_state do
```
robacourt, line 197: The shape log collector is a pretty big module and does lots of things. Could this relation logic be extracted into a new module?

icehaunter: Probably could, but I'm on vacation next week, and it's already quite late. Can i makr that as follow up pr?

robacourt: There's no urgency as far as Trigger.dev is concerned, it'd take 60 days for their memory to run out and unfortunately they'll have WAL lag issues before then. I'd say it's better to get it refactored before it's merged. It'd also help me assess this PR.

robacourt (approved):
Looks great! Nice work. Thanks for the refactoring!

PR #2457, opened 2025-03-17
feat: add `params` support to TS, Elixir clients & docs

icehaunter:


packages/typescript-client/src/client.ts
```diff
@@ -729,3 +740,17 @@ function validateOptions<T>(options: Partial<ShapeStreamOptions<T>>): void {
...
 }
+
+function setQueryParam(
+  url: URL,
+  key: string,
+  value: Record<string, string> | string
+): void {
+  if (typeof value === `string`) {
+    url.searchParams.set(key, value)
+  } else {
```
samwillis, line 751: I know our API is typed so that it only excepts string values, but I think we should maybe handle numbers in here and convert to strings. Currently they would fall through to the object handling and throw with a weird error (ignoring the type error). If we just call `toString()` on numbers I think thats fine. We leave the types the same, it maybe just prevents a foot gun?

packages/typescript-client/test/stream.test.ts
```diff
@@ -120,4 +120,43 @@ describe(`ShapeStream`, () => {
...
+    }
+
+    const aborter = new AbortController()
+    const stream = new ShapeStream({
+      url: shapeUrl,
+      params: {
+        table: `foo`,
+        where: `a=$1 and b=$2`,
+        columns: [`id`],
+        params: {
```
samwillis, line 142: I initially thought that we would want to provide an array of positional params, but that would create a disconnect between js array indexes and Postgres positional arguments (1 based). I think this is the correct strategy.

samwillis (approved):
This looks good to me - I think the API for the params as an object, rather than array, is portably the better tradeoff.

We need to add docs for this to the typescript client docs.

packages/typescript-client/src/client.ts
icehaunter: Done, added a fallback case with a `.toString()` so that it's a bit more defensive

packages/elixir-client/lib/electric/client/shape_definition.ex
```diff
@@ -34,6 +34,12 @@ defmodule Electric.Client.ShapeDefinition do
               doc:
                 "The namespace the table belongs to. If `nil` then Postgres will use whatever schema is the default (usually `public`)."
             ],
+            params: [
+              type: {:or, [nil, {:map, :pos_integer, :string}, {:list, :string}]},
```
kevin-dp, line 38: Is `:pos_integer` including zero or not?
iirc they are indexed from one so we may want to ensure that `:pos_integer` also rejects 0.

packages/elixir-client/lib/electric/client/shape_definition.ex
```diff
@@ -34,6 +34,12 @@ defmodule Electric.Client.ShapeDefinition do
...
+              default: nil,
+              doc:
+                "List of parameters to include in the shape. If `nil` this is equivalent to all parameters (`SELECT *`)"
```
kevin-dp, line 41: The 2nd part of this sentence looks wrong? Seems like a copy/paste issue from the `columns` option.

packages/typescript-client/src/client.ts
```diff
@@ -64,6 +65,14 @@ export interface PostgresParams {
   /** The where clauses for the shape */
   where?: string
 
+  /**
+   * Positional where clause paramater values. These will be passed to the server
+   * and will substitute `$1` parameters in the where clause.
```
kevin-dp, line 70: nit: `$i` is better than `$1` here

packages/typescript-client/src/client.ts
```diff
@@ -64,6 +65,14 @@ export interface PostgresParams {
...
+   *
+   * If where clause is `id = $1 or id = $2`, params must have keys `"1"` and `"2"`.
+   */
+  params?: Record<string, string>
```
kevin-dp, line 74: I don't like this. I know that this might be common for query parameters in a URL, but this doesn't make much sense in JS. An object where the keys are indices is just an array. As a user i would be annoyed if i had to pass an object like this (with quotes in the string which is very likely to forget!). I'll strongly argue that these need to be passed as an array (like all PG client libraries do) and that we do the index trickery behind the scenes.

kevin-dp (changes requested):
Nice work!
Most comments are minor, just left one comment about the format of params in the TS library which i don't like.

packages/elixir-client/lib/electric/client/shape_definition.ex
icehaunter: `pos_integer` is not a `non_neg_integer` :D (so it will reject, yes)

PR #2458, opened 2025-03-17
chore: Rename server header to electric-server

kevin-dp:
In certain environments, e.g. when running in a cloudflare tunnel, the `server` header that we set to `"ElectricSQL/#{version}"` gets overwritten by cloudflare with the value `"cloudflare"`. This PR renames the response header to `electric-version` to avoid this and improve visibility.

**Should we mark this as a breaking change or not?**
If someone relies on this header it is in theory a breaking change.
Or we could keep the `server` header and also introduce the `electric-server` header.

### DIFF
--- a/packages/sync-service/lib/electric/plug/router.ex
+++ b/packages/sync-service/lib/electric/plug/router.ex
@@ -45,7 +45,7 @@ defmodule Electric.Plug.Router do
   match _, do: send_resp(conn, 404, "Not found")
 
   def server_header(conn, version),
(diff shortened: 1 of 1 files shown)


alco:
Not a serious breaking change, nobody should have business logic hanging off the server header.

I think it's good that we use a custom header only for this, this doesn't unnecessarily condition clients to rely on the presence of the Server header where it does not reliably show the server version in all environments.

alco (approved):

PR #2460, opened 2025-03-17
fix: allow arrays for where clause params

icehaunter:


packages/typescript-client/src/client.ts
```diff
@@ -69,9 +69,12 @@ export interface PostgresParams {
    * Positional where clause paramater values. These will be passed to the server
    * and will substitute `$1` parameters in the where clause.
    *
-   * If where clause is `id = $1 or id = $2`, params must have keys `"1"` and `"2"`.
+   * It can be an array (note that positional arguments start at 1, the array will be mapped
```
kevin-dp, line 72: Nit: can you replace `$1` by `$i` in the comment above please.

packages/typescript-client/src/client.ts
```diff
@@ -761,3 +764,14 @@ function setQueryParam(
     url.searchParams.set(key, value.toString())
   }
 }
+
+function convertWhereParamsToObj(
+  options: ExternalParamsRecord
+): ExternalParamsRecord {
+  if (Array.isArray(options.params)) {
+    options.params = Object.fromEntries(
```
kevin-dp, line 772: Are we sure we want to mutate the object?
I didn't double check but i'm assuming this object has not been copied and is the one passed by the user.
The user may not expect that we modify their object.
I'd consider returning a modified copy of the object.

packages/typescript-client/src/client.ts
```diff
@@ -761,3 +764,14 @@ function setQueryParam(
     url.searchParams.set(key, value.toString())
   }
 }
+
+function convertWhereParamsToObj(
+  options: ExternalParamsRecord
```
kevin-dp, line 769: nit: this is not the `options` object but the `options.params` object, perhaps `paramsOptions` would be more descriptive. It's pretty minor but since we have `options.params.params` i think it is a good thing to be more explicit about it.

website/docs/api/clients/typescript.md
```diff
@@ -101,12 +101,15 @@ export interface ShapeStreamOptions<T = never> {
     where?: string
 
     /**
-     * Params for the positional parameters (`$1`) in the where clause. Must have keys
-     * matching the used positional parameters in the where clause.
+     * Positional where clause paramater values. These will be passed to the server
+     * and will substitute `$1` parameters in the where clause.
```
kevin-dp, line 105: i would replace `$1` by `$i` here.

kevin-dp (approved):
LGTM!
Left a few minor comments

packages/typescript-client/src/client.ts
icehaunter: I think type is very explicit already, but I've edited the name

PR #2466, opened 2025-03-18
Move liveview example to phoenix.sync

magnetised:


kevin-dp (approved):
LGTM although it's probably best to get a 2nd opinion for an Elixir expert.

PR #2467, opened 2025-03-19
[WIP] SQLite-sync

balegas:
Sync ElectricSQL shapes into a SQLite database. Based on the great [pglite-sync](https://github.com/electric-sql/pglite/tree/main/packages/pglite-sync) extension for PGlite.

Thin-layer that applies shapes into your local SQLite database. Works with your preferred SQLite build and driver, just implement the generic `SQLiteWrapper` driver interface. 

Works with the experimental `MultiShapeStream` for synching multiple tables into local database.

TODO
- [ ] README
- [ ] Document constraints and limitations
- [ ] Maybe some more wrappers
- [ ] Wait for pglite-sync to actually move into monorepo
- [ ] Integration test

### DIFF
--- a/packages/sqlite-sync/src/sync.ts
+++ b/packages/sqlite-sync/src/sync.ts
@@ -242,9 +242,8 @@ export const makeElectricSync = (
             debug,
           })
         }
-        // TODO need to add explicit rollback inside transaction
         if (unsubscribed) {
-          throw new Error(`Unsubscribed while committing`)
+          tx.rollback()
         }
       })
       if (debug) console.timeEnd(`commit`)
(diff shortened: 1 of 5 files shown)


KyleAMathews:
Talking to Traun Leyden about schema migrations.

We're thinking the answer is a callback that lets the app code decide what to do. The app could decide to immediately do a schema change or just flag it up until the new time the app loads, etc.

KyleAMathews:
Another option is the app sends an app version number with each request & the proxy rejects requests for app versions that are too old.

icehaunter:
@balegas said this is not longer relevant

PR #2470, opened 2025-03-20
Removing file writes from the main WAL processing loop

robacourt:
Speed up replication processing by removing file writes from the main processing loop:
- persisting the tracked relations now only happens on a relation change, not every transaction
- `last_processed_lsn` is no longer persisted, but it is worked out from the last lsn seen in the shape logs. This also avoids a race condition where the lsn is written in the shape logs but not the PersistedReplicationState.

As part of this PR, this also fixes a bug we found where `last_processed_lsn` was always being set to `Lsn.from_integer(0)` on start-up rather than reading from disk.

Please note that as part of this PR, `last_processed_lsn` and `global_last_seen_lsn` no longer include LSNs from transactions that do not affect any shapes.

Here we can see a 20% increase in WAL processing speed:
Screenshot 2025-03-24 at 15 11 29

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -23,6 +23,7 @@ end
 defmodule Electric.ShapeCache do
   use GenServer
 
+  alias Electric.Postgres.Lsn
   alias Electric.Replication.LogOffset
   alias Electric.ShapeCache.ShapeStatus
   alias Electric.Shapes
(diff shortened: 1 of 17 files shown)


robacourt:
benchmark this

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -293,21 +297,25 @@ defmodule Electric.ShapeCache do
...
     |> state.shape_status.list_shapes()
-    |> Enum.each(fn {shape_handle, shape} ->
+    |> Enum.flat_map(fn {shape_handle, shape} ->
       try do
-        {:ok, _latest_offset} = start_shape(shape_handle, shape, state)
+        {:ok, latest_offset} = start_shape(shape_handle, shape, state)
 
         # recover publication filter state
         publication_manager.recover_shape(shape, publication_manager_opts)
+        [LogOffset.extract_lsn(latest_offset)]
```
robacourt, line 306: @msfstef I reverted the change that made these offsets rather than LSNs, for a couple of reasons:
- LSNs were not hidden from ShapeCache, it still had to convert to an LSN to talk to start processing on the ShapeLogCollector
- LogOffset.max doesn't make sense when comparing offsets from different shapes, the op_opset is different in each shape

In short, LSN is the only thing that makes sense comparing globally, hiding them in offsets for an incomplete decoupling seems to just add confusion and not much benefit to me. But happy to revert back if you disagree.

robacourt:
benchmark this

magnetised (approved):
Looks, good. Nice change. Look forward to those benchmarks...

Issue #2471, opened 2025-03-20
Document the `api_secret` Query Parameter.


yvesnrb:
The API configuration page describes how to set [ELECTRIC_SECRET](https://electric-sql.com/docs/api/config#electric-secret), which Electric will use to authorize shape requests. However from looking at the website and OpenAPI spec, it is unclear how this secret should be provided to Electric in the requests. From poking around in the source a bit I discovered it should be a query parameter named `api_secret`. Should this be documented on the API spec?

giladrom:
Came here to look for this - still seems to be missing in the documentation.

PR #2472, opened 2025-03-20
docs: document `api_secret` query parameter (#2471)

yvesnrb:
Closes isse #2471

### DIFF
--- a/website/electric-api.yaml
+++ b/website/electric-api.yaml
@@ -16,6 +16,20 @@ servers:
   - url: http://localhost:3000
     description: Local server
 
+components:
+  parameters:
+    api_secret:
+      name: api_secret
+      in: query
+      schema:
+        type: string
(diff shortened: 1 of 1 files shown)


yvesnrb:
Also, I've noticed the comment reading `Query parameters` seems to be duplicated on lines 44 and 63 and again on lines 484 and 501. Seems like it should only be on lines 44 and 484.

kevin-dp:
Thanks for catching this and opening a PR to fix it!
The duplicated comment is indeed not necessary.
Any reason you are using `$ref` instead of inlining it like the other parameters?

yvesnrb:
> Thanks for catching this and opening a PR to fix it! The duplicated comment is indeed not necessary. Any reason you are using `$ref` instead of inlining it like the other parameters?

Just thought I'd avoid duplicating the same thing on both the get and delete methods. I got a commit here inlining it if you'd prefer.

kevin-dp (approved):

Issue #2474, opened 2025-03-23
Return 409 status w/ must-refetch control on handle/shape def mismatch


KyleAMathews:
A pattern that keeps coming up is that a proxy can decide what the shape definition is, not the client. This means however that the shape definition request can change from request to request — in which case the client should refetch. But this isn't handled correctly atm as we just return a 400 w/ an error when there's a mismatch as we thought that'd be a really unexpected event.

But now it seems like it's a fairly normal event so we'll return 409 & must-fetch like other errors of this sort where the fix is for the client to refetch.

samwillis:
Agree this should be the behaviour.

While looking at it yesterday for the cloud dashboard I found this old PR where the current behaviour was discussed and implemented: https://github.com/electric-sql/electric/pull/1880

In particular this part of the thread: https://github.com/electric-sql/electric/pull/1880#discussion_r1810760885

I can't think of any edge cases off the top of my head where a 409 for a valid, but unmatched, definition and handle would result in any adverse behaviour.

I hacked in 409 to the cloud dashboard proxy, and the clone behaved as expected.

PR #2476, opened 2025-03-24
fix: Return 409 must-refetch instead of 400 on shape definition and handle mismatch

kevin-dp:
Fixes https://github.com/electric-sql/electric/issues/2474

Currently, Electric returns a 400 error code when the shape definition and shape handle do not match.
However, we discovered a pattern for handling dependent shapes on the proxy where shape B is constructed based on the results for shape A. For example, if we want to load all teams for a user we would first create a shape to get all team_ids a user belongs to and then another shape that selects all teams where `team_id IN (team_id_1, ..., team_id_n)`. The definition of the second shape changes whenever the results of the first shape change. For this to work properly, we need Electric to return 409 must-refetch with the handle of the new shape when the shape definition changes.


kevin-dp:
There's a failing TS integration test "should chunk a large log with reasonably sized chunks" which keeps failing.
Although it seems totally unrelated to the changes introduced in this PR.

msfstef (approved):
Looks good! I wasn't around for the discussion of this change but I'm not sure how I feel about us always creating new shape with changes in the definition, which allows the clients a lot of flexibility but also allows them a lot of room to blow up the server because of client errors (changing the shape definition  continuously for a given shape).

I do like that it simplifies the logic though, and if we have a good way to garbage collect shapes and make them even cheaper than they are now then this makes sense.

PR #2477, opened 2025-03-24
Add CI smoke test for Phoenix.Sync compatibility

magnetised:
Uses the `examples/phoenix-liveview` Phoenix app to check if the current Electric still works with `Phoenix.Sync`.

Depends on #2466 which migrates the liveview example to Phoenix.Sync


magnetised:
Closing in favour of merging all this into #2466

Issue #2469, opened 2025-03-19
Empty live 204 responses with body failing with strict clients


KyleAMathews:
It seems we're not handling this correctly.

> Initial shape works just updates that dont work, this is just:
id: uuid
workspace_id: ref to another table
content: binary
timestamp: bigint
sequence_number: int
> 
> I never do updates no an existing row, only inserts, im pretty sure its the binary serialisation or something similar because ive tried with other tables and it works just fine! Im evaluating electric sql for my app digitalnotes.com, currently have my own sync mechanism on top of redis streams

https://discord.com/channels/933657521581858818/1351967731817119764/1351975066547519548

flaviouk:
```
09:36:39.292 [info] Received relation "public"."content_updates"
09:36:39.305 [debug] Relation received in ShapeLogCollector: %Electric.Replication.Changes.Relation{id: 18664, schema: "public", table: "content_updates", columns: [%Electric.Replication.Changes.Column{name: "workspace_id", type_oid: 2950}, %Electric.Replication.Changes.Column{name: "content", type_oid: 17}, %Electric.Replication.Changes.Column{name: "timestamp", type_oid: 20}, %Electric.Replication.Changes.Column{name: "sequence_number", type_oid: 23}, %Electric.Replication.Changes.Column{name: "id", type_oid: 2950}]}
09:36:39.305 [info] Schema for the table "public"."content_updates" changed - terminating shape 43627731-1742463352996
09:36:39.306 [info] Schema for the table "public"."content_updates" changed - terminating shape 55449792-1742463238528
09:36:39.375 [debug] 1 consumers of replication stream
09:36:39.387 [debug] 0 consumers of replication stream
09:36:39.387 [debug] Updating last processed lsn to 6959815832
09:36:39.388 [debug] [SET] /persistent/state/electric-embedded:last_processed_lsn
09:36:39.388 [info] Received transaction 3036979 from Postgres at 1/9ED814E8
[...]

PR #2480, opened 2025-03-25
fix: Add tests for binary columns

msfstef:
Was triaging https://github.com/electric-sql/electric/issues/2469 and I added two integration tests:

1. Ensure most types get replicated correctly, both for the initial snapshot and subsequent log updates
2. Ensure large binaries (>1 MB) get replicated correctly as well

Both tests are passing but I think they are worth including in the test suite.

I think we need to change the title of the above issue as looking at it more closely the problem is that we return a body with a 204 response, which is against the spec (and is what the error is reporting as well).


robacourt (approved):
Nice! ⭐

PR #2481, opened 2025-03-25
llms: build and expose `/llms.txt`.

thruflo:
https://deploy-preview-2481--electric-next.netlify.app/llms.txt

Without:

With:

Claude wrote the llms.js script. Basically pulling in the content from the docs, so it doesn't get out of date. With some horrible hacks to slim down and get the nesting and spacing right.

Don't judge me. I was just following the vibes.

### DIFF
--- a/scripts/llms.js
+++ b/scripts/llms.js
@@ -42,8 +42,6 @@ function replaceVersionPlaceholders(content) {
   for (const [packageName, placeholder] of Object.entries(packageVersionPlaceholders)) {
     const version = getPackageVersion(packageName);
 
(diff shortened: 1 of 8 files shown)


KyleAMathews:
This is ~38k tokens according to https://tokenizer.streamlit.app/

Which is ~fine — that fits within most models. It's a bit expensive but until various coding helpers have better tools for feeding custom docs into them, this seems like a good first step.

KyleAMathews (approved):

thruflo:
Hmm, that's more than I thought. I'll have a go at slimming down a bit.

PR #2486, opened 2025-03-27
fix!: Return 200 for empty live response

msfstef:
Closes https://github.com/electric-sql/electric/issues/2469

Need to address whether this is a breaking change, if we want to do it experimentally with a flag, whether we instead want to keep 204 responses and move info to headers.

Opening this PR so we can have the scope of the changes ready.

BREAKING CHANGE: community clients relying on `204` alone for up-to-date logic might break - live responses now always return a `200` with a body/

### DIFF
--- a/.changeset/beige-rivers-shout.md
+++ b/.changeset/beige-rivers-shout.md
@@ -3,4 +3,6 @@
 "@core/sync-service": patch
 ---
 
-Convert live responses with no changes from `204` to `200`.
+fix!: Convert live responses with no changes from `204` to `200`.
+
+BREAKING CHANGE: community clients relying on `204` alone for up-to-date logic might break - live responses now always return a `200` with a body/


thruflo:
Discussion https://discord.com/channels/933657521581858818/959442635972563005/1354736341375651962

thruflo:
Can I note that fixing the Typescript client does not all the clients fix. We have an official Elixir client. We also have community contributed clients, like for Swift and Flutter, and documentation on writing your own client.

I know I'm commenting on a draft / initial sketch for comments but it's important not to be too tunnel vision on the TS client.

msfstef:
> We have an official Elixir client.

I am relying on the lack of mention of 204s in the implementation + the test suite passing without any changes on that

> We also have community contributed clients, like for Swift and Flutter, and documentation on writing your own client.

I would be happy to take a look personally to see if they use the 204 status code - I changed the documentation mentioning it in this draft as well as the API spec, and even our TS client actually used it but would work without it, it was mostly the TS client tests that I fixed that simulated 204 responses.


It is unfortunate that this is a breaking change but simultaneously I'm fairly confident it is not as breaking to sanely implemented clients, whereas we saw that clients using http libraries that want to respect the http spec are already breaking.

msfstef:
FWIW our TS client test suite on `main` passes when ran against the Elixir sync service in this branch - I think for all intents and purposes our own clients are compatible with this change and is thus not breaking, the earlier we commit to a solution the fewer community driven clients will have trouble with this.

icehaunter (approved):

KyleAMathews:
Using a 204 was more of a rhetorical flourish than needed so I'm fine removing it 👍

kevin-dp (approved):
LGTM!

jtkeyva:
> Can I note that fixing the Typescript client does not all the clients fix. We have an official Elixir client. We also have community contributed clients, like for Swift and Flutter, and documentation on writing your own client.
> 
> I know I'm commenting on a draft / initial sketch for comments but it's important not to be too tunnel vision on the TS client.

can you please share the dart/flutter solution?

PR #2490, opened 2025-03-27
fix: Synchronously refresh publication and clear cache if failed

msfstef:
Followup to https://github.com/electric-sql/electric/pull/2487

In the above PR I fixed an issue that would cause the publication manager to exit, which made Electric go into a restart loop.

There were two issues that compounded that problem:
1. I had previously converted the recovery of the shapes on the publication manager asynchronous, using `cast` instead of `call`, which means that Electric went ahead with initialisation and marked itself healthy despite not managing to actually recover - if it had waited for the recovery to finish and fail the container would never have been marked as healthy and never deployed.
2. If a recovery fails, for whatever reason, it might enter into an infinite recovery loop as it is likely some of the persisted state from which we are trying to recover is causing the issue. We should be able to recover from that since we consider the storage/cache disposable.

In order to solve the above issues, I implement the following:
1. Convert the calls back to synchronous calls, increasing the timeout for `ShapeCache` initialisation slightly, as that was the original purpose of making them asynchronous (it might take a while to recover things).
  a. I'm not entirely sure if this is a bit of an anti-pattern - we could place the initialisation in a `{:continue, :restore_shapes}` after the init - my concern is that if the rest of Electric continues on and initialises everything else before the restore is done we might have a problem.
2. Use `try..catch..after` blocks to ensure cleanup occurs if any recovery operation fails, such that Electric will still manage to start after the supervisor restarts the process and the old, problematic state is cleaned up.
  a. It's a bit of a brutal approach, I've tested it by tweaking the publication manager to fail every now and then and it works really nicely, where it will erase the stored shapes and restart and continues on like normal.
  b. We could theoretically place this at a higher level, such as the `start_link` of the `ShapeCache`, so that we always clear the cache if initialisation fails, but I wanted to keep the scope tight and only to things that we have observed during real operation.

I can spend time to add a sort of integration test with a failing `PublicationManager` to better simulate this situation in a test suite rather than ad hoc.

Tagging @balegas as well as we discussed this change

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -292,7 +292,10 @@ defmodule Electric.ShapeCache do
         shape_handle
       )
     after
-      # ensure data is cleaned even if consumer exits mid-call
+      # another failsafe for ensuring data is cleaned if consumer fails to
+      # clean itself - this is not guaranteed to run in case of an actual exit
+      # but provides an additional safeguard
+
       shape_handle
       |> Electric.ShapeCache.Storage.for_shape(state.storage)
       |> Electric.ShapeCache.Storage.unsafe_cleanup!()
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -218,8 +218,12 @@ defmodule Electric.ShapeCache do
 
     try do
       :ok = publication_manager.refresh_publication(publication_manager_opts)
+    rescue
+      error ->
+        clean_up_all_shapes(state)
+        reraise error, __STACKTRACE__
     catch
-      kind, reason when kind in [:exit, :error] ->
+      :exit, reason ->
         clean_up_all_shapes(state)
         exit(reason)
     end
@@ -321,7 +325,7 @@ defmodule Electric.ShapeCache do
       catch
         kind, reason when kind in [:exit, :error] ->
           Logger.error(
-            "Failed to recover shape #{shape_handle}: #{inspect(kind)} #{inspect(reason)}"
+            "Failed to recover shape #{shape_handle}: #{Exception.format(kind, reason, __STACKTRACE__)}"
           )
 
           # clean up corrupted data to avoid persisting bad state
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -88,7 +88,10 @@ defmodule Electric.ShapeCache do
       GenServer.start_link(
         __MODULE__,
         Map.new(opts) |> Map.put(:db_pool, db_pool) |> Map.put(:name, name),
-        name: name
+        name: name,
+        # allow extra time for shape cache to initialise as it needs to
+        # restore existing shapes
+        timeout: Keyword.get(opts, :timeout, 20_000)
       )
     end
   end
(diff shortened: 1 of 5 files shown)


robacourt (approved):
LGTM

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -209,9 +212,17 @@ defmodule Electric.ShapeCache do
...
     {publication_manager, publication_manager_opts} = opts.publication_manager
-    publication_manager.refresh_publication(publication_manager_opts)
+
+    try do
+      :ok = publication_manager.refresh_publication(publication_manager_opts)
+    catch
+      kind, reason when kind in [:exit, :error] ->
+        clean_up_all_shapes(state)
+        exit(reason)
+    end
```
icehaunter, line 225: I don't think we should be converting exits to errors. I think a cleaner approach would be to add a separate `rescue` clause:


```suggestion
    rescue
      error -> 
        clean_up_all_shapes(state)
        reraise error, __STACKTRACE__
    catch
      :exit, reason ->
        clean_up_all_shapes(state)
        exit(reason)
    end
```

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -270,15 +281,18 @@ defmodule Electric.ShapeCache do
...
-      state.consumer_supervisor,
-      state.stack_id,
+    try do
+      Electric.Shapes.DynamicConsumerSupervisor.stop_shape_consumer(
+        state.consumer_supervisor,
+        state.stack_id,
+        shape_handle
+      )
+    after
+      # ensure data is cleaned even if consumer exits mid-call
```
icehaunter, line 291: This comment is either false, or not precise enough: exits are not caught by `after`. 

> An after clause allows you to define cleanup logic that will be invoked both when the block of code passed to try/1 succeeds and also when an error is raised. Note that the process will exit as usual when receiving an exit signal that causes it to exit abruptly and so the after clause is not guaranteed to be executed.

- https://hexdocs.pm/elixir/1.18.1/Kernel.SpecialForms.html#try/1

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -304,9 +318,11 @@ defmodule Electric.ShapeCache do
         # recover publication filter state
         publication_manager.recover_shape(shape, publication_manager_opts)
         [LogOffset.extract_lsn(latest_offset)]
-      rescue
-        e ->
-          Logger.error("Failed to recover shape #{shape_handle}: #{inspect(e)}")
+      catch
+        kind, reason when kind in [:exit, :error] ->
+          Logger.error(
+            "Failed to recover shape #{shape_handle}: #{inspect(kind)} #{inspect(reason)}"
```
icehaunter, line 324: It would be better to format the error/exit in a standard way using [`Exception.format/3`](https://hexdocs.pm/elixir/1.18.1/Exception.html#format/3)

icehaunter (approved):

packages/sync-service/lib/electric/shape_cache.ex
alco: At the risk of being overly pedantic, the `after` clause does get executed if `exit()` is called inside the current process.

Given this test module:
```elixir
$ cat test_mod.exs 
defmodule M do
  def test(arg) do
    try do
      case arg do
        :raise -> raise "Error"
        :throw -> throw "Thrown"
        :exit -> exit(:test_exit)
      end
    after
      IO.puts "I'm inside 'after'"
    end
  end
end

IO.puts "current process: #{inspect self()}"
M.test :exit
IO.puts "current process: #{inspect self()}"
```

Here's what gets displayed in the shell:
```
$ elixir test_mod.exs 
current process: #PID<0.95.0>
I'm inside 'after'
** (exit) :test_exit
    test_mod.exs:7: M.test/1
    test_mod.exs:16: (file)
```

The second `IO.puts()` call is not reached.

msfstef: Yeah the idea is only to cleanup in case of an exit in the called consumer process (which was called to clean itself up), not in the calling shape cache process itself.

Issue #2493, opened 2025-03-28
Using array spreads for cases that can easily exceed 120k values


AntonOfTheWoods:
In various places in the code, developers have used the modern JS syntax `...` for destructuring arrays. This uses the stack in many implementations. Many implementations are relatively restricted (~120k in Chrome, for example).

https://github.com/electric-sql/electric/blob/main/packages/experimental/src/multi-shape-stream.ts#L186, or https://github.com/electric-sql/pglite/pull/571/files#diff-0b1e57b1dd80601f11d4fcd31ce6223f3ad5d79d474598869a2ab830ae69c635L164, for example.

This causes many cases where the stack is exhausted, and crashes are caused. It seems there is a lack of testing for these cases.

PR #2498, opened 2025-03-31
fix: Avoid function arg destructuring to not hit stack limits

msfstef:
Closes https://github.com/electric-sql/electric/issues/2493

We should avoid this pattern for functions that are to be used with unbounded argument sizes (if not avoid it all together given this issue)

### DIFF
--- a/packages/experimental/src/multi-shape-stream.ts
+++ b/packages/experimental/src/multi-shape-stream.ts
@@ -312,9 +312,9 @@ export class MultiShapeStream<
     // Min of all the lastSyncedAt values
     const shapeEntries = this.#shapeEntries()
     if (shapeEntries.length === 0) return
-    return shapeEntries
-      .map(([_, shape]) => shape.lastSyncedAt() ?? Infinity)
-      .reduce((a, b) => (a < b ? a : b))
+    return shapeEntries.reduce((minLastSyncedAt, [_, shape]) => {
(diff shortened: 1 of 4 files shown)


samwillis (approved):
LGTM 👍 

Thanks for picking this up!

packages/experimental/src/multi-shape-stream.ts
```diff
@@ -310,11 +310,11 @@ export class MultiShapeStream<
...
-    return Math.min(
-      ...this.#shapeEntries().map(
-        ([_, shape]) => shape.lastSyncedAt() ?? Infinity
-      )
-    )
+    const shapeEntries = this.#shapeEntries()
+    if (shapeEntries.length === 0) return
+    return shapeEntries
+      .map(([_, shape]) => shape.lastSyncedAt() ?? Infinity)
+      .reduce((a, b) => (a < b ? a : b))
```
kevin-dp, line 317: Can simplify to `.reduce(Math.min)`

kevin-dp (approved):

packages/experimental/src/multi-shape-stream.ts
kevin-dp: Btw, this is going to loop twice over the entries. If it's a concern that there may be a lot of entries, let's perhaps do it in one go with reduce.

kevin-dp (approved):

PR #2499, opened 2025-03-31
fix: skip updates that don't change any columns

### LINKED ISSUE #2452: Updates with no changes at all get let through only on shapes without `columns`
When you write to Postgres the same row it has, it generates an update event nonetheless. However currently our filtering is slightly misconfigured, and when such an update comes, if the shape doesn't have a `columns` property, then the update is added to the shape, but if the shape does have a `columns` property, then no updated columns means no intersection with "columns of interest", means it doesn't get added. 

I'm not sure what's the more correct behaviour here, but with consistency and LSN considerations (cc @samwillis), it feels like we should let them through always.

icehaunter:
Closes #2452

### DIFF
--- a/.changeset/short-crabs-glow.md
+++ b/.changeset/short-crabs-glow.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: skip updates that don't change any columns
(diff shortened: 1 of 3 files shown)


robacourt (approved):
👍

Issue #2497, opened 2025-03-31
Handle (or documented) body request timeout failures


AntonOfTheWoods:
Currently several failure/retry scenarios are managed by the standard JS  `ShapeStream` (https://github.com/electric-sql/electric/blob/15e6d1d81a542cb6ec093178c8519b5e0bceab1a/packages/typescript-client/src/client.ts#L313). Body timeouts (e.g, `nginx`'s `client_body_timeout`), however, are not. In these cases, failure happens [here](https://github.com/electric-sql/electric/blob/15e6d1d81a542cb6ec093178c8519b5e0bceab1a/packages/typescript-client/src/client.ts#L550) and there seems to be no standard path for managing this kind of error mentioned in the docs.

The errors appear in the dev console as `TypeError: Failed to fetch`.

If there is any easy work-around, it might be sufficient to document this, possibly in a FAQ.

msfstef:
The top-level error should be manageable via the `onError` callback, however I think it makes sense to treat errors while consuming the body of the response as `FetchError`s, since they are still within the scope of consuming from the network, and publish them to the individual subscribers as errors as well.

I've opened https://github.com/electric-sql/electric/pull/2500 to address this

AntonOfTheWoods:
Thanks for this @msfstef ! I am test driving it and I'm still getting the same error after the timeout, just now it's called `FetchError: Failed to fetch` and it's being thrown in a different place.

I am trying to use the `pglite-sync` module, which uses the new (experimental) `MultiShapeStream` (though I'm using it in single shape mode) but ultimately exposes an `onError` method, like you get [here](https://electric-sql.com/docs/guides/auth#example-1). Presumably there is something I can add to this method that will allow me to get the failed connection to retry? I have spent a while trying to get it to work but I can't get it to retry without it trying to re-insert values that have already been inserted (from the initial sync, which is what is failing due to having non-negligeable data to insert and timing out). 

Should it be retrying already or am I holding it wrong?

msfstef:
The [ `onError` callback](https://electric-sql.com/docs/api/clients/typescript#options) allows you to specify an async handler with a return value that will determine the behaviour - as you've pointed out in the example but pasting it here for convenience as well:

```ts
  /**
   * A function for handling errors.
   * This is optional, when it is not provided any shapestream errors will be thrown.
   * If the function returns an object containing parameters and/or headers
   * the shapestream will apply those changes and try syncing again.
   * If the function returns void the shapestream is stopped.
[...]

PR #2500, opened 2025-03-31
fix: Catch and surface body consumption errors in regular flow

msfstef:
Closes https://github.com/electric-sql/electric/issues/2497

- Wrap fetch such that we ensure that we can not only complete the request but also consume the fetch body in full - and any errors arising from that are still surfaced as `FetchError`
- Always throw errors from our fetch even if the status code is not in the `>= 400 && < 500` range - if we don't, it fails in the next step as `response` is undefined.

In this way, errors during the consumption of the body stream are handled the same as other network errors, interrupting the stream, notifying subscribers, etc.


kevin-dp (commented):
I'm not sure about stopping the loop on 5xx errors.
IIRC the idea was that 5xx errors are transient so we want to keep retrying.
I understand that we want to expose those errors somehow to the client, but i think we should then do that without stopping the loop.

msfstef:
> IIRC the idea was that 5xx errors are transient so we want to keep retrying.

That still happens within the fetch backoff wrapper - if the retries end, we throw `FetchBackoffError` which just breaks out of the loop. However if we have caught an error in the fetch, we will fail it anyway because we have no response to continue fetching. This isn't changing our looping at all, that condition I've removed was simply never failing.

packages/typescript-client/src/client.ts
```diff
@@ -510,11 +513,11 @@ export class ShapeStream<T extends Row<unknown> = Row>
             this.#reset(newShapeHandle)
             await this.#publish(e.json as Message<T>[])
             continue
-          } else if (e.status >= 400 && e.status < 500) {
+          } else {
             // Notify subscribers
             this.#sendErrorToSubscribers(e)
 
-            // 400 errors are not actionable without additional user input,
+            // these errors are not actionable without additional user input,
```
kevin-dp, line 520: Might need clarification here that these will all be 4xx errors (because of how fetchWithBackoff works) which are not actionable without user input.

packages/typescript-client/src/fetch.ts
```diff
@@ -95,6 +95,32 @@ export function createFetchWithBackoff(
...
+    try {
+      if (res.body === null) return res
+      const text = await res.text()
+      return new Response(text, res)
+    } catch (err) {
+      throw new FetchError(
+        res.status,
+        undefined,
+        undefined,
+        Object.fromEntries([...res.headers.entries()]),
```
kevin-dp, line 112: Why do we need to make a copy?

kevin-dp (approved):
Changes look good, thanks!

packages/typescript-client/src/client.ts
msfstef: with this change they might not be - any error that has reached here is not actionable (e.g. the response body somehow is corrupted when reading it) - but I can make the comment a bit more expanded

packages/typescript-client/src/fetch.ts
msfstef: to avoid shared ownership of the headers map with the response - it's not an expensive operation and it's what we do in `FetchError.fromResponse`, I've only added this here since we're doing something slightly different than that static method

PR #2501, opened 2025-03-31
feat: Do not cache >= 400 repsonses and non-GET methods

msfstef:
We currently cache _everything_, including 400 responses which can lead to bad behaviours in long delays in recovering.

This PR sets an explicit `no-cache` header for anything but `GET` and `OPTIONS` requests, and any response that is `>= 400`.

It might be worth discussing what to do with 409 responses - perhaps we actually might want to cache something but for a very short time.


icehaunter (approved):
lgtm

KyleAMathews (approved):
Good catch 👍 

Why would we cache 409s for a short time?

KyleAMathews:
For re-fetch loops?

KyleAMathews:
Ah yeah, good point as 409s are basically a permanent redirect.

msfstef:
Perhaps a `max-age=60, must-revalidate` could work for this? Maybe even shorter - they act as redirects but they are still relatively stateful responses, but by having even a little bit we enable request collapsing and dramatically reduce any traffic surges because of shape invalidations etc

KyleAMathews:
Are they stateful though?

PR #2507, opened 2025-03-31
fix: disallow generated columns in shapes

### LINKED ISSUE #2448: Subscribing to a table that has a generated primary key column causes the shape to get invalidated after every change
## Reproduction steps

1. Create a test table:

       CREATE TABLE tab(val jsonb);

2. Add a generated PK column to it:

       ALTER TABLE tab
         ADD COLUMN id uuid PRIMARY KEY GENERATED ALWAYS AS ((val->>'id')::uuid) STORED;

3. Create a shape subscription:

       curl -i http://localhost:3000/v1/shape?table=tab&offset=-1
       HTTP/1.1 200 OK
       electric-handle: 106269577-1741874073474
       electric-offset: 0_0
       electric-schema: {"id":{"type":"uuid","not_null":true,"pk_index":0},"val":{"type":"jsonb"}}

       []

4. Insert a row into the table:

       INSERT INTO tab VALUES (('{"id": "' || gen_random_uuid() || '"}')::json);

## Observed behaviour

Observe the following in the server log output:

```
14:55:09.053 pid=<0.3985.0> [info] Received relation "public"."tab"

14:55:09.066 pid=<0.3993.0> shape_handle=106269577-1741874073474 [info] Schema for the table "public"."tab" changed - terminating shape 106269577-1741874073474

14:55:09.094 pid=<0.3985.0> [info] Received transaction 756 from Postgres at 0/19A7780
```

Trying to GET the shape returns a 409 response, even though the `electric-schema` value remains the same:

```
HTTP/1.1 409 Conflict
electric-handle: 106269577-1741874145416
electric-schema: {"id":{"type":"uuid","not_null":true,"pk_index":0},"val":{"type":"jsonb"}}
location: /v1/shape?handle=106269577-1741874145416&offset=-1&table=tab

[
    {
        "headers": {
            "control": "must-refetch"
        }
    }
]
```

If I send a request with the new shape handle but with `offset=0_0&live=true`, the request succeeds and waits for changes. Inserting a new row or modifying an existing one will again result in retirement of the shape handle, termination of the shape consumer, but the request will not be interrupted and will eventually receive a 204 response from the server after the live request timeout expires.

## Expected behaviour

The shape shouldn't get recycled on every change to the table.

Live request should be interrupted when the underlying shape consumer terminates.

## Notes

This behaviour only manifests itself when the table has a generated column. With regular columns PG does not preface every row change with a relation message and therefore no shape recycling occurs.

icehaunter:
Closes #2448 

Generated columns cannot be included in a publication by PostgreSQL, so we disallow them in shapes. When doing `CREATE PUBLICATION p FOR TABLE t` without a column list, generated columns are omitted. When doing an explicit `FOR TABLE t (id)` (if `id` is generated), then it says `ERROR:  cannot use generated column "id" in publication column list`. 

We could, later, allow generated columns based on the subset of where clause functions we support by filling the stored values on our side too, but that's a way off.

### DIFF
--- a/.changeset/forty-buttons-accept.md
+++ b/.changeset/forty-buttons-accept.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: disallow generated columns in shapes
--- a/packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
+++ b/packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
@@ -88,6 +88,7 @@ defmodule Electric.Postgres.Inspector.DirectInspector do
       attndims as array_dimensions,
       atttypmod as type_mod,
       attnotnull as not_null,
+      attgenerated != '' as is_generated,
       pg_type.typname as type,
       pg_type.typtype as type_kind, -- e.g. an enum is kind 'e'
       elem_pg_type.typname as array_type, -- type of the element inside the array or nil if it's not an array
--- a/packages/sync-service/lib/electric/shapes/shape.ex
+++ b/packages/sync-service/lib/electric/shapes/shape.ex
@@ -174,13 +174,18 @@ defmodule Electric.Shapes.Shape do
           [String.t(), ...] | nil
         ) ::
           {:ok, [String.t(), ...] | nil} | {:error, {:columns, [String.t()]}}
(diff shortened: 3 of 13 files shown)


robacourt (approved):
Nice!

PR #2510, opened 2025-04-01
fix: refetch schema info when seeing a relation

### LINKED ISSUE #2444: Column changing nullability isn't reflected in the shape schema header
Reported @ https://discord.com/channels/933657521581858818/933657523049885698/1349496269311180984

> does electric cache schema definitons? i had a column that used to be notnull that i changed to nullable ( i checked rds and it seems to have done the change ) but when i use useShape i get the error that column cant be null

They got this error `ParserNullValueError: Column "age" does not allow NULL values` despite the column now being nullable.

The error comes from https://github.com/electric-sql/electric/blob/eb8167a1d180ba010ff58511ac85c89cf3019f1d/packages/typescript-client/src/parser.ts#L171

icehaunter:
Closes #2444

There will be a follow-up to refetch a schema even in absence of updates

### DIFF
--- a/.changeset/rich-glasses-run.md
+++ b/.changeset/rich-glasses-run.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: refetch schema from DB when seeing a relation message to get recent info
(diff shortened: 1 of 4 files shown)


robacourt (approved):
Nice

PR #2512, opened 2025-04-01
[Fly.io docs] set `DATABASE_URL` as a secret

kylemclaren:
It's not recommended to set sensitive values in the fly.toml env section.

### DIFF
--- a/website/docs/integrations/fly.md
+++ b/website/docs/integrations/fly.md
@@ -33,7 +33,7 @@ Fly's [Supabase Postgres](https://fly.io/docs/supabase) is a managed Postgres se
 
 ### Deploy Electric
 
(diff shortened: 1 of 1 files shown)


alco:
Hey @kylemclaren! Thanks for the suggestion.

We agree in principle, but it is important for us as developers of Electric that users have as few hurdles as possible on the path of getting Electric up and running on their preferred platform. Without the `DATABASE_URL` setting in place, Electric won't start and its healthcheck endpoint will keep responding with errors, so it won't be clear for the user whether they did something wrong and how to go about fixing it.

Of course in a production setting sensitive values like DATABASE_URL and ELECTRIC_SECRET must be defined as secrets. We are working on the assumption that someone deploying Electric to Fly is aware of that.

PR #2514, opened 2025-04-01
feat(sync-service): Add LRU shape expiry

robacourt:
With this PR, once the number of shapes exceeds `max_shapes` shapes are expired, least recently used first.

`max_shapes` is currently fixed and specified with the new `ELECTRIC_EXPERIMENTAL_MAX_SHAPES` environment variable. The env var has `EXPERIMENTAL` in it's name to convey that we will not be supporting this variable in the future since we would prefer that `max_shapes` was dynamically calculated. However for trigger.dev we can see setting max_shapes to 8000 should prevent WAL lag and shapes that have not been used for about two hours are expired which suits their needs.

A shapes `last_read` time is not currently persisted to disk. That has been left to a later PR. For trigger.dev this is fine since they do not currently recover shapes on restart. In the event that shape recovery was used with this implementation of cache expiry, shapes not used since restoration would have a `last_read` time of the time of recovery and so maybe be expired in an unexpected order.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -312,7 +312,7 @@ defmodule Electric.ShapeCache do
             shape_handle: shape.shape_handle,
             max_shapes: max_shapes,
             shape_count: shape_count,
-            elapsed_ms_since_use: shape.elapsed_ms_since_use
+            elapsed_minutes_since_use: shape.elapsed_minutes_since_use
           ],
           fn ->
             Logger.info(
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -156,6 +156,7 @@ defmodule Electric.ShapeCache do
 
       true ->
         server = Electric.Shapes.Consumer.name(stack_id, shape_handle)
+        send(self(), :maybe_expire_shapes)
         GenServer.call(server, :await_snapshot_start, 15_000)
     end
   end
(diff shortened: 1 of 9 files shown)


robacourt:
benchmark this

robacourt:
Since expiry is triggered by shape creation, if there was performance impact we'd expect to see it in shape creation. The effect seems to be minimal though.

For v1.0.2:
Screenshot 2025-04-02 at 09 37 04

With this PR and `max_shapes` set to 10:
Screenshot 2025-04-02 at 09 36 35

Perhaps there is a little difference, it's tricky to tell what's statistically significant. In any case I feel at this stage it's performant enough, and as we develop the cache expiry further we can optimise it further.

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -144,6 +146,10 @@ defmodule Electric.ShapeCache do
     shape_status = Access.get(opts, :shape_status, ShapeStatus)
     stack_id = Access.fetch!(opts, :stack_id)
 
+    shape_status.update_last_read_time_to_now(table, shape_handle)
+    server = Access.get(opts, :server, name(opts))
+    GenServer.cast(server, :maybe_expire_shapes)
```
icehaunter, line 151: I feel like adding this cost (not inline, I get it, we don't wait) to every single request that accesses any part of any shape which has a memory copy + sort under the hood is suboptimal. I'd rather expiration be periodic tbh

icehaunter (approved):
LGTM

PR #2516, opened 2025-04-02
Improve retry behaviour for Elixir client

magnetised:
Replaces retry count with a wall-clock based timeout, with a default of 5 minutes. 

Also add ability to add static headers to every request.

Document fetch options and validate them when creating the client.


packages/elixir-client/lib/electric/client/fetch/http.ex
```diff
@@ -53,4 +215,27 @@ defmodule Electric.Client.Fetch.HTTP do
   defp jitter() do
     1 - 0.1 * :rand.uniform()
   end
+
+  defp now, do: System.monotonic_time(:second)
+
+  defp transient?(%Req.Response{status: status}) when status in [408, 429, 500, 502, 503, 504] do
```
msfstef, line 221: any reason not to include all `>=500` responses? in the issue we ourselves ran into it's a 530 error from Cloudflare so this would not retry on that, and there isn't really an option to set this?

packages/elixir-client/lib/electric/client/fetch/http.ex
```diff
@@ -44,6 +120,92 @@ defmodule Electric.Client.Fetch.HTTP do
...
+            ") ",
+            Exception.message(exception)
+          ]
+
+        response ->
+          ["retry: got response with status #{response.status}, "]
+      end
+    end)
+
+    Logger.warning(fn ->
```
msfstef, line 192: would it make sense to provide another config option for handling these errors? or generally be able to provide a fully custom retry function if someone wants to (e.g. see the issue with over-specifying _which_ 5xx errors are transient)

msfstef (approved):
Looks good, although my understanding is that this will still not do any retries for e.g. the 530 errors that triggered this change in the first place as the 5xx errors are overspecified? Perhaps we could do `>=500` and exclude known ones, or at least provide some way to easily configure that if you still prefer a more conservative default.

packages/elixir-client/lib/electric/client/fetch/http.ex
magnetised: yeah. maybe the ability to pass a custom function for the "is transient?" test for responses. good idea. hadn't picked up the 530 error

msfstef: yeah I think a custom `is_transient?` method might do the trick

msfstef (approved):
Looks good! Would appreciate a test or two around the custom transient options given it's a fresh public API

magnetised:
> Looks good! Would appreciate a test or two around the custom transient options given it's a fresh public API

yeah, actually because of the way it's wired, the current retry behaviour tests do actually use the public api, but I'll make it explicit

PR #2521, opened 2025-04-03
chore: Rename api_secret query param to secret

kevin-dp:
Fixes https://github.com/orgs/electric-sql/projects/13/views/1?pane=issue&itemId=104307653&issue=electric-sql%7Celectric%7C2475

Accompanying docs PR: https://github.com/electric-sql/electric/pull/2547

### DIFF
--- a/integration-tests/tests/secure-mode.lux
+++ b/integration-tests/tests/secure-mode.lux
@@ -51,11 +51,11 @@
   ??HTTP/1.1 401 Unauthorized
 
   # rejected if secret is wrong
-  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&api_secret=wrong-secret"
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&secret=wrong-secret"
   ??HTTP/1.1 401 Unauthorized
 
   # accepted if secret is provided and correct
-  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&api_secret=test-secret"
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&secret=test-secret"
   ??HTTP/1.1 200 OK
 
 [cleanup]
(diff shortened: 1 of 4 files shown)


msfstef (approved):
Looks good!

packages/sync-service/lib/electric/plug/router.ex
```diff
@@ -60,7 +60,7 @@ defmodule Electric.Plug.Router do
     else
       conn = conn |> fetch_query_params()
 
-      case conn.query_params["api_secret"] do
+      case conn.query_params["secret"] do
```
KyleAMathews, line 63: we do want to keep backwards compatibility. So keep also looking for `api_secret` but mark it as deprecated and that we'll remove it in v2.

KyleAMathews (changes requested):

packages/sync-service/lib/electric/plug/router.ex
kevin-dp: Ok, i modified the router to fall back to `api_secret` if `secret` is not provided.

integration-tests/tests/secure-mode.lux
```diff
@@ -51,11 +51,11 @@
   ??HTTP/1.1 401 Unauthorized
 
   # rejected if secret is wrong
-  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&api_secret=wrong-secret"
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&secret=wrong-secret"
   ??HTTP/1.1 401 Unauthorized
 
   # accepted if secret is provided and correct
-  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&api_secret=test-secret"
+  !curl -v -X GET "http://localhost:3000/v1/shape?table=items&offset=-1&secret=test-secret"
```
KyleAMathews, line 58: I'd add a test for api_secret to make sure we don't inadvertently break it (until v2).

KyleAMathews (approved):

integration-tests/tests/secure-mode.lux
kevin-dp: I did add a unit test for this to the packages/sync-service/test/electric/plug/router_test.exs

KyleAMathews: 👍

PR #2526, opened 2025-04-03
fix: better error handling for timing issues when accessing PG

### LINKED ISSUE #2502: Unformatted 500 if altering table for a snapshot is too slow
https://github.com/electric-sql/electric/blob/844a54f74139be268799049646e47130a37aa153/packages/sync-service/lib/electric/shape_cache.ex#L153

This call exits on a timeout, causing HTTP process to exit ungraciously 

--- 

Found with Antithesis under network latency between Electric and PG

### LINKED ISSUE #2503: Unformatted 500 when getting information about a table in Postgres is too slow
https://github.com/electric-sql/electric/blob/main/packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex#L59

This call exits on a timeout, causing HTTP process to exit ungraciously

--- 

Found with Antithesis under network latency between Electric and PG

### LINKED ISSUE #2505: Output stream is cut after sending 200 if file creation takes too long
We're returning a stream that doesn't open the file until read, sending a 200 chunked, and then start reading the file. If file never appears, we then cut the response unexpectedly. We can do better by only sending a 200 once the snapshot creation actually starts (i.e. we get the first row). It slightly slows down time to first byte, but makes errors more explicit

---

Found with Antithesis under heavy latency between Electric and Postgres

icehaunter:
- Catch the exits when altering tables to account for GenServer timeouts as well as DBConnection raises:
  - Closes #2502 
  - Closes #2503 
- Make sure to mark snapshot as started only after getting first row from PG to avoid race condition where the file wasn't created yet
  - Closes #2505

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -161,6 +161,18 @@ defmodule Electric.Plug.ServeShapePlug do
   end
 
   @impl Plug.ErrorHandler
+  def handle_errors(conn, %{kind: :error, reason: exception, stack: stack})
+      when is_exception(exception, DBConnection.ConnectionError) do
+    OpenTelemetry.record_exception(:error, exception, stack)
+
+    error_str = Exception.format(:error, exception)
+
+    conn
+    |> fetch_query_params()
+    |> assign(:error_str, error_str)
+    |> send_resp(503, Jason.encode!(%{error: "Database is unreachable"}))
+  end
+
   def handle_errors(conn, error) do
     OpenTelemetry.record_exception(error.kind, error.reason, error.stack)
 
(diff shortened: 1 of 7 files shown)


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -11,17 +11,17 @@ defmodule Electric.Plug.ServeShapePlug do
 
   require Logger
 
-  plug :fetch_query_params
+  plug(:fetch_query_params)
```
robacourt, line 14: What's happened to your formatter?

robacourt (approved):
Looks great! Just left a tiny nitpick

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
icehaunter: /shrug

PR #2532, opened 2025-04-03
fix: Maintain backwards compatibility with 204s

msfstef:
Closes https://github.com/electric-sql/electric/issues/2538

Maintain backwards compatibility after https://github.com/electric-sql/electric/pull/2486

The change was forwards compatible (old clients work with new servers) but not backwards compatible (new clients didn't work with old servers - old enough that they wouldn't return a body with the 204).

NOTE: The change was actually both forwards and backwards compatible as of the official 1.0.0 release, as we were already (wrongly) including a body with 204 responses at that point. I think that we should not reinsert this for the sake of backwards compatibility for versions prior to the official 1.0.0.

### DIFF
--- a/.changeset/chatty-pens-notice.md
+++ b/.changeset/chatty-pens-notice.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
+Maintain backwards compatibility of client for 204 responses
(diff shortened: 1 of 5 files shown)


KyleAMathews (approved):

KyleAMathews:
nice work! Love all the new tests & good job tracking down all the no-body codes — the http world is big and weird so it's good to lock things down as much as possible.

PR #2546, opened 2025-04-06
WIP feat(client): experimental support for streaming SSE in live mode

samwillis:
Probably needs some tests

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -39,6 +39,7 @@ import {
   TABLE_QUERY_PARAM,
   REPLICA_PARAM,
   FORCE_DISCONNECT_AND_REFRESH,
+  EXPERIMENTAL_LIVE_SSE_QUERY_PARAM,
 } from './constants'
 
 const RESERVED_PARAMS: Set<ReservedParamKeys> = new Set([
@@ -244,6 +245,11 @@ export interface ShapeStreamOptions<T = never> {
    */
   subscribe?: boolean
 
+  /**
+   * Experimental support for Server-Sent Events (SSE) for live updates.
+   */
+  experimentalLiveSse?: boolean
+
   signal?: AbortSignal
   fetchClient?: typeof fetch
   backoffOptions?: BackoffOptions
@@ -281,8 +287,9 @@ export interface ShapeStreamInterface<T extends Row<unknown> = Row> {
 }
 
 /**
- * Reads updates to a shape from Electric using HTTP requests and long polling. Notifies subscribers
- * when new messages come in. Doesn't maintain any history of the
+ * Reads updates to a shape from Electric using HTTP requests and long polling or
+ * Server-Sent Events (SSE).
+ * Notifies subscribers when new messages come in. Doesn't maintain any history of the
  * log but does keep track of the offset position and is the best way
  * to consume the HTTP `GET /v1/shape` api.
  *
@@ -294,6 +301,14 @@ export interface ShapeStreamInterface<T extends Row<unknown> = Row> {
  * const stream = new ShapeStream(options)
  * stream.subscribe(messages => {
  *   // messages is 1 or more row updates
+ * })
+ * ```
+ *
+ * To use Server-Sent Events (SSE) for real-time updates:
+ * ```
+ * const stream = new ShapeStream({
+ *   url: `http://localhost:3000/v1/shape`,
+ *   liveMode: 'sse'
  * })
  * ```
  *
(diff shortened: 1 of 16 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -714,6 +749,108 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+                const lastMessage = batch[batch.length - 1]
+                if (isUpToDateMessage(lastMessage)) {
+                  const upToDateMsg = lastMessage as typeof lastMessage & {
+                    headers: { global_last_seen_lsn: string }
+                  }
+                  this.#lastSyncedAt = Date.now()
+                  this.#isUpToDate = true
+                  this.#lastOffset =
+                    `${upToDateMsg.headers.global_last_seen_lsn}_0` as Offset
+                  // TODO: we also need the cache buster `cursor` value
```
thruflo, line 811: Is it on the live SSE response as a header? I’m thinking that’s fine, as in, the client only needs to update the cursor when the response is empty, so it’s fine to get it from the header.

thruflo:
I think we did want to keep using SSE (just with the slightly different client behaviour) rather than switching to an alternative server response protocol, because using SSE proper triggers the desired request collapsing behaviour from CDNs.

balegas:
Did you check that keeping the regular http response open for a bit would mess with the CDN request collapsing?

kevin-dp (changes requested):
Looking into the implementation it seems that it does not properly handle corner cases such as 409. Any non 200 status code would basically be thrown whereas in our long polling implementation we handle those status codes appropriately.

Also this currently only works in the browser in other environments like NodeJS the EventSource API is not defined and we have to polyfill it. However, as pointed out by @samwillis the browser's event source API is very limited and we would want to use the [fetch-event-source library](https://github.com/Azure/fetch-event-source) instead. That will solve the polyfill issue.

As a final remark, i don't like the complexity of the `ShapeStream.start` method. It was already complex and is now becoming even more complex. I have ideas to refactor this code to make it more comprehensible.

KyleAMathews:
Shipped in https://github.com/electric-sql/electric/pull/2776

thruflo:
One of the nuances here that I was thinking about recently is how the SSE client handles interruptibility.

I.e.: we have the ability for the backend to interrupt a long poll request. But for an SSE request, if it's interrupted, how does the SSE client handle this and is there potential for some kind of duplicate message delivery? Or is the SSE client advancing its offset with every message it receives?

kevin-dp:
> But for an SSE request, if it's interrupted, how does the SSE client handle this and is there potential for some kind of duplicate message delivery?

The SSE approach should behave the same way as the long poll when the frontend or the backend interrupt the request. When that happens, the fetch (note that both long poll and SSE use fetch) is aborted and an error is thrown. That error is caught here: https://github.com/electric-sql/electric/blob/572a93587cda94745a8b7865d9eb515e7035ca97/packages/typescript-client/src/client.ts#L444-L460
We will continue to the next iteration of the loop which will try resuming from where we left it.

>  Or is the SSE client advancing its offset with every message it receives?

Yes, each time we receive a message it is processed by this method:

```ts
async #onMessages(messages: string, schema: Schema, isSseMessage = false) {
    const batch = this.#messageParser.parse(messages, schema)

    // Update isUpToDate
    if (batch.length > 0) {
      const lastMessage = batch[batch.length - 1]
      if (isUpToDateMessage(lastMessage)) {
        if (isSseMessage) {
          // Only use the offset from the up-to-date message if this was an SSE message.
          // If we would use this offset from a regular fetch, then it will be wrong
          // and we will get an "offset is out of bounds for this shape" error
          const offset = getOffset(lastMessage)
          if (offset) {
            this.#lastOffset = offset
          }
        }
        this.#lastSyncedAt = Date.now()
        this.#isUpToDate = true
      }

      await this.#publish(batch)
    }
  }
```
So we're using the offset that is provided in the up-to-date message to advance the offset.

thruflo:
@kevin-dp my concern is:

- accumulate some messages (no uptodate)
- error, reconnect
- accumulate the same messages again
- uptodate -> commit with duplicate data

Seeing the message processing code, I think this can be a possibility, if the SSE client maintains the accumulated messages when handling an interrupt.

Maybe this isn’t an issue because the client clears the accumulator each time it re-connects on error? If so, it’s worth knowing that it’s important to interrupt with an error.

I had considered a 200 response with @magnetised for the Phoenix.Sync interrupt. We’ve moved past and we’re not doing it any more but for a while it could have snuck in there!

kevin-dp:
@thruflo right, that's a valid concern. So as you described, if we receive an update and then before we get the up to date message the connection breaks, we will reconnect and start from the old offset and get the update a second time. So the shapestream consumer will receive the update twice (or more, depending on how many times this happens). This happens because we immediately publish every incoming message. So, if the connection breaks before it gets uptodate, we will try again with the same offset and get some updates we already got before. Even though this is not 100% right, i think it's fine in most cases because the consumer would usually just materialize the stream into a map and applying an insert/update/delete twice or more is idempotent. But if the consumer does any additional side effects like counting the number of updates then it will be wrong.

We can solve this problem by changing the SSE `onmessage` callback such that it buffers messages until we get the `uptodate` message. That way we process only complete batches of operations like we do in regular mode (by virtue of Electric sending out complete batches that end with an up to date message).

KyleAMathews:
Yeah we should keep the same semantics so buffering until up-to-date makes sense

PR #2547, opened 2025-04-07
docs: Mark api_secret as deprecated in OpenAPI spec

kevin-dp:
Introduces the `secret` query param in favor of `api_secret` which is deprecated and will be removed in v2.


KyleAMathews (approved):
:shipit:

kevin-dp:
Should only ship it once we release the PR that introduces the `secret` param though.

PR #2549, opened 2025-04-07
chore: Trigger Cloud workflow for updating Electric on release

kevin-dp:
This PR extends the changesets workflow such that we trigger the Cloud workflow for updating the version of Electric that is used by the cloud.

### DIFF
--- a/.github/workflows/changesets_release.yml
+++ b/.github/workflows/changesets_release.yml
@@ -43,3 +43,23 @@ jobs:
         run: node scripts/tag-latest.mjs
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      # Get the Electric version of the Docker image
+      - name: Get sync-service version
+        if: steps.changesets.outputs.published == 'true'
+        id: sync_version
+        run: |
+          VERSION=$(jq -r '.version' packages/sync-service/package.json)
+          echo "version=${VERSION}" >> $GITHUB_OUTPUT
+      # Trigger cloud update workflow
+      - name: Trigger cloud update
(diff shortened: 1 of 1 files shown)


.github/workflows/changesets_release.yml
```diff
@@ -43,3 +43,23 @@ jobs:
         run: node scripts/tag-latest.mjs
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      # Get the Electric version of the Docker image
+      - name: Get sync-service version
+        if: steps.changesets.outputs.published == 'true'
+        id: sync_version
+        run: |
+          VERSION=$(jq -r '.version' packages/sync-service/package.json)
```
msfstef, line 51: I guess the unfortunate thing is we can't easily pick up the canary docker SHA for the admin electric - but perhaps it's better to keep it on stable releases

msfstef (approved):
Looks good! I would prefer this as a separate job within the GH workflow that [depends on the `changesets` job](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-jobs-in-a-workflow#example-requiring-successful-dependent-jobs) - so that if changesets are released and the second part fails that is clear and can be retried etc though

.github/workflows/changesets_release.yml
kevin-dp: Indeed, it's a bit unfortunate but we'll also be able to trigger the workflow from the GH UI. So if we explicitly want to adopt a canary release, we could manually trigger the workflow with the hash. I'll make sure it supports that.

PR #2554, opened 2025-04-07
Make password obfuscation more implicit

alco:
In the spirit of the changes made in
https://github.com/electric-sql/electric/pull/2375, this change goes further in lifting the requirement of explicitly calling the obfuscate_password() function.

The calls we still had in config/runtime.exs were redundant because StackSupervisor performs the obfuscation anyway.


msfstef (approved):
Thank you for reviewing this ! This was the reason I brought it up, as I remembered we were obfuscating in multiple places

nit: needs changeset

Issue #2552, opened 2025-04-07
Handle database does not exist errors


kevin-dp:
Sentry has some issues related to some sources whose database no longer exists. When the database no longer exists Postgrex throws the following error:
```json
{
  "application": "db_connection",
  "crash_reason": "{%Postgrex.Error{message: nil, postgres: %{code: :internal_error, message: \\"database \\\\\\"write-patterns-pr-2479\\\\\\" does not exist\\", severity: \\"ERROR\\", pg_code: \\"XX000\\"}, connection_id: nil, query: nil}, []}",
  "domain": [
    "elixir"
  ],
  "file": "lib/db_connection/connection.ex",
  "gl": "#PID<0.2886.0>",
  "line": 107,
  "mfa": "{DBConnection.Connection, :handle_event, 4}",
  "pid": "#PID<0.675873.0>",
  "time": 1744020828402441
}
```

I have also seen this happening to one of my own sources:

![Image](https://github.com/user-attachments/assets/2b0fa81e-d92a-47c7-a8cb-84d4510795f9)
![Image](https://github.com/user-attachments/assets/0bf5f628-4505-43a5-9955-dab8060a5c47)

Should we automatically delete a source if its associated database no longer exist?
However, that means it may automatically disappear from the user's dashboard which may be unexpected (cc @KyleAMathews).

kevin-dp:
It seems that the pattern where this occurs is always the same. The source for the Sentry issue is: https://stratovolt.electric-sql.cloud/sources/a94167d1-e2c7-46ef-8771-0555664f5626

It follows the same pattern as the screenshots above. The DB is deleted and Electric looses connection with the DB.
Electric then:
- tries to reconnect to the DB a number of times, all reconnections fail
- Electric then tries to clean up the tenant
- Cleanup fails with "database does not exist" error

kevin-dp:
Note that when this happens to an Electric cloud user, they will see that there source is in "error" state but they can't delete the source because we do not provide that functionality yet.

kevin-dp:
Since we cannot recover from this error, we may be able to fix this simply by recognizing it as a fatal error in `stop_if_fatal_error` which currently only recognizes slot invalidation errors:

```ex
defp stop_if_fatal_error(
         %Postgrex.Error{
           postgres: %{
             code: :object_not_in_prerequisite_state,
             detail: "This slot has been invalidated" <> _,
             pg_code: "55000"
           }
         } = error,
         state
       ) do
[...]

PR #2561, opened 2025-04-08
fix: Handle db does not exist error as a fatal error

kevin-dp:
This PR handles the Postgres "database does not exist" error which can occur when a user deletes their database. This is a fatal error we can't recover from so we now also handle it like that. I modified the shutdown method to log a warning rather than an error since the shutdown is not unexpected since we are shutting it down on purpose as a response to the DB no longer existing. This will also avoid that this error pops up in Sentry for our cloud.

Fixes https://github.com/electric-sql/electric/issues/2552

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -720,18 +720,7 @@ defmodule Electric.Connection.Manager do
          } = error,
          state
        ) do
-    Electric.StackSupervisor.dispatch_stack_event(
-      state.stack_events_registry,
-      state.stack_id,
-      {:database_slot_invalidated, %{error: error}}
-    )
-
-    # Perform supervisor shutdown in a task to avoid a circular dependency where the manager
-    # process is waiting for the supervisor to shut down its children, one of which is the
-    # manager process itself.
-    Task.start(Electric.Connection.Supervisor, :shutdown, [state.stack_id, error])
-
-    {:noreply, state}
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -734,6 +734,31 @@ defmodule Electric.Connection.Manager do
...
+  defp stop_if_fatal_error(
+         %Postgrex.Error{
+           postgres: %{
+             code: :internal_error,
+             pg_code: "XX000"
+           }
+         } = error,
+         state
+       ) do
+    if Regex.match?(~r/database ".*" does not exist$/, error.postgres.message) do
```
robacourt, line 746: If this doesn't match, should it still be fatal?

kevin-dp: Oh god, great catch!!

robacourt (approved):
Nice!

msfstef (approved):
Looks good!

Would love a test around this but we don't currently have the connection manager easily set up for integration tests - perhaps in router tests or something?

alco (approved):
Nice!

Let's not copy-paste code that can be encapsulated in a small helper function though.

kevin-dp:
> Nice!
> 
> Let's not copy-paste code that can be encapsulated in a small helper function though.

Thanks. I wanted to extract it into a function but wasn't sure if it was worth it yet.

PR #2566, opened 2025-04-08
Tudor/upgrade path pglite

tdrz:
DO NOT MERGE!

This is an example of how PGlite data can be upgraded via `pg_dump`. See `Migrate` option bottom left corner.


icehaunter:
Hey @tdrz can this be closed?

PR #2570, opened 2025-04-08
Add note to website about pricing for electric cloud

KyleAMathews:


website/product/cloud.md
```diff
@@ -142,6 +142,10 @@ export async function GET(req: Request) {
 }
 ```
 
+### Pricing
+
+Electric Cloud is free while we work on pricing. We'll be launching pricing in the next month or two (May/June 2025). There'll always be generous free tier. If you plan to use Electric in a larger app (1000+ MAU) please reach out to make sure we can fully support you & to get a sense of what pricing will be like.
```
kevin-dp, line 147: you're missing "a" in:
> There'll always be generous free tier.

Issue #2574, opened 2025-04-09
Failed to start shape supervisor: {{:shutdown, {:failed_to_start_child, Electric.ShapeCache, :timeout}}


magnetised:
https://electricsql-04.sentry.io/issues/36559224/events/4ed2e6e0ba8b45b799f61c9df32cdfa8/

`ShapeCache.start_link/1` is timing out probably due to delays loading existing shapes

PR #2575, opened 2025-04-09
set shape cache restore timeout per-shape, not globally

magnetised:
Currently we have a hard-coded timeout of 20s for `ShapeCache.init/1` to complete. Since we're starting all existing shape consumers within this callback 20s may not be long enough if there are enough shapes.

Rather than limit the duration of the entire shape recovery process, this PR wraps each shape-restart in a timeout (currently 5s) and leaves the (default) timeout for `init/1` as `:infinity`.

Currently, in the spirit of the other error handling in this bit of code, shapes that timeout are deleted rather than the entire shape cache init failing. I'm not sure this is the right approach TBH -- some critical problem in the storage layer or elsewhere could result in all shapes being removed by this timeout and the system starting "normally" but with some fundamental problem at the storage or pg layer.

Options are:

- as currently just drop shapes that timeout (with an error in the logs)
- allow some number of failures but raise after that
- crash with any timeout

Feedback welcome.

Fixes https://github.com/electric-sql/electric/issues/2574

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -243,7 +243,6 @@ defmodule Electric.ShapeCache do
         reraise error, __STACKTRACE__
     catch
       :exit, reason ->
-        dbg(reason)
         clean_up_all_shapes(state)
         exit(reason)
     end


packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -242,6 +243,7 @@ defmodule Electric.ShapeCache do
         reraise error, __STACKTRACE__
     catch
       :exit, reason ->
+        dbg(reason)
```
msfstef, line 246: nit: remove

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -375,34 +375,61 @@ defmodule Electric.ShapeCache do
...
+  # the shape cache loads existing shapes within its init/1 callback
+  # which is useful because we know that when the start_link call on the
+  # stack completes the system is fully booted and all shape consumers are
+  # running.
+  #
+  # rather than set a global timeout for the `init/1` function we leave the
+  # `start_link/1` timeout as `:infinity` and instead have a timeout for every
+  # shape
+  defp start_shape_with_timeout(shape_handle, shape, state, timeout) do
+    task = Task.async(fn -> start_and_recover_shape(shape_handle, shape, state) end)
```
msfstef, line 396: I always want to double check - are these async tasks linked, and would an exit from them make the `ShapeCache` fall over? Or are we handling exits implicitly via the synchronous `Task.yield` call further down?

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -90,10 +94,7 @@ defmodule Electric.ShapeCache do
       GenServer.start_link(
         __MODULE__,
         Map.new(opts) |> Map.put(:db_pool, db_pool) |> Map.put(:name, name),
-        name: name,
-        # allow extra time for shape cache to initialise as it needs to
-        # restore existing shapes
-        timeout: Keyword.get(opts, :timeout, 20_000)
```
msfstef, line 96: so this will go on forever? is there no default/implicit timeout in the process initialization `start_link`? I think the initial reason I had added this was because that's what I had observed

msfstef (approved):
Looks good - also unsure if we should throw or fail a shape but I think our design philosophy has been going towards aggressive clearing of the cache, so that's better than relying on people deploying it to have proper orchestration etc.

I have left a comment for clarification about the timeout of `start_link`

packages/sync-service/lib/electric/shape_cache.ex
magnetised: There's a default timeout of `:infinity` for the `start_link` call. it will go on forever, hence the monitoring of the time taken to start each shape consumer.

packages/sync-service/lib/electric/shape_cache.ex
magnetised: They are linked: https://hexdocs.pm/elixir/Task.html#async/1-linking

Since we catch errors in the `start_and_recover_shape` function a crash when starting the shape consumer won't crash the shapecache - exactly as before.

Issue #2492, opened 2025-03-27
Explicitly set needed headers for Access-Control-Expose-Headers


KyleAMathews:
Currently we return a `*` from the Electric server for the value of this header — this isn't ideal though as `*` isn't allowed (or is just ignored) in a few scenarios e.g. if the client sends custom headers or cookies (`credentials: 'include'` for fetch).

This has the unpleasant affect that CORS just works with Electric... until it doesn't. Which is generally when people go from testing to using it for real as that's when they add auth of some sort and stuff breaks.

The fix for sending credentials/custom headers (like `authorization`) is to explicitly set which headers are allowed in `Access-Control-Expose-Headers`. We suggest people do that themselves but there's no reason Electric can't just always return the list as we of course know what the list is.

The list of headers we need to expose are:
- `electric-cursor`
- `electric-handle`
- `electric-offset`
- `electric-schema`
- `electric-up-to-date`

PR #2579, opened 2025-04-09
Make explicit the headers that clients need to be able read

KyleAMathews:
Fix https://github.com/electric-sql/electric/issues/2492

### DIFF
--- a/packages/sync-service/lib/electric/plug/utils.ex
+++ b/packages/sync-service/lib/electric/plug/utils.ex
@@ -157,10 +157,7 @@ defmodule Electric.Plug.Utils do
       do:
         conn
         |> put_resp_header("access-control-allow-origin", get_allowed_origin(conn, opts))
-        |> put_resp_header(
-          "access-control-expose-headers",
-          Electric.Shapes.Api.Response.access_control_expose_headers()
-        )
+        |> put_resp_header("access-control-expose-headers", headers_to_expose())
         |> put_resp_header("access-control-allow-methods", get_allowed_methods(conn, opts))
 
     defp get_allowed_methods(_conn, opts), do: Access.get(opts, :methods, []) |> Enum.join(", ")
@@ -174,6 +171,10 @@ defmodule Electric.Plug.Utils do
           [] -> "*"
         end
       )
+    end
+
+    defp headers_to_expose do
+      Enum.join(Electric.Shapes.Api.Response.electric_headers(), ",")
     end
   end
 
--- a/packages/sync-service/lib/electric/plug/utils.ex
+++ b/packages/sync-service/lib/electric/plug/utils.ex
@@ -159,7 +159,7 @@ defmodule Electric.Plug.Utils do
         |> put_resp_header("access-control-allow-origin", get_allowed_origin(conn, opts))
         |> put_resp_header(
           "access-control-expose-headers",
-          "electric-cursor, electric-handle, electric-offset, electric-schema, electric-up-to-date"
+          Electric.Shapes.Api.Response.access_control_expose_headers()
         )
         |> put_resp_header("access-control-allow-methods", get_allowed_methods(conn, opts))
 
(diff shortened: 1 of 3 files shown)


thruflo:
Maybe worth adding a test?

KyleAMathews:
We can assert against the headers but other than that it'd need a test w/ a real browser which we don't have atm. So probably too much work for that.

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -157,7 +157,10 @@ defmodule Electric.Plug.Utils do
       do:
         conn
         |> put_resp_header("access-control-allow-origin", get_allowed_origin(conn, opts))
-        |> put_resp_header("access-control-expose-headers", "*")
+        |> put_resp_header(
+          "access-control-expose-headers",
+          "electric-cursor, electric-handle, electric-offset, electric-schema, electric-up-to-date"
```
msfstef, line 162: The names of all of these headers are defined in `Electric.Shapes.Api.Response` - I'd avoid duplicating them for the risks of changing something, adding one more header, etc and forgetting it.

I would define all of the names explicitly in `Electric.Shapes.Api.Response` (perhaps as module attributes) and have a method `put_access_control_expose_headers()` to respect the same pattern we use for all headers including the cache.

msfstef (approved):
Looks good but I would prefer if we followed the existing pattern and put this logic in `Electric.Shapes.Api.Response` and avoid string duplication - and I agree that a simple test checking that this header is included in the serve shape plug tests would be nice

packages/sync-service/lib/electric/plug/utils.ex
KyleAMathews: @msfstef windsurf appreciated the feedback and I pushed its new work!

packages/sync-service/lib/electric/shapes/api/response.ex
```diff
@@ -307,4 +323,12 @@ defmodule Electric.Shapes.Api.Response do
       etag
     end
   end
+
+  @doc """
+  Returns a comma-separated list of all Electric headers that need to be exposed for CORS.
+  This is used by the CORSHeaderPlug to set the Access-Control-Expose-Headers header.
+  """
+  def access_control_expose_headers do
+    Enum.join(@electric_headers_to_expose, ", ")
```
msfstef, line 332: nit: I'd name this `put_access_control_expose_headers` and use `Plug.Conn.put_resp_header` the same as the rest of this module - I'd also remove documentation as it's not idiomatic (no need to mention where it is used, and the name of the function is clear enough)

msfstef: (also the test is failing)

packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
```diff
@@ -109,6 +109,47 @@ defmodule Electric.Plug.ServeShapePlugTest do
       :ok
     end
 
+    test "sets correct CORS headers for Access-Control-Expose-Headers", ctx do
```
msfstef, line 112: nit: I don't think this test should be at the top of this test suite - there should be a more appropriate location in this test suite where we check other headers. In fact we could even have this check as part of an existing test checking headers

packages/sync-service/lib/electric/shapes/api/response.ex
alco: I disagree with the responsibility allocation: the Api.Response module deals with building responses. It is implicit in its implementation that it is the source of truth for all Electric-specific headers. So it should only make those headers available to other modules, nothing more. Than it's the PLug.Utils.CORSHeaderPlug's job to take those headers and expose them for CORS appropriately.

packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
alco: Fixed it.

alco:
@msfstef Great feedback. I have addressed your comments and fixed the test. Not going to approve this PR myself because it now includes my changes. Whenever you have a spare minute, thanks.

packages/sync-service/lib/electric/shapes/api/response.ex
msfstef: You've convinced me haha that makes more sense

msfstef (approved):
Ready to merge, thanks @alco !

PR #2580, opened 2025-04-10
fix: Replace client ID and secret with auth token for examples

msfstef:
Fixing example deployments by using auth tokens rather than client ID and secrets.


alco (approved):
:+1:

PR #2586, opened 2025-04-11
Improve proxy docs around removing content-length/content-encoding

KyleAMathews:
Tighten the code pattern & also add it to the cloud docs as well where it was missing.


thruflo (approved):
Be nice to somehow dry up these snippets.

samwillis (approved):
👍 lgtm

PR #2590, opened 2025-04-14
Return 409 immediately to live requests when shape invalidated

### LINKED ISSUE #2585: A live shape request doesn't get interrupted by shape invalidation
**Versions**

Reproducible on the latest [`main`](https://github.com/electric-sql/electric/commit/ed165ec9e0a716a49de1970efcfa443333521300) version of the server.

**Bug description**

It is a minor bug.

Create a table with one row in it, create a shape and start a live request for it:
```
=# CREATE TABLE test(payload jsonb, id uuid primary key);
CREATE TABLE

=# INSERT INTO test(payload, id)
     SELECT ('{"id": "' || id || '", "description": "..."}')::json, id FROM gen_random_uuid() id;
INSERT 0 1

$ curl -i http://localhost:3000/v1/shape\?offset=-1&table=test
HTTP/1.1 200 OK
electric-server: ElectricSQL/1.0.5
electric-handle: 109192296-1744370730054
electric-schema: {"id":{"type":"uuid","not_null":true,"pk_index":0},"payload":{"type":"jsonb"}}
electric-offset: 0_0

[{"key":"\"public\".\"test\"/\"feb3925b-143a-4f44-9fcb-89eeb864f4d1\"","value":{"id":"feb3925b-143a-4f44-9fcb-89eeb864f4d1","payload":"{\"id\": \"feb3925b-143a-4f44-9fcb-89eeb864f4d1\", \"description\": \"...\"}"},"headers":{"operation":"insert","relation":["public","test"]}}

$ curl -i http://localhost:3000/v1/shape\?offset=0_0&table=test&handle=109192296-1744370730054&live=true
```

While the live request is waiting for data, alter the table schema and insert a new row:
```
=# alter table test drop column id;
ALTER TABLE

=# INSERT INTO test(payload)
     SELECT ('{"id": "' || id || '", "description": "..."}')::json FROM gen_random_uuid() id;
INSERT 0 1
```

The Electric server detects the schema change and invalidates the shape. However, the live request doesn't get interrupted and returns an `up-to-date` response after the long-poll timeout is reached:
```
# $ curl -i http://localhost:3000/v1/shape\?offset=0_0&table=test&handle=109192296-1744370730054&live=true
HTTP/1.1 200 OK
electric-server: ElectricSQL/1.0.5
electric-cursor: 15938780
electric-handle: 109192296-1744370730054
electric-up-to-date: 
electric-offset: 0_inf

[{"headers":{"control":"up-to-date","global_last_seen_lsn":"0"}}]
```

Only the subsequent request using the same shape handle returns a `must-refetch` response:
```
$ curl -i http://localhost:3000/v1/shape\?offset=0_inf&table=test&handle=109192296-1744370730054&live=true
HTTP/1.1 409 Conflict
electric-server: ElectricSQL/1.0.5
location: /v1/shape?handle=105980062-1744370777033&offset=-1&table=test
electric-handle: 105980062-1744370777033
electric-schema: {"payload":{"type":"jsonb"}}

[{"headers":{"control":"must-refetch"}}]
```

**Expected behavior**

While the bug is minor and shape data streaming is healed after the client retries the live request using the new URL from the `location` header, there is a delay between the table in the database getting a new row inserted and the client receiving it, which can be as long as the full duration of a single live request.

magnetised:
There's no need to reply with a 200 and then require a second request to get the 409 status. we can just send the 409 immediately and save a round-trip. the two are functionally identical - a live request returns a 409, only the duration of that request changes.

Fixes #2585

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -111,7 +111,7 @@ defmodule Electric.Shapes.Consumer do
 
   def handle_call(:clean_and_stop, _from, state) do
     state =
-      reply_to_snapshot_waiters({:error, "Shape terminated before snapshot completed"}, state)
+      reply_to_snapshot_waiters(state, {:error, "Shape terminated before snapshot completed"})
 
     # TODO: ensure cleanup occurs after snapshot is done/failed/interrupted to avoid
     # any race conditions and leftover data
@@ -169,7 +169,7 @@ defmodule Electric.Shapes.Consumer do
           "Snapshot creation failed for #{shape_handle} because of:\n#{Exception.format(:error, error, stacktrace)}"
         )
 
-    state = reply_to_snapshot_waiters({:error, error}, state)
+    state = reply_to_snapshot_waiters(state, {:error, error})
     cleanup(state)
     {:stop, :normal, state}
   end
@@ -182,7 +182,7 @@ defmodule Electric.Shapes.Consumer do
 
   def terminate(reason, state) do
     state =
-      reply_to_snapshot_waiters({:error, "Shape terminated before snapshot was ready"}, state)
+      reply_to_snapshot_waiters(state, {:error, "Shape terminated before snapshot was ready"})
 
     if is_error?(reason) do
       cleanup(state)
@@ -216,12 +216,10 @@ defmodule Electric.Shapes.Consumer do
     Inspector.clean(root_table, inspector)
 
     state =
-      reply_to_snapshot_waiters(
-        {:error, "Shape relation changed before snapshot was ready"},
-        state
-      )
-
-    cleanup(state)
+      state
+      |> reply_to_snapshot_waiters({:error, "Shape relation changed before snapshot was ready"})
+      |> notify_shape_rotation()
+      |> cleanup()
 
     {:stop, :normal, state}
   end
@@ -316,7 +314,6 @@ defmodule Electric.Shapes.Consumer do
       log_state: log_state,
       chunk_bytes_threshold: chunk_bytes_threshold,
       shape_status: {shape_status, shape_status_state},
-      registry: registry,
       storage: storage
     } = state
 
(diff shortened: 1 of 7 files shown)


packages/sync-service/timeout.exs
```diff
@@ -0,0 +1,57 @@
+defmodule W do
```
alco, line 1: Did you commit this by mistake? Otherwise, a moduledoc would be useful to have here.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -216,12 +216,10 @@ defmodule Electric.Shapes.Consumer do
...
     state =
-      reply_to_snapshot_waiters(
-        {:error, "Shape relation changed before snapshot was ready"},
-        state
-      )
-
-    cleanup(state)
+      state
+      |> reply_to_snapshot_waiters({:error, "Shape relation changed before snapshot was ready"})
+      |> notify_shape_rotation()
```
alco, line 221: Was this the missing piece without which the live request would wait until the timeout runs out before returning a response?

packages/elixir-client/set_dependency_version.ex
```diff
@@ -0,0 +1,61 @@
+Mix.install([{:sourceror, "~> 1.7"}, {:jason, "~> 1.4"}])
```
alco, line 1: This change doesn't seem to belong in this PR.

alco (approved):
:speedboat:

msfstef (approved):
Looks good! Just +1-ing @alco 's comments mostly

packages/sync-service/timeout.exs
magnetised: oops!

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: yes, that's all it was

packages/elixir-client/set_dependency_version.ex
magnetised: i seem to have been a bit heavy-handed with my "add to commit" flow

Issue #2589, opened 2025-04-11
Don't revalidate old live requests


KyleAMathews:
**Versions**
- Client: Any
- Electric: latest 

**Bug description**
I was helping debug an odd problem reported by a user on Discord — when he started his app he's seeing ~100 live requests with empty body responses before it stabilizes into normal long polling.

At first I suspected it was a proxy caching the data but two things that disproved that theory was a) the response headers had a cloudflare cdn header saying the request was `REVALIDATED` and then also as he was on Electric Cloud, I could see in the backend that all the quickly repeating calls were actually arriving to the cloud electric instance.

Looking closer at the headers arriving in the backend — all the quickly resolved requests have the same `If-None-match` request header which the server immediately returned a 304 from.

The root issue is we're expecting unique cursors to serve as cache busters. But caches/proxies *can* sometimes ignore changing query parameters.

So we should also add the cursor to the etags we generate as another way of preventing caches from holding onto live requests. And also pass the etag cursor to the cursor generating function to prevent us from returning the same etag similar to https://github.com/electric-sql/electric/pull/1856

Possibly, in retrospect, that's all we need and we don't need to set the cursor in the query param 🤔 — my initial PR for reference https://github.com/electric-sql/electric/pull/1826 — something to test to possibly remove the cursor query param for v2.

**Expected behavior**
No proxy will cache or revalidate live requests.

msfstef:
Is there a reason we're adding etags and revalidating live requests at all? We could just skip the `if-none-match` check and not attach etags to live responses.

PR #2593, opened 2025-04-14
feat: Do not cache empty live responses

msfstef:
Fixes https://github.com/electric-sql/electric/issues/2589

~We introduced the concept of a cursor in https://github.com/electric-sql/electric/pull/1826 in order to avoid infinite loops of clients running into cached live responses by artificially "moving the cache forward" via the time based, coordinated cache buster.~

~However we should not need it, as our cache buster is already the `offset` parameter. We were running into this issue because we were caching live responses that do not move the cache forward, i.e. empty live responses, so clients would continuously hit the same cache over and over again.~

~Even with the "cursor" fix we still run into this issue but in a different way - empty live responses create a "chain" rather than a loop of cache hits, that can be arbitrarily long as we allow these cached live responses to be revalidated as well. This means someone who is at the tip of the log might end up following a huge chain of responses with no changes in them, and each of those requests made might trigger a separate revalidation request to the origin.~

~Request collapsing on cache misses works regardless of the cache policy you set on a response (since the CDN does not yet know what the caching policy on the response will be). This is [a canonical example](https://developers.cloudflare.com/cache/concepts/revalidation/#example-2) from Cloudflare.~

~Request collapsing on stale cache hits, as discussed [in the other Cloudflare example](https://developers.cloudflare.com/cache/concepts/revalidation/#example-1) will send back only a single revalidation out of the collapsed requests.~

~Therefore to avoid these infinite loops, we can simply _not cache_ live responses with no changes in them, which retains the request collapsing behaviour without creating any infinite loops.~

~For live responses that _do_ contain changes, we set the usual 5 second lifetime + 5 second stale lifetime, so that clients that are slightly behind can catch up using the cache, but the cursor is not needed as these cached live responses move the `offset` forward.~

~If we go ahead with this change we do need to keep the cursor header present as we require it in our official client, although I've ripped out any logic for it since it won't actually be used and it is better to not change it between requests to ensure cache consistency - we can discuss a path to deprecation~

### UPDATE

We just use an etag that is always different for live responses that contain no changes to ensure they never get revalidated.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -551,9 +551,10 @@ defmodule Electric.Shapes.Api do
     %{response: response, global_last_seen_lsn: global_last_seen_lsn} =
       update_attrs(request, %{ot_is_empty_response: true})
 
-    %{
+    %Response{
       response
       | status: 200,
+        no_change: true,
         body: encode_log(request, [up_to_date_ctl(global_last_seen_lsn)])
     }
   end
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -554,7 +554,7 @@ defmodule Electric.Shapes.Api do
     %Response{
       response
       | status: 200,
-        no_change: true,
+        no_changes: true,
         body: encode_log(request, [up_to_date_ctl(global_last_seen_lsn)])
     }
   end
--- a/packages/sync-service/lib/electric/shapes/api/response.ex
+++ b/packages/sync-service/lib/electric/shapes/api/response.ex
@@ -1,4 +1,5 @@
 defmodule Electric.Shapes.Api.Response do
+  alias Electric.Plug.Utils
   alias Electric.Shapes.Api
   alias Electric.Shapes.Shape
   alias Electric.Telemetry.OpenTelemetry
@@ -27,7 +28,7 @@ defmodule Electric.Shapes.Api.Response do
     api: %Api{},
     chunked: false,
     up_to_date: false,
-    no_change: false,
+    no_changes: false,
     params: %Api.Params{},
     status: 200,
     trace_attrs: %{},
@@ -44,7 +45,7 @@ defmodule Electric.Shapes.Api.Response do
           chunked: boolean(),
           params: Api.Params.t(),
           up_to_date: boolean(),
-          no_change: boolean(),
+          no_changes: boolean(),
           status: pos_integer(),
           trace_attrs: %{optional(atom()) => term()},
           body: Enum.t()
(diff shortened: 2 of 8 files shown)


thruflo:
Very interesting! Have you verified the request collapsing behaviour with other reverse proxies / CDNs?

msfstef:
@thruflo I've added a request collapsing test that we didn't have before to our typescript client that uses Nginx's `proxy_cache_lock` collapsing behaviour - it asserts that if you run 10 clients at tip of stream and insert a row, you get 10 requests with 9 cache hits.

I haven't run this on Cloudflare but I definitely will, although I'd like to get @KyleAMathews 's insight on how he tested this before to ensure we get the same behaviour. Ideally we get some more caching behaviour tests in our suite as well.

KyleAMathews:
The behavior for Fastly/Cloudflare that I've seen is that once a request gets back `no-cache` then all subsequent requests to that URL won't get cache busted.

All CDNs/proxies behave a bit differently — which is why there's multiple layers of protection here. Writing actual tests that we can verify against all of them would be excellent. I just did a lot of manual testing w/ custom SST setups deploying to Fastly/Cloudflare.

KyleAMathews:
For more context, we didn't see any of these issues until I started testing on CDNs.

msfstef:
@KyleAMathews you're right and indeed Fastly collapses requests but then "uncollapses" them if the response is not cacheable:

image
from [their docs](https://www.fastly.com/documentation/guides/concepts/edge-state/cache/request-collapsing/)

Therefore if we _need_ to make the response cacheable we can never make sure we don't run into a cycle :/ so I think I can reinstate the cursor, which makes it safer to make this change as well, but still make sure we never _revalidate_ requests with no changes, which we can do by not including etags on them.

We could also skip revalidation for all live requests.

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -554,9 +554,10 @@ defmodule Electric.Shapes.Api do
     %{response: response, global_last_seen_lsn: global_last_seen_lsn} =
       update_attrs(request, %{ot_is_empty_response: true})
 
-    %{
+    %Response{
       response
       | status: 200,
+        no_changes: true,
```
KyleAMathews, line 560: what is this?

packages/sync-service/lib/electric/shapes/api/response.ex
```diff
@@ -298,13 +293,21 @@ defmodule Electric.Shapes.Api.Response do
     Plug.Conn.assign(conn, :streaming_bytes_sent, bytes_sent)
   end
 
-  def etag(%__MODULE__{handle: handle, offset: offset, params: params} = _response, opts \\ []) do
-    etag = "#{handle}:#{params.offset}:#{offset}"
+  def etag(response, opts \\ [])
+
+  def etag(%__MODULE__{handle: handle, offset: offset, params: params, no_changes: true}, opts) do
+    "#{handle}:#{params.offset}:#{offset}:#{System.monotonic_time()}"
```
KyleAMathews, line 299: add a comment

packages/typescript-client/test/cache.test.ts
```diff
@@ -105,6 +107,79 @@ describe(`HTTP Proxy Cache`, { timeout: 30000 }, () => {
     expect(getCacheStatus(cachedRes)).toBe(CacheStatus.HIT)
   })
 
+  it.only(`should collapse requests in live mode`, async ({
```
KyleAMathews, line 110:
```suggestion
  it(`should collapse requests in live mode`, async ({
```

Is this test still useful?

msfstef: I think so, good to have as a safety check like the other cache ones - I mostly added it cause I realised we didn't have that anywhere

icehaunter (approved):

PR #2599, opened 2025-04-15
Sync publication state to active shapes on load

### LINKED ISSUE #1719: Electric fails to detect when its persistent storage goes away and prevents Postgres from cleaning up WAL files
Imagine a simple scenario where Electric is running inside a Kubernetes pod with no persistent storage. Some shapes get created, so Electric creates a publication in Postgres and starts processing transactions.

When the pod is restarted, a new file system is created for it with no traces of the previous shape storage. Electric will no longer be able to process incoming transactions from Postgres, causing the latter to build up its WAL backlog indefinitely.

Electric should be able to detect this failure mode and drop transactions when there is no active shape collector process instead of keeping those transactions around by refusing to advance the replication slot.

magnetised:
Fixes #1719 - I think.

Don't know if we've made other changes to the replication system since after 1719 but things seem robust against a storage reset.

This pr does fix the synchronization of the publication state to the active shape state, so if electric loses its persistent storage between restarts it will reset the publication state (tables plus filter clauses). Previously the shapes that were active are not restored on restart the outdated publication state would persist. To do this I've made the initial sync call to the publication manager synchronous 

1. start electric
2. create shape
3. stop electric
4. write data to shape
5. delete the persistent state
6. start electric

on restart:

1. replication client receives pending transactions from old shape table but discards them because there are no active consumers
2. old shape table is removed from  the publication
3. replication proceeds as for a new/blank electric

1 and 2 could happen in any order but the end result is the same. Nothing builds up in the replication slot as the replication client is active all the time.

I'm not doing any special checks for ownership because we're not doing anything that electric wouldn't normally do to the publication that it assumes it owns according to it's configuration.


packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -380,7 +382,10 @@ defmodule Electric.ShapeCache do
     |> Enum.flat_map(fn {shape_handle, shape} ->
       start_shape_with_timeout(shape_handle, shape, state, timeout)
     end)
-    |> Lsn.max()
+    |> Enum.unzip()
+    |> then(fn {lsns, shapes} ->
+      {Lsn.max(lsns), shapes}
```
msfstef, line 387: what are we returning the shapes for though?

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -186,21 +188,25 @@ defmodule Electric.Replication.PublicationManager do
...
     {:noreply, state}
   end
 
-  def handle_call(:refresh_publication, from, state) do
-    state = add_waiter(from, state)
-    state = schedule_update_publication(state.update_debounce_timeout, state)
-    {:noreply, state}
+  # This returns immediately and the actual sync is done asynchronously to
+  # separate boot of shape cache from refreshing publication. This means errors
+  # generated by the initial sync won't show up in the stack init, but only
```
msfstef, line 204: I feel this goes counter to the work in https://github.com/electric-sql/electric/pull/2490 - the idea was that if we cannot recover the stored shapes fully, including the publication, then Electric fails to start. This was introduced because we got into a situation where a deployed Electric was considered "healthy" but was subsequently failing.

msfstef (commented):
I think that this is superceded by https://github.com/electric-sql/electric/pull/2634 which introduces a `forced?` flag - I'm concerned about making this async, perhaps this can be addressed within the context of that PR?

alco:
The problem of keeping the publication up-to-date has been addressed by other changes, as mentioned above.

Regardless, if there's a need to "clean the publication up" in response to the shape storage going away, the cleanest way to do it would be to recreate it? And let the connection setup logic implemented in https://github.com/electric-sql/electric/pull/2706 take care of recreating the slot as well.

PR #2604, opened 2025-04-16
fix: improve error handling around a periodic wal size check

icehaunter:
This fixes an error we're seeing sometimes in Sentry around monitoring. Not sure about the root cause of process being unresponsive, but it might be that it's a slow query against Postgres that's blocking the entire thing. This PR moves the wal query into a task so that the manager can continue to handle `DOWN` messages and such

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -144,7 +144,16 @@ defmodule Electric.Connection.Manager do
   end
 
   def report_retained_wal_size(server) do
-    GenServer.call(server, :report_retained_wal_size)
+    try do
+      GenServer.call(server, :report_retained_wal_size)
+    catch
+      :exit, exit_reason ->
+        Logger.warning(
+          "Failed to report retained WAL size: #{inspect(exit_reason)}. Current queue size for #{inspect(server)} is #{inspect(Process.info(server, :message_queue_len))}"
+        )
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -225,10 +234,15 @@ defmodule Electric.Connection.Manager do
...
-  def handle_call(:report_retained_wal_size, _from, state) do
+  def handle_call(:report_retained_wal_size, _from, %{pool_pid: pool, stack_id: stack_id} = state) do
     if state.monitoring_started? do
       slot_name = Keyword.fetch!(state.replication_opts, :slot_name)
-      query_and_report_retained_wal_size(state.pool_pid, slot_name, state.stack_id)
+
+      # We don't want to block the manager on a PG query in case the network connection
+      # is slow, so we run it in a separate task.
+      Task.start(fn ->
+        query_and_report_retained_wal_size(pool, slot_name, stack_id)
```
magnetised, line 244: is there any danger of these stacking up if things are slow?

magnetised (approved):
lovely

packages/sync-service/lib/electric/connection/manager.ex
icehaunter: Probably there is! I'll add a timeout that's less than polling period

PR #2616, opened 2025-04-17
fix: shape cleanup should delete the shape handle too

icehaunter:
We had an issue, where when we cleaned up shapes (e.g. on `purge_all_shapes` which happened to be triggered for everyone because of `timeline null` issue), if the processes were never started (again, in particular for `purge_all_shapes`, then the shape handle failed to be cleaned up, and it got our system into an incosistent state where the ETS said shape exists, but no associated processes were ever started


alco (approved):
Perfect!

PR #2617, opened 2025-04-17
fix: make sure wal size queries don't block or crash

icehaunter:


packages/sync-service/lib/electric/telemetry/stack_telemetry.ex
```diff
@@ -291,5 +291,49 @@ with_telemetry [OtelMetricExporter, Telemetry.Metrics] do
...
+              deadline: 3_000
+            )
+
+          # The query above can return `-1` which I'm assuming means "up-to-date".
+          # This is a confusing stat if we're measuring in bytes, so normalise to
+          # [0, :infinity)
+
+          :telemetry.execute(
+            [:electric, :postgres, :replication],
+            %{wal_size: max(0, wal_size) |> dbg},
```
alco, line 322: Forgot to remove the `dbg` call here.

alco (approved):
Great!

I don't think an explicit lookup of the process PID is necessary. I removed it and did a bit of cleanup in a separate PR based on this branch - https://github.com/electric-sql/electric/pull/2618.

PR #2623, opened 2025-04-21
Yjs Provider and example

balegas:
TODO:
- [x] readme and better documentation
- [x] publish y-electric as separate package
- [x] fix demo deployment
- [x] ~~try different approach for storing offline changes that only keeps state vectors~~


examples/yjs/src/y-electric.ts
```diff
@@ -0,0 +1,495 @@
...
+    operations,
+    awareness,
+    resumeStateProvider,
+    databaseProvider,
+    connect = true,
+    onSendError,
+    fetchClient,
+  }: {
+    doc: Y.Doc
+    operations: {
```
samwillis, line 100: I would be tempted to name this `operationsShape` and the other `awarenessShape`, it makes it more apparent what they are.

In the PGLite sync extension the equvilent param is named `shape` (we one need one): https://pglite.dev/docs/sync#syncshapetotable-api

examples/yjs/src/y-electric.ts
```diff
@@ -0,0 +1,495 @@
...
+      options: OperationsShapeOptions
+      endpoint: string
+    }
```
samwillis, line 103: I think these can be better named, `options` is a shape options and should be named as such. And make it clear the endpoint is the url that you sending the operation to.

In PGLite-sync we named the shape options param `shape` maybe flow the same pattern here?

For the endpoint url, maybe `sendUrl`? Under the shape options we use `url` rather than `endpoint` so i think it's better to use url for consistency.

```suggestion
    operations: {
      shape: OperationsShapeOptions
      sendUrl: string
    }
```

examples/yjs/src/y-electric.ts
```diff
@@ -0,0 +1,495 @@
...
+    awareness?: {
+      options: AwarenessShapeOptions
+      endpoint: string
+      protocol: awarenessProtocol.Awareness
+    }
+    resumeStateProvider?: ResumeStateProvider
```
samwillis, line 109: I think these can be better named, `options` is a shape options and should be named as such. And make it clear the endpoint is the url that you sending the operation to.

In PGLite-sync we named the shape options param `shape` maybe flow the same pattern here?

For the endpoint url, maybe `sendUrl`? Under the shape options we use `url` rather than `endpoint` so i think it's better to use url for consistency.

```suggestion
    operations: {
      shape: OperationsShapeOptions
      sendUrl: string
    }
    awareness?: {
      shape: AwarenessShapeOptions
      sendUrl: string
      protocol: awarenessProtocol.Awareness
    }
```

examples/yjs/src/y-electric.ts
```diff
@@ -0,0 +1,495 @@
...
+      if (origin === `server`) {
+        return
+      }
+
+      this.batch(update)
+      this.sendOperations()
+    }
+
+    this.awarenessHandler = (update, origin) =>
+      this.sendAwarenessState(update, origin)
```
samwillis, line 154: Same here, why not a method and why alias?

examples/yjs/src/y-electric.ts
```diff
@@ -0,0 +1,495 @@
...
+      }
+    }
+
+    if (databaseProvider) {
+      databaseProvider.once(`synced`, () => {
+        this.ready = true
+        if (connect) {
+          this.connect()
+        }
+      })
```
samwillis, line 168: If I understand correctly, it's still up to the user to connect their `databaseProvider` to their document,  we don't do it for them? They can pass it to the ElectricProvider so that it's not connected until they have loaded the initial state from the db?

I think this make sense, we don't want to handle connecting the other provider - its normal in Yjs to have the user wire up all their providers.

Question is, do we need this? Can the user not just wait until the initial sync on their provider before then connecting the Electric one?

examples/yjs/src/y-electric.ts
```diff
@@ -0,0 +1,495 @@
...
+        this.awareness.protocol,
+        [this.doc.clientID],
+        `local`
+      )
+
+      this.awareness.protocol.setLocalState({})
+    }
+
+    this.emit(`connection-close`, [])
+
```
samwillis, line 264: We should consider if we want to await any pending updates to be sent before the `desconnect()` method returns.
Alternatively we could abort any in progress send calls?

examples/yjs/src/y-electric.ts
```diff
@@ -0,0 +1,495 @@
...
+  }
+
+  private batch(update: Uint8Array) {
+    if (this.resume.batching) {
+      this.resume.batching = Y.mergeUpdates([this.resume.batching, update])
+    } else {
+      this.resume.batching = update
+    }
+    this.resumeStateProvider?.save(this.resume)
+  }
```
samwillis, line 225: If I understand correctly, we batch all updates from Yjs to a merged `batching` buffer.
The `sendUpdates` method is looping reading from this buffer and sending batches.
So if there are multiple local updates while a `fetch` is in-flight they are batched together. i.e. there is only ever one in flight update fetch call at once?
I like this, it works well.

We are then also optionally persisting that batch to local storage, essentially as a WAL.

examples/yjs/src/local-storage-persistence.ts
```diff
@@ -0,0 +1,65 @@
...
+        ...state,
+        batching: state.batching ? Array.from(state.batching) : undefined,
+        sending: state.sending ? Array.from(state.sending) : undefined,
+      }
+      // We need to convert the Uint8Array to a regular array for storage
+      localStorage.setItem(this.key, JSON.stringify(dataToStore))
+      // TODO: check that we're not logging to much
+      // console.log(`Resume state saved to localStorage`, dataToStore)
+    } catch (error) {
+      console.error(`Failed to save resume state to localStorage`, error)
```
samwillis, line 33: Fail silently... I see the todo, but we need to think what the fallback is here.

examples/yjs/src/local-storage-persistence.ts
```diff
@@ -0,0 +1,65 @@
...
+        parsedData.batching = new Uint8Array(parsedData.batching)
+      }
+
+      if (parsedData.sending) {
+        parsedData.sending = new Uint8Array(parsedData.sending)
+      }
+
+      return parsedData
+    } catch (error) {
+      console.error(`Failed to load resume state from localStorage`, error)
```
samwillis, line 61: I would be tempted to not silently fail here.

examples/yjs/src/y-electric.ts
```diff
@@ -0,0 +1,495 @@
...
+
+  private operationsShapeHandler(
+    messages: Message<OperationMessage>[],
+    offset: Offset,
+    handle: string
+  ) {
+    for (const message of messages) {
+      if (isChangeMessage(message) && message.value.op) {
+        const decoder = message.value.op
```
samwillis, line 356: We currently use the message from the shape directly - the schema of the table is essentially fixed by the clients usage. Maybe we could provide a callback that enables a user to transform a shape messages into the required for the implementation, a `getOpFromMessage` callback.

Maybe v2...

samwillis (commented):
This is great!

Left a bunch of comments / thoughts.

The thing we dont demo is how to compact a Ydoc from all operations down to a compacted state with GC. But maybe thats out of scope for now.

examples/yjs/src/y-electric.ts
balegas: Now the wiring is down outside Electric. I don't like much that I need a provider for storage and a provider for the shape resume state. I think this can be a later improvement.

The ElectricResumeStateProvider now has async and sync interfaces.

examples/yjs/src/y-electric.ts
balegas: yep, I'm using the merge updates to batch all updates at once, this will also work for debouncing if I add it.

I've stopped persisting the batched at the expense of re-sending some changes back to the server. We've discussed that compaction will end up cleaning the duplicate state from the stream.

examples/yjs/src/y-electric.ts
balegas: yep, I've left that as a TODO.

examples/yjs/src/y-electric.ts
```diff
@@ -0,0 +1,495 @@
...
+    this.emit(`status`, [{ status: `connecting` }])
+  }
+
+  private async sendOperations() {
+    if (!this.connected || this.resume.sending) {
+      return
+    }
+
+    try {
+      while (this.resume.batching && this.resume.batching.length > 2) {
```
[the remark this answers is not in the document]
balegas: I've modified the coce slightly and is now more readable

examples/yjs/src/y-electric.ts
balegas: I was being a bit lazy. I've added the callbacks.

examples/yjs/src/local-storage-persistence.ts
balegas: everything throws now.... although not very easy to catch.

examples/yjs/src/local-storage-persistence.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  save(state: ResumeState): void {
+    try {
+      const dataToStore = {
+        ...state,
+        batching: state.batching ? Array.from(state.batching) : undefined,
+        sending: state.sending ? Array.from(state.sending) : undefined,
+      }
+      // We need to convert the Uint8Array to a regular array for storage
+      localStorage.setItem(this.key, JSON.stringify(dataToStore))
```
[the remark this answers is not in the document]
balegas: I've separate the two parts and encoded as b64

balegas (commented):
Thanks a lot for the comments, I've addressed the review. Please let me know if I missed something.

There is still a bit of work to do; 
-I need to fix tests 
- handle demo deployment 
- I'm leaving merging updates on the server as a follow up

I'll publish the demo first and then extract the provider, package it and release it!

samwillis (approved):
LGTM 👍

PR #2624, opened 2025-04-21
fix: Electric restart shouldn't cause requests with `0_inf` to fail

### LINKED ISSUE #2596: Error: `offset out of bounds for this shape` when `offset=0_inf`
**Versions**
- Client: PGlite sync
- Electric: Cloud + Local

**Bug description**
We have had multiple reports of users seeing this error when using the PGlite sync extension:

```
_FetchError: HTTP Error 400 at http://localhost:5155/v1/shape?columns=id%2Cname%2Cdescription%2Cuser_id%2Corg_id&cursor=15990&handle=7968192-1744424581840&live=true&offset=0_inf&table=projects&where=1%3D0: {"message":"Invalid request","errors":{"offset":["out of bounds for this shape"]}}
```

This breaks the sync as it's unrecoverable.

Chatting to @balegas we noticed that it only seems to happen when `live=false`, and has only started happening with PGlite and the new sync extension that supports multiple shapes.

Our working theory is that if you are syncing multiple tables, and the *abort a `live=true` request and connect with `live=false`* dance to check if we really are up to date, will trigger this code path.

Looking at traces from cloud, it looks like the code path starts with `live=false`, but when it hits the plug it swaps to `live=true`, that seems odd?

There is a question here on what the current behaviour is if we are correct. The disconnect dance is done in the `MultiShapeStream` api in the experimental package in this repo which PGlite then depends on.

Related thread with further information : https://discord.com/channels/933657521581858818/1361250652302479502

icehaunter:
Issue was only with shapes that didn't observe any writes after being created. When creating a snapshot, we don't know ahead of reading how many total rows we would read from DB, so we created a shape with an in-memory value for last known offset of `0_inf` - after all real snapshot chunks (denoted as `0_1`, `0_2`, ..., `0_n`), but before real LSNs with `x_y` where x > 0. On a restart however we'd read actually created chunk lists and fill in "last known" offset to be say `0_2` instead. Which meant suddenly clients who previously read through the snapshot and were just hanging waiting for new changes got the 400 because requested `0_inf` was indeed larger than presumed maximum of `0_2`. The PR just makes it so that the state of the world after a restart is consistent with the one we achieve while creating a fresh shape.

Closes #2596

### DIFF
--- a/.changeset/sour-poems-beg.md
+++ b/.changeset/sour-poems-beg.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: make sure Electric restart doesn't cause requests with `0_inf` to fail
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -4,6 +4,7 @@ defmodule Electric.Shapes.Consumer do
     significant: true
 
   import Electric.Postgres.Xid, only: [compare: 2]
+  import Electric.Replication.LogOffset, only: [is_virtual_offset: 1, last_before_real_offsets: 0]
 
   alias Electric.Shapes.Api
   alias Electric.LogItems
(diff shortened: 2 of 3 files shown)


alco (approved):
Brilliant!

I have pushed a simpler way to stop processes in the test, avoiding kills and a flake-inducing sleep.

icehaunter:
@KyleAMathews No, issue was only with shapes that didn't observe any writes after being created. When creating a snapshot, we don't know ahead of reading how many total rows we would read from DB, so we created a shape with an in-memory value for last known offset of `0_inf` - after all real snapshot chunks (denoted as `0_1`, `0_2`, ..., `0_n`), but before real LSNs with `x_y` where x > 0. On a restart however we'd read actually created chunk lists and fill in "last known" offset to be say `0_2` instead. Which meant suddenly clients who previously read through the snapshot and were just hanging waiting for new changes got the 400 because requested `0_inf` was indeed larger than presumed maximum of `0_2`. The PR just makes it so that the state of the world after a restart is consistent with the one we achieve while creating a fresh shape.

msfstef (commented):
Nice catch!

PR #2626, opened 2025-04-21
elixir-client: Stop explicitly obfuscating the DB password

alco:
Mirror PR: https://github.com/electric-sql/phoenix_sync/pull/23

### DIFF
--- a/.changeset/brave-jars-speak.md
+++ b/.changeset/brave-jars-speak.md
@@ -1,0 +1,5 @@
+---
+"@core/elixir-client": patch
+---
+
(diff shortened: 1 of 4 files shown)


alco:
@magnetised Please take a look and see if it makes sense. 

`config :electric_client, database_config: connection_opts` is only used in tests, as far as I can tell, so it can be deobfuscated there, i.e. no need for connection opt processing in the config file.

`config electric, replication_connection_opts: ...` is passed to StackSupervisor which does the obfuscation, so again -- no need for the explicit call in the config file.

PR #2628, opened 2025-04-21
Add Posthog analytics to website

KyleAMathews:


website/.vitepress/theme/Layout.vue
```diff
@@ -1,18 +1,37 @@
...
+import { useSidebar } from "vitepress/theme"
+const router = useRouter()
 
-import DefaultTheme from 'vitepress/theme-without-fonts'
+if (typeof window !== "undefined") {
+  watch(
+    () => router.route.data.relativePath,
+    (path) => {
+      posthog.init("phc_o4xENyuuSCdNPG2CWtfdqzYYXs6v8SbmVDzm3CP0Qwn", {
+        api_host: `https://admin.electric-sql.cloud/api/ph`,
```
thruflo, line 13: It appears that there are two API hosts here.

website/.vitepress/theme/Layout.vue
```diff
@@ -1,18 +1,37 @@
...
-import { useData } from 'vitepress'
-import { useSidebar } from 'vitepress/theme'
+import { watch } from "vue"
+import { useData, useRouter } from "vitepress"
+import { posthog } from "posthog-js"
+import { useSidebar } from "vitepress/theme"
+const router = useRouter()
 
-import DefaultTheme from 'vitepress/theme-without-fonts'
+if (typeof window !== "undefined") {
```
thruflo, line 8: Should this be in an onMounted vue lifecycle event handler? That also takes care of ensuring the code is executed client side.

Also, I suspect we should limit this to only-in-production.

thruflo (approved):

PR #2634, opened 2025-04-23
fix: handle publication/schema alterations on silent PG

### LINKED ISSUE #1733: Automatically delete shape(s) when a table is dropped
When a table is dropped from Postgres, users currently need to manually delete all shapes that are defined on that table. This is needed to avoid any problems that would occur when recreating the shape (e.g. with a different schema or different data) as Electric would still contain outdated data for that table potentially even in the old (invalid) format.

Currently, Electric is subscribed to a Postgres publication that streams changes (DML statements) about all tables involved in one or more shapes. However, Postgres does not stream DDL statements (such as `DROP TABLE`), which currently prevents Electric from detecting dropped tables and hence Electric can't automatically delete the corresponding shapes.

**The goal of this issue is to discuss/propose ways of detecting tables being dropped in order to automate shape deletion.**

[@thruflo proposed a very neat approach ](https://github.com/electric-sql/electric/issues/1689#issuecomment-2352845354) of using an event trigger that calls a special "purge" function when a table is dropped. That special function would write a row into a meta table, which will come through the replication stream. We could check for this special row on the replication stream to detect a table being dropped right in time. The only downside here is that event triggers require superuser access and don't work in single-user mode (e.g. which is the case of PGlite iirc).

### LINKED ISSUE #2449: Dropping and recreating a table that has an active shape subscription causes all subsequent updates to be invisible to Electric
When a table is dropped, it is removed from the Postgres publication. Subsequent recreation of the table doesn't automatically add it back to the publication. As a result, Electric remains oblivious of all new updates to the table and always responds with "up-to-date" to client requests that expect to fetch the new rows.

One of the possible solutions for this being discussed is polling the database (the publication in this case) to detect such discrepancies and restore the publication state asynchronously.

### LINKED ISSUE #2517: Shapes not invalidated when table is changed but no writes happen
This is mostly a "uncomfortable in development" bug - where as you're iterating a migration is not propagated to Elixir.

### LINKED ISSUE #2558: Snapshot creation fails when cached schema is stale
When trying to create a snapshot, we trust the shape definition, which is built with reliance on the cached DB schema. If the schema has changed, but we've not seen that (due to lack of writes on the affected table), then we may allow shape definition which is actually wrong, and that causes an error while querying Postgres.

---
Sentry Issue: [ELECTRIC-2KF](https://electricsql-04.sentry.io/issues/36362727/?referrer=github_integration)

```
Snapshot creation failed for 79993341-1744059563239 because of:
** (Postgrex.Error) ERROR 42703 (undefined_column) column "some column" does not exist

    query: SELECT <our_snapshot_query>
    (db_connection 2.7.0) lib/db_connection.ex:725: DBConnection.prepare!/3
    (electric 1.0.2) lib/electric/shapes/querying.ex:33: anonymous fn/4 in Electric.Shapes.Querying.stream_initial_data/4
    (opentelemetry 1.5.0) /app/deps/opentelemetry/src/otel_tracer_default.erl:47: :otel_tracer_default.with_span/5
    (electric 1.0.2) lib/electric/telemetry/open_telemetry.ex:87: anonymous fn/3 in Electric.Telemetry.OpenTelemetry.do_with_span/4
    (telemetry 1.3.0) /app/deps/telemetry/src/telemetry.erl:324: :telemetry.span/3
    (electric 1.0.2) lib/electric/shapes/consumer/snapshotter.ex:186: anonymous fn/8 in Electric.Shapes.Consumer.Snapshotter.query_in_readonly_txn/7
    (opentelemetry 1.5.0) /app/deps/opentelemetry/src/otel_tracer_default.erl:47: :otel_tracer_default.with_span/5
    (electric 1.0.2) lib/electric/telemetry/open_telemetry.ex:87: anonymous fn/3 in Electric.Telemetry.OpenTelemetry.do_with_span/4

```

icehaunter:
When PG is silent - there are no updates to tables that might interest us - the we're not busting the cache of the schema, not invalidating shapes for deleted tables, etc. This PR addresses that.

Practical impact:
- `PublicationManager` now tracks table OIDs along with the schema/name pair.
- When altering the publication, we check for dropped, recreated, or renamed tables. For any that match, we change the altering of the publication to not touch the affected tables, and instead force a clear on affected shapes. This is needed despite a periodic check, because there is a window where this may come up between the checks.
- `PublicationManager` now has a bit of coupling to `ShapeCache` to notify all the shapes that should be invalidated.
- For invalidated shapes, we're sending 409s, because maybe it's just a table recreation and clients can start from scratch, or maybe it's a table drop and further requests will yield a 400 when table is actually no longer found.

Issue impact:
- Closes #2449 - decision here is that if the table has been recreated, we shouldn't re-add it to the publication automatically, but rather invalidate the shapes and creation of a new shape initiated by the client might re-add the table if it need to.
- Closes #2558 - previous PR already closed this issue, but this PR makes the window where the guard is necessary smaller by polling the schema periodically
- Closes #1733 - addresses it in two ways, by periodic polling as well as double-checking when actually doing the publication altering to have a transactionally-consistent view
- Closes #2517 - schema changes on a silent DB are caught within a polling period


packages/sync-service/lib/electric/replication/schema_reconciler.ex
```diff
@@ -0,0 +1,97 @@
+defmodule Electric.Replication.SchemaReconciler do
+  @moduledoc false
+  _ = """
```
robacourt, line 3:
```suggestion
  @moduledoc """
```

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -7,40 +7,101 @@ defmodule Electric.Postgres.Configuration do
...
+    |> case do
+      {:ok, missing_relations} ->
+        missing_relations
+
+      {:error, reason} ->
+        raise reason
+    end
+  end
+
+  defp list_missing_relations(previous_relations, conn) do
```
msfstef, line 74:
```suggestion
  defp list_missing_relations(relations, conn) do
```
nit: technically this helper doesn't know anything about previous or next relations - just whether the relations are present/the same

also technically this would be better if called `list_changed_relations`?

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -111,57 +165,28 @@ defmodule Electric.Postgres.Configuration do
...
+    oids =
       relation_filters
       |> Map.keys()
-      |> Enum.unzip()
-      |> Tuple.to_list()
+      |> Enum.map(&elem(&1, 0))
 
     %Postgrex.Result{rows: rows} =
-      Postgrex.query!(conn, query_for_relations_without_full_identity, params)
+      Postgrex.query!(conn, query_for_relations_without_full_identity, [oids])
```
msfstef, line 189: do we have to worry at all about this list of oids becoming _too_ large? it scales with the number of relations so I doubt this is a problem, just making sure

packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
```diff
@@ -1,36 +1,66 @@
...
   def load_relation({schema, name}, conn) when is_binary(schema) and is_binary(name) do
     query = load_relation_query("format('%I.%I', $1::text, $2::text)::regclass")
-    do_load_relation(conn, query, [schema, name])
+
+    case do_load_relation(conn, query, [schema, name]) do
+      {:ok, []} ->
+        {:error, "No relation found"}
+
+      {:ok, relations} ->
+        {:ok, hd(relations)}
```
msfstef, line 37: nit: potentially need to add a similar comment to the one that was removed from `do_load_relation` here about expecting a single relation?

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -250,25 +298,28 @@ defmodule Electric.Replication.PublicationManager do
            publication_name: publication_name,
            db_pool: db_pool,
            pg_version: pg_version,
-           configure_tables_for_replication_fn: configure_tables_for_replication_fn
+           configure_tables_for_replication_fn: configure_tables_for_replication_fn,
+           next_update_forced?: forced?
```
msfstef, line 302: I'd appreciate a small comment somewhere about when a `forced?` update might be necessary

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -207,40 +226,69 @@ defmodule Electric.Replication.PublicationManager do
...
     case update_publication(state) do
-      {:ok, state} ->
+      {:ok, state, missing_relations} ->
+        if missing_relations != [] do
+          Logger.info(
+            "Relations dropped/renamed since last publication update: #{inspect(missing_relations)}"
+          )
+
+          {mod, args} = state.shape_cache
+          mod.clean_all_shapes_for_relations(missing_relations, args)
```
msfstef, line 236: I don't love this coupling to the shape cache and cleaning of shapes - I'm assuming you wanted to avoid extra indirections though

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -75,12 +82,13 @@ defmodule Electric.Replication.PublicationManager do
             stack_id: [type: :string, required: true],
             publication_name: [type: :string, required: true],
             db_pool: [type: {:or, [:atom, :pid, @name_schema_tuple]}],
+            shape_cache: [type: :mod_arg, required: false],
```
msfstef, line 85: slightly worried about this coupling - would some kind of event subscription method for "missing relations" be too cumbersome?

packages/sync-service/lib/electric/postgres/inspector.ex
```diff
@@ -104,6 +107,10 @@ defmodule Electric.Postgres.Inspector do
     end)
   end
 
+  def list_diverged_relations({module, opts}) do
```
msfstef, line 110: I think `diverged_relations` is not clear enough as a name - perhaps `list_invalid_cached_relations` or `list_relations_for_cache_invalidation` or something to that effect would be more clear

otherwise if you don't agree I'd appreciate a bit more documentation on the API for what it is as I don't think it's immediately obvious what the divergence refers to, and it implies a sort of statefulness that is only present for the cached inspectors (which is why I think it would be better to make a direct reference to the caching)

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -7,40 +7,101 @@ defmodule Electric.Postgres.Configuration do
...
-      fn conn ->
-        configure_tables_for_replication_internal!(
-          conn,
-          relation_filters,
-          pg_version,
-          publication_name
-        )
+          non_neg_integer(),
+          String.t()
+        ) :: [Electric.oid_relation()]
```
msfstef, line 34: nit: worth mentioning in the docstring that what the return relations are - as the `configure_publication!` call returning those is a bit unintuitive (perhaps returning a map with a `changed_relations` or `unpublished_relations` key would help?)

packages/sync-service/lib/electric/replication/schema_reconciler.ex
```diff
@@ -0,0 +1,97 @@
...
+    # We essentially never want to fail here, as this is a periodic task.
+    # If it fails, we'll just try again next time, so no additional retries are implemented
+    with {:ok, diverged_relations} <- Inspector.list_diverged_relations(state.inspector),
+         :ok <-
+           shape_cache_mod.clean_all_shapes_for_relations(diverged_relations, shape_cache_args),
+         :ok <-
+           Enum.each(diverged_relations, fn {_, rel} ->
+             Inspector.clean(rel, state.inspector)
+           end),
+         :ok <- PublicationManager.refresh_publication(stack_id: state.stack_id, forced?: true) do
```
msfstef, line 81: are we worried about any race conditions here with requests coming in for shapes or for creating shapes using the cached inspector info etc between cleaning the shapes for relations and cleaning the inspector cache?

the publication update should be safe regardless, but making sure we don't introduce any races that don't resolve themselves without being too loud

msfstef (approved):
Looks great!

My only concern is the `ShapeCache` coupling in the `PublicationManager`, which I get is the most straightforward way but just makes the manager manage more than just the publication

packages/sync-service/lib/electric/postgres/configuration.ex
icehaunter: Still a single variable interpolation, and it's a list of all tables. If you postgres has like 100k tables than it might be a problem, but I doubt that this particular line would be the first problem we hit. I don't think there are interpolation limits on oid checks, it's a primary key for the table we're checking, and biggest issue with this is query transfer size, but again, likely not an issue.

packages/sync-service/lib/electric/replication/publication_manager.ex
icehaunter: Yeah, I though about it too. We could go through smth like an event registry, but this is essentially a clean `Process.send` under the hood, so coupling is fairly loose. I think added complexity otherwise is not worth it, and I don't think it's possible to draw a one-way-only hierarchy unfortunately

packages/sync-service/lib/electric/replication/schema_reconciler.ex
icehaunter: Nah, shapes cleaning themselves will bulk-schedule a refresh for the publication which should be then hit in the same window here and invalidate the rest of the shapes that we didn't get otherwise. Even if we refresh twice in a short succession, no harm there afaict

PR #2635, opened 2025-04-23
fix: Properly interpolate variable in GH workflow

kevin-dp:
Fixes https://github.com/electric-sql/stratovolt/issues/530


KyleAMathews (approved):
:shipit:

Issue #2612, opened 2025-04-17
Surface Postgrex errors


balegas:
Postgrex errors are unhandled throwing at the top level. We want to handle these errors and surface errors that might relevant to the user

PR #2637, opened 2025-04-24
Surface more connection-related events as stack events

alco:
Addresses #2612 and https://github.com/electric-sql/stratovolt/issues/421.

Here's an example of what it looks like and PG is stopped and later restarted:
Screenshot from 2025-04-24 02-39-45

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -247,12 +247,7 @@ defmodule Electric.Connection.Manager do
           |> mark_connection_succeeded()
           |> update_connection_opts(connection_opts)
 
-        Electric.StackSupervisor.dispatch_stack_event(
-          state.stack_events_registry,
-          state.stack_id,
-          :waiting_for_connection_lock
-        )
-
+        dispatch_stack_event(:waiting_for_connection_lock, state)
         schedule_periodic_connection_status_log(:log_lock_connection_status)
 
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/connection/supervisor.ex
```diff
@@ -97,7 +97,7 @@ defmodule Electric.Connection.Supervisor do
       Electric.StackSupervisor.dispatch_stack_event(
         opts[:stack_events_registry],
         stack_id,
-        :ready
+        :stack_is_running
```
msfstef, line 100: any particular reason to change this? this will require changes across cloud and other tests so would rather avoid it if it isn't strictly necessary

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -422,6 +426,8 @@ defmodule Electric.Connection.Manager do
         inspect(reason, pretty: true) <> "\n\n" <> inspect(state, pretty: true)
     )
 
+    dispatch_stack_event({:database_connection_severed, format_exit_reason(reason)}, state)
```
msfstef, line 429: we need to document all possible stack events and handle them in cloud appropriately - perhaps we need them centrally defined somewhere for ease of documentation?

msfstef (approved):
Looks good! As noted, we need to explicitly handle these extra events on Cloud.

PR #2638, opened 2025-04-24
fix: don't send a 500 on a shape delete race condition

icehaunter:
We're currently sending a 500 by raising an error when a shape has been deleted in-between us getting a shape handle and asking for the log of that shape. It's more reasonable to send a 409 and let the client deal with an actual error of table being deleted or some such.


msfstef (approved):
Makes a lot of sense!

Would like a test as well if possible just for good measure?

PR #2651, opened 2025-04-28
fix: ensure continued replication when WAL slot size is exceeded

### LINKED ISSUE #2640: Autorecover when pg says a slot has "exceeded the maximum reserved size"
**Versions**
- Electric: latest

**Bug description**
Currently if PG shuts down a publication slot because of the error: "This slot has been invalidated because it exceeded the maximum reserved size." Electric just shuts down. This is really disruptive in production because it means downtime.

**Expected behavior**
Electric should tell the user that they need to increase their `max_slot_wal_keep_size` setting to avoid hitting this error and then delete the slot, local shapes, and reconnect to start over.

icehaunter:
Closes #2640

### DIFF
--- a/.changeset/ninety-toes-talk.md
+++ b/.changeset/ninety-toes-talk.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: ensure continued replication (via shape reset) when WAL slot size is exceeded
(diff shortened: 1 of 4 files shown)


alco (approved):
Perfect!

PR #2652, opened 2025-04-28
Refactor and fix connection initialization/reconnection backoff

alco:
The highlight of this PR is the code refactoring inside Connection.Manager module to make the state machine that it implements more explicit and transitions between steps better documented.

Not too long ago we made the initialization of ReplicationClient asynchronous which had been necessitated by the possibility of its startup getting blocked on a pending DB transaction. This PR also makes LockConnection's initialization async for consistency. Both types of processes now return early from their respective `init()` callbacks and then cast messages to the connection manager when reaching a new milestone (connected, lock_acquired, ready_to_stream, streaming_started).

What prompted this refactoring was a new failure state we had discovered. If there's a pending transaction that blocks Electric's initialization for the first time and Electric is then shut down, Postgres will still have a running backend blocked on the `CREATE_REPLICATION_SLOT` statement. When Electric starts again, it will be able to complete the full connection setup (because the replication slot has already been created in Postgres) but will fail when switching ReplicationClient into streaming mode because it's aready actively used by a different OS process (the `walsender` backend in Postgres stuck from Electric's first launch).

Electric's failure to handle this specific scenario resulted in that it got stuck in a reconnection loop without exponential backoff, always trying to reconnect every 2 seconds. Additionally, it resulted in a snowballing of warning logs with each new logger call getting added on every reconnection attempt, due to the absence of any way to keep track of periodic connection status checks and the logs they were emitting.

The PR fixes both the runaway logging and the error handling of this new failure state. It is now simulated in the new integration test defined in `replication-client-self-conflict.lux`.

A new unit test is added that shows the minimum required configuration to start a Connection.Supervisor in isolation. Trimming it down further to only start the Connection.Manager process would require some mocking which I dislike.

As a nice bonus, I discovered a bug in reading the connection options and pool options from application env: the app env wasn't actually being consulted. Probably not a problem in Electric Cloud but it did result in the `ELECTRIC_QUERY_DATABASE_URL` config to be ignored.

For code reivew, I suggest opening the updated code for Connection.Manager and reading it top down. This should be easier than trying to make sense of the diff view.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -1047,9 +1047,9 @@ defmodule Electric.Connection.Manager do
   defp update_connection_opts(
          :start_replication_client,
          conn_opts,
-         %State{shared_connection_opts: nil} = state
+         %State{shared_connection_opts: nil, replication_opts: replication_opts} = state
        ) do
-    %State{state | replication_opts: put_in(state.replication_opts[:connection_opts], conn_opts)}
+    %State{state | replication_opts: put_in(replication_opts, [:connection_opts], conn_opts)}
   end
 
   defp update_connection_opts(_step, conn_opts, %State{shared_connection_opts: nil} = state) do
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -27,8 +27,55 @@ defmodule Electric.Connection.Manager do
       Supervisor.start_link(children, strategy: :one_for_one)
   """
 
+  # We model the connection manager as a state machine with 2 levels of state. On the 1st
+  # level, it can be in one of several phases:
+  #
+  #   - :connection_setup
+  #   - :running
+  #   - :restarting_replication_client
+  #
+  # Connection manager starts in the :connection_setup phase and moves into the :running phase
+  # after all processes have been started and the replication client is given the command to
+  # start streaming.
+  #
+  # If at any point after that the replication client process exits, connection manager
+  # transitions into the :restarting_replication_client phase and tries to restart the client.
+  #
+  # If any other connection process exits, connection manager itself will shut down to start
+  # from scratch, as there could be too many failure states to address each one individually.
+  #
+  # The 2nd level of the state machine splits the current phase into a series of steps which
+  # the process goes through until it finishes the connection setup phase and transitions into
+  # the running phase.
+  #
+  # Function clauses match on the current phase and step as a way to both assert on the current
+  # state as well as make the code more self-documenting.
+
   defmodule State do
+    @type phase :: :connection_setup | :running | :restarting_replication_client
+    @type step ::
+            {:start_lock_connection, nil}
+            | {:start_lock_connection, :connecting}
+            | {:start_lock_connection, :acquiring_lock}
+            # Steps that start with {:start_replication_client, ...} are pertinent to both the
+            # :connection_setup phase and the :restarting_replication_client phase
+            | {:start_replication_client, nil}
+            | {:start_replication_client, :connecting}
+            | {:start_replication_client, :configuring_connection}
+            | :start_connection_pool
+            | :start_shapes_supervisor
+            | {:start_replication_client, :start_streaming}
+            # Steps of the :running phase:
+            | nil
+            | :waiting_for_streaming_to_start
+
     defstruct [
+      # The phase the connection manager is in. It defines which actions are taken if any of
+      # the connection processes exit.
+      :current_phase,
+      # The current step defines what the connection manager does in the current phase and
+      # which step will be taken next.
+      :current_step,
       # Database connection opts to be passed to Postgrex modules
       :connection_opts,
       # Replication options specific to `Electric.Postgres.ReplicationClient`
@@ -46,18 +93,18 @@ defmodule Electric.Connection.Manager do
       :shape_cache_opts,
       # PID of the replication client
       :replication_client_pid,
+      # Timer reference for the periodic replication client status check
+      :replication_client_timer,
       # PID of the Postgres connection lock
       :lock_connection_pid,
+      # Timer reference for the periodic lock status check
+      :lock_connection_timer,
       # PID of the database connection pool
       :pool_pid,
       # PID of the shape log collector
       :shape_log_collector_pid,
       # Backoff term used for reconnection with exponential back-off
       :connection_backoff,
-      # Flag indicating whether the lock on the replication has been acquired
-      :pg_lock_acquired,
-      # This flag is set to true when the replication connection completes its setup procedure.
-      :replication_connection_established,
       # PostgreSQL server version
       :pg_version,
       # PostgreSQL system identifier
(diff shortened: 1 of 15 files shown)


alco:
I wanted to note that with the LockConnection's initialization now also async, the following error may get logged more frequently:

```
17:23:08.783 pid=<0.386.0> [debug] Starting replication client for stack single_stack

17:23:08.795 pid=<0.697.0> [error] :gen_statem {:"Elixir.Electric.ProcessRegistry:single_stack", {Electric.Postgres.ReplicationClient, nil}} terminating
** (Postgrex.Error) ssl not available
    (stdlib 6.2) gen_statem.erl:3864: :gen_statem.loop_state_callback_result/11
    (stdlib 6.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
Process Label: {:replication_client, "single_stack"}
Queue: [internal: {:connect, :init}]
Postponed: []
State: {:no_state, %Postgrex.ReplicationConnection{protocol: nil, state: {Electric.Postgres.ReplicationClient, %Electric.Postgres.ReplicationClient.State{stack_id: "single_stack", connection_manager: #PID<0.386.0>, transaction_received: {Electric.Replication.ShapeLogCollector, :store_transaction, [{:via, Registry, {:"Elixir.Electric.ProcessRegistry:single_stack", {Electric.Replication.ShapeLogCollector, nil}}}]}, relation_received: {Electric.Replication.ShapeLogCollector, :handle_relation_msg, [{:via, Registry, {:"Elixir.Electric.ProcessRegistry:single_stack", {Electric.Replication.ShapeLogCollector, nil}}}]}, publication_name: "electric_publication_default", try_creating_publication?: true, start_streaming?: false, slot_name: "electric_slot_default", slot_temporary?: false, display_settings: ["SET bytea_output = 'hex'", "SET DateStyle = 'ISO, DMY'", "SET TimeZone = 'UTC'", "SET extra_float_digits = 1", "SET IntervalStyle = 'iso_8601'"], origin: "postgres", txn_collector: %Electric.Postgres.ReplicationClient.Collector{transaction: nil, tx_op_index: nil, relations: %{}}, step: :disconnected, applied_wal: 0}}, auto_reconnect: false, reconnect_backoff: 500, streaming: nil}}
Callback mode: :handle_event_function, state_enter: false

17:23:08.799 pid=<0.386.0> [warning] Failed to connect to the database using SSL. Trying again, using an unencrypted connection.
```

We need to make sure it's ignored in Sentry. The is logged when DATABASE_URL or ELECTRIC_QUERY_DATABASE_URL is configured without `sslmode` or with `sslmode=prefer/require` while the Postgres cluster is not configured to serve over SSL. The LockConnection process exits and the exit reason is logged by OTP machinery separately from our handling of that exit.

The reason you may have not seen these logs earlier was because our LockConnection initialization was synchronous, and so this error was returned from its `start_link()` call as opposed to getting logged by OTP, and also because when only DATABASE_URL configuration was set, Electric would perform fallback to noSSL once, and so by the time it got to asynchronously initialization the ReplicationClient process, its connection opts would already have `ssl: false`.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -226,149 +273,246 @@ defmodule Electric.Connection.Manager do
...
+      )
 
-      {:error, reason} ->
-        handle_connection_error(reason, state, @replication_mode)
-    end
+    # NOTE(alco): We're jumping ahead of ourselves here a bit because at this point we don't
+    # yet have a confirmation that the connection pool has succeeded in opening a database
+    # connection. But since we already have a lock connection and a replication connection
+    # open, it's likely the connection pool will also succeed.
+    Electric.StatusMonitor.mark_connection_pool_ready(state.stack_id, pool_pid)
```
robacourt, line 380: So is this state useful to have on the StatusMonitor? What does it being true tell us?

If nothing then maybe it should be removed. Albeit that would mean the status would be unaffected if the pool pid died, but that shouldn't happen on it's own 🤔

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -203,9 +202,26 @@ defmodule Electric.Postgres.ReplicationClient do
...
+  # check whether the START_REPLICATION_SLOT statement succeeded before switching the
+  # connection into streaming mode. Returning {:query, "START_REPLICATION_SLOT ...", state}
+  # works fine when the query result is an error: it is then passed to the handle_result()
+  # callback. But if streaming starts without issues, a function clause error is encountered
+  # inside Postgrex.ReplicationConnection because it expects the connection to already have
+  # been switched into streaming mode by returning {:stream, "START_REPLICATION_SLOT ...", [], state}.
+  #
+  # Hence this function clause of `handle_data()` that notifies the connection manager about
+  # successful streaming start as soon as it receives the first replication message from
+  # Postgres.
```
robacourt, line 215: What if there are no changes? Conn.Man would get stuck in a the `waiting_for_streaming_to_start` step. Or are we guaranteed to have changes?

packages/sync-service/test/electric/db_connection_error_test.exs
```diff
@@ -48,7 +48,7 @@ defmodule Electric.FatalErrorTest do
       }
 
       assert %DbConnectionError{
-               message: "User does have the REPLICATION attribute. " <> _,
+               message: "User does not have the REPLICATION attribute. " <> _,
```
robacourt, line 51: An important word to leave out 🤣 . Good spot!

robacourt (approved):
Awesome! As well as some important fixes, Conn.Man seems much clearer and robust now. Great work, Oleksii! I've just left a couple of comments

packages/sync-service/lib/electric/postgres/replication_client.ex
alco: This is not a problem. The step `:waiting_for_streaming_to_start` belongs to the `:running` phase. The replication client has been marked as ready with the StatusMonitor earlier, when it notified the connection manager that it finished connection setup and was ready to start streaming.

If you look inside the `Connection.Manager.handle_cast()` callback that processes the `:replication_client_started_streaming` notification, you'll see that it's only affecting the exponential backoff state. I know there's a disconnect in reasoning between the two modules here but I'm not sure if adding more doc comments can fix that.

packages/sync-service/lib/electric/connection/manager.ex
alco: Connection pool status is a bit wishy-washy right now. Here we're optimistically setting it to ready. If a connection error is countered when both the lock connection and the replication client are already up, we will mark the connection pool as errored with StatusMonitor.

I think this can be improved when we start listening for individual pool connection exits. We will then be able to count the number of alive pool connections and update the pool status accordingly.

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: So that will mean and exponential backoff will not be reset by database that's not being written to. So if it loses connection before it gets a write it may have to wait a while before reconnecting.

Also, potentially the step names could be clearer:
`:waiting_for_streaming_to_start` -> `:awaiting_first_change`
Just because it's a bit surprising that it may never "start" leaving you to think something is wrong.

alco: One thing that helps us here is the keepalive message, it gets streamed regularly when writes happen inside Postgres even if not related to the published tables.

I'm done a couple of renamings. As for resetting the backoff state, I guess we can achieve that with a timer that, when it runs out, will mark the replication connection as succeeded.

PR #2653, opened 2025-04-29
elixir-client: Include old values in change message

magnetised:


.changeset/lemon-numbers-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/elixir-client": patch
```
msfstef, line 2: nit: technically a minor change since we're adding a feature? not sure how we're following versioning now

packages/elixir-client/lib/electric/client/message.ex
```diff
@@ -104,24 +105,36 @@ defmodule Electric.Client.Message do
...
       }
     end
+
+    defp old_value(%{"old_value" => old_value}, value_mapping_fun) when is_map(old_value) do
+      map_values(old_value, value_mapping_fun)
+    end
+
+    defp old_value(_msg, _value_mapping_fun), do: nil
+
+    @compile {:inline, map_values: 2}
```
msfstef, line 124: nice trick, added to the arsenal! is there measurable performance benefit from this though? (just worried about any potentially odd behaviours this might produce?)

msfstef (approved):
Nice!

.changeset/lemon-numbers-rhyme.md
magnetised: hmm. good point. forgot we're 1.0 now. will update to a minor

PR #2654, opened 2025-04-29
feat(sync-service): Friendly error messages when failing to connect to the database

robacourt:


icehaunter (approved):
LGTM, nice work! Great use of significant children for this. Just a couple of nitpicks around tests, code structure seems fine.

icehaunter (approved):

.changeset/strange-walls-tease.md
```diff
@@ -0,0 +1,8 @@
+---
+"@core/sync-service": patch
+---
+
+Electric will now shut down and provide an error message if there is a critical error connectecting to the database
```
alco, line 5:
```suggestion
Electric will now shut down and provide an error message if there is a critical error connecting to the database
```

alco (commented):
Great work!

I only have one nitpick with regards to the phrasing "Timeout waiting...". It's not super clear to me whether it describes a fact or an ongoing process.

In the case of  "Timeout waiting for Postgres lock acquisition", what is actually timed out here? The client's request? But it does get a response on time, it just happens to be an error response. And inside the server nothing has timed out, it will keep waiting for the lock to be acquired indefinitely.

I think dropping "timeout" and starting every one of those messages with "Waiting for..." removes the ambiguity.

robacourt:
> Great work!
> 
> I only have one nitpick with regards to the phrasing "Timeout waiting...". It's not super clear to me whether it describes a fact or an ongoing process.
> 
> In the case of "Timeout waiting for Postgres lock acquisition", what is actually timed out here? The client's request? But it does get a response on time, it just happens to be an error response. And inside the server nothing has timed out, it will keep waiting for the lock to be acquired indefinitely.
> 
> I think dropping "timeout" and starting every one of those messages with "Waiting for..." removes the ambiguity.

It's a timeout in that we wait for 5 seconds and if the stack's not ready in that time we return the error. "Waiting for" doesn't sound like an error to me, but maybe I'm missing something. For now I'll merge this PR so it's not holding up your PR, and I'll ask the question in the team chat about copy. Thanks for the review!

PR #2655, opened 2025-04-29
fix: load instance_id from opts for configuring application_telemetry

balegas:
`instance_id` is required in telemetry_opts but in application_telemetry supervision tree we load it from Electric (supervision tree?) configuration, which is not started with the top-level application in Cloud.

```elixir
defp telemetry_opts(opts) do
    [
      instance_id: Keyword.fetch!(opts, :instance_id),
     ...
    ]
  end
```
This was causing issues when we enable the call_home_telemetry in Cloud since it emits some metrics for the instance even without any tenants. Hope this is right, but I'm happy to learn if there is a better way of doing it.

### DIFF
--- a/.changeset/purple-snakes-poke.md
+++ b/.changeset/purple-snakes-poke.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: load instance_id from opts for configuring application_telemetry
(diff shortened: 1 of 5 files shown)


icehaunter (changes requested):
I think `StackTelemetry` also suffers from the same function call, and it won't work without an appropriate change there as well

balegas:
So I ended up removing ```Electric.instance_id()``` at all.

icehaunter (approved):
Yup, LGTM

Issue #2597, opened 2025-04-15
Electric goes into an infinite loop trying to serve a shape request


alco:
**Versions**
Latest `main` version is affected.

**Bug description**
Continuing on from my investigation of Postgres' walsender odd behaviour in https://github.com/electric-sql/electric/issues/2595, I killed the walsender process and tried to request a shape from Electric:
```
$ curl -i http://localhost:3000/v1/shape\?table=items&offset=-1
```
As a result, Electric has fallen into an infinite loop, consuming ~130% CPU continuously and the curl request is blocked forever waiting for a response. Looking at the Erlang processes inside Electric with observer, I can see that it's stuck inside the `Electric.ShapeCache.await_snapshot_start()` recursive function loop. [This change](https://github.com/electric-sql/electric/commit/dea3e1b54d83552ab2597d12020d395b4fc7e967#diff-613b95ec7d19b7239b13453ced0e98420c40a7dc1e22f9c4aeb3141139a07aac) introduced the possibility for Electric to end up in an infinite loop when there's no shape consumer process alive.

I've yet to describe concise reproduction steps to force Electric into this state to begin with.

davidchalifoux:
I just ran into this myself.

The app I'm making is multi-tenant, with the shapes being filtered by the user's current `org_id`. It started an infinite loop after I tested switching between organizations.

PR #2656, opened 2025-04-29
fix: Ensure shape cleanup after consumer unexpected shutdown

msfstef:
Closes https://github.com/electric-sql/electric/issues/2597

The infinite loop would occur in the case where a consumer would not longer exist but the `ShapeStatus` table still had an entry for the corresponding shape.

After going through the code and consulting with @icehaunter as well, the issue was that there are cases where the internal cleanup run by the consumer itself in the `terminate` callback would not always occur - the only surefire way to go about this is to monitor the consumers and trigger cleanups when unexpected shutdowns occur.

I initially added two failing tests covering this situation, which do indeed get fixed by introducing this monitoring + cleanup of shape consumers in the `ShapeCache`.

Technically we should be able to get rid of the cleanups in the `terminate` callback in the consumer now, but I figured that whenever the consumer can cleanup after itself it's less load on the shape cache process, and since the shape cache will only attempt a cleanup if the ETS entry is still present (i.e. the consumer failed to clean itself up), I decided to keep both in. Also for the sake of fewer changes and simpler review, I can create a new PR to clean up the unnecessary code and fix tests appropriately.

I introduced an asynchronous publication cleanup call as well, as there is no reason to block the shape cache process  which is a central one waiting for the publication update.

I also cleaned up a few things to make use of the behaviours we've defined and fixed some dialyzer/type issues in the process.

Finally, I added a 50ms wait when retrying `await_snapshot_start` - we should no longer get into this infinite loop situation since this fix is trying to address the root cause, but in case there's more edge cases we at least avoid busy waiting and taking up too much CPU time. Not adding finite retries yet however as we should never need them - let the request time out.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -7,6 +7,7 @@ defmodule Electric.ShapeCacheBehaviour do
 
   @type shape_handle :: String.t()
   @type shape_def :: Shape.t()
+  @type xmin :: non_neg_integer()
 
   @callback get_shape(shape_def(), opts :: Access.t()) ::
               {shape_handle(), current_snapshot_offset :: LogOffset.t()} | nil
@@ -456,7 +457,7 @@ defmodule Electric.ShapeCache do
   end
 
   defp start_shape(shape_handle, shape, state, otel_ctx \\ nil) do
-    with {:ok, _pid} <-
+    with {:ok, pid} <-
            Electric.Shapes.DynamicConsumerSupervisor.start_shape_consumer(
              state.consumer_supervisor,
              stack_id: state.stack_id,
@@ -474,6 +475,7 @@ defmodule Electric.ShapeCache do
              create_snapshot_fn: state.create_snapshot_fn,
              otel_ctx: otel_ctx
            ) do
+      ref = Process.monitor(pid)
       consumer = Shapes.Consumer.name(state.stack_id, shape_handle)
       {:ok, latest_offset} = Shapes.Consumer.initial_state(consumer)
       {:ok, latest_offset}
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -73,6 +73,10 @@ defmodule Electric.ShapeCache do
             ]
           )
 
+  defguardp is_expected_consumer_shutdown?(reason)
+            when reason in [:normal, :killed, :shutdown] or
+                   (is_tuple(reason) and elem(reason, 0) == :shutdown and tuple_size(reason) == 2)
+
   def name(stack_id) when not is_map(stack_id) and not is_list(stack_id) do
     Electric.ProcessRegistry.name(stack_id, __MODULE__)
   end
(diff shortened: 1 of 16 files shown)


packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -150,7 +156,8 @@ defmodule Electric.ShapeCache.ShapeStatus do
         state.shape_meta_table,
         [
           {{{@shape_meta_data, shape_handle}, :_, :_, :_, :_}, [], [true]},
-          {{{@shape_hash_lookup, :_}, shape_handle}, [], [true]}
+          {{{@shape_hash_lookup, :_}, shape_handle}, [], [true]},
+          {{{@shape_consumer_ref_lookup, :_}, shape_handle}, [], [true]}
```
icehaunter, line 160: This might compound quite harshly actually on a restart - and old code does that too, but for old code we can (& should) pass in the hash to match on full key.

What I mean is in case of mass consumer death, we have N partial table walks. If we have like a million shapes eventually, that's a million items in ETS to walk for every shape removed. I think we can allow to pass in a ref for a "good" cleanup, and only do a walk if it's not provided.

msfstef: Yeah I was concerned about how these are getting deleted - but the suggestion of passing the ref I think is a good compromise (only available when cleaned up by the monitoring process though)

An alternative would be to simply have a separate `remove_consumer_ref` method that always gets called on `:DOWN` messages - it won't be in the same transaction as removing the entire shape but it would be cheap.

robacourt (changes requested):
I think rather than overloading shape cache/status, we need a new process that handles cleanup. All cleanup is done there (not in consumer). It has it's own map of refs to shape handle etc. That seems much simpler to me.

msfstef:
Based on a discussion with @robacourt I've created a separate `ShapeCleaner` process that handles the cleanups, and opened a followup PR https://github.com/electric-sql/electric/pull/2663 that removes the cleanup logic from everywhere else.

packages/sync-service/lib/electric/shapes/shape_cleaner.ex
```diff
@@ -0,0 +1,129 @@
...
+  end
+
+  @impl true
+  def handle_call({:monitor_shape, shape_handle}, _from, state) do
+    # monitor the consumer to ensure we clean up killed shapes
+    shape_consumer_pid = Electric.Shapes.Consumer.whereis(state.stack_id, shape_handle)
+    ref = Process.monitor(shape_consumer_pid)
+
+    {:reply, :ok,
+     %{state | ref_to_shape_handle: Map.put(state.ref_to_shape_handle, ref, shape_handle)}}
```
robacourt, line 73: You can avoid needing `ref_to_shape_handle` if you tag the monitor with the shape_handle:
```elixir
Process.monitor(shape_consumer_pid, tag: {:consumer_down, shape_handle})
```
then
```elixir
def handle_info({{:consumer_down, shape_handle}, _ref, :process, _pid, reason}, state)
```

Docs [here](https://www.erlang.org/doc/apps/erts/erlang.html#monitor/3:~:text=Reference%20Manual.-,%7Btag%2C%20UserDefinedTag%7D,--%20Replace%20the%20default)

Just realised this after we last spoke!

packages/sync-service/lib/electric/shapes/shape_cleaner.ex
```diff
@@ -0,0 +1,129 @@
...
+            when reason in [:normal, :killed, :shutdown] or
+                   (is_tuple(reason) and elem(reason, 0) == :shutdown and tuple_size(reason) == 2)
+
+  @impl true
+  def handle_info({:DOWN, ref, :process, _pid, reason}, state)
+      when not is_expected_consumer_shutdown?(reason) do
+    case Map.get(state.ref_to_shape_handle, ref) do
+      nil ->
+        # shape already cleaned up by consumer itself
+        {:noreply, state}
```
robacourt, line 86: This can only happen if `:DOWN` being called twice, right? Which shouldn't happen. The monitor tag I mention in another comment should nicely remove this case.

robacourt (approved):
Looking much nicer, thanks for separating the ShapeCleaner! Just left a couple of tiny nits.

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -171,7 +172,8 @@ defmodule Electric.ShapeCache do
           :exit, {:noproc, _} ->
             # The fact that we got the shape handle means we know the shape exists, and the process should
             # exist too. We can get here if registry didn't propagate registration across partitions yet, so
-            # we'll just retry.
+            # we'll just retry after waiting for a short time to avoid busy waiting.
+            Process.sleep(50)
```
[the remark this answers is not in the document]
msfstef: The 50ms delay is mostly there to avoid something getting stuck here hogging all of the CPU (as what was observed was busy waiting as all checks were in-memory ones) - the assumption I've made here is that eventually the request would be cancelled.

I don't think this should have any side effects like booting up shapes and consumers if they aren't up, as we do not expect this to fail ever other than small timing issues as @icehaunter discovered and led him to add this loop here, so we could just have a number of limited attempts and throw an appropriate error?

I avoided adding it in this PR as I mostly wanted to 1. add tests for reproducing the error and 2. address what I thought was the root cause.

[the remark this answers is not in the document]
alco: Sounds good, thanks for elaborating.

PR #2660, opened 2025-04-29
fix: correctly handle an error while schema recon fails

icehaunter:


balegas (approved):
I trust that the failure is a benign errors, but I can't stop being curious about how the system recovers from that :).

packages/sync-service/lib/electric/replication/schema_reconciler.ex
```diff
@@ -82,8 +82,11 @@ defmodule Electric.Replication.SchemaReconciler do
       :ok
     else
       {:error, reason} ->
-        Logger.error("Schema reconciliation failed: #{reason}")
+        Logger.warning("Schema reconciliation failed: #{reason}")
         :error
+
+      :error ->
+        Logger.warning("Schema reconciliation failed while fetching fresh relations")
```
msfstef, line 89: If we're downgrading this to a warning I agree with Valter that a short explanation of why this is potentially expected? How would we detect if it _always_ fails and never works as opposed to failing because of some Postgres connection error or something?

msfstef (approved):
+1 for handling all error cases but error-level visibility on these errors is something we might want to retain?

packages/sync-service/lib/electric/replication/schema_reconciler.ex
icehaunter: There's a log inside the ETSCache listing with failure details, so log context would be the answer

PR #2662, opened 2025-04-30
fix: Add retries on shape file cleanup operation to prevent `EEXIST` race

msfstef:
We observed the `unsafe_cleanup!` call on shape files failing with an `:eexist` reason on the `File.rm_rf!` calls, which didn't make much sense.

Looking at the Elixir source we suspect that the issue is `rm_dir` failing due to some race occurring:
https://github.com/elixir-lang/elixir/blob/e35ffc5a903bff3b595e323eb1ac12c4ecd515ad/lib/elixir/lib/file.ex#L1308-L1312

Here's a discussion with Gemini about this as well https://gemini.google.com/share/d36919a2b507

Potentially the directory is assumed to be empty after files have been deleted but something causes it to not actually be fully empty, especially on weirder filesystems like AWS' EFS which is networkd and distributed.

This PR suggests adding some arbitrary retries with small waits in an attempt to mitigate this issue - I'm open to suggestions about different wait times (if any, potentially we just retry and since they are syscalls they already will take some time) and retry attempts.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/file_storage.ex
@@ -709,7 +709,6 @@ defmodule Electric.ShapeCache.FileStorage do
       # due to some FS race the deletion fails with EEXIST. Very hard to test
       # and prevent so we mitigate it with arbitrary retries.
       {:error, :eexist, _} when attempts_left > 0 ->
-        Process.sleep(1)
         unsafe_cleanup_with_retries!(opts, attempts_left - 1)
 
       {:error, reason, path} ->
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/shape_cache/file_storage.ex
```diff
@@ -695,10 +695,26 @@ defmodule Electric.ShapeCache.FileStorage do
...
+      # There is a very unlikely but observed scenario where the rm_rf call
+      # tries to delete a directory after having deleted all its files, but
+      # due to some FS race the deletion fails with EEXIST. Very hard to test
+      # and prevent so we mitigate it with arbitrary retries.
+      {:error, :eexist, _} when attempts_left > 0 ->
+        Process.sleep(1)
+        unsafe_cleanup_with_retries!(opts, attempts_left - 1)
+
+      err ->
+        raise "Failed to clean up shape data: #{inspect(err)}"
```
icehaunter, line 716: nitpick: I would rather raise the actual error instead of a `RuntimeError` wrapper: 
```suggestion
      {:error, reason, path} ->
        raise File.Error,
          reason: reason,
          path: path,
          action: "remove files and directories recursively from"
```
in order to be consistent with actual raises from these functions

icehaunter (approved):

robacourt (approved):
LGTM

PR #2672, opened 2025-05-05
Prevent errors when terminating an active shape

magnetised:
This fixes a few race/error conditions that arise when deleting an active shape.

- There's a race condition in the shape status which can lead to exceptions
- This adds a count of active readers on a shape and holds the shape processes live until they've all terminated, otherwise shapes can terminate while clients are reading the data resulting in a 500
- Shape data cleanup is done by a separate async task that only begins once the shape supervisor has terminated

Incidentally I've also enabled the use of a custom connection pool for the elixir client.

I will have a separate pr with the shape-creation/deletion churn generator


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -101,6 +101,11 @@ defmodule Electric.Replication.ShapeLogCollector do
     }
   end
 
+  def terminate(_reason, state) do
+    LsnTracker.terminate(state.stack_id)
```
msfstef, line 105: is this necessary? the table is owned by the process so would go away with it no? has this caused any issues?

packages/sync-service/lib/electric/shape_cache/file_storage.ex
```diff
@@ -695,28 +696,8 @@ defmodule Electric.ShapeCache.FileStorage do
...
-      {:error, :eexist, _} when attempts_left > 0 ->
-        unsafe_cleanup_with_retries!(opts, attempts_left - 1)
-
-      {:error, reason, path} ->
-        raise File.Error,
-          reason: reason,
-          path: path,
-          action: "remove files and directories recursively from"
-    end
+    {:ok, _} = File.rm_rf(opts.data_dir)
```
msfstef, line 699: is there a reason this gets rid of the retries? the EEXIST error occurred on EFS with an rm_rf call which is why that was put in place, what was the reason for removing it?

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -149,8 +151,8 @@ defmodule Electric.ShapeCache.ShapeStatus do
       :ets.select_delete(
         state.shape_meta_table,
         [
-          {{{@shape_meta_data, shape_handle}, :_, :_, :_, :_}, [], [true]},
-          {{{@shape_hash_lookup, :_}, shape_handle}, [], [true]}
+          {{{@shape_hash_lookup, :_}, shape_handle}, [], [true]},
```
msfstef, line 154: nit: since we now have the shape itself, we can recompute the hash to perform a cheaper delete on the shape hash lookup.

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -421,8 +435,6 @@ defmodule Electric.Shapes.Api do
           serve_shape_log(request)
       end
 
-    clean_up_change_listener(request)
```
msfstef, line 424: this was here because we might never encode a response message (e.g. `if_not_modified` is true and therefore we  return a 304 without any messages) - how are you ensuring this is always cleaned up now?

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -593,20 +615,48 @@ defmodule Electric.Shapes.Api do
   def stack_id(%Api{stack_id: stack_id}), do: stack_id
   def stack_id(%{api: %{stack_id: stack_id}}), do: stack_id
 
-  defp encode_log(%Request{api: api}, stream) do
-    encode(api, :log, stream)
+  # append the cleanup operations onto the end of the stream for every response
+  # by concating a dummy stream with only an `after` function. by making it
+  # part of consuming the response stream rather than an explicit call we
+  # simplify the API as it cleans up after itself when the response is read
```
msfstef, line 621: I disagree with the assumption that `unsubscribe_after_stream` always gets called - we have potential responses that never go through this path (e.g. 304 not modified). I don't think we are truly simplifying the API if we are explicitly registering listeners and then implicitly cleaning them up.

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -334,6 +335,21 @@ defmodule Electric.Shapes.Api do
...
   end
 
+  defp register_shape_subscriber(%Request{handle: nil} = request) do
+    request
+  end
+
+  defp register_shape_subscriber(%Request{} = request) do
+    %{api: %{stack_id: stack_id}, handle: handle} = request
+    # The assumption here is that all api requests live in a transient process
+    # that will exit once the request is complete.
```
msfstef, line 345: This assumption does not hold, which is why the explicit cleanup calls were added - Bandit reuses the same process for [all requests on the same connection](https://github.com/mtrudel/bandit/tree/main/lib/bandit/http2#process-model) and @robacourt had observed issues arising from the fact that requests seemed to be reusing the same process.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -375,11 +399,9 @@ defmodule Electric.Shapes.Consumer do
           "Truncate operation encountered while processing txn #{txn.xid} for #{shape_handle}"
         )
 
-        state
-        |> cleanup()
-        |> notify_shape_rotation()
+        cleanup(state, {:shutdown, :truncate})
```
msfstef, line 402: the resulting updated state is ignored - is that ok in this case? feels wrong to need to propagate it only some of the time?

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -368,32 +372,51 @@ defmodule Electric.ShapeCache do
...
   end
 
-  defp clean_up_all_shapes(state) do
-    shape_handles =
-      state.shape_status_state |> state.shape_status.list_shapes() |> Enum.map(&elem(&1, 0))
+  defp purge_shape(state, shape_handle) do
+    deregister_shape(shape_handle, state)
+
+    # Cleanup the storage for the shape
+    shape_handle
```
msfstef, line 403: should this not occur via the shape monitor in some way? having shape cleanup logic spread out over various processes feels avoidable, and shape cache in particular would benefit from avoiding IO operations since it is on the critical request path

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -488,17 +510,53 @@ defmodule Electric.Shapes.Consumer do
     %{state | pg_snapshot: pg_snapshot}
   end
 
-  defp cleanup(state) do
+  # cleanup is now done in stages.
+  # 1. register that we want the shape data to be cleaned up.
+  # 2. request a notification when all active shape data reads are complete
+  # 3. exit the process when we receive that notification
+  defp cleanup(state, reason \\ :normal) do
```
msfstef, line 517: The work in https://github.com/electric-sql/electric/pull/2656 adds a couple of tests and guards against the situation where essentially the shape consumer dies without being able to clean itself up - since the `ShapeStatus` ETS table is owned by `ShapeCache` and its state is crucial for how the API behaves, having the cleanup guaranteed by a monitoring process that is somehow related to the `ShapeCache` supervision-wise would be safer (e.g. the CleanupTask you have could also be ensuring the shape is not in the ShapeStatus table)

msfstef (changes requested):
Awesome work - left some questions about some of the changes

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: I was trying to get `mix test ... --repeat-until-failure` to work but it fails eventually with a duplicate table name error, i.e. the next test starts before the previous test has finished cleaning up. this change didn't actually fix that so I could get rid of this.

packages/sync-service/lib/electric/shape_cache/file_storage.ex
magnetised: i want to assert that the rm_rf is now working without issue now we're waiting for process termination before deleting things. i don't want to hide that error because it shows that something is not working  -- the retry was a hack, this pr is a fix, or at least it should be

packages/sync-service/lib/electric/replication/shape_log_collector.ex
msfstef: Ah yes I've seen that failure - it would be surprising cause all tests should have unique table names - most likely some of the tests that restart processes (like restarting the shape cache) within the _same_ test are at fault

yeah I would remove this termination as it is unnecessary, I remember @robacourt working on this and removing terminations explicitly because of this

packages/sync-service/lib/electric/shapes/api.ex
magnetised: i've wrapped the empty response in the cleanup stream.

packages/sync-service/lib/electric/shape_cache/file_storage.ex
msfstef: right - but did you manage to reproduce that error before with your testing and then resolve it with your changes? cause it most likely was FS related rather than process related - `rm_rf` should never raise `EEXIST` so it's more like a freak error. Even if the process is dead the FS might not have flushed things and we might still run into it.

packages/sync-service/lib/electric/shapes/api.ex
magnetised: the thing is that the cleanup has to happen after the response body has been read, not when we return the response body stream from the api. if we just de-register when returning the response then the active client count will be wrong because we'll have reduced the client count but still be streaming data out. i had missed terminating the 304 response -- good catch, but i don't see a neater way of doing this without explicit (and not obvious) callbacks that hook into the actual server mechanism (there's no plug after_body_read hook).

packages/sync-service/lib/electric/shapes/api.ex
magnetised: yeah, i wrote that then realised it was wrong and forgot to delete it

packages/sync-service/lib/electric/shapes/api.ex
msfstef: By empty response do you mean `no_change_response` by virtue of the `encode_log` call? The 304 from `if_not_modified` does not call `no_change_response` since it does not include any encoded body - am I missing something?

packages/sync-service/lib/electric/shapes/api.ex
msfstef: Ah I had not considered the streaming aspect - that makes a lot of sense and in that case we _have_ to chain it to that... in that case even though I don't like the implicit nature of it we should ensure all responses are generated with a function that includes the cleanup (since there's no after body read hook as you mentioned, not sure if we can just implement a small one?)

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: I'm doing a belt-and-braces thing. The consumer removes itself from the shape status asap if possible -- it's a good idea to remove the shape from the active path asap -- but the async cleanup also does it.

msfstef: yep I was going to say that after your issue this morning in standup - most things rely on the in-memory status changing so that should happen ASAP, I'm assuming doing that fixed a lot of tests as well

packages/sync-service/lib/electric/shapes/api.ex
magnetised: in my recent commits i've wrapped the empty body of the 304 response in the termination handling stream

packages/sync-service/lib/electric/shape_cache/file_storage.ex
magnetised: after all that I've restored the retry as the :eexist issue is happening in tests.

packages/sync-service/lib/electric/lsn_tracker.ex
```diff
@@ -28,10 +28,17 @@ defmodule Electric.LsnTracker do
...
-    |> :ets.new([:protected, :named_table])
+    table = table(stack_id)
+
+    case :ets.info(table, :id) do
+      :undefined ->
+        :ets.new(table, [:protected, :named_table])
+
+      ref when is_reference(ref) ->
+        :ok
+    end
```
robacourt, line 41: Why is this needed? I would say you want this to error if it's already got created because that means there's an issue with your test

msfstef: +1, I imagine the errors in tests are because of individual process restarts within the same test, I would try to isolate the cases and fix the test helpers/setup

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -198,9 +209,12 @@ defmodule Electric.Shapes.Consumer do
           error
       end
 
-    state = reply_to_snapshot_waiters(state, {:error, error})
-    cleanup(state)
-    {:stop, :normal, state}
+    state =
+      state
+      |> reply_to_snapshot_waiters({:error, error})
+      |> terminate_safely()
```
msfstef, line 215: technically this now also notifies about shape rotations which is new behaviour for this case - but callers would be waiting for the snapshot to start so they wouldn't reach the receive call for the shape rotation notification - I assume that was the reasoning?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -120,17 +130,17 @@ defmodule Electric.Shapes.Consumer do
...
+  def handle_call(:stop_and_clean, _from, state) do
     # Waiter will receive this response if the snapshot wasn't done yet, but
     # given that this is definitely a cleanup call, a 409 is appropriate
     # as old shape handle is no longer valid
+
     state =
-      reply_to_snapshot_waiters(state, {:error, Api.Error.must_refetch()})
+      state
+      |> reply_to_snapshot_waiters({:error, Api.Error.must_refetch()})
+      |> terminate_safely()
```
msfstef, line 141: shape rotation is notified - this is new behaviour but I assume this is intended? Any requests waiting on data should be notified about this

packages/sync-service/lib/electric/stack_supervisor.ex
```diff
@@ -327,13 +338,25 @@ defmodule Electric.StackSupervisor do
...
+
     children =
       [
         {Electric.ProcessRegistry, partitions: registry_partitions, stack_id: stack_id},
         {Registry,
          name: shape_changes_registry_name, keys: :duplicate, partitions: registry_partitions},
         {Electric.Postgres.Inspector.EtsInspector, stack_id: stack_id, pool: db_pool},
-        {Electric.Connection.Supervisor, new_connection_manager_opts}
+        {Electric.Connection.Supervisor, new_connection_manager_opts},
+        {Electric.Shapes.Monitor,
```
msfstef, line 354: shouldn't this live _before_ the connection supervisor? I haven't followed all of the code fully yet but shape cache and conumers rely on the monitor being up (?)

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
```diff
@@ -0,0 +1,89 @@
...
+    if consumer_alive?(stack_id, shape_handle) do
+      {:error, "Expected shape #{shape_handle} consumer to not be alive before cleaning shape"}
+    else
+      {publication_manager, publication_manager_opts} = publication_manager_impl
+      {shape_status, shape_status_state} = shape_status_impl
+
+      Logger.debug("cleaning shape data #{inspect(shape_handle)}")
+
+      shape_status.remove_shape(shape_status_state, shape_handle)
+      publication_manager.remove_shape(shape_handle, shape, publication_manager_opts)
```
msfstef, line 76: this is currently a `GenServer.call` - in my PR I had converted this to a cast so as to not delay the data removal. My worry is that if the publication update fails because of e.g. PG connection issues, this call fails and we don't reach the data cleanup.

Either adding an async call for removing shapes or potentially rescuing failed calls to ensure the data cleanup also occurs would be nice I think

packages/sync-service/lib/electric/lsn_tracker.ex
magnetised: If you run `mix test --repeat-until-failure 1000` then you get exceptions when it tries to create an already existing table. we have tests that restart processes that end up resetting the replication stream somehow, then calling `ShapeLogCollector.start_processing`without terminating the existing process which then attempts to re-create an existing table.

packages/sync-service/lib/electric/shape_cache/file_storage.ex
```diff
@@ -234,39 +239,62 @@ defmodule Electric.ShapeCache.FileStorage do
...
+      opts.db
+      |> validate_db_process!()
+      |> CubDB.has_key?(@snapshot_started_key)
+    rescue
+      Storage.Error -> false
+    end
+  end
+
+  defp validate_db_process!(name) do
+    if pid = GenServer.whereis(name) do
```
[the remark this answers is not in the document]
magnetised: it's just doing explictly what cubdb/genserver would do anyway -- if you're calling a named process it's an ets lookup (GenServer.whereis) then a call to the resolved pid - so basically i'm doing that by hand so-to-speak, so I can validate the pid exists.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -583,6 +593,10 @@ defmodule Electric.Shapes.Consumer do
     ]
   end
 
+  defp calculate_replication_lag(%Transaction{commit_timestamp: nil}) do
+    0
```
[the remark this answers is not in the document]
magnetised: think it has to be a number or we'd get telemetry crashes -- this only happens in tests so I just wanted to do the least odd thing

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -368,32 +373,54 @@ defmodule Electric.ShapeCache do
...
-    after
-      # another failsafe for ensuring data is cleaned if consumer fails to
-      # clean itself - this is not guaranteed to run in case of an actual exit
-      # but provides an additional safeguard
+    # remove the shape immediately so new clients are redirected elsewhere
+    deregister_shape(shape_handle, state)
+
+    # Rather than just stop here, tell the consumer to stop
+    # using the shape.status active connection monitor
+    # to wait for clients to finish
```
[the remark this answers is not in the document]
magnetised: confuses me too! don't think it adds any value

packages/sync-service/lib/electric/lsn_tracker.ex
robacourt: It would be better to fix those specific tests. They'll be running two `ShapeLogCollector`s with the same stack ID so they're probably still flakey

packages/sync-service/lib/electric/shape_cache/file_storage.ex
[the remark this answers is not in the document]
msfstef: sounds good - wasn't familiar with GenServer internals as long as it is an in memory lookup it's whatever

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
magnetised: what we really need is some persistent queue for these things.... deletion of data should happen even if we restart the system

msfstef: it _will_ happen, eventually, due to us being essentially a cache and having garbage collection. Not against persisting the "delete" requests though, might be quite simple to do, but definitely not a priority I think.

msfstef:
This closes https://github.com/electric-sql/electric/issues/2597 - I've closed https://github.com/electric-sql/electric/pull/2656 in favour of this PR.

packages/sync-service/lib/electric/stack_supervisor.ex
magnetised: true, in reality the connection supervisor startup is slow compared to the monitor, but yes.

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: i had a version that aborted downloads when the shape was terminated but got rid of that because it caused json decoding errors in the clients, which was noisy in terms of "no errors". in reality i don't think it matters if we wait for readers to terminate, but happy to add it back in if we're happy with exceptions on the client-side

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: i don't see how there's any new behaviour here tbh, apart from the consumer not stopping immediately in this case

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -368,32 +373,54 @@ defmodule Electric.ShapeCache do
...
 
-    :ok
+    state
   end
 
-  defp clean_up_all_shapes(state) do
-    shape_handles =
-      state.shape_status_state |> state.shape_status.list_shapes() |> Enum.map(&elem(&1, 0))
+  defp purge_shape(state, shape_handle, shape) do
+    case Electric.Shapes.ConsumerSupervisor.stop_and_clean(state.stack_id, shape_handle) do
```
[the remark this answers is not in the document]
magnetised: the consumer is not now exiting unless there's some error (it used to exit after the clean call). I will have a look though -- we can't just schedule a clean up because the whole point is that the cleanup only happens once the shape is un-referenced. the only other option would be to expand the cleanup task to do the reference following outside the consumer completely. i've never seen this call raise in my very long running tests but we could add that

packages/sync-service/lib/electric/lsn_tracker.ex
magnetised: i don't think that's the case -- there was only one but with duplicate calls to the start_processing function. i can add an issue to clean this up if you're ok with just merging this monster with this caveat.

packages/sync-service/lib/electric/shape_cache.ex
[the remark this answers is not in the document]
msfstef: Actually nevermind since we call `register_writer` in the `init` call, and that monitors the consumer and cleans up the shape in case it exits unexpectedly, the case I'm talking about is covered.

packages/sync-service/lib/electric/shapes/consumer.ex
msfstef: I agree with not interrupting readers immediately, I'd rather we wait for a short second for cleaner behaviour.

My point was that `terminate_safely` now internally also calls `notify_shape_rotation` whereas before we did not do that. As stated in the other comments the requesters/readers would be stuck waiting for the snapshot and would terminate the request path with the `reply_to_snapshot_waiters` call, but a subsequent message would also be sent about the shape rotation which we didn't send before. Not new behaviour effectively, but just noting that a shape rotation message is now being sent from this whereas it wasn't before.

msfstef (approved):

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: ok. gotcha

packages/sync-service/lib/electric/shapes/monitor.ex
```diff
@@ -0,0 +1,122 @@
+defmodule Electric.Shapes.Monitor do
```
alco, line 1: Please include Supervisor in the module name. It's easy to get the wrong impression from the name Monitor.

alco: Alternatively, this module can be made process-less and a separate small module `Electric.Shapes.Monitor.Supervisor` can host just the supervisor bits (start_link and init).

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
```diff
@@ -0,0 +1,117 @@
...
+               )
+               |> case do
+                 :ok ->
+                   :ok
+
+                 {:error, reason} ->
+                   Logger.error(["Failed to clean shape #{shape_handle}: ", reason])
+               end
+             catch
+               exception -> Logger.error(exception)
```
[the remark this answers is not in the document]
icehaunter: It would also be cleaner to handle the error inside the `cleanup/6` call and rely on logging from an error tuple

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
```diff
@@ -0,0 +1,117 @@
...
+      |> Enum.find(:ok, &match?({:error, _}, &1))
+    end
+  end
+
+  defp consumer_alive?(stack_id, shape_handle) do
+    !is_nil(Electric.Shapes.Consumer.whereis(stack_id, shape_handle))
+  end
+
+  defp perform_async_catching_errors(fun, message) do
+    Task.async(fn ->
```
icehaunter, line 106: Why are we spawning tasks outside the supervision tree from inside the supervision tree? Maybe put them under the same supervisor?

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
```diff
@@ -0,0 +1,117 @@
...
+      remove_shape =
+        perform_async_catching_errors(
+          fn ->
+            publication_manager.remove_shape(shape_handle, shape, publication_manager_opts)
+          end,
+          "Failed to remove shape #{shape_handle} from publication"
+        )
+
+      [data_cleanup, remove_shape]
+      |> Task.await_many(60000)
```
icehaunter, line 96: Yeah, this call exits after a timeout, and the `catch` above will not handle it in it's current form. This may happen especially with `remove_shape` call for the publication manager

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
```diff
@@ -0,0 +1,117 @@
...
+        stack_id,
+        storage_impl,
+        publication_manager_impl,
+        shape_status_impl,
+        shape_handle,
+        shape,
+        on_cleanup \\ fn _ -> :ok end
+      ) do
+    with {:ok, _pid} <-
+           Task.Supervisor.start_child(name(stack_id), fn ->
```
alco, line 30: How does this work exactly? Neither the type spec not the documentation for [`Task.Supervisor.start_child()`](https://hexdocs.pm/elixir/Task.Supervisor.html#start_child/2) mention the ability to pass an anonymous function in place of a child spec.

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
alco, line 50: With this syntax only `throw`s are caught. To catch errors, throws and exits, you must explicitly match on two arguments:
```
catch
  _type, _val ->
```

The `Logger.error()` call needs explicit formatting because as is it doesn't produce pretty output:
```
iex(2)> try do
...(2)> raise "oopsie"
...(2)> catch
...(2)> :error, exception -> Logger.error(exception)
...(2)> end

13:21:52.494 [error] [message: "oopsie", __struct__: RuntimeError, __exception__: true]
```

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
alco, line 96: Why 60 seconds? This could be turned into a module attribute with an explainer comment.

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
```diff
@@ -0,0 +1,117 @@
...
+      |> Task.await_many(60000)
+      |> Enum.find(:ok, &match?({:error, _}, &1))
+    end
+  end
+
+  defp consumer_alive?(stack_id, shape_handle) do
+    !is_nil(Electric.Shapes.Consumer.whereis(stack_id, shape_handle))
+  end
+
+  defp perform_async_catching_errors(fun, message) do
```
alco, line 105: It looks odd to be using `Task.async` and catching errors inside the spawned task. `Task.Supervisor.async_nolink` is the go-to way to be able to await a task but isolate the caller process from errors.

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    # log some debug stats
+    # send(self(), :log_active)
+
+    {:ok, state}
+  end
+
+  @impl GenServer
+  def handle_call({:register_reader, handle, pid}, _from, state) do
```
icehaunter, line 155: I'm worried about this module and this `GenServer.call`.

We're calling `register_reader` for every incoming (valid) HTTP request. That means we're bottle-necking entire HTTP throughput on this single GenServer's mailbox. It's also essentially bottle-necked twice on throughput, because our clients are guaranteed to be transient tasks which will be `DOWN`, so we get at least 2 messages per HTTP connection into this process'es mailbox

What's more, we're storing reference data and handles in `:ets`, which is then not used for quick out-of-process reads: only read I can see here is `statistics` which is relatively rare. So we don't get any ETS parallelization benefits, but we do get out-of-process memory copy in/out drawback.

packages/sync-service/test/electric/shapes/monitor_test.exs
```diff
@@ -0,0 +1,462 @@
...
+        )
+
+      assert_receive :ready, 100
+      Monitor.notify_reader_termination(stack_id, handle1, :my_reason)
+      send(subscriber1, {:subscribe, handle1})
+      refute_receive {Monitor, :reader_termination, _handle1, _}, 100
+    end
+  end
+
+  # defp wrap_storage(%{storage: storage}) do
```
alco, line 378: Leftover by mistake?

packages/sync-service/lib/electric/stack_supervisor.ex
```diff
@@ -86,7 +87,15 @@ defmodule Electric.StackSupervisor do
                  storage: [type: :mod_arg, required: true],
                  chunk_bytes_threshold: [
                    type: :pos_integer,
-                   default: Electric.ShapeCache.LogChunker.default_chunk_size_threshold()
+                   default: LogChunker.default_chunk_size_threshold()
+                 ],
+                 monitor_opts: [
```
icehaunter, line 92: So these aren't used in actual code, and aren't functionally part of public API of stack supervisor, as callbacks are scarcely actionable (e.g. they don't even specify what got removed, reader or writer). Since these are test-only, they should probably live under `tweaks` key

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
icehaunter: What do you mean, it's used like that in the docs:
```
Task.Supervisor.start_child(MyTaskSupervisor, fn ->
  IO.puts("I am running in a task")
end)
```

packages/sync-service/lib/electric/shape_cache/file_storage.ex
```diff
@@ -234,39 +241,62 @@ defmodule Electric.ShapeCache.FileStorage do
...
+      pid
```
icehaunter, line 289: This is again prone to a race between a `whereis` and the found process dying. Agree that this is a rare condition in this particular instance, but on a shutdown sequence it's possible. We need to catch the missing process at the call point, not before to be fully correct. I also don't see what error relabeling gives us. We have a raised error instead of a `:noproc` exit, but what's the benefit?

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
alco: Ah, my bad, I ended up on the wrong doc page inadvertently.

The `with` expression threw me off here. We're not supposed to get an error return value from the `start_child()` call here, an `{:ok, pid}` should be returned as soon as a new process is started, regardless of how the function execution goes. It's better to assert `{:ok, _pid} = Task.Supervisor.start_child()` here than rely on the caller performing a match if we want to fail when a cleanup process spawning fails.

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
magnetised: Yeah, I had the same concern but in reality it doesn't seem to be a problem. The calls are quick and our systems are designed to be put behind a CDN so the actual load isn't high. My stress tester, which drove most of these fixes, was running 400 simultaneous connections with a high churn - clients would be killed after ~2s and shapes deleted after ~2s - and there were no issues with slowdowns on the registration calls. 

(There are 2 calls to this genserver, but they're not usually due to :DOWN messages - most connection processes are persistent and re-used by Bandit - but instead de-registration at the end of the request.)

Having said that, looking at the final version of the code, it would work as a pure ets update function (except in the case of a re-register of the same pid under a new handle, which is an edge case anyway), which I don't think was true for most of its lifetime.

integration-tests/run.sh
```diff
@@ -1,10 +1,16 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
```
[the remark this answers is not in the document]
magnetised: oops!

packages/sync-service/test/electric/shapes/monitor_test.exs
magnetised: yeah, api changes...

packages/sync-service/lib/electric/shape_cache/file_storage.ex
magnetised: The re-labeling is just to make clearer the problem in exceptions. It's not perfect as you say but it will make it easier to diagnose problems trying to read/write to storage that's no longer running.

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
magnetised: I started with ets with the goal of making these registrations parallelisable but in the end evolved a design that didn't work purely with the ets version. However I think I can make many aspects of the process parallelisable or at least async keeping the ets

magnetised: no. I was wrong about that. there's no way of really speeding things up using parallel writes to the ets because of the need to monitor the process. using an ets does reduce gc churn though.

icehaunter:
Benchmark this

icehaunter:
Benchmark this

magnetised:
Benchmark this

magnetised:
Benchmark this

magnetised:
Benchmark this

magnetised:
Benchmark this

magnetised:
Benchmark this

PR #2675, opened 2025-05-06
Add default issue types in issue templates

alco:
Not sure this is how it works but we can at least try.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -3,6 +3,7 @@ name: Bug report
 about: Create a report to help us improve
 title: ''
 labels: bug, triage
+type: bug
 assignees: ''
 
 ---
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -3,7 +3,7 @@ name: Bug report
 about: Create a report to help us improve
 title: ''
 labels: bug, triage
-type: bug
+type: Bug
 assignees: ''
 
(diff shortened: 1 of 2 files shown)


KyleAMathews (approved):
I'm not sure how this works either!

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -3,6 +3,7 @@ name: Bug report
 about: Create a report to help us improve
 title: ''
 labels: bug, triage
+type: bug
```
msfstef, line 6:
```suggestion
type: Bug
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -3,6 +3,7 @@ name: Feature request
 about: Suggest an idea for this project
 title: 'Feature request: '
 labels: feature request, triage
+type: feature
```
msfstef, line 6:
```suggestion
type: Feature
```

msfstef (changes requested):
Looking at [the relevant docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#top-level-syntax) this is indeed the right top-level key.


> The issue type that will be automatically added to issues created with this template. Issue types are defined at the organization level and can be used to create a shared syntax across repos.

However looking at our existing issue types, we seem to have them capitalised, and since the `type` is any string and when I filter on our existing types it uses the capitalised version, I think we need to specify it like that

msfstef (approved):

PR #2677, opened 2025-05-06
Fix issues in Prometheus metric collection

### LINKED ISSUE #2674: Our custom measurements of system metrics erase system measurements
**Versions**
The latest `main` is affected. The main issue was introduced by https://github.com/electric-sql/electric/commit/7600746eb95616fd3e6d8ba142c788ac70157f70 on March 3. Other metrics may also be malfunctioning in the same manner.

**Bug description**
We use the `telemetry_poller` library to define a bunch of [telemetry](https://hexdocs.pm/telemetry/readme.html) metrics that are gathered at regular intervals (polling).

The problem lies in the custom measurements that we define for a couple of metrics. The biggest culprit is the `vm.memory` metric for which we add a measurement to capture memory usage of different process types:

https://github.com/electric-sql/electric/blob/c8f2100c59591fa817795d4f78671bbbcf60f66a/packages/sync-service/lib/electric/telemetry/application_telemetry.ex#L236-L238

Most of the measurements we poll for are of type `last_value()` and those measurements literally save the last value that was reported for any given metric. So the above `:telemetry.execute()` call preserves only the `processes_by_type` measurement for the `vm.memory` metric, while erasing the other ones like `total`, `ets` and the rest.

I noticed the problem by seeing the following log output getting dumped every poll internval:

```
17:53:12.414 pid=<0.451.0> [debug] Measurement not found, expected: total. metric_name:=[:vm, :memory, :total]
17:53:12.415 pid=<0.451.0> [debug] Measurement not found, expected: processes_used. metric_name:=[:vm, :memory, :processes_used]
17:53:12.416 pid=<0.451.0> [debug] Measurement not found, expected: binary. metric_name:=[:vm, :memory, :binary]
17:53:12.417 pid=<0.451.0> [debug] Measurement not found, expected: ets. metric_name:=[:vm, :memory, :ets]
17:53:12.417 pid=<0.451.0> [debug] Measurement not found, expected: total. metric_name:=[:vm, :memory, :total]
17:53:12.418 pid=<0.451.0> [debug] Measurement not found, expected: processes_used. metric_name:=[:vm, :memory, :processes_used]
17:53:12.418 pid=<0.451.0> [debug] Measurement not found, expected: binary. metric_name:=[:vm, :memory, :binary]
17:53:12.418 pid=<0.451.0> [debug] Measurement not found, expected: ets. metric_name:=[:vm, :memory, :ets]
17:53:12.419 pid=<0.451.0> [debug] Measurement not found, expected: total. metric_name:=[:vm, :memory, :total]
17:53:12.419 pid=<0.451.0> [debug] Measurement not found, expected: processes_used. metric_name:=[:vm, :memory, :processes_used]
17:53:12.419 pid=<0.451.0> [debug] Measurement not found, expected: binary. metric_name:=[:vm, :memory, :binary]
17:53:12.420 pid=<0.451.0> [debug] Measurement not found, expected: ets. metric_name:=[:vm, :memory, :ets]
17:53:12.420 pid=<0.451.0> [debug] Measurement not found, expected: total. metric_name:=[:vm, :memory, :total]
17:53:12.420 pid=<0.451.0> [debug] Measurement not found, expected: processes_used. metric_name:=[:vm, :memory, :processes_used]
17:53:12.421 pid=<0.451.0> [debug] Measurement not found, expected: binary. metric_name:=[:vm, :memory, :binary]
17:53:12.421 pid=<0.451.0> [debug] Measurement not found, expected: ets. metric_name:=[:vm, :memory, :ets]
```

But those logs on their own don't show the underlying problem. It becomes obvious when we augment the log message with the list of available measurements for the affected metric:

```
[debug] Measurement not found, expected: processes_by_type. metric_name:=[:vm, :memory, :processes_by_type], measurements=%{atom: 1130777, atom_used: 1107630, binary: 7863784, code: 22683862, ets: 2915016, processes: 35222440, processes_used: 35182360, system: 62048728, total: 97271168}

[debug] Measurement not found, expected: processes_used. metric_name:=[:vm, :memory, :processes_used], measurements=%{processes_by_type: 4707980}
[debug] Measurement not found, expected: binary. metric_name:=[:vm, :memory, :binary], measurements=%{processes_by_type: 4707980}
[debug] Measurement not found, expected: ets. metric_name:=[:vm, :memory, :ets], measurements=%{processes_by_type: 4707980}
[debug] Measurement not found, expected: total. metric_name:=[:vm, :memory, :total], measurements=%{processes_by_type: 4707980}
```

The first log entry above shows the correct set of measurements obtained from a call to `:erlang.memory()`. We only see the log entry about the `processes_by_type` measurement not found because at that point our custom measurement hasn't yet overriden the system ones.

The following log entries show that the map of measurements only keeps our custom `processes_by_type` measurement and so an error is logged for the other ones.

**Expected behavior**
We want measurements to be set-additive, not exclusive, so that executing a single measurement such as `vm.memory.processes_by_type` preserves the other measurements obtained from a different call to `:telemetry.execute()`.

alco:
Fixes #2674.

Trying to report out custom per-process-type memory metrics as part of telemetry poller's builtin `memory` probe was a mistake, as explained in the linked issue. This PR defines a separate event for our custom memory probes.

Now, instead of exporting `vm.memory.processes_by_type` we have the  `process.memory.total` measurement which is tagged with the process type/name. Such measurements are still reported only for the top N processes.

From what I've been able to find, the `vm.memory.processes_by_type` metric has only been referenced in Honeycomb by @robacourt's query that showcased the mem usage grouped by process type/name.

### DIFF
--- a/.changeset/happy-cats-do.md
+++ b/.changeset/happy-cats-do.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix the issue with failing to report vm memory metrics. Change the name of the "vm.memory.processes_by_type" metric to "process.memory.total".
--- a/packages/sync-service/lib/electric/plug/utility_router.ex
+++ b/packages/sync-service/lib/electric/plug/utility_router.ex
@@ -10,4 +10,6 @@ defmodule Electric.Plug.UtilityRouter do
   else
     get "/metrics", do: resp(conn, 200, "[]")
   end
+
+  get _, do: resp(conn, 404, "Not found")
 end
--- a/packages/sync-service/lib/electric/telemetry/application_telemetry.ex
+++ b/packages/sync-service/lib/electric/telemetry/application_telemetry.ex
@@ -159,10 +159,10 @@ with_telemetry [Telemetry.Metrics, OtelMetricExporter] do
 
     defp prometheus_metrics do
       [
+        last_value("process.memory.total", tags: [:process_type], unit: :byte),
         last_value("system.cpu.core_count"),
         last_value("system.cpu.utilization.total"),
         last_value("vm.memory.processes_used", unit: :byte),
-        last_value("vm.memory.processes_by_type", tags: [:process_type], unit: :byte),
         last_value("vm.memory.binary", unit: :byte),
         last_value("vm.memory.ets", unit: :byte),
         last_value("vm.system_counts.process_count"),
@@ -197,7 +197,7 @@ with_telemetry [Telemetry.Metrics, OtelMetricExporter] do
 
     defp memory_by_process_type_metrics(%{otel_per_process_metrics?: true}) do
       [
-        last_value("vm.memory.processes_by_type", tags: [:process_type], unit: :byte)
+        last_value("process.memory.total", tags: [:process_type], unit: :byte)
       ]
     end
 
@@ -218,7 +218,7 @@ with_telemetry [Telemetry.Metrics, OtelMetricExporter] do
         # Our custom measurements:
         {__MODULE__, :uptime_event, []},
         {__MODULE__, :cpu_utilization, []},
-        {__MODULE__, :memory_by_process_type, [opts]},
+        {__MODULE__, :process_memory, [opts]},
         {__MODULE__, :get_system_load_average, []},
         {__MODULE__, :get_system_memory_usage, []}
       ]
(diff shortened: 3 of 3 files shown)


msfstef:
This moves the measurement to a separate key, but in the issue you describe the expected behaviour to be that we "add" to the existing measurement values - is that not possible and this is the workaround for now?

Also, does this issue affect any of the other `vm.X` measurements in a similar way?

alco:
@msfstef 

> in the issue you describe the expected behaviour to be that we "add" to the existing measurement values - is that not possible and this is the workaround for now?

We can disable the collection of the `vm.memory` metric in `telemetry_poller` and define our own `vm.memory` metric and include all the info there. But there's a mismatch in the usage between the system-wide measurements such as `vm.memory.total`, `vm.memory.ets` and our `processes_by_type` because we add tags to our measurements. So to keep using `vm.memory.*` we'd need to duplicate vm-wide metrics for each process tag and that doesn't make much sense.

> Also, does this issue affect any of the other vm.X measurements in a similar way?

My code search didn't turn up any other cases where our custom metrics conflict with those defined by `telemetry_poller`.

msfstef (approved):

PR #2679, opened 2025-05-06
chore (examples): Deploy example apps in parallel

kevin-dp:
Currently, example apps are deployed sequentially. This takes a very long time (> 40 minutes). This is not very practical when we want to quickly test changes to one of these example apps. This PR modifies the workflow to **deploy** the **changed examples in parallel**.

### DIFF
--- a/.github/workflows/deploy_all_examples.yml
+++ b/.github/workflows/deploy_all_examples.yml
@@ -8,10 +8,27 @@ concurrency:
   group: ${{ github.event_name == 'push' && 'prod-deploy-group' || format('examples-pr-{0}', github.event.number) }}
 
 jobs:
-  deploy-examples:
-    name: Deploy All Examples to Production
+  deploy:
+    name: Deploy ${{ matrix.example.name }}
     environment: "Production"
     runs-on: ubuntu-latest
+    continue-on-error: true
+    needs: changed-files
+    strategy:
+      matrix:
+        example:
+          - name: yjs
+            path: examples/yjs
+          - name: linearlite-read-only
+            path: examples/linearlite-read-only
+          - name: write-patterns
+            path: examples/write-patterns
+          - name: nextjs
+            path: examples/nextjs
+          - name: todo-app
+            path: examples/todo-app
+          - name: proxy-auth
+            path: examples/proxy-auth
 
     env:
       DEPLOY_ENV: "production"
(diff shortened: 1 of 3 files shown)


alco:
@kevin-dp example compete for the same stage when running `sst deploy`. Each example needs to use a different stage in order for them to be deployed in parallel.

msfstef:
Can confirm that stage names do not conflict since these are all different SST applications - the stage is like the environment, the state files are stored under [app name] -> [stage name]

alco:
Sorry, ignorance on my part.

.github/workflows/deploy_examples.yml
```diff
@@ -46,125 +92,66 @@ jobs:
...
-            - nextjs: ${nextjs}
-            - todoapp: ${todoapp}
-            - yjs: ${yjs}
-            - writes: ${writes}
-            - auth: ${auth}
-            `;
+            // Get the matrix of examples that were deployed
+            const matrix = JSON.parse(${{ toJSON(needs.changed-files.outputs.matrix) }})
+            
+            const urls = {
```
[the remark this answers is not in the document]
kevin-dp: I tried to do this but GH doesn't give you access to `needs.deploy.outputs`, instead you have to read specific outputs...

.github/workflows/deploy_all_examples.yml
```diff
@@ -43,34 +60,13 @@ jobs:
...
-        run: pnpm sst deploy --stage production
-
-      - name: Deploy Yjs example
-        working-directory: ./examples/yjs
-        run: pnpm sst deploy --stage production
+      - name: Deploy
+        id: deploy
+        working-directory: ./${{ matrix.example.path }}
+        run: |
+          pnpm --filter @electric-sql/client --filter @electric-sql/experimental --filter @electric-sql/react run build
```
msfstef, line 71: how come this is necessary when it wasn't before? was the previous CI workflow not actually complete? (it would be nice if this part could be shared by all of them but not worth it for this ad-hoc workflow)

msfstef (approved):
🚀

.github/workflows/deploy_all_examples.yml
kevin-dp: It was already there in the "Deploy Write Patterns example" step. And since they ran sequentially, it means that from that point on all the next deployments also had access to it (for some reason the deployments before that one didn't need it).

PR #2681, opened 2025-05-06
Fix the gating of telemetry deps behind a Mix target

alco:
This PR fixes the typo in the `targets` option for telemetry deps. All it does is make sure those deps aren't compiled when Mix target is set to anything other than `application`.

We also disable compilation of telemetry modules in the test env to make sure the stub code is exercised. A new CI step has been added specifically for the tests that exercise ApplicationTelemetry.


msfstef (approved):
Looks good!

PR #2682, opened 2025-05-06
fix: Use regular connection for advisory lock instead of pooled one

msfstef:
I'm investigating some issues around the release of the advisory lock - sometimes I've noticed that we've terminated the Electric process but when trying to grab the relevant advisory lock elsewhere we fail as the lock is still in use.

I'm not sure how quickly Postgres should be responding to the session ending and releasing the lock, I assume very quickly, but I started thinking about what happens with connection pooling.

For example, pgBouncer [does not support session locks](https://www.pgbouncer.org/features.html) when in transaction pooling mode, and generally unless advisory locks are explicitly supported poolers would mess with the expected behaviour of the advisory lock.

For that reason I suggest using the unpooled connection for the advisory lock. I also thought that we might want to explicitly unlock it but according to PG docs:
> pg_advisory_unlock_all () → void : Releases all session-level advisory locks held by the current session. (This function is implicitly invoked at session end, even if the client disconnects ungracefully.)

### DIFF
--- a/.changeset/mighty-otters-fry.md
+++ b/.changeset/mighty-otters-fry.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Use non-pooled connection for grabbing advisory lock to avoid unexpected behaviour.
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -277,12 +277,13 @@ defmodule Electric.Connection.Manager do
         :start_lock_connection,
         %State{
           current_phase: :connection_setup,
-          current_step: {:start_lock_connection = step, _},
+          current_step: {:start_lock_connection, _},
           lock_connection_pid: nil
         } = state
       ) do
     opts = [
-      connection_opts: connection_opts(step, state),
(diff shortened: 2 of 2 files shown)


alco:
Great thinking!

The changes need to be mirrored for the `update_connection_opts()` function as well because there's actually a coupling between the two: when we start any connection and the process exits, we will try to repair the connection opts for that step. So if the two functions aren't in sync, bad things may happen.

This indicates that more refactoring is in order... But as a quick fix, I'm suggesting the following changes to make the code a tiny bit more self-documenting.

alco (approved):

balegas:
That sounds good. Expanding on the hypothesis, I'm wondering if makes sense that the lock is not released because the pooler takes longer to close the session when the client disconnects... hence wouldn't it be good to explicitly call the release?

msfstef:
> the lock is not released because the pooler takes longer to close the session when the client disconnects...

I tried to understand if that was the case, and my conclusion was that it depends on the pooler, on the pooler configuration, and that ultimately the advisory (and specifically session level) lock behaviour is not going to be predictable. For that reason I think this change is the immediate fix - alternatively we can potentially find a different locking mechanism.

PR #2684, opened 2025-05-07
fix: Only log unexpected lock acquisition errors

msfstef:
The lock acquisition might take a while, and many PG providers might time out the statement execution - we're fine to just keep trying but we should only log unexpected errors.

We could argue that we don't even want to backoff and retry, and just immediately retry, but I'm worried that since the providers are timing statements out we might be better off not retrying immediately.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/lock_connection.ex
+++ b/packages/sync-service/lib/electric/postgres/lock_connection.ex
@@ -101,9 +101,12 @@ defmodule Electric.Postgres.LockConnection do
     {time, backoff} = :backoff.fail(backoff)
     tref = :erlang.start_timer(time, self(), :acquire_lock)
 
-    Logger.error(
+    log_message =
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/postgres/lock_connection.ex
```diff
@@ -101,9 +101,12 @@ defmodule Electric.Postgres.LockConnection do
     {time, backoff} = :backoff.fail(backoff)
     tref = :erlang.start_timer(time, self(), :acquire_lock)
 
-    Logger.error(
+    log_message =
```
robacourt, line 104: Inspecting the error looks quite messy in the logs, I'd only show it if the error is unexpected.

robacourt (approved):
LGTM. Just left a tiny nit

icehaunter (approved):

Issue #2691, opened 2025-05-08
Exponential backoff does not kick in if replication fails after it has successfully started streaming


alco:
**Versions**
Current Electric version.

**Bug description**
[The recent refactoring](https://github.com/electric-sql/electric/pull/2652) fixed the issue for the case where attempting to start streaming from a repication slot fails and caused the replication client to reconnect. However, there may be other repeating failure conditions that are only reachable after streaming has successfully started, such as the case when Electric's publication is removed and a row is updated in a table that used to be included in the publication.

**Expected behavior**
Once the replication client is up and running and has received the first data message, Electric needs to wait for a set time interval before marking the connection as successful. Any failure within that time interval will result in the reconnection with the exponential backoff. As soon as the set time interval passes with no issues, the connection manager can reset the backoff state.

PR #2693, opened 2025-05-08
fix: Add timeout before considering replication connection successful

msfstef:
Closes https://github.com/electric-sql/electric/issues/2691

Added a 5 second timeout - what we've observed is setup errors would come through within a few milliseconds, but I think allowing a few seconds is reasonable. A connection that falls over within 5 seconds of being established should be considered unsuccessful IMO.

I've avoided using timers and storing timer references, as I think we can get the job done by relying on the fact that if the replication client pid is the same between receiving the first message and the timeout passing, we can consider the connection successful.

Will test this ad hoc with @alco 's reproduction steps - we don't really have proper testing around the ConnMan so not sure what to do with this.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -142,8 +142,10 @@ defmodule Electric.Connection.Manager do
 
   @connection_status_check_interval 10_000
 
-  # Time after establishing replication connection before we consider it successful,
-  # to allow for setup errors sent over the stream to be received.
+  # Time after establishing replication connection before we consider it successful
+  # from a retrying perspective, to allow for setup errors sent over the stream
+  # to be received. Any failure within this period will trigger a retry within
+  # the same reconnection period rather than a new one.
   @replication_liveness_confirmation_duration 5_000
 
   def child_spec(init_arg) do
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -537,13 +537,16 @@ defmodule Electric.Connection.Manager do
   # reconnection attempt will start from the minimum timeout and grow exponentially from
   # there.
   def handle_info(
-        {:replication_liveness_timeout, replication_client_pid},
+        {:timeout, _tref, {:replication_liveness_check, replication_client_pid}},
         %State{replication_client_pid: replication_client_pid} = state
       ),
       do: {:noreply, mark_connection_succeeded(state)}
 
-  def handle_info({:replication_liveness_timeout, _replication_client_pid}, state),
-    do: {:noreply, state}
+  def handle_info(
+        {:timeout, _tref, {:replication_liveness_check, _replication_client_pid}},
+        state
+      ),
+      do: {:noreply, state}
 
   # Special-case the explicit shutdown of the supervision tree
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -142,6 +142,10 @@ defmodule Electric.Connection.Manager do
 
   @connection_status_check_interval 10_000
 
+  # Time after establishing replication connection before we consider it successful,
+  # to allow for setup errors sent over the stream to be received.
+  @replication_liveness_timeout 5_000
```
robacourt, line 147: It's not really a timeout because it's not an error if it goes over. Maybe `confirm_live_duration`?

robacourt, line 147: 5 seconds seems like a crazy long time to me, for the cloud UI to update and API calls to be ready. This will be seen with new users and sluggishness will not give a good impression. 1 second max I would say, perhaps even less if you can get away with it.

robacourt (approved):
Nice. Just left a couple of nits

packages/sync-service/lib/electric/connection/manager.ex
msfstef: To clarify - we're not waiting 5 seconds to confirm the stack is ready and start doing things - we're waiting 5 seconds to consider any connection failure to belong to a separate "run" of failures and restart the backoff.

Basically what I'm saying here is that if a failure occurs within 5 seconds, we retry with an increased backoff timeout and consider it a retry. If it fails after 5 seconds, the retries are starting from scratch with the initial timeout and backoff.

robacourt: Ah, ok, gotcha. 5s is fine then 👍

msfstef: I left a larger comment explaining this a bit better cause I understand the confusion

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -748,15 +767,19 @@ defmodule Electric.Connection.Manager do
...
+    # to start replication.
     # This is the only way to be sure because it can still fail after we issue the
     # start_streaming() call, so marking it as having succeeded earlier would result in a
     # reconnection loop with no exponential backoff.
-    state = mark_connection_succeeded(state)
+    Process.send_after(
+      self(),
+      {:replication_liveness_timeout, state.replication_client_pid},
+      @replication_liveness_confirmation_duration
+    )
```
alco, line 781: Just a nitpick: I would make this consistent with the other timers in this module and use the `:erlang.start_timer()` function. Its benefit is a uniform message format whereas with `send_after()` every message tends to be unique.

```suggestion
    :erlang.start_timer(
      @replication_liveness_confirmation_duration,
      self(),
      {:replication_liveness_check, state.replication_cliend_pid}
    )
```

and then in `handle_info` you'll need to match on

```
  def handle_info(
        {:timeout, _tref, {:replication_liveness_check, replication_client_pid}},
        %State{replication_client_pid: replication_client_pid} = state) do
...
```

On the technical side, there's barely any difference between the two, `Process.send_after()` also returns a timer ref.

alco (approved):
Perfect!

Only one nit from me.

PR #2695, opened 2025-05-08
chore (examples): playwright test for remix example

kevin-dp:


balegas (approved):

msfstef (approved):
Nice work, looks good! Why not make the test a separate step in the CI workflow though? (i.e. `pnpm test:browser` if the example has this script)

I can envision us running e2e tests on a schedule rather than just on deploy

PR #2701, opened 2025-05-12
chore (examples): deploy react example

kevin-dp:


examples/react/sst.config.ts
```diff
@@ -0,0 +1,61 @@
...
+  async run() {
+    const dbName = isProduction() ? `react-app` : `react-app-${$app.stage}`
+
+    const { sourceId, sourceSecret } =
+      createDatabaseForCloudElectric({
+        dbName,
+        migrationsDirectory: `./db/migrations`,
+      })
+
+    if (!process.env.ELECTRIC_API) {
```
msfstef, line 34: nit: I would move this check at the start, before we create a database - I see you might be following a pattern in other examples but I think we should slowly move the env check at the start for all of them.

msfstef (approved):
Looks good but is it not supposed to have add item and clear item buttons?

image

kevin-dp:
> Looks good but is it not supposed to have add item and clear item buttons?

I also found it a bit weird but it never had. This was known as the "basic" example so i was assuming the purpose of the example is just to show the `useShape` hook to sync data.

PR #2703, opened 2025-05-12
(fix)sync-service: ensure stack messages are json serialisable

robacourt:


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -931,16 +941,15 @@ defmodule Electric.Connection.Manager do
...
-        In order to recover consistent replication, the slot will be dropped along with all existing shapes.
-        If you're seeing this message without having recently stopped Electric for a while,
-        it's possible either Electric is lagging behind and you might need to scale up,
-        or you might need to increase the `max_slot_wal_keep_size` parameter of the database.
-    """)
+    Logger.warning(error.message)
 
     dispatch_stack_event(
-      {:database_slot_exceeded_max_size, %{error: error.original_error}},
+      {:warning,
```
msfstef, line 947: is `:warning` the right message for this? it feels like it needs a user action as well? I guess we might be reading off of warnings as well to do stuff

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -862,7 +872,7 @@ defmodule Electric.Connection.Manager do
     dispatch_stack_event(
       {:database_connection_failed,
```
msfstef, line 873: is this going to go under `:connection_issue` in a separate PR?

msfstef (approved):
Looks good!

packages/sync-service/lib/electric/connection/manager.ex
robacourt: yes, I can't change it yet because of backwards compatibility

packages/sync-service/lib/electric/connection/manager.ex
robacourt: It auto-recovers so it's just a warning. the plan for now with warnings is just show them in the logs, but at some point they might be emailed to the customer as well.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -621,8 +623,12 @@ defmodule Electric.Connection.Manager do
     )
 
     dispatch_stack_event(
-      {:database_connection_severed,
-       %{error: error.original_error, type: error.type, message: error.message}},
+      {:connection_issue,
```
alco, line 626: `:connection_error` would be more appropriate IMO, to leave any doubt of a serious problem even before turning one's attention to the stack event's details.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -931,16 +941,15 @@ defmodule Electric.Connection.Manager do
...
-    """)
+    Logger.warning(error.message)
 
     dispatch_stack_event(
-      {:database_slot_exceeded_max_size, %{error: error.original_error}},
+      {:warning,
+       %{
+         type: :database_slot_exceeded_max_size,
+         message: error.message,
+         error: inspect(error.original_error, pretty: true)
```
alco, line 951: Perhaps this could be encapsulated into a function in the `DbConnectionError` module to avoid repetition? E.g. `DbConnectionError.format_original_error(error)`.

alco (approved):

packages/sync-service/lib/electric/connection/manager.ex
robacourt: done

packages/sync-service/lib/electric/connection/manager.ex
robacourt: done

PR #2705, opened 2025-05-12
feat: Idempotent `PublicationManager` calls

msfstef:
Ensures calls to `PublicationManager` are made with an associated shape handle to ensure idempotency and proper cleanups.

This prevents cases where the following happens:
1. Shape with handle X and shape definition A is added
2. Shape with handle Y != X and shape definition A is added
3. Shape X is removed/deleted - cleanup ends up trying to clean the associated publication more than once
4. Shape Y gets accidentally removed from the publication because of Shape X's cleanup

To avoid such cases, we also keep track of the shape handles we track - this is not a large memory overhead and ensures we no longer have to worry about this. We can improve the approach but tracking shape handles in the publication is something we definitely want to do.

### DIFF
--- a/.changeset/dirty-cherries-poke.md
+++ b/.changeset/dirty-cherries-poke.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Make `PublicationManager` calls idempotent based on shape handles to ensure proper cleanups.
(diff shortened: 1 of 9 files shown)


magnetised (approved):
Lovely

Issue #2692, opened 2025-05-08
When Electric needs to recreate the slot or the publication, the other one must also be recreated


alco:
Currently Electric will create the publication when it doesn't find one and do the same for the replication slot.

However, we have encountered a scenario in which using the old slot after the publication has been dropped and recreated results in repeated faliure to stream from Postgres. There appears to be a lingering reference to the old publication in Postgres' internal cache. The only remedy @msfstef and I have come up with is to recreate both as a unit to bring the replication state back into working condition.

To reproduce:
- start with a clean database
- start Electric
- create a table and a shape for it
- drop the publication (`DROP PUBLICATION electric_publication_default`)
- insert into the table

The replication client in Electric exits and the connection manager then starts it again.

<details>
<summary>Log output</summary>

```
12:45:14.130 pid=<0.738.0> [error] :gen_statem {:"Elixir.Electric.ProcessRegistry:single_stack", {Electric.Postgres.ReplicationClient, nil}} terminating
** (Postgrex.Error) ERROR 42704 (undefined_object) publication "electric_publication_default" does not exist
    (stdlib 6.2) gen_statem.erl:3864: :gen_statem.loop_state_callback_result/11
    (stdlib 6.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
Process Label: {:replication_client, "single_stack"}
Queue: [info: {:tcp, #Port<0.17>, <<69, 0, 0, 0, 239, 83, 69, 82, 82, 79, 82, 0, 86, 69, 82, 82, 79, 82, 0, 67, 52, 50, 55, 48, 52, 0, 77, 112, 117, 98, 108, 105, 99, 97, 116, 105, 111, 110, 32, 34, 101, 108, 101, 99, 116, 114, ...>>}]
Postponed: []
[...]

PR #2706, opened 2025-05-12
Recreate the replication slot when the publication is recreated

alco:
Implements #2692.

- [x] Add an integration test


msfstef (approved):
Looks good! Adding an integration test that recreates the described situation might be worth it for this case I think (and we can extend that test with the OID and LSN detection)

Issue #2559, opened 2025-04-08
In-flight requests crash when database is deleted


kevin-dp:
When Electric has one or more in-flight requests (e.g. a live request) for a given database and the database gets deleted while the requests are still in flight then the request returns a 500:

```json
{
  "error": "** (ArgumentError) errors were found at the given arguments:\n\n  * 1st argument: the table identifier does not refer to an existing ETS table\n"
}
```

And the full error from the Electric logs:

```
10:32:26.777 pid=<0.797.0> request_id=GDRKu_EjNPhbQdkAAAIE [info] Sent 500 in 20013ms

10:32:26.780 pid=<0.797.0> request_id=GDRKu_EjNPhbQdkAAAIE [error] ** (Plug.Conn.WrapperError) ** (ArgumentError) errors were found at the given arguments:

  * 1st argument: the table identifier does not refer to an existing ETS table

    (stdlib 6.2) :ets.lookup(:"single_stack:lsn_tracker", :last_processed_lsn)
    (electric 1.0.5) lib/electric/lsn_tracker.ex:26: Electric.LsnTracker.get_last_processed_lsn/1
    (electric 1.0.5) lib/electric/shapes/api.ex:571: Electric.Shapes.Api.get_global_last_seen_lsn/1
    (electric 1.0.5) lib/electric/shapes/api.ex:371: Electric.Shapes.Api.determine_global_last_seen_lsn/1
    (electric 1.0.5) lib/electric/shapes/api.ex:545: Electric.Shapes.Api.hold_until_change/1
    (electric 1.0.5) lib/electric/shapes/api.ex:414: anonymous fn/1 in Electric.Shapes.Api.serve_shape_log/1
    (opentelemetry_api 1.4.0) src/otel_tracer_noop.erl:59: :otel_tracer_noop.with_span/5
    (electric 1.0.5) lib/electric/telemetry/open_telemetry.ex:87: anonymous fn/3 in Electric.Telemetry.OpenTelemetry.do_with_span/4
    (telemetry 1.3.0) /Users/kevin/Documents/Electric/development/electric/packages/sync-service/deps/telemetry/src/telemetry.erl:324: :telemetry.span/3
    (electric 1.0.5) lib/electric/shapes/api.ex:427: Electric.Shapes.Api.serve_shape_log/2
    (electric 1.0.5) lib/electric/plug/serve_shape_plug.ex:1: Electric.Plug.ServeShapePlug.plug_builder_call/2
[...]

PR #2712, opened 2025-05-13
fix: Ensure stack is active after long poll timeout

msfstef:
Fixes https://github.com/electric-sql/electric/issues/2559

The steps to reproduce this issue, along with a host of other ones, was to essentially delete the database while a request was in-flight and waiting for updates (either new changes or a shape rotation).

If the database is deleted or the connection fails for any other reason, the stack is retrying and trying to set itself up, which means that the `LsnTracker` ETS table is not live, and thus the `global_last_seen_lsn` cannot be read.

Initially I tried to address that problem in isolation, but more kept coming, and I thought that the cleanest thing would be to ensure that the stack is actually up and running after a long poll timeout and return the same class of stack not ready errors if it is not.

This means potentially adding, at worst, an additional `stack_ready_timeout` ms on top of the `long_poll_timeout` to the response, the defaults are 5s and 20s respectively. I could modify the code to use a short `stack_ready_timeout` just for that case but I think keeping things simple for handling things falling over is cleaner. I think it's fine to provide a little bit of time if the stack is actually recovering rather than eagerly return 500s.

An alternative would be to subscribe to stack status notifications (I'd need to write the code for `StatusMonitor` to handle that) and _immediately_ return a 500, but I think that's not appropriate, as the stack might fall over due to a connection issue and recover within a few seconds - I think long poll requests should be able to wait quietly for the stack to potentially get back up and send over changes processed within the long poll timeout. That being said, it might be a cleaner approach (and more in line with our 'is stack ready' checks before starting to handle requests).

I've added a check that the registry exists as from my tests potentially the whole of the application might fall over (e.g. due to fatal connection errors) and therefore there is no registry from which to "clean up" the change listener, and we can ensure this way that decent/consistent error responses are sent back.

### DIFF
--- a/.changeset/hot-carpets-destroy.md
+++ b/.changeset/hot-carpets-destroy.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Ensure stack is active after long poll timeout
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -534,17 +534,29 @@ defmodule Electric.Shapes.Api do
     after
       # If we timeout, return an up-to-date message
       long_poll_timeout ->
-        request
-        |> update_attrs(%{ot_is_long_poll_timeout: true})
-        |> determine_global_last_seen_lsn()
-        |> no_change_response()
+        # Ensure stack is ready after a long poll timeout, as it might
+        # have failed during this period
+        case hold_until_stack_ready(request.api) do
+          :ok ->
+            request
+            |> update_attrs(%{ot_is_long_poll_timeout: true})
+            |> determine_global_last_seen_lsn()
+            |> no_change_response()
+
+          {:error, response} ->
+            response
+        end
     end
   end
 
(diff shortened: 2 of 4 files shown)


robacourt (approved):
I like that as a solution, sure it may extend the timeout but only in quite unusual circumstances and not by much, and it's super simple. Great work!

alco:
Both the stack readiness and the liveness of the registry process may flip just after our checks, in other words the code is still vulnerable to races. So if the goal is to return a meaningful error to the client instead of a generic 500, we need to wrap those expressions with `try do`.

msfstef:
> Both the stack readiness and the liveness of the registry process may flip just after our checks

You're right, although this is true of all parts of the API not just after the long poll timeout - this change is mostly asserting that after a large period of time that is the long poll timeout, we should treat the request the same way we do when receiving the request.

For other races I think we should have a more generic check in the Plug error handler that can either have a more generic message or even check the `StatusMonitor` and return a nice 503 error if the stack is inactive - treating any unexpected error as being due to the stack being down (as far as the requester is concerned).

alco:
@msfstef yeah, I like your thinking there.

alco (approved):

Issue #2504, opened 2025-03-31
Remove location header on 409


kevin-dp:
Currently 409 returns a location header but this is a potential security leak when combined with a proxy pattern that adds WHERE clauses based on authentication.

Let's remove the location header.

@KyleAMathews comment:
> We thought about using this in clients early on but never ended up using it — so not a break change to remove.

msfstef:
As described in https://github.com/electric-sql/electric/pull/2713 which closes this, we currently include the `location` header in our OpenAPI spec so this could potentially be considered a breaking change. Probably a decision to make sooner rather than later.

PR #2713, opened 2025-05-13
fix: Remove location header from 409 responses

msfstef:
Closes https://github.com/electric-sql/electric/issues/2504

Technically we were documenting this in our OpenAPI spec, so even though we don't use it in our clients and we don't advertise it much in our docs this _is_ a bit of a breaking change?

I'll let @KyleAMathews decide if we want to move forward with this, I've prepared this PR to get rid of it just to accelerate a decision on this.

### DIFF
--- a/.changeset/olive-garlics-clap.md
+++ b/.changeset/olive-garlics-clap.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Remove `location` header from 409 resopnses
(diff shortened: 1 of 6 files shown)


KyleAMathews (approved):
:shipit: 

Yeah, a breaking change is one that breaks someone — so it seems unlikely this will break anyone given the official clients don't use it.

And regardless, breaking changes are more acceptable for closing potential security problems.

kevin-dp (approved):
LGTM!

PR #2714, opened 2025-05-13
chore (deployments): Pin changed-files action

kevin-dp:
We're pinning this action because some versions were compromised and we want to be sure we don't accidentally use one of those versions. This hash corresponds to the latest release: v46.0.5


msfstef (approved):
Thank you @icehaunter for bringing attention to [this](https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066)

Issue #2571, opened 2025-04-08
Feature request: pass through more fetch options to ShapeStream


KyleAMathews:
We're adding them one by one — we include params & headers & signal. We should probably just include the rest that apply e.g.:

- credentials
- redirect
- referrer

PR #2715, opened 2025-05-13
feat: Add more standard `fetch` options to `ShapeStreamOptions`

msfstef:
Closes https://github.com/electric-sql/electric/issues/2571

Adds more standard `fetch` options to `ShapeStreamOptions` since we already had `headers` and `signal`.

Technically already possible via a custom `fetchClient` but this PR addresses the mentioned issue to ensure we make a decision on whether we want this or not.

I don't love that we're "polluting" the `ShapeStreamOptions` at the root level with more fetch options, as we could potentially want some of these keys for our own purposes (and we could've instead had a `requestOpts` key with all of these).

Happy to hear feedback on this.

### DIFF
--- a/.changeset/shaggy-nails-juggle.md
+++ b/.changeset/shaggy-nails-juggle.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
+Add more `fetch` options to `ShapeStreamOptions` that get carried forward to requests.
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -17,11 +17,13 @@ import {
   ReservedParamError,
 } from './error'
 import {
+  AllowedRequestOptions,
   BackoffDefaults,
   BackoffOptions,
   createFetchWithBackoff,
   createFetchWithChunkBuffer,
   createFetchWithConsumedMessages,
+  createFetchWithRequestOptions,
   createFetchWithResponseHeadersCheck,
 } from './fetch'
 import {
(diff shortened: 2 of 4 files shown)


kevin-dp (commented):
Not sure how i feel about this one. As you said i don't like polluting the options, especially since users can achieve this themselves if they really need to by providing their own fetch client. A separate requestOptions key would perhaps be a tad better.

KyleAMathews:
I guess I don't feel strongly about the extra options as I see ShapeStream as basically a fancy fetch client specific to subscribing to Electric shapes. Also lots of libraries have lots of options that nobody uses unless they need them so in practice, it's not a big deal.

But I can see both sides. It does seem a bit late though to add a sub-object given the other ones are on the top level?

msfstef:
> It does seem a bit late though to add a sub-object given the other ones are on the top level?

We could add a sub-object now, like `requestOptions`, which either includes `headers` and `signal` which are the only two we have (and `headers` is a custom one for us as we have reserved keys) and we respect the `requestOptions` one over the root level one, marking them as deprecated (or vice versa) - or the `requestOptions` one initially just doesn't have `signal` and `headers`.

I'm not against `signal` and `headers` being at the root and the rest being in a sub object because the `signal` is meant to be a signal for the _stream_ rather than an individual request's options, and since we have custom `headers` and reserved keys it makes sense to me to have those at the root. The rest are just options that will be passed to `fetch` as they are.

I think it's worth thinking about an appropriate name for the sub-object thorugh, like `additonalFetchParams`, `additionalRequestOptions`, `extraFetchParams`, `fetchOptions` - unsure what to pick, but mind you we already have a `fetchClient` so something like `fetchParams` or `fetchAdditionalOptions` might be good.

msfstef:
Closing as stale - we can revisit if this is still relevant, the approach wasn't ideal either.

Issue #2437, opened 2025-03-11
Log slow queries (when creating a new shape)


KyleAMathews:
Electric initializes a new shape by running a query. It streams the results back to the client and also to disk to start the shape log. Subsequent requests are resolved from disk so much faster.

Some shape initializations can be quite slow for normal SQL select reasons e.g. it's returning lots of rows, the where clause isn't against an index so it requires a full table scan, etc.

To make this clear — we should log whenever a query takes more than perhaps 1s with the duration, the shape definition / url, and the query that Electric runs.

Then people can run the query directly against PG to debug why it's running slow.

msfstef:
Additionally, we now no longer have a single "snapshot" file - we stream the query result into chunks - so I'm assuming by slow query we really mean from performing the query to starting to receive the result stream.

msfstef:
Opened https://github.com/electric-sql/electric/pull/2716 that logs a warning with the actual query used - this measures the time from doing the initial query until the stream is stored, so this includes IO time as well.

The query alone to get the stream won't take longer than 200ms (P999 on Cloud), so the actual stream to file needs to be included, which does indeed have P90 in the seconds.

erik-the-implementer:
Closing this out — it looks resolved on `main` / in a released version.

**Why:** PR #2716 'logs a warning with the actual query used' for slow snapshot creation (measures from initial query until stream stored), opened by the assignee msfstef directly in response to this issue (2025-05-13).

If you can still reproduce it on the latest release, please drop a note here and we'll reopen.

<sub>Triaged as part of an open-issue sweep. Apologies for the bulk notification.</sub>

PR #2716, opened 2025-05-13
feat: Log slow snapshot queries for ease of debugging

msfstef:
Closes https://github.com/electric-sql/electric/issues/2437

Implementing a feature request by @KyleAMathews - logs slow queries (that take longer than 1 second) as a warning, along wit the actual query executed so it can be more easily debugged.

I don't have a good way to test this however, any advice appreciated.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
@@ -212,8 +212,9 @@ defmodule Electric.Shapes.Consumer.Snapshotter do
 
               Logger.warning(
                 "Slow query for handle #{shape_handle} took #{total_query_time}ms, shape: #{inspect(shape)}, query: #{query_str}",
-                query_str: query_str,
-                stack_id: stack_id
+                stack_id: stack_id,
+                shape_handle: shape_handle,
+                query_str: query_str
               )
             end
 
--- a/packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
@@ -8,6 +8,10 @@ defmodule Electric.Shapes.Consumer.Snapshotter do
 
   require Logger
 
+  # The threshold in milliseconds for an initial data query to stream out
+  # to be considered slow and for it to be logged as a warning
+  @slow_query_threshold_ms 1_000
+
   def name(%{
         stack_id: stack_id,
         shape_handle: shape_handle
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
```diff
@@ -197,7 +203,22 @@ defmodule Electric.Shapes.Consumer.Snapshotter do
 
             # could pass the shape and then make_new_snapshot! can pass it to row_to_snapshot_item
             # that way it has the relation, but it is still missing the pk_cols
-            Storage.make_new_snapshot!(stream, storage)
+            result = Storage.make_new_snapshot!(stream, storage)
+
+            total_query_time = System.monotonic_time(:millisecond) - query_start_time
+            # in case the query took a long time, log it to facilitate debugging
+            if total_query_time > @slow_query_threshold_ms do
```
icehaunter, line 210: I don't like it that we're using a constant value here as a comparison when it would be proportional to the amount of rows queried in addition to plain query complexity. And influenced by how weak or overloaded the PG instance is. Given we can relatively easily make the storage return an ok-tuple with the written row count, we should do that and scale our "slow" marker accordingly. 

I'm also not sure who this log is aimed at, with it logging entire weirdly built select. The json forming is functionally a constant overhead for every row, which means relative slowness marking a perceived "slow" query is likely to come from either where clause complexity or PG being overloaded.

Because of that, I suggest we instead log not the executed query, but a fictional one that reads as equivalent. So, for a `Shape.new(columns: [:id, :text, :other_text], where: "text ~ 'hello world'")` we would log smth like this, replacing dirty manual json forming with an easier to read fake function
```sql
-- make_storage_json is temporary and built at runtime
SELECT make_storage_json(id, text, other_text) WHERE text ~ 'hello world'
```

msfstef: Agree on a dynamic threshold with row count - we can do a linear scaling with a baseline that we can determine from honeycomb (might be worth adding number of rows written to the span as well).

And practically agree on a better formed query that's actually readable as well - with a small clause that if it's not _exactly_ the same then we might miss potentially weird reasons for slow downs. I think that relates to your question about who this is for - my interpretation of the issue is that it should be for anyone hosting Electric (including ourselves) to more easily debug why a shape might be so slow to start. With that in mind I think potentially missing an unknown weird bug is worth it to make the logged query be something reasonable.

icehaunter:
Superseded by #3157 and closed in #3156

PR #2717, opened 2025-05-13
fix: on-load reconciliation by adding on-disk storage

### LINKED ISSUE #2689: Seemingly infinite loop of 409 must-refetch responses when table dropped and recreated
**Versions**
- Client: typescript, 1.0.4
- Electric: 1.0.12 

**Bug description**
I'm seeing a seemingly infinite loop of 409 responses and requests from the typescript client when a table is dropped and recreated.

I've created a repo that reproduces this: https://github.com/kohdan/electric-repro

<details>

<summary>docker-compose.yml</summary>
I had to modify it to wait for healthcheck since electric will crash otherwise in 1.0.12.

```yaml
name: "electric_quickstart"

services:
  postgres:
    image: postgres:17-alpine
    environment:
      POSTGRES_DB: electric
      POSTGRES_USER: postgres
      POSTGRES_PASSWORD: password
    ports:
      - 54321:5432
    tmpfs:
      - /var/lib/postgresql/data
      - /tmp
    command:
      - -c
      - listen_addresses=*
      - -c
      - wal_level=logical
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U postgres -d electric"]
      interval: 5s
      timeout: 5s
      retries: 5

  electric:
    image: electricsql/electric:1.0.12
    environment:
      DATABASE_URL: postgresql://postgres:password@postgres:5432/electric?sslmode=disable
      # Not suitable for production. Only use insecure mode in development or if you've otherwise secured the Electric API.
      # See https://electric-sql.com/docs/guides/security
      ELECTRIC_INSECURE: true
    ports:
      - "3000:3000"
    depends_on:
      postgres:
        condition: service_healthy
```

</details>

<details>

<summary>main.ts</summary>

```typescript
import { ShapeStream, Shape } from "@electric-sql/client";

  url: `http://localhost:5173/v1/shape`,
  params: {
    table: "items",
  },
});
const shape = new Shape(stream);

// The callback runs every time the Shape data changes.
shape.subscribe((data) => console.log(data));
```

</details>

I did some curl requests and it seems like at a certain point electric won't stop returning 409s.

**Expected behavior**
More graceful handling of schema changes. I was hoping upgrading to 1.0.12 would fix the issue of stale data staying around after schema changes. Maybe I'm missing something.

icehaunter:
Closes #2689


packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
```diff
@@ -279,4 +313,24 @@ defmodule Electric.Postgres.Inspector.EtsInspector do
     stack_id = Access.fetch!(opts, :stack_id)
     :"#{stack_id}:relation_table"
   end
+
+  @spec persist_data(map()) :: :ok
+  defp persist_data(state) do
+    info = :ets.tab2list(state.pg_info_table)
+    relations = :ets.tab2list(state.pg_relation_table)
```
alco, line 320: Is it going to be a problem if the two tables are out of sync? I'm only just noticing that this module maintains two tables and runs inserts on them sequentially. If all updates stay within one process, then we're kinda fine. But the ETS tables are configured to be `public`, so in principle they could see concurrent writes from anywhere.

alco (approved):

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
icehaunter: We can make them protected if we move cleanup in-process. That's actually in a next pr, so it'll be addressed!

PR #2723, opened 2025-05-14
fix: Do not fail fast on matrix jobs

msfstef:
Avoid fail fast strategy for the following:

- elxir tests on various PG versions: we might suceed in one and fail in another, or might flake in one and not in another in our case, we should not be cancelling the rest
- deploy examples (dispatch and regular): if one example fails to deploy it should not cancel the other deployments - in fact this could lead to a bad SST state @kevin-dp


alco (approved):
:+1:

kevin-dp (approved):

PR #2726, opened 2025-05-15
chore (examples): Use a shared PG for production deployment of examples

kevin-dp:
This PR is a first step towards using a shared PG DB for all production deployments of the examples.
Currently, it only uses this shared DB for the React example. After verifying that this works, we can open a follow up PR to use this shared DB for all examples that use the items table.


examples/.shared/lib/infra.ts
```diff
@@ -13,5 +15,45 @@ export function getSharedCluster(serviceName: string): sst.aws.Cluster {
   })
 }
 
+/**
+ * Returns the shared example database if we are in the production stage.
+ * Otherwise, it creates a new database with the given name.
+ */
+export function getExampleDatabase(dbName: string) {
```
msfstef, line 22: nit: `getExampleSource` in this case?

msfstef (commented):
Looks good but I'm confused about the migrations - would you pool all example migrations under the shared folder? How would that then work for PR environments that deploy non-shared DBs and sources?

I figured that we would be running each example's migration separately and "assume" they are orthogonal to each other.

balegas:
I'm not sure we need to extend this to all examples. I think it's good if we use the same database for the examples that share the same schema, but sharing across different schemas, might make things a bit more complicated

(the same argument for examples that use the same/different) backends.

examples/.shared/lib/infra.ts
balegas: Running examples migrations agains a single database is not going to work well with `pg-migrations`.

kevin-dp:
> I'm not sure we need to extend this to all examples. I think it's good if we use the same database for the examples that share the same schema, but sharing database across examples with different schemas, might make things a bit more complicated.
> 
> (the same argument for examples that use the same/different) backends.

Yeah plan was to start with the things that are common. I was thinking that it might make sense to also share the DB if they have different schemas, essentially all we have to do is to define all the schema in the shared folder. But wen can decide that later.

examples/.shared/lib/infra.ts
kevin-dp: @balegas can you elaborate why? I don't see why it would be a problem, we're running a single shared migration file that creates the items table.

kevin-dp:
> Looks good but I'm confused about the migrations - would you pool all example migrations under the shared folder? How would that then work for PR environments that deploy non-shared DBs and sources?
> 
> I figured that we would be running each example's migration separately and "assume" they are orthogonal to each other.

Basic plan is to have a shared migration just for the items table. The examples that have other schemas would still define their own migrations. Although we could probably move it into the shared folder too and apply them on the shared DB but we can decide on that later.

balegas:
> Although we could probably move it into the shared folder too

The general pattern is to define the schema in the example folder --- we want to keep it simple; We treat the items schema specially because we have many examples using it.

examples/.shared/lib/infra.ts
balegas: We've cleared it out 1-1. We're only going to do this for the examples that use the items schema for now.

balegas (approved):
Thanks for doing this

PR #2728, opened 2025-05-15
feat: Add CodeCov to repo

msfstef:
Adds CodeCov for the sync service

Coverage is merged between regular unit tests and telemetry unit tests before being uploaded.

For test analysis, two junit reports are generated and uploaded.

Only thing that I don't like is that when only specific tags are included, the JUnit reports show the rest of the tests as skipped, so uploading these two reports shows this as running all tests twice but most are skipped (rather than only including the ones that run).

Not sure if this affects us in any way as this is technically correct, I'm willing to see how this looks in our reports before doing any more xml fiddling.


balegas:
It's already showing it's value :D

icehaunter (approved):

Issue #2475, opened 2025-03-23
Rename api_secret param to just secret


KyleAMathews:
Also update cloud to just secret.

We'll make this backwards compatible as it's easy to support both params.

kevin-dp:
I opened a PR for renaming the `api_secret` query param to `secret`.
@KyleAMathews what do you mean with:
> Also update cloud to just secret.

I don't think our cloud is using this secret since we are not publicly exposing our cloud Electrics.

KyleAMathews:
We use `source_secret` in cloud as the parameter — so the idea is to also change that to just `secret`.

thruflo:
Actually @kevin-dp I see it the other way. They are the same thing. You always provide a secret when running in secure mode and with the cloud you just also provide a `source_id`. Yes they need to match and may be used differently our-side but from a user's perspective in the proxy, they always add the `secret` param whether it's self-host or cloud.

(As per recent Discord thread we should sort all this out and make sure the docs are accurate and clear).

kevin-dp:
Indeed they both serve as a secret but they protect different things, one protects the entire (self-hosted) Electric API, the other one protects a single instance and is different per instance! I'm afraid that using the same name for both makes it extra hard to document. For instance, a user looking at the OpenAPI spec will see the `secret` parameter which says that this corresponds to the `ELECTRIC_SECRET` set when starting Electric. But should not be confused with the `secret` parameter in cloud Electric, which is the source's secret. If phrased like that perhaps it clears out confusion, not sure.

thruflo:
My argument is that it should be "confused" with the `secret` parameter in cloud Electric. They are both the secret that needs to be provided to authenticate. In the cloud case, you also provide a `source_id` and the secret is validated together with it. In the OS case, the secret is validated on its own.

I think this is a feature, not a bug.

kevin-dp:
@thruflo https://github.com/electric-sql/stratovolt/pull/456

burakakca:
i have this docker compose yaml and i get error how can i fix that? i'm i missing? i'm using db as supabase

```bash
REPOSITORY                              TAG                IMAGE ID       CREATED        SIZE
electricsql/electric                    latest             2c3e1f2b8aa2   3 days ago     166MB
```

```yaml
version: "3.3"
name: "electricsql_quickstart"

services:
  electric:
    image: electricsql/electric
    environment:
      # Updated DATABASE_URL to point to your Supabase Postgres instance.
[...]

PR #2729, opened 2025-05-16
docs: Update source_secret to secret in proxy example

otterDeveloper:
It seems `source_secret` was renamed to `secret` #2475
But this proxy example was not updated accordingly.

Updating this example should help users avoid confusion and potential 401 errors.

### DIFF
--- a/examples/proxy-auth/app/shape-proxy/route.ts
+++ b/examples/proxy-auth/app/shape-proxy/route.ts
@@ -14,7 +14,7 @@ export async function GET(request: Request) {
 
   if (process.env.ELECTRIC_SOURCE_SECRET) {
     originUrl.searchParams.set(
-      `source_secret`,
+      `secret`,
       process.env.ELECTRIC_SOURCE_SECRET
     )
   }


balegas:
I think these changes are not necessary,```source_secret```is a custom parameter that we use for Cloud deployments. Examples should work following the instructions in the README.md.

otterDeveloper:
It's `secret` for the docker image, it should be documented somewhere.
The guide on proxy auth https://electric-sql.com/docs/guides/auth#proxy-auth links to this example.
And it seems the cloud and self hosted version vary in the secret param?
Now I am confussed.
But these differences should de documented.

The guide in the docs never actually mentions how to auntheticate to the electric instance.

KyleAMathews:
@balegas `source` is what we document got cloud too so this makes sense to me. Either works on cloud but only secret works for docker.

balegas:
@KyleAMathews I was not aware Cloud would either accept `source_secret` and `secret`. Cloud [docs](https://electric-sql.com/product/cloud#making-api-requests) mention `source_secret`. But I just confirmed that you're correct. 

I'll open an issue to change `source_secret` to `secret` in all examples and docs.

> The guide in the docs never actually mentions how to authenticate to the electric instance, either self-hosted or cloud.

The guide indeed predates `ELECTRIC_SECRET` and doesn't mention it. We could improve the guide to mention the mechanism, since it is only documented in the [configuration](https://electric-sql.com/docs/api/config#electric-secret) and [api](https://electric-sql.com/openapi.html#/paths/~1v1~1shape/get)

icehaunter:
Closed via https://github.com/electric-sql/electric/pull/2758, sorry about the troubles

PR #2731, opened 2025-05-17
docs: replace source_secret with secret

balegas:
Since our Cloud accepts the query param `secret`, we should just remove mentions to the legacy `source_secret` in our docs. See #2729 for context.


icehaunter:
Superseded and closed via https://github.com/electric-sql/electric/pull/2758

PR #2736, opened 2025-05-19
chore (examples): shared db for item examples

kevin-dp:
This PR modifies the NextJS, Remix, and TanStack examples to use the shared PG DB.


msfstef (approved):
Looks good!

PR #2738, opened 2025-05-19
chore (examples): Add playwright tests to common examples

kevin-dp:


.github/workflows/test_examples.yml
```diff
@@ -1,33 +1,175 @@
...
-        required: true
-        type: string
+  schedule:
+    - cron: '0 9 * * *'  # Runs daily at 9 AM UTC
+  workflow_dispatch:  # Allow manual triggering
 
 jobs:
   test:
-    name: Run Browser Tests
+    name: Test ${{ matrix.example.name }}
```
msfstef, line 10: this shouldn't reference a matrix right? should be something like `Run e2e tests on examples`

examples/nextjs/playwright.config.ts
```diff
@@ -0,0 +1,72 @@
...
+ * https://github.com/motdotla/dotenv
+ */
+// import dotenv from 'dotenv';
+// import path from 'path';
+// dotenv.config({ path: path.resolve(__dirname, '.env') });
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
```
msfstef, line 14: we could have a shared config in `.shared` that examples reference for brevity - and each example just re-exports the shared config

.github/workflows/deploy_examples.yml
```diff
@@ -138,13 +138,125 @@ jobs:
             exit 123
           fi
 
-  test-remix:
+  test-examples:
```
msfstef, line 141: nit: add a `name: Test examples` or `name: Run e2e tests on examples` or smth (currently looks inconsistent with other jobs on this workflow)

msfstef (approved):
Very nice, awesome job!

I've left some comments around 1) workflow job names and 2) potentially sharing a common playwright config rather than fully duplicating it

PR #2740, opened 2025-05-19
sync-service: Use separate mock modules per-test

magnetised:
keeping the same Mock.* naming

This prevents expectations defined in simultaneously executing tests from clashing. I think this is the cause of various flakes in the CI - we define expectations on `Support.Mock.Behaviour` and occasionally a parallel test hits that expectation but fails because it's coming from an unregistered pid.

This pr gives every test module its own set of `__MODULE__.Mock.Behaviour` mocks to define expectations on but keeps the same usage in the tests.

### DIFF
--- a/packages/sync-service/test/electric/plug/delete_shape_plug_test.exs
+++ b/packages/sync-service/test/electric/plug/delete_shape_plug_test.exs
@@ -1,12 +1,12 @@
 defmodule Electric.Plug.DeleteShapePlugTest do
   use ExUnit.Case, async: true
+  use Support.Mock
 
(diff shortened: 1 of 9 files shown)


msfstef (approved):
Nice - the more isolation the better

robacourt (commented):
This should not be needed. Mox says it supports async tests and uses Nimble Ownership under the hood. We must be using it wrong

magnetised:
> This should not be needed. Mox says it supports async tests and uses Nimble Ownership under the hood. We must be using it wrong

Perhaps.

PR #2749, opened 2025-05-20
Try partitioning shape monitor

magnetised:
Builds on shape cleanup race-condition fixes with a partitioned shape monitor to distribute process registrations across multiple processes


magnetised:
Benchmark this

PR #2753, opened 2025-05-21
fix(sync-service):  Fix ipv6 to ipv4 fallback in the case of 'host is unreachable'

robacourt:


msfstef (approved):
Looks good! The issue with these fallbacks is that if someone configures Electric to use IPv6, it might be that it ends up using IPv4 because of some wrong configurations - but since we have these fallbacks in place we should probably cover most cases.

PR #2756, opened 2025-05-21
chore (examples): Rename source_secret to secret

kevin-dp:
This PR modifies the examples to use the `secret` parameter when interacting with Electric instead of the deprecated `source_secret`. Fixes https://github.com/orgs/electric-sql/projects/16/views/1?pane=issue&itemId=111149403&issue=electric-sql%7Celectric%7C2732

### DIFF
--- a/.github/workflows/teardown_examples_pr_stack.yml
+++ b/.github/workflows/teardown_examples_pr_stack.yml
@@ -67,21 +67,17 @@ jobs:
           export PR_NUMBER=${{ github.event.number }}
           echo "Removing stage pr-$PR_NUMBER"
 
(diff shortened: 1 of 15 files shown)


kevin-dp:
Deployments are failing to create the Neon DB because Neon doesn't seem to like us creating the DBs concurrently. Note that all these DBs are being created on the same Neon PG cluster. Hence, Neon is returning a 423 error code:
```
HTTP 423 (Locked): This status code indicates that the requested resource is locked, usually due to concurrent access or editing restrictions. It implies a temporary unavailability due to ongoing processes.
```

PR #2757, opened 2025-05-21
fix: refactor inspector to ensure correct cleanup & persistance

icehaunter:
This is refactoring of `EtsInspector` and `Inspector` behaviour driven by a bug I've found in the cleanup

In short, I was assuming that passing in a `{schema, table}` tuple, as specified in the spec, did the cleanup, but it did it only partially, leading to issues when doing the reconciliation, and further issues now that I've introduced the persistence.

This PR:
- Normalizes the functions to same return values, explicitly differentiating a missing table from general PG errors
- Removed the Inspector code assumption and logic that it's being passed in not-normalized tables - we do validate and downcase identifiers at entry. This also simplifies internal storage structures
- Normalizes the API of the inspector to essentially only work with OIDs. That's the only true stable identifier, so it should act as the only cache key.
- Removes unused "truncate" parameter for identifier parsing. Instead, to avoid silent truncation and weird injection issues, we complain with a 400 that the identifier is too long.
- Extends test coverage of the EtsInspector to include more edge cases & assumptions

Updated EtsInspector is best read top-down and not as a diff


robacourt (approved):
Nice work! LGTM

PR #2771, opened 2025-05-23
Upgrade asdf tool versions and Elixir/Erlang versions in Dockerfile

alco:
This PR upgrades Elixir to 1.18.4 and Erlang/OTP to 28.0.1, latest available releases.

Elixir Client tests will keep failing until we release an updated version of Electric and bump the dependency version. This is due to the change in how compiled regexes are treated as module attributes in OTP 28.0.


alco:
@msfstef Yep, thanks for the advice.

I noticed an upgrade prompt from pnpm and decide to try out upgrading everything as an exercise to see what would fail. I can slice this up into multiple steps to reduce the risk of stuff breaking.

alco:
I've reverted the upgrade of tool versions for now, instead opting for a 2-phase approach. In phase 1 we get the code changes merged without breaking elixir-client. Once these changes are released and the Electric dependency version is updated in elixir-client, then we upgrade Elixir and Erlang in phase 2.

PR #2544, opened 2025-04-06
Add experimental support for streaming SSE in live mode.

thruflo:
If you add `experimental_live_sse=true` to your live requests, then the server streams SSEs rather than returning immediately when there's new data.

Requests are closed after 60 seconds, in order to support request collapsing. We also diverge slightly from default SSE behaviour by requiring the client to re-connect on a new URL once the request is closed. Because we require the client to honour our API mechanism of advancing the offset.

This can be worked around using a standard JS EventStream client by closing in the event of error and reconnecting manually. Just a rough sketch for reference:

```js
const url = `http://localhost:3000/v1/shape?table=items&live=true&experimental_live_sse=true&handle=${handle}&offset=${offset}`
const es = new EventSource(url)
es.onerror = () => {
  es.close()
}
```

Note that the HTTP headers are returned at the start of the response. This means that the current header mechanism to return the next offset isn't valid. At the moment, you get a response like this:

```
curl -v 'http://localhost:3000/v1/shape?table=items&handle=22194952-1743787019364&live=true&experimental_live_sse=true&offset=5501319691220011464_0'
> GET /v1/shape?table=items&handle=22194952-1743787019364&live=true&experimental_live_sse=true&offset=5501319691220011464_0 HTTP/1.1
> Host: localhost:3000
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 200 OK
< transfer-encoding: chunked
< date: Sun, 06 Apr 2025 11:12:45 GMT
< cache-control: public, max-age=59
< x-request-id: GDO2V0hAm4_Dp5cAABBj
< electric-server: ElectricSQL/1.0.5
< access-control-allow-origin: *
< access-control-expose-headers: *
< access-control-allow-methods: GET, HEAD, DELETE, OPTIONS
< content-type: text/event-stream
< electric-cursor: 15505980
< etag: "22194952-1743787019364:5501319691220011464_0:5501319691220011464_0"
< electric-handle: 22194952-1743787019364
< electric-up-to-date:
< electric-offset: 5501319691220011464_0
< connection: keep-alive
<
data: {"value":{"id":"9a33ad9e-39a6-4ab4-aebf-4a4c54c8dbc6"},"key":"\"public\".\"items\"/\"9a33ad9e-39a6-4ab4-aebf-4a4c54c8dbc6\"","headers":{"last":true,"relation":["public","items"],"operation":"insert","lsn":"5501319691220014840","op_position":0,"txids":[792]}}

[...]

PR #2776, opened 2025-05-26
feat: support for Server-Sent Events (SSE)

kevin-dp:
This is a follow up PR on https://github.com/electric-sql/electric/pull/2546 and https://github.com/electric-sql/electric/pull/2544. It solves a bug related with 409s (must refetch) in SSE mode and it replaces the EventSource browser API by the [fetch-event-source](https://github.com/Azure/fetch-event-source) library. I refactored the `ShapeStream.#start` method which was becoming very big and complex. To this end, i split the logic into helper methods that handle the different parts that need to happen (building the shape URL, making the request, parsing the response headers, handling the response body, etc.).

I had to patch the [fetch-event-source](https://github.com/Azure/fetch-event-source) library because it relies on browser-specific features such as `document` and `window` (cf. https://github.com/Azure/fetch-event-source/pull/41). But we want our client to also work in server-side JS environments.

I also had to patch the `fetch-event-source` library because it does not abort the fetch when you pass an already aborted signal. A complete description of the bug and the fix can be found here: https://github.com/Azure/fetch-event-source/issues/98.

To use, add `experimentalLiveSSE` to your shape options e.g.

```tsx
useShape({
   url,
   experimentalLiveSSE: true
 })
```

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -672,17 +672,14 @@ export class ShapeStream<T extends Row<unknown> = Row>
     requestAbortController: AbortController
     headers: Record<string, string>
   }): Promise<void> {
-      if (
-        this.#isUpToDate &&
-        this.options.experimentalLiveSse &&
-        !this.#isRefreshing
-      ) {
-          opts.fetchUrl.searchParams.set(
-            EXPERIMENTAL_LIVE_SSE_QUERY_PARAM,
-            `true`
-          )
-        return this.#requestShapeSSE(opts)
-      }
+    if (
+      this.#isUpToDate &&
+      this.options.experimentalLiveSse &&
+      !this.#isRefreshing
+    ) {
+      opts.fetchUrl.searchParams.set(EXPERIMENTAL_LIVE_SSE_QUERY_PARAM, `true`)
+      return this.#requestShapeSSE(opts)
+    }
 
     return this.#requestShapeLongPoll(opts)
   }
@@ -703,7 +700,7 @@ export class ShapeStream<T extends Row<unknown> = Row>
 
     const schema = this.#schema! // we know that it is not undefined because it is set by `this.#onInitialResponse`
     const res = await response.text()
-    const messages = (res) || `[]`
+    const messages = res || `[]`
 
     await this.#onMessages(messages, schema)
   }
@@ -719,7 +716,7 @@ export class ShapeStream<T extends Row<unknown> = Row>
       await fetchEventSource(fetchUrl.toString(), {
         headers,
         fetch,
-        onopen: async (response) => {
+        onopen: async (response: Response) => {
           this.#connected = true
           await this.#onInitialResponse(response)
         },
--- a/packages/typescript-client/test/client.test.ts
+++ b/packages/typescript-client/test/client.test.ts
@@ -83,73 +83,79 @@ describe(`Shape`, () => {
     }
   )
 
-  it.for(fetchAndSse)(`should continually sync a shape/table (liveSSE=$experimentalLiveSse)`, async ({ experimentalLiveSse }, {
-    issuesTableUrl,
-    insertIssues,
-    deleteIssue,
-    updateIssue,
-    waitForIssues,
-    aborter,
-  }) => {
-    const [id] = await insertIssues({ title: `test title` })
-
-    const expectedValue1 = [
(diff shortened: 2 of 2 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -41,6 +41,10 @@ import {
   FORCE_DISCONNECT_AND_REFRESH,
   EXPERIMENTAL_LIVE_SSE_QUERY_PARAM,
 } from './constants'
+import {
+  EventSourceMessage,
```
KyleAMathews, line 45: Let's not load this node polyfill in the browser. It can make behavior inconsistent and it's extra code. We can detect when we're in node and conditionally load the polyfill.

KyleAMathews (changes requested):

packages/typescript-client/src/client.ts
kevin-dp: This is not a node polyfill. This is a library that implements SSE via standard fetch rather than via the browser's EventSource API which is very limited.

packages/typescript-client/src/helpers.ts
```diff
@@ -51,3 +51,15 @@ export function isUpToDateMessage<T extends Row<unknown> = Row>(
 ): message is ControlMessage & { up_to_date: true } {
   return isControlMessage(message) && message.headers.control === `up-to-date`
 }
+
+/**
+ * Parses the LSN from the up-to-date message and turns it into an offset.
+ * The LSN is only present in the up-to-date control message when in SSE mode.
+ * If we are not in SSE mode this function will return undefined.
+ */
+export function getOffset(message: ControlMessage): Offset | undefined {
```
samwillis, line 60: I'm not sure we should be using the LSN to compute an offset in the client. This is an internal server side implementation detail that is quite likely to change.

Would it be better to include the offset on each message from the server when in sse mode?

kevin-dp: That would indeed be better. However, this is what @thruflo documented the client should do in https://github.com/electric-sql/electric/pull/2544#issue-2974965001 and you also did this in your original implementation: https://github.com/electric-sql/electric/pull/2546. Do you want me to change the server such that the server returns an `electric-offset` in the body rather than the current `global_last_seen_lsn` ? (cc @samwillis  @thruflo)

packages/typescript-client/src/client.ts
kevin-dp: @KyleAMathews I don't think anyone has looked into how SSE will work with a proxy. It will likely be more complicated than the current proxy pattern but it should be feasible to accept incoming SSE requests, then use `fetch-event-source` to proxy it through, and then in the `onopen`, `onmessage`, and `onclose` handlers forward those SSE events to the user.

thruflo: If the proxy is just passing the headers and streaming the response through then it should just work™️. I.e.: the fact that the connection stays open and the response streams through in parts should be transparent to most proxies?

packages/typescript-client/src/helpers.ts
thruflo: Would this require additional deserialisation / serialisation? As in, to special case it for the SSE? Where are we on providing this info generally -- perhaps we can push the change out of scope and revisit the SSE implementation when we firm up how we expose the offset generally?

packages/typescript-client/src/client.ts
KyleAMathews: Yeah I did some research and there's good support. And given this will always be an optional "fast" mode — electric will always work in basic http long-polling.

wadefletch:
Is there a timeline or plan for releasing this functionality, even if behind a flag?

alco:
@kevin-dp I suspect this was supposed to be merged into `main` but it ended up getting merged into @samwillis's branch and never made its way to `main`. Though it is a bit odd there's no changeset in this PR.

KyleAMathews:
Oh I thought this was released — @kevin-dp what's the next step here?

kevin-dp:
That's odd. Let's re-open @samwillis' PR in which we merged this one and then merge that one into main? @KyleAMathews

KyleAMathews:
👍 yeah sounds good :shipit:

kevin-dp:
@wadefletch I merged SSE support into main last week, cf. https://github.com/electric-sql/electric/pull/2856

PR #2779, opened 2025-05-26
fix: account for hash collisions in across shape definitions

### LINKED ISSUE #2778: Shape hash collisions shouldn't yield same shape handle
**Versions**
- Electric: 1.0.16 & earlier

**Bug description**
When 2 shapes have the same hash, they resolve to the same (cached) shape handle.

**Expected behavior**
We must never return same shape handle for different shapes, as that leads to data from another shape being shown.

icehaunter:
Closes #2778

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -93,7 +93,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
 
   @spec add_shape(t(), Shape.t()) :: {:ok, shape_handle()} | {:error, term()}
   def add_shape(state, shape) do
-    {hash, shape_handle} = Shape.generate_id(shape)
+    {_, shape_handle} = Shape.generate_id(shape)
     # For fresh snapshots we're setting "latest" offset to be a highest possible virtual offset,
     # which is needed because while the snapshot is being made we DON'T update this ETS table.
     # We could, but that would required making the Storage know about this module and I don't like that.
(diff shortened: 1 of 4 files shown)


icehaunter:
Benchmark this

alco (approved):

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -322,11 +322,10 @@ defmodule Electric.ShapeCache.ShapeStatus do
         state.shape_meta_table,
         Enum.concat([
           Enum.flat_map(shapes, fn {shape_handle, shape} ->
-            hash = Shape.hash(shape)
             relations = Shape.list_relations(shape)
 
             [
-              {{@shape_hash_lookup, hash}, shape_handle},
+              {{@shape_hash_lookup, Shape.comparable(shape)}, shape_handle},
```
msfstef, line 328: does this have a significant memory and/or performance impact? We are now keeping the whole shape as a key, and I don't know if it also impacts lookup performance in ETS in any way?

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -103,7 +103,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
       :ets.insert_new(
         state.shape_meta_table,
         [
-          {{@shape_hash_lookup, hash}, shape_handle},
+          {{@shape_hash_lookup, Shape.comparable(shape)}, shape_handle},
```
msfstef, line 106: nit: can you add this to `remove_shape` to avoid walking the table as well as part of this? (the `:ets.select_delete` call)

msfstef (approved):
Looks good! This looks backwards compatible, but just to double check would previous shape handles be untouched and can be maintained as they are (from disk)?

PR #2780, opened 2025-05-26
fix: Avoid going to CubDB to calculate disk usage

msfstef:
Closes https://github.com/electric-sql/electric/issues/2769

Avoids going o CubDB to calculate the disk usage, thus not requiring the process registry to be alive and we don't have to deal with all the ways in which the genserver call could go wrong.

An alternative would be to move the registry first, then telemetry, then the rest of the stack.


icehaunter (approved):

alco (approved):
:+1:

PR #2791, opened 2025-05-30
Fix lock semantics for the exclusive connection lock when running against a pooler

alco:
We have recently observed some source in Electric Cloud struggle due to lock and replication slot contention issues which in theory shouldn't have happened at all. The cause of the issues was an imperfect validation logic applied to the creation of new sources. As a consequence, some sources get created with a pooled connection URL.

More concretely, our validation can successfully weed out pooled connection strings from Supabase, but Neon's pool allows establishing replication connections by proxying them directly to the DB. As a result, Electric ends up acquiring an advisory lock on a pooled connection where the pooler runs in transaction mode, leading to a bunch of undesirable effects that prevent Electric's normal operation.

---

This PR addresses the cause of the issues by making the `LockConnection` process open a replication connection to the DB or fail if that cannot be achieved. In this way, either the advisory lock will be taken on a long-living connection as expected, or the connection won't be able to open and the source will immediately transition into an error state, instead of assigning an advisory lock to a random pooled connection as it currently does.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -111,6 +111,8 @@ defmodule Electric.Connection.Manager do
       :pg_system_identifier,
       # PostgreSQL timeline ID
       :pg_timeline_id,
+      # PostgreSQL's current WAL flush position at lock connection startup
+      :pg_wal_flush_lsn,
       # ID used for process labeling and sibling discovery
       :stack_id,
       # Registry used for stack events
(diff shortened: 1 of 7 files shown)


magnetised (approved):
Fantastic stuff.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -832,24 +838,35 @@ defmodule Electric.Connection.Manager do
...
   end
 
-  def handle_cast({:pg_info_looked_up, {server_version, system_identifier, timeline_id}}, state) do
+  def handle_cast({:pg_system_info_obtained, info}, state) do
+    {:noreply,
+     %State{
+       state
+       | pg_system_identifier: info.system_identifier,
+         pg_timeline_id: info.timeline_id,
+         pg_wal_flush_lsn: info.current_wal_flush_lsn
```
msfstef, line 847: since we never update this value (?), would we want to somehow specify this parameter a bit more like `pg_initial_wal_flush_lsn` or `pg_wal_flush_lsn_at_connection_time` or something like that?

alco: Good point! It'll be less confusing to just not have it in the state to begin with. https://github.com/electric-sql/electric/pull/2818

PR #2793, opened 2025-06-02
wip: respect order_by and limit on initial query.

thruflo:
The server-side of the on-site hackathon work on respecting order_by and limit on initial shape query. @balegas added equivalent client-side support.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api/params.ex
+++ b/packages/sync-service/lib/electric/shapes/api/params.ex
@@ -45,6 +45,11 @@ defmodule Electric.Shapes.Api.Params do
     field(:handle, :string)
     field(:live, :boolean, default: false)
     field(:where, :string)
+
+    # XXX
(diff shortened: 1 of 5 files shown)


balegas:
do we want to merge this now? I'll can add the ts side of it after fixing tests

thruflo:
No, we definitely don't want to merge this as is. I think the generic sub-query approach we designed on day 2 is a much more viable approach. I've pushed this just so the code exists somewhere as a reference if we want to see what was involved in threading the new query params through.

msfstef:
Would this also not imply that the `order_by` and `limit` are respected by subsequent log updates? It's simple to apply it to the initial snapshot but that doesn't feel like much of a guarantee, especially when we compact etc

icehaunter:
I think this is superseded by the RFC proposal and implementation in #3159. Please reopen if I'm mistaken

PR #2796, opened 2025-06-02
elixir-client: Add support for array params

magnetised:
Fixes https://github.com/electric-sql/phoenix_sync/issues/57


packages/elixir-client/test/electric/client/ecto_adapter/postgres_test.exs
```diff
@@ -47,9 +47,65 @@ defmodule Electric.Client.EctoAdapter.PostgresTest do
...
+    )
+
+    assert_where(
+      from(t in TestTable) |> where([t], t.mm in ^values_mm),
+      ~s|("mm" IN (3140000,2710000,6626000))|
+    )
+
+    assert_where(
+      from(t in TestTable) |> where([t], t.mm in ^values_mm or t.mm in ^values_mm2),
+      ~s|("mm" IN (3140000,2710000,6626000) OR "mm" IN (9990000))|
```
alco, line 89: Why do values like `Decimal.new("3.14")` get converted to `3140000` and similar in the resulting SQL?

magnetised: @alco The `Money` type (which is an Ecto.ParameterizedType` impl) dumps its values as 

```
decimal_val
|> Decimal.mult(1_000_000)
|> Decimal.to_integer()
```

so we're testing that the shape where clauses are generated using the values as cast by the ecto types.

alco: Ah, I get it now. Thanks.

PR #2799, opened 2025-06-03
feat: add gc metrics

icehaunter:


packages/sync-service/lib/electric/debug/process.ex
```diff
@@ -24,7 +24,7 @@ defmodule Electric.Debug.Process do
     |> Enum.take(count)
   end
 
-  defp type_and_memory(pid) do
+  def type_and_memory(pid) do
```
robacourt: This seems an odd function to make public, especially as all you're really after is the type. There used to be a public function `process_type/1` for this exact purpose, but it looks like it's become `process_type/2`. I would reintroduce `process_type/1` rather than expose `type_and_memory/1`

robacourt: I've taken the liberty of fixing this myself. Feel free to revert/change if you wish

PR #2801, opened 2025-06-04
elixir-client: Build shape definition from changeset

magnetised:
Allows for generating a ShapeDefinition from an Ecto.Changeset

Also bundles a few quality-of-life changes:

- ShapeDefinition.params/1 returns a `%{binary => binary}` map in query string mode, rather than a mix of atom and binary keys
- Widen accepted types for where parameters from `:string` to [:string, :integer, :float, :boolean]` to make life simpler for user. these types are trivial to convert to strings so why not do that for people
- Give ShapeDefinitions a `replica` setting. This is how it should have been to begin with -- the replica mode is a property of the shape, not of the stream
- Expose the ShapeDefinition validation schema so we can split options and validate partial shape parameters (Phoenix.Sync.PredefinedShape accepts a big bag of options that we need to assign appropriately)

### DIFF
--- a/packages/elixir-client/lib/electric/client/shape_definition.ex
+++ b/packages/elixir-client/lib/electric/client/shape_definition.ex
@@ -3,12 +3,12 @@ defmodule Electric.Client.ShapeDefinition do
   Struct for defining a shape.
 
       iex> ShapeDefinition.new("items", where: "something = true")
-      {:ok, %ShapeDefinition{table: "items", where: "something = true"}}
+      {:ok, %ShapeDefinition{table: "items", where: "something = true", replica: :default}}
   """
 
   alias Electric.Client.Util
 
-  @public_keys [:namespace, :table, :columns, :where, :params]
+  @public_keys [:namespace, :table, :columns, :where, :params, :replica]
   @derive {Jason.Encoder, only: @public_keys}
   @enforce_keys [:table]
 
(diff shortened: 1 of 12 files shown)


.github/workflows/elixir_tests.yml
```diff
@@ -102,7 +102,7 @@ jobs:
         run: |
           cp -R _build/test _build/application_test
           rm -rf _build/application_test/lib/electric
-          MIX_TARGET=application mix coveralls.json --only telemetry_target --import-cover cover
+          MIX_TARGET=application mix do deps.get + coveralls.json --only telemetry_target --import-cover cover
```
alco: This isn't needed. We copy built deps from the test environment to make them available for the application target. And locally I can run telemetry tests just fine without any extra shenanigans with deps.

magnetised: It worked for me locally so thought it might help here

packages/elixir-client/lib/electric/client/shape_definition.ex
```diff
@@ -14,6 +14,9 @@ defmodule Electric.Client.ShapeDefinition do
 
   defstruct @public_keys ++ [parser: {Electric.Client.ValueMapper, []}]
 
+  # only allow things that are trivially convertable to strings
+  @params_types {:or, [:string, :integer, :float, :boolean]}
```
alco, line 18: Shouldn't types like Decimal or Money be part of this too? Or are they handled elsewhere?

alco (approved):
Cool!

packages/elixir-client/lib/electric/client/shape_definition.ex
magnetised: I expanded the support from just strings to allow for some simple types to save the user having to do e.g. `to_string(true)`. Might be possible to just accept anything that implements String.Chars or any Ecto type but for the moment I'm stopping here...

Issue #2727, opened 2025-05-15
Handle errors from Postgres pool


balegas:
We have a few issues related with handling errors from inside Postgres Pool. We're not sure if we can setup handlers for them of we need to filter them out from Sentry. Also we have a few different ideas how we could filter them out.

Let's research and find a solution

PR #2804, opened 2025-06-05
Catch pooled connection errors in Connection.Manager process

### LINKED ISSUE #1554: Improve db connection error visibility
With the move to per-shape storage and consumers, deleting a shape while the snapshot is being produced will result in the associated db connection process exiting. 

Our previous strategy of just allowing the entire pool to shutdown with a connection error was hence a problem, so we reverted back to the default exponential backoff, to prevent connection errors from bringing down the entire pool.

This means that we don't get good information about connection errors any more, which could be important. 

Some ideas re getting connection failure errors:

> From the docs it seems that the connection pool is a regular OTP supervisor and connection processes are its children. See the note in the last paragraph of this function's doc - https://hexdocs.pm/db_connection/DBConnection.html#disconnect_all/3.
> 
> So if we handle the connection process exit in ConnectionManager and if it's not fatal we keep the pool running, the latter should just start a new connection to go back to the steady state.

https://hexdocs.pm/db_connection/2.4.1/DBConnection.html#start_link/2-telemetry

> A [:db_connection, :connection_error] event is published whenever a connection checkout receives a %DBConnection.ConnectionError{}.

alco:
Fix #1554, #2727.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -87,6 +87,8 @@ defmodule Electric.Connection.Manager do
       :shared_connection_opts,
       # Database connection pool options
       :pool_opts,
+      # Unique ref for matching on :connected and :disconnected events the connection pool
+      :pool_ref,
       # Options specific to `Electric.Timeline`
       :timeline_opts,
       # Options passed to the Replication.Supervisor's start_link() function
@@ -247,6 +249,7 @@ defmodule Electric.Connection.Manager do
         current_phase: :connection_setup,
         current_step: {:start_lock_connection, nil},
         pool_opts: pool_opts,
+        pool_ref: make_ref(),
         timeline_opts: timeline_opts,
         shape_cache_opts: shape_cache_opts,
         connection_backoff: {ConnectionBackoff.init(1000, 10_000), nil},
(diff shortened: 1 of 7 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -683,6 +691,43 @@ defmodule Electric.Connection.Manager do
...
+      {:connection_error,
+       %{
+         error: DbConnectionError.format_original_error(error),
+         type: error.type,
+         message: error.message
+       }},
+      state
+    )
+
+    # If the error is of an unknown type, it would have already been logged by DbConnectionError itself.
```
balegas, line 720: We've been chatting back and forth about disabling library logging. I think with this PR and catching all the unknown errors from the lib we can disable it safely. Shall we do that as part of this PR? cc @robacourt

robacourt: Sounds like a good idea to me

alco: Sorry, I'm not sure I get what is meant by disabling library logging here.

robacourt: 1. Set `capture_log_messages` to false 
2. Come up with an alternative means to send errors to Sentry when we want to (when we currently just log an error)

Because 2 is not trivial, perhaps it should be a separate PR.

PR #2815, opened 2025-06-09
fix(elixir-client): Add handling of Ecto.ULID types

magnetised:
ULID fields expect to `cast/1` a base-32 encoded value but since the backing column type is uuid, the values in the replication stream are the string representation of a uuid, so trying to cast these raises. This pr adds special handling for custom types that use UUID-type columns and attempt to simulate a load of a raw uuid value from the db.

Fixes https://github.com/electric-sql/phoenix_sync/issues/24


packages/elixir-client/lib/electric/client/ecto_adapter.ex
```diff
@@ -118,13 +118,77 @@ if Code.ensure_loaded?(Ecto) do
...
+    defp d(?D), do: 13
+    defp d(?E), do: 14
+    defp d(?F), do: 15
+    defp d(?a), do: 10
+    defp d(?b), do: 11
+    defp d(?c), do: 12
+    defp d(?d), do: 13
+    defp d(?e), do: 14
+    defp d(?f), do: 15
+    defp d(_), do: throw(:error)
```
alco: Since the function is inlined, it could perhaps be defined in a more concise way?
```
def d(char) when char in ?0..?9, do: char - ?0
def d(char) when char in ?a..?f, do: char - ?a + 10
def d(char) when char in ?A..?F, do: char - ?A + 10
```

On a separate note, the `throw(:error)` looks out of place here since it's not caught by a `try ... catch`. It shouldn't it be `raise Ecto.CastError` instead?

packages/elixir-client/lib/electric/client/ecto_adapter.ex
```diff
@@ -239,29 +239,16 @@ if Code.ensure_loaded?(Ecto) do
...
-    defp d(?E), do: 14
-    defp d(?F), do: 15
-    defp d(?a), do: 10
-    defp d(?b), do: 11
-    defp d(?c), do: 12
-    defp d(?d), do: 13
-    defp d(?e), do: 14
-    defp d(?f), do: 15
-    defp d(_), do: throw(:error)
+    for {r, o} <- [{?0..?9, 0}, {?A..?F, 10}, {?a..?f, 10}], {c, i} <- Enum.with_index(r, o) do
```
alco, line 263: I feel like you're mocking me here... point taken.

alco (approved):

packages/elixir-client/lib/electric/client/ecto_adapter.ex
magnetised: no mocking intended! far from it. I just like playing with code code generation and went with the most fun impl...

alco: I see. To me, this looks like you've taken it a bit too far with code density. But my original comment was too subjective already, the code had been fine as originally written, tbh.

PR #2819, opened 2025-06-10
Retire database_connection_failed in favour of connection_error

robacourt:


msfstef (approved):
Looks good and good catch on the total retry time!

We should make sure we don't merge this to cloud before it starts handling `connection_error` though

Also, needs a changeset

PR #2822, opened 2025-06-11
sync-service: Only warn when publication manager exits during shape cleanup

magnetised:
Either the publication manager has crashed and everything will be restored once it starts up or we're removing a source and the publication manager has been shutdown before or during this task run

Fixes https://github.com/electric-sql/stratovolt/issues/676

### DIFF
--- a/packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
+++ b/packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
@@ -128,14 +128,20 @@ defmodule Electric.Shapes.Monitor.CleanupTaskSupervisor do
     !is_nil(Electric.Shapes.Consumer.whereis(stack_id, shape_handle))
   end
 
-  # don't spam test logs with failures due to process shutdown
   if @env == :test do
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
```diff
@@ -128,14 +128,20 @@ defmodule Electric.Shapes.Monitor.CleanupTaskSupervisor do
...
 
-  # don't spam test logs with failures due to process shutdown
   if @env == :test do
+    # don't spam test logs with failures due to process shutdown
     defp log_error(:exit, _message) do
       :ok
     end
+  else
+    # don't spam sentry with errors caused by shutdown order (i.e. when the
+    # publication manager has been shutdown)
```
msfstef, line 138: I think we want to be more particular with what we log as a warning (?) - exits because of shutdowns or no proc are expected and harmless, but if an exit is caused by an actual bug when calling `remove_shape` we would still want it to show up as an error no? as there is a chance that it can repeatedly exit because of an actual bug and we won't have visibility over that

msfstef (approved):

PR #2824, opened 2025-06-11
fix: account for ETS matching maps when comparing shapes

icehaunter:


packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -68,13 +74,10 @@ defmodule Electric.Shapes.Shape do
...
 
   def comparable(%__MODULE__{} = shape) do
-    shape
-    |> Map.drop([:table_info, :storage])
-    |> Map.update!(:where, fn
-      nil -> nil
-      %Expr{} = expr -> Electric.Shapes.Shape.Comparable.comparable(expr)
-      statement when is_binary(statement) -> statement
-    end)
+    {:shape, {shape.root_table_id, shape.root_table}, shape.root_pk,
```
msfstef, line 87: nit: this is the only thing I was slightly worried about - we need to make sure this tuple is kept up to date with any more fields/info we end up adding to the shape. Maybe a docstring with a little bit of an explanation might help here?

msfstef (approved):
Looks good, nice catch ! 🚀 

Fixes https://github.com/electric-sql/electric/issues/2820

packages/sync-service/lib/electric/shapes/shape.ex
icehaunter: Done

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -67,14 +73,21 @@ defmodule Electric.Shapes.Shape do
...
+     Comparable.comparable(shape.where), shape.selected_columns,
+     Enum.flat_map(shape.flags, fn {k, v} -> if(v, do: [k], else: []) end) |> Enum.sort(),
```
alco, line 89: Not in scope for this PR but it would be better to turn `shape.flags` into a keyword list that gets sorted and normalized whenever it is updated, instead of doing this mapping and sorting every time we do any lookups for the shape.

balegas: if we don't open an issue for doing it, we'll never do it :)

PR #2829, opened 2025-06-12
feat: allow publication alter debounce to be configurable

icehaunter:
- Introduce a tweak variable to debounce `ALTER PUBLICATION` calls
  - This is currently required for some setups against AWS aurora
- Fix weird startup order issue
  - We did start schema reconcilier, then schema cache, but essentially first functional line of the reconciler was to call into the schema cache, which was blocked on startup until it could refresh the publication and such. Putting the reconcilier behind schema cache in order makes it so our startup is more linear, and any shape requests that come in during startup (that need a publication alter) will be debounced together with the schema reconciler normalizing the publication

### DIFF
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -202,7 +202,13 @@ config :electric,
   profile_where_clauses?: env!("ELECTRIC_PROFILE_WHERE_CLAUSES", :boolean, false),
   persistent_kv: persistent_kv_spec,
   listen_on_ipv6?: env!("ELECTRIC_LISTEN_ON_IPV6", :boolean, nil),
-  secret: secret
+  secret: secret,
+  publication_alter_debounce_ms:
+    env!(
+      "ELECTRIC_PUBLICATION_ALTER_DEBOUNCE_TIME",
+      &Electric.Config.parse_human_readable_time!/1,
+      nil
+    )
 
 if Electric.telemetry_enabled?() do
   config :sentry,
--- a/.changeset/tidy-days-matter.md
+++ b/.changeset/tidy-days-matter.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+feat: allow publication alter debounce to be configurable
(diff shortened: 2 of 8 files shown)


packages/sync-service/config/runtime.exs
```diff
@@ -202,7 +202,8 @@ config :electric,
   profile_where_clauses?: env!("ELECTRIC_PROFILE_WHERE_CLAUSES", :boolean, false),
   persistent_kv: persistent_kv_spec,
   listen_on_ipv6?: env!("ELECTRIC_LISTEN_ON_IPV6", :boolean, nil),
-  secret: secret
+  secret: secret,
+  publication_alter_debounce_ms: env!("ELECTRIC_PUBLICATION_ALTER_DEBOUNCE_MS", :integer, nil)
```
alco: We have some inconsistency in the values time-based config settings expect. There's `ELECTRIC_SHAPE_HIBERNATE_AFTER`, `ELECTRIC_SYSTEM_METRICS_POLL_INTERVAL`, `ELECTRIC_OTEL_EXPORT_PERIOD`, all of which are not documented and use `Electric.Config.parse_human_readable_time!()` to parse time. So configuring them with a bare number means milliseconds.

Then there are two public settings `ELECTRIC_CACHE_MAX_AGE` and `ELECTRIC_CACHE_STALE_AGE`, both can only take a number and that value is interpreted as seconds.

Introducing a configuration setting with the units in its name is a double-edged sword that in this case would also add one more convention into the mix. Could the name for this new setting be `ELECTRIC_PUBLICATION_REFRESH_AFTER` or `ELECTRIC_PUBLICATION_UPDATE_DELAY`. If it's important to keep the word `debounce` in the name, I'd suggest the `DEBOUNCE_TIME` suffix.

alco (approved):
Great!

The only feedback I've got is a bit of rambling on configuration setting naming.

packages/sync-service/config/runtime.exs
icehaunter: Yeah, fair point! I'll probably switch to `DEBOUNCE_TIME` as you suggest, but also will probably switch to throttle instead of debounce as suggested by @KyleAMathews

PR #2836, opened 2025-06-17
feat(sync-service): Configure GC telemetry via env vars

robacourt:


icehaunter (approved):
God these names are getting ridiculous

PR #2837, opened 2025-06-17
feat(sync-service): GenStage telemetry

robacourt:
Current replication processing time telemetry records overall time and breaks down by it's various parts, however there are gaps in this breakdown and data from trigger.dev shows that a significant amount of time is spent in these gaps.

This PR times all the following 7 intervals (in this order):
1. message_to_collector
2. new_txn_handle_call
3. gen_stage_receive_txn
4. partitions.handle_event
5. dispatcher.dispatch
6. gen_stage.post_dispatch
7. message_from_collector

3 of these intervals were already available: message_to_collector, partitions.handle_event and dispatcher.dispatch however are now using the new IntervalTimer for consistency.

The new IntervalTimer (and it's sibling ProcessIntervalTimer) have been introduced as many of these intervals are impossible to put spans around due to way GenStage code is structured. ProcessIntervalTimer additionally uses process memory for the times when the timer state cannot be passed around (again due to how GenStage code is structured)

### DIFF
--- a/packages/sync-service/lib/electric/telemetry/interval_timer.ex
+++ b/packages/sync-service/lib/electric/telemetry/interval_timer.ex
@@ -1,0 +1,31 @@
+defmodule Electric.Telemetry.IntervalTimer do
+  @state_key :timed_intervals
+
+  def state do
+    Process.get(@state_key, [])
+  end
+
+  def set_state(state) do
+    Process.put(@state_key, state)
+  end
+
+  def start(interval) do
+    [{interval, time()} | state()]
+    |> set_state()
+  end
+
+  def intervals do
+    intervals([time() | state()])
+    |> Enum.reverse()
+  end
+
+  defp intervals([end_time, {interval, start_time} | rest]) do
+    [{interval, end_time - start_time} | intervals([start_time | rest])]
+  end
+
+  defp intervals([_end_time]), do: []
+
+  defp time do
+    System.monotonic_time(:microsecond)
+  end
+end
--- a/packages/sync-service/lib/electric/telemetry/interval_timer.ex
+++ b/packages/sync-service/lib/electric/telemetry/interval_timer.ex
@@ -6,12 +6,12 @@ defmodule Electric.Telemetry.IntervalTimer do
 
   @default_state []
 
-  def start_interval(state \\ nil, interval) do
-    [{interval, time()} | state || @default_state]
+  def start_interval(state \\ nil, interval_name) do
+    [{interval_name, time()} | state || @default_state]
   end
 
   def durations(state) do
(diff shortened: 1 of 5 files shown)


robacourt:
benchmark this

packages/sync-service/lib/electric/telemetry/interval_timer.ex
```diff
@@ -0,0 +1,31 @@
...
+    |> set_state()
+  end
+
+  def intervals do
+    intervals([time() | state()])
+    |> Enum.reverse()
+  end
+
+  defp intervals([end_time, {interval, start_time} | rest]) do
+    [{interval, end_time - start_time} | intervals([start_time | rest])]
```
alco: Could you explain the recursive logic here in prose? I'm failing to guess how this is used in practice.

packages/sync-service/test/electric/telemetry/interval_timer_test.exs
```diff
@@ -0,0 +1,32 @@
...
+        IntervalTimer.start("C")
+        Process.sleep(@c_time)
+
+        IntervalTimer.intervals()
+      end)
+
+    assert [{"A", a_time}, {"B", b_time}, {"C", c_time}] = intervals
+
+    assert a_time >= @a_time * @μs_per_ms
+    assert b_time >= @b_time * @μs_per_ms
```
alco, line 28: I think this needs to also assert that `b_time < (@b_time + @a_time) * @μs_per_ms` to verify that interval "B" started after the `@a_time` delay.

packages/sync-service/lib/electric/telemetry/interval_timer.ex
alco: Ah, I think I get it: if we're tracking multiple intervals one after the other, the moment when a new interval is started is considered that `end_time` for the previous interval.

robacourt: It does look confusing. I've just refactored it and added a comment. Hopefully it's more readable now.

alco: Looks great. Thanks!

packages/sync-service/test/electric/telemetry/interval_timer_test.exs
robacourt: As long as `a_time + b_time + c_time <= total + @rounding_error` consistently passes, I'm not sure how that could fail.

alco: Makes sense.

robacourt:
Closed since #2845 allows for much simpler telemetry

PR #2839, opened 2025-06-18
Dispatch a stack event when the replication client is blocked on a pending transaction

alco:


robacourt (approved):

balegas:
This sound useful to help the user understand why the source is not finished connecting.

PR #2840, opened 2025-06-18
fix(sync-service): Fix issue where replication traffic slows down shape restoration

### LINKED ISSUE #2767: Replication traffic slows down shape restoration
Currently Electric does not wait for shape restoration to finish before starting the replication stream. This leads to the ShapeLogCollector being simultaneously bombarded by replication messages (which are just buffered on the ShapeLogCollector) and shape consumer subscriptions, slowing down the restoration time and potentially leading to a restoration timeout or even out of memory on the ShapeLogCollector buffer.

The replication stream should wait for replication to finish before starting streaming.

robacourt:
Fixes #2767 by not starting the replication client streaming until the consumers are ready. This avoids the need for the consumers to buffer messages until ready, which also reduces memory consumption and uses one less feature of GenStage making it easier to replace.


alco (approved):
Nice.

PR #2841, opened 2025-06-18
chore: Ensure test processes are supervised by the test supervisor

msfstef:
Ensures that all processes started are started under the ExUnit test supervisor, in the hope that this gives us better guarantees about cleanups occurring at the right time and we reduce flakes that seem to potentially be related to timing issues.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -215,6 +215,12 @@ defmodule Electric.Postgres.ReplicationClient do
     {:noreply, state}
   end
 
+  # This callback is invoked when the connection process receives a shutdown signal.
+  def handle_info({:EXIT, _pid, :shutdown}, _state) do
+    Logger.debug("Replication client #{inspect(self())} received shutdown signal, stopping")
+    {:disconnect, :shutdown}
+  end
+
   # The implementation of Postgrex.ReplicationConnection doesn't give us a convenient way to
   # check whether the START_REPLICATION_SLOT statement succeeded before switching the
   # connection into streaming mode. Returning {:query, "START_REPLICATION_SLOT ...", state}
--- a/packages/sync-service/test/electric/shape_cache_test.exs
+++ b/packages/sync-service/test/electric/shape_cache_test.exs
@@ -1223,8 +1223,12 @@ defmodule Electric.ShapeCacheTest do
       )
     end
 
-    defp stop_shape_cache(_ctx) do
-      for name <- [ShapeCache, ShapeLogCollector, Shapes.DynamicConsumerSupervisor] do
+    defp stop_shape_cache(ctx) do
+      for name <- [
+            ctx.shape_cache_opts[:server],
+            ctx.consumer_supervisor,
+            ctx.shape_log_collector
+          ] do
         stop_supervised(name)
       end
     end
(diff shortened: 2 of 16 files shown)


packages/sync-service/test/electric/shape_cache_test.exs
```diff
@@ -1224,50 +1224,12 @@ defmodule Electric.ShapeCacheTest do
...
-          {:DOWN, _, :process, ^pid, :killed} -> :process_killed
-        after
-          500 -> raise "#{name} process not killed"
-        end
+      for name <- [
+            ctx.shape_cache_opts[:server],
+            ctx.consumer_supervisor,
+            ctx.shape_log_collector
+          ] do
+        stop_supervised(name)
```
robacourt, line 1232: Hell yeah!

robacourt (approved):
Nice work!

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -215,6 +215,12 @@ defmodule Electric.Postgres.ReplicationClient do
     {:noreply, state}
   end
 
+  # This callback is invoked when the connection process receives a shutdown signal.
+  def handle_info({:EXIT, _pid, :shutdown}, _state) do
```
alco, line 219: Why is this needed and how does this work? I don't see the `trap_exit` process flag getting set neither here nor in Postgrex.ReplicationConnection.

Is it set via the `Postgrex.Protocol.connect()` call inside `Postgrex.ReplicationConnection`? But then it's an implementation detail. Why does it have to manifest itself in our code?

msfstef: Should have left a more explicit comment - this was added because [`start_link_supervised!`](https://hexdocs.pm/ex_unit/main/ExUnit.Callbacks.html#start_link_supervised!/2) propagates the a `:shutdown` exit signal, and apparently both `Postgrex.ReplicationConnection` and `Postgrex.SimpleConnection` see this. The trapping of exits is being set by `Postgrex.Protocol.connect()` as you pointed out for both of them.

> def connect(opts) do
    # Trap exits so that DBConnection calls `disconnect` on unexpected shutdowns

For our `LockConnection` tests I avoided this by using `start_supervised!` with `restart: :temporary` since the relevant tests would fail their assertions if the lock connection died anyway, but for the replication client tests (that were more numerous) I wanted to ensure they would fail if it crashed for whatever reason, so I figured I'd implement the handling for this shutdown signal in the most reasonable way possible. (edit: I now realise that with this handler it might not be failing appropriately anyway in case of an exit propagation ? edit on edit: no it would still fail as an unexpected crash would not have the `:shutdown` exit signal)

I don't love that test behaviour has led to implementation changes, but ultimately it seems these processes are trapping exits anyway and figured this is a reasonable way to handle this particular exit signal. Happy to change it to use `start_supervised!` instead and ensure the tests fail appropriately though.

alco: Thanks for explaining that.

It's fine to have this code, I guess. I don't fully understand what happens during stack shutdown, to be honest. I have tried to debug it not too long ago with the help of logger calls but I couldn't observe either `Connection.Manager` or `ReplicationClient` receiving the `:shutdown` EXIT signal, even though both processes do shut down when I explicitly stop the connection supervisor. I've posted a few examples in https://github.com/electric-sql/electric/pull/2844.

PR #2845, opened 2025-06-19
feat(sync-service) Replace GenStage with a simple bespoke message publishing system

### LINKED ISSUE #2670: GenStage alternative POC
We're observing accumulating WAL when the number of shapes starts to increase in some usage scenarios, is hard to explain why from traces. All points to something about how GenStage works. We're not making the best use of GenStage as we use only a tiny fraction of its capabilities, so we want to check how Electric performs with a custom thing for dispatching messages to shape processes.

robacourt:
Fixes #2670 by replacing GenStage with a simple message publishing system. The approach used offers some significant simplifications over GenStage:
- Very easy to understand. The sending process is [20 lines of code](https://github.com/electric-sql/electric/blob/5f3a10fe0eafdbe986d61151417b03d2a6c142ad/packages/sync-service/lib/electric/publisher.ex#L11) rather than a 10k line library
- Sending scales to 0 on no matching shapes
- Much easier to profile. Sending is just `Publisher.publish(matched_subscribers, message)` which you can put a span around, rather than returning a list of what to send and the sending is lost in the library
- Subscriptions are decoupled from process initialisation, so for example we could initialise the shape consumers in parallel but subscribe in series (which is beneficial since subscribing in on a single process)
- Producer and consumer processes can both be simple `GenServer`s rather than having to be `GenStage`s. This means any process can be a producer or consumer. So for example currently the `ReplicationClient` and `ShapeLogCollector` are separate processes since `ReplicationClient` is a `Postgrex.ReplicationConnection` and `ShapeLogCollector` is a `GenStage`. All transactions have to be copied between these two processes with some performance cost. After this PR we could make them one process eliminating this copy.


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -56,14 +59,20 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+    GenServer.call(server, {:new_txn, txn, trace_context, call_time}, :infinity)
   end
 
   def handle_relation_msg(%Changes.Relation{} = rel, server) do
     trace_context = OpenTelemetry.get_current_context()
     GenServer.call(server, {:relation_msg, rel, trace_context}, :infinity)
   end
 
+  def subscribe(server, shape) do
+    # TODO: will all subscribers be subscribing async on init? if so this may need have a larger timeout
```
msfstef: is there a chance that the consumer init itself might also fail because of timeouts even if we increase this timeout? Or is the suggestion here to make the subscribe call a cast, and rely on the `set_last_processed_lsn` timeout as the time required until all subscribe messages are processed?

I prefer if the subscribe call is sync and the `set_last_processed_lsn` does not require a large timeout, I think it's much clearer behaviour, but I do wonder about the timeouts involved.

robacourt: Oops, I meant to do that TODO, not leave it in the PR 😄 I'll look at that now

robacourt:
benchmark this

packages/sync-service/lib/electric/replication/shape_log_collector.ex
robacourt: I've removed the TODO. For the record the consumers are started synchronously, each will finish subscribing before the next is started, so no timeout increase is needed.

msfstef: > each will finish subscribing before the next is started

IIRC on restore we start them all asynchronously, which means that their init calls will be called concurrently and will all get queued at the `ShapeLogCollector`. My concern is that the `start_link` call will start timing out for some of them if e.g. we're restoring thousands of shape consumers on startup, because it'll take too long to finish the `init` call since it's waiting on all the others.

I can't imagine that the subscribe call requires long to handle but just checking this is not an issue

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -34,12 +37,10 @@ defmodule Electric.Replication.ShapeLogCollector do
   end
 
   def set_last_processed_lsn(server, last_processed_lsn) do
-    # Allow 60s for this call as it may need to wait for thousands of restored shapes
-    # to subscribe before it returns.
-    GenStage.call(server, {:set_last_processed_lsn, last_processed_lsn}, 60_000)
+    GenServer.call(server, {:set_last_processed_lsn, last_processed_lsn})
```
robacourt, line 40: This long timeout is no longer necessary as consumers subscribe synchronously as part of start-up and will have completed before set_last_processed_lsn/2 is called. Unlike GenStage which would still be subscribing after all the consumers had started.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
robacourt: As discussed they are currently synchronous, a Task is used but yields before moving onto the next. If at some point we want to load each shape from disk asynchronously we could still have synchronous subscriptions since the subscription mechanism is now decoupled from consumer initialisation.

alco (approved):
Brilliant!

PR #2847, opened 2025-06-24
fix: `OR` operator in where clauses should not be strict

msfstef:
Was trying to get a where clause in a shape to work, where either one field would match a value OR another field would, and in practice I would set one field to null and move the value over to the other (meaning that it should trigger an update) and I was seeing a delete coming through.

Digging through the code I found the issue to be that we were treating `AND`, `OR`, and `NOT` all as `strict?` operators, meaning that if any of their arguments are `NULL` we collapse the result to `NULL` also. However this is not the correct behaviour on postgres for the `OR` operator, as shown below:

```sql
select null and true; # -> null
select not null; # -> null
select null or true; # -> t
select null or false; # -> null
```

I've changed the `OR` operator to not be strict along with a simple test for it. I wonder if we should populate our where clause execution tests a bit more, we can probably generate a good suite of tests for it.

edit: I've added a custom `or` operator after @icehaunter 's note about the odd behaviour, and also had to fix up some "bool chains" we were parsing.

The affected operators are:
- `OR`  - before `true OR null` would be `null` rather than `true`
- `IN` - before `1 IN (1, null)` would be `null` rather than `true`, since `IN` is just a chain of `OR`s
- `BETWEEN SYMMETRIC` - no behaviour change as the underlying operations are comparisons that will always either both yield `null` or a boolean, but it now uses the correct `or` operator either way

### DIFF
--- a/packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
+++ b/packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
@@ -125,4 +125,32 @@ defmodule Electric.Replication.PostgresInterop.Casting do
   end
 
   def ilike?(text, pattern), do: like?(text, pattern, true)
+
+  @doc """
+  The Postgres OR operator, which has some specific behaviour when
+  comparing NULLs with booleans.
+
+  ## Examples
+
+      iex> pg_or(true, false)
+      true
+
+      iex> pg_or(false, false)
+      false
+
+      iex> pg_or(nil, true)
+      true
+
+      iex> pg_or(nil, false)
+      nil
+
+      iex> pg_or(nil, nil)
+      nil
+  """
+  @spec pg_or(boolean() | nil, boolean() | nil) :: boolean() | nil
+  def pg_or(a, b)
+  def pg_or(nil, true), do: true
+  def pg_or(nil, _b), do: nil
+  def pg_or(a, nil), do: pg_or(nil, a)
+  def pg_or(a, b), do: Kernel.or(a, b)
 end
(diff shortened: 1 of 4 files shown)


icehaunter:
Not sure if worth correcting further, but `SELECT null or false` yields `null`, not `false` in postgres /shrug

msfstef:
> Not sure if worth correcting further, but SELECT null or false yields null, not false in postgres /shrug

Ugh I did not think of checking that case - that's disgusting. It's probably worth fixing as I'm worried it might lead to other unexpected issues like the one I had down the line. I guess I can special case the execution of the OR function, not great but this "semi-strict" behaviour is probably not worth encoding otherwise?

edit: @icehaunter added a test, thankfully `Kernel.or/2` has exactly this behaviour:
```elixir
iex> Kernel.or(false,nil)
nil
iex> Kernel.or(true,nil)
true
```

icehaunter:
@msfstef your test work with this argument order but not the other way around I believe - in Elixir `nil or false` will coalesce to last false arg - `false`

msfstef:
Fixed it up with a custom `pg_or` function, added more tests and included reversed arguments for safety.

Turns out the `IN` operator was also wrong, since it chains `OR`s and thus `1 IN (1, NULL)` was returning `NULL` rather then true.

packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
```diff
@@ -125,4 +125,32 @@ defmodule Electric.Replication.PostgresInterop.Casting do
...
+      nil
+
+      iex> pg_or(nil, nil)
+      nil
+  """
+  @spec pg_or(boolean() | nil, boolean() | nil) :: boolean() | nil
+  def pg_or(a, b)
+  def pg_or(nil, true), do: true
+  def pg_or(nil, _b), do: nil
+  def pg_or(a, nil), do: pg_or(nil, a)
```
icehaunter: Let's write it out, it's literally one more case, but one less delegation
```suggestion
  def pg_or(true, nil), do: true
  def pg_or(_, nil), do: nil
```

msfstef: can't help it, I just want to say it's commutative and let a smart compiler deal with it haha

PR #2848, opened 2025-06-24
fix: Account for race between loading shape and listening for changes

msfstef:
This PR started as an attempt to fix a flaky TS test, that basically just tested that it received updates to a shape.

After some investigation, I found that the reason it was flaking was that sometimes the updates arrived at the shape consumer the moment between the request:
1. Loaded the shape (its handle and latest offset)
2. Started listening for changes to the shape (which we only do if the request's offset is larger than or equal to the read latest offset)

In this case, the `:new changes` message would not have been received by the request process as the changes were processed before it started listening, and the latest offset is loaded is now out of date.

This leads to the request hanging for 20 seconds waiting for new updates (despite the update already having been processed). The next request would pick it up, so this is not a correctness issue, but it does lead to cases where live updates might take an upper bound of the long poll timeout to arrive to clients, which is not ideal.

Possible solutions:
1. Start listening for changes before we load the shape's latest offset
 1a. Not sure how expensive registering basically _all_ requests would be for changes, and potentially deregistering them early if it is not necessary
 2b. We would have to separately load the handle and the offset, as the changes are keyed on the handle, which would mean two separate ETS reads anyway up front, on top of "flooding" the registry
2. Do an additional check for the latest offset after we start listening for changes
 2a. This ensures that we capture any missed changes, and can occur at any point after we have registered the request for new changes, and the cost is just an additional ETS read.
 2b. We can do the manual check early or as late as when we block on the receive for new changes
 2c. We can even delay the second check such that it happens at e.g. 200ms of waiting, basically setting an upper bound on the amount of time a request can be delayed because of a missed change, just for the sake of reducing the number of additional ETS reads, but I would avoid overoptimising to avoid the read.
3. Somehow always read up to the latest part of the log when we don't find a chunk boundary, rather than up to the latest offset that was loaded at the start of the request.
 3a. I don't think this approach is actually viable because we'd have to introspect the log, which we generally stream out after already having sent the headers

I opted for option 2, as I think it is the most explicit and requires the least amount of changes, and it comes at the cost of a simple additional ETS read.

## Implementation
1. Check for any messages in the mailbox (receive do with 0 timeout)
2. If new change messages arrived, terminate there
3. If no messages arrived, do a lookup to check manually if we are up to date.
4. If we are not up to date (shape updated or rotated), send ourselves a message
5. Wait for changes again with a long poll timeout

This currently just checks the mailbox immediately, but we could do this additional read with an "upper bound" of time we're willing to wait for changes, in order to reduce the number of additional lookups we do. I doubt it's worth it though.

Note, I've added 2 failing tests, one for missing changes and one for missing a shape rotation, although I'm not sure if we could ever miss the shape rotation as we do interact with the shape process after listening for changes (and if it had rotated between loading and listening for changes, the subsequent interaction would crash) - perhaps the rotation edge case is causing a different kind of crash elsewhere that might be worth investigating. The rotation race is an argument for doing the ETS lookup _immediately_ after registering for changes, to appropriately handle the request.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -566,11 +566,62 @@ defmodule Electric.Shapes.Api do
   defp hold_until_change(%Request{} = request) do
     %{
       new_changes_ref: ref,
+      last_offset: last_offset,
       handle: shape_handle,
-      api: %{long_poll_timeout: long_poll_timeout}
+      params: %{shape_definition: shape_def},
+      api: %{long_poll_timeout: long_poll_timeout} = api
     } = request
 
     Logger.debug("Client #{inspect(self())} is waiting for changes to #{shape_handle}")
+
+    # Between loading the shape info and registering as a listener for new changes,
+    # there is a short time period where information might be lost, so we check our
+    # mailbox and then do an explicit check if nothing is present
+    hold_until_change_with_timeout(
+      request,
+      0,
+      fn ->
+        case Shapes.get_shape(api, shape_def) do
+          {^shape_handle, ^last_offset} ->
+            # no-op, shape is still present and unchanged
+            nil
+
+          {^shape_handle, latest_log_offset}
+          when is_log_offset_lt(last_offset, latest_log_offset) ->
+            send(self(), {ref, :new_changes, latest_log_offset})
+
+          {other_shape_handle, _} when other_shape_handle != shape_handle ->
+            send(self(), {ref, :shape_rotation, other_shape_handle})
+
+          nil ->
+            send(self(), {ref, :shape_rotation})
+        end
+
+        hold_until_change_with_timeout(
+          request,
+          # If we timeout, return an up-to-date message
+          long_poll_timeout,
+          fn ->
+            # Ensure stack is ready after a long poll timeout, as it might
+            # have failed during this period
+            case hold_until_stack_ready(request.api) do
+              :ok ->
+                request
+                |> update_attrs(%{ot_is_long_poll_timeout: true})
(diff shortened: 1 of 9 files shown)


packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -566,12 +566,63 @@ defmodule Electric.Shapes.Api do
...
+        case Shapes.get_shape(api, shape_def) do
+          {^shape_handle, ^last_offset} ->
+            # no-op, shape is still present and unchanged
+            nil
+
+          {^shape_handle, latest_log_offset}
+          when is_log_offset_lt(last_offset, latest_log_offset) ->
+            send(self(), {ref, :new_changes, latest_log_offset})
+
+          {other_shape_handle, _} when other_shape_handle != shape_handle ->
```
magnetised:
```suggestion
          {other_shape_handle, _} ->
```

no need for the extra assertion and the var name is nice and explicit about what's going on

magnetised (approved):
Good catch. Race conditions... yuck.

Don't love the end result but I have no suggestions for how to improve it...

packages/sync-service/lib/electric/shapes/api.ex
msfstef: I have a `when is_log_offset_lt(last_offset, latest_log_offset)` above, so technically we _could_ end up with the same shape handle and a lower log offset (?) and it would fall into this pattern

That's why I added the explicit comparison, and if the above ever happens it should crash

msfstef:
Simplified the code as per @robacourt and @icehaunter suggestion to skip the mailbox check, which makes the change smaller and simpler. Thanks for the reviews!

PR #2850, opened 2025-06-25
elixir-client: Support specifying column subsets via Ecto.Query.select/3

### LINKED ISSUE #2843: Feature request: Support `select` when inferring shape from Ecto query
I would like to be able to sync only a subset of columns with a `select` like shown below.
```elixir
def sync(conn, %{"table" => "some_table"} = params) do
  user_id = conn.assigns.user.id

  sync_render(
    conn,
    params,
    from(x in MyApp.SomeSchema,
      where: x.user_id == ^user_id,
      select: [:id, :user_id, :some_column]
    )
  )
end
```

In my local copy of `electric_client`, I was able to achieve it like below. I'm sure these could be more robust but it has worked fine for me so far.
```elixir
# works when selection is like `[:id, :user_id, :some_column]`
defp query_columns(%{select: %Ecto.Query.SelectExpr{take: %{0 => {:any, columns}}}}) do
  Enum.map(columns, &to_string(&1))
end

# works when selection is like `[x.id, x.user_id, x.some_column]`
defp query_columns(%{
        select: %Ecto.Query.SelectExpr{
          expr: [{{:., [], [{:&, [], [0]}, _column_id]}, [], []} | _] = selections
        }
      }) do
  for {{:., [], [{:&, [], [0]}, column_id]}, [], []} <- selections,
      do: to_string(column_id)
end

# existing implementation selecting all columns
defp query_columns(%{from: %{source: {_table_name, struct}}}) do
  Enum.map(
    struct.__schema__(:fields),
    &to_string(struct.__schema__(:field_source, &1))
  )
end
```

magnetised:
Fixes #2843

### DIFF
--- a/.changeset/violet-wasps-agree.md
+++ b/.changeset/violet-wasps-agree.md
@@ -1,0 +1,5 @@
+---
+"@core/elixir-client": patch
+---
+
+Support specifying column subsets via Ecto.Query.select/3
--- a/packages/elixir-client/lib/electric/client/ecto_adapter.ex
+++ b/packages/elixir-client/lib/electric/client/ecto_adapter.ex
@@ -110,11 +110,28 @@ if Code.ensure_loaded?(Ecto) do
       {table_name, query_prefix || source_prefix, struct}
     end
 
+    defp query_columns(%{select: %Ecto.Query.SelectExpr{take: %{0 => {action, columns}}}})
+         when action in [:any, :map, :struct] do
+      Enum.map(columns, &to_string/1)
+    end
+
+    defp query_columns(%{select: %Ecto.Query.SelectExpr{expr: [{{:., _, _}, [], []} | _] = expr}}) do
+      map_expr_columns(expr)
+    end
+
+    defp query_columns(%{select: %Ecto.Query.SelectExpr{expr: {:{}, [], [_ | _] = expr}}}) do
+      map_expr_columns(expr)
+    end
+
     defp query_columns(%{from: %{source: {_table_name, struct}}}) do
       Enum.map(
         struct.__schema__(:fields),
         &to_string(struct.__schema__(:field_source, &1))
       )
+    end
+
+    defp map_expr_columns(expr) do
+      for {{:., [], [{:&, [], [0]}, name]}, [], []} <- expr, do: to_string(name)
     end
 
     @doc false
--- a/packages/elixir-client/lib/electric/client.ex
+++ b/packages/elixir-client/lib/electric/client.ex
@@ -387,13 +387,40 @@ defmodule Electric.Client do
         iex> Elixir.Client.shape!(&MyApp.Todo.changeset/1)
         %Electric.Client.ShapeDefinition{table: "todos", columns: ["title", "completed"]}
 
-    Passing a changeset will filter the table columns according to the applied
+    Values from the Electric change stream will be mapped to instances of the
+    passed `Ecto.Schema` module.
+
+    ## Column subsets
+
+    Specifying a subset of columns to stream can be done in two ways:
+
+    1. Passing a changeset will filter the table columns according to the applied
     validations so if you want a shape to include a column, you must ensure
     that it has some kind of validation, using e.g.
     `Ecto.Changeset.validate_required/2`
 
-    Values from the Electric change stream will be mapped to instances of the
-    passed `Ecto.Schema` module.
+    2. Using `Ecto.Query.select/3` to select a subset of columns within the query:
+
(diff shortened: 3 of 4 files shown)


.changeset/violet-wasps-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/elixir-client": patch
```
msfstef, line 2: as @icehaunter pointed out we might be able to afford being more generous with minor updates now when new features are added

packages/elixir-client/lib/electric/client/ecto_adapter.ex
```diff
@@ -110,13 +110,30 @@ if Code.ensure_loaded?(Ecto) do
...
+
     defp query_columns(%{from: %{source: {_table_name, struct}}}) do
       Enum.map(
         struct.__schema__(:fields),
         &to_string(struct.__schema__(:field_source, &1))
       )
     end
 
+    defp map_expr_columns(expr) do
+      for {{:., [], [{:&, [], [0]}, name]}, [], []} <- expr, do: to_string(name)
```
msfstef, line 134: I'm not familiar with Ecto but this looks more like [brainfuck](https://esolangs.org/wiki/Brainfuck) code than Elixir lol

alco: This is [Elixir AST](https://hexdocs.pm/elixir/quote-and-unquote.html), though in this case it's highly specific to Ecto's internal representation of its query API. I can't say I've seen such destructuring of Ecto's query expressions before, so it does look a bit dense without accompanying comments.

alco: > I'm not familiar with Ecto but this looks more like [brainfuck](https://esolangs.org/wiki/Brainfuck) code than Elixir lol

Fun comparison, it had never occurred to me. FWIW, Elixir's AST is actually a blessing to work with thanks to its regular structure. In [Erlang AST](https://www.erlang.org/doc/apps/erts/absform.html#expressions), on the other hand, every expression has its own bespoke representation. In any case, it's quite a mindbending experience working directly with AST. One of my early encounters with Elixir was reading its Erlang code responsible for [translating](https://github.com/elixir-lang/elixir/blob/main/lib/elixir/src/elixir_erl_pass.erl) Elixir AST into Erlang AST; I couldn't make out what any of that was about at all.

alco (approved):
This is great!

I think there needs to be a disclaimer, at least as code comment, highlighting which features of Ecto's extensive `select` API are supported and which aren't. For example, on the [doc page](https://hexdocs.pm/ecto/Ecto.Query.html#select/3) we see expressions such as 
```
from(c in City, select: c) # returns the schema as a struct
from(c in City, select: %{n: c.name, answer: 42})
from(c in City, select: %{c | alternative_name: c.name})
from(c in City, select: %Data{name: c.name})
```
which aren't covered by tests added in this PR.

Then there are [dynamic parts](https://hexdocs.pm/ecto/Ecto.Query.html#select/3-dynamic-parts) and [select_merge](https://hexdocs.pm/ecto/Ecto.Query.html#select_merge/3) which would probably raise if used with elixir-client (?), so it'd be nice to know that their omission is intentional and whether they can be supported at all at some point.

packages/elixir-client/lib/electric/client/ecto_adapter.ex
msfstef: Ah I see, so Ecto is moving the AST representations around and "executing/unquoting" them when necessary - that makes sense, thanks for all the references, good to read up on it!

As I mentioned in the top level comment it does worry me if we are matching on the internal representation though, as I would imagine how exactly Ecto uses the internal structure (and thus AST representation) is not part of a stable public API ?

alco: Yeah, that's a valid concern. There have been some adapters for Ecto targeting other DBs, such as [Mongo.Ecto](https://hexdocs.pm/mongodb_ecto/Mongo.Ecto.html). I'm not sure if they also pattern-match on Ecto.Query.*Expr structs or use other clever tricks. One thing we can rely on is that Ecto's query API is unlikely to change, but as I've pointed out in my comment, there's more to `select` and related functionality, such that supporting all of it won't be feasible with this brute-force pattern-matching-on-ast approach.

balegas:
@magnetised would be nice to merge this

magnetised:
@alco I added some docs around supported `select/3` styles. could you have a look pls.

PR #2851, opened 2025-06-25
chore(sync-service): Fix flakey Publisher tests

robacourt:
These tests can occasionally fail with:

```
 Assertion failed, no matching message after 5000ms
     The process mailbox is empty.
     code: assert_receive :publish_finished
```

The race condition occurs if the TestSubscriber receives the `:finish_processing_message` message before it receives the publisher call, in which case it swallows the message since `handle_info` is not implemented on TestSubscriber. A GenServer would log that a message had been received and not handled, but as `use GenServer` was omitted from the TestSubscriber the message was swallowed silently.

This PR:
- Adds `use GenServer` to the TestSubscriber as this is best practice (the race condition would have been obvious had it been included in the first place)
- Waits for `:message_received` from each of the subscribers before sending `:finish_processing_message`. This eliminates the race condition.
- Renames `event` to `message` for consistency  

Credit to ChatGPT which worked out what the race condition was. It's solution was terrible, but the hard part was knowing what the race condition was, so that was super helpful.

### DIFF
--- a/packages/sync-service/test/electric/publisher_test.exs
+++ b/packages/sync-service/test/electric/publisher_test.exs
@@ -5,6 +5,8 @@ defmodule Electric.PublisherTest do
   alias Electric.PublisherTest.TestSubscriber
 
   defmodule TestSubscriber do
+    use GenServer
+
     def start_link(on_message) do
       GenServer.start_link(__MODULE__, on_message)
     end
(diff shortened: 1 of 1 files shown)


robacourt:
> Very nice, that's what I'm talking about! I wonder if we have any other such "mock" processes that could benefit from implementing GenServer

Good shout! I just checked and we don't currently. It would be nice if a static analysis tool like credo could ensure things like this. I might look into that at some point.

PR #2854, opened 2025-06-26
feat(sync-service): Add interval timer for more precise replication processing telemetry

robacourt:
When profiling a process to see where time is being spent, using spans or `timed_fun` is problematic:
- Gaps between the timed funs are not included leading to the parts not adding up to the whole. 
- `timed_fun` requires you to wrap a section of code in the fun which is not always possible (e.g. how long does it take for a message to get to a process).
- Code needs to be wrapped in funs and any variables needed outside of the timed section need to be returned which can lead to complicated looking code.

This PR introduced the interval timer, which can be used via the `OpenTelemetry` module.

e.g.

```elixir
OpenTelemetry.start_interval("quick_sleep")
Process.sleep(1)
OpenTelemetry.start_interval("longer_sleep")
Process.sleep(2)
OpenTelemetry.stop_and_save_intervals(total_attribute: "total_sleep_µs")
```

will add the following attributes to the current span:
- quick_sleep.duration_µs: 1000
- longer_sleep.duration_µs: 2000
- total_sleep_µs: 3000

The interval timer guarantees that the parts add up to the total and there are no gaps. When presented with confusing results having these strong guarantees reduces the explanations that you need to consider which can make profiling significantly simpler. 

This PR also uses the interval timer to profile replication stream processing.


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -90,32 +99,37 @@ defmodule Electric.Replication.ShapeLogCollector do
         persistent_replication_data_opts: persistent_replication_data_opts,
         tracked_relations: tracker_state,
         partitions: Partitions.new(Keyword.new(opts)),
-        filter: Filter.new(Keyword.new(opts))
+        filter: Filter.new(Keyword.new(opts)),
+        timer: nil
```
msfstef: nit: perhaps we could name this in a way that it is more explicit that this timer is for telemetry (and not for any functional puprose) - like `telemetry_timer` or something

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -146,39 +159,45 @@ defmodule Electric.Replication.ShapeLogCollector do
...
     Logger.debug(fn -> "Txn received in ShapeLogCollector: #{inspect(txn)}" end)
 
-    OpenTelemetry.timed_fun("shape_log_collector.handle_transaction.duration_µs", fn ->
-      handle_transaction(txn, from, state)
-    end)
+    state = handle_transaction(state, txn)
+
+    OpenTelemetry.start_interval("shape_log_collector.transaction_message_response")
+
+    {:reply, OpenTelemetry.extract_interval_timer(), %{state | timer: nil}}
```
msfstef: I see we set `timer: nil` here but I cannot see where we set it to anything else, or where we use it?

robacourt: Oops, I meant to remove that. It's gone now. Good spot, thanks!

packages/sync-service/lib/electric/replication/shape_log_collector.ex
robacourt: It's not needed, I've removed it.

PR #2856, opened 2025-06-27
feat: experimental support for streaming SSE in live mode

kevin-dp:
Rebases https://github.com/electric-sql/electric/pull/2544 on latest main

### DIFF
--- a/.changeset/thick-bottles-act.md
+++ b/.changeset/thick-bottles-act.md
@@ -1,0 +1,6 @@
+---
+"@electric-sql/client": patch
+"@core/sync-service": patch
+---
+
+Add experimental SSE support.


alco:
This test keeps failing no matter how many times I rerun it. Doesn't look like flake.

```
HTTP Proxy Cache > should get fresh response when age is passed the stale age 5015ms
     → Test timed out in 5000ms.
```

kevin-dp:
> This test keeps failing no matter how many times I rerun it. Doesn't look like flake.
> 
> ```
> HTTP Proxy Cache > should get fresh response when age is passed the stale age 5015ms
>      → Test timed out in 5000ms.
> ```

Yes, this test started failing when i upgraded Vitest to v3. If i downgrade it back to v2 it passes again..

kevin-dp:
@alco i fixed the failing unit test. It seems that Vitest v3 ignores the "timeout" configuration on the `describe` block. So i removed it and added a custom timeout for the caching test. It's green now.

PR #2859, opened 2025-06-30
Fix flaky ShapeCache test

alco:
Before this change, it was easy to see the test fail after just a handful of runs. The reason was that the simulated snapshotter process would be resumed before the test process had started waiting on the snapshot.

This change makes sure there's a process in the consumer's waiting list before resuming the simulated snapshotter, though it does make the test aware of implementation details.

This test has failed in two CI jobs spawned by the [commit in `main`](https://github.com/electric-sql/electric/commit/fbbcad415db1a719e2acef50e337641086d30609) that was supposed to publish a new version of Electric and other packages.


robacourt (approved):
Nice one!

PR #2862, opened 2025-06-30
fix: Ensure PG expression execution in Electric matches PG

msfstef:
Introduces some `StreamData` property tests for our PG expression parser and executor.

Not the prettiest but I did already find several issues, some of which I would need confirmation that they were not intentional:

1. `AND` should handle `false and null` case as `false`
2. Array concatenation with `null` should return the same array
3. `array_prepend` and `array_append` work with `null`s, and they actually add the null to the array
4. Array prepends and appends using `||` work with `null` but they keep the array the same, and they do not work with strings
5. Array contains and is contained have specific behaviours with `null` - if the array that should be contained has a `null` then it will never pass the check - whether the array containing the other has a `null` or not.
6. The `^` operator should handle `0^0`, which `Float.pow` did not - not sure if this affects performance significantly @icehaunter 
7. Added bitwise not operator - not necessary, can remove, was just checking most operators

I've also found that various comparison operators don't work on booleans and arrays, but decided to leave that as separate work and instead try to get this merged as it is an improvement over what we had before.

The tests generate arbitrary expressions and then use postgres as the oracle to ensure that the values we produce are the same.

If both our runner and postgres error, we ignore the result.

If our runner resolves to nil but postgres errors I also ignore that result, as we have this concept of `strict?` execution that would need examination in lots of parts to ensure we have the same behaviour as postgres exactly. Can be done iteratively.

### DIFF
--- a/.changeset/few-eggs-relax.md
+++ b/.changeset/few-eggs-relax.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix parsing and execution for some edge casses of where clause operators.
--- a/.changeset/few-eggs-relax.md
+++ b/.changeset/few-eggs-relax.md
@@ -2,4 +2,4 @@
 "@core/sync-service": patch
 ---
 
-Fix parsing and execution for some edge casses of where clause operators.
+Fix parsing and execution for some edge cases of where clause operators.
--- a/packages/sync-service/test/support/pg_expression_generator.ex
+++ b/packages/sync-service/test/support/pg_expression_generator.ex
@@ -1,0 +1,253 @@
+defmodule Support.PgExpressionGenerator do
+  @moduledoc """
+  Generates Postgres expressions for testing purposes.
+  """
+
+  import StreamData
+
+  ## TYPE GENERATORS
+
+  defp null_gen, do: constant("NULL")
+  defp bool_gen, do: member_of(["TRUE", "FALSE"])
+
+  defp int_gen, do: integer() |> map(&Integer.to_string/1)
+  defp double_gen, do: float(min: -1.0e6, max: 1.0e6) |> map(&Float.to_string/1)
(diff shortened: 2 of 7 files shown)


.changeset/few-eggs-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix parsing and execution for some edge casses of where clause operators.
```
robacourt:
```suggestion
Fix parsing and execution for some edge cases of where clause operators.
```

packages/sync-service/test/support/pg_expression_generator.ex
```diff
@@ -0,0 +1,241 @@
...
+
+  defp expression_gen(type_gen, op_generators) do
+    type_gen = nullable_type_gen(type_gen)
+
+    op_generators
+    |> Enum.concat([
+      {:unary_op, is_null_op_gen()}
+      # {:comparison_op, comparison_op_gen()},
+      # {:range_op, range_comparison_op_gen()},
+      # {:membership_op, membership_op_gen()}
```
robacourt: Are these meant to be commented out?

msfstef: Whoops - yeah these are supposed to work with all datatypes but they don't so I've added them to the ones that we do support instead (and forgot to comment them out).

We need to make them work for booleans basically.

msfstef:
> It seems like we can extend this approach some more? My Claude discussion — https://claude.ai/share/a5325fda-8dc3-4d62-af86-fdd9d759d25e

@KyleAMathews I had to limit things to operations and types we do support - I've added nested bool expressions for good measure in this PR but I think the rest I'd need to pair with @icehaunter to avoid making too many implementation changes while testing things.

One thing I've omitted from the tests that we do support is timestamps, but I think we can add it as a subsequent PR.

PR #2863, opened 2025-06-30
sync-service: Add missing default for http_api_num_acceptors

magnetised:
`http_api_num_acceptors` was added as a configuration option with no default value which meant that it *had* to be configured or the `api_server/1` call would raise. This PR adds the missing default (`nil` so that we don't set a custom value by default) and includes a smoke test of the default configuration to avoid this happening again.

Also enables the external configuration of the `inspector` and `registry` components which is needed for current work on `Phoenix.Sync`


robacourt (approved):

alco (approved):
Beat me to it. Thanks!

packages/sync-service/test/electric/config_test.exs
```diff
@@ -28,4 +28,39 @@ defmodule Electric.ConfigTest do
       assert :ok = Electric.Config.validate_security_config!(nil, true)
     end
   end
+
+  describe "defaults" do
+    # want to know what happens without our runtime environment
+    setup do
```
alco, line 36: I like the idea but tests without assertions throw me off. Perhaps a short comment explaining the purposes of writing these tests in this manner is in order.

PR #2864, opened 2025-06-30
fix: Use separate stack event for pool connection errors

msfstef:
Fixes https://github.com/electric-sql/stratovolt/issues/697

We need to classify these pool connection errors differently than the errors that cause either a reconnection or a failure to electric.

We can do that either with a separate stack event as suggested by @alco and done here with `:pool_connection_error`, or adding a property to indicate that we can ignore this. I am of the opinion that this should be a clearly different event since it is not fatal or causes a full reconnection, but am happy to hear other thoughts - my understanding is the pool connections will be reinstated?

### DIFF
--- a/.changeset/brown-bugs-shout.md
+++ b/.changeset/brown-bugs-shout.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Use `:pool_connection_error` stack event for pool errors that we do not retry anything on.
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -725,7 +725,7 @@ defmodule Electric.Connection.Manager do
     error = DbConnectionError.from_error(exit_reason)
 
     dispatch_stack_event(
-      {:connection_error,
+      {:pool_connection_error,
        %{
          error: DbConnectionError.format_original_error(error),
          type: error.type,


robacourt (approved):
👍

alco:
> my understanding is the pool connections will be reinstated?

Yes. To put it simply, `Postgrex.start_link()` which is called inside Connection.Manager is a wrapper around `DBConnection.start_link()`.  The latter starts a supervisor process implemented in `DBConnection.ConnectionPool.Pool` which is a regular one_for_one supervisor: so any connection process gets restarted when it stops.

If pooled connections keep crashing right after getting restarted, the max_restarts limit within max_seconds will be reached and cause the supervisor itself to crash, then the Connection.Manager process will get EXIT message about it and will shut down itself.

So classifying pooled conection errors as something that doesn't cause any state transitions in the cloud is the right call here. :+1:

alco (approved):
:heart:

balegas:
We shouldn't log them at all to the user dashboard, since they might create more confusion since they are non-actionable.

msfstef:
> We shouldn't log them at all to the user dashboard, since they might create more confusion since they are non-actionable.

In that case we have to either filter them at the ingestion level or at the emit level

balegas:
As discussed many times, `pool_connection_error` are never actionable or handled by Electric, I think we could not emit events for the entire category of errors. Do we see any downsides with this?

msfstef:
So are you suggesting we don't emit any stack events at all for this? I think I agree, I'm just not sure why we were emitting one in the first place

balegas:
I think for completeness/on the expectation that an upper-layer would prevent them from appearing in the dashboard.
The only situation I'm not 100% sure we need the `pool_connection_errors` is during connection, but my understanding it that this is already covered by electric who needs to handle the errors explicity.

If we're all agreeing. I'd ask not to emit these errors. They are really confusing and noisy for the user

msfstef:
I still don't understand if you think we should not be emitting these at the Electric level or if we should ignore them at the cloud level.

I've opened a PR that removes the stack event altogether https://github.com/electric-sql/electric/pull/2865

PR #2866, opened 2025-07-01
fix: Ensure `LockConnection` can get properly restart on `ConnMan` restarts

msfstef:
While working on https://github.com/electric-sql/stratovolt/issues/699 I noticed that when the connection manager gets restarted by its supervisor, when it attempts to start the `LockConnection`, it fails with `{:already_started, ...}`.

The lock connection process is linked to ConnMan, but it seems like there's a delay between ConnMan dying and the lock connection (which under the hood is a `:gen_statem`) is killed, so when ConnMan tries to start a new lock connection with the same name it runs into this issue, which causes ConnMan to die and restart and run into the same issue until the supervisor kills it altogether.

Three suggested solutions:
1. Don't register the lock connection process with a global `name`
 1a. We don't use it anywhere, so restarts always manage to start a lock and the timing issue is sidestepped
 1b. The negative is that we now have lock processes not globally addressed and we might end up trying to hold the lock twice somewhere (?)
2. Add a retry mechanism, either internally to `LockConnection` or in `ConnMan`, so that if we get `:already_started` we wait a few milliseconds and try a few times
 2a. I tried this and this works fine, but I don't like the arbitrary retries and timing attempts
3. Ensure lock is killed before trying to start it
 3a. Strongly dislike this approach - feels forced

I've opted for 1. in this PR, after having tried 2. which worked but I don't love retries. Happy to hear opinions.  

## UPDATE
After discussions with @robacourt  (thank you for pointing out that `terminate` still runs when exiting the same process!), I've made conn man behave slightly more like a supervisor, following loosely what the OTP supervisor does but with fewer requirements.

I've added the relevant pids of "children" processes that need to be shut down before conn man exits - this solves the issue. Happy to remove the extra processes and only keep the lock connection in there to solve the immediate bug, but this seems to do the trick.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -107,8 +107,6 @@ defmodule Electric.Connection.Manager do
       :lock_connection_timer,
       # PID of the database connection pool
       :pool_pid,
-      # PID of the shape supervisor
-      :shapes_supervisor_pid,
       # PID of the shape log collector
       :shape_log_collector_pid,
       # Backoff term used for reconnection with exponential back-off
@@ -482,8 +480,7 @@ defmodule Electric.Connection.Manager do
 
     state = %{
       state
-      | shapes_supervisor_pid: shapes_sup_pid,
-        shape_log_collector_pid: log_collector_pid,
+      | shape_log_collector_pid: log_collector_pid,
         current_step: :waiting_for_consumers,
         purge_all_shapes?: false
     }
@@ -948,15 +945,15 @@ defmodule Electric.Connection.Manager do
     # connection manager is able to start the processes in a clean state.
     Logger.debug("Terminating connection manager with reason #{inspect(reason)}.")
 
+    Electric.Connection.Supervisor.stop_shapes_supervisor(state.stack_id)
+
     %{
-      shapes_supervisor_pid: shapes_supervisor_pid,
       pool_pid: pool_pid,
       replication_client_pid: replication_client_pid,
       lock_connection_pid: lock_connection_pid
     } = state
 
     [
-      shapes_supervisor_pid,
       pool_pid,
       replication_client_pid,
       lock_connection_pid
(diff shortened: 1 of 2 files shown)


robacourt (changes requested):
As discussed, ConnMan is a bespoke supervisor and should behave like a normal supervisor as much as possible. Normal supervisors trap exits and shutdown children on terminate, so I suggest ConnMan does the same. Also ConnMan should `use GenServer,  shutdown: :infinity` like a normal supervisor to ensure terminate is given the time it needs to close it's children.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -937,6 +940,37 @@ defmodule Electric.Connection.Manager do
...
+      replication_client_pid,
+      lock_connection_pid
+    ]
+    |> Enum.filter(&is_pid/1)
+    |> Enum.map(fn pid ->
+      Process.monitor(pid)
+      Process.exit(pid, :shutdown)
+
+      receive do
+        {:DOWN, _ref, :process, ^pid, _reason} -> :ok
```
robacourt:
```suggestion
      ref = Process.monitor(pid)
      Process.exit(pid, :shutdown)

      receive do
        {:DOWN, ref, :process, ^pid, _reason} -> :ok
```

packages/sync-service/lib/electric/postgres/lock_connection.ex
```diff
@@ -32,8 +32,8 @@ defmodule Electric.Postgres.LockConnection do
     ]
   end
 
-  def name(stack_id) do
-    Electric.ProcessRegistry.name(stack_id, __MODULE__)
+  def name(opts) do
+    Electric.ProcessRegistry.name(Access.fetch!(opts, :stack_id), __MODULE__)
   end
```
robacourt: Why this change?

msfstef: Ah I had made it initially to make it consistent with other modules in order to call this across all of them but it's no longer needed - I'll revert it

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -937,6 +940,49 @@ defmodule Electric.Connection.Manager do
...
+
+    receive do
+      {:DOWN, ^ref, :process, ^pid, _reason} ->
+        :ok
+    after
+      5000 ->
+        Process.exit(pid, :kill)
+
+        receive do
+          {:DOWN, _ref, :process, ^pid, _reason} ->
```
robacourt:
```suggestion
        receive do
          {:DOWN, ^ref, :process, ^pid, _reason} ->
```

robacourt (approved):
Lovely! Thanks for updating it

robacourt (approved):
Great work!

PR #2871, opened 2025-07-03
fix: Always include latest versions

msfstef:
With our current approach we end up skipping Docker image updates because the commits get overwritten - I propose always sending the latest versions of both the commit sha and the docker version and we just ensure we don't do unnecessary updates at the target.

### DIFF
--- a/.github/workflows/changesets_release.yml
+++ b/.github/workflows/changesets_release.yml
@@ -56,14 +56,13 @@ jobs:
           fetch-depth: 0
       # Get the Electric version of the Docker image
       - name: Get Electric version
-        if: needs.changesets.outputs.published == 'true'
         id: sync_version
         run: |
           VERSION=$(jq -r '.version' packages/sync-service/package.json)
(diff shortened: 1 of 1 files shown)


.github/workflows/changesets_release.yml
```diff
@@ -56,14 +56,13 @@ jobs:
...
-        if: needs.changesets.outputs.published == 'true'
         id: sync_version
         run: |
           VERSION=$(jq -r '.version' packages/sync-service/package.json)
           echo "version=${VERSION}" >> $GITHUB_OUTPUT
-      - name: Trigger cloud update (on release)
+
+      - name: Trigger cloud update
         uses: peter-evans/repository-dispatch@v2
-        if: needs.changesets.outputs.published == 'true'
```
kevin-dp, line 66: I believe that `steps.sync_version.outputs.version` will be empty if we are in a PR instead of on a release because the changesets won't have run so it won't have published to docker. Can you double check?

kevin-dp (commented):
I have a concern regarding the docker image, see my comment.

PR #2875, opened 2025-07-07
fix: account for PG14 when calculating if publication needs updating

icehaunter:


packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -407,13 +408,16 @@ defmodule Electric.Replication.PublicationManager do
         pg_version,
         publication_name
       )
+      |> dbg
```
msfstef:
```suggestion
```

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -407,13 +408,16 @@ defmodule Electric.Replication.PublicationManager do
...
 
     {:ok, state, missing_relations}
   rescue
     # if we are unable to do row filtering for whatever reason, fall back to doing only
     # relation-based filtering - this is a fallback for unsupported where clauses that we
     # do not detect when composing relation filters
     err ->
+      dbg()
```
msfstef:
```suggestion
```

packages/sync-service/test/electric/replication/publication_manager_test.exs
```diff
@@ -236,6 +237,33 @@ defmodule Electric.Replication.PublicationManagerTest do
       refute_receive {:filters, _}, 500
     end
 
+    @tag update_debounce_timeout: 100
+    @tag pg_version: 14_0001
+    test "should not update publication if new shape adds nothing when where clauses aren't considered",
```
msfstef, line 242: nit: test title a bit confusing - aren't we just testing that pg14 never updates the where clause?

msfstef (approved):
Nice !

packages/sync-service/test/electric/replication/publication_manager_test.exs
msfstef: nevermind, test makes sense in its entirety

icehaunter: Yup, but that test failed - we called the configure regardless of the tested PG version which led to debouncing on every request because the flag that where clauses shouldn't be considered was never set

PR #2878, opened 2025-07-08
fix: Handle more pool connection exit reasons

msfstef:
We are occasionally getting crashes because there is no function clause matching for `:pool_conn_down` messages with reason `:shutdown`, as we only cover `{:shutdown, reason}` (see [this Sentry issue](https://electricsql-04.sentry.io/issues/48398126/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream&stream_index=20))

I've generalized the handler for `:pool_conn_down` to handle all exit reasons, since they are mostly no-ops anyway.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -720,13 +720,12 @@ defmodule Electric.Connection.Manager do
 
   # When a pooled connection terminates, we log its exit reason, but more connections will
   # be started by the connection pool supervisor, so we don't need to do anything else.
-  def handle_info({:pool_conn_down, _ref, :process, _pid, reason}, state) do
-    error =
-      case reason do
-        {:shutdown, exit_reason} -> exit_reason
-        exit_reason -> exit_reason
-      end
-      |> DbConnectionError.from_error()
+  def handle_info({:pool_conn_down, _ref, :process, _pid, :shutdown}, state) do
+    {:noreply, state}
+  end
+
+  def handle_info({:pool_conn_down, _ref, :process, _pid, {:shutdown, exit_reason}}, state) do
+    error = DbConnectionError.from_error(exit_reason)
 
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -720,8 +720,13 @@ defmodule Electric.Connection.Manager do
...
   # be started by the connection pool supervisor, so we don't need to do anything else.
-  def handle_info({:pool_conn_down, _ref, :process, _pid, {:shutdown, exit_reason}}, state) do
-    error = DbConnectionError.from_error(exit_reason)
+  def handle_info({:pool_conn_down, _ref, :process, _pid, reason}, state) do
+    error =
+      case reason do
+        {:shutdown, exit_reason} -> exit_reason
+        exit_reason -> exit_reason
+      end
+      |> DbConnectionError.from_error()
```
robacourt: So given the specific [Sentry error](https://electricsql-04.sentry.io/issues/48398126/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream&stream_index=20) you saw, that would do a `DbConnectionError.from_error(:shutdown)` which would be an `:unknown` error  since we only recognise `:shutdown` when `Mix.env() == :test`.

It looks like we're just using DbConnectionError here to see if the error is recognised, and log it if not, so personally I wouldn't change DbConnectionError, I would just add another function head:

```suggestion
  def handle_info({:pool_conn_down, _ref, :process, _pid, :shutdown}, state) do
    {:noreply, state}
  end

  def handle_info({:pool_conn_down, _ref, :process, _pid, {:shutdown, exit_reason}}, state) do
    error = DbConnectionError.from_error(exit_reason)
```

msfstef: fair enough

robacourt (approved):

PR #2881, opened 2025-07-08
feat(sync-service): Allow configuration of TCP send timeout

robacourt:
AntonOfTheWoods experiences the following error message:
```
Error while streaming response: :timeout
```

This is most probably due to slow response processing on the client side causing TCP backpressure. AntonOfTheWoods has asked for configuration to increase the timeout, so this PR introduces `ELECTRIC_TCP_SEND_TIMEOUT` to allow configuration of the TCP/SSL send timeout.

The default is `30s` as it is right now, but for example the timeout can be set to 60 seconds with:
```
ELECTRIC_TCP_SEND_TIMEOUT=60s
```

### DIFF
--- a/.changeset/chilled-oranges-type.md
+++ b/.changeset/chilled-oranges-type.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Add ELECTRIC_TCP_SEND_TIMEOUT to allow configuration of TCP/SSL send timeout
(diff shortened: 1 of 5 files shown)


msfstef (approved):
Nice, would it be easy to add a test for this or is it not easily reproducible?

robacourt:
> Nice, would it be easy to add a test for this or is it not easily reproducible?

I time-boxed a day to try, but I couldn't in that time. I've verified that the option is working in a rather roundabout way, but it's not suitable for an automated test.

> Also, should this be added to the docs, or should we leave it undocumented for now?

Good idea. Will do

PR #2887, opened 2025-07-10
fix: Require pg version for publication manager

msfstef:
No reason we should be seeing [errors like these](https://electricsql-04.sentry.io/issues/14716077/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream&stream_index=1) when we already have a `pg_version` available from the replication client.

Removing code that's unnecessary, simplifying the publication manager.

### DIFF
--- a/.changeset/tough-cows-kick.md
+++ b/.changeset/tough-cows-kick.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Ensure publication manager is given pg_version to remove redundant queries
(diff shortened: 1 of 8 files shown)


alco (approved):
Nice!

I've pushed two cleanup commits. @robacourt if you could take a look at those, we can merge this into main.

robacourt:
> I've pushed two cleanup commits. @robacourt if you could take a look at those, we can merge this into main.

LGTM

PR #2888, opened 2025-07-11
chore: update quickstart compose file

abn:
This changes uses fully qualified names for container images in the quickstart compose file.

Additionally, a healthcheck is added to the postgres service to allow electric service to wait till the database is ready before starting.

### DIFF
--- a/website/public/docker-compose.yaml
+++ b/website/public/docker-compose.yaml
@@ -1,15 +1,14 @@
-version: "3.3"
 name: "electric_quickstart"
 
 services:
   postgres:
-    image: postgres:16-alpine
+    image: docker.io/postgres:16-alpine
     environment:
       POSTGRES_DB: electric
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: password
     ports:
-      - 54321:5432
+      - "54321:5432"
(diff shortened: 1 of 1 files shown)


thruflo:
Thanks for the contribution. I think integrating the migrations into this docker compose file is good for the quickstart flow.

However, the same docker compose file is also linked to and used independently as the easiest way to fire up an Electric. In those cases, I don’t think it makes sense to run the quickstart migration.

abn:
Ah fair enough. I guess there are two options:

1. I can roll back the config map change and keep just the health check and the FQDN for the container images.
2. Create two compose files `docker-compose.quickstart.yml` and `docker-compose.yml` that allows for both usecases.

Do you have a preference?

KyleAMathews:
The docker health check / FQDN seem like a great idea but I'd rather we keep running the migrations manually as is — we already require people to use psql for the quickstart. The point is it's rather manual but shows exactly what's going on, no magic.

abn:
@KyleAMathews makes sense to me. I have updated the branch to just keep the changes to health check, service condition and FQDN.

KyleAMathews (approved):
Thanks!

PR #2894, opened 2025-07-16
Use descriptive names for CI jobs

alco:
We have been erroneously treating job naming as if namespaced under the workflow name. It doesn't always work, like when viewing the CI runs for an already merged PR. And in general it's not too helpful to have different jobs all called "Build and test".

image


robacourt (approved):
Much nicer, thanks!

PR #2895, opened 2025-07-17
Benchmark replication throughput with publication row filters enabled/disabled

alco:
## Benchmarking results

All benchmark runs were using the `postgres:17-alpine` Docker image for the database.

### Comparison 1: Static or indexed where clauses

| Parameter | Value |
------------|--------------
| row_size | 100 |
| tx_count | 100, 1000 |
| tx_row_count | 100 |
| shape_count | 5, 2000, 4000, 8000, 16000 |
| where_clause | `name = '#{name}'`, `name > 'az'` |

**Green** - row filters enabled, **Purple** - row filters disabled

470-471-row-filter-comparison

### Comparison 2: Large number of shapes, unindexed where clause

| Parameter | Value |
------------|--------------
| row_size | 50 |
| tx_count | 100 |
| tx_row_count | 200 |
| shape_count | 10000, 20000, 30000, 60000 |
| where_clause | `name = '#{name}'`, `name ILIKE '#{name}%'` |

**Green** - row filters enabled, **Purple** - row filters disabled

472-490-more-shapes

---

**N.B.** The last curve seems to be affected by the maximum runtime cap of a client request. Here's an updated chart for the large number of shapes with the runtime cap fixed.

This one has fewer data points because the last two benchmark runs failed to process the shape request with the `name ILIKE '#{name}%'` where clause with row filtering disabled. This is concerning...

511-504-more-shapes-no-cap

### Comparison 3: Large number of shapes, revisited

Here's the final comparison between enabled and disabled row filters. I had to lower the number of txs and the number of rows per tx for each test run to obtain data points that were missing from the last chart above.

| Parameter | Value |
------------|--------------
| row_size | 50 |
| tx_count | 50 |
| tx_row_count | 20 |
| shape_count | 10000, 20000, 30000, 60000 |
| where_clause | `name = '#{name}'`, `name ILIKE '#{name}%'` |

**Green** - row filters enabled, **Purple** - row filters disabled

512-513-more-shapes-all-data-points

### DIFF
--- a/.github/workflows/benchmarking.yml
+++ b/.github/workflows/benchmarking.yml
@@ -10,6 +10,7 @@ defaults:
 env:
   PROJECT_ID: vaxine
   REGISTRY: europe-docker.pkg.dev/vaxine/electric
+  ISSUE_NUMBER: ${{ github.event.pull_request.number }}
 
 jobs:
   build:
@@ -51,8 +52,8 @@ jobs:
         run: |
           docker pull ${{ env.REGISTRY }}/electric:canary-builder || true
           docker pull ${{ env.REGISTRY }}/electric:canary-runner-base || true
-          docker pull ${{ env.REGISTRY }}/electric:pr-${{ github.event.issue.number }}-builder || true
-          docker pull ${{ env.REGISTRY }}/electric:pr-${{ github.event.issue.number }}-runner-base || true
+          docker pull ${{ env.REGISTRY }}/electric:pr-${{ env.ISSUE_NUMBER }}-builder || true
+          docker pull ${{ env.REGISTRY }}/electric:pr-${{ env.ISSUE_NUMBER }}-runner-base || true
       - name: Build base images
         env:
           DOCKER_BUILDKIT: "1"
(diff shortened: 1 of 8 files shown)


balegas:
Links seem to be broken

alco:
> Links seem to be broken

I have access to benchmark results via Livebook files in the benchmarking-fleet repo. As soon as the last benchmark run has finished, I'll post the last perf comparison charts here.

alco:
4 out of 8 test runs failed for 2471d8c817094b89ea63d5ce93dab7511f8735a9. Rerunning it once again.

alco:
I've added a new chart to the description of the PR. Since it's not showing all the data points we're looking for, I'm scheduling another benchmark run using fewer transactions and fewer rows per transaction.

alco:
@balegas @KyleAMathews I've just added the latest benchmarking results to the PR description.

icehaunter:
@alco about the benchmarks - "processing time per change" implies it was measured on Electric. Can you run the indexed/unindexed * filtered/unfilted benchmark matrix for write falling into one of 10k shapes measured as a full round-trip? I'm interested in how much the where clause on PG affects the speed of writing to WAL as measured together with Electric

balegas:
So, the reading of the benchmarks is that it is the publication row filtering still give us and advantage particularly for unindexed where clauses, with the advantage that we filter out data at the source. It also didn't blow Postgres, which make it sound that the where clause filtering is optimized beyond our imagination :).

I'm also interested on the load on Postgres, or any other statistical information that would allow us to evaluate if we're straining Postgres more when using the publication filters. However, the number are so much worse without publication row filtering for the unindexed benchmark that we could already conclude that we can't disable it.

KyleAMathews:
> However, the number are so much worse without publication row filtering for the unindexed benchmark that we could already conclude that we can't disable it.

That's only considering part of the question.

Look also at how much *faster* the vast majority of shapes will get if we remove this. Electric is increasingly useless as the number of shapes gets larger and that's a big problem as we want to support 100k+ shapes. 

Another important perspective is that only large users have 10k+ shapes. These comparison charts don't apply at all to 99% of users. So we have to shift our thinking from "oh we might accidentally slow down a random user's instance" to "only large users are going to care about indexed vs unindexed where clauses and large users tend to be more sophisticated and/or engage with us for support so are going to do benchmarking and read the docs more throughly so we can direct them to shape where clauses that are fast while we index more of the remaining."

Finally, deleting a bunch of fragile code makes our codebase easier to work in and reduces a big surface area for bugs or strange behavior — *especially* again for the only group that actually cares about this, large users doing unpredictable things.

balegas:
Oleksii is having a look into indexing ILIKE where clauses. If we get them indexed, we'll disable publication filtering for where clauses (keeping tables). Otherwise the best of both worlds would be to disable publication filtering only when we have indexed shapes.

We though Pg would blow with more WHERE clauses on the publication but it is doing quite well. So, if we only use the publications where clause filtering when we can't index (and we know these will be a small minority) I don't see a big reason to have a regression in latency for those shapes.

KyleAMathews:
One, Two, Trie, go!

alco:
Usage of publication row filtering has been disabled in https://github.com/electric-sql/electric/pull/2914.

PR #2903, opened 2025-07-22
Prevent Electric from crashing if it doesn't have the privilege to create a publication

alco:
Instead, check if the publication already exists in the database and that it publishes all the required operations: INSERT, UPDATE, DELETE, TRUNCATE.


alco:
Subsumed by https://github.com/electric-sql/electric/pull/2942.

PR #2914, opened 2025-07-29
Stop utilizing publication row filtering

alco:
This removes the code that was responsible for adding and maintaining the row filter on the publication (aka the WHERE clause).

We're still waiting on the final benchmark results from https://github.com/electric-sql/electric/pull/2895#issuecomment-3139023498 to have a better understanding of transaction processing times with row filtering enabled and disabled.

### DIFF
--- a/.changeset/late-berries-do.md
+++ b/.changeset/late-berries-do.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Disable the use of row filters in the Postgres publication.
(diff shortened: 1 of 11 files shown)


balegas:
Sanity check that we still keep tables in the publication?

alco:
I've just rebased this on top of the manual publication management branch. As soon as that one's merged, we can merge this one.

alco:
> Sanity check that we still keep tables in the publication?

All good on that front. I've added a test to verify that a table remains in the publication as long as there's at least one shape mentioning it.

Issue #2900, opened 2025-07-19
Force reset the stack when encountering a update with no old version


icehaunter:
**Versions**
- Electric: 1.1 & below 

**Bug description**
We can encounter updates with no changes - for example when table was added to the publication without altering the replica identity. Or there is a bug or misconfiguration on postgres side. In any case, we [print the error](https://github.com/electric-sql/electric/blob/46bf14453f190b2bee1c6c8d341d9b07330f9a97/packages/sync-service/lib/electric/postgres/replication_client/collector.ex#L85-L93) - but then fail later on because code assumes that this cannot happen.

**Expected behavior**
We should instead trigger a full stack failure, resetting the replication slot, dropping all shapes and starting from scratch, as that's the only way to make sure we're not missing any updates and all shapes have consistent and complete view of the data.

msfstef:
Is this related to https://github.com/electric-sql/electric/issues/2911 ?

Issue #2101, opened 2024-12-05
Add a hard limit to the transaction size


balegas:
Electric loads entire transactions from Postgres logical replication into memory before starting to write to shape logs. For very large transactions, this might exhaust memory, crashing the server.

While we don't have to handle arbitrarily large transactions, we shall set a limit to the amount of memory a transaction can use and handle erros. 

Handling errors is not trivial, since Electric can't continue consuming the replication stream. Dropping all shapes and recreating the replication slot might be a way of getting out of that state.

**Exhausting memory shouldn't happen very often and does not affect initial sync of shapes**

KyleAMathews:
Nice — yeah if we can't handle it gracefully then we really have no alternative but to dropping the transaction and killing all shapes and starting over.

We could probably set the limit based on the amount of memory of the server? If someone has trouble with big transactions then the fix is just getting a bigger box.

PR #2915, opened 2025-07-29
fix: Restart stack on irrecoverable replication issues

msfstef:
Fixes https://github.com/electric-sql/electric/issues/2900
Fixes https://github.com/electric-sql/electric/issues/2101

We will now be handling the following two cases:
1. Transaction comes in with a replica identity not set to `FULL` - something got misconfigured, and the stream is no longer valid since changes don't contain all information we need
2. Transaction gets collected and its data content surpasses `max_txn_size`, which defaults to 250 MB (I can make it env var configurable)

When these things happen, we exit the replication client with an `:irrecoverable_slot` error, which gets parsed as a `DbConnectionError`.

I've added a `drop_replication_slot?` field on these errors to make it clearer whether an error should trigger slot drop, and subsequently a full stack reset. Combined with a `retry_may_fix?` flag, this causes the slot and shapes to be cleaned up and thus we replicate from scratch, so the above 2 issues should resolve as the large transaction will be now part of the snapshot, and the replica will be reconfigured by the publication manager.

I've also renamed `:replication_slot_invalidated` to `:database_slot_exceeded_max_size` to make it consistent across the code, and made that error retryable with a replication slot drop - a warning stack event is still fired but no reason we shouldn't just restart things.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -295,8 +295,6 @@ defmodule Electric.Postgres.ReplicationClient do
   end
 
   defp process_x_log_data(data, _wal_end, %State{stack_id: stack_id} = state) do
-    msg_size = byte_size(data)
-
     OpenTelemetry.timed_fun("decode_message_duration", fn -> decode_message(data) end)
     # # Useful for debugging:
     # |> tap(fn %struct{} = msg ->
@@ -308,7 +306,7 @@ defmodule Electric.Postgres.ReplicationClient do
     #       message_type <> " :: " <> inspect(Map.from_struct(msg))
     #   )
     # end)
-    |> Collector.handle_message_with_size(msg_size, state.txn_collector)
+    |> Collector.handle_message(state.txn_collector)
     |> case do
       {:error, reason, _} ->
         exit({:irrecoverable_slot, reason})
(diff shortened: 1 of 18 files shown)


robacourt (approved):
Nice!

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -291,8 +308,11 @@ defmodule Electric.Postgres.ReplicationClient do
     #       message_type <> " :: " <> inspect(Map.from_struct(msg))
     #   )
     # end)
-    |> Collector.handle_message(state.txn_collector)
+    |> Collector.handle_message_with_size(msg_size, state.txn_collector)
     |> case do
+      {:error, reason, _} ->
+        exit({:irrecoverable_slot, reason})
```
alco: Nit: it's unusual to see a process call `exit()` on itself. Any specific reason this is not a `raise`?

msfstef: no specific reasons - still learning what's idiomatic!
would a `raise {:irrecoverable_slot, reason}` trigger the same `{:EXIT, {:irrecoverable_slot, reason}}` message for the process trapping them?

msfstef: I've made it raise a `DbConnectionError` directly, and made it so that `DbConnectionError.from_error` called on a `DbConnectionError` is  a no-op.

I could create a separate exception for the replication client and make `DbConnectionError` parse that but I figured we might as well use it directly - is this better?

alco: Thanks! If anything, it is now more consistent with what we generally do to make an error noticeable in the code. Let's leave it at that.

PR #2916, opened 2025-07-29
feat(sync-service) Sample telemetry at source for faster replication stream processing

robacourt:
Screenshot 2025-07-29 at 15 01 56

This PR speeds up replication processing by 65%. The above benchmarks were run via docker which slows them down a bit, without docker processing time came down to 23µs (43.5K TPS) which is about [as fast as Postgres can go](https://www.enterprisedb.com/blog/performance-comparison-major-PostgreSQL-versions) (on a 72 CPU  144 GB RAM machine)

The performance improvements come from:
- Bespoke upfront span sampling. Standard OpenTelemetry sampling acts once the spans have been created. This bespoke mechanism samples at source to reduce the overhead of `with_span` from 13µs to practically nothing.
- Transaction logging is now at debug level
- Transaction metrics are sampled at source
- The `pg_txn.replication_client.process_x_log_data` span which measured decoding duration and other durations before the transaction was formed has been replaced by attributes on the `pg_txn.replication_client.transaction_received` which not only includes decoding time, it has full granular interval timing for message collection time, time spent waiting for postgres, updateing the applied WAL, etc. 

These different changes can be seen separately in the rebased commits of this PR.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -12,6 +12,7 @@ defmodule Electric.Postgres.ReplicationClient do
   alias Electric.Postgres.ReplicationClient.ConnectionSetup
   alias Electric.Replication.Changes.Relation
   alias Electric.Telemetry.OpenTelemetry
+  alias Electric.Telemetry.Sampler
 
   require Logger
 
@@ -318,17 +319,19 @@ defmodule Electric.Postgres.ReplicationClient do
 
         {m, f, args} = state.transaction_received
 
-        :telemetry.execute(
-          [:electric, :postgres, :replication, :transaction_received],
-          %{
-            monotonic_time: System.monotonic_time(),
-            receive_lag: DateTime.diff(DateTime.utc_now(), txn.commit_timestamp, :millisecond),
-            bytes: byte_size(data),
-            count: 1,
-            operations: txn.num_changes
-          },
-          %{stack_id: state.stack_id}
-        )
+        if Sampler.sample?() do
+          :telemetry.execute(
+            [:electric, :postgres, :replication, :transaction_received],
+            %{
+              monotonic_time: System.monotonic_time(),
+              receive_lag: DateTime.diff(DateTime.utc_now(), txn.commit_timestamp, :millisecond),
+              bytes: byte_size(data),
+              count: 1,
+              operations: txn.num_changes
+            },
+            %{stack_id: state.stack_id}
+          )
+        end
 
         # this will block until all the consumers have processed the transaction because
         # the log collector uses manual demand, and only replies to the `call` once it
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -283,7 +283,9 @@ defmodule Electric.Postgres.ReplicationClient do
         <<@repl_msg_x_log_data, _wal_start::64, _wal_end::64, _clock::64, data::binary>>,
         %State{stack_id: stack_id} = state
       ) do
-    decode_message(data)
+    data
+    |> OpenTelemetry.start_interval("replication_client.decode_message")
+    |> decode_message()
     # # Useful for debugging:
     # |> tap(fn %struct{} = msg ->
     #   message_type = struct |> to_string() |> String.split(".") |> List.last()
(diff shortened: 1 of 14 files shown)


robacourt:
benchmark this

robacourt:
benchmark this

robacourt:
> Benchmark results, triggered for 82c35

I believe the differences shown in these benchmarks are not statistically significant, but I am in the process of verifying this.

balegas:
Great work. Amazing how much performance we can get by optimizing telemetry and logging. Every µs counts!

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -279,8 +268,13 @@ defmodule Electric.Postgres.ReplicationClient do
...
   end
 
-  defp process_x_log_data(data, _wal_end, %State{stack_id: stack_id} = state) do
-    OpenTelemetry.timed_fun("decode_message_duration", fn -> decode_message(data) end)
+  def handle_data(
+        <<@repl_msg_x_log_data, _wal_start::64, _wal_end::64, _clock::64, data::binary>>,
+        %State{stack_id: stack_id} = state
+      ) do
+    data
+    |> OpenTelemetry.start_interval("replication_client.decode_message")
```
alco, line 287: I wonder if importing and inlining this function could yield a few more microseconds of improvement.

Erlang by default does no optimization on external functions to suppor code reloading. But we know that in our case it's the time that matters, not the ability to reload code at runtime.

packages/sync-service/lib/electric/telemetry/open_telemetry.ex
```diff
@@ -55,13 +55,27 @@ defmodule Electric.Telemetry.OpenTelemetry do
...
 
+  @doc """
+  Creates a span providing there is a parent span in the current context.
+  If there is no parent span, the function `fun` is called without creating a span.
+
+  This is necessary for the custom way we do sampling, if the parent span is not sampled, the child span
+  will not be created either.
+  """
+  def with_child_span(name, attributes, stack_id \\ nil, fun) do
+    case get_current_context() do
```
alco: We could use a specially crafter function to check if there's current context and avoid calling `:otel_baggage.get_all()` which happens inside `get_current_context()`.

alco (approved):
Beautiful!

packages/sync-service/lib/electric/telemetry/open_telemetry.ex
robacourt: Great idea! Done

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: As discussed, will look at in separate PR

alco: Disregard my suggestion. Neither importing nor inlining makes any difference on the generated code.

I've found a [confirmation from the Erlang team](https://erlangforums.com/t/when-exactly-does-the-erlang-compiler-inline-code/1302/4) that inlining only works within the module boundary, so only local function calls can be inlined. And importing a function from a module is syntactic sugar, both in Elixir and Erlang. In the generated code it's still a fully-qualified function call.

E.g. for this Elixir code:
```
    timer = start_interval("shape_log_collector.transaction_message")
    trace_context = get_current_context()

    timer = GenServer.call(server, {:new_txn, txn, trace_context, timer}, :infinity)

    OpenTelemetry.stop_and_save_intervals(
      timer: timer,
      total_attribute: :"shape_log_collector.transaction.total_duration_µs"
    )

    :ok
  end
```
I get this generated code (read from the module's .beam file and pretty-printed as Erlang code):
```
store_transaction(#{'__struct__' :=
                        'Elixir.Electric.Replication.Changes.Transaction'} =
                      _txn@1,
                  _server@1) ->
    _timer@1 =
        'Elixir.Electric.Telemetry.IntervalTimer':start_interval(<<"shape_log_collector.transaction_message">>),
    _trace_context@1 =
        'Elixir.Electric.Telemetry.OpenTelemetry':get_current_context(),
    _timer@2 = 'Elixir.GenServer':call(_server@1,
                                       {new_txn,
                                        _txn@1,
                                        _trace_context@1,
                                        _timer@1},
                                       infinity),
    'Elixir.Electric.Telemetry.OpenTelemetry':stop_and_save_intervals([{timer,
                                                                        _timer@2},
                                                                       {total_attribute,
                                                                        'shape_log_collector.transaction.total_duration_�s'}]),
    ok.
```

PR #2922, opened 2025-07-30
fix: move metadata & some chunks into memory to avoid CPU overhead for PureStorage

icehaunter:
- Moves some commonly-accessed metadata into memory (snapshot complete, etc)
- Moves 3 latest chunks into memory which means no files opened for example when requesting `0_inf` offset on a shape that's not seen any changes (previously it was 3 `fopen` syscalls)

### DIFF
--- a/.changeset/swift-vans-battle.md
+++ b/.changeset/swift-vans-battle.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: improve CPU usage of PureFileStorage by lowering syscall count


packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -567,8 +575,8 @@ defmodule Electric.ShapeCache.PureFileStorage do
        write_position: json_file_size,
        key_file_write_pos: key_file_size,
        bytes_in_chunk: json_file_size - position,
-       chunk_started?: chunk_status != :complete
-     ), suffix}
+       chunk_started?: match?({{_, nil}, _, _}, List.last(chunks))
```
msfstef: nit: could extract necessary data from line 567 to determine this?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -647,32 +745,58 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
-        {offset, filename, latest_name(opts)}
+        {latest_name(opts), compaction_boundary(opts), {nil, []}}
     end
   end
 
   defp open_files(%__MODULE__{} = opts, suffix) do
     {
       File.open!(json_file(opts, suffix), [:append, :raw]),
-      File.open!(chunk_file(opts, suffix), [:append, :raw]),
+      File.open!(chunk_file(opts, suffix), [:append, :raw, :sync]),
```
msfstef, line 803: was this something that was just missed before? what's the reason to make the chunk file sync (and not the other ones)? I think it warrants a comment

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -717,9 +863,16 @@ defmodule Electric.ShapeCache.PureFileStorage do
         %__MODULE__{} = opts
       )
       when not is_real_offset(min_offset) do
-    if not snapshot_started?(opts), do: raise(Storage.Error, message: "Snapshot not started")
+    # Single ETS lookup to get both snapshot_started? and last_snapshot_chunk
+    metadata =
+      read_multiple_cached_metadata(opts, [:snapshot_started?, :last_snapshot_chunk])
+
+    snapshot_started = Keyword.get(metadata, :snapshot_started?) || false
```
msfstef, line 874:
```suggestion
    snapshot_started = Keyword.get(metadata, :snapshot_started?, false)
```
nit: why not a default? could we be storing a nil?

packages/sync-service/lib/electric/shape_cache/pure_file_storage/chunk_index.ex
```diff
@@ -122,6 +122,59 @@ defmodule Electric.ShapeCache.PureFileStorage.ChunkIndex do
     end)
   end
 
+  def read_last_n_chunks(chunk_file_path, n) do
```
msfstef, line 125: A test on this would be good IMO as it's doing some work that might be prone to misalignment issues

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -595,6 +603,63 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+
+        [storage_meta() = meta] ->
+          # Extract all requested values from the single ETS record
+          Enum.map(keys, fn key ->
+            value =
+              case key do
+                :snapshot_started? -> storage_meta(meta, :snapshot_started?)
+                :compaction_started? -> storage_meta(meta, :compaction_started?)
+                :last_snapshot_chunk -> storage_meta(meta, :last_snapshot_chunk)
+                :last_seen_txn_offset -> storage_meta(meta, :last_seen_txn_offset)
```
msfstef: confused as to why we can look up the cache for these keys with `read_multiple_cached_metedata` but not with `read_cached_metadata`?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -910,10 +1069,57 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+    |> add_opening_chunk_boundary_to_cache(offset, pos)
+    |> update_chunk_boundaries_cache(opts)
+  end
+
+  defp add_opening_chunk_boundary_to_cache(
+         writer_acc(cached_chunk_boundaries: {boundary, cached_chunks}) = acc,
+         offset,
+         pos
+       )
+       when length(cached_chunks) < 3 do
```
msfstef, line 1091:
```suggestion
       when length(cached_chunks) < @num_chunks_cached do
```
Might be good to set this as a module attribute, and add a comment describing the caching strategy (N last written chunks)?

msfstef (approved):
Looks good ! Left a few minor comments

icehaunter:
Benchmark this

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: Yeah, it was. I think this is just cleaner

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: tbh this part was autogenerated, we only really care about the relevant keys, I'll trim it down

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: It's read off-disk as nil if it's missing

PR #2923, opened 2025-07-30
chore: Fix more flakes

msfstef:
- Clean up per-test databases at the end of the suite to avoid errors from hanging connections from the pooler etc - just requires increasing the replication slots, but the finished test databases should be idle anyway
- Ensure consumers are down with a cleanup instruction before asserting storage failures
- Assert storage failures with retries and a timeout to avoid arbitrary large sleeps
- Always wait for shapes to start their snapshot even in mock to avoid missing ETS table errors (because of consumers dying after the shape cache)

## ACTUAL SOURCE CHANGES
- If a log stream read request comes for the last virtual offset before the reals, skip the snapshot logic entirely and try to read the real log directly - fixes `handling with real storage transactions are buffered until snapshot xmin is known` consumer test which was trying to add 1 to the op offset `:infinity`
- When doing SSE requests, we should apply the same logic for checking if updates came in between the request starting and subscribing to changes - this had fixed the flake in `should correctly select columns for initial sync and updates` without SSE but the SSE test was flaking for this reason as it did not replicate the fix. I've abstracted it out.

### DIFF
--- a/.changeset/nine-knives-serve.md
+++ b/.changeset/nine-knives-serve.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Check for changes since request start in SSE requests.
--- a/.changeset/stupid-glasses-sin.md
+++ b/.changeset/stupid-glasses-sin.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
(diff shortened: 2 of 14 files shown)


robacourt (approved):
Fantastic! That's a great set of fixes. Great work!

PR #2942, opened 2025-08-05
Alco/manual table publishing

alco:


packages/sync-service/config/runtime.exs
```diff
@@ -225,7 +225,14 @@ config :electric,
   process_registry_partitions: env!("ELECTRIC_TWEAKS_PROCESS_REGISTRY_PARTITIONS", :integer, nil),
   http_api_num_acceptors: env!("ELECTRIC_TWEAKS_HTTP_API_NUM_ACCEPTORS", :integer, 100),
   tcp_send_timeout:
-    env!("ELECTRIC_TCP_SEND_TIMEOUT", &Electric.Config.parse_human_readable_time!/1, nil)
+    env!("ELECTRIC_TCP_SEND_TIMEOUT", &Electric.Config.parse_human_readable_time!/1, nil),
+  manual_table_publishing?: env!("ELECTRIC_MANUAL_TABLE_PUBLISHING", :boolean, nil),
```
msfstef, line 245: nit: could you add this in the docs in the website as part of this PR?

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -184,6 +188,8 @@ defmodule Electric.Replication.PublicationManager do
       publication_name: opts.publication_name,
       db_pool: opts.db_pool,
       pg_version: opts.pg_version,
+      can_alter_publication?: Map.get(opts, :can_alter_publication?, true),
+      manual_table_publishing?: Map.get(opts, :manual_table_publishing?, false),
```
msfstef: nit: do we need a fallback since we have defined a default in the nimble schema? seems redundant

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -285,6 +294,27 @@ defmodule Electric.Replication.PublicationManager do
              prepared_relation_filters: committed_filters
          }}
 
+      {:error, reason} when is_atom(reason) ->
```
msfstef: this check feels a bit counterintuitive and hard to reason w.r.t. to the following cases - could we either 1. have a guard for the reason being one of the specific ones and 2. for the `manual_table_publishing?` and `can_alter_publication?` to be part of the guard for the case?

The reason being an atom feels like it could override the next two cases from occurring and without a clear enough reason

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -346,6 +377,34 @@ defmodule Electric.Replication.PublicationManager do
...
+       )
+       when not state.can_alter_publication? or state.manual_table_publishing? do
+    # We cannot modify the publication, so all that remains is to check whether the
+    # publication is in the right state for the set of currently active relation filters.
+    if not forced? and Map.keys(current_filters) == Map.keys(committed_filters) do
+      Logger.debug("No changes to publication, skipping checkup")
+      {:ok, state, []}
+    else
+      with {:ok, modified_relations} <-
+             Configuration.check_publication_for_missing_relations(
```
msfstef: it feels counter intuitive that we call `update_publication` when we can't update the publication - which introduces the confusion in the cases discussed above - could we perhaps check these flags earlier on and call a `check_publication` function instead?

packages/sync-service/test/electric/plug/low_privilege_router_test.exs
```diff
@@ -0,0 +1,48 @@
...
+  describe "/v1/shapes" do
+    setup %{db_conn: conn, escaped_db_name: db_name} do
+      Postgrex.query!(conn, "GRANT CREATE ON DATABASE \"#{db_name}\" TO unprivileged", [])
+      %{connection_opt_overrides: [port: 54321, username: "unprivileged"]}
+    end
+
+    setup [:with_complete_stack, :with_router]
+
+    @tag with_sql: ["INSERT INTO items VALUES (gen_random_uuid(), 'test value 1')"]
+    test "GET fails to create initial snapshot when Electric does not own the user table ", %{
```
msfstef, line 30: very nice test setup! might be worth adding a few more here? (I see you've added a few integration tests but the router tests are easier to read/modify/run)

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -11,6 +11,65 @@ defmodule Electric.Postgres.Configuration do
 
   @pg_15 150_000
 
+  @spec check_publication_for_missing_relations(
+          Postgrex.conn(),
+          String.t(),
+          [Electric.oid_relation()],
+          filters()
+        ) :: :ok | {:error, :misconfigured_replica_identity | :table_missing_from_publication}
+  def check_publication_for_missing_relations(
```
msfstef: nit: this implies that it is checking for missing relations, but it is actually checking that the given relations are present and configured correctly, so perhaps a `check_publication_validity` or `correctness` or `check_publication_configuration` to mirror the `configure_publication!` counterpart would be more clear?

msfstef (commented):
Very nice work - glad to see this approach working!

Left some comments specifically w.r.t. to the publication manager code being more explicit + adding the config to the website as part of this PR!

packages/sync-service/config/runtime.exs
alco: I have created a [separate PR](https://github.com/electric-sql/electric/pull/2936) for the docs because code changes need to wait for a release before becoming public but docs are published as soon as the PR is merged into main.

packages/sync-service/lib/electric/replication/publication_manager.ex
alco: Right, this is tautologic. I'm removing these inline defaults.

packages/sync-service/lib/electric/replication/publication_manager.ex
alco: Sounds fair.

At this point there cannot be another atom returned from `update_publication()` because this error is then `raised` by the waiting process. Trying to raise an atom itself crashes, that's how I know.

```
iex(1)> raise :test
** (UndefinedFunctionError) function :test.exception/1 is undefined (module :test is not available)
    :test.exception([])
    iex:1: (file)
```

But I get what you're saying. I'm trying to solve this by eliding the `update_publication` call in case is cannot actually be updated.

packages/sync-service/test/electric/plug/low_privilege_router_test.exs
alco: Yeah, adding integration tests felt more straightforward at the time. I've since added one more test in https://github.com/electric-sql/electric/pull/2947 that covers a couple more edgecases.

alco:
Thanks for the valuable feedback, @msfstef . I have addressed all of it.

balegas:
Please add a reference in configuration docs. It's probably also worth mentioning it here https://electric-sql.com/docs/guides/deployment#_1-running-postgres

alco:
@balegas I've added a paragraph to the Deployment guide - https://github.com/electric-sql/electric/pull/2936/commits/faddc6e264b204fa96a84c91feb30e3a998d6360

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -44,9 +114,7 @@ defmodule Electric.Postgres.Configuration do
       # "New filters" were configured using a schema read in a different transaction (if at all, might have been from cache)
       # so we need to check if any of the relations were dropped/renamed since then
       changed_relations =
-        (previous_relations ++ Map.keys(new_filters))
-        |> Enum.uniq()
-        |> list_changed_relations(conn)
+        list_changed_relations(conn, known_relations(previous_relations, Map.keys(new_filters)))
```
icehaunter: nitpick: why did you reorder arguments to this function that made it unpipable in this case? I think it reads worse like this, at least to me, and the other usage place doesn't gain anything

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -11,6 +11,76 @@ defmodule Electric.Postgres.Configuration do
...
+  If any of the relations in the filters are missing from the publication or don't have their
+  replica identity set to full, an error is returned.
+
+  If some of the relations included in the publication have been modified (e.g. the table name
+  has changed), those will be returned as a list of modified relations from this function to
+  allow for the cleanup of their corresponding shapes.
+  """
+  @spec check_publication_relations_and_identity(
+          Postgrex.conn(),
+          String.t(),
```
icehaunter: wrong spec, I believe
```suggestion
          [Electric.oid_relation()],
```

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -11,6 +11,76 @@ defmodule Electric.Postgres.Configuration do
...
+          [Electric.oid_relation()],
+          filters()
+        ) :: :ok | {:error, :misconfigured_replica_identity | :table_missing_from_publication}
+  def check_publication_relations_and_identity(
```
icehaunter, line 35: This function does 4 PG queries - is it safe that they're not in a transaction?

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -11,6 +11,76 @@ defmodule Electric.Postgres.Configuration do
...
+        new_relations,
+        publication_name
+      ) do
+    known_relations = known_relations(previous_relations, new_relations)
+    changed_relations = list_changed_relations(conn, known_relations)
+
+    published_relations = lookup_published_relations(conn, known_relations, publication_name)
+    {_oids, schemas, tables} = known_relations
+
+    with :ok <- check_relations_in_publication(Enum.zip(schemas, tables), published_relations),
```
icehaunter: I think there's a flaw in logic here, correct me if I'm wrong:
If a table has been dropped, it'll still be present in the list of `known_relations`, but won't be in the publication and need not be in the publication, but it'll still present with an error.

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -11,6 +11,76 @@ defmodule Electric.Postgres.Configuration do
...
+          into: MapSet.new() do
+        relation
+      end
+
+    diff_set = MapSet.difference(known_relations_set, published_relations_set)
+
+    if MapSet.size(diff_set) == 0 do
+      :ok
+    else
+      {:error, :tables_missing_from_publication}
```
icehaunter: I think it'd be more useful in both of these errors to return a list of stuff that's wrong and use it in the error message instead of a generic "Database table is missing from the publication and ..."

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -545,13 +545,19 @@ defmodule Electric.Shapes.Api do
         Logger.warning("Schema changed while creating snapshot for #{shape_handle}")
         Response.error(request, error.message, status: error.status)
 
+      {:error, %SnapshotError{type: :missing_privilege} = error} ->
+        Logger.warning("Failed to create snapshot for #{shape_handle}: #{error.message}")
+        message = "Unable to create initial snapshot: " <> error.message
+        Response.error(request, message, status: 403)
```
icehaunter: nitpick: I don't think 403 is valid, as this is not a client lacking permissions, but entire Electric being misconfigured. I'd rather not confuse Electric's clients with authentication/authorization level codes because they might think this is a genuine "client lacking permissions" issue. I think `500` is the best we can do here.

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -251,39 +257,58 @@ defmodule Electric.Replication.PublicationManager do
...
+      case Configuration.check_publication_relations_and_identity(
+             state.db_pool,
+             Map.keys(committed_filters),
+             Map.keys(current_filters),
+             state.publication_name
+           ) do
+        {:ok, modified_relations} ->
+          update_relation_filters(state, modified_relations)
+
+        {:error, reason} ->
```
icehaunter, line 297: question: I don't see shapes being cleaned in this codepath. Is this intentional? Because if some of them have had a table removed from the publication, we cannot continue them even after they're re-added to the publication because we've missed a data chunk

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -334,6 +402,7 @@ defmodule Electric.Replication.PublicationManager do
   # to the DB.
   @spec update_publication(state()) ::
           {:ok, state(), [Electric.oid_relation()]} | {:error, term()}
+
```
icehaunter, line 431: nitpick

```suggestion
```

packages/sync-service/lib/electric/postgres/configuration.ex
alco: It's idiomatic to pass the value representing a resource or a "thing unto which the action is applied" as a first argument in Elixir. All the other functions in this module that take a conn expect it in the 1st argument position.

I can agree that this particular call site might not look as pretty as before because of this, but it would be the wrong thing to optimize for.

packages/sync-service/lib/electric/postgres/configuration.ex
alco: Yeah, forgot to update the spec after changing the place of the publication_name argument.

packages/sync-service/lib/electric/postgres/configuration.ex
icehaunter: I'd argue you could think that `list_changed_relations` acts upon known relations and `conn` is a side-effect information - it doesn't return the conn after all, but I won't push this

packages/sync-service/lib/electric/replication/publication_manager.ex
alco: Brilliant observation!

packages/sync-service/lib/electric/postgres/configuration.ex
alco: You're absolutely correct.

I'm adding test coverage for this and fixing this edge case.

packages/sync-service/lib/electric/postgres/configuration.ex
alco: Agreed. Fixing.

packages/sync-service/lib/electric/replication/publication_manager.ex
alco: Nah, this blank line is fine since the spec is followed by more than one function clause.

packages/sync-service/lib/electric/shapes/api.ex
alco: Good point. I'm changing it to 503, known_error.

packages/sync-service/lib/electric/postgres/configuration.ex
alco: Hm, I only see two queries: one for the list of changed relations and another one to check the publication tables. By the way, I have updated those queries in the latest commits, so make sure you check them out.

Anyway, wrapping SELECTs in a transaction wouldn't help with anything unless you mean a `REPEATABLE READ` transaction. But we don't need that since we query for changed relations at the current point in time and then query for the list of publication tables. If the DB admin alters the publication between those two, we still get what we want in the end:
- The publication state is validated against known and changed relations. If it is valid at this point in time, it is valid period.
- We still return the list of changed relations to publication manager so it can invalidated matching shapes.
- I've added shape invalidation in the case where this function returns an error, as you've pointed out.

alco:
@icehaunter Thanks for the extremely useful insights! I have addressed your comments and added new commits to this PR. I'm not 100% sure about the correctness of the last one. I had to do that because otherwise the shape consumer would be the first to return a "must refetch" error to the snapshot waiter and the error from the publication manager would be ignored, breaking quite a large chunk of error reporting logic introduced in this PR.

icehaunter (approved):

PR #2943, opened 2025-08-05
feat: improve startup times for existing shapes

### LINKED ISSUE #2928: Faster shape loading
Our startup times are a bit slow when there are a lot of pre-existing shapes. We can benchmark it and make it faster by parallelising

icehaunter:
Closes #2928

Init goes from ~17s to ~7s on MacBook M4 for 10000 shapes. I'll open a follow-up PR with further improvements, but this is big on it's own

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -64,7 +64,7 @@ defmodule Electric.Connection.Manager do
             | {:start_replication_client, :configuring_connection}
             | :start_connection_pool
             | :start_shapes_supervisor
-            | :waiting_for_consumers
+            | {:waiting_for_consumers, integer()}
             | {:start_replication_client, :start_streaming}
             # Steps of the :running phase:
             | :waiting_for_streaming_confirmation
@@ -194,8 +194,8 @@ defmodule Electric.Connection.Manager do
     GenServer.cast(manager, :lock_connection_started)
   end
 
-  def consumers_ready(stack_id) do
-    GenServer.cast(name(stack_id), :consumers_ready)
+  def consumers_ready(stack_id, total_recovered, total_failed_to_recover) do
+    GenServer.cast(name(stack_id), {:consumers_ready, total_recovered, total_failed_to_recover})
   end
 
   def exclusive_connection_lock_acquisition_failed(manager, error) do
@@ -456,6 +456,8 @@ defmodule Electric.Connection.Manager do
       )
     end
 
+    start_time = System.monotonic_time()
+
     shapes_sup_pid =
       case Electric.Connection.Supervisor.start_shapes_supervisor(
              stack_id: state.stack_id,
@@ -482,7 +484,7 @@ defmodule Electric.Connection.Manager do
     state = %{
       state
       | shape_log_collector_pid: log_collector_pid,
-        current_step: :waiting_for_consumers,
+        current_step: {:waiting_for_consumers, start_time},
         purge_all_shapes?: false
     }
 
@@ -883,12 +885,24 @@ defmodule Electric.Connection.Manager do
   end
 
   def handle_cast(
-        :consumers_ready,
+        {:consumers_ready, total_recovered, total_failed_to_recover},
         %State{
           current_phase: :connection_setup,
-          current_step: :waiting_for_consumers
+          current_step: {:waiting_for_consumers, start_time}
         } = state
       ) do
+    duration = System.monotonic_time() - start_time
+
+    Logger.notice(
+      "Consumers ready in #{System.convert_time_unit(duration, :native, :millisecond)}ms (#{total_recovered} shapes, #{total_failed_to_recover} failed to recover)"
+    )
+
+    :telemetry.execute(
+      [:electric, :connection, :consumers_ready],
+      %{duration: duration, total: total_recovered, failed_to_recover: total_failed_to_recover},
+      %{stack_id: state.stack_id}
+    )
+
     state = %State{state | current_step: {:start_replication_client, :start_streaming}}
     {:noreply, state, {:continue, :start_streaming}}
   end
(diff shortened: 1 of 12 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -883,12 +885,24 @@ defmodule Electric.Connection.Manager do
...
         } = state
       ) do
+    duration = System.monotonic_time() - start_time
+
+    Logger.warning(
+      "Consumers ready in #{System.convert_time_unit(duration, :native, :millisecond)}ms (#{total_recovered} shapes)"
+    )
+
+    :telemetry.execute(
+      [:electric, :connection, :consumers_ready],
```
msfstef, line 901: very nice! let's start measuring - should we also somehow also measure how many shapes were found to compare to how many were actually recovered?

packages/sync-service/lib/electric/replication/supervisor.ex
```diff
@@ -34,6 +34,7 @@ defmodule Electric.Replication.Supervisor do
       )
 
     children = [
+      {Task.Supervisor, name: Electric.ProcessRegistry.name(stack_id, :task_supervisor)},
```
msfstef: is this name unique enough? feels potentially a bit generic and there's other task supervisors around

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -715,7 +715,7 @@ defmodule Electric.ShapeCache.PureFileStorage do
     path = Path.join(metadata_dir, "shape_definition.json")
 
     with {:ok, contents} <- File.open(path, [:read, :raw, :read_ahead], &IO.binread(&1, :eof)),
-         {:ok, decoded} <- Jason.decode(contents),
+         {:ok, decoded} <- Jason.decode(if(is_binary(contents), do: contents, else: "")),
```
msfstef, line 718: what could the result be other than a binary, and is it safe to continue the execution with an empty string rather than failing somehow?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -85,30 +103,19 @@ defmodule Electric.Shapes.Consumer do
...
-        monitors: [],
-        cleaned?: false,
-        writer: writer
-      })
-
-    :ok =
-      Electric.Shapes.Monitor.register_writer(config.stack_id, config.shape_handle, config.shape)
-
-    ShapeLogCollector.subscribe(producer, config.shape)
+    ShapeLogCollector.subscribe(producer, state.shape)
```
msfstef, line 112: now that this call lives in a continue, how are we still guaranteed that the stream won't be started before all shapes are registered?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: `:eof` is the other result on an empty file (I'm not using `File.read/1` because it doesn't take `:raw` option). This will fail to decode and will be handleded properly instead of raising

packages/sync-service/lib/electric/replication/supervisor.ex
icehaunter: Yeah, lemme make it more specific

packages/sync-service/lib/electric/connection/manager.ex
icehaunter: Nice idea

packages/sync-service/lib/electric/shapes/consumer.ex
icehaunter: The `start_shape` call in`ShapeCache` is blocking via a GenServer.call to the consumer. I've added a comment there

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -883,12 +885,24 @@ defmodule Electric.Connection.Manager do
...
       ) do
+    duration = System.monotonic_time() - start_time
+
+    Logger.notice(
+      "Consumers ready in #{System.convert_time_unit(duration, :native, :millisecond)}ms (#{total_recovered} shapes, #{total_failed_to_recover} failed to recover)"
+    )
+
+    :telemetry.execute(
+      [:electric, :connection, :consumers_ready],
+      %{duration: duration, total: total_recovered, failed_to_recover: total_failed_to_recover},
```
msfstef, line 902:
```suggestion
      %{duration: duration, recovered: total_recovered, failed_to_recover: total_failed_to_recover},
```
The `total` keyword is a bit confusing w.r.t. to the failed ones (might think it's "total attempted")

packages/sync-service/test/electric/shape_cache_test.exs
```diff
@@ -1235,7 +1235,8 @@ defmodule Electric.ShapeCacheTest do
       for name <- [
             ctx.shape_cache_opts[:server],
             ctx.consumer_supervisor,
-            ctx.shape_log_collector
+            ctx.shape_log_collector,
+            "shape_task_supervisor"
```
msfstef, line 1239: nit: why not follow the same pattern and pass it in the context?

msfstef (approved):
Looks good ! Excited to see this in production

packages/sync-service/lib/electric/connection/manager.ex
icehaunter: The whole metric is going to be `electric.connection.consumers_ready.total`, which reads to me very clearly as "total ready consumers", not "total attemted"

packages/sync-service/test/electric/shape_cache_test.exs
icehaunter: It's just easier to use a constant here tbh

PR #2951, opened 2025-08-05
Remove extra query params from starter

KyleAMathews:


KyleAMathews:
Yeah, I didn't want this code here — we want people to treat shape calls like API calls where only the essential data is passed through the call — so in this case, the table & user.id were already being loaded on the backend (as they should) so it was confusing having this code there.

PR #2952, opened 2025-08-05
Teach Electric to verify database certificate by verifying it against…

alco:
… the root certificate


ConProgramming:
@alco Awesome! Thank you. I'll try when I can. I'm curious why you're using `certfile` here and the Supabase fix used `cacertfile` - what am I missing? https://github.com/supabase/realtime/pull/1473/files

alco:
@ConProgramming We're passing the same `cacertfile` option to Erlang's SSL library as Supabase Realtime. Are you referring to the fact that `certfile` is used in variable names in the code? No particular reason.

`psql` has an connection option called `sslrootcert` which is used for the same purpose. I guess we need to be more precise in the configuration option naming here to avoid causing confusion. Let me ponder whether CA or ROOT would be more fitting for Electric and I'll propose an updated naming.

ConProgramming:
@alco ah I see now. This makes sense to me then - no notes here!

alco:
@ConProgramming Thanks for the feedback! I have changed the option name to `ELECTRIC_DATABASE_CA_CERTIFICATE_FILE` and updated the description in the docs in the hopes of reducing ambiguity.

ConProgramming:
> @ConProgramming Thanks for the feedback! I have changed the option name to `ELECTRIC_DATABASE_CA_CERTIFICATE_FILE` and updated the description in the docs in the hopes of reducing ambiguity.

Amazing! Excited to see this roll out 🙏

PR #2960, opened 2025-08-06
fix: Add health endpoints to yjs and write patterns examples

kevin-dp:
The yjs and write patterns examples were not deploying properly because the `/health` endpoint was missing from the backend API and thus the ECS health checks would fail and force a roll back after failing for a certain amount of time.

### DIFF
--- a/examples/write-patterns/shared/backend/api.js
+++ b/examples/write-patterns/shared/backend/api.js
@@ -178,6 +178,10 @@ app.post('/changes', async (req, res) => {
   return res.status(200).json({ status: 'OK' })
 })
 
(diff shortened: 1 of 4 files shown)


msfstef (approved):
Looks good! Would be good to add health endpoints on the other examples as well (tanstack,todo-app, or whichever other example is deployed)

kevin-dp:
@msfstef The other ones do have health check endpoints (i explicitly checked it for the tanstack example and the others are green on ECS so they must also have the endoint).

msfstef (approved):

Issue #2967, opened 2025-08-07
Pool connections fail to connect, leading to `:killed` messages and failures to checkout connections


msfstef:
**Bug description**
We occasionally see the connection pool being killed with reason `:killed` which seemed concerning.

Looking into it more, it seems that when the pool connections keep dying before they manage to actually connect with reasons like "failed to connect (too_many_connections)", as the target PG can't accept more, and eventually they reach their supervisor's `:max_restarts`/`:max_seconds` leading it to die.

When the pool conn supervisor dies, the `DbConnection.Watcher` sends a kill to the main pool process that leads  to what we observe in the ConnMan.

Our connection listener that looks for `:pool_conn_down` doesn't fire as the connections never actually connect and thus don't fire `:connected` or `:disconnected` events, so these errors only show up in logs.

**Expected behavior**
We should 1) expect to be able to grab a connection in the pool before we consider the stack to actually be ready and 2) see stack events and proper error reports/logs for why the connection pool is failing.

**Proposed solution**
We monitor pool connections via the `:configure` option on the DB pool so that we can detect errors before they manage to connect, and also not consider the connection pool ready until at least one pool connection connects.

PR #2969, opened 2025-08-07
feat: Monitor pooled connections fully and handle failures gracefully

msfstef:
Fixes https://github.com/electric-sql/electric/issues/2967

After some extensive research with @alco on why we were seeing `:killed` messages from the connection pool, we found that it was due to the pool connections failing to connect (see linked issue description).

I started working entirely within ConnMan but was disgusted at how monolithic it was becoming so I decided to split the pool logic out to its own process.

The `Pool` process now handles the actual DbConnection pool, monitors the pool connections fully by linking them with the "supervisor" process and doing some bookkeeping.

I keep track of the most recent pool connection error and return that in case the entire pool dies - this piece of logic can and should become more fleshed out (perhaps track errors with a timeout?) but I find it to be _much_ better than what we currently do and it does the trick.

You can reproduce the problem described in the issue by setting PG's `max_connections` to something lower than 20, which makes the pool unable to become healthy and thus fail with a `:killed` message.

Through this change, ConnMan now also waits for all the pool connections to become healthy, and it also handles restarts of the pool with a backoff, and along with the reporting of the individual pool connection errors this should sort out Cloud reporting as well.

I intend to add an integration test as well for the case where we run out of connections and encounter this, but ran out of time. Happy for someone else to do that or for me to pick this back up when I'm back.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager/pool.ex
+++ b/packages/sync-service/lib/electric/connection/manager/pool.ex
@@ -1,0 +1,226 @@
+defmodule Electric.Connection.Manager.Pool do
+  @moduledoc """
+  A connection pool for managing multiple connections to a PostgreSQL database.
+  """
+
+  use GenServer
+  require Logger
+  alias Electric.DbConnectionError
+
+  @type pool_status :: :starting | :ready | :repopulating
+
+  @type connection_status :: :starting | :connected | :disconnected
+
+  @type t :: %__MODULE__{
+          stack_id: Electric.stack_id(),
+          pool_ref: reference(),
+          pool_pid: pid(),
+          pool_size: non_neg_integer(),
+          connection_manager: GenServer.server(),
+          status: pool_status(),
+          connection_pids: %{pid() => connection_status()}
+        }
+
+  defstruct [
+    :stack_id,
+    :pool_ref,
+    :pool_pid,
+    :pool_size,
+    :connection_manager,
+    status: :starting,
+    connection_pids: %{}
+  ]
+
+  def name(stack_id) when not is_map(stack_id) and not is_list(stack_id) do
+    Electric.ProcessRegistry.name(stack_id, __MODULE__)
+  end
+
+  def name(opts) do
+    name(Access.fetch!(opts, :stack_id))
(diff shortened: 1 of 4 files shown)


msfstef:
> But feels a bit scary to merge this without a test (at least manual) reproducing the pool connection failure case (e.g. due to too_many_connections).

In order to implement the fix I've manually reproduced this by setting the max_connections in the postgres config to a number lower than our pool config - now that I'm back I'll also add some tests so we can merge this

packages/sync-service/lib/electric/connection/manager/pool.ex
```diff
@@ -0,0 +1,259 @@
...
+        connection_listeners: {[self()], pool_ref},
+        # Assume the manager connection might be pooled, so use unnamed prepared
+        # statements to avoid issues with the pooler
+        #
+        # See https://hexdocs.pm/postgrex/0.19.3/readme.html#pgbouncer
+        prepare: :unnamed
+      ]
+
+    {:ok, pool_pid} =
+      pool_mod.start_link(pool_config ++ pool_opts ++ conn_opts)
```
icehaunter, line 102: We're essentially starting the pool without a supervisor and relying on this link to kill the pool & all connections, correct? Slightly concerning because we're essentially not implementing the time-to-kill logic of a normal supervisor if the pool is stuck

icehaunter (approved):
Looks good, but I have one question

packages/sync-service/lib/electric/connection/manager/pool.ex
msfstef: I have reimplemented the same Supervisor shutdown logic that I had added to ConnMan in the `terminate` handler. Unfortunately a lot of our logic in conn man was already doing that and I had added workarounds like this to imitate proper supervisor behaviour, but this is why I want to revisit all of ConnMan with OTP standards.

I considered splitting this into two processes with a supervisor (start the pool manager, then start the actual pool and pass the manager's PID for connection listeners etc) - but it was complicating things when we're already using this pattern now, I'd rather revisit all of conn man together.

PR #2971, opened 2025-08-08
fix starter command in readme.md

FriendlyUsername:
just stumbled upon this, correct starter command should be 
npx gitpick electric-sql/electric/tree/main/examples/tanstack-db-web-starter my-tanstack-db-project


KyleAMathews (approved):
Thanks!

PR #2973, opened 2025-08-09
Add lost env file in tanstack example

ivklgn:
these small changes fix the problem of launching dev mode in `examples/tanstask`


KyleAMathews:
We don't want to checkin a `.env` file — the instructions say to copy the `.env.example` over to `.env`. Thanks for the PR! https://github.com/electric-sql/electric/tree/main/examples/tanstack-db-web-starter#create-a-new-project

PR #2974, opened 2025-08-11
feat: correctly track disk flushes across all shapes

### LINKED ISSUE #2925: Correct acknowledgements on flush
We need to make sure that we acknowledge a transaction as "flushed" only once we've actually flushed it to all shapes - because of a delayed fsync

icehaunter:
Closes #2925

### DIFF
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -52,7 +52,8 @@ defmodule Electric.Postgres.ReplicationClient do
       # "applied" offsets but we only keep track of the "applied" offset which we define as the
       # end LSN of the last transaction that we have successfully processed and persisted in the
       # shape log storage.
-      applied_wal: 0
+      received_wal: 0,
+      flushed_wal: 0
     ]
 
     @type t() :: %__MODULE__{
@@ -70,7 +71,8 @@ defmodule Electric.Postgres.ReplicationClient do
             txn_collector: Collector.t(),
             step: Electric.Postgres.ReplicationClient.step(),
             display_settings: [String.t()],
-            applied_wal: non_neg_integer()
+            received_wal: non_neg_integer(),
+            flushed_wal: non_neg_integer()
           }
 
     @opts_schema NimbleOptions.new!(
@@ -218,6 +220,13 @@ defmodule Electric.Postgres.ReplicationClient do
   end
 
   @impl true
+  def handle_info({:flush_boundary_updated, lsn}, state) do
+    state = %{state | flushed_wal: lsn}
+
+    {:noreply, [encode_standby_status_update(state)], state}
+  end
+
+  @impl true
   def handle_info(:start_streaming, %State{step: :ready_to_stream} = state) do
     ConnectionSetup.start_streaming(state)
   end
@@ -271,7 +280,7 @@ defmodule Electric.Postgres.ReplicationClient do
       # with keepalives to avoid it getting filled with irrelevant changes, like
       # heartbeats from the database provider
       1 ->
-        state = update_applied_wal(state, wal_end)
+        state = update_received_wal(state, wal_end)
         {:noreply, [encode_standby_status_update(state)], state}
 
       0 ->
@@ -367,12 +376,12 @@ defmodule Electric.Postgres.ReplicationClient do
 
             case apply(m, f, [txn | args]) do
               :ok ->
-                OpenTelemetry.start_interval("replication_client.update_applied_wal")
+                OpenTelemetry.start_interval("replication_client.update_received_wal")
                 # We currently process incoming replication messages sequentially, persisting each
                 # new transaction into the shape log store. So, when the applied function
                 # returns, we can safely advance the replication slot past the transaction's commit
                 # LSN.
-                state = update_applied_wal(state, Electric.Postgres.Lsn.to_integer(txn.lsn))
+                state = update_received_wal(state, Electric.Postgres.Lsn.to_integer(txn.lsn))
                 response = [encode_standby_status_update(state)]
 
                 OpenTelemetry.stop_and_save_intervals(
@@ -407,9 +416,9 @@ defmodule Electric.Postgres.ReplicationClient do
   defp encode_standby_status_update(state) do
     <<
       @repl_msg_standby_status_update,
-      state.applied_wal + 1::64,
-      state.applied_wal + 1::64,
-      state.applied_wal + 1::64,
+      state.received_wal + 1::64,
+      state.flushed_wal + 1::64,
+      state.flushed_wal + 1::64,
       current_time()::64,
       0
     >>
@@ -418,10 +427,10 @@ defmodule Electric.Postgres.ReplicationClient do
   @epoch DateTime.to_unix(~U[2000-01-01 00:00:00Z], :microsecond)
   defp current_time(), do: System.os_time(:microsecond) - @epoch
 
-  defp update_applied_wal(state, wal) when is_number(wal) and wal >= state.applied_wal,
-    do: %{state | applied_wal: wal}
+  defp update_received_wal(state, wal) when is_number(wal) and wal >= state.received_wal,
+    do: %{state | received_wal: wal}
 
-  defp update_applied_wal(state, wal) when is_number(wal), do: state
+  defp update_received_wal(state, wal) when is_number(wal), do: state
 
   defp notify_connection_opened(%State{connection_manager: manager} = state) do
     :ok = Electric.Connection.Manager.replication_client_started(manager)
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -6,6 +6,8 @@ defmodule Electric.Replication.ShapeLogCollector do
   use GenServer
 
   require Electric.Postgres.Lsn
+  alias Electric.Postgres.ReplicationClient
+  alias Electric.Replication.ShapeLogCollector.FlushTracker
   alias Electric.LsnTracker
   alias Electric.Replication.ShapeLogCollector.AffectedColumns
   alias Electric.Postgres.Lsn
@@ -78,6 +80,10 @@ defmodule Electric.Replication.ShapeLogCollector do
     GenServer.call(server, {:subscribe, shape})
   end
 
+  def notify_flushed(server, _shape_handle, offset) do
+    GenServer.cast(server, {:writer_flushed, self(), offset})
+  end
+
   def init(opts) do
     Process.set_label({:shape_log_collector, opts.stack_id})
     Logger.metadata(stack_id: opts.stack_id)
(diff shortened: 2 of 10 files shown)


packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
```diff
@@ -0,0 +1,256 @@
...
+          },
+          min_incomlete_flush_tree: :gb_trees.tree(LogOffset.t_tuple(), MapSet.t(shape_id())),
+          notify_fn: (non_neg_integer() -> any())
+        }
+
+  @doc """
+  Create a new flush tracker to figure out when flush boundary moves up across all writers
+
+  When doing a flush across N shapes, it might be delayed on different cadences depending on the amount of data we’re writing.
+  It also might be worth it eventually to break lock-step writes. We need to tell Postgres accurately (enough) when we’ve actually flushed the data it sent us.
```
robacourt, line 29: > accurately (enough)

Is this saying it's not totally accurate because flushes might not have actually written to disk, as mentioned by @alco in #2925 ?

We should probably mention this potential bug here and raise an issue for it.

At least now we can put in a way to verify that the write has actually been written to disk and it needn't be that fast as it will just reduce the offset that's being ACKed rather than holding up replication processing.

packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
```diff
@@ -0,0 +1,256 @@
...
+  2. If Mapping is empty after the update, we’re globally caught up - set `last_global_flushed_offset` to equal `last_seen_offset`
+  3. Otherwise:
+    1. Determine the new global flushed offset:
+       `last_global_flushed_offset = max(last_global_flushed_offset, min(for {_, {_, last_flushed}} <- Mapping, do: last_flushed))`
+       We take the maximum between the already last flushed offset, and the lowest flushed offset across shapes that
+       had not caught up. Because this `min` is expected to be called very often, there should exist a lookup structure to get this `min` in a fast manner
+  4. On last_global_flushed_offset update - notify the replication client with actual transaction LSN:
+    1. If flushes are caught up (i.e. Mapping is empty), then notify with LSN = tx_offset of the last flushed offset
+    2. Otherwise, it’s complicated to determine which transactions have been flushed completely without keeping track of
+       all intermediate points, so notify with LSN = tx_offset - 1, essentially lagging the flush by one transaction just in case.
```
robacourt, line 72: > so notify with LSN = tx_offset - 1, essentially lagging the flush by one transaction just in case

Why is this safe?

packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
```diff
@@ -0,0 +1,256 @@
...
+  end
+
+  defp update_global_offset(
+         %__MODULE__{
+           min_incomlete_flush_tree: min_incomlete_flush_tree,
+           last_global_flushed_offset: prev_last_global_flushed_offset,
+           last_flushed: last_flushed
+         } = state
+       ) do
+    if last_flushed == %{} do
```
robacourt: This could be a separate function head, which would get rid of the nested `if`s

packages/sync-service/lib/electric/shape_cache/file_storage.ex
```diff
@@ -410,6 +410,10 @@ defmodule Electric.ShapeCache.FileStorage do
         end
 
       CubDB.put_multi(db, [{@log_chunk_size_info, new_chunk_size} | entries])
+
+      send(self(), {Storage, :flushed, elem(List.last(log_items), 0)})
```
robacourt, line 414: Do we not need to do this for `InMemoryStorage` too?

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -99,15 +105,27 @@ defmodule Electric.Replication.ShapeLogCollector do
...
         tracked_relations: tracker_state,
         partitions: Partitions.new(Keyword.new(opts)),
-        filter: Filter.new(Keyword.new(opts))
+        filter: Filter.new(Keyword.new(opts)),
+        flush_tracker:
+          FlushTracker.new(
+            notify_fn: fn lsn ->
+              case GenServer.whereis(ReplicationClient.name(opts.stack_id)) do
+                nil -> :ok
+                pid -> send(pid, {:flush_boundary_updated, lsn})
```
robacourt, line 114: It would seem more idiomatic for this to be a `cast`

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -407,9 +416,9 @@ defmodule Electric.Postgres.ReplicationClient do
   defp encode_standby_status_update(state) do
     <<
       @repl_msg_standby_status_update,
-      state.applied_wal + 1::64,
-      state.applied_wal + 1::64,
-      state.applied_wal + 1::64,
+      state.received_wal + 1::64,
+      state.flushed_wal + 1::64,
```
robacourt, line 420: For now this is an improvement, but I think there may be benefits in tracking the applied_wal separately from the flushed_wal to distinguish between what's applied (and so readable by a client) and what's flushed. I'm assuming we can use the same algorithm you used in the flush tracker again but have the txn boundries as the trigger rather than the flush points.

packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
```diff
@@ -0,0 +1,256 @@
...
+    1. If flushes are caught up (i.e. Mapping is empty), then notify with LSN = tx_offset of the last flushed offset
+    2. Otherwise, it’s complicated to determine which transactions have been flushed completely without keeping track of
+       all intermediate points, so notify with LSN = tx_offset - 1, essentially lagging the flush by one transaction just in case.
+
+  Normalization of the txn flush on a writer:
+  1. On incoming transaction, store a mapping of last offset that’s meant to be written by this writer to the whole txn offset
+  2. On a flush, the writer should remove from the mapping all elements that are less-then-or-equal to last flushed offset, and then
+    1. If last removed element from the mapping is equal to the flushed, then use the whole txn offset instead to notify the sender
+    2. Otherwise, use actual last flushed offset to notify the sender.
+  """
```
robacourt: I got confused about this for a while, but I understand it now. Perhaps rather than normalizing calling it `align_to_transaction_boundary` or something like that.

robacourt:
```suggestion
  Aligning the writers flushed offset with the transaction boundary:
  1. On incoming transaction, store a mapping of last offset that’s meant to be written by this writer to the last offset for the txn
  2. On a flush, the writer should remove from the mapping all elements that are less-then-or-equal to last flushed offset, and then
    1. If last removed element from the mapping is equal to the flushed, then use the transaction last offset instead to notify the sender
    2. Otherwise, use actual last flushed offset to notify the sender.
  """
```

packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
```diff
@@ -0,0 +1,256 @@
...
+  When doing a flush across N shapes, it might be delayed on different cadences depending on the amount of data we’re writing.
+  It also might be worth it eventually to break lock-step writes. We need to tell Postgres accurately (enough) when we’ve actually flushed the data it sent us.
+
+  Main problem is that we’re not only flushing on different cadences, but also each shape might not see every operation, so our flush acknowledgement
+  should take a complicated minimum across all shapes depending on what they are seeing. What’s more is that we want to normalize the acknowledged WALs
+  to transaction boundaries, because that’s how PG is sending the data.
+
+  It’s important to note that because shapes are not seeing all operations, they don’t necessarily see the last-in-transaction operation, while the
+  sender doesn’t know how many operations will be sent upfront. Because of that it’s up to the writer to acknowledge the intermediate flushes but also
+  to normalize the last-seen operation to the transaction offset so that the sender can be sure the writer has caught up.
```
robacourt:
```suggestion
  Main problem is that we’re not only flushing on different cadences, but also each shape might not see every operation, so our flush acknowledgement
  should take a complicated minimum across all shapes depending on what they are seeing. What’s more is that we want to align the acknowledged WALs
  to transaction boundaries, because that’s how PG is sending the data.

  It’s important to note that because shapes are not seeing all operations, they don’t necessarily see the last-in-transaction operation, while the
  sender doesn’t know how many operations will be sent upfront. Because of that it’s up to the writer to acknowledge the intermediate flushes but also
  to align the last-seen operation to the transaction offset so that the sender can be sure the writer has caught up.
```

packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
icehaunter: No, it's more about that we might be acknowledging the flush one flush behind. "fsync not fsyncing" is fairly hard to  find I think, although we can learn from PG

packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
icehaunter: This is safe because we'll always ignore already-stored transactions, so our continuations are always better off restarting in the recent past because then we're sure we're not missing anything

packages/sync-service/lib/electric/shape_cache/file_storage.ex
icehaunter: oh yeah, forgor 💀

packages/sync-service/lib/electric/postgres/replication_client.ex
icehaunter: We don't see a real distinction on txn boundaries because it normalizes to the txn boundary anyway - individual operations don't have a WAL position associated, and we're acknowledging an LSN of a COMMIT message essentially

packages/sync-service/lib/electric/replication/shape_log_collector.ex
icehaunter: `ReplicationClient` implements the client thingy from Postgrex which doesn't expose a `handle_cast` /shrug

packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
robacourt: Ok, cool. I've raised an issue, #2982 , for the fsync not fsyncing part.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
robacourt: So it doesn't!

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: Do we flush at the end of every transaction then? I'm surprised that's not slow

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -561,4 +576,14 @@ defmodule Electric.Shapes.Consumer do
     now = DateTime.utc_now()
     Kernel.max(0, DateTime.diff(now, commit_timestamp, :millisecond))
   end
+
+  defp normalize_offset(%{txn_offset_mapping: txn_offset_mapping} = state, offset) do
```
robacourt:
```suggestion
  defp align_to_txn_boundary(%{txn_offset_mapping: txn_offset_mapping} = state, offset) do
```

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -99,15 +105,27 @@ defmodule Electric.Replication.ShapeLogCollector do
         persistent_replication_data_opts: persistent_replication_data_opts,
         tracked_relations: tracker_state,
         partitions: Partitions.new(Keyword.new(opts)),
-        filter: Filter.new(Keyword.new(opts))
+        filter: Filter.new(Keyword.new(opts)),
+        flush_tracker:
+          FlushTracker.new(
+            notify_fn: fn lsn ->
+              case GenServer.whereis(ReplicationClient.name(opts.stack_id)) do
```
alco, line 112: Why do this extra step if we can send a message and not worry about the process being alive to receive it?

PR #2983, opened 2025-08-13
feat(sync-service): Sample metrics with separate env var

robacourt:
Although sampling transaction metrics improves replication processing speed by 10%, @icehaunter [has questioned](https://github.com/electric-sql/stratovolt/issues/742#issuecomment-3173968555) whether the 5µs it saves per transaction is really worth the extra complexity of reading sampled metrics. This PR keeps the ability to sample metrics (set the ENV VAR `ELECTRIC_METRICS_SAMPLING_RATIO=0.01`) for power users but defaults to `ELECTRIC_METRICS_SAMPLING_RATIO=1` (i.e. 100% of metrics are kept, no sampling) for the average user and for our cloud service at this time.

### DIFF
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -191,6 +191,7 @@ config :electric,
   otel_export_period: otel_export_period,
   otel_per_process_metrics?: env!("ELECTRIC_OTEL_PER_PROCESS_METRICS", :boolean, nil),
   otel_sampling_ratio: env!("ELECTRIC_OTEL_SAMPLING_RATIO", :float, nil),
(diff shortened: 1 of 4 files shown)


icehaunter (approved):
That's awesome, but also can we keep around the `with_span` metrics? That call does 2 things, right? it starts the OTEL span, which may or may not be sampled, but also it generates telemetry events that we can act upon separately. I'm happy taking this discussion into a separate PR

robacourt:
> That's awesome, but also can we keep around the `with_span` metrics? That call does 2 things, right? it starts the OTEL span, which may or may not be sampled, but also it generates telemetry events that we can act upon separately. I'm happy taking this discussion into a separate PR

So are you suggesting in `with_span` we should do the `:telemetry.span` even if the sampling doesn't include the span?

PR #2992, opened 2025-08-14
fix: make sure the flush acks are advancing if Electric is up-to-speed

icehaunter:


packages/sync-service/lib/electric/telemetry/stack_telemetry.ex
```diff
@@ -337,6 +337,8 @@ with_telemetry [OtelMetricExporter, Telemetry.Metrics] do
             deadline: 3_000
           )
 
+        dbg(wal_size)
```
robacourt: remove

PR #3001, opened 2025-08-18
fix: Handle `:cpu_sup` failed measurements gracefully

msfstef:
Fixes [Sentry issue](https://electricsql-04.sentry.io/issues/34167118/)

### DIFF
--- a/packages/sync-service/lib/electric/telemetry/application_telemetry.ex
+++ b/packages/sync-service/lib/electric/telemetry/application_telemetry.ex
@@ -317,11 +317,10 @@ with_telemetry [Telemetry.Metrics, OtelMetricExporter] do
                 Map.put(acc, probe, 100 * (value / 256 / cores))
             end
           end)
-          |> then(fn load_map ->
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/telemetry/application_telemetry.ex
```diff
@@ -307,10 +307,21 @@ with_telemetry [Telemetry.Metrics, OtelMetricExporter] do
...
+
+              value ->
+                Map.put(acc, probe, 100 * (value / 256 / cores))
+            end
+          end)
+          |> then(fn load_map ->
+            unless Enum.empty?(load_map) do
+              :telemetry.execute([:system, :load_percent], load_map)
+            end
           end)
```
icehaunter: nitpick: `unless` is deprecated, and I think it'll read better as a case

```suggestion
          |> case do
             x when x == %{} -> :ok
             map -> :telemetry.execute([:system, :load_percent], map)
          end
```

PR #3003, opened 2025-08-18
add an `enabled` option to `useShape` hook that allows conditionally disabling shape synchronization

samwillis:
This addresses the reports we have had from users where they want a way to disable a useShape hook until they are ready to define the params for it.

This should maintain full backwards compatibly on the types as the return type of `useShape` is now a discriminating union with an `isEnabled` flag with overloads that ensure that you get the old type when enabled isn't set or is false.


samwillis:
Closed in favour of steering users to TanStack DB which has this functionality.

PR #3005, opened 2025-08-18
feat: add support for subqueries behind a feature flag

icehaunter:
This PR allows initial, under-optimised implementation of where clauses in the form of `column IN (SELECT id FROM parent WHERE ...)`

It has some limitations:
- Inner select can only have PKs as selected columns
- Composite PKs are not supported
- Runtime evaluation is unindexed, so it'll be fairly slow on a lot of dependent shapes
- Any change inside the inner select (move-in/out) will immediately cause all dependent shapes to be dropped
- Inner select is done as an (unexposed) shape, which is materialized. This materialization is in memory for now, so it's not memory-efficient
- Only 1 inner select is supported for now

This PR is mergable because it doesn't change current behaviour, but adds a feature flag as ENV variable that allows creations of these shapes, to allow us to iterate in main codebase and make this reviewable piece by piece

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -108,7 +108,6 @@ defmodule Electric.Replication.ShapeLogCollector do
         tracked_relations: tracker_state,
         partitions: Partitions.new(Keyword.new(opts)),
         dependency_layers: DependencyLayers.new(),
-        shape_handles_by_pid: %{},
         pids_by_shape_handle: %{},
         filter:
           opts
@@ -129,18 +128,18 @@ defmodule Electric.Replication.ShapeLogCollector do
     {:ok, state}
   end
 
-  def handle_info({:unsubscribe, ref, :process, pid, _reason}, state) do
+  def handle_info({{:unsubscribe, shape_handle}, ref, :process, pid, _reason}, state) do
     OpenTelemetry.with_span("shape_log_collector.unsubscribe", [], state.stack_id, fn ->
       {:noreply,
        state
-       |> remove_subscription({pid, ref})
+       |> remove_subscription({pid, ref}, shape_handle)
        |> Map.update!(:flush_tracker, &FlushTracker.handle_shape_removed(&1, pid))}
     end)
   end
 
   def handle_call({:subscribe, shape_handle, shape}, {pid, _ref}, state) do
     OpenTelemetry.with_span("shape_log_collector.subscribe", [], state.stack_id, fn ->
-      ref = Process.monitor(pid, tag: :unsubscribe)
+      ref = Process.monitor(pid, tag: {:unsubscribe, shape_handle})
       from = {pid, ref}
 
       state =
@@ -148,7 +147,6 @@ defmodule Electric.Replication.ShapeLogCollector do
           state
           | partitions: Partitions.add_shape(state.partitions, shape_handle, shape),
             filter: Filter.add_shape(state.filter, shape_handle, shape),
-            shape_handles_by_pid: Map.put(state.shape_handles_by_pid, pid, shape_handle),
             pids_by_shape_handle: Map.put(state.pids_by_shape_handle, shape_handle, pid),
             dependency_layers:
               DependencyLayers.add_dependency(state.dependency_layers, shape, shape_handle)
@@ -332,7 +330,7 @@ defmodule Electric.Replication.ShapeLogCollector do
     end
   end
 
-  defp remove_subscription(%{subscriptions: {count, set}} = state, {pid, _} = from) do
+  defp remove_subscription(%{subscriptions: {count, set}} = state, from, shape_handle) do
     subscriptions =
       if MapSet.member?(set, from) do
         {count - 1, MapSet.delete(set, from)}
@@ -344,14 +342,11 @@ defmodule Electric.Replication.ShapeLogCollector do
         {count, set}
       end
 
-    shape_handle = Map.fetch!(state.shape_handles_by_pid, pid)
-
     %{
       state
       | subscriptions: subscriptions,
         filter: Filter.remove_shape(state.filter, shape_handle),
         partitions: Partitions.remove_shape(state.partitions, shape_handle),
-        shape_handles_by_pid: Map.delete(state.shape_handles_by_pid, pid),
         pids_by_shape_handle: Map.delete(state.pids_by_shape_handle, shape_handle),
         dependency_layers:
           DependencyLayers.remove_dependency(state.dependency_layers, shape_handle)


packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -509,6 +500,44 @@ defmodule Electric.ShapeCache do
...
+
+          ConsumerSupervisor.start_materializer(%{
+            stack_id: state.stack_id,
+            shape_handle: shape_handle,
+            storage: state.storage,
+            columns: inner_shape.selected_columns,
+            materialized_type: materialized_type
+          })
+
+          Materializer.wait_until_ready(%{
```
msfstef: This seems extremely blocking, as it runs within `ShapeCache` and waits not only for each dependent shape process to start but also for it to fully materialise (basically blocking on an arbitrary snapshot) with an infinite timeout.

I wonder if perhaps the consumer itself should be starting and waiting for these, without blocking the shape cache? Unsure, but definitely some way to keep this flowing better

I realise this might be just for the sake of getting a POC out, just checking if this is a known limitation

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -323,11 +344,17 @@ defmodule Electric.Replication.ShapeLogCollector do
         {count, set}
       end
 
+    shape_handle = Map.fetch!(state.shape_handles_by_pid, pid)
```
msfstef: nit: the `unsubscribe` call is triggered via a tagged `Process.monitor` call in the `subscribe` handler which now has the shape handle - we might as well tag the monitor with `{:unsubscribe, shape_handle}` and get rid of the `shape_handles_by_pid` index?

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -255,7 +270,13 @@ defmodule Electric.Replication.ShapeLogCollector do
 
     OpenTelemetry.start_interval("shape_log_collector.publish")
     context = OpenTelemetry.get_current_context()
-    Publisher.publish(affected_shapes, {:handle_event, event, context})
+
+    for layer <- DependencyLayers.get_for_handles(state.dependency_layers, affected_shapes) do
+      # Each publish is synchronous, so layers will be processed in order
+      layer
+      |> Enum.map(&Map.fetch!(state.pids_by_shape_handle, &1))
```
msfstef, line 275: Do we need `pids_by_shape_handle`? Would it not be enough to just use `Consumer.name(stack_id, shape_handle)` instead and avoid maintaining another lookup?

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -163,19 +167,48 @@ defmodule Electric.Shapes.Shape do
     end
   end
 
+  # We get (table, columns, where) -> we prepare SELECT columns FROM table WHERE where
+  # We parse the complete SELECT statement, and get a topologically sorted list of subqueries
+  # (including the root query)
+  # Then for each subquery we create a shape struct, and put it into context so that next shape
+  #
```
msfstef: supernit: slightly better formatting of the comment (remove empty line, next to function)

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -192,21 +225,92 @@ defmodule Electric.Shapes.Shape do
...
+    Enum.reduce_while(subqueries, {:ok, []}, fn subquery, {:ok, acc} ->
+      case new(Map.put(shared_opts, :select, subquery)) do
+        {:ok, %__MODULE__{root_pk: pk, selected_columns: selected_columns} = shape} ->
+          if Enum.sort(pk) == Enum.sort(selected_columns) do
+            {:cont, {:ok, [shape | acc]}}
+          else
+            {:halt,
+             {:error,
+              {:where,
+               "Subquery has selected #{selected_columns |> Enum.join(", ")} but primary key is #{pk |> Enum.join(", ")}. `IN (SELECT ...)` is supported only for primary key columns."}}}
```
msfstef, line 270: nit: might be worth leaving a comment here as to why we are only limited to selecting primary key columns, for ease of understanding the limitation

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -192,21 +225,92 @@ defmodule Electric.Shapes.Shape do
...
+      with {:ok, column_info, _} <- load_column_info(relation, inspector) do
+        column_info
+        |> Enum.filter(&(&1.name in shape.selected_columns))
+        |> Inspector.columns_to_expr()
+        |> Map.to_list()
+        |> case do
+          [{_, type}] ->
+            {:ok, Map.put(acc, ["$sublink", "#{i}"], {:array, type})}
+
+          [_, _] ->
```
msfstef: why is the other case two columns rather than an arbitrary number? wouldn't a catchall match be better here?

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -163,19 +167,48 @@ defmodule Electric.Shapes.Shape do
...
+         opts = Map.new(opts),
          inspector <- Access.fetch!(opts, :inspector),
          {:ok, {oid, table} = relation} <- validate_relation(opts, inspector),
          {:ok, column_info, pk_cols} <- load_column_info(relation, inspector),
          {:ok, selected_columns} <-
-           validate_selected_columns(column_info, pk_cols, Access.get(opts, :columns)),
+           validate_selected_columns(column_info, pk_cols, Map.get(opts, :columns)),
          refs = Inspector.columns_to_expr(column_info),
-         {:ok, where} <- maybe_parse_where_clause(Access.get(opts, :where), opts[:params], refs),
+         {:ok, parsed_where} <- maybe_parse_where(Map.get(opts, :where)),
```
msfstef: this `with` is generally getting very confusing, and the `maybe_parse_where` and `maybe_parse_where_clause` functions being different are also hard to follow and distinguish.

I wouldn't say this flow is particularly self-documenting, so I think either a fuller description in a comment or a refactor into a clearer flow/namings would help IMO

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -366,7 +413,18 @@ defmodule Electric.Shapes.Consumer do
 
     Logger.debug(fn -> "Txn received in Shapes.Consumer: #{inspect(txn)}" end)
 
-    case filter_changes(changes, shape) do
+    extra_refs =
+      shape.shape_dependencies_handles
```
msfstef: nit: I've seen this exact pattern twice - might be worth refactoring into a shape method or elsewhere? (the desired form of the refs is very specific and am averse to having it written out in various places)

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -426,7 +484,11 @@ defmodule Electric.Shapes.Consumer do
     end
   end
 
-  defp notify_new_changes(state, latest_log_offset) do
+  defp notify_new_changes(state, changes, latest_log_offset) do
+    if state.materializer_subscribed? do
+      Materializer.new_changes(state, changes)
```
msfstef: nit: the passing of the entire process state as is to the materializer module feels oddly coupled - if all it needs is the stack id/shape handle I feel it could be more explicit?

msfstef (approved):
Nice work, love that it's fairly contained!

I've left some comments mostly about 1) some lookups that I think might be avoidable and 2) the `Shape.new` code becoming very hard to follow and understand.

Other limitations are known, I did want to highlight the blocking of `ShapeCache` for materializing subqueries which feels like it should be noted even for the feature flag.

packages/sync-service/lib/electric/shape_cache.ex
icehaunter: Yeah, good point on it blocking the shape cache. I'll think about it before merging

packages/sync-service/lib/electric/shapes/shape.ex
icehaunter: I fixed the naming, and I'll think on how to make this better, good point

packages/sync-service/lib/electric/shapes/consumer.ex
icehaunter: eh, was just easier than `Map.take(state, [:stack_id, :shape_handle])` :D

msfstef: fair enough - I was thinking something along the lines of a `state.materializer` field with what's necessary but that would also be unnecessarily verbose

packages/sync-service/lib/electric/replication/shape_log_collector.ex
robacourt: Great spot! Done

packages/sync-service/lib/electric/shape_cache.ex
icehaunter: Addressed by making it non-blocking

KyleAMathews:
ChatGPT Pro's review https://chatgpt.com/share/68a4839c-5294-800c-9d71-b79d0e231ae3

robacourt:
benchmark this

PR #3008, opened 2025-08-19
feat(sync-service): Include span metrics even if the OTEL span has been not included because of sampling

robacourt:
#2916 introduced upfront sampling, however this also sampled the span metrics. This PR keeps span metrics even if the OTEL span has not been included because of sampling.

Keeping the metrics does slow down replication processing, replication throughput benchmarks show a 10% slow down from 27µs/txn to 30µs/txn because of this PR, however we may consider this acceptable for now given how useful the metrics are.

### DIFF
--- a/.changeset/wild-cameras-move.md
+++ b/.changeset/wild-cameras-move.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Include span metrics even if the OTEL span has been not included because of sampling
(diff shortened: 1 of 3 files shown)


icehaunter (approved):
I like it as-is, but shouldn't this respect your "sample metrics" flag?

robacourt:
> I like it as-is, but shouldn't this respect your "sample metrics" flag?

Oh, as in `with_child_span` should be checking if the child span is sampled as well as if it's in a span context. Good point! I will fix

balegas:
Thanks for fixing that!

Issue #2984, opened 2025-08-13
Skip integrity checks on clean shutdown


balegas:
Electric performs file-based integrity checks on all persistent shapes during startup, causing slow startup due to excessive syscalls.

We should track clean shutdowns to be able to skip file validation and just load the metadata from ETS in the expected case.

This will speedup startup time while keeping the same integrity guarantees.

elimelt:
+1

balegas:
If the shutdown was not clean, we should clear all shape data for safety. The follow up task is #2982 where we would only drop shape data for the shape logs that are dirty.

PR #3012, opened 2025-08-20
chore: Refactor `ShapeStatus` table to begin separately before `ShapeCache` and consumers

msfstef:
This work precedes the work I'm doing on https://github.com/electric-sql/electric/issues/2984

The `ShapeStatus` ETS table was being initialised by `ShapeCache`, but consumers also use it, and the consumer supervisor lives before the shape cache, causing issues in tests during shutdowns etc.

This changes _no_ behaviour in Electric (or at least it shouldn't!). I've refactored the tests and everything relevant to use shape cache in its new form.

I still don't love how it looks and I think it can be simplified further but this addresses the initial concern of moving it up front, and simplifies it as a mod arg passed at the stack supervisor level and then passed as is to appropriate modules, with a dedicated `ShapeStatusAgent` initializing and owning the table.

### DIFF
--- a/.changeset/smooth-ducks-eat.md
+++ b/.changeset/smooth-ducks-eat.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Move `ShapeStatus` ETS table to initialise before all consumers and `ShapeCache`.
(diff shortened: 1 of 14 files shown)


robacourt (approved):
Nice, that works!

PR #3014, opened 2025-08-20
Add vary header for examples w/ auth

KyleAMathews:
Follow up to #3013 — fix examples/starter to follow this pattern.


thruflo (approved):
Any of the example backends need to also return the right vary header?

KyleAMathews:
No the others don't have auth

PR #3016, opened 2025-08-21
fix: clarify wording, correct grammar, and improve formatting in Electric docs

dev-shahed:
fix: clarify wording, correct grammar, and improve formatting in Electric docs


icehaunter:
Thanks for the contribution, I think we're happy with the wording for now but we'll give it another pass

Issue #3020, opened 2025-08-21
Return consistent cache pointer (cursor) on 409 responses to avoid request cycles


msfstef:
**Bug description**
When we get back 409 responses without a new shape ID, the new `offset=-1` request is made without a cache buster that essentially leads to a cycle of requests up to the same 409 - we see this often in cached setups.

**Expected behavior**
The 409 responses should _always_ provide a way for the next `offset=-1` request to either reach the server or a cached response for the _new_ shape, not the invalidated one.

Some discussion have led us to believe that adding the existing `cursor` parameter to 409 response would allow us to have a consistent cache buster on 409s.

An alternative could be the 409 invalidation should let the client use some other consistent cache buster like the shape ID of the invalidated shape.

Also worth making 409s without a shape id/cache buster not be cached, such that these chains always end up reaching the server and get back a 409 with a shape id instead.

alco:
Related issue - https://github.com/electric-sql/electric/issues/2934

PR #3021, opened 2025-08-21
fix: Consistent cache busting on 409s

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3020

Before adapting the cursor for the purpose of 409s, I decided to use the handle itself which we were already using.

Here is my reasoninig:

1. Our main issue is that we are caching 409s (with or without a next handle) for 60 seconds - we are caching them as they are part of our regular request flow and it would create unnecessary load to break request coalescing and go to origin for them
  1a. One immediate fix is to change the caching for 409s _without_ a handle to 1 second, the minimum to enable request coalescing. This means that on the next request cycle they would most likely run into a 409 _with_ a shape handle anyway.
  1b. We could disable caching on 409s without a handle altogether - I _think_ that would would happen is that the 409 is return to one of the "held" requests, and the next one is picked as the "champion", and that one goes to the origin and it would _now_ receive a 409 _with_ a shape handle - so perhaps this is the better option.
2. I'm explicitly removing the shape handle as part of the `offset=-1` requests, so it really now only acts as a cache buster and it is not respected.
  2a. This allows us to "abuse" the `handle` parameter for `offset=-1` requests as a cache buster (which we were already sort of doing) - and rather than rely on a cursor we can make clients just request the same handle with a `-next` suffix, which is already consistent as per @samwillis ' suggestion and does not require coordination.

The reason I'm avoiding reusing the cursor is that it is used very explicitly in clients for _live_ requests, and ultimately it wouldn't give us a big benefit since all we really want is for clients to go on to the "next" chain of requests from the one they are on - whether that's via an explicit new handle that is given or via a consistent cache buster without coordination is irrelevant. I'd rather not add more complexity to the protocol.

## IMPORTANT

This change basically makes the `offset=-1` calls _ignore_ the `handle` parameter, so if you provide an invalid handle we will still match to an existing shape and return the valid one. If this is not how we would like the API to behave then I will scrap this change and use a different approach (with the cursor).

### DIFF
--- a/.changeset/purple-ligers-play.md
+++ b/.changeset/purple-ligers-play.md
@@ -1,0 +1,7 @@
+---
+"@electric-sql/client": patch
+"@core/elixir-client": patch
+"@core/sync-service": patch
+---
+
+Ensure 409s do not lead to infinite request cycles because of caching.
--- a/packages/elixir-client/lib/electric/client/stream.ex
+++ b/packages/elixir-client/lib/electric/client/stream.ex
@@ -189,11 +189,12 @@ defmodule Electric.Client.Stream do
   end
 
   # 409: Upon receiving a 409, we should start from scratch with the newly
-  #      provided shape handle
+  #      provided shape handle or with a fallback pseudo-handle to ensure
+  #      a consistent cache buster is used
   defp handle_response({:error, %Fetch.Response{status: status} = resp}, stream)
        when status in [409] do
     %{value_mapper_fun: value_mapper_fun} = stream
-    handle = shape_handle(resp)
+    handle = shape_handle(resp) || "#{stream.shape_handle}-next"
 
     stream
     |> reset(handle)
--- a/packages/sync-service/lib/electric/shapes/api/params.ex
+++ b/packages/sync-service/lib/electric/shapes/api/params.ex
@@ -135,7 +135,7 @@ defmodule Electric.Shapes.Api.Params do
     offset = fetch_change!(changeset, :offset)
 
     if offset == LogOffset.before_all() do
-      changeset
+      delete_change(changeset, :handle)
     else
       validate_required(changeset, [:handle], message: "can't be blank when offset != -1")
     end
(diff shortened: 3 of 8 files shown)


balegas:
Using a cache buster after receiving a 409 sounds good. I think we need to document that the cache-buster needs to be stable for cache-friendliness and we can leave to the implementer how to do it. I think that reusing the handler is a bit hacky and we could use a more explicit parameter since we ignore it anyways.

>  We could disable caching on 409s without a handle altogether - I think that would would happen is that the 409 is return to one of the "held" requests, and the next one is picked as the "champion", and that one goes to the origin and it would now receive a 409 with a shape handle - so perhaps this is the better option.

This seems more transparent and works with request coalescing. Is the next request guaranteed to receive the shape handle? In that case, I don't see any downside with this approach.

msfstef:
> I think that reusing the handler is a bit hacky and we could use a more explicit parameter since we ignore it anyways.

True - in this case I just wanted to avoid introducing new parameters to clients, and all the logic to add and remove the paramater appropriately (whereas the handle will get appropriately replaced once a new one is returned). While the handle as a cache buster is a bit of an abuse of the parameter, it does lend itself really well for this job.

> This seems more transparent and works with request coalescing. Is the next request guaranteed to receive the shape handle? In that case, I don't see any downside with this approach.

Yeah once the shape has been invalidated subsequent requests that reach the origin _will_ receive a 409 with the new shape handle - I just wasn't sure if request coalescing would work with a `no-cache` header for most CDNs. If you are confident it does then I'm happy to turn caching off for 409s without a shape handle.

balegas:
> Yeah once the shape has been invalidated subsequent requests that reach the origin will receive a 409 with the new shape handle - I just wasn't sure if request coalescing would work with a no-cache header for most CDNs. If you are confident it does then I'm happy to turn caching off for 409s without a shape handle.

You're right. I'm not sure. We need to check that with CF (and maybe sanity check with Varnish or Nginx?) because if it does, I think this is the preferred solution

balegas:
After discussion on standup today, we've decided to go with the cache buster using the shape-handler. We decided against changing the cache headers for 409s without an handler as this doesn't remove the need of using the cache buster.

@msfstef I'll reiterate that we need to document the need of using the cache buster after the 409 and its properties. We can give our implementation as an example. I'd like that to be reflected here https://electric-sql.com/openapi#/paths/~1v1~1shape/get as this is the reference (beside our own client) someone should use to implement a client for Electric.

PR #3024, opened 2025-08-22
Rename file->legacy_file and pure_file->file

alco:
Since the introduction of pure_file, our reference documentation for `ELECTRIC_STORAGE` has been wrong. This change aims to make it so the `PureFileStorage` implementation is given the weight it deserves while still making it possible to choose the legacy `FileStorage` implementation.

No documentation changes are needed since we don't want to advertise the use of the legacy file storage.


icehaunter:
Switching the default it one thing, but a breaking change switching the meaning of a feature flag feels excessive

alco:
> Switching the default it one thing, but a breaking change switching the meaning of a feature flag feels excessive

What is the meaning of that flag, though? The docs say that `ELECTRIC_STORAGE` is set to `FILE` by default and we know that means the `PureFileStorage` implementation is in use.

alco:
An alternative renaming I can suggest is as follows:
- `file` -> [`file`, `legacy_file`]. So do not change the original name but only ever talk about using `legacy_file` explcitly, in docs and in chat
- `pure_file` -> `optimized_file` or `fast_file`. Calling it "pure file" is meaningless to users, as they don't have the context to understand what "pure" means in the name.

icehaunter:
Sure, `fast_file` seems better

alco:
@icehaunter Updated. Please take another look.

msfstef (approved):
Looks good! Do we want to rename things internally as well or perhaps at another time (i.e. `PureFileStorage` to `FastFileStorage`)?

alco:
@msfstef I wouldn't want to rename the module in this PR since there's work happening in that module in other branches.

PR #3033, opened 2025-08-22
example: update tanstack-db-web-starter.

thruflo:
Minor README tweaks, updated the README live query example and bumped the dependency versions.


examples/tanstack-db-web-starter/src/routes/_authenticated/project/$projectId.tsx
```diff
@@ -25,8 +25,8 @@ function ProjectPage() {
   const [newTodoText, setNewTodoText] = useState("")
 
   const { data: todos } = useLiveQuery(
-    (q) =>
-      q
+    (query) =>
```
KyleAMathews: we use `q` everywhere else in the tanstack docs — why do this?

thruflo: It was just half and half in this repo. Can revert -- I was actually mainly trying to adjust to:

```ts
const { data: todos } = useLiveQuery(
    (query) => (
      query
        .from(...)
        .where(({ foo }) => eq(foo.bar, true))
    ),
    [dep]
)
```

From:

```ts
const { data: todos } = useLiveQuery(
    (query) =>
      query
        .from(...)
        .where(({ foo }) => eq(foo.bar, true)),
    [dep]
)
```

Because I find the trailing comma very hard to spot and easy to read as a comma separator between the clauses in the latter. I.e.: when you have a dependencies array, I much prefer the explicit outer `()`.

However then I ran `pnpm format` and it stripped out those changes and left just the arg rename :/

KyleAMathews: Oh haha. So that was what Claude did before 🤷‍♂️

I guess I don't really care

KyleAMathews (approved):

PR #3041, opened 2025-08-25
Name Support.Repo differently for different test modules to avoid races

alco:
Failure showcase:
* https://github.com/electric-sql/electric/actions/runs/17210761452/job/48821677710
* https://github.com/electric-sql/electric/actions/runs/17207937307/job/48812570629

### DIFF
--- a/packages/elixir-client/test/electric/client/ecto_adapter/enum_decoding_test.exs
+++ b/packages/elixir-client/test/electric/client/ecto_adapter/enum_decoding_test.exs
@@ -1,11 +1,11 @@
 defmodule Electric.Client.EctoAdapter.EnumDecodingTest do
   use ExUnit.Case, async: true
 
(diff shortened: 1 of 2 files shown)


msfstef (approved):
I'm surprised this works (is Support.Repo not shutting down properly/completely?) - have you observed a reduction in flakes somehow?

alco:
@msfstef Your intuition has proven to be correct: the problem stemmed from the fact that two different test modules were starting `Support.Repo` before each test. It was difficult to repro because `Electric.Client.EctoAdapter.EnumDecodingTest` only had 1 test. After I had disabled all other test modules and added a bunch of dummy test to this latter one, I could consistently reproduce the race.

I've redone the fix to use a differenet name for the support repo in each test module.

msfstef (approved):

PR #3042, opened 2025-08-25
feat: Faster recovery times with single point controlled shutdown recovery

msfstef:
Closes https://github.com/electric-sql/electric/issues/2984

We had observed that loading stored shapes took a long time, especially with larger number of shapes.

Doing some research most of the time was spent waiting on the various reads being done per-shape in order to load it up. @icehaunter is working on reducing those, but either way this would always scale with the number of shapes (at least 2-3 calls per shape). On slow FS like networked storage the time spent waiting on FS calls racks up pretty quickly.

After testing per-shape caching of metadata (reducing basically things to a single read per shape load), I concluded that to see a significant enough speedup we would need a global index that would avoid scaling FS operations with shapes recovered as much as possible.

Ultimately, the `ShapeStatus` table is started before everything else and it has now been assigned its own process as well, so it's the perfect place to do a controlled shutdown backup and restore.

## Backup procedure
- On `ShapeStatusOwner` shutdown, we dump the ETS table to a "backup" directory in our storage dir (with a 60 sec shutdown timeout, should be enough for this)
- On every shape consumer's controlled shutdown, we terminate the writer, flushing the state, closing file handles, and finalising metadata, and if the shape is still being tracked in `ShapeStatus`, we store the "storage recovery state" in the same ETS table - this will get dumped in the `ShapeStatusOwner` shutdown, as consumers all die before that process.
- We dump the ETS table with a basic validation of object count

## Restore procedure
- The `ShapeStatusOwner` attempts to restore the table on init from the backup
 - On top of ETS' own object count validation, I've added an additional integrity check that compares the shape handles found on the file system (a single `ls` call) with the ones found in the restored ETS table - if they do not match the backup is discarded.
 - If the backup restore fails for any reason, we fallback to our regular recovery mechanism with FS as the source of truth
 - On every backup restore (or attempted restore), we delete the backup to avoid any loops or stale state - since the backups are versioned we basically remove a shape status backup dir in its entirety.
- On every consumer startup, "consume" the shape storage state that was stored in ETS to avoid doing any FS operations to initialise the writer
 - No integrity checks are made - we just ensure the stored version and metadata structure matches, otherwise discard
 - We do not verify any files exist, directory structure exists, or anything else - it would defeat the purpose of a controlled restore. Failures will lead to a shape cleanup, nothing catastrophic.
 - I've added `pg_snapshot` in the cached metadata list to avoid all reads, even for the `get_current_position` one.
 - Shape restore has been reduced to just grabbing the file handles.

## Performance 
On my MacBook Air M2, 10k shapes recovery went from ~13sec to ~3sec. I haven't tested this on EFS but I suspect due to the significant reduction in FS/syscalls we're going to see a large speedup (as 10k shapes are taking minutes to load up currently).

## TODO
[] add more tests for covering FS edge cases of failed recovery

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/file_storage.ex
@@ -167,7 +167,7 @@ defmodule Electric.ShapeCache.FileStorage do
     case File.ls(shapes_dir) do
       {:ok, shape_handles} ->
         shape_handles
-        |> Enum.reject(&match?(@metadata_storage_dir, &1))
+        |> Enum.reject(&String.starts_with?(&1, "."))
         |> Enum.reject(&exists?(deletion_marker_path(shapes_dir, &1)))
         |> then(&{:ok, MapSet.new(&1)})
 
@@ -208,7 +208,7 @@ defmodule Electric.ShapeCache.FileStorage do
   end
 
   @impl Electric.ShapeCache.Storage
-  def metadata_backup_dir(%{base_path: base_path} = opts) do
+  def metadata_backup_dir(%{base_path: base_path}) do
     base_path |> Path.join(@metadata_storage_dir) |> Path.join("backups")
   end
 
--- a/packages/sync-service/lib/electric/shape_cache/file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/file_storage.ex
@@ -17,6 +17,7 @@ defmodule Electric.ShapeCache.FileStorage do
   @version_key :version
 
   @shape_definition_file_name "shape_defintion.json"
+  @metadata_storage_dir ".meta"
 
   @xmin_key :snapshot_xmin
   @pg_snapshot_key :pg_snapshot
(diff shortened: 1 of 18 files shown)


packages/sync-service/lib/electric/shape_cache/file_storage.ex
```diff
@@ -184,15 +202,16 @@ defmodule Electric.ShapeCache.FileStorage do
...
-        {:ok, %{}}
-
       {:error, reason} ->
         {:error, reason}
     end
   end
 
+  @impl Electric.ShapeCache.Storage
+  def metadata_backup_dir(%{base_path: base_path}) do
+    base_path |> Path.join(@metadata_storage_dir) |> Path.join("backups")
```
alco: Nitpick
```suggestion
    Path.join([base_path, @metadata_storage_dir, "backups"])
```

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -171,25 +175,44 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
         end)
         |> then(&{:ok, &1})
+
+      {:error, reason} ->
+        {:error, reason}
     end
   end
 
+  def metadata_backup_dir(%{base_path: base_path}) do
+    base_path |> Path.join(@metadata_storage_dir) |> Path.join("backups")
```
alco:
```suggestion
    Path.join([base_path, @metadata_storage_dir, "backups"])
```

icehaunter (approved):

PR #3045, opened 2025-08-26
fix(sync-service): Gracefully handle termination of un-ready stack

### LINKED ISSUE #3040: StatusMonitor shutdown
Uncaught exit - {:shutdown, {GenServer, :call, [{:via, Registry, {:"Electric.ProcessRegistry:8b4e58da-efdd-4716-b4bf-9b668079cc8a", {Electric.StatusMonitor, nil}}}, {:wait_until_active, 5000}, :infinity]}}

https://electricsql-04.sentry.io/issues/40928520/events/28bd261af2b9418a8b5e308fb012a99c/?query=

This shouldn't be caught at this level. You should handle the error in the right place

magnetised:
Fixes #3040

### DIFF
--- a/packages/sync-service/lib/electric/status_monitor.ex
+++ b/packages/sync-service/lib/electric/status_monitor.ex
@@ -110,7 +110,7 @@ defmodule Electric.StatusMonitor do
             "Stack ID not recognised: #{stack_id}"
           )
       catch
-        :exit, reason ->
+        :exit, _reason ->
           maybe_retry_wait_until_active(
             stack_id,
             timeout,


packages/sync-service/lib/electric/status_monitor.ex
```diff
@@ -102,6 +102,9 @@ defmodule Electric.StatusMonitor do
             timeout,
             "Status monitor not found for stack ID: #{stack_id}"
           )
+
+        :exit, reason ->
+          {:error, "Stack #{inspect(stack_id)} has terminated: #{inspect(reason)}"}
```
msfstef: why not retry on `:shutdown` the same way we do for `:noproc`? It's called `wait_until_active` so I would expect it to wait until the timeout is up, is there a reason we're considering this as a different case than the ETS table not being set up and the process not being alive?

(edit: still happy to keep this fallback in here as well for other exits)

magnetised: My thinking was that if the stack has been terminated then that's pretty, well, terminal. The other cases handle transient timing issues/race conditions whereas the stack being terminated is not transient. I guess there could be situations where the status monitor pid crashes, and will be restarted so this would be transient but it feels very unlikely. The StackSupervisor's restart strategy is `one_for_one` which means that the status monitor will only exit on an error in it's process or the stack being shutdown.

msfstef: That's why I was thinking that I'd rather wait for the `StatusMonitor` to get back up so we can get a proper response for the state of the stack - without it then we just have a generic "stack is down" without a more accurate state.

magnetised:
> but not for a shutdown (which implicitly would lead to a restart)

a shutdown doesn't implicitly lead to a restart... if the process is exiting with reason `:shutdown` then it's being taken down by the supervisor, not crashing.

I see your point though and will retry for all `:exit`s

msfstef:
> a shutdown doesn't implicitly lead to a restart...

Agreed, but the same holds for `:noproc` and no ETS table - for the sake of consistency though I think it'd make sense to use a single strategy. The `maybe_retry` blah blah already takes care of timing things out so it's safe to use for all.

packages/sync-service/lib/electric/status_monitor.ex
```diff
@@ -95,16 +110,12 @@ defmodule Electric.StatusMonitor do
...
           )
       catch
-        :exit, {:noproc, _} ->
-          # This happens when the StatusMonitor process has not been started yet
+        :exit, reason ->
           maybe_retry_wait_until_active(
             stack_id,
             timeout,
-            "Status monitor not found for stack ID: #{stack_id}"
+            "Stack #{inspect(stack_id)} has terminated with reason: #{inspect(reason)}"
```
msfstef: nit: we are exposing this message to clients so I think we should probably avoid printing the entire error

PR #3050, opened 2025-08-26
fix: flush tracker was using wrong shape identity

icehaunter:


icehaunter:
@msfstef I've added a unit test that's using the public APIs so it should be nice & stable

PR #3052, opened 2025-08-26
Msfstef/fix erlang errors from connections

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3049

The error that is hidden is a `:noproc` `ErlangError` when trying to do `Process.link` from the connection being set up to a supervisor that is no longer alive.

The underlying DBConnection application has various processes living in its own extra application handling connection lifecycles with some levels of indirection so we cannot rely on the pool being terminated to ensure there's no new connections popping up when the supervisor is dead.

Instead, we check if the supervisor is alive at the time of the connection being configured, and if it is not we exit.

I've reproduced the error manually by randomly killing off the supervisor with timeouts while the pool connections are setting up and was consistently reproducing this - the suggested fix does the trick. I was also thinking of rescuing the `ErlangError` but the `alive?` check feels more explicit (even if it technically has a gap between the check and the link op?). I don't _love_ it but it does take care of the issue and I can't find a better approach currently.

While running tests I also noticed some errors that I figured were because of an incorrect order of processes in our replication supervisor. The consumer supervisor should be started after the shape status table, the shape log collector, and the publication manager, as consumers make use of all three during their lifetime, so during controlled shutdowns we want this order to be maintained.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager/pool.ex
+++ b/packages/sync-service/lib/electric/connection/manager/pool.ex
@@ -243,7 +243,12 @@ defmodule Electric.Connection.Manager.Pool do
   # and log any issues before and after the connection is established.
   def configure_pool_conn(opts, supervisor_pid) do
     send(supervisor_pid, {:pool_conn_started, self()})
-    Process.link(supervisor_pid)
+
+    # If supervisor process not alive when initializing connection, abort it
+    if Process.alive?(supervisor_pid),
+      do: Process.link(supervisor_pid),
+      else: exit({:shutdown, {:supervisor_not_alive, supervisor_pid}})
+
     opts
   end
 
--- a/.changeset/rotten-birds-argue.md
+++ b/.changeset/rotten-birds-argue.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Prevent pool connections from starting if supervisor is found to already be dead.
--- a/.changeset/tame-donkeys-approve.md
+++ b/.changeset/tame-donkeys-approve.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Change order of replication supervisor processes to ensure consumers can perform all operations.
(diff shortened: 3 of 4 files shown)


packages/sync-service/lib/electric/connection/manager/pool.ex
```diff
@@ -243,7 +243,12 @@ defmodule Electric.Connection.Manager.Pool do
   # and log any issues before and after the connection is established.
   def configure_pool_conn(opts, supervisor_pid) do
     send(supervisor_pid, {:pool_conn_started, self()})
-    Process.link(supervisor_pid)
+
+    # If supervisor process not alive when initializing connection, abort it
+    if Process.alive?(supervisor_pid),
+      do: Process.link(supervisor_pid),
+      else: exit({:shutdown, {:supervisor_not_alive, supervisor_pid}})
```
alco: We can avoid the "gap" between the `alive?()` and `link()` calls by catching the exit after the fact:

```suggestion
    try do
      Process.link(supervisor_pid)
    catch
      :error, :noproc ->
        exit({:shutdown, {:supervisor_not_alive, supervisor_pid}})
    end
```

alco (approved):
:+1:

packages/sync-service/lib/electric/connection/manager/pool.ex
msfstef: That approach was my initial one (although what you've written doesn't quite work):
```elixir
try do
  Process.link(supervisor_pid)
rescue
  ErlangError ->
    exit({:shutdown, {:supervisor_not_alive, supervisor_pid}})
end
```

But I felt that it was slightly less explicit. It _is_ more solid as there's no "gap" though so maybe I keep it like this

alco: Why doesn't my approach work? I've tested it in IEx with a simple process. It does make it more explicit about which error we're catching as opposed to the catch-all `rescue ErlangError`.

msfstef: Your approach absolutely works and I must've mistyped it when testing it! Tried it myself in IEx and then rewrote it and now it works haha - my bad!

PR #3057, opened 2025-08-27
fix(sync-service): Monitor child processes synchronously within the Connection.Manager

### LINKED ISSUE #3018: FunctionClauseError: no function clause matching in Electric.Connection.Manager.handle_cast/2
Sentry Issue: [ELECTRIC-4R5](https://electricsql-04.sentry.io/issues/58716006/?referrer=github_integration)

This is a bug. Needs to be fixed even without a refactoring of the connection manager.

```
FunctionClauseError: no function clause matching in Electric.Connection.Manager.handle_cast/2
  File "lib/electric/connection/manager.ex", line 700, in Electric.Connection.Manager.handle_cast/2
  File "gen_server.erl", line 2460, in :gen_server.try_handle_cast/3
  File "gen_server.erl", line 2418, in :gen_server.handle_msg/3
  File "proc_lib.erl", line 333, in :proc_lib.init_p_do_apply/3
```

magnetised:
Some kind of race condition is the probable explanation for #3018 - somehow the ShapeCache has been restarted without the Connection.Manager process knowing about it, and so its state goes out of sync with the actual process tree.

Fixes #3018

### DIFF
--- a/.changeset/clever-maps-chew.md
+++ b/.changeset/clever-maps-chew.md
@@ -2,4 +2,4 @@
 "@core/sync-service": patch
 ---
 
-Fix potential race conditions when monitoring the log collector process
+Fix potential race conditions when monitoring the log collector process and improve connection manager's handling of shape process crashes.


alco:
There's definitely a race but this doesn't fix it.

First, I really dislike the `receive` call inside a `GenServer` callback. It means the message will be processed out of order if there are already messages from other processes in Connection.Manager's inbox. It makes for a more confusing control flow and can introduce other races.

I've manually tested this change and encountered the following failure mode:

```
13:50:41.046 pid=<0.603.0> [error] GenServer {:"Electric.ProcessRegistry:single_stack", {Electric.Connection.Manager, nil}} terminating
** (FunctionClauseError) no function clause matching in Electric.Connection.Manager.handle_info/2
    (electric 1.1.6) lib/electric/connection/manager.ex:515: Electric.Connection.Manager.handle_info({:process_monitored, Electric.Replication.ShapeLogCollector, #PID<0.640.0>, #Reference<0.2928434455.3442475009.72659>}, %Electric.Connection.Manager.State{...}
```

So having `receive` called synchronously after starting the replication supervisor doesn't solve the issue.

---

I can reproduce the original issue by performing these steps:
- create a table named `items` in Postgres
- start Electric
- drop `electric_publication_default` in Postgres
- request a shape for `items`

Running this on `main` results in a Connection.Manager crash with reason:
```
13:50:12.710 pid=<0.348.0> [error] GenServer {:"Electric.ProcessRegistry:single_stack", {Electric.Connection.Manager, nil}} terminating
** (FunctionClauseError) no function clause matching in Electric.Connection.Manager.handle_cast/2
    (electric 1.1.6) lib/electric/connection/manager.ex:713: Electric.Connection.Manager.handle_cast({:consumers_ready, 0, 0}
```

By this point, Connection.Manager has completed its setup and is in its steady state, so the problem is not asynchrnous monitoring of the shape log collector during initialization but the fact that the shape cache gets restarted at all. When the shape cache process restarts, Connection.Manager can be in any phase:
- `:connection_setup` if it hasn't started streaming yet
- `:restarting_replication_client` if the replication client died before the `:consumers_ready` message from shape cache got sent
- `:running` if its in its steady state when an external reason caused the shape cache to restart, as I demonstrated above

magnetised:
> First, I really dislike the receive call inside a GenServer callback. It means the message will be processed out of order if there are already messages from other processes in Connection.Manager's inbox. It makes for a more confusing control flow and can introduce other races.

yup. gotcha. there's a requirement to make sure the manager has the log collector's pid before it can receive the down from it exiting - hence pushing that receive up the queue. but dammit yes, I missed the restart condition and the accompanying message. I'll re-think.

> By this point, Connection.Manager has completed its setup and is in its steady state, so the problem is not asynchrnous monitoring of the shape log collector during initialization but the fact that the shape cache gets restarted at all. When the shape cache process restarts, Connection.Manager can be in any phase

So, with your vastly greater understanding of the connection manager, what should happen after the shape cache restart when in this steady state? Is it just a no-op since the purpose of the original message is to kick the manager onto a step we've already taken?

msfstef:
> When the shape cache process restarts, Connection.Manager can be in any phase:

My understanding was that if shape cache dies for whatever reason, the `Replication.Supervisor` would take care to tear down everything else including the shape log collector as well before spinning it back up - which would fire a `:DOWN` message that would bring the ConnMan to a specific state. Am I missing something?

alco:
> what should happen after the shape cache restart when in this steady state? Is it just a no-op since the purpose of the original message is to kick the manager onto a step we've already taken?

As I understand it, what happens in my reproduction example is that all children of `Electric.Replication.Supervisor` are terminated in response to the publication manager shutdown. Connection.Manager receives a DOWN message about the shape log collector and restarts the replication client. The difference between restarting the replication client from the original initialization sequence that Connection.Manager went through at startup is that this time the shape cache can send the `consumers_ready` message at any point during replication client initialization.

The `consumers_ready` step in Connection.Manager was added in https://github.com/electric-sql/electric/pull/2840. By design, it was supposed to be sent by the shape cache after the replication client has finished initializing and before it started streaming. But in the above case this no longer holds because it's no longer Connection.Manager starting the shape cache.

I would say ignoring this message in any phase other than `connection_setup` and in any step other than `{:waiting_for_consumers, _}` is fine. Better than complicating this state machine further by trying to predict all the steps at which this message may arrive or by adding more coordination between Connection.Manager and the shape cache.

alco:
> > When the shape cache process restarts, Connection.Manager can be in any phase:
> 
> My understanding was that if shape cache dies for whatever reason, the `Replication.Supervisor` would take care to tear down everything else including the shape log collector as well before spinning it back up - which would fire a `:DOWN` message that would bring the ConnMan to a specific state. Am I missing something?

I was highlighting the fact that the space of possible states is larger than what we would like to focus on when thinking about normal operation of the system.

Is there anything preventing the shape cache from dying after Connection.Manager has started it (via the Replication.Supervisor) but before it has transitioned from the `:connection_setup` phase into the `:running` phase?

The `handle_info()` handler in Connection.Manager for `:DOWN` message from the shape log collector doesn't make any state transitions.

magnetised:
@alco have updated after your feedback. Please have a look.

Incidentally I followed your reproduction steps:

>     create a table named items in Postgres
>    start Electric
>    drop electric_publication_default in Postgres
>    request a shape for items

and electric never recovered:

```
Stack not ready after 5000ms. Reason: Timeout waiting for replication client to be ready. Check that you don't ha
ve pending transactions in the database. Electric has to wait for all pending transactions to commit or rollback before it can create the replication slot.
```

On main the crash in the connection.manager actually helps to recover the system... Maybe we should just restart when we recieve that consumers_ready message. Going to poke at things to see if I can not crash and recover the replication stream

magnetised:
So, there was a problem with the handling of the log collector exit message that was being obscured by the crash. I've put in a fix, which does result in a resumption of streaming after the deletion of the publication.

We had a catch all no-op handler for all `:EXIT` messages with reason `:shutdown` but in the case of say, the publication manager crashing, we would receive a `:DOWN` message for the shape log handler with reason `:shutdown` because the log collector was being restarted by the supervisor because of the `one_for_all` strategy. 

We pass this reason onto the replication client in the stop/2 call and thus receive the exit signal with reason `shutdown` ... which we were ignoring and so the replication client was not being restarted.

I've added an explicit handler for the replication client exiting with reason `shutdown` and using it to trigger a restart of the replication client. 

I may be botching things... so please let me know.

This does not solve all issues -- for example with the reproduction steps above, there are race conditions where the publication manager is restarted before the replication client or the re-creation of the slot triggers a purge_all_shapes call to the shape cache before it's restarted/is running.

Basically its messy but it gets there. Think fixing that is out of scope of this pr though.

alco:
@magnetised This looks great.

Yeah, we won't be able to make it race free right here and right now. Connection.Manager's role and the relationship between Connection.Supervisor processes and Replication.Supervisor processes need to be rethought and refactored. Out of scope for this PR for sure.

alco:
I think we can keep things simpler but replacing the special-casing of the way `:shutdown` exit is handled for the replication client by just explicitly providing a different reason when Connection.Manager stops the client. I've pushed this change to your branch.

packages/sync-service/lib/electric/connection/manager/process_monitor.ex
```diff
@@ -0,0 +1,72 @@
+defmodule Electric.Connection.Manager.ProcessMonitor do
+  # Exists to allow child processes to synchronously register themselves and be
```
msfstef: nit: turn it into a moduledoc?

packages/sync-service/lib/electric/connection/supervisor.ex
```diff
@@ -58,7 +58,8 @@ defmodule Electric.Connection.Supervisor do
 
     children = [
       {Electric.StatusMonitor, opts[:stack_id]},
-      {Electric.Connection.Manager, opts}
+      {Electric.Connection.Manager, opts},
+      {Electric.Connection.Manager.ProcessMonitor, opts[:stack_id]}
```
msfstef: I understand that `ProcessMonitor` expects to find ConnMan, and that's why it's the last child, but it only expects to find it in response to monitor calls (which come from conn man)

Would it not make more sense to have it start _before_ conn man? And ensure this "proxied" monitoring can always occur during conn man's initialization (even beyond the initial init call)

With a `rest_for_one` strategy I feel like `ProcessMonitor` could die and restart, without conn man restarting, and now connman thinks it is monitoring processes but it actually isn't because `ProcessMonitor` "lost" the monitoring during the restart.

packages/sync-service/lib/electric/connection/manager/process_monitor.ex
```diff
@@ -0,0 +1,72 @@
...
+
+  @impl GenServer
+  def init(stack_id) do
+    Process.set_label({:connection_manager_process_monitor, stack_id})
+
+    {:ok, %{stack_id: stack_id}}
+  end
+
+  @impl GenServer
+  def handle_call({:monitor, module, pid}, _from, state) do
```
msfstef, line 45: would it be possible to set up a basic test for this module?

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -894,6 +904,11 @@ defmodule Electric.Connection.Manager do
     {:noreply, state, {:continue, :start_streaming}}
   end
 
+  def handle_cast({:consumers_ready, _recovered, _failed} = msg, state) do
+    Logger.debug("received #{inspect(msg)} in phase #{state.current_phase}: ignoring")
```
msfstef:
```suggestion
    Logger.debug("Received #{inspect(msg)} in phase #{state.current_phase}: ignoring")
```
supernit: all our debug logs are sentence cased

packages/sync-service/lib/electric/db_connection_error.ex
```diff
@@ -326,6 +326,15 @@ defmodule Electric.DbConnectionError do
     end
   end
 
+  def from_error(:shape_log_collector_down) do
+    %DbConnectionError{
+      message: "Linked ShapeLogCollector process terminated",
```
msfstef:
```suggestion
      message: "Internal stream processing encountered an error",
```
nit: since these errors are exposed as stack stack events as well, I'd prefer if the message was more descriptive of what has occurred in a more general sense rather than being tied to the internals in names and terminology. Suggestion is just to highlight what I mean.

packages/sync-service/lib/electric/connection/manager.ex
magnetised: I always think that debug logs should be lower case -- like caps are too important for them :)

packages/sync-service/lib/electric/connection/supervisor.ex
magnetised: I'll link them both together because we're both right...

magnetised:
> I think we can keep things simpler but replacing the special-casing of the way `:shutdown` exit is handled for the replication client by just explicitly providing a different reason when Connection.Manager stops the client. I've pushed this change to your branch.

Agreed, thanks

msfstef (approved):
Nice work, hopefully this works better than before🤞

PR #3058, opened 2025-08-27
fix: ensure correct processing of skipped txns for flushes

icehaunter:


packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
```diff
@@ -146,6 +146,13 @@ defmodule Electric.Replication.ShapeLogCollector.FlushTracker do
            |> add_to_tree(last_flushed_offset, shape_id)}
       end
 
+    min_incomlete_flush_tree =
```
alco:
```suggestion
    min_incomplete_flush_tree =
```

msfstef (approved):

Issue #3063, opened 2025-08-28
Turn request/response logs (e.g. "sent 200") to debug level


msfstef:
Our current request logs are way too verbose at the info level - we should lower some of them to debug level, we already have sampled telemetry of all requests available outside of logs.

PR #3064, opened 2025-08-28
chore: Lower log level of request handler loggers to `debug`

msfstef:
Closes https://github.com/electric-sql/electric/issues/3063

- Lowers `Plug.Logger`'s log level to `debug`
- Turns the query string logging to `debug`
- To offset the lack of request logs at `info` log, I've added an additional log line for new shapes being created, so that info logs have more continuity

This has been discussed a few times and I have the feeling that there is agreement on this, but if anyone would like to weigh and argue for keeping request logs in `info` please do so.

Having an HTTP service without request logs at `info` level does feel odd so I'm not feeling 100% sure on this one.


alco (approved):

balegas:
Sounds good to me... too noisy.

balegas (approved):

PR #3065, opened 2025-08-28
fix: Remove shape log collector monitoring altogether

msfstef:
Followup to https://github.com/electric-sql/electric/pull/3057

Simplifies connection manager as we do not need to tie ShapeLogCollector's lifetime to the replication client - the replication client will crash in absence of the shape log collector either way (following @alco 's advice).

I've linked and added a special handling for the shapes supervisor exits - if the supervisor itself exits (which it always down with `:shutdown` which we were previously ignoring), we need to shutdown the connection manager. This would only happen in cases of infinite restarts etc but good to not miss this IMO.

Because the shapes supervisor is started as a temporary child to the connection manager supervisor, if it itself dies the connection manager wouldn't know (other than by crashing by virtue of the shape log collector missing, in case we try to process transactions).


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -584,6 +579,13 @@ defmodule Electric.Connection.Manager do
       ),
       do: {:noreply, state}
 
+  # Special-case any exit of the shapes supervisor - if it happens we need to restart
+  # the whole connection manager, as it means it has given up restarting its own
+  # children for whatever reason.
+  def handle_info({:EXIT, pid, reason}, %{shapes_supervisor_pid: pid} = state) do
+    {:stop, {:shutdown, reason}, state}
```
magnetised: If the reason is `:shutdown` then the exit reason of this process will be `{:shutdown, :shutdown}` - think it's worth propagating the reason. 

Plus, if the shapes supervisor has crashed because of an error, is it not worth crashing this process (using `{:stop, error, state}` to see if the resulting restart can fix the situation? The closer we stick to the behaviour of a real supervisor the better IMHO.

magnetised (commented):
Do we try to restart on a crash of the shapes supervisor, or just stop? 

Otherwise looks really good.

packages/sync-service/lib/electric/connection/manager.ex
msfstef: I agree, I guess I'm following the pattern on line 658 where we propagate it via `{:shutdown, reason}` - that would still trigger a restart for the supervisor. But that clause already assumes the pure `:shutdown` has been "filtered" out, so I can special case it.

In reality since the replication supervisor is a supervisor, it's only really going to exit with `:shutdown`, so not much to propagate.

msfstef: Ah I just re-read the code and we actually start the replication supervisor under the same supervisor as the connection manager - so this whole clause I've added is irrelevant, the process is not linked and we'd have to monitor it.

Good thing I decided to add a test.

magnetised: Right, so the Connection.Supervisor will restart it unless its shutting down.

msfstef: It _should_ but we set its restart policy to `:temporary` - so no

msfstef: The way I see it there are two options:

1. We monitor the shapes supervisor in the connection manager in order to shut it down if it dies (it's a dynamic child with rest_for_one policy, I'd rather not let the supervisor restart it by setting its policy to `:transient` or something)
2. Do not start it under the connection supervisor, instead start it as a linked process to the connection manager like we do for the rest of the processes we start.

msfstef: I've opted to keep the supervisor structure, but I'm now linking the shapes supervisor explicitly in the connection manager to capture the case of it dying. Added a failing test case for it as well. @alco opinions?

alco:
First of all, it feels good to get rid of the monitoring code and get closer back to the realm of supervision trees. But now we're still preventing ourselves from fully relying on supervision by linking the replication supervisor to connection manager and manually handling its EXIT signal.

Let's try to reason about the requirements in supervision terms: we want to keep the current configuration for `Electric.Connection.Supervisor` where `Connection.Manager` can be restarted for many reasons but `StatusMonitor` keeps chugging along. And if it's the `StatusMonitor` process that terminates, we also want to bring `Connection.Manager` down. This is already achieved with the `rest_for_one` supervision strategy.

Now, at some point in `Connection.Manager`'s initialization flow we want to start `Electric.Replication.Supervisor` and tie its lifetime to `Connection.Manager`'s lifetime, such that terminating either one of the two will cause the other one to terminate as well.

This can be modelled by a multi-level supervision approach (pseudo-code for illustrative purposes; I think every supervisor should be defined in its own module for clarity):

```elixir
    children = [
      {Electric.StatusMonitor, opts[:stack_id]},
      %{
        id: Electric.ConnectionManagerAndReplicationSupervisor,
        start: {
          Supervisor,
          :start_link,
          [
            [{Electric.Connection.Manager, opts}], [strategy: :one_for_all]
          ]
        },
        type: :supervisor,
        name: {Electric.ConnectionManagerAndReplicationSupervisor, stack_id}
      }
    ]
```

With this in place, when `Connection.Manager` is ready to start `Electric.Replication.Supervisor`, it starts it under this `Electric.ConnectionManagerAndReplicationSupervisor`.

The code may need to be tweaked but the general principle is this: remove the explicit link and rely on supervision tree modelling to achieve the desired shutdown/restart behaviour between multiple processes.

msfstef:
@alco I completely agree - but I figured this will be part of the work on revisiting the supervision strategy. This PR was mostly to fix the issue that is currently merged in main (with an error logged on every controlled shutdown because of shape log collector) + a tweak to at least never run into a situation where something is not running and never restarting

If you can either tweak this PR yourself with the proposed change, or remove the shape supervisor linking altogether since we weren't doing that in the first place, or merge it in its current state and I can start the work on more proper supervision next week I would prefer that, but I'm eager to get this thing in before any release or Cloud update is made.

alco:
@msfstef Sounds good. I have pushed a set of changes that adds an intermediate supervisor to the picture, ensuring that both Connection.Manager and Replication.Supervisor terminate when either one of them dies. Then Connection.Supervisor will restart Connection.Manager and the system recovers on from there.

alco (approved):
Self-congratulatory approval.

PR #3068, opened 2025-08-29
feat: support non-PK selects in subqueries

### LINKED ISSUE #3037: Subqueries on non-pk columns

icehaunter:
Closes #3037

Keeps around 2 structures - a reference counter for the values, and the key => value mapping, because on our replica modes we might not see old values that need to be counted down, especially on deletes.

Allows following one-to-many relations:
```sql
CREATE TABLE comments (..., task_id INT NOT NULL, ...);
CREATE TABLE tasks (id INT PRIMARY KEY, ...);
```

You can define a shape with `where: "id IN (SELECT task_id FROM comments WHERE ...)"` - or other applications of course.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
@@ -180,6 +180,7 @@ defmodule Electric.Shapes.Consumer.Materializer do
       Enum.reduce(changes, {state.index, {state.value_counts, []}}, fn
         %Changes.NewRecord{key: key, record: record}, {index, counts_and_events} ->
           value = cast!(record, state)
+          if is_map_key(index, key), do: raise("Key #{key} already exists")
           index = Map.put(index, key, value)
 
           {index, increment_value(counts_and_events, value)}
--- a/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
@@ -56,9 +56,10 @@ defmodule Electric.Shapes.Consumer.Materializer do
     end)
   end
 
-  def subscribe(stack_id, shape_handle) do
-    GenServer.call(name(stack_id, shape_handle), :subscribe)
-  end
+  def subscribe(opts), do: GenServer.call(name(opts), :subscribe)
+
+  def subscribe(stack_id, shape_handle),
+    do: subscribe(%{stack_id: stack_id, shape_handle: shape_handle})
 
   def start_link(opts) do
     GenServer.start_link(__MODULE__, opts, name: name(opts))
@@ -73,7 +74,8 @@ defmodule Electric.Shapes.Consumer.Materializer do
 
     state =
       Map.merge(opts, %{
-        view: %{},
+        index: %{},
+        value_counts: %{},
         offset: LogOffset.before_all(),
         ref: nil,
         subscribers: MapSet.new()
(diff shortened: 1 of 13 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -694,11 +694,19 @@ defmodule Electric.Connection.Manager do
...
     # try-catch block.
 
+    reason =
+      case reason do
+        :normal -> :normal
+        :shutdown -> :shutdown
+        {:shutdown, term} -> {:shutdown, term}
+        _ -> :shape_log_collector_down
+      end
+
```
robacourt: how is this change related to this PR?

packages/sync-service/test/electric/shapes/consumer/materializer_test.exs
```diff
@@ -64,6 +64,43 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
     assert Materializer.get_link_values(ctx) == MapSet.new([1, 2, 3])
   end
 
+  test "materializer correctly maps non-pk changes", %{shape_handle: shape_handle} = ctx do
```
robacourt: Please could we add some tests that show that if we get unexpected changes (e.g. a delete for a PK that does not exist or an insert for a PK that already exists) that we throw an error as opposed to making the ref counts inconsistent.

packages/sync-service/test/electric/shapes/consumer/materializer_test.exs
```diff
@@ -64,6 +64,43 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
...
+
+    Materializer.new_changes(ctx, [
+      %Changes.UpdatedRecord{key: "2", record: %{"value" => "3"}, old_record: %{"value" => "2"}},
+      %Changes.DeletedRecord{key: "3", old_record: %{"value" => "1"}},
+      %Changes.UpdatedRecord{key: "1", record: %{"other" => "1"}, old_record: %{"other" => "0"}}
+    ])
+
+    assert Materializer.get_link_values(ctx) == MapSet.new([1, 3])
+
+    assert_receive {:materializer_changes, ^shape_handle, move_out: 2, move_in: 3}
```
robacourt: It would be nice to break these into smaller tests that test the different scenarios, with a nice test name that describes the scenario, e.g. insert causes a move in, update causes move in and move out, etc.

Also I'm wondering if there's scenarios not tested that could be additional tests:
- insert for existing value does not cause a move in
- update causing a move in and out when replica mode = default (so no old value available)

robacourt (changes requested):
Lovely! I've just left a few nits

packages/sync-service/lib/electric/connection/manager.ex
icehaunter: In no way, but logs are very loud in tests without this

msfstef: this whole handle info clause no longer exists in conn man anyway

icehaunter: I'll rebase :D

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -193,4 +174,56 @@ defmodule Electric.Shapes.Consumer.Materializer do
...
+    end
+  end
+
+  defp decrement_value({value_counts, events}, value) do
+    # If we're decrementing, it must have been added before
+    case Map.fetch!(value_counts, value) do
+      1 ->
+        {Map.delete(value_counts, value), [{:move_out, value} | events]}
+
+      count ->
```
msfstef, line 226: should never happen (TM), but should we explicitly handle the case of `:error`, even if it's just raising? I don't like the idea on relying on the bad arithmetic to throw for that case

icehaunter: No, we discussed with Rob and this is "just raise" because it should indeed never happen and we should investigate if it does

msfstef: Agree - but to clarify we expect a `bad argument in arithmetic expression` for this case? I was thinking we capture `:error` and raise a runtime error

msfstef: ah sorry nevermind I just noticed we're doing `fetch!` - ignore me

robacourt (approved):
Excellent!

whollacsek:
Hello, is this deployed in the cloud? I'm getting "Subqueries are not supported" error:

```js
{"message":"Invalid request","errors":{"where":["Subqueries are not supported"]}}], "isLoading": true}
```

KyleAMathews:
@whollacsek it's behind a feature flag still on cloud. We're hoping to lift that soon. Reach out about your use case and we'll see if we can add you earlier.

whollacsek:
Thanks for the reply. I was hoping to use sub-query to reimplement my RLS.
I have a series of tables that allows me to implement flexible ABAC

PR #3072, opened 2025-08-29
feat: support 3+ layers of subqueries

### LINKED ISSUE #3036: Subqueries should work across 3+ levels

icehaunter:
Closes #3036

### DIFF
--- a/.changeset/spotty-wombats-relax.md
+++ b/.changeset/spotty-wombats-relax.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+feat: support 3+ layers of subqueries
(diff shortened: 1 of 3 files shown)


robacourt (approved):
Nice!

icehaunter:
@msfstef wdym "only thing that was needed" - for 3 layers to work? If so, yes.

msfstef:
@icehaunter was it not working before because terminations due to materializer changes were not notifying the entire dependency chain or something? just looking for some more context from the change to the 3+ layers support unblock

icehaunter:
@msfstef yup exactly. All the logic was in place, but propagation wasn't hooked up correctly

PR #3073, opened 2025-08-29
feat: close files on inactive shapes and don't open all on startup

### LINKED ISSUE #3039: Close open files on unused shapes
Currently, we're holding open a write handle per shape regardless of action and even of going into hibernation. We should instead close writer file handles while the shape isn't actively written to

icehaunter:
Closes #3039. I'm planning to make a follow-up PR that's going to also lower the amount of open file handles per active shape. Also our hibernate timer is 30 seconds - we may want to actually raise that to like 1-2 minutes?

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
@@ -84,6 +84,7 @@ defmodule Electric.ShapeCache.PureFileStorage do
     :write_timer,
     :open_files,
     :ets,
+    :latest_name,
     :opts
   ]
 
@@ -414,7 +415,8 @@ defmodule Electric.ShapeCache.PureFileStorage do
     end)
 
     writer_state(state,
-      open_files: open_files(opts, new_latest_suffix),
+      open_files: nil,
+      latest_name: new_latest_suffix,
       writer_acc: adjust_write_positions(writer_acc, -log_file_pos, -key_file_pos)
     )
   end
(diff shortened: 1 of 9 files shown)


packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -892,6 +910,9 @@ defmodule Electric.ShapeCache.PureFileStorage do
     {latest_name, {compaction_boundary, compacted_name}, {cached_min, chunks}} =
       boundary_info || get_read_source_info(opts)
 
+    # Any virtual offsets are handled elsewhere - normalize them to the last before real offsets for main log
+    offset = if is_virtual_offset(offset), do: LogOffset.last_before_real_offsets(), else: offset
```
msfstef, line 914: this confuses me slightly - if virtual offsets are handled elsewhere, and this is a private function, how do we ever end up handling a virtual offset here, and how do we know its safe to cast it into a last before real offsets? Should we not just add a guard to the function and let it fail if we don't expect it to ever get here? (or fix whatever code path leads here with a virtual offset?)

msfstef (approved):
Very nice! Will hopefully help a lot with the file handle usage

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: It was normalized to 0,inf in one call path but not the other. I prefer it normalized here in one place

Issue #3011, opened 2025-08-20
Investigate test flakes related to shape cleanups


msfstef:
Some Electric sync service tests seem to be flaking on CI and _very often_ the failure is a timeout on waiting for a shape cleanup assertions from the monitor process.

Some examples of tests flaking:
- `Elixir.Electric.Shapes.ConsumerTest::test transaction handling with real storage restarting a consumer doesn't lower the last known offset when only snapshot is present`
- `Elixir.Electric.ShapeCacheTest::test after restart deletes shapes that fail to initialise within a timeout`
- `Elixir.Electric.ShapeCacheTest::test get_or_create_shape_handle/2 shape initialization expires shapes if shape count has gone over max_shapes`

It's worth investigating if there are cases where the shape doesn't get cleaned up at all, or if it is purely a timing issue.

If it is the former, then there might be a serious underlying bug in the cleanup logic that needs to be addressed.

If it is the latter, we need to determine an upper bound for the cleanup to occur and use appropriate timeouts everywhere - and if possible configure the test setup to have a lower upper bound (e.g. if there's a debounce or throttle for cleaning up the shapes or something).

PR #3085, opened 2025-09-01
fix: Eliminate monitor cleanup assertion flakes

msfstef:
Closes https://github.com/electric-sql/electric/issues/3011

The async shape cleanup procedure is set to have a timeout of 60s, so as to not hang on FS being slow for _too_ long. However this upper bound is too high for tests, and at least on CI it seems that occasionally it might get into that range (when running everything concurrently).

It's still worth investigating how it gets bottlenecked this bad, but reducing the timeout does at least ensure that the tests are asserting the semantics that cleanup is being called, and they don't flake because of timeouts due to slow FS.

I was thinking of passing the timeout as a function argument but given the module is already using `Mix.env()` to configure the code this approach felt like the simplest, most self-contained one.

P.S. also made the slot names unique per test rather than uniform random as some test failures seem to be related to occasional collisions of slot names

With these changes all known (at least to me and codecov) flakes should be covered.

### DIFF
--- a/packages/sync-service/test/support/component_setup.ex
+++ b/packages/sync-service/test/support/component_setup.ex
@@ -248,10 +248,20 @@ defmodule Support.ComponentSetup do
     %{shape_log_collector: name}
   end
 
-  def with_slot_name_and_stream_id(_ctx) do
-    # Use a random slot name to avoid conflicts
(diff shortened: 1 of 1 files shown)


robacourt (approved):
Lovely!

PR #3088, opened 2025-09-02
feat: Feature flag as stack configuration

msfstef:
In order to be able to configure feature flags per stack we need them to not be a global config.

Trying to thread things through turned out to be a bit messy - perhaps we should revisit the number of configuration calls we're making as I feel we should be able to get them down to 2 (stack config and API config I suppose?). This is ultimately what we're doing but there's many layers of reutilization (`core_configuration` `build_shared_opts` etc) which make it slightly confusing IMO.

One thing worth noting is that I added a default to the `Shape` schema to be `Electric.Config.get_env(:feature_flags)` - otherwise in all tests we'd have to explicitly specify the feature flags. I could make this default a test only thing, let me know your thoughts.

Ultimately outside of tests, `Shape.new` is only called in one place in the API, but in tests we call it _everywhere_.

### DIFF
--- a/.changeset/tame-items-wait.md
+++ b/.changeset/tame-items-wait.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Turn `feature_flags` into an explicit API+StackSupervisor configuration.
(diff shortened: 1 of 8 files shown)


alco (approved):
Looks good!

Perhaps you can add a single unit test that verifies that a descendant process of StackSupervisor can indeed have a different value of `feature_flags` from the one defined in Application env?

msfstef:
> Perhaps you can add a single unit test that verifies that a descendant process of StackSupervisor can indeed have a different value of feature_flags from the one defined in Application env?

Annoyingly the only thing that uses feature flags at the moment is the API which is not part of the stack supervisor - I've added it as part of the shared options as I think it's appropriate in the way that we use it to define the API (both here and in the cloud).

PR #3097, opened 2025-09-04
fix: Do not crash `ReplicationClient` on `ShapeLogCollector` missing

msfstef:
Closes https://github.com/electric-sql/stratovolt/issues/807

I have tried two approaches to this:
- [This commit](https://github.com/electric-sql/electric/commit/f49e8b1989c0352819466225647ca4e3ffac4d61) encapsulates things into the shape log collector and uses a readiness signal
 - Requires some level of extra coupling which we're trying to avoid, ended up not loving it
- [This commit](https://github.com/electric-sql/electric/pull/3097/commits/92361e80fc8279d96a17c9c7e7bf7168f0df6d4f) moves the responsibility to the replication client, reuses the `:consumers_ready` message and `start_streaming` semantics
 - I prefer this cause it reuses existing messaging stemming from our supervision strategy - any time the shape log collector dies it will take down consumers with it, so when everything is back up and ready to process again the replication client can be given a `start_streaming` message which it will ignore if not doing anything, or use to retry processing a transaction

I think the latter approach is cleaner, uses existing semantics, keeps the separation we need between modules and processes.

### DIFF
--- a/packages/sync-service/test/electric/postgres/replication_client_test.exs
+++ b/packages/sync-service/test/electric/postgres/replication_client_test.exs
@@ -194,9 +194,12 @@ defmodule Electric.Postgres.ReplicationClientTest do
 
       send(pid, {:flush_boundary_updated, tx_lsn})
 
-      Process.exit(pid, :some_error)
+      interrupt_val = "interrupt #{inspect(pid)}"
+      insert_item(conn, interrupt_val)
+      refute_received {:from_replication, _}, 50
+      Process.exit(pid, :some_reason)
 
-      assert_receive {:DOWN, ^monitor, :process, ^pid, :some_error}, @assert_receive_db_timeout
+      assert_receive {:DOWN, ^monitor, :process, ^pid, :some_reason}, @assert_receive_db_timeout
 
       refute_received _
 
@@ -205,6 +208,7 @@ defmodule Electric.Postgres.ReplicationClientTest do
       start_client(ctx)
 
       assert %NewRecord{record: %{"value" => "return: not ok"}} = receive_tx_change()
+      assert %NewRecord{record: %{"value" => ^interrupt_val}} = receive_tx_change()
 
       refute_receive _
     end
@@ -638,6 +642,16 @@ defmodule Electric.Postgres.ReplicationClientTest do
       "return: " <> val ->
         send(test_pid, {:from_replication, transaction})
         val
+
+      "interrupt #PID" <> pid_str ->
+        pid = pid_str |> String.to_charlist() |> :erlang.list_to_pid()
+
+        if pid == self() do
+          raise "Interrupting transaction processing abnormally"
+        else
+          send(test_pid, {:from_replication, transaction})
+          :ok
+        end
 
       _ ->
         send(test_pid, {:from_replication, transaction})


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -850,6 +850,16 @@ defmodule Electric.Connection.Manager do
     {:noreply, state, {:continue, :start_streaming}}
   end
 
+  def handle_cast(
+        {:consumers_ready, _recovered, _failed},
+        %{replication_client_pid: replication_client_pid} = state
+      )
+      when is_pid(replication_client_pid) do
```
msfstef: we _could_ potentially assume the replication client pid is always present - I suspect with the change in supervision strategy this will be made clearer

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -361,4 +372,19 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+
+  # Call into the server with an infinite timeout, and if the server is
+  # not available or shutting down, assume that we are listening for it to
+  # come back up and call again.
+  defp call_until_ready(server, call_arg) do
+    GenServer.call(server, call_arg, :infinity)
+  catch
+    :exit, {reason, _} = err when reason in [:noproc, :shutdown] ->
+      receive do
+        {:shape_log_collector_ready, new_server} -> call_until_ready(new_server, call_arg)
```
alco: For this message to be delivered here, the shape log collector must itself be a target.

packages/sync-service/lib/electric/connection/manager.ex
alco: This clause sidesteps the state machine and makes the one above almost redundant, except that the above clause updates the state machine.

Does this imply that we can receive `consumers_ready` in any state?

Why not implement this clause as returning `{:continue, :start_streaming}` so that the manager's state could transition to `waiting_for_streaming_confirmation` which is requried for it to process the `replication_client_streamed_first_message` cast correctly?

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -458,6 +453,20 @@ defmodule Electric.Postgres.ReplicationClient do
     >>
   end
 
+  defp call_until_ready_for_processing(mfa) do
```
alco: This name is a bit ambiguous. I would call it `apply_with_retries` or `apply_until_successful`.

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -458,6 +453,20 @@ defmodule Electric.Postgres.ReplicationClient do
...
+    {m, f, args} = mfa
+    apply(m, f, args)
+  catch
+    :exit, {reason, _} = err when reason in [:noproc, :shutdown] ->
+      Logger.debug(
+        "Call #{inspect(mfa)} failed with #{inspect(err)}, waiting for :start_streaming message before retrying"
+      )
+
+      receive do
+        :start_streaming -> call_until_ready_for_processing(mfa)
```
alco: Let's move this `:start_streaming` constant into a module attribute since it's referenced in more than one place in the module.

alco: Also, it may not be immediately obvious that a failed call will result in an eventual reception of this `:start_streaming` message. It makes a bunch of assumption about what needs to happen in other modules before replication client can retry the call. It also overloads the `:start_streaming` message so that it now not only means that the replication client needs to issue a `START_REPLICATION_SLOT` statement but it can also just unblock it when it's already streaming.

I'm not insisting on it, but I think making a `call()` to Connection.Manager here and letting the connection manager explicitly unblock the replication client with a `GenServer.reply()` would make the sequencing clearer.

At the very least we should document in a comment here why it is reasonable to expect a `:start_streaming` message.

alco (commented):
I like the idea of blocking inside the replication client but I have reservations about piggying back on the `:start_streaming` message like that. I think this opens up new states in Connection.Manager and makes the client unnecessarily coupled to Connection.Manager's logic by assuming it will send a `:start_streaming` message to a replication client process that's already streaming, after it has restarted the shape log collector.

packages/sync-service/lib/electric/postgres/replication_client.ex
msfstef: I like the idea of using an explicit blocking call to "wait" for readiness - I will say however that we currently do not track the replication supervisor's "readiness" in the connection manager for it to know when to answer to this call.

We would need to reinstate some monitoring of e.g. the shape log collector in the connection manager to use this approach - I'm happy to do that and would prefer it (so as to make fewer assumptions in the replication client), but checking if that makes sense to you

alco: > I will say however that we currently do not track the replication supervisor's "readiness" in the connection manager for it to know when to answer to this call.

We kind of do, by way of waiting on the "consumers_ready" event and using it as a trigger to start streaming. I don't think we need to add anything beyond that, just make sure the replication client has an explicit call to Connection.Manager as opposed to assuming it will get the "start_streaming" message when
- it is already streaming
- it just so happens that the function it's trying to call is currently crashing

msfstef: My worry is that if conn man receives `consumers_ready` before it receives the call from the replication client, how does conn man know it can tell the client to resume? We basically need a flag we toggle, and we can toggle it on with `consumers_ready` but not off now that we don't monitor anything

alco: Fair point.

At this point I'm shifting more towards your first suggestion where ShapeLogCollector notifies the interested processes when it's ready.

Right now we use dynamic function references in the replication client that it calls when it gets a new relation or a new transaction. But this is uncommon within our code base where, for the most part, we use plain module references for calls between modules. So I think removing this dynamic dispatch and replacing it with static ShapeLogCollector module references inside ReplicationClient doesn't increase coupling in this case but makes their coupling less obscured.

Now, with that in mind we can state that ReplicationClient directly depends on the connection manager (with which is exchanges a few messages) and on the shape log collector (which it explicitly calls). The invariant we want to implement is this: "replication client survives restarts of the shape log collector". We're just looking for the best mechanism to implement the retry logic when a call to shape log collector fails.

To avoid involving connection manager in the mechanism, we'd either need for the replication client to wait for a new shape log collector process to be come alive (via polling or introduction of another intermediate process on which the replication client can block) or block on receive and let shape log collector unblock it, like you implemented in the earlier commit with "on_ready_targets".

On the other hand, making the replication client call into connection manager involves bringing back the logic in the latter that monitors the shape log collector process.

I'm voicing all of this just to make sure we're on the same page. I'm more inclined to have a direct communication between replication client and shape log collector. But perhaps instead of introducing the waiting logic we could resue the approach we have implemented once in Electric.ShapeCache where it expects the consumer process to be up and retries the call after a 50ms delay in case the process is not there yet. This falls out of the expectation that if the process is not there we expect it to start up shortly.

msfstef: Thank you for laying it all out - we are indeed on the same page.

I agree that we could make the dependency more explicit since it exists either way.

That being said - perhaps a small retry logic like the one in ShapeCache as you mentioned keeps things nicely decoupled while providing significant resilience. We would just have to settle on a max time to wait.

msfstef: On second thought - the retries is a bit annoying because we also have to wait for a readiness step on the shape log collector (set_last_processed_lsn) - so for retries to work we would have to make the shape log collector reject calls while it is still not ready to process them.

Not an unreasonable thing to enforce on the shape log collector though,

msfstef: I've changed things around to retry on all exits and errors for a total of 1 second, with 50ms retries interruptible by exits.

alco (approved):
Looks good!

I didn't think the time remaining was necessary because either the shape log collector becomes available again or the replication supervisor eventually goes down after exhausting restart attempts and takes the connection manager supervisor with it.

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -453,6 +453,25 @@ defmodule Electric.Postgres.ReplicationClient do
...
+  defp apply_with_retries(mfa, time_remaining \\ 1_000) do
+    start_time = System.monotonic_time(:millisecond)
+    {m, f, args} = mfa
+
+    try do
+      apply(m, f, args)
+    catch
+      _, _ when time_remaining > 0 ->
+        receive do
+          # on receiving an exit while holding processing, we should respect the exit
```
alco, line 470: Nice touch!

msfstef:
> I didn't think the time remaining was necessary because either the shape log collector becomes available again or the replication supervisor eventually goes down after exhausting restart attempts and takes the connection manager supervisor with it.

So do you suggest retrying infinitely? I figured I'd add a max retry time for the sake of maintaining the decoupling i.e. not making too many assumptions about the structure of everything else and the replication client still making sense.

I could increase the time remaining to something that would effectively be infinite (given what you mentioned would happen within that time)

edit: in fact, the 1 second I put is not nearly enough as the replication client will wait for the shape log collector to not only be up, but also to become ready, which requires the shape cache to load all consumers, and that potentially takes a while. Perhaps I should remove the time remaining altogether? I've set it to 10 minutes just as a sanity check.

alco:
@msfstef I like the latest edits. Great work!

PR #3098, opened 2025-09-04
fix: reduce file handle usage by 1 on PFS

icehaunter:
There will be a follow-up PR that'll reduce the usage to 1 FH, but it's down to 2 currently.

I've also cleaned up the code here, moving all write loop logic into a separate module because it was unwieldy

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
@@ -27,6 +27,7 @@ defmodule Electric.ShapeCache.PureFileStorage do
   - In case the writer is offline and in-memory ETS/buffer is not present, reads still succeed using on-disk information (i.e. last persisted full txn offset).
   """
 
+  alias Electric.ShapeCache.PureFileStorage.WriteLoop
   alias Electric.ProcessRegistry
   alias Electric.ShapeCache.PureFileStorage.ActionFile
   alias Electric.ShapeCache.PureFileStorage.KeyIndex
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
```diff
@@ -154,27 +154,35 @@ defmodule Electric.ShapeCache.PureFileStorage.LogFile do
...
-             <<json::binary-size(json_size)>> <- IO.binread(file, json_size) do
-          entry =
-            {LogOffset.new(tx_offset, op_offset), key_size, key, op_type, flag, json_size, json}
-
-          {[{entry, position}], {file, expected_position(position, entry)}}
-        else
-          _ -> {:halt, {file, position}}
-        end
+      fn
+        {file, position} when position >= end_position ->
```
msfstef, line 171: not a huge fan of the implicit comparison with an atom - I know that all atoms are larger than numbers in Elixir but using that fact feels hard to read. I'd prefer an explicit check for whether `position` or `end_position` are `:eof` perhaps?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -876,22 +797,17 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
-  defp maybe_open_files(writer_state(open_files: x) = state) when not is_nil(x), do: state
+  defp open_file(opts, suffix, :json_file),
+    do: File.open!(json_file(opts, suffix), [:append, :raw])
 
-  defp maybe_open_files(writer_state(opts: opts, latest_name: latest_name) = state) do
-    {f1, f2, f3} = open_files(opts, latest_name)
-    writer_state(state, open_files: {f1, f2, f3})
-  end
+  defp open_file(opts, suffix, :chunk_file),
+    do: File.open!(chunk_file(opts, suffix), [:append, :raw, :sync])
```
msfstef, line 822: nit: why not carry over the comment on why we use sync mode for chunk files
```
# We're opening the chunk file in sync mode because writes there are rare but we prefer for them
# to be atomic
```

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -387,19 +344,12 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
-      -log_file_pos
-    )
-
     ChunkIndex.copy_adjusting_positions(
       chunk_file(opts, current_latest_suffix),
       chunk_file(opts, new_latest_suffix),
       offset,
       -log_file_pos,
-      -key_file_pos
+      0
```
msfstef, line 364: so we still write key positions in the chunk index but they are entirely unused - is this to ensure some form of compatibility? If we end up deploying this and then rolling back wouldn't we be able to read this but get wrong results as they key positions are nil? Perhaps we remove key positions from the chunk index altogether and bump the `@version` tag for the file storage?

packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
icehaunter: Eof and infinity are very normal and conventional, I'd rather keep this

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: I don't want to blow up shapes just for the sake of it, so I chose to keep zeroes for now. We can make a breaking change later if we want to

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: Forgot

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
msfstef: But it's not just for the sake of it - am I wrong in my understanding that this is only forwards but not backwards compatible? We can fix forward if there's an issue, but my point is that anyone creating files with this version of the file system should not be using them with a previous version.

icehaunter: No, the key position part of chunk indices is just not read at all for the main part of the log, so it's fine to have it whatever it was before this update. Rollback will however break compaction, that's true

msfstef: Hmm ok - compaction is still marked as experimental so I _guess_ it's fine if it ends up breaking for this edge case. Tagging @balegas for a decision on that.

icehaunter:
Benchmark this

icehaunter:
I'm not sure why the comment failed to post, but here are the graphs:

balegas:
Appreciate the speedup on write fan-out, but we got a bit of slow down on the expected case (writes to 1 shape). What's the reason for this?

msfstef:
>Appreciate the speedup on write fan-out, but we got a bit of slow down on the expected case (writes to 1 shape). What's the reason for this?

This is comparing to 1.1.0 so the slow down might have occurred at any point since then I assume

balegas:
only read the comment about compaction now. For Cloud users this is less of a problem since we're busting shape cache on deploys now.

PR #3103, opened 2025-09-06
feat: Add client-side cache buster for expired shapes to prevent 409s

### LINKED ISSUE #3100: Feature request: Client stores if a shape 409s so then it can add a cacheBuster on future requests
**Is your feature request related to a problem? Please describe.**
Electric responds with 409 when part of a shape log is requested that no longer exists. The client then re-requests the shape to get the new shape. This is robust but adds extra latency each time the app loads until the original cached shape response expires.

E.g.

page loads
request to /threads?offset=-1
<- 200 OK (from disk cache). Instant response
electric-handle response header = 43061132-1757029388081097
request to /threads?handle=43061132-1757029388081097&offset=0_0
<- 409 back. Not instant response
electric-handle response header = 106759320-1757086121951522
request to /threads?handle=106759320-1757086121951522&offset=-1
<- 200 back (from disk cache if not first time)
request to /threads?handle=106759320-1757086121951522&offset=0_0
<- 200 back. Not instant response

**Describe the solution you'd like**
The client stores when there's a 409 returned for a shape and in subsequent app loads, the client generates a cacheBuster and uses that when constructing URLs.

This problem is most relevant on the web so we could use localStorage for this.

KyleAMathews:
## Summary
- Implements client-side caching of 409'd shapes using localStorage
- Adds cache_buster parameter to prevent redundant 409 responses  
- Reduces app loading latency when shapes expire

## Changes
- Add `SHAPE_CACHE_BUSTER_QUERY_PARAM` constant for expired shape requests
- Store expired shape handles in localStorage when 409 responses occur
- Check for expired shapes during URL construction and add cache buster
- Gracefully handle localStorage availability and errors

## Test plan
- [ ] Test 409 response handling stores shape in localStorage
- [ ] Verify subsequent requests include cache_buster parameter
- [ ] Confirm behavior works correctly when localStorage unavailable
- [ ] Test multiple expired shapes don't interfere with each other

Fixes #3100

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -28,6 +28,7 @@ import {
   CHUNK_LAST_OFFSET_HEADER,
   LIVE_CACHE_BUSTER_HEADER,
   LIVE_CACHE_BUSTER_QUERY_PARAM,
+  SHAPE_CACHE_BUSTER_QUERY_PARAM,
   COLUMNS_QUERY_PARAM,
   LIVE_QUERY_PARAM,
   OFFSET_QUERY_PARAM,
@@ -517,6 +518,12 @@ export class ShapeStream<T extends Row<unknown> = Row>
         // with the newly provided shape handle, or a fallback
         // pseudo-handle based on the current one to act as a
         // consistent cache buster
+        
+        // Store the current shape handle as expired to avoid future 409s
+        if (this.#shapeHandle) {
+          this.#markShapeExpired(this.#shapeHandle)
+        }
+        
         const newShapeHandle =
           e.headers[SHAPE_HANDLE_HEADER] || `${this.#shapeHandle!}-next`
         this.#reset(newShapeHandle)
@@ -600,6 +607,14 @@ export class ShapeStream<T extends Row<unknown> = Row>
     if (this.#shapeHandle) {
       // This should probably be a header for better cache breaking?
       fetchUrl.searchParams.set(SHAPE_HANDLE_QUERY_PARAM, this.#shapeHandle!)
+      
+      // Add cacheBuster for shapes known to be expired to prevent 409s
+      if (this.#isShapeExpired(this.#shapeHandle)) {
+        fetchUrl.searchParams.set(
+          SHAPE_CACHE_BUSTER_QUERY_PARAM,
+          `expired_${Date.now()}`
+        )
+      }
     }
 
     // sort query params in-place for stable URLs and improved cache hits
@@ -940,6 +955,37 @@ export class ShapeStream<T extends Row<unknown> = Row>
     this.#isUpToDate = false
     this.#connected = false
     this.#schema = undefined
+  }
+
+  /**
+   * Get the localStorage key for a specific shape handle
+   */
+  #getShapeStorageKey(shapeHandle: string): string {
+    return `electric_expired_shape_${shapeHandle}`
+  }
+
+  /**
+   * Check if a shape is marked as expired in localStorage
+   */
+  #isShapeExpired(shapeHandle: string): boolean {
+    if (typeof localStorage === `undefined`) return false
+    try {
+      return localStorage.getItem(this.#getShapeStorageKey(shapeHandle)) === `true`
+    } catch {
+      return false
+    }
+  }
+
+  /**
+   * Mark a shape as expired in localStorage
+   */
+  #markShapeExpired(shapeHandle: string): void {
+    if (typeof localStorage === `undefined`) return
+    try {
+      localStorage.setItem(this.#getShapeStorageKey(shapeHandle), `true`)
+    } catch {
+      // Ignore localStorage errors
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -941,6 +956,39 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+    }
+  }
+
+  /**
+   * Mark a shape as expired in localStorage
+   */
+  #markShapeExpired(shapeHandle: string): void {
+    if (typeof localStorage === `undefined`) return
+    try {
+      localStorage.setItem(this.#getShapeStorageKey(shapeHandle), `true`)
```
kevin-dp: nit: we can remove the if test if we do `localStorage?.setItem(...)`

packages/typescript-client/src/constants.ts
```diff
@@ -5,6 +5,7 @@ export const SHAPE_SCHEMA_HEADER = `electric-schema`
 export const CHUNK_UP_TO_DATE_HEADER = `electric-up-to-date`
 export const COLUMNS_QUERY_PARAM = `columns`
 export const LIVE_CACHE_BUSTER_QUERY_PARAM = `cursor`
+export const SHAPE_CACHE_BUSTER_QUERY_PARAM = `cache_buster`
```
kevin-dp: Do we need to introduce a separate one? Could we not use the `cursor` also for this one? Since up till now we only use `cursor` for cache busting live requests, it seems that we can reuse it to bust non-live shape requests too? Just needs a renaming from `LIVE_CACHE_BUSTER_QUERY_PARAM` to `CACHE_BUSTER_QUERY_PARAM`.

kevin-dp (approved):
Looks good, left a few very minor comments.

samwillis:
I'm slightly nervous of accumulating these expired shape handles in local storage as individual keys, while it's true we can probably hold 10s k in it, it seems a little messy. 

Could we:
- constrain the list to the last 100
- store them in a single key in local storage
- load all into memory on start
- when we add a new one push out the oldest (when > 100) and serialise + save
- watch for the storage change event from other tabs to load the set when it changes:

```
window.addEventListener("storage", (event) => {
  if (event.key === "electric_expired_shape_handles") {
    // reload expired handles
  }
});
```

KyleAMathews:
@samwillis hmm yeah a little LRU cache would be needed for very large apps w/ lots of churn. Easy enough to add...

kevin-dp:
> I'm slightly nervous of accumulating these expired shape handles in local storage as individual keys, while it's true we can probably hold 10s k in it, it seems a little messy.

Keeping track of a set of expired shape handles seems like the perfect use case for a bloom filter. It will tell us whether the shape is possibly expired or definitely not expired. If it tells us it is possibly expired, we add a cache buster (no real harm if it would be a false positive).

Might be overkill for apps with "few" shapes as we can easily store several thousand shape handles in local storage but might be worth it for apps with huge amounts of shapes.

packages/typescript-client/src/constants.ts
KyleAMathews: The LLM thought the same thing 😆 

I suppose in theory we could because this new cache busting is only relevant when we're not live and reading cached values & the live cache cursor is needed for preventing CDNs from returning cached live requests 🤔 

But they are different. This PR introduces an actual normal cache buster where the live stuff is a cursor — so cache busting but different flavor as it's helping ensure we move forward through log time.

KyleAMathews:
> perfect use case for a bloom filter

Yeah with a ton more shapes that could make sense — I'm more worried though about all the extra code for a bloom filter. I'm thinking of capping the number of shapes at 250 as that's a lot more than we support right now anyways. That'll be cheap still to load (10-15kb) and iterate over.

KyleAMathews:
This is ready for another review.

packages/typescript-client/src/expired-shapes-cache.ts
```diff
@@ -0,0 +1,67 @@
+interface ExpiredShapeCacheEntry {
+  expiredHandle: string
+  lastUsed: number
+}
+
+/**
+ * LRU cache for tracking expired shapes with automatic cleanup
+ */
+export class ExpiredShapesCache {
+  private data: Record<string, ExpiredShapeCacheEntry> = {}
```
samwillis, line 10: We are only keeping the last expired handle per url? Is that enough?

samwillis (commented):
I think we are close, and maybe there, but wanted to clarify some thinking.

You are saving one expired handle per url, but as we have seen there can be multiple that cause weird loops. What was your thinking here?

In my mind, as handles are uuids we don't need to bucket them by url, just have a LRU cache of the last n of them. I would keep this as a list with a max length serialised to storage, and as a set in memory for fast lookup.

It may well be that per url is ok, in which case this is good to go I think.

Once thing I think thats worth following up on in a further PR is standardising a key-value metadata interface that `ShapeStream` takes as a config option. This can default to using a thin wrapper around localStorage in the browser, but provides a route to other storages for React Native / Capacitor / Electric / Node.

packages/typescript-client/src/expired-shapes-cache.ts
KyleAMathews: Yeah the idea is that a given shape definition's browser cache might no longer be valid — so the cache key is by shape definition and we only need to store the last-invalidated handle to get out of that handle's cache. We only load one shape / shape definition so no need to store multiple handles or worry about additional URLs from other timelines of the shape that we might have cached in the past.

Issue #2904, opened 2025-07-22
Feature request: add transform option to shape options.


thruflo:
Users would like to transform rows. For example to map snake_case to camelCase field names.

We provide the parser option to transform individual values. We don’t provide a row-level transform function.

**Describe the solution you'd like**

I propose we add a transform or transformRow option to the shape options. This would be a function that takes a row value and potentially transforms it.

I think this should be in the client.

In addition, we could also provide a documentation / example code snippet for a node/worker proxy that transforms rows when proxying.

This would demonstrate how to do the transformation on the server (behind or in front of the CDN) without needing to build this into core Electric.

**Additional context**

https://discord.com/channels/933657521581858818/933657523049885698/1396235679427461122

KyleAMathews:
Fixed by https://github.com/electric-sql/electric/pull/3105

PR #3104, opened 2025-09-06
Optional transformRow function in Shape constructor.

stereosteve:
Can be used to convert object keys to camelCase.

Possible solution to: https://github.com/electric-sql/electric/issues/2904

Questions:
* Perhaps this should be an option on `ShapeStreamOptions` so that it is exposed to the `useShape` hook?

### DIFF
--- a/packages/typescript-client/src/shape.ts
+++ b/packages/typescript-client/src/shape.ts
@@ -11,6 +11,10 @@ export type ShapeChangedCallback<T extends Row<unknown> = Row> = (data: {
 
 type ShapeStatus = `syncing` | `up-to-date`
 
+export type TransformRowFunction<T extends Row<unknown> = Row> = (
+  message: T
+) => T
+
 /**
  * A Shape is an object that subscribes to a shape log,
  * keeps a materialised shape `.rows` in memory and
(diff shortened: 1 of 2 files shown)


packages/typescript-client/src/shape.ts
```diff
@@ -11,6 +11,10 @@ export type ShapeChangedCallback<T extends Row<unknown> = Row> = (data: {
 
 type ShapeStatus = `syncing` | `up-to-date`
 
+export type TransformRowFunction<T extends Row<unknown> = Row> = (
```
KyleAMathews, line 14: Let's just call it `transform` — in the future Electric will support non-sql/table data sources so it won't always be "rows" people are transforming.

KyleAMathews (commented):
Looks great!

Could you also add a changeset & update the docs?

stereosteve:
@KyleAMathews yes can do!

After looking at it a bit more, I think https://github.com/electric-sql/electric/pull/3105 might be a better approach?

stereosteve:
closing in favor of: https://github.com/electric-sql/electric/pull/3105

PR #3105, opened 2025-09-06
feat: Add transformer to ShapeStreamOption.

stereosteve:
`messageParser` will call `transformer` with the complete `message.value`.  Can be used to convert object keys to camelCase.

Possible solution to: https://github.com/electric-sql/electric/issues/2904.

This is an alternative to https://github.com/electric-sql/electric/pull/3104 that would make it an option on the ShapeStreamOptions, so it can be exposed to the `useShape` hook without extra work.

### DIFF
--- a/.changeset/bright-fishes-attack.md
+++ b/.changeset/bright-fishes-attack.md
@@ -1,5 +1,6 @@
 ---
 "@electric-sql/client": patch
+"@electric-sql/react": patch
 ---
 
 Add `transformer` function to `ShapeStreamOptions` to support transforms like camelCase keys.


packages/typescript-client/src/parser.ts
```diff
@@ -94,11 +98,16 @@ export function pgArrayParser<Extensions>(
 
 export class MessageParser<T extends Row<unknown>> {
   private parser: Parser<GetExtensions<T>>
-  constructor(parser?: Parser<GetExtensions<T>>) {
+  private transformer?: TransformFunction
+  constructor(
+    parser?: Parser<GetExtensions<T>>,
+    transformer?: TransformFunction
```
KyleAMathews:
```suggestion
    transformer?: TransformFunction<T>
```

packages/typescript-client/src/parser.ts
```diff
@@ -19,6 +19,10 @@ export type Parser<Extensions = never> = {
   [key: string]: ParseFunction<Extensions>
 }
 
+export type TransformFunction = (
```
KyleAMathews:
```suggestion
export type TransformFunction<T extends Record<string, unknown>> =
```

packages/typescript-client/src/parser.ts
```diff
@@ -19,6 +19,10 @@ export type Parser<Extensions = never> = {
...
+  message: Record<string, unknown>
+) => Record<string, unknown>
```
KyleAMathews:
```suggestion
) => T
```

packages/typescript-client/src/client.ts
```diff
@@ -259,6 +259,7 @@ export interface ShapeStreamOptions<T = never> {
   fetchClient?: typeof fetch
   backoffOptions?: BackoffOptions
   parser?: Parser<T>
+  transformer?: TransformFunction
```
KyleAMathews:
```suggestion
  transformer?: TransformFunction<T>
```

packages/typescript-client/src/parser.ts
```diff
@@ -118,6 +127,8 @@ export class MessageParser<T extends Row<unknown>> {
         Object.keys(row).forEach((key) => {
           row[key] = this.parseRow(key, row[key] as NullableToken, schema)
         })
+
+        if (this.transformer) value = this.transformer(value)
```
KyleAMathews, line 131: Adding a try/catch here w/ a custom error would be good as this is user-supplied code

stereosteve: Do you think the catch should do a `console.log` or just return the un-transformed value?

stereosteve:
@KyleAMathews I tried tracking functions by reference with the WeakMap... but it caused existing tests with a `fetchClient` to fail since the callback is inline.

We could skip `fetchClient` specifically when computing hash... but I thought users might get surprised if they specify `transformer` inline and end up with a different `shapeStream` on every call.

So for function values it'll return the text of the function body, which should avoid these surprises.  Does that seem okay to you?

website/docs/api/clients/typescript.md
```diff
@@ -238,6 +238,12 @@ export interface ShapeStreamOptions<T = never> {
    */
   parser?: Parser<T>
 
+  /**
+   * A function to transform the Message value before emitting to subscribers.
+   * This can be used to camelCase keys or rename fields.
+   */
+  transformer?: TransformFunction
```
KyleAMathews:
```suggestion
  transformer?: TransformFunction<T>
```

packages/typescript-client/src/parser.ts
KyleAMathews: oh we're fine actually — this try/catch should catch any thrown errors https://github.com/electric-sql/electric/blob/8623e7320c91bf1298c4e2896dbf6fcd323f5fdf/packages/typescript-client/src/client.ts#L429-L451

The idea is that all errors during fetching should be exposed to an optional `onError` callback

KyleAMathews:
> So for function values it'll return the text of the function body, which should avoid these surprises. Does that seem okay to you?

Yup! That works too

packages/typescript-client/src/parser.ts
stereosteve: OK will remove!

.changeset/bright-fishes-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"@electric-sql/client": patch
```
KyleAMathews, line 2: we also need to add `@electric-sql/react` here

Issue #3102, opened 2025-09-05
Ensure flush tracking acknowledges minimum of last flushed shapes


msfstef:
**Versions**
- Electric: 1.1.9

**Bug description**
Flush acknowledgement is really delayed and WAL builds up on Postgres when high number of transactions are coming through.

**Expected behavior**
Should be acknowledging LSNs on a more rolling basis - handle cases where transactions have come in before acknowledgements from previous transactions have come back.

PR #3106, opened 2025-09-07
fix: Flush tracker acknowledgements under heavy load

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3102

The main issue was that whenever we starter tracking flushes for a shape that we were not already tracking, we assumed its `last_flushed_offset` was the `last_global_flushed_offset`. This sounds like a reasonable assumption, and it is a valid lower bound for the new shape's last flushed offset, but it leads to the following scenario covered by the new test:

## Problematic Case

0. Initial state is `last_global_flushed_offset` is at transaction 9.
1. Transaction 10 for Shape A is ingested
2. Transaction 11 for Shape B is ingested
3. Flush ack for transaction 10 by Shape A is received
  3a. `last_global_flushed_offset` is not updated as Shape B's most recent flush  was set to txn 9
4. Transaction 12 for Shape A is ingested
  4a. We set Shape A's most recent flush as `last_global_flushed_offset` which is txn 9, despite the fact that we had just acknowledged its flush of transaction 10 moments ago.
5. Flush ack for transaction 11 by Shape B is received
  5a. Again no updates to `last_global_flushed_offset` as Shape A's most recent one is still at txn 9

This can go on with multiple shapes, and with enough transactions the `last_flushed` mapping is never empty and thus `last_global_flushed_offset` is never updated.

## Proposed Solution

The solution I have implemented is changing our assumption of what the last flushed transaction for a newly tracked shape is. Instead of using the `last_global_flushed_offset`, we assume the shape has flushed all transactions up to the previous one from the one that is being ingested now.

I claim that this is a safe assumption and a safe upper bound, because if the shape was not being tracked when a transaction is ingested, it means that:
1. Either the shape was previously tracked, and was flushed up to the most recently seen transaction and removed from the mapping
2. Or the shape is new and this is the first relevant transaction ingested for the shape

In either case, we can assume that all relevant transactions to the shape before the current one have been flushed. If the transaction right before this one was not relevant to the shape, and thus was not sent to the shape, it is safe to say that it is effectively flushed, as flushing a transaction for a shape that has nothing to write from it involves doing nothing (it is "flushed" by virtue of it being irrelevant).

## Attempts to cover more issues

I spent some time thinking about what we could do to avoid more issues. I tried writing a property test but it ended up not being as simple as I was initially envisioning it.

One thing we could do is block the ingestion of new transactions if we have not acknowledged anything for a while (or if acknowledged seen LSN is getting significantly larger than the ackowledged flushed LSN), so as to give time for shapes to flush and acknowledge things fully. Basically create backpressure based on how far the seen and flushed acknowledged LSNs are getting.

P.S. I also made a bunch of tests async as it was missing and I didn't see any particular reason they weren't before.

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
@@ -1,4 +1,5 @@
 defmodule Electric.Replication.ShapeLogCollector.FlushTracker do
+  require Electric.Replication.LogOffset
   alias Electric.Replication.LogOffset
   alias Electric.Replication.Changes.Transaction
 
@@ -52,7 +53,7 @@ defmodule Electric.Replication.ShapeLogCollector.FlushTracker do
   2. Determine affected shapes
   3. For each shape,
     1. If Mapping already has the shape, update `last_sent` to the max offset of the transaction
-    2. If Mapping doesn’t have the shape, add it with `{last_sent, last_global_flushed_offset}`
+    2. If Mapping doesn’t have the shape, add it with `{last_sent, prev_log_offset}` where `prev_log_offset` is the offset just before this transaction
   4. If Mapping is empty after this update, then we’re up-to-date and should consider this transaction immediately flushed. Update `last_global_flushed_offset` and `maxFlushedOffset` to be the max offset of the transaction
   5. Wait for the writers to send the flushed offset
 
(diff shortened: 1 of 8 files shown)


packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
```diff
@@ -88,40 +93,50 @@ defmodule Electric.Replication.ShapeLogCollector.FlushTracker do
...
       ) do
-    mapset = tree_get_shape_set(state, last_global_flushed_offset)
+    # for affected shapes that are not currently tracked, assume their last flushed offset
+    # is the transaction before this one, which is a safe upper bound
+    prev_log_offset = %LogOffset{tx_offset: last_log_offset.tx_offset - 1}
 
-    {last_flushed, mapset} =
-      Enum.reduce(affected_shapes, {last_flushed, mapset}, fn shape, {acc, mapset} ->
+    {last_flushed, new_shape_ids} =
+      Enum.reduce(affected_shapes, {last_flushed, MapSet.new()}, fn shape, {acc, new_shape_ids} ->
```
magnetised: nit: the renaming is a good idea, but why not go futher and rename `acc` to `last_flushed`or `new_last_flushed`

magnetised (approved):
This is the first time I've delved into the flushtracker so I'm no expert but the reasoning re lsns seems sound to me and you've definitely caught a bug if that new test was failing.

balegas:
I'd really like to dig more into the algorithm. It does seem complex :)

PR #3108, opened 2025-09-08
Avoid invoking PublicationManager from SchemaReconciler when there are no diverged relations

alco:
Just getting rid of a no-op function invocation that pollutes the log with periodic log entries.

### DIFF
--- a/integration-tests/tests/not-owner-of-the-table.lux
+++ b/integration-tests/tests/not-owner-of-the-table.lux
@@ -15,7 +15,7 @@
   "-v $(realpath ../scripts/low_privilege_db.sh):/docker-entrypoint-initdb.d/initdb-low_privilege_db.sh"]
 
 ## Start the sync service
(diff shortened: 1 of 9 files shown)


alco:
As the failing integration test has shown, the change in the 1st commit alone wasn't enough. I have pushed another change that moves the responsibility of refreshing the publication with `forced?=true` every minute into the publication manager itself, since it's not the responsibility of SchemaReconciler.

As for the need to fetch the publication state from the database once every minute, it could be debated. One thing it does for us is repair any meddling with the publication in Postgres outside of Electric. For example, if the admin adds a table to the publication that Electric can't read from (or can't update the replica identity off), the periodic `refresh_publication` call will remove that table from the publication since it's not used by any shape.

alco:
Rebased the branch on top of https://github.com/electric-sql/electric/pull/3119

alco:
Rebased once more on top of https://github.com/electric-sql/electric/pull/3152.

Issue #3047, opened 2025-08-26
Some (all?) metrics do not get their stack_id metadatum renamed to source_id


alco:
<img width="1775" height="1026" alt="Image" src="https://github.com/user-attachments/assets/5c9220c4-07e8-48f9-a11e-b7cf95961b81" />

alco:
This issue hasn't been fixed. None of the metrics we export to Honeycomb have `source_id` attribute.

alco:
The problem is that the otel_metric_exporter lib does not read stack_id or source_id from the measurement metadata. It takes `stack_id` from the resource config which is passed to it at process initialization time.

We need to clean up the way OTEL is configured because we have multiple resource configurations and detections throughout the codebase.

PR #3110, opened 2025-09-08
fix: Ensure all stack telemetry respects `telemetry_span_attrs`

msfstef:
Closes https://github.com/electric-sql/electric/issues/3047

- Fixes `electric.postgres.replication.transaction_received.bytes` to use actual byte size of transaction
  - I made the collector return an extra element in the tuple that is transaction meta to avoid adding more properties on the transaction itself unless they are actually needed in the shapes - no reason to copy over more things to all consumers
- Implemented thin wrapper over `:telemetry.execute` in OpenTelemetry module which we use for spans so that any stack related telemetry uses that and respects `telemetry_span_attrs`
  - This ensures all stack metrics actually end up having a `source_id` on them in Cloud

### DIFF
--- a/.changeset/early-rules-smile.md
+++ b/.changeset/early-rules-smile.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Ensure all stack telemetry adds `telemetry_span_attrs` to its metadata.
--- a/.changeset/giant-students-juggle.md
+++ b/.changeset/giant-students-juggle.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix `electric.postgres.replication.transaction_received.bytes` metric to use actual transaction size.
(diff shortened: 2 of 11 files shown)


magnetised (approved):
Nice

msfstef:
>I'm minorly worried that wrapping :telemetry.execute() calls inside Electric.Telemetry.OpenTelemetry obscures the fact that those metrics are used not only by our OpenTelemetry exporter but laos the Prometheus and CallHomeReporter ones. At the same time the name get_stack_span_attrs() doesn't reflect the fact that those attrs are also put on metrics, not only spans.
>
>But we could make it an issue for the future to adjust the naming.

Agree on all fronts - I'll merge this and open the next issue immediately - I'll quote your concern as is if you don't mind haha

msfstef:
Opened issue https://github.com/electric-sql/electric/issues/3111

PR #3113, opened 2025-09-08
fix(sync-service): Split pg connections across two pools

magnetised:
Separating long-running snapshot connections from the various admin/metadata queries

Also move connection opts discovery into external process

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -82,11 +82,6 @@ defmodule Electric.Connection.Manager do
       :connection_opts,
       # Replication options specific to `Electric.Postgres.ReplicationClient`
       :replication_opts,
-      # Connection options that are shared between regular connections and the replication
-      # connection. If this is set to `nil` post-initialization, it means that regular
-      # connections and the replication connection have been configured using different
-      # connection URLs.
-      :shared_connection_opts,
       # Database connection pool options
       :pool_opts,
       # Options specific to `Electric.Timeline`
@@ -127,12 +122,14 @@ defmodule Electric.Connection.Manager do
       :tweaks,
       :persistent_kv,
       :purge_all_shapes?,
+      validated_connection_opts: %{replication: nil, pool: nil},
       drop_slot_requested: false
     ]
   end
 
   use GenServer, shutdown: :infinity
   alias Electric.Connection.Manager.ConnectionBackoff
+  alias Electric.Connection.Manager.ConnectionResolver
   alias Electric.DbConnectionError
   alias Electric.StatusMonitor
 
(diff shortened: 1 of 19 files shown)


packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -154,7 +154,9 @@ defmodule Electric.Replication.PublicationManager do
       stack_id = Keyword.fetch!(opts, :stack_id)
 
       name = Keyword.get(opts, :name, name(stack_id))
-      db_pool = Keyword.get(opts, :db_pool, Electric.Connection.Manager.pool_name(stack_id))
+
+      db_pool =
+        Keyword.get(opts, :db_pool, Electric.Connection.Manager.pool_name(stack_id, :metadata))
```
msfstef: nit: should this really be called a "metadata" pool if we use it to alter publications? perhaps a "control" pool or something would be more appropriate?

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -181,12 +181,13 @@ defmodule Electric.Connection.Manager do
...
-  def pool_name(stack_id) when not is_map(stack_id) and not is_list(stack_id) do
-    Electric.ProcessRegistry.name(stack_id, @db_pool_ephemeral_module_name)
+  def pool_name(stack_id, role)
+      when not is_map(stack_id) and not is_list(stack_id) and role in [:metadata, :snapshot] do
+    Electric.ProcessRegistry.name(stack_id, @db_pool_ephemeral_module_name, role)
   end
 
   def pool_name(opts) do
-    pool_name(Access.fetch!(opts, :stack_id))
+    pool_name(Access.fetch!(opts, :stack_id), Access.get(opts, :role, :snapshot))
```
msfstef: should we enforce a role is provided with `fetch!`? this way we don't accidentally end up creating contention where we shouldn't, I prefer that over convenience

packages/sync-service/lib/electric/connection/manager/pool.ex
```diff
@@ -70,6 +74,7 @@ defmodule Electric.Connection.Manager.Pool do
     Logger.metadata(stack_id: opts[:stack_id])
     Electric.Telemetry.Sentry.set_tags_context(stack_id: opts[:stack_id])
 
+    role = Access.fetch!(opts, :role)
```
msfstef: nit: add it in the `Logger.metadata` so we can more easily distinguish which pool errors and logs are coming from. Otherwise I'd also advise to change all other logs to include the role which is a bit of a pain.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -237,8 +238,20 @@ defmodule Electric.Connection.Manager do
...
   end
 
-  def connection_pool_ready(manager) do
-    GenServer.cast(manager, :connection_pool_ready)
+  def connection_pool_ready(manager, role) do
+    GenServer.cast(manager, {:connection_pool_ready, role})
+  end
+
+  def pool_sizes(total_pool_size) do
+    metadata_pool_size = if(total_pool_size >= 4, do: 4, else: total_pool_size)
```
msfstef: nice - so we configure a general pool size and we just "steal" a few for metadata, sounds reasonable

msfstef:
> Messes with some assumptions in the connection manager I think. Is this critical considering the current re-write-in-progress?

The way you've written it it feels like it maintains exactly the same semantics as with the one pool:
- If it exits during connection setup, try falling back on ssl/ipv4 and crash conn man stack otherwise (now doing this for either pool)
  - Here the assumption that I think is broken is that if the first pool crashes because of SSL issues, the second is guaranteed to also crash, but the first crash will "fall back" the connection options and the second one will notice that conn options already did the fallback and crash the entire stack - this is _already_ an issue and reading your code made me realise that, it's why the ipv6 fallback was not working.
- If it exits after that, crash the conn man stack and restart

So apart from an already existing bug in our fallback logic, which will become worse/more frequent, I think the rest is sound?

packages/sync-service/lib/electric/replication/publication_manager.ex
alco: I would make separate functions on `Connection.Manager`: `admin_pool()`, `data_pool()`, etc. They don't have to include the word "name" in their names. It's going to be clearer this way. It's not like we're going to programmatically choose a pool at runtime and therefore make the type of pool a function argument.

msfstef: Agreed both on the naming and using separate functions

msfstef (approved):

PR #3115, opened 2025-09-09
Fix transformer option typing

stereosteve:
I tried some TypeScript tricks to make `TransformFunction` generic, but found it hard to write a version that would satisfy type constraints after narrowing in the `ShapeStream` class.

But I think this is okay, since the call site for `transformer` is in the land of `JSON.parse` and the final value is returned `as Result` that `Record<string, unknown>` is the most accurate typing available in that context.

This also saves end users from having to return with `as Row` in their `transformer` implementation.  It seems like the lowest friction signature for users.

I tested with:

```
cd packages/typescript-client && pnpm typecheck && pnpm build && cd -
cd packages/react-hooks && pnpm build && cd -
cd examples/nextjs && pnpm typecheck && cd -
cd examples/proxy-auth && pnpm typecheck && cd -
```

### DIFF
--- a/packages/typescript-client/src/parser.ts
+++ b/packages/typescript-client/src/parser.ts
@@ -19,9 +19,9 @@ export type Parser<Extensions = never> = {
   [key: string]: ParseFunction<Extensions>
 }
 
-export type TransformFunction<Extensions = never> = (
+export type TransformFunction = (
   record: Record<string, unknown>
-) => Row<Extensions>
+) => Record<string, unknown>
 
 const parseNumber = (value: string) => Number(value)
 const parseBool = (value: string) => value === `true` || value === `t`
@@ -98,10 +98,10 @@ export function pgArrayParser<Extensions>(
 
 export class MessageParser<T extends Row<unknown>> {
   private parser: Parser<GetExtensions<T>>
-  private transformer?: TransformFunction<GetExtensions<T>>
+  private transformer?: TransformFunction
   constructor(
     parser?: Parser<GetExtensions<T>>,
-    transformer?: TransformFunction<GetExtensions<T>>
+    transformer?: TransformFunction
   ) {
     // Merge the provided parser with the default parser
     // to use the provided parser whenever defined
--- a/packages/typescript-client/src/parser.ts
+++ b/packages/typescript-client/src/parser.ts
@@ -20,7 +20,7 @@ export type Parser<Extensions = never> = {
 }
 
 export type TransformFunction<Extensions = never> = (
-  message: Row<Extensions>
+  record: Record<string, unknown>
 ) => Row<Extensions>
 
 const parseNumber = (value: string) => Number(value)
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -259,7 +259,7 @@ export interface ShapeStreamOptions<T = never> {
   fetchClient?: typeof fetch
   backoffOptions?: BackoffOptions
   parser?: Parser<T>
-  transformer?: TransformFunction<T>
+  transformer?: TransformFunction
 
   /**
    * A function for handling shapestream errors.
--- a/packages/typescript-client/test/client.test.ts
+++ b/packages/typescript-client/test/client.test.ts
@@ -127,13 +127,19 @@ describe.for(fetchAndSse)(
     }) => {
       const [id] = await insertIssues({ title: `test title` })
 
+      type CustomRow = {
+        ID: number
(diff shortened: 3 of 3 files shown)


packages/typescript-client/src/parser.ts
```diff
@@ -98,10 +98,10 @@ export function pgArrayParser<Extensions>(
 
 export class MessageParser<T extends Row<unknown>> {
   private parser: Parser<GetExtensions<T>>
-  private transformer?: TransformFunction<GetExtensions<T>>
+  private transformer?: TransformFunction
   constructor(
     parser?: Parser<GetExtensions<T>>,
-    transformer?: TransformFunction<GetExtensions<T>>
```
KyleAMathews, line 104: Gonna tag @kevin-dp on this as he added all these types

kevin-dp (commented):
I'm not sure which problem this solves, could you please clarify this with some code examples.

I wrote the original version of the parser with the extensions and a community member added support for transformers with types based on those of the parser. I do think that the transformer type is correct and precise:
```ts
export type TransformFunction<Extensions = never> = (
  message: Row<Extensions>
) => Row<Extensions>
```
The input type is a `Row` that can contain extensions (e.g. if the user is using Dates or other types that aren't built into our ts client). The input and output of the transformer are both typed as `Row<Extensions>` but that doesn't mean they need to be the same type. The input and output can be different as long as they are both rows (optionally with types from the extensions).

This PR removes the `Extensions` from the transformer and types it as `Record<string, unknown>`. This is a more general type and i prefer to stick to the more precise type whenever possible.

stereosteve:
Yeah sorry, I didn't provide much context in the PR description.

Yesterday made [PR 3105](https://github.com/electric-sql/electric/pull/3105) to let the user specify arbitrary record transforms at parse time... the motivation being converting keys from `snake_case` to `camelCase` for this issue https://github.com/electric-sql/electric/issues/2904.

Given the primary use case is to rename fields, I think at least the `transformer` input should be `Record<string, unknown>` even if the output is a generic `CustomRow` type.

But when merged to main, the vanilla usage in `nextjs` and `proxy-auth` examples got this type error:

```
Types of property 'transformer' are incompatible.
    Type 'TransformFunction<never> | undefined' is not assignable to type 'TransformFunction<string | number> | undefined'.
      Type 'TransformFunction<never>' is not assignable to type 'TransformFunction<string | number>'.
        Types of parameters 'message' and 'message' are incompatible.
          Type 'Row<string | number>' is not assignable to type 'Row<never>'.
            Type 'string | number' is not assignable to type 'never'.
              Type 'string' is not assignable to type 'never'.
```

The tricky part is we want to add an optional `transformer` function to `ShapeStreamOptions` which is defined as:

```
export interface ShapeStreamOptions<T = never>
```

In the `ShapeStream` constructor this gets narrowed to:

```
ShapeStreamOptions<GetExtensions<T>>
```

So I couldn't figure out a way to define a generic `TransformFunction` interface that could be used on the `ShapeStreamOptions` definition and also in `MessageParser` after the options have been narrowed to `ShapeStreamOptions<GetExtensions<T>>`.

Granted I'm at the edge of my TypeScript knowledge.  @kevin-dp let me know if you have ideas about how to correctly type in both the `ShapeStreamOptions` and `MessageParser`.

stereosteve:
Also worth noting that this is primarily an escape hatch for users who might have existing types from a REST API and have to conform the postgres schema so they can introduce electric within an existing codebase.

I was trying to make the interface simple so the user didn't have to deal with internal type details that might be confusing.  There's no real internal use of this `TransformFunction` option.

It'd be unfortunate if the user had to do this to satisfy the typing along the way:

```ts
      // transformer example: uppercase keys
      const uppercaseKeys: TransformFunction<GetExtensions<CustomRow>> = (
        row
      ) =>
        Object.fromEntries(
          Object.entries(row).map(([k, v]) => [k.toUpperCase(), v])
        ) as CustomRow

      const shapeStream = new ShapeStream<CustomRow>({
        url: `${BASE_URL}/v1/shape`,
        params: {
          table: issuesTableUrl,
        },
        transformer: uppercaseKeys,
      })
```

stereosteve:
closed by: https://github.com/electric-sql/electric/pull/3117

kevin-dp:
Hi @stereosteve, thanks for elaborating on the matter. Your transformer function unraveled a problem with our implementation of `GetExtensions`. As you noticed i have fixed that problem and i believe this solves the problems you had with the typing of the `TransformFunction`. Let me know if that's not the case.

PR #3117, opened 2025-09-09
Fix GetExtensions to infer minimal Extensions

kevin-dp:
Fixes the issue with the transformer which is blocking CI due to failing typecheck.

The problem was that the `GetExtensions` type was inferring types that aren't extensions:
```ts
GetExtensions<{
  id: number
  name: string
  org_id: number
}>
```
The snippet above was being resolved to `string | number` instead of `never`. That was wrong because `string` and `number` are built-in types so they don't count as extensions. Since this object type has no extensions it should return `never`.

Fixed that by changing the `GetExtensions` type to explicitly check whether the type extends the base `Row` type. If that's the case, then there are no extensions and we just return `never`. Otherwise, we infer the extensions.

### DIFF
--- a/.changeset/spotty-oranges-pretend.md
+++ b/.changeset/spotty-oranges-pretend.md
@@ -1,5 +1,0 @@
----
-"@electric-sql/client": patch
----
-
-Fix problem with types of transformer due to a problem with the GetExtensions type.
--- a/.changeset/spotty-oranges-pretend.md
+++ b/.changeset/spotty-oranges-pretend.md
@@ -1,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
+Fix problem with types of transformer due to a problem with the GetExtensions type.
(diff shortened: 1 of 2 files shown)


.changeset/spotty-oranges-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@electric-sql/client": patch
+---
+
+Fix problem with types of transformer due to a problem with the GetExtensions type.
```
msfstef: nit we can probably remove the changeset since this is amending the issue introduced by an existing changeset

msfstef (approved):

.changeset/spotty-oranges-pretend.md
kevin-dp: Don't we want this changeset such that from the release notes it is clear that this issue is fixed in this release? Or we didn't release yet?

PR #3118, opened 2025-09-09
fix(sync-service): Move connection opts discovery into external process

### LINKED ISSUE #3099: Ensure IPv6 options and fallbacks work for both replication and pooled connection
We've noticed a discrepancy between IPv6 fallback logic between the replication connection and pooled connection - ensure that it works and is tested properly for both.

magnetised:
Fixes #3099

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -82,11 +82,6 @@ defmodule Electric.Connection.Manager do
       :connection_opts,
       # Replication options specific to `Electric.Postgres.ReplicationClient`
       :replication_opts,
-      # Connection options that are shared between regular connections and the replication
-      # connection. If this is set to `nil` post-initialization, it means that regular
-      # connections and the replication connection have been configured using different
-      # connection URLs.
-      :shared_connection_opts,
       # Database connection pool options
       :pool_opts,
       # Options specific to `Electric.Timeline`
@@ -321,27 +316,16 @@ defmodule Electric.Connection.Manager do
     in_connection_opts = Keyword.fetch!(opts, :connection_opts)
     in_replication_opts = Keyword.fetch!(opts, :replication_opts)
 
-    # If we see that both top-level connection opts and replication connection opts have been
-    # initialized from the same kwlist, we'll skip the extra work and only perform the no-ssl
-    # and ipv4 fallbacks once.
-    shared_connection_opts = Keyword.fetch!(in_replication_opts, :connection_opts)
+    replication_connection_opts = Keyword.fetch!(in_replication_opts, :connection_opts)
 
-    connection_opts = if is_nil(shared_connection_opts), do: in_connection_opts
+    connection_opts = in_connection_opts || replication_connection_opts
 
     replication_opts =
       in_replication_opts
       |> Keyword.put(:start_streaming?, false)
       |> Keyword.put(:connection_manager, self())
-      |> Keyword.update!(:connection_opts, fn in_connection_opts ->
-        if is_nil(shared_connection_opts), do: in_connection_opts
-      end)
 
-    %{
-      state
-      | shared_connection_opts: shared_connection_opts,
-        connection_opts: connection_opts,
-        replication_opts: replication_opts
-    }
+    %{state | connection_opts: connection_opts, replication_opts: replication_opts}
   end
 
   defp validate_connection(conn_opts, type, state) do
@@ -1205,13 +1189,10 @@ defmodule Electric.Connection.Manager do
   end
 
   defp pooled_connection_opts(state) do
-    state.shared_connection_opts || state.connection_opts
+    state.connection_opts
   end
 
-  defp replication_opts(%State{shared_connection_opts: nil} = state), do: state.replication_opts
-
-  defp replication_opts(%State{shared_connection_opts: conn_opts} = state),
-    do: Keyword.put(state.replication_opts, :connection_opts, conn_opts)
+  defp replication_opts(%State{} = state), do: state.replication_opts
 
   defp drop_slot(%State{pool_pids: %{admin: {pool_pid, _}}} = state) when is_pid(pool_pid) do
     pool = pool_name(state.stack_id, :admin)
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -127,12 +127,14 @@ defmodule Electric.Connection.Manager do
       :tweaks,
       :persistent_kv,
       :purge_all_shapes?,
+      validated_connection_opts: %{replication: nil, pool: nil},
       drop_slot_requested: false
     ]
   end
 
   use GenServer, shutdown: :infinity
   alias Electric.Connection.Manager.ConnectionBackoff
+  alias Electric.Connection.Manager.ConnectionResolver
   alias Electric.DbConnectionError
   alias Electric.StatusMonitor
 
@@ -322,20 +324,16 @@ defmodule Electric.Connection.Manager do
     # If we see that both top-level connection opts and replication connection opts have been
     # initialized from the same kwlist, we'll skip the extra work and only perform the no-ssl
     # and ipv4 fallbacks once.
-    shared_connection_opts =
-      if in_connection_opts == Keyword.fetch!(in_replication_opts, :connection_opts) do
-        populate_connection_opts(in_connection_opts)
-      end
+    shared_connection_opts = Keyword.fetch!(in_replication_opts, :connection_opts)
 
-    connection_opts =
-      if is_nil(shared_connection_opts), do: populate_connection_opts(in_connection_opts)
+    connection_opts = if is_nil(shared_connection_opts), do: in_connection_opts
 
     replication_opts =
       in_replication_opts
       |> Keyword.put(:start_streaming?, false)
       |> Keyword.put(:connection_manager, self())
       |> Keyword.update!(:connection_opts, fn in_connection_opts ->
-        if is_nil(shared_connection_opts), do: populate_connection_opts(in_connection_opts)
+        if is_nil(shared_connection_opts), do: in_connection_opts
       end)
 
     %{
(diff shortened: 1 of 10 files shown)


balegas:
Shall we expedite testing  this in staging Cloud. @msfstef we can check what users have pooled connection enabled and test with supabase to understand if this would work with the users that have pooled connections

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
```diff
@@ -0,0 +1,234 @@
...
+
+  def start_link(stack_id, opts) do
+    GenServer.start_link(__MODULE__, {stack_id, opts}, name: name(stack_id))
+  end
+
+  def validate(stack_id, db_connection, opts \\ []) do
+    with {:ok, pid} <- start_link(stack_id, opts),
+         result = GenServer.call(pid, {:validate, db_connection}, :infinity) do
+      Process.unlink(pid)
+      send(pid, :shutdown)
```
msfstef: nit: why not a `GenServer.stop`? also, if ConnMan abruptly terminates, it's not aware of this genserver to wait for it to shut down like it does for the rest

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -322,20 +324,16 @@ defmodule Electric.Connection.Manager do
     # If we see that both top-level connection opts and replication connection opts have been
     # initialized from the same kwlist, we'll skip the extra work and only perform the no-ssl
     # and ipv4 fallbacks once.
-    shared_connection_opts =
-      if in_connection_opts == Keyword.fetch!(in_replication_opts, :connection_opts) do
-        populate_connection_opts(in_connection_opts)
-      end
+    shared_connection_opts = Keyword.fetch!(in_replication_opts, :connection_opts)
```
alco: This completely changes the logic below. `shared_connection_opts` can no longer be `nil` so there's no point in running `is_nil(shared_connection_opts)` checks below.

And I don't see a reason to keep `shared_connection_opts` if there's now a separate `validate()` call for every type of connection opts. So even if `connection_opts == replication_opts[:connection_opts]`, we'll still run separate checks for each because of different `type` arguments passed into `validate()`.

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
```diff
@@ -0,0 +1,234 @@
...
+  def name(stack_id) when is_binary(stack_id) do
+    Electric.ProcessRegistry.name(stack_id, __MODULE__, System.unique_integer([:positive]))
+  end
+
+  def start_link(stack_id, opts) do
+    GenServer.start_link(__MODULE__, {stack_id, opts}, name: name(stack_id))
+  end
+
+  def validate(stack_id, db_connection, opts \\ []) do
+    with {:ok, pid} <- start_link(stack_id, opts),
```
msfstef: Why not start the `ConnectionResolver` in the supervisor before the connection manager and just use it rather than starting a link to it every time? This way we do not need to worry about its correct shutdown, and we can always assume we have it

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
```diff
@@ -0,0 +1,234 @@
...
+      {:ok, []}
+    end
+
+    def notify(_channel, _payload, _state) do
+      :ok
+    end
+  end
+
+  def name(stack_id) when is_binary(stack_id) do
+    Electric.ProcessRegistry.name(stack_id, __MODULE__, System.unique_integer([:positive]))
```
msfstef: If the resolver is started before the conn man we can give this a simpler unique name

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
```diff
@@ -0,0 +1,234 @@
...
+    end
+  end
+
+  @impl GenServer
+  def init({stack_id, opts}) do
+    # ignore exits from connection processes that fail to start due to
+    # connection errors
+    Process.flag(:trap_exit, true)
+
+    Process.set_label({:connection_discovery, opts[:stack_id]})
```
alco: Should be `:connection_resolver`.

alco (commented):
Looks good. There's some confusion about `shared_connection_options` in the new code, I don't think we need them anymore.

It'd be nice to have an actual test for exercising IPv6 fallback working in a real connection. Why can't we have that tested here, since we're already running real Postgres for tests? Is the DNS request for AAAA on localhost going to resolve to a valid IPv6 address?

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
alco: Having connection resolver linked to ConnMann should be sufficient for correct shutdown. Why would ConnMann wait for connection resolver?

balegas:
> My main worry is why are we not starting the connection resolver as a process in the supervisor before the connection manager, so we don't have to start link and kill it every time?

A comment from 10000 feet high: please make sure the exists don't get to Sentry.

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
alco: I supposed because this starts a Postgrex.SimpleConnection which ties the procecss lifetime to connection lifetime.

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
msfstef: I guess since we're starting them with unique names then it's fine - but we would otherwise need to because conn man is not a supervisor and thus in its `terminate` callback it waits for the processes it's linked to to go down, otherwise upon restarting it might run into `:already_started` errors - being linked doesn't guarantee that the linked processes will be dead by the time conn man restarts

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
msfstef: It's trapping its exits though ignoring them - so I don't think it's actually linked

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
```diff
@@ -0,0 +1,234 @@
...
+  use GenServer, restart: :temporary
+
+  require Logger
+
+  defmodule Connection do
+    @moduledoc false
+    @behaviour Postgrex.SimpleConnection
+
+    def init(_opts) do
+      {:ok, []}
```
msfstef, line 12:
```suggestion
      Logger. metadata(is_connection_process?: true, stack_id: opts[:stack_id])
      Electric.Telemetry.Sentry.set_tags_context(stack_id: opts[:stack_id])
      {:ok, []}
```

We need all processes we start that are going to be logging tons of errors to 1) be tagged with the appropriate stack ID and 2) to have the `is_connection_process?` tag so that we can ignore them/filter them appropriately.

magnetised:
> Looks good. There's some confusion about `shared_connection_options` in the new code, I don't think we need them anymore.

Yeah, in the interests of speed (on the clock with this one) I r tried not to go down too many rabbit holes. 

> It'd be nice to have an actual test for exercising IPv6 fallback working in a real connection. Why can't we have that tested here, since we're already running real Postgres for tests? Is the DNS request for AAAA on localhost going to resolve to a valid IPv6 address?

`localhost` resolves to `[::1]` for me and connects so I don't think this is entirely straightforward -- I have the mocked out test which exercises the path.

packages/sync-service/lib/electric/connection/manager.ex
magnetised: yeah, again i didn't delve too deeply into the shared_connection_opts -- just did a superficial pass to remove the obviously unnecessary calls

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
magnetised: it is linked so the `:shutdown` should bring the connection procs down too. we're just not crashing if the connection procs crash

magnetised:
@msfstef have restructured as a permanent sibling process of the connection manager as you suggest

@alco let's push some of the cleanup around shared opts to another pr -- don't want to get into the weeds here

there is a reliable failure of one of the integration tests looking at behaviour when there are insufficient connections available. must be something to do with the splitting the connections between the two pools interfering with the restart mechanism somehow. am investigating...

msfstef:
> there is a reliable failure of one of the integration tests looking at behaviour when there are insufficient connections available. must be something to do with the splitting the connections between the two pools interfering with the restart mechanism somehow. am investigating...

Yeah that seems to be failing on the pool PR (and the failure translates to this one) - the test manually fixes the number of available connections on postgres so that the pool fails to start expecting:

```
Database connection in connection_pool mode failed: sorry, too many clients already
```

I assume you need to match to a different log now that you have named the pools

packages/sync-service/lib/electric/connection/manager/connection_resolver.ex
msfstef: Yeah sorry I meant "linked" in the language sense of the lifetime being linked

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -316,34 +313,26 @@ defmodule Electric.Connection.Manager do
...
-        if is_nil(shared_connection_opts), do: populate_connection_opts(in_connection_opts)
-      end)
 
-    %{
-      state
-      | shared_connection_opts: shared_connection_opts,
-        connection_opts: connection_opts,
-        replication_opts: replication_opts
-    }
+    %{state | connection_opts: connection_opts, replication_opts: replication_opts}
```
msfstef, line 324: I like the cleaner separation, but we could just have a simpler approach to sharing things if unavailable - if replication and connection opts are the same, then just leave `connection_opts` as nil and fallback to the replication conn opts - and use a getter for that?

msfstef, line 324: given that all unit tests missed the fact that we were not using a pooled connection string it would be good if we can have some sort of manager_test that checks that the pooled connection is actually used if provided

alco: I agree. Let me write it up as a follow-up issue.

magnetised: there are validations on the config that fill both the connection opts and the replication connection opts are present. we could optimize the resolution process by only doing it once if both connection opts are identical though

alco: We now store validated connection opts in a separate map, so we now also need to take the `type` into account when deciding which connection opts are shared.

I think it should be a follow-up change.

magnetised: that is a really good point

alco (approved):

packages/sync-service/lib/electric/connection/manager.ex
msfstef: Yes good point - let's merge something that works and improve very minor unnecessary latency in a followup. Yaks will be shaved another day.

PR #3119, opened 2025-09-09
chore: Simplify `PublicationManager` to only track relations

msfstef:
Overdue change - simplifies `PublicationManager`.

No more complicated multi-index ref counting - just relations.

I've moved most things to use mapsets as they are appropriate for the set of relations we are tracking - we were doing lots of `Enum.uniq` calls in various places anyway. Not that the number of relations should ever be so large that this becomes a concern.

### DIFF
--- a/integration-tests/tests/not-owner-of-the-table.lux
+++ b/integration-tests/tests/not-owner-of-the-table.lux
@@ -62,15 +62,14 @@
   ??{"message":"Unable to create initial snapshot: must be owner of table items"}
 
 [shell electric]
-  ??[info] Configuring publication electric_publication_integration to drop [] tables, and add [] tables
 	??[info] Altering identity of public.items to FULL
   ?\[error\] Failed to configure publication: %Postgrex.Error\{message: nil, postgres: %\{code: :insufficient_privilege,.*message: "must be owner of table items"
 
 ## Wait for the schema reconciler to run and verify that the table is removed from the publication
 [shell electric]
   ??[info] Configuring publication electric_publication_integration to drop ["public.items"] tables, and add [] tables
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -107,7 +104,9 @@ defmodule Electric.Postgres.Configuration do
     else
       {:error,
        {:misconfigured_replica_identity,
-        Enum.map(bad_relations, fn {oid, relation, _ident} -> {oid, relation} end)}}
+        bad_relations
+        |> Enum.map(fn {oid, relation, _ident} -> {oid, relation} end)
+        |> MapSet.new()}}
```
magnetised:
```suggestion
        MapSet.new(bad_relations, fn {oid, relation, _ident} -> {oid, relation} end)}}
```

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -125,32 +124,39 @@ defmodule Electric.Postgres.Configuration do
...
-        |> list_changed_relations(known_relations(previous_relations, Map.keys(new_filters)))
+        |> list_changed_relations(known_relations(previous_relations, new_relations))
         |> Enum.map(&trim_changed_relation/1)
+        |> MapSet.new()
 
-      used_filters = Map.drop(new_filters, changed_relations)
+      used_relations = MapSet.difference(new_relations, changed_relations)
 
-      if changed_relations != [],
+      if MapSet.size(changed_relations) == 0,
```
magnetised:
```suggestion
      if MapSet.size(changed_relations) > 0,
```

Issue #2792, opened 2025-05-30
Feature request: Drop replication connection when there's no active shapes and there's not a bigger backlog of wal to process


KyleAMathews:
One annoying thing about electric right now is it keeps a constant connection to the database. Which for scale to zero services like neon and Prisma, they can never scale down which makes them a lot more expensive for low traffic projects.

What we can do is drop the connection when there's no shape subscriptions and periodically check the wal size and wake up the publication to process data when it grows to large (as well as when there's active shapes).

alco:
But wouldn't checking the WAL size periodically cause the database server to wake up anyway?

msfstef:
> But wouldn't checking the WAL size periodically cause the database server to wake up anyway?

I'm assuming that if this check is infrequent enough this could still work (e.g. once per day or every few hours or something like that, not sure how the "scale to zero" works for these various providers), but there won't be a persistent active connection at least?

alco:
@msfstef From the original description:

> periodically check the wal size and wake up the publication to process data when it grows to large

this is highly dependent on the write rate in the user database. Once an hour might not be enough in some cases. So setting some default periodicity might be inadequate for some users while fine for others.

Most managed Postgres services provided some sort of visibility into the replication lag, i.e. how much WAL is not getting discarded because it is claimed by a replication slot. Ideally, we would hook into that and wake Electric up whenever a threshold of WAL buildup is reached. But in practice I don't see that happen.

A more practical approach would be to maintain statistics of the incoming flow rate of transactions in our replication client and set a dynamic "wake up alert" based on that before closing the replication connection to let the database server scale down. But we'd still maintain the compute active without a more direct integration with the runtime platform that would have allowed the Electric process to shut down as well until woken up by an incoming HTTP request or an alarm.

KyleAMathews:
> not sure how the "scale to zero" works for these various providers

Neon scales down after 5 minutes of inactivity. Prisma Postgres is the only other scale-to-zero service I know of — they don't seem to say anywhere how long they wait but presumably it's pretty quick too.

So checking once every two hours would allow it to be inactive ~95% of the time assuming no other use.

We could also probably just check their API to see if the instance is awake — that wouldn't wake it up. And if it's down, we just let it keep sleeping.

[...]

PR #3126, opened 2025-09-10
feat(sync-service): Scale down database connections after a lull in streamed transactions

alco:
Part of #2792.

This PR adds an internal API for shutting connection manager down (which closes all database connections) and bringing it back up (which is done by restarting both the connection manager and the replication supervisor). A new configuration option called `ELECTRIC_DATABASE_SCALE_DOWN_ON_IDLE_TIMEOUT` is introduced to allow users to configure the idle timeout, after which Electric will shut down its database connections if it hasn't seen a transaction in that time interval.

When a shape request comes in, the database connections are scaled back up to allow logical streaming to resume and populate the shape log with new transactions.

The first optimization we can implement that would help delay the scaling up is to serve shapes from the shape logs without proactively scaling up and only do so if the request is asking for the bleeding edge, i.e. beyond the last offset available in the log.

~~I had to make `StatusMonitor` aware of the scaled-down state and say that the stack is `active` for Electric to be able to continue serving shape requests for shapes it already has cached. New shape requests will currently fail with snapshot timeout but we can make it so they cause the connection manager to "wake up" as the next iteration.~~

From our [brief discussion on Discord](https://discord.com/channels/933657521581858818/972052396605837342/1400968235968495648) from some time ago:

> So basically the first step could be adding a setting like ELECTRIC_LOGICAL_REPLICATION_IDLE_TIMEOUT. When Electric doesn't receive a transaction from PG for that amount of time, it closes its DB connections (the replication connection, the lock connection and its internal DB pool connections).
> 
> Currently closing DB connections also implies shutting down the HTTP API but we've been planning to decouple the "write to shape logs" and "read from shape logs" parts of Electric, so doing that could be a bonus subtask under this scale-to-zero task.

~~In this PR we keep the ability to serve shape requests for known shapes. We can do more later when we get more coupling between read and write facets of Electric.~~

~~I have deliberately *not* made any shape request "wake up" the connection manager because that would defeat the purpose of the feature. Imagine a dev leaving their demo app running overnight or publishing it online. With Electric's long-polling going, the replication connection would never go to sleep if it's woken up every time by an incoming shape request. The same reasoning applies if we start scaling down connections for sources in cloud.~~

Remaining things:
- [x] integration test that verifies shutting down Connection.Manager causes all Electric's database connection to close
- [x] configuration option for setting the idle timeout: after that amount of time has passed with no new transactions, Electric will scale down
- ~~[ ] configuration option for the WAL size check period: while scaled down, Electric will check for new transactions with this period to see if it needs to restore the replication connection~~

### DIFF
--- a/packages/sync-service/lib/electric/config.ex
+++ b/packages/sync-service/lib/electric/config.ex
@@ -48,6 +48,7 @@ defmodule Electric.Config do
     replication_slot_temporary?: false,
     replication_slot_temporary_random_name?: false,
     max_txn_size: 250 * 1024 * 1024,
+    scale_down_on_idle_timeout: 0,
     manual_table_publishing?: false,
     ## HTTP API
     # set enable_http_api: false to turn off the HTTP server totally
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -674,10 +674,31 @@ defmodule Electric.Connection.Manager do
     shutdown_or_reconnect(reason, pid_type, state)
   end
 
+  # The replication client exited because it hasn't streamed any new transactions for a while.
+  # This is a signal for all database connections to close and transition Electric into a
+  # scaled down mode.
+  def handle_info(
+        {:EXIT, pid, {:shutdown, {:connection_idle, time}}},
+        %State{replication_client_pid: pid, current_phase: :running} = state
+      ) do
+    time_s = System.convert_time_unit(time, :millisecond, :second)
+
+    Logger.warning(
+      "Closing all database connections after the replication stream has been idle for #{time_s} seconds"
+    )
+
+    # Perform shutdown in a task so that connection manager doesn't end up blocking the
+    # supervisor trying to shut it down.
+    Task.start(Electric.StackSupervisor, :shutdown_database_connections, [state.stack_id])
+
+    {:noreply, state}
+  end
+
   # The replication client exited after the connection setup has completed, it can be restarted
-  # independently of the lock connection and the DB pool. On the other hand, if any of the
-  # latter two shut down, Connection.Manager will itself terminate to be restarted by its
-  # supervisor in a clean state.
+  # independently of the lock connection and the DB pool.
+  #
+  # On the other hand, if any of the latter two shut down, Connection.Manager will itself
+  # terminate to be restarted by its supervisor in a clean state.
   def handle_info(
         {:EXIT, pid, reason},
         %State{replication_client_pid: pid, current_phase: :running} = state
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -524,21 +524,31 @@ defmodule Electric.Connection.Manager do
       )
     end
 
+    repl_sup_opts = [
+      stack_id: state.stack_id,
+      shape_cache_opts: shape_cache_opts,
+      pool_opts: state.pool_opts,
+      replication_opts: state.replication_opts,
+      tweaks: state.tweaks,
+      can_alter_publication?: state.can_alter_publication?,
+      manual_table_publishing?: state.manual_table_publishing?,
+      persistent_kv: state.persistent_kv
+    ]
+
     start_time = System.monotonic_time()
 
-    with {:error, reason} <-
-           Electric.Connection.Manager.Supervisor.start_replication_supervisor(
-             stack_id: state.stack_id,
-             shape_cache_opts: shape_cache_opts,
-             pool_opts: state.pool_opts,
-             replication_opts: state.replication_opts,
-             tweaks: state.tweaks,
-             can_alter_publication?: state.can_alter_publication?,
-             manual_table_publishing?: state.manual_table_publishing?,
-             persistent_kv: state.persistent_kv
-           ) do
-      Logger.error("Failed to start shape supervisor: #{inspect(reason)}")
-      exit(reason)
+    case Electric.Connection.Manager.Supervisor.start_replication_supervisor(repl_sup_opts) do
+      {:ok, _pid} ->
+        :ok
+
+      {:error, {:already_started, _pid}} ->
+        # Ask the existing shape cache to clean itself up and stuff.
+        # Update last processed lsn based on the the latest replication client state ???
+        :ok
+
+      {:error, reason} ->
+        Logger.error("Failed to start shape supervisor: #{inspect(reason)}")
+        exit(reason)
     end
 
     state = %{
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -968,6 +968,8 @@ defmodule Electric.Connection.Manager do
       {:replication_liveness_check, state.replication_client_pid}
     )
 
+    Logger.debug("Replication client started streaming")
+
     state = %{state | current_step: :streaming}
     {:noreply, state}
   end
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -44,7 +44,6 @@ defmodule Electric.Postgres.ReplicationClient do
       :display_settings,
       :txn_collector,
       :publication_owner?,
-      origin: "postgres",
       step: :disconnected,
       # Cache the end_lsn of the last processed Commit message to report it back to Postgres
       # on demand via standby status update messages -
@@ -71,7 +70,6 @@ defmodule Electric.Postgres.ReplicationClient do
             start_streaming?: boolean(),
             slot_name: String.t(),
             slot_temporary?: boolean(),
-            origin: String.t(),
             txn_collector: Collector.t(),
             step: Electric.Postgres.ReplicationClient.step(),
             display_settings: [String.t()],
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -44,6 +44,7 @@ defmodule Electric.Postgres.ReplicationClient do
       :display_settings,
       :txn_collector,
       :publication_owner?,
+      :max_idle_time,
       step: :disconnected,
       # Cache the end_lsn of the last processed Commit message to report it back to Postgres
       # on demand via standby status update messages -
@@ -56,6 +57,7 @@ defmodule Electric.Postgres.ReplicationClient do
       received_wal: 0,
       flushed_wal: 0,
       last_seen_txn_lsn: Lsn.from_integer(0),
+      last_seen_txn_timestamp: nil,
       flush_up_to_date?: true
     ]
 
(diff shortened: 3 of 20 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -524,21 +524,31 @@ defmodule Electric.Connection.Manager do
...
-           ) do
-      Logger.error("Failed to start shape supervisor: #{inspect(reason)}")
-      exit(reason)
+    case Electric.Connection.Manager.Supervisor.start_replication_supervisor(repl_sup_opts) do
+      {:ok, _pid} ->
+        :ok
+
+      {:error, {:already_started, _pid}} ->
+        # Ask the existing shape cache to clean itself up and stuff.
+        # Update last processed lsn based on the the latest replication client state ???
```
msfstef, line 546: imo we should avoid doing this - we should never end up starting the replication supervisor if it's already started, I think it makes more sense to exit and let things restart the right way?

I've made changes so that we never make the shape cache clean itself up, a full cleanup becomes a file system rename before starting the supervisor.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -664,10 +674,31 @@ defmodule Electric.Connection.Manager do
...
+    Task.start(Electric.StackSupervisor, :shutdown_database_connections, [state.stack_id])
+
+    {:noreply, state}
+  end
+
   # The replication client exited after the connection setup has completed, it can be restarted
-  # independently of the lock connection and the DB pool. On the other hand, if any of the
-  # latter two shut down, Connection.Manager will itself terminate to be restarted by its
-  # supervisor in a clean state.
+  # independently of the lock connection and the DB pool.
```
msfstef, line 698: this is outdated now as I have removed the indepndent replication client restarts

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -359,7 +382,14 @@ defmodule Electric.Postgres.ReplicationClient do
...
 
       {%Transaction{} = txn, txn_meta, %Collector{} = txn_collector} ->
-        state = %{state | txn_collector: txn_collector, last_seen_txn_lsn: txn.lsn}
+        timestamp = System.monotonic_time()
+
+        state = %{
+          state
+          | txn_collector: txn_collector,
+            last_seen_txn_lsn: txn.lsn,
+            last_seen_txn_timestamp: timestamp
```
msfstef, line 391:
```suggestion
            last_seen_txn_timestamp: System.monotonic_time()
```
nit: seems clearer than binding a generic `timestamp` var

packages/sync-service/lib/electric/stack_supervisor.ex
```diff
@@ -371,6 +372,8 @@ defmodule Electric.StackSupervisor do
              Keyword.take(monitor_opts, [:on_remove, :on_cleanup]),
              Keyword.take(shape_cache_opts, [:publication_manager])
            ])},
+          {Electric.ShapeCache.ShapeStatusOwner,
```
msfstef, line 375: Now that we have moved this here, how do we ensure it gets reset in case of full shape invalidation? Before we were 1. shutting down the replication supervisor which took in mem stuff out and 2. deleting storage which took away the backup (although the backup would fail anyway since it would fail the validation)

packages/sync-service/lib/electric/status_monitor.ex
```diff
@@ -27,25 +31,44 @@ defmodule Electric.StatusMonitor do
...
-        shape_log_collector_ready: {true, _}
-      } ->
-        :active
-
-      _ ->
+  @spec status(String.t(), [Atom.t()]) :: status()
+  def status(stack_id, opts \\ []) do
+    if db_scaled_down?(stack_id) do
+      if @wake_up_if_scaled_down_opt in opts do
+        Electric.StackSupervisor.restore_database_connections(stack_id)
```
msfstef, line 38: I personally don't like the idea of the status monitor having such an active role in modifying the system. It should never affect the state of the system itself, just report on its state.

packages/sync-service/lib/electric/status_monitor.ex
```diff
@@ -91,7 +114,7 @@ defmodule Electric.StatusMonitor do
   end
 
   def wait_until_active(stack_id, timeout) do
-    if status(stack_id) == :active do
+    if status(stack_id, [@wake_up_if_scaled_down_opt]) == :active do
```
msfstef, line 117: I don't think it makes sense to wake up the connection on _any_ request, and not on a `wait_until_active`.

I think we can immediately detect if a request is for a new shape (`offset=-1`) or an existing one - if it's for an existing one, we should be able to serve it without database connections - if it's a new shape request we can fire off a wake up signal, and instead make `wait_until_active` accept whether it's waiting for `active` or just `read_only` status?

packages/sync-service/lib/electric/status_monitor.ex
```diff
@@ -27,25 +31,44 @@ defmodule Electric.StatusMonitor do
...
+  @spec status(String.t(), [Atom.t()]) :: status()
+  def status(stack_id, opts \\ []) do
+    if db_scaled_down?(stack_id) do
+      if @wake_up_if_scaled_down_opt in opts do
+        Electric.StackSupervisor.restore_database_connections(stack_id)
         :starting
+      else
+        # If there's no explicit request to wake the DB connections up, we assume this call is
+        # coming from the health check endpoint and report the servicec as active.
+        :active
```
msfstef, line 43: I think rather than return `:starting` or `:active`, we should return a new status like `:read_only`  if the db is scaled down, and leave the wake up signal as a responsibility to some other process (see comment above). This way we can decide if requests need `active` or `read_only` and only wait for the appropriate status to serve it.

packages/sync-service/lib/electric/config.ex
```diff
@@ -48,6 +48,7 @@ defmodule Electric.Config do
     replication_slot_temporary?: false,
     replication_slot_temporary_random_name?: false,
     max_txn_size: 250 * 1024 * 1024,
+    scale_down_on_idle_timeout: 0,
```
msfstef, line 51: I'm not necessarily against this, but we should decide on a convention on things like this - sometimes we use a timeout of 0 as literally immediate timeout (e.g. for debounce timeouts), and in this case it means "never time out". Should we perhaps accept a different value for "do not scale down" - or have a separate flag for enabling scale down and also a timeout configuration?

packages/sync-service/lib/electric/stack_supervisor.ex
```diff
@@ -76,6 +76,7 @@ defmodule Electric.StackSupervisor do
                      slot_temporary?: [type: :boolean, default: false],
                      try_creating_publication?: [type: :boolean, default: true],
                      max_txn_size: [type: {:or, [:non_neg_integer, nil]}, default: nil],
+                     max_idle_time: [type: :non_neg_integer, default: 0],
```
msfstef, line 79: nit: whenever I try to add something to config, I find that we set defaults in multiple places - should we not recover the default here from `Electric.Config` rather than setting it explicitly? (same for the others, work for a separate PR, but wanted to raise this)

msfstef (changes requested):
Great work!

I have some concerns regarding:
1. maintaining the correct state in `ShapeStatus` in cases of shape invalidations
2. not waking up the stack on all requests (perhaps just new shapes)
3. not waking stack up from the `StatusMonitor`

I also share the concern about the WAL buildup, I understand that in cases like Neon waking up the database would effectively be what we're trying to avoid.

Also, if live shape requests coming in do not wake up the db connection as I suggested above, the only way this would work would be to have some way to check the wal buildup.

Given the above, rather than waking up on _every_ request, would it not be better if we did the following:
1. on new shape requests, recover db connections (and hold the requests until active)
2. on live shape requests, perform (throttled) checks of the WAL to decide whether we ned to wake up db conns
3. with a relatively long interval, check the WAL for build up

for 2 and 3 we can have a single WAL checker process that handles things

packages/sync-service/lib/electric/connection/manager.ex
alco: Yeah, I think this is leftover from my initial implementation where I tried to restart the connection manager alone without taking the replication supervisor down as well. I don't see how we can end up in this state with the current approach where the whole `Connection.Manager.Supervisor` is stopped in order to scale the system back up.

We will inevitably make it so the replication supervisor survives any connection manager restarts (as Rob's working on it) but then it should no longer be the connection manager's responsibility to start the replication supervisor at any point in the application lifetime.

packages/sync-service/lib/electric/config.ex
alco: This is a good point. I was basing my thinking on PGBouncer's config - https://www.pgbouncer.org/config.html#dangerous-timeouts. It doesn't seem as controversial to me because the setting defines an action that happens after the system has been idle for some time. We don't usually associate idleness with immediacy. And secondly, it doesn't make sense for Electric to scale down DB connections after every transaction, so, in my mind, this prevents any confusion of using the 0 value for disabling the scale down.

I think we should define a minimum allowed value for it, e.g. 1 minute.

msfstef: I agree that it makes sense and is a common pattern in configurations of this type, I'm just worried about inconsistent conventions causing confusion.

packages/sync-service/lib/electric/stack_supervisor.ex
alco: Addressed this by adding `reset()` function to ShapeStatus in the [new PR](https://github.com/electric-sql/electric/pull/3257/commits/46bb3bc13d38afc3d6f4500d9354d9d1a0e08e5a).

packages/sync-service/lib/electric/status_monitor.ex
alco: I have moved the stop/restart logic into a new module called [`Electric.Connection.Restarter`](https://github.com/electric-sql/electric/pull/3257/commits/e51544e0f05619d85748b1837c817c4c3c051041#diff-50611e1444fb61913be1715c567e1bd4e4e93197540dd407b4a551bdaa24f2fa).

packages/sync-service/lib/electric/status_monitor.ex
alco: Addressed this in the description of the [new PR](https://github.com/electric-sql/electric/pull/3257#issue-3493419738).

packages/sync-service/lib/electric/status_monitor.ex
alco: I've added another status value `:db_conn_sleeping` which is used to determine whether the stack is active or in the scaled down mode.

packages/sync-service/lib/electric/config.ex
alco: I have renamed the configuration option to `ELECTRIC_REPLICATION_IDLE_TIMEOUT` and this fields to `replication_idle_timeout`. I don't see a big issue in using 0 to mean "disabled" here. Would hate to introduce a second flag just to work around this.

packages/sync-service/lib/electric/stack_supervisor.ex
alco: Our config validation, passing and defaulting is a mess. I've updated this a bit by calling `Electric.Config.default()`. But repeated default values are a common sight in our codebase regardless.

alco:
Closing in favor of https://github.com/electric-sql/electric/pull/3257

PR #3128, opened 2025-09-10
[TMP]: Investigating NFS performance

msfstef:
Making changes to try out performance on NFS storage like AWS EFS.


msfstef:
Closing - issue was with EFS on AWS not doing attribute caching on Fargate, and our file operations do many metadata access operations (each open comes with a `stat` as well).

PR #3129, opened 2025-09-11
Replace Caddy with `@electric-sql/vite-plugin-trusted-https`.

thruflo:
Adds `vite-plugin-trusted-https` package and ported the TanStack DB web starter to use it, as an initial example. The idea is no faffing with system packages, no sudo needed and uses Vite's built in HTTPS/HTTP2 support.

Example usage:
https://github.com/electric-sql/electric/blob/thruflo/replace-caddy-with-node-http2/examples/tanstack-db-web-starter/vite.config.ts

This needs a fair bit of manual testing to make sure it actually works before releasing into the wild. If it does work, we can use it to swap out Caddy from most examples.

It also needs the `workspace:*` taken out of the starter dept once the plugin is published.

Also made some minor changes to the copy and content in the starter README: https://github.com/electric-sql/electric/blob/thruflo/replace-caddy-with-node-http2/examples/tanstack-db-web-starter/README.md


thruflo:
This is superseded by the far simpler #3221.

PR #3131, opened 2025-09-11
fix(sync-service): Handle shutdown of blocked connection manager

magnetised:
Connection manager waiting for connection resolution to finish is not receiving signals from the supervisor, so a terminate call while blocked in resolving an unresponsive host will only succeed once the postgres connection times out.

This solves that by making the shutdown behaviour of the connection  resolver process is `brutal_kill` so the call from the conn manager exits and unblocks and the supervisor shutdown continues normally.

This requires a re-ordering of the child startup and so introduces a new call back to start the conn manager startup process once the resolver is up

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -845,8 +845,6 @@ defmodule Electric.Connection.Manager do
       state.replication_client_pid
     )
 
-    dispatch_stack_event(:replication_client_ready, state)
-
     state = %{state | replication_client_blocked_by_pending_transaction?: false}
 
     case phase do
--- a/packages/sync-service/test/electric/connection/manager_test.exs
+++ b/packages/sync-service/test/electric/connection/manager_test.exs
@@ -237,9 +237,6 @@ defmodule Electric.Connection.ConnectionManagerTest do
     test "failure to get pooled connection results in retries", ctx do
       %{stack_id: stack_id} = ctx
 
-      # make sure we've reached the connection pool stage
-      assert_receive {:stack_status, _, :replication_client_ready}, 1000
-
       ref = Process.monitor(GenServer.whereis(Electric.Connection.Manager.name(stack_id)))
 
       refute_receive {:DOWN, ^ref, :process, _pid, _reason}, 1000
@@ -318,7 +315,6 @@ defmodule Electric.Connection.ConnectionManagerTest do
       )
 
       assert_receive {:stack_status, _, :connection_lock_acquired}, 1000
-      assert_receive {:stack_status, _, :replication_client_ready}, 1000
 
       {time, log} =
         :timer.tc(


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -996,6 +1006,8 @@ defmodule Electric.Connection.Manager do
   defp handle_connection_pool_ready(state) do
     case state.pool_pids do
       %{admin: {pid1, true}, snapshot: {pid2, true}} when is_pid(pid1) and is_pid(pid2) ->
+        dispatch_stack_event(:connection_succeeded, state)
```
alco: This is too generic. And it will confuse customers who get these stack events in their event log.

By the way, I've just noticed that this `handle_connection_pool_ready()` function obscures the fact that a `{:continue, ...}` value is returned from the GenServer callback that calls the helper function. I'm strongly for keeping the return value of `handle_*` callbacks inside the callback itself. Adding indirection just puts more cognitive load on us poor devs.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -815,6 +845,8 @@ defmodule Electric.Connection.Manager do
       state.replication_client_pid
     )
 
+    dispatch_stack_event(:replication_client_ready, state)
```
msfstef: Is this necessary? We are not emitting events for every internal process we start, I'd be reluctant to introduce more stack events now before we revisit this more thoroughly

packages/sync-service/test/electric/connection/manager_test.exs
```diff
@@ -255,6 +286,55 @@ defmodule Electric.Connection.ConnectionManagerTest do
...
+          fn ->
+            ExUnit.CaptureLog.capture_log(fn ->
+              :ok = Supervisor.stop(Connection.Manager.Supervisor.name(stack_id: ctx.stack_id))
+            end)
+          end,
+          :millisecond
+        )
+
+      assert time < 1000
+      refute log =~ "Electric.DBConnection unknown error"
```
msfstef, line 330: Could we generalize this to no error logs in general?

magnetised: no, because the dbconnection processes error as they're unable to connect.

packages/sync-service/lib/electric/connection/manager.ex
magnetised: i had this feeling that extra messages would provoke dissent... i think I did need a little more info for initial versions of the test but actually maybe I can drop it if it's controversial. I personally don't see why giving people more progress information is bad

msfstef: I don't think it's bad but currently we are forwarding all of them in cloud rather than picking a subset to store, and imo once we introduce them they are hard to remove cause we start building automations on them, so I'd rather be cautious about adding things that might end up being used or relied on

packages/sync-service/test/electric/connection/manager_test.exs
msfstef: Ah fair - ok yeah those errors should get filtered by virtue of the metadata we've added - just making sure we don't have any other stray errors that end up flooding sentry

PR #3134, opened 2025-09-12
Log insufficient_privilege errors at the warning level

alco:
Logging them at the error level spams our sentry with alerts that aren't actionable.


alco:
@magnetised I wrote one user affected by this but never got a response.

The whole issue might have been constrainted to the Shape Explorer creating shapes for any table you click on. It has since been changed to only show tables that the user is the owner of.

alco:
This hasn't been a problem after we have passed the phase where those issues were first seen on Sentry an caused repeated Rootly alerts. And I don't like this implementation, it needs a revisit. Closing this for now.

PR #2584, opened 2025-04-10
docs: sketch out walkthrough tutorials concept.

thruflo:
https://deploy-preview-2584--electric-next.netlify.app/docs/walkthroughs/

PR #3138, opened 2025-09-12
docs: sync stacks page.

thruflo:
Was starting on the tutorial and went back and checked the previous "Sync busters!" concept in https://github.com/electric-sql/electric/pull/2584 in case it had anything useful to keep / port over. Played with it and ended up converting into this "Sync stacks" page. 

The idea is to have a page that emphasises "composable" + "recommend TanStack DB" and has the common content we often relay about choice of stack / options / what's best for app dev, how do TanStack DB and PGlite relate etc.

=> https://deploy-preview-3138--electric-next.netlify.app/docs/stacks

I would envisage publishing alongside the new Quickstart and Tutorial, below the Tutorial in the left hand nav. Sharing early / separately because (a) it's largely self-contained and (b) for feedback around whether this page adds value/clarity or reduces it.


watadarkstar:
@thruflo Some feedback if I may:

- In the "Sync stacks" section, are you referring to the entire stack or just a specific part of it? I’m still a bit confused — is it TanStack DB OR Phoenix OR PgLite OR Yjs, or is it meant to be an AND in that section?
- I personally think the "Sync stacks" sub-section creates more confusion and can be removed. Or maybe rename this section to something more clear, but I can't think of a good name for this section yet. However, I feel like the page is better without this section and stands on its own.
- I also would love to know which technology is meant for which part of the stack i.e. TanstackDB is for the client. PGLite is for the database? Pheonix is for the backend (but also what part, the proxy? the sync engine?)
- Also may want to name drop Elecrtic Cloud for the backend sync engine somewhere on this page

Other thoughts:

- I'd also love a section on recommending a cloud provider for the database i.e. Supabase or Railway but also being clear that any PG database would work as long as it has replica permissions, etc. I personally struggled with Railway and ElectricSQL 
- Also a section on a recommendation for the Proxy i.e. Next.js or Express but also being clear that any HTTP endpoint would work (as far as I understand)

Just suggestions, thanks again for the great work your doing! I ❤️ ElectricSQL!

Screenshot 2025-09-14 at 10 51 21 AM

thruflo:
Thanks @watadarkstar have tried to address (without entirely re-engineering -- I've left the sync stacks summary section there but have tried to make it clearer what's in the stacks so it's less confusing).

PR #3141, opened 2025-09-16
feat(sync-service): Improve performance of LRU shape expiry

robacourt:
Turbo.ai have been seeing 2-8s latency on requests when the number of shapes exceeds the maximum number of shapes and LRU expiry kicks in. This is due to the shape cache process being blocked while the expiry happens, which can take a few seconds.

This PR moves expiry out of the ShapeCache process and into a new ExpiryManager process. When a new shape is added, the ShapeCache calls `ExpiryManager.notify_new_shape_added/1` which is non-blocking and debounced, and if max_shapes has been set, triggers finding the LRU shapes and expiring them.

### DIFF
--- a/packages/sync-service/lib/electric/replication/supervisor.ex
+++ b/packages/sync-service/lib/electric/replication/supervisor.ex
@@ -47,7 +47,8 @@ defmodule Electric.Replication.Supervisor do
       publication_manager,
       consumer_supervisor,
       shape_cache,
-      schema_reconciler
+      schema_reconciler,
+      {Electric.ShapeCache.ExpiryManager, stack_id: opts[:stack_id]}
     ]
 
     Supervisor.init(children, strategy: :one_for_all)
--- a/packages/sync-service/lib/electric/replication/supervisor.ex
+++ b/packages/sync-service/lib/electric/replication/supervisor.ex
@@ -37,6 +37,7 @@ defmodule Electric.Replication.Supervisor do
     consumer_supervisor = Keyword.fetch!(opts, :consumer_supervisor)
     shape_cache = Keyword.fetch!(opts, :shape_cache)
     schema_reconciler = Keyword.fetch!(opts, :schema_reconciler)
+    expiry_manager = Keyword.fetch!(opts, :expiry_manager)
     stack_id = Keyword.fetch!(opts, :stack_id)
 
     children = [
@@ -48,7 +49,7 @@ defmodule Electric.Replication.Supervisor do
       consumer_supervisor,
       shape_cache,
       schema_reconciler,
-      {Electric.ShapeCache.ExpiryManager, stack_id: opts[:stack_id]}
+      expiry_manager
     ]
 
     Supervisor.init(children, strategy: :one_for_all)
--- a/packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
+++ b/packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
@@ -3,6 +3,7 @@ defmodule Electric.ShapeCache.ExpiryManager do
 
   @schema NimbleOptions.new!(
             stack_id: [type: :string, required: true],
+            shape_status: [type: :mod_arg, required: true],
             max_shapes: [type: {:or, [:non_neg_integer, nil]}, default: nil]
           )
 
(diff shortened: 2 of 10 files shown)


balegas:
This sounds good. Deleting the files will be handled at `ExpiryShapeManager` speed and off-the-critical path.

packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
```diff
@@ -0,0 +1,142 @@
...
+      OpenTelemetry.with_span(
+        "expiry_manager.expire_shapes",
+        [
+          max_shapes: max_shapes,
+          shape_count: shape_count,
+          number_to_expire: number_to_expire
+        ],
+        fn ->
+          shapes_to_expire
+          |> Enum.each(fn shape -> expire_shape(shape, state) end)
```
alco, line 87: Is it safe to replace this with `Task.async_stream()`? It is worth expiring shapes in parallel with max concurrency set to e.g. 10-20.

alco (approved):
Lovely !

packages/sync-service/lib/electric/replication/supervisor.ex
```diff
@@ -47,7 +48,8 @@ defmodule Electric.Replication.Supervisor do
       publication_manager,
       consumer_supervisor,
       shape_cache,
-      schema_reconciler
+      schema_reconciler,
+      expiry_manager
```
alco: Forgot to mention one thing: is there any harm in moving the startup of this process to happen before schema_cache?

SchemaCache restores shapes inside its `init()` callback and then notifes the shape log collector, at which point the stack becomes ready and shape requests start being served. We might get a new shape request when we're already at max_shapes and the expiry manager hasn't started yet.

balegas:
A couple of comments after discussion:
- we want to add a bit of headroom whenever we clear shapes. So instead of bringing back the number of shapes to the max limit. We'll add a second variable which is the target number of shapes after a cleanup
- In the telemetry we want to know the number of shapes we've cleaned up and the duration of a cleaning round. This will allow us to analyze if things are slow because they are slow, or just because we're deleting lots of shapes

We had a discussion about we're sending a message to the ExpiryManager on every new shape to check if we need to cleanup vs. making the cleanup run periodically. There are trade-offs, but we expect that the number of new shapes not to be that large that would become an issue. It does seem that we should have a reasonable debouncing window for checking number of shapes if the counting of shapes is a bit compute intensive.

alco:
@balegas @robacourt 

> It does seem that we should have a reasonable debouncing window for checking number of shapes if the counting of shapes is a bit compute intensive.

If ExpiryManager gets a message for each new shape, let's make it keep the running count of shapes in its state, just for its internal use. Then it won't need to go through the list of all shapes ever after doing that once during initialization.

packages/sync-service/lib/electric/replication/supervisor.ex
robacourt: done

balegas:
I thought about that too, but I was not sure if we deleted shapes in any other way like 409s etc, that would not go through this code path

packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
robacourt: will look at in another PR

robacourt:
> * we want to add a bit of headroom whenever we clear shapes. So instead of bringing back the number of shapes to the max limit. We'll add a second variable which is the target number of shapes after a cleanup

I've added `ELECTRIC_EXPERIMENTAL_EXPIRY_RATIO` with a default of 0.2

> * In the telemetry we want to know the number of shapes we've cleaned up and the duration of a cleaning round. This will allow us to analyze if things are slow because they are slow, or just because we're deleting lots of shapes

yup, that's in there

> We had a discussion about we're sending a message to the ExpiryManager on every new shape to check if we need to cleanup vs. making the cleanup run periodically. There are trade-offs, but we expect that the number of new shapes not to be that large that would become an issue. It does seem that we should have a reasonable debouncing window for checking number of shapes if the counting of shapes is a bit compute intensive.

I've added a `recheck_delay_ms` of 1 second so the count will not be checked more than once a second. This is useful on top of debouncing when under the number of shapes is less than the max_shapes.

PR #3146, opened 2025-09-17
feat: Simplify purging of shapes in case of timeline or slot change

msfstef:
This is in response to slow purge observed by @alco 

- In case of timeline change or slot change, simply delete the entire shape directory
- If publication manager refresh fails during shape cache init, synchronously wait for all shapes to be cleaned up rather than async purging
- Do not do early "deregistrations" of shapes from shape status as it might mess with operations happening on the consumer. The only ones deregistering the shape should be 1) the consumer dying or 2) the cleanup task supervisor that asserts the consumer is dead before doing so. (@magnetised please confirm cause you added the early deregistration logic)


packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
```diff
@@ -319,6 +319,11 @@ defmodule Electric.ShapeCache.InMemoryStorage do
         do: ignoring_exceptions(fn -> :ets.delete(table) end, ArgumentError)
   end
 
+  @impl Electric.ShapeCache.Storage
+  def cleanup_all!(_opts) do
+    :ok
```
magnetised, line 331: really? shouldn't we actually empty the ets table? seems like not doing that is bad and will bite us. also shouldn't some test fail if we don't empty the storage..?

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -269,11 +263,11 @@ defmodule Electric.ShapeCache do
       :ok = publication_manager.refresh_publication(publication_manager_opts)
     rescue
       error ->
-        purge_all_shapes(state)
+        clean_up_all_shapes(state)
         reraise error, __STACKTRACE__
     catch
       :exit, reason ->
-        purge_all_shapes(state)
+        clean_up_all_shapes(state)
```
magnetised, line 270: clean_up_all_shapes is very slow  - this is the call that was blocking restarts. on my machine it was taking 9s to clean 5,000 shapes, and 194s to clean 20,000. couldn't we do the equivalent here - just emptying the ets and deleting the storage dir (after stopping the consumers)?

magnetised (commented):
I'm worried about still relying on clean_all_shapes in the shape cache since it's blocking and slow. is there some equivalent of the `mv` call we do on the filesystem? like create a new shape status ets then just send the existing data to some background process to handle termination? i'm making things up here I know...

perhaps this isn't the pr to do it in but the thing you did in the connection manager - making the shape cleanup O(1) - we need to do in the shape cache.

packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
msfstef: This is supposed to be called when no consumers are alive (as it touches the data they actively use, especially if files were used), and with in memory storage if no consumers are alive then no data is present since there is no permanence, and the ets tables are per-shape and linked to the consumer lifetimes.

The only way around this would be to have a stack wide ETS that keeps track of what ETS tables we have started so that I can delete all of them. Happy to do that but figured that we should never need that.

packages/sync-service/lib/electric/shape_cache.ex
msfstef: You're right - I'll change the "clean up all shapes" call to shut down consumers, clean shape status, and delete the directory.

packages/sync-service/lib/electric/shape_cache/in_memory_storage.ex
magnetised: The semantics of that aren't guaranteed - how do we know that that call will only be used before storage has started and that the ets is definitely blank. "no consumers alive" doesn't mean that no consumers have ever lived.

msfstef: Added a cleanup method for in mem storage - that being said, the same argument about not knowing how it will be used applies to `cleanup!` which we already use. We make assumptions about how storage is going to be used already.

packages/sync-service/lib/electric/shape_cache.ex
msfstef: After discussing with @magnetised we decided that this might be worth investigating in a subsequent PR.

msfstef:
> I think we'll want to explore the idea of adding a generation number to our persistent storage, such that creating a new generation should let Electric start up in a clean state, while all cleanup of old data can be done in background. This is kind of a similar idea to Postgres' timeline ID.

The alternative is to do something similar to what we do on cloud - we have a "trash" dir that we move the directory to instead of deleting it.

I will open a followup to this for this sort of deletion.

PR #3149, opened 2025-09-18
feat: Simplify `PublicationManager` to not require full shape for removal

msfstef:
Now that we have simplified the publication manager, it lends itself to easier cleanups without needing the full shape.

This is part of the work of simplifying, unifying, and speeding up cleanups.

By making `remove_shape` only require the shape handle it simplifies the required information for cleaning up shapes - we only need the handle.


alco (approved):
:+1: :+1:

PR #3152, opened 2025-09-22
fix: Fix premature replies to concurrent publication updates for same shape handle

msfstef:
Working on simplifying publication manager and noticed that we had incorrect behaviour for concurrent updates to the same shape handle - requests after the first one would immediately reply without waiting for the publication to actually update.

Unlikely we're seeing this in production as we generally don't call ops on the same shape handle multiple times, perhaps only removals which don't particularly matter (not as much as adding a shape that is important for keeping track of the replication stream)

### DIFF
--- a/.changeset/rude-cars-beam.md
+++ b/.changeset/rude-cars-beam.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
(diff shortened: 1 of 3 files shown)


alco (approved):
Brilliant!

PR #3155, opened 2025-09-22
feat: Simplify `Connection.Manager` restart logic in case of replication stream failure

msfstef:
So far we've been independently restarting the replication client without killing any other parts of the infrastructure, but only doing that for the replication client and not for the other connection processes.

During that independent restart, we are not currently serving any requests (serving 503s) as we have not decoupled the storage/serving of existing data from connection processes yet, so this change should not affect downtime.

In fact, the problem with live and independente restart of the replication client is that it was calling `ShapCache.clean_all_shapes` in case of a slot invalidation, which was extremely slow and inefficient, causing timeouts and other issues as it scrambles to kill a large number of shapes.

It was also extra complexity in the connection manager state machine that did not offer many benefits in its current state. The main benefit to be gained is to decouple the shape processing stack ("shapes supervisor") from the connection processes uptime.

I've modified `shutdown_and_reconnect` to handle proper restarts and config errors during runtime as well, it's worth giving extra care to this change as it was surprisingly easy to alter these things without breaking any tests.

Main drawback of this change is that we are no longer doing exponential backoff retries in case of a replication stream failure, as it is no longer during the connection setup - that being said any connection error worth retrying with a backoff will cause a restart and subsequent retry backoff loop in the next setup phase.

### DIFF
--- a/integration-tests/tests/recreation-of-replication-slot.lux
+++ b/integration-tests/tests/recreation-of-replication-slot.lux
@@ -46,7 +46,7 @@
 
   ??[error] :gen_statem {:"Electric.ProcessRegistry:single_stack", {Electric.Postgres.ReplicationClient, nil}} terminating
   ??** (Postgrex.Error) ERROR 42704 (undefined_object) publication "electric_publication_integration" does not exist
-  ??[warning] Electric.Connection.Manager is restarting after it has encountered an error in replication mode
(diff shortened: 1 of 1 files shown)


alco (approved):
Lovely

robacourt (approved):
Nice

PR #3156, opened 2025-09-22
feat: add timeout to first data on snapshots to avoid long-running queries

icehaunter:
Closes #856 by adding a 30s timeout to first data batch - otherwise query is considered too slow for snapshots.

Timeout is counted only after we've made a connection and started a transaction - pool queueing and other waits are not included. We're waiting until first data (or end-of-data in case of absence) for those 30s for the query to be considered "good enough". Probably can make that timer even lower. The query itself is still unbounded - if the snapshot is multiple GBs that are being copied, we'll not interrupt it in this PR.


icehaunter:
@msfstef opinions on configurability?

packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
```diff
@@ -136,13 +139,106 @@ defmodule Electric.Shapes.Consumer.Snapshotter do
...
+          start_time + timeout_to_first_data,
+          abs: true
+        )
+
+        receive do
+          {:stream_timeout, task_pid} ->
+            Process.demonitor(ref, [:flush])
+            Task.Supervisor.terminate_child(supervisor, task_pid)
+
+            raise SnapshotError,
```
msfstef: nit: can you create a utility in the `SnapshotError` module for this error just so we have a central point where it's easy to see all types of this error?

packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
```diff
@@ -136,13 +139,106 @@ defmodule Electric.Shapes.Consumer.Snapshotter do
...
+
+      {^ref, result} ->
+        Process.demonitor(ref, [:flush])
+        handle_task_exit(result)
+
+      {:DOWN, ^ref, :process, _, reason} ->
+        case reason do
+          {err, stacktrace} when is_exception(err) -> reraise err, stacktrace
+          reason -> exit(reason)
+        end
```
msfstef, line 220: supernit: this duplication doesn't look particularly pretty, would a small private macro like `handle_premature_task_end" or smth like that be an overkill for this?

msfstef:
Forgot to comment on configurability - I think perhaps we can hold off on exposing an env var for this, 30 seconds to first data once a connection is already established I think is already quite conservative

packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
icehaunter: I'd rather this ugly but explicit for now

PR #3158, opened 2025-09-23
feat: Faster large cleanups using renames and background batch deletes

msfstef:
For large numbers of shapes, deleting all of them takes a long while (doing an Elixir/Erlang rm_rf fo 10k shapes is in the order of tens of seconds) - `rm_rf`ing recursively lists into directories and deletes things which for large directories like deleting all shapes or all files for a shape can be slow.

Using a different `rm_rf` implementation like the regular util is faster (twice as fast) but messes with the BEAM and we would have to use ports/nifs/dirty schedulers which I want to avoid for the time being for the purpose of keeping things as portable and embeddable as possible. Also Erlang's `rm_rf` implementation is safe and should play nice with schedulers if done in the background without blocking our critical processes.

I've implemented an `AsyncDeleter` process that lives in the top level of our stack - all it does is it maintains a temporary trash directory (stack scoped), it has a public API called `delete` which all it does is:
1. Ensure trash directory exists with `mkdir_p!` to guarantee move will work
2. Attempt to rename desired path to delete into trash directory (with random suffix to guarantee uniqueness)
  2a. If file already does not exist when attempting to rename it, ignore it and move on
3. If rename succeeds, cast a message to `AsyncDeleter` to schedule a cleanup
4. `AsyncDeleter` performs a cleanup of the entire stack trash directory every `cleanup_interval_ms`

I've added some tests, including one to ensure concurrent deletes and cleanups work, since the renames happen in the calling process, and the cleanup can occur at any time.

For now I've only made `cleanup!` and `cleanup_all!` storage methods use the async deleter, which are the main ones that are slow, but we can potentially use it for more deletes in our system (compaction does a lot of deletes for example).

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
@@ -612,7 +612,7 @@ defmodule Electric.ShapeCache.PureFileStorage do
 
     if on_disk_version != opts.version or not snapshot_complete?(opts) or
          is_nil(read_metadata!(opts, :pg_snapshot)),
-       do: rm_rf!(opts.data_dir)
+       do: cleanup!(opts)
 
     create_directories!(opts)
     write_metadata!(opts, :version, @version)


icehaunter:
Erlang's rm -rf blocks the file server genserver unfortunately which is a bit of a bottleneck, but the pure storage shouldn't be going through that fileserver anyway

msfstef:
> but the pure storage shouldn't be going through that fileserver anyway

All of our `File.open` calls in our storage are opened with `:raw` so that should be fine - however we do use other operations like `File.rename` for atomic metadata updates which will go through the file server I assume. That might run into a bottleneck.

Compaction also uses several `File` operations that would also run into the bottleneck.

Ultimately I think the better approach is to have a sidecar process entirely outside of the BEAM for cleaning up the trash directory, but we can do that as followup work as I'm worried how that'll play with portability.

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -0,0 +1,153 @@
...
+  end
+
+  defp random_suffix, do: System.unique_integer([:positive]) |> to_string()
+
+  @impl true
+  def handle_info(:perform_delete, state) do
+    {:noreply, do_cleanup(state)}
+  end
+
+  defp trash_dir(stack_id), do: Path.join(System.tmp_dir!(), "#{@trash_dir_base}_#{stack_id}")
```
alco: Wouldn't `#{@trash_dir_base}/#{stack_id}` be cleaner in the sense of not polluting the systemp tmp dir too much?

msfstef: yep indeed changed it already!

PR #3159, opened 2025-09-23
feat: allow snapshot-less shapes & subset snapshots

icehaunter:
Adds:
- `log` query parameter, values `full | changes_only`. Latter adds the new behaviour where we're not asking for an initial snapshot at all from Postgres
- `subset__` parameter group (`where/order_by/limit/offset`) that, if present, alters the response form to be
```typescript
{
  metadata: SnapshotMetadata,
  data: ChangeMessage
}
 
 export type SnapshotMetadata = {
  xmin: number
  xmax: number
  xip: number[]
  /** Random number that's reflected in the `snapshot_mark` header on the snapshot items. */
  snapshot_mark: number
  database_lsn: string
}
```
- `subset__where` is validated much the same as normal where clause, with an additional assertion on lack of subqueries
- Support for `offset=now` on the server to skip historical data

Where data contains the `snapshot_mark` in the headers if the client needs to match back the results after passing them through some pipeline

Updates the client:
- Adds a `log` parameter, values `lazy | eager`
- Addes a `requestSnapshot` function that returns data/metadata described above AND injects the data into the shape stream
- Shape stream then drops any changes deemed to be already seen via the snapshot according to xmin/xmax/xip logic

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -5,9 +5,11 @@ import {
   Row,
   MaybePromise,
   GetExtensions,
+  ChangeMessage,
+  SnapshotMetadata,
 } from './types'
 import { MessageParser, Parser, TransformFunction } from './parser'
-import { getOffset, isUpToDateMessage } from './helpers'
+import { getOffset, isUpToDateMessage, isChangeMessage } from './helpers'
 import {
   FetchError,
   FetchBackoffAbortError,
@@ -43,12 +45,19 @@ import {
   PAUSE_STREAM,
   EXPERIMENTAL_LIVE_SSE_QUERY_PARAM,
   ELECTRIC_PROTOCOL_QUERY_PARAMS,
+  LOG_MODE_QUERY_PARAM,
+  SUBSET_PARAM_WHERE,
+  SUBSET_PARAM_WHERE_PARAMS,
+  SUBSET_PARAM_LIMIT,
+  SUBSET_PARAM_OFFSET,
+  SUBSET_PARAM_ORDER_BY,
 } from './constants'
 import {
   EventSourceMessage,
   fetchEventSource,
 } from '@microsoft/fetch-event-source'
 import { expiredShapesCache } from './expired-shapes-cache'
+import { SnapshotTracker } from './snapshot-tracker'
 
 const RESERVED_PARAMS: Set<ReservedParamKeys> = new Set([
   LIVE_CACHE_BUSTER_QUERY_PARAM,
@@ -58,6 +67,7 @@ const RESERVED_PARAMS: Set<ReservedParamKeys> = new Set([
 ])
 
 type Replica = `full` | `default`
+export type LogMode = `changes_only` | `full`
 
 /**
  * PostgreSQL-specific shape parameters that can be provided externally
@@ -114,11 +124,20 @@ export type ExternalParamsRecord<T extends Row<unknown> = Row> = {
   [K in string]: ParamValue | undefined
 } & Partial<PostgresParams<T>> & { [K in ReservedParamKeys]?: never }
 
+export type SubsetParams = {
+  where?: string
+  params?: Record<string, string>
+  limit?: number
+  offset?: number
+  orderBy?: string
+}
+
 type ReservedParamKeys =
   | typeof LIVE_CACHE_BUSTER_QUERY_PARAM
   | typeof SHAPE_HANDLE_QUERY_PARAM
   | typeof LIVE_QUERY_PARAM
   | typeof OFFSET_QUERY_PARAM
+  | `subset__${string}`
 
 /**
  * External headers type - what users provide.
@@ -258,6 +277,11 @@ export interface ShapeStreamOptions<T = never> {
    */
   experimentalLiveSse?: boolean
 
+  /**
+   * Initial data loading mode
+   */
+  mode?: LogMode
+
   signal?: AbortSignal
   fetchClient?: typeof fetch
   backoffOptions?: BackoffOptions
@@ -293,8 +317,20 @@ export interface ShapeStreamInterface<T extends Row<unknown> = Row> {
   lastOffset: Offset
   shapeHandle?: string
   error?: unknown
+  mode: LogMode
 
   forceDisconnectAndRefresh(): Promise<void>
+
+  requestSnapshot(params: {
+    where?: string
+    params?: Record<string, string>
+    limit: number
+    offset?: number
+    orderBy: string
+  }): Promise<{
+    metadata: SnapshotMetadata
+    data: Array<Message<T>>
+  }>
 }
 
 /**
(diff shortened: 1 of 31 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -58,6 +67,7 @@ const RESERVED_PARAMS: Set<ReservedParamKeys> = new Set([
 ])
 
 type Replica = `full` | `default`
+type LogMode = `eager` | `lazy`
```
KyleAMathews: I actually like the names you picked for the server better — "mode" as an option is pretty opaque and eager/lazy describe more the benefit you get from picking them vs. what they actually are. And it'd be nice to have the client option translate directly to the API call to the server internals.

So `logMode: "changes_only" | "full"` would work

packages/typescript-client/src/snapshot-tracker.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * Remove a snapshot from tracking
+   */
+  removeSnapshot(snapshotMark: number): void {
+    this.activeSnapshots.delete(snapshotMark)
+  }
+
+  /**
+   * Check if a change message should be filtered based on active snapshots
```
KyleAMathews:
```suggestion
   * Check if a change message should be filtered because its already in an active snapshot
```

packages/typescript-client/src/client.ts
```diff
@@ -295,6 +319,17 @@ export interface ShapeStreamInterface<T extends Row<unknown> = Row> {
...
+
+  requestSnapshot(params: {
+    where?: string
+    params?: Record<string, string>
+    limit: number
+    offset?: number
+    orderBy: string
+  }): Promise<{
+    metadata: SnapshotMetadata
+    data: Array<{ key: string; value: T }>
```
KyleAMathews: this should be ChangeMessage ?

packages/typescript-client/src/client.ts
icehaunter: Done

packages/typescript-client/src/client.ts
```diff
@@ -114,11 +124,20 @@ export type ExternalParamsRecord<T extends Row<unknown> = Row> = {
   [K in string]: ParamValue | undefined
 } & Partial<PostgresParams<T>> & { [K in ReservedParamKeys]?: never }
 
+export type ExternalSubsetParamsRecord = {
```
kevin-dp: nit: the name of this type is a bit cryptic. What about `SubsetParams` or `SnapshotParams`?
Is there any reason we specifically talk about subsets instead of snapshots?

packages/typescript-client/src/client.ts
```diff
@@ -293,8 +317,20 @@ export interface ShapeStreamInterface<T extends Row<unknown> = Row> {
   lastOffset: Offset
   shapeHandle?: string
   error?: unknown
+  mode: `changes_only` | `full`
```
kevin-dp: Let's use the `LogMode` type here

packages/typescript-client/src/client.ts
```diff
@@ -293,8 +317,20 @@ export interface ShapeStreamInterface<T extends Row<unknown> = Row> {
...
 
   forceDisconnectAndRefresh(): Promise<void>
+
+  requestSnapshot(params: {
```
kevin-dp, line 324: For maintainability reasons I'd prefer if we reuse the `ExternalSubsetParamsRecord` type here and modify it such that limit and orderBy are required. To simplify this we can define a helper type:
```ts
type RequireProps<T, K extends keyof T> = 
  Omit<T, K> & Required<Pick<T, K>>;
```
and then this type becomes: `params: RequireProps<ExternalSubsetParamsRecord, "limit" | "orderBy">`

packages/typescript-client/src/client.ts
```diff
@@ -978,6 +1046,81 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+    if (this.#mode === `full`) {
+      throw new Error(
+        `Snapshot requests are not supported in ${this.#mode} mode, as the consumer is guaranteed to observe all data`
+      )
+    }
+    // We shouldn't be getting a snapshot on a shape that's not started
+    if (!this.#started) await this.#start()
+
+    // Pause the stream if this is the first snapshot request
+    this.#activeSnapshotRequests++
```
kevin-dp, line 1112: Should we move this line and the if test below into the try block to be 100% sure that the activeSnapshotRequests is always decremented?
Right now it would not be decremented if e.g. `this.#pause()` would throw. I know that right now that's not possible but if we change it in the future we may forget about this so it looks safer to move it into the try block.

packages/typescript-client/src/snapshot-tracker.ts
```diff
@@ -0,0 +1,86 @@
+import { Row, SnapshotMetadata } from './types'
+import { ChangeMessage } from './types'
+
+/**
+ * Determines if a change message should be filtered based on active snapshots
```
kevin-dp: This docstring seems to document `shouldRejectMessage` rather than this class.

packages/typescript-client/src/snapshot-tracker.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+      }
+    }
+
+    // If this change is skippable via any snapshot, skip it.
+    for (const { xmin, xmax, xip, keys } of this.activeSnapshots.values()) {
+      if (!keys.has(message.key)) continue
+      if (xid < xmin) return true
+      if (xid < xmax && !xip.includes(xid)) return true
+      if (xid < xmax) continue
```
kevin-dp: These last 2 if statements are obsolete

packages/typescript-client/src/types.ts
```diff
@@ -11,6 +11,7 @@ export type Value<Extensions = never> =
   | Extensions
   | Value<Extensions>[]
   | { [key: string]: Value<Extensions> }
+  | number[]
```
kevin-dp: This is not needed, it's covered by the recursive case `| Value<Extensions>[]`

packages/typescript-client/src/client.ts
```diff
@@ -114,11 +124,20 @@ export type ExternalParamsRecord<T extends Row<unknown> = Row> = {
...
+  offset?: number
+  orderBy?: string
+}
+
 type ReservedParamKeys =
   | typeof LIVE_CACHE_BUSTER_QUERY_PARAM
   | typeof SHAPE_HANDLE_QUERY_PARAM
   | typeof LIVE_QUERY_PARAM
   | typeof OFFSET_QUERY_PARAM
+  | `subset[${string}]`
```
samwillis: I think this should be this?
```suggestion
  | `subset__${string}`
```

packages/typescript-client/src/client.ts
icehaunter: Those are no longer required anyway, so it's all good

PR #3160, opened 2025-09-23
feat(sync-service) Optimise removing a shape from the where clause filter indexes

robacourt:
Turbo.ai are seeing ShapeLogCollector.unsubscribe take 100ms on average with a P99 of 300ms. This blocks the replication stream and given that 50 shapes can be expired at a time can cause many seconds of lag.

We've seen from telemetry that it's removing from the where clause filter indexes that is the bottleneck.

This PR speeds up removing a shape from the where clause filter indexes. For 60K shapes, it speeds it up 12800x. It achieves this by using the shape data to selectively remove from the index tree rather than traversing the whole tree looking for the shape handles.

This PR also lowers the memory footprint of the filter. A naive implementation would have increased the memory footprint since the shape data is needed for removal now, but since we're trying to lower memory use I've taken this opportunity to remove the duplication of shape data in the filter so now the memory use is less than before.

```
# before
Removing 100 shapes from 60000: each took 12.805ms mean, P99 48.303ms

:erts_debug.flat_size(filter) #=> 12581546

# this PR
Removing 100 shapes from 60000: each took 0.001ms mean, P99 0.008ms

:erts_debug.flat_size(filter) #=> 12324026
```

### DIFF
--- a/packages/sync-service/lib/electric/shapes/filter.ex
+++ b/packages/sync-service/lib/electric/shapes/filter.ex
@@ -23,7 +23,7 @@ defmodule Electric.Shapes.Filter do
 
   require Logger
 
-  defstruct tables: %{}, refs_fun: nil
+  defstruct tables: %{}, refs_fun: nil, shapes: %{}
 
   @type t :: %Filter{}
   @type shape_id :: any()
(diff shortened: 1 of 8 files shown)


packages/sync-service/lib/electric/shapes/filter.ex
```diff
@@ -60,16 +61,20 @@ defmodule Electric.Shapes.Filter do
...
-      filter
-      | tables:
-          filter.tables
-          |> Enum.map(fn {table_name, condition} ->
-            {table_name, WhereCondition.remove_shape(condition, shape_id)}
-          end)
-          |> Enum.reject(fn {_table, condition} -> WhereCondition.empty?(condition) end)
-          |> Map.new()
-    }
+    shape = Map.fetch!(filter.shapes, shape_id)
```
msfstef, line 64: if the memory overhead for this is large, we do have a centralised `shape_handle -> shape` index in `ShapeStatus` (although that includes an ets read which for something as performance sensitive as this might not be worth ti?)

robacourt: The issue is the shape has been wiped from the shape cache by this point. We'd need to keep it somehow.

For this PR I've actually reduced the memory footprint so I'm not worried for now. I think there are plenty of optimisations that we can do to reduce memory footprint of the filter such as shape IDs being Int64s before we need to go down the central storage route.

balegas: let's run some benchmarks just to sanity check

balegas:
> For 60K shapes, it speeds it up 12800x.

The best trick for really huge speedups is to have really bad base implementations :D

robacourt:
benchmark this

packages/sync-service/lib/electric/shapes/filter.ex
robacourt: The benchmarks show there's no statistical difference in memory footprint between the two
Screenshot 2025-09-23 at 15 45 58

PR #3161, opened 2025-09-23
feat: Use separate `ShapeCleaner` process for shape removals to avoid blocking `ShapeCache`

msfstef:
We had spread out deletions a bit too much over the code base, so this is a first attempt at unifying that again.

The main purpose is to move shape cleaning operations away from `ShapeCache`, and especially `clean_all_shapes_for_relations` which was running in a blocking manner, and on a relation change could affect potentially thousands of shapes and block the critical path unnecessarily.

I've kept the actual cleanup in `Electric.Shapes.Monitor` to avoid making this PR too large, and also because it lives outside of the shape supervision and is thus able to clean up storage even during stack restarts due to connection issues, although I'm not entirely sure that's good as it makes it hard to reason about race conditions (especially with restoring backed up state).

- Now `SchemaReconciler`, `ExpiryManager`, `PublicationManager` and `ShapeCache` all point to the `ShapeCleaner` for both stopping and cleaning shapes.
- The `ShapeCleaner` in turn ensures the shape consumer is dead, and if it is already dead schedules a shape "purging" via the monitor as we are already doing.
- On a sync `remove_shape` call, we immediately remove the shape from shape status if the consumer is already dead - this shouldn't really happen as we should not have a shape withotu a consumer in our index, but it's there for safety.
- On a `remove_shapes_for_relations`, rather than do things synchronously and block the process, removals are queued up and done one by one to allow `remove_shape` calls to be interleaved by the `ExpiryManager` and `ShapeCache`'s failed recoveries and starts.

### DIFF
--- a/.changeset/odd-kangaroos-hunt.md
+++ b/.changeset/odd-kangaroos-hunt.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Move shape deletion operations into separate process to avoid blocking `ShapeCache` on critical path.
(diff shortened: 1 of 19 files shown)


icehaunter (approved):
Really good refactoring, no comments I think

PR #3163, opened 2025-09-24
feat: add a control message at the end of snapshot

icehaunter:
@samwillis requested this as an explicit boundary so that it's present in the replication stream and allows clients to reason about transactions that have been incorporated into the initial snapshot, which is important for write-then-read pattern in Electric, where we're waiting for the txid returned from a read.

This add a new control message - `snapshot-end`, with 3 additional fields: xmin/xmax/xip which should be plenty.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
@@ -284,9 +284,15 @@ defmodule Electric.Shapes.Consumer.Snapshotter do
 
             send(task_parent, {:ready_to_stream, self(), System.monotonic_time(:millisecond)})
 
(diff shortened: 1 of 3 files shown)


robacourt (approved):
Nice

samwillis:
@icehaunter this is great, and matches what we need. two questions:

1. Are these the 32bit or 64bit txids? is `number` sufficiently large enough of should they be strings?

2. You mentioned having an `isVisibleInSnapshot` export from the client with a function that handled comparisons with these, is it worth adding that to this PR?

icehaunter:
@samwillis yeah, good notice - they are indeed int64. I'll adjust the code here & in the subsets PR. As for the export - yeah, I can do it here.

PR #3164, opened 2025-09-24
feat(sync-service): Handle pool timeouts and disconnections more gracefully

robacourt:


.changeset/shiny-falcons-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Handle pool timeouts and disconnections more gracefully
```
msfstef:
```suggestion
Handle pool timeouts and disconnections in the DB Inspector more gracefully
```
nit: you're probably following up with more graceful handling in shape log collector

PR #3165, opened 2025-09-24
fix: Handle errors caused by stack termination

magnetised:
At any point in the API request cycle the referenced stack could go down: either due to some problem with the database connection or because of some cloud restart.

We should handle errors due to the stack disappearing without obscuring errors arising from other issues/bugs in the request handling.

This PR wraps the `API.validate/2` and `API.serve_shape_log/1` calls in exception handling and a re-test of the stacks availability. 

If the stack is down then we return the correct 503 response. If its up then we retry the call on the understanding that if the stack was the problem it's now up again and things should work but if the problem was somewhere else the same request will raise the same exception which we won't catch/retry.

Fixes https://github.com/electric-sql/stratovolt/issues/845

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -454,8 +454,8 @@ defmodule Electric.Shapes.Api do
     validate_serve_usage!(request)
 
     with_span(request, "shape_get.plug.serve_shape_log", fn ->
-      request
-      |> serve_shape_log_catching_stack_shutdown()
+      request.api
+      |> handling_stack_down(fn -> do_serve_shape_log(request) end)
       |> ensure_response_cleanup()
     end)
   end
@@ -519,12 +519,6 @@ defmodule Electric.Shapes.Api do
           message:
             "Request.serve/1 must be called from the same process that called Request.validate/2"
     end
-  end
-
-  defp serve_shape_log_catching_stack_shutdown(%Request{api: api} = request) do
-    handling_stack_down(api, fn ->
-      do_serve_shape_log(request)
-    end)
   end
 
   defp do_serve_shape_log(%Request{} = request) do


packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -357,20 +359,24 @@ defmodule Electric.Shapes.Api do
...
   end
 
-  defp register_shape_subscriber(%Request{} = request) do
+  defp register_shape_subscriber(%Request{registered_pid: nil} = request) do
     %{api: %{stack_id: stack_id}, handle: handle} = request
     :ok = Electric.Shapes.Monitor.register_reader(stack_id, handle)
 
     Logger.debug(fn -> "Registering subscriber for shape #{inspect(handle)}" end)
 
+    %{request | registered_pid: self()}
```
msfstef: gotta love idempotency - small nit is `registered_pid` is a bit too generic, maybe `shape_subscriber_pid` is more clear?

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -496,6 +514,12 @@ defmodule Electric.Shapes.Api do
     end
   end
 
+  defp serve_shape_log_catching_stack_shutdown(%Request{api: api} = request) do
```
msfstef: nit: do we need this helper method if the only place we call this in is in `serve_shape_log`?

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -179,11 +179,13 @@ defmodule Electric.Shapes.Api do
   @spec validate(t(), %{(atom() | binary()) => term()}) ::
           {:ok, Request.t()} | {:error, Response.t()}
   def validate(%Api{} = api, params) when is_configured(api) do
-    with :ok <- hold_until_stack_ready(api),
-         {:ok, request} <- validate_params(api, params),
-         {:ok, request} <- load_shape_info(request) do
-      {:ok, seek(request)}
-    end
+    handling_stack_down(api, fn ->
```
msfstef, line 185: I think we might want to add this in a couple more places where we do `hold_until_stack_ready`, namely:
- `predefined_shape`
- `validate_for_delete`

Basically anywhere we do `hold_until_stack_ready`, except for after a long poll timeout (which we could and should remove since we have added this retry logic IMO), we would use this as well.

EDIT:
we should not remove the `hold_until_stack_ready` after the long poll timeout, cause a retry might cause the long poll to run again and the request ends up taking at worst 5sec + 20 sec + 5 sec + 20 sec total of 50 seconds to complete - what do you think?

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -938,6 +962,24 @@ defmodule Electric.Shapes.Api do
...
+  defp handling_stack_down(api, action_fun) do
+    try do
+      action_fun.()
+    rescue
+      e ->
+        Logger.warning(
+          "Got exception serving API request: #{Exception.format(:error, e, __STACKTRACE__)}. Retrying..."
+        )
+
+        with :ok <- hold_until_stack_ready(api) do
```
msfstef, line 980: love that we're encapsulating all stack down retries in one place

I'm not sure about swallowing errors in case the stack was up to begin with - some errors might occur one time and not the next time due to race conditions, and we'll be swallowing them silently by virtue of essentially always having at least one retry (although it might silently affect P99 latencies)

On the other hand even if we check and the stack is "up" at this point, it might have been down for a short time when the error occurred.

Perhaps a good compromise is to add a property to the telemetry span here marking that the request is retried, so we can measure how often this happens easily?

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -938,6 +962,24 @@ defmodule Electric.Shapes.Api do
     nil
   end
 
+  defp handling_stack_down(api, action_fun) do
+    try do
+      action_fun.()
+    rescue
```
msfstef: One of our most common errors are uncaught genserver exits (see [this issue](https://electricsql-04.sentry.io/issues/46809791/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream) for example)  which this will not catch - should we turn this into a `try...catch` instead (and add a test for that as well)?

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -938,6 +962,24 @@ defmodule Electric.Shapes.Api do
...
+      e ->
+        Logger.warning(
+          "Got exception serving API request: #{Exception.format(:error, e, __STACKTRACE__)}. Retrying..."
+        )
+
+        with :ok <- hold_until_stack_ready(api) do
+          # if the stack is up then either the error was transient caused by the stack restarting
+          # or its permanent: either the stack is just gone or there's some underlying bug that
```
msfstef: there's still the edge case of the stack going up and down multiple times during this retries but I'd rather see that happen before overengineering something (and also something else will probably be seriously wrong with the stack)

packages/sync-service/lib/electric/shapes/api.ex
magnetised: and in validate/2...

msfstef: we don't call this in `validate/2` no? I'm looking through the diff and the only usage of `serve_shape_log_catching_stack_shutdown` is in `serve_shape_log/2`? am I reading this wrong?

msfstef:
> To me issue is that the stack goes down with connection errors which are expected. If we didn't we probably wouldn't need this PR. I wrote up my thoughts on this [here](https://electric-sql.slab.com/posts/inspector-to-gracefully-handle-connection-failures-ki3n6row?shr=e6eh1npcco4bpadfa51ohn6q). I'm currently working on the ShapeLogCollector part.

It's true, but until we are able to serve shapes with no connections at all this is _very_ relevant. But even beyond that it would still be relevant for cases where things can go down temporarily and brought back up by a supervisor, but the requests passed a check right before that happened. I think having _at least_ a single retry will save us so much trouble.

>Hiding the first error may be problematic as it may be the one with the most useful information, subsequent errors may just be secondary effects. Sure the first error is a warning in the logs but that may be missed.

Agree although the idea is that if the error was because of the stack being down, we will have seen an error about why the stack is down elsewhere - the request should receive an appropriate response about the stack being down and no need to log errors for it failing because it's trying to interact with a failing stack.

I still would like some observability which is why I'm suggesting having at least a span attribute so we can measure how often this happens - as Garry pointed out the only errors that would get swallowed up are transient ones.

An alternative which I know Garry is opposed to is having a small list of errors we _know_ we can ignore and not log to Sentry: `:enoent` when reading files, `ArgumentError` when reading ETS tables, GenServer exits because of shutdowns/timeouts. It's not ideal but we could preserve observability over unknown errors.

>If you do decide to proceed with this approach note that issues with the stack often manifest as exits rather than exceptions.

Yep indeed, mentioned it here https://github.com/electric-sql/electric/pull/3165#discussion_r2376041602

robacourt:
> It's true, but until we are able to serve shapes with no connections at all this is _very_ relevant. But even beyond that it would still be relevant for cases where things can go down temporarily and brought back up by a supervisor, but the requests passed a check right before that happened. I think having _at least_ a single retry will save us so much trouble.

Fair enough. I can certainly see it's merit as a quick fix for now. I do worry about it making debugging more complicated but it sounds like you're thinking of lots of ways of mitigating that. If we get to the stage that the stack only goes down on unexpected errors that we do need to be alerted about, a cascade of errors is to be expected and not necessarly an issue, so perhaps this could be removed then if it's costs no longer outweigh it's benefits.

magnetised:
> Fair enough. I can certainly see it's merit as a quick fix for now. I do worry about it making debugging more complicated but it sounds like you're thinking of lots of ways of mitigating that. If we get to the stage that the stack only goes down on unexpected errors that we do need to be alerted about, a cascade of errors is to be expected and not necessarly an issue, so perhaps this could be removed then if it's costs no longer outweigh it's benefits.

One of the prompts for this fix were the thousands of errors generated in the cloud by a stack restart. So we'll always have to handle the stack going down unexpectedly in the middle of a request.

packages/sync-service/lib/electric/shapes/api.ex
magnetised: my bad. mis-read the location of the comment. i didn't want to over-complicate the do_serve_shape_log function, which is already complex enough, though another layer of indirection is annoying too.

balegas:
@magnetised have you identified the classes of errors that are more noisy? Would there be more precise ways of handling those than doing a catch all? 

If we can catch errors originating from http requests, we would return a proper error message on the response and avoid the mass of errors flowing to Sentry. I think it's okay we still get some errors for unhandled race conditions. I'd prefer making things better case by case than a catch-all.

magnetised:
Supplanted by improved stack down recognition

PR #3168, opened 2025-09-24
fix: Remove `refresh_publiation` API form PubMan

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3167

We started seeing this error with an update made on 22nd of September, and until then one of the relevant changes included is this: https://github.com/electric-sql/electric/pull/3108

Since @alco has introduced periodic forced publication updates anyway, and upon reconciling the schema we also trigger dropping relevant shapes (which would remove them from the publication on cleanup), we no longer need to refresh the publication manually IMO - not much to be gained other than unnecessary dependencies between our processes.

It's still worth investigating why we're seeing these timeouts (but was harder to debug because we were missing sentry metadata, see https://github.com/electric-sql/electric/pull/3166) - but ultimately we should simplify things.

1. Schema gets reconciled
2. Shapes using diverged/stale relations get dropped
3. Shape cleanup will involve removing them from the publication which should cause a publication update (as all relevant shapes for a given relation will be removed)
4. Even if (3) doesn't occur, which it will, we now periodically refresh the publication anyway

In fact, I would go even further to say that since we're periodically refreshing the publication and thus detecting missing relations there, we should simply add an inspector cleanup and get rid of the schema reconciler in its entirety - what do you think @alco @icehaunter ?

### DIFF
--- a/.changeset/little-seals-try.md
+++ b/.changeset/little-seals-try.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Remove unncecessary `refresh_publication` API from `PublicationManager`.
--- a/integration-tests/tests/not-owner-of-the-table.lux
+++ b/integration-tests/tests/not-owner-of-the-table.lux
@@ -69,10 +69,8 @@
 [shell electric]
   ??[info] Configuring publication electric_publication_integration to drop ["public.items"] tables, and add [] tables
 
-  # Call refresh_publication() once more to ensure the previous update has completed. This is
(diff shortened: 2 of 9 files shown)


icehaunter:
> In fact, I would go even further to say that since we're periodically refreshing the publication and thus detecting missing relations there, we should simply add an inspector cleanup and get rid of the schema reconciler in its entirety

Yeah, sure, I've missed where we've added periodicity to the publication manager. If it's already touching the DB, no reason to not pool that logic there.

PR #3172, opened 2025-09-25
hack(sync-service): restart on max shapes

robacourt:


balegas:
Hopefully we don't need this anymore

PR #3174, opened 2025-09-25
fix: Prevent race between shape adding and removing

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3173

Our `remove_shape` call uses a `select_delete` to remove all relevant records from the table for a given shape.

Because `select_delete` does a traversal and does not guarantee that concurrent operations won't see mixed results during it (i.e. it is not atomic and isolated), we ran into the following scenario:

1. We check `get_existing_shape` to determine whether a handle exists for a given shape spec
2. That checks the `shape -> handle` lookup, which returned a handle
3. That _subsequently_ checks the `handle -> metadata` lookup for the latest offset on the shape, which returns nil
4. The overall `get_existing_shape` call returns nil (assuming that if anything is missing, all records are missing)
5. We get to `add_shape` which expects `:ets.insert_new` to succeed for the entry to the `shape -> handle` lookup
6. The operation fails as the `shape -> handle` lookup entry is still present (see step 2)

I've opted for the following solution:
- In `remove_shape`, ensure the `shape -> handle` lookup entry gets deleted _first_, as it is the pointer from a shape def to any existing shapes.
  - The shape cleanup can be delayed (similarly for the files as well), but the first thing that needs to happen is to "deprecate" it such that the shape spec is no longer pointing to its handle
- In `get_existing_shape`, rather than swallowing the error in case the `shape -> handle` entry is present, but the `handle -> metadata` entry is not, I'm asserting that it _should_ be present.
  - The swallowing of the error in that call was essentially what was causing this race silently
  - An alternative solution would have been for `get_existing_shape` to call itself recursively until it returns a consistent result, but I thought that would be disgusting.

### DIFF
--- a/.changeset/modern-chairs-speak.md
+++ b/.changeset/modern-chairs-speak.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix race condition between removing and adding new shapes concurrently that led to crashes.
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -190,10 +190,14 @@ defmodule Electric.ShapeCache.ShapeStatus do
           @shape_meta_shape_pos
         )
 
+      # Always delete the hash lookup first, so that we guarantee that no shape spec
+      # is ever matched to a handle with incomplete information, since deleting with
+      # select_delete can lead to inconsistent state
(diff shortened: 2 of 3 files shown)


robacourt (approved):
Great spot and neat solution!

balegas:
Thanks @msfstef, great find. 

It's really hard to catch issues like this in testing environments, but if we have a test suite and we know what we're looking for, we might end up finding other things. I want for us to write an RFC next week about doing automated load testing.

PR #3179, opened 2025-09-25
feat: remove a stuck lock if underlying slot is not active

icehaunter:
Closes https://github.com/electric-sql/stratovolt/issues/837

### DIFF
--- a/packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
+++ b/packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
@@ -1,0 +1,115 @@
+defmodule Electric.Postgres.LockBreakerConnection do
+  @moduledoc """
+  A Postgres connection that is used to break an abandoned lock.
+  """
+  alias Electric.Connection
+
+  require Logger
+
+  @behaviour Postgrex.SimpleConnection
+
+  @type option ::
+          {:connection_opts, Keyword.t()}
+          | {:stack_id, String.t()}
+
+  @type options :: [option]
+
+  def name(stack_id) do
+    Electric.ProcessRegistry.name(stack_id, __MODULE__)
+  end
+
+  @spec start_link(options()) :: {:ok, pid()} | {:error, Postgrex.Error.t() | term()}
+  def start_link(opts) do
+    {connection_opts, init_opts} = Keyword.pop(opts, :connection_opts)
+
+    connection_opts = Electric.Utils.deobfuscate_password(connection_opts)
+
+    stack_id = Keyword.fetch!(opts, :stack_id)
+
+    Postgrex.SimpleConnection.start_link(
+      __MODULE__,
+      init_opts |> Keyword.put(:database, connection_opts[:database]),
+      [
+        auto_reconnect: false,
+        sync_connect: true,
+        name: name(stack_id)
+      ] ++
+        connection_opts
+    )
+  end
+
+  def stop_backends_and_close(server, lock_name, lock_connection_backend_pid \\ nil) do
+    send(server, {:stop_backends_and_close, lock_name, lock_connection_backend_pid})
+  end
+
+  @impl true
+  def init(opts) do
+    opts = Map.new(opts)
+
+    Process.set_label({:lock_breaker_connection, opts.stack_id})
+
+    metadata = [
+      # flag used for error filtering
+      is_connection_process?: true,
+      stack_id: opts.stack_id
+    ]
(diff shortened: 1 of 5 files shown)


packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  @impl true
+  def init(opts) do
+    opts = Map.new(opts)
+
+    Process.set_label({:lock_breaker_connection, opts.stack_id})
+
+    metadata = [
+      # flag used for error filtering
```
msfstef:
```suggestion
```
supernit: I ended up removing this comment - the flag is just indicating this is a connectin process, obviously useful for any kind of log/error processing

packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
```diff
@@ -0,0 +1,115 @@
...
+  @behaviour Postgrex.SimpleConnection
+
+  @type option ::
+          {:connection_opts, Keyword.t()}
+          | {:stack_id, String.t()}
+
+  @type options :: [option]
+
+  def name(stack_id) do
+    Electric.ProcessRegistry.name(stack_id, __MODULE__)
```
msfstef: since we're not keeping track of the process in connection manager to tear it down in `terminate`, I think giving this a unique name would be better so that we would never run into `:already_started` errors in case of restarts.

we had observed this issue with the lock connection before we added shutdowns in terminate to simulate supervisor behaviour.

packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
```diff
@@ -0,0 +1,115 @@
...
+  def handle_result([%Postgrex.Result{columns: ["pg_terminate_backend"]} = result], state) do
+    if result.num_rows == 0 do
+      Logger.debug("No stuck backends found")
+    else
+      Logger.notice(
+        "Terminated a stuck backend to free the lock #{state.lock_name} because slot with same name was inactive"
+      )
+    end
+
+    exit(:shutdown)
```
msfstef, line 86:
```suggestion
    {:disconnect, :shutdown}
```
why not use this? (I think this is the right API)

packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
```diff
@@ -0,0 +1,115 @@
...
+  def handle_result(%Postgrex.Error{} = error, _) do
+    raise error
+  end
+
+  defp lock_breaker_query(lock_name, lock_connection_backend_pid, database) do
+    """
+    WITH inactive_slots AS (
+        select slot_name
+        from pg_replication_slots
+        where active = false and database = '#{database}' and slot_name = '#{lock_name}'
```
msfstef, line 107: so to clarify - the way we ensure we're not breaking other electric's locks is by virtue of the slot being inactive?

I think it's worth adding more info in the moduledoc about this, the context of why we need to break these locks I think is important to include more fully, otherwise it's not clear why we have locks to begin with.

packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
icehaunter: No, `SimpleConnection`'s [behaviour](https://hexdocs.pm/postgrex/Postgrex.SimpleConnection.html#c:handle_info/2) [specifies](https://github.com/elixir-ecto/postgrex/blob/b419affa58b61cc67eba400de6e086ae964fe723/lib/postgrex/simple_connection.ex#L438-L442) only noreply and query tuples as valid return values:

```elixir
    case apply(mod, fun, args) do
      {:noreply, mod_state} ->
        {:keep_state, %{state | state: {mod, mod_state}}, {:timeout, state.idle_interval, nil}}
      {:query, query, mod_state} ->
        # ...
    end
```

packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
icehaunter: Yeah, I've removed the name

packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
icehaunter: Done

PR #3183, opened 2025-09-26
Partition DynamicConsumerSupervisor to improve startup and shutdown times

alco:
This adds a supervisor layer where a `PartitionSupervisor` now carries the `DynamicConsumerSupervisor.name(stack_id)` name and it starts a number of `DynamicConsumerSupervisor` processes to shard load between them. This change is transparent for the rest of the code that uses `DynamicConsumerSupervisor`'s public API.

I've done some measurements by repeatedly starting Electric, waiting for it to start all consumer processes for 50K shapes, then terminating it with the `SIGTERM` signal. I have done 5 attempts for the current implementation in `main` and for this PR, measuring both the startup and shutdown time.

|  | Attempt 1 | Attempt 2 | Attempt 3 | Attempt 4 | Attempt 5 |
| --- | --- | --- | --- | --- | --- |
| PartitionSupervisor startup | 28.8s | 12.5s | 11.8s | 11.0s | 11.8s |
| DynamicSupervisor startup | 20.6s | 20.0s | 20.0s | 19.7s | 20.5s |

|  | Attempt 1 | Attempt 2 | Attempt 3 | Attempt 4 |
| --- | --- | --- | --- | --- |
| PartitionSupervisor shutdown | 4.8s | 10.1s | 11.0s | 11.0s |
| DynamicSupervisor shutdown | 27.0s | 27.4s | 28.2s | 28.6s |

The 1st attempt can be considered an outlier since timings have become consistent afterwards.

When a PartitionSupervisor is used, all processes shut down within 2-3 seconds and the rest of the shutdown time is dominated by the shape status table backup.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/dynamic_consumer_supervisor.ex
+++ b/packages/sync-service/lib/electric/shapes/dynamic_consumer_supervisor.ex
@@ -14,8 +14,15 @@ defmodule Electric.Shapes.DynamicConsumerSupervisor do
 
   The number of dynamic supervisors is equal to the number of CPU cores.
   """
-  def partition_supervisor_spec(stack_id) do
-    {PartitionSupervisor, child_spec: {__MODULE__, [stack_id: stack_id]}, name: name(stack_id)}
+  def child_spec(opts) do
+    stack_id = Keyword.fetch!(opts, :stack_id)
+    name = Keyword.get(opts, :name, name(stack_id))
+
+    # We're overriding Electric.Shapes.DynamicConsumerSupervisor's child_spec() function here
+    # to make the usage of PartitionSupervisor transparent to the callers. As a consequence, we
+    # need to call `super()` to obtain the original DynamicSupervisor child_spec() to pass as an option to
+    # PartitionSupervisor.
+    PartitionSupervisor.child_spec(child_spec: super(opts), name: name)
   end
 
   def name(stack_id) do
--- a/packages/sync-service/lib/electric/shapes/dynamic_consumer_supervisor.ex
+++ b/packages/sync-service/lib/electric/shapes/dynamic_consumer_supervisor.ex
@@ -8,6 +8,12 @@ defmodule Electric.Shapes.DynamicConsumerSupervisor do
 
   require Logger
 
+  @doc """
+  Returns a child spec for the PartitionSupervisor that starts a pool of
+  DynamicConsumerSupervisor procecesses to shard child processes across.
+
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/shapes/dynamic_consumer_supervisor.ex
```diff
@@ -8,22 +8,47 @@ defmodule Electric.Shapes.DynamicConsumerSupervisor do
...
   def start_shape_consumer(name, config) do
-    Logger.debug(fn -> "Starting consumer for #{Access.fetch!(config, :shape_handle)}" end)
+    Logger.debug(fn -> "Starting consumer for #{Keyword.fetch!(config, :shape_handle)}" end)
 
-    DynamicSupervisor.start_child(name, {ConsumerSupervisor, config})
+    # Use a random integer as the routing key to achieve balanced sharding of child processes
+    # across all dynamic supervisors. The top limit for the key is picked to future-proof it
+    # for cases where Electric runs on a CPU with many cores. 256 should be sufficient for the
+    # foreseeable future.
+    key = :rand.uniform(256)
```
magnetised: According to the docs, PartitionSupervisor will do `rem(key, partitions)` to pick a partition so why not just `phash2` the shape handle and remove any assumptions.

magnetised (approved):
Nice. Looking forward to trying this out locally.

packages/sync-service/lib/electric/shapes/dynamic_consumer_supervisor.ex
alco: Fair point!

I thought calling `:erlang.phash2()` or `Integer.parse()` would be slower than just picking a random number but I've run [some benchmarks](https://gist.github.com/alco/4f0d75dc1c23837b7e2106a0361ec409) and `:erlang.phash2()` is consistently faster:

```
Name                       ips        average  deviation         median         99th %
:erlang.phash2()      367.92 K        2.72 μs   ±461.13%        2.60 μs        3.71 μs
:rand.uniform()       228.41 K        4.38 μs   ±420.16%        4.10 μs        7.11 μs
Integer.parse()       124.97 K        8.00 μs   ±121.39%        7.76 μs       14.43 μs

Comparison: 
:erlang.phash2()      367.92 K
:rand.uniform()       228.41 K - 1.61x slower +1.66 μs
Integer.parse()       124.97 K - 2.94x slower +5.28 μs
```

PR #3187, opened 2025-09-29
fix: Minor publication and connection error changes

msfstef:
- make `econnrefused` a retryable error, server might be overloaded, temporarily not running, we should not be shutting down so eagerly
- parse server login failures by poolers and make them retryable (currently only [this pgbouncer error](https://electricsql-04.sentry.io/issues/66120509/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream) captures but we can add more if we see them
- Remove dependency injection from pubman since we were already referencing `Configuration` module for manual publication updates - we use Repatch in tests instead.
- Fix type

### DIFF
--- a/.changeset/empty-spiders-peel.md
+++ b/.changeset/empty-spiders-peel.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Parse pooler login errors as retryable errors and `econnrefused` as retryable.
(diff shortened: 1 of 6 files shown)


msfstef:
> PlanetScale Postgres has a helpful [catalog of errors](https://planetscale.com/docs/postgres/connecting/client-error-reference#backend-connection-failures) with mitigation startegy advice in included.

that will be very useful, although generally errors with codes are something we can more easily deal with - for now i'm just adding things as they come on sentry

PR #3197, opened 2025-09-30
fix(sync-service): Keep trash directory in same filesystem as shape storage

### LINKED ISSUE #3196: AsyncDeleter cannot mv a storage dir across filesystems
**Bug description**

We attempt to make storage file deletion atomic/O(1) by renaming to a temp directory (using `System.tmp_dir!/0`). Linux by default uses a separate, special, mount for `/tmp` which means our rename fails as you cannot move a file between filesystems.

The rename should instead move the file within the storage dir which is guaranteed to be on the same filesystem.

magnetised:
Fixes #3196

### DIFF
--- a/packages/sync-service/lib/electric/async_deleter.ex
+++ b/packages/sync-service/lib/electric/async_deleter.ex
@@ -25,7 +25,6 @@ defmodule Electric.AsyncDeleter do
   ]
 
   @trash_dir_base ".electric_trash"
-
   @default_cleanup_interval_ms 10_000
 
   def name(stack_id) when is_binary(stack_id),
@@ -117,7 +116,16 @@ defmodule Electric.AsyncDeleter do
     {:noreply, do_cleanup(state)}
   end
 
-  def trash_dir(stack_id), do: Path.join([System.tmp_dir!(), @trash_dir_base, stack_id])
+  def trash_dir(stack_id) do
+    trash_base_dir =
+      Application.get_env(:electric, :trash_dir) ||
+        Path.join(
+          Application.get_env(:electric, :storage_dir, System.tmp_dir!()),
+          @trash_dir_base
+        )
+
+    Path.join(trash_base_dir, stack_id)
+  end
 
   defp do_rename(path, stack_id) do
     trash_dir = trash_dir(stack_id)
--- a/packages/sync-service/lib/electric/async_deleter.ex
+++ b/packages/sync-service/lib/electric/async_deleter.ex
@@ -136,7 +136,7 @@ defmodule Electric.AsyncDeleter do
         trash_dir
 
       [] ->
-        raise RuntimeError, message: ""
+        raise RuntimeError, message: "#{inspect(__MODULE__)} not running for stack #{stack_id}"
     end
   end
 
--- a/packages/sync-service/lib/electric/async_deleter.ex
+++ b/packages/sync-service/lib/electric/async_deleter.ex
@@ -90,8 +90,6 @@ defmodule Electric.AsyncDeleter do
       pending: []
     }
 
-    File.mkdir_p!(trash_dir)
-
     {:ok, state, {:continue, :initial_cleanup}}
   end
 
@@ -147,7 +145,8 @@ defmodule Electric.AsyncDeleter do
   defp do_rename(path, trash_dir) do
     dest = unique_destination(trash_dir, Path.basename(path))
 
-    with :ok <- File.rename(path, dest) do
+    with :ok <- File.mkdir_p(trash_dir),
+         :ok <- File.rename(path, dest) do
       {:ok, dest}
     end
   end
--- a/packages/sync-service/lib/electric/async_deleter.ex
+++ b/packages/sync-service/lib/electric/async_deleter.ex
@@ -32,6 +32,10 @@ defmodule Electric.AsyncDeleter do
 
   def name(opts), do: name(opts[:stack_id])
 
+  def ets_name(stack_id) when is_binary(stack_id) do
+    :"Electric.AsyncDeleter-#{stack_id}"
+  end
+
   def start_link(opts) do
     GenServer.start_link(__MODULE__, opts, name: name(opts[:stack_id]))
   end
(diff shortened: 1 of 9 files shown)


packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -117,7 +116,16 @@ defmodule Electric.AsyncDeleter do
     {:noreply, do_cleanup(state)}
   end
 
-  def trash_dir(stack_id), do: Path.join([System.tmp_dir!(), @trash_dir_base, stack_id])
+  def trash_dir(stack_id) do
+    trash_base_dir =
+      Application.get_env(:electric, :trash_dir) ||
```
msfstef: we don't use `Application.get_env` anywhere within the stack as we configure things per-stack in the stack supervisor - this needs to be part of our configuration and stack supervisor, same as our storage dir

msfstef (changes requested):
The trash dir needs to be passed as an option somewhere as we never use application config since we want things to be configured per stack.

Also, this can be a separate PR, but if you do a quick search for `Systemp.tmp_dir!` you will find we use it in one more place in utils for storage for compaction - might be worth changing that as well

packages/sync-service/lib/electric/async_deleter.ex
magnetised: That's a good point. Was being lazy...

magnetised:
> Also, this can be a separate PR, but if you do a quick search for `Systemp.tmp_dir!` you will find we use it in one more place in utils for storage for compaction - might be worth changing that as well

tmp_dir is only a problem if we attempt to `File.rename` into or out of it .. will double check that

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -43,11 +42,12 @@ defmodule Electric.AsyncDeleter do
   """
   def delete(path, opts) when is_binary(path) do
     stack_id = opts[:stack_id]
+    trash_dir = Keyword.get(opts, :trash_dir, trash_dir(stack_id))
```
alco:
```suggestion
    trash_dir = Keyword.get_lazy(opts, :trash_dir, fn -> trash_dir(stack_id) end)
```

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -130,16 +130,17 @@ defmodule Electric.AsyncDeleter do
...
 
   defp do_cleanup(state) do
-    # Remove the entire trash dir contents in one go
     start_time = System.monotonic_time(:millisecond)
 
-    try do
-      clean_dir!(state.trash_dir)
-    rescue
-      e -> Logger.warning("AsyncDeleter: rm_rf failed: #{inspect(e)}")
+    for dir <- MapSet.to_list(state.pending) do
```
alco: It looks odd to have `state.pending` changed to a map set only to be converted to list here. Are you worried about the same path ending up multiple times on the list?

I'm a bit out of the loop regarding why we cannot rm -rf the whole trash dir in one go, but if this is the way it's supposed to be, `Enum.each(state.pending, ...` would look more idiomatic than the current convertion to list and accumulation of return values from `unsafe_cleanup_with_retries!()` by the `for` expression.

alco (approved):

packages/sync-service/lib/electric/async_deleter.ex
magnetised: Yeah, I was pre-emptively avoiding duplicates in the pending list. Probably unlikely tbh, considering everything is given a unique name. Will stop worrying and just make it a list.. 🤷🏻 

> I'm a bit out of the loop regarding why we cannot rm -rf the whole trash dir in one go

Yeah. This change makes that harder too since we end up not knowing what the trash dir is. I was trying to avoid race conditions where we delete the trash dir then write to it. In the end everything's an explicit  recursive delete in elixir anyway so it really doesn't make a huge difference.

Will change to Enum.each - hadn't considered the implicit `reduce` of `for`

msfstef: +1 deleting the whole dir introduces races for ensuring the dir exists when renaming to it - and as garry said elixir rm rf is ls -> rm recursion anyway

unless we want to move to a port program for removing trash I think this doesn't matter

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -43,11 +42,12 @@ defmodule Electric.AsyncDeleter do
...
-    case do_rename(path, stack_id) do
-      {:ok, _dest} ->
+    case do_rename(path, trash_dir) do
+      {:ok, dest} ->
         server = opts[:server] || name(stack_id)
-        GenServer.cast(server, {:schedule_cleanup, path})
+        GenServer.cast(server, {:schedule_cleanup, dest})
         :ok
 
       {:error, :enoent} ->
```
msfstef, line 60: nit: I wrote this code but wanted to metnion that annoyingly whether the thing you're renaming is missing or the target is missing, you get the same enoent error -_- given the issue that prompted this fix I think we should somehow know if we're failing to rename things

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -130,16 +129,17 @@ defmodule Electric.AsyncDeleter do
   end
 
   defp do_cleanup(state) do
-    # Remove the entire trash dir contents in one go
-    start_time = System.monotonic_time(:millisecond)
+    if state.pending != [] do
```
msfstef, line 156: I disagree with using the in memory pending list - how do you ensure that the trash directory is fully cleaned up upon a restart, and trash doesn't just build up? Can we at least populate a pending list upon initialising?

packages/sync-service/lib/electric/async_deleter.ex
magnetised: I've just added a mkdir_p(trash_dir) to every rename op so either that errors, which won't be with an `:enoent` error..., or the problem is that the source `path` is gone.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -186,10 +189,15 @@ defmodule Electric.ShapeCache.PureFileStorage do
 
     try do
       case File.touch(marker_file_path) do
-        :ok -> Electric.AsyncDeleter.delete(opts.data_dir, stack_id: opts.stack_id)
+        :ok ->
+          Electric.AsyncDeleter.delete(opts.data_dir, stack_id: opts.stack_id)
```
msfstef, line 193: grrr... indents....

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -117,29 +130,38 @@ defmodule Electric.AsyncDeleter do
...
-  def trash_dir(stack_id), do: Path.join([System.tmp_dir!(), @trash_dir_base, stack_id])
+  def trash_dir!(stack_id) do
+    case :ets.lookup(ets_name(stack_id), :trash_dir) do
+      [{:trash_dir, trash_dir}] ->
+        trash_dir
 
-  defp do_rename(path, stack_id) do
-    trash_dir = trash_dir(stack_id)
+      [] ->
+        raise RuntimeError, message: ""
```
msfstef: can we add a message here?

magnetised: doh

PR #3198, opened 2025-09-30
Slim down PureFileStorage config's memory footprint + refactor a part of stack configuration sharing between processes

alco:
### TL;DR

- removing some internal configuration fields which are stack-accessible anyway and are simply passed between processes
- removing some instances of dependency injection pattern which does little more than obscure inter-module dependencies
- introduce a stack-wide `StackConfig` process + ETS table to keep static configuration for the stack
- strip away all path-related fields from PureFileStorage's struct to reduce per-consumer-process memory footprint

### Background

After I inspected which Elixir processes in Electric were taking up the most memory, cloned configuration options that are repeated for every shape were a clear offender. Most prominent of those is shape storage config that contains multiple paths, all sharing the same prefix.

<details>
<summary>Here's the state of a single consumer process running in production</summary>

```
%{
  monitors: [],
  buffer: [],
  registry: :"Registry.ShapeChanges:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
  otel_ctx: %{
    :"$__otel_baggage_ctx_key" => %{
      "stack_id" => {"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx", []}
    },
    {:otel_tracer, :span_ctx} => {:span_ctx,
     224916200833174271611003428090384065735, 13323822870313796073, 0,
     {:tracestate, []}, true, true, false,
     {:otel_span_ets, #Function<2.22926739/1 in :otel_tracer_server.on_end/1>}}
  },
  hibernate_after: 30000,
  writer: {Electric.ShapeCache.PureFileStorage,
   {:writer_state,
    {:writer_acc, [], [], 0, [], 0, 0, LogOffset.last_before_real_offsets(),
     LogOffset.last_before_real_offsets(), LogOffset.last_before_real_offsets(),
     LogOffset.last_before_real_offsets(), 0, 0, 0, false,
     {LogOffset.last_before_real_offsets(), []}}, nil, nil,
    #Reference<1.702501262.1906180100.191398>, "latest.0",
    %Electric.ShapeCache.PureFileStorage{
      buffer_ets: nil,
      base_path: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
      data_dir: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/000000000-9999999999999999",
      tmp_dir: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/.tmp/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
      metadata_dir: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/000000000-9999999999999999/metadata",
      log_dir: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/000000000-9999999999999999/log",
      stack_id: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
      stack_ets: :"Elixir.Electric.ShapeCache.PureFileStorage:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
      stack_task_supervisor: {:via, Registry,
       {:"Electric.ProcessRegistry:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
        {Electric.ShapeCache.PureFileStorage.TaskSupervisor, nil}}},
      shape_handle: "000000000-9999999999999999",
      chunk_bytes_threshold: 10485760,
      flush_period: 1000,
      compaction_config: %{period: 600000, keep_complete_chunks: 2},
      version: 1
    }}},
  storage: {Electric.ShapeCache.PureFileStorage,
   %Electric.ShapeCache.PureFileStorage{
     buffer_ets: nil,
     base_path: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
     data_dir: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/000000000-9999999999999999",
     tmp_dir: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/.tmp/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
     metadata_dir: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/000000000-9999999999999999/metadata",
     log_dir: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/000000000-9999999999999999/log",
     stack_id: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
     stack_ets: :"Elixir.Electric.ShapeCache.PureFileStorage:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
     stack_task_supervisor: {:via, Registry,
      {:"Electric.ProcessRegistry:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
       {Electric.ShapeCache.PureFileStorage.TaskSupervisor, nil}}},
     shape_handle: "000000000-9999999999999999",
     chunk_bytes_threshold: 10485760,
     flush_period: 1000,
     compaction_config: %{period: 600000, keep_complete_chunks: 2},
     version: 1
   }},
  stack_id: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
  chunk_bytes_threshold: 10485760,
  inspector: {Electric.Postgres.Inspector.EtsInspector,
   [     
     stack_id: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
     server: {:via, Registry,
      {:"Electric.ProcessRegistry:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
       {Electric.Postgres.Inspector.EtsInspector, nil}}}
   ]},
  shape_status: {Electric.ShapeCache.ShapeStatus,
   %Electric.ShapeCache.ShapeStatus{
     shape_meta_table: :"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx:shape_meta_table",
     storage: {Electric.ShapeCache.PureFileStorage,
      %{
        tmp_dir: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/.tmp/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
        base_path: "./persistent/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/shapes/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
        stack_id: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
        chunk_bytes_threshold: 10485760,
        stack_ets: :"Elixir.Electric.ShapeCache.PureFileStorage:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
        stack_task_supervisor: {:via, Registry,
         {:"Electric.ProcessRegistry:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
          {Electric.ShapeCache.PureFileStorage.TaskSupervisor, nil}}},
        flush_period: 1000,
        compaction_config: %{period: 600000, keep_complete_chunks: 2}
      }}
   }},
  db_pool: {:via, Registry,
   {:"Electric.ProcessRegistry:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
    {Electric.DbPool, :snapshot}}},
  publication_manager: {Electric.Replication.PublicationManager,
   [stack_id: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"]},
  shape_handle: "000000000-9999999999999999",
  log_producer: {:via, Registry,
   {:"Electric.ProcessRegistry:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
    {Electric.Replication.ShapeLogCollector, nil}}},
  run_with_conn_fn: &Electric.Shapes.Consumer.Snapshotter.run_with_conn/2,
  create_snapshot_fn: &Electric.Shapes.Consumer.Snapshotter.query_in_readonly_txn/7,
  shape: Shape.new!({00000, "public.table"}, 
    where: "...",
    columns: [...]
  ),
  latest_offset: LogOffset.last_before_real_offsets(),
  pg_snapshot: %{
    filter_txns?: true,
    xip_list: [],
    xmax: 0,
    xmin: 0
  },
  snapshot_started: true,
  txn_offset_mapping: [],
  materializer_subscribed?: false,
  awaiting_snapshot_start: [],
  hibernate_timer: nil,
  cleaned?: false
}
```

</details>

Reducing that in-memory size of that state is the main focus of this PR.

***N.B.** The size of the pretty-printed state above doesn't directly map to its in-memory size because the majority of values there are atoms (which are interned in BEAM) and some maps that are printed more than once may actually be stored once in memory. But due to the data immutability and each Elixir process having independent heap, there's not a lot of structured sharing happening.*

### Overview of code changes

The key change is made to the PureFileStorage struct: I've removed all of its fields that can be calculated from the shared options and refactored field accesses into function calls in the storage implementation.

Another significant refactoring is avoiding passing `shape_status_state` everywhere since it's static per stack. I haven't been able to remove it from `Electric.Shapes.Consumer` because of how pervasively mocks are used in consumer tests.

Repairing tests after doing the refactorings has eaten a large chunk of the development time. It is in large part due to how configuration is passed between modules dynamically, with ad-hoc Nimble schemas, default values being set in multiple places, etc. We should really centralize stack-wide config at some point which is now possible to introduce gradually with `Electric.StackConfig`.

### Note on the use of ETS for `Electric.StackConfig`

My original goal was to persist the shared configuration into `:persistent_term` because it's very fast for reads and doesn't copy data into process heaps. But there's a dealbreaker in the way `:persistent_term` works: modifying or deleting a value stored in it results in the BEAM VM scheduling a heap scanning + GC pass for _all running processes_.

### Mem footprint comparisons

Following are comparisons done in different environments.

First, a **comparison of OS stats** when running Electric locally using the `main` branch vs this PR.

main_shared_config_htop_custom_storage_cmp

Based on these numbers, the difference appears to be small. I'm attributing it to the fact that BEAM retains memory it has previously requested from the OS and doesn't immediately release it. Take a look at the **memory usage numbers reported by the BEAM VM** itself below (screenshots are from Erlang's Observer tool):

main_shared_config_sys_stats_custom_storage_cmp

Here the difference is more pronounced.

Finally, a **comparison of OTEL metrics gathered in staging env** based on charts that show memory footprint in relation to the total number of shapes, within a 1.5-hour time window.

[`main` branch](https://ui.honeycomb.io/electric-sql-06/environments/pr-846/datasets/electric-region/board-query/cDij6tGm74Z/result/xbb766Tqwxb?vs=hideCompare&cs_2=omitMissingValues):
main_staging_mem

[PR branch](https://ui.honeycomb.io/electric-sql-06/environments/pr-846/datasets/electric-region/board-query/cDij6tGm74Z/result/Dd69w6wUvdu?vs=hideCompare&cs_2=omitMissingValues):
shared_config_staging_mem

Key points:
- at 65K shapes `main` shows combined process heap at 2.32GB and the PR branch is at 1.81GB
- when shape expiry eventually brings the number of shapes down to 60K, `main` sits at 2.1GB whereas the PR branch is at 1.73GB.

### Takeaways

- In an architecture where a large number of user-defined resources (shapes) are apped to Elixir processes, we need to be very precise about what to keep in the process state.
- Shared configuration should be centralized as much as possible as opposed to duplicating it in tens of thousands of processes.
- There's also the cost of copying the same terms from one memory location to another every time a new process is started that uses that term.
- The configuration options passed to supervisor children are also kept in the supervisor process' memory as part of child specs. This is especially noticeable in the case of DynamicConsumerSupervisor, which ends up using 200-300MB just to keep shared config and shapes that are already stored in ETS tables. This is addressed in https://github.com/electric-sql/electric/pull/3230.
- The next big improvement in memory efficiency is going to be found in transaction processing and ShapeLogCollector. We need to investigate the impact of streaming transaction data and decoding only the few columns from each successive row that are needed for shape index look up first, before we decode the whole transaction received from Postgres into memory. The rows themselves can be more efficiently written into shape logs without fully materializing them in memory and passing around as messages. The ShapeLogCollector's index could be eventually moved to disk, with an in-memory cache for highly active shapes.

### DIFF
--- a/.github/workflows/elixir_tests.yml
+++ b/.github/workflows/elixir_tests.yml
@@ -95,22 +95,24 @@ jobs:
             ${{ runner.os }}-sync-service-build-
 
       - name: Install dependencies
-        run: mix deps.get && mix deps.compile
+        run: |
+          mix deps.get && mix deps.compile
+          MIX_TARGET=application mix deps.get && mix deps.compile
 
       - name: Compile
         # don't bail on compile warnings - let's get the results of the tests
         # and let the formatting task check for compilation warnings
-        run: mix compile
+        run: |
+          mix compile
+          MIX_TARGET=application mix compile
 
       - name: Run tests
         run: mix coveralls.json --include slow --cover --export-coverage test
 
       - name: Run telemetry tests
-        run: |
-          rm -rf _build/application_test
-          cp -R _build/test _build/application_test
-          rm -rf _build/application_test/lib/electric
-          MIX_TARGET=application mix coveralls.json --only telemetry_target --import-cover cover
+        env:
+          MIX_TARGET: application
+        run: mix coveralls.json --only telemetry_target --import-cover cover
 
       - name: Upload coverage reports to CodeCov
         uses: codecov/codecov-action@ad3126e916f78f00edff4ed0317cf185271ccc2d
--- a/packages/sync-service/mix.exs
+++ b/packages/sync-service/mix.exs
@@ -149,7 +149,7 @@ defmodule Electric.MixProject do
   defp telemetry_deps() do
     [
       {:sentry, "~> 11.0"},
-      {:opentelemetry, "~> 1.5"},
+      {:opentelemetry, "~> 1.6"},
       {:opentelemetry_exporter, "~> 1.8"},
       {:otel_metric_exporter, "~> 0.3.11"},
       # For debugging the otel_metric_exporter check it out locally and uncomment the line below


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -44,8 +44,8 @@ defmodule Electric.Replication.ShapeLogCollector do
     Electric.ProcessRegistry.name(stack_id, __MODULE__)
   end
 
-  def set_last_processed_lsn(server, last_processed_lsn) do
-    GenServer.call(server, {:set_last_processed_lsn, last_processed_lsn})
+  def set_last_processed_lsn(term, last_processed_lsn) do
```
magnetised: `term` doesn't feel like the right word here. what about `server_ref` or something that explicitly declares what the arg is.

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -90,36 +76,32 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
-  def opts(opts) do
-    opts = NimbleOptions.validate!(opts, @schema)
-    struct(__MODULE__, opts)
-  end
-
-  @impl true
-  def initialise(
-        %__MODULE__{shape_meta_table: meta_table, shape_last_used_table: last_used_table} = state
-      ) do
+  def initialise(term, storage) do
```
magnetised: again term feels unhelpful. is this `stack_ref`?

packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
```diff
@@ -25,11 +25,13 @@ defmodule Electric.Shapes.Consumer.Snapshotter do
     GenServer.call(name(stack_id, shape_handle), :start_snapshot, 1_000)
   end
 
-  def start_link(config) do
+  def start_link(config) when is_map(config) do
     GenServer.start_link(__MODULE__, config, name: name(config))
   end
 
   def init(config) do
+    test_callback_hook()
```
magnetised: Could you add a comment explaining this pls.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -68,6 +68,8 @@ defmodule Electric.Shapes.Consumer do
   @impl GenServer
 
   def init(config) do
+    test_callback_hook()
```
magnetised: similarly... a little bit of context for this would be useful.

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -256,7 +234,18 @@ defmodule Electric.ShapeCache do
...
+
+    # Empirical evidence shows that after recovering 50K shapes ShapeStatusOwner and ShapeCache
+    # each take up 200+MB of memory. Explicitly running garbage collection for both immediately
+    # takes that down to 4-5MB.
+    shape_status_owner_pid =
+      state.stack_id
+      |> Electric.ShapeCache.ShapeStatusOwner.name()
+      |> GenServer.whereis()
+
+    :erlang.garbage_collect(shape_status_owner_pid)
```
magnetised: could we move this gc to the status owner process?

packages/sync-service/test/support/test_utils.ex
```diff
@@ -126,4 +126,43 @@ defmodule Support.TestUtils do
...
+    :ok
+  end
+
+  # This function is normally called inside the init() callback of an OTP behaviour to inherit any
+  # patched functions from the test process.
+  #
+  # It looks up the test process' PID in the caller process' dictionary and executes the
+  # function under the `:callback_for_descendant_procs` key matching the caller's module (if
+  # any).
+  def execute_test_callback(caller_mod) do
```
magnetised: this is why I hate mocking.

magnetised (approved):
This is fab. Fantastic work. 

Especially agree re simplifying the config as you mention in the description: something I've hit with Phoenix.Sync a lot.

Could we get rid of the explicit mocking integration via callbacks? If we register `:"$ancestors"` and `:"$callers"` into the process dictionary we can use the same mechanism as e.g. Task.

alco:
> Could we get rid of the explicit mocking integration via callbacks? If we register `:"$ancestors"` and `:"$callers"` into the process dictionary we can use the same mechanism as e.g. Task.

I'm not following. There's no caller relationship between the test process and the descendant process we want to mock. But luckily for us, Elixir already maintains a chain of ancestors when it starts each new one. That's how we're able to get hold of the test process pid from the depths of the supervision tree.

packages/sync-service/lib/electric/shape_cache/shape_status.ex
alco: Good idea! Updated.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
alco: Great suggestion! Updated.

packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
alco: I've renamed these functions. The one in `Support.TestUtils` which is called from this has a comment explaining its purpose.

I can't think of a nicer way to achieve this. I could define a macro but it would still need to be explicitly `require`d and called inside `init()`.

packages/sync-service/lib/electric/shape_cache.ex
alco: Good call. Updated.

packages/sync-service/test/support/test_utils.ex
alco: Me too!

PR #3199, opened 2025-09-30
fix: Only register the writer consumer when terminating

### LINKED ISSUE #3191: `Electric.Shapes.Monitor` slowing startup and shutdown
**Bug description**

As every consumer process registers itself with `Electric.Shapes.Monitor` we have consumer startup serialised through a single process. Also we have the result of this delaying shutdown as every consumer process is monitored and so a shutdown of N consumers results in N down messages to the monitor process.  

Suggested fix is to only register the consumer when it's wanting to invoke the clean shutdown process -- a supervisor `:shutdown` doesn't go through this clean shutdown anyway so there's no need to introduce the dependency unecessarily.

magnetised:
Currently we register all shape consumer processes eagerly with the monitor and then use this to perform a cleanup when they terminate with the reason `{:shutdown, :cleanup}`.

In reality we don't need to register all the writers, only the ones who are requesting a cleanup. So rather than register in the `init/1` callback, this pr merges the writer registration with the request for reader termination notification.

With this a supervisor kill all -> restart won't even touch the monitor but shapes that have requested a clean termination will be properly handled.

Fixes #3191


packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
```diff
@@ -38,19 +38,13 @@ defmodule Electric.Shapes.Monitor.RefCounter do
   @doc false
   @spec register_reader(stack_id(), shape_handle(), pid()) :: :ok
   def register_reader(stack_id, shape_handle, pid \\ self()) do
-    GenServer.call(name(stack_id), {:register_reader, shape_handle, pid})
+    GenServer.call(name(stack_id), {:register_reader, shape_handle, pid}, :infinity)
```
msfstef: what is the reasoning for an infinity timeout on registering/unregistering readers? it seems like this makes the time to process a request unbounded? i.e. if this is solving an issue it already means that we're waiting >5 seconds for registering with a monitor, perhaps it would be better to see timeout errors about this than hide it in unbounded latency

magnetised: oh! good catch. I put that in locally - have a bunch of patches that I'm trying to juggle. will remove. thanks

needed it because the monitor process was overwhelmed with writer registrations, since that's no longer the case the patch isn't needed for my stress tests anyway

packages/sync-service/test/electric/shapes/consumer_test.exs
```diff
@@ -860,7 +860,6 @@ defmodule Electric.Shapes.ConsumerTest do
       # Stop the consumer and the shape cache server to simulate a restart
       stop_supervised!(ctx.consumer_supervisor)
       assert_receive {:DOWN, ^ref, :process, _pid, _reason}, 1000
-      assert_receive {Electric.Shapes.Monitor, :remove, ^shape_handle}
```
msfstef, line 863: nit: why remove this assertion?

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
```diff
@@ -291,6 +295,24 @@ defmodule Electric.Shapes.Monitor.RefCounter do
...
+      state,
+      consumer_pid,
+      ConsumerSupervisor.whereis(state.stack_id, shape_handle),
+      shape_handle
+    )
+  end
+
+  defp monitor_consumer_processes(state, consumer_pid, supervisor_pid, shape_handle) do
+    if supervisor_pid,
+      do: Process.monitor(supervisor_pid, tag: {:down, :writer_supervisor, shape_handle})
```
msfstef: do things work without monitoring the supervisor/ receiving a down message for it?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -294,6 +292,13 @@ defmodule Electric.Shapes.Consumer do
 
   @impl GenServer
   def terminate(reason, state) do
+    :ok =
```
msfstef, line 295: I'm not sure this makes sense? If you call this in the `terminate` callback rather than `terminate_safely`, it means that it won't actually wait for the reader termination message any more at all, since as soon as the terminate call ends the process dies (?)

Before we were calling it in a `terminate_safely` call that would essentially do nothing and the consumer would just handle killing itself when it receives the notification from the monitor. Am I missing something? Are we actually testing the logic that the consumer won't terminate until readers are done reading anywhere, cause I feel like this should fail?

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
```diff
@@ -188,40 +185,35 @@ defmodule Electric.Shapes.Monitor.RefCounter do
     {:reply, :ok, state}
   end
 
-  def handle_call({:register_writer, handle, pid}, _from, state) do
-    if supervisor = ConsumerSupervisor.whereis(state.stack_id, handle) do
+  def handle_call({:notify_reader_termination, shape_handle, pid, reason}, _from, state) do
```
msfstef, line 191: we could remove the entire `notify_reader_termination` logic and just make `handle_writer_termination` an infinite (or large) timeout call that returns when readers are done, and just block on it in the terminate callback?

packages/sync-service/test/electric/shapes/consumer_test.exs
magnetised: Because the shutdown is normal we no longer register the consumer pid with the monitor (which is the whole point of this pr) so it doesn't do anything and so doesn't send this callback message

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
magnetised: good point.

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
msfstef: side note - if we do block on it, it means that shape log collector will also block on processing a transaction that includes this shape, as it will wait for this to process a message, which it won't be able to process cause it's blocked on termination, but hasn't notified that it has died so shape log collector doesn't send it messages.

So perhaps shape log colletor needs to be sent an unsubscribe message rather than rely purely on monitoring if we want to do it this way

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: This call handles the case where the consumer is crashing. The `terminate_safely` route is for the consumer being cleanly removed. This call is a no-op for non-crash reasons such as a supervisor :shutdown and I'll make it a no-op for a `terminate_safely` shutdown too. 

This does the other half of this systems job: to cleanup shape data when the writer crashes

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
magnetised: we could, yes.

magnetised:
> I feel like we're not actually testing our real consumer processes with the monitor process - the monitor has unit tests with toy consumer processes and the consumer does not test that it waits to exit until readers are done.
> 
> Unless I'm wrong, the logic here (moving the call to `terminate`) makes the whole thing not actually do what it's supposed to, and no tests fail, which is concerning.

The monitor tests use "toy" processes but there are consumer tests that excercise the full machinery -- they were failing before the addition of the `terminate` call. the reason no tests fail is because I fixed them...

packages/sync-service/lib/electric/shapes/consumer.ex
msfstef: right - but as it stands now, would this ever wait for readers to terminate before shutting down? is there a test that asserts that with the actual consumer implementation?

msfstef: also, if the cleanup is actually handled by the monitor process, perhaps we should always immediately shut down the consumer and just let the monitor process wait for readers before cleaning it up, since its the one responsible for cleaning anyway?

msfstef (approved):
@magnetised misread the code, thought you removed the `notify_reader_termination` call from the consumer whereas it was just the register writer one - sorry!

Looks good!

PR #3202, opened 2025-10-01
docs: added docs for offset=now, subset snapshots and mode=changes_only

icehaunter:
Docs for #3159

### DIFF
--- a/website/docs/api/clients/typescript.md
+++ b/website/docs/api/clients/typescript.md
@@ -224,6 +224,19 @@ export interface ShapeStreamOptions<T = never> {
   subscribe?: boolean
 
   /**
+   * Initial data loading mode.
+   *
+   * When `mode` is `full` (the default), the server creates an initial snapshot
+   * of all data matching the shape definition before delivering real-time updates.
+   *
+   * When `mode` is `changes_only`, the server skips the initial snapshot creation.
+   * The client will only receive changes that occur after the shape is established,
+   * without seeing the base data. In this mode, you can use `requestSnapshot()` to
+   * fetch subsets of data on-demand.
+   */
+  mode?: "full" | "changes_only"
+
+  /**
    * Signal to abort the stream.
    */
   signal?: AbortSignal
@@ -348,7 +361,26 @@ export type ChangeMessage<T extends Row<unknown> = Row> = {
 }
 ```
 
-Or a `ControlMessage`, representing an instruction to the client, as [documented here](../http#control-messages).
+Or a `ControlMessage`, representing an instruction to the client:
+
+```ts
+export type ControlMessage = {
+  headers:
+    | (Header & {
+        control: `up-to-date` | `must-refetch`
+        global_last_seen_lsn?: string
+      })
+    | (Header & { control: `snapshot-end` } & PostgresSnapshot)
+}
+```
+
+Control messages include:
+
+- `up-to-date` - Indicates the client has received all available data
+- `must-refetch` - Indicates the client must discard local data and re-sync from scratch
+- `snapshot-end` - Marks the end of a subset snapshot, includes PostgreSQL snapshot metadata (xmin, xmax, xip_list) for tracking which changes to skip
+
+See the [HTTP API control messages documentation](../http#control-messages) for more details.
 
 #### Parsing and Custom Parsing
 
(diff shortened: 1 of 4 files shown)


website/docs/api/clients/typescript.md
```diff
@@ -589,4 +621,97 @@ The following error types may be encountered:
...
+  params: {
+    table: "items",
+  },
+  mode: "changes_only", // Skip initial snapshot
+})
+```
+
+In `changes_only` mode, the client only receives changes that occur after the shape is established. This is useful for:
+
+- Event streams where historical data isn't needed
```
samwillis: We need to be slightly cautious of calling it an event stream as we don't really guarantee seeing every event. - particularly with compaction.

website/docs/api/http.md
```diff
@@ -110,6 +126,58 @@ The `live` parameter puts the server into live mode, where it will hold open the
...
+
+#### Full mode (default)
+
+When using `log=full` (the default), the server creates an initial snapshot of all data matching the shape definition and streams it to the client before delivering real-time updates. This is the standard mode where you get the complete current state followed by live changes.
+
+#### Changes-only mode
+
+When using `log=changes_only`, the server skips creating an initial snapshot. The client will only receive changes that occur after the shape is established, without seeing the base data. This mode is useful for:
+
+- Event streams where historical data isn't needed
```
samwillis: Came comment as above

website/docs/api/clients/typescript.md
```diff
@@ -589,4 +621,97 @@ The following error types may be encountered:
...
+})
+```
+
+In `changes_only` mode, the client only receives changes that occur after the shape is established. This is useful for:
+
+- Places where historical data isn't needed
+- Applications that fetch their initial state through other means
+- Reducing initial sync time when you don't need the complete dataset
+
+#### Starting from 'now'
```
thruflo, line 646: I'm curious about the interplay between changes-only mode and offset now.

If you don't set offset now, then what do you get in changes only mode?
If you're in full mode, what happens with offset now?

website/docs/api/clients/typescript.md
```diff
@@ -589,4 +621,97 @@ The following error types may be encountered:
...
+  mode: "changes_only", // Skip initial snapshot
+})
+```
+
+In `changes_only` mode, the client only receives changes that occur after the shape is established. This is useful for:
+
+- Places where historical data isn't needed
+- Applications that fetch their initial state through other means
+- Reducing initial sync time when you don't need the complete dataset
+
```
thruflo, line 643: I would add a summary note about enabling Requesting subset snapshots which allows shape contents to be varied dynamically at runtime, enabling advanced dyanmic/progressive data loading strategies.

We can link to the shapes guide for a more narrative once updated.

website/docs/api/http.md
```diff
@@ -98,6 +98,22 @@ Note that the other control message is `must-refetch` which indicates that the c
 
 :::
 
+::: info Snapshot-end
```
thruflo, line 101: Do we need to similarly update the "writing your own client guide"? Can this message safely be ignored?

website/docs/api/http.md
```diff
@@ -110,6 +126,58 @@ The `live` parameter puts the server into live mode, where it will hold open the
 
 The server holds open the request until either a timeout (returning `200` with only an up-to-date message) or when new data is available, which it sends back as the response. The client then reconnects and the server blocks again for new content. This way the client is always updated as soon as new data is available.
 
+### Log modes
+
+Electric supports two log modes for syncing shapes, controlled by the `log` query parameter:
+
+#### Full mode (default)
+
+When using `log=full` (the default), the server creates an initial snapshot of all data matching the shape definition and streams it to the client before delivering real-time updates. This is the standard mode where you get the complete current state followed by live changes.
```
thruflo, line 135: The parameter name was `mode` above. 

Personally I prefer `log` as more descriptive.

website/docs/api/http.md
```diff
@@ -110,6 +126,58 @@ The `live` parameter puts the server into live mode, where it will hold open the
...
+```
+
+This is particularly useful when combined with `log=changes_only` mode and `replica=full` for applications that don't keep state and need to start fresh upon reload without historical data.
+
+### Subset snapshots
+
+When using `changes_only` mode, you can request subset snapshots to fetch specific portions of data on-demand. This is done by adding `subset__*` query parameters to your request:
+
+```sh
+curl -i 'http://localhost:3000/v1/shape?table=foo&offset=123_4&handle=abc-123&subset__where=priority=high&subset__order_by=created_at&subset__limit=10'
```
thruflo, line 162: What's the cache behaviour here? This is a GET but has the side effect of varying the shape definition.

website/docs/api/clients/typescript.md
```diff
@@ -223,6 +223,19 @@ export interface ShapeStreamOptions<T = never> {
...
+   *
+   * When `mode` is `full` (the default), the server creates an initial snapshot
+   * of all data matching the shape definition before delivering real-time updates.
+   *
+   * When `mode` is `changes_only`, the server skips the initial snapshot creation.
+   * The client will only receive changes that occur after the shape is established,
+   * without seeing the base data. In this mode, you can use `requestSnapshot()` to
+   * fetch subsets of data on-demand.
+   */
+  mode?: "full" | "changes_only"
```
thruflo, line 237: This is called `mode` whereas we use `log` in the http spec.

thruflo (commented):
Great stuff 👍 left a few comments.

website/docs/api/clients/typescript.md
icehaunter: 1) You'll get a stream of changes with no baseline snapshots. You're meant to discard changes until you've executed some subset snapshots and have a baseline to apply updates to - or do something else entirely like ts/db does. Given a stateful client that's executed and applied some snapshots, the client wouldn't use a `now` and instead would use a normal offset continuation point
2) Kinda same thing as a above, but another client that's using the same shape will be able to get the "normal" view of historical data

website/docs/api/http.md
icehaunter: Currently I  we're setting nocache headers on those subset responses. Those are not stored server-side, and they don't interfere with normal flow of data

website/docs/api/clients/typescript.md
icehaunter: Renamed to log everywhere

website/docs/api/http.md
icehaunter: I've added a note to client dev that all unknown control messages should be ignored

PR #3203, opened 2025-10-01
fix(sync-service): Add stack canary process to detect shutdown

### LINKED ISSUE #3190: StatusMonitor reports a stack as :ready even if it's in a crashed state
**Bug description**

If we have a crash in a process supervised by `Electric.Replication.Supervisor` then the entire stack is taken down because of the `:all_for_one` strategy.

Unfortunately only a subset of these processes are monitored, so if e.g. the `ShapeCache` crashes the StatusMonitor only finds out about it when the ShapeLogCollector is terminated which, critically, is **after** the consumer supervisor has taken down all the consumers (which can take a while, for various reasons).

So for the entire time we're tearing down consumers, we're accepting requests to a stack that's had some critical process crash and is actively being stopped.

**Expected behavior**

A restart of the processes managed by the `Replication.Supervisor` should mark the stack down until it's got to a reasonable state.

magnetised:
The last process in a supervisor's children list will be the first to be terminated in a restart. This canary process registers itself with the stack status and any shutdown of the Recplication.Supervisor, either due to a crash or a stack restart, will mark the stack as down immediately.

Seems to prevent a lot of errors on stack termination locally: only active requests raise due to the exit of the "crashed" process rather than every request after the process crash until the consumers have been terminated.

Fixes #3190

### DIFF
--- a/packages/sync-service/test/electric/connection/manager_test.exs
+++ b/packages/sync-service/test/electric/connection/manager_test.exs
@@ -182,6 +182,26 @@ defmodule Electric.Connection.ConnectionManagerTest do
 
       assert StatusMonitor.status(stack_id) in [:waiting, :starting]
     end
+
+    test "backtracks the status when the canary goes down", %{stack_id: stack_id} do
+      wait_until_active(stack_id)
+
+      # should backtrack the status by virtue of the shape log collector being shut down
+      # by the replication supervisor
+      monitor =
+        stack_id
+        |> Electric.Replication.Supervisor.canary_name()
+        |> GenServer.whereis()
+        |> Process.monitor()
+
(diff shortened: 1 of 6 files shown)


magnetised:
> Easy win! Would like to see a test in the `manager_test.exs` where we test some of these things

Your wish is my command.

packages/sync-service/test/electric/connection/manager_test.exs
```diff
@@ -182,6 +182,26 @@ defmodule Electric.Connection.ConnectionManagerTest do
 
       assert StatusMonitor.status(stack_id) in [:waiting, :starting]
     end
+
+    test "backtracks the status when the canary goes down", %{stack_id: stack_id} do
```
msfstef, line 186: 🔥

balegas:
great. doesn't look super elegant but happy it works :D

PR #3204, opened 2025-10-01
fix: rename `ELECTRIC_QUERY_DATABASE_URL` env var

### LINKED ISSUE #3171: Dropping replication slot with ELECTRIC_QUERY_DATABASE_URL fails
**Versions**

- Client: platform, 1.0.10
- Electric: 1.0.11

**Bug description**
Electric attempts to drop replication slot but fails as it seems to be using the query database url when it is set instead of `DATABASE_URL`.

Here is the logs from my backend: https://gist.github.com/jay-babu/e2b2efc4d900b4e9c22dccbe5b93ec95. This issue resulted in electric restarting every 5 minutes which led to all clients getting a 409 and then needing a full resync. This was an endless cycle. The fix was to unset the URL and redeploy. This recreated the slot. NOTE: I had ELECTRIC_MANUAL_TABLE_PUBLISHING set to true, which did not let the tables to automatically be added that were there before.

**Expected behavior**
Dropping of replication slot succeeds. And hopefully the tables being added back correctly.

icehaunter:
Closes #3171

### DIFF
--- a/.changeset/tall-bottles-give.md
+++ b/.changeset/tall-bottles-give.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+fix: rename `ELECTRIC_QUERY_DATABASE_URL` to `ELECTRIC_POOLED_DATABASE_URL` env variable to avoid confusion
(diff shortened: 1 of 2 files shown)


balegas:
Don't we want to deprecate it first? Not everyone will read the changelogs.

balegas:
We should also also update docs to make sure it's clear that we should use the same user for the two connections

icehaunter:
@balegas it's configured to read the old value still. I'll update the docs in a separate pr

PR #3207, opened 2025-10-02
fix(sync-service): Don't send flushed message if there's nothing to flush

### LINKED ISSUE #3188: High CPU usage for an Electric instance with no incoming transactions
**Bug description**

For an electric instance with a large number of shapes connected to a db with no activity at all, the base CPU sits at ~100% (so one core is very busy).

From a quick profile, it seems that the culprit is the `ShapeLogCollector` which is being spammed with thousands of calls to `notify_flushed/3`. ~~This is probably due to faulty logic in the `hibernate` handling within the consumers which seem to never actually hibernate but just constantly say they're hibernating.~~

Looks like the consumers hibernate properly but are constantly being woken up by messages from the storage backend.

magnetised:
Electric is showing high CPU in an idle state caused by constant shape consumer activity. This was due to the following loop:

1. consumer gets `:hibernate` timeout,
2. consumer calls `Storage.hibernate`,
3. `PureFileStorage` hibernates, flushes and sends a `:flushed` message to the consumer,
4. consumer hibernates,
5. consumer wakes up due to the `:flushed` message,
6. GOTO 1

This pr fixes this loop by only sending the `:flushed` message if there was any pending data to write.

Fixes #3188

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage/write_loop.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage/write_loop.ex
@@ -303,10 +303,13 @@ defmodule Electric.ShapeCache.PureFileStorage.WriteLoop do
   @doc """
   Flush the buffer if it's not empty
   """
+  def flush_buffer(writer_acc(buffer_size: 0) = acc, _state) do
+    acc
+  end
+
   def flush_buffer(
         writer_acc(
           buffer: buffer,
-          buffer_size: buffer_size,
           last_seen_offset: last_seen_offset,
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/shape_cache/pure_file_storage/write_loop.ex
```diff
@@ -318,22 +318,24 @@ defmodule Electric.ShapeCache.PureFileStorage.WriteLoop do
     if buffer_size > 0 do
```
msfstef: nit: why not make this a function head?

PR #3209, opened 2025-10-02
fix: ensure correct tagging of live requests in metrics

### LINKED ISSUE #3123: We've stopped emitting live_requests metric
We stopped emitting live requests metric. We should either:
 1. check if current sync_requests tracks sync + live requests
 2. send sync and live request separately
 3. or make sync_requests track sync + live request

icehaunter:
Closes #3123

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -132,9 +132,8 @@ defmodule Electric.Plug.ServeShapePlug do
         duration: System.monotonic_time() - conn.private[:electric_telemetry_span][:start_time]
       },
       %{
-        live: assigns[:live],
-        shape_handle:
-          conn.query_params["handle"] || assigns[:active_shape_handle] || assigns[:handle],
+        live: get_live_mode(assigns),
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -146,6 +145,14 @@ defmodule Electric.Plug.ServeShapePlug do
     conn
   end
 
+  defp get_handle(%{response: %{shape_handle: shape_handle}}), do: shape_handle
+  defp get_handle(%{request: %{shape_handle: shape_handle}}), do: shape_handle
+  defp get_handle(_), do: nil
+
+  defp get_live_mode(%{response: %{params: %{live: live}}}), do: live
+  defp get_live_mode(%{request: %{params: %{live: live}}}), do: live
```
msfstef, line 153: if a request comes in as a live request, but we instead do not serve them in live mode, should we mark the request as live in metrics?

msfstef (approved):

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
icehaunter: I think that's OK

PR #3210, opened 2025-10-02
chore: ensure correct publishing order for JS packages

icehaunter:
We have a breaking publish job every time the client is updated, because by default `changeset publish` uses a non `-r` version of pnpm publish and thus ignores topological sorting. This change keeps the tagging and all that, but instead uses `-r` and thus should publish packages in correct order and avoid errors


icehaunter:
It was always first publishing, then tagging, just under the hood @msfstef

PR #3211, opened 2025-10-02
chore(sync-service): Add test that validates consumer remains hibernated

magnetised:
As @msfstef pointed out, we didn't have a test for the consumer's hibernation.

This required properly wiring the hibernation timeout through the stack config.

I also took the opportunity to use GenServer's built-in timeout functionality.


magnetised:
> I chose not to use the built-in timeout because it felt that a generic `timeout` message is harder to reason about being tied to hibernation after some time, so for readabilty. I'm ok with this change if you disagree

Right, getcha. That makes sense. The `:timeout` is a bit unhelpful. Despite that I think it's better to use the built in functionality. If only you could add a `:tag` like `Process.monitor/2`.

Issue #3186, opened 2025-09-29
Publication updates fail for all shapes if one is failing


msfstef:
**Bug description**
Since publication updates are batched, failures to update the publication for a single relation will cause all other relations to also fail for that update - if requests for a relation that can't be added keep coming then all other relations never make it either and errors keep getting returned.

Also if the ability to add a relation changes during the lifetime of the sync service it gets into a state where it can't update anything. This issue also requires investigating this a bit more as we've seen production error.

**Expected behavior**
Relations that can't make it into the publication should fail independently and the appropriate errors returned only to requests relating to it.

PR #3218, opened 2025-10-02
Msfstef/publication update independent failures

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3186

This PR includes significant changes to `Electric.Postgres.Configuration` and `Electric.Replication.PublicationManager`.

### Changes
- Ensure PubMan handles per-relation errors such that only relevant shapes fail
- If any individual relation fails to be configured, relevant shapes are terminated
- If global error (e.g. connection error) occurs, propagate error to all requesters but do not invalidate shapes explicitly
- Got rid of retries - we don't do them everywhere, e.g. snapshots, and should be handled at network layer (see @robacourt 's work)
- Configuration now returns map with configuration result for each relation it either touched or detected to be invalid
- Validating the configuration returns the same result for simplicity of API, with different error codes since it cannot actually modify the relations
- PubMan now also checks that the publication 1) exists, 2) is configured to publish all operations, and 3) is owned by the Electric PG user.
  - These are all checks done once at the start of setting up the replication connection and publication, but are replicated in PubMan which will be doing these before every publication update/check, which also occur periodically, so we can detect something going wrong
  - These checks will also set the `can_alter_publication?` flag if the ownership changes so appropriate results are returned for shape requests
- Only log errors for unknown errors when failing to configure a relation, everything else is a warning
- _Slightly_ more generous with calling `ALTER TABLE` and `ALTER PUBLICATION` when fixing misconfigured relations just for the sake of API simplicity - no reason to be so cheap for something that doesn't happen often at the expense of clarity IMO.

## TODOs
- If the publication does not exist, we should notify the connection manager so that the whole stack is restarted. We are not doing this right now, we are just shutting down the publication manager which just restarts the replication supervisor but doesn't actually fix the problem. I'll add this as a separate PR.
- Cover case where we don't have ownership over the publication, but we do over the table, so we would need to set the replica identity without trying to add it to the publication.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/configuration.ex
+++ b/packages/sync-service/lib/electric/postgres/configuration.ex
@@ -9,13 +9,13 @@ defmodule Electric.Postgres.Configuration do
 
   @type relation_filters :: PublicationManager.relation_filters()
   @type relations_configured :: %{
-          Electric.oid_relation() => :ok | {:error, :relation_invalidated | term()}
+          Electric.oid_relation() => :ok | {:error, :schema_changed | term()}
         }
   @type relations_checked :: %{
           Electric.oid_relation() =>
             :ok
             | {:error,
-               :relation_invalidated
+               :schema_changed
                | :relation_missing_from_publication
                | :misconfigured_replica_identity}
         }
@@ -101,7 +101,7 @@ defmodule Electric.Postgres.Configuration do
         Enum.map(valid, &{&1, :ok}),
         Enum.map(to_add, &{&1, {:error, :relation_missing_from_publication}}),
         Enum.map(to_fix, &{&1, {:error, :misconfigured_replica_identity}}),
-        Enum.map(to_invalidate, &{&1, {:error, :relation_invalidated}})
+        Enum.map(to_invalidate, &{&1, {:error, :schema_changed}})
       ]
       |> List.flatten()
       |> Map.new()
@@ -174,7 +174,7 @@ defmodule Electric.Postgres.Configuration do
         Enum.map(to_add, &{&1, add_table_to_publication(conn, publication_name, &1)})
       )
       |> Enum.concat(Enum.map(valid, &{&1, :ok}))
-      |> Enum.concat(Enum.map(to_invalidate, &{&1, {:error, :relation_invalidated}}))
+      |> Enum.concat(Enum.map(to_invalidate, &{&1, {:error, :schema_changed}}))
       |> Map.new()
 
     configuration_result
--- a/packages/sync-service/lib/electric/postgres/configuration.ex
+++ b/packages/sync-service/lib/electric/postgres/configuration.ex
@@ -91,7 +91,7 @@ defmodule Electric.Postgres.Configuration do
     %{
       valid: valid,
       to_add: to_add,
-      to_fix_replica_identity: to_fix,
+      to_reconfigure: to_reconfigure,
       to_invalidate: to_invalidate
     } =
       determine_publication_relation_actions!(conn, publication_name, expected_rels)
@@ -100,13 +100,13 @@ defmodule Electric.Postgres.Configuration do
       [
         Enum.map(valid, &{&1, :ok}),
         Enum.map(to_add, &{&1, {:error, :relation_missing_from_publication}}),
-        Enum.map(to_fix, &{&1, {:error, :misconfigured_replica_identity}}),
+        Enum.map(to_reconfigure, &{&1, {:error, :misconfigured_replica_identity}}),
         Enum.map(to_invalidate, &{&1, {:error, :schema_changed}})
       ]
       |> List.flatten()
       |> Map.new()
 
-    if MapSet.size(to_add) == 0 and MapSet.size(to_fix) == 0 do
+    if MapSet.size(to_add) == 0 and MapSet.size(to_reconfigure) == 0 do
       Logger.info(fn ->
         tables = for {_, rel} <- valid, do: Utils.relation_to_sql(rel)
 
@@ -133,6 +133,10 @@ defmodule Electric.Postgres.Configuration do
     Postgrex.transaction(conn, &do_configure_publication!(&1, publication_name, new_relations))
     |> case do
       {:ok, relations_configured} ->
+        Logger.debug(
+          "Publication configuration committed - result: #{inspect(relations_configured)}"
+        )
+
         relations_configured
 
       {:error, reason} ->
@@ -147,24 +151,27 @@ defmodule Electric.Postgres.Configuration do
       valid: valid,
       to_add: to_add,
       to_drop: to_drop,
-      to_fix_replica_identity: to_fix,
+      to_reconfigure: to_reconfigure,
       to_invalidate: to_invalidate
     } =
       determine_publication_relation_actions!(conn, publication_name, new_publication_rels)
 
-    to_add = MapSet.union(to_add, to_fix) |> MapSet.to_list()
-    to_drop = MapSet.to_list(to_drop)
-    to_invalidate = MapSet.to_list(to_invalidate)
+    # Re-add tables that need fixing
+    to_add = MapSet.union(to_add, to_reconfigure)
 
-    if to_drop != [] or to_add != [] do
+    if MapSet.size(to_add) > 0 or MapSet.size(to_drop) > 0 do
+      to_add_list = for {_, rel} <- to_add, do: Utils.relation_to_sql(rel)
+      to_drop_list = for {_, rel} <- to_drop, do: Utils.relation_to_sql(rel)
+      to_invalidate_list = for {_, rel} <- to_invalidate, do: Utils.relation_to_sql(rel)
+
       Logger.info(
         "Configuring publication #{publication_name} to " <>
-          "drop #{inspect(to_drop)} tables, and " <>
-          "add #{inspect(to_add)} tables " <>
-          "- skipping altered tables #{inspect(to_invalidate)}",
-        publication_alter_drop_tables: to_drop,
-        publication_alter_add_tables: to_add,
-        publication_alter_invalid_tables: to_invalidate
+          "drop #{inspect(to_drop_list)} tables, and " <>
+          "add #{inspect(to_add_list)} tables " <>
+          "- skipping altered tables #{inspect(to_invalidate_list)}",
+        publication_alter_drop_tables: to_drop_list,
+        publication_alter_add_tables: to_add_list,
+        publication_alter_invalid_tables: to_invalidate_list
       )
     end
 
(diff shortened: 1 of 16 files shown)


alco:
> Cover case where we don't have ownership over the publication, but we do over the table, so we would need to set the replica identity without trying to add it to the publication.

Why is this a problem that needs addressing? We're talking about Electric's ownership over its own publication vs user tables. If the user's permissions policy doesn't prevent them from giving Electric the DB role that owns user tables, there's no reason for them to manually manage the publication. I just want to highlight that this is talking about Electric doing _more work_ for a case where it can access user's DB resources but cannot access its own publication.

The one case this would address is where Electric has ownership of _some_ tables. But in that case, the publication already has to be manually managed and the other tables must have their replica identity changed by the user. It should be the user who changes the replica identity for all of their tables.

msfstef:
> The one case this would address is where Electric has ownership of some tables. But in that case, the publication already has to be manually managed and the other tables must have their replica identity changed by the user. It should be the user who changes the replica identity for all of their tables.

Yes I was thinking of the case where 1) the publication cannot be altered by Electric (`can_alter_publication?` is false) but 2) Electric has permission over _some_ tables, and could thus correctly set their replica identity. The `can_alter_publication? = false` case is basically a fallback for Electric to work even without setting the manual publishing flag, which made me think that we should also try to maximally configure things to work as much as our permissions allow us.

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -7,296 +7,308 @@ defmodule Electric.Postgres.Configuration do
...
+                raise Electric.DbConfigurationError.publication_not_owned(publication_name)
+            end
+
+          _ ->
+            raise Electric.DbConfigurationError.publication_missing_operations(publication_name)
+        end
+
+      %{num_rows: 0} ->
+        raise Electric.DbConfigurationError.publication_missing(publication_name)
+    end
```
robacourt, line 62: These queries should deal with the database pool not existing, catching the exits much like EtsInspector does

msfstef: I'm now catching and rescuing all relevant errors but not parsing them in the same way as ETS inspector - it would be good to unify connection error handling (either with a macro or a wrapper library?) so that we can consistently handle these errors as well

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -7,296 +7,308 @@ defmodule Electric.Postgres.Configuration do
...
+  end
```
icehaunter, line 63: I don't see a reason to make this 2 queries (or so nested)

```suggestion
  def check_publication_status!(conn, publication_name) do
    query =
      """
      SELECT
        pubinsert AND pubupdate AND pubdelete AND pubtruncate as all_operations,
        pg_get_userbyid(p.pubowner) = current_role as is_owned
      FROM pg_publication WHERE pubname = $1
      """

    Postgrex.query!(conn, query, [publication_name])
    |> one_or_nil()
    |> case do
      %{"all_operations" => true, "is_owned" => true} ->
        :ok

      %{"is_owned" => false} ->
        raise Electric.DbConfigurationError.publication_not_owned(publication_name)

      %{"all_operations" => false} ->
        raise Electric.DbConfigurationError.publication_missing_operations(publication_name)

      _ ->
        raise Electric.DbConfigurationError.publication_missing(publication_name)
    end
  end

  defp one_or_nil(%Postgrex.Result{rows: [row], columns: cols}),
    do: Enum.zip(cols, row) |> Map.new()

  defp one_or_nil(_), do: nil
```

packages/sync-service/test/electric/replication/publication_manager_db_test.exs
```diff
@@ -50,6 +53,18 @@ defmodule Electric.Replication.PublicationManagerDbTest do
       end
     )
 
+    # send message when publication is configured to avoid timing issues
```
alco: Presumably this can be achieved with `Repatch.spy()` but I personally haven't been able to get it to work when I was working on something.

packages/sync-service/test/electric/replication/publication_manager_db_test.exs
```diff
@@ -58,7 +73,7 @@ defmodule Electric.Replication.PublicationManagerDbTest do
     %{pub_mgr_opts: pub_mgr_opts, relation: relation, relation_with_oid: {relation_oid, relation}}
   end
 
-  describe "add_shape()" do
+  describe "add_shape/3" do
```
alco, line 73: I never understood this convention. More often than note, when a function's arity changes in the module, nobody remembers to update it in the test's `describe` block.

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -128,19 +122,14 @@ defmodule Electric.Replication.PublicationManager do
   def init(opts) do
     opts = Map.new(opts)
 
+    Process.flag(:trap_exit, true)
```
alco: I know we aren't diligent about it everywhere, but IMO every setting of trap_exit to true should be accompanied by a comment explaining the reasoning.

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -155,192 +144,186 @@ defmodule Electric.Replication.PublicationManager do
...
-          |> MapSet.to_list()
-          |> ShapeCleaner.remove_shapes_for_relations(stack_id: state.stack_id)
-
-          tables = Enum.map(relations, fn {_oid, relation} -> Utils.relation_to_sql(relation) end)
-          message = publication_error_message(error_type, tables, state)
-          error = %Electric.DbConfigurationError{type: reason, message: message}
-
-          state = reply_to_waiters({:error, error}, state)
-          {:ok, %{state | next_update_forced?: false}}
+  def handle_info({:EXIT, _, reason}, state), do: {:stop, reason, state}
```
alco: Would be nice to know which processes we're expecting an exit signal from. In this case I think it's only the supervisor?

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -7,296 +7,308 @@ defmodule Electric.Postgres.Configuration do
...
+  """
+  @spec check_publication_status!(Postgrex.conn(), String.t()) :: :ok
+  def check_publication_status!(conn, publication_name) do
+    case Postgrex.query!(
+           conn,
+           "SELECT pubinsert, pubupdate, pubdelete, pubtruncate FROM pg_publication WHERE pubname = $1",
+           [publication_name]
+         ) do
+      %{num_rows: 1, columns: cols, rows: [row]} ->
+        publication = Enum.zip(cols, row) |> Map.new()
```
alco: Nit: IMO this is redundant. We already know the order of the columns in the query result because we define it ourselves in the query. So it can be simply annotated in the code in your preferred way:

```
# [pubinsert, pubupdate, pubdelete, pubtruncate]
[true, true, true, true]
```
or
```
[true = _pubinsert, true = _pubupdate, true = _pubdelete, true = _pubtruncate]
```
or
```
Enum.all?(cols, & &1 == true)
```

packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -7,296 +7,308 @@ defmodule Electric.Postgres.Configuration do
...
+      determine_publication_relation_actions!(conn, publication_name, expected_rels)
+
+    configuration_result =
+      [
+        Enum.map(to_preserve, &{&1, {:ok, :validated}}),
+        Enum.map(to_add, &{&1, {:error, :relation_missing_from_publication}}),
+        Enum.map(to_reconfigure, &{&1, {:error, :misconfigured_replica_identity}}),
+        Enum.map(to_invalidate, &{&1, {:error, :schema_changed}})
+      ]
+      |> List.flatten()
```
alco:
```suggestion
      |> Stream.concat()
```

packages/sync-service/test/electric/replication/publication_manager_db_test.exs
msfstef: while the convention makes sense as it uniquely identifies the function based on arity, I agree that more often than not it goes out of sync - I think we could go for an entirely semantic name without parentheses or arity

that being said, I renamed it just for consistency with most of our test suites cause I don't like inconsistency haha

packages/sync-service/lib/electric/replication/publication_manager.ex
msfstef: you're right I'll add a comment - tbh having to trap exits to deal with connection issues is not something I particularly like, but until we have a unified way of dealing with them it's what we need to do I guess

packages/sync-service/lib/electric/replication/publication_manager.ex
msfstef: yes exactly - figured I'd add a catch all because we're now trapping exits to still simulate proper behaviour - do you reckon it's better if I explicitly only handle the `:shutdown` clause so we can see errors if anything else happens?

packages/sync-service/test/electric/replication/publication_manager_db_test.exs
msfstef: Good point - it's `Repatch.notify` and it returns a notification message format whenever it is called - will use that instead, cleaner

packages/sync-service/lib/electric/postgres/configuration.ex
msfstef: good idea !

packages/sync-service/lib/electric/replication/publication_manager.ex
msfstef: I've added a comment but am still cpaturing all exit signals here - let me know if you think we should do something different.

packages/sync-service/test/electric/replication/publication_manager_db_test.exs
msfstef: (resolving the comment but noting this, mayb we can open a discussion on conventions?)

packages/sync-service/lib/electric/replication/publication_manager.ex
alco: Hmm, after reading the new comments I still don't understand that point of trapping exits in this process. If I'm reading this right ,when any linked process exits, PubMan will receive an `EXIT` message and will stop itself with the same reason. But isn't it how exit signal propagation already works when trap_exit=false?

msfstef: Yes - the problem is being able to catch exits in `try...catch` clauses, specifically when doing connection operations. If we try to checkout a connection and the pool no longer exists, it would exit, and we can't handle that exit in a `try...catch` unless we are trapping exits (I'm pretty sure at least?)

msfstef: ah I just tried for this particular error it seems that it works without trapping exits - so we don't need to trap exits for it? tagging @robacourt to check this makes sense, this is for `:exit, {_, {DBConnection.Holder, :checkout, _}}` exits

msfstef: I've removed exit trapping - I was confused from previous times I had encountered similar cases, but I understand that exit calls within the `try...catch` will be caught even if the process as a whole is not trapping exits.

alco: > Yes - the problem is being able to catch exits in `try...catch` clauses, specifically when doing connection operations. If we try to checkout a connection and the pool no longer exists, it would exit, and we can't handle that exit in a `try...catch` unless we are trapping exits (I'm pretty sure at least?)

Catching exits with `try ... catch` is orthogonal to the `trap_exit` flag. So if this is the only reason you've added trapping exits to PubMan, feel free to remove it.

I can elaborate but it'll take me some time to write it up so I'll add that as a separate comment.

msfstef: Will appreciate it thank you!

PR #3221, opened 2025-10-02
Add `shardSubdomain` shape option.

thruflo:
Adds an optional `shardSubdomain` shape option to auto-shard the url subdomain in development:

```ts
const url = isProd
  ? 'https://example.com' 
  : 'http://localhost:3000'

  url,
  shardSubdomain: 'localhost'
})
```

```ts
/**
 * Enable subdomain sharding to bypass browser HTTP/1.1 connection limits.
 * This is useful in local development and is enabled by default for localhost URLs.
 *
 * See https://electric-sql.com/docs/guides/troubleshooting#slow-shapes-mdash-why-are-my-shapes-slow-in-the-browser-in-local-development
 *
 * When sharded, each shape stream gets a unique subdomain (e.g., `a7f2c.localhost`),
 * which bypasses the browser HTTP/1.1 connection limits. This avoids the need to serve
 * the development server over HTTP/2 (and thus HTTPS) in development.
 *
 * Options:
 * - `'localhost'` - Automatically shard `localhost` and `*.localhost` URLs (the default)
 * - `'always'` - Shard URLs regardless of the hostname
 * - `'never'` - Disable sharding
 * - `true` - Alias for `'always'`
 * - `false` - Alias for `'never'`
 *
 * @default 'localhost'
 *
 * @example
 * { url: 'http://localhost:3000/v1/shape', shardSubdomain: 'localhost' }
 * // → http://a1c2f.localhost:3000/v1/shape
 *
 * @example
 * { url: 'https://api.example.com', shardSubdomain: 'localhost' }
 * // → https://api.example.com
 *
 * @example
 * { url: 'https://localhost:3000', shardSubdomain: 'never' }
 * // → https://localhost:3000
 *
 * @example
 * { url: 'https://api.example.com', shardSubdomain: 'always' }
 * // → https://b2d3g.api.example.com
 */
shardSubdomain?: ShardSubdomainOption
```

This solves the [slow shapes in development](https://electric-sql.com/docs/guides/troubleshooting#slow-shapes-mdash-why-are-my-shapes-slow-in-the-browser-in-local-development) problem without needing http2 or system level deps like Caddy or mkcert.

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -290,28 +290,38 @@ export interface ShapeStreamOptions<T = never> {
 
   /**
    * Enable subdomain sharding to bypass browser HTTP/1.1 connection limits.
+   * This is useful in local development and is enabled by default for localhost URLs.
    *
-   * Each shape gets a unique subdomain (e.g., `a7f2c.localhost:3000`),
-   * providing independent connection pools. Eliminates need for HTTP/2,
-   * Caddy, or trusted certificates in development.
+   * See https://electric-sql.com/docs/guides/troubleshooting#slow-shapes-mdash-why-are-my-shapes-slow-in-the-browser-in-local-development
+   *
+   * When sharded, each shape stream gets a unique subdomain (e.g., `a7f2c.localhost`),
+   * which bypasses the browser HTTP/1.1 connection limits. This avoids the need to serve
+   * the development server over HTTP/2 (and thus HTTPS) in development.
    *
    * Options:
-   * - `'localhost'` - Only shard localhost and *.localhost URLs (recommended for development)
-   * - `'always'` - Shard all URLs (use with custom infrastructure that supports wildcard subdomains)
-   * - `'never'` - Disable sharding (default)
+   * - `'localhost'` - Automatically shard `localhost` and `*.localhost` URLs (the default)
+   * - `'always'` - Shard URLs regardless of the hostname
+   * - `'never'` - Disable sharding
    * - `true` - Alias for `'always'`
    * - `false` - Alias for `'never'`
    *
-   * @default 'never'
-   * @example
-   * // Recommended: Opt-in for localhost development
-   * { url: 'http://localhost:3000/v1/shape', shardSubdomain: 'localhost' }
-   * // → http://a7f2c.localhost:3000/v1/shape
+   * @default 'localhost'
    *
    * @example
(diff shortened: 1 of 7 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -282,6 +288,33 @@ export interface ShapeStreamOptions<T = never> {
...
+   * Caddy, or trusted certificates in development.
+   *
+   * Options:
+   * - `'localhost'` - Only shard localhost and *.localhost URLs (recommended for development)
+   * - `'always'` - Shard all URLs (use with custom infrastructure that supports wildcard subdomains)
+   * - `'never'` - Disable sharding (default)
+   * - `true` - Alias for `'always'`
+   * - `false` - Alias for `'never'`
+   *
+   * @default 'never'
```
KyleAMathews: Why default true for localhost? A lot of people will never discover this otherwise and we want this problem just solved.

thruflo: I was thinking some people may already have https://localhost and just generally not to change default behaviour. Who knows what it might break, e.g.: some user code expecting a certain URL.

Hence opt-in.

What I'm thinking is, if this looks good to go, we can publish the patch release and then I'll go through the examples and configure them to use it, so in practice everyone trying Electric will just get the fix built in.

Perhaps I'm being too conservative?

KyleAMathews: It'd only break someone who's explicitly filtering on the host

https://claude.ai/share/1cfd5c47-0d1a-490f-bb64-20e753207062

Which... could be happening I suppose so yeah, we can update starters and docs (we can replace the caddy recommendation now). Maybe make it default in a major release.

thruflo: I've been eyeballing this when looking at all the places we would need to change the code and docs in the examples to apply this across the monorepo. I've swung into the YOLO the default to 'localhost' camp.

If we leave this as is up-to the developer then they have to remember to add the shardSubdomain config to every shape definition. Whereas the current Caddy approach is faff-once and forget. We then have the question of pollute every shape example in the docs with this param or risk people copying snippets from docs and forgetting to add the param in local dev.

So, I think it's going to cause much less confusion and grief to default this to `'localhost'`. That way we can just strip out the Caddy stuff and everything will just work. No need to change code snippets.

KyleAMathews: Yeah good point. It would be tedious and ugly to always add this. And there should be few to no people who are filtering localhost API calls by host.

PR #3223, opened 2025-10-03
Strip out caddy-for-http2 usage and docs.

thruflo:
Update the docs and examples to stop using Caddy for HTTP/2, following on from #3221.


KyleAMathews (approved):

xamgore:
I suggest looking into https://github.com/vitejs/vite/issues/4184#issuecomment-3276550729, they just merged it.

PR #3228, opened 2025-10-04
Revise starter to use localhost port sharding.

thruflo:
This replaces the reverted subdomain sharding with port sharding. It also keeps the changes purely scoped to the starter example for now, so no changes needed to the typescript client.

Works fine with the cookie auth:

image


thruflo:
Suggested plan to deploy the starter demo https://github.com/electric-sql/electric/pull/3634

PR #3230, opened 2025-10-05
WIP Further reduce memory footprint by collapsing consumer processes and deduplicating shape structs

alco:
This is a logical continuation of https://github.com/electric-sql/electric/pull/3198. The separate PR is for ease of review.

Key changes made here:
- The ConsumerSupervisor tree is collapsed into a single Consumer process (@magnetised's brilliant idea)
- Stack-wide config is stored in ETS and is looked up directly by Consumer processes when needed. This avoid copying the same config through ShapeCache -> DynamicConsumerSupervisor
- Shape structs are looked up in ShapeStatus directly by Consumer processes. This significantly reduces the mem footprint of DynamicConsumerSupervisor (which keeps every child's config in its state) and allows us to eliminate more copying by no longer copying the whole ShapeStatus state into every task that starts one Consumer process.

Here's a visual progression that demonstrates preliminary gains (I still have to fix tests and refactor config passing). In all of the following, Electric is started with 50K shapes in its persistent storage; I wait for its uptime to reach 2 minutes, take a screenshot and shut it down with a SIGTERM.

`main`:
mem_stats_main

`alco/config-slimdown` branch which is the previous PR https://github.com/electric-sql/electric/pull/3198:
mem_stats_pre_optim

After 43a6b26a8d0cbce215a39fb4047c769785dfd0c2 (collapsing consumer processes):
mem_stats_without_consumer_sup

After 196a21e402053eeb2bb6812149245edec0004f1b (trim down DynamicConsumerSupervisor's state):
mem_stats_with_less_copying


magnetised:
Closing in favour of #3389 and #3390

PR #3234, opened 2025-10-06
Unlink the lock breaker connection

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3232

I've tested this manually by raising in the lock breaker connection and reproduced the error.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -582,6 +582,10 @@ defmodule Electric.Connection.Manager do
           stack_id: state.stack_id
         )
 
+      # unlink the lock breaker so that if it crashes it does not affect the manager,
+      # since it is a one shot fix attempt anyway
+      Process.unlink(breaker_pid)
+
       lock_name = Keyword.fetch!(state.replication_opts, :slot_name)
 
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -582,6 +582,10 @@ defmodule Electric.Connection.Manager do
           stack_id: state.stack_id
         )
 
+      # unlink the lock breaker so that if it crashes it does not affect the manager,
+      # since it is a one shot fix attempt anyway
+      Process.unlink(breaker_pid)
```
alco: Shouldn't we do this inside `LockBreaker.start_link()` and rename that function it just `start()` or `spawn()`?

alco (approved):
Good catch!

packages/sync-service/lib/electric/connection/manager.ex
msfstef: yeah I thought of doing that and was hoping simple conn would have a method - I'll do that since you had the same feeling

msfstef: Renamed it to `start` with the unlink inside of it

PR #3238, opened 2025-10-06
feat(sync-service):  Keep consumers on loss of replication client

robacourt:


packages/sync-service/lib/electric/application.ex
```diff
@@ -44,12 +44,6 @@ defmodule Electric.Application do
     # server and the database connection manager. The latter is responsible for establishing
     # all needed connections to the database (acquiring the exclusive access lock, opening a
     # replication connection, starting a connection pool).
-    #
-    # Once there is a DB connection pool running, Connection.Manager will start the singleton
-    # `Electric.Replication.Supervisor` which is responsible for starting the shape log collector
-    # and individual shape consumer process trees.
-    #
-    # See the moduledoc in `Electric.Connection.Supervisor` for more info.
```
alco, line 52: This whole comment is outdated and badly misplaced, I would suggest removing it and adding the following paragraph to the comment block just above the `def children_application` line:
```
    # The root application supervisor starts the core service processes, such as the HTTP
    # server for the HTTP API and telemetry exporters, and a single StackSupervisor, basically
    # making the application run in a single-tenant mode where all API requests are forwarded
    # to that sole tenant.
```

alco (commented):
Amazing work!

>    Replication.Supervisor (or perhaps it should be called the ShapeSystemSupervisor)
>    Connection.Manager.Supervisor (or perhaps it should be called the ConnectionSystemSupervisor)

Sounds good to me, but I would keep it consistent with the code comments and call them `ShapeSubsystemSupervisor` and `ConnectionSubsystemSupervisor`.

---

The only change that bugs me is that the whole replication supervisor has to be started before the connection manager starts initializing database connections. Since the two subsystems can now fail independently of each other, it would only be fair for them to start in parallel and reduce the time until the stack becomes ready.

The consumers_ready message needs to be transformed into a StatusMonitor entry which the connection manager can look up and wait on.

If the connection manager detects a timeline change,  there's no other recourse than terminating the shape subsystem (even if it's still starting up) and starting it with a clean state. I don't particularly see the reason to treat the case of count_shapes==0 differently: this case seems like a rare outlier.

---

Regarding the coordination with my PR, I can rebase yours on top of mine myself, if mine is merged first. I don't think it makes sense for you to switch context to the new way of config passing here, you can just keep working in context of `main` and bring the PR to completion.

It will be easier for me to rebase it later since that's basically what I've already done in my PR, a number of times actually as I've had to resolve conflicts with multiple other PRs that got merged to main.

robacourt:
> I don't particularly see the reason to treat the case of count_shapes==0 differently: this case seems like a rare outlier.

It's not rare at all, it happens when the system first starts up because the slot is new. It would seem odd to start the replication client twice on first start-up. I specifically added that condition for this case.

robacourt:
> The only change that bugs me is that the whole replication supervisor has to be started before the connection manager starts initializing database connections. Since the two subsystems can now fail independently of each other, it would only be fair for them to start in parallel and reduce the time until the stack becomes ready.

So there's no way as far as I know to do that in the supervision tree, so I assume you mean for ShapeCache to load the shapes as part of a `:continue` rather than as part of `init`?

alco:
> > I don't particularly see the reason to treat the case of count_shapes==0 differently: this case seems like a rare outlier.
> 
> It's not rare at all, it happens when the system first starts up because the slot is new. It would seem odd to start the replication client twice on first start-up. I specifically added that condition for this case.

Ah, I see. Because the new slot creation sets `purge_all_shapes` to true.

I don't think we should reset storage in this case. When the system first starts up, it doesn't have any timeline stored, so the timeline check returns `:ok`. We should be able to identify this state and skip resetting the storage.

So what should be happening is this:
- timeline check returns :no_previous_timeline or similar
- conn man skips checking state.purge_all_shapes and just moves on

If we do have a previous timeline stored, then state.purge_all_shapes is checked together with the timeline check to decide whether to reset the storage.

alco:
> > The only change that bugs me is that the whole replication supervisor has to be started before the connection manager starts initializing database connections. Since the two subsystems can now fail independently of each other, it would only be fair for them to start in parallel and reduce the time until the stack becomes ready.
> 
> So there's no way as far as I know to do that in the supervision tree, so I assume you mean for ShapeCache to load the shapes as part of a `:continue` rather than as part of `init`?

Yes, that seems to be the way to do it. But we can no longer just send a message to connection manager because it may not be up by that time. So the coordination about consumers being ready needs to happen via StatusMonitor, IMO.

robacourt:
> I don't think we should reset storage in this case. When the system first starts up, it doesn't have any timeline stored, so the timeline check returns `:ok`. We should be able to identify this state and skip resetting the storage.

I like the idea of skipping the reset, but it's not the timeline that triggers it, it's `replication_client_created_new_slot` being sent by the ReplicationClient. I'm not quite sure at this stage how I would stop that.

(edit) Oh, actually you're saying I could use the `:no_previous_timeline` to not check `purge_all_shapes`. Ok, that could work. Thanks!

robacourt:
> The consumers_ready message needs to be transformed into a StatusMonitor entry which the connection manager can look up and wait on.

Where we called `consumers_ready` we calling `ShapeLogCollector.set_last_processed_lsn` which in turn calls `StatusMonitor.mark_shape_log_collector_ready` so I assume that should be enough unless I'm missing something?

alco:
@robacourt oh yes, this way it's even better: when Connection Manager is ready to start streaming, it checks whether the ShapeLogCollector itself is ready. If not, it waits for it to become ready and then instructs the replication client to start streaming. You've hit the nail on the head!

robacourt:
implemented as #3344

PR #3239, opened 2025-10-06
chore: Fix test flakes

msfstef:


packages/sync-service/test/electric/postgres/inspector/ets_inspector_test.exs
```diff
@@ -500,16 +500,45 @@ defmodule Electric.Postgres.Inspector.EtsInspectorTest do
...
+               send(test_pid, :pool_busy)
+               Postgrex.query!(conn, "SELECT PG_SLEEP(10)", [])
+             end
+           )
+         end}
+      )
 
-    setup %{db_conn: conn} do
-      start_link_supervised!({Task, fn -> Postgrex.query!(conn, "SELECT PG_SLEEP(10)", []) end})
+      assert_receive :pool_busy
```
robacourt, line 539: If we `wait_for_pool_to_be_busy`, do we really need to wait for this as well?

msfstef: because in `wait_for_pool_to_be_busy` we do a postgrex query, which might occur before the task actually checks out the conn and runs the sleep command, and then the task fails because it can't check out a conn

robacourt: shouldn't the wait function handle that error and retry? the issue with what you have is that there's a race condition where we can receive the message then the wait function errors before the sleep command has started

PR #3241, opened 2025-10-06
docs: Add type-safe WHERE clause generation to auth guide

KyleAMathews:
Update auth guide and examples to demonstrate type-safe WHERE clause generation using Drizzle and Kysely query builders.

Changes:
- Add new section on type-safe WHERE clause generation with Drizzle and Kysely
- Update main proxy auth example to use Drizzle with type-safe operators
- Update tanstack-db examples (projects.ts, todos.ts) to use Drizzle with:
  - sql.identifier() and column.name for unqualified column names
  - Proper parameterized query handling via params[1], params[2], etc.
- Add comment to proxy-auth example pointing to type-safety docs
- Explain how Electric expects WHERE clauses without table prefixes
- Document parameter passing format for Electric's HTTP API

### DIFF
--- a/examples/proxy-auth/app/shape-proxy/route.ts
+++ b/examples/proxy-auth/app/shape-proxy/route.ts
@@ -41,7 +41,7 @@ export async function GET(request: Request) {
 
   // Only query orgs the user has access to.
   if (!user.isAdmin) {
-    originUrl.searchParams.set(`where`, `"org_id" = "${user.org_id}"`)
+    originUrl.searchParams.set(`where`, `"org_id" = ${user.org_id}`)
   }
 
   const response = await fetch(originUrl)


website/docs/guides/auth.md
```diff
@@ -166,6 +175,99 @@ export async function GET(request: Request) {
...
+  // ... setup code ...
+
+  if (!user.roles.includes("admin")) {
+    // TypeScript will error if you reference invalid columns
+    const query = db
+      .selectFrom("users")
+      .selectAll()
+      .where((eb) =>
+        eb.and([eb("org_id", "=", user.org_id), eb("status", "=", "active")])
+      )
```
igalklebanov: equivalent:

```suggestion
      .where("org_id", "=", user.org_id)
      .where("status", "=", "active")
```

KyleAMathews: thanks!

examples/proxy-auth/app/shape-proxy/route.ts
```diff
@@ -41,7 +41,10 @@ export async function GET(request: Request) {
 
   // Only query orgs the user has access to.
   if (!user.isAdmin) {
-    originUrl.searchParams.set(`where`, `"org_id" = ${user.org_id}`)
+    // Note: For production, use a type-safe query builder like Drizzle or Kysely
```
thruflo: Perhaps "it's possible to". I.e.: not everyone must use a type safe clause generator. It's also fine to use simple strings in many cases.

examples/tanstack-db-web-starter/src/routes/api/projects.ts
```diff
@@ -13,8 +16,26 @@ const serve = async ({ request }: { request: Request }) => {
 
   const originUrl = prepareElectricUrl(request.url)
   originUrl.searchParams.set("table", "projects")
-  const filter = `owner_id = '${session.user.id}' OR '${session.user.id}' = ANY(shared_user_ids)`
-  originUrl.searchParams.set("where", filter)
+
```
thruflo: Generally, I feel that the readability of the example has reduced here.

```ts
const filter = `owner_id = '${session.user.id}' OR '${session.user.id}' = ANY(shared_user_ids)`
originUrl.searchParams.set("where", filter)
```

Is pretty easy for me to scan.

```ts
  // Generate type-safe WHERE clause using ANY operator for array membership
  // Note: We use column.name to get unqualified column names since Electric expects
  // WHERE clauses without table prefixes (just column names)
  const whereExpr = sql`${sql.identifier(projectsTable.owner_id.name)} = ${session.user.id} OR ${session.user.id} = ANY(${sql.identifier(projectsTable.shared_user_ids.name)})`

  const qb = new QueryBuilder()
  const { sql: query, params } = qb
    .select()
    .from(projectsTable)
    .where(whereExpr)
    .toSQL()

  const fragment = query.replace(/^SELECT .* FROM .* WHERE\s+/i, "")
  originUrl.searchParams.set("where", fragment)

  // Add params as individual query parameters: params[1]=value, params[2]=value, etc.
  params.forEach((value, index) => {
    originUrl.searchParams.set(`params[${index + 1}]`, String(value))
  })
```

Is more code with a fair bit more complexity. Perhaps pulling out the `sql.identifier(projectsTable.owner_id.name)` things into named variables could help make the clause easier to parse?

website/docs/guides/auth.md
```diff
@@ -166,6 +175,98 @@ export async function GET(request: Request) {
 }
 ```
 
+#### Type-safe where clause generation
```
thruflo, line 170: This section is great. I wonder if we can just have simple string examples and a comment that says something along the lines of "this is a simple where clause, if you'd like type safe where clause generation, see <link to doc page section>"

thruflo (approved):
I suspect you'll feel that type safe sql building is best-practise enough to want to default to / always present. In which case, this all looks great 👍 

However, as per my comments, I feel the examples are harder to understand and it may be better to document and link to type safe clause generation as an option / advanced pattern.

examples/proxy-auth/app/shape-proxy/route.ts
KyleAMathews: yeah — I don't feel that strongly about using this approach or not — making it software is a good idea.

examples/tanstack-db-web-starter/src/routes/api/projects.ts
KyleAMathews: Yeah true — the string version is very easy to read whereas the latter is harder unless you know drizzle well. So I'll move the basic example back and then say "type safety is an option — see here"

PR #3242, opened 2025-10-07
feat(sync-service): repopulate log collectors state from the status ets

magnetised:
Don't rely on consumer processes existing in order to initialise the shape log collector -- all the information we need is in the shape status ets.

We still need the consumer processes to register in order to synchronise the filters etc with the existing shapes but we only need to do that if the consumer was not in the original state.

This is handled by having an extra `action` param to the subscribe call. Shape consumers being restored don't add their shape to the shape log collector's internal filters while those being newly created do.

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -443,6 +443,6 @@ defmodule Electric.Replication.ShapeLogCollector do
       Support.TestUtils.activate_mocked_functions_for_module(__MODULE__)
     end
   else
-    def test_callback_hook, do: :noop
+    def activate_mocked_functions_from_test_process, do: :noop
   end
 end


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -122,7 +124,32 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+            fn {shape_handle, shape}, {partitions, filter, layers} ->
+              {
+                Partitions.add_shape(partitions, shape_handle, shape),
+                Filter.add_shape(filter, shape_handle, shape),
+                DependencyLayers.add_dependency(layers, shape, shape_handle)
+              }
+            end
+          )
+
+        {:noreply, %{state | partitions: partitions, filter: filter, dependency_layers: layers}}
```
alco, line 152: There is no need to update the state map for every shape. Seems more natural to me to keep these lists in a tuple for the duration of Enum.reduce() and assign them to the state afterwards.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -136,7 +136,7 @@ defmodule Electric.Shapes.Consumer do
       Materializer.subscribe(state.stack_id, shape_handle)
     end
 
-    ShapeLogCollector.subscribe(state.stack_id, state.shape_handle, state.shape)
+    ShapeLogCollector.subscribe(state.stack_id, state.shape_handle, state.shape, state.phase)
```
alco: Call it a nitpick but it doesn't sit right with me that we're assigning a phase to the consumer process for its entire lifetime by storing it in `state`. It would be more appropriate to pass the phase as an argument to the `handle_continue` callback and avoid storing it in the process state. Or perhaps we should just call `ShapeLogCollector.subscribe` inside the `init()` callback? I don't see how this call is related to the initialisation of storage.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: Maybe you're commenting on an old version but I'm doing exactly what you describe...

alco: You're goddamn right! I must have been hallusinatic when I wrote this.

PR #3255, opened 2025-10-07
chore(sync-service): Remove FileStorage

### LINKED ISSUE #3244: Deprecate FileStorage
As part of #3243 we are no longer starting processes for a shape. FileStorage requires an active process to perform reads so will require a lot of special casing to support under a model where we no longer definitely have any running processes for a shape. Seems like a good time to deprecate this and avoid the overhead.

magnetised:
Fixes #3244


packages/sync-service/config/runtime.exs
```diff
@@ -126,7 +126,9 @@ storage_spec =
           {Electric.ShapeCache.InMemoryStorage, []}
 
         legacy_file when legacy_file in ["file", "legacy_file"] ->
-          {Electric.ShapeCache.FileStorage, storage_dir: shape_path}
+          raise RuntimeError,
```
alco, line 129: nit: writing `RuntimeError` and `message` is redundant. This is equivalent to `raise "..."`. But if you simply prefer to be explicit, never mind.

packages/sync-service/lib/electric/application.ex
```diff
@@ -186,6 +186,15 @@ defmodule Electric.Application do
       get_env_lazy(opts, :storage, fn ->
         Electric.Config.Defaults.storage(Keyword.take(opts, [:storage_dir]))
       end)
+      |> case do
+        {Electric.ShapeCache.FileStorage, _opts} ->
```
alco, line 198: Just a reminder that neither `config/runtime.exs` nor this function are invoked in Cloud. We'll need to simply remove mentions of `FileStorage` there without any exceptions.

magnetised: Yeah, I'm hedging around open-source usage. Either via e.g. Phoenix.Sync or via the Docker container

PR #3257, opened 2025-10-08
feat(sync-service): Scale down database connections on replication idle timeout

alco:


balegas:
@alco I agree with your decisions. The main target of this feature is devs and small projects that are tolerant to slow requests. This also looks like the easier approach and we can make the wake up smarter --- it's not a problem to wake up the database at all, we just want to ensure we don't spend hours of inactvity.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -678,6 +687,24 @@ defmodule Electric.Connection.Manager do
...
+  # The replication client exited because it hasn't streamed any new transactions for a while.
+  # This is a signal for all database connections to close and transition Electric into a
+  # scaled down mode.
+  def handle_info(
+        {:EXIT, pid, {:shutdown, {:connection_idle, time}}},
+        %State{replication_client_pid: pid, current_phase: :running} = state
+      ) do
+    time_s = System.convert_time_unit(time, :millisecond, :second)
+
+    Logger.warning(
```
icehaunter: This is an important message, but not sure what it's warning about

```suggestion
    Logger.notice(
```

alco:
I have pushed a commit that makes the return value of `StatusMonitor.status()` return separate status for the connection and shape subsystems. This solves the problem where the replication supervisor's true state was masked by the scaled-down connection mode: `status()` would return `:db_conn_sleeping` regardless of the state of the replication supervisor.

packages/sync-service/lib/electric/connection/restarter.ex
```diff
@@ -0,0 +1,75 @@
...
+    Electric.ProcessRegistry.name(stack_id, __MODULE__)
+  end
+
+  def name(opts), do: name(opts[:stack_id])
+
+  def stop_connection_subsystem(stack_id) do
+    GenServer.cast(name(stack_id), :stop_connection_subsystem)
+  end
+
+  def restart_connection_subsystem(stack_id) do
```
msfstef, line 51: supernit: `restart` also implies stopping and starting, perhaps `wake_connection_subsystem` is more explicit that this only does something when it is sleeping? obviously feel free to ignore this

edit: given we also have `stop_connection_subsystem` I understand that it works for this API - will keep this comment here as the `sleep`/`wake` `stop`/`restart` dualities slightly confused me

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -405,7 +433,10 @@ defmodule Electric.Postgres.ReplicationClient do
                 # new transaction into the shape log store. So, when the applied function
                 # returns, we can safely advance the replication slot past the transaction's commit
                 # LSN.
-                state = update_received_wal(state, Electric.Postgres.Lsn.to_integer(txn.lsn))
+                state =
+                  state
+                  |> update_received_wal(Electric.Postgres.Lsn.to_integer(txn.lsn))
```
msfstef, line 438: love the mini formatting convention fights happening throughout all our PRs

packages/sync-service/lib/electric/replication/schema_reconciler.ex
```diff
@@ -71,9 +71,15 @@ defmodule Electric.Replication.SchemaReconciler do
   end
 
   defp handle_reconcile(state) do
-    state.inspector
-    |> Inspector.list_relations_with_stale_cache()
-    |> handle_diverged_relations(state)
+    case Electric.StatusMonitor.status(state.stack_id) do
```
msfstef, line 74: pubman and schemareconciler I suppose fall into this odd space between connection subsystem and shape subsystem, as they are periodic on a connection and are useless otherwise, but are intimately related with the state of our shapes.

I suppose for just two of them this is a clean enough approach to interrupt scheduled operations rather than reconsider their placement in the supervision tree or something like that

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -761,18 +777,10 @@ defmodule Electric.Shapes.Api do
     after
       # If we timeout, return an up-to-date message
       long_poll_timeout ->
-        # Ensure stack is ready after a long poll timeout, as it might
-        # have failed during this period
-        case hold_until_stack_ready(request.api) do
```
msfstef, line 766: very nice that this is no longer needed

packages/sync-service/lib/electric/shapes/api/params.ex
```diff
@@ -296,6 +296,13 @@ defmodule Electric.Shapes.Api.Params do
 
           message, changeset when is_binary(message) ->
             add_error(changeset, field, message)
+
+          :connection_not_available, changeset ->
```
msfstef: I had added a clause in line 285 that should technically cover this - how did this come up again here, are we not covering all cases? Or can it fail in a deeper validation step (i.e. succeed in a table validation and fail in another column lookup)

packages/sync-service/lib/electric/connection/restarter.ex
alco: Yeah, I'm not happy with the "sleep/wakeup" terminology myself. Using "put to sleep" instead of "stop" in reference to the connection subsystem is confusing because it is reminiscent of process hibernation. On the other hand, using "scale down" is confusing in the sense that the Postgres server doesn't scale its compute down immediately, we're only scaling down Electric's connections to the database.

Anyway, I'm not attached to any of those names but feel like I don't have any better alternatives than this. I'm adding docstrings to `stop_connection_subsystem` and `restart_connection_subsystem` to clear this up a bit.

packages/sync-service/lib/electric/shapes/api/params.ex
alco: I had made this change a while ago, before you made yours. Retested this case just now and it appears this is no longer necessary. Removing.

I tested it by commenting out the lines in `Electric.Shapes.Api` that restart the connection subsystem and hold the request until stack ready. Then I scaled Electric down and sent it a shape request for a new table:

```
$ curl -i http://localhost:3000/v1/shape?table=itemz&offset=-1
HTTP/1.1 503 Service Unavailable
[...]
electric-schema: null

{"message":"Cannot connect to the database to verify the shape. Please try again later."}
```

N.B. Sending a request for an existing shape returns from the shape log without problems.

packages/sync-service/lib/electric/replication/schema_reconciler.ex
alco: Yeah, the unfortunate siblings who cannot associate themselves with neither subsystem fully..

PR #3263, opened 2025-10-08
Add Prettier formatting and CI check

KyleAMathews:
## Summary
- Install prettier as dev dependency
- Add `format` and `format:check` scripts to root package.json
- Create `.prettierignore` for common build artifacts
- Add `format:check` step to TS test CI workflow
- Format all files with prettier


KyleAMathews:
Yeah sounds good — I prefer single quotes & back ticks too — I also realized that I dislike having to format locally and push that commit — so I'm going to add a precommit hook which does it automatically for us.

PR #3265, opened 2025-10-09
fix: Ensure pool process exit does not log unnecessary error

msfstef:
Fixes [this issue](https://electricsql-04.sentry.io/issues/47739402/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream) where the pool exiting logs an error because we are not using a `{:shutdown, error}` tuple.

I've made it exit with an expected shutdown and we strip the reason from the tuple in the manager instead.

I've taken the opportunity to also parse [this Postgrex error](https://electricsql-04.sentry.io/issues/60380806/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream) about not being able to connect to a compute node as retryable.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -984,7 +984,7 @@ defmodule Electric.Connection.Manager do
   defp shutdown_or_reconnect(error, pid_type, state) do
     error =
       error
-      |> strip_shutdown_exit_reason()
+      |> strip_shutdown_atom()
       |> strip_exit_signal_stacktrace()
       |> DbConnectionError.from_error()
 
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -984,6 +984,7 @@ defmodule Electric.Connection.Manager do
   defp shutdown_or_reconnect(error, pid_type, state) do
     error =
       error
+      |> strip_shutdown_exit_reason()
```
robacourt: `strip_exit_signal_stacktrace()` removes the stack trace, so `strip_shutdown_exit_reason()` sounds like it removes the exit reason, whereas it actually keeps it. Maybe it should be called, `strip_shutdown_atom` or `get_reason`

msfstef: makes sense, words do mean something haha

PR #3267, opened 2025-10-09
feat: Release live SSE mode from experimental to available

msfstef:
I've added a patch update for both client and server but this could be a minor as this is adding a new feature.

I'm slightly confused as to how errors are being handled in SSE mode, as there's a reference to a flag that does not exist as well. @balegas not sure how well tested this is especially with the various errors, garbage collections, timeouts that can occur for shapes?

This also contains a fix for our must refetch messages that were always being sent as a list of messages even when `in_sse?` was true outside of the live sse clause, which breaks old clients, so the new server will now not be compatible with old clients :/ @balegas what do you reckon is the best thing to do here - IMO since this was an experimental feature we should be fine with breaking old clients using it before it was marked as released.

## UPDATE

We are no longer deprecating the experimental flag, we are completely removing it, so now this update requires anyone using SSE to upgrade both their server and clients.

## UPDATE 2

We are no longer sending the `must-refetch` message over the SSE event stream, instead letting it come through a regular 409 response that also contains the new shape handle in the headers and processed by the client the same way.

This maintains backwards compatibility, and thus I reinstated the deprecations of the experimental flags and changed back the changeset message to not be a breaking one (although still a minor release).

Electric has a test ensuring the the `experimental_live_sse` flag still works and can thus serve both old and new client versions, and the client always sends both to ensure it works with both old and new server versions.

@KyleAMathews @balegas let me know if you agree with this.

### DIFF
--- a/.changeset/few-bananas-relate.md
+++ b/.changeset/few-bananas-relate.md
@@ -1,6 +1,6 @@
 ---
-"@electric-sql/client": minor
-"@core/sync-service": minor
+'@electric-sql/client': minor
+'@core/sync-service': minor
 ---
 
 Deprecate `experimental_live_sse` and introduce proper `live_sse` flag for sending live updates as server sent events.
--- a/.changeset/loud-moles-grab.md
+++ b/.changeset/loud-moles-grab.md
@@ -1,5 +1,5 @@
 ---
-"@electric-sql/client": patch
+'@electric-sql/client': patch
 ---
 
 Handle 409 must refetch message from both SSE and long poll fetches correctly.


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -12,17 +12,17 @@ defmodule Electric.Plug.ServeShapePlug do
 
   require Logger
 
-  plug :fetch_query_params
+  plug(:fetch_query_params)
```
icehaunter: formatter quirk?

msfstef: whoops

packages/typescript-client/src/client.ts
```diff
@@ -815,13 +821,15 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+    const useSse = this.options.liveSse ?? this.options.experimentalLiveSse
     if (
       this.#isUpToDate &&
-      this.options.experimentalLiveSse &&
+      useSse &&
       !this.#isRefreshing &&
       !opts.resumingFromPause
     ) {
       opts.fetchUrl.searchParams.set(EXPERIMENTAL_LIVE_SSE_QUERY_PARAM, `true`)
+      opts.fetchUrl.searchParams.set(LIVE_SSE_QUERY_PARAM, `true`)
```
kevin-dp, line 838: I guess `EXPERIMENTAL_LIVE_SSE_QUERY_PARAM` is kept in order to keep new client versions compatible with old backends?

msfstef: yes exactly

msfstef: well we just decided to not deprecate and instead remove the experimental flag altogether so this is no longer relevant haha

KyleAMathews: Wait why? We don't do breaking changes now that we're past 1.0.

balegas: The old and new protocol are not compatible because of the changes to `must-refetch`. SSE feature was marked experimental, so it seems too much of a burden to keep compatibility between old and new flags. We're leaving a bit note about upgrading client and server.

It's true that it might break clients if the server is upgraded before clients. If you'd like to keep compatibility, I think the best is to fork the behaviour and deprecate _experimental_.

msfstef: Looking at the code, although we send a `must-refetch` message through SSE we never action it, so we could maintain compatibility by keeping the `must-refetch` as a list of events, the clients still work, and compatibility is maintained

It would require changing the original implementation intent of sending the `must-refetch` message through the SSE stream but it was never really used either way.

balegas: That would be better! thanks

PR #3273, opened 2025-10-13
Document the new ELECTRIC_REPLICATION_IDLE_TIMEOUT config

alco:


website/docs/api/config.md
```diff
@@ -128,6 +128,29 @@ Suffix for the logical replication publication and slot name.
...
+### ELECTRIC_REPLICATION_IDLE_TIMEOUT
+
+<EnvVarConfig
+    name="ELECTRIC_REPLICATION_IDLE_TIMEOUT"
+    defaultValue="0"
+    example="5min">
+
+After seeing no activity on the logical replication stream for this long, Electric will close all of its database connections. This allows the database server to scale-to-zero on supported providers.
+
+While Electric is in the scaled-down mode, an incoming shape request will cause is to reopen database connections and restart the logical replication stream. The request itself will be held until it can be processed as usual to return a proper response.
```
msfstef:
```suggestion
While Electric is in the scaled-down mode, an incoming shape request will cause it to reopen database connections and restart the logical replication stream. The request itself will be held until it can be processed as usual to return a proper response.
```

msfstef:
I suppose we want to merge this after we release a version with idle conns?

alco:
@msfstef yep

PR #3274, opened 2025-10-13
Add two new configuration options for periodic retained WAL size check in scaled down mode

### LINKED ISSUE #3260: Add a configuration option to set the WAL check period in scaled down mode
Just to avoid Electric going to sleep forever, we should have it wake up periodically and make sure there's no pathological WAL buildup over time.

This can be implemented as a timer on the `Electric.Connection.Restarter` process that counts down the time until the next wakeup. The period should be configurable via a new config option named, say, `ELECTRIC_REPLICATION_WAL_SIZE_CHECK_PERIOD`.

Currently, connection manager validates database connection options (separately for the direct connection and for the pooled connection) and stores the validated ones in its state. One downside of this is that revalidation happens every time the Manager restarts. The other downside is that when the Manager itself is down, the Restarter process won't have access to validated connection options.

This part needs to be refactored such that validated connection options are persistent in StackConfig.

We should also build a utility wrapper around Postgrex.SimpleConnection to be used for one-off queries, such as terminating the backend holding onto the lock, validating connection options and fetching the WAL size reserved by the replication slot:

```ELIXIR
#
## LockBreakerConnection
#
"WITH inactive_slots AS (..."
|> Electric.OneOffQuery.async(pool_connection_opts, max_reconnects: 0)

#
## WAL check
#
async_query = 
  {"""
      SELECT
        pg_wal_lsn_diff(pg_current_wal_lsn(), confirmed_flush_lsn)::int8
      FROM
        pg_replication_slots
      WHERE
        slot_name = $1
  """, [slot_name]}
  |> Electric.OneOffQuery.async(pool_connection_opts, max_reconnects: 1, reconnect_backoff: 500)

case Electric.OneOffQuery.await(async_query) do
  {:ok, result} -> ...
  {:error, reason} -> ...
end

#
## ConnectionResolver
#
# We're not interested in running any queries here. Just want to make sure a connection
# can be established using the provided connection opts.
Electric.OneOffQuery.connect(pool_connection_opts, max_reconnects: 1, reconnect_backoff: 500)
|> case do
  {:ok, valited_conn_opts} -> ...
  {:error, reason} -> ...
end
```

alco:
This PR introduces periodic monitoring of retained WAL size when Electric has scaled down its database connections due to inactivity. If during one of these periodic checks the retained WAL is detected to have grown beyond the configured threshold, Electric wakes up the connection subsystem to resume replication stream processing.

## Core changes

`Connection.Restarter`
- When connections scale down, starts a periodic timer to check retained WAL size
- Queries Postgres for the retained WAL size using `pg_wal_lsn_diff(pg_current_wal_lsn(), restart_lsn)`
- If size exceeds threshold, restarts the connection subsystem
- Logs checks with human-readable sizes and intervals

`Connection.Manager`
When it validates connection options, it stores the validated ones in StackConfig. This can then be used by `Connection.Restarter` for one-off DB queries to check the retained WAL size. When `Connection.Manager` restarts (after an error or when the connection subsystem is woken up), it erases these options from StackConfig and repeats the validation process as before. This is to prevent the system from ending up in an invalid state by ensuring that at `Connection.Manager` startup it will always get the same starting connection options to work with.

Added two new configuration options:
- `ELECTRIC_REPLICATION_IDLE_WAL_SIZE_CHECK_PERIOD`: How often to check retained WAL size when scaled down (default: 1 hour)
- `ELECTRIC_REPLICATION_IDLE_WAL_SIZE_THRESHOLD`: WAL size threshold that triggers reconnection (default: 100 MiB)

Both options support human-readable time/size formats.

## Refactoring

Introduced a new module named `OneOffConnection` that wrap `Postgrex.SimpleConnection` and provides a simple API for opening a one-off DB connection, running a query (using the simple protocol) and getting the result back, all synchronously.

Reimplemented the lock breaking logic using `OneOffConnection` and removed `LockBreakerConnection` since it was no longer necessary to have as a separate module.

Refactored `ConnectionResolver`, replacing its ad-hoc wrapping of `Postgrex.SimpleConnection` with `OneOffConnection`.

## Testing

Added a new integration test (`integration-tests/tests/wal-size-check-while-scaled-down.lux`) that verifies Electric's handling of two cases during its periodic WAL size check: 1) WAL size under the threshold; 2) WAL size has exceeded the threshold.

---

Closes https://github.com/electric-sql/electric/issues/3260.


packages/sync-service/lib/electric/connection/restarter.ex
```diff
@@ -104,4 +117,40 @@ defmodule Electric.Connection.Restarter do
...
+       when is_integer(period) and period > 0 do
+    timer = Process.send_after(self(), :check_wal_size, period)
+    %{state | wal_size_check_timer: timer}
+  end
+
+  defp schedule_wal_size_check(state), do: state
+
+  defp query_retained_wal_size(_state) do
+    # FIXME: placeholder
+    0
```
msfstef: I assume this doesn't work until this is fixed

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -339,12 +340,28 @@ defmodule Electric.Connection.Manager do
...
+    opts =
+      Map.get(state.validated_connection_opts, type) ||
+        Electric.StackConfig.get(state.stack_id, config_key)
+
+    if opts do
       {:ok, opts, state}
     else
       try do
         with {:ok, validated_opts} <- ConnectionResolver.validate(state.stack_id, conn_opts) do
+          Electric.StackConfig.put(state.stack_id, config_key, validated_opts)
```
msfstef: I don't love the dynamic configuration on this in the sense that the only reason to have it is for something to use it outside of the connection manager, in which case it doesn't really know when it is ready?

Also if the connection manager crashes, which could be because of a change in the connection configurations, would we not perhaps _want_ to revalidated the connection options passed to us?

msfstef (commented):
Looks good but I assume this does not work since there's no querying of the wal size (and no tests which I assume is for the above reason?)

packages/sync-service/lib/electric/connection/manager.ex
alco: This is a great point. The original implementation of connection option validation was modelled as a uni-directional flow: i.e. once we fall back from SSL to noSSL or from SSL to unencrypted, there was no going back. I've always thought of the revalidation when connection manager restarts as a redundant step. But you're proposing that it is, in fact, an opportunity to redo the validation?

Given that Electric.StackConfig lives under the StackSupervisor, any change in the stack configuration will erase its cached connection options and so there's no problem with having stale validated options.

From my perspective, doing validation once is the goal. We could implement it by running connection resolver for replication and pool connections upfront, before proceeding to initialize connection manager for the first time. Conceptually, this would not be as controversial, I think, but it would be at the cost of increased latency on the first connection setup.

packages/sync-service/lib/electric/connection/restarter.ex
alco: Yeah. I'll add the query and extend the integration test in a follow-up PR to be merged into this one.

packages/sync-service/lib/electric/connection/restarter.ex
```diff
@@ -75,30 +81,42 @@ defmodule Electric.Connection.Restarter do
      %{
        stack_id: Keyword.fetch!(opts, :stack_id),
        stack_events_registry: Keyword.fetch!(opts, :stack_events_registry),
-       wait_until_conn_up_ref: nil
+       slot_name: Keyword.fetch!(opts, :slot_name),
+       # NOTE(alco): These defaults are really only used in testing when these config options
+       # aren't set. The actual real-world defaults are defined in Electric.Config.
```
msfstef, line 86: wouldn't it be better to assert their presence and in any unit tests we explicitly set the parameters? we've used this pattern so far I think, and used `component_setup.ex` for utilities to start these

packages/sync-service/lib/electric/connection/restarter.ex
```diff
@@ -10,12 +10,18 @@ defmodule Electric.Connection.Restarter do
     - publication manager
     - schema reconciler
 
+  Once the connection subsystem is scaled down, Restarter starts a timer to check the retained
+  WAL size periodically. If the size exceeds the configured threshold, Restart will restart the
```
msfstef, line 14:
```suggestion
  WAL size periodically. If the size exceeds the configured threshold, Restarter will restart the
```

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -1275,4 +1295,67 @@ defmodule Electric.Connection.Manager do
...
+        Logger.debug("No stuck backends found")
+
+      {:ok, %Postgrex.Result{columns: ["pg_terminate_backend"]}} ->
+        Logger.notice(
+          "Terminated a stuck backend to free the lock #{lock_name} because slot with same name was inactive"
+        )
+
+      {:error, reason} ->
+        # No retries here since this is a one-shot attempt at fixing a lock
+        Logger.warning("Failed try and break stuck lock connection: #{inspect(reason)}")
```
msfstef, line 1326: nit: perhaps we need to be logging this as an error so they show up on sentry (?)

packages/sync-service/lib/electric/connection/restarter.ex
```diff
@@ -108,11 +126,92 @@ defmodule Electric.Connection.Restarter do
...
   def handle_info({ref, :ok}, %{wait_until_conn_up_ref: ref} = state) do
     {:noreply, %{state | wait_until_conn_up_ref: nil}}
   end
+
+  # The timer has already been cancelled and reset, ignore this message.
+  def handle_info(:check_wal_size, %{wal_size_check_timer: nil} = state) do
+    {:noreply, state}
+  end
+
+  def handle_info(:check_wal_size, state) do
```
msfstef, line 142: I've seen the integration test - would also appreciate a Restarter only integration test as part of our elixir test suite if it's not too difficult to set up!

msfstef (approved):
Love the one off connection utility!

Looks great, would appreciate a couple of happy path elixir tests isolating the restarter

Issue #3245, opened 2025-10-07
Load PublicationManager initial state from ETS


magnetised:
Don't require every consumer to register itself on restoration (currently via the snapshotter). Instead we can get the list of publication changes directly from the shape status ets. Only newly created shapes should call into the publication manager.

PR #3275, opened 2025-10-13
feat: Restore `PublicationManager` state from `ShapeStatus` ETS on startup

magnetised:
## Edited by @msfstef 

Closes https://github.com/electric-sql/electric/issues/3245

1. On bootup, pub man filters state is recovered directly from `ShapeStatus`
2. It immediately goes into a `:continue` clause to update the configuration
3. Keeps track of a `restore_complete?` flag as well as `restore_waiters` from the new `wait_for_restore` sync API
  3a. I decided to add waiters as I want the update to be able to fail due to e.g. network errors and still accept requests and handle them without crashing the shape subsystem, in the spirit of previous work.
4a. If restore succeeds, we reply to the waiters
4b. If restore fails, we go into a refresh loop with a shorter timeout than our regular refresh timeout until restore completes
5. We `wait_for_restore` in the `ShapeCache` before we mark consumers as ready, meaning we block there until we manage to recover the publication.

The last step is still coupling and blocking ShapeCache to PubMan but in the spirit of maintaining the current behaviour but avoiding the onslaught of recovery messages from all shapes I think this is a good first step, and @magnetised feel free to move things around in your work to decouple things.

The main thing to note is that _we should not_ start streaming changes until we know for sure that the publication has actually been restored - but I believe the shape subsystem should be able to handle existing shape data while we wait for that to happen.

I've also made it so that we only call `add_shape` to the publication manager if a snapshot _has not_ been started, so that most restored shapes don't end up calling the pubman (@magnetised was this the intention? if not, I can let you sort it out in your own PRs)

### DIFF
--- a/packages/sync-service/test/electric/replication/publication_manager_db_test.exs
+++ b/packages/sync-service/test/electric/replication/publication_manager_db_test.exs
@@ -34,6 +34,19 @@ defmodule Electric.Replication.PublicationManagerDbTest do
   ]
 
   setup ctx do
+    if ctx[:existing_where_clauses] do
+      for where_clause <- ctx.existing_where_clauses do
+        relation = {"public", "items"}
+        relation_oid = lookup_relation_oid(ctx.pool, relation)
+        shape = generate_shape({relation_oid, relation}, where_clause)
+        {:ok, _shape_handle} = Electric.ShapeCache.ShapeStatus.add_shape(ctx.stack_id, shape)
+      end
+    end
+
+    :ok
+  end
+
+  setup ctx do
     %{publication_manager: {_, pub_mgr_opts}} =
       with_publication_manager(%{
         module: ctx.module,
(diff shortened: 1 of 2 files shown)


packages/sync-service/test/electric/shape_cache_test.exs
```diff
@@ -56,13 +55,22 @@ defmodule Electric.ShapeCacheTest do
...
+
+    def remove_shape(_handle, _opts), do: :ok
+
+    def wait_for_restore(opts) do
+      send(opts[:test_pid], {:called, :wait_for_restore})
+    end
   end
 
-  setup :verify_on_exit!
+  # setup :verify_on_exit!
```
alco: Hm?

msfstef: whoops

PR #3276, opened 2025-10-13
fix: Add stack id to logger metadata for call home telemetry

msfstef:
Sometimes call home telemetry is stack-scoped, but we don't set the logging metadata for it, so log filtering becomes a bit hard for it

### DIFF
--- a/.changeset/clever-boxes-shop.md
+++ b/.changeset/clever-boxes-shop.md
@@ -1,5 +1,5 @@
 ---
-"@core/sync-service": minor
+'@core/sync-service': minor
 ---
 
 Remove old FileStorage implementation


packages/sync-service/lib/electric/telemetry/call_home_reporter.ex
```diff
@@ -24,12 +24,23 @@ with_telemetry Telemetry.Metrics do
...
-      GenServer.start_link(
-        __MODULE__,
-        {metrics, first_report_in, reporting_period, name, static_info, reporter_fn},
-        name: name
-      )
+      stack_id = Keyword.get(opts, :stack_id)
+
+      if stack_id = Keyword.get(opts, :stack_id),
+        do:
+          GenServer.start_link(
```
alco: This looks odd. We were starting CallHomeReporter unconditionally before. Why is it conditioned on the presence of stack_id now?

packages/sync-service/lib/electric/telemetry/call_home_reporter.ex
```diff
@@ -47,12 +58,27 @@ with_telemetry Telemetry.Metrics do
...
+        reporter_fn: reporter_fn,
+        stack_id: stack_id
+      } = opts
+
       # We need to trap exits here so that `terminate/2` callback has more chances to run
       # and send data before crash/shutdown
       Process.flag(:trap_exit, true)
       Process.set_label({:call_home_reporter, name})
 
+      if not is_nil(stack_id) do
```
alco: It's perfectly idiomatic to write `if stack_id do`, and you're already relying on this implicit falsiness of `nil` above, in the `start_link()` function.

packages/sync-service/lib/electric/telemetry/call_home_reporter.ex
alco: I suppose this change's goal is to disable application-wide CallHomeReporter? I don't see why we cannot enforce the presence of `stack_id` and remove the `CallHomeReporter` spec from `ApplicationTelemetry`.

msfstef: No sorry this is wrong accidentally committed

msfstef: I had written it in the start link first and moved it to init, but forgot to remove the full line and the formatter thought this made sense

Issue #2783, opened 2025-05-26
Fire stack events as part of `StatusMonitor` state transitions


msfstef:
While working on https://github.com/electric-sql/electric/pull/2781 to fix an issue where we would sometimes omit to fire a `:ready` stack event message when the replication client alone recovered from a disconnection, @alco and I discussed and noticed that the separation between `StatusMonitor`'s state and the stack events can be confusing.

The suggestion here is to essentially move most of the important stack events to be fired from within `StatusMonitor` so that they are consistent with Electric's actual status and behaviour (health check + responses).

This could be in the form of monitoring internal state transitions and firing stack events for particular ones that we already fire events for, but this requires some investigation.

PR #3277, opened 2025-10-13
Dispatch the "scaled_down_database_connections" stack event from Restarter

alco:


alco:
> I do want to point to #2783 as we might potentially want to move more stack events to the status monitor (including this one) rather than spread them out throughout the code base.

I'm refraining from doing this change here because currently we don't have any stack events dispatched from StatusMonitor.  Before this change, all stack events are dispatched from Connection.Manager. I can adjust my change in one of these two ways to keep the dispatch contained to the connection manager:
- call `Electric.StackSupervisor.dispatch_stack_event(()` from `Connection.Manager.Supervisor.stop_connection_manager()`
- add a custom call in `Connection.Manager` that will dispatch this event and then ask the supervisor to shut itself down like it's currently done in `Connection.Manager.dispatch_fatal_error_and_shutdown()`.

msfstef:
@alco I think we can keep this exactly as is, but mostly mentioned it in that it seems like we might want to look at moving stack events to status monitor in a separate PR

PR #3282, opened 2025-10-14
feat(sync-service): Expire excess shapes

robacourt:


robacourt:
> Nice!
> 
> Does this render the batch expiry size obsolete?

Yes, but I've not removed it yet to not break backwards compatibility. I need to remove it from the cloud UI first

PR #3285, opened 2025-10-14
fix: Fix publication update deadlocks

msfstef:
We observed deadlocks during publication updates in high traffic usage.

The protection against concurrent transactions causing deadlocks was removed here: https://github.com/electric-sql/electric/pull/3218

The idea was that since we linearise everything behind the publication manager, this protection is redundant, but it seems this can still occur (and DDL operations on Postgres have much weaker guarantees).

I've added a slow test (tagged slow to only run on CI) that managed to reproduce the deadlocks, although not always but that's the problem with concurrency bugs, and restoring the logic of updating replica identities only after touching the publication seems to have done away with the error completely.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/configuration.ex
+++ b/packages/sync-service/lib/electric/postgres/configuration.ex
@@ -222,12 +222,32 @@ defmodule Electric.Postgres.Configuration do
           {:ok, :configured} | {:error, term()}
   defp set_table_replica_identity_full(conn, oid_relation) do
     {_oid, relation} = oid_relation
+    {schema, name} = relation
     table = Utils.relation_to_sql(relation)
 
-    Logger.debug("Setting #{table} replica identity to FULL")
+    case Postgrex.query(
+           conn,
+           """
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -168,11 +168,28 @@ defmodule Electric.Postgres.Configuration do
       )
     end
 
+    drop_results =
+      Enum.map(to_drop, &{&1, drop_table_from_publication(conn, publication_name, &1)})
+
+    # `ALTER TABLE` should be after the publication altering, because it takes out an exclusive lock over this table,
+    # but the publication altering takes out a shared lock on all mentioned tables, so a concurrent transaction will
+    # deadlock if the order is reversed.
```
alco: Would be nice to mention here that even though publication manager serialises updates we still see deadlocks in prod.

msfstef: good idea!

PR #3289, opened 2025-10-15
fix: Properly handle reconnections in connection manager

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3288

We see tons of timer errors for clauses not existing in deploys on cloud, see issue above.

I've noticed that for connection pools, we only allow starting them if the pool pids are set to none, which is only ever true on the first try, so the reconnection logic never really worked for them.

I've made it so that we initialise `pool_pids` with admin and snapshot and start the appropriate ones.

I've also made it so that we only attempt reconnections if we fail during one of the setup steps of the process that actually failed, cause the logic seemed to basically retry the current step on _any_ connection error, so the pools might fail and it might retry the replication client.

The assumption there was that things would only ever exit and reconnect during their connection step, but the reality is that they can exit with a connection error _at any point_ during the setup.

I continue to believe that our connection manager is a liability right now because it's far from fully mapped out.

I haven't added tests for this issue which is really annoying, but opening this PR regardless to move work on this issue. Feel free to contribute to it.

edit: I've manually reproduced the error by sporadically killing the pools after they are ready and observing startup - this fix indeed works and if the pools crash after having become ready all of the connection manager is brought down but in a graceful way.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -469,7 +469,7 @@ defmodule Electric.Connection.Manager do
 
             %{
               state
-              | pool_pids: Map.update!(state.pool_pids, pool_role, fn _ -> {pool_pid, false} end)
+              | pool_pids: Map.put(state.pool_pids, pool_role, {pool_pid, false})
             }
           end)
 


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -442,39 +442,38 @@ defmodule Electric.Connection.Manager do
...
+                stack_id: state.stack_id,
+                role: pool_role,
+                connection_manager: self(),
+                pool_opts: Keyword.put(state.pool_opts, :pool_size, pool_size),
+                conn_opts: conn_opts
+              )
+
+            %{
+              state
+              | pool_pids: Map.update!(state.pool_pids, pool_role, fn _ -> {pool_pid, false} end)
```
magnetised: Nit: This is just a Map.put, no need for Map.update

msfstef:
> I'm assuming no tests !== no testing right?

Wrong assumption! It's messy to reproduce this and don't want to add more flaky tests that are going to overlap with your and Rob's work - this is a blind fix on something that looks obviously wrong.

I could potentially manually kill and stall things for the sake of reproducing the error and checking the fix.

msfstef:
@magnetised I've come back with manual testing. I'm killing the pool with random chance after initialising successfully and I can reproduce the error seen on Sentry.

With my fixes, if the process dies outside of its own initialisation procedure, conn man restarts as part of its regular flow without logging spurious errors.

alco (approved):
Nice!

PR #3290, opened 2025-10-15
feat(sync-service): Lazily start shape writers

### LINKED ISSUE #3243: lazily start shape consumers on boot

### LINKED ISSUE #3294: Application/stack startup is aborted when restoring shapes into ShapeStatus due to a non-existing atom error
I have created a number of shapes, then killed the Electric app with Ctrl-C, then restarted:

```elixir
$ iex -S mix
Erlang/OTP 28 [erts-16.0.1] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit:ns]


13:26:23.186 [warning] Electric is running in insecure mode - this risks exposing your database - only use insecure mode in development or if you've otherwise secured the Electric API.
** (Mix) Could not start application electric: Electric.Application.start(:normal, []) returned an error: shutdown: failed to start child: Electric.StackSupervisor
    ** (EXIT) shutdown: failed to start child: Electric.ShapeCache.ShapeStatusOwner
        ** (EXIT) an exception was raised:

  * 1st argument: not an already existing atom

                :erlang.binary_to_existing_atom("uuid")
                (electric 1.1.14) lib/electric/replication/eval/expr.ex:62: anonymous fn/1 in Electric.Replication.Eval.Expr.from_json_safe/1
                (elixir 1.18.4) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
                (elixir 1.18.4) lib/map.ex:262: Map.new_from_enum/2
                (electric 1.1.14) lib/electric/replication/eval/expr.ex:62: Electric.Replication.Eval.Expr.from_json_safe/1
                (electric 1.1.14) lib/electric/shapes/shape.ex:618: Electric.Shapes.Shape.from_json_safe/1
                (electric 1.1.14) lib/electric/shape_cache/pure_file_storage.ex:174: anonymous fn/3 in Electric.ShapeCache.PureFileStorage.get_all_stored_shapes/1
                (elixir 1.18.4) lib/enum.ex:2546: Enum."-reduce/3-lists^foldl/2-0-"/3
                (electric 1.1.14) lib/electric/shape_cache/pure_file_storage.ex:173: Electric.ShapeCache.PureFileStorage.get_all_stored_shapes/1
                (electric 1.1.14) lib/electric/shape_cache/shape_status.ex:374: Electric.ShapeCache.ShapeStatus.load/3
                (electric 1.1.14) lib/electric/shape_cache/shape_status_owner.ex:43: Electric.ShapeCache.ShapeStatusOwner.init/1
                (stdlib 7.0.1) gen_server.erl:2276: :gen_server.init_it/2
                (stdlib 7.0.1) gen_server.erl:2236: :gen_server.init_it/6
                (stdlib 7.0.1) proc_lib.erl:333: :proc_lib.init_p_do_apply/3
```

And this happens at every subsequent attempt to restart the app.

Doesn't reproduce when there's a shape status backup available following a clean shutdown. But it is sufficient to remove the backup to force Electric to recover shapes from shape logs for this issue to get reproduced.

magnetised:
Only launch a writer/consumer process for an existing shape when receiving a write to that shape.

- Fixes exceptions caused when trying to get shape log data from shapes with only snapshots (e.g. https://electricsql-04.sentry.io/issues/53066466/?environment=production&query=is%3Aunresolved%20enoent&referrer=issue-stream)
- Fixes snapshot started state not being restored into the status ets from the disk
- Fixes an occasional issue in dev caused by trying to load uuid columns (see comment)

Fixes #3243
Fixes #3294

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -355,7 +355,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
   def snapshot_started?(stack_ref, shape_handle) do
     case :ets.lookup(shape_meta_table(stack_ref), {@snapshot_started, shape_handle}) do
       [] -> false
-      [{{@snapshot_started, ^shape_handle}, state}] -> state
+      [{{@snapshot_started, ^shape_handle}, started?}] -> started?
     end
   end
 


packages/sync-service/lib/electric/replication/eval/expr.ex
```diff
@@ -71,7 +71,7 @@ defmodule Electric.Replication.Eval.Expr do
   defp type_from_json_safe(["row", types]), do: {:row, Enum.map(types, &type_from_json_safe/1)}
   defp type_from_json_safe(["internal", type]), do: {:internal, type_from_json_safe(type)}
   defp type_from_json_safe(["enum", type]), do: {:enum, type_from_json_safe(type)}
-  defp type_from_json_safe(type) when is_binary(type), do: String.to_existing_atom(type)
+  defp type_from_json_safe(type) when is_binary(type), do: String.to_atom(type)
```
magnetised: Don't know what changed, but running locally was occasionally raising when trying `String.to_existing_atom("uuid")`.

Since we're de-serialising data from postgres, not un-trusted user input this feels safe

alco: https://github.com/electric-sql/electric/issues/3294

I don't think this is needed. As I've written in that issue, this only affects dev/test. I would instead add a noop call to `Electric.Postgres.supported_types` inside ShapeStatus, just before loading shapes from disk, with a comment explaining that we make sure all supported types are loaded by the time we start parsing shape definitions.

If shape definition is trusted data, why not write it using Erlang's external term format instead of JSON?

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -75,8 +76,25 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+  def subscribe(_server_ref, _shape_handle, _shape, :restore) do
+    :ok
+  end
+
+  # new shapes -- created after boot -- do need to be added
+  def subscribe(server_ref, shape_handle, shape, :create) do
+    GenServer.call(server(server_ref), {:subscribe, shape_handle, shape})
+  end
+
+  def unsubscribe(server_ref, shape_handle) do
```
robacourt: This is no longer subscribing, perhaps the functions should be called `add_shape` and `remove_shape`

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -75,8 +76,25 @@ defmodule Electric.Replication.ShapeLogCollector do
     :ok = GenServer.call(server, {:relation_msg, rel, trace_context}, :infinity)
   end
 
-  def subscribe(server_ref, shape_handle, shape, action) when action in [:restore, :create] do
-    GenServer.call(server(server_ref), {:subscribe, shape_handle, shape, action})
+  # shapes that are being restored are already in the filters
+  # because they were restored from the ets at startup
+  def subscribe(_server_ref, _shape_handle, _shape, :restore) do
```
robacourt: Minor Nit. Either ShapeLogCollector needs to know the difference between `:create` and `:restore` or Consumer does. I'd say it's better that Consumer does as it already knows about action. If it's a `:restore` then not to call `add_shape`. That way ShapeLogCollector doesn't even need to know about actions.

packages/sync-service/lib/electric/shape_cache/pure_file_storage/chunk_index.ex
```diff
@@ -455,4 +462,16 @@ defmodule Electric.ShapeCache.PureFileStorage.ChunkIndex do
...
+    File.open!(path, opts)
+  end
+
+  defp open!(path, opts, fun) do
+    File.open!(path, opts, fun)
+  end
+
+  defp open(path, opts, fun) do
+    File.open(path, opts, fun)
+  end
```
robacourt: Is this just to make the code more concise? If so you could `import File, only: [open!: 2, open!: 3, open: 3]`

magnetised:
I need to do a bit more work on the cleanup -- some paths I missed...

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -195,6 +200,14 @@ defmodule Electric.ShapeCache do
     end
   end
 
+  # this is race-condition free because it's only every called from the shape log collector
+  # which processes each transaction in serial
+  # and if it's calling this function then it knows that no consumer exists for the shape
```
robacourt: This is a slightly confusing comment as in the handle_call you mention race conditions

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -260,69 +272,75 @@ defmodule Electric.ShapeCache do
...
+    # basically ignores the {:error, :no_shape} result - excluding the shape handle
+    # from the broadcast
+    case ShapeStatus.get_shape_by_handle(state.stack_id, shape_handle) do
+      {:ok, shape} ->
+        # TODO: otel ctx from shape log collector?
+        {:ok, pid} = start_shape(shape_handle, shape, state, nil, :restore)
+        {:reply, {:ok, pid}, state}
+
+      {:error, _reason} ->
+        {:reply, {:error, :no_shape}, state}
```
robacourt: Surely we should be more explicit with the reason here? Otherwise we're just swallowing any error

packages/sync-service/lib/electric/replication/eval/expr.ex
msfstef: I agree that I'd rather we address the root cause than blindly create atoms from disk files.

Also good point about the external term format - I guess it's convenient for it to be easily human readable, or at least it was when we were initially developing things. Perhaps it might be worth revisiting to simplify and remove unnecessary json serialization/deserialization code.

packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -0,0 +1,198 @@
+defmodule Electric.Shapes.ConsumerRegistry do
```
robacourt, line 1: Does this really need to be a separate process? It could just be a module who's state is held by the ShapeLogCollector.  It can still be tested separately. I don't think there would be any performance difference. I mention it as it would make concurrency would be easier to reason about, rather than having to know that it only gets called synchronously by the ShapeLogCollector, it's explicit. Less processes to worry about in the supervision tree too.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -75,8 +76,29 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+    :ok
+  end
```
robacourt, line 83: Either ShapeLogCollector needs to know the difference between :create and :restore or Consumer does. I'd say it's better that Consumer does as it already knows about action. If it's a :restore then it should not call subscribe. That way ShapeLogCollector doesn't even need to know about actions.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -75,8 +76,29 @@ defmodule Electric.Replication.ShapeLogCollector do
     :ok = GenServer.call(server, {:relation_msg, rel, trace_context}, :infinity)
   end
 
-  def subscribe(server_ref, shape_handle, shape, action) when action in [:restore, :create] do
-    GenServer.call(server(server_ref), {:subscribe, shape_handle, shape, action})
+  # shapes that are being restored are already in the filters
+  # because they were restored from the ets at startup
+  def subscribe(_server_ref, _shape_handle, _shape, :restore) do
```
msfstef, line 81: If this action is now a no-op, should we instead guard against calling it at all in the consumer if the action is `:restore`? So as to simplify the shape log collector interface and remove no-op code and instead move the responsibility to the caller

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -75,8 +76,29 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+    # This has to be async otherwise the system will deadlock -
+    # - a consumer being cleanly shutdown may be waiting for a response from ShapeLogCollector
+    #   while ShapeLogCollector is waiting for an ack from a transaction event, or
+    # - a consumer that has crashed will be waiting in a terminate callback
+    #   for a reply from the unsubscribe while the ShapeLogCollector is again
+    #   waiting for a txn ack.
+    GenServer.cast(server(server_ref), {:remove_shape, shape_handle})
+  end
+
+  def remove_shape_sync(server_ref, shape_handle) do
```
msfstef: I see that we only use this in the cleanup code - do we _need_ the response? what could go wrong with cleaning up with an async cast? If our core runtime clenaup logic relies on the async version, should we just let the cleanup also use that one just for the purpose of having less code?

robacourt (approved):
Very nice! Great it addresses the file enoent issues too. I've left a comment questioning if it really needs an extra process, and I've left a few nits.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -365,62 +396,73 @@ defmodule Electric.Replication.ShapeLogCollector do
...
-      %{subscriptions: {0, _}} ->
+      %{subscriptions: 0} ->
         Logger.debug(fn ->
           "Dropping relation message for #{inspect(rel.schema)}.#{inspect(rel.table)}: no active consumers"
         end)
 
         %{state | tracked_relations: tracker_state}
 
       _ ->
+        # relation changes will also start consumers if they're not running
```
msfstef, line 400: nit: relation changes can only result in consumers cleaning themselves up if they are affected - since things like migrations might affect tons of consumers, might be worth in a followup to clean them up in a cheaper way?

I do appreciate the simplicity of letting them handle relations themselves and if this is not an issue at scale then ignore me

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -365,62 +396,73 @@ defmodule Electric.Replication.ShapeLogCollector do
...
 
-  defp remove_subscription(%{subscriptions: {count, set}} = state, from, shape_handle) do
-    OpenTelemetry.start_interval("unsubscribe_shape.remove_subscription")
+  defp remove_subscription(%{subscriptions: count} = state, shape_handle) do
+    OpenTelemetry.with_span(
+      "shape_log_collector.remove_shape",
+      [shape_handle: shape_handle],
+      state.stack_id,
+      fn ->
+        if Filter.has_shape?(state.filter, shape_handle) do
```
msfstef, line 411: before we were not making this check - if all removals are idempotent do we need this? is this a performance optimisation or was there some sort of race/bug this fixes?

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -365,62 +396,73 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+             | subscriptions: count - 1,
+               filter: filter,
+               partitions: partitions,
+               pids_by_shape_handle: pids_by_shape_handle,
+               dependency_layers: dependency_layers,
+               flush_tracker: flush_tracker
+           }
+           |> log_subscription_status()}
+        else
+          Logger.warning("Received unsubscribe from unknown consumer: #{inspect(shape_handle)}")
```
msfstef:
```suggestion
          Logger.error("Received unsubscribe from unknown consumer: #{inspect(shape_handle)}")
```
We were logging an error for this case before, which I think I've never seen on Sentry, so perhaps before downgrading it we should let it be an error if it's never supposed to happen?

extra nit: perhaps worth adding a span attribute to mark if this removal was unexpected/unknown if we want to keep this as a warning so we can maintain observability of these cases

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -10,6 +10,8 @@ defmodule Electric.ShapeCacheBehaviour do
 
   @callback get_shape(shape_def(), opts :: Access.t()) ::
               {shape_handle(), current_snapshot_offset :: LogOffset.t()} | nil
+  @callback get_shape_by_handle(shape_handle(), opts :: Access.t()) ::
```
msfstef: why is this defined on `ShapeCache` if it's a `ShapeStatus` API? Doing a quick search I don't see it being used anywhere as part of the `ShapeCache` interface so perhaps there's not need for this?

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -106,6 +105,12 @@ defmodule Electric.ShapeCache do
     ShapeStatus.get_existing_shape(table, shape)
   end
 
+  @impl Electric.ShapeCacheBehaviour
+  def get_shape_by_handle(handle, opts) do
```
msfstef: see comment above - is this needed as part of the `ShapeCache` module?

packages/sync-service/lib/electric/shape_cache.ex
msfstef:
```suggestion
  # this is race-condition free because it's only every called from the shape log collector
  # which processes each transaction serially
  # and if it's calling this function then it knows that no consumer exists for the shape
```

+1 on @robacourt 's comment - I'm assuming you mean that this can be a `call` rather than a `cast` because there's no race in that regard, but this comment only makes sense in the context of the work you've done and not as a standalone, and conflicts with the `handle_call` - perhaps you can omit this comment altogether as we only need to point out why we make thins asynchronous rather than synchronous.

packages/sync-service/lib/electric/shape_cache.ex
msfstef: The only possible error `get_shape_by_handle` that can be thrown is the shape not being found - but I agree that I'd rather we make the `get_shape_by_handle` API return a `:not_found` atom so we can be explicit in this case as without inspecting the underlying code it looks like we're swallowing errors

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -260,69 +272,75 @@ defmodule Electric.ShapeCache do
...
+  end
+
   defp shape_handles(state) do
     ShapeStatus.list_shapes(state.stack_id)
   end
 
-  # Timeout is per-shape, not for the entire function
-  defp recover_shapes(state, timeout) do
+  defp recover_shapes(state) do
+    import Electric.Postgres.Lsn, only: [is_larger: 2]
```
msfstef, line 295: we never do inline imports anywhere in the codebase - is this idiomatic? why not import it at the top level with everything else? I assume it would be equivalent at compilation time?

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -378,8 +396,15 @@ defmodule Electric.ShapeCache do
            snapshot_timeout_to_first_data: state.snapshot_timeout_to_first_data,
            action: action
          ) do
-      {:ok, _supervisor_pid} ->
-        :ok
+      {:ok, supervisor_pid} ->
+        # TODO: we will just get the consumer pid when we simplify the consumer process tree
+        consumer_pid =
+          Enum.find_value(Supervisor.which_children(supervisor_pid), fn
```
msfstef, line 401: nit: would be nicer if this was a `ConsumerSupervisor` API instead - but given the TODO feel free to ignore this

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -173,7 +173,8 @@ defmodule Electric.ShapeCache.PureFileStorage do
         |> Enum.reduce(%{}, fn opts, acc ->
           case read_shape_definition(opts) do
             {:ok, shape} ->
-              Map.put(acc, opts.shape_handle, shape)
+              snapshot_started? = snapshot_started?(opts)
```
msfstef, line 180: nice catch, although this does mean twice as many reads and we are still keeping the requirements to repopulate the `ShapeStatus` ETS table very uncoupled to this API - perhaps we need to define a better interface for recovering from storage so we never miss any values/flags that need to be repopulated

packages/sync-service/lib/electric/shape_cache/pure_file_storage/chunk_index.ex
msfstef: +1

packages/sync-service/lib/electric/shape_cache/pure_file_storage/chunk_index.ex
```diff
@@ -269,16 +267,25 @@ defmodule Electric.ShapeCache.PureFileStorage.ChunkIndex do
 
         nil ->
           if file_complete?,
-            do: :error,
+            do: :not_found,
```
msfstef, line 272:
```suggestion
            do: :incomplete_chunk_file,
```
nit: was slightly confusing to see a `:not_found` in a file that was found, might be easier to understand the logic with more explicit names

also, is an incomplete chunk file a recoverable error? should we be propagating an error reason so the caller can decide if a cleanup is needed?

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -241,6 +243,19 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
 
+  @impl true
+  def get_shape_by_handle(stack_ref, shape_handle) do
+    case :ets.lookup_element(
+           shape_meta_table(stack_ref),
+           {@shape_meta_data, shape_handle},
+           @shape_meta_shape_pos,
+           nil
+         ) do
+      nil -> {:error, "no shape found for handle #{shape_handle}"}
```
msfstef: nit: I would prefer a `:not_found` or `{:error, :not_found}` or even `nil` return so we can more explicitly handle it, see comment above in `ShapeCache` by both me and @robacourt  - the message is unnecessary since the caller already knows which shape handle they are asking for

We are slightly inconsistent in this module, `get_existing_shape` for example returns `nil` while `remove_shape` returns something like this. I would prefer an explicit atom indicating that it's not there for all of them.

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -323,7 +338,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
   def snapshot_started?(stack_ref, shape_handle) do
     case :ets.lookup(shape_meta_table(stack_ref), {@snapshot_started, shape_handle}) do
       [] -> false
-      [{{@snapshot_started, ^shape_handle}, true}] -> true
+      [{{@snapshot_started, ^shape_handle}, state}] -> state
```
msfstef:
```suggestion
      [{{@snapshot_started, ^shape_handle}, started?}] -> started?
```
supernit - we often use `state` in genservers etc and it confuses me when I see it

packages/sync-service/lib/electric/shapes/filter.ex
```diff
@@ -41,6 +51,8 @@ defmodule Electric.Shapes.Filter do
   """
   @spec add_shape(Filter.t(), shape_id(), Shape.t()) :: Filter.t()
   def add_shape(%Filter{} = filter, shape_id, shape) do
+    if is_map_key(filter.shapes, shape_id), do: raise("duplicate shape #{shape_id}")
```
msfstef:
```suggestion
    if has_shape?(filter, shape_id), do: raise("duplicate shape #{shape_id}")
```
might as well use the new API!

packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -0,0 +1,198 @@
...
+      :ets.lookup_element(table, handle, 2)
+    rescue
+      _e in ArgumentError -> nil
+    end
+  end
+
+  defp start_consumer!(handle, %__MODULE__{} = state) do
+    %__MODULE__{stack_id: stack_id, start_consumer_fun: start_consumer_fun} = state
+
+    case start_consumer_fun.(handle, stack_id: stack_id) do
```
robacourt: It would be quicker to follow if we didn't have this level of indirection here. Maybe just use Repatch instead of an injected fun.

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
```diff
@@ -233,7 +234,10 @@ defmodule Electric.Shapes.Monitor.RefCounter do
   end
 
   def handle_call({:handle_writer_termination, shape_handle, consumer_pid}, _from, state) do
-    # only monitor if the consumer hasn't already registered via notify_reader_termination
+    ShapeLogCollector.remove_shape(state.stack_id, shape_handle)
```
msfstef, line 237: we're already calling this in `terminate_safely` so wouldn't this be called twice and trigger the warning in the `remove_shape` in the shape log collector I've commented on above?

I think we either make this entirely idempotent and fine to call multiple times, or perhaps we need to only call this here if the consumer wasn't registered via `notify_reader_termination`?

packages/sync-service/lib/electric/shapes/monitor/cleanup_task_supervisor.ex
```diff
@@ -87,12 +93,30 @@ defmodule Electric.Shapes.Monitor.CleanupTaskSupervisor do
...
   end
 
+  defp cleanup_shape_log_collector(stack_id, shape_handle) do
+    perform_reporting_errors(
+      fn ->
+        case Electric.Replication.ShapeLogCollector.remove_shape_sync(stack_id, shape_handle) do
+          :ok ->
+            Logger.debug("Removed shape #{shape_handle} from ShapeLogCollector")
+
+          {:error, _reason} ->
```
msfstef: I don't love that we assume that any error from this API means this was just a duplicate call - either we return a specific atom specifying that that was the case, or we convert this to an async call here as well as mentioned above and perhaps we just handle removals idempotently without logging warnings?

packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -0,0 +1,198 @@
...
+    {:ok, n}
+  end
+
+  @spec publish([shape_handle()], term(), registry_state()) :: :ok
+  def publish(shape_handles, event, registry_state) do
+    %{table: table} = registry_state
+
+    shape_handles
+    |> Enum.flat_map(fn handle ->
+      (consumer_pid(handle, table) || start_consumer!(handle, registry_state)) |> List.wrap()
```
msfstef, line 49: so my understanding is that consumers will be started serially - if a transaction ends up waking up lots of consumers at once (like a migration), we would encounter a bottleneck here starting them up one by one no? Should we not be starting them asynchronously like we did when recovering them in `ShapeCache`?

packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -0,0 +1,198 @@
...
+      {:error, reason} ->
+        raise RuntimeError,
+          message:
+            "Stack #{stack_id} unable to start consumer process for shape handle #{handle}: #{inspect(reason)}"
+    end
+  end
+
+  @doc false
+  def registry_table(stack_id) do
+    table = :ets.new(ets_name(stack_id), [:public, :named_table, write_concurrency: :auto])
```
msfstef: We would only need write concurrency if we were starting consumers asynchronously in their own tasks - otherwise registering all happens in this process and removing happens in the shape log collector.

There's also the argument @robacourt made about this potentially not being a process and just being a module that can be used by the shape log collector

packages/sync-service/lib/electric/shapes/consumer_registry.ex
msfstef: +1

packages/sync-service/lib/electric/replication/shape_log_collector.ex
msfstef: +1

packages/sync-service/lib/electric/shape_cache/pure_file_storage/chunk_index.ex
magnetised: I used these in order to more easily trace file creation when debugging the :enoent stuff. Left them in in case they were useful in the future. Can remove if this causes confusion

packages/sync-service/lib/electric/shape_cache.ex
magnetised: Good point! The later me found a race condition that the earlier me hadn't seen.

packages/sync-service/lib/electric/shape_cache.ex
magnetised: The only non-raising error possible from this call is not-found. I'll remove the `reason` completely to match `Access.fetch` and rename to `fetch_shape_by_handle`

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: I see what you're saying -- I just saw it from a different angle. The consumer doesn't need to know if it's being created or restored but to the shapelogcollector it makes a difference so only the shape log collector should care about the action.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: Rob mentioned the same thing -- to me it's the responsibility of the shape log collector to know about this difference. The consumer shouldn't know the details of how the S.L.C maintains its state.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: yeah. i flip-flopped on this. i had to use async on the other cleanup calls because of very real deadlocks but didn't want to. this was the one place I could be sure the remove call wouldn't deadlock so got lured into making it sync.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: Definitely worth a followup I just left it like this for ease. https://github.com/electric-sql/electric/issues/3303

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: We were using the subscription map to test if a shape was registered but then just doing the removal anyway. Seemed a simple change to just not do anything if there was nothing to do.

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
magnetised: This call is only made for exit reasons that aren't either a supervisor shutdown or a `{:shutdown, :cleanup}` (see the guards on `handle_writer_termination/4`) so it wouldn't be called twice in that case.

packages/sync-service/lib/electric/shapes/consumer_registry.ex
magnetised: Since generally transactions only touch a single consumer, which you can see in honeycomb, I've just kept this simple. We can always refactor later.

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
msfstef: ah I got confused by the same thing twice then whoops - ignore me

packages/sync-service/lib/electric/shapes/consumer_registry.ex
msfstef: this is true during regular operations - we have observed migrations causing a bit of a mess since they touch everything, but perhaps we can think about that case separately

packages/sync-service/lib/electric/shape_cache/pure_file_storage/chunk_index.ex
robacourt: +1

packages/sync-service/lib/electric/shape_cache/pure_file_storage/chunk_index.ex
robacourt: yeah, i'd get rid of them

packages/sync-service/lib/electric/shape_cache/shape_status.ex
robacourt: +1

packages/sync-service/lib/electric/shape_cache.ex
alco: I guess this is imported here because `is_larger` is too generic at the module level? But it has to be imported to be used as a guard below.

packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -60,7 +60,7 @@ defmodule Electric.Shapes.ConsumerRegistry do
     :ets.delete(table, shape_handle)
 
     table
-    |> :ets.update_counter(@count_key, -1)
+    |> :ets.update_counter(@count_key, {2, -1, 0, 0})
```
alco: How is this possible? The call to `remove_consumer()` is gated behind a filter check in ShapeLogConsumer. We shouldn't see multiple attempts to remove the same consumer.

In any case, I don't see why we need to manually keep track of the numbeer of records in the ETS table. In `active_consumer_count()` we should just query the table size like we do in `ShapeStatus`:
```
    :ets.info(shape_last_used_table(stack_ref), :size)
```

packages/sync-service/lib/electric/shape_cache.ex
msfstef: Yeah that makes sense - I read in the Elixir docs that this is the intended use so ignore my comment on whether it is idiomatic, I just had not seen it before in our codebase - will resolve!

packages/sync-service/lib/electric/shapes/consumer_registry.ex
magnetised: > How is this possible? The call to remove_consumer() is gated behind a filter check in ShapeLogConsumer

we have various removal paths, some aren't gated behind the shape log collector so I saw locally the count going below 0

Yeah, we could just drop it and use that info call. You're right

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -75,14 +76,35 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+  def subscribe(_server_ref, _shape_handle, _shape, :restore) do
+    :ok
+  end
+
+  # new shapes -- created after boot -- do need to be added
+  def subscribe(server_ref, shape_handle, shape, :create) do
+    GenServer.call(server(server_ref), {:subscribe, shape_handle, shape})
+  end
+
+  def remove_shape(server_ref, shape_handle) do
```
icehaunter, line 90: question: given this is called from the `ref_counter`, and on a system shutdown functionally causes a deluge of messages to this process - would it be prudent to pass in a reason and ignore `:normal | :shutdown | {:shutdown, reason} when reason != :cleanup` reasons as they're expected to be followed by log collector dying? I think this might conflict with ongoing transaction processing, but still might be worth thinking about?

magnetised: This is only called from the ref counter in the case of a shutdown-with-shape clean -- the guards you're looking for are on the `RefCounter.handle_writer_termination/4` function:

```elixir
 def handle_writer_termination(_stack_id, _shape_handle, reason, _pid)
     when is_consumer_shutdown_with_data_retention?(reason),
     do: :ok

 # this is the clean shutdown route so the writer has already registered itself
 def handle_writer_termination(_stack_id, _shape_handle, {:shutdown, :cleanup}, _pid), do: :ok

 def handle_writer_termination(stack_id, shape_handle, _reason, pid) do
   GenServer.call(name(stack_id), {:handle_writer_termination, shape_handle, pid})
 end

```

basically its a no-op in the case of a supervisor-invoked `:shutdown`

packages/sync-service/lib/electric/shape_cache.ex
magnetised: My current PR removes this

PR #3291, opened 2025-10-15
chore: Make periodic measurements customizable

alco:
Disable ApplicationTelemetry.process_memory() periodic measurement by default.

This is a costly probe that we've seen take 1.3 seconds to complete. At 5-second periodicity, it runs 26% of the time. There is an additional cost of ensuring the return value of Process.list() is a consistent snapshot of the system processes, we do not really need it.

Top processes by reductions or memory can be observed under human supervision by running `etop` ([<cloud PR>](https://github.com/electric-sql/stratovolt/pull/899)), as in

```elixir
:etop.start(lines: 40, sort: :memory, interval: 2)
```

This PR also introduces the possibility of passing custom periodic measurements to be run at the same cadence as the builtin ones. This is done to make it easier to add probes in production without havgint to update the Electric dependency version every time.

### DIFF
--- a/.changeset/seven-hairs-applaud.md
+++ b/.changeset/seven-hairs-applaud.md
@@ -1,5 +1,0 @@
----
-'@core/sync-service': patch
----
-
-Disable the top-processes-by-memory measurement by default.
--- a/packages/sync-service/lib/electric/telemetry/application_telemetry.ex
+++ b/packages/sync-service/lib/electric/telemetry/application_telemetry.ex
@@ -247,6 +247,7 @@ with_telemetry [Telemetry.Metrics, OtelMetricExporter] do
         [
           {__MODULE__, :uptime_event, [opts]},
           {__MODULE__, :cpu_utilization, [opts]},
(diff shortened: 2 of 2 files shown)


msfstef:
Would we be able to still maintain a per-process probe without manual intervention? It is useful to have some stats on this over time and over various releases - if it's costly perhaps it could/should run in a much sparser frequency? Or if there's a way to not pay the cost for the consistent snapshot perhaps there's a cheaper alternative?

alco:
@msfstef sorry for making it unnecessarily confusing. I have reverted the change that disabled memory collection as it didn't have a deciding impact on performance. We can make it even less noticeable by using process iteration in the same way as `:recon.proc_count` does.

I've opened [an issue](https://github.com/electric-sql/electric/issues/3309) for that.

msfstef (approved):
@alco thanks for clarifying as I couldn't match the PR desc to the code haha - this looks good!

alco:
Incomplete and no longer relevant. See https://github.com/electric-sql/electric/pull/3412.

PR #3293, opened 2025-10-16
chore(sync-service): Recover shapes without holding supervisor up

robacourt:
This is a small part of the shape subsystem work. This PR recovers shapes without holding it's supervisor up. Eventually this will mean we can recover shapes asynchronously while the database connection is being established.

The effect of this currently is that the ExpiryManager will start before the shapes have been restores so this PR also makes the ExpiryManager check if the stack is ready before expiring.

I've also used this opportunity to re-write the ExpiryManager tests that were a mess (I wrote them)

The other process this affects is the SchemaReconciler, it may now reconcile before the shapes have been restored but given that it loads its own data from PersistentKV I don't think this is an issue (but please let me know if you know otherwise!)

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -84,21 +84,17 @@ defmodule Electric.ShapeCache.ShapeStatus do
   @snapshot_started :snapshot_started
 
   @impl true
-  def initialise(stack_ref, storage) do
+  def initialise(stack_ref, storage \\ nil) do
     last_used_table = shape_last_used_table(stack_ref)
     :ets.new(last_used_table, [:named_table, :public, :ordered_set])
 
     meta_table = shape_meta_table(stack_ref)
 
-    case load_table_backup(meta_table, storage) do
-      {:ok, ^meta_table, path} ->
-        Logger.info("Loaded shape status from backup at #{path}")
-        :ok
-
-      _ ->
-        Logger.debug("No shape status backup loaded, creating new table #{meta_table}")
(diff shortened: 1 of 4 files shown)


robacourt:
> Nice. Some sweet and juicy merge conflicts for me in there...

Yeah, sorry! I've been up to my eyeballs in merge conflicts recently, seems I'm changing what everyone else is changing! My strategy now is to make my PRs as small as possible to minimise and simplify conflicts. Hope it's ok!

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -84,21 +84,17 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
-        :ok
-
-      _ ->
-        Logger.debug("No shape status backup loaded, creating new table #{meta_table}")
-        :ets.new(meta_table, [:named_table, :public, :ordered_set])
-        load(meta_table, last_used_table, storage)
+    if storage do
+      load_from_storage(stack_ref, storage)
+    else
+      :ets.new(meta_table, [:named_table, :public, :ordered_set])
```
alco, line 96: This table creation is now duplicated between this function and `load_from_storage()`.

Where did the need to pass `nil` storage come from? Is it done just for tests? I would argue for either one of two things to be done:
1. restore the logic where we try loading table backup first and then fallback to creating a new ETS table, with all code located in this `initialise()` function. `nil` storage can be handled inside `load_table_backup()`.
2. create a separate clause of `initialise` just for the test env:

```
if Mix.env() == :test do
  def initialise(stack_id) when is_binary(stack_id) do
    create_last_used_table(stack_id)
    create_meta_table(stack_id)
    :ok
  end
end
```

PR #3295, opened 2025-10-16
Hibernate processes that produce garbage during initialisation just once at the end of init()

alco:
Hibernating once at the end of initialisation feels less dirty than explicit calls to `:erlang.garbage_collect()`.

### DIFF
--- a/.changeset/smooth-carrots-press.md
+++ b/.changeset/smooth-carrots-press.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Hibernate some processes after init() that tend to accumulate garbage during initialisation. Because the process heap doesn't keep growing at the same high rate post-initialisation, GC never runs for it and the garbage never gets collected.


msfstef:
I don't quite understand, even without much activity, wouldn't the VM collect that memory if it was under memory pressure? It still feels arbitrary to collect or hibernate stuff on startup and I'm just wondering if this is actually improving performance or whether it's optimising numbers that the VM should be able to take care of in case of need?

alco:
@msfstef The VM does not do GC periodically. And it doesn't try to GC or kill processes when running low on memory: contrary to the OS, BEAM treats all of its processes as essential for the OTP applications running in it.

Each process has its own heap with is garbage-collected independently. Process stack and process heap share the same memory area and are growing towards each other. When they meet, BEAM runs garbage collection and afterwards assesses if there's sufficient free space left in the heap. If not, it grows the heap.

If a lot of garbage is generated during process initialization, the heap may grow a few times (BEAM uses the Fibonacci sequence as the growth factor initially). But if the process requires significantly less dynamic memory afterwards, its heap will keep its size and won't be garbage-collected until the heap and the stack pointer meet again.

This memory remains allocated for the process and is not reusable by the VM.

I agree with you that it's not desirable to have random process hibernation "sprinkled" over the code base. But it seems to be more efficient than doing explicit GC because the heap gets compacted during hibernation. I can try to come up with some way to estimate the amount of garbage a process is holding on to (by comparing the size of its state vs the total heap size in use) and only then doing GC/hibernation. But I'm not sure it's going to yield useful results.

msfstef:
@alco I see that makes sense - I wonder if perhaps then since the large amount of garbage seems to collect on processes doing shape restoration, perhaps we should run it in a Task and just collect the results? Would that eliminate the need for garbage collection by avoiding the buildup of garbage in the long lived process?

msfstef (approved):
Forgot to approve - this is better than/equivalent to the arbitrary garbage collect call regardless!

alco:
> @alco I see that makes sense - I wonder if perhaps then since the large amount of garbage seems to collect on processes doing shape restoration, perhaps we should run it in a Task and just collect the results? Would that eliminate the need for garbage collection by avoiding the buildup of garbage in the long lived process?

Yeah, good point. We should definitely strive to encapsulate garbage-generating code in a process. I was lazy and wanted to have a quick win but after observing the production since and after looking at Gary's work on starting consumer process lazily, I'm abandoning this initiative for now.

I've opened a [new PR](https://github.com/electric-sql/electric/pull/3326) that removes explicit GC calls. We should do more extensive simulation testing and only add GC or hibernation if we succeed to reliably reproduce these memory waste conditions.

Issue #3220, opened 2025-10-02
Support PostgreSQL 18 generated columns with publish_generated_columns=stored


evanheckert:
Electric currently rejects shapes containing generated columns, even when the publication is properly configured with publish_generated_columns=stored on PostgreSQL 18.

## Current Behavior

When trying to sync a table with generated columns, Electric returns an error:

> An error occurred while syncing collection: manufacturer-container-cost-version, 
> it has been marked as ready to avoid blocking apps waiting for '.preload()' to finish. 
> You can provide an 'onError' handler on the shapeOptions to handle this error, and this message will not be logged. FetchError: HTTP Error 400 at `http://localhost:5133/v1/shape?offset=-1&table=materials_manufacturercontainercostversion`: `{"message":"Invalid request","errors":{"columns":["The following columns are generated and cannot be included in the shape: cost_per_estimate_unit. You can exclude them from the shape by explicitly listing which columns to fetch in the 'columns' query param"]}}`

## Requested Behavior

Electric should support syncing generated columns when:
- PostgreSQL version ≥ 18.0
- Publication has `publish_generated_columns = stored` enabled
- Generated columns are `STORED` (not `VIRTUAL`)

PostgreSQL 18 added support for replicating stored generated columns via logical replication. Electric should be able to consume these columns from the replication stream.

### Additional Context

PostgreSQL 18 documentation on generated column replication: https://www.postgresql.org/docs/18/logical-replication-gencols.html

Kyle [mentioned](https://discord.com/channels/933657521581858818/1422991098581614744/1423350572794380298) it's currently a static validation check in Electric.

thruflo:
Thanks for flagging this up 🙏

thruflo:
Related https://github.com/electric-sql/electric/pull/3206

balegas:
Electric being too conservative with that check. Now that Postgres supports generated_columns in PG18, we should check if we can set `publish_generated_columns=stored` for the publication before issuing the error.

Implementation note: we should not try to set this configuration if manual publication is enabled and fallback using the same code path. cc @msfstef

PR #3297, opened 2025-10-16
feat: Support generated columns in Postgres 18

msfstef:
Closes https://github.com/electric-sql/electric/issues/3220

## Overview of changes

I've made the following core changes to support this:

- When creating a new publication, if on PG18 and above, always set `publish_generated_columns = 'stored'`, see [PG docs](https://www.postgresql.org/docs/current/sql-createpublication.html#SQL-CREATEPUBLICATION-PARAMS-WITH-PUBLISH-GENERATED-COLUMNS) for details
  - If we find an existing publication, whether owned or not, we do not alter it.
  - For people upgrading Electric on a PG18 db, they will have to manually set the publication parameter or somehow reset the electric slot/publication/timeline to get this feature. I could add a step to always enable it if we can, but ideally in a separate PR.
- In the `PublicationManager`, before each update, we check the publication status which now includes whether generated columns are replicated
  - We cache the result, and we initialise with an optimistic assumption that generated columns are supported
- In PubMan, we now keep track of which shape handles use generated columns, and in case the configuration changes we do an O(N) scan over our tracked shapes and schedule removals for any shape that used generated columns
  - Slow scan is fine, it's just a list traversal that happens _only_ if the publication configuration is disabled at runtime, which might be the case for manually configured publications but never for our own owned publication.
  - Any shapes pending creation that require generated columns would also receive an error reply
  - Once the configuration result is cached, shape creations immediately get an error answer from the publication manager until a check confirms that generated columns are supported
- The Inspector now loads `supported_features` from Postgres, so at the validation stage we can determine if we should block generated column shapes the same way we have until now, or whether we should let them through and let the PublicationManager determine if they can be created.
  - We will now get different errors if you are on e.g. PG17 and request a shape with generated columns, and if you are on PG18 and above. In the latter you now receive:
  ```
  Publication "{pub_name}" does not publish generated columns. This is a feature introduced in PostgreSQL 18 and requires setting the publication parameter 'publish_generated_columns' to 'stored'. Alternatively, you can exclude them from the shape by explicitly listing which columns to fetch in the 'columns' query param.
  ```
  - Refactored the Inspector ETS table name to `inspector_table` rather than `relation_table` for more generality as it now holds more Postgres inspection information - I think it makes sense to reuse the Inspector for this to keep the infrastructure decoupled, this is purely information about what Postgres supports as part of a validation step.
- Parse PublicationManager error, which gets turned into a `SnapshotError`, at the API level to return 400 in case the feature is supported in the database but not configured.
- Some minor typespec fixes and cleanups of unnecessary info (e.g. `can_alter_publication?` flag in ConnMan - the pubman will determine it by itself at runtime)

## Potential alternatives
- Rather than return errors from the publication manager, we can just clean up shapes when the configuration changes and set the flag appropriately somewhere so that we continue to reject generated columns in the same way at the shape creation validation stage
  - PRO: simplify handling of cases by introducing a sort of "supported features" table for the validation to lookup
  - PRO: by separating into a separate "supported features" table we could offload it from the inspector (although it would still need to check postgres somehow, so I don't like this)
  - CON: I think it's important to separate when a feature is not available and when it is not enabled by choice - by letting the publication manager reply, we are notifying developers that they _could_ use generated columns if they wanted to if they are using PG18 and above.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/configuration.ex
+++ b/packages/sync-service/lib/electric/postgres/configuration.ex
@@ -39,6 +39,7 @@ defmodule Electric.Postgres.Configuration do
 
   @doc """
   Check whether the publication with the given name exists, and return its status.
+
   The status includes whether the publication is owned, whether it publishes all operations,
   and whether it publishes generated columns (if supported by the Postgres version).
   """
--- a/packages/sync-service/lib/electric/postgres/replication_client/connection_setup.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client/connection_setup.ex
@@ -75,7 +75,6 @@ defmodule Electric.Postgres.ReplicationClient.ConnectionSetup do
     # elsewhere. See `Electric.Replication.PublicationManager`.
     query = "CREATE PUBLICATION #{Utils.quote_name(state.publication_name)}"
 
-    # if PG version is 18 or higher, we want to include generated columns in the publication
     query =
       if state.pg_version >= @pg18_version,
         do: query <> " WITH (publish_generated_columns = stored)",
--- a/packages/sync-service/lib/electric/replication/publication_manager.ex
+++ b/packages/sync-service/lib/electric/replication/publication_manager.ex
@@ -329,7 +329,7 @@ defmodule Electric.Replication.PublicationManager do
         can_alter_publication?: can_alter_publication?,
         publishes_generated_columns?: publishes_generated_columns?
       } ->
-        # if the publication has switched form being able to publish generated columns
+        # if the publication has switched from being able to publish generated columns
         # to not being able to publish them, we need to fail any shapes that depend on
         # that feature
         state =


packages/sync-service/lib/electric/postgres/configuration.ex
```diff
@@ -32,41 +38,43 @@ defmodule Electric.Postgres.Configuration do
   @typep publication_relation :: {Electric.relation_id(), Electric.relation(), <<_::8>>}
 
   @doc """
-  Check whether the publication with the given name exists, is configured to
-  publish all required operations, and is owned by the current user.
+  Check whether the publication with the given name exists, and return its status.
```
alco, line 41:
```suggestion
  Check whether the publication with the given name exists, and return its status.
  
```

packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
```diff
@@ -230,6 +230,17 @@ defmodule Electric.Postgres.Inspector.DirectInspector do
   defp parse_type_kind("r"), do: :range
   defp parse_type_kind("m"), do: :multirange
 
+  @impl Electric.Postgres.Inspector
+  def load_supported_features(conn) do
+    with {:ok, %{rows: [[pg_version]]}} <-
+           Postgrex.query(conn, "SELECT current_setting('server_version_num')::int", []) do
+      {:ok,
+       %{
+         supports_generated_column_replication: pg_version >= 180_000
```
alco, line 239: I'm slightly worried that this might not work as expected with Postgres-compatible providers such as Aurora. Ideally we would be able to detect the exact feature needed and not rely on the assumption that a certain version includes it. But that's all out of scope for this PR, just thinking out loud.

packages/sync-service/lib/electric/postgres/replication_client/connection_setup.ex
```diff
@@ -61,16 +61,26 @@ defmodule Electric.Postgres.ReplicationClient.ConnectionSetup do
...
   ###
 
+  @pg18_version 180_000
   defp create_publication_query(state) do
     Logger.debug("ReplicationClient step: create_publication_query")
     # We're creating an "empty" publication here because synced tables are added to it
     # elsewhere. See `Electric.Replication.PublicationManager`.
     query = "CREATE PUBLICATION #{Utils.quote_name(state.publication_name)}"
+
+    # if PG version is 18 or higher, we want to include generated columns in the publication
```
alco: Totally a nitpick but this comment doesn't add anything that's not already there in the code below it.

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -288,23 +316,41 @@ defmodule Electric.Replication.PublicationManager do
...
+        # configure them automatically in case it is owned
+        {:error,
+         Electric.DbConfigurationError.publication_missing_operations(state.publication_name),
+         state}
+
+      %{
+        can_alter_publication?: can_alter_publication?,
+        publishes_generated_columns?: publishes_generated_columns?
+      } ->
+        # if the publication has switched form being able to publish generated columns
```
alco:
```suggestion
        # if the publication has switched from being able to publish generated columns
```

packages/sync-service/lib/electric/postgres/inspector/direct_inspector.ex
msfstef: I agree, that's the reason I've made the API from the inspector return a map of supported features - we can manage the internal implementation of how we check those features ourselves.

One way would be to inspect the schema of the `pg_publication` for the generated col attribute

PR #3301, opened 2025-10-20
Upgrade Elixir to v1.19.1 / OTP 28.1.1

alco:


integration-tests/tests/invalidated-replication-slot.lux
```diff
@@ -4,13 +4,6 @@
 
 [global pg_container_name=replication-slot-invalidated-while-running__pg]
 
-[my invalidated_slot_error=
-  """
-  ** (Postgrex.Error) ERROR 55000 (object_not_in_prerequisite_state) cannot read from logical replication slot "electric_slot_integration"
```
magnetised: Is postgrex not raising as much in 1.19?

magnetised (approved):
Fab. Looking forward to 1.19...

integration-tests/tests/invalidated-replication-slot.lux
alco: It does. The problem was in the `handle_otp_reports` setting. See https://github.com/electric-sql/electric/pull/3301/commits/1995580282ea583d62a6b55bdb3bc1568a03a412. It's going to become a point of controversy in this PR probably. I now need to check if setting it to `true` will give us more headache in Sentry.

alco: To be clear, I have removed the commit that removed those matches from integration tests and instead have set `handle_otp_reports` to `true`.

msfstef (approved):
Nice!

packages/sync-service/config/runtime.exs
```diff
@@ -30,7 +30,7 @@ config :logger, :default_formatter,
 sasl? = env!("ELECTRIC_LOG_OTP_REPORTS", :boolean, false)
 
 config :logger,
-  handle_otp_reports: false,
+  handle_otp_reports: true,
```
msfstef, line 33: would definitely want to know more on the implications of this and how this ties in to our reporting on Sentry

alco: I think we've been living with this option enabled, since before Elixir v1.19 setting it to any value didn't affect anything. The documentation for Logger mentions that it defaults to true.

I had to set it to `true` after upgrading Elixir to v1.19 because we are relying on logs from the `application_controller` process about an application stopping or starting. After the upgrade those logs were no longer emitted.

So in summary, changing the value of this setting seemingly doesn't change the logging behaviour for us, but rather keeps it the same as before.

msfstef: awesome - perhaps we can revisit what setting this to false would mean separately

PR #3302, opened 2025-10-20
fix(tests): unlink component setups to avoid teardown races

icehaunter:
When running `mix test test/electric/shape_cache_test.exs` in particular, there were a lot of error-level logs on consumer failing to register itself with a shape status. It was happening after the test passed. I found that this happens because components we're starting in `setup` part of the test were using `start_link_supervised` instead of `start_supervised`, which caused them to be shut down all together when the test completed, instead of being shut down in correct supervision order as they would in a real system. This PR replaces all `start_link_supervised` in the component setup.


msfstef (approved):
Ah nice catch - I was getting frustrated with this as I was expecting them to tear down in order

Issue #3306, opened 2025-10-20
Make polling interval configurable per metric type


alco:
Currently all polling-based metrics are driven by a global `telemetry_poller` process running under `ApplicationTelemetry` and stack-specific ones running under each live `StackSupervisor`. Therefore they are all sampled at the same interval.

We want to be able to configure polling interval for specific metrics based on the need and cost of each one.

Issue #3305, opened 2025-10-20
Detect ETS memory growth or leaks


alco:
Extracted from https://github.com/electric-sql/electric/issues/3237.

The idea is to sample the byte size and number of records in each of the stack's ETS tables to detect runaway growth over time. This needs to be done inside `StackTelemetry`.

The few tables defined globally can also be sampled inside `ApplicationTelemetry`.

Issue #3304, opened 2025-10-20
Identify processes with large message queues


alco:

alco:
We were already tracking those using `:erlang.system_monitor`. I've made some adjustments to the existing implementation in https://github.com/electric-sql/electric/pull/3308.

Issue #3237, opened 2025-10-06
Improve BEAM Runtime Observability for Electric


balegas:
> I've done a research on observability practices for large BEAM production systems and it told me a lot of interesting stories about production issues. I think it's worth a read: https://gist.github.com/balegas/ca6047972f5cecfa74009404643895a8

# Problem

Electric’s sync engine relies on the BEAM virtual machine to manage large numbers of lightweight processes efficiently.
While we already export application-level metrics and distributed traces to Honeycomb via OpenTelemetry, we currently lack direct visibility into BEAM runtime behavior. This makes it difficult to determine whether performance issues such as latency spikes, throughput drops, or increased resource usage originate from application logic or from underlying VM saturation.

In particular, we have limited insight into:
- Scheduler and run queue contention.
- Memory distribution across processes, ETS tables, and binaries.
- Runaway process or mailbox growth over time.

Without these metrics, we cannot accurately assess system efficiency, detect early signs of saturation, or evaluate the resource impact of architectural decisions (e.g., process-per-tenant patterns, fanout strategies, or memory allocation behavior).

# Expected Gains

Improved runtime observability will enable us to:
- Detect scheduler contention and saturation before it impacts performance.
 - Identify memory leaks, fragmentation, and inefficient allocation patterns.
- Track process and ETS growth to prevent runaway resource usage.
- Quantify the effect of design decisions on BEAM resource efficiency.
- Establish meaningful alert thresholds and capacity planning baselines.

# Metrics Roadmap

All metrics will be emitted periodically via Telemetry and exported to Honeycomb using the existing OpenTelemetry pipeline.

Rank | Metric | Purpose | Source | Suggested Interval
-- | -- | -- | -- | --
1 | beam.scheduler.utilization | Detect scheduler contention and core saturation | erlang:statistics(scheduler_wall_time) | 15s
2 | beam.run_queue.length | Measure runnable process backlog | erlang:statistics(run_queue_lengths) | 15s
3 | beam.memory.by_type | Observe memory usage by category (processes, binaries, ETS, atoms) | erlang:memory/0 | 15s
4 | beam.process.count | Track process growth and supervision load | erlang:system_info(process_count) | 30s
5 | beam.gc.pause_ms | Measure garbage collection frequency and pause times | erlang:statistics(garbage_collection) | 30s
6 | beam.process.mailbox_topN | Identify processes with large message queues | recon:proc_count(message_queue_len, 10) | 30s
[...]

PR #3308, opened 2025-10-20
Expose more BEAM metrics in ApplicationTelemetry

alco:
Part of #3237.

This PR exposes the following metrics:
- individual scheduler utilization, as measured by taking two samples with 100ms between them
- invididual scheduler run queue lengths
- VM-wide garbage collection stats (number of runs and total bytes reclaimed)
- VM-wide reduction stats (on every measurement we report the delta from the previous one)

Memory metrics, process/atom/etc counts were already being collected. I've added a few more memory stats to the list of exported metrics.

This is all implemented in as custom probes on `telemetry_poller`, so all are gathered at the same interval. I've created a separate issue #3306 for making the polling interval configurable.

Another two metrics are tracked separately: Identify processes with large message queues (#3304) and Detect ETS memory growth or leaks #3305.

### DIFF
--- a/.changeset/smooth-oranges-camp.md
+++ b/.changeset/smooth-oranges-camp.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Expose more BEAM VM metrics in ApplicationTelemetry that can be exported via OTEL.


packages/sync-service/lib/electric/telemetry/stack_telemetry.ex
```diff
@@ -224,7 +224,6 @@ with_telemetry [OtelMetricExporter, Telemetry.Metrics] do
         last_value("electric.postgres.replication.wal_size", unit: :byte, keep: for_stack(opts)),
         last_value("electric.storage.used", unit: {:byte, :kilobyte}, keep: for_stack(opts)),
         last_value("electric.shapes.total_shapes.count", keep: for_stack(opts)),
-        last_value("electric.postgres.replication.wal_size", unit: :byte, keep: for_stack(opts)),
```
msfstef, line 227: how come we're removing this here? it's generally one we actively use for monitoring

msfstef (approved):
Very nice! look forward to having all these metrics!

Noted the removal of the wal size metric, am I misreading something?

packages/sync-service/lib/electric/telemetry/stack_telemetry.ex
alco: This line is a copy of line 224 above. I'm just removing the duplication here.

msfstef: ah missed that! good catch

PR #3312, opened 2025-10-21
chore(sync-service): Remove can_alter_publication? from Connection.Manager

robacourt:
Part of a series of PRs to separate the shape and connection subsystems.

This PR removes the `can_alter_publication?` variable from the Connection.Manager state that it uses to initialise the PublicationManager (via the ReplicationSupevisor) since in an upcoming PR Connection.Manager will no longer be starting the ReplicationSupevisor and PublicationManager can already work out `can_alter_publication?` for itself.


msfstef:
I just merged a PR that removes this https://github.com/electric-sql/electric/pull/3297

PR #3315, opened 2025-10-21
(chore)sync-service: Do not clear storage when initializing

robacourt:
Part of a series of PRs to separate the shape and connection subsystems.

This PR removes the unnecessary storage reset when initializing.

### DIFF
--- a/.changeset/nice-roses-pretend.md
+++ b/.changeset/nice-roses-pretend.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Remove unnecessary clearing of storage on initialization"


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -487,13 +487,16 @@ defmodule Electric.Connection.Manager do
...
         {state.pg_system_identifier, state.pg_timeline_id},
         state.timeline_opts
-      ) == :timeline_changed
+      )
+
+    timeline_changed? = timeline_check == :timeline_changed
+    initializing? = timeline_check == :no_previous_timeline
 
-    if timeline_changed? or state.purge_all_shapes? do
+    if timeline_changed? or (state.purge_all_shapes? and not initializing?) do
```
msfstef, line 497: why do we need this extra logic here? surely if we're initializing then resetting empty storage is a no-op and simpler to think about? with the splitting of the subsystems we'll have to requesti t from the shape subsystem anyway

robacourt: At the moment resetting storage on is basically a no-op because you delete nothing. Once the shape subsystem is started independently however we need to stop the shape subsystem, then delete the files, then restart the shape subsystem. So this is to avoid the unnecessary restart of the shape subsystem on initialisation.

The other way to do it is in the reset function to only restart the shape subsystem if shape_count > 0 which is what I did in my original PR, and perhaps it's worth seeing @alco's comments on that here: https://github.com/electric-sql/electric/pull/3238#issuecomment-3376927532 my interpretation being that we skip when explicitly initializing rather than shape_count=0 being the special case.

msfstef: sounds reasonable!

Issue #3286, opened 2025-10-15
Ensure a given Electric instance holds the lock to operate on publication manager


msfstef:
**Versions**
- Electric: a057f9c7dc07743c8be687ba5e45ce17fb9172db

**Bug description**
At any point our lock connection might get lost, due to network partitions or anything else, and another Electric  B waiting (e.g. in a rolling deploy) can immediately grab it and start operating on the slot and publication.

When the lock connection is lost in Electric A, until everything gets shut down and restarted, it might still be operating on the slot and publication without a lock, causing inconsistencies across the two Electrics.

**Expected behavior**
No modifying operations should be made by an Electric if it's not holding the lock.

**Suggested solution**

- Replication connection modifications
  - Solved by moving the lock inside the replication connection https://github.com/electric-sql/stratovolt/issues/811
- Replication slot dropping on cleanup
  - Since we do this in the connection manager, we can ensure the lock is alive before doing that, otherwise the same solution as for the publication manager below
- Publication manager modifications
  - Before any modification to the publication, check that 1) the lock is active, and 2) it is owned by the stack itself, either by matching the backend pid or investigate if the advisory lock can be given some extra metadata to match on.

alco:
> Before any modification to the publication, check that 1) the lock is active, and 2) it is owned by the stack itself, either by matching the backend pid or investigate if the advisory lock can be given some extra metadata to match on.

This doesn't prevent the case where the lock is dropped after the check and acquired by a different instance of Electric.

A fool-proof way to achieve mutual exclusion would be to make any publication modifications in a transaction and acquire the lock once again for the duration of the transaction. From [PostgreSQL docs](https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS):

> A lock can be acquired multiple times by its owning process; for each completed lock request there must be a corresponding unlock request before the lock is actually released. Transaction-level lock requests, on the other hand, behave more like regular lock requests: they are automatically released at the end of the transaction, and there is no explicit unlock operation. This behavior is often more convenient than the session-level behavior for short-term usage of an advisory lock. 

> If a session already holds a given advisory lock, additional requests by it will always succeed, even if other sessions are awaiting the lock; this statement is true regardless of whether the existing lock hold and new request are at session level or transaction level.

[...]

PR #3319, opened 2025-10-21
feat: Move connection lock inside replication connection

msfstef:
Closes https://github.com/electric-sql/stratovolt/issues/811

Removes `LockConnection` as a separate connection process altogether, reducing the number of wal sender processes required to run electric from 2 to 1 and simplifying the connection manager further.

## Overview of changes
- Acquiring the lock is now part of our regular `ConnectionSetup` state machine that runs on the replication connection
- System identification is also now part of the regular state machine setup
- The backend pid is queried along with other `pg_info` to reduce number of separate queries
- Lock acquisition query failures now immediately retry (removed backoff logic internal to LockConnection)
  - I had added backoff logic but having investigated logs of the past 2 months on cloud we never fail the query for any reason other than statement timeouts, which should immediately be retried.
  - Other failures will trigger a full retry and reestablishing the connection, like any connection failure would do
- On `terminate` of the connection manager, if the admin pool is available we forcibly kill the replication conn backend (which includes both the lock and the replication slot usage)
- Since killing the replication client also loses the lock, we cannot guarantee that `drop_slot` runs while we hold the lock
  - This was not a guarantee before either, so I decided to not concern myself too much with this undocumented feature anyway - it works with how we use it internally in cloud
  - I ensure the publication is dropped while we hold the lock such that anyone who grabs the lock will be forced to recreate the slot anyway, as well as ensuring there's no unnecessary buildup

## Followup work
I believe that we should not rely on the admin pool to do cleanups like killing the replication client backend and dropping the slot, since the admin pool is only started after we do initial configuration of the replication connection which includes the lock and replication slots/publication.

We should be performing these cleanups with a one-off, single conn pool connection so we can always at least attempt them upon shutdown.

Ensuring the "drop slot on shutdown" is done in a "protected" way would require us doing a dance with multiple locks like the one I intend to do for https://github.com/electric-sql/electric/issues/3286 - grab a cleanup transaction lock, stop the replication client (and lose the general lock), do the clenaup, release the transaction lock. Other electrics check for the cleanup lock before setting up publication/slot to avoid overlaps. All of this might also be overkill as we cannot guarantee cleanup of a slot in case of unexpected kills/shutdowns.

## Notes on ConnMan
With this change, conn man has become so simple that I think we're very close to just using a regular supervisor, which will also allow us to rethink how we want to do cleanups as we can have a separate monitor process that takes care of these cleanups with its own ephemeral connection (if able to establish it).

### DIFF
--- a/integration-tests/tests/_macros.luxinc
+++ b/integration-tests/tests/_macros.luxinc
@@ -24,7 +24,7 @@
       -p $pg_host_port:5432 \
       -v $(realpath ../scripts/reset_wal.sh):/docker-entrypoint-initdb.d/initdb-reset_wal.sh \
       ${extra_volumes} \
-      postgres:14-alpine \
+      postgres:18-alpine \
         -c wal_level=logical ${config_opts}
 
     [timeout 15]
@@ -56,7 +56,7 @@
       -v $(realpath ../support_files/server.crt):/etc/server.crt:ro \
       -v $(realpath ../support_files/server.key):/etc/server.key:ro \
       --entrypoint entrypoint-ssl.sh \
-      postgres:17-alpine \
+      postgres:18-alpine \
         -c config_file=/etc/postgresql.conf
 
     [timeout 15]


robacourt (approved):
Awesome! Great work!

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -752,31 +700,19 @@ defmodule Electric.Connection.Manager do
...
-        %State{
-          current_phase: :connection_setup,
-          current_step: {:start_replication_client, :connecting}
-        } = state
-      ) do
     tref = schedule_periodic_connection_status_check(:replication_client)
 
+    # As soon as we acquire the connection lock, we try to configure the replication connection
+    # first because it requires additional privileges compared to regular "pooled" connections,
+    # so failure to configure a replication connection should be reported ASAP.
```
alco: This comment is no longer relevant. We've already started the replication connection by this point, there's no dilemma anymore between starting it or the pools first.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -1119,6 +1064,7 @@ defmodule Electric.Connection.Manager do
     )
 
     drop_slot(state)
+    drop_publication(state)
```
alco, line 1064: Above you're dropping the publication before the slot with the reason:
```
    # drop publication before replication client so that we ensure anyone
    # acquiring the lock next will recreate the replication slot
```

Doesn't it apply here too?

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -371,6 +371,11 @@ defmodule Electric.Connection.Manager do
 
         {:noreply, state}
 
+      # the ConnectionResolver process was killed, as part of the application
+      # shutdown in which case we'll be killed next so just return here
+      {:error, :killed} ->
+        {:noreply, state}
```
alco, line 375: We do we need to handle this? Seems like going against the regular OTP supervisor behaviour and trying to be overly defensive.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -945,43 +883,55 @@ defmodule Electric.Connection.Manager do
...
-        :ok
+      shutdown_child(replication_client_pid, reason, 1_000)
     end
 
-    # We brutally kill the lock connection process as it might hang on waiting
-    # to establish a connection and can't be gracefully killed
-    if is_pid(lock_connection_pid), do: shutdown_child(lock_connection_pid, :kill)
+    # ensure the replication backend is killed to avoid orphaned locks and
+    # to ensure dropping the slot is possible if requested
+    state = kill_replication_backend(state)
```
alco, line 907: brutal :see_no_evil:

msfstef: completely - but we had discussed that we want to do this as we often see things hanging after we terminate

packages/sync-service/lib/electric/connection/manager.ex
msfstef: not sure this is replicating code from the lock connection validation and without it tests fail @magnetised

PR #3320, opened 2025-10-21
fix: Handle more inspector connection failures

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3316

This PR also semi-addresses [this Sentry issue](https://electricsql-04.sentry.io/issues/62065592/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream) - this is again a symptom of in-flight requests during a stack shutdown, they will still fail but with less misleading errors.

In our shape validation steps, the `:connection_not_available` errors are handled more gracefully but with a lot of code. This time in the schema header there wasn't an easy way to thread things through, but we should be able to just raise appropriate errors and let the error handler return the appropriate 503 errors.

We do indeed do that, but only for `DBConnection.ConnectionError` - perhaps we should rethink it a bit.

I've also considered refactoring the code in the shape validation handling `:connection_not_available` gracefully and instead just raising `DBConnection.ConnectionError` so that a 503 is returned. Would like to hear the reviewer's thoughts.

This particular fix is hard to put in a test cause it requires losing both the cache and connection after a request has been validated and processed but before the schema header is put in - I could do it with Repatch but decided against adding flimsy tests and just make it make sense, I've tested it manually ofc.

p.s. I also took the opportunity to fix the flake in the nullability router integration test

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -1031,9 +1031,16 @@ defmodule Electric.Shapes.Api do
     # (e.g. column nullability changes but the type remains the same), we might return the new
     # version if it's invalidated in ETS or server is restarted.
     case Inspector.load_column_info(shape.root_table_id, inspector) do
-      {:ok, columns} -> Electric.Schema.from_column_info(columns, shape.selected_columns)
-      {:error, :connection_not_available} -> nil
-      :table_not_found -> nil
+      {:ok, columns} ->
+        Electric.Schema.from_column_info(columns, shape.selected_columns)
+
+      {:error, :connection_not_available} ->
+        # TODO: we currently only convert DBConnection errors to proper 503s, we should
+        # handle a custom error we can more easily propagate
+        raise %DBConnection.ConnectionError{message: "Cannot connect to the database."}
+
+      :table_not_found ->
+        nil
     end
   end
 


robacourt (approved):
Nice!

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
```diff
@@ -364,6 +364,8 @@ defmodule Electric.Postgres.Inspector.EtsInspector do
       :not_in_cache -> :not_in_cache
       oid -> {:ok, {oid, relation}}
     end
+  rescue
+    ArgumentError -> :not_in_cache
```
robacourt, line 368: If the ETS table is not existing then then the process is not going to exist, so won't this mean that the next call errors with noproc? Would it not be better to return an error and not try to call the process?

msfstef: Indeed it moves the error from a table not existing to a noproc call exit - I just think that the noproc error is much clearer in terms of what actually happened, and it could be that the process was restarting and the table was unavailable momentarily.

I think ultimately how we then handle the noproc would fall under the issue of how do we handle inflight requests during stack restarts/shutdowns

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -1031,11 +1031,9 @@ defmodule Electric.Shapes.Api do
...
     # version if it's invalidated in ETS or server is restarted.
     case Inspector.load_column_info(shape.root_table_id, inspector) do
-      {:ok, columns} ->
-        Electric.Schema.from_column_info(columns, shape.selected_columns)
-
-      :table_not_found ->
-        nil
+      {:ok, columns} -> Electric.Schema.from_column_info(columns, shape.selected_columns)
+      {:error, :connection_not_available} -> nil
+      :table_not_found -> nil
```
alco: Why not make `:table_not_found` just another error reason?

msfstef: I agree - will handle in separate PR

PR #3321, opened 2025-10-21
fix: Upgrade PG in integration tests to 18

msfstef:
Seems like our slot invalidation integration tests fail with PG18.

We no longer receive an explicit "wal size exceeded" error (I think [this thread](https://www.postgresql.org/message-id/CAD21AoCh%2BDxBVaRQ3Uz3OzSgtoNZ-ZAjyH%2Be_6OZkqg4g5P6hQ%40mail.gmail.com) is relevant) and instead just receive that we can no longer access the slot and that it has been invalidated.

I started by parsing this error into `:database_slot_max_size_exceeded` as well but decided to spplit into a new emore generic `:database_slot_invalidated` error with:
```Couldn't start replication: slot has been invalidated and can no longer be accessed.```
but still keeping a hint about potentially max size being exceeded - it's probably better to respect the error since it deliberately hides the reason for invalidation and the user can inspect their own database logs. I'd love your opinion @alco on this as I think you know more on the replication/pg side of things.

The other big change seems to be that removing the publication does not trigger the replication connection to fail, no error is transmitted over the stream. We currently don't manually handle the slot missing although I have a TODO in the publication manager to handle that, but this fixes the intent of the integration test. Opened an issue https://github.com/electric-sql/electric/issues/3322 as well.

### DIFF
--- a/integration-tests/tests/recreation-of-replication-slot.lux
+++ b/integration-tests/tests/recreation-of-replication-slot.lux
@@ -43,6 +43,9 @@
 
 [shell electric]
   -
+  ## TODO: we should detect the publication missing and trigger a restart automatically,
+  ## but that requires a manual check since PG18 - this is tracked in the following issue:
+  ## https://github.com/electric-sql/electric/issues/3322
   !:ok = Application.stop(:electric)
   !:ok = Application.start(:electric)
   ??[debug] Starting replication client

### CODE REVIEW

**Comment by alco** [comment]
> I started by parsing this error into `:database_slot_max_size_exceeded` as well but decided to spplit into a new emore generic `:database_slot_invalidated` error

:+1:

File: packages/sync-service/lib/electric/db_connection_error.ex [hunk 1]
```diff
@@ -123,6 +123,33 @@ defmodule Electric.DbConnectionError do
...
 
+  # From PostgreSQL 18 we no longer get the "exceeded the maximum reserved size" detail error
+  # but only a more generic "wal_removed" and that we can no longer access the replication slot.
+  # We can still hint that it is probably due to exceeding max size, but it's safer to not be
+  # so specific given the error deliberately obscures that.
+  def from_error(
+        %Postgrex.Error{
+          postgres: %{
+            code: :object_not_in_prerequisite_state,
+            detail: "This replication slot has been invalidated due to \"wal_removed\"."
```
alco: I think it pays to avoid a literal match here and do a prefix match instead, such that we can cover more possible reasons for invalidation and only include the actual reason in the error message stored on DbConnectionError.

Here's the [commit](https://github.com/postgres/postgres/commit/f41d8468ddea34170fe19fdc17b5a247e7d3ac78#diff-f6d164d3eee939f49155ccd85170350685be75578fc9b0a1135fdb819e32c609R573-R578) in Postgres that changed this error message.

integration-tests/tests/recreation-of-replication-slot.lux
```diff
@@ -43,11 +43,9 @@
 
 [shell electric]
   -
-
-  ??[error] :gen_statem {:"Electric.ProcessRegistry:single_stack", {Electric.Postgres.ReplicationClient, nil}} terminating
-  ??** (Postgrex.Error) ERROR 42704 (undefined_object) publication "electric_publication_integration" does not exist
-  ??[warning] Electric.Connection.Manager is restarting after it has encountered an error
-  ??[debug] Starting replication client for stack single_stack
+  !:ok = Application.stop(:electric)
+  !:ok = Application.start(:electric)
```
alco, line 50: This change neuters the test, it no longer verifies what it is supposed to verify. What do you think about disabling it (e.g. by changing the file extension) until it can be properly fixed or removed as no longer useful?

packages/sync-service/lib/electric/db_connection_error.ex
msfstef: Good idea! We can keep the hint about checking the max wal size regardless of hte invalidation reason I guess, although the reason I was only capturing "wal_removed" was because I think this is the one we'd see for that case and worth adding a hint for.

For other reasons we potentially don't want to include it, what's your opinion?

integration-tests/tests/recreation-of-replication-slot.lux
msfstef: Well the test actually tests: `Verify that Electric recreates the replication slot after it has had to create a new publication`

At least nominally it doesn't test that it fails if the publication is not present, but that if Electric restarts it recovers from a missing publication - I think it's worth keeping this test like this here and I can add a note both on the test and on the issue I created for tirggerring a restart in case the publication is missing!

packages/sync-service/lib/electric/db_connection_error.ex
alco: We can adapt [these explanations](https://github.com/postgres/postgres/blob/2d7b247cb414ccbc052c485fd82a841477a7e1ff/src/backend/replication/slot.c#L1631-L1667) from Postgres source for our use case, to provide basic explanation to the user about the invalidation reason.

To be clear, your current implementation is fine. max_wal_size is by far the only reason we've seen. Though now that PG 18 has idle_replication_slot_timeout, some may people may start playing with it and it will inevitably lead to some suprise invalidations.

integration-tests/tests/recreation-of-replication-slot.lux
alco: You're goddamn right! I got the wrong impression when I took a cursory look through the test.

packages/sync-service/lib/electric/db_connection_error.ex
msfstef: Ah great find! I've opened this issue https://github.com/electric-sql/electric/issues/3325 to track this work separately so as to not block this, thank you for finding this in the pg codebase!

Issue #3322, opened 2025-10-21
Replication not interrupted when publication is dropped


msfstef:
**Bug description**
If the publication is dropped entirely, at least on PG18, the replication connection is not interrupted or sent an error and instead just stops replicating transactions silently.

**Expected behavior**
We should detect the publication does not exist and trigger a reset of the stack

PR #3324, opened 2025-10-22
fix: Force stack restart in case of fatal publication misconfiguration

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3322

This addresses issues raised in https://github.com/electric-sql/electric/pull/3321 and will fix the corresponding integration test once this is merged as well.

I've renamed `restart_connection_subsystem` to `restore_connection_subsystem` as it only does something if it is "sleeping", and added a call API for restarting to actually restart the subsystem. I don't think this is abusing the restarter as I think it is the right layer of indirection to perform this operation.

We could in the future be more precise with how we do this recovery but I think for these fatal misconfigurations this is good enough.

### DIFF
--- a/packages/sync-service/test/electric/replication/publication_manager_test.exs
+++ b/packages/sync-service/test/electric/replication/publication_manager_test.exs
@@ -210,6 +210,9 @@ defmodule Electric.Replication.PublicationManagerTest do
     setup do
       test_pid = self()
 
+      # We don't have the Restarter process running in this test, or the enitre connection subsystem for that matter.
+      # So this mock is just to verfiy the correct function is called when publication manager encounters
+      # a fatal error.
       Repatch.patch(
         Electric.Connection.Restarter,
         :restart_connection_subsystem,


packages/sync-service/test/electric/replication/publication_manager_test.exs
```diff
@@ -207,30 +207,52 @@ defmodule Electric.Replication.PublicationManagerTest do
   end
 
   describe "publication misconfiguration" do
+    setup do
+      test_pid = self()
+
+      Repatch.patch(
```
alco, line 216: Instead of patching the code, we could fetch the supervisor pid before and after the restart and assert that they are not the same.

Even more fidelity can be achieved with a Process.monitor but I think the boilerplate overhead would defeat the benefit in this case.

msfstef: In this test suite we don't actually have any of that part of the stack running - it is almost entirely isolated

alco, line 216:
```suggestion
      # We don't have the Restarter process running in this test, or the enitre connection subsystem for that matter.
      # So this mock is just to verfiy the correct function is called when publication manager encounters
      # a fatal error.
      Repatch.patch(
```

alco: I see. So this is simply testing that a call to `Electric.Connection.Restarter.restart_connection_subsystem()` is made.

msfstef: Yes indeed - adding the comment makes sense

PR #3326, opened 2025-10-22
Remove explicit calls to :erlang.garbage_collect()

alco:
The code is not sitting still and at least one of these calls is becoming obsolete with Gary's work on starting consumer processes lazily. Moreover, the current startup of consumer processes in ShapeCache already spawns a separate `Task` process for each consumer, so it's not that clear where all the garbage is coming from.

In any case, we should refrain from adding explicit GC unless we can reliably reproduce the situtations the show its unquestionable benefit.


msfstef:
I think the garbage comes from listing out all the shapes from ETS in order to start recovering them - we do that 1) in the shape status owner where we load them from disk and 2) in ShapeCache in order to start the consumers

Basically the recovery itself should be done in a task

PR #3327, opened 2025-10-22
fix: Fix relation ref counting in publication manager

msfstef:
We were removing a relation from prepared relations manually rather than waiting for all appropriate shape removals, causing issues in checks we do with ref counting.

We should rely solely on clients removing themselves as we do in the rest of the publication manager code and rely on the ref counting.

### DIFF
--- a/packages/sync-service/test/electric/replication/publication_manager_test.exs
+++ b/packages/sync-service/test/electric/replication/publication_manager_test.exs
@@ -345,27 +345,6 @@ defmodule Electric.Replication.PublicationManagerTest do
 
       refute_receive {:filters, _}, 200
     end
-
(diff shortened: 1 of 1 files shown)


msfstef:
For some further context - whenever we update the ref counting for relations tracked, at the end we check if the relation being update is no longer part of our "prepared" filters for adding/keeping in the publication and if it is not, we reply to all waiters for that relation with a runtime error that "shape removed before updating publication"

However when getting an error during configuring the publication for a particular relation, we were arbitrarily removing it form the prepared filters (but not from our ref counting), and because the ref counting logic does not always re-add things to our filter since it assumes it is "untouched", it ends up firing this error for incoming requests due to errors from previous requests.

I will make the ref counting a bit more resilient as well as part of this.

PR #3328, opened 2025-10-22
feat(sync-service) Gracefully handle no connection while processing txn in ShapeLogCollector

robacourt:
Part of the shape subsystem work:

This PR gracefully handles connection issues while processing transactions in the ShapeLogCollector. It simply returns a connection error which the ReplicationClient responds to by waiting for the connection to be re-established and then retrying.

Please note that the ShapeLogCollector will still crash if there is no connection while subscribing, and will be shut down by the connection manager if the replication client losing connection. These scenarios will be handled by later PRs.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -466,10 +466,12 @@ defmodule Electric.Postgres.ReplicationClient do
     >>
   end
 
-  # Retry applying the given MFA for up to @max_retry_time milliseconds, used for
-  # processing transactions without crashing if the processor is down. The max retry
-  # time is only there to avoid a worst case scenario, as the processor being unable
-  # to process things for this long should lead to the replication client being killed.
+  # Retry applying the given MFA
+  # A retry may need to happen if the connection is available or the collector is not ready yet.
+  # In those instances we wait until the stack is ready and retry, and will go on retrying forever.
+  # We may also get a process down, and we retry here too but with a timeout since processes should
+  # be bought back up by the supervisor and if this carries on for longer than the timeout there may
+  # be a more serious issue.
   @retry_time 10 * 60_000
   @effectively_infinity 100 * 365 * 24 * 60 * 60 * 1000
   @spin_prevention_delay 50


packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -484,12 +471,26 @@ defmodule Electric.Postgres.ReplicationClient do
   # time is only there to avoid a worst case scenario, as the processor being unable
   # to process things for this long should lead to the replication client being killed.
   @max_retry_time 10 * 60_000
-  defp apply_with_retries(mfa, time_remaining \\ @max_retry_time) do
+  @effectively_infinity 100 * 365 * 24 * 60 * 60 * 1000
```
msfstef, line 476: love the name! but can we not just set it to `:infinity`? I guess the API does not support it but it feels like it'd be better if it's part of the status monitor API

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -484,12 +471,26 @@ defmodule Electric.Postgres.ReplicationClient do
...
-      apply(m, f, args)
+      case apply(m, f, args) do
+        :ok ->
+          :ok
+
+        {:error, error} when error in [:not_ready, :connection_not_available] ->
+          Process.sleep(100)
+
+          Electric.StatusMonitor.wait_until_active(state.stack_id,
+            timeout: @effectively_infinity,
```
msfstef, line 491: we're waiting forever on the stack being active but `max_retry_time` is still only 60 seconds  - should we remove `max_retry_time` and instead just retry forever?

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -484,12 +471,26 @@ defmodule Electric.Postgres.ReplicationClient do
...
     {m, f, args} = mfa
 
     try do
-      apply(m, f, args)
+      case apply(m, f, args) do
+        :ok ->
+          :ok
+
+        {:error, error} when error in [:not_ready, :connection_not_available] ->
+          Process.sleep(100)
```
msfstef: nit: we're doing 50 milliseconds for other retries - should we set an attribute like `@retry_timeout 50` and use that both here and in the `receive do ... after` timeout?

also, if we're going to call `wait_until_active` is this sleep still necessary?  worst case scenario it will retry a few times before it actually holds on `wait_until_active`, but that way we're optimizing for the smallest possible amount of downtime at the cost of a few tight loops

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -484,12 +471,26 @@ defmodule Electric.Postgres.ReplicationClient do
...
     start_time = System.monotonic_time(:millisecond)
     {m, f, args} = mfa
 
     try do
-      apply(m, f, args)
+      case apply(m, f, args) do
+        :ok ->
+          :ok
+
+        {:error, error} when error in [:not_ready, :connection_not_available] ->
```
msfstef, line 487: what happens if we get any other error - I suppose we crash the replication client on a match error? Before we were just logging and continuing (with a TODO that's been there for >6 months lol)

robacourt: yeah, the logging and continuing was not used in practice and if it ever was it would be a bad idea!

basically we want to crash the ShapeLogCollector if it's something we don't expect. I was leaving it as a match error

msfstef: I assume you mean crash the replication client - I think it makes sense yeah if it's an unexpected error I suppose we can't do much

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: damn you not liking my hack! it is a little fiddly to do it right but that's no excuse. mind if I do it in a separate PR?

msfstef: haha almost got me with the cool attribute name - sure yeah separate PR sounds good

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: they could be the same value indeed. i think it's still worth having something to prevent spinning and the CPU maxing out for no reason

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: well, exits I don't think we should wait for ever, the shape log collector process should be back up in no time and if it isn't there's a problem. it could even be a shorter time than 10 mins but I didn't set that time and don't know if there's implications with shutdown

packages/sync-service/lib/electric/postgres/replication_client.ex
msfstef: fair enough, a 50-100ms timeout should be manageable

packages/sync-service/lib/electric/postgres/replication_client.ex
msfstef: sounds good - might be worth leaving a comment explaining the different retries though for posterity

robacourt: i've added a comment

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: i've added an attribute to make it clearer

PR #3332, opened 2025-10-23
feat(sync-service): Simplify consumer processes

magnetised:
This is a continuation of with work @alco started in #3230

Instead of a supervisor plus snapshotter, storage and the consumer process, just start the consumer process directly and have it launch a snapshotter if needed.

> The ConsumerSupervisor tree is collapsed into a single Consumer
process
>
> Stack-wide config is stored in ETS and is looked up directly by Consumer processes when needed. This avoid copying the same config through ShapeCache -> DynamicConsumerSupervisor
>
> Shape structs are looked up in ShapeStatus directly by Consumer processes. This significantly reduces the mem footprint of DynamicConsumerSupervisor (which keeps every child's config in its state) for sources that define large
shapes (e.g. with long columns lists and/or where clauses) and allows us to eliminate more copying by no longer copying the whole ShapeStatus state into every task that starts one Consumer process.

### DIFF
--- a/packages/sync-service/test/electric/shape_cache/shape_status_test.exs
+++ b/packages/sync-service/test/electric/shape_cache/shape_status_test.exs
@@ -1,14 +1,12 @@
 defmodule Electric.ShapeCache.ShapeStatusTest do
   use ExUnit.Case, async: true
-  use Support.Mock
+  use Repatch.ExUnit, assert_expectations: true
 
   alias Electric.Replication.LogOffset
   alias Electric.ShapeCache.ShapeStatus
   alias Electric.Shapes.Shape
 
-  import Mox
-
-  setup :verify_on_exit!
+  import Support.TestUtils, only: [expect_calls: 3, patch_calls: 3]
 
   @inspector {__MODULE__, []}
 
@@ -52,9 +50,11 @@ defmodule Electric.ShapeCache.ShapeStatusTest do
   end
 
   defp new_state(_ctx, opts \\ []) do
-    Mock.Storage
-    |> stub(:metadata_backup_dir, fn _ -> nil end)
-    |> expect(:get_all_stored_shapes, 1, fn _ -> {:ok, Access.get(opts, :stored_shapes, %{})} end)
+    stub_storage([force: true], metadata_backup_dir: fn _ -> nil end)
+
+    expect_storage(
+      get_all_stored_shapes: fn _ -> {:ok, Access.get(opts, :stored_shapes, %{})} end
+    )
 
     state = shape_status_opts(opts)
 
@@ -69,6 +69,14 @@ defmodule Electric.ShapeCache.ShapeStatusTest do
       end
 
     {:ok, state, shape_handles}
+  end
+
+  defp stub_storage(opts \\ [], stubs) do
+    patch_calls(Electric.ShapeCache.Storage, opts, stubs)
+  end
+
+  defp expect_storage(opts \\ [], expectations) do
+    expect_calls(Electric.ShapeCache.Storage, opts, expectations)
   end
 
   test "starts empty", ctx do
@@ -381,10 +389,12 @@ defmodule Electric.ShapeCache.ShapeStatusTest do
       table = table_name()
 
       # First lifecycle: no shapes in storage, start empty, add a shape, terminate to create backup
-      Mock.Storage
-      |> stub(:metadata_backup_dir, fn _ -> backup_base_dir end)
-      |> expect(:get_all_stored_shapes, 1, fn _ -> {:ok, %{}} end)
-      |> stub(:get_all_stored_shape_handles, fn _ -> {:ok, MapSet.new()} end)
+      stub_storage(
+        metadata_backup_dir: fn _ -> backup_base_dir end,
+        get_all_stored_shape_handles: fn _ -> {:ok, MapSet.new()} end
+      )
+
+      expect_storage(get_all_stored_shapes: fn _ -> {:ok, %{}} end)
 
       state = shape_status_opts(table: table)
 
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -365,50 +332,39 @@ defmodule Electric.ShapeCache do
...
       materialized_type =
         shape.where.used_refs |> Map.fetch!(["$sublink", Integer.to_string(index)])
 
-      ConsumerSupervisor.start_materializer(%{
-        stack_id: state.stack_id,
+      Shapes.DynamicConsumerSupervisor.start_materializer(state.consumer_supervisor, %{
+        stack_id: stack_id,
         shape_handle: shape_handle,
-        storage: state.storage,
+        storage: ShapeCache.Storage.for_stack(stack_id),
```
alco, line 345: Since this storage can be looked up by `stack_id`, this call can be moved inside `Materializer`'s `init()` callback. That way we'll avoid duplicating the stack storage struct for each Materializer child of DynamicConsumerSupervisor.

packages/sync-service/lib/electric/shapes/monitor/ref_counter.ex
```diff
@@ -298,23 +273,13 @@ defmodule Electric.Shapes.Monitor.RefCounter do
...
 
-  defp monitor_consumer_processes(state, consumer_pid, supervisor_pid, shape_handle) do
-    if !supervisor_pid,
-      do:
-        raise(RuntimeError,
-          message: "No ConsumerSupervisor process found for shape #{shape_handle}"
-        )
-
-    Process.monitor(supervisor_pid, tag: {:down, :writer_supervisor, shape_handle})
+  defp monitor_consumer_processes(state, consumer_pid, shape_handle) do
```
alco, line 282:
```suggestion
  defp monitor_consumer_process(state, consumer_pid, shape_handle) do
```

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -1,89 +1,107 @@
...
-  end
-
-  def await_snapshot_start(consumer, timeout) do
-    GenServer.call(name(consumer), :await_snapshot_start, timeout)
+  @spec await_snapshot_start(Electric.stack_id(), ShapeCache.shape_handle(), timeout()) ::
+          :started | {:error, any()}
+  def await_snapshot_start(stack_id, shape_handle, timeout \\ @default_snapshot_timeout)
+      when is_binary(stack_id) and is_binary(shape_handle) do
+    stack_id
+    |> whereis(shape_handle)
```
alco, line 39: I'm not a fan of duplicating this lookup all over the module. Could you define a private function `consumer(stack_id, shape_handle)`?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -1,89 +1,107 @@
...
   def whereis(stack_id, shape_handle) do
-    GenServer.whereis(name(stack_id, shape_handle))
+    ConsumerRegistry.whereis(stack_id, shape_handle)
+  end
+
+  def stop_and_clean(stack_id, shape_handle) do
+    # if consumer is present, terminate it gracefully
+    stack_id
+    |> whereis(shape_handle)
+    |> GenServer.call(:stop_and_clean, 30_000)
```
alco, line 70: This timeout could also be defined as a module attribute at the top of the module.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -1,89 +1,107 @@
...
 
     {action, config} = Map.pop!(config, :action)
+    # The existing body of consumer tests made it impossible to replace this dynamic
+    # setting with a static module alias. But do note that this is only set to anything
+    # other than Electric.ShapeCache.ShapeStatus in the test environment.
+    shape_status_mod = Map.get(config, :shape_status_mod) || Electric.ShapeCache.ShapeStatus
+    {:ok, shape} = shape_status_mod.fetch_shape_by_handle(stack_id, shape_handle)
+
+    consumer_supervisor =
+      Map.get(config, :consumer_supervisor, Shapes.DynamicConsumerSupervisor.name(stack_id))
```
alco, line 100: I take it that we cannot use a static module reference here because of tests?

We know from our supervision tree architecture that the DynamicConsumerSupervisor process is a parent of the consumer process by definition. Perhaps, we should define a private function that performs a lookup for the first `$ancestor` in the process dictionary of the consumer process when it needs to get hold of the supervisor? All in the name of reducing the amount of state that is passed down from supervisors to children and stored in children's state, even though it's only used during process initialization.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -309,9 +362,10 @@ defmodule Electric.Shapes.Consumer do
   end
 
   # We're trapping exists so that `terminate` is called to clean up the writer,
-  # otherwise we respect the OTP exit protocol.
-  def handle_info({:EXIT, _from, reason}, state) do
-    Logger.debug("Caught EXIT: #{inspect(reason)}")
+  # otherwise we respect the OTP exit protocol. Since nothing is linked to the consumer
```
alco, line 365: BTW, you have removed the cub-db-backed file storage. So there's no longer the need for `Storage.start_link()` that gets called on line 132 above?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -358,15 +412,16 @@ defmodule Electric.Shapes.Consumer do
...
-      state
+    %{shape: %Shape{root_table_id: root_table_id, root_table: root_table}} = state
 
     Logger.info(
       "Schema for the table #{Utils.inspect_relation(root_table)} changed - terminating shape #{state.shape_handle}"
     )
 
     # We clean up the relation info from ETS as it has changed and we want
     # to source the fresh info from postgres for the next shape creation
+    inspector = Inspector.for_stack(state.stack_id)
```
alco, line 423: We look up the stack-wide inspector here. But in StackSupervisor's initialization I see inspector being passed as an option to shapecache. Different use cases because of tests?

magnetised:
Moved to #3389 and #3390

PR #3333, opened 2025-10-23
feat(sync-service): Gracefully handle connection not available in subscribe and relation changes

robacourt:
Part of the shape subsystem work:

This PR gracefully handles connection issues while:
-  processing relations in the ShapeLogCollector. It simply returns a connection error which the ReplicationClient responds to by waiting for the connection to be re-established and then retrying.
- processing shape consumer subscriptions. The shape consumer responds by shutting down and removing the shape.

Please note that the ShapeLogCollector will still be shut down by the connection manager if the replication client losing connection. This scenarios will be handled by later PR.

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -398,7 +398,7 @@ defmodule Electric.Replication.ShapeLogCollector do
           "Dropping relation message for #{inspect(rel.schema)}.#{inspect(rel.table)}: no active consumers"
         end)
 
-        %{state | tracked_relations: tracker_state}
+        {:ok, %{state | tracked_relations: tracker_state}}
 
       _ ->
         case Partitions.handle_relation(state.partitions, updated_rel) do


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -169,8 +169,10 @@ defmodule Electric.Replication.ShapeLogCollector do
           |> Enum.reduce(
             {state.partitions, state.filter, state.dependency_layers, 0},
             fn {shape_handle, shape}, {partitions, filter, layers, count} ->
+              {:ok, partitions} = Partitions.add_shape(partitions, shape_handle, shape)
```
msfstef, line 172: so if it fails on restoration we would just crash - which is fine cause it's initialisation

robacourt: yes. i would like this to be more graceful too but not quite sure how to do that yet, and in any case will be a separate PR

robacourt:
> Looks good! I feel like it might be cleaner if the inspector calls were somehow extracted from partitions so it's not polluted from connection related code - but not sure if we can gain much from that

Yeah, I like that idea in principal, make Partitions entirely functional and just pass in the relation info/oid data structure. There is still the call to Inspector.clean which would be more complicated to remove. I guess I feel Parititions looks quite manageable to me and I'd rather keep the complexity out of the currently over complex ShapeLogCollector. I'd like to do some work simplifying ShapeLogCollector so perhaps we should consider this after that.

PR #3334, opened 2025-10-23
chore(sync-service): Allow StatusMonitor.wait_until_active to take infinite timeout

robacourt:
Based on @msfstef 's feedback in [#3328](https://github.com/electric-sql/electric/pull/3328#discussion_r2454814647) this PR allows StatusMonitor.wait_until_active to have an infinite timeout

### DIFF
--- a/packages/sync-service/lib/electric/status_monitor.ex
+++ b/packages/sync-service/lib/electric/status_monitor.ex
@@ -185,19 +185,19 @@ defmodule Electric.StatusMonitor do
     end
   end
 
-  @retry_time 10
+  @spin_prevention_delay 10
   defp maybe_retry_wait_until_active(_stack_id, _opts, timeout, last_error)
-       when timeout <= @retry_time do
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/status_monitor.ex
```diff
@@ -193,7 +193,14 @@ defmodule Electric.StatusMonitor do
 
   defp maybe_retry_wait_until_active(stack_id, opts, timeout, _) do
     Process.sleep(@retry_time)
```
alco: This is not part of your changes but it looks odd that in this context `@retry_time` means "the amount of time to sleep before the next attempt" whereas in the replication client it means "the total time alloted for retries".

This value in StatusMonitor seems more similar to `@spin_prevention_delay` in the replication client.

robacourt: Good point. I've renamed it to `@spin_prevention_delay`

PR #3336, opened 2025-10-24
Update Elixir deps

alco:


msfstef (approved):
Nothing that looks particularly dangerous - fire away 🚀

PR #3337, opened 2025-10-24
Use builtin :json.encode() function where possible

alco:


msfstef:
Seems like tuples are not supported natively so we might need to have a wrapper module to provide a tuple encoder

alco:
@msfstef Apart from tuples, we have a bunch of structs getting encoded as JSON. To support them with the builtin JSON, we'd need to use a wrapper function everywhere that passes a custom encoding function to `:json.encode()`.

I tried making this change as an experiment. I don't think it is a good idea to switch after all. Jason's protocol-based DX is nice.

PR #3344, opened 2025-10-26
feat(sync-service): Make shape subsystem resilient to connection failures

robacourt:
This PR allows the shape subsystem (processes under the soon to be renamed Replication.Supervisor) to stay running even when the connection subsystem is torn down due to a connection failure.

I've run a bit of chaos testing, dropping the connection randomly while creating shapes. We still get 500s for the snapshot  connection failure which I'll address on a separate PR, but otherwise it looks good.

### DIFF
--- a/integration-tests/tests/db-connection-scaledown.lux
+++ b/integration-tests/tests/db-connection-scaledown.lux
@@ -124,7 +124,6 @@
 
   ??[info] Acquiring lock from postgres
   ??[info] Starting replication from postgres
-  ??[debug] Starting consumer for $handle
 
   ??[debug] Chunked 200 in
 


integration-tests/tests/db-connection-scaledown.lux
```diff
@@ -124,7 +124,6 @@
 
   ??[info] Acquiring lock from postgres
   ??[info] Starting replication from postgres
-  ??[debug] Starting consumer for $handle
```
robacourt, line 127: We no longer stop the shape subsystem when the database is idle (hence this change in the lux tests) however we should for stop it for scale to zero. I aim to address this in a separate PR.

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -479,10 +484,18 @@ defmodule Electric.Connection.Manager do
...
 
-    with {:error, reason} <-
-           Electric.Connection.Manager.Supervisor.start_replication_supervisor(repl_sup_opts) do
-      Logger.error("Failed to start shape supervisor: #{inspect(reason)}")
-      exit(reason)
+    case Electric.CoreSupervisor.start_replication_supervisor(repl_sup_opts) do
+      {:ok, _pid} ->
+        :ok
+
+      {:error, {:already_started, _pid}} ->
```
msfstef, line 491: I'm confused by this comment and clause - this can happen if we didn't explicitly stop the replication supervisor in L448 above.

Is this meant to be here for your next PRs when you remove the stopping of the replication supervisor? And if so, would it not make more sense to add this clause in that PR?

msfstef (approved):
Looks good! Left a comment about restarting

packages/sync-service/lib/electric/connection/manager.ex
robacourt: The only reason we stop the replication supervisor in L448 is if we're resetting the storage which is an edge case, so I'm not considering that case in this comment. 

Ignoring that case, the main reason why you get `already_started` or not is whether the Conn.Man is restarting or not. If it's starting for the first time it'll need to start the shape subsystem too, if it's restarting it'll already be up.

> next PRs when you remove the stopping of the replication supervisor

There's no such planned PR. The stopping has to stay so we can reset the storage.

msfstef: Aaaah sorry misread it and missed that we only stop it for cleanups! Yes makes sense

PR #3348, opened 2025-10-27
Implement load shedding with Retry-After 

KyleAMathews:
### Problem

Electric currently lacks critical load shedding mechanisms to handle traffic spikes and prevent cascading failures. During boot storms or database outages:

- **Client**: Does not honor `Retry-After` headers, has no jitter in backoff, and allows unlimited retries that amplify load
- **Server**: Has no admission control, never returns 429/503 with `Retry-After`, and allows unbounded concurrency that can starve the WAL consumer

This creates a "thundering herd" problem where retry storms amplify load and prevent recovery.

### Solution

This PR implements Priority 1 load shedding improvements based on industry best practices (AWS exponential backoff with jitter, Finagle retry budgets, Google SRE overload handling):

**Client (TypeScript):**
- Honor `Retry-After` headers from 429/503 responses
- Add full jitter to prevent synchronized retries
- Implement retry budget (10% of traffic) to prevent retry storms
- Add configurable max retries (default: 10)

**Server (Elixir):**
- Return `Retry-After` headers on all 503 responses (5-10 seconds)
- Implement admission control with ETS-based concurrency limits (default: 1000 per stack)
- Fail fast with 503 + adaptive `Retry-After` when overloaded
- Add `max_concurrent_requests` configuration

#### Client Changes (`packages/typescript-client/src/fetch.ts`)

**1. Retry-After Header Support**
- Parse `Retry-After` header in both seconds and HTTP-date formats
- Wait at least the server-specified duration before retrying
- Apply jitter on top of `Retry-After` to prevent thundering herd

**2. Retry Budget**
- Track retry rate in 60-second rolling windows
- Limit retries to 10% of total requests (configurable via `retryBudgetPercent`)
- Prevent retry storms from amplifying outages
- Allow first 10 requests to bypass budget (cold start protection)

**3. Max Retries**
- Add `maxRetries` option (default: 10)
- Give up after max attempts to prevent infinite retry loops

**4. Full Jitter**
- Implement AWS recommended full jitter pattern
- Spreads retry attempts across time to avoid synchronization

#### Server Changes

**1. Response Module (`packages/sync-service/lib/electric/shapes/api/response.ex`)**
- Add `:retry_after` field to Response struct
- Add `put_retry_after_header/2` to set `Retry-After` HTTP header
- Include header in response pipeline

**2. API Module (`packages/sync-service/lib/electric/shapes/api.ex`)**
- Stack not ready → `Retry-After: 5`
- Long poll timeout → `Retry-After: 10`
- Snapshot errors → `Retry-After: 10`

**3. Error Handler (`packages/sync-service/lib/electric/plug/serve_shape_plug.ex`)**
- Database unreachable errors → `Retry-After: 10`
- Add `check_admission` plug to request pipeline
- Calculate adaptive retry-after with jitter (5-10 seconds)

**4. Admission Control (`packages/sync-service/lib/electric/admission_control.ex`)** ⭐ NEW
- ETS-based concurrent request counter per stack
- `try_acquire/2` - atomically acquire permit or return `:overloaded`
- `release/1` - release permit when request completes
- Zero database overhead, cheap atomic operations
- Supports monitoring via `get_current/1`

**5. Application Setup (`packages/sync-service/lib/electric/application.ex`)**
- Add `Electric.AdmissionControl` to supervision tree
- Add `max_concurrent_requests` to API configuration
- Pass config to plugs

**6. Configuration (`packages/sync-service/lib/electric/config.ex`)**
- Add `max_concurrent_requests: 1000` default

### Benefits

✅ **Coordinated backoff** - Server tells clients when to retry, clients honor it  
✅ **Prevent thundering herd** - Jitter spreads retry attempts across time  
✅ **Prevent retry storms** - Retry budget caps retries at 10% of traffic  
✅ **Protect WAL consumer** - Admission control prevents unbounded concurrency  
✅ **Fail fast** - Return 503 immediately instead of queuing indefinitely  
✅ **Observable** - Retry-After headers make backoff behavior visible  
✅ **Zero DB overhead** - ETS-based admission control, no database queries  

### Configuration

**Server (environment variable or config):**
```bash
# Max concurrent requests per stack (default: 1000)
export ELECTRIC_MAX_CONCURRENT_REQUESTS=1000
Or in config:

config :electric, :max_concurrent_requests, 1000
Client (TypeScript):

  backoff: {
    maxRetries: 10,           // Max retry attempts (default: 10)
    retryBudgetPercent: 0.1,  // 10% retry budget (default: 0.1)
    initialDelay: 100,        // Initial delay in ms (default: 100)
    maxDelay: 10_000,         // Max delay in ms (default: 10000)
    multiplier: 1.3,          // Backoff multiplier (default: 1.3)
    debug: true               // Log retry attempts
  }
})
Testing Recommendations
Load Testing:

Boot storm: 1000 clients starting simultaneously

Before: Server collapses, WAL consumer falls behind
After: Admission control sheds load, clients back off with jitter
Sustained overload: Send > 1000 concurrent requests

Verify: 503 responses with Retry-After headers
Verify: Clients wait before retrying
Database outage: Kill Postgres for 30 seconds

Verify: 503 + Retry-After: 10 responses
Verify: Retry budget prevents amplification
Retry budget: Trigger many errors

Verify: Retry rate stays ≤ 10% of total requests
Manual Testing:

# Check admission control is running
curl http://localhost:3000/v1/shape?table=items -I
# Should see Retry-After header when overloaded

# Check concurrent request count
# Look for admission control logs in server output
Backward Compatibility
✅ Fully backward compatible

Client: All new options are optional with sensible defaults
Server: Only returns new headers, doesn't break existing clients
Configuration: New max_concurrent_requests has default value
Related Documentation
See LOAD_SHEDDING_EVALUATION.md for:

Detailed evaluation and analysis
Priority 2 improvements (adaptive backoff, per-tenant rate limiting)
Priority 3 improvements (bootstrap coalescing, NGINX edge limiting)
Industry references (AWS, Google SRE, Netflix, Finagle)
Future Work (Not in this PR)
Priority 2 (next sprint):

Adaptive Retry-After based on P95 latency and replication lag
Per-tenant rate limiting with token buckets
Bootstrap coalescing endpoint for CDN
Priority 3 (polish):

NGINX edge rate limiting
Metrics and observability dashboards
Load testing framework

### DIFF
--- a/packages/sync-service/lib/electric/admission_control.ex
+++ b/packages/sync-service/lib/electric/admission_control.ex
@@ -42,6 +42,11 @@ defmodule Electric.AdmissionControl do
     GenServer.start_link(__MODULE__, [], name: __MODULE__)
   end
 
+  @allowed_kinds ~w|initial existing|a
+  for {kind, pos} <- Enum.with_index(@allowed_kinds, 2) do
+    defp tuple_pos(unquote(kind)), do: unquote(pos)
+  end
+
   @doc """
   Try to acquire a permit for the given stack_id.
 
@@ -60,7 +65,7 @@ defmodule Electric.AdmissionControl do
       {:error, :overloaded}
 
   """
-  def try_acquire(stack_id, opts \\ []) do
+  def try_acquire(stack_id, kind, opts \\ []) when kind in @allowed_kinds do
     max_concurrent =
       Keyword.get_lazy(opts, :max_concurrent, fn ->
         Electric.Config.get_env(:max_concurrent_requests)
@@ -68,24 +73,29 @@ defmodule Electric.AdmissionControl do
 
     # Atomically increment counter, but cap at max_concurrent
     # ETS update_counter format: {position, increment, threshold, set_value}
-    # Position 2 is the counter value in tuple {stack_id, counter}
     current =
       :ets.update_counter(
         @table_name,
         stack_id,
-        {2, 1, max_concurrent, max_concurrent},
-        {stack_id, 0}
+        {tuple_pos(kind), 1, max_concurrent, max_concurrent},
+        {stack_id, 0, 0}
       )
 
     if current >= max_concurrent do
       # At or over capacity, decrement back and reject
-      :ets.update_counter(@table_name, stack_id, {2, -1, 0, 0}, {stack_id, 0})
+      :ets.update_counter(@table_name, stack_id, {tuple_pos(kind), -1, 0, 0}, {stack_id, 0, 0})
 
       # Emit telemetry event
       :telemetry.execute(
         [:electric, :admission_control, :reject],
         %{count: 1},
-        %{stack_id: stack_id, reason: :overloaded, current: current, limit: max_concurrent}
+        %{
+          stack_id: stack_id,
+          reason: :overloaded,
+          kind: kind,
+          current: current,
+          limit: max_concurrent
+        }
       )
 
       {:error, :overloaded}
@@ -95,7 +105,7 @@ defmodule Electric.AdmissionControl do
       :telemetry.execute(
         [:electric, :admission_control, :acquire],
         %{count: 1, current: current},
-        %{stack_id: stack_id, limit: max_concurrent}
+        %{stack_id: stack_id, kind: kind, limit: max_concurrent}
       )
 
       :ok
--- a/packages/sync-service/lib/electric/config.ex
+++ b/packages/sync-service/lib/electric/config.ex
@@ -66,8 +66,8 @@ defmodule Electric.Config do
     stack_ready_timeout: 5_000,
     send_cache_headers?: true,
     max_shapes: nil,
-    # This value should be tuned for the hardware it's running on
-    max_concurrent_requests: 4000,
+    # This value should be tuned for the hardware it's running on.
+    max_concurrent_requests: 1000,
     ## Storage
     storage_dir: "./persistent",
     storage: &Electric.Config.Defaults.storage/0,
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -5,6 +5,7 @@ defmodule Electric.Plug.ServeShapePlug do
   # The halt/1 function is redefined further down below
   import Plug.Conn, except: [halt: 1]
 
+  alias Electric.Replication.LogOffset
   alias Electric.Utils
   alias Electric.Shapes.Api
   alias Electric.Telemetry.OpenTelemetry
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: packages/typescript-client/src/fetch.ts [hunk 1]
```diff
@@ -21,24 +21,75 @@ import {
 // want to retry
 const HTTP_RETRY_STATUS_CODES = [429]
 
+// Retry budget tracking (module-level)
```
icehaunter: I don't like this being module-level instead of class-instance or closure variables. Feels unnecessary and harder to test

LOAD_SHEDDING_EVALUATION.md
```diff
@@ -0,0 +1,933 @@
+# Electric Load Shedding Evaluation & Implementation Plan
```
icehaunter: I don't think this or retry strategy doc should be committed

packages/sync-service/lib/electric/admission_control.ex
```diff
@@ -0,0 +1,150 @@
...
+
+      iex> Electric.AdmissionControl.try_acquire("stack-123", max_concurrent: 1000)
+      :ok
+
+      iex> Electric.AdmissionControl.try_acquire("stack-123", max_concurrent: 1)
+      {:error, :overloaded}
+
+  """
+  def try_acquire(stack_id, opts \\ []) do
+    max_concurrent = Keyword.get(opts, :max_concurrent, 1000)
```
icehaunter: Is 1000 concurrent requests enough with our live requests being fairly long-running? That means 1000 concurrent clients waiting for silent shapes will functionally halt the stack from serving any new clients, which feels too low to me

packages/sync-service/lib/electric/admission_control.ex
```diff
@@ -0,0 +1,150 @@
...
+        %{count: 1},
+        %{stack_id: stack_id, reason: :overloaded, current: current, limit: max_concurrent}
+      )
+
+      {:error, :overloaded}
+    else
+      # Successfully acquired permit
+      # Emit telemetry for current concurrency level
+      :telemetry.execute(
+        [:electric, :admission_control, :acquire],
```
icehaunter, line 106: We want to add this telemetry event to exported list under `StackTelemetry`

LOAD_SHEDDING_EVALUATION.md
KyleAMathews: yeah, that was claude code web pushing it out

packages/sync-service/lib/electric/admission_control.ex
KyleAMathews: I think @msfstef suggested last night that we could tie this number to the pool size of the source.

packages/typescript-client/src/fetch.ts
KyleAMathews: Yeah for sure 👍

KyleAMathews: Also you can have many instances of the client going

packages/sync-service/lib/electric/admission_control.ex
icehaunter: Pool size is one thing, live requests to existing stable shapes is another. I don't think they're necessarily correlated

KyleAMathews: Oh right sorry I meant a pool just for creating new shapes as yeah reading existing shapes has different constraints

PR #3349, opened 2025-10-28
Add write_concurrency to ShapeStatus ETS tables

KyleAMathews:
Enables `write_concurrency: true` on ShapeStatus LastUsedTable and MetaTable to reduce lock contention during concurrent shape operations.

## Problem
ShapeStatus tables experience high concurrent write activity:
- 5-50+ API clients updating timestamps via `await_snapshot_start()`
- 10-20+ consumers updating offsets/xmins during transaction processing
- Shape deletes blocked waiting for global write lock
- Performance degradation with large numbers of shapes

## Solution
Add `write_concurrency: true` to both ShapeStatus tables since they have:
- Multiple concurrent writers to distributed keys (different shape_handle per operation)
- High write frequency under load
- No hot key contention

## Expected Impact
- Faster shape deletion operations
- Better scalability with increasing concurrent shapes
- Reduced lock contention on API operations

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -412,7 +412,6 @@ defmodule Electric.ShapeCache.ShapeStatus do
       :named_table,
       :public,
       :ordered_set,
-      read_concurrency: true,
       write_concurrency: true
     ])
 


packages/sync-service/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @core/sync-service
 
+## Unreleased
+
+### Patch Changes
+
+- Enable read and write concurrency for ETS tables to improve performance under concurrent workloads. Added `write_concurrency: true` to ShapeStatus tables (LastUsedTable and MetaTable) to reduce lock contention during concurrent shape operations, addressing slow deletes with large numbers of shapes. Also added appropriate concurrency options to other high-traffic tables including StatusMonitor, EtsInspector, CallHomeReporter, RefCounter, StackConfig, InMemoryStorage, and PureFileStorage.
```
msfstef: is this supposed to be here? I've never seen this, the changelog is supposed to get updated by the changesets publish PR

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -408,14 +408,28 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
 
   defp create_meta_table(stack_ref) do
     meta_table = shape_meta_table(stack_ref)
 
-    :ets.new(meta_table, [:named_table, :public, :ordered_set])
+    :ets.new(meta_table, [
+      :named_table,
+      :public,
+      :ordered_set,
+      read_concurrency: true,
```
msfstef: I'm not entirely sure if we should enable read concurrency, see [docs](https://www.erlang.org/doc/apps/stdlib/ets.html#new/2):

>You typically want to enable this option when concurrent read operations are much more frequent than write operations, or when concurrent reads and writes comes in large read and write bursts (that is, many reads not interrupted by writes, and many writes not interrupted by reads).
>
>You typically do not want to enable this option when the common access pattern is a few read operations interleaved with a few write operations repeatedly. In this case, you would get a performance degradation by enabling this option.

I think that for `last_used` table, we can keep concurrent reads on - the only process reading is the expiry manager and it does so on a set period, so matches the large read bursts description.

But for our meta table, I don't actually think we do any read bursts, instead we continuously read and write to it concurrently, so I would be worried that enabling read concurrency might not be ideal given what the docs say.

packages/sync-service/CHANGELOG.md
KyleAMathews: oh no — Claude got confused

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -408,14 +408,27 @@ defmodule Electric.ShapeCache.ShapeStatus do
   defp create_last_used_table(stack_ref) do
     last_used_table = shape_last_used_table(stack_ref)
 
-    :ets.new(last_used_table, [:named_table, :public, :ordered_set])
+    :ets.new(last_used_table, [
+      :named_table,
+      :public,
+      :ordered_set,
+      read_concurrency: true,
+      write_concurrency: true
```
robacourt:
```suggestion
      write_concurrency: true
```

Last used table is being written to much more than it's being read and only read by a single process, the ExpiryManager, so it shouldn't have `read_concurrency`

robacourt (approved):

alco:
I thought we were going to set it to `write_concurrency: :auto`, since according to the docs

> The `auto` alternative for the `write_concurrency` option is similar to the `true` option but automatically adjusts the synchronization granularity during runtime depending on how the table is used. This is the recommended `write_concurrency` option when using Erlang/OTP 25 and above as it performs well in most scenarios.

robacourt:
> I thought we were going to set it to `write_concurrency: :auto`

Yeah, that sounds better

msfstef:
> I thought we were going to set it to write_concurrency: :auto

Ah yes good point - we should open a new PR

alco:
No worries, I can make one.

PR #3353, opened 2025-10-29
feat: re-add serverside admission control

icehaunter:
This adds back the code + the tests to make sure counters go down correctly

Redo of https://github.com/electric-sql/electric/pull/3348 which we reverted

### DIFF
--- a/packages/sync-service/lib/electric/admission_control.ex
+++ b/packages/sync-service/lib/electric/admission_control.ex
@@ -82,19 +82,11 @@ defmodule Electric.AdmissionControl do
         |> Map.fetch!(kind)
       end)
 
-    # Atomically increment counter
-    # ETS update_counter format: {position, increment}
-    current =
-      :ets.update_counter(
-        table_name,
-        stack_id,
-        {tuple_pos(kind), 1},
-        {stack_id, 0, 0}
-      )
+    current = incr(table_name, stack_id, kind)
 
     if current > max_concurrent do
       # At or over capacity, decrement back and reject
-      :ets.update_counter(table_name, stack_id, {tuple_pos(kind), -1, 0, 0}, {stack_id, 0, 0})
+      decr(table_name, stack_id, kind)
 
       # Emit telemetry event
       :telemetry.execute(
@@ -141,7 +133,7 @@ defmodule Electric.AdmissionControl do
   """
   def release(stack_id, kind, opts \\ []) when kind in @allowed_kinds do
     table_name = Keyword.get(opts, :table_name, @table_name)
-    :ets.update_counter(table_name, stack_id, {tuple_pos(kind), -1, 0, 0}, {stack_id, 0, 0})
+    decr(table_name, stack_id, kind)
     :ok
   end
 
@@ -185,5 +177,13 @@ defmodule Electric.AdmissionControl do
 
     Logger.info("Admission control initialized with table: #{table_name}")
     {:ok, %{table_name: table_name}}
+  end
+
+  defp incr(table_name, stack_id, kind) do
+    :ets.update_counter(table_name, stack_id, {tuple_pos(kind), 1}, {stack_id, 0, 0})
+  end
+
+  defp decr(table_name, stack_id, kind) do
+    :ets.update_counter(table_name, stack_id, {tuple_pos(kind), -1, 0, 0}, {stack_id, 0, 0})
   end
 end
--- a/packages/sync-service/lib/electric/shapes/api/response.ex
+++ b/packages/sync-service/lib/electric/shapes/api/response.ex
@@ -279,7 +279,7 @@ defmodule Electric.Shapes.Api.Response do
        when status >= 400 do
     conn
     |> put_cache_header("cache-control", "no-store", api)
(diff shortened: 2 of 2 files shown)


packages/sync-service/lib/electric/admission_control.ex
```diff
@@ -0,0 +1,189 @@
...
+      :ets.update_counter(
+        table_name,
+        stack_id,
+        {tuple_pos(kind), 1},
+        {stack_id, 0, 0}
+      )
+
+    if current > max_concurrent do
+      # At or over capacity, decrement back and reject
+      :ets.update_counter(table_name, stack_id, {tuple_pos(kind), -1, 0, 0}, {stack_id, 0, 0})
```
msfstef: nit: I would extract the increment and decrement to one line private functions just because of the verbosity of ETS syntax - we already call this decrement twice in code

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -51,6 +57,64 @@ defmodule Electric.Plug.ServeShapePlug do
...
+
+      {:error, :overloaded} ->
+        # Calculate adaptive retry-after based on load
+        retry_after = calculate_retry_after(stack_id, max_concurrent)
+
+        response =
+          Api.Response.error(
+            get_in(config, [:api]),
+            %{code: "overloaded", message: "Server is currently overloaded, please retry"},
+            status: 503,
```
msfstef, line 97: why not 429? it seems like the most appropriate one here since this is a rate limiter and is unrelated to service unavailability

don't remember if clients handle it correctly though

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -180,24 +244,45 @@ defmodule Electric.Plug.ServeShapePlug do
 
     error_str = Exception.format(:error, exception)
 
+    conn = fetch_query_params(conn)
+    ensure_admission_control_release(conn)
```
msfstef, line 248: why are we calling this here? we use `register_before_send` which I'm pretty sure covers all cases

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -180,24 +244,45 @@ defmodule Electric.Plug.ServeShapePlug do
...
+    )
   end
 
   def handle_errors(conn, error) do
     OpenTelemetry.record_exception(error.kind, error.reason, error.stack)
 
     error_str = Exception.format(error.kind, error.reason)
 
+    conn = fetch_query_params(conn)
+    ensure_admission_control_release(conn)
```
msfstef, line 267: same here - we use `register_before_send` which should cover all cases

packages/sync-service/lib/electric/shapes/api/response.ex
```diff
@@ -272,10 +274,12 @@ defmodule Electric.Shapes.Api.Response do
   end
 
   # All other 4xx and 5xx responses should never be cached
+  # Use no-store (not no-cache) to prevent CDN caching of error responses
   defp put_cache_headers(conn, %__MODULE__{status: status, api: api})
        when status >= 400 do
     conn
-    |> put_cache_header("cache-control", "no-cache", api)
+    |> put_cache_header("cache-control", "no-store", api)
+    |> Plug.Conn.put_resp_header("surrogate-control", "no-store")
```
msfstef: why not `put_cache_header` which respects the `send_cache_headers?` flag?

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
icehaunter: 429 is a per-client, 503 is server timeout over a lot of clients

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
icehaunter: No. `register_before_send` adds a property to the `Conn` struct, but the `try ... catch` block in `use Plug.ErrorHandler` will call this function in a catch with a different, not-filled `Conn` struct (hence our need to `fetch_query_params` again)

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
msfstef: ah good point - ignore me

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
msfstef: ugh - makes sense - but perhaps the admission control should be a plug outside and before the `serve_shape_plug` in which case the conn read by the handle error will have the registration already?

I don't love it cause we can't as easily reuse it in cloud but I think admission control doesn't have to be part of serve shape plug

icehaunter: I think it does, as it's stack-dependent and bounded by type that makes sense only in context of a shape serve

PR #3354, opened 2025-10-29
fix: re-add reverted fetch client update

icehaunter:
reverted version of the client wasn't published so this just makes sure code matches the published version, hence no changeset

### DIFF
--- a/packages/typescript-client/src/fetch.ts
+++ b/packages/typescript-client/src/fetch.ts
@@ -38,21 +38,8 @@ export interface BackoffOptions {
    * Maximum number of retry attempts before giving up.
    * Set to Infinity (default) for indefinite retries - needed for offline scenarios
    * where clients may go offline and come back later.
-   *
-   * The retry budget provides protection against retry storms even with infinite retries.
    */
   maxRetries?: number
-  /**
-   * Percentage of requests that can be retries (0.1 = 10%)
-   *
-   * This is the primary load shedding mechanism. It limits the *rate* of retries,
-   * not the total count. Even with infinite retries, at most 10% of your traffic
-   * will be retries, preventing retry storms from amplifying server load.
-   *
-   * The budget resets every 60 seconds, so a temporary spike of errors won't
-   * permanently exhaust the budget.
-   */
-  retryBudgetPercent?: number
 }
 
 export const BackoffDefaults = {
@@ -60,7 +47,31 @@ export const BackoffDefaults = {
   maxDelay: 60_000, // Cap at 60s - reasonable for long-lived connections
   multiplier: 1.3,
   maxRetries: Infinity, // Retry forever - clients may go offline and come back
-  retryBudgetPercent: 0.1, // 10% retry budget prevents amplification
+}
+
+/**
+ * Parse Retry-After header value and return delay in milliseconds
+ * Supports both delta-seconds format and HTTP-date format
+ * Returns 0 if header is not present or invalid
+ */
+export function parseRetryAfterHeader(retryAfter: string | undefined): number {
+  if (!retryAfter) return 0
+
+  // Try parsing as seconds (delta-seconds format)
+  const retryAfterSec = Number(retryAfter)
+  if (Number.isFinite(retryAfterSec) && retryAfterSec > 0) {
+    return retryAfterSec * 1000
+  }
+
+  // Try parsing as HTTP-date
+  const retryDate = Date.parse(retryAfter)
+  if (!isNaN(retryDate)) {
+    // Handle clock skew: clamp to non-negative, cap at reasonable max
+    const deltaMs = retryDate - Date.now()
+    return Math.max(0, Math.min(deltaMs, 3600_000)) // Cap at 1 hour
+  }
+
+  return 0
 }
 
 export function createFetchWithBackoff(
@@ -74,37 +85,7 @@ export function createFetchWithBackoff(
     debug = false,
     onFailedAttempt,
     maxRetries = Infinity,
-    retryBudgetPercent = 0.1,
   } = backoffOptions
-
-  // Retry budget tracking (closure-scoped)
-  // Resets every minute to prevent retry storms
-  let totalRequests = 0
-  let totalRetries = 0
-  let budgetResetTime = Date.now() + 60_000
-
-  function checkRetryBudget(percent: number): boolean {
-    const now = Date.now()
-    if (now > budgetResetTime) {
-      totalRequests = 0
-      totalRetries = 0
-      budgetResetTime = now + 60_000
-    }
-
-    totalRequests++
-
-    // Allow retries for first 10 requests to avoid cold start issues
-    if (totalRequests < 10) return true
-
-    const currentRetryRate = totalRetries / totalRequests
-    const hasCapacity = currentRetryRate < percent
-
-    if (hasCapacity) {
-      totalRetries++
-    }
-
-    return hasCapacity
-  }
   return async (...args: Parameters<typeof fetch>): Promise<Response> => {
     const url = args[0]
     const options = args[1]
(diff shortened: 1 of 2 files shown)


packages/typescript-client/src/fetch.ts
```diff
@@ -28,17 +28,39 @@ export interface BackoffOptions {
...
+  /**
+   * Maximum number of retry attempts before giving up.
+   * Set to Infinity (default) for indefinite retries - needed for offline scenarios
+   * where clients may go offline and come back later.
+   *
+   * The retry budget provides protection against retry storms even with infinite retries.
+   */
+  maxRetries?: number
+  /**
+   * Percentage of requests that can be retries (0.1 = 10%)
```
KyleAMathews: /**
   * Percentage of retry attempts that will be allowed through (0.1 = 10%)
   *
   * CRITICAL: This budget is shared across ALL requests made by this fetch client instance.
   * If you have 50 concurrent shape subscriptions all hitting errors, the budget limits
   * the aggregate retry rate across all 50 streams, not per-stream.
   *
   * This is the primary load shedding mechanism. When the server is struggling:
   * - 50 streams fail and want to retry
   * - Budget allows ~5 retries through immediately (10% of 50 retry attempts)
   * - Remaining 45 wait maxDelay (60s) before checking budget again
   *
   * This prevents retry storms from amplifying server load during incidents.
   * Even with infinite retries enabled, at most 10% of retry *attempts* will
   * execute immediately - the rest are throttled.
   *
   * The budget resets every 60 seconds, so temporary error spikes don't
   * permanently block retries.
   *
   * If you need independent budgets, create separate fetch client instances.
   */
  retryBudgetPercent?: number

packages/typescript-client/src/fetch.ts
```diff
@@ -62,7 +115,11 @@ export function createFetchWithBackoff(
     while (true) {
       try {
         const result = await fetchClient(...args)
-        if (result.ok) return result
+        if (result.ok) {
+          // Reset backoff on successful request
+          delay = initialDelay
```
msfstef: this doesn't make sense to me, the `delay` variable is scoped per request so there's no reason to reset it?

packages/typescript-client/src/fetch.ts
```diff
@@ -51,7 +73,38 @@ export function createFetchWithBackoff(
     multiplier,
     debug = false,
     onFailedAttempt,
+    maxRetries = Infinity,
+    retryBudgetPercent = 0.1,
   } = backoffOptions
+
+  // Retry budget tracking (closure-scoped)
+  // Resets every minute to prevent retry storms
+  let totalRequests = 0
```
msfstef: this does not make sense in the way that we use `createFetchWithBackoff` - we create an instance of this fetcher for every instance of `ShapeStream` so there's no global tracking of retries - this will only track retries per individual shape stream so all of this budget code does not make sense to me

packages/typescript-client/src/fetch.ts
```diff
@@ -51,7 +73,38 @@ export function createFetchWithBackoff(
...
+
+  function checkRetryBudget(percent: number): boolean {
+    const now = Date.now()
+    if (now > budgetResetTime) {
+      totalRequests = 0
+      totalRetries = 0
+      budgetResetTime = now + 60_000
+    }
+
+    totalRequests++
```
msfstef: why are we incrementing `totalRequests` in a check function? it feels counterintuitive as a check might be done multiple times

packages/typescript-client/src/fetch.ts
```diff
@@ -51,7 +73,38 @@ export function createFetchWithBackoff(
...
+
+    // Allow retries for first 10 requests to avoid cold start issues
+    if (totalRequests < 10) return true
+
+    const currentRetryRate = totalRetries / totalRequests
+    const hasCapacity = currentRetryRate < percent
+
+    if (hasCapacity) {
+      totalRetries++
```
msfstef: incrementing total retries in a check seems counterintuitive - the check should return the check and the retrier should increase the retries

packages/typescript-client/src/fetch.ts
```diff
@@ -80,17 +137,77 @@ export function createFetchWithBackoff(
           // Any client errors cannot be backed off on, leave it to the caller to handle.
           throw e
         } else {
-          // Exponentially backoff on errors.
-          // Wait for the current delay duration
-          await new Promise((resolve) => setTimeout(resolve, delay))
+          // Check retry budget and max retries
+          attempt++
+          if (attempt >= maxRetries) {
```
msfstef: this means that if you set `maxRetries` to 1, since we're incrementing the attempt before this check, it will immediately match and throw, thus doing 0 retries. Off by one error.

packages/typescript-client/src/fetch.ts
```diff
@@ -80,17 +137,77 @@ export function createFetchWithBackoff(
...
+              console.log(
+                `Retry budget exhausted (attempt ${attempt}), backing off`
+              )
+            }
+            // Wait for maxDelay before checking budget again
+            // This prevents tight retry loops when budget is exhausted
+            await new Promise((resolve) => setTimeout(resolve, maxDelay))
+            // Don't throw - continue retrying after the wait
+            // This allows offline clients to eventually reconnect
+            continue
```
msfstef: nit: `continue` in a large while loop complicates reading the code significantly IMO - abstracting out utilities like getting the wait timeout and then doing these checks at the end to avoid continue and only have a single await would be better

packages/typescript-client/src/fetch.ts
```diff
@@ -80,17 +137,77 @@ export function createFetchWithBackoff(
...
+          }
+
+          // Calculate wait time honoring server-driven backoff as a floor
+          // Precedence: max(serverMinimum, min(clientMaxDelay, backoffWithJitter))
+
+          // 1. Parse server-provided Retry-After (if present)
+          let serverMinimumMs = 0
```
msfstef: this looks like it would be clearer if it lived in a dedicated parser utility and tested in isolation

packages/typescript-client/src/fetch.ts
```diff
@@ -80,17 +137,77 @@ export function createFetchWithBackoff(
...
+                  // Handle clock skew: clamp to non-negative, cap at reasonable max
+                  const deltaMs = retryDate - Date.now()
+                  serverMinimumMs = Math.max(0, Math.min(deltaMs, 3600_000)) // Cap at 1 hour
+                }
+              }
+            }
+          }
+
+          // 2. Calculate client backoff with full jitter
+          const jitter = Math.random() * delay
```
msfstef: 100% jitter feels like too much? it means that if the delay was determined to be 20 seconds it could delay anywhere between 0 and 20 seconds - jitter should be a jitter, small noise variation from the overall delay

packages/typescript-client/src/fetch.ts
msfstef: The fetch client is not shared between shapes.

msfstef (changes requested):
I think this fundamentally doesn't do what it is supposed to do - the `createFetchWithBackoff` is called per shape stream and thus all the "global retry budget" logic is not really doing anything.

And if we _do_ move to global retry budget I would advise at least scoping it per origin, although ultimately if we go via a proxy we don't know where it could be sending it.

I would prefer if the global retry budget discussion was a separate PR and instead this only added max delay and respecting the `Retry-After` headers

packages/typescript-client/src/fetch.ts
KyleAMathews: This is based on this article from AWS where they show "full jitter" significantly speeds up recovery https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

KyleAMathews:
I'm fine with removing the global retry budget for now. Faster backoff and respecting the retry-after header is higher priority.

packages/typescript-client/src/fetch.ts
msfstef: interesting - might be worth linking in the code as it's definitely counter intuitive

PR #3358, opened 2025-10-30
Fix multiple renders from cached up-to-date messages on page refresh

KyleAMathews:
When a shape receives multiple updates within the HTTP cache window (60s), each update ends with an up-to-date control message that gets cached. On page refresh, these cached responses replay rapidly, causing multiple renders. This change implements cursor-based detection to suppress cached up-to-date notifications until a fresh response (with a new cursor) arrives from the server, ensuring only one render occurs.

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -442,7 +442,7 @@ export class ShapeStream<T extends Row<unknown> = Row>
   #activeSnapshotRequests = 0 // counter for concurrent snapshot requests
   #midStreamPromise?: Promise<void>
   #midStreamPromiseResolver?: () => void
-  #replayMode: false | string = false // False or last seen cursor when replaying cached responses
+  #lastSeenCursor?: string // Last seen cursor from previous session (used to detect cached responses)
   #currentFetchUrl?: URL // Current fetch URL for computing shape key
   #lastSseConnectionStartTime?: number
   #minSseConnectionDuration = 1000 // Minimum expected SSE connection duration (1 second)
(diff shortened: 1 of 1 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -441,6 +442,8 @@ export class ShapeStream<T extends Row<unknown> = Row>
   #activeSnapshotRequests = 0 // counter for concurrent snapshot requests
   #midStreamPromise?: Promise<void>
   #midStreamPromiseResolver?: () => void
+  #replayMode: false | string = false // False or last seen cursor when replaying cached responses
```
kevin-dp: nit: this variable serves 2 purposes: storing the last seen cursor and indicating whether we're in replay mode or not. I'd go for `#lastSeenCursor: string | undefined` and then we can derive replay mode, e.g. `get #replayMode() = #lastSeenCursor !== undefined`.

packages/typescript-client/src/client.ts
```diff
@@ -834,6 +837,35 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
         this.#midStreamPromiseResolver?.()
+
+        // Check if we should suppress this up-to-date notification
+        // to prevent multiple renders from cached responses
+        if (this.#replayMode && !isSseMessage) {
+          // We're in replay mode (replaying cached responses during initial sync).
+          // Check if the cursor has changed - cursors are time-based and always
+          // increment, so a new cursor means fresh data from the server.
+          const currentCursor = this.#liveCacheBuster
+          const lastSeenCursor =
```
kevin-dp: Why do we need this logic? The if test on L843 `if (this.#replayMode && ...)` checks the replay mode, so if we're here, the #replayMode must be a string. So the ternary seems unnecessary and the `lastSeenCursor` check in in the if test on L851 seems unnecessary too.

PR #3359, opened 2025-10-30
feat: Ensure `PublicationManager` handles relations individually and concurrently with requests

msfstef:
This work was prompted by timeouts and errors observed on stacks with a high number of relations involved.

As soon as the sync service was up, many requests would come in for all relations, or a recovery would be attempted, and we would attempt to configure all relations in a single transaction.

That would occasionally time out, but also we would often see some tables being slow to update because e.g. one particular table was holding the lock perhaps due to db administration occurring, leading to _no_ tables being configured.

In this PR I am converting PublicationManager to a supervisor of two processes, a `RelationTracker` and a `Configurator`.

## `RelationTracker`

Responsible for reference counting and request handling. This never blocks and is always able to reply to shapes wanting to be registered or deregistered from the publication.

It keeps track of `prepared_relation_filters`, `submitted_relation_filters`, `committed_relation_filters`, which represent the three possible states of a relation.

When a relation needs to be in the publication, it will be in the `prepared` set due to the reference counting logic.

If the `prepared` set differs from the `submitted` or `committed` set, then a publication update is requested asynchronously from the `Configurator`. As soon as the update is requested, the `submitted` set is set to be the `prepared` set.

As the `Configurator` configures relations individually, it sends back results of whether it succeeds or fails. Any relation that has successfully configured gets added to the `committed` set.

The `RelationTracker` will immediately reply to requests to add a relation when the relevant relation is in both the `submitted` and `committed` set - as in this case we know that the relation has both been committed and is maintained in any current "in flight" requests to update it.

## `Configurator`

I have essentially destructured the logic in the `Electric.Postgres.Configuration` module so that we run each relation individually in the `Configurator` process.

We first check the publication status, returning any errors or relevant states like generated column support back to `RelationTracker`.

Then we determine "relation actions" by listing relations in the publication, replica identities, any invalid relations, etc.

Finally we schedule each action to run via `handle_continue` clauses terminating when actions are done. We still don't process any messages while this is happening, but this process only does the publication update so it should not be a problem.

## `PublicationManager`

This is now a supervisor with a `:one_for_one` strategy - if the `Configurator` dies, we can just bring it back up again as it holds no important state other than caching some properties of the publication and it pulls the filters from the `RelationTracker`. The `RelationTracker` holds the reference counting but it restores it from the in memory shape status cache, so in case of a restart it will get back to where it is, and it can handle updates it does not know about from the configurator.

Important to note that I've also reinstated all shapes asking the publication to be added regardless of whether they have a snapshot as we are not lazily loading shapes and it is not longer a problem doing this.

### DIFF
--- a/packages/sync-service/test/electric/replication/publication_manager_test.exs
+++ b/packages/sync-service/test/electric/replication/publication_manager_test.exs
@@ -497,6 +497,47 @@ defmodule Electric.Replication.PublicationManagerTest do
     end
   end
 
+  describe "component restarts" do
+    test "handles relation tracker restart", ctx do
+      notify_alter_queries()
+      shape = generate_shape(ctx.relation_with_oid, @where_clause_1)
+
+      {:ok, shape_handle} = Electric.ShapeCache.ShapeStatus.add_shape(ctx.stack_id, shape)
+
+      run_async(fn ->
+        assert_raise RuntimeError, fn ->
+          PublicationManager.add_shape(shape_handle, shape, ctx.pub_mgr_opts)
+        end
+      end)
+
+      relation_tracker_name = PublicationManager.RelationTracker.name(ctx.stack_id)
+
+      receive do
+        {:alter_publication, _, _} -> GenServer.stop(relation_tracker_name)
+      end
+
+      assert_pub_tables(ctx, [ctx.relation])
+
+      # after restart, the publication manager should repopulate the publication
+      PublicationManager.remove_shape(shape_handle, ctx.pub_mgr_opts)
+      assert_pub_tables(ctx, [])
+    end
+
+    @tag update_debounce_timeout: 100
+    test "handles configurator restart", ctx do
+      notify_alter_queries()
+      shape = generate_shape(ctx.relation_with_oid, @where_clause_1)
+
+      configurator_name = PublicationManager.Configurator.name(ctx.stack_id)
+
+      assert :ok = PublicationManager.add_shape(@shape_handle_1, shape, ctx.pub_mgr_opts)
+      assert_pub_tables(ctx, [ctx.relation])
+      run_async(fn -> PublicationManager.remove_shape(@shape_handle_1, ctx.pub_mgr_opts) end)
+      GenServer.stop(configurator_name)
+      assert_pub_tables(ctx, [])
+    end
+  end
+
   defp assert_pub_tables(ctx, expected_tables, timeout \\ 500) do
     start_time = :erlang.monotonic_time(:millisecond)
     pub_tables = fetch_pub_tables(ctx)


packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -1,86 +1,22 @@
 defmodule Electric.Replication.PublicationManager do
```
robacourt, line 1: Supervisors should be named as such in my opinion

packages/sync-service/lib/electric/replication/publication_manager/configurator.ex
```diff
@@ -0,0 +1,400 @@
...
+  def handle_info(:debounced_update_publication, state) do
+    {
+      :noreply,
+      %{state | scheduled_filters: nil, scheduled_update_ref: nil},
+      {:continue, {:check_publication, state.scheduled_filters}}
+    }
+  end
+
+  @impl true
+  def handle_continue({:check_publication, filters}, state) do
```
magnetised, line 112: This feels like an over-use of continues. A list of function calls within a `with` would be much easier to read and have essentially the same behaviour.

packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
```diff
@@ -0,0 +1,498 @@
...
+    )
+  end
+
+  @impl true
+  def handle_call({:add_shape, shape_handle, publication_filter}, from, state) do
+    {oid_rel, with_gen_cols} = publication_filter
+
+    # if the relation is already committed AND part of the last made
+    # update submission, we can consider it ready
+    relation_ready =
```
magnetised:
```suggestion
    relation_ready? =
```

packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
```diff
@@ -0,0 +1,498 @@
...
+          :reply,
+          {:error,
+           Electric.DbConfigurationError.publication_missing_generated_columns(
+             state.publication_name
+           )},
+          state,
+          state.publication_refresh_period
+        }
+
+      relation_ready ->
```
magnetised:
```suggestion
      relation_ready? ->
```

packages/sync-service/lib/electric/replication/publication_manager.ex
```diff
@@ -91,658 +37,41 @@ defmodule Electric.Replication.PublicationManager do
...
+    children = [
+      {__MODULE__.RelationTracker, opts},
+      {__MODULE__.Configurator, opts}
+    ]
 
-    %Electric.DbConfigurationError{
-      type: :schema_changed,
-      message: "Database table #{table} has been dropped or renamed"
-    }
+    Supervisor.init(children, strategy: :rest_for_one)
```
robacourt: So if the Configurator dies, the RelationTracker will not be restarted which if it's mid configuring then RelationTracker will not retry calling the Configurator until it polls, so waiters may be waiting the poll interval. Is that ok? Making this a `:one_for_all` would mean the waiters crash instead, which may be more explicit (crashes are easier to debug than hanging) but then maybe waiting the poll time is fine and it's better to have resilience

robacourt (approved):
Great work! Just had a couple of comments

packages/sync-service/lib/electric/replication/publication_manager.ex
msfstef: I've followed the pattern used by @magnetised in `Electric.Shapes.Monitor` and sort of in `PureFileStorage` with `stack_start_link` - I liked the idea that in this case the clients of PubMan are unaware of the supervision tree under the hood given that they will never call any of the internal processes directly

packages/sync-service/lib/electric/replication/publication_manager.ex
msfstef: Good point - a pull based approach would also work, and the `Configurator` asks for filters the moment it initialises. It's better to update more than to update less.

packages/sync-service/lib/electric/replication/publication_manager/configurator.ex
msfstef: I get what you mean but IMO because `check_publication` has several side effects (notifying the relation tracker on success, notifying for errors) which I want to occur as soon as possible, I think continues "read" better

Basically any time we have to use a connection with postgres, do something, and notify with results, I assign it its own continue to handle that. If side effects were not there and it was purely a functional piping of information a with would read better.

I'd prefer not to hide the side effects (the various `notify` calls) inside private function calls further down this file as it makes it hard to follow what is actually happening - any notification call occurs in a `handle_continue` body.

packages/sync-service/lib/electric/replication/publication_manager.ex
robacourt: Yeah, I don't like it in `Electric.Shapes.Monitor` either 😄 Keep the supervisor separate from the interface and you can get the encapsulation you are after. `PublicationManager` should be the context module with the interface and you can have a separate `PublicationManager.Supervisor`. 

`PureFileStorage` is vaguely forgivable because it's doing it to be behind a common storage interface, but there's probably some better ways here too.

packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
robacourt: +1

packages/sync-service/lib/electric/replication/publication_manager.ex
msfstef: Yeah as I typed this I thought of what you were describing and it makes sense - I'll move the supervisor itself as a supervisor and keep the interface there!

packages/sync-service/lib/electric/replication/publication_manager.ex
msfstef: On initialisation, `Configurator` now calls `fetch_current_filters` in a continue clause which will run the first update.

The `RelationTracker` also has an iniit continue clause that recovers its filters but I've removed the scheduling of an update at the end of the recovery since `Configurator` will pull it. A test covers the recovery process.

Technically if the relation tracker also scheduled an update after recovering (and even if it did not honestly) we could actually have a `:one_for_one` strategy

robacourt: > On initialisation, Configurator now calls fetch_current_filters in a continue clause which will run the first update.

Fab, that's very nice

> Technically if the relation tracker also scheduled an update after recovering (and even if it did not honestly) we could actually have a :one_for_one strategy

Isn't the issue that the Configurator could message the RelationTracker about a filter that's not in it's prepared set? That sounds like it might cause issues.

msfstef: We discussed in Disc as well but the relation tracker is resilient to receiving updates for relations it is not tracking - it might over-update in case of a crash but that works perfectly fine.

I've converted it to a `:one_for_one` and added a couple of tests where I'm killing components of the system

PR #3363, opened 2025-11-03
Don't reset the offset when recovering from an error

KyleAMathews:
When the `onError` handler returns new params/headers to retry after an error (like a 401), the ShapeStream was calling `#reset()` which reset the offset to `-1`, causing the stream to refetch all data from the beginning. This resulted in duplicate insert operations being emitted to Shape subscribers, causing "already exists" errors in collections.

The fix removes the `#reset()` call during error recovery. Instead, the stream now:
- Updates only the params/headers as needed
- Continues from the current offset (not -1)
- Avoids re-fetching data that was already processed

This matches the expected behavior where the client only refetches from the beginning when the backend explicitly sends a `must-refetch` control message (e.g., on 409 errors when the shape is deprecated).

Added integration test to verify that after error recovery:
- The offset is preserved (not reset to -1)
- No duplicate operations are emitted
- Shape data remains correct

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -511,7 +511,8 @@ export class ShapeStream<T extends Row<unknown> = Row>
       // Check if onError handler wants to retry
       if (this.#onError) {
         const retryOpts = await this.#onError(err as Error)
-        if (typeof retryOpts === `object`) {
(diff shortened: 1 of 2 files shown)


kevin-dp (approved):
Perfect! You'll need to run prettier though in order for CI to turn green.

PR #3364, opened 2025-11-03
Fix infinite loop in ShapeStream live_sse mode

KyleAMathews:
Fixes an issue where SSE connections that close immediately (due to caching or proxy misconfiguration) would cause an infinite loop of rapid requests.

The fix implements:
1. Connection duration tracking for SSE requests
2. Detection of consecutively short-lived connections (< 1 second)
3. Automatic fallback to long polling after 3 short connections
4. Console warning with configuration guidance
5. Delay between retry attempts to prevent tight loops
6. Reset of fallback state on shape reset

This addresses the issue reported where misconfigured reverse proxies (e.g., Caddy without proper SSE configuration) or cached responses cause the client to repeatedly reconnect without delay.

Related: https://github.com/electric-sql/electric/blob/a5a0443ebce2a5fbf80e670b60f5cd80aff7d85e/packages/typescript-client/src/client.ts#L632

### DIFF
--- a/packages/typescript-client/test/client.test.ts
+++ b/packages/typescript-client/test/client.test.ts
@@ -2072,3 +2072,276 @@ it(
   },
   { timeout: 15000 }
 )
+
+it(
+  `should reset SSE fallback state after shape rotation (409)`,
+  async ({ issuesTableUrl, aborter }) => {
+    let sseRequestCount = 0
+    let shouldReturn409 = false
+    const requestUrls: string[] = []
+
+    // Mock console.warn to capture the fallback warning
+    const originalWarn = console.warn
+    const warnMock = vi.fn()
+    console.warn = warnMock
+
+    try {
+      const fetchClient = async (
+        input: string | URL | Request,
+        init?: RequestInit
+      ) => {
+        const url = input instanceof Request ? input.url : input.toString()
+        requestUrls.push(url)
+
+        const urlObj = new URL(url)
+        const isSSE = urlObj.searchParams.get(`live_sse`) === `true`
+        const hasHandle = urlObj.searchParams.has(`handle`)
+
+        // Return 409 to trigger shape rotation after fallback occurs
+        if (shouldReturn409 && hasHandle) {
+          shouldReturn409 = false // Only return 409 once
+          return new Response(null, {
+            status: 409,
+            statusText: `Conflict`,
+            headers: new Headers({
+              'electric-handle': `new-handle-${Date.now()}`,
+            }),
+          })
+        }
+
+        if (isSSE) {
+          sseRequestCount++
+          // First 3 SSE requests: return short connections to trigger fallback
+          if (sseRequestCount <= 3) {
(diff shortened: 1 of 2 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -916,6 +926,42 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+          this.#consecutiveShortSseConnections >= this.#maxShortSseConnections
+        ) {
+          // Too many short connections - fall back to long polling
+          this.#sseFallbackToLongPolling = true
+          console.warn(
+            `[Electric] SSE connections are closing immediately (possibly due to caching or proxy misconfiguration). ` +
+              `Falling back to long polling. ` +
+              `Please check your proxy/CDN configuration for SSE support. ` +
+              `For proxies: Nginx requires 'X-Accel-Buffering: no', Caddy requires 'flush_interval -1'. ` +
+              `All proxies should set 'Cache-Control: no-cache' and avoid buffering SSE responses.`
```
samwillis: The reverse proxy should not cache, but there can (should?) be cache headers to trigger request collapsing at the CDN.

kevin-dp:
> Automatic fallback to long polling after 3 short connections

I'm not sure about this. If someone configures their app to use SSE, and there is a problem with SSE, then i would expect it to keep trying to connect with SSE (albeit with exponential backoff) rather than switch to long polling which will hide the problem and the dev may think it works fine if they miss the console warning.

packages/typescript-client/src/client.ts
```diff
@@ -916,6 +926,42 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+          )
+        } else {
+          // Add a delay to prevent tight infinite loop while we're still trying SSE
+          await new Promise((resolve) =>
+            setTimeout(
+              resolve,
+              this.#minSseConnectionDuration - connectionDuration
```
kevin-dp: This is a very small delay (< 1 sec). What about `2^retry`, e.g. `2^0 = 1` on first retry, `2^1` on 2nd retry, `2^2` on third retry.

kevin-dp (approved):
LGTM. Left just one nit. Seems that we broke an existing test and the new test also fails. We need to fix that before we can merge.

KyleAMathews:
> If someone configures their app to use SSE, and there is a problem with SSE, then i would expect it to keep trying to connect with SSE (albeit with exponential backoff) rather than switch to long polling which will hide the problem and the dev may think it works fine if they miss the console warning.

They would eventually notice it's broken. And I'm sure they'd appreciate that their app kept working during that time 😆 

I'm normally a big fan of crash only programming but SSE is just an enhancement on top of long-pollling so I think it's perfectly fine falling back when something is off.

packages/typescript-client/src/client.ts
KyleAMathews: good catch, fixed

PR #3366, opened 2025-11-03
fix: Parse Prisma badly formatted Postgres errors into known `DbConnectionError`

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3362

I don't love that we're parsing a failed parsing of a Postgres error, but for the sake of capturing major PG providers this will do.

@balegas I'm not entirely convinced this is an actual fix we should merge - the badly formatted PG code is failing on various levels:

image

### DIFF
--- a/packages/sync-service/lib/electric/db_connection_error.ex
+++ b/packages/sync-service/lib/electric/db_connection_error.ex
@@ -367,8 +367,20 @@ defmodule Electric.DbConnectionError do
   end
 
   # Fix for badly formatted errors from Prisma's managed Postgres
-  def from_error({:badkey, :code, %{message: _message}} = error) do
-    maybe_failed_to_identify_database(error) || unknown_error(error)
+  def from_error({:badkey, :code, %{message: message}} = error) do
+    if Regex.match?(
+         ~r/^Failed to identify your database: Your account has restrictions: [\w]+./,
+         message
+       ) do
+      %DbConnectionError{
+        message: message,
+        type: :endpoint_not_found,
+        original_error: error,
+        retry_may_fix?: false
+      }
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/db_connection_error.ex
```diff
@@ -366,6 +366,11 @@ defmodule Electric.DbConnectionError do
     }
   end
 
+  # Fix for badly formatted errors from Prisma's managed Postgres
+  def from_error({:badkey, :code, %{message: _message}} = error) do
+    maybe_failed_to_identify_database(error) || unknown_error(error)
+  end
```
robacourt:
```suggestion
  def from_error({:badkey, :code, %{message: _message}} = error) do
    if Regex.match?(
         ~r/^Failed to identify your database: Your account has restrictions: [\w]+./,
         message
       ) do
      %DbConnectionError{
        message: message,
        type: :endpoint_not_found,
        original_error: error,
        retry_may_fix?: false
      }
    else
      unknown_error(error)
    end
  end
```

Incredibly small nit, but the function can just be inlined as it's not reused. It just avoids a bit of indirection.

robacourt (approved):
👍

balegas:
> @balegas I'm not entirely convinced this is an actual fix we should merge - the badly formatted PG code is failing on various levels

We discussed 1) can't do anything about library errors, but we can filter them out at Sentry-level; 2) check what is the behavior of the code if we do minimal parsing without looking into the error message.

msfstef:
Will merge this to cover these errors for the time being - our code already retries for unknown errors so we're covered regardless.

PR #3372, opened 2025-11-03
fix: Only include schema in 200 responses

msfstef:
We've seen a large amount of errors on Sentry from the `Api.schema` being called when there is no connection for tons of responses.

Looking at the code it seems we _always_ call `put_schema_header` as part of `put_resp_header` even for errors - and when the error is that the stack is not up and no connection is available, an error from the schema trying to be read is guaranteed to happen.

We need to examine `put_resp_header` being called on _all_ responses as I suspect we are including too many headers on error responses, but for the schema one in particular that requires a live stack and potentially a connection, we should either only include it in 200 responses (what this PR does), or we should just omit it if it is not available.

I decided not to omit it as I am not sure what it is used for and if it is an integral part of the protocol, so for now I am just guarding it to be added on 200 responses only.

### DIFF
--- a/.changeset/unlucky-pants-cheat.md
+++ b/.changeset/unlucky-pants-cheat.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Only include schema in 200 responses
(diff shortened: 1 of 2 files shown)


magnetised (approved):
Agree that this is the right approach -- the schema header is certainly used and since we don't know other peoples client implementations it's definitely safer to keep it in. 

Do we only return 200 statuses?

msfstef:
> Do we only return 200 statuses?

AFAIK if our response contains data, it will be a 200. We occasionally return some 4xx codes but those will not contain data, so I believe that a schema is unnecessary.

PR #3379, opened 2025-11-04
fix: Guard against missing connection pool

msfstef:
Fixes unnecessary error logs for cases like [this Sentry issue](https://electricsql-04.sentry.io/issues/74399805/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream) where the `execute_and_log_errors` fails because of a connection issue - in that case we log a warning. It also handles a missing pool as seen in [this Sentry error](https://electricsql-04.sentry.io/issues/74702354/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream)

Fixes missing connection pool not being handled gracefully in snapshotter, see [this Sentry issue](https://electricsql-04.sentry.io/issues/72247640/?environment=production&project=4508410462404688&query=is%3Aunresolved&referrer=issue-stream)

Ultimately we should handle connection errors and a missing pool uniformly across the sync service but this is an immediate fix and we should refactor it after we discuss and decide on what the uniform approach should be. Tagging @robacourt for the relevant connection/operations split.

### DIFF
--- a/packages/sync-service/test/support/test_utils.ex
+++ b/packages/sync-service/test/support/test_utils.ex
@@ -156,7 +156,8 @@ defmodule Support.TestUtils do
     callback_fun = fn pid -> Repatch.allow(self_pid, pid) end
 
     # The descendant process running module `mod` will look up this callback in its root ancestor and execute it.
-    Process.put(:callback_for_descendant_procs, {mod, callback_fun})
+    old_cbs = Process.get(:callbacks_for_descendant_procs) || []
+    Process.put(:callbacks_for_descendant_procs, [{mod, callback_fun} | old_cbs])
 
     :ok
   end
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -1158,9 +1158,17 @@ defmodule Electric.Connection.Manager do
...
+        Logger.warning("Failed to execute query: #{query}\nError: #{inspect(error)}")
+
       {:error, error} ->
         Logger.error("Failed to execute query: #{query}\nError: #{inspect(error)}")
     end
+  catch
+    :exit, {_, {DBConnection.Holder, :checkout, _}} ->
+      Logger.warning(
+        "Failed to execute query: #{query}\nError: Database connection not available"
+      )
```
robacourt, line 1171: Ultimately I think we should have `query /3` as part of the connection subsystem interface and the exit is handled inside that, but this is good for now

msfstef: Not sure if that would be enough as we use the whole Postgrex interface `query/3` `query!/3` `transaction/2` etc we might have to replicate some parts of it if we want to go with that approach.

robacourt: Yeah, I was thinking of having them all. 

Or another approach is having modules a bit like the Configuration module used to be, no process but does the Postgrex stuff, does the catching and returns a result. The module would be in the connection namespace but called from the shape namespace. Basically the Gateway pattern. Perhaps that would be neater in this instance.

It still doesn't get rid of the code repetition, but we could have a macro for that.

PR #3383, opened 2025-11-04
feat: Always handle `AsyncDeleter` requests to avoid message queue buildup

msfstef:
Followup to https://github.com/electric-sql/electric/pull/3380

The above PR solves the issue of the file server being overloaded and all shapes being blocked on removing themselves.

This PR addresses the buildup of remove messages on the AsyncDeleter itself.

Whenever the removal is triggered, which is an `rm_rf` of an arbitrarily large directory that runs recursively, no messages to the AsyncReceiver can be processed in the meantime, building up the message queue and bad things can happen (e.g. calling processes get their reductions penalised by the beam)

In this PR I am splitting the request handling and file removal into two separate concurrent actors.

In [this commit](https://github.com/electric-sql/electric/commit/7783f38402c4f1aed2c1ef6962cc76188f61ebe3) I followed a much more verbose approach, with a dedicated task supervisor under an overall async deleter supervisor etc etc - I think this is overkill and unnecessary so I reverted it.

The current approach just runs the `rm_rf` in a task - whenever the task is done, we schedule a new cleanup. If there's nothing to cleanup, no task is started and no new cleanup is scheduled until more requests come in.

This ensures that the process can always handle requests and message queues don't build up, and deletion occurs asynchronously.

### DIFF
--- a/packages/sync-service/lib/electric/async_deleter.ex
+++ b/packages/sync-service/lib/electric/async_deleter.ex
@@ -14,7 +14,6 @@ defmodule Electric.AsyncDeleter do
        first queued delete before the batch is removed. Defaults to 10000 ms.
   """
 
-
   use GenServer
   require Logger
 


packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -90,8 +92,8 @@ defmodule Electric.AsyncDeleter do
     {:noreply, do_cleanup(state)}
```
alco, line 92: This `handle_continue(:initial_cleanup, state)` clause must be leftover from an earlier version of the code? It's always a noop since it's called right after `init()` when the list of pending deletes is always empty.

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -118,8 +120,53 @@ defmodule Electric.AsyncDeleter do
   defp random_suffix, do: System.unique_integer([:positive]) |> to_string()
 
   @impl true
-  def handle_info(:perform_delete, state) do
-    {:noreply, do_cleanup(state)}
+  def handle_info(:perform_delete, %{cleanup_task: nil} = state) do
```
alco, line 123: This match should be turned into an `if` inside the function body. Or you need to add a noop clause for handling the `:perform_delete` message when there's already a cleanup task running.

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -118,8 +120,53 @@ defmodule Electric.AsyncDeleter do
...
+    Logger.warning(
+      "AsyncDeleter: failed to delete #{length(state.pending)} paths " <>
+        "for stack #{state.stack_id} after #{duration}ms with reason: #{inspect(reason)}" <>
+        " - will retry on next scheduled cleanup."
+    )
+
+    {:noreply,
+     %{
+       state
+       | pending: Enum.concat(state.in_progress, state.pending),
```
alco: nit: Could be a simple `++` since both fields are known to be lists.

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -140,29 +187,38 @@ defmodule Electric.AsyncDeleter do
...
-    end
+  defp do_cleanup(%{pending: []} = state), do: state
 
-    %{state | pending: [], timer_ref: nil}
+  defp do_cleanup(state) do
+    start_time = System.monotonic_time(:millisecond)
+
+    task =
+      Task.async(fn ->
+        Process.set_label({:async_deleter_cleanup_task, state.stack_id})
```
alco: It's a nitpick in this case because AsyncDeleter's state is small, but BEAM will copy the whole `state` into the task process only to access the `stack_id` field. Better assign `state.stack_id` to a local variable before starting the task and reference that variable inside the task function.

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -140,29 +187,38 @@ defmodule Electric.AsyncDeleter do
...
+        Logger.debug("AsyncDeleter: Cleaning trash directory #{inspect(trash_dir)}")
+
+        try do
+          clean_dir!(trash_dir)
+        rescue
+          e -> Logger.warning("AsyncDeleter: rm_rf failed: #{inspect(e)}")
+        end
+      end)
+
+    Process.unlink(task.pid)
```
alco, line 220: This looks a bit off. We start an async task and unlink it from the deleter process. But then we have explicit handling of task shutdown in the `terminate()` callback. It's like we reimplement process linking halfway.

On one hand, we could remove the task shutdown call from `terminate()` and let the task run to completion, regardless of whether AsyncDeleter is alive or not. Since we want those files deleted, we could give it the best chance we can even if the stack starts shutting down in parallel.

On the other hand, if we really want to terminate the task together with AsyncDeleter but shield the latter from any errors raised in the task, it seems more in line with OTP principles to make AsyncDeleter trap exits and write a `handle_info({:EXIT, ...` with two clauses: one matches on the task pid and does nothing, the other would exit the AsyncDeleter process with whatever reason the incoming `:EXIT` message contains (basically emulating the ordinary exit signal propagation).

I'm leaning more towards the first option.

packages/sync-service/lib/electric/async_deleter.ex
msfstef: good point - I think we should still ensure a cleanup on startup so I'll fix this

packages/sync-service/lib/electric/async_deleter.ex
msfstef: The flow of the code ensures that we should never receive a `:perform_delete` message when a cleanup task is running - a timer is only started if there's not task running. I had intentionally left it like this to let it crash in case it runs in unexpected ways

packages/sync-service/lib/electric/async_deleter.ex
msfstef: Ah nice catch and good thing to keep in mind going forward - thank you

packages/sync-service/lib/electric/async_deleter.ex
msfstef: I would much prefer the first option.

The reason I added this is that the task being linked (and not supervised at the AsyncDeleter level) means that:

1. If AsyncDeleter dies and gets restarted, it might trigger a new cleanup task that might, if even for a short time, run concurrently with the previous one, as the link does not guarantee that it will be dead by the time AsyncDeleter is brought back up. If we don't see any issues with concurrent `rm_rf` operations running on the trash dir then this is fine, perhaps I'm being _too_ conservative.
2. If the task dies, it will take down the `AsyncDeleter` process with it and it feels unnecessary, since all I would want to do is reschedule it (as I do in the `:DOWN` message handling for non-normal exits) - in this case I would need to trap exits to handle them and it loses the simplicity of the first option you describe

Do you think we should perhaps just let the task take down the AsyncDeleter process if it exits unexpectedly? There's very little that can go wrong but we are not catching/guarding against _all_ possible failures in the task.

packages/sync-service/lib/electric/async_deleter.ex
msfstef: I fixed this by populating `pending` with a `File.ls` at the start, so we have log completeness as well.

packages/sync-service/lib/electric/async_deleter.ex
alco: I thought that could be the case. Just noting here that `do_cleanup` is also called outside of the timer-driven logic (the one after the `init()`). So, even though unlikely, it's not theoretically implausible to receive this message from the timer when there's already a cleanup task running. Future changes to this module may be less diligent with regards to maintaining this invariant that `:perform_delete` only arrives when `cleanup_task` is `nil`.

I just think making this message handler's do nothing in case there is already a task running is making the code a bit more idempotent. No longer rely on the strict sequencing between different callback functions and instead realy on the fact that if there's a cleanup task running, once it's done we know for sure it'll schedule another cleanup attempt.

packages/sync-service/lib/electric/async_deleter.ex
alco: I haven't actually thought about the case of two async tasks concurrently running `rm_rf` on the same dir. I think that makes the current implementation that uses an explicit call to `Task.shutdown` perfectly reasonable. Maybe just add a code comment above that line to explain this possibility.

packages/sync-service/lib/electric/async_deleter.ex
msfstef: I'm happy to make that change and I see your point - although the flipside of that is that with code changes this idempotency might turn into a silent swallowing of cleanup not running because the cleanup task is never removed from the state, and all subsequent actions never perform a cleanup - that was my reason for a stricter assertion of the code flow

alco (approved):
:+1:

packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -13,14 +13,19 @@ defmodule Electric.AsyncDeleter do
     * `:cleanup_interval_ms` - interval in milliseconds after the
        first queued delete before the batch is removed. Defaults to 10000 ms.
   """
+
+  require Logger
```
magnetised: dupe
```suggestion
```

magnetised (approved):
Lovely.

PR #3385, opened 2025-11-04
feat(sync-service): Speed up shape meta data removal

robacourt:
For 50K shapes and concurrent read and writes of the ShapeStatus ETS tables, `ShapeStatus.remove_shape` takes 50ms on Ampere, 23ms locally. This only gets slower as the number of shapes increases. This PR speeds up that process by 450% to 0.05ms. 

It does mean that the metadata delete is not atomic, but as far as I can tell there are no race conditions that this introduces.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -202,11 +202,8 @@ defmodule Electric.ShapeCache.ShapeStatus do
       # is ever matched to a handle with incomplete information, since deleting with
       # select_delete can lead to inconsistent state
       :ets.delete(meta_table, {@shape_hash_lookup, Shape.comparable(shape)})
-
       :ets.delete(meta_table, {@shape_meta_data, shape_handle})
-
       :ets.delete(meta_table, {@shape_storage_state_backup, shape_handle})
-
       :ets.delete(meta_table, {@snapshot_started, shape_handle})
 
       Enum.each(Shape.list_relations(shape), fn {oid, _} ->


packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -203,16 +203,21 @@ defmodule Electric.ShapeCache.ShapeStatus do
       # select_delete can lead to inconsistent state
       :ets.delete(meta_table, {@shape_hash_lookup, Shape.comparable(shape)})
```
msfstef, line 204: as long as this is the first thing that gets deleted, the rest of the entries are basically orphaned since there is no way for someone to go from a shape definition to this shape handle - and a new shape will be created for the same shape definition since this entry is no longer here.

it was already not atomic as we were first deleting this lookup, and then the rest atomically - which means that if we crashed between deleting the lookup and the rest we would still have orphaned data, but at least the orphaned data would be "consistent" but unreachable.

I'm not against this approach but worth noting that we might need to somehow measure and/or deal with orphaned metadata?

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -203,16 +203,21 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
 
+      :ets.select_delete(meta_table, [
```
msfstef: how did this speed things up? was it that it's no longer grabbing multiple locks over large parts of the table in order to make this operation atomic?

at some point we discussed doing `:ets.delete` for every entry, since this is looking basically like we are doing that perhaps that might be even faster?

msfstef (approved):
wow crazy that this speeds things up so much - I assume it's because trying to delete so many entries (which are spread out all over the table due to their keys not being close to each other) ends up creating a lot of contention?

and what is the reason not to move to an explicit `:ets.delete` of every entry as we had discussed in the past at some point?

packages/sync-service/lib/electric/shape_cache/shape_status.ex
robacourt: I've updated it to use `:ets.delete`. It's no faster but looks neater.

packages/sync-service/lib/electric/shape_cache/shape_status.ex
robacourt: I'm can't actually thing of a scenario where we get orphaned data without having a bug. Maybe if the calling process was brutally killed, but that would only be at shutdown surely. And if it's a bug we should fix the bug. I rather not have a orphaned data monitor to cope with us not fixing bugs.

packages/sync-service/lib/electric/shape_cache/shape_status.ex
msfstef: I think ultimately `:ets.delete/2` will be cleaner in the underlying code as well: https://github.com/erlang/otp/blob/bc3132beee5483db4a83a39d9987bb9efe2429ac/erts/emulator/beam/erl_db.c#L3251-L3273

It uses the key for a direct lookup and erase, whereas `:ets.select_delete/2` will actually compile the match specification and run it:
https://github.com/erlang/otp/blob/bc3132beee5483db4a83a39d9987bb9efe2429ac/erts/emulator/beam/erl_db.c#L3353-L3398

following the code is doing my head in but ultimately select will have to run the match spec against records, and since doing them all together is like an OR semantic of "this record matches any of these" I suspect it ends up being less than optimal

msfstef:
From [ETS docs](https://www.erlang.org/doc/apps/stdlib/ets.html#table-traversal) on table traversal:

>Traversals using match and select functions may not need to scan the entire table depending on how the key is specified. A match pattern **with a fully bound key (without any match variables) will optimize the operation to a single key lookup** without any table traversal at all. For ordered_set a partially bound key will limit the traversal to only scan a subset of the table based on term order. A partially bound key is either a list or a tuple with a prefix that is fully bound.

>**with a fully bound key (without any match variables) will optimize the operation to a single key lookup**

Given that we're seeing the same performance as with `:ets.delete` I suspect that this optimization was happening when we split the select, but not when we had it altogether. So we were likely doing full table traversals.

PR #3386, opened 2025-11-04
docs: document how to enable live SSE mode.

thruflo:
Document SSE support for live mode in:
- HTTP API page: Added section explaining SSE as alternative to long polling
- TypeScript client page: Documented liveSse option with usage examples
- OpenAPI spec: Added live_sse and experimental_live_sse parameters

Key additions:
- SSE message format and protocol details
- Benefits comparison: SSE vs long polling
- Proxy configuration requirements (Nginx, Caddy, Apache)
- Automatic fallback behavior in TypeScript client
- Keep-alive mechanism (21-second intervals)
- SSE response examples in text/event-stream format

SSE provides more efficient real-time updates with a persistent connection that reduces request overhead and latency compared to long polling.

### DIFF
--- a/website/docs/api/clients/typescript.md
+++ b/website/docs/api/clients/typescript.md
@@ -148,8 +148,8 @@ stream.subscribe((messages) => {
 ```
 
 **Benefits of SSE:**
-- Single persistent connection for all live updates - the client doesn't need to reconnect after each message
-- Lower latency when messages arrive frequently (equivalent to long polling if messages are 100ms+ apart)
+- Fewer HTTP requests - the client doesn't need to reconnect after each message
+- Lower latency for small messages arriving frequently (<100ms apart, such as token streaming)
 - Reduced bandwidth (no request overhead per update)
 - More efficient for frequent updates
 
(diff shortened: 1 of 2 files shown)


website/docs/api/clients/typescript.md
```diff
@@ -127,6 +127,41 @@ stream.subscribe((messages) => {
...
+})
+
+stream.subscribe((messages) => {
+  // Receive real-time updates via SSE
+})
+```
+
+**Benefits of SSE:**
+- Single persistent connection for all live updates
+- Lower latency (server pushes changes immediately)
```
KyleAMathews: I'd put this as "the client doesn't need to reconnect after each message"

KyleAMathews: also it's lower latency for many fast messages. It's the same latency if messages are 100ms+ apart

thruflo: Thanks -- updated

KyleAMathews (approved):
:shipit:

PR #3389, opened 2025-11-05
chore(sync-service): Reduce consumer memory usage

magnetised:
Reduces per-shape memory usage: 

1. Spawns only a single process for each shape instead of 2 by removing the consumer supervisor
2. Reduces the config for each consumer process down to the absolute minimum to reduce the redundant duplication of per-stack config between consumers in the supervisor.


packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -387,29 +383,24 @@ defmodule Electric.ShapeCache do
...
+             registry: state.registry,
+             db_pool: state.db_pool,
+             hibernate_after: state.shape_hibernate_after,
+             otel_ctx: otel_ctx,
+             snapshot_timeout_to_first_data: state.snapshot_timeout_to_first_data,
+             action: action
+           }
          ) do
-      {:ok, supervisor_pid} ->
-        # TODO: we will just get the consumer pid when we simplify the consumer process tree
```
msfstef, line 406: gotta love it when a plan comes together

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -104,10 +118,17 @@ defmodule Electric.Shapes.Consumer do
...
   def handle_continue({:init_consumer, action}, state) do
     %{
-      storage: storage,
+      storage: stack_storage,
       shape_status_mod: shape_status_mod
     } = state
 
+    storage = ShapeCache.Storage.for_shape(state.shape_handle, stack_storage)
+
+    case ShapeCache.Storage.start_link(storage) do
```
msfstef, line 130: nit: I think this is removed in another PR, cause this not being supervised with the consumer is sus - leaving comment for confirmation feel free to resolve

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -136,10 +157,32 @@ defmodule Electric.Shapes.Consumer do
     if all_materializers_alive?(state) && subscribe(state, action) do
       Logger.debug("Writer for #{state.shape_handle} initialized")
 
-      Snapshotter.start_snapshot(state.stack_id, state.shape_handle)
+      # We start the snapshotter even if there's a snapshot because it also performs the call
+      # to PublicationManager.add_shape/3. We *could* do that call here and avoid spawning a
+      # process if the shape already has a snapshot but the current semantics rely on being able
+      # to wait for the snapshot asynchronously and if we called publication manager here it would
+      # block and prevent await_snapshot_start calls from adding snapshot subscribers.
```
msfstef, line 170: thank you for the comment - and let's improve this in a separate PR for ease of reviewability!

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -285,9 +331,10 @@ defmodule Electric.Shapes.Consumer do
   end
 
   # We're trapping exists so that `terminate` is called to clean up the writer,
-  # otherwise we respect the OTP exit protocol.
-  def handle_info({:EXIT, _from, reason}, state) do
-    Logger.debug("Caught EXIT: #{inspect(reason)}")
+  # otherwise we respect the OTP exit protocol. Since nothing is linked to the consumer
+  # we shouldn't see this...
```
msfstef, line 334: if we're not trapping exits, and shouldn't see this, why not remove this clause altogether?

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -7,7 +7,7 @@ defmodule Electric.Shapes.Consumer.Materializer do
 
   # NOTES:
   # - Consumer does txn buffering until pg snapshot is known
-  use GenServer
+  use GenServer, restart: :transient
```
msfstef: to confirm - we're monitoring the materializer in the consumers that use it, and would crash them if this goes down - but until now it lived in a supervisor that would get brought down by the significant process dying and thus wouldn't get revived whereas now it lives in the dynamic supervisor and should die for good?

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -91,10 +91,13 @@ defmodule Electric.Shapes.Consumer.Materializer do
   end
 
   def handle_continue({:start_materializer, storage}, state) do
-    _ = Consumer.await_snapshot_start(state)
-    Consumer.subscribe_materializer(state)
+    %{stack_id: stack_id, shape_handle: shape_handle} = state
 
-    Process.monitor(GenServer.whereis(Consumer.name(state)),
+    :started = Consumer.await_snapshot_start(stack_id, shape_handle)
```
msfstef: I've opened a PR that would return a failure to this if it waits for longer than some time for the snapshot data to start https://github.com/electric-sql/electric/pull/3391 - perhaps I have to rethink it a bit, but noting that if for whatever reason this times out, like running into the 15 sec call timeout which we see a lot of requests run into, this would fall over despite the shape still trying to create the snapshot, so we should at least put an infinite timeout here?

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -180,16 +183,23 @@ defmodule Electric.Shapes.Consumer.Materializer do
...
     {:noreply, state}
   end
 
   def handle_info({{:consumer_down, _}, _ref, :process, _pid, _reason}, state) do
-    {:noreply, state}
+    for pid <- state.subscribers do
+      send(pid, {:materializer_shape_invalidated, state.shape_handle})
+    end
+
+    {:stop, :shutdown, state}
```
msfstef, line 208: in the original code, if the reason is `{:shutdown, :cleanup}` we don't shutdown the process here - was this change in behaviour intended?

robacourt:
> As part of this the existing manual addition of a consumer to the ConsumerRegistry ets is replaced with a standard erlang name registration into ConsumerRegistry because it removes race conditions but we still manage the de-registration via the ShapeLogCollector so that the process is registered (even if dead) until the SLC receives the remove shape message. This keeps the existing flow and deals with race conditions when removing shapes.

What race conditions does this remove?

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: we are trapping exits -- we still need the terminate/2 callback

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: yeah. the only storage that has any functionality in start_link is in-memory, so this won't last long. I have an open issue to remove Storage.start_link

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
magnetised: In the original version the materializer is supervised alongside the consumer, which is marked as significant, so if the consumer exits for any reason then the materializer would be automatically shutdown too.

So I think this code is wrong -- the materializer should shutdown if the consumer shuts down (except in the future when consumers shutdown through inactivity).

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: I think now you've polished the publication manager add_shape always returns immediately - is that right? in which case bringing that call into this function would be ok. 

Making the PubManager.add_shape call sync in here basically broke some assumptions about certain failure conditions when we don't have perms on a table that I wasn't comfortable changing.

msfstef: `add_shape` currently will still go through the pub man, but the process handling requests 1) will have already populated its filters from ShapeStatus and 2) can always handle requests as the publication updating is done in a separate process.

However, the call may still block until we have confirmed the publication actually has the specified relation configured. And if the table does not have the appropriate terms, it _will_ raise.

I suggest we keep it in the snapshotter and visit it separately

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
magnetised: The materializer wasn't marked as significant, so would have been revived on a crash. Only the consumer dying would have brought it down.

magnetised:
> What race conditions does this remove?

IIRC there were problems with tests where the gap between the consumer starting and the name registration happening in ShapeLogCollector was causing problems -- the gap between the process existing and being able to resolve it was painful. In production use there isn't really a problem.

icehaunter (approved):

PR #3390, opened 2025-11-05
chore(sync-service): Reduce memory consumption of consumer supervisor

magnetised:
A continuation of #3332 reducing the amount of memory used by the consumer supervisor. Because supervisors hold the process and the config needed to start it, the trick is to pass the absolute minimum required config to the supervisor and have the process itself lookup anything else it needs based on that.

By removing the dependency injection config from the consumer state and relying on Repatch in tests this also virally spread to require the removal of all Mox usage because if Repatch is called on a module defining a behaviour it alters that module so that Mox no longer recognised it when attempting to define functions from the defined behaviour.

I've tried to simplify setup and remove the need for mocking where possible by relying on the stack id to separate processes, so even in a test doing `GenServer.call(name(stack_id), ...)` will work. I think a lot of these custom names were holdovers from the pre-stack id days, when processes weren't so perfectly sandboxed.

So instead of using say `my_function(arg, server: server)` I've just replaced the server reference with a stack id as much as possible. I think we can eventually get everything down to `my_function(stack_id, ...args)`.

I also took the opportunity to remove all the duplicate handling of keyword lists and maps in the various `name/1` functions as `ProcessRegistry.name/3` already has this.

I've tried to chase all the rabbits in this, but may have missed some.


packages/sync-service/lib/electric/replication/publication_manager/configurator.ex
```diff
@@ -44,10 +44,7 @@ defmodule Electric.Replication.PublicationManager.Configurator do
           scheduled_update_ref: nil | reference()
         }
 
-  def name(stack_id) when not is_map(stack_id) and not is_list(stack_id),
-    do: Electric.ProcessRegistry.name(stack_id, __MODULE__)
-
-  def name(opts), do: name(Access.fetch!(opts, :stack_id))
+  def name(stack_ref), do: Electric.ProcessRegistry.name(stack_ref, __MODULE__)
```
robacourt, line 47: How come we're calling it `stack_ref` now now `stack_id`?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -95,9 +95,14 @@ defmodule Electric.Shapes.Consumer do
     Electric.Telemetry.Sentry.set_tags_context(metadata)
 
     {action, config} = Map.pop!(config, :action)
+    # The existing body of consumer tests made it impossible to replace this dynamic
+    # setting with a static module alias. But do note that this is only set to anything
+    # other than Electric.ShapeCache.ShapeStatus in the test environment.
```
robacourt: This comment is redundant now

packages/sync-service/lib/electric/replication/publication_manager/configurator.ex
msfstef: I had the same question, but I think stack_ref is shorthand for either stack_id binary or options/state that contains the stack id, and the Electric.ProcessRegistry.name does the underlying logic of taking that out (see PR desc).

I don't love the stack_ref name either as it confuses me, sounds like an alias of stack_id, but I get the concept.

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -468,4 +417,12 @@ defmodule Electric.ShapeCache do
 
     {descendents ++ [{handle, shape} | siblings], known}
   end
+
+  if Mix.env() == :test do
```
msfstef, line 421: nit: maybe we should use a macro for this since it seems it's starting to show up in various modules - ignore for this PR, bringing it up for discussion

packages/sync-service/mix.exs
```diff
@@ -118,7 +118,6 @@ defmodule Electric.MixProject do
       {:dialyxir, "~> 1.4", only: [:test], runtime: false},
       {:excoveralls, "~> 0.18", only: [:test], runtime: false},
       {:junit_formatter, "~> 3.4", only: [:test], runtime: false},
-      {:mox, "~> 1.1", only: [:test]},
```
msfstef, line 121: beautiful 👌

msfstef (approved):
Beautiful!

We should get more eyes on this and merge it ASAP as it touches _everything_

packages/sync-service/lib/electric/replication/publication_manager/configurator.ex
robacourt: Oh, I see

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
```diff
@@ -20,12 +20,8 @@ defmodule Electric.Postgres.Inspector.EtsInspector do
   @behaviour Electric.Postgres.Inspector
 
   ## Public API
-  def name(opts) do
-    Keyword.get(
-      opts,
-      :name,
```
alco, line 26: Have you checked that nowhere in the code do we use inspector opts with the `name` key in them? `Electric.ProcessRegistry.name()` doesn't check for the presense of this key.

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -228,7 +226,8 @@ defmodule Electric.Shapes.Api do
 
   @spec delete_shape(Request.t()) :: Response.t()
   def delete_shape(%Request{handle: handle} = request) when is_binary(handle) do
-    :ok = Shapes.clean_shape(handle, request.api)
+    %{api: %{stack_id: stack_id}} = request
+    :ok = Shapes.clean_shape(handle, stack_id)
```
alco: It's somewhat unexpected to have different functions in `Shapes` take `stack_id` in different places. The other functions from `Shapes` called in this module pass `stack_id` as the first arg.

alco (approved):
:trophy: :medal_sports: :gift: :heart:

packages/sync-service/lib/electric/shapes/api.ex
magnetised: Yeah, sigh. I didn't want to take on normalizing the args order in this pr as well as everything else. IMHO we should just settle on `(stack_id, shape_handle, ...)` everywhere and lose this "stack/server/opts reference as the last args" style.

`Shapes` is particularly weird though...

Issue #3365, opened 2025-11-03
Faster snapshot creation timeout handling


balegas:
When we receive a new shape request, Electric creates the resources for the shape before queryinh the snapshot from Postgres. If this request times out, electric will cleanup all the resources --- which takes some.

We've had an incident where Electric was not able to keep up with WAL because of the amount of work that was put into handling new shape requests and failing.

We have to optimize this path of code to reduce the amount of time spent in teardown on timeouts or be able to pick up on previous attempt to create a snapshot that failed.

msfstef:
I think the simplest approach to this, on top of the admission control/rate limiter which https://github.com/electric-sql/electric/pull/3353 adds, is to simply not tear down shapes because of a snapshot connection queue timeout and instead retry until we get it.

Subsequent requests for that shape will simply hold and wait for that shape a bit, and all shapes will queue up to be created.

Worst case scenario is we end up with loads of shapes waiting for pool connections, and requests being served 503s from timeouts waiting for the shapes to be created, but we shouldn't be seeing an infinite loop of the system creating and deleting shapes and we should observe it eventually catching up.

msfstef:
We have some new developments:
- we are now reducing consumer process spin up to just one process in https://github.com/electric-sql/electric/pull/3389
- memory involved in spinning them up and down is reduced in https://github.com/electric-sql/electric/pull/3390
- faster ETS removal https://github.com/electric-sql/electric/pull/3385
- issue opened on faster/async removal w.r.t. shape monitor https://github.com/electric-sql/electric/issues/3398

Given all of these, I think we should not concern ourselves with the cost of bringing a consumer up and down again. However, I have the following concerns.

1. We currently have a default "timeout to first data" on snapshots of 30 seconds, meaning that we will not fail snapshots waiting to see data until 30 seconds have passed, and for those 30 seconds + however long it takes to stream the data the pool connection will be taken.
  1a. We have seen on boot when surge of new shapes come in that snapshot requests start taking longer and longer, with some taking >20 sec. In fact many requests time out because of this, and I've opened a PR so that we can terminate requests after some waiting time without killing the shape https://github.com/electric-sql/electric/pull/3391 - but perhaps we should just terminate the shapes earlier if they see no data?
[...]

PR #3391, opened 2025-11-05
feat: Fail requests waiting for snapshot after a short time without crashing shape

msfstef:
Relevant to https://github.com/electric-sql/electric/issues/3365

We have observed lots of timeouts on requests calling `await_snapshot_start`, which fail with a `RuntimeError` because the call times out after 15 seconds. We log those as errors on Sentry and return 500s, but we should break down what happens more cleanly.

Looking at honeycomb I've noticed at high peak load we sometimes see queries taking >20-25 seconds, and I believe that it takes that long to see the first data appear (e.g. because the query is unindexed/on a large table).

Our default snapshot "timeout" to first data is 30 seconds, as we want to allow snapshot to take long, but requests for it time out before we error the snapshot, ending up as 500s, although they really should be 503s.

I've added a `snapshot_waiting_for_data_interval` option and set that to 5 seconds. As soon as the query is executed, we send a `waiting_for_snapshot_data` message to the consumer every 5 seconds, and fail any snapshot waiters with an appropriate error.

This solves being able to quickly and correctly reply to requests without holding them for too long, without sacrificing the ability to let a shape take longer than how long we want requests to be held for.

With this, we can observe if we still get snapshot timeouts 500s, at which point I can only think that it might happen because the consumers take _too long_ to subscribe to a busy ShapeLogCollector, but I have no evidence for that so I'd like to filter out the cases where the database being the actual bottleneck.

## NOTE

Currently if a request comes in the _moment_ a `waiting_for_snapshot_data` message is sent, it will immediately be failed, so the interval timeout is not per request. However that is already the behaviour for our overall snapshot timeout - if a request comes in right before the overall timeout runs out it is immediately failed.

I could make the request holding per-request by adding "request_arrived_at" timestamps on the waiters and doing appropriate calculations, but I'd prefer to keep things simple, especially for cases like these where it's already _very_ slow snapshots.

There is also an alternative of just setting the timeout for the snapshot to get data to a much smaller time, but it _will_ start failing actual queries that can be fulfilled that just take long.

### DIFF
--- a/.changeset/nine-walls-carry.md
+++ b/.changeset/nine-walls-carry.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Fail requests waiting for snapshot to start after shorter time without failing entire snapshot.
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -65,6 +65,10 @@ defmodule Electric.ShapeCache do
             snapshot_timeout_to_first_data: [
               type: {:or, [:non_neg_integer, {:in, [:infinity]}]},
               default: :timer.seconds(30)
+            ],
+            snapshot_waiting_for_data_interval: [
+              type: :non_neg_integer,
+              default: :timer.seconds(5)
             ]
           )
 
@@ -171,13 +175,16 @@ defmodule Electric.ShapeCache do
         try do
           Electric.Shapes.Consumer.await_snapshot_start(
             %{stack_id: stack_id, shape_handle: shape_handle},
-            15_000
+            Access.get(opts, :timeout, 15_000)
           )
         catch
           :exit, {:timeout, {GenServer, :call, _}} ->
             # Please note that :await_snapshot_start can also return a timeout error as well
             # as the call timing out and being handled here. A timeout error will be returned
-            # by :await_snapshot_start if the PublicationManager queries take longer than 5 seconds.
+            # by :await_snapshot_start if the PublicationManager queries take too long,
+            # or if the snapshot itself takes too long to return any data. If the call itself
+            # times out it is likely that there is a surge of new shapes and consumers are blocked
+            # on registering themselves with the shape log collector.
             Logger.error("Failed to await snapshot start for shape #{shape_handle}: timeout")
             {:error, %RuntimeError{message: "Timed out while waiting for snapshot to start"}}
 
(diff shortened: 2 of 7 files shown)


msfstef:
@alco @icehaunter what are your thoughts on not doing any of this and instead either
1. Reduce the "time to first data" to something much smaller, like 10 seconds, or
2. Simply assume that if the `await_snapshot_start` call is done with a timeout, and it times out, we convert the timeout to a snapshot error that states that the snapshot is taking too long to start/the system is overloaded/make sure you have indices on your tables

Solution 2. is by far the simplest and also has the added benefit of being a per-client timeout.

alco:
What are the reasons for avoiding the same timeout on awaiting snapshot by the clients and snapshot's time to first data? My thinking is that if we have a timeout on snapshot creation, the client who initiates that shape creation should wait exactly as long before getting back a SnapshotError.

Do we have an upper bound on how long a client can wait for a response that is lower than the snapshot's time to first data timeout?

msfstef:
> Do we have an upper bound on how long a client can wait for a response that is lower than the snapshot's time to first data timeout?

On our current cloud infrastructure, the [CF default is 120 seconds](https://developers.cloudflare.com/support/troubleshooting/http-status-codes/cloudflare-5xx-errors/error-524/#common-causes) and configurable, so perhaps it's not that much of an issue.

But I think more generally we would want our requests to _have_ an upper bound on how much we hold them for, similar to the 20 second (but confiugrable) long poll we use.

I suppose my main concern is the coupling of the upper bound of the requests to the upper bound we wait on the snapshot for.

icehaunter (approved):
I think main reason for crashes was that pool checkout timeout was larger than message waiting. this PR seems like a good approach. I also like just using the plain timeout within `await_snapshot_start`. Functionally the only way for it to be slow is on a saturated pool, so error conversion is straightforward

msfstef:
> Functionally the only way for it to be slow is on a saturated pool, so error conversion is straightforward

The problem is, and the reason I've opened this PR as well, that it might also be the consumer being very slow to initialise, like the blocking call to subscribe to ShapeLogCollector, and I am trying to confirm that these errors are not all just timeout mismatches.

I'm thinking of ditching this PR and instead initially just increasing the timeout on the `await_snapshot_start` call to something like 45-60 seconds. Smaller change that should filter out any timeouts because of mismatches.

Basically the changes in this PR break any usage of `await_snapshot_start` that is willing to wait an arbitrary amount of time for it, like the call in the materializer.

msfstef:
Closing as I opted to just increase the timeout https://github.com/electric-sql/electric/pull/3401

Will monitor production behaviour for a while longer and see if we still observe call timeouts, in which case we need to consider the blocking of consumer initialization on subscribing to the shape log consumer.

Issue #3262, opened 2025-10-08
Emit electric.postgres.replication.wal_size with source_id


balegas:
`electric.postgres.replication.wal_size` needs to be tagged with `source_id` to be able to correlate with other metrics. It's one of the few metrics that we track that is still only emitted with `stack_id`.

PR #3392, opened 2025-11-05
 Support externally defined OTEL resource attributes in the metrics exported from ApplicationTelemetry and StackTelemetry

alco:
This change is needed in Electric to allow Cloud to set its own resource attributes for metrics.

Fixes https://github.com/electric-sql/electric/issues/3262.

### DIFF
--- a/.changeset/late-cows-itch.md
+++ b/.changeset/late-cows-itch.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Support externally defined OTEL resource attributes in the metrics exported from ApplicationTelemetry and StackTelemetry


packages/sync-service/lib/electric/telemetry/stack_telemetry.ex
```diff
@@ -70,10 +70,14 @@ with_telemetry [OtelMetricExporter, Telemetry.Metrics] do
...
        metrics: otel_metrics(opts),
        export_period: opts.otel_export_period,
-       resource: %{
-         stack_id: opts.stack_id,
-         instance: %{installation_id: Map.get(opts, :installation_id, "electric_default")}
-       }}
+       resource:
+         Map.merge(
+           %{
+             stack_id: opts.stack_id,
```
robacourt, line 76: If this is to get `source_id` then `stack_id` is `source_id` so why do we need `otel_resource_attributes`? And for the application telemetry, that can't have a `source_id` because it's not source/stack specific

alco: Yeah, it is a bit of a brainmelt.

When Electric runs in the standalone mode, we want every metric to be tagged with `stack_id`. This is doing that.

When Electric runs in the library mode as part of a parent OTP application, the parent app may pass extra resource attributes. That is how we add `source_id` to metrics in Cloud, implemented in https://github.com/electric-sql/stratovolt/pull/931.

The end result is metrics get both `stack_id` and `source_id`. I'm hoping to resolve this in my work on extracting telemetry into a separate library that can then be separately configured by OSS Electric and Cloud. It's taking me a while to finish that so I've decided to propose this interim change so that we could finally have correct tags in our charts.

alco: ApplicationTelemetry won't get `source_id` because it's passed a different set of telemetry options. In Cloud's code we only include `source_id` in those telemetry options that are passed to StackSupervisor. ApplicationTelemetry gets passed a different set of telemetry options.

PR #3396, opened 2025-11-05
Fix documentation links leading to localhost

xamgore:
It's probably a mistake due to copy-paste from a browser.


alco (approved):
Good catch. Thanks for the contribution!

PR #3399, opened 2025-11-05
feat(sync-service): Send only rows for matching shapes instead of transactions

### LINKED ISSUE #3373: Send only rows for matching shapes instead of transactions
When a row in a transaction matches a shapes we're sending the entire transaction payload to the Shape Consumer. Shape Consumers will claim enough memory to hold the entire transaction which is not released until the process hibernates or terminates. 

One hypothesis would be to have the shape matching algorithm sending only the rows, the advantage of this is that processes would possibly hold less memory, the downside is that we need to do a bit more processing sequentially. We want to test and evaluate this hypothesis.

robacourt:
Fixes #3373

Reduces the amount of data that needs to be copied and processed by the shape consumers by filtering the transaction and only including the changes that affect the shape. This should reduce the size of transactions that a shape consumer receives and so potentially reduce it's memory footprint.

Since a transaction may affect many shapes and each shape may take a different subset of the transaction, the `Filter` now returns a map of shape handles to the event to send to each.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer_registry.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer_registry.ex
@@ -88,7 +88,7 @@ defmodule Electric.Shapes.ConsumerRegistry do
   end
 
   @doc """
-  Calls many GenServers asynchronously with different messages and waits
+  Calls many GenServers asynchronously with their message and waits
   for their responses before returning.
 
   Returns `:ok` once all GenServers have responded or have died.


robacourt:
benchmark this

packages/sync-service/lib/electric/shapes/filter.ex
```diff
@@ -176,4 +186,33 @@ defmodule Electric.Shapes.Filter do
...
+         transaction_map
+       ) do
+    transaction_map =
+      filter
+      |> affected_shapes(change)
+      |> Enum.reduce(transaction_map, fn shape_id, transaction_map ->
+        Map.update(
+          transaction_map,
+          shape_id,
+          %Transaction{transaction | changes: [change]},
```
msfstef, line 209: the `%Transaction{}` struct contains metadata like `num_changes`, `last_log_offset`, `affected_relations`

these are all still valid as far as the transaction is concerned, but technically no longer match the actual changes the consumer will see - does this perhaps call for a different struct or anything that will make this more clear, like a `FilteredTransaction`?

msfstef, line 209: Will this cause memory usage amplification since with this change we are duplicating a lot of the data (+ transaction metadata) for each shape inside the SLC process? Or is the underlying memory usage smart enough to not cause duplications and only build up the necessary memory for the wrappers like the structs and lists?

robacourt: Hm. I think github has mixed up the versions. I think you're commenting on an earlier version. This has been resolved already but your comment is not marked as `outdated` 🤷

robacourt: It should be smart enough to not duplicate the data and the only extra memory used will be some lists of refs.

msfstef: I'm going through the commit history and I'm not sure what you mean that it is addressed? We are keeping all of the transaction metadata, but the metadata is wrong (num changes does not match the changes contained, last_log_offset is not the last log offset carried in the transaction, etc)

could you point me to the commit?

robacourt: Ah, ok I misread your message. I thought you were saying we didn't preserve those fields, which I hadn't in a previous version.

But yes, you are right. Those fields either need to be updated or I should use a different struct. I'll have a look at how they're used and see what makes most sense.

msfstef: Ultimately I think we want to move to streaming individual changes to consumers rather than complete transactions, so whatever you come up with I wouldn't worry _too much_ about it - I just feel that a "filtered" transaction and a transaction are not the same, and using the same struct might be confusing.

robacourt:
this will now be implemented as part of #3414

PR #3400, opened 2025-11-06
chore: Remove storage metadata backup logic

msfstef:
Now that lazy shape loading is in place, backing up and reloading storage metadata in the already overloaded ShapeStatus table does not offer any significant benefits.

The only reason it was there was because on startup if we had lots of shapes to recover, we would have thousands of FS operations queued up for reading lots of different metadata from files, and that would block recovery and the ability of the system to start serving requests.

With lazy shape loading, we no longer have an onslaught of FS operations blocking the system recovery, instead they occur on demand and proportionally to shapes requested after a reboot. Holding the cache in memory until a shape is loaded feels unnecessary and one more thing that can go wrong.

### DIFF
--- a/.changeset/angry-monkeys-destroy.md
+++ b/.changeset/angry-monkeys-destroy.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Remove shape storage metadata backup mechanism now that shape lazy loading is in place.


magnetised (approved):
That's a layer of complexity that I'm glad to see the back of.

(Thanks for the merge conflicts...)

alco:
Geez, I was just going to suggest that you wait for Garry to merge his PRs. We'd rather have fewer conflicts for him to deal with to speed up the merge.

PR #3402, opened 2025-11-06
Add query fallback when Electric disconnects

KyleAMathews:


balegas:
Mixed feelings about this one. Obviously it's nice to continue serving data, but wouldn't it provide inconsistent experience? How does it work to retrieve snapshots but not have live updates? E.g. you get the first page, no more live updates for it but then you get a second page that is more recent in time. 

I understand we do progressive snapshots, but I presume we don't want to rely on that. Even if we do, it means we're putting more and more load on the database with fallback query for each page.

msfstef:
I feel that the PG replication is as reliable as a PG read replica - with the exception that we currently don't do read only mode yet so as to serve data while the replication stream is inactive.

If we knew a replication slot _exists_ and where we last left it, we could also create new shapes (i.e. just the snapshots) without replication active since we know that we can "resume" them once we resume replication, and that brings us closer to a read replica behaviour as well.

In short I think we should aim to provide what someone would expect from a read replica, but in change stream format.

KyleAMathews:
> If we knew a replication slot exists and where we last left it, we could also create new shapes (i.e. just the snapshots) without replication active since we know that we can "resume" them once we resume replication, and that brings us closer to a read replica behaviour as well.

Ooo yes! That's a great point.

KyleAMathews:
Closing this PR and converting it to issue #3470 for further discussion and design consideration. The issue includes the original RFC content, team feedback from @balegas and @msfstef, and additional context from internal discussions about resync constraints and architectural alternatives.

PR #3403, opened 2025-11-10
fix: Avoid going through erl file server process for per-shape ops

msfstef:
Started tracing calls to the file server with `:sys.trace` and running Electric with new shapes being created and found a few spots we missed that build up a queue on the file server.

If we want shapes to be scalable we need to avoid blocking on single processes unless strictly necessary (e.g. currently for building the index on the shape log collector)

Took the opportunity to remove unused code.

### DIFF
--- a/packages/sync-service/lib/electric/async_deleter.ex
+++ b/packages/sync-service/lib/electric/async_deleter.ex
@@ -133,7 +133,7 @@ defmodule Electric.AsyncDeleter do
     duration = System.monotonic_time(:millisecond) - start_time
 
     Logger.debug(
-      "AsyncDeleter: deleted #{length(state.pending)} paths " <>
+      "AsyncDeleter: deleted #{length(state.in_progress)} paths " <>
         "for stack #{state.stack_id} in #{duration}ms"
     )
 


packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -109,7 +109,7 @@ defmodule Electric.AsyncDeleter do
   defp unique_destination(trash_dir, base) do
     attempt = Path.join(trash_dir, base <> "_" <> random_suffix())
 
-    if File.exists?(attempt) do
+    if File.exists?(attempt, [:raw]) do
```
msfstef, line 112: This check is potentially not necessary - each delete is now an existence syscall + a rename syscall which feels like too much.

A unique destination should be _unique_ and we can ensure it with super high probability (e.g. unique int + timestamp) - but also worst case scenario we basically just overwrite a destination that was already in the trash?

msfstef: update: if the directory is non-empty it would be a problem, and even if it isn't we don't want to actually trigger any removals on this rename - either way, the point still stands that our uniqueness should be strong enough and we should only fallback to try another unique name in case of a failure perhaps? I'll do that on a separate PR

icehaunter: I agree on random suffix being enough to avoid an existence check if prefixed with a native-precision timestamp

icehaunter (approved):
Good catch

packages/sync-service/lib/electric/async_deleter.ex
icehaunter: Actually if we use shape handle along with a precise timestamp that guarantees uniqueness because same shape can't be deleted twice anyway

msfstef: The AsyncDeleter should be agnostic to what it's removing, but you're right that shape folders are unique anyway - should be fine with a timestamp for the rest

alco:
Brilliant changes. May I suggest the idea of defining our own module for file ops to provide a reliable abstraction for raw file access over using `prim_file`, `file`, and `File`?

msfstef:
> May I suggest the idea of defining our own module for file ops to provide a reliable abstraction for raw file access over using prim_file, file, and File?

I think `Electric.ShapeCache.PureFileStorage.FileInfo` has taken on that role but I think somehow inconsistently + it covers more actions than you would expect it to - we can refactor things into a raw file utility module I agree

PR #3412, opened 2025-11-10
Replace direct telemetry deps with the electric_telemetry library

alco:
Extract telemetry code from Electric into a separate package. On top of that, the new package accepts additional user-provided periodic measurements and metric definitions for otel/prometheus exporters; both exporters now use the same list of metrics.

One thing that's been lost in this move is the `stack_id` sentry tag inside the CallHomeRepoter process. Sentry is a dependency of Electric and there's no straightforward way of propagating it into CallHomeReporter which is a module in the electric_telemetry dependency now. I don't feel like it's a big loss because we set stack_id in the logger metadata for that process, as well as mention it in the process name and label.

With this merged, we can delete the electric-sql/electric-telemetry repo.

### DIFF
--- a/packages/sync-service/dev/otel-collector-honeycomb-config.yaml
+++ b/packages/sync-service/dev/otel-collector-honeycomb-config.yaml
@@ -4,8 +4,8 @@ exporters:
     headers:
       x-honeycomb-team: ${env:ELECTRIC_HNY_API_KEY}
       x-honeycomb-dataset: ${env:ELECTRIC_HNY_DATASET}
-  logging:
-    loglevel: info
+  debug:
+    verbosity: normal
 receivers:
   otlp:
     protocols:
@@ -54,4 +54,4 @@ service:
         - batch
       exporters:
         - otlp/honeycomb
-        - logging
+        - debug


msfstef:
Why not keep https://github.com/electric-sql/electric-telemetry as part of our mono-repo as another elixir project? Easier to access and work with.

alco:
@msfstef Main reason is decoupling.

Our primary use case for telemetry is in Electric Cloud. So if we're going to make any changes to the telemetry code, it'll be informed by Cloud and should be tested first in Cloud. All while Electric keeps using its own stable version of the lib.

How can we achieve that?

1. Keep electric-telemetry in a separate repo and pin Cloud and Electric to different commits.
2. Publish electric-telemetry on Hex and pin different versions of the package in the two repos.

Moving electric-telemetry into Electric's repo would work fine with the 2nd option but I'm against publishing our telemetry as a Hex package since it is only ever going to work as part of Electric.

With the 1st option it becomes awkward: we cannot use electric-telemetry in Electric as a path dependency because then Electric will necessarily always use the latest version of it. And the alternative is to pull it from Github (i.e. `{:electric_telemetry, github: "electric-sql/electric", path: "packages/electric-telemetry", ref: "..."}`), even though it's part of the same repo as Electric itself?

packages/sync-service/lib/electric/telemetry/processes.ex
msfstef, line 1: We have a bunch of tests still in place for these, and I suspect they are still passing because we're reusing the `Electric.Telemetry` namespace, whereas I think this should now be in `ElectricTelemetry`, and the tests should move to the new repo, no?

packages/sync-service/config/runtime.exs
```diff
@@ -274,42 +269,20 @@ config :electric,
     )
 
 if Electric.telemetry_enabled?() do
+  # Disable the default telemetry_poller process since we start our own in
+  # `ElectricTelemetry.{ApplicationTelemetry, StackTelemetry}`.
+  config :telemetry_poller, default: false
```
magnetised, line 277: could we set this config value as a default in `packages/electric-telemetry/mix.exs`?

packages/sync-service/lib/electric/telemetry/open_telemetry/config.ex
```diff
@@ -0,0 +1,36 @@
...
+        {:otel_batch_processor, %{}}
+      end
+
+    otel_simple_processor =
+      if otel_debug? do
+        # In this mode, each span is printed to stdout as soon as it ends, without batching.
+        {:otel_simple_processor, %{exporter: {:otel_exporter_stdout, []}}}
+      end
+
+    config :opentelemetry,
```
magnetised, line 27: Find this a bit weird, but I get it.

packages/sync-service/lib/electric/stack_supervisor.ex
```diff
@@ -389,4 +378,101 @@ defmodule Electric.StackSupervisor do
 
     Supervisor.init(children, strategy: :one_for_one, auto_shutdown: :any_significant)
   end
+
+  defp telemetry_children(%{stack_telemetry: stack_telemetry}), do: [stack_telemetry]
```
magnetised: Nit: could this telemetry stuff be moved to a child module to keep this cleaner?

packages/sync-service/mix.exs
```diff
@@ -61,9 +74,7 @@ defmodule Electric.MixProject do
 
   def application do
     [
-      extra_applications: [:logger, :os_mon, :runtime_tools],
```
magnetised, line 64: This is all worth it to get rid of `os_mon` from the base app... ❤️

packages/sync-service/config/runtime.exs
alco: I don't know what you mean here.

The problem is that `telemetry_poller` gets started before `electric_telemetry` because it's a dependency. And it looks up the `:default` key in its app env at startup. So the key has to be set before `telemetry_poller` gets started and `config/runtime.exs` of the parent app is the only place to have this config in.

packages/sync-service/lib/electric/stack_supervisor.ex
alco: Done!

packages/sync-service/config/runtime.exs
magnetised: You can add `env` configuration to the `application` callback in mix.exs: https://hexdocs.pm/elixir/1.19.3/Application.html#module-the-application-environment

If we put this required config there it would remove the need to add it at every point of use

alco: But that configuration is for the app owns the mix.exs. We cannot put arbitrary config for other apps, such as `telemetry_poller`, in there.

PR #3418, opened 2025-11-11
fix: Ensure `consumer_pid` returns nil if table missing

msfstef:
If the shape log collector is down, then the consumer supervisor is down and no consumers are alive, and thus we can safely return `nil` when looking up a consumer process.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer_registry.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer_registry.ex
@@ -120,6 +120,8 @@ defmodule Electric.Shapes.ConsumerRegistry do
 
   defp consumer_pid(handle, table) do
     :ets.lookup_element(table, handle, 2, nil)
+  rescue
+    ArgumentError -> nil
   end
 
   defp start_consumer!(handle, %__MODULE__{stack_id: stack_id} = state) do
--- a/.changeset/rude-olives-pay.md
+++ b/.changeset/rude-olives-pay.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Ensure consumer registry does not crash on a lookup when registry table is missing.


packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -120,6 +120,8 @@ defmodule Electric.Shapes.ConsumerRegistry do
 
   defp consumer_pid(handle, table) do
     :ets.lookup_element(table, handle, 2, nil)
+  rescue
+    ArgumentError -> nil
```
robacourt, line 124: if this returns nil, it'll try to start a consumer and fail, so you'll still get an exception. I guess your argument is that it's an easier to interpret error message? I would argue that it's better to raise here with a nice readable message, that way you get your readable message but without swallowing the error which could cover up bugs

msfstef: This utility is used to go from consumer handle -> consumer pid if it exists. It is used by the shape cleaner, by the monitor, in order to check if a consumer is alive.

My argument is that this API should return a consumer pid if the consumer for a given handle is alive, and nil if it is not. If the shape log collector is dead (no table), then we know that consumers are also dead, and thus returning nil is correct, and users of this API will see the correct behaviour.

If something else crashes along the way it should not be because this returns nil, as this would be the correct reply by this public API I think.

robacourt: Ah, ok, I was just thinking about how it's used in `publish/3`. As a public API that makes sense.

robacourt (approved):

packages/sync-service/lib/electric/shapes/consumer_registry.ex
msfstef: Yeah - `publish/3` is called from within the shape log collector so it should never fail with an argument error, so that didn't concern me much. What concerned me was that we were killing vital processes because of this raise that is not strictly necessary from this API.

msfstef:
@magnetised I heard your argument about controlling de-registration, but apart from the shape log collector that needs to send data to shapes, why would we want to control deregistration for processes that try to terminate the consumer, like shape cleaner and monitor?

magnetised:
We need to avoid the race condition where the SLC has a shape in its filter so tries to send data to it. The consumer has been stopped and automatically removed itself from the registration ets then the SLC will try to start it again. So we need to have the SLC as authoritative over the existing consumers and only remove the pid registration when its removed the shape from its filters.

Note that this doesn't prevent us stopping the consumer process just from de-registering it. monitoring a dead process just immediately sends a DOWN message, which is covered by the publish call.

Does that answer your question?

msfstef:
> We need to avoid the race condition where the SLC has a shape in its filter so tries to send data to it.

I understand that - my question is that if this is an SLC concern, why has this SLC API been coupled to the `Consumer.whereis` API? If the monitor or shape cleaner (or any other process) is checking if a given consumer process is alive, is there a reason they need to respect the SLC's view of the world?

The `Consumer.whereis` and `Consumer.stop_and_clean` APIs are used in various other processes, and all of these are now coupled to the shape log collector being alive - do they need to be, or could the SLC just check its own registry for its own publishing of data?

magnetised:
> The Consumer.whereis and Consumer.stop_and_clean APIs are used in various other processes, and all of these are now coupled to the shape log collector being alive - do they need to be, or could the SLC just check its own registry for its own publishing of data?

I feel like there's some underlying concern in what you're asking that's implicit in the questions but not actually asked. But, there is no reason why the SLC should own the consumer registry table, in my original impl it was owned by another process but in the review it was strongly suggested that the SLC should own the table.

The only coupling between the consumer process status api and the SLC is in the ownership of the ets by the SLC except in the sense that the SLC has to be the one to remove registered processes.

msfstef:
> I feel like there's some underlying concern in what you're asking that's implicit in the questions but not actually asked.

Sorry, to clarify - the concern is that these APIs are used by processes that live above the SLC in the supervision tree and we observed errors because they try to look up consumers on a registry that does not exist.

This PR addresses that by making the public API for the lookup return `nil` if the registry does not exist, which works with the implicit knowledge of how the supervision tree is structured (i.e. if the registry is dead, consumers are also definitely dead).

My concern is just with the fact that we have processes outside of that supervision tree using these consumer APIs, and given that consumers are no longer registered on a global registry that feels potentially unsafe - but this might just be a gut reaction to these errors. Perhaps the shape cleaner can just be brought up after the log collector, and the monitor is sort of going away anyway.

magnetised:
We could just make the ShapeStatusOwner owner of the consumer registry table too.

Issue #3416, opened 2025-11-11
Call home telemetry metric aggregation is too memory intensive


msfstef:
**Bug description**
On busy instances with lots of traffic, the call home reporter seems to accumulate a non-trivial amount of memory to produce reports - and similarly for its ETS table.

On further inspection, it seems that the metrics it produces are exact, meaning that it maintains all data necessary to produce exact percentile information, uniqueness checks, etc.

**Expected behavior**
Metrics should be produced in a streaming fashion where possible with near constant memory overhead, and using probabilistic methods elsewhere to maintain the constant memory. We can use something like [t-digest](https://github.com/tdunning/t-digest) for percentiles, and somethine like [HyperLogLog](https://en.wikipedia.org/wiki/HyperLogLog) for cardinality of a set of unique values.

PR #3422, opened 2025-11-11
feat: Memory efficient call home telemetry

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3416

## Intent

The problem I was trying to solve is high memory usage by the call home reporter (along with its ETS table), because it basically collected all data to calculate histogram statistics (min, max, mean, median, mode) and count unique occurrences. The amount of memory needed to do this scaled with the amount of data produced, which is not great.

The solution to this would be for our telemetry to use appropriate estimates for these statistics using methods that are much more memory efficient, and meant for streaming datasets.

## Approach

I wanted to ensure that whatever method I ended up using for calculating these statistics was tested, but currently the call home reporter is doing everything in one module which is not easy to test because of its inclusion of requests, telemetry logic, etc.

I decided to first split up the actual measurement logic into its own module (see https://github.com/electric-sql/electric/commit/994f0df444ef6a9533bc4cef066bc98585e29493) while keeping the same calculation methods, then write a test suite for it, with the help of Claude of course (see https://github.com/electric-sql/electric/commit/5ffdb037a694d8fc15bd526d65e5612c90c099d7).

With the calculation logic split out and tested, I then proceeded to replace the metrics that were causing issues, while trying not to break the test suite too much. First I started with unique element counting, and then with the summary statistics.

I opted to try and keep the same approach that is using an ETS table for each process to update its own metrics concurrently, rather than make everything cast messages to the reporter and calculate statistics in a linearized, sequential manner. This complicates things significantly but I think it's a good first step to avoid changing too many things.

## Unique Element Counting
The ideal algorithm for this is [HyperLogLog](https://en.wikipedia.org/wiki/HyperLogLog), which has a decent [Elixir implementation](https://hexdocs.pm/hypex/readme.html#usage), but it would mean adding a dependency and it is not ETS-native, such that every process can update the metrics itself concurrently with low overhead.

The algorithm is simple enough and I have a pathway to doing an ETS-native approach that would work, but it would require significantly more testing, is a bit cryptic and hard to follow, which I think would end up being 1) bug prone, and 2) a pain to maintain.

Instead I opted for [cardinality estimation with linear probabilistic counting](https://www.waitingforcode.com/big-data-algorithms/cardinality-estimation-linear-probabilistic-counting/read) - basically allocating a bitmap of a fixed size, and using a hash of the same size of the items tracked to "fill" the appropriate bits.

The percentage of the bitmap that is "filled" can then be used to estimate the cardinality of the set that we're counting. The accuracy of that estimate is related to the size of the allocated bitmap, and I've included an error calculation formula in code. 

I've opted for ~64kbit (or 8kb) bitmaps, so every count unique metric will use up a fixed memory of 8kb, which should give use very good accuracy for unique counts of ~100k. Our reporting interval is 30 minutes, I think this is a pretty good balance. We could make this configurable per-metric, but I think it would be overkill.

We can consider using HyperLogLog again if we want higher cardinality measurements, but for now the bitmap approach is much easier to follow and implement.

## Summary Metrics
For min, max, and mean, it's fairly trivial to just keep a running tally of items streaming through (min, max, count, sum) and calculate the mean at the end. I'm using ETS match specs with `:ets.select_replace` for this, which is pretty powerful, and should do the trick.

In this iteration, I've completely removed the **median** and **mode**. If they are very important to us right now, it would be good to know and we can put the effort into making them work - but from what I could gather the telemetry reporters I found don't use them or don't even report summary metrics at all.

For percentiles, we can use a [t-digest](https://github.com/tdunning/t-digest), and for the median alone we could do something simpler like [P^2 quantile estimation](https://aakinshin.net/posts/p2-quantile-estimator-intro/). Neither are easily implementable in the pattern we have now, they would be much easier if done in-process with linearized stream of items. I did not want to use that approach for now.

For the mode, there's several memory-friendly ways to calculate it ([space saving alg](https://imoumoulidou.github.io/SpaceSaving_Presentation.pdf), and [count-min sketch](https://en.wikipedia.org/wiki/Count%E2%80%93min_sketch)) but again they are not easily computable with ETS and concurrent updates - they would need a linearized approach.

Therefore to avoid changing the concurrency profile of how we collect these metrics, I've just opted to remove them altogether.

```elixir
# before
run_queue_total %{max: 0, min: 0, mode: 0, mean: 0.0, median: 0},
# after
run_queue_total: %{max: 0, min: 0, mean: 0.0}
```

## Potential Investigation

If instead of using ETS concurrent updates we just let each process send their telemetry events to a telemetry processor process, we can much more easily leverage existing implementations of these various data structures for streaming statistics, and keep the percentiles, the modes, cardinality counts for arbitrary sizes, etc.

I decided to not introduce new dependencies and keep the same approach we are currently using and focus on solving the memory buildup only.

## TODO
[] Test the memory buildup of the call home reporter with these changes (lol)

### DIFF
--- a/packages/sync-service/lib/electric/telemetry/measurement.ex
+++ b/packages/sync-service/lib/electric/telemetry/measurement.ex
@@ -113,6 +113,7 @@ defmodule Electric.Telemetry.Measurement do
           {
             {
               {:const, key},
+              :"$1",
               {:min, :"$2", {:const, value}},
               {:max, :"$3", {:const, value}},
               {:+, :"$4", 1},


packages/sync-service/lib/electric/telemetry/measurement.ex
```diff
@@ -0,0 +1,157 @@
...
+        end
+    end
+  end
+
+  def calc_metric(%__MODULE__{table: table}, key, default \\ nil) do
+    case :ets.lookup(table, key) do
+      [] ->
+        default
+
+      [{^key, value}] ->
```
magnetised: this feels fragile. is it not worth tagging the ets tuples with a type? so this would be e.g. `{^key, :counter, count}`

packages/sync-service/test/electric/telemetry/measurement_test.exs
```diff
@@ -0,0 +1,553 @@
...
+    end
+
+    test "calc_metric returns default if table doesn't exist" do
+      fake_measurement = %Measurement{table: :nonexistent_table}
+      assert Measurement.calc_metric(fake_measurement, :any_key, 0) == 0
+    end
+
+    test "counter increments are atomic" do
+      measurement = Measurement.init(:test_counter_atomic)
+      # Simulate concurrent increments
```
magnetised, line 66: claude was here, stating the obvious

packages/sync-service/test/electric/telemetry/measurement_test.exs
```diff
@@ -0,0 +1,553 @@
...
+      measurement = Measurement.init(:test_sum_atomic)
+      # Simulate concurrent sums
+      tasks =
+        for i <- 1..50 do
+          Task.async(fn -> Measurement.handle_sum(measurement, :concurrent_sum, i) end)
+        end
+
+      Task.await_many(tasks)
+
+      # Sum of 1..50 is 1275
```
magnetised: though, appreciate this one.. 😉

packages/sync-service/test/electric/telemetry/measurement_test.exs
```diff
@@ -0,0 +1,553 @@
...
+    end
+
+    test "estimates large number of unique values", %{measurement: measurement} do
+      # Add 100 unique values
+      for i <- 1..100 do
+        Measurement.handle_unique_count(measurement, :large_set, i)
+      end
+
+      count = Measurement.calc_metric(measurement, :large_set)
+      # With 1024-bit bitmap, estimate should be reasonably accurate for 100 unique values
```
magnetised: thought it was 8192 bits?

msfstef: claude was here and I missed this comment - the reference to numbers in comments is killing me

packages/sync-service/lib/electric/telemetry/measurement.ex
msfstef: I hear that - I was starting to add that and decided not to as it's extra memory - but I agree that clarity and robustness are probably more important. Added.

PR #3424, opened 2025-11-12
feat(sync-service): Remove Shapes.Monitor

### LINKED ISSUE #3398: Consumer.Monitor.RefCounter is slowing down shape deletion
In order to prevent errors from reads to a shapes files we have a process that waits until all api read requests for a shape have terminated before terminating the shape consumer and deleting its data.

This requires serialising all deletions through a single genserver that tracks termination requests and api requests which is a major bottleneck when deleting lots of shapes.

Instead of this I propose simply delaying the deletion of the shape's data by some number of minutes (say 5). This will ensure that all active reads will have terminated (with nearly 100% probability) before deleting the data without the need to track state in a genserver.

magnetised:
Shapes.Monitor existed in order to be able to shutdown shape processes without returning errors to clients actively reading that shape's data.

However both the registration of reader processes and the coordination with the consumer shutdown were throttling the system at points of high load.

This PR removes this Monitor system, replacing it with careful handling of file deletions within the storage system and a streamlined, simplified shape removal API via ShapeCleaner.remove_shape and friends.

Shape removal is split into critical immediate deregistration steps, each with a deterministic and short duration, and longer less critical steps, handled via an async Task process. This is much as before except the steps are all performed within the remove_shape/2 function, rather than spread throughout the code.

Other tweaks:

- Fix bug with Shapes.Api.Delete not passing the stack_id
- Fix argument ordering in ShapeCleaner.remove_shapes* and AsyncDeleter.delete
- Remove deletion marker file from PureFileStorage.cleanup - AsyncDeleter does an immediate rename of the storage directory which is atomic under linux so there is no intermediate, half-deleted, state to catch with the deletion marker, and not having it saves us 2 syscalls
- Split StackConfig.lookup into a raising and non-raising version (with a default) rather than just a raising version.
- Fix Storage.cleanup!/{1,2} implementations for TestStorage and InMemoryStorage

Fixes #3398

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
@@ -846,10 +846,10 @@ defmodule Electric.ShapeCache.PureFileStorage do
     metadata =
       read_multiple_cached_metadata(opts, [:snapshot_started?, :last_snapshot_chunk])
 
-    snapshot_started = Keyword.get(metadata, :snapshot_started?) || false
+    snapshot_started? = Keyword.get(metadata, :snapshot_started?) || false
     last_snapshot_chunk = Keyword.get(metadata, :last_snapshot_chunk)
 
-    if !snapshot_started && !shape_gone?(opts) do
+    if not snapshot_started? and not shape_gone?(opts) do
       raise(Storage.Error, message: "Snapshot not started")
     end
 
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -645,6 +645,10 @@ defmodule Electric.Shapes.Consumer do
     {reason, state}
   end
 
+  defp stop_and_clean(state) do
+    {:stop, @stop_and_clean_reason, mark_for_removal(state)}
+  end
+
   defp reply_to_snapshot_waiters(%{awaiting_snapshot_start: []} = state, _reply) do
     state
   end
@@ -803,10 +807,6 @@ defmodule Electric.Shapes.Consumer do
       {:shutdown, false} -> {:stop, reason, state}
       _ -> stop_and_clean(state)
     end
-  end
-
-  defp stop_and_clean(state) do
-    {:stop, @stop_and_clean_reason, mark_for_removal(state)}
   end
 
   defp terminate_writer(state) do


packages/sync-service/lib/electric/async_deleter.ex
```diff
@@ -40,28 +43,27 @@ defmodule Electric.AsyncDeleter do
...
-  Returns {:ok, output} on success or {:error, reason} on failure.
-  """
-  def delete(path, opts) when is_binary(path) do
-    stack_id =
-      opts[:stack_id] ||
-        raise ArgumentError, message: "Missing required :stack_id in opts: #{inspect(opts)}"
+  Deletes the given directory by first renaming it into the stack's trash directory
+  then asynchronously removing the trash entry using rm -rf.
 
+  Returns `:ok` on success or `{:error, reason}` on failure.
```
msfstef: nit: was claude here? (would prefer a `@spec`)

packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
```diff
@@ -259,7 +259,11 @@ defmodule Electric.Replication.PublicationManager.RelationTracker do
     state = reply_to_relation_waiters(oid_rel, {:error, error}, state)
 
     if not is_struct(error, DBConnection.ConnectionError) do
-      ShapeCleaner.remove_shapes_for_relations([oid_rel], stack_id: state.stack_id)
+      ShapeCleaner.remove_shapes_for_relations(
+        state.stack_id,
+        [oid_rel],
+        {:error, Electric.SnapshotError.from_error(error)}
```
msfstef, line 265: why convert this to a snapshot error in the publication manager?

if the shape was in the process of creating a snapshot, then `reply_to_relation_waiters` will return the error and the snapshot procedure will convert it to an appropriate error

if the shape _was not_ in the process of creating a snapshot, then we should propagate the error as is - if the issue is how we handle them at the API level I'd prefer we sort that out instead

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -119,7 +119,7 @@ defmodule Electric.ShapeCache do
   @impl Electric.ShapeCacheBehaviour
   def clean_shape(shape_handle, stack_id)
       when is_shape_handle(shape_handle) and is_stack_id(stack_id) do
-    ShapeCleaner.remove_shape(shape_handle, stack_id)
+    ShapeCleaner.remove_shape(stack_id, shape_handle)
```
msfstef, line 122: could you please make changes like argument switching that span the whole codebase in separate PRs or at least separate commits than the core changes you try to make? it makes it very confusing when reviewing and easy to miss things

ideally these changes are pure refactors and should not be mixed in the same PRs as the ones doing actual changes to how the service works

packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
```diff
@@ -74,6 +74,8 @@ defmodule Electric.ShapeCache.ExpiryManager do
     number_to_expire = shape_count - state.max_shapes
     shapes_to_expire = least_recently_used(state, number_to_expire)
 
+    latest = Enum.at(shapes_to_expire, -1)
```
msfstef: nit: can we add a doc somewhere (on `least_recently_used`) that guarantees this is ordered from less to most recent? We don't have a test around that functionality that I can see and this relies on that explicitly

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -1156,4 +1149,49 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+      raise File.Error, path: path, reason: :enoent
+    end
+  end
+
+  defp handle_open_error!(%__MODULE__{}, path, {:error, reason}) do
+    raise File.Error, path: path, reason: reason
+  end
+
+  defp shape_gone?(stack_id, shape_handle) do
+    case Electric.ShapeCache.ShapeStatus.get_existing_shape(stack_id, shape_handle) do
```
msfstef: rather than couple the storage to `ShapeStatus` and how those pointers work, I would much rather keep this encapsulated and it can check if the shape folder/shape definition file exists - it's an additional (cheap) syscall for edge cases but it keeps the storage entirely separate to our indexing and shape metadata store, I think it's worth it

since our deletes atomically remove the entire directory (logic that is encapsulated in storage), if a file is missing because the shape has been removed, the entire folder is also guaranteed to be missing.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -1156,4 +1149,49 @@ defmodule Electric.ShapeCache.PureFileStorage do
 
   @doc false
   def tmp_file(%__MODULE__{} = opts, filename), do: Path.join(tmp_dir(opts), filename)
+
+  @doc false
+  # Safely opens a shape file checking for shape presence if the file is not found
+  def open_file_stream(%__MODULE__{} = opts, path, modes, open_fun \\ &File.open/2, stream_fun)
```
msfstef:
```suggestion
  def open_file_stream!(%__MODULE__{} = opts, path, modes, open_fun \\ &File.open/2, stream_fun)
```
nit: consistent naming - on error we would raise, right?

msfstef: Also, I can't see any mention of this function being used anywhere in these changes

packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
```diff
@@ -138,28 +140,12 @@ defmodule Electric.ShapeCache.PureFileStorage.LogFile do
...
-  end
-
   @spec stream_entries(
           log_file_path :: String.t(),
           start_position :: non_neg_integer(),
           excl_end_pos :: non_neg_integer() | :eof
-        ) ::
-          Enumerable.t({log_item_with_sizes(), non_neg_integer()})
+        ) :: Enumerable.t({log_item_with_sizes(), non_neg_integer()})
+  # compaction only
```
msfstef: nit: comment above spec?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -1156,4 +1149,49 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+    case open_fun.(path, modes) do
+      {:ok, file} ->
+        stream_fun.(file)
+
+      {:error, _reason} = error ->
+        handle_open_error!(opts, path, error)
+    end
+  end
+
+  def stream_open_file!(%__MODULE__{} = opts, path, modes, open_fun \\ &File.open/2) do
```
msfstef: nit: this function does not "stream an open file" - it opens a file and streams it (?)

packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
```diff
@@ -188,42 +174,61 @@ defmodule Electric.ShapeCache.PureFileStorage.LogFile do
...
-        {:error, reason} ->
-          raise File.Error,
-            path: log_file_path,
-            reason: reason,
-            action: "pread(#{start_position}, #{end_position - start_position})"
-      end
-    after
-      File.close(file)
+    case stream_open_file!(opts, log_file_path, [:read, :raw]) do
+      {:halt, :shape_gone} ->
```
msfstef: the "stream open file" returning an error like `{:halt, :shape_gone}` feels like confusing API design which warrants at least some specs on them.

I would prefer if this returned something like `{:halt, :file_removed}` - this indicates that the file is not "missing", indicating a potential error, but that it was removed and we can handle that as we wish. It is transparent w.r.t. to what is happening and lets the API user handle it appropriately.

packages/sync-service/lib/electric/shape_cache/shape_cleaner.ex
```diff
@@ -4,134 +4,146 @@ defmodule Electric.ShapeCache.ShapeCleaner do
 
   This process ensures removing of shapes does not block critical path of shape creation.
   """
-  use GenServer
+
+  # use GenServer
```
msfstef: is this intentional?

This is no longer a process so might be worth updating the module doc with how it works and what it does

packages/sync-service/lib/electric/shape_cache/shape_cleaner.ex
```diff
@@ -4,134 +4,146 @@ defmodule Electric.ShapeCache.ShapeCleaner do
...
-    # schedule these shape removals one by one to avoid blocking the GenServer
-    # for too long to allow interleaved sync removals
+      remove_shapes(stack_id, affected_shapes, reason)
+    end)
+  end
 
-    new_queue = state.queued_removals ++ affected_shapes
+  def handle_writer_termination(_stack_id, _shape_handle, reason)
+      when reason in [:normal, :killed, :shutdown] or
+             (is_tuple(reason) and elem(reason, 0) == :shutdown) do
```
msfstef, line 89: should we not only be ok with `{:shutdown, :cleanup}` rather than all shutdown tuples? an unexpected shutdown tuple should fall under the abnormal shutdown (?)

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -65,13 +66,29 @@ defmodule Electric.Shapes.Consumer do
...
+    :ok
+  end
+
+  def stop(pid, reason) when is_pid(pid) do
+    # consumers call ShapeCleaner.remove_shape themselves, in which case a call
+    # here would deadlock because the consumer would be calling itself (and in
+    # fact the beam/elixir prevents a process calling itself) so instead we
+    # cast here
+    if pid == self() do
+      GenServer.cast(pid, {:stop, reason})
```
msfstef: this seems confusing to me and a slightly undefined behaviour of this API - we would expect a call to `stop` to return `:ok` once the process is dead, but if used by the process itself (indirectly via another API, in a circular way), then it returns `:ok` before the process is actually stopped.

This means that `ShapeCleaner.remove_shape` will start clearing the storage, remove the shape from the log collector, etc, while there might be more messages in the consumer message queue to handle before reaching the `:stop` cast that might interact with the storage etc

Why not exit here explicitly with a shutdown tuple we want to handle and let the `terminate` callback + `handle_writer_termination` by ShapeCleaner sort things out, terminate the writer properly, and delete the files in a more guaranteed order?

I appreciate the goal of having a single call to remove the shape, but it feels like this is a weird pattern to use for the sake of achieving that arbitrary goal - ultimately we're special casing the consumer removing itself anyway, why hide it in here rather than make it more explicit?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -560,22 +576,6 @@ defmodule Electric.Shapes.Consumer do
...
-  defp notify_shape_rotation(state) do
-    Registry.dispatch(
-      Electric.StackSupervisor.registry_name(state.stack_id),
-      state.shape_handle,
-      fn registered ->
-        Logger.debug(fn ->
-          "Notifying ~#{length(registered)} clients about removal of shape #{state.shape_handle}"
-        end)
-
-        for {pid, ref} <- registered, do: send(pid, {ref, :shape_rotation})
```
msfstef, line 572: if we're not longer notifying for shape rotation, how are live requesters notified that the shape has rotated now? we only seem to reply to snapshot waiters for the shape termination, would this not leave live requesters hanging for the entire 20 seconds despite the shape already having been deleted?

packages/sync-service/lib/electric/shape_cache/shape_cleaner.ex
magnetised: I'm planning ahead to something like `{:shutdown, :suspend}` or whatever we use for the "stop but don't clean" (preserving `:shutdown` for supervisor initiated stops)

packages/sync-service/lib/electric/core_supervisor.ex
```diff
@@ -47,7 +47,9 @@ defmodule Electric.CoreSupervisor do
 
     consumer_supervisor_spec = {Electric.Shapes.DynamicConsumerSupervisor, [stack_id: stack_id]}
 
-    shape_cleaner_spec = {Electric.ShapeCache.ShapeCleaner, stack_id: stack_id}
+    shape_cleaner_spec =
+      {Electric.ShapeCache.ShapeCleaner,
+       Keyword.merge(Keyword.get(tweaks, :shape_cleaner_opts, []), stack_id: stack_id)}
```
alco: This looks a bit heavy.

```suggestion
       [stack_id: stack_id] ++ Keyword.get(tweaks, :shape_cleaner_opts, [])}
```

packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
alco: Yeah. Another thing that's less obvious here is that `shapes_to_expire` is obtained by loading all entries from an ETS table into the process mem, sorting them and mapping. Now this adds another traversal of the whole list.

Since the shapes to expire are stored in a sorted ETS table, it'd be better to define a function on ShapeStatus that specifically returns the latest value.

packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
```diff
@@ -84,23 +86,13 @@ defmodule Electric.ShapeCache.ExpiryManager do
...
+        number_to_expire: number_to_expire,
+        elapsed_minutes_since_use: latest.elapsed_minutes_since_use
       ],
       fn ->
-        Electric.ShapeCache.ShapeCleaner.remove_shape(
-          shape.shape_handle,
-          state.stack_id
+        Electric.ShapeCache.ShapeCleaner.remove_shapes(
+          state.stack_id,
+          Enum.map(shapes_to_expire, & &1.shape_handle)
```
alco: Aha! Clever optimisation. Take it just one step further and change the return value of `ShapeStatus.least_recently_used` to only return shape handles. Then we won't need this `Enum.map()` here.

packages/sync-service/lib/electric/shape_cache/shape_cleaner.ex
```diff
@@ -4,134 +4,146 @@ defmodule Electric.ShapeCache.ShapeCleaner do
...
 
   require Logger
 
   @type shape_handle() :: Electric.ShapeCacheBehaviour.shape_handle()
+  @type stack_id() :: Electric.stack_id()
 
-  @schema NimbleOptions.new!(stack_id: [type: :string, required: true])
+  def child_spec(args) do
+    CleanupTaskSupervisor.child_spec(args)
+  end
```
alco: Instead of obscuring which process gets started like this, we could just pass `CleanupTaskSupervisor` to whichever supervisor is the parent of this.

packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
msfstef: We're also doing an `:ets.tab2list` there which I think is an unnecessary memory spike - I might open a PR to do a fold over the table and build an ordered list

I agree with having an API for the latest as the table is ordered and thus the latest should be O(1) cheap to grab

packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
magnetised: I *knew* you'd hate this... 

Because we no longer wait for requests to a given shape to terminate before cleaning the shape, the `remove_shapes_for_relations` is actually able to terminate the consumer before this error propagates back to it from the snapshotter, which means that the consumer terminates with `the consumer exited before completing the snapshot` (or whatever the message is in `Consumer.terminate/2` in the `reply_to_snapshot_waiters` call).

So instead of passing this generic error back to the requests, I intercept a bit of the snapshotter error handling code in order to return the same error that the snapshotter would have returned and thus returning a more useful error message back to the client.

It's not pretty but it maintains the current semantics without gymnastics (I'm a poet, I know it). Rather than figure out how to do this I think the better solution is the one I proposed on Discord: validate that we're able to add a table to the publication before even getting to the point of starting a consumer.

I think this little hack seems like a reasonable compromise for the moment I feel. The alternative would involve a lot of fragile coordination.

packages/sync-service/lib/electric/async_deleter.ex
magnetised: Me not claude unfortunately. I was documenting for myself I think. Will convert to a spec.

packages/sync-service/lib/electric/shape_cache.ex
magnetised: Since I was refactoring the shapecleaner (and use of the shapecleaner), reordering the args felt very much like part of that.

packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
magnetised: I don't share any particular performance concerns here -- iterating a list is pretty fast -- but can certainly refactor to remove the need.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
magnetised: That is a good idea. It does tie the test to the exact functionality of the AsyncDeleter but I guess that's a pretty stable bet

packages/sync-service/lib/electric/shape_cache/shape_cleaner.ex
magnetised: OOps. missed that.

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: Yes, good point. I was reluctant to mess with the existing termination and did want to keep the single consistent call but you're right.

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: I moved this to the shape cleaner.

packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
magnetised: I don't think performance is critical here -- it's run every 30 seconds or something -- but can certainly remove all the iterations which might also give us a memory win

packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
msfstef: I see - I appreciate the rhyme and the argument. Happy to keep it in and agree that it would be better to move the whole thing "earlier" and we catch two birds with one stone

packages/sync-service/lib/electric/shapes/consumer.ex
msfstef: Ah I missed it entirely sorry - having that notification as an async cleanup task might be a bit confusing but it makes sense to defer it a bit. Would be happy with just some more moduledocs for what responsibilities each thing has

packages/sync-service/lib/electric/shape_cache/expiry_manager.ex
alco: We can handle this in a follow-up PR as Stefanos has already put his eye on it.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
magnetised: it opens a file for a stream... I'll rename to `safe_open_file` or something

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -867,7 +849,9 @@ defmodule Electric.ShapeCache.PureFileStorage do
     snapshot_started = Keyword.get(metadata, :snapshot_started?) || false
     last_snapshot_chunk = Keyword.get(metadata, :last_snapshot_chunk)
 
-    if not snapshot_started, do: raise(Storage.Error, message: "Snapshot not started")
+    if !snapshot_started && !shape_gone?(opts) do
```
msfstef:
```suggestion
    if not snapshot_started and not shape_gone?(opts) do
```
nit: I tend to prefer `not` `and` `or` for boolean operations, they raise argument errors for non bool args and IMO the intention reads more clearly

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -1156,4 +1151,38 @@ defmodule Electric.ShapeCache.PureFileStorage do
 
   @doc false
   def tmp_file(%__MODULE__{} = opts, filename), do: Path.join(tmp_dir(opts), filename)
+
+  @type file_opener() :: (Path.t(), [File.mode()] ->
+                            {:ok, File.file_descriptor()} | {:error, File.posix()})
+  @doc false
```
msfstef, line 1157: out of curiosity - why `@doc false` and not have this as part of the doc?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -793,10 +801,28 @@ defmodule Electric.Shapes.Consumer do
       {_, true} -> {:noreply, state}
       {{:shutdown, _}, false} -> {:stop, reason, state}
       {:shutdown, false} -> {:stop, reason, state}
-      _ -> {:noreply, terminate_safely(state)}
+      _ -> stop_and_clean(state)
     end
   end
 
+  defp stop_and_clean(state) do
```
msfstef: nit: for the sake of locality of code it would be nice if the `stop_with_reason`, `mark_for_removal` were close together so all removal code is easy to read

msfstef (approved):
Beautiful! Nice to see removals becoming less and less blocking!

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
magnetised: `@doc false` means not part of the public API which this function definitely isn't.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
magnetised: I mis-remembered or mis-read some docs which recommended `||` etc. I see you're right

PR #3425, opened 2025-11-12
chore: Update `otel_telemetry_exporter`

msfstef:
Updating the dep to include changes I made over there.

Running `mix deps.get` also seems to have updated `protobuf` a minor version - I think you had an issue with that @alco or is this update ok to merge in? If not I can roll that one back.


alco:
So, I updated protobuf in otel_metric_exporter because it's generally a good thing to keep deps up to date. But then a problem appeared when building Electric'a Docker image: it fails because protobuf 0.14+ and protox (brought in by pg_query_ex) have overlapping modules. I think this problem arises when assembling an Elixir release.

I solved it in my PR by pinning protobuf version in Electric, since that's where the problem is.

msfstef:
@alco thanks for the pointer I remembered roughly that you had an issue with it but not exactly what - I've pinned it in this PR as well just to unblock getting those changes in

alco (approved):

PR #3427, opened 2025-11-13
feat: proper move ins/outs for subqueries without 409s

### LINKED ISSUE #3330: Basic move support, with single subquery

icehaunter:
Closes #3330

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
@@ -149,6 +149,11 @@ defmodule Electric.Shapes.Consumer.Materializer do
           end
 
         %{"headers" => %{"event" => "move-out", "patterns" => patterns}} ->
+          patterns =
+            Enum.map(patterns, fn %{"pos" => pos, "value" => value} ->
+              %{pos: pos, value: value}
+            end)
+
           %{headers: %{event: "move-out", patterns: patterns}}
       end)
       |> apply_changes(state)
@@ -375,21 +380,21 @@ defmodule Electric.Shapes.Consumer.Materializer do
 
   defp add_row_to_tag_indices(tag_indices, key, move_tags) do
     # For now we only support one move tag per row (i.e. no `OR`s in the where clause if there's a subquery)
-    Enum.reduce(move_tags, tag_indices, fn [val1], acc ->
-      Map.update(acc, val1, MapSet.new([key]), &MapSet.put(&1, key))
+    Enum.reduce(move_tags, tag_indices, fn tag, acc when is_binary(tag) ->
+      Map.update(acc, tag, MapSet.new([key]), &MapSet.put(&1, key))
     end)
   end
 
   defp remove_row_from_tag_indices(tag_indices, key, move_tags) do
-    Enum.reduce(move_tags, tag_indices, fn [val1], acc ->
-      case Map.fetch(acc, val1) do
+    Enum.reduce(move_tags, tag_indices, fn tag, acc when is_binary(tag) ->
+      case Map.fetch(acc, tag) do
         {:ok, v} ->
           new_mapset = MapSet.delete(v, key)
 
           if MapSet.size(new_mapset) == 0 do
-            Map.delete(acc, val1)
+            Map.delete(acc, tag)
           else
-            Map.put(acc, val1, new_mapset)
+            Map.put(acc, tag, new_mapset)
           end
 
         :error ->
@@ -400,8 +405,9 @@ defmodule Electric.Shapes.Consumer.Materializer do
 
   defp pop_keys_from_tag_indices(tag_indices, patterns) do
     # This implementation is naive while we support only one tag per row and no composite tags.
-    Enum.reduce(patterns, {MapSet.new(), tag_indices}, fn [val1], {keys, acc} ->
-      case Map.pop(acc, val1) do
+    Enum.reduce(patterns, {MapSet.new(), tag_indices}, fn %{pos: _pos, value: value},
+                                                          {keys, acc} ->
+      case Map.pop(acc, value) do
         {nil, acc} -> {keys, acc}
         {v, acc} -> {MapSet.union(keys, v), acc}
       end
--- a/packages/sync-service/lib/electric/shapes/consumer/move_handling.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/move_handling.ex
@@ -5,7 +5,7 @@ defmodule Electric.Shapes.Consumer.MoveHandling do
   alias Electric.Shapes.PartialModes
   alias Electric.Shapes.Shape
   alias Electric.Shapes.Shape.SubqueryMoves
-  alias Electric.Shapes.Consumer.MoveHandlingState
+  alias Electric.Shapes.Consumer.MoveIns
 
   @spec process_move_ins(State.t(), Shape.handle(), list(term())) :: State.t()
   def process_move_ins(state, _, []), do: state
@@ -42,7 +42,7 @@ defmodule Electric.Shapes.Consumer.MoveHandling do
       )
 
     move_handling_state =
-      MoveHandlingState.add_waiting_move_in(state.move_handling_state, name, pg_snapshot)
+      MoveIns.add_waiting(state.move_handling_state, name, pg_snapshot)
 
     %{state | move_handling_state: move_handling_state}
   end
@@ -53,14 +53,21 @@ defmodule Electric.Shapes.Consumer.MoveHandling do
 
   def process_move_outs(state, dep_handle, removed_values) do
     message =
-      SubqueryMoves.make_move_out_control_message(state.shape, [{dep_handle, removed_values}])
+      SubqueryMoves.make_move_out_control_message(
+        state.shape,
+        state.stack_id,
+        state.shape_handle,
+        [
+          {dep_handle, removed_values}
+        ]
+      )
 
     {{_, upper_bound}, writer} = Storage.append_control_message!(message, state.writer)
 
     {%{state | writer: writer}, {[message], upper_bound}}
   end
 
-  @spec query_complete(State.t(), MoveHandlingState.move_in_name(), list(String.t())) ::
+  @spec query_complete(State.t(), MoveIns.move_in_name(), list(String.t())) ::
           {State.t(), notification :: term()}
   def query_complete(%State{} = state, name, key_set) do
     # 1. Splice the stored data into the main log
@@ -69,7 +76,7 @@ defmodule Electric.Shapes.Consumer.MoveHandling do
 
     # 2. Remove this entry from "waiting" and add to "filtering"
     move_handling_state =
-      MoveHandlingState.change_move_in_to_filtering(state.move_handling_state, name, key_set)
+      MoveIns.change_to_filtering(state.move_handling_state, name, key_set)
 
     state = %{state | move_handling_state: move_handling_state, writer: writer}
 
--- a/packages/sync-service/lib/electric/shapes/consumer/state.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/state.ex
@@ -1,7 +1,7 @@
 defmodule Electric.Shapes.Consumer.State do
   @moduledoc false
-  alias Electric.Shapes.Consumer.MoveHandlingState
-  alias Electric.Shapes.Consumer.InitialSnapshotState
+  alias Electric.Shapes.Consumer.MoveIns
+  alias Electric.Shapes.Consumer.InitialSnapshot
   alias Electric.Shapes.Shape
   alias Electric.Replication.Changes.Transaction
   alias Electric.Postgres.SnapshotQuery
@@ -18,8 +18,8 @@ defmodule Electric.Shapes.Consumer.State do
     :latest_offset,
     :storage,
     :writer,
-    initial_snapshot_state: InitialSnapshotState.new(nil),
-    move_handling_state: MoveHandlingState.new(),
+    initial_snapshot_state: InitialSnapshot.new(nil),
+    move_handling_state: MoveIns.new(),
     buffer: [],
     txn_offset_mapping: [],
     materializer_subscribed?: false,
@@ -116,11 +116,11 @@ defmodule Electric.Shapes.Consumer.State do
   #       }
 
   defguard is_snapshot_started(state)
-           when is_struct(state.initial_snapshot_state, InitialSnapshotState) and
+           when is_struct(state.initial_snapshot_state, InitialSnapshot) and
                   state.initial_snapshot_state.snapshot_started?
 
   defguard needs_initial_filtering(state)
-           when is_struct(state.initial_snapshot_state, InitialSnapshotState) and
+           when is_struct(state.initial_snapshot_state, InitialSnapshot) and
                   state.initial_snapshot_state.filtering?
 
   @spec new(Electric.stack_id(), Shape.handle(), Shape.t()) :: uninitialized_t()
@@ -156,7 +156,7 @@ defmodule Electric.Shapes.Consumer.State do
         do: LogOffset.last_before_real_offsets(),
         else: latest_offset
 
-    initial_snapshot_state = InitialSnapshotState.new(pg_snapshot)
+    initial_snapshot_state = InitialSnapshot.new(pg_snapshot)
 
     %__MODULE__{
       state
@@ -164,7 +164,7 @@ defmodule Electric.Shapes.Consumer.State do
         storage: storage,
         writer: writer,
         initial_snapshot_state: initial_snapshot_state,
-        buffering?: InitialSnapshotState.needs_buffering?(initial_snapshot_state)
+        buffering?: InitialSnapshot.needs_buffering?(initial_snapshot_state)
     }
   end
 
@@ -195,7 +195,7 @@ defmodule Electric.Shapes.Consumer.State do
   def add_waiter(%__MODULE__{initial_snapshot_state: initial_snapshot_state} = state, from) do
     %{
       state
-      | initial_snapshot_state: InitialSnapshotState.add_waiter(initial_snapshot_state, from)
+      | initial_snapshot_state: InitialSnapshot.add_waiter(initial_snapshot_state, from)
     }
   end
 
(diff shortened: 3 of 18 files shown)


packages/sync-service/lib/electric/replication/log_offset.ex
```diff
@@ -212,7 +212,13 @@ defmodule Electric.Replication.LogOffset do
       iex> compare(new(10, 1) |> increment(4), new(10, 5))
       :eq
   """
-  def increment(%LogOffset{op_offset: op_offset} = log_offset, increment \\ 1) do
+  def increment(log_offset, increment \\ 1)
+
+  def increment(%LogOffset{op_offset: :infinity} = log_offset, increment) when increment > 0 do
+    %{log_offset | tx_offset: 1, op_offset: increment - 1}
+  end
```
robacourt, line 219: This reads as ♾️ + 1 = 0, but I'm guessing this is no longer an increment function and something more akin to a `next` function so perhaps it should be named as such (perhaps keeping and calling the increment for the non infinity case). Also tx_offset gets set to 1 regardless of what it was before, presumably because the assumption is infinity is only on tx_offset 0, so perhaps that should be asserted. Or perhaps this isn't a general function of LogOffset at all and should just be kept in the module where it's used.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -850,6 +853,71 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+               <<json::binary-size(json_size)>> <- IO.binread(file, json_size) do
+            {[{offset, key_size, key, op_type, 0, json_size, json}],
+             {file, LogOffset.increment(offset)}}
+          else
+            {:meta, :eof} ->
+              {:halt, {file, offset}}
+
+            _ ->
+              raise Storage.Error,
+                message: "Incomplete move-in snapshot file at #{shape_snapshot_path(opts, name)}"
```
robacourt, line 945: Shouldn't we be logging what the error was for ease of debugging?

packages/sync-service/lib/electric/shapes/consumer/move_handling.ex
```diff
@@ -0,0 +1,72 @@
...
+
+  @spec query_complete(State.t(), State.move_in_name(), list(String.t())) ::
+          {State.t(), notification :: term()}
+  def query_complete(state, name, key_set) do
+    # 1. Splice the stored data into the main log
+    {{_, upper_bound} = bounds, writer} =
+      Storage.append_move_in_snapshot_to_log!(name, state.writer)
+
+    # 2. Remove this entry from "waiting" and add to "filtering"
+    state = State.change_move_in_to_filtering(%{state | writer: writer}, name, key_set)
```
robacourt: Consumer.State seems like a pretty odd abstraction to me. Could move handling not have it's own state, and then the consumer could contain that state?

icehaunter: I was annoyed at the amount of keys getting assigned and the buffering logic becoming less clear. I wanted smaller public functions with docs that handle state transitions

robacourt: That doesn't really answer my question. But maybe I haven't explained myself well. Couldn't move handling have a `%MoveHandling{}` state struct with just what it needs from `Consumer.State` and then in Consumer:

```
{move_handling, notification} =
      state.move_handling
      |> MoveHandling.process_move_ins(dep_handle, move_in)
      |> MoveHandling.process_move_outs(dep_handle, move_out)

    notify_new_changes(%{state | move_handling: move_handling}, notification)
 ```
 This would decouple the move handling from the rest of the consumer logic and make move handling more cohesive.

File: packages/sync-service/lib/electric/replication/changes.ex [hunk 4]
```diff
@@ -107,7 +127,8 @@ defmodule Electric.Replication.Changes do
       :log_offset,
       :key,
       :old_key,
-      tags: [],
+      move_tags: [],
```
msfstef, line 155: does this change from `tags` to `move_tags` break anything? I assume this was still undocumented/experimental?

packages/sync-service/lib/electric/replication/eval.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Electric.Replication.Eval do
+  def type_to_pg_cast({:array, type}), do: "#{type_to_pg_cast(type)}[]"
+  def type_to_pg_cast({:enum, name}), do: to_string(name)
+  def type_to_pg_cast({:row, _}), do: raise("Unsupported type: row")
+  def type_to_pg_cast({:internal, _}), do: raise("Unsupported type: internal")
+  def type_to_pg_cast(type), do: to_string(type)
```
msfstef: nit: worth a doc + doctest perhaps - also, should it not raise for arbitrary tuples? I assume this last clause is intended to match any "simple" type it would be nice if it was honing in on that

packages/sync-service/lib/electric/replication/eval/env.ex
```diff
@@ -170,6 +170,27 @@ defmodule Electric.Replication.Eval.Env do
     end
   end
 
+  def const_to_pg_string(%__MODULE__{}, value, type) when type in @text_types, do: value
```
msfstef, line 173: nit: would appreciate type spec/small doc/doctest on public api

packages/sync-service/lib/electric/replication/log_offset.ex
msfstef: I personally don't mind this too much, but would like to see a doctest for this case in particular

the ♾️ is really just a stand-in for the "supremum" of a given `tx_offset`, not an actual infinity, so adding one makes sense in that it moves to the next `tx_offset`. I see the point in calling this `next` but probably worth doing separately in a single PR.

That being said, I think I would prefer if this was also overspecified to the `last_before_real_offsets` which has `tx_offset` of 0 rather than any `tx_offset` - the only times we use infinity is for the supremum of the virtual offsets (which can move to the real offsets), and the last of the real offsets, which _should not_ move on to the next tx offset.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -850,6 +853,71 @@ defmodule Electric.ShapeCache.PureFileStorage do
     write_cached_metadata!(opts, :last_snapshot_chunk, LogOffset.new(0, last_chunk_num))
   end
 
+  def write_move_in_snapshot!(stream, name, %__MODULE__{} = opts) do
+    path = shape_snapshot_path(opts, name)
+    File.mkdir_p!(shape_snapshot_dir(opts))
```
msfstef:
```suggestion
    FileInfo.mkdir_p!(shape_snapshot_dir(opts))
```
This call goes through the file server - but also, do we think it is necessary? shouldn't that directory exist by virtue of the writer having been initialized?

I think we should try and not be too defensive if it saves us a few syscalls

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -850,6 +853,71 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+
+    stream
+    |> Stream.map(fn [key, json] ->
+      [<<byte_size(key)::32>>, <<byte_size(json)::64>>, ?i, key, json]
+    end)
+    |> Stream.into(File.stream!(path))
```
msfstef:
```suggestion
    |> Stream.into(File.stream!(path, [:delayed_write]))
```
is there a reason to avoid delayed writes in this case? I suppose we want to ensure everything is written, but the docs say:

>The buffered data is also flushed before some other file operation than [write/2](https://www.erlang.org/doc/apps/kernel/file.html#write/2) is executed.

So by the moment we want to actually "append" this to the log it will have been sorted out (?) there should be no other concurrent readers of it like there are of the main log

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -850,6 +853,71 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+              {:halt, {file, offset}}
+
+            _ ->
+              raise Storage.Error,
+                message: "Incomplete move-in snapshot file at #{shape_snapshot_path(opts, name)}"
+          end
+        end,
+        fn {file, _} ->
+          File.close(file)
+          File.rm!(shape_snapshot_path(opts, name))
```
msfstef:
```suggestion
          :file.delete(shape_snapshot_path(opts, name))
```
This call goes through the file server

packages/sync-service/lib/electric/shapes/partial_modes.ex
```diff
@@ -1,30 +1,85 @@
...
+  alias Electric.Shapes.Shape
   alias Electric.Postgres.Lsn
   alias Electric.Shapes.Querying
   alias Electric.Connection.Manager
+  alias Electric.Postgres.SnapshotQuery
 
-  def query_subset(shape, subset, opts) do
+  def query_subset(shape_handle, %Shape{} = shape, subset, opts) do
     pool = Manager.pool_name(opts[:stack_id], :snapshot)
+    mark = Enum.random(1..(2 ** 31))
```
msfstef:
```suggestion
    mark = Enum.random(0..(2 ** 31 - 1))
```
nit: ensuring the number is in 0-indexed 31 bit precision

packages/sync-service/lib/electric/shapes/partial_modes.ex
```diff
@@ -1,30 +1,85 @@
...
+        Querying.query_subset(conn, shape, subset, headers)
+        |> Enum.to_list()
+      end,
+      stack_id: opts[:stack_id],
+      query_reason: "subset_query"
+    )
+    |> case do
+      {:ok, result} ->
+        metadata =
+          receive(
```
msfstef: should this become an non-blocking receive with 0 timeout and a raise if the message is not present? if the function has returned this message _should_ already be in the mailbox

packages/sync-service/lib/electric/shapes/partial_modes.ex
```diff
@@ -1,30 +1,85 @@
...
+            send(parent, {:query_move_in_complete, opts[:move_in_name], result})
+          end
+        )
+      rescue
+        error ->
+          send(parent, {:query_move_in_error, opts[:move_in_name], error, __STACKTRACE__})
+      end
     end)
+
+    receive do
```
msfstef, line 81: nit: should this be given a timeout with a raise and a task shutdown for safety?

packages/sync-service/lib/electric/shapes/querying.ex
```diff
@@ -126,26 +154,25 @@ defmodule Electric.Shapes.Querying do
...
     headers =
-      case Keyword.get(opts, :snapshot_mark) do
-        nil -> headers
-        mark -> Map.put(headers, :snapshot_mark, mark)
-      end
+      headers
+      |> Map.merge(additional_headers)
       |> Jason.encode!()
       |> Utils.escape_quotes(?')
+      |> String.replace(~r/\$\$\{(\w+)\}/, ~s[' || \\1::text || '])
```
msfstef: not entirely sure what this is doing (?)

packages/sync-service/lib/electric/shapes/shape/subquery_moves.ex
```diff
@@ -0,0 +1,101 @@
+defmodule Electric.Shapes.Shape.SubqueryMoves do
+  alias Electric.Replication.Eval
+  alias Electric.Replication.Eval.Walker
+  alias Electric.Shapes.Shape
+
+  def move_in_where_clause(
```
msfstef, line 30: would appreciate a docstring here (or maybe a module doc) explaining the strategy for building the move in where clauses - there are a couple of steps here that make sense in the wider move in logic but reading it in isolation is very hard to parse and guide understanding (e.g. the shape handle provided is that of a dependency of the provided shape, and the move_ins are for that given shape handle, and this is the where clause to populate with results as a result of this move in)

packages/sync-service/lib/electric/shapes/consumer/move_handling.ex
robacourt: Claude came up with some nice ideas on how to break things up in different areas of responsibility, and reduce State to a Context that is just a data structure: https://github.com/electric-sql/electric/compare/main...claude/refactor-move-handling-017uQXxZHdB3FBEmm7QMYXh1?expand=1

packages/sync-service/lib/electric/shapes/shape/subquery_moves.ex
```diff
@@ -0,0 +1,101 @@
...
+  """
+  @spec make_move_out_control_message(Shape.t(), [
+          {dep_handle :: String.t(), gone_values :: String.t()},
+          ...
+        ]) :: map()
+  # Stub guard to allow only one dependency for now.
+  def make_move_out_control_message(shape, [_] = move_outs) do
+    %{
+      headers: %{
+        event: "move_out",
```
msfstef: nit: so far we have used dash separated names like `must-refetch` and `up-to-date` in our control messages - I realise this is not exactly the same header but it'd be good to keep the same pattern

packages/sync-service/lib/electric/shapes/shape/subquery_moves.ex
```diff
@@ -0,0 +1,101 @@
...
+    # Patterns are a list of tuples (represented as lists for JSON serialization)
+    Enum.map(gone_values, &List.wrap/1)
+  end
+
+  def move_in_tag_structure(%Shape{} = shape)
+      when is_nil(shape.where)
+      when shape.shape_dependencies == [],
+      do: []
+
+  def move_in_tag_structure(shape) do
```
msfstef, line 139: here as well I would really appreciate a doc describing what the structure will look like - this looks like it builds a list with one element, that is a list of either column names or lists of column names

I've followed the discussion so I understand that this is what the tags will look like, but following this code is very unintuitive without a description of what I should expect

packages/sync-service/lib/electric/shapes/consumer/snapshotter.ex
```diff
@@ -210,116 +210,59 @@ defmodule Electric.Shapes.Consumer.Snapshotter do
...
+            # available in `Stream.transoform/5`.
+            acc
           end
         )
-      end,
-      timeout: :infinity
-    )
-  catch
-    :exit, {_, {DBConnection.Holder, :checkout, _}} ->
-      raise SnapshotError.connection_not_available()
```
msfstef, line 288: this has been removed and as far as I can tell has not been put in place anywhere again (?)

packages/sync-service/lib/electric/shapes/consumer/move_handling.ex
msfstef: reading through the code I agree with rob's point - it feels like the move in logic could be further encapsulated into a module with its own state and set of public APIs

given the expansion of responsibilities of the consumer I prefer a tradeoff of verbosity for the sake of clarity of responsibility (and testability)

packages/sync-service/lib/electric/shapes/consumer/state.ex
```diff
@@ -0,0 +1,304 @@
...
+          filtering_move_ins: list(Shape.handle()),
+          move_in_buffering_snapshot: nil | pg_snapshot(),
+          hibernate_after: non_neg_integer(),
+          latest_offset: nil,
+          initial_pg_snapshot: nil,
+          storage: nil,
+          writer: nil
+        }
+
+  @typedoc """
```
msfstef, line 57: really appreciate the documentation here (and throughout this module) - out of curiosity, why a type doc and not a module doc for this?

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -169,16 +185,25 @@ defmodule Electric.Shapes.Consumer.Materializer do
...
-        send(pid, {:materializer_changes, state.shape_handle, events})
+  def handle_call({:new_changes, {range_start, range_end}}, _from, state) do
+    stack_storage = Storage.for_stack(state.stack_id)
+    storage = Storage.for_shape(state.shape_handle, stack_storage)
+
+    Storage.get_log_stream(range_start, range_end, storage)
+    |> Stream.map(&Jason.decode!/1)
+    |> Stream.map(fn %{"key" => key, "value" => value, "headers" => %{"operation" => operation}} ->
+      case operation do
+        "insert" -> %Changes.NewRecord{key: key, record: value}
```
msfstef, line 201: do these not also need to carry over any tags in the log? what makes this logic different than the initial stream read in the `{:read_stream, storage}` continue?

packages/sync-service/lib/electric/shapes/consumer/initial_snapshot_state.ex
```diff
@@ -0,0 +1,94 @@
+defmodule Electric.Shapes.Consumer.InitialSnapshotState do
```
robacourt: This could just be `InitialSnapshot`, no? It's a bit confusing having all these thing called `State`

packages/sync-service/lib/electric/shapes/consumer/move_handling_state.ex
```diff
@@ -0,0 +1,116 @@
+defmodule Electric.Shapes.Consumer.MoveHandlingState do
```
robacourt: Perhaps this could be just `MoveIns`? That highlights it's just to do with move ins (not move outs) and it's less confusing than having another module called state. It would also allow you to remove `move_in` from the function names

packages/sync-service/lib/electric/replication/changes.ex
icehaunter: It was a layover from last-last-last electric, and was an unused field

packages/sync-service/lib/electric/shapes/consumer/move_handling.ex
icehaunter: I've done the encapsulation, hope it's better now

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: It kind of doesn't matter which error it is, as any incomplete file is just wrong. I don't think this can be meaningfully more precise

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: Yup, you're right, just missed the flag

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
icehaunter: `:file.delete` does too, for the record, but I've addressed

packages/sync-service/lib/electric/shapes/partial_modes.ex
icehaunter: I don't think so, inner functions already handle timings & crashes

packages/sync-service/lib/electric/shapes/consumer/state.ex
icehaunter: I avoid moduledoc because it's then visible in the hexdocs for the package, and I don't like that :D

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
icehaunter: Yup, I've addressed that

packages/sync-service/lib/electric/replication/log_offset.ex
robacourt: supremium or infinity, they both have the same issue, it's not a real increment. i still think `next` makes more sense but happy for that to be addressed in a separate PR

Issue #3414, opened 2025-11-11
Send individual changes to consumers


robacourt:
Part of #3413 . The shape log collector should send individual changes, transaction starts and commits to the shapes that they affect. The consumers can then buffers to form a filtered transaction from the changes and then process as normal. This will mean a consumers working memory is not expanded to the size of the full transaction when the shape only needs a subset of it. It also paves the way for the rest of #3413 

Note: For shapes that depend on subqueries this will mean the Filter may have out of date materialised views at the point of sending the change to the consumer (as it does now) however this does not matter as:
1. For move ins the missing data will be fetched from Postgres
2. For move outs, the extra data should be filtered out by the consumer (as it does now)

Commit messages should be sent in dependency layers (as the transactions are done now) and not proceed to the next layer until a response from the commit has been received. Only commit messages will have responses, other changes are just cast to the consumer. The ShpeLogCollector will not proceed to the next transaction until all responses (or exit signals) to all commit messages sent have been received.

Transaction start and commit messages will need to be sent to all consumers who receive at least 1 change from the transaction.

Performance should not be compromised by this change, we may need to buffer messages to not bare the cost of the additional context switching from sending multiple messages rather than a whole transaction in 1 message.

PR #3429, opened 2025-11-13
feat(sync-service): Send individual operations to ShapeLogCollector

robacourt:
Part of #3414, this PR sends individual operations from the ReplicationClient to ShapeLogCollector. This should reduce the working memory used by the ReplicationClient, but the ultimate aim is to send individual operations to storage and so not buffer at any point (unless subqueries are involved). This PR just moves the transaction consolidation one step, from ReplicationClient to ShapeLogCollector to keep the PR small. Other PRs will move the consolidation further, from ShapeLogCollector to Consumer, then finally from to Consumer to storage (except with subqueries are involved).

Currently the replication stream take two forms:
1. `Replication.Messages` such as `Messages.Begin`, `Messages.Commit`, `Messages.Insert` etc.
2. `Changes.Transaction` (or `Changes.Relation`)

And `Collector` converts one to the other.

This PR introduces an intermediate form, Operations, as Messages are coupled to the postgres replication protocol and are not optimised for use with Shapes (for example they do not include the relation name) and Transactions are whole transactions. So with this PR we have:
1. `Replication.Messages` such as `Messages.Begin`, `Messages.Commit`, `Messages.Insert` etc.
2. Operation (spec: `Changes.operation()`) such as `Changes.Begin`. `Changes.Commit`, `Changes.NewRecord` etc.
4. `Changes.Transaction` (or `Changes.Relation`) which I'm calling database actions (spec: `Changes.action()`)

`Collector` has been replaced by `MessageConverter` that converts 1 to 2 and `TransactionBuilder` that coverts 2 to 3.

Benchamarking this PR validated @icehaunter 's hypothesis that extra messages would slow down small transactions, so this PR also includes a `OperationBatcher` that batches operations, which the benchmarks confirm eliminate the slow down.

### DIFF
--- a/packages/sync-service/lib/electric/postgres/replication_client.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client.ex
@@ -107,6 +107,13 @@ defmodule Electric.Postgres.ReplicationClient do
                    max_txn_size: [type: {:or, [:non_neg_integer, nil]}, default: nil]
                  )
 
+    # Making the batch size small results in more message passing which
+    # can have a performance impact. The larger the batch size the more memory
+    # is used to hold the operations in memory before sending them off to be processed.
+    # For local testing batch sizes of 3 and above overcome the performance hit of message
+    # passing, but as we're not currently worried about the memory consumption of the
+    # replication client and don't want to risk any performance degradation in production
+    # it has been set arbitrarily high to 100. We can tune this figure later if needed.
     @default_max_operation_batch_size 100
 
     @spec new(Access.t()) :: t()
--- a/packages/sync-service/lib/electric/postgres/replication_client/message_converter.ex
+++ b/packages/sync-service/lib/electric/postgres/replication_client/message_converter.ex
@@ -1,7 +1,13 @@
 defmodule Electric.Postgres.ReplicationClient.MessageConverter do
   @moduledoc """
   Conversion of incoming Postgres logical replication messages
-  to internal change representation.
+  to internal operatrion representation.
+
+  It is stateful, consuming the replication messages in sequential order
+  to keep track of the relation and transaction information needed
+  to form the operations.
+
+  It also enforces a maximum transaction size if configured to do so.
   """
 
   require Logger


packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -118,7 +117,12 @@ defmodule Electric.Postgres.ReplicationClient do
 
       struct!(
         __MODULE__,
-        opts ++ [txn_collector: %Collector{max_tx_size: max_txn_size}]
+        opts ++
+          [
+            message_converter:
+              MessageConverter.new(max_tx_size: max_txn_size, stack_id: opts[:stack_id]),
```
msfstef, line 132: nit: should we be grabbing stack id with a `fetch!` operation?

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -108,6 +105,8 @@ defmodule Electric.Postgres.ReplicationClient do
                    max_txn_size: [type: {:or, [:non_neg_integer, nil]}, default: nil]
                  )
 
+    @max_operation_batch_size 3
```
msfstef: nit: worth adding a comment on how this number was decided (I'm assuming via benchmarks) and why we batch in the first place

packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -352,120 +356,63 @@ defmodule Electric.Postgres.ReplicationClient do
 
   def handle_data(
         <<@repl_msg_x_log_data, _wal_start::64, _server_wal_end::64, _clock::64, data::binary>>,
-        %State{stack_id: stack_id} = state
+        %State{} = state
       ) do
-    data
-    |> OpenTelemetry.start_interval("replication_client.decode_message")
```
msfstef, line 358: are we getting rid of this timer? were we not using it to begin with?

robacourt:
benchmark this

packages/sync-service/lib/electric/postgres/replication_client/message_converter.ex
```diff
@@ -0,0 +1,261 @@
+defmodule Electric.Postgres.ReplicationClient.MessageConverter do
+  @moduledoc """
+  Conversion of incoming Postgres logical replication messages
+  to internal change representation.
```
msfstef: might be worth mentioning that even though this is a converter, it is stateful and aware of transaction boundaries and collects information (e.g. to track max transaction size)

packages/sync-service/lib/electric/postgres/replication_client/message_converter.ex
```diff
@@ -0,0 +1,261 @@
...
+      | tx_size: state.tx_size + bytes,
+        change_count: state.change_count + 1,
+
+        # We're adding 2 to the op index because it's possible we're splitting some of the operations before storage.
+        # This gives us headroom for splitting any operation into 2.
+        tx_op_index: state.tx_op_index + 2
+    }
+  end
+
+  defp log_txn_received(%LR.Commit{} = msg, %__MODULE__{} = state) do
```
msfstef, line 249: I don't love the idea of the converter being responsible for logging a received transaction - the process actually receiving it should

perhaps we can have an API to return "current txn stats" that the replication client can use to log things? or when a commit message is converted it emits an extra piece of info that can be used for logging?

packages/sync-service/lib/electric/replication/operation_batcher.ex
```diff
@@ -0,0 +1,46 @@
...
+  @spec new(non_neg_integer()) :: t()
+  def new(max_batch_size) do
+    %__MODULE__{max_batch_size: max_batch_size, buffer: [], flush?: false}
+  end
+
+  @doc """
+  Batch operations according to the following rules:
+  - A batch is flushed when a Commit operation is seen
+  - A batch is flushed when a Relation operation is seen
+  - A batch is flushed when the max batch size is reached
```
msfstef, line 21: This is simple and easy to understand and implement - I wonder if there would be any advantage to doing batching with a max byte size instead (using the existing converter's running tally of transaction size accumulated)

not a request for change for this PR, just opening the discussion

robacourt: It would involve keeping byte size on the operations, which I think we need anyway, but I'll leave that for another PR

msfstef: The message converter already keeps a running tally of the total txn size - it can be used and "diffed" for the batcher

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: The interval timings need redesigning with this new structure. My aim was to do that as part of a separate PR. As far as I know I was the only one using the interval timings, and to not have them for a bit doesn't seem like the end of the world.

packages/sync-service/lib/electric/postgres/replication_client/message_converter.ex
robacourt: good idea. done

packages/sync-service/lib/electric/postgres/replication_client/message_converter.ex
robacourt: I don't like it either. Mind if I address that in a separate PR though?

packages/sync-service/lib/electric/replication/operation_batcher.ex
robacourt: Still, I'll do it in a separate PR

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: Done

packages/sync-service/lib/electric/postgres/replication_client.ex
robacourt: No, it's optional. As per your other comment, this module shouldn't need the stack_id, so I'll remove this in that later PR.

packages/sync-service/lib/electric/postgres/replication_client/message_converter.ex
msfstef: sounds good!

PR #3430, opened 2025-11-13
Don't create dynamic query keys for subset_params

KyleAMathews:
Users discovered that proxy configurations couldn't match subset__params because it was using deepObject style (subset__params[1], subset__params[2]) which creates dynamic parameter names.

This change serializes subset__params as a JSON string instead:
- Client: sends subset__params={"1":"value1","2":"value2"}
- Server: parses JSON string into map with new JsonOrMapStringParams type
- Proxy configs: can now match constant parameter name "subset__params"

The server-side implementation maintains backward compatibility by accepting both JSON strings and map objects (for tests).

Updated:
- TypeScript client to use JSON.stringify for subset params
- Elixir server to parse JSON with custom Ecto type
- OpenAPI spec to reflect JSON string format
- HTTP API documentation with examples

### DIFF
--- a/.changeset/brave-dolphins-heal.md
+++ b/.changeset/brave-dolphins-heal.md
@@ -3,6 +3,6 @@
 '@electric-sql/client': patch
 ---
 
-Fix subset__params to use constant parameter name for proxy configurations
+Fix subset\_\_params to use constant parameter name for proxy configurations
 
-Changed subset__params from deepObject style (subset__params[1], subset__params[2]) to JSON serialization (subset__params={"1":"value1","2":"value2"}). This allows proxy configurations to match the constant parameter name "subset__params" in ELECTRIC_PROTOCOL_QUERY_PARAMS without needing dynamic pattern matching.
+Changed subset**params from deepObject style (subset**params[1], subset**params[2]) to JSON serialization (subset**params={"1":"value1","2":"value2"}). This allows proxy configurations to match the constant parameter name "subset\_\_params" in ELECTRIC_PROTOCOL_QUERY_PARAMS without needing dynamic pattern matching.


samwillis:
@KyleAMathews the normal `params` is the same (`params[0]=xxx`), so this change would introduce an inconstancy in the API.

The difference is in `params` is generally constructed in the proxy, rather than proxied...

image

samwillis:
But I also suppose that `subset__params` is generally auto created by a client such as TanStack DB,so maybe an inconstancy is not a problem

samwillis:
I'm happy with the ts code and api change.

KyleAMathews:
@icehaunter can you review also (esp. elixir code)?

KyleAMathews:
@icehaunter yeah it's coming up now as the client autogenerates these subset params & the older where clause params are handwritten, generally in the proxy.

I'm not entirely sure I understand your question 😅 but the key thing here is that `subset_params` is a static param key so that our `ELECTRIC_PROTOCOL_QUERY_PARAMS` can know to pass it through automatically for people. So encoding the value as JSON means all the variability gets hidden in that.

icehaunter:
@KyleAMathews what about consistency between this and root params?

icehaunter:
Also versioning, it's a breaking change on a non-flagged feature

KyleAMathews:
It's a "compile" target not something users write so we're not overly worried about API compatibility — https://github.com/electric-sql/electric/pull/3430#issuecomment-3529225928

It's a very new feature & very few people are using it still so I'm not worried about breaking people that much — especially given it doesn't work with the proxy pattern we document and people are using w/o workarounds.

Also the server will accept inputs from older clients while they upgrade. So we only break people who a) have already started using this feature and b) upgrade their client & not their server & c) who have rolled this into production — which is probably nobody at this point.

.changeset/brave-dolphins-heal.md
```diff
@@ -0,0 +1,8 @@
+---
+'@core/sync-service': patch
+'@electric-sql/client': patch
+---
+
+Fix subset\_\_params to use constant parameter name for proxy configurations
+
+Changed subset**params from deepObject style (subset**params[1], subset**params[2]) to JSON serialization (subset**params={"1":"value1","2":"value2"}). This allows proxy configurations to match the constant parameter name "subset\_\_params" in ELECTRIC_PROTOCOL_QUERY_PARAMS without needing dynamic pattern matching.
```
alco, line 8: I worry that, as developers push AI-generated text without reviewing it more and more, fewer humans will be reading it since it just results in pure cognitive load for cognitive load's sake.

This changelog entry renders like this in our changelog:
image

PR #3431, opened 2025-11-13
Fix Electric SQL collections getting stuck after rapid tab changes

KyleAMathews:
reported at https://discord.com/channels/719702312431386674/1438326657063850035/1438326657063850035


kevin-dp:
This fix looks good to me. It will handle the case where we quickly resume the stream before the pause fully finished. The old pause shouldn't interfere with the quick resume because of the extra check that's added in the if test. It would be nice to have preview packages in Electric such that the OP can test the fix and confirm it works before we actually merge it.

PR #3434, opened 2025-11-13
Add explicit column transform so we can convert back

KyleAMathews:
Implements a new `columnMapper` API that handles both encoding (TS → DB) and decoding (DB → TS) of column names. This solves the problem of using camelCase column names in TypeScript while querying snake_case database columns.

Key features:
- `snakeCamelMapper()` - Built-in snake_case ↔ camelCase converter
- `createColumnMapper(mapping)` - Custom column name mapping
- Automatic encoding of WHERE clauses and ORDER BY clauses
- Backward compatible - existing `transformer` API still works

The new API deprecates `transformer` in favor of bidirectional mapping that works for both query results and filter parameters.

Example usage:
```typescript
import { snakeCamelMapper } from '@electric-sql/client'

  params: { table: 'todos' },
  columnMapper: snakeCamelMapper()
})

// Now you can filter with camelCase column names:
// WHERE userId = $1 → automatically encoded to → WHERE user_id = $1
```

Addresses Discord feature request for reverse-transform/encode function to complement the existing transformer prop for TanStack integration.

### DIFF
--- a/packages/typescript-client/src/column-mapper.ts
+++ b/packages/typescript-client/src/column-mapper.ts
@@ -1,5 +1,8 @@
 import { Schema } from './types'
 
+type DbColumnName = string
+type AppColumnName = string
+
 /**
  * A bidirectional column mapper that handles transforming column **names**
  * between database format (e.g., snake_case) and application format (e.g., camelCase).
@@ -20,13 +23,13 @@ export interface ColumnMapper {
    * Transform a column name from database format to application format.
    * Applied to column names in query results.
    */
-  decode: (dbColumnName: string) => string
+  decode: (dbColumnName: DbColumnName) => AppColumnName
 
   /**
    * Transform a column name from application format to database format.
    * Applied to column names in WHERE clauses and other query parameters.
    */
-  encode: (appColumnName: string) => string
+  encode: (appColumnName: AppColumnName) => DbColumnName
 }
 
 /**


kevin-dp:
@KyleAMathews what's the status of this PR? iirc `transformer` is useful e.g. for encryption but this PR marks the transformer as deprecated.

If we keep both `transformer` and `columnMapper` then we need to determine in which order they run when decoding and document that, e.g. `tansformer(decode(serverMsg))` or `decode(transformer(serverMsg))`.

packages/typescript-client/src/client.ts
```diff
@@ -728,7 +778,13 @@ export class ShapeStream<T extends Row<unknown> = Row>
     // Add PostgreSQL-specific parameters
     if (params) {
       if (params.table) setQueryParam(fetchUrl, TABLE_QUERY_PARAM, params.table)
-      if (params.where) setQueryParam(fetchUrl, WHERE_QUERY_PARAM, params.where)
+      if (params.where) {
+        // Apply column name encoding if columnMapper is provided
+        const encodedWhere = this.options.columnMapper
+          ? encodeWhereClause(params.where, this.options.columnMapper.encode)
```
kevin-dp: Type error here because `encodeWhereClause` expects the `whereClause` to be a `string` but the `params.where` argument is of type `string | Record<string, string>`.

kevin-dp: Also, instead of writing a ternary here, i would rather modify `encodeWhereClause` such that it allows the encoder to be `undefined`.
If it's undefined it returns the where clause without any changes, so we essentially move that ternary logic inside the `encodeWhereClause` function.
So it becomes:
```ts
const encodedWhere = encodeWhereClause(params.where, this.options.columnMapper?.encode)
```

packages/typescript-client/src/client.ts
```diff
@@ -749,16 +805,29 @@ export class ShapeStream<T extends Row<unknown> = Row>
     }
 
     if (subsetParams) {
-      if (subsetParams.where)
-        setQueryParam(fetchUrl, SUBSET_PARAM_WHERE, subsetParams.where)
+      if (subsetParams.where) {
+        // Apply column name encoding if columnMapper is provided
+        const encodedWhere = this.options.columnMapper
+          ? encodeWhereClause(subsetParams.where, this.options.columnMapper.encode)
```
kevin-dp: Would remove the ternary if we move the logic inside `encodeWhereClause`

packages/typescript-client/src/client.ts
```diff
@@ -749,16 +805,29 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
       if (subsetParams.limit)
         setQueryParam(fetchUrl, SUBSET_PARAM_LIMIT, subsetParams.limit)
       if (subsetParams.offset)
         setQueryParam(fetchUrl, SUBSET_PARAM_OFFSET, subsetParams.offset)
-      if (subsetParams.orderBy)
-        setQueryParam(fetchUrl, SUBSET_PARAM_ORDER_BY, subsetParams.orderBy)
+      if (subsetParams.orderBy) {
+        // Apply column name encoding to ORDER BY clause if columnMapper is provided
+        const encodedOrderBy = this.options.columnMapper
+          ? encodeWhereClause(
```
kevin-dp: Same here with the ternary

packages/typescript-client/src/column-mapper.ts
```diff
@@ -0,0 +1,248 @@
...
+/**
+ * A bidirectional column mapper that handles transforming column names
+ * between database format (e.g., snake_case) and application format (e.g., camelCase).
+ */
+export interface ColumnMapper<Extensions = never> {
+  /**
+   * Transform a row from database format to application format (decode).
+   * Applied to data received from Electric.
+   */
+  decode: (row: Row<Extensions>) => Row<Extensions>
```
kevin-dp: This function conflates the decoding of the column name with the old transformer function we had.
Perhaps best to keep the old `transformer` and have it separate from column mapping.
So `decode` would then be of type `(dbColumnName: string) => string`.
We could define some aliases to make this even more clear on the types:
```ts
type DbColumnName = string
type AppColumnName = string
decode: (dbColName: DbColumnName) => AppColumnName
encode: (appColName: AppColumnName) => DbColumnName
```

packages/typescript-client/src/column-mapper.ts
```diff
@@ -0,0 +1,248 @@
...
+   * Transform column names from application format to database format (encode).
+   * Applied to column references in WHERE clauses and other query parameters.
+   */
+  encode: (columnName: string) => string
+
+  /**
+   * Optional reverse mapping for debugging/introspection.
+   * Maps application column names to database column names.
+   */
+  mapping?: Record<string, string>
```
kevin-dp: I don't see the need for this. `encode` and `decode` cover both directions.

packages/typescript-client/src/column-mapper.ts
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * Converts a snake_case string to camelCase.
+ *
+ * @example
+ * snakeToCamel('user_id') // 'userId'
+ * snakeToCamel('project_id') // 'projectId'
+ * snakeToCamel('created_at') // 'createdAt'
+ */
+export function snakeToCamel(str: string): string {
```
kevin-dp, line 52: There are a bunch of corner cases that are not handled by this function: leading underscores, trailing underscores, multiple underscores, segments starting with uppercase.
```ts
snakeToCamel(`user_id`)
'userId' // ok
snakeToCamel(`_user`)
'User' // wrong
snakeToCamel(`user_`)
'user_' // should we drop trailing underscores?
snakeToCamel(`user_Id`)
'user_Id' // wrong
snakeToCamel(`user__id`)
'user_Id' // wrong
```

I asked chatgpt how these corner cases are usually handled, this is what it replied:

# 🐍➡️🐫 Snake Case to Camel Case — Recommended Rules

## **1. Preserve leading underscores**

Leading underscores often carry semantic meaning (e.g., privacy indicators).
Always keep them exactly as they are.

```
_test → _test
__my_var → __myVar
```

---

## **2. Drop trailing underscores**

Trailing underscores rarely encode meaning and typically cause ambiguity, so remove them during conversion.

```
test_ → test
test__ → test
```

---

## **3. Split on underscores and normalize segments**

Break the string on underscores.
Ignore empty segments (caused by multiple underscores).

Normalize each segment to lowercase before camelization.

```
test_column → testColumn
test__column → testColumn
```

---

## **4. Camelize by lowercasing the first segment and capitalizing subsequent ones**

After normalization:

* First segment → lowercase
* All other segments → `Capitalize`

```
test_column_name → testColumnName
```

---

## **5. Normalize segments that start with uppercase**

Uppercase snake segments should be treated the same as lowercase ones: normalize to lowercase first.

```
test_Column → testColumn
test_XML_value → testXmlValue
```

packages/typescript-client/src/column-mapper.ts
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * Converts a camelCase string to snake_case.
+ *
+ * @example
+ * camelToSnake('userId') // 'user_id'
+ * camelToSnake('projectId') // 'project_id'
+ * camelToSnake('createdAt') // 'created_at'
+ */
+export function camelToSnake(str: string): string {
```
kevin-dp, line 91: A basic correctness criteria is: `camelToSnake(snakeToCamel(str)) = str`. We should add that to the tests.

packages/typescript-client/src/column-mapper.ts
kevin-dp: This `mapping` should not be in the `ColumnMapper` type. The `ColumnMapper` consists of the `encode` and `decode` functions.
If we want to create a column mapper from an explicit map of column names, then we can use the `createColumnMapper` function which will turn that map into an `encoder` and `decoder` function.
But that map itself should not be part of this interface because it is not used by the `client.ts` (which relies solely on the `encode` and `decode` functions).

packages/typescript-client/src/column-mapper.ts
```diff
@@ -0,0 +1,248 @@
...
+ * - Qualified names: table.columnName
+ * - Operators: columnName IS NULL, columnName IN (...)
+ *
+ * Note: This uses regex-based replacement which works for most common cases
+ * but may not handle all complex SQL expressions perfectly. For complex queries,
+ * test thoroughly or use database column names directly in WHERE clauses.
+ *
+ * @internal
+ */
+export function encodeWhereClause(
```
kevin-dp, line 165: This may transform literal values if they are not passed as parameters, e.g.:
```ts
encodeWhereClause("user_id = 'user_id'", snakeToCamel)
// returns "userId = 'userId'"
// instead of: "userId = 'user_id'"
```

kevin-dp: Another tricky problem is if users have $ signs in their column names and provide a custom encoder to handle it. That would probably not work because we ignore matches that start with `$`.

packages/typescript-client/test/column-mapper.test.ts
```diff
@@ -0,0 +1,213 @@
...
+import {
+  snakeToCamel,
+  camelToSnake,
+  createColumnMapper,
+  snakeCamelMapper,
+  encodeWhereClause,
+} from '../src/column-mapper'
+import type { Schema } from '../src/types'
+
+describe(`snakeToCamel`, () => {
```
kevin-dp, line 11: Should also have a describe block that checks that `camelToSnake(snakeToCamel(str)) = str`

packages/typescript-client/test/column-mapper.test.ts
```diff
@@ -0,0 +1,213 @@
...
+    expect(camelToSnake(`updatedAt`)).toBe(`updated_at`)
+  })
+
+  it(`should handle single words`, () => {
+    expect(camelToSnake(`id`)).toBe(`id`)
+    expect(camelToSnake(`name`)).toBe(`name`)
+  })
+
+  it(`should handle multiple capital letters`, () => {
+    expect(camelToSnake(`userProfileImageURL`)).toBe(`user_profile_image_u_r_l`)
```
kevin-dp: This is clearly wrong, nobody would name their column `user_profile_image_u_r_l`

kevin-dp (changes requested):
I don't think we should deprecate `transformer` and the current snake to camelcase mapper has some problems with corner cases.

kevin-dp:
Don't forget to add a changeset.

kevin-dp:
@KyleAMathews could you summarize the changes you made in response to my review? Would be useful to know the state of this PR without having to look into all the details of the latest commits.

KyleAMathews:
@claude please summarize the commits after @kevin-dp's first review

KyleAMathews:
@kevin-dp claudes asleep — but basically fixed the edge cases you identified (and a few more) & added back-and-forth tests for snake <--> camel case.

kevin-dp:
> @kevin-dp claudes asleep — but basically fixed the edge cases you identified (and a few more) & added back-and-forth tests for snake <--> camel case.

So, do we agree on keeping `transformer` and changing the interface of the mapper such that decoding only does the column name decoding (and no transformations) ?

KyleAMathews:
@kevin-dp yeah — fixed the type https://github.com/electric-sql/electric/pull/3434/commits/b295dc6f0a48f6356f912160fd52c9f8202a23f8

packages/typescript-client/src/column-mapper.ts
```diff
@@ -0,0 +1,374 @@
...
+ *   columnMapper: snakeCamelMapper() // Column renaming
+ * })
+ * ```
+ */
+export interface ColumnMapper {
+  /**
+   * Transform a row from database format to application format (decode).
+   * Applied to data received from Electric.
+   */
+  decode: (row: Record<string, unknown>) => Record<string, unknown>
```
kevin-dp: This is not what i suggested. It rewrote `Row<Extensions>` to `Record<string, unknown>` it basically didn't do anything meaningful, it just rewrote the extentensions to `unknown`. Please have a look at my previous comment on this line. `decode` should have type signature `(dbColumnName: string) => string`. Also consider the type aliases i proposed in my previous review comment on this line.

kevin-dp (changes requested):

packages/typescript-client/src/column-mapper.ts
```diff
@@ -0,0 +1,354 @@
...
+ * For value transformations (e.g., encryption), use the `transformer` option.
+ *
+ * @example
+ * ```typescript
+ * const mapper = snakeCamelMapper()
+ * mapper.decode('user_id') // 'userId'
+ * mapper.encode('userId') // 'user_id'
+ * ```
+ */
+export interface ColumnMapper {
```
kevin-dp, line 21: @Claude i would introduce type aliases here to make more clear which string we expect.
`type DbColumnName = string` and `type AppColumnName = string` and use those aliases in encode and decode below.

kevin-dp: Claude is being lazy... i did it myself

PR #3436, opened 2025-11-14
Remove invalid subset loading check

KyleAMathews:
Removed the check that prevented requestSnapshot from being called in full mode. It's valid to load subsets while loading the full log, which is necessary for LiveQuery joins with collections using progressive syncMode.

Fixes the error: "Snapshot requests are not supported in full mode, as the consumer is guaranteed to observe all data"

### DIFF
--- a/packages/typescript-client/test/client.test.ts
+++ b/packages/typescript-client/test/client.test.ts
@@ -1711,6 +1711,44 @@ describe.for(fetchAndSse)(
       expect(titles).toEqual([`c`, `b`])
     })
 
+    it(`requestSnapshot works in full mode (progressive sync)`, async ({
+      issuesTableUrl,
+      insertIssues,
+      aborter,
+    }) => {
+      await insertIssues({ title: `A` }, { title: `B` }, { title: `C` })
+
+      // Create a shape in full mode (default/progressive sync)
+      const shapeStream = new ShapeStream({
+        url: `${BASE_URL}/v1/shape`,
+        params: { table: issuesTableUrl },
+        // No log parameter = full mode (default)
+        liveSse,
+        signal: aborter.signal,
+      })
+      const shape = new Shape(shapeStream)
+
+      // Wait for initial sync to complete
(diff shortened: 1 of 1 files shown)


kevin-dp:
Should we have an e2e test for this?

icehaunter:
Out of curiosity, what's the use of a snapshot in a full mode, given the client already holds all the same data?

KyleAMathews:
@icehaunter for the "progressive mode" https://tanstack.com/blog/tanstack-db-0.5-query-driven-sync#progressive-sync-fast-initial-paint--instant-client-queries

@kevin-dp you mean in the tanstack db repo?

samwillis:
@KyleAMathews This will not fix the issue with Tanstack DB progressive mode, we need to fix it on the DB side rather than on the Electric Client side.

The electric client ensures that the results of a `requestSnapshot` call are injected into the shape stream at the point where they are causally correct - this is important for `on-demand` mode in DB to insure that live updates are applied on top of the snapshot and it does not overwrite newer changes. With this fix the snapshots will still not be injected until the end of the stream, not optimistically as we would like.

The intention of `progressive` mode is that while the initial sync is happening we use and show snapshots. We need to have a discussion on this mode as currently nothing is applied to the collection until the first `up-to-date`.

icehaunter:
> @icehaunter for the "progressive mode" https://tanstack.com/blog/tanstack-db-0.5-query-driven-sync#progressive-sync-fast-initial-paint--instant-client-queries
> 
> @kevin-dp you mean in the tanstack db repo?

No, I got the general idea, I'm just not sure why additional snapshots are needed on top of already full data. I thought the idea for progressive mode was to only use these subset snapshots, otherwise it seems to make little sense if the client already has all the data. Am I missing something?

samwillis:
@icehaunter 

> I'm just not sure why additional snapshots are needed on top of already full data

DB `progressive` mode is supposed to show the snapshots before we load all the data. `on-demand` is working with the new scheme perfectly.

PR #3437, opened 2025-11-16
Fix retry-after handling for retries

KyleAMathews:
The server was sending the retry-after header to clients during overload scenarios (503 responses), but the header was not included in the access-control-expose-headers CORS header. This caused browsers to block access to the retry-after value, preventing the client from honoring the server's backoff directive.

As a result, clients would ignore the server's retry-after header and use only exponential backoff with jitter, which could result in very short retry delays (as low as 0-100ms on early retries), leading to intense retry loops 

This fix adds "retry-after" to the list of headers exposed via CORS, allowing clients to properly read and honor the server's retry-after directive.

Also adds a test to verify that the TypeScript client properly honors the retry-after header during the retry loop.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api/response.ex
+++ b/packages/sync-service/lib/electric/shapes/api/response.ex
@@ -12,6 +12,7 @@ defmodule Electric.Shapes.Api.Response do
   @electric_schema_header "electric-schema"
   @electric_up_to_date_header "electric-up-to-date"
   @electric_known_error_header "electric-internal-known-error"
+  @retry_after_header "retry-after"
 
   # List of all Electric-specific headers that may be included in API responses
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/shapes/api/response.ex
```diff
@@ -20,7 +20,8 @@ defmodule Electric.Shapes.Api.Response do
     @electric_offset_header,
     @electric_schema_header,
     @electric_up_to_date_header,
-    @electric_known_error_header
+    @electric_known_error_header,
+    "retry-after"
```
kevin-dp: Let's introduce an `@retry_after_header` module attribute for this in order to be consistent with the rest of the code.

PR #3443, opened 2025-11-17
feat: Reduce memory buildup when calculating least recently used shapes

msfstef:
Doing `tab2list` means we copy the entire table into process memory and perform an accumulation there - as the table scales with the number of shapes the memory pressure becomes non-negligible.

In this PR we iterate over individual entries using `:ets.foldl` and build up a balanced tree for maintaining the oldest N entries, such that the memory used never exceeds the specified number of least recently used shapes requested.

Added some tests to assert that it returns ordered results, as well as some edge cases and a happy path with more shapes than one.

Disclaimer: PR uses 🤖 help

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -301,8 +301,20 @@ defmodule Electric.ShapeCache.ShapeStatus do
     :ok
   end
 
+  @doc """
+  Updates the last read time for the given shape to the current time.
+  """
   def update_last_read_time_to_now(stack_ref, shape_handle) do
-    :ets.insert(shape_last_used_table(stack_ref), {shape_handle, System.monotonic_time()})
+    update_last_read_time(stack_ref, shape_handle, System.monotonic_time())
+  end
+
+  @doc """
+  Sets the last read time for the given shape to the provided time.
+
+  Used for tests, otherwise prefer `update_last_read_time_to_now/2`.
+  """
+  def update_last_read_time(stack_ref, shape_handle, time) do
+    :ets.insert(shape_last_used_table(stack_ref), {shape_handle, time})
   end
 
   def least_recently_used(stack_ref, shape_count) do
--- a/packages/sync-service/test/electric/shape_cache/shape_status_test.exs
+++ b/packages/sync-service/test/electric/shape_cache/shape_status_test.exs
@@ -286,9 +286,10 @@ defmodule Electric.ShapeCache.ShapeStatusTest do
       {:ok, state, []} = new_state(ctx)
       {:ok, shape1} = ShapeStatus.add_shape(state, shape!())
       {:ok, shape2} = ShapeStatus.add_shape(state, shape2!())
-      ShapeStatus.update_last_read_time_to_now(state, shape2)
-      Process.sleep(10)
-      ShapeStatus.update_last_read_time_to_now(state, shape1)
+
+      now = System.monotonic_time()
+      ShapeStatus.update_last_read_time(state, shape2, now)
+      ShapeStatus.update_last_read_time(state, shape1, now + 10)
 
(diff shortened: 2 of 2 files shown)


packages/sync-service/test/electric/shape_cache/shape_status_test.exs
```diff
@@ -316,6 +316,66 @@ defmodule Electric.ShapeCache.ShapeStatusTest do
...
+
+    test "returns correct N shapes when N < total shapes", ctx do
+      {:ok, state, []} = new_state(ctx)
+
+      # Add 5 shapes with staggered updates
+      shapes =
+        for i <- 1..5 do
+          {:ok, handle} = ShapeStatus.add_shape(state, shape!("test_#{i}"))
+          Process.sleep(5)
+          ShapeStatus.update_last_read_time_to_now(state, handle)
```
alco: It would be better to expose an API for tests that allows for setting the last read time to a user-provided value. This will help us get rid of the sleep in the test.

packages/sync-service/test/electric/shape_cache/shape_status_test.exs
```diff
@@ -316,6 +316,66 @@ defmodule Electric.ShapeCache.ShapeStatusTest do
...
+      assert length(result) == 3
+
+      result_handles = Enum.map(result, & &1.shape_handle)
+
+      assert Enum.at(shapes, 0) in result_handles
+      assert Enum.at(shapes, 1) in result_handles
+      assert Enum.at(shapes, 2) in result_handles
+
+      refute Enum.at(shapes, 3) in result_handles
+      refute Enum.at(shapes, 4) in result_handles
```
alco: Since we control all the timestamps in the tests, we could make this more concise by doing a list equality comparison against an expected result.

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -306,16 +306,41 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
+                acc = :gb_trees.delete(max_time, acc)
+                :gb_trees.insert(last_read, handle, acc)
+              else
+                acc
+              end
+          end
+        end,
+        :gb_trees.empty(),
+        table
+      )
```
robacourt, line 347: I find that quite hard to follow. Maybe splitting this into smaller functions would help?

msfstef: I think the logic is fairly concise (reduction with condition on accumulating) and I want to avoid adding helper functions that will pass around an arbitrary tree structure

would refactoring slightly to improve readability, better names, and adding more comments explaining the logic be ok?

alco: Looks good to me as is. Perhaps add a comment above each case clause, e.g. "Fill the tree with `shape_count` entries first" and "Maintain the current count by replacing more recently used shape handles with older ones".

packages/sync-service/test/electric/shape_cache/shape_status_test.exs
msfstef: Added update_last_read_time/3 API and replaced all sleeps

packages/sync-service/test/electric/shape_cache/shape_status_test.exs
msfstef: Got lazy with tests and let Claude keep the odd assertion format - will try to keep those more concise as well

packages/sync-service/lib/electric/shape_cache/shape_status.ex
msfstef: Converted the reduction logic to two if clauses for clarity, along with comments similar to the ones suggested by @alco

robacourt: Yeah, much nicer. Thanks 👍

robacourt (approved):
Nice!

PR #3445, opened 2025-11-17
Consolidate linting tools into single CI job

KyleAMathews:
Previously, linting/formatting errors would cause multiple CI jobs to fail simultaneously (often 7+ jobs), making it confusing to identify the actual issue and slowing down CI feedback.

Changes:
- Created new autoformat.yml workflow that runs all linting/formatting checks in a single job:
  - TypeScript prettier formatting check
  - TypeScript eslint checks (all packages)
  - Elixir mix format check
  - Elixir compilation warnings check

- Removed redundant linting checks from other workflows:
  - ts_test.yml: Removed stylecheck and format:check from package and example checking jobs (kept typecheck for actual type errors)
  - elixir_tests.yml: Removed the separate formatting job
  - integration_tests.yml: Removed --warnings-as-errors flag from compile step

Now linting/formatting errors will only cause the single autoformat job to fail, while other test jobs focus on functional testing.

### DIFF
--- a/.github/workflows/autoformat.yml
+++ b/.github/workflows/autoformat.yml
@@ -41,9 +41,35 @@ jobs:
 
       - name: Check Prettier formatting
         run: pnpm run format:check
+        continue-on-error: true
+        id: prettier_check
+
+      - name: Show formatting help message
+        if: steps.prettier_check.outcome == 'failure'
+        run: |
+          echo "::error::Prettier formatting check failed"
+          echo ""
+          echo "Note: For pull requests, this check runs on a merge commit between"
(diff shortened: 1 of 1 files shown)


.github/workflows/autoformat.yml
```diff
@@ -0,0 +1,79 @@
...
+    paths-ignore:
+      - 'website/**'
+      - '**/README.md'
+
+permissions:
+  contents: read
+
+jobs:
+  typescript_formatting:
+    name: TypeScript/JavaScript formatting and linting
```
kevin-dp: Let's shorten to "TypeScript formatting and linting"

alco (approved):
Nice!

PR #3449, opened 2025-11-18
chore: Fix sync service test flakes

msfstef:
- Fixed timeout when lock is taken test
- More reliable assertion for AsyncDeleter tests
- Handle holding connection with `receive ... do` blocks
- Run test tasks under test supervisor
- Terminate PG backends when rerunning tests to avoid "another session is using X db"

I've ran the tests on CI many times to reproduce issues - there's a couple more flakes but I think this irons out the main ones

### DIFF
--- a/packages/sync-service/test/electric/shape_cache_test.exs
+++ b/packages/sync-service/test/electric/shape_cache_test.exs
@@ -494,7 +494,7 @@ defmodule Electric.ShapeCacheTest do
       alias Electric.Replication.Eval.Parser
       where_clause = Parser.parse_and_validate_expression!("TRUE", refs: %{})
       # Insert a fake slow query
-      where_clause = %{where_clause | query: "PG_SLEEP(10)::text ILIKE ''"}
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/replication/publication_manager/configurator.ex
```diff
@@ -269,7 +269,7 @@ defmodule Electric.Replication.PublicationManager.Configurator do
 
   @spec do_publication_update(
           :add_and_configure | :add | :configure | :drop,
-          PublicationManager.RelationTracker.relation_filter(),
+          Electric.oid_relation(),
           state()
         ) :: {:ok, :configured | :dropped} | {:error, any()}
   defp do_publication_update(action, filter, state)
```
robacourt, line 275: Should `filter` not be called `oid_relation` too?

msfstef: well a relation is a filter so it's used interchangeable - a set of relations is called `relation_filters`

Issue #3447, opened 2025-11-18
Make individual shape consumers able to idempotently consume transactions


msfstef:
Shape consumers/writers have all the necessary information to know whether a transaction has already been ingested into the log or not, and can thus decide whether to discard it.

Currently this check is only done at the `ShapeLogCollector` level, but doing it at the consumer level will allow consumers themselves to idempotently ingest transactions using the monotonically increasing LSN, and relax the requirements for global LSN tracking across shapes for ingestion.

Issue #3442, opened 2025-11-17
Minimise overhead for startup shape recovery


msfstef:
The concern of the work involved in this issue is ensuring shapes can independently be recovered at runtime and as little work as possible is done upon startup to allow for requests to be served.

The startup overhead for recovering existing shape should scale as little as possible with the actual number of shapes to recover.

Our current recovery process involves the following:

- We have a `ShapeStatus` table that is our central shape metadata store, which contains 1) an index from shape definitions to shape handles (our pointers), and 2) actual shape metadata. On controlled shutdowns, we backup this table and attempt to recover it to accelerate recovery.
  - We are too defensive in aborting this recovery on mismatches between files and the table, falling back to reading all shape definitions from files rather than trusting the backup and cleaning up files that don't match
  - We should separate the pointer table from the shape metadata table, such that the shape metadata table can be recovered lazily
- We go over all recovered shapes and extract their latest LSN in order to determine the global last seen LSN
  -  This involves file operations that scale with the number of shapes before we can begin processing the replication stream, which could be avoided with 1) a central global last seen LSN persister and 2) shape consumer idempotent transaction handling.

Addressing the sub-points specified above will significantly improve recovery, decoupling the scaling of operations required to perform recovery with number of shapes stored and aggressively trimming shapes that cannot be easily recovered.

I will create sub-issues for these in order to split up the work into relatively independent chunks to minimise overlap with other work being done on consumers and replication stream processing. This might lead to some redundant operations occurring in the sync service until the work is finalised (e.g. central global LSN persister while we still rely on shape recovery for it), but redundant is better than lacking/incomplete.

balegas:
Thanks for opening this issue. 

> On controlled shutdowns, we backup this table and attempt to recover it to accelerate recovery

Does this mean that rn if Electric crashes we always lose all shapes? Sounds this could make this a periodic operation (I'm asking to understand, and just on shutdown is perfectly fine for now)

> which could be avoided with 1) a central global last seen LSN persister and 2) shape consumer idempotent transaction handling.

To clarify, the global last seen LSN persister would basically persist it periodically, or when take the snapshot before shutdown?

msfstef:
> Does this mean that rn if Electric crashes we always lose all shapes? Sounds this could make this a periodic operation (I'm asking to understand, and just on shutdown is perfectly fine for now)

[...]

PR #3452, opened 2025-11-18
feat: Consumers idempotently handle transactions

msfstef:
Closes https://github.com/electric-sql/electric/issues/3447

We were already filtering transactions at the SLC level, but for the sake of loosening the requirements of having a central "latest LSN processed" node for filtering we move the filter at the consumer level.

For the sake of maintaining what we already have in place, I'm simply comparing the latest log offset of the transaction compared to the latest log offset tracked by the consumer - the flush acknowledgements are separate anyway, if a flush ack is not sent for a transaction then it will be sent again anyway.

I've also fixed the `latest_offset` state variable kept in the consumer which was not being updated.

### DIFF
--- a/.changeset/brave-crews-help.md
+++ b/.changeset/brave-crews-help.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Ensure shape consumers idempotently handle transactions using log offset comparisons.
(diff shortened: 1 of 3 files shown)


robacourt (approved):
Lovely! Nice and small too 🙏

PR #3455, opened 2025-11-18
feat(sync-service): Terminate shape consumer processes after timeout

### LINKED ISSUE #3253: Inactive consumers should be put offline
Hibernated consumers should be put offline after some timeout, so replacing the process hibernation with process shutdown (without shape removal).

This will reduce memory usage to some subset of the defined shapes, the size of which will depend on the timeout duration.

Shapes which are part of a sub-query chain should remain active (but hibernated). The existing dependency structure would not survive the restart of one of its members. Shutting down and restarting shapes with dependencies can be part of later work.

magnetised:
Closes #3253

### DIFF
--- a/packages/sync-service/test/electric/shapes/consumer_test.exs
+++ b/packages/sync-service/test/electric/shapes/consumer_test.exs
@@ -952,7 +952,7 @@ defmodule Electric.Shapes.ConsumerTest do
         %Commit{lsn: lsn1}
       ]
 
-      assert :ok = ShapeLogCollector.store_transaction(txn, ctx.producer)
+      assert :ok = ShapeLogCollector.handle_operations(txn, ctx.producer)
 
       assert_receive {:flush_boundary_updated, 300}, 1_000
 


packages/sync-service/config/runtime.exs
```diff
@@ -172,6 +172,8 @@ call_home_telemetry_url =
 shape_hibernate_after =
   env!("ELECTRIC_SHAPE_HIBERNATE_AFTER", &Electric.Config.parse_human_readable_time!/1, nil)
 
+shape_enable_suspend? = env!("ELECTRIC_SHAPE_SUSPEND_CONSUMER", :boolean, nil)
```
msfstef, line 175: if this is supposed to be a feature flag we might want to add it under the existing feature flags env var? it'll make easy to configure it on cloud as well

packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -150,7 +179,12 @@ defmodule Electric.Shapes.ConsumerRegistry do
 
   @doc false
   def registry_table(stack_id) do
-    :ets.new(ets_name(stack_id), [:public, :named_table, write_concurrency: :auto])
+    :ets.new(ets_name(stack_id), [
+      :public,
+      :named_table,
+      write_concurrency: true,
```
msfstef: why not keep it `:auto`? do we not trust OTP enough, or do OTP registries themselves use `true` explicitly?

packages/sync-service/config/runtime.exs
magnetised: I don't think this is a feature flag -- as in it's not a wip feature that will eventually just be turned on. I think it's a feature that may be enabled or disabled on a per-install basis on an on-going basis.

re cloud: can we only toggle feature flags or do we get a per instance ENV override too?

packages/sync-service/lib/electric/shapes/consumer_registry.ex
magnetised: My thinking was: now we know that multiple processes **will** be concurrently writing why not just enable it. Reading the docs (😲 ) I see `:auto` is the recommended setting. I should read more docs.

packages/sync-service/config/runtime.exs
msfstef: Ah ok - I figured that if this works we would just have it always one and not configurable

re cloud: we can do a per-source config setting just fine, but it involves updating validation of parameters in code etc which is a bit of a pain - so for features that are turned on or off that we're trying out I'd rather reuse a single mechanism to avoid the overhead of UI toggles, parsing, validation, etc

magnetised: Ok. In that case I'll make it a feature flag so we can try it out easily then we can move it to the config later.

magnetised: @msfstef I've doubled up: kept the config env setting with a default true but added the feature flag `suspend_consumers` too, so both have to be enabled for the feature to work

msfstef: makes sense - if the default is true it works for us, and enabling the feature flag makes it possible to configure it if someone doesn't want it

if it ends up being perfect we can remove the feature flag

PR #3458, opened 2025-11-18
fix: Always cleanup replication slots on graceful shutdown

### LINKED ISSUE #3284: Explicit Postgres connection cleanup on graceful shutdown
During graceful shutdowns, Electric doesn't explicitly clean up its Postgres backend connections. We've had issues where we've seen hanging connections and we can't quite explain why. We want to explicitly shutdown the connection to prevent any potential inconsistent states. While this doesn't address crash scenarios, it might help preventing accumulation of stale connections during normal deployment cycles and planned restarts .

**Proposed Solution**: Track backend PIDs and explicitly terminate them during shutdown sequence using a separate pool connection.

## Requirements

- Capture Postgres backend conneciton PID when establishing replication connection
- On graceful shutdown, terminate the backend process
- Defensively clean up replication slots before termination
- Ensure cleanup happens before releasing advisory lock

kennyu:
Remove conditional checks in Connection.Manager.terminate/2 to ensure replication slots and publications are always cleaned up during graceful shutdown, preventing resource leaks during deployment cycles.

Fixes #3284

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -880,7 +880,8 @@ defmodule Electric.Connection.Manager do
 
     # drop publication before replication client so that we ensure anyone
     # acquiring the lock next will recreate the replication slot
-    if state.drop_slot_requested, do: drop_publication(state)
+    # Always perform defensive cleanup during graceful shutdown
+    drop_publication(state)
 
     if is_pid(replication_client_pid) do
       # if we are acquiring the lock, we should kill the replication client brutally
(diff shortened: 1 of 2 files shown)


msfstef:
It is currently undocumented but you an set an environment variable `CLEANUP_REPLICATION_SLOTS_ON_SHUTDOWN` to true and it will create the slot as a [temporary slot](https://www.postgresql.org/docs/current/view-pg-replication-slots.html#VIEW-PG-REPLICATION-SLOTS) which is automatically cleaned up when the connection is lost.

As @icehaunter pointed out, cleaning up the slot means you cannot recover any data, so the above might be what you need.

balegas:
Agree with both. way to achieve this is to use the temporary slot.

Issue #3448, opened 2025-11-18
Separate responsibility of global LSN tracking


msfstef:
Currently upon recovery we go through every individual shape to read the last processed LSN to derive the global last processed LSN.

We should instead have a central periodic persister, which can be achieved if shapes can idempotently consume transactions they've already seen and if we relax the requirement for the reported global last seen LSN.

PR #3462, opened 2025-11-19
feat: LSN tracker persistence through replication slot confirmed flush position

msfstef:
Closes https://github.com/electric-sql/electric/issues/3448

The initial idea was to have a separate, periodic persister of the last processed LSN, but as I was working through the code I realised that we might not even need that.

Since now consumers are idempotent in the transactions they handle (see https://github.com/electric-sql/electric/pull/3452), even the our global last processed LSN is slightly behind any given shape that is ok.

Instead of persisting it, we actually recover it from the replication slot's confirmed flush LSN.
- If the slot is new, we reset the last processed LSN to the new slot's LSN since all shapes will be invalidated
- If the slot is an existing one, we initialise it to the confirmed flush LSN
  - If the reconnection to the slot happens at runtime (e.g. because of network partition), we preserve the current last processed LSN
  - If the reconnection happens at startup, we initialise the LSN from the slot. This value might be _slightly_ behind if shapes processed transactions that they did not manage to acknowledge back, but we can tolerate this.

This way, shape recovery does not have to read all shapes' last processed LSN during startup.

If we need stronger guarantees of the global processed lsn being the max, we can also attempt to set it everytime we lazily recover an existing shape (if the shape's max lsn is larger than the current max processed lsn). Happy to include this in this PR or a next one.

### DIFF
--- a/packages/sync-service/test/electric/postgres/lock_breaker_connection_test.exs
+++ b/packages/sync-service/test/electric/postgres/lock_breaker_connection_test.exs
@@ -9,6 +9,7 @@ defmodule Electric.Postgres.LockBreakerConnectionTest do
   setup [
     :with_unique_db,
     :with_stack_id_from_test,
+    :with_lsn_tracker,
     :with_slot_name
   ]
 


packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -256,52 +254,19 @@ defmodule Electric.ShapeCache do
...
-        ordered: false
-      )
-      |> Enum.reduce({Lsn.from_integer(0), 0}, fn
-        {:ok, {_handle, :error}}, acc -> acc
-        {:ok, {_handle, lsn}}, {max, recovered} when is_larger(lsn, max) -> {lsn, recovered + 1}
-        _, {max, recovered} -> {max, recovered + 1}
-      end)
-
-    total_failed_to_recover = length(all_handles_and_shapes) - total_recovered
+    total_recovered = length(all_handles_and_shapes)
```
magnetised: This feels a bit pointless now. Why even do this? the only output is a log message which is now telling us how long it takes to list the shapes and count the list. We can get the number of "recovered" shapes on init of the shape status table.

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -201,23 +199,23 @@ defmodule Electric.ShapeCache do
   @impl GenServer
   def handle_continue(:recover_shapes, state) do
```
magnetised: we're not recovering anything here... see below for my comment on `recover_shapes/1` -- all feels a bit performative for a function that calls `PublicationManager.wait_for_restore/1` and `ShapeLogCollector.mark_as_ready/1`

packages/sync-service/test/electric/replication/shape_log_collector_test.exs
```diff
@@ -20,6 +20,7 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
     only: [
       with_in_memory_storage: 1,
       with_shape_status: 1,
+      with_lsn_tracker: 1,
```
magnetised: we could just drop this `only` clause -- in other tests we just import the whole module

packages/sync-service/lib/electric/shape_cache.ex
msfstef: I plan to do an integrity check that will fail some shapes if they do not match so theres' going to be something to report - but you're right I'll make this more compact and rewrite whatever is necessary in other PRs

packages/sync-service/lib/electric/shape_cache.ex
msfstef: Renamed it to `:wait_for_restore` - I still prefer this in the `handle_continue` because we don't want to wait for the publication manager in the init, but I simplified the other calls, there's no explicit restoration reference.

packages/sync-service/lib/electric/postgres/replication_client/connection_setup.ex
```diff
@@ -284,6 +286,32 @@ defmodule Electric.Postgres.ReplicationClient.ConnectionSetup do
 
   ###
 
+  defp query_slot_flushed_lsn_query(%State{slot_name: slot_name} = state) do
+    Logger.debug("ReplicationClient step: query_slot_flushed_lsn")
+
+    query = """
+      SELECT confirmed_flush_lsn
+      FROM pg_replication_slots
+      WHERE slot_name = #{Utils.quote_string(slot_name)}
```
alco, line 295: nit: just as a double-checking mechanism, we could add `AND NOT active` to the where clause here to make sure we're not reading LSN from a slot that's used by another consumer. We already ensure exclusive access with an advisory lock but I just feel like between any two independent queries unexpected stuff may happen...

So if the slot is active, we'd get back 0 rows and that should be treated as a crash of the replication client.

alco: Or maybe I'm overthinking too much and crashing the client here would open up a way for another Electric to steal the thunder.... I don't know.

msfstef: I think that we should rely on the advisory lock since we do so in various places - if we're querying this it means that we fully intend on using this replication slot as well. If it ends up being already active then the replication client will fail when it attempts to start streaming anyway

PR #3463, opened 2025-11-19
expose ShapeStream.fetchSnapshot as public api

samwillis:
This exposes the `ShapeStream.fetchSnapshot` as a public api that can be used to fetch a snapshot without it being injected into the emitted stream of change messages. This is useful for cases where the user wants to handle the application of these snapshot in a custom way, as we need to in Tanstack DB for the `progressive` mode.

While working on this I also took the opportunity to tidy up the parsing of the messages, in the old implementation we would JSON.stringify the messages, only to the parse them again. I refactored the parser to be able to work on already parsed (JSON -> basic JS) messages.

First commit is the core change to expose `fetchSnapshot`, second commit is the refactor of message parsing.

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -1338,15 +1338,15 @@ export class ShapeStream<T extends Row<unknown> = Row>
         url: fetchUrl.toString(),
       })
 
-    const responseData = await response.json()
-    const batch = this.#messageParser.parseSnapshotData<ChangeMessage<T>>(
-      responseData.data,
+    const { metadata, data: rawData } = await response.json()
+    const data = this.#messageParser.parseSnapshotData<ChangeMessage<T>>(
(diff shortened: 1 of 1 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -1338,14 +1338,14 @@ export class ShapeStream<T extends Row<unknown> = Row>
         url: fetchUrl.toString(),
       })
 
-    const { metadata, data } = await response.json()
-    const batch = this.#messageParser.parse<Array<ChangeMessage<T>>>(
-      JSON.stringify(data),
+    const responseData = await response.json()
```
kevin-dp: nit: we could destructure it here such that we don't need to access it below (i.e. just pass `data` to the fun call and use `metadata` in the returned object).

kevin-dp (approved):

Issue #3456, opened 2025-11-18
`ShapeStatus.count_shapes/1` does not tally with `Electric.ShapeCache.ShapeStatus.list_shapes() |> length()` on fresh boot


magnetised:
On a system with `49599` defined shapes which has been correctly shutdown with `System.stop/0` the ShapeStatus data does not tally on restart.

```elixir
iex> Electric.ShapeCache.ShapeStatus.count_shapes("single_stack")
0
iex> Electric.ShapeCache.ShapeStatus.shape_meta_table("single_stack") |> Electric.ShapeCache.ShapeStatus.list_shapes() |> length()
49599
```

When applying load, the system correctly recognises existing shape handles and records lots of `Started consumer for existing handle` messages and the `count_shapes` tally starts to converge with the result of the list_shapes.

The system logs the following on boot (key lines isolated):

```
15:01:08.694 pid=<0.312.0> [info] Starting ElectricSQL 1.2.5
15:01:08.715 pid=<0.316.0> [info] Admission control initialized with table: electric_admission_control

15:01:10.135 pid=<0.374.0> [info] Loaded shape status from backup at /tmp/electric/shapes/single_stack/.meta/backups/shape_status_backups/shape_status_v1.ets.backup

15:01:10.189 pid=<0.313.0> [info] Running Electric.Plug.Router with Bandit 1.8.0 at 0.0.0.0:3000 (http)
15:01:10.229 pid=<0.696.0> [info] Acquiring lock from postgres with name electric_slot_default
15:01:10.229 pid=<0.385.0> [info] Postgres server version = 170002, system identifier = 7574075528495947811, timeline_id = 1
15:01:10.229 pid=<0.696.0> [info] Lock acquired from postgres with name electric_slot_default
15:01:10.286 pid=<0.717.0> [info] Connection pool [admin: 127.0.0.1:5555] is ready with 4 connections
15:01:10.286 pid=<0.698.0> [info] Connection pool [snapshot: 127.0.0.1:5555] is ready with 16 connections
15:01:10.296 pid=<0.385.0> [info] Connected to Postgres 7574075528495947811 and timeline 1
15:01:10.297 pid=<0.725.0> [info] Starting shape replication pipeline
15:01:10.324 pid=<0.385.0> [info] Starting replication from postgres

15:01:11.195 pid=<0.728.0> [info] Restored filters for 49599 shapes
[...]

Issue #3450, opened 2025-11-18
Separate `shape def -> handle` index table from metadata


msfstef:
The pointers from a shape definition to an existing shape handle are the most important thing that defines whether a shape "exists" as far clients are concerned. Deleting a pointer entry means that the rest of the metadata can be cleaned up asynchronously, since it is theoretically unreachable.

I believe that this index should be its own separate table, as lean and fast as possible, and in future tasks revisit how it is backed up and its consistency checked against existing files.

PR #3466, opened 2025-11-19
feat: Separate `shape def -> handle` table from metadata

msfstef:
Closes https://github.com/electric-sql/electric/issues/3450
Fixes https://github.com/electric-sql/electric/issues/3456 as a bonus

I have separated the `shape def -> handle` lookup into its own table - this alone is not necessarily an improvement but it is the precursor to lazily loading the shape metadata table as requests or data for shapes comes in.

In order to split up the work, despite separating the tables, I'm still backing up both of them to maintain the same functionality as before. In the next steps I will only be backing up the hash lookup.

As part of this work I have also ensured the last used table is repopulated upon restoring from the backup with new dates - I don't think it's worth backing that table up - and I have replaced the `count` API to use the hash lookup table instead.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -22,7 +22,8 @@ defmodule Electric.ShapeCache.ShapeStatusBehaviour do
 
   @callback initialize_from_storage(stack_ref(), Electric.ShapeCache.Storage.storage()) ::
               :ok | {:error, term()}
-  @callback terminate(stack_ref(), String.t()) :: :ok | {:error, term()}
+  @callback terminate(stack_ref(), Electric.ShapeCache.Storage.storage()) ::
+              :ok | {:error, term()}
   @callback list_shapes(stack_ref()) :: [{shape_handle(), Shape.t()}]
   @callback count_shapes(stack_ref()) :: non_neg_integer()
   @callback get_existing_shape(stack_ref(), Shape.t() | shape_handle()) ::
@@ -104,7 +105,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
   @impl true
   def terminate(stack_ref, storage) do
     case backup_dir(storage) do
-      nil -> :ok
+      nil -> {:error, :no_backup_dir_configured}
       backup_dir -> store_backup(stack_ref, backup_dir)
     end
   end


packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -199,7 +207,8 @@ defmodule Electric.ShapeCache.ShapeStatus do
       # Always delete the hash lookup first, so that we guarantee that no shape spec
       # is ever matched to a handle with incomplete information, since deleting with
       # select_delete can lead to inconsistent state
-      :ets.delete(meta_table, {@shape_hash_lookup, Shape.comparable(shape)})
+      :ets.delete(shape_hash_lookup_table(stack_ref), Shape.comparable(shape))
+
       :ets.delete(meta_table, {@shape_meta_data, shape_handle})
       :ets.delete(meta_table, {@snapshot_started, shape_handle})
```
magnetised, line 214: why not merge this snapshot started value into the metadata for a shape handle?

msfstef: ah you were right - I was mistaken and thought that we just had it as part of the metadata

if you don't mind I'll sort it out in the next PR where I'll be lazy loading things

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -406,6 +403,20 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
+
+  def shape_hash_lookup_table(opts) when is_list(opts),
+    do: shape_hash_lookup_table(opts[:stack_id])
+
+  def shape_hash_lookup_table(stack_id) when is_stack_id(stack_id),
+    do: :"shape_hash_lookup_table:#{stack_id}"
+
+  if Mix.env() == :test do
+    def shape_hash_lookup_table(state) when is_map(state), do: state.shape_hash_lookup_table
+  end
```
robacourt, line 416: Why not just use the test's stack_id?

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -123,14 +119,13 @@ defmodule Electric.ShapeCache do
   @impl Electric.ShapeCacheBehaviour
   def await_snapshot_start(shape_handle, stack_id)
       when is_shape_handle(shape_handle) and is_stack_id(stack_id) do
-    meta_table = ShapeStatus.shape_meta_table(stack_id)
-    ShapeStatus.update_last_read_time_to_now(meta_table, shape_handle)
```
msfstef, line 127: @magnetised @robacourt I realised just now that this means that the LRU cache was not actually doing LRU so far - we were inserting the LRU records in the wrong ETS table

therefore so far the expiry shape manager was just expiring the oldest shapes regardless of usage

packages/sync-service/lib/electric/shape_cache/shape_status.ex
msfstef: yep I also wondered the same - I'll fix up these tests in the next PR

Issue #3451, opened 2025-11-18
Lazily load shape metadata from files


msfstef:
Rather than trying to backup in-memory metadata and then doing complicated consistency checks, if the `shape def -> handle` index is the only backup, we can lazily load the shape's metadata from the files when needed (e.g. request for it comes in)

This way there is no need to ensure consistency between the backed up metadata and the files - the only consistency check done is which shapes are available.

msfstef:
Deprioritising and eventually closing in favour of moving to an SQLite-backed metadata store implementation (details/plan available soon by @magnetised), at which point "lazily loading" from disk is irrelevant.

PR #3471, opened 2025-11-19
chore: Clean up `ShapeStatus` implementation

msfstef:
This work is in preparation of https://github.com/electric-sql/electric/issues/3451

- Removes `set_snapshot_xmin` and `snapshot_xmin` API as it is basically unused
- Convert `snapshot_started` metadata into positional entry in metadata tuple (and rename setter API to `mark_snapshot_as_started` for consistency)
- ETS table names only accept options with a stack id or a stack id, rather than an arbitrary table name, eliminating the possibility for issues like the one encountered for the LRU cache @robacourt 
- Remove table names from public API and behaviour - this is entirely internal now
- Make listing shapes for given relations a part of the behaviour
- Fix up tests
  - Remove unused tests
  - Use test stack IDs where possible rather than arbitrary table names 
  - Use stub inspector rather than arbitrary module functions
  
  
I considered (and in fact have a commit stashed for it) splitting the relation lookup table from the metadata table, into a table of type `:bag`, but then realised that it would also have to be an `:ordered_set` since for a given relation we might have an arbitrary number of shapes attached to it, and we need ordering on the key for efficient per-shape removals.

The only argument I still potentially have for splitting it out is that the relation lookups have a very different access pattern - we add entries when a shape is created, we look them up only for the sake of shape deletions, we remove them when a shape is removed. The read pattern is exclusively range reads on a given relation oid partial key head. This _might_ create contention on an otherwise critical table.

I would like to hear opinions on splitting it out @alco @robacourt 

edit: I've opened the PR for splitting out the relation lookup table as I've convinced myself https://github.com/electric-sql/electric/pull/3472

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -39,10 +39,6 @@ defmodule Electric.ShapeCache.ShapeStatusBehaviour do
   @callback snapshot_started?(stack_ref(), shape_handle()) :: boolean()
   @callback remove_shape(stack_ref(), shape_handle()) :: {:ok, Shape.t()} | {:error, term()}
   @callback reset(stack_ref()) :: :ok
-
-  @callback shape_hash_lookup_table(stack_ref()) :: atom()
-  @callback shape_meta_table(stack_ref()) :: atom()
-  @callback shape_last_used_table(stack_ref()) :: atom()
 end
 
 defmodule Electric.ShapeCache.ShapeStatus do
@@ -71,6 +67,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
 
   @behaviour Electric.ShapeCache.ShapeStatusBehaviour
 
+  @typep stack_ref() :: Electric.ShapeCache.ShapeStatusBehaviour.stack_ref()
   @type shape_handle() :: Electric.ShapeCacheBehaviour.shape_handle()
   @type table() :: atom() | reference()
   @type t() :: Keyword.t() | binary() | atom()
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -401,25 +401,19 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
-
   def shape_hash_lookup_table(opts) when is_list(opts) or is_map(opts),
     do: shape_hash_lookup_table(Access.fetch!(opts, :stack_id))
 
   def shape_hash_lookup_table(stack_id) when is_stack_id(stack_id),
     do: :"shape_hash_lookup_table:#{stack_id}"
 
   @impl true
-  def shape_meta_table(table) when is_atom(table), do: table
-
```
robacourt: 🙏

magnetised:
> The only argument I still potentially have for splitting it out is that the relation lookups have a very different access pattern - we add entries when a shape is created, we look them up only for the sake of shape deletions, we remove them when a shape is removed. The read pattern is exclusively range reads on a given relation oid partial key head. This _might_ create contention on an otherwise critical table.

You didn't ask for my opinion but... AFAIK the only time we need to look up a shape by relation is when the publication manager hits a problem and issues a delete for all associated shapes. My question would be why is the performance of this so critical? It's a fairly edge case. If splitting out the relation info improves the system and the only reason not to do it is this one *potential* performance issue that's not really in the critical path (ets is fast...) then I'd say go for it. The bag type might be even quicker than the current select anyway (it's just a `lookup` on a key right?)

msfstef:
> You didn't ask for my opinion but...

Always appreciated, I don't tag everyone just to not be annoying!

> My question would be why is the performance of this so critical?

The performance of these lookups in particular is anything _but_ critical, which is why I agree that even the suspicion that the access patterns for this might impact the critical read and write path is enough to justify splitting it out.

> The bag type might be even quicker than the current select anyway (it's just a lookup on a key right?)

I initially implemented it with a bag but realised that the _most_ common operation is just individual shape removal, rather than removal of all shapes under a relation, and removing an object under a single key in a bag would scale with O(N_rel) where N_rel is the number of shapes under a given relation.

Using an ordered set allows us to still have O(1) removals which are common, and range scans for a given relation are still very efficient since the keys are ordered by the relations first.

Feel free to give https://github.com/electric-sql/electric/pull/3472 a review !

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -491,8 +502,8 @@ defmodule Electric.ShapeCache.ShapeStatus do
 
           meta_tuples =
             [
-              {{@shape_meta_data, shape_handle}, shape, nil, LogOffset.first()},
-              {{@snapshot_started, shape_handle}, snapshot_started?}
+              {{@shape_meta_data, shape_handle}, shape, snapshot_started?, nil,
+               LogOffset.last_before_real_offsets()}
```
alco, line 476: Why the change from `LogOffset.first()` here?

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -235,6 +235,17 @@ defmodule Electric.ShapeCache.ShapeStatus do
     :ok
   end
 
+  @doc """
+  Removes all ETS tables associated with the given stack reference.
+  Used in tests for tearing down.
```
alco, line 238: Shouldn't the tests terminate the shapestatusowner process instead for a clean teardown of tables?

packages/sync-service/lib/electric/shape_cache/shape_status.ex
msfstef: I don't know why `LogOffset.first()` was there in the first place - there is a whole line of reasoning documented in the shape initialisation as to why we use `last_before_real_offsets`, which I believe applies to this loading as well - so I chose to keep it consistent.

Ultimately the shape needs to actually populate this offset for readers to be able to access the log properly anyway, and I'm not sure if it works properly right now - this will all be irrelevant after https://github.com/electric-sql/electric/issues/3451

packages/sync-service/lib/electric/shape_cache/shape_status.ex
msfstef: In these tests there is no `ShapeStatusOwner` process, the shape status module is tested and intiialised within the test process

Because the shape status owner might have slightly some more responsibilities (e.g. backing things up periodically, and on shutdown) I'd rather it is tested separately for just those responsibilities

PR #3472, opened 2025-11-19
feat: Split out relation to shape lookup ETS table in `ShapeStatus`

msfstef:
See discussion in https://github.com/electric-sql/electric/pull/3471

This is also in preparation of https://github.com/electric-sql/electric/issues/3451

Basically we only use the relation lookup for deleting shapes and we do large range scans over the table which might lock it up or create contention, even if the table is partitioned via `:write_concurrency`. Currently it is colocated with the metadata table which is on the critical read and write path.

By splitting it into its own table, we ensure that low priority scans for deleting shapes never end up blocking reads or writes to the metadata table.

As a bonus, until I actually implement lazy recovery of shape metadata, by moving this into a separate table we also no longer persist these as our backup as they can be recovered from the shape itself

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -78,6 +78,8 @@ defmodule Electric.ShapeCache.ShapeStatus do
 
   @shape_hash_lookup_handle_pos 2
 
+  @shape_last_used_time_pos 2
+
   @shape_meta_shape_pos 2
   @shape_meta_snapshot_started_pos 3
   @shape_meta_latest_offset_pos 4
@@ -320,7 +322,11 @@ defmodule Electric.ShapeCache.ShapeStatus do
   Used for tests, otherwise prefer `update_last_read_time_to_now/2`.
   """
   def update_last_read_time(stack_ref, shape_handle, time) do
-    :ets.insert(shape_last_used_table(stack_ref), {shape_handle, time})
+    :ets.update_element(
+      shape_last_used_table(stack_ref),
+      shape_handle,
+      {@shape_last_used_time_pos, time}
+    )
   end
 
   def least_recently_used(stack_ref, shape_count) do


packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -183,33 +175,27 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
       # is ever matched to a handle with incomplete information, since deleting with
       # select_delete can lead to inconsistent state
       :ets.delete(shape_hash_lookup_table(stack_ref), Shape.comparable(shape))
 
-      :ets.delete(meta_table, {@shape_meta_data, shape_handle})
+      :ets.delete(meta_table, shape_handle)
 
       Enum.each(Shape.list_relations(shape), fn {oid, _} ->
-        :ets.delete(meta_table, {@shape_relation_lookup, oid, shape_handle})
+        :ets.delete(shape_relation_lookup_table(stack_ref), {oid, shape_handle})
```
magnetised: nit: why not move the table name gen out of the loop?

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -183,33 +175,27 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
-          meta_table,
-          {@shape_meta_data, shape_handle},
-          @shape_meta_shape_pos
-        )
+      shape = :ets.lookup_element(meta_table, shape_handle, @shape_meta_shape_pos)
 
       # Always delete the hash lookup first, so that we guarantee that no shape spec
       # is ever matched to a handle with incomplete information, since deleting with
       # select_delete can lead to inconsistent state
       :ets.delete(shape_hash_lookup_table(stack_ref), Shape.comparable(shape))
```
magnetised, line 197: since we now have the "critical" delete isolated is it worth splitting the deletion into an immediate plus async call much like I've done with the shapecleaner. that way we can cleanup the ets asynchronously knowing that the shape has effectively, and quickly, been "removed".

magnetised (approved):
ShapeStatus is starting to look very sane, which is saying something!

packages/sync-service/lib/electric/shape_cache/shape_status.ex
msfstef: that's the plan - but I want to do it in a separate PR!

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -557,14 +586,22 @@ defmodule Electric.ShapeCache.ShapeStatus do
           do: :ets.rename(recovered_hash_lookup_table, hash_lookup_table)
 
         last_used_table = create_last_used_table(stack_ref)
+        relation_lookup_table = create_relation_lookup_table(stack_ref)
 
         # repopolate last used table with current time
```
alco: Typo: should be `repopulate`.

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -506,7 +535,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
         handles = Enum.map(handles, &elem(&1, 0))
         shape = %Shape{shape | shape_dependencies_handles: handles}
 
-        :ets.update_element(meta_table, {@shape_meta_data, handle}, {2, shape})
+        :ets.update_element(meta_table, handle, {2, shape})
```
alco: We're using module attributes for offsets in other `:ets.update_element` calls in this module. It would be nice to use one here too.

msfstef: oops missed that! agree

packages/sync-service/lib/electric/shape_cache/shape_status.ex
msfstef: italian-core typo

packages/sync-service/lib/electric/shape_cache/shape_status.ex
msfstef: Opened task for this as well https://github.com/electric-sql/electric/issues/3474

PR #3477, opened 2025-11-20
fix: Avoid edge case with 409 causing more SSE requests

msfstef:
Fix test flake (see [this run](https://github.com/electric-sql/electric/actions/runs/19533935381/job/55923347981) for example) caused by 409s for the shape resetting the fallback from SSE to long polling.

Not sure if 409s _should_ reset this fallback but it does, and the test should account for it as we occasionally see these 409s from the tests overlapping each other.

### DIFF
--- a/packages/typescript-client/test/client.test.ts
+++ b/packages/typescript-client/test/client.test.ts
@@ -2276,7 +2276,7 @@ it(
   async ({ issuesTableUrl, aborter }) => {
     let requestCount = 0
     let sseRequestCount = 0
-    let initialHandle: string | null = null
+    let initialHandle: string | undefined
     const requestUrls: string[] = []
 
     // Mock console.warn to capture the fallback warning
@@ -2300,7 +2300,7 @@ it(
         const reqHandle = urlObj.searchParams.get(SHAPE_HANDLE_HEADER)
 
         if (isSSE && (!initialHandle || reqHandle === initialHandle)) {
-          initialHandle ??= reqHandle
+          initialHandle ??= reqHandle!
           sseRequestCount++
           // Handle up to 4 SSE requests (we expect 3, but might see 4 due to timing)
           if (sseRequestCount <= 4) {


packages/typescript-client/test/client.test.ts
```diff
@@ -2275,6 +2276,7 @@ it(
   async ({ issuesTableUrl, aborter }) => {
     let requestCount = 0
     let sseRequestCount = 0
+    let initialHandle: string | null = null
```
kevin-dp: Why `string | null` and not `string | undefined`, then we could shorten it to `let initialHandle: string | undefined`

packages/typescript-client/test/client.test.ts
```diff
@@ -2314,7 +2319,7 @@ it(
               status: 200,
               headers: new Headers({
                 'electric-offset': `0_0`,
-                'electric-handle': `test-handle`,
+                'electric-handle': reqHandle!,
```
kevin-dp, line 2322: nit: i think the `!` operator would be more appropriate in the definition of `const reqHandle` rather than here.

msfstef: I disagree - there will be requests made with the fetch client that do not include a handle (initial request)
If it is an SSE live request, it is guaranteed to have a handle, but I'd rather not do a nested check

packages/typescript-client/test/client.test.ts
msfstef: because I'm setting `intialHandle ??= reqHandle`, and `reqHandle` can be `string | null` at that point.

that being said I assert it later one so I could just assert it here.

kevin-dp: > because I'm setting `intialHandle ??= reqHandle`, and `reqHandle` can be `string | null` at that point.
> 
> that being said I assert it later one so I could just assert it here.

Ah ok, leave it like this than! :)

packages/typescript-client/test/client.test.ts
kevin-dp: Oh ok missed that, judging by the code i though it would always be present.

PR #3479, opened 2025-11-20
chore(sync-service): Improve `PureFileStorage.init_writer/1` performance

magnetised:
With shape consumer processes coming and going, this storage init call is being done a lot more. A simple timer showed it taking 1-4 milliseconds in a running system.

This pr refactors the filesystem initialization code to remove filesystem accesses as much as possible and cuts the `init_writer` call by ~50%.

It also fixes a bug where new storage instances would immediately call `cleanup!/1` on themselves (even though they didn't exist).

[Benchmark code](https://gist.github.com/magnetised/f4214110bea23896d6194846dec108db) shows the following improvement:

Before:

```
##### With input empty #####
Name                  ips        average  deviation         median         99th %
init_writer        1.14 K      880.05 μs     ±8.89%      872.38 μs     1073.97 μs

##### With input used #####
Name                  ips        average  deviation         median         99th %
init_writer        900.92        1.11 ms    ±17.00%        1.16 ms        1.43 ms
```

After:

```
##### With input empty #####
Name                  ips        average  deviation         median         99th %
init_writer        1.79 K      559.23 μs    ±12.67%      558.63 μs      716.82 μs

##### With input used #####
Name                  ips        average  deviation         median         99th %
init_writer        1.70 K      589.83 μs    ±30.58%      637.21 μs      931.23 μs
```

Not earth shattering but this is very much in the critical path.

### DIFF
--- a/.changeset/ten-ladybugs-argue.md
+++ b/.changeset/ten-ladybugs-argue.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Improve storage initialization performance


msfstef (approved):
Very nice - cutting down on FS ops is significant

PR #3480, opened 2025-11-20
chore(sync-service): Fix more test flakes

magnetised:


packages/sync-service/lib/electric/postgres/lock_breaker_connection.ex
```diff
@@ -99,18 +101,20 @@ defmodule Electric.Postgres.LockBreakerConnection do
   @impl true
   def notify(_, _, _), do: :ok
 
-  defp lock_breaker_query(lock_name, lock_connection_backend_pid, database) do
+  defp lock_breaker_query(lock_name, lock_connection_backend_pid, database)
+       when is_integer(lock_connection_backend_pid) or is_nil(lock_connection_backend_pid) do
```
msfstef, line 105: if this is fixing anything it might be worth a changeset since it's not test code

magnetised: It's not really fixing anything just asserting that we don't need to quote the value

PR #3482, opened 2025-11-20
feat: Speed up and correct shape recovery into `ShapeStatus`

msfstef:
Part of a larger piece of work in https://github.com/electric-sql/electric/issues/3442

Precursor to https://github.com/electric-sql/electric/issues/3451

I was looking into _how_ slow recovering all of the shapes into shape status is to determine whether it is worth doing lazily, and discovered 2 very important things:

1. We were sequentially loading all the shapes, so we were super slow. I've parallelised the loading and saw significant gains.
2. As pointed out by @alco in a previous PR, we were loading an arbitrary offset into the shape status, cause we were relying on recovered shapes starting consumers that would set _the correct one_. With lazy shapes, we never fixed this, and I believe it breaks readers unless a shape receives new data.

For 1. the fix was fairly easy, start async tasks under the stack supervisor and load up. An additional performance hack I put in is that for boolean metadata we can simply have or not have a file, and use a simple `stat` check that does not use a file handle and is much faster to do.

For 2., the fix was to load the latest metadata from storage as well as the snapshot started that we were already loading. This slows down recovery significantly as expected, making it clear that lazily loading metadata is important.

## Performance
I was recovering 50k shapes on my machine (m2 macbook air):
- Before optimization (loading shape def + snapshot started, sequentially) : ~10-20k ms
- After optimization (same behaviour, in parallel): ~ ~2-3k ms
- After optimization (with latest offset, in parallel): ~4-6k ms

We saw a ~5-6x speed gain just by parallelising loading, but adding latest offset roughly doubled the time needed (as it was one more read per shape).

I think these numbers, and how they scale, make it clear that lazy loading thatd data is the way to go. I have an approach that will keep it entirely encapsulated to `ShapeStatus` so I think it will be a decent implementation.

Also worth noting that while writing this I still think we're missing significant test coverage for recovery. We have a basic crash recovery test but it doesn't cover e.g. recovering a shape that has more than just its snapshot available.

edit: created companion PR with failing integration test https://github.com/electric-sql/electric/pull/3485

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
@@ -181,7 +181,7 @@ defmodule Electric.ShapeCache.PureFileStorage do
               _ -> []
             end
           end,
-          timeout: 30_000,
+          timeout: :infinity,
           ordered: false
         )
         |> Enum.flat_map(fn {:ok, res} -> res end)


packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -171,30 +172,48 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+        Task.Supervisor.async_stream(
+          stack_task_supervisor(stack_opts.stack_id),
+          shape_handles,
+          fn handle ->
+            case recover_stored_shape(stack_opts, handle) do
+              {:ok, shape_data} -> [{handle, shape_data}]
+              _ -> []
+            end
+          end,
+          timeout: 30_000,
```
magnetised: timeout makes me nervous. it's like saying 'crash if we have "too many" shapes'...

magnetised (changes requested):
This is great, but let's get rid of the timeout for loading the shapes

magnetised (approved):

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
msfstef: > `:timeout` - the maximum amount of time to wait (in milliseconds) without receiving a task reply (across all running tasks). Defaults to 5000.

From [the docs](https://hexdocs.pm/elixir/Task.Supervisor.html#async_stream/6) - I explicitly wrote `:infinity` and removed it because of how this was written but should've just tested it. The timeout is _per task_, but the docs made it seem like "if no reply is received from _any_ task within X time, time out" 😨 should've gone with my gut

alco: If you read the next item in the docs:
> `:on_timeout` - what do to when a task times out. The possible values are:
> * `:exit` (default) - the process that spawned the tasks exits.
> * `:kill_task` - the task that timed out is killed. The value emitted for that task is `{:exit, :timeout}`.

So Garry's gut feeling was partly correct: setting a timeout here would mean the whole recovery fails if one shape recovery takes long for any reason.

We really need to be intentional about whether we abort shape recovery if one shape misbehaves.

PR #3483, opened 2025-11-20
(chore) Refactor operation batching

robacourt:
Addressing @msfstef 's feedback on #3465 this PR:
- Separates concerns of metrics and message converting
- Makes txn handling in Replication client more explicit
- Removes the case where `handle_operations` is called with an empty list of operations

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -59,9 +59,6 @@ defmodule Electric.Replication.ShapeLogCollector do
   # the new txn to disk, instead the storage backend is responsible for
   # determining how long a write should reasonably take and if that fails
   # it should raise.
-  @spec handle_operations([Changes.operation()], any()) :: :ok | {:error, any()}
-  def handle_operations([], _server), do: :ok
-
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/postgres/replication_client.ex
```diff
@@ -423,6 +424,20 @@ defmodule Electric.Postgres.ReplicationClient do
   # Since we process a message at a time and flush the buffer
   # on commit, we will only ever have one commit here
   defp acknowledge_transactions([commit], state) do
+    if Sampler.sample_metrics?() do
+      OpenTelemetry.execute(
+        [:electric, :postgres, :replication, :transaction_received],
+        %{
+          monotonic_time: System.monotonic_time(),
+          receive_lag: DateTime.diff(DateTime.utc_now(), commit.commit_timestamp, :millisecond),
```
alco, line 432: Something to think about: we have two calls to get the current time per incoming Commit message. I think we could bring them down to one. We could also quantize it by getting the current time using a timer, to ensure we only do it once every, say, 10ms.

Not for this PR, just a thought I've had.

robacourt: Good idea. One for another PR!

PR #3485, opened 2025-11-20
fix: Crash recovery fails because of invalid restored offset

msfstef:
Companion to https://github.com/electric-sql/electric/pull/3482

Updated the integration test to show that it currently fails to read recovered shapes, and it only worked because it happened to only test the snapshot.

The recovered shapes are not woken up and thus their latest offset is not updated, causing the readers to fail with out of bounds error.


magnetised (approved):
Nice

Issue #3484, opened 2025-11-20
Bug Report: Shape Cache Collision Between Different log_mode Values


KyleAMathews:
## Problem Description

We've discovered a critical bug where shapes with different `log_mode` values (`:full` vs `:changes_only`) but otherwise identical parameters are treated as the same shape by the shape cache. This causes data availability issues in production.

### Reported Symptoms (from Slack)

From Marius on staging:

> A shape doesn't return any data (it returns the up-to-date status, but it's definitely not up-to-date)
Key observations:
- Issue appeared after using on-demand mode (which uses `log=changes_only`)
- Clearing the cache through the cloud console temporarily fixed it
- Even on a branch where on-demand mode was disabled, shapes still exhibited `changes_only` behavior
- The shape returned `snapshot-end` but no data rows

## Root Cause Analysis

The `Shape.comparable()` function (`packages/sync-service/lib/electric/shapes/shape.ex:98-103`) is used as the cache key for shape lookups in ETS. Currently it includes:

- `root_table_id` / `root_table`
- `root_pk`
- `where` clause
- `selected_columns`
- `flags` (sorted)
- `replica` type

**But it does NOT include `log_mode`.**

This means:
```elixir
# These two shapes are considered IDENTICAL by the cache:
shape1 = Shape.new!("offers", log_mode: :changes_only)
shape2 = Shape.new!("offers", log_mode: :full)

Shape.comparable(shape1) == Shape.comparable(shape2)  # TRUE - BUG!
```

### Cache Lookup Flow

1. `ShapeStatus.add_shape/2` inserts: `{Shape.comparable(shape), shape_handle}` into ETS
2. `ShapeStatus.get_existing_shape/2` looks up by: `Shape.comparable(shape)`
3. If found, returns existing handle

Source: `packages/sync-service/lib/electric/shape_cache/shape_status.ex:136,250`

## Reproduction Scenario

1. **Client A** requests `GET /v1/shape?table=offers&log=changes_only`
   - Shape created with `log_mode: :changes_only`
   - Handle "12345-1732123456789012" assigned
   - Initial snapshot is empty (by design - see `Querying.stream_initial_data/4:59-61`)
   - Cache stores: `{comparable(shape_with_changes_only), "12345-..."}`

2. **Client B** requests `GET /v1/shape?table=offers` (defaults to `log=full`)
   - ShapeCache calls `get_existing_shape()`
[...]

PR #3486, opened 2025-11-20
fix: Shape comparable should include log mode to avoid collisions

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3484

What I was worried about happened - the comparable was not updated as we updated the shape. I think we should have a better more solid way to define the comparable.

Also updating this will kind of invalidate existing stored shapes if using shape status backup, but not quite sure what the best way to do that is.

I've invalidated it by bumping the shape status backup version so this works now.

### DIFF
--- a/packages/sync-service/test/electric/shapes/shape_test.exs
+++ b/packages/sync-service/test/electric/shapes/shape_test.exs
@@ -785,5 +785,23 @@ defmodule Electric.Shapes.ShapeTest do
         end
       end
     end
+
+    test "log_mode affects the equivalence", %{inspector: inspector} do
+      shape_fun = fn mode ->
+        Shape.new(~S|the_table|,
+          inspector: inspector,
+          log_mode: mode
+        )
+      end
+
+      {:ok, shape1} = shape_fun.(:full)
+      {:ok, shape2} = shape_fun.(:changes_only)
+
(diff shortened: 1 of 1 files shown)


balegas:
Tests are already broken so they are doing their work :D. Claude-generated fix incoming

balegas:
I've pushed a fix for the tests. When deleting a shape we were always passing `table`, but now since we also have the `log` mode in the comparison the shape would not  match. We could pass the `log` mode explicitly, but it seems that when we have a shape handle there is no good reason for passing the shape definition in the delete, so I just stopped passing it.

msfstef:
>Wat no test? Could we have some property test that mutates all the keys and asserts the comparable-ness?

Sorry yeah opened the PR last thing before leaving haha just for reference of what the fix is - property test was exactly what I was thinking as well

>I'm also really nervous about this need to keep updating @backup_version... At least some scary comment in Shape.comparable about the dependency (or could we automate it? Calculate the version from the comparable fields or something?)

I'm updating it even without releasing electric just for the sake of consistency of the commits and the need to update it. I agree that automating it somehow would be better. Either way recovery right now is broken anyway (see https://github.com/electric-sql/electric/pull/3482)

I have an idea to not rely on shape comparable in general but will elaborate when back

PR #3496, opened 2025-11-25
chore(sync-service): Reduce PublicationManager memory usage

### LINKED ISSUE #3492: PublicationManager.RelationTracker high memory usage
On a system with 150k defined shapes the RelationTracker is using 275MB of memory within the BEAM. This seems excessive considering that the only unique information it holds are reference counts for table -> defined shapes and whether a given shape's relation has been added to the publication.

magnetised:
Track state using just the table `oid`, not an `{oid, {schema, table}}` tuple and maintain a separate map of `oid => {schema, table}`.

Also build initial state in an ephemeral task so that we immediately garbage collect the results of loading the shape status table into RAM.

On a system with 150,000 defined shapes, this reduces the initial memory usage of PublicationManager from a steady state ~70MB (~270MB before GC) to a consistent ~21MB.

Also found and fixed a bug when cleaning relations -- we missed an argument change in the call to ShapeCleaner.cleanup_shapes_async/2 -- and fixed the behaviour of waiting for publication updates when failing (the start time was re-calculated every attempt, so the test never timed out).

Fixes #3492

### DIFF
--- a/packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
+++ b/packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
@@ -144,6 +144,8 @@ defmodule Electric.Replication.PublicationManager.RelationTracker do
       [],
       state.stack_id,
       fn ->
+        # Build initial state in an ephemeral Task process so that to avoid
+        # retaining the data from list_shapes in this process's heap.
         state =
           Task.async(fn ->
             state.stack_id
@@ -475,7 +477,6 @@ defmodule Electric.Replication.PublicationManager.RelationTracker do
       end)
 
     if not Enum.empty?(to_fail) do
-      dbg(to_fail)
       # schedule removals for any tracked shapes that require generated columns
       handles = for {handle, _} <- to_fail, do: handle
 


packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
```diff
@@ -141,18 +145,21 @@ defmodule Electric.Replication.PublicationManager.RelationTracker do
...
-            state,
-            fn {shape_handle, shape}, state ->
-              add_shape_to_publication_filters(
-                shape_handle,
-                get_publication_filter_from_shape(shape),
-                state
-              )
-            end
-          )
+          Task.async(fn ->
```
alco, line 150: Please add a comment explaining that this is done in a task to prevent the main process' heap from growing and staying high.

packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
```diff
@@ -419,31 +452,34 @@ defmodule Electric.Replication.PublicationManager.RelationTracker do
...
         else
           waiters
         end
       end)
 
-    # schedule removals for any tracked shapes that require generated columns
-    for {handle, _} <- to_fail do
-      ShapeCleaner.remove_shape_async(handle, stack_id: state.stack_id)
+    if not Enum.empty?(to_fail) do
+      dbg(to_fail)
```
alco: leftover

magnetised: bugger

PR #3497, opened 2025-11-25
feat(client): add types for move messages

icehaunter:


packages/typescript-client/src/types.ts
```diff
@@ -43,10 +43,28 @@ export type NormalizedPgSnapshot = {
 }
 
 interface Header {
-  [key: Exclude<string, `operation` | `control`>]: Value
+  [key: Exclude<string, `operation` | `control` | `event`>]: Value
```
kevin-dp, line 46: The `Exclude` isn't doing anything here because `string` is too general. `Exclude<string, `operation` | `control` | `event`>` type checks to just `string`. For example, no type error is thrown here:
```ts
interface Header {
  [key: Exclude<string, `operation` | `control` | `event`>]: any
}

const test: Header = {
  a: 5,
  event: 'bar'
}
```

I'm not sure how it will show up in IntelliSense. If it shows the full excluded type, it is still useful as it serves some documentation purposes but it might be that it just shows up as `string` in IntelliSense.

packages/typescript-client/src/types.ts
```diff
@@ -43,10 +43,28 @@ export type NormalizedPgSnapshot = {
...
 }
 
 export type Operation = `insert` | `update` | `delete`
+/**
+ * A tag is a list of string identifying a reason for this row to be part of the shape.
```
kevin-dp: nit: typo "a list of string" -> "a list of strings". Also i'd rather say "array" than "list".

kevin-dp: Also "identifying a reason" but in the case of composites there are multiple reasons, so i would put this in plural.

packages/typescript-client/src/types.ts
```diff
@@ -43,10 +43,28 @@ export type NormalizedPgSnapshot = {
...
+ *
+ * Tags can be composite, which is why they are a list. Within one shape, tag width is fixed and
+ * is dependent on the where clause, so it's functionally always a tuple.
```
kevin-dp: Tuples isn't standard in JS. Probably clearer if we clarify "(i.e. a fixed size array)".

packages/typescript-client/src/types.ts
```diff
@@ -43,10 +43,28 @@ export type NormalizedPgSnapshot = {
...
+ */
+export type MoveTag = string[]
+
+/**
+ * A move-out pattern is a list of strings with at least one non-wildcard element specifying which
+ * rows will no longer be part of the shape.
+ *
+ * Actual values within the tuple must be considered opaque if they are not a literal `"*"`.
+ */
+export type MoveOutPattern = (string | `*`)[]
```
kevin-dp: The `string` type already encompasses the string `"*"`. So it's just `string[]`.

kevin-dp (approved):
LGTM! Left a few nits.

packages/typescript-client/src/types.ts
icehaunter: I know it does, but it still shows up like that and is reasonable for docs I think, even if the type unionizes into a sttring

PR #3498, opened 2025-11-25
Delete ETS entries in PureFileStorage for terminated shapes

### LINKED ISSUE #3491: mem leak: PureFileStorage's stack ETS table keeps accumulating entries without deleting them
This [recent change](https://github.com/electric-sql/electric/commit/6bb011b8d8411ec4710a918fdbd039af5f3a446c#diff-98aa3361de9882a623be71bfd7a9ee80a83a2e5721575a2320d0c645a1f2136fR628-R631) removed the call to `Storage.terminate/1` when shape consumer terminates. The unintended consequence of that is that an expired shape doesn't clean up its entry in PureFileStorage's stack ETS table. So as shapes get expired constantly, more and more garbage entries are accumulated in the table and are never cleaned up until the stack restart.

alco:
Fix #3491.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -638,7 +638,14 @@ defmodule Electric.Shapes.Consumer do
   # 2. request a notification when all active shape data reads are complete
   # 3. exit the process when we receive that notification
 
-  defp mark_for_removal(state), do: %{state | terminating?: true}
+  defp mark_for_removal(%{terminating?: true} = state) do
+    state
+  end
+
+  defp mark_for_removal(state) do
+    # remove the writer state to save on a pointless Storage.terminate/1 call
+    Map.delete(%{state | terminating?: true}, :writer)
+  end
 
   defp stop_with_reason(reason, state) do
     {reason, state} =


packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -638,14 +638,7 @@ defmodule Electric.Shapes.Consumer do
   # 2. request a notification when all active shape data reads are complete
   # 3. exit the process when we receive that notification
 
-  defp mark_for_removal(%{terminating?: true} = state) do
-    state
-  end
-
-  defp mark_for_removal(state) do
-    # remove the writer state to save on a pointless Storage.terminate/1 call
-    Map.delete(%{state | terminating?: true}, :writer)
```
magnetised, line 647: Why  re-factor the ets deletes to happen at cleanup and then re-enable the terminate call which will result in writes to files that are about to be deleted? Surely we can avoid properly terminating the consumer storage knowing that a) the data is irrelevant and can be discarded and b) the ets will be cleaned up by the Storage.cleanup! call

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -227,14 +245,19 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
-    Electric.AsyncDeleter.delete(
-      stack_opts.stack_id,
-      shape_data_dir(stack_opts.base_path, shape_handle)
-    )
+    with :ok <-
+           Electric.AsyncDeleter.delete(
+             stack_opts.stack_id,
+             shape_data_dir(stack_opts.base_path, shape_handle)
+           ) do
+      delete_shape_ets_entry(stack_opts.stack_id, shape_handle)
```
magnetised, line 253: moving the ets deletes to the cleanup is a great idea.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -227,14 +245,19 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+    end
   end
 
   def cleanup_all!(%{stack_id: stack_id, base_path: base_path}) do
-    Electric.AsyncDeleter.delete(stack_id, base_path)
+    with :ok <- Electric.AsyncDeleter.delete(stack_id, base_path) do
+      drop_all_ets_entries(stack_id)
```
magnetised, line 259: should we not just cleanup the ets even if the AsyncDeleter call fails?

magnetised (approved):
Looks good but see my question re terminate vs cleanup. @msfstef would want us to not write to the filesystem if we can avoid it... 😜

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
msfstef: I don't mind this approach as AsyncDeleter operations are atomic - so if they fail the whole cleanup should have basically touched nothing, which is a nice property I think

packages/sync-service/lib/electric/shapes/consumer.ex
msfstef: Agree that this seems like an unnecessary change (?) unless I'm missing something

msfstef (approved):
Important fix to get in!

PR #3499, opened 2025-11-25
chore(sync-service): Cleanup API struct and GC http handler processes

magnetised:
API instance was carrying a lot of fields it didn't need anymore. This doesn't affect memory usage unfortunately. Setting `fullsweep_after` for the Thousand Island acceptor processes should though -- since they're re-used they accumulate a lot of binaries. 

The lower `fullsweep_after` is the higher the cpu cost so I've set it to `10` as the default. `0` really cleans up the memory of the processes but is not reasonable for prod.

https://www.erlang.org/doc/apps/erts/erlang.html#spawn_opt/4

Update: after the discussion below I've changed this to use `max_requests` to shutdown http connection handler processes after some number of requests. I don't think this will have a huge effect on RAM usage overall but it will deal with the kind of problems we saw after the CF outage, where request processes accumulated messages

### DIFF
--- a/.changeset/tame-glasses-tan.md
+++ b/.changeset/tame-glasses-tan.md
@@ -2,4 +2,4 @@
 '@core/sync-service': patch
 ---
 
-Cleanup API config and set fullsweep_after for conn processes
+Cleanup API config and terminate request handler processes periodically
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -248,7 +248,7 @@ config :electric,
     ),
   process_registry_partitions: env!("ELECTRIC_TWEAKS_PROCESS_REGISTRY_PARTITIONS", :integer, nil),
   http_api_num_acceptors: env!("ELECTRIC_TWEAKS_HTTP_API_NUM_ACCEPTORS", :integer, 100),
-  conn_fullsweep_after: env!("ELECTRIC_TWEAKS_CONN_FULLSWEEP_AFTER", :integer, nil),
+  conn_max_requests: env!("ELECTRIC_TWEAKS_CONN_MAX_REQUESTS", :integer, nil),
   tcp_send_timeout:
     env!("ELECTRIC_TCP_SEND_TIMEOUT", &Electric.Config.parse_human_readable_time!/1, nil),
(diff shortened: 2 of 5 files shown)


alco:
The memory vs CPU tradeoff is worrying.

As suggested before during our call, I would try going the way of shutting down connection processes after handling N requests instead - see the `max_requests` option [here](https://hexdocs.pm/bandit/Bandit.html#t:http_1_options/0).

There is [this thread](https://elixirforum.com/t/what-do-you-think-could-be-the-best-way-to-garbage-collect-in-bandit/62144) on the Elixir Forum where the author of Bandit talks about adding an internal GC after handling every 5 requests in version 1.3.0. So, unless it's been reverted (which I haven't seen in the Changelog), we should have seen the effect of already?

This all seems to be specific to HTTP1. Can we sidestep the issue entirely by having Cloudflare/AWS LB access our endpoint using HTTP2? According to Bandit's [process model for h2](https://github.com/mtrudel/bandit/blob/main/lib/bandit/http2/README.md), there is no problem of long-lived connection processes there: Cloudflare Worker/AWS LB may keep a number of long-lived connections open but actual client requests should be encapsulated in HTTP streams and those are handled by short-lived processes inside Bandit.

magnetised:
@alco as usual thanks for the excellent research...

> There is this thread on the Elixir Forum where the author of Bandit talks about adding an internal GC after handling every 5 requests in version 1.3.0. So, unless it's been reverted (which I haven't seen in the Changelog), we should have seen the effect of already?

I don't know what's going on here but the heap sizes for request processes within e.g. turbo are ~400KB, 

output from observer_cli:

```
heap_size        | 85.6094 KiB  
total_heap_size  | 448.8125 KiB 
```

which adds up. @msfstef and I talked about killing procs after a number of requests after the CF outage, which showed handler processes with ~6 million messages in their queue.

I think our setup is complex because we have connections coming in via the CF tunnel, which means they're very long lived. Even so I would have expected the GC setting you point to having some effect.

Note that in the discussion you linked to they set the fullsweep_after to `0`, which, at least locally, made a massive difference to the mem usage of request processes, down from ~106K to a consistent ~30K. I opted for a less CPU intensive setting of 10, because, you know, compromise...

alco (approved):

PR #3505, opened 2025-11-26
chore(sync-service): Hibernate ShapeStatusOwner process

magnetised:
If the recovery process is unable to restore the status ets tables from backup files and has to iterate the filesystem then ShapeStatusOwner is left with a lot of accumulated memory (~300MB for a system with 200K shapes).

Release this by hibernating immediately (this process does nothing but own ets tables anyway).

### DIFF
--- a/.changeset/eleven-radios-sit.md
+++ b/.changeset/eleven-radios-sit.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Hibernate the shape status owner process to release any memory accumulated during startup
--- a/packages/sync-service/lib/electric/shape_cache/shape_status_owner.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status_owner.ex
@@ -43,7 +43,7 @@ defmodule Electric.ShapeCache.ShapeStatusOwner do
     :ok = ShapeStatus.initialize_from_storage(stack_id, config.storage)
     :ok = Electric.LsnTracker.initialize(stack_id)
 
-    {:ok, %{stack_id: stack_id, storage: config.storage}}
+    {:ok, %{stack_id: stack_id, storage: config.storage}, :hibernate}
   end
 
   @impl true


alco:
I've been there before - https://github.com/electric-sql/electric/pull/3295.

It feels a bit dirty, especially without any explanatory code comments.

I like the way you encapsulated temporary memory requirements by doing stuff in an async task in https://github.com/electric-sql/electric/pull/3496. I would prefer we do the same thing here:
- create an async task for loading data from backup and pass it the PID of the ShapeStatusOwner process
- if the ETS table is loaded from backup, call `:ets.give_away()` on it to transfer ownership to ShapeStatusOwner
- if new tables are created, set their `heir` appropriately so that ownership transfer happens after the async task terminates

magnetised:
@alco sorry I merged this before/as you were commenting. I read your previous pr and I think you were right with a lot of the work you did there. Reading your proposed alternative above does it not strike you that a simple `:hibernate` reply that does the same thing is not the better solution? This process literally does nothing after init so why not just hibernate it and be done.

alco:
> This process literally does nothing after init so why not just hibernate it and be done.

That's a perfect question to ask! Why do we need such a process at all?

All the tables it creates are public. We only want to make sure they survive unexpected restarts of both the connection subsystem and the shape subsystem. With the scaledown feature enabled, they also survive the scaledown of replication connection.

So their lifetime is tied to the stack itself. We could replace the `ShapeStatusOwner` process with `ShapeStatus.Initializer` that is either a wrapper for `Task` that executes the initialization code for shape status tables. In the end all table ownerships should be transferred to StackSupervisor or StackConfig, if you prefer not to add on stuff to a supervisor process.

magnetised:
I understand the need for the process and think we should retain the role (a process that just owns public tables for a stack is good), I guess I just don't understand the need for such complexity around memory management in this one case. And yes, we could refactor as you say and probably should at some point, but IMHO this works for now and with no risk of accidental bugs.

msfstef:
checking in to say that I agree that if the process only owns the tables we could move it to the stack supervisor, but also think that a single line change to hibernate a process that does that right now is the best immediate solution with no side effects. I would have appreciated a comment on the memory stuff though for posterity!

In the case of the changes you had made @alco, we were hibernating processes like the ShapeCache which would've been _immediately_ woken up, and therefore hibernating just for the memory management, which feels odd. In this case the process actually does nothing, so I think it makes sense both from a memory management perspective but also for the semantics of the process itself.

Issue #3511, opened 2025-11-27
Shape lookup data uses a lot of RAM


magnetised:
We currently validate very read request by using the request parameters to define a shape, then retrieving the shape's handle by using the shape as a key to lookup the handle in an ETS table. We do this for every request, even if the params include a `handle`.

This means that our api requires shape information to be held in fast in-memory storage.

However this shape data is massive. The shape metadata table and associated shape -> handle lookup table require hundreds of MB of RAM on a server with thousands of shapes.

We need to reduce this RAM usage and move to some form of on-disk storage of the shape information.

PR #3515, opened 2025-11-27
feat(sync-service): Validate request handles using shape hash

### LINKED ISSUE #3512: Validate API request handles without requiring a shape -> handle LUT
We can validate that a given handle matches the parameter shape definition using a simple check that the hash of the shape defined by the request parameters matches the hash for the shape under the given handle.

This reduces the requirement for a lookup of shape -> shape handle to the edges of a session: either the first request where we don't have a shape handle or the moment where a shape has been deleted.

magnetised:
This is part 1 of #3511.

Rather than loading a shape handle from the ShapeStatus LUTs using the Shape definition and then checking that it matches the one in the request params, we instead do a simple check that the request's handle and shape definition match using a hash equality check.

We only do a shape def -> handle lookup if the request doesn't have a handle parameter or the hash check fails.

I've also replaced use of `fetch_shape_by_handle` as the implementation of `has_shape?` with a lookup that just uses metadata.

This is an intermediate step - the next PR will move the shape out of the meta table completely and encapsulate shape lookups as a separate module.

Fixes #3512

### DIFF
--- a/packages/sync-service/lib/electric/shapes/shape.ex
+++ b/packages/sync-service/lib/electric/shapes/shape.ex
@@ -89,6 +89,7 @@ defmodule Electric.Shapes.Shape do
           shape_dependencies: [json_safe(), ...],
           log_mode: log_mode()
         }
+  @type comparable() :: term()
 
   @doc """
   Return a comparable representation of the shape.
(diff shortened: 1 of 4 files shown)


msfstef:
Beat me to it, this is exactly what I was also thinking while working on shape status! gmta

(can't review but saw this pop up and was happy haha)

packages/sync-service/lib/electric/shapes/shape.ex
```diff
@@ -110,10 +112,16 @@ defmodule Electric.Shapes.Shape do
   def hash(%__MODULE__{} = shape),
```
msfstef: I'm against having two separate places in code that return the "hash" of the shape, as they risk going out of sync with each other. I think we can deprecate this hash call and use `comparable_hash` everywhere we need the hash potentially. IMO a single public API returns the hash.

PR #3517, opened 2025-11-27
Parse feature flags into relevant config options

alco:
Can anyone name one good reason for threading feature flags as a list of strings through internal module function calls?

They are used as a convenient way to provide tweaks from the outside. Internally, they should be parsed into relevant options at the edge.

Unfortunately, Electric has multiple input edges, so the resulting code is not as simple as I'd have liked it to be.

### DIFF
--- a/packages/sync-service/lib/electric/application.ex
+++ b/packages/sync-service/lib/electric/application.ex
@@ -152,10 +152,15 @@ defmodule Electric.Application do
   # Gets the API-side configuration based on the same opts + application config
   # used for `configuration/1`
   defp api_configuration(opts) do
-    opts
-    |> core_configuration()
-    |> Electric.StackSupervisor.build_shared_opts()
+    {feature_flags, core_config} =
+      opts
+      |> core_configuration()
+      |> Electric.StackSupervisor.build_shared_opts()
+      |> Keyword.pop(:feature_flags)
+
+    core_config
     |> Keyword.merge(
+      allow_subqueries?: Electric.Config.feature_flag_allow_subqueries() in feature_flags,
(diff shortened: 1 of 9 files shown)


msfstef:
> Can anyone name one good reason for threading feature flags as a list of strings through internal module function calls?

I think the idea is that with a single property which is meant to hold temporary flags/guards for new features we can flexibly try out features without flooding our configs with flags that we plan to get rid of soon.

We should be validating the feature flags at the edge, but from there on I think it makes sense to keep it as a single thing (struct, list, string, whatever) to do additional guard checks for features as we try them. This way we don't add and remove flags in various parts of the system that we might end up forgetting to clean up etc.

magnetised:
@alco though I get your point, and your first line made me laugh, actually this is a good idea. Feature flags are temporary things, we're gating wip features per-deployment, so adding and removing them should be easy. If we have to wire the config through the stack for every temporary feature flag it will be a real burden.

alco:
I get the sentiment that we want to avoid adding config boilerplate for feature flags which are presumably temporary. But I disagree with the premise that passing flags around the codebase as a list of strings is good design.

The right mental model for feature flag is `Application` env: it's the globally available key-value store that any module can do lookups from. In our multitenant-capable app, `Electric.StackConfig` serves the equivalent role. So feature flags defined by the user should be parsed into named and typed config valuesl that are put into `Electric.StackConfig` and then looked up in those parts of the code that need them.

There should be no passing of `feature_flags` into any internal modules. Even though that's the easiest thing to do it also makes it easy to shoot oneself in the foot. Haven't we recently encountered a problem where the suspend_customers feature flags wasn't being applied after tenant restart in Cloud? :wink-wink:

PR #3518, opened 2025-11-27
chore(sync-service): Refactor ShapeStatus to encapsulate shape lookups

### LINKED ISSUE #3513: Isolate shape defn lookups within ShapeStatus
If we are to move to an on-disk format then we need to separate shape lookups (both shape defn to handle and handle to shape defn) from the other shape status information.

magnetised:
This is essentially what we have now but with a clear api between fast metadata lookups and shape -> handle and handle -> shape lookups.

This moves the %Shape{} instances out of the metadata table and gives a hint of the memory savings to come, this is for ~150K (very simple) shapes:

```
shapedb:shape_lookup:single_stack    | 149432       | 242.7575 MiB
shapedb:handle_lookup:single_stack   | 149432       | 83.1469 MiB
shape_meta_table:single_stack        | 149432       | 29.5633 MiB
shape_relation_lookup_table:single_s | 149432       | 19.3010 MiB
shape_last_used_table:single_stack   | 149432       | 15.8808 MiB
```

You can see that the shapedb tables are responsible for most of the ets usage and the metadata now only consumes ~30MiB, so a ~90% reduction, which means we shouldn't have to go any further and move more things out of ram.

The current api is not going to work for a sqlite backed version, but it's close I think.

Fixes #3513

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status/shape_db.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status/shape_db.ex
@@ -1,0 +1,181 @@
+defmodule Electric.ShapeCache.ShapeStatus.ShapeDb do
+  @moduledoc false
+
+  # Will eventually replace the current ETS lookup tables with a sqlite-backed
+  # shape db. Currently just an encapsulation of the shape-to-handle and
+  # handle-to-shape ETS lookups.
+
+  alias Electric.Shapes.Shape
+
+  import Electric, only: [is_stack_id: 1]
+
+  @type shape_handle() :: Electric.ShapeCacheBehaviour.shape_handle()
+  @type stack_id() :: Electric.stack_id()
+
+  # this is called if the load_backup call fails
+  def create(stack_id, _version) when is_stack_id(stack_id) do
+    Enum.each(tables(stack_id), fn table ->
+      create_table(table)
+    end)
+  end
+
+  @spec load(stack_id(), [{Shape.t(), Shape.comparable(), shape_handle()}]) :: :ok
+  def load(stack_id, shape_data) when is_stack_id(stack_id) do
+    {handle_lookup_data, shape_lookup_data} =
+      Enum.reduce(
+        shape_data,
+        {[], []},
+        fn {shape, comparable, shape_handle}, {handle_lookup_data, shape_lookup_data} ->
+          {
+            [{comparable, shape_handle} | handle_lookup_data],
+            [{shape_handle, shape} | shape_lookup_data]
+          }
+        end
+      )
+
+    :ets.insert(shape_to_handle_table(stack_id), handle_lookup_data)
+    :ets.insert(handle_to_shape_table(stack_id), shape_lookup_data)
+    :ok
+  end
+
+  def add_shape(stack_id, %Shape{} = shape, comparable_shape, shape_handle)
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/shape_cache/shape_status/shape_db.ex
```diff
@@ -0,0 +1,177 @@
+defmodule Electric.ShapeCache.ShapeStatus.ShapeDb do
+  @moduledoc false
```
msfstef, line 2: I'd appreciate a module doc explaining why this is extracted - I appreciate you have a plan but you know how it goes, short doc that explains the reason for this abstraction will go a long way.

msfstef (approved):
Refactoring looks good!

msfstef:
Pasting some thoughts here from personal chat:

I think as part of our wider design for this it's worth considering what these indexes are even for - cause the handle -> shape lookup is surprisingly underused.

It is basically only necessary for 1) consumer initialization (which right now could just read it from the shape def file directly) and 2) consumer re-initialization start_consumer_for_handle where it reads it in ShapeCache to restore dependencies, and I feel that we can definitely sort that out or incorporate it into (1) somehow as a single lookup

All other uses are just checks that the shape exists which I think you've already basically removed with your other PR that just hasn't been merged yet.

With that in mind, perhaps we don't need a handle -> shape lookup at all? we already have a file that you can get the definition from, we should examine the performance requirements and potentially keep it simple

And that gives us time to think about and design a central persisted store as a replacement of our current storage approach for metadata (rather than replacing it as a side effect)

Ultimately I think the idea of using sqlite for it is the right approach, it will give a great balance between in-memory and disk usage, great performance, my only concern is introducing it without a clear plan for replacing the current source of truth that is the files, and without having thought about the full consequences of atomicity of deletes etc

magnetised:
> With that in mind, perhaps we don't need a handle -> shape lookup at all? we already have a file that you can get the definition from, we should examine the performance requirements and potentially keep it simple

I see this from the other side. Rather than replacing lookups from some shape db to do the handle to shape lookup, with calls to the storage layer, we should be replacing the storage layer lookups with a centralised on-disk db.

I would like to see the removal of the shape data from the storage layer completely and asap and this is just part of an incremental move to removing this metadata responsibility from the storage api. 

We do need the handle -> shape mapping, even if it's not used much and I don't understand why we'd take a step backwards and replace a centralised db with a storage lookup knowing that that isn't the endgame.

> All other uses are just checks that the shape exists which I think you've already basically removed with your other PR that just hasn't been merged yet.

I've not removed it, just reduced the frequency of use and it's fundamental, unfortunately (in some form - we discussed changing the current approach but IIRC realised that at the end of the day we need some full comparable shape data somewhere for the shape -> handle lut).

>  the full consequences of atomicity of deletes etc

We don't have atomic deletes now either and rely on a carefully crafted deletion order across multiple ets tables. SQLite is atomic and will allow us to remove the handle -> shape and shape -> handle mappings atomically, which atm we can't do. so overall I think it's equivalent or a net win.

> And that gives us time to think about and design a central persisted store as a replacement of our current storage approach for metadata (rather than replacing it as a side effect)

This is a deliberate part of replacing the the storage for metadata, not a side effect. The question is I guess is how much of this change we put in a single pr and how much we spread it out. I do think we can do it incrementally.

PR #3521, opened 2025-11-28
Setup PR comment action for releases

KyleAMathews:
When there's a release, we write a comment to the related issues/PRs something like:

```
The PR fixing this issue (#123) has been released! :rocket:

The following packages include the fix:

- `@electric-sql/client@1.2.0`

Thanks for reporting!
```


KyleAMathews:
Yeah it's been really nice on tanstack/db

PR #3524, opened 2025-12-01
feat: added support for move tags on the shape class

icehaunter:


packages/typescript-client/src/shape.ts
```diff
@@ -176,15 +178,39 @@ export class Shape<T extends Row<unknown> = Row> {
...
+              if (message.headers.tags) {
+                // Remove old tags from index
+                const oldTags = this.#rowTags.get(message.key)
+                if (oldTags) {
+                  oldTags.forEach((tag) =>
+                    this.#removeTagFromIndex(tag, message.key)
+                  )
+                }
+                // Set new tags
+                const newTags = new Set(message.headers.tags)
```
kevin-dp, line 203: More efficient to add the incoming tags and remove only the removed_at tags instead of first removing all tags and then adding the new tags

kevin-dp (approved):

PR #3525, opened 2025-12-01
feat: better buffering on move-ins

icehaunter:
New buffering means that instead of buffering all transactions until move-in snapshot becomes known and then second time until results become available, essentially rebasing the stream on the move-in snapshot, we now do the opposite: rebase the move-in snapshot on whatever position the stream happens to be in. We achieve this by keeping track of any keys touched within a given snapshot window, and skipping those keys in the move-in snapshot because those changes have already been made visible from the stream.

What this achieves is we're buffering much less data (only keys for inserts/updates instead of full transaction bodies) and not blocking any consumers (which allows us to have 3+ layers of subqueries without consistency issues).

Aside from that, this PR also fixes:
- Issue with storage where reads near the start of the stream could yield a duplicate value (not reported before, but happened in tests for move-outs)
- Issue with shadowing where some operations that happened before a move-in was resolved were made visible when they shouldn't have
- Issue where a move-in followed by a move-out of the same value was still appended to the log

### DIFF
--- a/.changeset/blue-brooms-walk.md
+++ b/.changeset/blue-brooms-walk.md
@@ -1,0 +1,7 @@
+---
+'@core/sync-service': patch
+---
+
+fix: ensure correct log reading near the log start, especially when a move-in/out is a first thing in the shape log
+fix: ensure correct processing of move-in/move-out sequences affecting same values
+fix: ensure correct move-in handling without duplicated data
--- a/.changeset/eight-kids-flash.md
+++ b/.changeset/eight-kids-flash.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+feat: add support for better buffering strategy over move-ins, unlocking 3+ shape layers


packages/sync-service/lib/electric/shape_cache/crashing_file_storage.ex
```diff
@@ -27,6 +27,7 @@ defmodule Electric.ShapeCache.CrashingFileStorage do
   defdelegate hibernate(opts), to: PureFileStorage
   defdelegate compact(opts, keep_complete_chunks), to: PureFileStorage
   defdelegate append_move_in_snapshot_to_log!(name, writer_state), to: PureFileStorage
+  defdelegate append_move_in_snapshot_to_log_filtered!(name, writer_state, touch_tracker, snapshot), to: PureFileStorage
```
robacourt: Do we need both or can this new one replace the old one?

packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -2750,20 +2750,58 @@ defmodule Electric.Plug.RouterTest do
...
+                %{"headers" => %{"control" => "up-to-date"}}
+              ]} = shape_req(req, ctx.opts)
+    end
 
-      assert length(data1) == 4
+    @tag with_sql: [
+           "CREATE TABLE parent (id INT PRIMARY KEY, value INT NOT NULL)",
+           "CREATE TABLE child (id INT PRIMARY KEY, parent_id INT NOT NULL REFERENCES parent(id), value INT NOT NULL)"
+         ]
+    test "runtime move-ins/outs are handled correctly", ctx do
```
robacourt:
```suggestion
    test "move-outs while processing move-ins are handled correctly", ctx do
```

packages/sync-service/lib/electric/shapes/querying.ex
```diff
@@ -143,6 +144,22 @@ defmodule Electric.Shapes.Querying do
     ~s['#{escape_relation(root_table)}' || '/' || #{join_primary_keys(pk_cols)}]
   end
 
+  # Converts a tag structure to something PG select can fill, but returns a list of separate strings for each tag
+  # - it's up to the caller to interpolate them into thje query correctly
```
robacourt:
```suggestion
  # - it's up to the caller to interpolate them into the query correctly
```

Issue #3475, opened 2025-11-20
Restoring shape backup should aggressively handle untracked shapes


msfstef:
Currently, when restoring `ShapeStatus` backups, an integrity check is done by comparing the handles in the backup to the ones found in the file system. If they do not match up, the backup is discarded and the ones not in the file system are loaded from scratch.

We should instead prioritise the backup more aggressively, either cleaning up any untracked shapes on the file system or attempting to recover those, and only those, as well.

PR #3527, opened 2025-12-01
feat: Reconcile `ShapeStatus` ETS backup after integrity check

msfstef:
Closes https://github.com/electric-sql/electric/issues/3475

Part of a larger piece of work described in https://github.com/electric-sql/electric/issues/3442

Until now, we were doing a check of what shape handles we have stored with a cheap `ls` call on the shape directory, and if the handles did not match the ones found in the ETS backup, we would throw away the backup.

This is a bit _too_ aggressive, since the ETS backup is actually the best way to recover quickly, and is bound to fail often if we periodically back up the tables to hedge against unexpected crashes.

This PR allows for backup reconciliation. Rather than _either_ loading a full backup that is perfect _or_ loading all shapes from the files, it allows loading any shape handles not found in the backup from storage.

Basically, we load the ETS backup, we do an `ls` on the shapes dir, any handles not in the backup but in the storage are loaded from storage, and any handles in the backup but not in the storage are discarded.

### DIFF
--- a/.changeset/chilly-timers-divide.md
+++ b/.changeset/chilly-timers-divide.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Reconcile ShapeStatus ETS backup rather than invalidate when stored shape handles and backed up handles diverge.


magnetised (approved):
Now getting really aware of every call to the storage layer to recover things, but for where we are, this is a good move.

msfstef:
> Now getting really aware of every call to the storage layer to recover things

Yep - although if this code does what it says then worst case scenario we do our existing recovery mechanism, best case it loads few to no shapes at all.

PR #3528, opened 2025-12-02
chore(sync-service): Remove suspend_consumers feature flag

magnetised:
Behaviour is now determined solely by ELECTRIC_SHAPE_SUSPEND_CONSUMER setting, which defaults to false.

I've changed the behaviour when encountering unknown feature flag settings from an exception to a Logger.error so that changes to supported feature flags don't break deployments (and we don't have to keep old feature flags around once they're not needed).

I've updated the docs so that people can use this feature for their own deployments.

### DIFF
--- a/website/docs/api/config.md
+++ b/website/docs/api/config.md
@@ -288,7 +288,7 @@ Port that the [HTTP API](/docs/api/http) is exposed on.
     defaultValue="false"
     example="true">
 
-Whether to terminate Shape consumer processes after `ELECTRIC_SHAPE_HIBERNATE_AFTER` seconds. This saves on memory at the cost of slightly higher CPU usage. When receiving a transaction that contains changes matching a given shape, electric will start a consumer process to handle the update. If more transactions matching the shape appear within the time defined by `ELECTRIC_SHAPE_HIBERNATE_AFTER` then the consumer will remain active, if not it will be terminated.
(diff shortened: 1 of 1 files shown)


.changeset/tough-days-smoke.md
```diff
@@ -0,0 +1,6 @@
+---
+'@core/sync-service': patch
+'@electric-sql/docs': patch
```
msfstef, line 3: I don't actually know if this is necessary, I think the site always updates on deploy regardless - might be wrong though

website/docs/api/config.md
```diff
@@ -281,6 +281,32 @@ Port that the [HTTP API](/docs/api/http) is exposed on.
...
 </EnvVarConfig>
 
+### ELECTRIC_SHAPE_SUSPEND_CONSUMER
+
+<EnvVarConfig
+    name="ELECTRIC_SHAPE_SUSPEND_CONSUMER"
+    defaultValue="false"
+    example="true">
+
+Whether to terminate Shape consumer processes after `ELECTRIC_SHAPE_HIBERNATE_AFTER` seconds. This saves on memory at the cost of slightly higher CPU usage. When receiving a transaction that contains changes matching a given shape, electric will start a consumer process to handle the update. If more transactions matching the shape appear within the time defined by `ELECTRIC_SHAPE_HIBERNATE_AFTER` then the consumer will remain active, if not it will be terminated.
```
msfstef:
```suggestion
Whether to terminate shape consumer processes after `ELECTRIC_SHAPE_HIBERNATE_AFTER` seconds. This saves on memory at the cost of slightly higher CPU usage. When receiving a transaction that contains changes matching a given shape, the consumer process is started to handle the update. If more transactions matching the shape appear within the time defined by `ELECTRIC_SHAPE_HIBERNATE_AFTER` then the consumer will remain active, if not it will be terminated.
```

nit: consistent capitalisation of Shape
nit: avoiding references to Electric (or alternatively should capitalise)

Issue #3474, opened 2025-11-20
Lazily delete shape metadata after `def -> handle` pointer is gone


msfstef:
With the changes introduced in https://github.com/electric-sql/electric/issues/3450, we only need to remove the pointer from a shape definition to a given shape immediately, and the rest of the cleanup can happen asynchronously in its own time.

This is one more step in the work being done to make cleaning up shapes as lower priority to the system as possible, not blocking anything on the critical read or write path.

See https://github.com/electric-sql/electric/pull/3472#discussion_r2544957178 for argument by @magnetised

msfstef:
This is potentially still relevant despite the work on moving to an SQLite-backed metadata store (see @magnetised 's plan), as we can still clean up data other than the pointer outside of the critical path.

Perhaps still worth doing.

PR #3531, opened 2025-12-02
feat: Unlink and cleanup `ShapeStatus` metadata separately

msfstef:
Closes https://github.com/electric-sql/electric/issues/3474

Part of a larger body of work described in https://github.com/electric-sql/electric/issues/3442

This is the last of the tasks that might still be relevant even with @magnetised proposed move to an SQLite backed metadata store, as we still might want to have a "quick" unlinking of a handle from the shape definition synchronously in the critical removal path, and then schedule the orphaned metadata cleanup sweep later.

I decided to still implement this for the sake of closure on the changes I proposed that I still think are relevant, the only outstanding one being https://github.com/electric-sql/electric/issues/3451 which I do believe is not worth doing as lazy loading is a worse version of disk-backed DB.

I'm ok if we end up not merging this, but I think as a general API I think it might still make sense which is why I decided to at least prepare the PR.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -157,6 +157,13 @@ defmodule Electric.ShapeCache.ShapeStatus do
   def unlink_handle_from_shape(stack_ref, shape_handle) do
     stack_id = extract_stack_id(stack_ref)
     shape = ShapeDb.unlink_handle_from_shape!(stack_id, shape_handle)
+
+    :ets.update_element(
+      shape_meta_table(stack_ref),
+      shape_handle,
+      {@shape_meta_shape_hash_pos, nil}
+    )
+
     {:ok, shape}
   rescue
     ArgumentError ->
--- a/packages/sync-service/test/electric/shape_cache/shape_status_test.exs
+++ b/packages/sync-service/test/electric/shape_cache/shape_status_test.exs
@@ -114,6 +114,7 @@ defmodule Electric.ShapeCache.ShapeStatusTest do
 
     # After unlinking, the shape should not be findable via get_existing_shape
     refute ShapeStatus.get_existing_shape(state, shape)
+    refute ShapeStatus.has_shape_handle?(state, shape_handle)
 
     # But the shape should still be fetchable by handle (metadata still exists)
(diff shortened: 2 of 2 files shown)


packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -148,6 +148,17 @@ defmodule Electric.ShapeCache.ShapeStatus do
     :ets.select(shape_relation_lookup_table(stack_ref), patterns)
   end
 
+  @spec unlink_handle_from_shape(stack_ref(), shape_handle()) ::
+          {:ok, Shape.t()} | {:error, term()}
+  def unlink_handle_from_shape(stack_ref, shape_handle) do
+    stack_id = extract_stack_id(stack_ref)
```
magnetised, line 158: I think we also need to delete the metadata table entry here too. That's validating that a handle matches a shape for every request. If we remove the shape db entry then the metadata it would mean that no more reads come through for the shape. If we just remove the shapedb entry then a deleted handle will pass the `resolve_shape_handle` test without falling back to shape db.

msfstef: addressed by setting the shape hash to `nil` in the unlinking

As discussed this does make this whole thing less self-consistent because now `has_shape_handle?` will return false after unlinking, but `fetch_shape_by_handle` still returns the shape. I don't particularly like the API being so inconsistent so I'm contemplating whether this change is worth it given how things work now.

alco:
@msfstef what's the status of this PR?

msfstef:
@alco I was sort of hoping to get https://github.com/electric-sql/electric/pull/3572 in - but I think the core concept is still relevant, of deleting things async as long as the unlinking is done synchronously

I've refrained from merging this as it's not strictly necessary and don't want to step over @magnetised 's work

magnetised:
@msfstef  I think the right answer here is to add a tombstone table which guards all further shape lookups and allows for a single atomic "unlink" step that guards the rest of the deletion process which, being spread across multiple tables (and soon db tables), will never be consistent.

so I'm tempted to keep this work just for the new 2-phase delete.

magnetised (approved):

magnetised:
@msfstef Closing because the new sqlite version removes the need for this entirely -- the only non-sqlite data is the LRU information which is very quick to remove and hardly critical

PR #3540, opened 2025-12-02
Yield while parsing data so that the UI isn't blocked while loading larger shapes

KyleAMathews:
Add async parsing methods that yield to the main thread periodically, preventing the UI from becoming unresponsive when syncing shapes with many rows.

- Add yieldToMain() utility that uses scheduler.yield() when available, falling back to setTimeout(0)
- Add parseAsync() and parseSnapshotDataAsync() methods to MessageParser that yield every 1000 messages by default
- Update ShapeStream to always use async parsing for long-poll responses and snapshots
- Export yieldToMain and DEFAULT_YIELD_EVERY for users who need custom yielding in their own processing code

### DIFF
--- a/packages/typescript-client/src/parser.ts
+++ b/packages/typescript-client/src/parser.ts
@@ -167,13 +167,8 @@ export class MessageParser<T extends Row<unknown>> {
    *
    * @param messages - JSON string of messages to parse
    * @param schema - Schema for type parsing
-   * @param yieldEvery - Number of messages to process before yielding (default: 1000)
    */
-  async parseAsync<Result>(
-    messages: string,
-    schema: Schema,
-    yieldEvery: number = DEFAULT_YIELD_EVERY
-  ): Promise<Result> {
+  async parseAsync<Result>(messages: string, schema: Schema): Promise<Result> {
     // First, parse the JSON without transformation (this is fast)
     const parsed = JSON.parse(messages)
 
@@ -188,7 +183,7 @@ export class MessageParser<T extends Row<unknown>> {
       this.transformParsedMessage(parsed[i], schema)
 
       // Yield periodically to prevent blocking
-      if ((i + 1) % yieldEvery === 0) {
+      if ((i + 1) % DEFAULT_YIELD_EVERY === 0) {
         await yieldToMain()
       }
     }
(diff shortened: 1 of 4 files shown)


packages/typescript-client/src/parser.ts
```diff
@@ -157,6 +158,104 @@ export class MessageParser<T extends Row<unknown>> {
...
+      this.transformParsedMessage(parsed, schema)
+      return parsed as Result
+    }
+
+    // For arrays, transform each message with yielding
+    for (let i = 0; i < parsed.length; i++) {
+      this.transformParsedMessage(parsed[i], schema)
+
+      // Yield periodically to prevent blocking
+      if ((i + 1) % DEFAULT_YIELD_EVERY === 0) {
```
kevin-dp, line 186: nit: i'd prefer if `parseAsync` and `parseSnapshotDataAsync` would take the yield every configuration as an argument instead of relying on a global. We can still have a global configuration but i would pass that global as an argument when calling these functions.

KyleAMathews:
It's just a setTimeout which returns essentially immediately unless something else needs a bunch of CPU — so essentially free to do whenever. I'm not sure the interval really matters?

kevin-dp:
> It's just a setTimeout which returns essentially immediately unless something else needs a bunch of CPU — so essentially free to do whenever. I'm not sure the interval really matters?

I get what you mean about `setTimeout` returning quickly, but I still think the yield frequency is important.

Even if the yield itself is “cheap,” it isn’t free. Every time we yield we’re:

1. **Pausing the loop**
2. **Scheduling a microtask/macrotask**
3. **Letting the event loop swap us out and back in**

That context-switching overhead adds up if we do it *far* more often than necessary. Yielding every 1000 messages might mean we yield thousands of times per second on fast hardware or large batches, which is unnecessary churn.

On the other hand, if each message is expensive, yielding every 1000 iterations might mean we block the main thread for much longer than intended—possibly tens of milliseconds—which risks jank.

The browser’s main thread time budget is fundamentally **time-based**, not **iteration-based**, so a time-based yielding strategy matches the actual constraint:

- If messages are cheap, we’ll process more per slice.
- If messages are expensive, we’ll yield sooner.
- We stay within a predictable frame budget (~8–10ms), so we avoid blocking rendering.

The goal is not to yield “whenever,” but to yield “often enough to keep the app responsive, while not yielding so often that we waste time context switching.”

KyleAMathews:
Not entirely sure anymore this complex change is needed so closing.

Issue #3415, opened 2025-11-11
Avoid holding whole transaction in consumer memory


robacourt:
Part of #3413 and following on from #3414 . The consumer only buffers when the shape depends on subqueries. If the shape does not depend on subqueries, changes are written straight to storage. Storage will also need to be informed of the commit so that it can calculate chunk boundaries.

robacourt:
@icehaunter has pointed out this edge case that will need to be considered: Move-out/move-in results must never interleave a transaction. They use fake offsets for the data so if we write a move-out msg or move-in results to the storage, we'll use real offsets if we've not finished the transaction.

PR #3541, opened 2025-12-02
feat(sync-service): Send change batches to Consumer

robacourt:
Fix for #3414

This PR reduces the memory footprint of a consumer by only sending it changes that affect it. It does this by sending TransactionFragments to the Consumer, which also lays the groundwork to reduce the memory footprint even further (e.g. #3415 )

Performance is comparable to current (but there's a fair amount of variation in the benchmarks as shown by these two different runs!)
Screenshot 2025-12-04 at 13 06 11
Screenshot 2025-12-04 at 13 21 16

### DIFF
--- a/.changeset/hungry-terms-hide.md
+++ b/.changeset/hungry-terms-hide.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Reduce memory consumption by filtering transactions to only include the changes that affect the shape


packages/sync-service/lib/electric/postgres/replication_client/message_converter.ex
```diff
@@ -50,62 +55,85 @@ defmodule Electric.Postgres.ReplicationClient.MessageConverter do
...
+     %Relation{
+       id: id,
+       schema: ns,
+       table: name,
+       columns: Enum.map(cols, fn col -> %Column{name: col.name, type_oid: col.type_oid} end)
+     }, new_state}
   end
 
-  def convert(%LR.Insert{} = msg, %__MODULE__{} = state) do
+  def convert(%LR.Insert{} = msg, max_batch_size, %__MODULE__{} = state) do
```
msfstef: I don't understand why `max_batch_size` is not part of the `MessageConverter`'s state and is instead passed as an argument? We seem to define it when the replication client is started and never changed at runtime - which even if we did it would make sense to have an API on the message converter no?

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -311,40 +306,60 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+           last_log_offset: last_log_offset
+         }
+       )
+       when LogOffset.is_log_offset_lte(first_log_offset, last_processed_offset) and
+              LogOffset.is_log_offset_lt(last_processed_offset, last_log_offset) do
+    raise """
+    Received TransactionFragment that has already been partially processed.
+
+    This scenario is not currently supported. It could occur if the
+    batch size was changed while restarting the replication client.
```
msfstef, line 385: If the consumer (I suppose in the TransactionBuilder) simply discards parts of the transaction fragment it receives that it has already seen based on the changes' log offsets, wouldn't this not be a problem anymore?

A little while back I had modified the consumer to idempotently handle transactions so the collector can safely replay them and know that consumers can discard them - shouldn't consumers also idempotently handle transaction fragments regardless of how they are batched?

packages/sync-service/lib/electric/replication/transaction_builder.ex
```diff
@@ -62,33 +49,42 @@ defmodule Electric.Replication.TransactionBuilder do
...
-    {nil, %{state | transaction: txn}}
+    %{state | transaction: txn}
   end
 
-  defp build_from_change(%Relation{} = relation, state) do
-    {relation, state}
+  defp add_changes(%{transaction: txn} = state, %TransactionFragment{} = fragment) do
+    txn = %{
+      txn
+      | changes: Enum.reverse(fragment.changes) ++ txn.changes,
```
msfstef, line 58: adding from above - couldn't we keep track of the last collected change log offset and only append changes that are larger than that to make the transaction building resilient to transactions being fragmented in different ways?

msfstef (approved):
Very nice work! Would love to see benchmarks as well!

packages/sync-service/lib/electric/postgres/replication_client/message_converter.ex
robacourt: The idea was so we could change it at runtime, but you're right, we could equally have an API on the message converter. I'll update it

robacourt:
benchmark this

packages/sync-service/lib/electric/replication/shape_log_collector.ex
robacourt: Yeah, this PR was dragging on so this was one of the things I deferred.

packages/sync-service/lib/electric/replication/transaction_builder.ex
robacourt: It's a bit late to do it here, you need to guard against it at the beginning of ShapeLogCollector so that the EventRouter doesn't get messed up, and at the beginning of Consumer (potentially in the builder but the idea is we remove the builder wherever possible so it would be better to be decoupled)

PR #3543, opened 2025-12-03
Report the correct reatined WAL size confirmed_flush_lsn_lag metrics

alco:
Fix [#3544](https://github.com/electric-sql/electric/issues/3544).

The metric that is currently reported as `electric.postgres.replication.wal_size` is incorrectly named. Its name was changed from `electric.postgres.replication.lag` in https://github.com/electric-sql/electric/pull/2056.

Comparing `pg_current_wal_lsn()` to the replication slot's `confirmed_flush_lsn` shows how caught up the replcation slot consumer is to the latest database state. So it does measure the lag in terms of the total byte size of changes that had been sent to Electric over the replication connection and how much of those bytes Electric confirmed to have successfully flushed to disk. This values tends to 0 (or -1 but we can it at 0) as time progresses and no new changes appear in PG.

However, the other metric that is super critical to Electric's uptime is the difference between `pg_current_wal_lsn()` and the replication slot's `restart_lsn`. This shows how much WAL Postgres is forced to keep for the slot and this is the value that, when it exceeds the `max_slot_wal_keep_size` config, causes the replication slot to get invalidated. We have neglected collecting this metric until now.

### DIFF
--- a/packages/sync-service/lib/electric/stack_supervisor/telemetry.ex
+++ b/packages/sync-service/lib/electric/stack_supervisor/telemetry.ex
@@ -44,6 +44,9 @@ defmodule Electric.StackSupervisor.Telemetry do
       [
         Telemetry.Metrics.last_value("electric.shapes.total_shapes.count"),
         Telemetry.Metrics.last_value("electric.shapes.active_shapes.count"),
+        Telemetry.Metrics.last_value("electric.postgres.replication.pg_current_wal_lsn",
+          unit: :byte
+        ),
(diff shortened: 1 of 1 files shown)


alco:
Testing these metrics in our cloud staging:
image

msfstef (approved):
Perfect!

Would be nice to somehow have test coverage for these metrics

edit: just saw the staging test - very nice - still would be nice if we can just spin up a postgres and have a sanity check test

PR #3545, opened 2025-12-03
feat(sync-service): All configuring process GC settings for ShapeLogCollector

magnetised:
Add an env var for permanent configuration of per-process flags to configure gc settings and add to ShapeLogCollector

Also add a simple call to manually set values on the running SLC

### DIFF
--- a/packages/sync-service/lib/electric/config.ex
+++ b/packages/sync-service/lib/electric/config.ex
@@ -93,6 +93,11 @@ defmodule Electric.Config do
     conn_max_requests: 50,
     ## Performance tweaks
     publication_alter_debounce_ms: 0,
+    # allow for configuring per-process `Process.spawn_opt()`. In the form
+    # %{process_id :: atom() => [Process.spawn_opt()]}
+    # See `Process.flag/2`
+    #
+    # e.g. %{shape_log_collector: [min_heap_size: 1024 * 1024, min_bin_vheap_size: 1024 * 1024]}
     process_spawn_opts: %{},
     ## Misc
     process_registry_partitions: &Electric.Config.Defaults.process_registry_partitions/0,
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -93,6 +98,23 @@ defmodule Electric.Replication.ShapeLogCollector do
     GenServer.call(server(server_ref), :active_shapes)
   end
 
+  def set_process_flags(server_ref, flags) do
```
msfstef, line 112: would like some docs on these since the are arbitrary and potentially temporary

packages/sync-service/lib/electric/config.ex
```diff
@@ -93,6 +93,7 @@ defmodule Electric.Config do
     conn_max_requests: 50,
     ## Performance tweaks
     publication_alter_debounce_ms: 0,
+    process_spawn_opts: %{},
```
msfstef, line 101: worth a comment over this one as this is the main place we would look at to understand what this option is, what it does, and why it's here

packages/sync-service/lib/electric/config.ex
```diff
@@ -67,7 +67,7 @@ defmodule Electric.Config do
     send_cache_headers?: true,
     max_shapes: nil,
     # This value should be tuned for the hardware it's running on.
-    max_concurrent_requests: %{initial: 300, existing: 1000},
+    max_concurrent_requests: %{initial: 3000, existing: 10000},
```
msfstef: why was this changed in this PR? it seems worrying

msfstef (approved):
Lovely tweak addition, the per-process tweak will help to try things out! would like slightly more info/comments so that it makes some sense in the near future

packages/sync-service/lib/electric/config.ex
magnetised: oops! have to mod locally to get my stress-test to work.

PR #3547, opened 2025-12-03
perf(sync-service): Use ETS tables instead of maps in Filter to reduce GC pressure

robacourt:
Summary

  Migrates the Filter module and its namespace (WhereCondition, EqualityIndex, InclusionIndex) from
  Elixir maps to ETS tables to reduce garbage collection pressure with large numbers of shapes
  (200K+).

  Problem

  With 200K shapes being added and removed frequently, the map-based implementation causes large GC
  delays due to copying immutable map data structures on every add/remove operation.

  Solution

  Store filter data in ETS tables (outside the process heap) instead of nested Elixir maps:

  - 5 private ETS tables per filter: shapes_table, tables_table, where_cond_table, eq_index_table,
  incl_index_table
  - Same algorithmic complexity: O(1) for equality lookups, O(tree depth) for inclusion index
  - API unchanged: All existing tests pass without modification

  Performance Results (200K shapes)

  | Scenario               | ETS Implementation | Original (maps) | Target   |
  |------------------------|--------------------|-----------------|----------|
  | Equality index lookup  | 5.75µs             | 4.41µs          | <100µs ✅ |
  | Multiple matches (20)  | 13.82µs            | 9.07µs          | <100µs ✅ |
  | Post-churn performance | 9.49µs             | 4.69µs          | <100µs ✅ |

  The ETS implementation meets the <100µs target for affected_shapes with 200K equality-indexed
  shapes.

  Key Changes

  1. filter.ex: Changed struct to hold ETS table references; operations mutate ETS in-place
  2. where_condition.ex: WhereConditions now identified by refs and stored in ETS
  3. index.ex: Changed from protocol dispatch to direct function dispatch
  4. equality_index.ex: Stores entries in ETS with keys {where_cond_id, field, value}
  5. inclusion_index.ex: Stores tree nodes in ETS with keys {where_cond_id, field, path}

Screenshot 2025-12-04 at 17 06 47

### DIFF
--- a/packages/sync-service/lib/electric/shapes/filter/where_condition.ex
+++ b/packages/sync-service/lib/electric/shapes/filter/where_condition.ex
@@ -7,7 +7,7 @@ defmodule Electric.Shapes.Filter.WhereCondition do
   being on the same branch.
 
   Each WhereCondition is identified by a unique reference and stores:
-  - `index_keys`: list of {field, operation} tuples for indexed conditions
+  - `index_keys`: MapSet of {field, operation} tuples for indexed conditions
   - `other_shapes`: map of shape_id -> where_clause for non-optimized shapes
 
   The logic for specific indexes (equality, inclusion) is handled by dedicated modules that also use ETS.
@@ -26,7 +26,7 @@ defmodule Electric.Shapes.Filter.WhereCondition do
   require Logger
 
   def init(%Filter{where_cond_table: table}, condition_id) do
-    :ets.insert(table, {condition_id, {[], %{}}})
+    :ets.insert(table, {condition_id, {MapSet.new(), %{}}})
   end
 
   def add_shape(%Filter{where_cond_table: table} = filter, condition_id, shape_id, where_clause) do
@@ -49,14 +49,7 @@ defmodule Electric.Shapes.Filter.WhereCondition do
        ) do
     [{_, {index_keys, other_shapes}}] = :ets.lookup(table, condition_id)
     key = {optimisation.field, optimisation.operation}
-
-    index_keys =
-      if key in index_keys do
-        index_keys
-      else
-        [key | index_keys]
-      end
-
+    index_keys = MapSet.put(index_keys, key)
     :ets.insert(table, {condition_id, {index_keys, other_shapes}})
 
     Index.add_shape(filter, condition_id, shape_id, optimisation)
@@ -152,7 +145,7 @@ defmodule Electric.Shapes.Filter.WhereCondition do
       :deleted ->
         [{_, {index_keys, other_shapes}}] = :ets.lookup(table, condition_id)
         key = {optimisation.field, optimisation.operation}
-        index_keys = List.delete(index_keys, key)
+        index_keys = MapSet.delete(index_keys, key)
         update_or_delete_condition(table, condition_id, index_keys, other_shapes)
 
       :ok ->
@@ -160,8 +153,8 @@ defmodule Electric.Shapes.Filter.WhereCondition do
     end
   end
 
-  defp update_or_delete_condition(table, condition_id, [], other_shapes)
-       when other_shapes == %{} do
+  defp update_or_delete_condition(table, condition_id, index_keys, other_shapes)
+       when index_keys == %MapSet{} and other_shapes == %{} do
     :ets.delete(table, condition_id)
     :deleted
   end
@@ -193,7 +186,7 @@ defmodule Electric.Shapes.Filter.WhereCondition do
       [],
       fn ->
         [{_, {index_keys, _other_shapes}}] = :ets.lookup(table, condition_id)
-        OpenTelemetry.set_attribute(:index_count, length(index_keys))
+        OpenTelemetry.add_span_attributes(index_count: MapSet.size(index_keys))
 
         index_keys
         |> Enum.map(fn {field, operation} ->


robacourt:
benchmark this

packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
```diff
@@ -2,69 +2,181 @@ defmodule Electric.Shapes.Filter.Indexes.EqualityIndex do
...
-  alias Electric.Shapes.Filter.Index
-  alias Electric.Shapes.Filter.Indexes.EqualityIndex
+  alias Electric.Shapes.Filter
   alias Electric.Shapes.Filter.WhereCondition
 
-  defstruct [:type, :values]
+  @env Env.new()
+
+  @doc """
+  Check if the index for a field is empty.
```
magnetised: Smells like claude. Do we need the "empty? function checks if empty" type docs?

packages/sync-service/lib/electric/shapes/filter/indexes/inclusion_index.ex
```diff
@@ -12,226 +12,450 @@ defmodule Electric.Shapes.Filter.Indexes.InclusionIndex do
...
+        type,
+        array_value,
+        shape_id,
+        and_where
+      ) do
+    # Store type metadata
+    meta_key = {where_cond_id, field, :meta}
+    :ets.insert(table, {meta_key, {type}})
+
+    # Sort and deduplicate the array
```
magnetised: sigh

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
```diff
@@ -2,100 +2,148 @@ defmodule Electric.Shapes.Filter.WhereCondition do
...
+
+    :ets.insert(table, {where_cond_id, {updated_keys, other_shapes}})
+
+    # Add shape to the index
+    Index.add_shape(filter, where_cond_id, shape_id, optimisation)
   end
 
-  defp optimise_where(%Expr{eval: eval}), do: optimise_where(eval)
+  @doc false
+  def optimise_where(%Expr{eval: eval}), do: optimise_where(eval)
```
magnetised: why did you make this public?

msfstef:
This looks pretty good but a bit hard to review - in the meantime might be worth running a benchmark on it as well?

robacourt:
benchmark this

packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
```diff
@@ -2,69 +2,181 @@ defmodule Electric.Shapes.Filter.Indexes.EqualityIndex do
...
+              [{_, {_type, nested_where_cond_id}}] ->
+                WhereCondition.affected_shapes(filter, nested_where_cond_id, record, fn _shape ->
+                  %{}
+                end)
+            end
+
+          :error ->
+            raise RuntimeError,
+              message:
+                "Could not parse value for field #{inspect(field)} of type #{inspect(type)}"
```
robacourt: Should we be raising an error here?

packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
```diff
@@ -2,69 +2,181 @@ defmodule Electric.Shapes.Filter.Indexes.EqualityIndex do
...
+
+            case :ets.lookup(table, key) do
+              [] ->
+                MapSet.new()
+
+              [{_, {_type, nested_where_cond_id}}] ->
+                WhereCondition.affected_shapes(filter, nested_where_cond_id, record, fn _shape ->
+                  %{}
+                end)
+            end
```
robacourt: nasty nesting

packages/sync-service/lib/electric/shapes/filter/indexes/equality_index.ex
robacourt: removed

packages/sync-service/lib/electric/shapes/filter/indexes/inclusion_index.ex
robacourt: removed

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
robacourt: put back

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
```diff
@@ -2,67 +2,67 @@ defmodule Electric.Shapes.Filter.WhereCondition do
...
+         %Filter{where_cond_table: table} = filter,
+         condition_id,
+         shape_id,
+         optimisation
+       ) do
+    [{_, {index_keys, other_shapes}}] = :ets.lookup(table, condition_id)
+    key = {optimisation.field, optimisation.operation}
+
+    index_keys =
+      if key in index_keys do
```
msfstef: this looks like a linear lookup in the size of `index_keys`, whereas previously we were doing a map lookup on the key - is there a particular reason we're not using a MapSet or Map for this? Does the size of this array concern us at all?

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
```diff
@@ -153,30 +184,34 @@ defmodule Electric.Shapes.Filter.WhereCondition do
...
       # We can't tell which shapes are affected, the safest thing to do is return all shapes
-      all_shape_ids(condition)
+      all_shape_ids(filter, condition_id)
   end
 
-  defp indexed_shapes_affected(condition, record, shapes) do
+  defp indexed_shapes_affected(%Filter{where_cond_table: table} = filter, condition_id, record) do
     OpenTelemetry.with_child_span(
       "filter.filter_using_indexes",
-      [index_count: map_size(condition.indexes)],
```
msfstef, line 162: we could set the span property from within the span callback if we wanted to - perhaps we don't use it?

robacourt: Good spot! It is a useful stat, I've added it in the callback as you suggested

packages/sync-service/lib/electric/shapes/filter/where_condition.ex
robacourt: Good spot. Fixed.

KyleAMathews:
Just reading through the PR body again. It's probably worth calling out that the micro benchmarks are slower but when running, it's faster due to no/less gc pauses.

PR #3548, opened 2025-12-03
feat(sync-service): Add function to dynamically re-enable consumer suspend

magnetised:
ATM if you re-enable consumer suspend you don't gain much because existing, hibernated, consumers don't get the message (literally).

This ConsumerRegistry.enable_suspend/3 function is a single shot config plus consumer wake up call that we can use to sucessfully toggle suspension, mitigating the thundering herd effect.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer_registry.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer_registry.ex
@@ -163,14 +163,11 @@ defmodule Electric.Shapes.ConsumerRegistry do
   disabled, because the configuration message will have the side-effect of
   waking all consumers from hibernation.
 
-  The `max_timeout` value allows for spreading the suspension of existing
+  The `jitter_period` value allows for spreading the suspension of existing
   consumers over a large time period to avoid a sudden rush of consumer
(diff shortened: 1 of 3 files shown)


packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -147,6 +147,51 @@ defmodule Electric.Shapes.ConsumerRegistry do
...
   end
 
+  @doc """
+  Dynamically (re-)enable consumer suspension on all running consumers.
+
+  This allows for dynamically re-configuring consumer suspension even if it was
+  disabled, because the configuration message will have the side-effect of
+  waking all consumers from hibernation.
+
+  The `max_timeout` value allows for spreading the suspension of existing
```
robacourt: It seems a bit confusing calling this a timeout, sure it's how it's implemented, but from an API perspective it's a `max_stagger_delay` and you'd probably do something like `hibernate_after + Enum.random(0..max_stagger_delay)`

robacourt (approved):
Nice! Just left a little nit

PR #3549, opened 2025-12-03
feat: Batch SLC shape registrations and deregistrations in separate process

msfstef:
Follows from discussions on GC and SLC pressure.

The SLC is a very busy process, as it churns through all operations in the replication stream which can be extremely busy.

At the same time, there is a constant stream of shape registrations and deregistrations which are currently handled by the SLC as well, which get interleaved in the log processing. This might be causing lots of index building and rebuilding in betweeen processing transactions at an unnecessarily hight rate.

This PR introduces a Registrator process that basically batches shape registrations and deregistrations, and sends them batched to the SLC Processor for handling. While it is waiting for the SLC to reply that it has updated the index, it collects/batches shape registrations and deregistrations to send as soon as the previous batch is done.

It _could_ be doing a debounce, but I think having it serially update the index is ensuring that the SLC basically only ever sees 1 message for updating the index in its mailbox at a time. This basically prioritizes log consumption over new shape creation and shape removal.

This is also the preparatory work if we want to move in a sort of out-of-process index updating with deltas being prepared and sent to be applied much more quickly.

### DIFF
--- a/packages/sync-service/test/electric/replication/shape_log_collector/supervisor_test.exs
+++ b/packages/sync-service/test/electric/replication/shape_log_collector/supervisor_test.exs
@@ -1,0 +1,57 @@
+defmodule Electric.Replication.ShapeLogCollector.SupervisorTest do
+  use ExUnit.Case, async: true
+
+  import Support.ComponentSetup
+
+  alias Electric.Replication.ShapeLogCollector
+
+  @inspector Support.StubInspector.new(
+               tables: [{1234, {"public", "test_table"}}],
+               columns: [%{name: "id", type: "int8", pk_position: 0}]
+             )
+
+  setup [
+    :with_stack_id_from_test,
+    :with_in_memory_storage,
+    :with_shape_status,
+    :with_lsn_tracker,
+    :with_persistent_kv
+  ]
+
+  setup ctx do
+    {:ok, pid} =
+      start_supervised(
+        Supervisor.child_spec(
+          {ShapeLogCollector.Supervisor,
+           stack_id: ctx.stack_id,
+           inspector: {Mock.Inspector, elem(@inspector, 1)},
+           persistent_kv: ctx.persistent_kv},
+          restart: :temporary
+        )
+      )
+
+    ref = Process.monitor(pid)
+
+    %{monitor_ref: ref}
+  end
+
+  test "shuts down when processor is shut down normally", %{monitor_ref: ref} = ctx do
+    ctx.stack_id |> ShapeLogCollector.name() |> GenServer.stop()
+    assert_receive {:DOWN, ^ref, :process, _pid, _reason}, 1000
+  end
+
+  test "shuts down when processor is shut down abnormally", %{monitor_ref: ref} = ctx do
+    ctx.stack_id |> ShapeLogCollector.name() |> GenServer.stop(:whatever)
+    assert_receive {:DOWN, ^ref, :process, _pid, _reason}, 1000
+  end
+
+  test "shuts down when registrator is shut down normally", %{monitor_ref: ref} = ctx do
+    ctx.stack_id |> ShapeLogCollector.name() |> GenServer.stop()
+    assert_receive {:DOWN, ^ref, :process, _pid, _reason}, 1000
+  end
+
+  test "shuts down when registrator is shut down abnormally", %{monitor_ref: ref} = ctx do
+    ctx.stack_id |> ShapeLogCollector.name() |> GenServer.stop(:whatever)
+    assert_receive {:DOWN, ^ref, :process, _pid, _reason}, 1000
+  end
+end


msfstef:
benchmark this

packages/sync-service/lib/electric/core_supervisor.ex
```diff
@@ -63,7 +63,7 @@ defmodule Electric.CoreSupervisor do
        refresh_period: Keyword.get(tweaks, :publication_refresh_period, 60_000)}
 
     shape_log_collector_spec =
-      {Electric.Replication.ShapeLogCollector,
+      {Electric.Replication.ShapeLogCollector.Supervisor,
```
magnetised, line 66: This is going to change the stack restart behaviour if the SLC crashes. Currently if the SLC crashes the entire stack goes down. Now its under a supervisor, that supervisor will restart it and the stack will remain up.

Now I guess the question is: is that a bad thing? How much of the current behaviour depends on this one_for_all strategy? 

With this new strategy we definitely want to move the ownership of the ConsumerRegistry ets to something further up the stack, or some separate process. 

Might be worth maintaining the current restart strategy using `shutdown: :any_significant` on the SLC.Supervisor rather than address all that here, or not have a supervisor and just add the Registrator as a direct child of this supervisor.

Not pulling down the entire stack for a crash in the SLC is a good thing, and now we're using ets to restore state not process subscriptions, definitely more achievable, but this isn't maybe the place to do it.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -1,596 +1,60 @@
...
-  defp fill_keys_in_txn(txn, state) do
-    with {:ok, pk_cols_of_relations} <- pk_cols_of_relations(txn, state) do
-      txn =
-        Map.update!(txn, :changes, fn changes ->
-          Enum.map(changes, &Changes.fill_key(&1, pk_cols_of_relations[&1.relation]))
-        end)
+  This is typically called after the initial shape registrations
+  have been processed.
+  """
+  defdelegate mark_as_ready(stack_id), to: __MODULE__.Processor
```
magnetised: I don't think its necessary to move all the "processing" functions to a separate module. It's just cognitive overhead. We have the same structure in the PublicationManager and the indirection can be annoying.

packages/sync-service/lib/electric/shape_cache/shape_cleaner.ex
```diff
@@ -150,7 +150,7 @@ defmodule Electric.ShapeCache.ShapeCleaner do
              :ok <- Storage.cleanup!(stack_storage, shape_handle),
              OpenTelemetry.start_interval("remove_shape.shape_log_collector_remove"),
              :ok <-
-               Electric.Replication.ShapeLogCollector.remove_shape(stack_id, shape_handle) do
+               Electric.Replication.ShapeLogCollector.unsubscribe(stack_id, shape_handle) do
```
magnetised: subscribe/unsubscribe is the wrong terminology for what's happening so rather than rename `remove_shape` -> `unsubscribe` it would make more sense to rename `subscribe` to `add_shape`

packages/sync-service/lib/electric/config.ex
```diff
@@ -97,7 +97,7 @@ defmodule Electric.Config do
     # %{process_id :: atom() => [Process.spawn_opt()]}
     # See `Process.flag/2`
     #
-    # e.g. %{shape_log_collector: [min_heap_size: 1024 * 1024, min_bin_vheap_size: 1024 * 1024]}
+    # e.g. %{shape_log_collector_processor: [min_heap_size: 1024 * 1024, min_bin_vheap_size: 1024 * 1024]}
```
magnetised: we could just rename `ShapeLogCollector` to `ShapeLogProcessor` - uncomfortable with a 'collector processor'

packages/sync-service/lib/electric/stack_supervisor.ex
```diff
@@ -336,7 +333,7 @@ defmodule Electric.StackSupervisor do
         [
           stack_id: stack_id,
           handle_operations:
-            {Electric.Replication.ShapeLogCollector, :handle_operations, [shape_log_collector]}
+            {Electric.Replication.ShapeLogCollector, :handle_operations, [stack_id]}
```
magnetised: 🥳

packages/sync-service/test/electric/connection/manager_test.exs
```diff
@@ -170,7 +171,7 @@ defmodule Electric.Connection.ConnectionManagerTest do
       # by the replication supervisor
       monitor =
         stack_id
-        |> Electric.Replication.ShapeLogCollector.name()
+        |> Electric.Replication.ShapeLogCollector.Processor.name()
```
magnetised: even if we keep the Processor module, could we not just `defdelegate name/1`

packages/sync-service/lib/electric/replication/shape_log_collector/processor.ex
```diff
@@ -0,0 +1,617 @@
...
+    trace_context = OpenTelemetry.get_current_context()
+    GenServer.call(name(stack_id), {:handle_operations, operations, trace_context}, :infinity)
+  end
+
+  def handle_shape_registration_updates(stack_id, shapes_to_add, shapes_to_remove) do
+    pid = name(stack_id) |> GenServer.whereis()
+    call_ref = make_ref()
+
+    send(
+      pid,
```
magnetised: why fake the call behaviour?

packages/sync-service/lib/electric/replication/shape_log_collector/processor.ex
```diff
@@ -0,0 +1,617 @@
...
+             filter: filter,
+             dependency_layers: layers,
+             subscriptions: count
+         }}
+      end
+    )
+  end
+
+  def handle_call(
+        {:handle_shape_registration_updates, shapes_to_add, shapes_to_remove},
```
magnetised: why not do this work inside the Registrator and just send the updated structs to the processor?

as long as we deregister the consumer after the processor definitely has the updated filter then it should all work. The fewer things we can do in this process the better.

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
```diff
@@ -0,0 +1,179 @@
...
+        ) :: :ok
+  # shapes that are being restored are already in the filters
+  # because they were restored from the ets at startup
+  def subscribe(_stack_id, _shape_handle, _shape, :restore) do
+    :ok
+  end
+
+  # new shapes -- created after boot -- do need to be added
+  def subscribe(stack_id, shape_handle, shape, :create) do
+    GenServer.call(name(stack_id), {:subscribe, shape_handle, shape})
```
magnetised: see my other comment - the comments here suggest that this is a function that "adds a shape to the filters", it's not a "subscription" function.

Wouldn't it make sense to give this a timeout of `:infinity` since the actual reply depends on the SLC processor to get to the message.

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
```diff
@@ -0,0 +1,179 @@
...
+
+  @spec unsubscribe(Electric.stack_id(), Electric.shape_handle()) :: :ok
+  def unsubscribe(stack_id, shape_handle) do
+    # This has to be async otherwise the system will deadlock -
+    # - a consumer being cleanly shutdown may be waiting for a response from ShapeLogCollector
+    #   while ShapeLogCollector is waiting for an ack from a transaction event, or
+    # - a consumer that has crashed will be waiting in a terminate callback
+    #   for a reply from the unsubscribe while the ShapeLogCollector is again
+    #   waiting for a txn ack.
+    GenServer.cast(name(stack_id), {:unsubscribe, shape_handle})
```
magnetised: this could probably be a call now that we have the indirection.

packages/sync-service/lib/electric/replication/shape_log_collector/request_batcher.ex
```diff
@@ -0,0 +1,179 @@
...
+       | to_add: Map.put(state.to_add, shape_handle, shape),
+         to_remove: MapSet.delete(state.to_remove, shape_handle),
+         to_schedule_waiters: Map.put(state.to_schedule_waiters, shape_handle, from)
+     }, {:continue, :maybe_schedule_update}}
+  end
+
+  @impl true
+  def handle_cast({:unsubscribe, shape_handle}, state) do
+    if from = Map.get(state.to_schedule_waiters, shape_handle) do
+      GenServer.reply(
```
magnetised, line 123: nice

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
```diff
@@ -0,0 +1,179 @@
...
+        state.to_add,
+        state.to_remove
+      )
+
+    ack_waiters =
+      state.to_schedule_waiters
+      |> Enum.flat_map(fn
+        {_shape_handle, nil} -> []
+        {shape_handle, from} -> [{shape_handle, from}]
+      end)
```
magnetised:
```suggestion
      |> List.keydelete(nil, 1)
```

packages/sync-service/lib/electric/core_supervisor.ex
msfstef: ah yes I initially had `any_significant` but figured it could restart independently - but you're right this changes semantics in unpredictable ways, I'll use that instead and we can rethink that later.

packages/sync-service/lib/electric/config.ex
msfstef: I agree - but what would the `Processor` module be called now haha

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
msfstef: I've tried to keep the same behaviour as before - giving it a timeout of `:infinity` means that we can have an arbitrary number of consumers waiting to be subscribed, whereas with a timeout conusmers can fail if the shape log collector is so busy that they are failing to subscribe.

I'm happy to increase the timeout but I think it makes sense to have a sensible one.

packages/sync-service/lib/electric/replication/shape_log_collector/processor.ex
msfstef: we have this pattern elsewhere in the codebase and it avoids having to add logic to the processor to reply back to the registrator - I've tried to touch as little as possible in the processor in this PR for the sake of not introducing accidental side effects in the core logic

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
msfstef: ooh very nice thank you

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
msfstef: I don't think so, because ultimately:
```
#   a consumer being cleanly shutdown may be waiting for a response from ShapeLogCollector
#   while ShapeLogCollector is waiting for an ack from a transaction event, or
```

We could make this a call that returns without the shape having been removed however, which is perhaps what you mean. In that way we at least know that the "removal has been scheduled".

packages/sync-service/lib/electric/replication/shape_log_collector/processor.ex
msfstef: I've left a comment in the moduledoc:
> In the future, this could also create diffs to the shape filters
> instead of sending the full list of shapes to add/remove on each update.

I think this is work for a future PR and how we can improve this process and unload stuff from the SLC, but I think there's complexities in doing that.

packages/sync-service/test/electric/connection/manager_test.exs
msfstef: I've avoided that explicitly because "name" of a multi-process service doesn't quite make sense IMO.

Perhaps if use the `any_significant` shutdown then we can make it be the supervisor for the sake of monitoring, but still for mocking calls you need the specific process you want to patch.

packages/sync-service/lib/electric/config.ex
msfstef: Also I'd rather do that in a separate PR as it will touch many files and will have to change many comments ETC and would rather do that separately.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
msfstef: I like the indirection for the purpose of creating a public API of what is essentially the "ShapeLogCollector service" - anything that uses it should only use this API, and how we internally end up structuring the processes that run the service is a concern that is hidden from the clients.

I very much want to avoid other parts of the code explicitly calling the `Processor` or the `Registrator` as these should always be subject to change. I can't see a way of creating a strict contract without _some_ indirection but happy to hear alternatives.

The reason things are in another module is that if we want to respect conventions that 1) there is a supervisor that is named appropriately in its own file and module, and 2) the top level module should start the supervisor, we end up with a level of indirection. We had this discussion with PubMan which is why it is the way it is now, happy to discuss it again but we should decide on our conventions.

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
magnetised: Didn't the timeout lead to cascading failures though? Overworked SLC not accepting new consumers, consumers timing out on the `subscribe` call, thus crashing and calling `unsubscribe` as part of the cleanup leading to more messages in the SLC's mailbox... etc etc.

Perhaps a bit of state in the consumer that prevents unsubscribes without subscribes, that would alleviate that (I'll add it in a separate pr). Let's increase the default timeout though - 5s isn't very long considering the work the SLC's doing.

packages/sync-service/lib/electric/replication/shape_log_collector/processor.ex
magnetised: Ok, gotcha. Missed the comment. Seemed like the obvious move

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
msfstef: > Perhaps a bit of state in the consumer that prevents unsubscribes without subscribes, that would alleviate that (I'll add it in a separate pr).

In this PR the registrator cancels subscribe requests that haven't been sent yet if it receives unsubscribe requests.

This means that there is a layer that will dedupe these with its batching. It still always will send the unnecessary unsubscribe requests, but those are trivially handled by the SLC as the shape is not registered and the request discarded. The main thing is that via this batching the SLC doesn't have to handle an onslaught of registrations followed by deregistrations.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
magnetised: I guess my problem is that this "convention" leads to a lot of indirection. I did it myself in the Consumer.Monitor thing that was just a proxy to Consumer.Monitor.RefCounter, and the overhead of that indirection was painful.

Nothing about the structure you're proposing, even if we end up with a supervisor, requires this indirection and it just makes the system that bit more complex.

packages/sync-service/lib/electric/config.ex
magnetised: Yeah, my point being that the separate processor module is just unnecessary and the weird naming is just another way that manifests.

packages/sync-service/test/electric/connection/manager_test.exs
magnetised: Above you wanted abstraction because the processor & registrator modules could change but here we make that processor module explicit... Make your mind up! 🙃

packages/sync-service/lib/electric/replication/shape_log_collector/processor.ex
magnetised: Yeah. Didn't like it there either but left it so as not to change too many things. Feel like this genserver api is private and we shouldn't use it.

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
magnetised: It's more like we don't have a deadlock anymore if the call is done to the registrator so we can make it a proper sync call and when it returns we know the shape as been removed. Having said that a call does delay a shutdown which we probably don't want

packages/sync-service/test/electric/connection/manager_test.exs
msfstef: haha keep in mind this is uniquely a test and mocking concern - indeed I'm happy to keep a `name` in the interface and set it to whatever kills the whole process, I just think it is unnecessary (and not used outside of tests)

packages/sync-service/lib/electric/replication/shape_log_collector/processor.ex
msfstef: Happy to do a cast request -> cast reply instead as I had done for the publication manager

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
msfstef: I still think we have a deadlock - if the consumer is blocked waiting for a reply to the removal request (assuming the removal API actually returns when it is removed), and the SLC is waiting for a txn ack, the consumer will never reach the txn ack.

The indirection just allows us to change this to a call if the indirection layer replies before it actually knows that the shape has been removed (which is what I've done)

packages/sync-service/lib/electric/core_supervisor.ex
msfstef: Ended up doing `max_restarts: 0` as I think it is cleaner than setting every process as significant etc

Ultimately until we revisit if we can or cannot let the SLC restart independently, the supervisor acts as a coordinator for startup and shutdown.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
msfstef: I'm unclear on how we can do this without indirection - what would the clients be calling? could you please sketch out a module/file structure for it?

packages/sync-service/lib/electric/replication/shape_log_collector/registrator.ex
msfstef: I've bumped it to an arbitrary 45 seconds + the argument I've made above - resolving this but let me know if you disagree

msfstef:
benchmark this

packages/sync-service/test/electric/connection/manager_test.exs
msfstef: I've eliminated this reference in the test - now the only references to SLC internals are 1) in the internal modules themselves (the Processor, Registrator, Supervisor, and the Interface) and 2) the tests of the service and internal modules (for mocking and asserting purposes)

packages/sync-service/lib/electric/config.ex
msfstef: I'm proposing `ShapeLogManager` - ultimately the service might end up being composed of the "index builder" the "request handler" the "log processor" and various moving parts.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
robacourt: I have to say I'm not a fan either.

Supervisors can just be separate, that's just a standard elixir thing. And to be honest you don't need a separate supervisor, the Shapes.Supervisor is one_for_all and can do it.

The Registrator is fine to be called directly, maybe with a better name, say SubscriptionBatcher.

I do think we need some abstractions over subsections of our code, say for the whole connection subsystem, it'll add an annoying level of indirection but with benefits. I'm not sure there's enough benefits here.

packages/sync-service/lib/electric/config.ex
robacourt: i would just keep it `ShapeLogCollector` and not have the processor

packages/sync-service/lib/electric/replication/shape_log_collector.ex
msfstef: After chatting more with @magnetised and seeing your opinion @robacourt I've removed the layer of indirection.

We now have our `ShapeLogCollector` be the same process that it was, but I've kept some indirection. It has 2 `defdelegate` for the process now called `ShapeLogCollector.RequestBatcher`, and I've kept the `ShapeLogCollector.Supervisor` cause while I hear that we could just add it to the top level supervisor, I felt that starting yet another process that it so intimately tied to the SLC (basically a helper/utility process for it) not in tandem would make it harder to follow.

packages/sync-service/lib/electric/config.ex
msfstef: Removed the processor, we now have our `ShapeLogCollector`

packages/sync-service/lib/electric/replication/shape_log_collector/request_batcher.ex
```diff
@@ -0,0 +1,196 @@
...
+    GenServer.call(name(stack_id), {:add_shape, shape_handle, shape}, 45_000)
+  end
+
+  @doc """
+  Schedules a shape removal from the SLC, returns before the shape is
+  actually removed.
+  """
+  @spec remove_shape(Electric.stack_id(), Electric.shape_handle()) :: :ok
+  def remove_shape(stack_id, shape_handle) do
+    GenServer.call(name(stack_id), {:remove_shape, shape_handle})
```
magnetised, line 73: the sweet smell of compromise

packages/sync-service/lib/electric/replication/shape_log_collector/request_batcher.ex
```diff
@@ -0,0 +1,196 @@
+defmodule Electric.Replication.ShapeLogCollector.RequestBatcher do
```
magnetised, line 1: You need to rename the file too... I miss "registrator" already..

packages/sync-service/lib/electric/replication/shape_log_collector/supervisor.ex
```diff
@@ -0,0 +1,37 @@
...
+    Electric.Telemetry.Sentry.set_tags_context(stack_id: stack_id)
+
+    children = [
+      {ShapeLogCollector, opts},
+      {ShapeLogCollector.RequestBatcher, stack_id: stack_id}
+    ]
+
+    # Prevent any restarts until the whole system is capable of sustaining
+    # the SLC dying without any other shape machinery being restarted
+    Supervisor.init(children, strategy: :one_for_all, max_restarts: 0)
```
magnetised, line 35: sneaky. have we tested this?

magnetised (approved):
Fab. Nice change. You just need to rename `registrator.ex`

packages/sync-service/lib/electric/replication/shape_log_collector/request_batcher.ex
msfstef: I know it's not proper english, but if your system components don't sound like they might kill you when skynet takes over they're not as fun are they :(

packages/sync-service/lib/electric/replication/shape_log_collector/supervisor.ex
msfstef: I had added tests in the SLC test suite and removed them accidentally - added them back as supervisor tests in particular but yeah made sure it worked

https://github.com/electric-sql/electric/pull/3549/commits/7fd8bf2e868777dda0e8c429028e60e8a434700b

PR #3552, opened 2025-12-04
fix: Benchmarking image with right telemetry context

msfstef:
I think the main thing that it was missing the `--build-context electric-telemetry=../electric-telemetry` but I took the opportunity to move to the GH action instead which hopefully takes care of any extra cache optimisations


msfstef:
benchmark this

Issue #3488, opened 2025-11-23
"Could not select an operator overload" error with enum column and subset__where


fcannizzaro:
**Versions**

- Client: @tanstack/electric-db-collection: `0.2.5`
- Electric: `1.2.6` (docker)

Using the subset from this live query:

```ts
  const { data } = useLiveQuery(
    (q) => {
      return q
        .from({ lock: Collections.lock })
        .where(({ lock }) => eq(lock.challenge, challenge))
        .select(({ lock }) => ({
          school: lock.school,
          category: lock.category,
        }));
    },
    [challenge],
  );
```

the table is 

```ts
export const locks = pgTable(
  "lock",
  {
    challenge: challengeEnum().notNull(),
    category: categoryEnum().notNull(),
    school: integer()
      .notNull()
      .references(() => schools.id, onDeleteCascade),
    createdAt: timestamp({ withTimezone: true }).defaultNow().notNull(),
  },
  (t) => [
    primaryKey({
      name: "pl_locks_school_challenge_category",
      columns: [t.challenge, t.category, t.school],
    }),
  ],
);
```

url: `log=full&offset=-1&subset__params%5B1%5D=qualification&subset__where=%22challenge%22+%3D+%241`

| Key              | Value             |
|------------------|-------------------|
| log              | full              |
| offset           | -1                |
| subset__params[1]| qualification     |
| subset__where    | "challenge" = $1  |

**Bug description**

Electric replies with:

```json
{
    "message": "Invalid request",
    "errors": {
        "subset": {
            "where": [
                "At location 12: Could not select an operator overload"
            ]
        }
    }
}
```

This does not happen when using eager mode.

fcannizzaro:
reference repo: https://github.com/fcannizzaro/electric-overload-operator-error

KyleAMathews:
We need to add support for the function here I believe https://github.com/electric-sql/electric/blob/main/packages/sync-service/lib/electric/replication/eval/env/known_functions.ex

samwillis:
I think the underlying issue here is that the column you are filtering on is an enum, we need to fix that.

KyleAMathews:
When using enum columns in subset WHERE clauses, the previous fix in #3500 only treated enums as text during validation but the actual query sent to PostgreSQL still compared enum_column = 'value'::text which fails because PostgreSQL can't compare enum = text directly.

fcannizzaro:
the latest canary fixed the issue

PR #3554, opened 2025-12-04
fix(sync-service): fix enum params in subsets

robacourt:
Fix for #3488

### DIFF
--- a/packages/sync-service/test/electric/replication/eval/parser_test.exs
+++ b/packages/sync-service/test/electric/replication/eval/parser_test.exs
@@ -684,6 +684,22 @@ defmodule Electric.Replication.Eval.ParserTest do
       assert {:error, "At location 17: Could not select an operator overload"} =
                Parser.parse_and_validate_expression(~S/'{1,2,3}'::int[] || '{2,1,2}'::text[]/)
     end
+
+    test "rejects comparison of mismatched enum types" do
+      refs = %{
+        ["col1"] => {:enum, "test_enum1"},
+        ["col2"] => {:enum, "test_enum2"}
+      }
+
+      env = %{Env.new() | allow_enums: true}
+
+      assert {:error, "At location 5: Could not select an operator overload"} =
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/replication/eval/env/known_functions.ex
```diff
@@ -147,6 +147,10 @@ defmodule Electric.Replication.Eval.Env.KnownFunctions do
       do: datetime |> DateTime.shift_zone!(tz) |> DateTime.to_naive()
   end
 
+  ## Enum operators
+  defpostgres("anyenum = anyenum -> bool", delegate: &Kernel.==/2)
+  defpostgres("anyenum <> anyenum -> bool", delegate: &Kernel.!=/2)
```
icehaunter, line 152: I think this leaves us open to a case we don't handle correctly, where the query passes our parsing but fails later (e.g. on wrong enum comparison)

```
electric=# CREATE TYPE test_enum1 AS ENUM ('a', 'b', 'c'); CREATE TYPE test_enum2 AS ENUM ('x', 'y', 'z'); CREATE TABLE test_enums (id SERIAL PRIMARY KEY, col1 test_enum1, col2 test_enum2);
CREATE TYPE
CREATE TYPE
CREATE TABLE
electric=# SELECT * FROM test_enums WHERE col1 = col2;
ERROR:  operator does not exist: test_enum1 = test_enum2
LINE 1: SELECT * FROM test_enums WHERE col1 = col2;
                                            ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

Or is this covered by the env flag? I don't think it is because example above doesn't parse enum consts

robacourt: I've checked and it seems fine, I've added this test: https://github.com/electric-sql/electric/pull/3554/commits/d9b0f2e1ec15508d3ef4604b2ca562f447db9755 Good to check!

Issue #3551, opened 2025-12-04
Out of bounds for this shape errors


noudadrichem:
**Versions**

- Client: Plain HTTP via ShapeStreams using PGLite Sync.
- Electric: 1.2.7

**Bug description**
Rolling over Electric (_e.g. to update version_) tasks in Fargate sometimes results in the following error on clients:
```json
{
    "message": "Invalid request",
    "errors": {
        "offset": [
            "out of bounds for this shape"
        ]
    }
}
```

### Context
We're running Electric in Fargate with EFS attached to it. With a configuration of always 1 running tasks, a re-deployment results in 2 tasks running side by side for some seconds as it's rolling over without downtime.

### Scenario is
1 task A with EFS is running
restart deployment which starts another task B beside A
2 tasks run at the same time for ±20/30 seconds. _(Here I'm expecting Electric to be smart enough to handle the lock swapping without corruping the shape logs on EFS.)_
1 old task A shuts down
The new task B is up and running successfully. But with 1 client seeing out of bounds errors while another client is fetching without issues.

**Expected behavior**
I expect 2 Electric tasks to be able to safely run side by side within a fargate cluster and EFS mounted. So I can safely rollover tasks or even better, run 2 electric instances side by side to prevent single point of failure.

P.S. Find me in the Discord if more context is needed.

balegas:
Hei, this should work out of the box.
We've recently fixed an issue with shape restoration that would raise "out of bound" errors. Patch was included in 1.2.7. Did you face this issue doing a deployment from electric 1.2.7 -> electric 1.2.7 or was this an upgrade from a previous version?

noudadrichem:
@balegas Mm, that's odd. This issue happened when running 1.2.7 and restarting to a new 1.2.7. 

So _not_ an upgrade from a previous version but rollover to the same version. Happy to hop on a call to show/clarify if needed.

msfstef:
We've identified the issue that occurs during these handovers - since we buffer things in memory and serve the data before it has been flushed to disk fully, there are edge cases that might lead to the new instance not having the data that a client has already seen, which leads to this error.

The data will be seen by the new instance almost immediately, since it will consume the replication slot (as the previous instance will not have advanced it by virtue of not having flushed the data), so this is more of an edge case and an eagerness on the server's part to enforce the limits of offsets it knows about.

We intend to give it a bit more breathing space before asserting that the request is out of bounds which should solve this issue. I should have a PR open for this this week.

[...]

PR #3563, opened 2025-12-08
fix: Handle out of bounds requests with a timeout

msfstef:
Fixes https://github.com/electric-sql/electric/issues/3551 on top of https://github.com/electric-sql/electric/pull/3562

I had pinpointed that there are cases in which a client might see a response with offset B from one Electric, but that one has not been flushed yet, and instead has only flushed the previous offset A.

At that point if there is a crash or handover, it is likely that the new Electric recovers the earlier offset A, and would eventually pick up the transaction with offset B from the replication slot. However when the client tries to transparently continue its shape with the new Electric with offset B, they occasionally run into an out of bounds error until the new Electric catches up on the replication slot.

My initial intended solution was to simply treat offsets higher than the latest seen from electric as offsets we hold in a live request, but after some discussions with @balegas we've concluded to handle them as errors after a timeout instead, allowing some time for Electric to catch up and potentially rescue these requests with offsets higher than what we've seen, but otherwise returning an error.

The idea would be that the error is also retryable, in which case we could skip the timeout altogether, but that would require changing the HTTP semantics of the API, and rather than return 400 non-retryable errors, return some sort of 5xx error (and basically always assume the client is right and we just don't have the offset they requested yet).

## Approach

I've first enhanced our rolling deploy integration test such that it consistently reproduced this issue, by shifting a client from one electric to the other mid-transaction ingestion.

Then I updated our API to no longer error when offsets are higher than the last seen, and instead continue as they would normally, but mark them as **_not up to date_**, and when it's time to serve the log, _always_ try to handle them as live requests, but also with a message sent to the request process itself after a timeout (half the long poll timeout) to error if the offset has not been caught up to yet.

Another significant change I've made is to the `get_log_stream` storage API, it now accepts a reverse range (the `up_to` being lower than the `since`) and returns an empty enum, rather than error. If we think this is not something we're ok with, I will have to rewire a few things - honestly the only place this is a pain to fix is for SSE events.

One final thing to consider is also that since we know request processes are long lived, we might need to be manually flushing/cleaning up the message queues at the start of each request.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -607,7 +607,7 @@ defmodule Electric.Shapes.Api do
     Process.send_after(
       self(),
       {ref, :out_of_bounds_timeout},
-      floor(request.api.long_poll_timeout * 0.5)
+      div(request.api.long_poll_timeout, 2)
     )
 
     handle_live_request(request)


packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -678,11 +697,17 @@ defmodule Electric.Shapes.Api do
   end
 
   defp handle_live_request(%Request{params: %{live_sse: true}} = request) do
-    stream_sse_events(request)
+    request
+    |> update_attrs(%{ot_is_immediate_response: false})
```
magnetised, line 701: why the refactor to add duplication?

msfstef: honestly I just shifted this duplication from https://github.com/electric-sql/electric/pull/3563/files#diff-7193a01db51187dc0d1ea9c61dfd7a1a9d305cdd75d49c0603d646714e8e7f5fR613 to `handle_live_request` cause I think it's more appropriate and easier to handle.

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -591,6 +597,22 @@ defmodule Electric.Shapes.Api do
...
+       when is_out_of_bounds(request) do
+    # treat out of bounds requests like live requests with a
+    # shorter timeout before failing them, as if the client happened
+    # to be slightly ahead because of a restart or handover the
+    # offset they have seen should show up shortly, otherwise we
+    # assume it is an actually invalid out of bounds request
+    Process.send_after(
+      self(),
+      {ref, :out_of_bounds_timeout},
+      floor(request.api.long_poll_timeout * 0.5)
```
alco:
```suggestion
      div(request.api.long_poll_timeout, 2)
```
more idiomatic this way, IMO.

PR #3564, opened 2025-12-08
Fix columnMapper to support loading subsets

KyleAMathews:
When using columnMapper with ShapeStream, the columns parameter was not being encoded from app column names to database column names. This fix ensures that when a user specifies columns using app format (e.g., camelCase like `userId`), they are properly converted to the database format (e.g., snake_case like `user_id`) before being sent to the server.

This makes columnMapper work correctly with column subsets.

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -863,16 +863,16 @@ export class ShapeStream<T extends Row<unknown> = Row>
         // Get original columns array from options (before toInternalParams converted to string)
         const originalColumns = await resolveValue(this.options.params?.columns)
         if (Array.isArray(originalColumns)) {
-          // Quote each column name to handle special characters (commas, etc.)
           // Apply columnMapper encoding if present
-          const serializedColumns = originalColumns
-            .map((col) => {
-              const colStr = String(col)
-              const encoded = this.options.columnMapper
-                ? this.options.columnMapper.encode(colStr)
-                : colStr
-              return quoteIdentifier(encoded)
-            })
+          let encodedColumns = originalColumns.map(String)
+          if (this.options.columnMapper) {
+            encodedColumns = encodedColumns.map(
+              this.options.columnMapper.encode
+            )
+          }
+          // Quote each column name to handle special characters (commas, etc.)
+          const serializedColumns = encodedColumns
+            .map(quoteIdentifier)
             .join(`,`)
           setQueryParam(fetchUrl, COLUMNS_QUERY_PARAM, serializedColumns)
         } else {


packages/typescript-client/src/column-mapper.ts
```diff
@@ -285,6 +285,23 @@ export function encodeWhereClause(
...
+ * @returns The encoded columns string
+ *
+ * @internal
+ */
+export function encodeColumns(
+  columns: string | undefined,
+  encode?: (columnName: string) => string
+): string {
+  if (!columns || !encode) return columns ?? ``
+  return columns.split(`,`).map(encode).join(`,`)
```
kevin-dp: What if a column name contains a comma? Postgres allows quoted names to contain special characters like commas.

KyleAMathews:
@kevin-dp I realized we were doing this wrong — we have access to the original column data so don't need to bother with parsing issues.

packages/typescript-client/src/client.ts
```diff
@@ -855,8 +855,21 @@ export class ShapeStream<T extends Row<unknown> = Row>
         )
         setQueryParam(fetchUrl, WHERE_QUERY_PARAM, encodedWhere)
       }
-      if (params.columns)
-        setQueryParam(fetchUrl, COLUMNS_QUERY_PARAM, params.columns)
+      if (params.columns) {
+        // Get original columns array from options (before toInternalParams converted to string)
+        const originalColumns = await resolveValue(this.options.params?.columns)
+        if (Array.isArray(originalColumns)) {
+          const encodedColumns = this.options.columnMapper
```
kevin-dp: nit: the ternary is a bit harder to read than:
```ts
let encodedColumns = originalColumns
if (this.options.columnMapper) {
  encodedColumns = originalColumns.map((col) =>
    this.options.columnMapper!.encode(String(col))
  )
}
```

packages/typescript-client/src/client.ts
```diff
@@ -855,8 +855,21 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+            ? originalColumns.map((col) =>
+                this.options.columnMapper!.encode(String(col))
+              )
+            : originalColumns
+          setQueryParam(fetchUrl, COLUMNS_QUERY_PARAM, encodedColumns.join(`,`))
```
kevin-dp: Similar concern to before, what if the columns contain a comma. Imagine a column named `"foo,bar"` and a column `"baz"`. After this `encodedColumns.join(',')` it will send `"foo,bar,baz"` to the server which will incorrectly interpret it as 3 columns.

KyleAMathews: as it turns out we've always had this looming edge case — https://github.com/electric-sql/electric/blob/ceef72cd68f51689109d960f439b6565863b264c/packages/typescript-client/src/client.ts#L189 — but yes, we should fix it — lemme consolidate the logic & then have it fix it

KyleAMathews:
@kevin-dp ok I think we're actually handling things now

PR #3567, opened 2025-12-09
Limit matrix job parallelism on CI

alco:


msfstef (approved):
Looks good, but what's the motivation?

alco:
@msfstef this is a draft PR, was just testing the behaviour.

The context behind this is the intent to migrate to 3rd-party runners for actions which, as a rule, offer better perf and caching but have limited job parallelism out of the box.

The other motivating factor is getting results from one specimen of each job type faster. It is desirable to run one of each job type first, before proceeding to filling the run queue with N types of "deploy TS example" or "Autoformat X", for example.

alco:
Blacksmith does not have concurrency limits, so this is not relevant for us now.

PR #3570, opened 2025-12-09
fix: Ensure `ShapeStatus` backup is saved and restored at appropriate timings

msfstef:
Also fixes https://github.com/electric-sql/electric/issues/3551 

The underlying issue is fixed by https://github.com/electric-sql/electric/pull/3563 but this PR basically ensures that when doing controlled restarts and rolling deploys with shared storage, the backups are stored at the appropriate time, and loaded at the appropriate time (after the lock is acquired).

After moving `ShapeStatusOwner` outside of the replication supervisor (either @alco or @robacourt's work I don't remember exactly), the backup was created after the instance was completely shut down, and loaded way before we even connect to the postgres DB. This means that we _never_ actually load a backup, since the new instance will already have loaded files before acquiring the lock, which also leads to inconsistencies and issues like the one in the issue above.

I've augmented the rolling deploy integration test to assert a shape status backup is loaded.

I also took the opportunity to remove the explicit storage dependency on the `ShapeStatus` APIs and instead leverage `Electric.StackConfig.lookup!`, or `Storage.for_stack`.

### DIFF
--- a/packages/sync-service/lib/electric/connection/manager.ex
+++ b/packages/sync-service/lib/electric/connection/manager.ex
@@ -887,8 +887,8 @@ defmodule Electric.Connection.Manager do
     # acquiring the lock next will recreate the replication slot
     if state.drop_slot_requested, do: drop_publication(state)
 
-    # perform a backup of our shape metadata after the snapshots and publication
-    # have been removed to guarantee a consistent checkpoint
+    # perform a backup of our shape metadata after the snapshotter processes and
+    # publication have been removed to guarantee a consistent checkpoint
     Electric.ShapeCache.ShapeStatus.save_checkpoint(state.stack_id)
 
     if is_pid(replication_client_pid) do


packages/sync-service/lib/electric/stack_supervisor.ex
```diff
@@ -376,7 +376,7 @@ defmodule Electric.StackSupervisor do
         Electric.ShapeCache.Storage.stack_child_spec(storage),
         {Electric.Postgres.Inspector.EtsInspector,
          stack_id: stack_id, pool: metadata_db_pool, persistent_kv: config.persistent_kv},
-        {Electric.ShapeCache.ShapeStatusOwner, [stack_id: stack_id, storage: storage]},
+        {Electric.ShapeCache.ShapeStatusOwner, [stack_id: stack_id]},
```
robacourt: Does it still make sense to be supervised at this level? Should it not be in Shapes.Supervisor? If this process went down would we not have to restart a whole load of other stuff as well?

robacourt (approved):
Nice! Just left one comment

packages/sync-service/lib/electric/stack_supervisor.ex
msfstef: This process really only holds on to ETS tables (and now initializes them) - the reason we moved it up here is because we want it to persist any restarts of the shapes subsystem in its entirety.

That being said, I think you're right and I'll move it in the monitored core supervisor which has a rest for one strategy - if the shape status dies, we take down the rest of the conn and shape subsystem

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -883,6 +883,10 @@ defmodule Electric.Connection.Manager do
     # acquiring the lock next will recreate the replication slot
     if state.drop_slot_requested, do: drop_publication(state)
 
+    # perform a backup of our shape metadata after the snapshots and publication
```
alco: I got confused by the phrasing `after the snapshots have been removed` here. Then noticed the `# snapshots can immediately be killed as they don't hold any locks` above.

It would be clearer to use the term `snapshotter` or `snapshotter process` in those cases, IMO.

packages/sync-service/lib/electric/stack_supervisor.ex
alco: I moved this process up to be a child of the StackSupervisor before Rob's refactoring of the connection and shape subsystems. The goal was to be able to handle incoming shape requests in the connection scaledown mode.

When an incoming requests causes the connection subsystem to wake up again, the connection manager supervisor gets restarted. In the past, this would also cause a restart of the shapes supervisor but it's no longer the case. So, now that the connection and shape subsystems are more independent, I think the ShapeStatusOwner process can be moved back under the shapes supervisor where it originally lived.

msfstef: For now I've moved it into monitored core supervisor because it has rest for one semantics and I'd like to avoid the one for all strategy of the replication supervisor. This PR is already doing quite a bit as it is and I'd like to avoid another layer of complexity.

packages/sync-service/lib/electric/connection/manager.ex
msfstef: I've made the comment more clear following your suggestion

PR #3572, opened 2025-12-09
feat: Rely on storage for metadata persistence and caching instead of `ShapeStatus`

msfstef:
Closes https://github.com/electric-sql/electric/issues/3451

With @magnetised 's move of the handle -> shape and shape -> handle lookups into the `ShapeDb` abstraction, we actually no longer need to persist anything else in ShapeStatus.

For `snapshot_started?` we can directly read from the storage, which has its own read-through cache, and since this is only called in `await_snapshot_start?` we can just directly call it there and get rid of all the unused APIs.

For `latest_offset`, I've taken that out of ShapeStatus and instead we read it from storage with the same read-through cache. The `ShapeCache` APIs are responsible for collating shape handles found in the indices with an offset from the storage to maintain the same outward facing API behaviour.

With this change, `ShapeStatus` no longer either recovers any of the metadata (getting rid of 66% of reads on a cold restore),  nor persists any thing other than the lookups. Everything else is populated upon recovery.

I've gotten rid of all unused APIs and calls in the consumer and shape status, and I took the chance to do some API renaming for consistent behaviour (e.g. `fetch_handle_by_shape` and `fetch_shape_by_handle`, and both do `{:ok, res}` or `:error`).

I've also split `Storage.get_current_position` into `Storage.get_latest_offset` and `Storage.get_pg_snapshot` to avoid doing two reads when only one is needed, since in most places we need either one or the other. Only the consumer on initialization needs both, and it can simply call both APIs with the same performance.

I've also changed ETS tables for most things to `set` from `ordered_set`, basically all except the relation lookups since most don't do any ordered/range lookups.

If I have time I will also benchmark cold shape recovery for 50k shapes with this change.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
@@ -42,7 +42,6 @@ defmodule Electric.ShapeCache.PureFileStorage do
 
   import LogOffset
   import Electric.ShapeCache.PureFileStorage.SharedRecords
-  import Record
 
   import File, only: [write!: 3]
 
@@ -58,7 +57,6 @@ defmodule Electric.ShapeCache.PureFileStorage do
     :shape_handle,
     :stack_id,
     :stack_ets,
-    :stack_cache_ets,
     snapshot_file_timeout: :timer.seconds(5),
     version: @version
   ]
@@ -81,37 +79,20 @@ defmodule Electric.ShapeCache.PureFileStorage do
   # For `snapshot_started?` we only have the metadata file if it has been set.
   @boolean_stored_keys [:snapshot_started?]
 
-  @cached_keys [
-    :snapshot_started?,
-    :pg_snapshot,
-    :compaction_started?,
-    :last_snapshot_chunk,
-    :last_seen_txn_offset,
-    :last_persisted_txn_offset,
+  # keys to populate an empty cache with and to preserve on writer termination
+  # for access via the read path
+  @read_path_keys [
+    :cached_chunk_boundaries,
+    :compaction_boundary,
     :last_persisted_offset,
-    :compaction_boundary,
-    :latest_name,
-    :cached_chunk_boundaries
-  ]
-
-  # Keys that are cached but read through to avoid disk reads
-  # when writer is not active
-  @stack_cached_keys [
-    :latest_name,
-    :compaction_boundary,
     :last_persisted_txn_offset,
-    :snapshot_started?,
-    :last_snapshot_chunk
+    :last_seen_txn_offset,
+    :last_snapshot_chunk,
+    :latest_name,
+    :snapshot_started?
   ]
 
-  # Record that caches read-through metadata to avoid disk reads
-  defrecord :storage_meta_stack,
-    shape_handle: nil,
-    latest_name: :not_cached,
-    compaction_boundary: :not_cached,
-    last_persisted_txn_offset: :not_cached,
-    snapshot_started?: :not_cached,
-    last_snapshot_chunk: :not_cached
+  @clean_cache_keys storage_meta_keys() -- @read_path_keys
 
   # Directory for storing metadata
   @metadata_storage_dir ".meta"
@@ -149,8 +130,7 @@ defmodule Electric.ShapeCache.PureFileStorage do
       chunk_bytes_threshold: stack_opts.chunk_bytes_threshold,
       shape_handle: shape_handle,
       stack_id: stack_id,
-      stack_ets: stack_ets,
-      stack_cache_ets: stack_cache_ets(stack_id)
+      stack_ets: :ets.whereis(stack_ets)
     }
   end
 
@@ -160,12 +140,11 @@ defmodule Electric.ShapeCache.PureFileStorage do
 
   @doc false
   def stack_ets(stack_id), do: :"Electric.ShapeCache.PureFileStorage:#{stack_id}"
-  def stack_cache_ets(stack_id), do: :"Electric.ShapeCache.PureFileStorage:cache:#{stack_id}"
 
   defp stack_task_supervisor(stack_id),
     do: ProcessRegistry.name(stack_id, __MODULE__.TaskSupervisor)
 
-  defp shape_data_dir(%__MODULE__{} = shape_opts) do
+  def shape_data_dir(%__MODULE__{} = shape_opts) do
     shape_data_dir(shape_opts, [])
   end
 
@@ -203,15 +182,6 @@ defmodule Electric.ShapeCache.PureFileStorage do
              :set,
              :public,
              keypos: storage_meta(:shape_handle) + 1,
-             read_concurrency: true,
-             write_concurrency: :auto
-           ])
-
-           :ets.new(stack_cache_ets(opts.stack_id), [
-             :named_table,
-             :set,
-             :public,
-             keypos: storage_meta_stack(:shape_handle) + 1,
              read_concurrency: true,
              write_concurrency: :auto
            ])
@@ -268,28 +238,9 @@ defmodule Electric.ShapeCache.PureFileStorage do
     Path.join([base_path, @metadata_storage_dir, "backups"])
   end
 
-  def delete_shape_ets_entry(stack_id, shape_handle) do
-    try do
-      :ets.delete(stack_ets(stack_id), shape_handle)
-      :ok
-    rescue
-      ArgumentError -> :ok
-    end
-  end
-
-  defp delete_shape_stack_cache_entry(stack_id, shape_handle) do
-    try do
-      :ets.delete(stack_cache_ets(stack_id), shape_handle)
-      :ok
-    rescue
-      ArgumentError -> :ok
-    end
-  end
-
   def drop_all_ets_entries(stack_id) do
     try do
       :ets.delete_all_objects(stack_ets(stack_id))
-      :ets.delete_all_objects(stack_cache_ets(stack_id))
       :ok
     rescue
       ArgumentError -> :ok
@@ -310,8 +261,8 @@ defmodule Electric.ShapeCache.PureFileStorage do
              stack_opts.stack_id,
              shape_data_dir(stack_opts.base_path, shape_handle)
            ) do
-      delete_shape_ets_entry(stack_opts.stack_id, shape_handle)
-      delete_shape_stack_cache_entry(stack_opts.stack_id, shape_handle)
+      :ets.delete(stack_ets(stack_opts.stack_id), shape_handle)
+      :ok
     end
   end
 
@@ -528,7 +479,8 @@ defmodule Electric.ShapeCache.PureFileStorage do
     write_metadata!(opts, :compaction_boundary, boundary)
   end
 
-  def latest_name(%__MODULE__{} = opts), do: read_metadata!(opts, :latest_name) || "latest.0"
+  def latest_name(%__MODULE__{} = opts),
+    do: read_cached_metadata(opts, :latest_name) || "latest.0"
 
   def set_latest_name(%__MODULE__{stack_ets: stack_ets} = opts, name) do
     :ets.update_element(
@@ -550,13 +502,15 @@ defmodule Electric.ShapeCache.PureFileStorage do
   end
 
   defp read_latest_offset(%__MODULE__{} = opts) do
-    metadata =
-      read_multiple_cached_metadata(opts, [
-        :last_seen_txn_offset,
-        :last_persisted_txn_offset,
-        :last_snapshot_chunk
-      ])
+    read_multiple_cached_metadata(opts, [
+      :last_seen_txn_offset,
+      :last_persisted_txn_offset,
+      :last_snapshot_chunk
+    ])
+    |> latest_offset()
+  end
 
+  defp latest_offset(metadata) do
     latest_offset =
       Keyword.get(metadata, :last_seen_txn_offset) ||
         Keyword.get(metadata, :last_persisted_txn_offset)
@@ -599,9 +553,18 @@ defmodule Electric.ShapeCache.PureFileStorage do
     close_all_files(state)
   end
 
-  def terminate(writer_state(opts: opts) = state) do
+  def terminate(writer_state(ets: ets_table, opts: opts) = state) do
     close_all_files(state)
-    delete_shape_ets_entry(opts.stack_id, opts.shape_handle)
+    try(do: :ets.delete(ets_table), rescue: (_ -> true))
+    clean_shape_ets_entry(opts)
+  end
+
+  # remove cached values not needed for the read path
+  defp clean_shape_ets_entry(%__MODULE__{} = opts) do
+    write_metadata_cache(
+      opts,
+      for(k <- @clean_cache_keys, do: {k, storage_meta_unset(k)})
+    )
   end
 
   defp close_all_files(writer_state(writer_acc: acc) = state) do
@@ -627,14 +590,15 @@ defmodule Electric.ShapeCache.PureFileStorage do
       write_shape_definition!(opts, shape_definition)
     end
 
-    suffix = latest_name(opts) || write_metadata!(opts, :latest_name, "latest.0")
+    suffix =
+      read_cached_metadata(opts, :latest_name) || write_metadata!(opts, :latest_name, "latest.0")
 
     {last_persisted_txn_offset, json_file_size, chunks} =
       if initialize? do
         {LogOffset.last_before_real_offsets(), 0, []}
       else
         last_persisted_txn_offset =
-          read_metadata!(opts, :last_persisted_txn_offset) ||
+          read_cached_metadata(opts, :last_persisted_txn_offset) ||
             LogOffset.last_before_real_offsets()
 
         trim_log!(opts, last_persisted_txn_offset, suffix)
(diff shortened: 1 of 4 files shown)


msfstef:
benchmark this

packages/sync-service/lib/electric/shapes/api/delete.ex
```diff
@@ -40,7 +40,11 @@ defmodule Electric.Shapes.Api.Delete do
 
   defp get_shape_handle(%Request{} = request) do
     %{params: %{shape_definition: shape}, api: api} = request
-    Shapes.get_shape(api.stack_id, shape)
+
+    case Shapes.fetch_handle_by_shape(api.stack_id, shape) do
```
magnetised, line 44: thanks for doing this rename -- the existing apis were very unclear

msfstef:
One thing I figured out is that the read through cache is not actually active if the shape consumer is not "alive", because when a consumer is shut down we also take down the stack ETS entry:

https://github.com/electric-sql/electric/blob/c6fb88596e927d063c8f645c864f245ecad78d3e/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex#L537-L540

https://github.com/electric-sql/electric/blob/c6fb88596e927d063c8f645c864f245ecad78d3e/packages/sync-service/lib/electric/shapes/consumer.ex#L818-L830

So somehow we need to keep alive this read-through cache, or use a different one since this one is also not alive if the consumer is never started (which would be common with lazy initialization). Open to ideas, I would like to have a single cache ideally rather than a separate one. Perhaps upon recovering shapes we do an initialization on them that isn't necessarily starting a whole process?

@magnetised would like to hear your thoughts

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -618,11 +603,6 @@ defmodule Electric.Shapes.Consumer do
     state
   end
 
-  defp mark_snapshot_started(%State{stack_id: stack_id, shape_handle: shape_handle} = state) do
-    :ok = ShapeCache.ShapeStatus.mark_snapshot_as_started(stack_id, shape_handle)
```
magnetised, line 620: re our conversation yesterday about giving shape db some kind of validity flag, is it worth keeping this?

packages/sync-service/test/electric/shapes/consumer_test.exs
```diff
@@ -567,39 +539,6 @@ defmodule Electric.Shapes.ConsumerTest do
       refute_receive {Electric.ShapeCache.ShapeCleaner, :cleanup, @shape_handle2}
     end
 
-    test "unexpected error while handling events stops affected consumer and cleans affected shape",
-         ctx do
-      expect_shape_status(
-        set_latest_offset: fn _, @shape_handle1, _ ->
-          raise "The unexpected error"
```
magnetised, line 584: could we keep this test but just move the crash onto a function that **is** called?

magnetised:
benchmark this

packages/sync-service/lib/electric/shapes/consumer.ex
msfstef: I think not because the validity will be reliant on a `snapshot_finished?` rather than this snapshot started, so it needs to be implemented in a different way anyway.

magnetised:
> So somehow we need to keep alive this read-through cache, or use a different one since this one is also not alive if the consumer is never started (which would be common with lazy initialization). Open to ideas, I would like to have a single cache ideally rather than a separate one. Perhaps upon recovering shapes we do an initialization on them that isn't necessarily starting a whole process?

Good catch. We need some stack-level equivalent to the shape status ets for the storage. We have the `Storage.stack_start_link/1` function so that shouldn't be too hard to wrangle in

packages/sync-service/test/electric/shapes/consumer_test.exs
msfstef: I tried but couldn't find something that made sense - we are literally crashing the process in the next test (which a raise does exactly that, sends an exit signal to itself AFAIK) so I figured this is unnecessary coverage

msfstef:
benchmark this

msfstef:
@magnetised I've added a read-through cache on the storage layer so that readers directly go from memory when the shape is "sleeping"

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -223,6 +212,15 @@ defmodule Electric.ShapeCache.PureFileStorage do
     end
   end
 
+  defp delete_shape_ets_read_through_cache_entry(stack_id, shape_handle) do
```
magnetised: nit: I'm all for clear function names but this is a little verbose, no?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -517,6 +516,9 @@ defmodule Electric.ShapeCache.PureFileStorage do
       WriteLoop.cached_chunk_boundaries(initial_acc)
     )
 
+    # remove any existing read-through cache now that writer is active
```
magnetised: because?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -1087,7 +1160,7 @@ defmodule Electric.ShapeCache.PureFileStorage do
       case :ets.lookup(stack_ets(opts.stack_id), handle) do
         [] ->
           # Writer's not active, only disk reads are possible
-          offset = get_latest_offset(opts)
+          {:ok, offset} = get_latest_offset(opts)
```
magnetised: nit: `{:ok, value}` returning functions should be `fetch...`

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -662,33 +727,47 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
-    :cached_chunk_boundaries
-  ]
+  # Read metadata with read-through cache in ETS to minimise disk reads when
+  # writer is not active
+  defp read_metadata_with_read_through_cache(%__MODULE__{shape_handle: handle} = opts, key)
+       when key in @read_through_cached_keys do
+    case :ets.lookup_element(
+           stack_ets(opts.stack_id),
+           {:read_through_cache, handle},
+           get_read_through_cache_key_pos(key) + 1,
```
magnetised: nice use of records

packages/sync-service/test/electric/shape_cache/pure_file_storage_test.exs
```diff
@@ -78,8 +78,11 @@ defmodule Electric.ShapeCache.PureFileStorageTest do
       # This deregisters the writer from the ETS
       PureFileStorage.terminate(writer)
 
-      assert PureFileStorage.get_current_position(opts) ==
-               {:ok, LogOffset.new(11, 0), %{xmin: 100}}
+      assert PureFileStorage.get_latest_offset(opts) ==
+               {:ok, LogOffset.new(11, 0)}
+
+      assert PureFileStorage.get_pg_snapshot(opts) ==
+               {:ok, %{xmin: 100}}
```
magnetised, line 99: nit: `fetch_pg_snapshot`?

packages/sync-service/test/electric/shape_cache/pure_file_storage_test.exs
```diff
@@ -78,8 +78,11 @@ defmodule Electric.ShapeCache.PureFileStorageTest do
       # This deregisters the writer from the ETS
       PureFileStorage.terminate(writer)
 
-      assert PureFileStorage.get_current_position(opts) ==
-               {:ok, LogOffset.new(11, 0), %{xmin: 100}}
+      assert PureFileStorage.get_latest_offset(opts) ==
```
magnetised: nit: `fetch_latest_offset`?

magnetised (approved):
Looks good, though I'm confused why we're only reading from the read through cache when the writer is inactive. Feels messy to have two modes. Would it be horrible to remove the split?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
msfstef: very verbose - but couldn't think of a better name that retains the clarity

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
msfstef: the read through cache is now unnecessary, and will also be outdated, so I'd rather remove it rather than leave outdated data hanging in memory

msfstef:
> Looks good, though I'm confused why we're only reading from the read through cache when the writer is inactive. Feels messy to have two modes. Would it be horrible to remove the split?

@magnetised the read through cache is a cache that gets populated on read - the `storage_meta` cache is a purely write-populated cache. I was slightly worried about any potential conflicts, but that's easily resolvable.

Because `storage_meta` stores a bunch of things that are only relevant when a writer is active, I wasn't sure if it was reasonable to keep the whole thing around for as long as the shape itself is around. I think you're right that we can consolidate the two, I was kind of trying to not "touch" the main implementation too much in implementing this I suppose, while implementing the desired behaviour.

If you're willing to have a go at consolidating them I'd be happy to accept some help on this. We could also merge this and do it in a separate PR (?)

magnetised:
@msfstef I switched things so that some keys are always written to the stack cache, even if there's a writer active. Simplifies things at the expense of splitting reads/writes across two tables.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -1352,30 +1358,26 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
-        %__MODULE__{shape_handle: handle} = opts,
+        %__MODULE__{} = opts,
         last_persisted_txn_offset,
         last_persisted_offset,
         last_seen_txn_offset,
         old_last_persisted_txn_offset
       ) do
     if old_last_persisted_txn_offset != last_persisted_txn_offset do
-      write_metadata!(opts, :last_persisted_txn_offset, last_persisted_txn_offset)
+      write_cached_metadata!(opts, :last_persisted_txn_offset, last_persisted_txn_offset)
```
msfstef: Not sure how this affects performance, I had decided to leave it as is purely for the purpose of not doing additional roundtrips to ETS as there seemed to have been an intentional effort to avoid them.

This call is on the hot path so it's an important thing to note.

packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -2702,7 +2702,7 @@ defmodule Electric.Plug.RouterTest do
           ],
           do: Postgrex.query!(ctx.db_conn, stmt)
 
-      Process.sleep(100)
+      Process.sleep(120)
```
msfstef, line 2705: we could alternatively just have a helper `assert_with_timeout` where it retries the assertion for a maximum of 500ms

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -57,10 +57,62 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+  @stack_cached_keys [
+    :latest_name,
+    :compaction_boundary,
+    :last_persisted_txn_offset,
+    :snapshot_started?,
+    :last_snapshot_chunk
+  ]
+
+  # Record that caches read-through metadata to avoid disk reads
+  defrecord :storage_meta_stack,
```
msfstef: the name `storage_meta_stack` as opposed to `storage_meta` I have to admit is a bit confusing - what was the reasoning for this name?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -810,6 +796,26 @@ defmodule Electric.ShapeCache.PureFileStorage do
     end
   end
 
+  defp write_metadata_cache(%__MODULE__{} = opts, key, value)
+       when key in @stack_cached_keys do
```
msfstef: if we're always writing to this, and _not_ the other, should the other not be storing them in ETS at all? This looks like a good change but I'm confused that it's leaving "code residue" in the existing ETS that's a bit hard to decode

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
magnetised: not sure what you mean by "as opposed to" here. I renamed it from `storage_meta_read_through` to `storage_meta_stack` - the reasoning is that these are the values maintained at the stack level, not by the writer. 

I'll admit that the naming isn't 100%. perhaps storage_meta_global or something

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
magnetised: the existing ets is still used for everything that isn't made available for the read path. also, what do you mean by "the other" here. wanna be sure what you mean.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
msfstef: That's what I figured, the confusing aspect is that `storage_meta` is also stored in what we call the "stack ets" - so the whole thing is a bit hard to decode

a reference to it being a cache record (similar to how you renamed the ets to a cache table) might work?

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
magnetised: this `last_persisted_txn_offset` is available to the read path, so if we don't do a write-through here then reads will go to the disk (or will be stale).
also note that the original does a write to the disk here, then a write to ets immediately below, so there's not a huge difference (apart from having to split the writes between tables, so not being able to issue a single `update_element` call

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
msfstef: by "the other" I do mean the original stack ets - basically I'm asking, given that we're storing the same keys in two different ETS tables and records, how are we ensuring they are never out of sync?

magnetised:
benchmark this

magnetised:
@msfstef this now uses the stack ets for the read cache and write cache, i've merged into this commit:

[`fe0d5c8` (#3572)](https://github.com/electric-sql/electric/pull/3572/commits/fe0d5c829cc3159d95af79815b60f4087bd57bcc)

the read path is careful about initializing the cache from disk -- theres a potential race with the writer initialising so it uses insert_new, the write path isn't - it just initializes the cached values because it knows that it's in charge. rather than deleting the cached values when the writer terminates, it just nils the cached values that aren't needed for the write path. 

i've looked at clean up as well and decided to always explictly delete the writer's ets table at the cost of a few extra file writes.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -710,51 +659,44 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
+    keys =
+      for key <- read_keys do
+        {key, read_metadata!(opts, key)}
+      end
+
+    meta = create_storage_meta([{:shape_handle, handle} | keys])
+
+    # prevent race conditions where a writer initialises the ets
+    # for a shape after we've checked for existence
+    :ets.insert_new(stack_ets, meta)
```
msfstef, line 696: very nice

packages/sync-service/lib/electric/shape_cache/pure_file_storage/shared_records.ex
```diff
@@ -26,4 +35,86 @@ defmodule Electric.ShapeCache.PureFileStorage.SharedRecords do
...
+    defp get_storage_meta_value(unquote(key), meta) do
+      {unquote(key), storage_meta(meta, unquote(key))}
+    end
+  end
+
+  for key <- @value_keys do
+    def storage_meta_key_pos(unquote(key)), do: storage_meta(unquote(key)) + 1
+  end
+
+  # def storage_meta_compare_and_swap(table, handle, key, expected_value \\ @unset) do
```
msfstef, line 82: remove commented code?

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -405,10 +405,12 @@ defmodule Electric.Shapes.Consumer do
       end
     end)
 
-    case ShapeCleaner.handle_writer_termination(state.stack_id, state.shape_handle, reason) do
-      :ok -> terminate_writer(state)
-      :removed -> :ok
-    end
+    # always need to terminate writer to remove the writer ets (which belongs
+    # to this process). leads to unecessary writes in the case of a deleted
```
msfstef, line 409: to clarify - the unnecessary writes you mean after the shape has been marked for removal, it might receive some events that it will write to the log? in a future pr we could probably just fix that since we mark the removal anyway

otherwise the `terminate_writer` technically just closes files (which is a no-op if they were already closed)

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -710,51 +659,44 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
-  rescue
-    ArgumentError -> read_metadata!(opts, key)
   end
 
-  defp read_metadata_with_stack_cache(%__MODULE__{} = opts, key),
-    do: read_metadata!(opts, key)
+  defp populate_read_through_cache!(%__MODULE__{} = opts, extra_keys) do
+    %{shape_handle: handle, stack_ets: stack_ets} = opts
+
+    read_keys = Enum.into(extra_keys, MapSet.new(@read_path_keys))
```
msfstef, line 685: so to clarify - if there's no active "cache", then on any request for a single key will trigger as many reads as there are `@read_path_keys`? I thought the idea was to only populate it with the desired key and everything else would have a placeholder `:not_cached`


Unless the thought here is that if someone is doing a read, they will probably need all the read path keys anyway, so might as well load them all up front if someone needs a read. Will appreciate to know what the logic is

packages/sync-service/lib/electric/shapes/consumer.ex
magnetised: a notionally deleted shape may have some buffered writes that the Storage.terminate call will flush to disk. we could save those writes, which is what I was doing by deleting the writer, but in the end I think it's worth knowing that the cleanup will always execute.

packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
magnetised: > Unless the thought here is that if someone is doing a read, they will probably need all the read path keys anyway, so might as well load them all up front if someone needs a read

That's the logic -- there is no partial read: if you need one value you'll need them all.

PR #3573, opened 2025-12-09
chore: Fix flaky restart test

msfstef:
Slightly higher timeout for a flaky test since it waits for PG stuff on CI which is slow.

Also, I'm removing the hack for uploading branch test results to a single branch on codecov cause they've finally added an "all branches" view allowing us to see flakes across PRs and main


robacourt (approved):
Nice!

PR #3575, opened 2025-12-10
fix: parse integers larger than int32 max in WHERE clauses

ibarakaiev:
Fixes a bug where WHERE clauses with integer literals larger than 2,147,483,647 (int32 max) would fail with an error like:

      * 1st argument: not a textual representation of a float

Root cause:
- libpg_query/PgQuery uses int32 for integer constants in the AST
- Integers exceeding int32 max are stored as Float nodes with string values (e.g., {:fval, "2793017076"}) - this is documented behavior
- The parser called String.to_float/1 on these strings, but Elixir's String.to_float/1 requires a decimal point and fails on "2793017076"

Solution:
- Use Integer.parse/1 to detect pure integer strings in fval values
- Parse them as int8 (bigint) instead of numeric
- This matches PostgreSQL's type inference for integer literals and ensures correct type matching with bigint columns (no implicit cast exists from numeric to int8)

Affected queries (now working):
- WHERE id = 2793017076
- WHERE id IN (2147483648, 3500000000)
- WHERE id > 2147483647

References:
- libpg_query wiki: https://github.com/pganalyze/libpg_query/wiki/Differences-between-JSON-output-formats ("This leads to problems with overly long integers or floats")

### DIFF
--- a/packages/sync-service/test/electric/replication/eval/parser_test.exs
+++ b/packages/sync-service/test/electric/replication/eval/parser_test.exs
@@ -673,6 +673,14 @@ defmodule Electric.Replication.Eval.ParserTest do
                )
 
       assert %Func{args: [%Ref{type: :int8}, %Const{type: :int8, value: -3_000_000_000}]} = result
+
+      # Actual floats should still be parsed as numeric
+      assert {:ok, %Expr{eval: result}} =
+               Parser.parse_and_validate_expression(~S|value = 2147483648.5|,
+                 refs: %{["value"] => :numeric}
+               )
+
+      assert %Func{args: [%Ref{type: :numeric}, %Const{type: :numeric}]} = result
     end
 
     test "implements common array operators: @>, <@, &&, ||" do


packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -1386,6 +1386,62 @@ defmodule Electric.Plug.RouterTest do
       assert op2_log_offset == LogOffset.increment(op1_log_offset, 2)
     end
 
+    @tag with_sql: [
+           "CREATE TABLE bigint_test (id BIGINT PRIMARY KEY, value TEXT NOT NULL)",
+           # Insert values larger than int32 max (2,147,483,647)
+           "INSERT INTO bigint_test VALUES (2147483648, 'just above int32'), (2793017076, 'large bigint'), (3500000000, 'another large')"
+         ]
+    test "GET supports WHERE clause with integers larger than int32 max", %{opts: opts} do
```
icehaunter: Let's remove this test, as these are integration tests which take a while, and a parser edge case is better covered by the unit test in the parser you have

packages/sync-service/lib/electric/replication/eval/parser.ex
```diff
@@ -1232,7 +1232,15 @@ defmodule Electric.Replication.Eval.Parser do
...
-        {:ok, %Const{type: :numeric, value: String.to_float(value), location: loc}}
+        # PgQuery/libpg_query stores integers larger than int32 max as {:fval, "string"}.
+        # We need to detect pure integer strings and parse them as int8.
+        case Integer.parse(value) do
+          {int_value, ""} when is_pg_int8(int_value) ->
+            {:ok, %Const{type: :int8, value: int_value, location: loc}}
+
+          _ ->
+            {:ok, %Const{type: :numeric, value: String.to_float(value), location: loc}}
+        end
```
icehaunter, line 1243: There's no reason to make this case, because there exists a `Float.parse/1` function which parses strings without a decimal point just fine

icehaunter:
You should also run `mix format` it seems

packages/sync-service/lib/electric/replication/eval/parser.ex
ibarakaiev: @icehaunter Happy to make this change, quick question:

If we have an integer like 2^53 = 9007199254740993 which gets parsed by `libpg_query` as a float string and `Float.parse("9007199254740993")` returns `{9007199254740992.0, ""}`, wouldn't that be an issue since we just lost precision due to it being outside of Elixir's float range (2^53)? The use case is if someone uses BIGINT with some random IDs up to 10^18.

ibarakaiev: Hey @icehaunter, just bumping this question, curious what you think.

ibarakaiev: Bump! @icehaunter

PR #3576, opened 2025-12-10
feat: add `snapshot-end` messages as boundaries after a move-in

### LINKED ISSUE #3569: Add "visibility boundary" message after move-in data to allow for transaction ID mapping
This is important for TanStack DB to be able to reason about the visibility of a given transaction if data appears as a move-in, not as an original txn.

icehaunter:
This is required for tanstack/db to correctly track optimistic write resolution. Closes #3569


robacourt (approved):
Nice!

PR #3585, opened 2025-12-11
feat: Reduce off heap allocations in `ShapeLogCollector`

msfstef:
After an investigation with @alco on memory allocations, we've observed that there's lots of strings being allocated off-heap by the SLC.

This is generally not an issue for most processes, but for a high-frequency, performance sensitive process like SLC we want to minimize unnecessary work and allocations and be extra careful about what code executes in the very tight loop.

I've painstakingly gone through every allocation happening by the process in the various steps of consuming a transaction, and have identified 3 places where we were allocating things when we didn't need to.

The main culprit is string interpolation, which allocates an intermediate representation to perform the interpolation.

1. In our `OpenTelemetry.start_interval`, we pass a string, and later on we do a string interpolation and atom creation to store  the attribute. We do this _for every_ interval type, _on every_ transaction fragment handling.
2. The flush tracker notifies the replication client, but we were calling `ReplicationClient.name` on _every_ notification call, which does a string interpolation + atom cast.
3. Updating the `LsnTracker` ETS table was using the stack id, which does a string interpolation for getting the table name.

These would generally be ok but at the rate at which they occur and for how long lived the process is, these are potentially impacting performance.

The solutions were easy:
1. Simply specify interval names as atoms, so we don't dynamically generate any (which is safer) and we don't incur the cost of string literals and concatenations - all compile time stuff.
2. Very simple, just generate the name at process initialization and pass the atom to the closure rather than generate it in the closure
3. A bit annoying, but make the API accept a `stack_ref` which can be either the stack ID or the table name, and add a `table_name` utility in order to optimize the calls for this.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -245,7 +245,7 @@ defmodule Electric.ShapeCache do
 
       {:ok, shape_handle} = ShapeStatus.add_shape(stack_id, shape)
 
-      Logger.debug("Creating new shape for #{inspect(shape)} with handle #{shape_handle}")
+      Logger.info("Creating new shape for #{inspect(shape)} with handle #{shape_handle}")
 
       {:ok, _pid} = start_shape(shape_handle, shape, state, otel_ctx, :create)
 


packages/sync-service/lib/electric/telemetry/open_telemetry.ex
```diff
@@ -179,9 +179,9 @@ defmodule Electric.Telemetry.OpenTelemetry do
   e.g.
 
   ```elixir
-  OpenTelemetry.start_interval("quick_sleep")
+  OpenTelemetry.start_interval(:quick_sleep)
```
robacourt:
```suggestion
  OpenTelemetry.start_interval(:"quick_sleep.duration_µs")
```
just so the comment below it makes sense and to show best practices

packages/sync-service/lib/electric/telemetry/open_telemetry.ex
```diff
@@ -179,9 +179,9 @@ defmodule Electric.Telemetry.OpenTelemetry do
...
   Process.sleep(1)
-  OpenTelemetry.start_interval("longer_sleep")
+  OpenTelemetry.start_interval(:longer_sleep)
```
robacourt:
```suggestion
  OpenTelemetry.start_interval(:"longer_sleep.duration_µs")
```

robacourt (approved):
Very nice! Just left a couple of suggestions

msfstef:
benchmark this

balegas:
Thanks for digging into this

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -245,7 +245,7 @@ defmodule Electric.ShapeCache do
 
       {:ok, shape_handle} = ShapeStatus.add_shape(stack_id, shape)
 
-      Logger.info("Creating new shape for #{inspect(shape)} with handle #{shape_handle}")
+      Logger.debug("Creating new shape for #{inspect(shape)} with handle #{shape_handle}")
```
alco: This is fine, albeit inconsitent given that there are other log entries tied to individual shapes that are logged at `info` level: the ones in `ShapeCleaner`, `ExpiryManager`,

msfstef: Reverting this change as it is inappropriate for this PR anyway - you're right, this was mostly an annoyance thing as in a surge of new shapes we overwhelm the logger at info level

msfstef:
> So the working theory is that BEAM allocates more space than is actually needed when doing string interpolation and this is how we end up with thousands of off-heap binaries that are all 256 bytes long?

That's exactly it yes - you can try doing string interpolation on a process and you'll notice 256 byte strings being allocated. I can't confirm this with source ATM, this is likely a minimum allocation for optimization purposes.

Either way, it ended up allocating at least 8 256 strings because of interpolations _on every_ transaction handled.

> I wonder if we could use a single process registry for all stacks.

It might be worth figuring out yeah - although we would still have the issue with the scoping of process names and in fact I'd like to take a look at other interpolations we do - e.g. the ShapeStatus calls are now all doing an interpolation _on every_ call and I wonder if that's an issue.

PR #3588, opened 2025-12-12
ci: Build Postgres images preconfigured with logical replication for use in CI workflows

alco:
This serves multiple goals:
- avoid repeating configuration code between CI workflows
- get rid of the need to run `psql` or `docker exec` commands in addition to defining Postgres as a service in those CI workflows that need it. See https://github.com/electric-sql/electric/pull/3600.
- https://github.com/electric-sql/electric/pull/3594 packages up sync-service as a Docker image for use in CI. However, it cannot be easily started as a service if the postgres service needs to be configured afterwards

In other words, this together with building a Docker image for sync service to use in CI workflows from the other PR helps avoid redoing some of the build steps between different CI workflows.

### DIFF
--- a/.github/workflows/build_postgres_image.yml
+++ b/.github/workflows/build_postgres_image.yml
@@ -1,0 +1,54 @@
+name: Postgres image for CI
+
+on:
+  # Run daily at 00:00 UTC. This allows the image to pick up updates from upstream and avoid
+  # going stale.
+  schedule:
+    - cron: '0 0 * * 0'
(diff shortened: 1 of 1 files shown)


alco:
I have [tested this workflow](https://github.com/electric-sql/electric/actions/runs/20186884899) by adding the `on.pull_request` trigger to the workflow file temporarily and later removing it.

robacourt (approved):
Nice!

PR #3594, opened 2025-12-12
ci: Build a Docker image for sync-service to reuse in CI workflows

alco:
Instead of duplicating caching and compilation logic for Electric in TS tests, build a Docker image from sync-service's source code and run it as a Docker container.

In a followup PR I'm going to also use the sync-service Docker container when running Lux integration tests.

One positive outcome from this is that we no longer need a dummy workflow to test that sync-service works when running in a Docker container. It is now implicitly verified by running it in a Docker container in existing workflows.

- [x] Replace the stub image tag derivation with a proper one that takes into account all files relevant for the Docker image build of sync-service

### DIFF
--- a/packages/sync-service/dev/docker-compose-electric.yml
+++ b/packages/sync-service/dev/docker-compose-electric.yml
@@ -1,10 +1,8 @@
 services:
   electric:
     image: $ELECTRIC_IMAGE
+    env_file: ../.env.dev
     environment:
-      ELECTRIC_INSECURE: true
-      ELECTRIC_USAGE_REPORTING: false
-      ELECTRIC_ENABLE_INTEGRATION_TESTING: true
       DATABASE_URL: 'postgresql://postgres:password@postgres/electric?sslmode=disable'
     ports:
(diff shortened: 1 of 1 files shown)


alco:
I've finished doing CI updates for this PR but there are three TS tests that keep failing consistently. Currently not sure what's causing that.

alco:
> I've finished doing CI updates for this PR but there are three TS tests that keep failing consistently. Currently not sure what's causing that.

Figured it out: had to include `packages/sync-service/.env.dev` in the compose file for Electric because it defines bespoke caching config for the dev env which TS tests rely on.

alco:
@msfstef I think performance difference will be negligible.

What this approach gives us is:
- less book-keeping of sync-service source and deps in TypeScript tests
- verifying that the Electric runs fine as a Docker container (we had issue with it in the past, that was the reason for creating the "Docker Image Smoketest" workflow)
- with a few upcoming customizations, the workflow that builds the image can also be used for building public Docker images for Electric

PR #3595, opened 2025-12-12
feat: add `@electric-sql/start` package that interacts with Cloud

thruflo:
@icehaunter I remember now that this WIP was me pulling out the start package specifics to a clean PR, then fixing stuff in it. See the TODO.md. When verified as working, this PR can be merged and the `@electric-sql/start` package published, and then the quickstart branch can be published with just the website changes. That's the idea ...

### DIFF
--- a/examples/tanstack-db-web-starter/src/vite-plugin-caddy.ts
+++ b/examples/tanstack-db-web-starter/src/vite-plugin-caddy.ts
@@ -13,7 +13,7 @@ interface CaddyPluginOptions {
 export function caddyPlugin(options: CaddyPluginOptions = {}): Plugin {
   const {
     host = `localhost`,
-    httpsPort = 4000,
+    httpsPort = 5173,
     encoding = true,
     autoStart = true,
     configPath = `Caddyfile`,


thruflo:
The trouble with the https://tanstack-start-db-electric-starter.localhost approach in the current starter is it doesn't work in Safari:

Screenshot 2025-12-22 at 11 19 00 AM

There's no actual need to use the subdomain like this. We can just run on a normal / default vite address.

In the port-sharing version of the start I did, https://github.com/electric-sql/electric/pull/3228, I solve this and also remove the need for Caddy and avoid the double security popup when installing. For me, running this starter from scratch on new macOS, I get two system-level security popups and a console warning when I run `pnpm run dev` on the starter.

thruflo:
There's also another DX problem which is that running the migrations after the dev server makes it likely to forget to run the migrations. You then open the web browser, try to login and get an auth error. I think it may be better to just do less magic and have explicit instructions in the README and docker version of the Quickstart to (a) ensure you have caddy and have run caddy trust (b) run the backend (c) then run the migrations (d) then run the dev server. This would mean adjusting the dev command to not run the backend services.

thruflo:
@icehaunter have updated the readme as envisaged on call earlier. Needs your changes for this to actually work.

thruflo:
Quickstart updated to match this: https://github.com/electric-sql/electric/pull/3114

packages/start/src/cli.ts
```diff
@@ -0,0 +1,66 @@
...
+    // Step 2: Setup TanStack Start template
+    console.log(`Setting up template...`)
+    await setupTemplate(appName, credentials)
+
+    // Step 3: Display completion message
+    console.log(`Setup complete`)
+    console.log(``)
+    console.log(`Next steps:`)
+    console.log(`  cd ${appName}`)
+    console.log(`  pnpm install`)
```
thruflo: The idea here: https://deploy-preview-3114--electric-next.netlify.app/docs/quickstart#get-started is that the start command wraps up install and migrate (because it can and it just simplifies the Quickstart flow).

(This also affects the `.stackblitzrc` as to whether it installs reps or not).

Any reason not to switch this over (aside from explicitness)?

thruflo:
@icehaunter see https://github.com/electric-sql/electric/pull/3595/changes/9f4db14fcc671759d2ff7057e642ca2b0ce6e952

packages/start/src/cli.ts
icehaunter: We can do a manual pnpm install & migrate if you want, although I slightly prefer the explicitness and not auto-installing stuff. Is it also worth noting in the quickstart guide the fact that NeonDB databases will expire in about an hour unless claimed?

icehaunter:
@thruflo - addressed your feedback in https://github.com/electric-sql/electric/pull/3595/commits/bb25a026cb8a49420423d3cf717ea51ac299c403

1. Only one dev command, `pnpm dev`, driven by `.env` file
2. `ELECTRIC_URL` being set takes the precedence - if you want
to run against docker, you need to unset it or set to the docker url
3. `@electric-sql/start` will run `pnpm install` & `pnpm migrate` immediately to match quickstart
4. Made sure Caddy runs on `localhost:4000` with an HTTPS cert instead of `.localhost` domain

PR #3598, opened 2025-12-12
change build target to ES2017 to get native async/await to fix generator bug

### LINKED ISSUE #3566: requestShape seems to create a chain of un-GC-able Promises
**Versions**

- Client: typescript, 1.2.0
- Electric: 1.2.7

**Bug description**
After being connected to the server for a decent period of time, on a dataset that has a fairly large amount of data (10s of thousands of rows) and is updating pretty often, we've started to see issues with tabs freezing or becoming unresponsive. After being very confused about this for a while, we realized you can pause these tabs' JS execution in the debugger while this is happening. Whenever we did that, execution was always inside of the electric-sql client with a long chain of promises in the call stack. This was suspicious, but not really a smoking gun since it may just be electric-sql is executing code often but not an issue.

I realized tonight that you can also get a heapdump while paused, and found this incredibly long chain of very small promises making up a much larger amount of retained memory:

Image

Image

The 3 closures there that are retaining these promises are here:

```ts
e0 = (e, t, n) => new Promise( (r, i) => {
            var o = e => {
                try {
                    a(n.next(e))
                } catch (e) {
                    i(e)
                }
            }
              , s = e => {
                try {
                    a(n.throw(e))
                } catch (e) {
                    i(e)
                }
            }
              , a = e => e.done ? r(e.value) : Promise.resolve(e.value).then(o, s);
            a((n = n.apply(e, t)).next())
        }
        )
```

This is the `__async` helper from tslib, sitting inside @electric-sql/client's compiled code. In unminified form:

```ts
var __async = (__this, __arguments, generator) => {
  return new Promise((resolve, reject) => {
    var fulfilled = (value) => {
      try {
        step(generator.next(value));
      } catch (e) {
        reject(e);
      }
    };
    var rejected = (value) => {
      try {
        step(generator.throw(value));
      } catch (e) {
        reject(e);
      }
    };
    var step = (x) => x.done ? resolve(x.value) : Promise.resolve(x.value).then(fulfilled, rejected);
    step((generator = generator.apply(__this, __arguments)).next());
  });
};
```

I think what is going on is, some of the async methods are recursive, e.g. [requestShape](https://github.com/electric-sql/electric/blob/e90e24e472e8194dc2f54863f1605574aea10800/packages/typescript-client/src/client.ts#L727). In actual async/await code, I don't think this is a problem: the Promises earlier in the chain can be GC'd once nobody holds a reference to them any more. With the __async compatibility shim, however, these get compiled down to generators, and we end up with a generator that returns a generator that returns a generator that returns a generator... with none of the initial generators able to fully complete because they are waiting on the one they called to yield.

I think there's a few potential fixes for this:

1) Do you actually need to support ES2016 (as is specified in your [tsconfig](https://github.com/electric-sql/electric/blob/main/packages/typescript-client/tsconfig.json)). If not, bumping that to at least ES2017 would make TS use native async/await (ES2018 for async iterators), and I don't think this would be an issue with that.

2) Don't use recursion, move to a top-level loop or similar

3) Break out of the stack in some way (queueMicrotask, etc.)

KyleAMathews:
Fixes #3566

The ES2016 target caused async/await to be transpiled using the `__async` helper which wraps each async function in a generator-based Promise. When recursive async functions like `requestShape()` call themselves, this creates nested Promise chains that cannot be garbage collected - each generator/Promise pair holds a reference to the next.

With ES2017+, native async/await is used which doesn't have this issue. The browser support difference is negligible (95.3% vs 95.6%), with the unsupported browsers being primarily IE11 (EOL June 2022).

### DIFF
--- a/.changeset/fix-memory-leak-es2017.md
+++ b/.changeset/fix-memory-leak-es2017.md
@@ -1,0 +1,7 @@
+---
+'@electric-sql/client': patch
+---
+
+Fix memory leak from recursive async functions by upgrading TypeScript target to ES2017.
+
+The ES2016 target caused async/await to be transpiled using the `__async` helper which creates nested Promise chains that cannot be garbage collected when recursive async functions like `requestShape()` call themselves. With ES2017+, native async/await is used which doesn't have this issue.


justinbelcher:
Thank you!

PR #3600, opened 2025-12-13
ci: Use the preconfigured Postgres image in CI workflows

alco:
Simplify starting a Postgres database on CI. Having a [preconfigured image available](https://github.com/electric-sql/electric/pull/3588) eliminates the need for manual configuration steps inside the CI job. This will also work great together with https://github.com/electric-sql/electric/pull/3594.


msfstef (approved):
Great idea

PR #3602, opened 2025-12-13
Debug and fix stuck test in PR

KyleAMathews:


alco:
I've cancelled the stuck job after I noticed it had been running for 1 hour this morning. Still the same OOM error from running the test. Do you think we have a memory leak somewhere?

KyleAMathews:
@alco looks like it's working now — the problem was a infinite loop caused by using Promise.resolve() (resolves inside a microtask before timers run) so test timeouts didn't run.

PR #3603, opened 2025-12-14
Add health check documentation to deployment guide

KyleAMathews:
Document the /v1/health endpoint for liveness and readiness probes, including HTTP status codes, response format, and note that it works without authentication when ELECTRIC_SECRET is set.


alco (approved):
Looks great!

PR #3605, opened 2025-12-15
Cache lux on CI

alco:


alco:
Included this change in https://github.com/electric-sql/electric/pull/3621.

PR #3611, opened 2025-12-16
fix(client): encode double-quoted identifiers in encodeWhereClause

ibarakaiev:
## Summary

Fix `encodeWhereClause` to encode double-quoted identifiers. This addresses https://github.com/TanStack/db/issues/927.

## Problem

Query builders like `@tanstack/electric-db-collection` generate WHERE clauses (for `subset__where`) with [double-quoted identifiers](https://github.com/TanStack/db/blob/f25f67259179f82c93af47995b4508754d1fc1c8/packages/electric-db-collection/src/sql-compiler.ts#L88):

```sql
"userId" = $1 AND "createdAt" > $2
```

Quoting is a SQL best practice to handle reserved words and preserve case. However, `encodeWhereClause` was intentionally skipping quoted identifiers (line 258 tracks both `'` and `"` as ranges to skip), presumably assuming that if you quoted it, you meant it literally.

This breaks `columnMapper` for any query builder that quotes identifiers - the column names never get transformed (e.g., `userId` → `user_id`).

## Solution

Pre-process double-quoted identifiers before the existing logic runs:

```typescript
whereClause = whereClause.replace(/"([^"]+)"/g, (_, identifier) => {
  return `"${encode(identifier)}"`
})
```

This should be safe because in PostgreSQL, double quotes are exclusively for identifiers (column/table names), never string literals (which use single quotes). So any `"something"` is a column name that should be encoded.

The rest of the function continues to handle unquoted identifiers and skip single-quoted string literals as before.

### DIFF
--- a/packages/typescript-client/test/column-mapper.test.ts
+++ b/packages/typescript-client/test/column-mapper.test.ts
@@ -363,12 +363,13 @@ describe(`encodeWhereClause`, () => {
     ).toBe(`first_name = 'John' AND last_name = 'Doe' AND user_id = $1`)
   })
 
-  it(`should not transform double-quoted identifiers`, () => {
-    // Postgres uses double quotes for case-sensitive identifiers
-    expect(encodeWhereClause(`"userId" = $1`, encode)).toBe(`"userId" = $1`)
+  it(`should transform double-quoted identifiers`, () => {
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Closed in favor of https://github.com/electric-sql/electric/pull/3629

PR #3614, opened 2025-12-16
fix: ensure correct change handling and transformations on active move-ins

icehaunter:
- moves change processing from consumer to a separate module for easier testing without stateful setup
- makes sure that changes only allowed to go through if they are properly covering the in-flight snapshot
- makes sure that updates that don't touch a referenced value are still converted from an update to an insert if it's the first time they are seen.

On the second part, logic is as follows:
- when a change comes, if it doesn't interact with in-flight move-in query results, then append using normal logic
- if a change references the value that was moved in (e.g. for `parent_id IN (SELECT id FROM parent)` there have appeared parents with ids `1` and `2`, and the change on the child table has `parent_id = 1`) (only in-flight queries are considered), then:
  - if the `pg_snapshot` is not known or is known and the txid is visible in the snapshot, then we know that when query results return they will have fresher view of the row and we need to keep the `NewRecord` from the query results; so we skip this change on the stream
    - we can apply this logic for when `pg_snapshot` is not known because that means the query hasn't started resolving yet, and is going to be definitely in the future from the current change.
  - otherwise, if the change is not visible in `pg_snapshot`, then this change has a fresher view of the row as compared to query results, and so we append the change to the log (possibly converting updates to inserts) and then, when applying query results, we'll skip the affected key.

### DIFF
--- a/packages/sync-service/lib/electric/replication/eval/expr.ex
+++ b/packages/sync-service/lib/electric/replication/eval/expr.ex
@@ -82,11 +82,11 @@ defmodule Electric.Replication.Eval.Expr do
   Wrap a parser part (Const, Ref, Func, Array, RowExpr) in an Expr struct, so that it can be evaluated on it's own.
 
   This is used when a subtree of our AST needs to be made evaluatable on it's own inside Electric. The `query` field
-  doesn't make sense in that context, as it's used when going back to postgres, so it's set to a useless value.
+  is not needed in that context, it's used when going back to postgres, so we don't bother calculating it.
   """
   def wrap_parser_part(expr) do
     %__MODULE__{
-      query: "",
+      query: "This should not be executed on the database",
       eval: expr,
       used_refs: Parser.find_refs(expr),
       returns: expr.type


packages/sync-service/lib/electric/shapes/consumer/change_handling.ex
```diff
@@ -0,0 +1,112 @@
...
+  end
+
+  # defp uses_moved_in_values?(referenced_values, state) do
+  #   Enum.any?(state.move_handling_state.in_flight_values, fn {path, values} ->
+  #     case Map.fetch(referenced_values, path) do
+  #       {:ok, used_value} -> MapSet.member?(values, used_value)
+  #       :error -> false
+  #     end
+  #   end)
+  # end
```
robacourt: Should this be removed?

packages/sync-service/lib/electric/replication/eval/expr.ex
```diff
@@ -78,6 +78,16 @@ defmodule Electric.Replication.Eval.Expr do
   defp type_from_json_safe(["enum", type]), do: {:enum, type_from_json_safe(type)}
   defp type_from_json_safe(type) when is_binary(type), do: String.to_existing_atom(type)
 
+  def wrap_eval_expr(expr) do
+    %__MODULE__{
+      # Intentionally wrong query to avoid evaluation
+      query: "0 / 0",
```
robacourt: Really? Surely there is a neater way!

packages/sync-service/lib/electric/shapes/consumer/change_handling.ex
```diff
@@ -0,0 +1,112 @@
+defmodule Electric.Shapes.Consumer.ChangeHandling do
```
robacourt, line 1: Nice refactoring to extract this. Great tests too

packages/sync-service/lib/electric/replication/eval/expr.ex
robacourt: Smells:
- a dummy query  - showing this data structure has a meaningful cohesive subset that should probably be separates
- naming: Eval.Expr.wrap_eval_expr - it's not wrapping an eval expr, at least not a Eval.Expr, it's wrapping Parser.X things which probably need a name

packages/sync-service/lib/electric/replication/eval/expr.ex
```diff
@@ -78,6 +78,21 @@ defmodule Electric.Replication.Eval.Expr do
   defp type_from_json_safe(["enum", type]), do: {:enum, type_from_json_safe(type)}
   defp type_from_json_safe(type) when is_binary(type), do: String.to_existing_atom(type)
 
+  @doc """
+  Wrap a parser part (Const, Ref, Func, Array, RowExpr) in an Expr struct, so that it can be evaluated on it's own.
+
+  This is used when a subtree of our AST needs to be made evaluatable on it's own inside Electric. The `query` field
+  doesn't make sense in that context, as it's used when going back to postgres, so it's set to a useless value.
```
robacourt:
```suggestion
  This is used when a subtree of our AST needs to be made evaluatable on it's own inside Electric. The `query` field
  is not needed in that context, it's used when going back to postgres, so we don't bother calculating it.
```

packages/sync-service/lib/electric/replication/eval/expr.ex
```diff
@@ -78,6 +78,21 @@ defmodule Electric.Replication.Eval.Expr do
...
+  """
+  def wrap_parser_part(expr) do
+    %__MODULE__{
+      query: "",
```
robacourt:
```suggestion
      query: "This should not be executed on the database",
```

robacourt: seems safer to have it fail with a readable error

PR #3620, opened 2025-12-16
Restore stack_id in CallHomeReporter's static info

alco:
It got lost in the extraction of telemetry code into a separate package.

See the old code [here](https://github.com/electric-sql/electric/blob/17132f5f3289b1c6f25181d2a445a74955e0523f/packages/sync-service/lib/electric/telemetry/stack_telemetry.ex#L97-L115), new code [here](https://github.com/electric-sql/electric/pull/3412/changes#diff-1d6e6e1bd32867603bef6bfdbcb4dee699519420fc63f53fcba32e0f62fcc62eR98-R115).

### DIFF
--- a/packages/electric-telemetry/lib/electric/telemetry/call_home_reporter.ex
+++ b/packages/electric-telemetry/lib/electric/telemetry/call_home_reporter.ex
@@ -176,7 +176,7 @@ defmodule ElectricTelemetry.CallHomeReporter do
     {:noreply, state}
   end
 
-  # Catch-all clauses to handle EXIT and DOWN messages from the async task started in `report_home()`.
+  # Catch-all clauses to handle the result, EXIT and DOWN messages from the async task started in `report_home()`.
   def handle_info({task_mon, %Req.Response{}}, state) when is_reference(task_mon) do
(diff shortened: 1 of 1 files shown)


balegas:
Do we have any test coverage for the call home reporter? It would be useful to check if the emitted metrics follow the specified format. We've broken metrics a few times.

balegas:
We were also not adding `electric_default`, iirc?

icehaunter (approved):

alco:
@balegas I have added separate tests that verify CallHomeReporter's payload when running 1) under ApplicationTelemetry and 2) StackTelemetry. In the latter case `stack_id` and `pg_version` will be included under `environment`. These tests only verify the payload shape, not the value of each field.

I had spent way too much time dealing with flakes before I finally got it to work by replacing `Task.start()` with `Task.async()` in `CallHomeReporter` and adding all the necessary clauses of `handle_info`.



> We were also not adding `electric_default`, iirc?

`data.environment.electric_installation_id = "electric_default"` both for application- and stack-based CallHomeReporter. The new tests verify this.

PR #3621, opened 2025-12-17
Enrich process telemetry with binary memory metrics and fix stack_id in OTEL resource

alco:
## Summary

- **New binary memory metrics**: Add separate `process.bin_memory.*` metrics that track off-heap (refc) binary memory per process group, including total bytes, max/avg binary count, and max/avg reference count. This gives visibility into binary memory pressure that was previously invisible — process heap memory alone doesn't show refc binary accumulation.

- **Fix stack_id in OTEL resource**: Stack telemetry was missing `stack_id` in its OTEL resource attributes, making it impossible to distinguish stacks in the metrics backend. Now merged into the resource so all stack metrics carry the identifier.

- **Configurable stack telemetry init delay**: New `ELECTRIC_STACK_TELEMETRY_INIT_DELAY` env var replaces the hardcoded 30s delay, enabling faster metric emission in tests and tunable startup behavior in production.

- **Richer `Processes` module**: Generalized `top_memory_by_type` into `top_by(sort_key)` supporting sorting by either process memory or binary memory. Added `{:at_least_bytes, n}` limit variant. Each process group now returns `proc_mem`, `binary_mem`, `max_bin_count`, `avg_bin_count`, `max_ref_count`, and `avg_ref_count`.

- **Expanded integration test**: Replaced `stack-telemetry.lux` with `otel-export.lux` that verifies both application-level metrics (including the new `process.bin_memory.*` family) and stack-level metrics (including `stack_id` in the resource and LSN/storage metrics).

- **CI: Cache lux** to avoid re-fetching it on every integration test run.

## Test plan

- [x] `electric-telemetry` unit tests pass (including new tests for `top_bin_memory_by_type` and `{:at_least_bytes, n}`)
- [x] `sync-service` compiles cleanly
- [ ] `otel-export.lux` integration test via CI


packages/sync-service/Dockerfile
```diff
@@ -25,7 +25,7 @@ COPY --from=electric-telemetry / /builder/electric-telemetry
 
 COPY mix.* /builder/electric/
 RUN mix deps.get
-RUN MIX_OS_DEPS_COMPILE_PARTITION_COUNT=4 mix deps.compile
```
magnetised, line 28: curious, why remove?

magnetised (approved):
excellent stuff.

packages/sync-service/Dockerfile
alco: For some reason this causes trouble on CI runners occasionally. I can revert it just before merging.

PR #3665, opened 2026-01-02
Expose long poll configuration

balegas:
Need this to increase polling timeouts in cloud


robacourt (approved):
🚀

msfstef:
Note: this change is not relevant to cloud as cloud does not use the application config - we can already change it via the build shared opts for the router

balegas:
Thanks @msfstef that's good to know, I didn't know that. I'll close this PR and open a new one in Cloud

PR #3673, opened 2026-01-06
Fix missing headers causing infinite loop

balegas:
When a electric server/proxy returns a 200 OK response but with missing required headers (like `electric-cursor`), the client would enter an infinite retry loop if the `onError` handler returned {} to request a retry.

The issue occurred because `MissingHeadersError` is thrown AFTER the backoff wrapper returns successfully (since the HTTP request itself succeeded), so no backoff delay was applied between retries.

Since `MissingHeadersError` indicates a configuration issue  that won't fix itself without manual intervention, we now explicitly refuse to retry on this error type and log a warning instead.

Also added a test case to verify this behavior.

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -669,7 +669,8 @@ export class ShapeStream<T extends Row<unknown> = Row>
       if (this.#onError) {
         const retryOpts = await this.#onError(err as Error)
         // Guard against null (typeof null === "object" in JavaScript)
-        if (retryOpts && typeof retryOpts === `object`) {
+        const isRetryable = !(err instanceof MissingHeadersError)
+        if (retryOpts && typeof retryOpts === `object` && isRetryable) {
           // Update params/headers but don't reset offset
           // We want to continue from where we left off, not refetch everything
           if (retryOpts.params) {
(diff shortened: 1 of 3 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -665,6 +665,18 @@ export class ShapeStream<T extends Row<unknown> = Row>
     } catch (err) {
       this.#error = err
 
+      // MissingHeadersError is fatal - call onError for notification but
+      // ignore return value since retrying won't fix a configuration issue.
+      if (err instanceof MissingHeadersError) {
```
msfstef: wouldn't this make more sense as a generic guard on the `if (this.#onError)` block for whether an error is actually retryable or not? basically on line 684 also check "error is retryable" - and if it is not the rest of the code executes as normal

this looks like unnecessary code duplication (and oddly specific perhaps?)

balegas: I agree, the code was a bit silly. I've moved to having the check in place. Would you prefer to extract the pattern? I think we could do it when we have a second nonRetryableError

balegas: Or maybe, we already have the same issue with other errors, so maybe we can classify all errors appropriately.

EDIT: The only other error that can reach this part of the code is `ParserNullValueError`, I prefer not to touch that.

msfstef: I think this is fine now! if more come up we can abstract, but this seems more like the minimal work to fix it

PR #3680, opened 2026-01-06
Fix run.sh script execution

KyleAMathews:
The stack-telemetry test was failing when the Docker image otel/opentelemetry-collector-contrib needed to be pulled for the first time. Add a 120-second timeout to allow for image download before expecting the "Starting HTTP server" message.


thruflo:
No workaround to the 120 delay?

Just because I assume it slows down completion for 2mins every time.

KyleAMathews:
Yeah 120s is just the timeout. If the docker image is cached, it'll finish much quicker.

PR #3682, opened 2026-01-07
feat(sync-service): SQLite-backed metadata storage

magnetised:
Improves memory usage and actually greatly simplifies the ShapeStatus code as it removes the need for a "restore" path at all, let alone 2 distinct versions based on clean vs unclean shutdown.

The only metadata left in RAM is the LRU information, which is just `{handle, timestamp}`.

The access is via two NimblePool instances, one for reads which has a "worker" per core and a single write worker which guarantees serialized writes.

Performance locally seems reasonable, though it will be slower since - even if the data is in sqlite's page cache - we're now loading shape data from disk and deserializing using `:erlang.binary_to_term/1`.

My (hopefully) only controversial move is to add snapshot status information to the shape data so that we can initialize the system and remove any shapes that were left in a half-baked state without querying the storage backend.

I think this is cleaner that the current system, which AFAICT has issues if the system is terminated before a snapshot is started (since the snapshotting requires an active consumer and consumers are now lazily started, meaning that `await_snapshot_start/2` will crash or hang).

The implications of this are:

- shape creation is going to be a bit slower. the benchmarks below show that.
- startup is a bit slower, even with 1 shape. So something like a 50ms base init time with 1 shape.
- iterating all defined shapes is going to be slower. Locally with 150,000 shapes I'm seeing ~4 seconds to restore publication filters and ~8s to re-build the filters. This slowdown I think is purely the cost of deserialising the data from sqlite and probably the more complex memory path of disk->sqlite->nif->erlang. 
 - shutdown seems more consistently quick - not having to dump the ets tables to disk may have been the problem there
 
I've let claude review and addressed its comments, it's conclusion:

>  This is a well-structured change that simplifies the codebase significantly. The removal of the separate backup/restore paths and ETS mirror tables reduces complexity. The  SQLite implementation follows best practices for connection pooling and threading.

### DIFF
--- a/packages/sync-service/lib/electric/shapes.ex
+++ b/packages/sync-service/lib/electric/shapes.ex
@@ -112,10 +112,17 @@ defmodule Electric.Shapes do
           shape_handle()
         ) :: :ok | {:error, term()}
   def make_new_snapshot!(stream, storage, stack_id, shape_handle) do
-    with :ok <- ShapeStatus.mark_snapshot_started(stack_id, shape_handle),
-         :ok <- Storage.make_new_snapshot!(stream, storage),
+    with :ok <- Storage.make_new_snapshot!(stream, storage),
          :ok <- ShapeStatus.mark_snapshot_complete(stack_id, shape_handle) do
       :ok
+    end
+  end
+
+  @spec mark_snapshot_started(Storage.shape_storage(), stack_id(), shape_handle()) ::
+          :ok | {:error, term()}
+  def mark_snapshot_started(storage, stack_id, shape_handle) do
+    with :ok <- Storage.mark_snapshot_as_started(storage) do
+      ShapeStatus.mark_snapshot_started(stack_id, shape_handle)
     end
   end
 
--- a/packages/sync-service/lib/electric/shapes/consumer/initial_snapshot.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/initial_snapshot.ex
@@ -75,10 +75,16 @@ defmodule Electric.Shapes.Consumer.InitialSnapshot do
     %{state | pg_snapshot: snapshot, filtering?: true}
   end
 
-  def mark_snapshot_started(%__MODULE__{snapshot_started?: true} = state, _), do: state
+  def mark_snapshot_started(
+        %__MODULE__{snapshot_started?: true} = state,
+        _stack_id,
+        _shape_handle,
+        _
+      ),
+      do: state
 
-  def mark_snapshot_started(%__MODULE__{} = state, storage) do
-    Storage.mark_snapshot_as_started(storage)
+  def mark_snapshot_started(%__MODULE__{} = state, stack_id, shape_handle, storage) do
+    Electric.Shapes.mark_snapshot_started(storage, stack_id, shape_handle)
     state = reply_to_waiters(state, :started)
     %{state | snapshot_started?: true}
   end
--- a/packages/sync-service/lib/electric/shapes/consumer/state.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/state.ex
@@ -247,7 +247,12 @@ defmodule Electric.Shapes.Consumer.State do
 
   def mark_snapshot_started(%__MODULE__{initial_snapshot_state: initial_snapshot_state} = state) do
     initial_snapshot_state =
-      InitialSnapshot.mark_snapshot_started(initial_snapshot_state, state.storage)
+      InitialSnapshot.mark_snapshot_started(
+        initial_snapshot_state,
+        state.stack_id,
+        state.shape_handle,
+        state.storage
+      )
 
     %{state | initial_snapshot_state: initial_snapshot_state}
   end
--- a/packages/sync-service/test/electric/shapes/consumer/initial_snapshot_test.exs
+++ b/packages/sync-service/test/electric/shapes/consumer/initial_snapshot_test.exs
@@ -1,11 +1,13 @@
 defmodule Electric.Shapes.Consumer.InitialSnapshotTest do
   use ExUnit.Case, async: true
+  use Repatch.ExUnit, assert_expectations: true
 
   alias Electric.Shapes.Consumer.InitialSnapshot
   alias Electric.Replication.Changes.Transaction
   alias Electric.ShapeCache.Storage
 
   import Support.ComponentSetup
+  import Support.TestUtils, only: [expect_shape_status: 1]
 
   @moduletag :tmp_dir
 
@@ -161,7 +163,7 @@ defmodule Electric.Shapes.Consumer.InitialSnapshotTest do
     end
   end
 
-  describe "mark_snapshot_started/2" do
+  describe "mark_snapshot_started/4" do
     setup [:with_stack_id_from_test, :with_in_memory_storage]
 
     setup %{storage: storage} do
@@ -170,15 +172,20 @@ defmodule Electric.Shapes.Consumer.InitialSnapshotTest do
       %{shape_storage: shape_storage}
     end
 
-    test "marks snapshot as started and replies to waiters", %{shape_storage: storage} do
+    test "marks snapshot as started and replies to waiters", %{
+      shape_storage: storage,
+      stack_id: stack_id
+    } do
       parent = self()
       ref = make_ref()
+
+      expect_shape_status(mark_snapshot_started: fn ^stack_id, "shape-handle" -> :ok end)
 
       state =
         InitialSnapshot.new(nil)
         |> InitialSnapshot.add_waiter({parent, ref})
 
-      state = InitialSnapshot.mark_snapshot_started(state, storage)
+      state = InitialSnapshot.mark_snapshot_started(state, stack_id, "shape-handle", storage)
 
       assert state.snapshot_started? == true
       assert state.awaiting_snapshot_start == []
@@ -188,14 +195,18 @@ defmodule Electric.Shapes.Consumer.InitialSnapshotTest do
       assert Storage.snapshot_started?(storage)
     end
 
-    test "is idempotent when already started", %{shape_storage: storage} do
+    test "is idempotent when already started", %{shape_storage: storage, stack_id: stack_id} do
       state = InitialSnapshot.new(nil)
 
-      state = InitialSnapshot.mark_snapshot_started(state, storage)
+      expect_shape_status(
+        mark_snapshot_started: {fn ^stack_id, "shape-handle" -> :ok end, exactly: 1}
+      )
+
+      state = InitialSnapshot.mark_snapshot_started(state, stack_id, "shape-handle", storage)
       assert state.snapshot_started? == true
 
       # Call again
-      state = InitialSnapshot.mark_snapshot_started(state, storage)
+      state = InitialSnapshot.mark_snapshot_started(state, stack_id, "shape-handle", storage)
       assert state.snapshot_started? == true
     end
   end


magnetised:
benchmark this

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
```diff
@@ -0,0 +1,452 @@
...
+    SELECT handle FROM shapes ORDER BY handle
+    """,
+    handle_exists: """
+    SELECT 1 FROM shapes WHERE handle = ?1
+    """,
+    handle_lookup: """
+    SELECT handle FROM shapes WHERE comparable = ?1 LIMIT 1
+    """,
+    shape_lookup: """
+    SELECT shape FROM shapes WHERE handle = ?1 LIMIT 1
```
robacourt: Should we not make handle unique? Otherwise this returns an arbitrary shape. It would also mean you could get rid of the `LIMIT 1`. Currently we're only unique on the composite `(comparable, handle)`

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
```diff
@@ -0,0 +1,452 @@
...
+    SELECT handle, shape FROM shapes ORDER BY handle
+    """,
+    list_handles: """
+    SELECT handle FROM shapes ORDER BY handle
+    """,
+    handle_exists: """
+    SELECT 1 FROM shapes WHERE handle = ?1
+    """,
+    handle_lookup: """
+    SELECT handle FROM shapes WHERE comparable = ?1 LIMIT 1
```
robacourt: Same for comparable. This should be unique too

packages/sync-service/lib/electric/shapes.ex
```diff
@@ -100,6 +101,24 @@ defmodule Electric.Shapes do
...
+          stack_id(),
+          shape_handle()
+        ) :: :ok | {:error, term()}
+  def make_new_snapshot!(stream, storage, stack_id, shape_handle) do
+    with :ok <- ShapeStatus.mark_snapshot_started(stack_id, shape_handle),
+         :ok <- Storage.make_new_snapshot!(stream, storage),
+         :ok <- ShapeStatus.mark_snapshot_complete(stack_id, shape_handle) do
+      :ok
+    end
+  end
```
robacourt, line 119: Before snapshots were only marked as started once the metadata was written to disk. Now they're marked as started before that metadata has been written, so when `Shapes.get_merged_log_stream/3` is called, it calls:

```elixir
      with :started <- ShapeCache.await_snapshot_start(shape_handle, stack_id) do
        storage = shape_storage(stack_id, shape_handle)
        {:ok, Storage.get_log_stream(offset, max_offset, storage)}
      end
```

and `Storage.get_log_stream/3` could fail because the metadata is missing

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db.ex
```diff
@@ -1,193 +1,287 @@
...
+                      modify(conn, insert_shape, [
+                        {:blob, shape_handle},
+                        {:blob, term_to_binary(shape)},
+                        {:blob, term_to_binary(comparable_shape)},
+                        shape_hash
+                      ]),
+                    {:ok, 1} <- modify(conn, increment_counter, [1]),
+                    :ok <-
+                      Enum.reduce_while(Shape.list_relations(shape), :ok, fn {oid, _name}, :ok ->
+                        case modify(conn, insert_relation, [{:blob, shape_handle}, oid]) do
```
robacourt: minor nit: it took me a while to work out how the db was being called, it feels like there's lots of levels of indirection. I get that you only want to prepare statements once per db. One idea would be to have a `ShapeDB.Queries` or `ShapeDB.Adapter` module that had the statements defined in it and functions such as:

```elixir
def insert_relation(conn, shape_handle, oid) do
    modify(conn, conn.insert_relation, [{:blob, shape_handle}, oid])
end
```

Which admittedly is another level of indirection but seems easier to follow to me.

robacourt (commented):
Very nice! 🚀  I've left some comments

balegas:
Codex review. Curious to know how did it understand things. I primed it latest 100 commits and an analysis on perf improvements we've done.

The PR **moves ShapeStatus metadata to SQLite** and keeps only the LRU map in ETS. It also introduces snapshot-state tracking in the shape metadata (0/1/2) and uses that to **remove “invalid” shapes at startup**. The infrastructure consists of a migrator and two NimblePool pools (read pool sized to schedulers; single-writer pool) with prepared statements and WAL mode. The new snapshot state is set via a wrapper around snapshot writes in `Electric.Shapes.make_new_snapshot!/4`.

------

## Potential issues / risks (bugs, races, invariants)

### 1) **Startup cleanup may delete valid shapes that never started snapshotting**

**What happens:** `ShapeStatus.initialize/1` calls `ShapeDb.validate_existing_shapes/1`, which deletes shapes whose `snapshot_state` is **0 or 1** (not started or started but not completed). It then **removes their storage asynchronously**.

**Why this is risky:**

- Snapshot state **0 (“not started”)** can occur for shapes that were created but whose snapshot hadn’t started yet (e.g., because consumers are lazily started). On restart, those shapes get deleted even though they may be valid and still needed.
- If a client had already received a handle, deleting the shape metadata and storage could cause “missing shape” errors or force handle re-creation (breaking handle stability).
- This is especially risky if shape creation can succeed and return a handle before the snapshot starts.

**Recommendation:** Consider narrowing “invalid” to **only `snapshot_state = 1`** (started-but-not-complete) and keep `0` unless you can guarantee no legitimate shapes exist in state 0 at shutdown. If the intent is to delete state-0 shapes, I’d want an explicit invariant documenting that a shape handle is never committed before the snapshot starts.

------

### 2) **`remove_shape_in_transaction/2` assumes relation rows always exist**

**What happens:** The removal flow deletes the shape row, decrements the count, and then deletes relations. It **expects** `DELETE FROM relations` to return `n > 0`, otherwise the `with` clause fails and returns `{:ok, 0}` instead of `:ok`. That return value can **short-circuit downstream cleanup** (because callers only accept `:ok`).

**Why this is risky:**

- If the relations table is missing entries (e.g., partial corruption, earlier failure, or a future shape type that doesn’t carry relations), removal will leave the shape partially deleted and skip storage cleanup.
- The current code path treats `{:ok, 0}` as a non-`:ok` result and fails `with :ok <- ...` in `ShapeStatus.remove_shape/2`, which bubbles as an error to callers and can skip cleanup work.

**Recommendation:** Make the relations delete **best-effort** (allow `n >= 0`) or explicitly handle `{:ok, 0}` as success. If relations are strictly required, return a clearer error and ensure callers clean up.

------

### 3) **Error propagation changes for shape lookup APIs**

**What happens:** `fetch_handle_by_shape/2` and `fetch_shape_by_handle/2` now return **whatever** `ShapeDb.handle_for_shape/2` or `ShapeDb.shape_for_handle/2` returns. Those functions return `{:ok, value}` or `:error`, but **can also return `{:error, reason}`** on SQLite failures (e.g., from `fetch_one/3`).

**Why this is risky:**

- Callers that pattern-match `:error` may now receive `{:error, reason}` and crash or treat it as success (depending on usage).
- The module specs still advertise `{:ok, shape_handle()} | :error` and `{:ok, Shape.t()} | :error`, so returning `{:error, reason}` violates the spec.

**Recommendation:** Normalize errors to `:error` (or update specs and callers) to keep behavior compatible. The simplest fix is to wrap `ShapeDb` results inside `ShapeStatus` and map `{:error, _}` to `:error`.

------

### 4) **Snapshot state transitions + failure behavior**

**What happens:** Snapshot creation now wraps `Storage.make_new_snapshot!/2` with `mark_snapshot_started` and `mark_snapshot_complete`. If snapshotting fails after `mark_snapshot_started`, the shape is left in state **1**, and on restart it is treated as “invalid” and deleted.

**Why this is risky:**

- This can cause **data loss** if a partial snapshot could have been resumed or retried in place.
- If transient snapshot failures occur (e.g., temporary DB issue), the system will wipe shape storage on next restart.

**Recommendation:** If deletion is intentional, document that **snapshot failures are destructive**. If not, consider a retry strategy or a grace period before eviction.

------

### 5) **`handle_exists?/2` can raise on SQLite errors**

**What happens:** `ShapeDb.handle_exists?/2` uses a `case` that only handles `{:ok, [1]}` and `:error`. If `fetch_one` returns `{:error, reason}`, it will raise a `CaseClauseError`. That can cascade into `ShapeStatus.has_shape_handle?/2` callers.

**Recommendation:** Handle `{:error, reason}` explicitly (e.g., log and return `false` or bubble a tagged error).

------

## Positive structural changes worth keeping

- **Reduced RAM usage** by eliminating large ETS tables and moving shape metadata to SQLite with WAL.
- **Explicit initialization path** that doesn’t depend on storage iteration on unclean shutdown.
- **Read/write pool separation** with a single writer to avoid SQLite busy errors.
- **Snapshot-state instrumentation** is a good way to reason about half-built shapes, as long as the invalidation logic is carefully scoped.
  These improvements are solid and align with the stated goals of memory reduction and simpler restore semantics.

------

## Summary recommendation

I’d approve the direction, but I’d **require fixes or documentation** around:

1. **`snapshot_state=0` deletion on startup** (risk of deleting valid shapes).
2. **Removal logic assuming relations exist** (returns non-`:ok`).
3. **Error normalization for lookup APIs** and `handle_exists?/2` to avoid crashes or spec violations.
4. **Explicit policy for snapshot failures** (destructive cleanup vs. recovery).

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
magnetised: good point. I've updated the indexes to add a unique constraint explictly on the handles and comparables. I've kept the `LIMIT 1` because I think it helps to clarify the expectations of the query - i.e. it acts as a separator between those queries that list things and those queries that fetch single rows

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db.ex
magnetised: I implemented this as you suggest which I hope will help with the clarity.

magnetised:
benchmark this

packages/sync-service/lib/electric/shapes.ex
magnetised: Good point. I've moved the update to shape status into the consumer at the point where it updates the storage metadata. This seems simplest and safest, if not the prettiest. I experimented with starting a new task process and using `ShapeCache.await_snapshot_start` but it meant even more processes and faff even if it improved encapsulation.

magnetised:
@balegas 

> I’d approve the direction, but I’d require fixes or documentation around:
>
>    snapshot_state=0 deletion on startup (risk of deleting valid shapes).

We don't really care about deleting a shape that hasn't got a snapshot this idea of a "valid shape" is a red herring

>    Removal logic assuming relations exist (returns non-:ok).
>    Error normalization for lookup APIs and handle_exists?/2 to avoid crashes or spec violations.

I think a crash at this point is the honest response, hiding an error behind a `false` is really dangerous. 

The comments about shapes without relations are just wrong. Shapes ATM always have 1 relation (though it's coded as if they could have more).

>    Explicit policy for snapshot failures (destructive cleanup vs. recovery).

Not sure what this means. Snapshot failures are already handled in other parts of the stack. The cleanup phase is to handle snapshots didn't complete because of e.g. a system crash.

PR #3692, opened 2026-01-10
Use object as `subscriptionId` in clients

tjenkinson:
This is such a small thing I spotted when looking into how things work I wondered if it was even worth opening. There's a _tiny_ chance of a collision with `Math.random()`, when that could be removed by using an object here, or maybe symbol/counter even 🤷

A while back we did discover some smart tv's which unbelievably returned a static number from `Math.random()` 😅


kevin-dp:
@tjenkinson indeed, we could also use a monotonically increasing counter.

PR #3696, opened 2026-01-12
Alco/elixir 1.20

alco:


alco:
superseded by github.com/electric-sql/electric/pull/3992

PR #3701, opened 2026-01-13
Handle stale cached responses with expired shape handles

KyleAMathews:
When a CDN/proxy is misconfigured and serves a stale cached response with an expired shape handle, the client would get into a broken state where:
1. The handle was rejected (matched expired cache entry)
2. But the offset was still advanced
3. Subsequent requests failed with "handle can't be blank when offset != -1"

This fix detects when a stale response is received (handle matches expired entry AND client has no handle yet) and triggers a retry with a random cache buster parameter (_cb) to bypass the misconfigured CDN cache.

Changes:
- Add StaleCacheError to signal need for cache-busted retry
- Add #staleCacheBuster field to store random cache buster
- Add CACHE_BUSTER_QUERY_PARAM constant for the _cb parameter
- Throw StaleCacheError in #onInitialResponse when detecting stale response
- Catch StaleCacheError in #requestShape and retry with cache buster
- Add _cb param to URL construction when set
- Add test for the stale cache retry scenario

## Summary by CodeRabbit

* **New Features**
  * Added cache-busting mechanism that automatically detects and recovers from stale CDN responses through intelligent retry logic with configurable retry limits.
  * New dedicated error type for stale cache scenarios.

* **Tests**
  * Added test coverage validating cache-busting retry behavior and error handling when maximum retries are exceeded.

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -1086,6 +1086,7 @@ export class ShapeStream<T extends Row<unknown> = Row>
             `This should not happen and indicates a proxy/CDN caching misconfiguration. ` +
             `The response contained handle "${shapeHandle}" which was previously marked as expired. ` +
             `Check that your proxy includes all query parameters (especially 'handle' and 'offset') in its cache key. ` +
+            `For more information visit the troubleshooting guide: https://electric-sql.com/docs/guides/troubleshooting ` +
             `Retrying with a random cache buster to bypass the stale cache (attempt ${this.#staleCacheRetryCount}/${this.#maxStaleCacheRetries}).`
         )
         // Generate a random cache buster for the retry
--- a/packages/typescript-client/src/constants.ts
+++ b/packages/typescript-client/src/constants.ts
@@ -28,7 +28,7 @@ export const SUBSET_PARAM_ORDER_BY = `subset__order_by`
 export const SUBSET_PARAM_WHERE_PARAMS = `subset__params`
 export const SUBSET_PARAM_WHERE_EXPR = `subset__where_expr`
 export const SUBSET_PARAM_ORDER_BY_EXPR = `subset__order_by_expr`
-export const CACHE_BUSTER_QUERY_PARAM = `_cb` // Random cache buster to bypass stale CDN responses
+export const CACHE_BUSTER_QUERY_PARAM = `cache-buster` // Random cache buster to bypass stale CDN responses
 
 // Query parameters that should be passed through when proxying Electric requests
 export const ELECTRIC_PROTOCOL_QUERY_PARAMS: Array<string> = [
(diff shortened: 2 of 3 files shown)


packages/typescript-client/src/client.ts
```diff
@@ -1031,6 +1056,45 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+              `Check that your proxy includes all query parameters (especially 'handle' and 'offset') in its cache key. ` +
+              `For more information visit the troubleshooting guide: https://electric-sql.com/docs/guides/troubleshooting`
+          )
+        }
+        console.warn(
+          `[Electric] Received stale cached response with expired shape handle. ` +
+            `This should not happen and indicates a proxy/CDN caching misconfiguration. ` +
+            `The response contained handle "${shapeHandle}" which was previously marked as expired. ` +
+            `Check that your proxy includes all query parameters (especially 'handle' and 'offset') in its cache key. ` +
+            `Retrying with a random cache buster to bypass the stale cache (attempt ${this.#staleCacheRetryCount}/${this.#maxStaleCacheRetries}).`
```
kevin-dp, line 1090: Perhaps also add the link to the troubleshooting docs in this warning.

packages/typescript-client/src/constants.ts
```diff
@@ -28,6 +28,7 @@ export const SUBSET_PARAM_ORDER_BY = `subset__order_by`
 export const SUBSET_PARAM_WHERE_PARAMS = `subset__params`
 export const SUBSET_PARAM_WHERE_EXPR = `subset__where_expr`
 export const SUBSET_PARAM_ORDER_BY_EXPR = `subset__order_by_expr`
+export const CACHE_BUSTER_QUERY_PARAM = `_cb` // Random cache buster to bypass stale CDN responses
```
kevin-dp: The actual name of this parameter `_cb`  is cryptic and not consistent wrt the other ones that are very explicit. Would be better to use something more descriptive like `cache-buster`.

packages/typescript-client/test/expired-shapes-cache.test.ts
```diff
@@ -364,4 +364,162 @@ describe(`ExpiredShapesCache`, () => {
...
+    // Small delay to let the response be processed
+    await new Promise((resolve) => setTimeout(resolve, 10))
+
+    // The client should have retried and gotten a fresh handle
+    expect(stream.shapeHandle).toBe(freshHandle)
+
+    // Verify the second request includes the cache buster parameter
+    expect(requestCount).toBeGreaterThanOrEqual(2)
+    const secondUrl = new URL(capturedUrls[1])
+    expect(secondUrl.searchParams.has(`_cb`)).toBe(true)
```
kevin-dp: Shouldn't hardcode `_cb` but use `CACHE_BUSTER_QUERY_PARAM`.

KyleAMathews:
@kevin-dp Good question about the Elixir client. I investigated and found it has a similar vulnerability but behaves differently:

**Elixir currently**: Accepts the stale handle → continues → server eventually rejects with 409 → resets (reactive recovery)

**TypeScript (this PR)**: Detects stale handle → retries with cache-buster → avoids the problem (proactive prevention)

The Elixir client lacks:
- Expired handles tracking
- Stale response detection
- Cache-buster retry logic

Created #3709 to track adding similar protection to the Elixir client.

PR #3703, opened 2026-01-13
Fix CI flake

KyleAMathews:
The SQLite-backed metadata storage (cb9f571) introduces initialization overhead that can cause a race condition in tests that check health status immediately after seeing "Starting replication" log.

The shape_log_collector_ready condition is set asynchronously in ShapeCache's handle_continue(:wait_for_restore), which now takes longer due to SQLite operations. This causes the health check to return "starting" instead of "active" if checked too quickly.

Fix by using the wait-for macro to poll for active status, allowing up to 10 seconds for shape initialization to complete.

### DIFF
--- a/integration-tests/tests/rolling-deploy.lux
+++ b/integration-tests/tests/rolling-deploy.lux
@@ -120,8 +120,8 @@
   ??[info] Starting replication from postgres
 
 [shell orchestator]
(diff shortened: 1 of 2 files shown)


magnetised (approved):
Fab, thanks

PR #3706, opened 2026-01-14
AI-friendly docs

balegas:


balegas:
I thinks it's a good idea to move this under internals, keep old agents and just leave a note where to look if the agent is looking for information about how electric is built

PR #3713, opened 2026-01-14
website: fix image previews.

thruflo:
This PR does four things:

0. it fixes the homepage title
1. it fixes website previews not showing on social media
2. it adds a couple of explicit Twitter meta tags
3. it vendors in the Image component from @unpic/vue to remove it as a dip to avoid a vue version clash.

It does not change the use of the netlify images service for the actual meta images. (I've reverted that change).

Previously the markdown serving edge function was serving twitter bots etc. with markdown; this was stopping them getting the meta tags in html they were expecting, thus our website pages and posts were not being expanded on Twitter etc.

The fix to this is slightly fragile, in that it's a whitelist of well known social media user agent strings. Other options were considered but they would compromise the optimization for tending to try to serve markdown to coding agents when at all possible.

## Summary by CodeRabbit

* **New Features**
  * Richer social preview metadata and a centralized meta-image URL builder for consistent sharing.
  * Netlify-powered responsive image component for optimized image delivery.
  * Server-side detection of social media bots to improve request handling.

* **Tests**
  * Added comprehensive unit tests and test configuration for image and bot-detection utilities.

* **Chores**
  * Added test scripts and dev tooling; updated Netlify dev config.
  * Removed legacy image library dependency.

### DIFF
--- a/website/.vitepress/config.mts
+++ b/website/.vitepress/config.mts
@@ -390,7 +390,9 @@ export default defineConfig({
     const fm = pageData.frontmatter
     const head = []
 
-    const title = `${fm.title || siteData.title} | ${fm.titleTemplate || 'ElectricSQL'}`
+    const pageTitle = fm.title || siteData.title
+    const titleTemplate = fm.titleTemplate || ':title | ElectricSQL'
+    const title = titleTemplate.replace(':title', pageTitle)
     const description = fm.description || siteData.description
 
     const PRODUCTION_URL = 'https://electric-sql.com'


thruflo:
Ok, it seems the issue is with Twitter, I can';t get any preview images to work. I suspect this PR may still be helpful though -- we don't need to route meta images through netlify and removing @unpic/vue fixes the website build.

KyleAMathews:
I'm confused — what's wrong with unpic? Why not route through Netlify for the social pic? It does the cropping/resizing automatically.

thruflo:
It has a direct dependency on vue which is incompatible with VueTweet. We can fix an override vue version but only in the workspace, which means it would be set for all examples etc.

It felt simpler to just vendor the actual unpic code we’re using and remove the dep.

On the meta image for Netlify, yes, I removed it in case somehow the Netlify image serving was what was breaking Twitter previews. I’ve realised that that wasn’t the problem, so it’s fine to use a Netlify image for the meta image. I’ll revert.

The real issue I’ve just finally got to the bottom of is that the Twitter bot is getting the markdown version of the pages. Hence no preview.

KyleAMathews (approved):

PR #3716, opened 2026-01-15
Move SQLite writes to a background thread and buffer read state

magnetised:
Reduce load on sqlite by moving the core metadata checked by most read requests into ETS and prevent timeouts during sustained parallel shape creation by writing to sqlite asynchronously in a background process. Writes are now done directly to ETS and flushed to disk in batches.

Only shape -> handle and handle -> shapes lookups are now handled by reading directly from sqlite. All other, small, metadata is held in an ets wriite-through cache and populated from sqlite on boot. 

Pending writes are flushed to disk on shutdown. This is quick, ~1,000 writes/s, but could cause delays to shutdown in an overloaded system. There is a risk of loss of in-flight writes if the system crashes.

Associated changes:

- Move ShapeCleaner up in the supervisor hierarchy so it's available before the ShapeStatus process attempts to cleanup incomplete/invalid shapes
- Allow for configuring load shedding via an environment variable - makes it easier to run load tests locally without maintaining a patch on the default configuration

## Summary by CodeRabbit

* **New Features**
  * SQLite-backed shape store with an ETS-backed write buffer, background flush controls (pause/resume), labeled DB checkout telemetry, pending-buffer telemetry, and optional max-concurrent-requests config.
* **Bug Fixes**
  * Reduced races between buffer and persistent store; more consistent snapshot lifecycle, LRU/meta handling, and removals.
* **Chores**
  * Supervision changes: added cleanup-task supervisor and removed legacy cleaner wiring.
* **Tests**
  * Expanded unit and integration tests for buffering, snapshot flows, crash recovery, and ordering.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status/shape_db.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status/shape_db.ex
@@ -82,6 +82,19 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb do
   Returns :error if the handle is tombstoned (being deleted).
   """
   def handle_for_shape(stack_id, %Shape{} = shape) when is_stack_id(stack_id) do
+    checkout_fun = &checkout!(stack_id, :handle_for_shape, &1)
+    handle_for_shape_inner(stack_id, shape, checkout_fun)
+  end
+
+  @doc """
+  Find a handle for a shape using the write connection to guarantee consistency.
+  """
+  def handle_for_shape_critical(stack_id, %Shape{} = shape) when is_stack_id(stack_id) do
+    checkout_fun = &checkout_write!(stack_id, :handle_for_shape_critical, &1)
+    handle_for_shape_inner(stack_id, shape, checkout_fun)
+  end
+
+  defp handle_for_shape_inner(stack_id, %Shape{} = shape, checkout_fun) do
     {comparable_shape, _shape_hash} = Shape.comparable_hash(shape)
 
     case WriteBuffer.lookup_handle(stack_id, comparable_shape) do
@@ -89,7 +102,7 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb do
         {:ok, handle}
 
       :not_found ->
-        case checkout!(stack_id, :handle_for_shape, &Query.handle_for_shape(&1, comparable_shape)) do
+        case checkout_fun.(&Query.handle_for_shape(&1, comparable_shape)) do
           {:ok, handle} ->
             if WriteBuffer.is_tombstoned?(stack_id, handle) do
               :error
--- a/packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
@@ -53,7 +53,9 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.WriteBuffer do
   import Electric, only: [is_stack_id: 1]
 
   @poll_interval 50
-  @max_drain_per_cycle 1000
+  # keep this low-ish so that this process yields the write connection
+  # to handle_for_shape_critical/2 reasonably often.
+  @max_drain_per_cycle 100
 
   def operations_table_name(stack_id), do: :"write_buffer_ops:#{stack_id}"
   def shapes_table_name(stack_id), do: :"write_buffer_shapes:#{stack_id}"
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -244,7 +244,9 @@ defmodule Electric.ShapeCache do
   end
 
   defp maybe_create_shape(shape, otel_ctx, %{stack_id: stack_id} = state) do
-    with {:ok, shape_handle} <- ShapeStatus.fetch_handle_by_shape(stack_id, shape),
+    # fetch_handle_by_shape_critical is a slower but guaranteed consistent
+    # shape lookup
+    with {:ok, shape_handle} <- ShapeStatus.fetch_handle_by_shape_critical(stack_id, shape),
          {:ok, offset} <- fetch_latest_offset(stack_id, shape_handle) do
       {shape_handle, offset}
     else
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -68,13 +68,14 @@ defmodule Electric.ShapeCache.ShapeStatus do
       # Add the lookup last as it is the one that enables clients to find the shape
       with {:ok, shape_hash} <- ShapeDb.add_shape(stack_id, shape, shape_handle) do
         # Cache shape metadata: {handle, hash, snapshot_started, last_read_time}
-        true =
-          :ets.insert_new(
-            shape_meta_table(stack_id),
-            {shape_handle, shape_hash, false, System.monotonic_time()}
-          )
-
-        {:ok, shape_handle}
+        if :ets.insert_new(
+             shape_meta_table(stack_id),
+             {shape_handle, shape_hash, false, System.monotonic_time()}
+           ) do
+          {:ok, shape_handle}
+        else
+          {:error, "duplicate shape #{inspect(shape_handle)}: #{inspect(shape)}"}
+        end
       end
     end)
   end
@@ -143,6 +144,24 @@ defmodule Electric.ShapeCache.ShapeStatus do
   def fetch_handle_by_shape(stack_id, %Shape{} = shape) when is_stack_id(stack_id) do
     OpenTelemetry.with_span("shape_status.fetch_handle_by_shape", [], stack_id, fn ->
       ShapeDb.handle_for_shape(stack_id, shape)
+    end)
+  end
+
+  @doc """
+  Where as `fetch_handle_by_shape/2` *may* under high-write load return stale
+  data -- not finding a shape that has been written -- due to SQLite's
+  cross-connection durability when in WAL mode, where a connection reads
+  against a snapshot of the data, this version does the lookup via the write
+  connection, which is guaranteed to see all writes (SQLite connections can
+  always see their own writes).
+
+  This guarantees that will will return consistent restults at the cost of slower
+  lookups as we're contending with access to the single write connection.
+  """
+  @spec fetch_handle_by_shape_critical(stack_id(), Shape.t()) :: {:ok, shape_handle()} | :error
+  def fetch_handle_by_shape_critical(stack_id, %Shape{} = shape) when is_stack_id(stack_id) do
+    OpenTelemetry.with_span("shape_status.fetch_handle_by_shape_critical", [], stack_id, fn ->
+      ShapeDb.handle_for_shape_critical(stack_id, shape)
     end)
   end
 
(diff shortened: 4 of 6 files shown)


magnetised:
@coderabbitai review

magnetised:
benchmark this

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
```diff
@@ -105,24 +103,24 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.Connection do
...
 
-      :checkout ->
-        :ok
-    end
+    {:ok, conn, conn, pool_state}
+  end
 
+  def handle_checkout({:checkout, _label}, _from, conn, pool_state) do
     {:ok, conn, conn, pool_state}
   end
```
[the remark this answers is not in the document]
magnetised: That function is reached when connections can be checked out without being queued.

balegas:
Concurrent shape creation looks significantly slower. Do you think this might cause timeouts with high load?

magnetised:
benchmark this

magnetised:
benchmark this

magnetised:
@coderabbitai review

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
```diff
@@ -0,0 +1,434 @@
...
+
+    # Queue operation for SQLite (contains full data including hash and relations)
+    true =
+      :ets.insert(
+        ops_table,
+        {ts, {:add, handle, shape_binary, comparable_binary, hash, relations}, false}
+      )
+
+    :ok
+  end
```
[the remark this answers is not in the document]
magnetised: handles are unique - they will never be removed and then re-added

integration-tests/tests/crash-recovery.lux
```diff
@@ -47,6 +47,9 @@
 
 ## Crash electric
 [shell electric]
+  # flush pending writes so that the shape we just created is definitely in the db
+  !Electric.ShapeCache.ShapeStatus.ShapeDb.WriteBuffer.flush_sync("single_stack")
```
msfstef: If we are first flushing the changes then we are not simulating crash recovery - the test should cover the behaviour of a crash, including unflushed write buffer changes, and if that breaks electric we should fix it no?

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
```diff
@@ -0,0 +1,434 @@
...
+  def pending_operations_count(stack_id) do
+    :ets.info(operations_table_name(stack_id), :size)
+  end
+
+  @doc "Returns all buffered shapes as a list of {handle, shape} tuples, excluding tombstoned handles"
+  def list_buffered_shapes(stack_id) do
+    tombstones = tombstones_table_name(stack_id)
+
+    shapes_table_name(stack_id)
+    |> :ets.tab2list()
```
msfstef: I would rather avoid tab2list where possible and instead do a reduce/fold - less memory pressure and scale independent?

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
```diff
@@ -0,0 +1,434 @@
...
+      {handle, :erlang.binary_to_term(shape_binary)}
+    end)
+  end
+
+  @doc "Returns the count of buffered shapes (excluding tombstoned)"
+  def buffered_shape_count(stack_id) do
+    tombstones = tombstones_table_name(stack_id)
+
+    shapes_table_name(stack_id)
+    |> :ets.tab2list()
```
msfstef: same as above - prefer reduce over tab2list

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
```diff
@@ -0,0 +1,434 @@
...
+
+    tombstones_table_name(stack_id)
+    |> :ets.tab2list()
+    |> Enum.count(fn {handle, _ts} ->
+      not has_buffered_add?(ops_table, handle)
+    end)
+  end
+
+  defp has_buffered_add?(ops_table, handle) do
+    :ets.match_object(ops_table, {:_, {:add, handle, :_, :_, :_, :_}, :_}) != []
```
msfstef: also full table scan (per-tombstone) - this is for telemetry only and again the buffer should be small but it is a bit scary to do a full scan per tombstone entry, surely there's more efficient ways of doing this (since we're doing tab2list again on the tombstones anyway) - we dont need to traverse the whole table every time. Having some sort of index might be worth it?

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
```diff
@@ -0,0 +1,434 @@
...
+        [] ->
+          :not_found
+      end
+    end
+  end
+
+  @doc "Look up a handle by comparable shape binary. Returns {:ok, handle} or :not_found."
+  def lookup_handle(stack_id, comparable_binary) do
+    tombstones = tombstones_table_name(stack_id)
+    pattern = {:_, :_, comparable_binary}
```
msfstef: to clarify, this is a full table scan - I guess the idea is that the buffer should be small and we can afford this, but if write intensity is high won't this be an issue? and it's on the critical path as well

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
```diff
@@ -0,0 +1,434 @@
...
+    tombstones_table_name(stack_id)
+    |> :ets.tab2list()
+    |> Enum.map(fn {handle, _ts} -> handle end)
+    |> MapSet.new()
+  end
+
+  # Write functions
+
+  @doc "Add a shape to the buffer"
+  def add_shape(stack_id, handle, shape_binary, comparable_binary, hash, relations) do
```
msfstef: To clarify - there is no limit to the size of this buffer, where it would force a flush and apply back pressure forcefully, right?

msfstef (commented):
Will give it a more detailed look tomorrow, main comments are 1) many full table scans and tab2lists on the buffer but its not guaranteed to be small (?) 2) crash test now doesn't actually test a crash (?) how do we handle the inflight requests being lost

will leave opus review as well below this

packages/sync-service/test/electric/replication/shape_log_collector/request_batcher_test.exs
```diff
@@ -115,6 +115,8 @@ defmodule Electric.Replication.ShapeLogCollector.RequestBatcherTest do
         end)
 
       assert_receive :adding_shape_2
+      # Small delay to ensure the add_shape GenServer.call is processed before remove
+      Process.sleep(5)
```
msfstef: why not wait for the `:processor_called` message here as well? after that we wait for the `processor_delay` set above so the deduplication should work fine - generally we patch the processor call entirely so I'm unclear on why we need a sleep here

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
magnetised: I've added a direct comparable -> handle lookup entry

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
magnetised: i used a match rather than a reduce in order to only extract the relevant entries in the buffered ops log, rather than reducing over the entire table

integration-tests/tests/crash-recovery.lux
magnetised: yeah, my bad. fixed.

integration-tests/tests/crash-recovery.lux
```diff
@@ -53,18 +53,22 @@
 ## Start the sync service again.
 [invoke setup_electric]
 [shell electric]
+  ??[info] Found 0 existing valid shapes and 1 shapes in an invalid state
   ??[info] Starting replication from postgres
 
 # Client should be able to continue same with old offset or 0_inf offset
```
msfstef, line 98: this looks better - would be good in this test to enhance it to actually have two shapes, one that is actually resumable (we wait for it to flush) and one that will be invalid, so both paths are tested

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
magnetised: there is no backpressure - we already have the load shedding. worst case we temporarily get similar memory usage to the previous pure-ets implementation... I think it's worth exploring the behaviour as-is rather than complicate it more.

msfstef: Makes sense!

integration-tests/tests/crash-recovery.lux
magnetised: done. i also added a "pause flush" function so we are not at the mercy of timing

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db.ex
```diff
@@ -90,71 +192,54 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb do
...
-    checkout!(stack_id, fn %Connection{} = conn ->
-      Query.count_shapes(conn)
-    end)
+    buffered_count = WriteBuffer.buffered_shape_count(stack_id)
+    sqlite_tombstone_count = WriteBuffer.sqlite_tombstone_count(stack_id)
+
+    case checkout!(stack_id, :count_shapes, &Query.count_shapes/1) do
+      {:ok, sqlite_count} -> {:ok, sqlite_count - sqlite_tombstone_count + buffered_count}
+      error -> error
+    end
```
msfstef, line 221: I might be overly cautious I'm just wondering if all of these operations are too expensive for the frequency we count our shapes at, and if they are worth the extra accuracy - it feels like we should have a cheap heuristic way to have a fairly accurate but approximate count of the shapes without scanning the table etc

If you don't believe this is in any way going to be a problem on performance with telemetry enabled with high load ignore this

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
```diff
@@ -0,0 +1,445 @@
...
+  ## Operations table format
+
+  `{{monotonic_time, unique_int}, operation, flushing}`
+
+  ## Shapes table key formats
+
+  - `{{:shape, handle}, shape_binary, comparable_binary}` - shape data
+  - `{{:comparable, comparable_binary}, handle}` - reverse index for O(1) lookup
+  - `{{:tombstone, handle}, timestamp}` - handles marked for deletion
+  """
```
msfstef, line 43: I think it would be really good to document here what happens on a crash/discarded in-flight requests and why it is a safe failure mode (no orphaned shapes, things will recover, etc)

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db.ex
magnetised: I don't think it'll be a problem to run this every 30s but I can add a counter to assuage doubts - we don't need it to be 100% accurate

magnetised:
benchmark this

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
```diff
@@ -9,7 +9,7 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.Connection do
 
   @behaviour NimblePool
 
-  @schema_version 2
+  @schema_version 3
```
msfstef, line 12: I assume this blows up storage right?

magnetised: yes.

PR #3718, opened 2026-01-15
fix: Race condition on shape removal

msfstef:
## Summary

Fixes a race condition in shape removal that crashes `ShapeLogCollector` with `ArgumentError: invalid destination`.

### The Race

During shape removal, operations happen in this order:
1. `ShapeStatus.remove_shape` - **immediate**
2. `Consumer.stop` - **immediate**  
3. `Storage.cleanup!` - **immediate**
4. `ShapeLogCollector.remove_shape` - **async** (batched via `RequestBatcher`)

If a transaction arrives between steps 1-3 and step 4:
- EventRouter still has the shape (removal pending)
- `ConsumerRegistry.publish` tries to start a consumer
- `start_consumer_for_handle` returns `{:error, :no_shape}` (shape gone from status)
- `start_consumer!` returns `nil`
- `broadcast()` calls `send(nil, msg)` → **crash**

### The Fix

Filter out nil pids in `broadcast/1` before attempting to send messages. This is safe because `start_consumer!` only returns `nil` when the shape has been removed.

```elixir
handle_event_pids
|> Enum.reject(fn {_handle, _event, pid} -> is_nil(pid) end)
|> Enum.map(fn {handle, event, pid} -> ... end)
```

---

**@magnetised** - The comment at `shape_cache.ex:226-230` mentions this race and claims ConsumerRegistry "excludes the shape handle from the broadcast", but that wasn't actually happening. This fix implements what the comment describes. 

Questions:
1. Is there any scenario where `start_consumer!` returning `nil` should *not* be silently skipped?
2. Should we also consider reordering the removal sequence (EventRouter first) as a more robust fix?

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition crash that occurred in the message broadcast system when a shape was removed during active transaction processing. The service now gracefully handles this edge case without crashing.

* **Tests**
  * Added test coverage for edge cases where process identifiers may be nil during concurrent broadcast operations, preventing similar race condition issues in the future.

### DIFF
--- a/.changeset/fix-shape-removal-race.md
+++ b/.changeset/fix-shape-removal-race.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Fix race condition crash in ConsumerRegistry when shape is removed during transaction processing


msfstef:
> reordering the operations would probably just lead to different race conditions...

Opus had suggested 3 fixes, including the reordering, but the reordering also would require the removal from SLC to be synchronous - which I really don't think it should be

PR #3719, opened 2026-01-15
fix: Race condition for reader after shape writer flushes data

msfstef:
## Summary

Fixes a race condition in `PureFileStorage` where readers can get empty or partial results when reading from ETS while a concurrent flush clears the buffer.

## The Bug

When a reader calls `get_log_stream`, it reads metadata (including `last_persisted_offset`) as a snapshot, then uses that to decide whether to read from disk or ETS. However, a concurrent flush can happen between these steps:

```
T1: Reader reads metadata: last_persisted=X, ets_ref=table
T2: Writer flushes: writes to disk, updates last_persisted=Y, clears ETS
T3: Reader decides to read from ETS (based on stale last_persisted=X)
    → ETS is empty or partially read → returns [] or incomplete data
```

The data exists on disk but the reader misses it because it made its read decision based on stale metadata.

### Two failure modes

1. **Empty ETS**: Flush completes entirely before reader's first `next_lookup` call → reader immediately gets `'$end_of_table'` and returns `[]`

2. **Partial ETS read**: Reader iterates with multiple `next_lookup` calls, and `delete_all_objects` runs between calls:
   ```
   next_lookup(ets, start) → {key1, value1}
   next_lookup(ets, key1)  → {key2, value2}
   ← delete_all_objects runs here (atomic) →
   next_lookup(ets, key2)  → '$end_of_table'
   ```
   Reader got entries 1-2 but missed entries 3-N.

**Note on ETS guarantees:** The buffer is an `ordered_set` table. Per [Erlang ETS docs](https://www.erlang.org/doc/apps/stdlib/ets.html), `delete_all_objects` is atomic and isolated—each individual `next_lookup` call sees either the full table or the empty table, never a mid-deletion state. However, a *sequence* of `next_lookup` calls can span the deletion, resulting in partial data. For `ordered_set` tables, `next_lookup` will not crash during concurrent deletion; it simply returns `'$end_of_table'` when the table is empty.

Both failure modes result in the client receiving less data than expected, but being told (via `electric-offset` header) that they received data up to the requested offset.

## The Fix

Modified `read_range_from_ets_cache` to return `{data, last_offset_read}` where `last_offset_read` is the tuple offset of the last entry actually read (or `nil` if empty). This allows detecting incomplete reads.

When an incomplete ETS read is detected (empty or partial), we fall back directly to reading from disk. This is safe because of the flush order in `write_loop.ex`:

1. `IO.binwrite` + `:file.datasync` — data written and synced to disk
2. `update_persistance_metadata` — updates `last_persisted_offset` in stack ETS
3. `trim_ets` — clears shape's ETS buffer via `:ets.delete_all_objects`

**If ETS is cleared, data is guaranteed to be on disk.** We can read directly using the existing `upper_read_bound` and `boundary_info` without re-reading metadata.

## Design Decision

The goal was to **minimally alter the existing code**. Alternative approaches like reader counters, delayed ETS clearing, or rearchitecting read/flush cycles were not considered—the fallback-to-disk approach requires only local changes to the reader path and doesn't affect the writer's eager memory cleanup behavior.

## Test Plan

- [x] Added test reproducing empty ETS race condition (fails without fix, passes with fix)
- [x] Added test reproducing partial ETS race condition (fails without fix, passes with fix)
- [x] All PureFileStorage tests pass (34 tests)
- [x] All LogOffset doctests pass (54 doctests)
- [x] Full test suite passes (1315 tests, 363 doctests, 8 properties)
- [x] No dialyzer warnings

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race in file-backed storage where concurrent flushes could produce empty or partial in-memory reads; readers now detect partial/cleared reads and retry using fresh metadata to ensure consistent reads from disk.

* **Tests**
  * Added tests simulating ETS read/write races to verify retry behavior and correct data recovery under concurrent flush scenarios.

### DIFF
--- a/packages/sync-service/test/electric/shape_cache/pure_file_storage_test.exs
+++ b/packages/sync-service/test/electric/shape_cache/pure_file_storage_test.exs
@@ -1,6 +1,5 @@
 defmodule Electric.ShapeCache.PureFileStorageTest do
   use ExUnit.Case, async: true
-  use Repatch.ExUnit
 
   import Support.ComponentSetup
   import Support.TestUtils
@@ -976,7 +975,7 @@ defmodule Electric.ShapeCache.PureFileStorageTest do
     setup :with_started_writer
     @describetag flush_period: 50
 
-    test "reader retries when ETS is unexpectedly empty due to concurrent flush", %{
+    test "reader falls back to disk when ETS is empty due to concurrent flush", %{
       writer: writer,
       opts: opts,
       stack_id: stack_id
@@ -994,7 +993,7 @@ defmodule Electric.ShapeCache.PureFileStorageTest do
       assert_receive {Storage, {PureFileStorage, :perform_scheduled_flush, [0]}}
       writer = PureFileStorage.perform_scheduled_flush(writer, 0)
 
-      # Get the fresh (correct) metadata after flush
+      # Get the fresh metadata after flush
       stack_ets = PureFileStorage.stack_ets(stack_id)
       [fresh_meta] = :ets.lookup(stack_ets, @shape_handle)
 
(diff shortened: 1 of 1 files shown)


packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
```diff
@@ -1135,23 +1135,45 @@ defmodule Electric.ShapeCache.PureFileStorage do
...
     cond do
       is_log_offset_lte(last_persisted, min_offset) and is_nil(ets) ->
         []
 
       is_log_offset_lte(last_persisted, min_offset) ->
-        read_range_from_ets_cache(ets, min_offset, upper_read_bound)
+        # Pure ETS read case
+        case read_range_from_ets_cache(ets, min_offset, upper_read_bound) do
+          {_data, last_offset}
+          when not retried? and (is_nil(last_offset) or last_offset < upper_read_bound_tuple) ->
```
msfstef: I'm contemplating whether we want to retry, or ti simply directly go to the main log since we know that this happens if we've flushed.

msfstef: I've updated the code to now directly read from the disk if the ETS cache is empty, since we know it will have flushed in that case

PR #3721, opened 2026-01-15
fix: Materializer startup race with consumer dying

msfstef:
## Summary

Fixes a race condition in the Materializer startup sequence where the Materializer crashes if the Consumer process dies during startup.

### The Race

The Materializer's `handle_continue(:start_materializer, ...)` calls `Consumer.await_snapshot_start` and `Consumer.subscribe_materializer` in sequence. Both functions use `GenServer.call` to the Consumer. If the Consumer dies during either call, the GenServer.call raises an `:exit` which crashes the Materializer.

### The Fix

Wrap both calls in a `try/catch` block to handle any `:exit` from GenServer.call gracefully:

```elixir
try do
  case Consumer.await_snapshot_start(stack_id, shape_handle, :infinity) do
    :started ->
      Consumer.subscribe_materializer(stack_id, shape_handle, self())
      Process.monitor(Consumer.whereis(stack_id, shape_handle), ...)
      {:noreply, state, {:continue, {:read_stream, shape_storage}}}
    {:error, _reason} ->
      {:stop, :shutdown, state}
  end
catch
  :exit, _ ->
    {:stop, :shutdown, state}
end
```

Instead of crashing, the Materializer now shuts down gracefully with `:shutdown` reason.

Note: `Process.monitor(nil, ...)` is safe - it returns a reference and immediately sends a `:DOWN` message, so no special handling is needed there.

## Test plan

- [x] Tests verify Materializer shuts down gracefully when Consumer dies during `await_snapshot_start`
- [x] Tests verify Materializer shuts down gracefully when Consumer dies during `subscribe_materializer`
- [x] Tests fail without the fix, pass with the fix
- [x] All existing materializer and shapes tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a startup race that could crash the Materializer when its dependent process terminated during initialization; Materializer now shuts down cleanly and logs a warning instead of crashing, improving stability.

* **Tests**
  * Added tests covering startup race scenarios to ensure graceful shutdown across timing-sensitive edge cases.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
@@ -16,6 +16,8 @@ defmodule Electric.Shapes.Consumer.Materializer do
   # and restarting would make no sense.
   use GenServer, restart: :temporary
 
+  require Logger
+
   alias Electric.Utils
   alias Electric.Replication.Changes
   alias Electric.Shapes.Consumer
@@ -124,7 +126,8 @@ defmodule Electric.Shapes.Consumer.Materializer do
       end
     catch
       # GenServer.call fails with :exit when Consumer is dead or dies mid-call
-      :exit, _ ->
+      :exit, reason ->
+        Logger.warning("Materializer startup failed with exit reason: #{inspect(reason)}")
         {:stop, :shutdown, state}
     end
   end


packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -108,17 +108,23 @@ defmodule Electric.Shapes.Consumer.Materializer do
...
 
-        {:noreply, state, {:continue, {:read_stream, shape_storage}}}
+          {:noreply, state, {:continue, {:read_stream, shape_storage}}}
 
-      {:error, _reason} ->
+        {:error, _reason} ->
+          {:stop, :shutdown, state}
+      end
+    catch
+      # GenServer.call fails with :exit when Consumer is dead or dies mid-call
```
msfstef, line 128: I went through a couple of approaches to avoid this but frankly this feels like the easiest, most code-localised approach that does the trick.

Although I realise it might be going against the "letting it fail" philosophy (?) happy to hear opinions

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -108,17 +108,23 @@ defmodule Electric.Shapes.Consumer.Materializer do
...
+      :exit, _ ->
```
robacourt: Maybe log a warning here inspecting the reason in case it's not the exit you were expecting.

msfstef: Added in fcab4ad13 - now logs the exit reason at warning level before shutting down.

PR #3730, opened 2026-01-17
website: update messaging and feature new products.

thruflo:
There is still more content to update/refactor but this may now be deployable, so we can ship and then iterate in place.

https://deploy-preview-3730--electric-next.netlify.app/


KyleAMathews:
Screenshot 2026-01-27 at 6 54 54 PM

these are confusing next to each other as it looks like they should be related

Issue #2322, opened 2025-02-10
Document how to unsubscribe to shapes


KyleAMathews:
We allude to it in the docs with documenting that you can use an AbortController but we should make it more explicit with some narrative and examples of how to you can use e.g. one controller per shape or one controller for a bunch of shapes or even a reference counter approach where it unsubscribes only when there's no other subscribers.

rgenge:
Unsubscribing Electric Shapes

If the ElectricSQL client is called repeatedly and the page is not refreshed, it doesn't automatically close previous subscriptions, leading to conflicts and performance issues due to lingering connections. This is particularly noticeable when using ShapeStream, as multiple subscriptions can accumulate without being properly aborted.

Here’s a working solution where we add an AbortController to the ShapeStream example. This ensures that each subscription is aborted once the data is fully loaded, preventing conflicts when multiple subscriptions are active. On the other hand, if you're using this connection to update pages in real-time, you might need to use some condition before aborting it:
```
import { ShapeStream, Shape } from '@electric-sql/client';

const abortController = new AbortController();
const signal = abortController.signal;

const stream = new ShapeStream({
  url: `http://localhost:3000/v1/shape`,
  params: {
    table: `foo`
  },
  signal
});
const shape = new Shape(stream);

// Returns promise that resolves with the latest shape data once it's fully loaded
await shape.rows;

// Optional: Abort the request when data is fully loaded
if (shape.currentRows && shape.currentRows.length > 0) {
  abortController.abort();
}
```

rgenge:
This is the first time I am trying to contribute to an project so feel free to say anything, I really do not know how to help and how to contribute but I want to learn so if this is bad just let me know how can I improve it.

thruflo:
Hey @rgenge thanks for this, really appreciate the input, thanks for the contribution :)

Generally the best way to do this is to make a Pull Request with the changes you're proposing. So in this case, there's a couple of pages in the docs that might be relevant:

- the Shapes guide https://github.com/electric-sql/electric/blob/main/website/docs/guides/shapes.md
- the Typescript client https://github.com/electric-sql/electric/blob/main/website/docs/api/clients/typescript.md
- also maybe the React integration https://github.com/electric-sql/electric/blob/main/website/docs/integrations/react.md

You could fork the repo, clone your fork, edit the markdown files locally, commit the changes and push to your fork. Then raise a PR. We can then review the proposed changes and merge the PR to update the docs.

[...]

PR #3732, opened 2026-01-18
fix(typescript-client): bundle patched version of fetch-event-source

pawelblaszczyk5:
## Summary

Bundles the patched `@microsoft/fetch-event-source` into `@electric-sql/client` dist output so consumers get the fixed version instead of the unpatched upstream package. Without this, `liveSse` shapes can't be aborted and throw errors in non-browser environments (SSR, Node, React Native).

## Root Cause

The monorepo applies a pnpm patch to `@microsoft/fetch-event-source` fixing document/window assumptions and abort signal bugs. But tsup treated it as an external dependency, so the published `@electric-sql/client` npm package emitted bare `import ... from '@microsoft/fetch-event-source'` statements. Consumers resolved this to the *unpatched* upstream version from npm, reintroducing all the bugs the patch fixed.

## Approach

Add `noExternal: ['@microsoft/fetch-event-source']` to the shared tsup config, which inlines the (patched) source into all four build outputs (ESM, CJS, legacy ESM, browser). Placed *after* the `...options` spread to ensure it always takes precedence.

### Key Invariants

- The patched fetch-event-source code must be present in all dist outputs
- No external `import`/`require` of `@microsoft/fetch-event-source` should appear in dist

## Verification

```bash
cd packages/typescript-client
pnpm build
pnpm vitest run --config vitest.unit.config.ts test/bundle.test.ts
```

The regression test reads all four dist outputs and asserts no external import/require of `@microsoft/fetch-event-source` exists.

## Files Changed

| File | Change |
|------|--------|
| `packages/typescript-client/tsup.config.ts` | Add `noExternal` after `...options` spread |
| `packages/typescript-client/test/bundle.test.ts` | Regression test asserting the dependency is inlined |
| `packages/typescript-client/vitest.unit.config.ts` | Register new test in unit config |
| `.changeset/slow-apricots-clean.md` | Changeset for patch bump |

---

Original PR by @pawelblaszczyk5 — see https://github.com/electric-sql/electric/issues/2322#issuecomment-3504810288 for the reported abort bug.

### DIFF
--- a/.changeset/slow-apricots-clean.md
+++ b/.changeset/slow-apricots-clean.md
@@ -1,0 +1,7 @@
+---
+'@electric-sql/client': patch
+---
+
+Properly bundle `fetch-event-source`, so consumer use patched version.
+
+When liveSse mode got introduced, it included `fetch-event-source` which is used instead of built-in `EventSource` because of richer capabilities. However, it had a few assumptions (document/window existence) + bugs, when it comes to aborting. This was patched, however, when building `typescript-client` patched version isn't included and when user uses it - they have unpatched version. 
--- a/packages/typescript-client/tsup.config.ts
+++ b/packages/typescript-client/tsup.config.ts
@@ -8,6 +8,7 @@ export default defineConfig((options) => {
     },
     tsconfig: `./tsconfig.build.json`,
     sourcemap: true,
+    noExternal: [`@microsoft/fetch-event-source`],
     ...options,
   }
 


pawelblaszczyk5:
I spotted someone running actions here and it failing with some unrelated golang error during build - I rebased the branch to current main, hope it helps 😄 Can I do anything more to help it move forward? Are my findings correct here, or should I dig deeper? Thanks a lot in advance!

pawelblaszczyk5:
Hello! Anything possible here to move this forward?

PR #3740, opened 2026-01-20
feat(sync-service): Write transaction fragments directly to storage

### LINKED ISSUE #3415: Avoid holding whole transaction in consumer memory

alco:
## Summary

Implements fragment-direct streaming for shape consumers, allowing transaction fragments to be written directly to storage as they arrive instead of buffering the entire transaction in memory until commit.

Closes #3415

## Motivation

Previously, the consumer would buffer all changes in a transaction in memory until a `Commit` message arrived. For large transactions, this could cause significant memory pressure. This PR enables shapes without subquery dependencies to stream fragments directly to storage, reducing memory usage while maintaining crash-safety guarantees.

## Key Changes

- **Fragment-direct mode**: Shapes without subquery dependencies now write transaction fragments directly to storage via `append_fragment_to_log!/2`
- **Crash-safety invariant**: `fetch_latest_offset` only returns committed transaction offsets, ensuring recovery correctness even if uncommitted fragment data is flushed to disk
- **Fallback to TransactionBuilder**: Shapes with dependencies, during initial filtering, or with materializer subscriptions continue using the existing buffered approach

## Limitations

Fragment-direct mode is only enabled when:
- Shape has no subquery dependencies
- Not buffering for initial snapshot
- No materializer subscribed (inner shapes need full transaction handling)
- Initial snapshot filtering is complete

### DIFF
--- a/packages/sync-service/agent-docs/flush-tracking-architecture.md
+++ b/packages/sync-service/agent-docs/flush-tracking-architecture.md
@@ -1,0 +1,368 @@
+# Flush Tracking Architecture
+
+This document explains how Electric's sync service tracks write progress across multiple shapes and coordinates acknowledgments back to Postgres. Understanding this architecture is essential for anyone working on storage, replication, or considering changes to the transaction processing pipeline.
+
+## Table of Contents
+
+1. [Overview](#overview)
+2. [LogOffset: The Fundamental Unit](#logoffset-the-fundamental-unit)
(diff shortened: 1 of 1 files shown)


alco:
Superseded by https://github.com/electric-sql/electric/pull/3783

PR #3743, opened 2026-01-20
ci(sync-service): trigger workflow when workflow file is modified

alco:
## Summary
- Add `.github/workflows/sync_service_tests.yml` to the workflow's path triggers
- Ensures the sync-service tests run when the workflow file itself is modified

## Summary by CodeRabbit

* **Chores**
  * Updated internal workflow configuration for testing processes.

### DIFF
--- a/.github/workflows/sync_service_tests.yml
+++ b/.github/workflows/sync_service_tests.yml
@@ -6,10 +6,12 @@ on:
     paths:
       - 'packages/electric-telemetry/**'
       - 'packages/sync-service/**'
+      - '.github/workflows/sync_service_tests.yml'
   pull_request:
     paths:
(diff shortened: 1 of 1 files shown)


alco:
@msfstef That workflow runs on a cron schedule. Are you suggesting adding another push trigger? I think one can be added temporarily for testing when making changes there. The purpose of the workflow is to run in background, not be associated with any specific commit in the repo.

msfstef:
@alco having a workflow trigger as well would be good so we can fix itwhen we want - e.g. now lots of tests are failing because of the limit on max connections, lt's not wait for the cron job!

edit: already has it - ignore me!

alco:
@msfstef I have rebuilt the Postgres image and rerun Elixir tests. All green now.

PR #3744, opened 2026-01-20
fix(sync-service): Fix hex.pm publishing for Electric package

### LINKED ISSUE #3737: Releases of Electric on hex.pm are broken since v1.2.4
You can see the latest published versions on [hex.pm](https://hex.pm/packages/electric).

The problem is that at one point (around https://github.com/electric-sql/electric/commit/c28e8ede9e67cd9ceabb69fb3384f5ba443e7565 and https://github.com/electric-sql/electric/commit/c4d0ea4b92f84ec4161d54c1792b2392eecbe5cc) we had to pin the version of `protobuf` because updating it would introduce a duplicate module problem between it and `protox`, another indirect dependency brought in by `pg_query_ex`.

Hex.pm doesn't allow publishing packages that have pinned dependencies, i.e. those with `override: true`. But since the invocation of the `ci:publish:hex-electric` script from the changesets_release workflow is trailed by `|| true`, the failure to publish it to hex.pm doesn't abort the overall publishing of packages.

---

To resolve this problem, we need to
- either address the code conflicts by removing duplicate modules from one of the lib
- or use only one of the lib both in [`pg_query_ex`](https://hex.pm/packages/pg_query_ex) and [`otel_metric_exporter`](https://hex.pm/packages/otel_metric_exporter).

robacourt:
Fixes #3737 
  
  Summary

  - Conditionally include telemetry dependencies only when building for MIX_TARGET=application
  - This allows mix hex.build and mix hex.publish to succeed

  Problem

  Releases of Electric on hex.pm have been broken since v1.2.4 (#3737) because:

  1. The protobuf dependency required override: true to avoid module conflicts with protox (both define Google.Protobuf.* modules starting from protobuf v0.14.0)
  2. The electric_telemetry dependency is a path dependency

  Hex.pm refuses to build packages with either of these conditions.

  Solution

  Since telemetry dependencies are only needed when building for production (MIX_TARGET=application), we now conditionally include them only for that target. When building for hex.pm (default target), these deps are excluded from the dependency list entirely.

  Test plan

  - mix hex.build successfully creates package
  - mix compile works without telemetry deps
  - MIX_TARGET=application mix compile works with all telemetry deps

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed hex.pm publishing for the Electric package.

* **Chores**
  * Simplified telemetry dependency configuration in the sync-service build with a clearer target-aware setup, reducing build complexity and improving maintainability.

### DIFF
--- a/packages/sync-service/mix.exs
+++ b/packages/sync-service/mix.exs
@@ -108,9 +108,6 @@ defmodule Electric.MixProject do
         {:tz, "~> 0.28"}
       ],
       dev_and_test_deps(),
-      # Only include telemetry deps when building for the telemetry target.
-      # This avoids issues with hex.pm when the deps include path dependencies or overrides
-      # that would otherwise prevent package building.
       telemetry_deps(Mix.target())
     ])
   end
@@ -127,6 +124,9 @@ defmodule Electric.MixProject do
     ]
   end
 
+  # Only include telemetry deps when building for the telemetry target.
+  # This avoids issues with hex.pm when the deps include path dependencies or overrides
+  # that would otherwise prevent package building.
   defp telemetry_deps(@telemetry_target) do
     [
       {:electric_telemetry, path: "../electric-telemetry"},


robacourt:
> I only have a nitpick: the conditional inside `deps` could look more natural in Elixir code if you instead passed `Mix.target()` as an argument to the `telemetry_deps()` function and matched on `@telemetry_target` in the function head.

Good point. Done.

PR #3745, opened 2026-01-20
fix(ci): Skip hex.pm publish when version already exists

### LINKED ISSUE #3736: changesets_release workflow ci:publish:hex-electric-client even when no new version is to be published
When an "update package versions" PR is merged and the changesets_release workflow runs, it fires of the `ci:publish:hex-electric-client` job without checking that there's a new version of the Elixir cilent. As a result, the publishing to hex.pm logs the following output in the CI log:
```
  ┃  . ci:publish:hex-electric-client: Publishing package...                                                                                                                                                         
  ┃  . ci:publish:hex-electric-client: [                         ] 0%                                                                                                                                                
  ┃  . ci:publish:hex-electric-client: [                         ] 0%                                                                                                                                                
  ┃  . ci:publish:hex-electric-client: [#####                    ] 23%                                                                                                                                               
  ┃  . ci:publish:hex-electric-client: [###########              ] 46%                                                                                                                                               
  ┃  . ci:publish:hex-electric-client: [#################        ] 69%                                                                                                                                               
  ┃  . ci:publish:hex-electric-client: [#######################  ] 93%                                                                                                                                               
  ┃  . ci:publish:hex-electric-client: Validation error(s)                                                                                                                                                           
  ┃  . ci:publish:hex-electric-client: [#########################] 100%Publishing failed                                                                                                                             
  ┃  . ci:publish:hex-electric-client:   inserted_at: can only modify a release up to one hour after publication                                                                                                     
  ┃  . ci:publish:hex-electric-client: Done 
```

It is not a problem because the invocation of the script is guarded with `|| true`. It's simply an inefficiency because the client code is built from scratch and failing to publish it is the very last step that discard all the building work done up to that point.

Example from a [recent run](https://github.com/electric-sql/electric/actions/runs/21003147814/job/60378379849#step:8:1005) on CI.

robacourt:
## Summary

- Check hex.pm API before attempting to publish `electric_client` and `electric` packages
- Skip unnecessary `deps.get` and compilation when the version is already published
- Applied to both `packages/elixir-client` and `packages/sync-service`

Fixes #3736

## Test plan

- [ ] Verify curl check works: `curl -sf "https://hex.pm/api/packages/electric_client/releases/0.8.0"` should succeed
- [ ] Verify non-existent version fails silently: `curl -sf "https://hex.pm/api/packages/electric_client/releases/99.99.99"` should fail
- [ ] CI should skip hex publish step on next changesets release if no version change

## Summary by CodeRabbit

* **Chores**
  * Improved package publishing workflow to prevent duplicate publishes of existing versions to Hex.pm registry.
  * Centralized publishing logic through a dedicated script for better management and consistency across packages.

### DIFF
--- a/scripts/publish_hex.sh
+++ b/scripts/publish_hex.sh
@@ -17,6 +17,10 @@ if [ -z "$1" ]; then
 fi
 
 HEX_PACKAGE_NAME="$1"
+
+# Note: This script must be run from the package directory containing package.json.
+# When invoked via `pnpm run publish:hex`, pnpm automatically sets the cwd to
+# the package directory, so this requirement is satisfied.
 VERSION=$(node -p "require('./package.json').version")
 
 echo "Checking if ${HEX_PACKAGE_NAME} version ${VERSION} exists on hex.pm..."


alco (commented):
I think I've already seen a similar suggestion from Claude Code before. In my case, though, I had told it upfront to check for the presence of a tag on GitHub, so it `curl`ed the GitHub API instead.

In any case, I think the "check the API" step would be more readable if it lived in a separate script file, such that it can also be easily invoked by hand to confirm it's workign as expected.

robacourt:
> I think I've already seen a similar suggestion from Claude Code before. In my case, though, I had told it upfront to check for the presence of a tag on GitHub, so it `curl`ed the GitHub API instead.
> 
> In any case, I think the "check the API" step would be more readable if it lived in a separate script file, such that it can also be easily invoked by hand to confirm it's workign as expected.

Good idea. Done!

scripts/publish_hex.sh
```diff
@@ -0,0 +1,30 @@
...
+#   ./scripts/publish_hex.sh electric
+
+if [ -z "$1" ]; then
+  echo "Usage: $0 <hex_package_name>"
+  echo "Example: $0 electric_client"
+  exit 1
+fi
+
+HEX_PACKAGE_NAME="$1"
+VERSION=$(node -p "require('./package.json').version")
```
alco, line 24: This step depends on the current directory but it's not mentioned in the docs/usage and it's quite subtle. Are you sure that when changesets runs the respective `publish:hex` scripts for different packages it does so from their respective directories?

alco: I mean, duh, it does run in the correct directory since right now the command that gets run is `mix deps.get + hex.publish`.

Nevermind. I just think this needs to be mentioned in a code comment because it's a subtle step.

robacourt: Good idea. I've added a comment

PR #3750, opened 2026-01-21
fix(sync-service): Compute correct tags for NULL column values in subqueries

robacourt:
## Summary

Fixes a crash when using on-demand sync with subqueries where rows have NULL values in the referenced column.

## Problem

When using shapes with subquery WHERE clauses like `task_id IN (SELECT id FROM tasks WHERE ...)`, if any row has a NULL value in the subquery column (`task_id`), Electric crashes with:

```
ArgumentError: not an iodata term
```

**Root cause:** In `make_tags` (querying.ex), tag generation uses SQL like:
```sql
md5('stack_id_handle' || "column"::text)
```

When the column is NULL:
1. `"column"::text` evaluates to NULL
2. `'...' || NULL` = NULL in PostgreSQL (concatenating anything with NULL produces NULL)
3. This NULL propagates through all string concatenation in the row's JSON construction
4. The entire row becomes `nil` in the query result
5. The encoder crashes trying to encode `nil` as iodata

## Solution

Namespace column values with a `v:` prefix, represent NULL as `NULL` (no prefix):
```sql
md5('stack_id_handle' || CASE WHEN col IS NULL THEN 'NULL' ELSE 'v:' || col END)
```

This ensures:
- NULL values don't propagate through concatenation
- NULL and the string literal `'NULL'` produce distinct hash values
- No restrictions on what values users can have in their columns

## Test plan

- [x] Added regression test verifying rows with NULL column values are handled correctly
- [x] All existing tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash in on‑demand sync when subquery columns contained NULLs by changing tag encoding: non‑NULL values are namespaced (prefixed) while NULL is represented distinctly, preventing NULL propagation, tag collisions, and incorrect JSON construction.

* **Tests**
  * Updated and added tests to reflect the new namespaced tag format and explicit NULL handling.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/querying.ex
+++ b/packages/sync-service/lib/electric/shapes/querying.ex
@@ -2,12 +2,11 @@ defmodule Electric.Shapes.Querying do
   alias Electric.ShapeCache.LogChunker
   alias Electric.Utils
   alias Electric.Shapes.Shape
+  alias Electric.Shapes.Shape.SubqueryMoves
   alias Electric.Telemetry.OpenTelemetry
 
-  # Value namespacing constants - MUST match SubqueryMoves.namespace_value/1
-  # See lib/electric/shapes/shape/subquery_moves.ex for the Elixir implementation
-  @value_prefix "v:"
-  @null_sentinel "NULL"
+  @value_prefix SubqueryMoves.value_prefix()
+  @null_sentinel SubqueryMoves.null_sentinel()
 
   def query_move_in(conn, stack_id, shape_handle, shape, {where, params}) do
     table = Utils.relation_to_sql(shape.root_table)
--- a/packages/sync-service/lib/electric/shapes/shape/subquery_moves.ex
+++ b/packages/sync-service/lib/electric/shapes/shape/subquery_moves.ex
@@ -4,10 +4,11 @@ defmodule Electric.Shapes.Shape.SubqueryMoves do
   alias Electric.Replication.Eval.Walker
   alias Electric.Shapes.Shape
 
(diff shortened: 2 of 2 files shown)


icehaunter:
I've asked Rob on Discord to address that rows with `value = NULL` and `value = '__NULL__'` in SQL will hash the same. By effectively name-spacing the real values using a prefix

packages/sync-service/lib/electric/shapes/querying.ex
```diff
@@ -150,10 +150,19 @@ defmodule Electric.Shapes.Querying do
...
+          namespaced = ~s[CASE WHEN #{col} IS NULL THEN 'NULL' ELSE 'v:' || #{col} END]
+          ~s[md5('#{stack_id}#{shape_handle}' || #{namespaced})]
 
         {:hash_together, columns} ->
-          column_parts = Enum.map(columns, &~s['#{&1}:' || #{pg_cast_column_to_text(&1)}])
+          column_parts =
+            Enum.map(columns, fn col_name ->
+              col = pg_cast_column_to_text(col_name)
+              ~s['#{col_name}:' || CASE WHEN #{col} IS NULL THEN 'NULL' ELSE 'v:' || #{col} END]
+            end)
```
alco, line 166: The building of the namespaced string could trivially be extracted into a helper function, to DRY it up, as it's easy to miss that the two occurrences must actually be the same when in the future a change is made to one of them.

it also feels icky that this SQL logic must be in sync with Elixir logic in the `subquery_moves.ex` file but there's no mention of that in the code.

robacourt: Good idea. Done

PR #3754, opened 2026-01-21
fix: Make rate limit error a known error

msfstef:
## Summary
- Marks the 503 rate limit/overloaded error as a known error by adding `known_error: true` to the error response
- This ensures rate limiting responses are treated as expected behavior rather than unexpected errors

## Test plan
- [ ] Verify rate limit responses include the known_error flag
- [ ] Confirm error handling/logging treats this as a known error


balegas:
this is right, but we should add an alarm that a source is being rate limited for a certain amount of time

PR #3761, opened 2026-01-22
fix(sync-service): Fix shape cleanup race condition

### LINKED ISSUE #3760: Race condition + potential storage corruption by shape cleanup.
I noticed a test failure on CI what I hadn't seen before. Let Opus 4.5 loose on it and it came back with the following analysis after it had tried to reproduce the failure locally, unsuccessfully.

## Prompt

Examine the test failure seen in this CI run - https://github.com/electric-sql/electric/actions/runs/21229289729/job/61084054610 (Build and test sync-service, pg15) and try reproducing it locally.

The failure I'm interested in is this:

1) test /v1/shapes - subqueries supports two subqueries at the same level but returns 409 on move-in (Electric.Plug.RouterTest)
Error:      test/electric/plug/router_test.exs:2985 

Reason about the code and offer your best guess as to why the test failed even if you don't manage to reproduce the same failure locally.

---

### The Bug

The failure is a **race condition** causing a `CaseClauseError` in `Electric.Shapes.Api.notify_changes_since_request_start/1` at line 729.

The error shows:
```elixir
** (CaseClauseError) no case clause matching:
{"14433500-1769037623985358", LogOffset.last_before_real_offsets()}
```

### Root Cause

The `notify_changes_since_request_start` function in `lib/electric/shapes/api.ex:729` has a `case` statement that handles four scenarios:

```elixir
case Shapes.resolve_shape_handle(stack_id, shape_handle, shape_def) do
  {^shape_handle, ^last_offset} ->        # 1. No changes
  {^shape_handle, latest_log_offset} when is_log_offset_lt(last_offset, latest_log_offset) -> # 2. New changes
  {other_shape_handle, _} when other_shape_handle != shape_handle -> # 3. Different handle
  nil ->                                   # 4. Shape not found
end
```

**Missing case**: The code doesn't handle when `resolve_shape_handle` returns `{shape_handle, latest_log_offset}` where `latest_log_offset < last_offset` (i.e., the shape's offset has gone BACKWARDS compared to the client's offset).

### How the Race Condition Occurs

1. **Test Setup**: A shape with two subqueries is created and has processed transactions (client offset = `465001208_0`)

2. **Move-in Event**: An INSERT into `project_members` triggers a move-in that should invalidate the shape (because multiple same-level subqueries can't handle move-ins correctly)

3. **Consumer Invalidation**: The consumer calls `stop_and_clean()` which:
   - Marks the consumer as terminating
   - Returns `{:stop, @shutdown_cleanup, state}`
   - In `terminate/2`, calls `handle_writer_termination` → `remove_shape_async`

4. **Async Cleanup**: `remove_shape_async` spawns an async task that will eventually:
   - Remove the shape from ShapeStatus
   - Stop the consumer
   - Clean up storage

5. **The Race Window**: Between when the consumer terminates and when the async cleanup removes the shape from ShapeStatus:
   - The writer ETS is deleted (during `terminate_writer`)
   - The stack ETS entry is cleaned but not removed
   - The shape is still "valid" in ShapeStatus

6. **API Request Wakes Up**: The live poll request wakes up and calls `notify_changes_since_request_start`

7. **resolve_shape_handle Returns Unexpected Value**:
   - `validate_shape_handle` succeeds (shape still in ShapeStatus)
   - `fetch_latest_offset` tries to read shape metadata
   - Due to the race, the storage metadata returns `nil` values
   - `latest_offset` returns `LogOffset.last_before_real_offsets()` (`{tx_offset: 0, op_offset: :infinity}`)

8. **Case Clause Failure**: The returned tuple `{shape_handle, last_before_real_offsets()}` doesn't match any case clause because:
   - First clause fails: `last_before_real_offsets() ≠ 465001208_0`
   - Second clause fails: `is_log_offset_lt(465001208_0, 0_infinity)` is false (465001208 > 0)
   - Third clause fails: `shape_handle == shape_handle`
   - Fourth clause fails: not `nil`

### Why It Doesn't Reproduce Locally

This race condition is extremely timing-sensitive. It requires the exact interleaving where:
1. The async cleanup has deleted or corrupted the storage metadata
2. BUT the ShapeStatus still has the shape entry

In CI environments (GitHub Actions), there may be more CPU contention, different scheduling, or I/O latency that makes this window more likely to hit.

### Suggested Fix

Add a case clause to handle the "shape went backwards" scenario, which should trigger a shape rotation (409):

```elixir
case Shapes.resolve_shape_handle(stack_id, shape_handle, shape_def) do
  {^shape_handle, ^last_offset} ->
    # no-op, shape is still present and unchanged
    nil

  {^shape_handle, latest_log_offset} when is_log_offset_lt(last_offset, latest_log_offset) ->
    send(self(), {ref, :new_changes, latest_log_offset})

  # NEW CASE: Handle shape offset going backwards (indicates shape was reset/invalidated)
  {^shape_handle, _latest_log_offset} ->
    # Shape was reset/invalidated - treat as rotation
    send(self(), {ref, :shape_rotation})

  {other_shape_handle, _} when other_shape_handle != shape_handle ->
    send(self(), {ref, :shape_rotation, other_shape_handle})

  nil ->
    send(self(), {ref, :shape_rotation})
end
```

This ensures that when the shape's offset appears to go backwards (indicating the shape was invalidated/reset), the client receives a 409 response prompting them to refetch.

magnetised:
Fixes #3760

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sync service reliability by properly addressing offset regression scenarios that can occur during live polling. The system now detects these edge cases and returns appropriate HTTP error responses to trigger necessary data refetch operations, preventing silent failures.

* **Tests**
  * Added comprehensive test coverage for offset regression edge cases during live polling operations.

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -734,6 +734,22 @@ defmodule Electric.Shapes.Api do
       {^shape_handle, latest_log_offset} when is_log_offset_lt(last_offset, latest_log_offset) ->
         send(self(), {ref, :new_changes, latest_log_offset})
 
+      {^shape_handle, _latest_log_offset} ->
+        # Fix for issue #3760: Handle offset regression during shape invalidation.
+        #
+        # This case handles a race condition where:
+        # 1. Shape invalidation spawns an async cleanup task
+        # 2. Between consumer termination and async cleanup completion, the writer
+        #    ETS gets deleted while ShapeStatus retains the shape entry
+        # 3. A pending API request queries the shape
+        # 4. Validation succeeds (shape exists in ShapeStatus), but metadata reading
+        #    fails, causing resolve_shape_handle to return LogOffset.last_before_real_offsets()
+        # 5. This offset is less than the client's stored offset (offset regression)
+        #
+        # When the offset goes backwards, the shape has effectively been invalidated
+        # and the client should refetch from the beginning.
+        send(self(), {ref, :shape_rotation})
+
       {other_shape_handle, _} when other_shape_handle != shape_handle ->
         send(self(), {ref, :shape_rotation, other_shape_handle})
 
--- a/.changeset/nervous-brooms-relax.md
+++ b/.changeset/nervous-brooms-relax.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Fix storage race condition when deleting shape during a live poll request
--- a/packages/sync-service/test/electric/shapes/api_test.exs
+++ b/packages/sync-service/test/electric/shapes/api_test.exs
@@ -1273,6 +1273,71 @@ defmodule Electric.Shapes.ApiTest do
     end
 
     @tag long_poll_timeout: 100
+    test "handles shape offset regression during live polling (issue #3760)", ctx do
+      # This test reproduces the CaseClauseError described in issue #3760.
+      # The bug occurs when:
+      # 1. A live request is waiting for changes
+      # 2. Shape invalidation spawns an async cleanup task
+      # 3. Between consumer termination and async cleanup completion,
+      #    the writer ETS gets deleted while ShapeStatus retains the shape entry
+      # 4. When notify_changes_since_request_start/1 calls resolve_shape_handle,
+      #    it returns {shape_handle, LogOffset.last_before_real_offsets()} as a fallback
+      # 5. This offset is *less than* the client's stored offset, causing a
+      #    CaseClauseError because no case clause handles offset regression
+
+      regressed_offset = LogOffset.last_before_real_offsets()
+
+      expect_shape_cache(
+        # First call during validation - returns valid offset
+        resolve_shape_handle: fn @test_shape_handle, @test_shape, _stack_id ->
+          {@test_shape_handle, @test_offset}
+        end,
+        # Second call during notify_changes_since_request_start - simulates
+        # the race condition where shape metadata was partially cleaned up,
+        # causing offset to regress to last_before_real_offsets()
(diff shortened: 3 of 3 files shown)


packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -734,6 +734,22 @@ defmodule Electric.Shapes.Api do
...
+      {^shape_handle, _latest_log_offset} ->
+        # Fix for issue #3760: Handle offset regression during shape invalidation.
+        #
+        # This case handles a race condition where:
+        # 1. Shape invalidation spawns an async cleanup task
+        # 2. Between consumer termination and async cleanup completion, the writer
+        #    ETS gets deleted while ShapeStatus retains the shape entry
+        # 3. A pending API request queries the shape
+        # 4. Validation succeeds (shape exists in ShapeStatus), but metadata reading
+        #    fails, causing resolve_shape_handle to return LogOffset.last_before_real_offsets()
```
msfstef, line 746: My agent had found this race as well but I had disliked this solution - what is it that is actually returning `last_before_real_offsets()`, and should it be returning something else instead rather than having this as a default? The default should perhaps be set when the shape is created, but if the shape is deleted we shouldn't fall back to the default

magnetised: This is the result of a race condition and we don't actually know that the shape is deleted at the point where we return this default offset, which is what you get with no ets table and no metadata files. Changing that default would break a lot of things. This feels like a pragmatic solution to the problem to me.

msfstef: fair enough - I'd prefer it if we somehow avoided this "default offset" being returned and handling the missing offset instead but I'm not sure how deep that rabbit hole is

PR #3764, opened 2026-01-23
fix: Carry over last change to next fragment to fix max_batch_size boundary bug

### LINKED ISSUE #3726: Batch INSERT Replication Fails for Sizes Divisible by 100
## Versions

- **Client**: Plain HTTP (integration test using Electric.Client)
- **Electric**: sync-service (Elixir backend) - current main branch

## Bug Description

Batch INSERT statements with row counts exactly divisible by 100 (100, 200, 300, 400, 500, etc.) fail to replicate. The database insert succeeds, but **zero messages** are received by clients instead of the expected number of change messages.

### Affected Batch Sizes
- ✅ **Works**: 25, 50, 75, 125, 150, 175, 225, 250, 275, 325, 350, 375, 425, 450, 475
- ❌ **Fails**: 100, 200, 300, 400, 500 (and presumably 600, 700, etc.)

## Expected Behavior

When executing a batch INSERT statement with N rows (e.g., `INSERT INTO items VALUES (...), (...), ...` with 100 rows), the Electric sync service should:

1. Successfully insert all N rows into the database (this works)
2. Replicate all N change messages to subscribed clients (this fails for N divisible by 100)
3. Clients should receive N `ChangeMessage` events with `operation: :insert`

## Steps to Reproduce

1. Set up a table and Electric shape subscription
2. Execute a batch INSERT with exactly 100 rows (or 200, 300, etc.)
3. Wait for replication
4. Observe that 0 messages are received instead of 100

### Test Case
See attached test file: `packages/sync-service/test/integration/batch_insert_test.exs`
- Test: `"receives all rows from batch insert (100 rows)"`
- Expected: 100 messages
- Actual: 0 messages

alco:
## Summary

Fix #3726.

Fixes the bug where transactions with a number of changes that is a multiple of `max_batch_size` (default: 100) had their commit fragment silently dropped by `ShapeLogCollector`.

### Root Cause

When a transaction has exactly N * `max_batch_size` changes (e.g. 100, 200, 300, ...):
1. The last change triggers `maybe_flush/1`, returning a fragment with `last_log_offset = (final_lsn, X)`
2. `maybe_flush` creates a new empty fragment but does NOT reset `state.last_log_offset`
3. When the Commit arrives, the final fragment gets the same `last_log_offset = (final_lsn, X)`
4. `ShapeLogCollector.handle_txn_fragment/2` drops it as a duplicate (offset <= last_processed_offset)

A secondary issue: when a fragment is flushed at exactly `max_batch_size`, the last change in that fragment is **not** marked with `last?=true`, which is incorrect since it should only be set on the very last change of the entire transaction.

### Solution

When flushing a fragment at the `max_batch_size` boundary, retain the last change from the flushed fragment and include it in the next fragment. This guarantees that even if the next replication message is a Commit, the resulting fragment will have at least one change with a `log_offset` greater than the previous fragment's `last_log_offset`.

This approach also fixes the `last?` marking issue since the carried-over change will be correctly annotated when its fragment is finalized.

### DIFF
--- a/.changeset/two-singers-repeat.md
+++ b/.changeset/two-singers-repeat.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Fix the issue where transactions that had exactly max_batch_size changes weren't written to the shape log.


robacourt (approved):
Nice solution. Great work!

PR #3774, opened 2026-01-26
Add runtime validation for columnMapper to catch common mistakes

KyleAMathews:
## Summary

Adds runtime validation for the `columnMapper` option in `ShapeStream` to catch a common developer mistake: passing the factory function (`snakeCamelMapper`) instead of calling it (`snakeCamelMapper()`).

## Root Cause

TypeScript's structural typing means that both `snakeCamelMapper` (the function) and `snakeCamelMapper()` (the result) type-check as valid `columnMapper` values at compile time. When users make this mistake, they get confusing runtime errors deep in the column mapping logic rather than a clear error at construction time.

## Approach

1. **Brand symbol** (`COLUMN_MAPPER_BRAND`) added to `ColumnMapper` interface to distinguish objects from functions at the type level
2. **Runtime validation** in `validateOptions()` checks:
   - If a function was passed → error with "Did you forget to call snakeCamelMapper()?"
   - If object lacks encode/decode → error with guidance to use factory functions
3. **Backwards compatible**: Custom mappers without the brand are still accepted if they have valid `encode`/`decode` methods

## Key Invariants

- Factory functions (`snakeCamelMapper()`, `createColumnMapper()`) always return branded objects
- Validation is duck-typed for backwards compatibility (brand not required at runtime)
- Error messages include the function name when available for actionable guidance

## Non-goals

- **Breaking custom mappers**: Users with plain `{ encode, decode }` objects should continue working
- **Type-level enforcement**: While the brand helps TypeScript, runtime validation is the primary defense

## Verification

```bash
# Type-check
cd packages/typescript-client && pnpm exec tsc --noEmit

# Run tests (requires Electric server)
pnpm test
```

## Files Changed

| File | Change |
|------|--------|
| `src/error.ts` | Add `InvalidColumnMapperError` class |
| `src/column-mapper.ts` | Add `COLUMN_MAPPER_BRAND`, `isValidColumnMapper()`, update factories |
| `src/client.ts` | Add validation in `validateOptions()` |
| `src/index.ts` | Export new error and validator |
| `test/column-mapper.test.ts` | Unit tests for validator and brand |
| `test/stream.test.ts` | Integration tests for validation errors |

---

### DIFF
--- a/packages/typescript-client/src/column-mapper.ts
+++ b/packages/typescript-client/src/column-mapper.ts
@@ -4,14 +4,6 @@ type DbColumnName = string
 type AppColumnName = string
 
 /**
- * Symbol used to brand ColumnMapper objects.
- * This helps TypeScript distinguish between a ColumnMapper object
- * and a function that returns a ColumnMapper.
- * @internal
- */
-export const COLUMN_MAPPER_BRAND = Symbol.for(`electric.columnMapper`)
-
-/**
  * Quote a PostgreSQL identifier for safe use in query parameters.
  *
  * Wraps the identifier in double quotes and escapes any internal
@@ -62,14 +54,6 @@ export function quoteIdentifier(identifier: string): string {
  */
 export interface ColumnMapper {
   /**
-   * Brand marker to distinguish ColumnMapper objects from functions.
-   * This helps catch the common mistake of passing `snakeCamelMapper`
-   * instead of `snakeCamelMapper()`.
-   * @internal
-   */
-  readonly [COLUMN_MAPPER_BRAND]: true
-
-  /**
    * Transform a column name from database format to application format.
    * Applied to column names in query results.
    */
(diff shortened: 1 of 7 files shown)


samwillis (commented):
Runtime validation looks good, but the `COLUMN_MAPPER_BRAND` looks entirely redundant. Suspect Claude tried to use it but could not for backwards compatibility?

packages/typescript-client/src/column-mapper.ts
```diff
@@ -57,6 +82,29 @@ export interface ColumnMapper {
...
 
+/**
+ * Validates that a value is a valid ColumnMapper object.
+ * Returns false if the value is a function (common mistake of passing
+ * the factory without calling it) or if it lacks encode/decode methods.
+ *
+ * Note: This intentionally does NOT check for COLUMN_MAPPER_BRAND to maintain
+ * backwards compatibility with custom ColumnMapper implementations that use
+ * plain objects with encode/decode methods. The brand is used for TypeScript
+ * type narrowing and documentation, not runtime enforcement.
```
samwillis: The `COLUMN_MAPPER_BRAND` seems entirely unused, except for in tests.

> The brand is used for TypeScript type narrowing and documentation

I don't see where its doing at.

KyleAMathews:
Good catch\! Removed the `COLUMN_MAPPER_BRAND` entirely in 0591f76f - it was adding complexity without benefit since we couldn't use it for validation (to maintain backwards compatibility with custom mappers). The runtime checks are what actually catch the common mistake.

KyleAMathews:
@samwillis ready for a re-review when you're ready

KyleAMathews:

PR #3778, opened 2026-01-27
fix(sync-service): Fix move-out race condition for rows entering shape in same txn

robacourt:
## Summary

- Fix race condition where rows entering a shape via FK update in the same transaction as the parent being deactivated were not correctly removed
- The move-out notification could arrive before the materializer subscribed or before the transaction was fully processed, causing the new row to be missed
- Added buffering for notifications when materializer isn't subscribed yet
- Added xid tracking to scope recently_removed values to the same transaction
- Added re-application of move-outs after transaction processing

## Test plan

- [x] Existing test at `test/integration/subquery_move_out_test.exs:343` now passes
- [x] All 11 tests in `subquery_move_out_test.exs` pass
- [ ] Manual testing of shapes with subqueries involving FK updates in same transaction as parent deactivation

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -210,7 +210,25 @@ defmodule Electric.Shapes.Consumer do
   def handle_call({:subscribe_materializer, pid}, _from, state) do
     Logger.debug("Subscribing materializer for #{state.shape_handle}")
     Process.monitor(pid, tag: :materializer_down)
-    {:reply, :ok, %{state | materializer_subscribed?: true}, state.hibernate_after}
+
+    # Flush any buffered notifications that arrived before the materializer subscribed
+    if state.pending_materializer_notifications != [] do
+      Logger.info(
+        "Flushing #{length(state.pending_materializer_notifications)} buffered notifications to materializer for #{state.shape_handle}"
+      )
+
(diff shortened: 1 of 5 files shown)


packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -456,11 +489,35 @@ defmodule Electric.Shapes.Consumer do
       TransactionBuilder.build(txn_fragment, state.transaction_builder)
 
     state = %{state | transaction_builder: transaction_builder}
-    handle_txns(txns, state)
+    state = handle_txns(txns, state)
+
+    # Re-apply recent move-outs. This handles the case where a row enters the shape
+    # in the same transaction that a dependency is removed. The move-out control message
+    # was written BEFORE the insert, so we need to write another move-out AFTER.
+    reapply_recent_move_outs(state)
```
robacourt, line 497: **Concern:** `recent_move_outs` is replayed on the next child transaction without any `xid`/boundary scoping. If the child doesn’t observe the same commit boundary as the dependency (e.g., no child table
  changes), this can leak across transactions or linger indefinitely. I think the contract should be: *if any dependency emits a commit, the child consumer must also observe that commit* — otherwise causality
  breaks. Can we either (a) enforce that commit‑notification invariant, or (b) scope `recent_move_outs` by `xid` and clear on any txn boundary (plus TTL fallback)?

PR #3782, opened 2026-01-27
Fix bug where the TypeScript client incorrectly updates its offset from stale CDN responses

KyleAMathews:
## Summary

Fix a bug where the TypeScript client incorrectly updates its offset from stale CDN responses. When a CDN returns a cached response with an expired shape handle, the client now ignores the entire response to prevent handle/offset mismatch.

## Root Cause

The stale cache detection logic in `#onInitialResponse` had three branches:
1. **New handle from response** → accept it
2. **Stale handle, client has no handle yet** → retry with cache buster
3. **Stale handle, client already has valid handle** → log warning, continue

Branch 3 correctly preserved the client's existing handle, but **the code continued to execute** and updated `#lastOffset` from the stale response. This created a state where:
- `handle = B` (correctly preserved)
- `offset = 0_inf` (incorrectly taken from stale response for handle A)

Subsequent requests with mismatched handle/offset caused server errors.

## Approach

Add an early `return` after logging the warning in branch 3, preventing any state updates from the stale response:

```typescript
} else {
  // We already have a valid handle, so ignore the stale response entirely
  // to prevent a mismatch between our current handle and the stale offset.
  console.warn(`[Electric] Received stale cached response...`)
  return  // ← THE FIX: skip all state updates
}
```

## Key Invariants

1. **Handle and offset must always be consistent** - they come from the same response
2. **Stale responses are fully ignored** - no partial state extraction
3. **Warning is logged** - operators can diagnose CDN misconfiguration

## Non-goals

- Not changing how stale responses are detected
- Not adding retry logic for branch 3 (client already has valid state)
- Not surfacing warning to `onError` callback (separate enhancement)

## Trade-offs

**Alternative**: Cancel response body and throw error to trigger retry.
**Why rejected**: Client already has valid state; continuing is the right behavior. The early return is simpler and sufficient.

## Verification

```bash
cd packages/typescript-client
pnpm vitest run test/expired-shapes-cache.test.ts
```

The test `should not update offset from stale response when client already has a handle` directly reproduces the bug scenario.

## Files Changed

| File | Change |
|------|--------|
| `src/client.ts` | Add early `return` after stale response warning to skip state updates |
| `test/expired-shapes-cache.test.ts` | Add regression test for the bug scenario |

---

### DIFF
--- a/.changeset/fix-stale-cache-offset-update.md
+++ b/.changeset/fix-stale-cache-offset-update.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/client': patch
+---
+
+Fix stale CDN response incorrectly updating client offset. When a CDN returns a cached response with an expired shape handle, the client now ignores the entire response (including offset) to prevent handle/offset mismatch that would cause server errors.


kevin-dp (approved):
Love a one line change that fixes the issue!

PR #3783, opened 2026-01-27
feat: Write transaction fragments directly to storage to reduce consumer memory footprint

### LINKED ISSUE #3415: Avoid holding whole transaction in consumer memory

alco:
Closes #3415

### Summary

For shapes without subquery dependencies (the vast majority), the consumer now streams each transaction fragment directly to storage as it arrives, instead of buffering the entire transaction in memory. This significantly reduces per-shape-consumer memory usage for large transactions.

Shapes with subquery dependencies continue to buffer the full transaction (`write_unit=txn`) because they need the complete change set for move-in/move-out reasoning.

### How it works

The consumer operates in one of two modes, determined at shape creation time:

- **`write_unit=txn_fragment`** (default): Each fragment is filtered, converted to log items, and written to storage immediately. A lightweight `PendingTxn` struct tracks only metadata (xid, offset, byte count, change count) across fragments. When the final fragment with a commit arrives, storage is notified to advance its transaction boundary.

- **`write_unit=txn`** (shapes with subquery dependencies): Fragments are accumulated into a complete transaction before writing, as before.

### Key additions

**`Consumer.PendingTxn`** — tracks in-progress transaction metadata during fragment streaming. It is the lightweight counterpart to `TransactionBuilder`: where `TransactionBuilder` accumulates all changes in memory, `PendingTxn` only tracks offsets and counters while actual data is written to storage and discarded immediately.

**New storage callbacks** — `append_fragment_to_log!/2` writes log items without advancing the persisted transaction offset (can be called multiple times per transaction), and `signal_txn_commit!/2` marks the transaction as complete so crash-recovery sees the correct committed boundary. The existing full-transaction write path is refactored to share the same underlying logic.

**`Support.StorageTracer`** — test helper using Erlang trace sessions to observe which storage functions are called from a process, enabling non-intrusive assertions on the write path taken without wrapper modules or function patching.

### Other notable changes

- Initial snapshot filtering (xid-based skip) is now applied at the first fragment, avoiding accumulation of fragments for transactions that will be discarded.
- The `FlushTracker` in `ShapeLogCollector` is now only invoked for fragments that include a commit, since mid-transaction fragments don't represent a flush boundary.
- Removed `CrashingFileStorage` (dead code).

### Expected memory footprint reduction

Previously, the consumer held an entire transaction's worth of change structs in memory until commit. Now, each fragment's changes are written to storage and released immediately. The consumer only retains the `PendingTxn` metadata struct (~100 bytes) per in-progress transaction, reducing peak memory by roughly the size of the buffered change data. The improvement scales linearly with transaction size.

### Testing

Consumer and storage tests have been expanded to cover both `write_unit` modes, including tracing-based assertions on which storage functions are called.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -353,10 +353,14 @@ defmodule Electric.ShapeCache do
       })
     end)
 
-    case Shapes.DynamicConsumerSupervisor.start_shape_consumer(
-           stack_id,
-           Map.put(opts, :shape_handle, shape_handle)
-         ) do
+    feature_flags = Electric.StackConfig.lookup(stack_id, :feature_flags, [])
+
+    start_opts =
+      opts
+      |> Map.put(:shape_handle, shape_handle)
+      |> Map.put(:subqueries_enabled_for_stack?, "allow_subqueries" in feature_flags)
+
+    case Shapes.DynamicConsumerSupervisor.start_shape_consumer(stack_id, start_opts) do
       {:ok, consumer_pid} ->
         # Now that the consumer process for this shape is running, we can finish initializing
         # the ShapeStatus record by recording a "last_read" timestamp on it.
--- a/packages/sync-service/lib/electric/shapes/consumer/state.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/state.ex
@@ -164,7 +164,7 @@ defmodule Electric.Shapes.Consumer.State do
   end
 
   @spec initialize_shape(uninitialized_t(), Shape.t(), map()) :: uninitialized_t()
-  def initialize_shape(%__MODULE__{} = state, shape, start_config) do
+  def initialize_shape(%__MODULE__{} = state, shape, opts) do
     %{
       state
       | shape: shape,
@@ -173,7 +173,8 @@ defmodule Electric.Shapes.Consumer.State do
         # Enable direct fragment-to-storage streaming for shapes without subquery dependencies
         # and if the current shape itself isn't an inner shape of a shape with subqueries.
         write_unit:
-          if shape.shape_dependencies != [] or Map.get(start_config, :is_subquery_shape?, false) do
+          if Map.get(opts, :subqueries_enabled_for_stack?, false) or
+               shape.shape_dependencies != [] or Map.get(opts, :is_subquery_shape?, false) do
             @write_unit_txn
           else
             @write_unit_txn_fragment
--- a/packages/sync-service/test/electric/shapes/consumer_test.exs
+++ b/packages/sync-service/test/electric/shapes/consumer_test.exs
@@ -597,6 +597,10 @@ defmodule Electric.Shapes.ConsumerTest do
         Map.get(ctx, :hibernate_after, 10_000)
       )
 
+      if not Map.get(ctx, :allow_subqueries, true) do
+        Electric.StackConfig.put(ctx.stack_id, :feature_flags, [])
+      end
+
       :ok
     end
 
@@ -671,6 +675,7 @@ defmodule Electric.Shapes.ConsumerTest do
                get_log_items_from_storage(LogOffset.last_before_real_offsets(), shape_storage)
     end
 
+    @tag allow_subqueries: false
     test "duplicate txn fragment handling is idempotent", ctx do
       {shape_handle, _} = ShapeCache.get_or_create_shape_handle(@shape1, ctx.stack_id)
       :started = ShapeCache.await_snapshot_start(shape_handle, ctx.stack_id)
@@ -908,7 +913,9 @@ defmodule Electric.Shapes.ConsumerTest do
       assert_receive {:flush_boundary_updated, ^tx_offset}
     end
 
-    @tag delay_snapshot_creation?: true, with_pure_file_storage_opts: [flush_period: 1]
+    @tag allow_subqueries: false,
+         delay_snapshot_creation?: true,
+         with_pure_file_storage_opts: [flush_period: 1]
     test "transaction fragments are buffered until snapshot xmin is known", ctx do
       register_as_replication_client(ctx.stack_id)
 
@@ -1119,7 +1126,9 @@ defmodule Electric.Shapes.ConsumerTest do
       assert {:ok, last_log_offset} == Storage.fetch_latest_offset(shape_storage)
     end
 
-    @tag pg_snapshot: {10, 13, [10]}, with_pure_file_storage_opts: [flush_period: 1]
+    @tag allow_subqueries: false,
+         pg_snapshot: {10, 13, [10]},
+         with_pure_file_storage_opts: [flush_period: 1]
     test "fragments that belong to transactions already included in the snapshot are skipped",
          ctx do
       register_as_replication_client(ctx.stack_id)
@@ -1535,7 +1544,7 @@ defmodule Electric.Shapes.ConsumerTest do
       assert [] == :ets.tab2list(table)
     end
 
-    @tag with_pure_file_storage_opts: [flush_period: 1]
+    @tag allow_subqueries: false, with_pure_file_storage_opts: [flush_period: 1]
     test "writes txn fragments to storage immediately but keeps txn boundaries when flushing",
          ctx do
       {shape_handle, _} = ShapeCache.get_or_create_shape_handle(@shape1, ctx.stack_id)


alco:
# Manual memory benchmark results

The tests below where done separately with electric@main and electric@this-branch. The database had a single table
```sql
CREATE TABLE IF NOT EXISTS public.users (
  id bigserial PRIMARY KEY,
  name TEXT NOT NULL,
  email TEXT NOT NULL                                                  
);
```

and Electric was initialized with 200 shapes, each shape having a `where` clause `N=N` where `N in 1..200`.

After creating the shapes and firing up Observer, I executed the following INSERT in the database:
```sql
insert into                              
  public.users(name, email)
    select
      'username_' || generate_series::text || '_' || repeat('x', 2000),
      'username_' || generate_series::text || '@example.com'
    from
      generate_series(1,10001);
```

## Memory charts

The total memory usage on `electric@main` grows almost to 9GB during transaction processing. For `electric@this-branch` the total memory usage stays below 500MB.

### electric@main
electric-main-mem-chart

### electric@this-branch
electric-fragment-mem-chart

## System stats comparison

For each run, I screenshotted the system stats in Observer both before the INSERT and after the memory usage had fallen back down (after GC and consumer hibernation).

The key thing to notice from these is that `electric@main` has residual process memory usage after the load testing which is not the case for `electric@this-branch`.

### electric@main
electric-main-stats

### electric@this-branch
electric-fragment-stats

alco:
@claude please review this PR. 

- Does it address the referenced issue?
- Do you see any obvious issue in the code?
- Lack of test coverage?
- Any gnarly or hairy code?

alco:
I've run the replication throughput benchmark on our cloud fleet with the following params, comparing `main` (i.e. `electricsql/electric:canary`) with this PR (pushed to Docker Hub as `alco/electric:fragment-direct-pr`):

```elixir
%{
  "row_size" => [150],
  "shape_count" => [5, 100, 1000, 2000],
  "tx_affects_shape" => [true, false],
  "tx_count" => [1000],
  "tx_row_count" => [90, 100, 120, 200, 220],
  "where_clause" => ["name = '\#{name}'", "name ILIKE '\#{name}%'"]
}
```

canvas

balegas:
Every microsecond counts!

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -455,33 +457,302 @@ defmodule Electric.Shapes.Consumer do
...
+
+  defp maybe_complete_pending_txn(%State{} = state, txn_fragment, persisted_changes) do
+    %{pending_txn: txn, writer: writer, shape: shape, shape_handle: shape_handle} = state
+
+    # Signal commit to storage to allow it to advance its internal txn offset
+    writer = ShapeCache.Storage.signal_txn_commit!(txn.xid, writer)
+
+    # Only notify if we actually wrote changes
+    if txn.num_changes > 0 do
+      notify_new_changes(state, persisted_changes, txn.last_log_offset)
```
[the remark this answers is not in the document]
robacourt: This ☝️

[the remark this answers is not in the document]
alco: When `notify_new_changes` is called from `maybe_complete_pending_txn`, that means `write_unit=txn_fragment` and so this shape has not subqueries and no materializers.

I could try and make it clearer in the code but I think it's a false concern.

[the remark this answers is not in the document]
robacourt: Not true. You're just checking if a shape has no dependencies. The innermost shape will have no dependencies (it IS the dependency) and will have a materializer.

e.g.
Outer shape: `where: id IN (SELECT id FROM accounts)` - dependencies: [Inner Shape]
Inner shape: `SELECT id FROM accounts` - dependencies: [] <-- this has a materilaizer

alco:
## Update (Feb 16)

Rebased on current main and pushed new commits addressing review feedback:

- **Split materializer and client notifications** (a93967e4b): Separated `notify_new_changes` into distinct functions for notifying the materializer vs. clients. This is needed to selectively notify the appropriate subscriber when processing individual txn fragments.

- **Use commit fragment offset for txn_offset_mapping and client notifications** (78a629acb): Fixes the flush alignment issue flagged in review — when the final commit fragment is filtered out, the mapping now uses the commit fragment's `last_log_offset` directly instead of the last relevant-change boundary. `PendingTxn` no longer needs to track `last_log_offset`.

- **Add `supports_txn_fragment_streaming?` callback to Storage behaviour** (4e630ab67): `InMemoryStorage` and `TestStorage` return `false`, `PureFileStorage` returns `true`. This allows runtime capability checking before enabling fragment streaming mode.

- **Downgrade `write_unit` to `:txn` when storage lacks fragment streaming support** (4714a7a02): On consumer init, validates that the storage backend supports txn fragment streaming. If unsupported (e.g. `InMemoryStorage`), falls back to full-transaction buffering with a warning log. This addresses the concern about `InMemoryStorage` crashing on fragment callbacks.

Still TODO: address @robacourt's feedback about shapes that are dependencies (innermost shapes with no dependencies themselves but that DO have a materializer) — these should probably use `write_unit=txn` to ensure the materializer gets the full transaction change set on commit.

Based on current implementation, Materializer sees changes from all txn fragments but when it then notifies the consumer with a `:materializer_changes` message in return, Consumer may still be assuming that a complete transaction has been processed.

alco:
## Latest changes (3 new commits since last push)

**Defer Materializer subscriber notifications until commit fragment**

Previously, when `write_unit=:txn_fragment`, the inner Consumer called `Materializer.new_changes()` per fragment, and the Materializer immediately sent `:materializer_changes` (move-in/move-out events) to the outer Consumer after each fragment. This could cause redundant intermediate notifications within a single transaction.

Changes:
- Added a `commit:` option to `Materializer.new_changes/3` (defaults to `true` for backward compatibility). When `commit: false`, events are accumulated in a new `pending_events` field on the Materializer state without notifying subscribers. When `commit: true`, accumulated events are flushed to subscribers.
- The fragment-streaming path in `consumer.ex` passes `commit: false` for every fragment. The flush happens via `notify_new_changes` in `maybe_complete_pending_txn` which uses `commit: true` (the default).
- Events from multiple fragments are merged by prepending newer events to older ones, matching the existing reverse-chronological ordering convention used by `increment_value`/`decrement_value`.
- Added a test verifying correct event ordering across uncommitted fragments.
- Moved `signal_txn_commit!` inside the `num_changes > 0` branch since there's no need to signal storage when no changes were written.

alco:
Regarding an erlier agentic comment about the potential impact of direct txn fragment writing to storage on `PureFileStorage.trim_log!`, I've asked Claude to analyse it and here's the result - https://gist.github.com/alco/588b3d7d7665f63d768d5f1b611499c7#file-fragment_storage_safety_analysis-md.

### TL;DR: Fragment Streaming Storage Safety

All 7 crash/consistency scenarios are **safe**. The three-offset architecture (`last_seen_offset` / `last_seen_txn_offset` / `last_persisted_txn_offset`) provides the core safety guarantee:

- **`trim_log!/2` correctly truncates uncommitted fragments** — it uses `last_persisted_txn_offset` (only advanced after a complete txn is finalized + flushed) to remove both chunk index entries and log file bytes beyond the committed boundary.
- **Readers never see uncommitted data** — `stream_main_log` caps reads at `last_seen_txn_offset`, which isn't advanced until `signal_txn_commit!` calls `finalize_txn`.
- **ETS consistency is maintained** — fragment entries stay in an in-memory buffer until `finalize_txn`; the disk/ETS split read strategy handles mid-transaction flushes correctly.
- **Compaction has one theoretical edge case** (transaction spanning 30+ MB creating 3+ chunks) where the compacted file could retain uncommitted data after a crash, but readers are still bounded by the txn offset, so no data corruption occurs. Practically, this is a near-zero likelihood scenario.

No code changes are needed — the existing invariants hold correctly with the fragment write path.

alco:
@robacourt That's dope! It has found one failure already after running for ~10sec. I'll look into fixing it and see if it can discover anything else.

alco:
Oops, actually the issue I've found is in `main` - https://github.com/electric-sql/electric/issues/3854. So it's difficult to see whether my PR has introduced new problems before that one is fixed.

alco:
> Oops, actually the issue I've found is in `main` - #3854. So it's difficult to see whether my PR has introduced new problems before that one is fixed.

Nevermind that issue, I've just closed it since its discription was hallucinated from incorrect facts and faulty assumptions.

I've played a bit with the oracle tests on `main` and the suprising thing I've discovered is that even at 2 shapes and 15 mutations per transaction it takes 2-10 runs to find a failing case. The more you increase the number of mutations the quicker it fails:

```
$ MUTATIONS_PER_TXN=15 SHAPE_COUNT=2 TXN_COUNT=10 \
            mix test --only oracle test/integration/oracle_property_test.exs
Running ExUnit with seed: 921473, max_cases: 32
Excluding tags: [:test, :slow]
Including tags: [:oracle]

[oracle] iteration_seed=1674
[oracle] Starting 2 shapes

=== SHAPES ===
  shape_1: value NOT LIKE 'v%'
  shape_2: level_3_id IN (SELECT id FROM level_3 WHERE active = false)

=== TRANSACTIONS (10 txns, 150 mutations) ===

[...]

[oracle]   txn_10 shape=shape_1 (unchanged) PASS
[oracle] View mismatch in step=txn_10 shape=shape_2 where=level_3_id IN (SELECT id FROM level_3 WHERE active = false) view_changed?=true

00:20:40.034 pid=<0.335.0> [error] GenServer #PID<0.335.0> terminating
** (ExUnit.AssertionError) 

Assertion with == failed
code:  assert materialized == oracle_after
left:  [%{"id" => "l4-11", "level_3_id" => "l3-4", "value" => "v561"}, %{"id" => "l4-14", "level_3_id" => "l3-4", "value" => "v395"}, %{"id" => "l4-16", "level_3_id" => "l3-1", "value" => "v15"}, %{"id" => "l4-19", "level_3_id" => "l3-1", "value" => "v104"}, %{"id" => "l4-6", "level_3_id" => "l3-1", "value" => "v443"}, %{"id" => "l4-9", "level_3_id" => "l3-4", "value" => "v382"}]
right: [%{"id" => "l4-11", "level_3_id" => "l3-4", "value" => "v561"}, %{"id" => "l4-12", "level_3_id" => "l3-2", "value" => "v802"}, %{"id" => "l4-14", "level_3_id" => "l3-4", "value" => "v395"}, %{"id" => "l4-16", "level_3_id" => "l3-1", "value" => "v15"}, %{"id" => "l4-17", "level_3_id" => "l3-2", "value" => "v16"}, %{"id" => "l4-19", "level_3_id" => "l3-1", "value" => "v104"}, %{"id" => "l4-4", "level_3_id" => "l3-2", "value" => "v331"}, %{"id" => "l4-6", "level_3_id" => "l3-1", "value" => "v443"}, %{"id" => "l4-9", "level_3_id" => "l3-4", "value" => "v382"}]

    (ex_unit 1.19.1) lib/ex_unit/assertions.ex:454: ExUnit.Assertions.assert/2
    (electric 1.4.4) test/support/oracle_harness/shape_checker.ex:277: Support.OracleHarness.ShapeChecker.assert_consistent!/4
    (electric 1.4.4) test/support/oracle_harness/shape_checker.ex:142: Support.OracleHarness.ShapeChecker.handle_call/3
    (stdlib 7.1) gen_server.erl:2470: :gen_server.try_handle_call/4
    (stdlib 7.1) gen_server.erl:2499: :gen_server.handle_msg/3
    (stdlib 7.1) proc_lib.erl:333: :proc_lib.init_p_do_apply/3
Last message: {:EXIT, #PID<0.324.0>, {%ExUnit.AssertionError{left: [%{"id" => "l4-11", "level_3_id" => "l3-4", "value" => "v561"}, %{"id" => "l4-14", "level_3_id" => "l3-4", "value" => "v395"}, %{"id" => "l4-16", "level_3_id" => "l3-1", "value" => "v15"}, %{"id" => "l4-19", "level_3_id" => "l3-1", "value" => "v104"}, %{"id" => "l4-6", "level_3_id" => "l3-1", "value" => "v443"}, %{"id" => "l4-9", "level_3_id" => "l3-4", "value" => "v382"}], right: [%{"id" => "l4-11", "level_3_id" => "l3-4", "value" => "v561"}, %{"id" => "l4-12", "level_3_id" => "l3-2", "value" => "v802"}, %{"id" => "l4-14", "level_3_id" => "l3-4", "value" => "v395"}, %{"id" => "l4-16", "level_3_id" => "l3-1", "value" => "v15"}, %{"id" => "l4-17", "level_3_id" => "l3-2", "value" => "v16"}, %{"id" => "l4-19", "level_3_id" => "l3-1", "value" => "v104"}, %{"id" => "l4-4", "level_3_id" => "l3-2", "value" => "v331"}, %{"id" => "l4-6", "level_3_id" => "l3-1", "value" => "v443"}, %{"id" => "l4-9", "level_3_id" => "l3-4", "value" => "v382"}], message: "Assertion with == failed", expr: {:assert, [line: 277], [{:==, [line: 277], [{:materialized, [line: 277], nil}, {:oracle_after, [line: 277], nil}]}]}, args: :ex_unit_no_meaningful_value, doctest: :ex_unit_no_meaningful_value, context: :==}, [{ExUnit.Assertions, :assert, 2, [file: ~c"lib/ex_unit/assertions.ex", line: 454]}, {Support.OracleHarness.ShapeChecker, ...}, ...]}}


  1) test shapes with generated where clauses and mutations (Electric.Integration.OraclePropertyTest)
     test/integration/oracle_property_test.exs:56
     ** (EXIT from #PID<0.324.0>) an exception was raised:
     
          Assertion with == failed
          code:  assert materialized == oracle_after
          left:  [
                   %{"id" => "l4-11", "level_3_id" => "l3-4", "value" => "v561"},
                   %{"id" => "l4-14", "level_3_id" => "l3-4", "value" => "v395"},
                   %{"id" => "l4-16", "level_3_id" => "l3-1", "value" => "v15"},
                   %{"id" => "l4-19", "level_3_id" => "l3-1", "value" => "v104"},
                   %{"id" => "l4-6", "level_3_id" => "l3-1", "value" => "v443"},
                   %{"id" => "l4-9", "level_3_id" => "l3-4", "value" => "v382"}
                 ]
          right: [
                   %{"id" => "l4-11", "level_3_id" => "l3-4", "value" => "v561"},
                   %{"id" => "l4-12", "level_3_id" => "l3-2", "value" => "v802"},
                   %{"id" => "l4-14", "level_3_id" => "l3-4", "value" => "v395"},
                   %{"id" => "l4-16", "level_3_id" => "l3-1", "value" => "v15"},
                   %{"id" => "l4-17", "level_3_id" => "l3-2", "value" => "v16"},
                   %{"id" => "l4-19", "level_3_id" => "l3-1", "value" => "v104"},
                   %{"id" => "l4-4", "level_3_id" => "l3-2", "value" => "v331"},
                   %{"id" => "l4-6", "level_3_id" => "l3-1", "value" => "v443"},
                   %{"id" => "l4-9", "level_3_id" => "l3-4", "value" => "v382"}
                 ]
          stacktrace:
            (ex_unit 1.19.1) lib/ex_unit/assertions.ex:454: ExUnit.Assertions.assert/2
            (electric 1.4.4) test/support/oracle_harness/shape_checker.ex:277: Support.OracleHarness.ShapeChecker.assert_consistent!/4
            (electric 1.4.4) test/support/oracle_harness/shape_checker.ex:142: Support.OracleHarness.ShapeChecker.handle_call/3
            (stdlib 7.1) gen_server.erl:2470: :gen_server.try_handle_call/4
            (stdlib 7.1) gen_server.erl:2499: :gen_server.handle_msg/3
            (stdlib 7.1) proc_lib.erl:333: :proc_lib.init_p_do_apply/3
```

At this point my main concern is: is it a bug in `main` or a bug in the test harness?

balegas:
> At this point my main concern is: is it a bug in main or a bug in the test harness?

We should be able to reproduce the same issue in main if that's the case. Let's do that experiment to be sure.

alco:
@robacourt Pushed https://github.com/electric-sql/electric/pull/3783/commits/751cf240461fa4d3b428f46aac88535520ef18fb to correctly incorporate your recent addition of cancelling out move-ins/move-outs.

alco:
I've just run oracle tests with max_batch_size=5 and mutations_per_txn=15 and can relibably reproduce this issue:
```
Poll error for shape=shape_9 where=(level_3_id = 'l3-5') OR (value BETWEEN 'v1' AND 'v10'): %Electric.Client.Error{message: %{"errors" => %{"offset" => ["out of bounds for this shape"]}, "message" => "Invalid request"}, resp: %Electric.Client.Fetch.Response{status: 400, last_offset: nil, shape_handle: nil, schema: nil, next_cursor: nil, request_timestamp: ~U[2026-02-18 18:57:16.838842Z], body: %{"errors" => %{"offset" => ["out of bounds for this shape"]}, "message" => "Invalid request"}, headers: %{"access-control-allow-methods" => ["GET, POST, HEAD, DELETE, OPTIONS"], "access-control-allow-origin" => ["*"], "access-control-expose-headers" => ["electric-cursor,electric-handle,electric-offset,electric-schema,electric-up-to-date,electric-internal-known-error,retry-after"], "cache-control" => ["no-store"], "content-type" => ["application/json; charset=utf-8"], "date" => ["Wed, 18 Feb 2026 18:57:16 GMT"], "electric-server" => ["ElectricSQL/1.4.4"], "surrogate-control" => ["no-store"], "vary" => ["accept-encoding"], "x-request-id" => ["GJVsPM1fclpQUIgAAEJC"]}}}
```

The key bit is `%Electric.Client.Error{message: %{"errors" => %{"offset" => ["out of bounds for this shape"]}, "message" => "Invalid request"}`

So apparently Electric returns incorrect offset that's not at transaction boundary.

Silver lining is that if I set `write_unit: :write_unit_txn` in this branch, that is stick to buffering txn fragments in memory, the oracle test harness only finds inconsistency errors similar to those that can be reproduced in `main`.

I may be able to fix the handling of multiple txn fragments for shapes without subqueries first and we ship those. Then deal with subqueries.

balegas:
:( if it's easy to reproduce, we should be able to figure the root cause. There is no rush to ship this PR, so let's do it properly when we know the fix.

alco:
I have pushed a fix for the issue described in https://github.com/electric-sql/electric/pull/3783#issuecomment-3922656123 and have set `write_unit=txn` even for leaf Consumer processes - those that are started for the innermost shape of a shape with subqueries.

Still it's impossible to be sure this branch doesn't add new failures on top of those already observed in `main` until we figure out the root cause of those. [This](https://github.com/electric-sql/electric/issues/3892) is just one example of different failures that are readily reproducible with oracle tests in `main`.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -456,64 +456,322 @@ defmodule Electric.Shapes.Consumer do
...
+      txn.consider_flushed? or fragment_already_processed?(txn_fragment, state) ->
+        skip_txn_fragment(state, txn_fragment)
+
+      # With write_unit=txn all fragments are buffered until the Commit change is seen. At that
+      # point, a transaction struct is produced from the buffered fragments and is written to
+      # storage.
+      state.write_unit == State.write_unit_txn() ->
+        {txns, transaction_builder} =
+          TransactionBuilder.build(txn_fragment, state.transaction_builder)
+
```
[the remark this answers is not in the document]
alco: By the way, the fact that a materializer can subscribe to consumer mid-transaction processing might also be causing [this issue](https://github.com/electric-sql/electric/issues/3891) (yet to be verified by a human).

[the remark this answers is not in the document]
alco: Fixed this by instead passing an `is_subquery_shape?` flag from `ShapeCache` into `Consumer.start_link` for non-root shapes that are part of a shape-with-subqueries hierarchy.

alco:
In order to get this PR merged, I have disabled individual txn fragment writing to storage when the `allow_subqueries` feature flag is present in the stack config.

I will undo that and will continue fixing edge cases for shapes with subqueries in this [follow-up PR](https://github.com/electric-sql/electric/pull/3906).

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -403,28 +409,37 @@ defmodule Electric.ShapeCache do
...
+  defp build_shape_dependencies([{handle, shape} | rest], root?, known) do
+    {siblings, known} = build_shape_dependencies(rest, false, MapSet.put(known, handle))
 
     {descendents, known} =
       Enum.zip(shape.shape_dependencies_handles, shape.shape_dependencies)
       |> Enum.reject(fn {handle, _shape} -> MapSet.member?(known, handle) end)
-      |> build_shape_dependencies(known)
+      |> build_shape_dependencies(false, known)
+
+    # Any inner shape of a root shape with subqueries must pass the is_subquery_shape? option
```
robacourt, line 433: I'm not sure why being a root is a special case. Any dependency should have the `is_subquery_shape?: true`

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -403,28 +409,37 @@ defmodule Electric.ShapeCache do
...
-  defp build_shape_dependencies([{handle, shape} | rest], known) do
-    {siblings, known} = build_shape_dependencies(rest, MapSet.put(known, handle))
+  defp build_shape_dependencies([{handle, shape} | rest], root?, known) do
+    {siblings, known} = build_shape_dependencies(rest, false, MapSet.put(known, handle))
 
     {descendents, known} =
       Enum.zip(shape.shape_dependencies_handles, shape.shape_dependencies)
       |> Enum.reject(fn {handle, _shape} -> MapSet.member?(known, handle) end)
-      |> build_shape_dependencies(known)
+      |> build_shape_dependencies(false, known)
```
robacourt, line 431: Can a shape become a subquery shape after it's created? So if `SELECT id FROM products` was created as a shape first and then `id IN (SELECT id FROM products)` later, would the 2nd shape use the first as the subquery shape? If so setting `is_subquery_shape?` on creation breaks. If this is the case personally I wouldn't be against ensuring that shapes and subquery shapes were distinct to prevent this from happening, it's only an optimisation to reuse shapes.

packages/sync-service/lib/electric/shape_cache.ex
robacourt: Oh, ignore me. You're just excluding the root from being `is_subquery_shape?`. Fair enough.

robacourt: The logic in my head seems more straightforward that you tell a shape whether it's a subquery shape or not, when building the root thats false and when it builds it's dependencies it's true. No need to even mention the root. Also `is_subquery_shape?` could always be set rather than some additional conditional logic. But this is a very minor nit!

packages/sync-service/lib/electric/shape_cache.ex
alco: I can address both your comments here.

You are right, a shape may start independent but then have a materializer subscribe to it at a later point in time. I attempted to identify inner shapes using the `is_subquery_shape?` option passed at consumer startup but then determined that it wouldn't be sufficient. If we want to enable txn fragment processing for stacks that have `allow_subqueries`, we need to make it work even for shape with subqueries properly, at least in the case where `shape.shape_dependencies == []`. At that point, the `is_subquery_shape?` flag becomes superfluous and will be removed from the codebase.

robacourt: Ok, so your plan for this PR is to not enable txn fragment processing for stacks that have allow_subqueries?

alco: Exactly.

Once this is merged, direct txn fragment writing to storage will only be enabled for stack that _don't_ have subqueries.

Issue #3787, opened 2026-01-28
ShapeLogCollector memory growth to 3.5GB causing OOM crash on maxwell


robacourt:
## Summary

On 27 Jan 2026 at ~18:09-18:11 UTC, the ShapeLogCollector on the maxwell production instance grew to 3.5GB of memory, causing the server to crash. Telemetry confirms the ShapeLogCollector was the primary memory consumer.

## Timeline

- **18:10:09.998** - Two new shapes with subquery dependencies created:
  - `97489818-1769537409997837` (offers shape with projects dependency)
  - `30042127-1769537409998609` (offer_items shape depending on offers)
- **18:10:10.437** - Materializer for shape `97489818` crashes with "Key already exists" error
- **18:10:10.449** - ShapeLogCollector crashes with `FunctionClauseError` in `DependencyLayers.add_after_dependencies/3`
- **18:10:10.45x** - Cascade of 80+ Materializer/Consumer shutdowns begins
- **18:10:29.702** - Memory alarm triggered: `{:process_memory_high_watermark, #PID<0.179627.0>}`
- **18:12** - Container restarts

### Bug 1: Materializer "Key already exists" race condition

The Materializer received a `NewRecord` for a key that already existed in its index. This happened because:

1. Shape `97489818` was created for `public.offers` with a subquery dependency on `public.projects`
2. The Materializer started, subscribed to the Consumer, then began reading from storage
3. A transaction arrived with a move-in event for offer `d3c8d8a5-5060-4a36-a67d-240de0c95a88`
4. The record was already in the snapshot (matched via `is_template = true` OR the subquery), AND was delivered via replication with `move_tags`
5. The Materializer's `apply_changes` raised at line 317:

```elixir
if is_map_key(index, key), do: raise("Key #{key} already exists")
```

The race window exists because the Materializer subscribes to the Consumer BEFORE reading from storage:

```elixir
# In handle_continue(:start_materializer, ...)
Consumer.subscribe_materializer(stack_id, shape_handle, self())  # <- Subscribes first
{:noreply, state, {:continue, {:read_stream, shape_storage}}}    # <- Then reads storage
```

### Bug 2: DependencyLayers missing function clause

When shape `30042127` (which depends on `97489818`) tried to register after `97489818` crashed:

```
Electric.Shapes.DependencyLayers.add_after_dependencies([], "30042127-...", MapSet.new(["97489818-..."]))
```

[...]

PR #3789, opened 2026-01-28
Fix Materializer to handle duplicate records idempotently

robacourt:
## Summary

Partial fix for #3787 - addresses the snapshot/replication race condition in the Materializer. The Dependency layers issue will be fixed in a separate PR.

### The race condition

The Materializer subscribes to the Consumer **before** reading from storage:

```elixir
# In handle_continue(:start_materializer, ...)
Consumer.subscribe_materializer(stack_id, shape_handle, self())  # <- Subscribes first
{:noreply, state, {:continue, {:read_stream, shape_storage}}}    # <- Then reads storage
```

During the window between subscribing and reading, any changes that arrive via `Consumer.new_changes()` will be delivered to the Materializer. If those changes include records that are also in the snapshot being read, the Materializer receives duplicates and crashes.

### Production example

From the maxwell instance crash on 27 Jan 2026:

```
** (RuntimeError) Key "public"."offers"/"d3c8d8a5-5060-4a36-a67d-240de0c95a88" already exists
```

The transaction that triggered it:
```elixir
%Electric.Replication.Changes.NewRecord{
  relation: {"public", "offers"},
  record: %{"id" => "d3c8d8a5-5060-4a36-a67d-240de0c95a88"},
  key: "\"public\".\"offers\"/\"d3c8d8a5-5060-4a36-a67d-240de0c95a88\"",
  move_tags: ["e12422d3af57a36d01a50b4645a517e4"]  # <- Move-in event
}
```

The record was already in the snapshot (matched via `is_template = true` OR the subquery), AND was delivered via replication with `move_tags` as a move-in event.

### The fix

Make the Materializer idempotent - when a duplicate record arrives, skip the insert but still track move_tags so subsequent move_out events work correctly.

## Test plan

- [x] Added tests verifying duplicate records are handled idempotently
- [x] Verified move_tags are tracked from duplicates so subsequent move_out works correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of duplicate records appearing in both snapshot and replication streams to prevent errors and ensure idempotent processing.


robacourt:
@coderabbitai review

icehaunter (changes requested):
Crashing versions of those functions were used explicitly because "not seeing an out of order row" is an assumed invariant
One case which led to the same crash was another bug which I've addressed.

When another "newRecord" arrives with some move tags, you don't know whether it's from snapshot, or actual move in or what, and if they are out of order/duplicated, you don't know which move tag set is valid, I don't think merging it is correct.

I would much rather keep the crash as an invariant insurance and fix only the underlying bug

PR #3790, opened 2026-01-28
Fix DependencyLayers race condition crash on missing dependencies

### LINKED ISSUE #3788: DependencyLayers crashes when dependency shape is removed before dependent shape is added
## Summary

When a shape with dependencies tries to register, but the dependency shapes have been removed (due to crash or cleanup), `DependencyLayers.add_after_dependencies/3` crashes with `FunctionClauseError` due to a missing function clause.

This is **Bug 2** from #3787.

## Root Cause

The race condition occurs because:

1. Dependency shape (A) is added to DependencyLayers
2. Dependent shape (B) starts its Consumer, checks `all_materializers_alive?` → passes
3. Shape A's Materializer crashes (e.g., due to #3787 Bug 1)
4. Shape A's Consumer terminates and calls `remove_shape`
5. Shape B's Consumer calls `add_shape`
6. RequestBatcher batches both operations together
7. ShapeLogCollector processes the batch - **removals first, then additions**
8. Shape A is removed from layers → layers becomes `[[]]` (empty set)
9. Shape B tries to add with dependency on A → CRASH

The crash happens because `add_after_dependencies/3` has no clause for when layers is exhausted but dependencies are still unfound:

```elixir
# This clause only matches when deps_to_find is empty
defp add_after_dependencies([], shape_handle, deps_to_find) when map_size(deps_to_find) == 0 do
  [MapSet.new([shape_handle])]
end

# MISSING: clause for when layers is [] but deps_to_find is NOT empty
```

## Evidence from #3787

```
** (RuntimeError) Key "public"."offers"/"..." already exists

18:10:10.449 [error] GenServer ShapeLogCollector terminating  
** (FunctionClauseError) no function clause matching in Electric.Shapes.DependencyLayers.add_after_dependencies/3
    (electric) lib/electric/shapes/dependency_layers.ex
    args: [[], "30042127-...", MapSet.new(["97489818-..."])]
```

The 12ms gap between crashes shows the sequence: Materializer crash → Consumer termination → remove_shape batched with add_shape → removal processed first → crash on addition.

## Proposed Fix

Instead of crashing, `add_dependency` should return an error tuple when dependencies are missing, allowing the caller to handle it gracefully:

```elixir
def add_dependency(layers, shape, shape_handle) do
  case shape.shape_dependencies_handles do
    [] ->
      {:ok, add_to_first_layer(layers, shape_handle)}

    dependency_handles ->
      case add_after_dependencies(layers, shape_handle, MapSet.new(dependency_handles)) do
        {:ok, layers} -> {:ok, layers}
        {:error, missing} -> {:error, {:missing_dependencies, missing}}
      end
  end
end

defp add_after_dependencies([layer | rest], shape_handle, deps_to_find) do
  remaining_deps = MapSet.difference(deps_to_find, layer)
  if MapSet.size(remaining_deps) == 0 do
    {:ok, [layer | add_to_first_layer(rest, shape_handle)]}
  else
    case add_after_dependencies(rest, shape_handle, remaining_deps) do
      {:ok, rest} -> {:ok, [layer | rest]}
      {:error, _} = error -> error
    end
  end
end

defp add_after_dependencies([], shape_handle, deps_to_find) when map_size(deps_to_find) == 0 do
  {:ok, [MapSet.new([shape_handle])]}
end

defp add_after_dependencies([], _shape_handle, deps_to_find) do
  {:error, deps_to_find}
end
```

The caller in `ShapeLogCollector` would then handle the error by failing the shape registration gracefully rather than crashing the entire collector.

## Impact

When this bug triggers, it crashes the ShapeLogCollector, which caused the cascade failure described in #3787 (80+ Consumer/Materializer shutdowns, 3.5GB memory growth, OOM crash).

## Related

- Parent issue: #3787
- Bug 1 (Materializer race condition) will be tracked separately

robacourt:
## Summary

Fixes #3788 — when a dependency shape is removed (due to crash/cleanup) before its dependent shape finishes registering, `DependencyLayers.add_after_dependencies/3` crashed with `FunctionClauseError`, taking down the `ShapeLogCollector` and cascading into OOM failures.

This is the companion fix to #3789, which addressed the Materializer duplicate-key crash that triggered the dependency removal in the first place.

### The race condition

1. Dependency shape A is added to DependencyLayers
2. Dependent shape B starts its Consumer, checks `all_materializers_alive?` → passes
3. Shape A's Materializer crashes (e.g. duplicate key, fixed by #3789)
4. Shape A's Consumer terminates and calls `remove_shape`
5. Shape B's Consumer calls `add_shape`
6. `RequestBatcher` batches both operations together
7. `ShapeLogCollector` processes the batch — **removals first, then additions**
8. Shape A is removed from layers → layers becomes `[[]]`
9. Shape B tries to add with dependency on A → **CRASH**

### ECS logs from production (27 Jan 2026)

```
** (RuntimeError) Key "public"."offers"/"d3c8d8a5-..." already exists

18:10:10.449 [error] GenServer ShapeLogCollector terminating
** (FunctionClauseError) no function clause matching in Electric.Shapes.DependencyLayers.add_after_dependencies/3
    (electric) lib/electric/shapes/dependency_layers.ex
    args: [[], "30042127-...", MapSet.new(["97489818-..."])]
```

The 12ms gap between crashes shows the sequence: Materializer crash → Consumer termination → `remove_shape` batched with `add_shape` → removal processed first → crash on addition.

### Changes

- **`DependencyLayers.add_dependency/3`** now returns `{:ok, layers}` or `{:error, {:missing_dependencies, missing}}` instead of crashing
- **`add_after_dependencies/3`** gets a new clause for exhausted layers with unfound dependencies
- **`ShapeLogCollector`** handles the error gracefully — logs a warning and fails the shape registration instead of crashing
- **`handle_continue(:restore_shapes)`** pattern-matches `{:ok, layers}` (crashes on error, which is correct since restored shapes should always have valid dependencies)

## Test plan

- [x] Updated existing `DependencyLayersTest` tests for new return type
- [x] Added error handling tests: missing dependency, removed dependency, partially missing dependencies, ok-tuple for no-dependency shapes
- [x] All 8 dependency layers tests pass
- [x] All 24 shape log collector tests pass

## Release Notes

* **Bug Fixes**
  * Fixed a race condition that caused crashes when shape dependencies were unavailable during registration or restoration
  * Improved error handling for missing dependencies with graceful recovery instead of system failures

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -251,18 +251,26 @@ defmodule Electric.Replication.ShapeLogCollector do
           |> Enum.reduce(
             {state.partitions, state.event_router, state.dependency_layers, 0},
             fn {shape_handle, shape}, {partitions, event_router, layers, count} ->
-              {:ok, partitions} = Partitions.add_shape(partitions, shape_handle, shape)
+              # Check dependencies first - if a parent shape failed to restore,
+              # we should skip this shape (and its children will also be skipped)
+              case DependencyLayers.add_dependency(layers, shape, shape_handle) do
+                {:ok, layers} ->
+                  {:ok, partitions} = Partitions.add_shape(partitions, shape_handle, shape)
 
-              # Crash if dependencies are missing during restore - all persisted shapes
-              # should have their dependencies present since they were valid when saved.
-              {:ok, layers} = DependencyLayers.add_dependency(layers, shape, shape_handle)
+                  {
+                    partitions,
+                    EventRouter.add_shape(event_router, shape_handle, shape),
+                    layers,
+                    count + 1
+                  }
 
-              {
-                partitions,
-                EventRouter.add_shape(event_router, shape_handle, shape),
-                layers,
-                count + 1
-              }
+                {:error, {:missing_dependencies, missing_deps}} ->
+                  Logger.warning(
+                    "Skipping shape #{shape_handle} during restore: missing dependencies #{inspect(MapSet.to_list(missing_deps))}"
+                  )
+
+                  {partitions, event_router, layers, count}
+              end
             end
           )
 


robacourt:
@coderabbitai review

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -253,10 +253,14 @@ defmodule Electric.Replication.ShapeLogCollector do
             fn {shape_handle, shape}, {partitions, event_router, layers, count} ->
               {:ok, partitions} = Partitions.add_shape(partitions, shape_handle, shape)
 
+              # Crash if dependencies are missing during restore - all persisted shapes
+              # should have their dependencies present since they were valid when saved.
+              {:ok, layers} = DependencyLayers.add_dependency(layers, shape, shape_handle)
```
robacourt: Is this a valid assumption?

icehaunter: I'd say yes but I'd prefer a different behaviour - there are cases where one shape might not get restored for one reason or another, and it seems like we should be dropping their children subtree, not saying that all persisted shapes have failed

packages/sync-service/lib/electric/shapes/dependency_layers.ex
```diff
@@ -17,14 +17,23 @@ defmodule Electric.Shapes.DependencyLayers do
...
 
+  # All dependencies found, add shape to new layer
   defp add_after_dependencies([], shape_handle, deps_to_find) when map_size(deps_to_find) == 0 do
-    [MapSet.new([shape_handle])]
+    {:ok, [MapSet.new([shape_handle])]}
+  end
+
+  # Dependencies are missing - return error with the missing dependency handles
+  defp add_after_dependencies([], _shape_handle, deps_to_find) do
+    {:error, {:missing_dependencies, deps_to_find}}
```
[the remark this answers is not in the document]
robacourt: good spot. fixed

packages/sync-service/lib/electric/replication/shape_log_collector.ex
robacourt: good idea @icehaunter . done

PR #3776, opened 2026-01-27
RFC: Move-in/Move-out Handling for Shapes with Subqueries

robacourt:
RFC: https://github.com/electric-sql/electric/blob/mob/arbitrary-boolean-expressions-with-subqueries/docs/rfcs/arbitrary-boolean-expressions-with-subqueries.md
PR: https://github.com/electric-sql/electric/pull/3791

PR #3791, opened 2026-01-28
feat: Support arbitrary boolean expressions with subqueries (OR, NOT)

robacourt:
## Summary

This PR implements [RFC "Arbitrary Boolean Expressions with Subqueries"](https://github.com/electric-sql/electric/pull/3776) which extends Electric's subquery support from single `IN (SELECT ...)` conditions to arbitrary boolean expressions with OR, NOT, and multiple subqueries.

### What's Changed
- **OR with multiple subqueries**: `WHERE project_id IN (SELECT ...) OR assigned_to IN (SELECT ...)`
- **NOT with subqueries**: `WHERE project_id NOT IN (SELECT ...)`
- **Complex expressions**: `WHERE (a IN sq1 AND b='x') OR c NOT IN sq2`

### Key Implementation Details

1. **DNF Decomposer** (`lib/electric/replication/eval/decomposer.ex`)
   - Converts WHERE clause AST to Disjunctive Normal Form
   - Applies De Morgan's laws for NOT handling
   - Assigns positions to atomic conditions for tracking

2. **Active Conditions** (`lib/electric/shapes/where_clause.ex`)
   - Computes `active_conditions` array indicating which atomic conditions are satisfied
   - Evaluates DNF against active conditions to determine inclusion
   - Used for both replication stream changes and initial snapshots

3. **Move-in/Move-out Handling** (`lib/electric/shapes/consumer/move_handling.ex`)
   - Position-based move-in/move-out with correct NOT inversion
   - Move-in to negated position triggers move-out
   - Move-out from negated position triggers move-in query
   - Deduplication logic prevents duplicate inserts for rows matching multiple disjuncts

4. **Message Format** (`lib/electric/log_items.ex`, `lib/electric/shapes/querying.ex`)
   - `active_conditions` array included in row message headers
   - SQL generation for initial snapshots includes `active_conditions`
   - Updated elixir-client to parse `active_conditions` field

## Test plan

- [x] All 1389 existing tests pass
- [x] 21 new integration tests for OR/NOT with subqueries (`test/electric/plug/subquery_router_test.exs`)
- [x] DNF decomposer tests (`test/electric/replication/eval/decomposer_test.exs`)
- [x] WhereClause tests for active conditions (`test/electric/shapes/where_clause_test.exs`)
- [x] Updated router tests no longer expect 409 for OR/NOT shapes

### Test Coverage
- Initial snapshot with OR subqueries
- Move-in/move-out with OR subqueries
- NOT IN shapes with correct inversion
- Row matching both disjuncts (deduplication)
- Row removed when all disjuncts become false
- Complex De Morgan expressions

## Related

- RFC: `docs/rfcs/arbitrary-boolean-expressions-with-subqueries.md`
- Implementation Plan: `packages/sync-service/IMPLEMENTATION_PLAN.md`

### DIFF
--- a/.changeset/arbitrary-boolean-expressions-with-subqueries.md
+++ b/.changeset/arbitrary-boolean-expressions-with-subqueries.md
@@ -1,0 +1,19 @@
+---
+"@core/sync-service": minor
+"@core/elixir-client": minor
+---
+
+Add support for arbitrary boolean expressions with subqueries.
+
+Previously, WHERE clauses with OR or NOT combined with subqueries would cause shape invalidation. This update implements RFC "Arbitrary Boolean Expressions with Subqueries" which enables:
+
+- OR with multiple subqueries: `WHERE project_id IN (SELECT ...) OR assigned_to IN (SELECT ...)`
+- NOT with subqueries: `WHERE project_id NOT IN (SELECT ...)`
+- Complex expressions: `WHERE (a IN sq1 AND b='x') OR c NOT IN sq2`
+
+Key features:
+- DNF (Disjunctive Normal Form) decomposition for WHERE clause analysis
+- `active_conditions` array in row messages indicating which atomic conditions are satisfied
+- Position-based move-in/move-out handling that correctly inverts behavior for negated positions
+- Deduplication logic to prevent duplicate inserts when rows match multiple disjuncts
+- Updated elixir-client to parse `active_conditions` field in message headers
--- a/docs/rfcs/arbitrary-boolean-expressions-with-subqueries.md
+++ b/docs/rfcs/arbitrary-boolean-expressions-with-subqueries.md
@@ -1,0 +1,627 @@
+---
+title: "RFC: Arbitrary Boolean Expressions with Subqueries"
+version: "1.0"
+status: draft
+owner: rob
+contributors:
+  - ilia
+  - kev
+created: 2026-01-27
+last_updated: 2026-02-16
+prd: N/A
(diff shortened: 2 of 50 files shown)


robacourt:
@coderabbitai review

kevin-dp:
Hi @robacourt, while implementing the DNF/active_conditions support in the ts/db client, I noticed that `tag_tracker.ex` stores `disjunct_positions` per-key in `key_data` and re-derives it every time a message with new tags arrives:

```elixir
disjunct_positions =
  case derive_disjunct_positions(normalized_new) do
    [] -> current_data && current_data.disjunct_positions
    positions -> positions
  end
```

This is redundant work — `disjunct_positions` is determined solely by the shape's WHERE clause (the DNF decomposition structure), which is fixed for the lifetime of a shape. Every row has the same tag structure; the hash values differ per row but which positions are `nil` vs participating is always identical. So there's no need to store it per-key or recompute it on each tagged message.

In the TS client we derive it once from the first tagged message and store it as a single variable shared across all rows. Might be worth doing the same in the Elixir client — just a single `disjunct_positions` field alongside `tag_to_keys`/`key_data` instead of inside each key's data, derived once on the first insert.

kevin-dp:
Also, i think it would be good to add the unit tests we have in ts/db for this feature (cf. https://github.com/TanStack/db/pull/1270/changes#diff-286083ece2106d963fcda491df8579332e28dd8219a45a51c967afff26d70747) to the Elixir client too.

PR #3794, opened 2026-01-28
Fix Materializer startup race condition via offset coordination

### LINKED ISSUE #3787: ShapeLogCollector memory growth to 3.5GB causing OOM crash on maxwell

robacourt:
## Summary

Fixes #3787 - addresses the snapshot/replication race condition in the Materializer that caused "Key already exists" crashes.

### The race condition

The Materializer subscribed to the Consumer **before** reading from storage:

```elixir
# BEFORE: In handle_continue(:start_materializer, ...)
Consumer.subscribe_materializer(stack_id, shape_handle, self())  # <- Subscribes first
{:noreply, state, {:continue, {:read_stream, shape_storage}}}    # <- Then reads ALL storage
```

During the window between subscribing and reading, any changes that arrive via `Consumer.new_changes()` would be delivered to the Materializer. If those changes included records that were also in the snapshot being read, the Materializer received duplicates and crashed.

### Production example (maxwell instance, 27 Jan 2026)

```
** (RuntimeError) Key "public"."offers"/"d3c8d8a5-5060-4a36-a67d-240de0c95a88" already exists
```

The transaction that triggered it:
```elixir
%Electric.Replication.Changes.NewRecord{
  relation: {"public", "offers"},
  record: %{"id" => "d3c8d8a5-5060-4a36-a67d-240de0c95a88"},
  key: "\"public\".\"offers\"/\"d3c8d8a5-5060-4a36-a67d-240de0c95a88\"",
  move_tags: ["e12422d3af57a36d01a50b4645a517e4"]  # <- Move-in event
}
```

The record was already in the snapshot (matched via `is_template = true` OR the subquery), AND was delivered via replication with `move_tags` as a move-in event.

### The fix

Consumer now returns its current offset when a Materializer subscribes. The Materializer reads storage **only up to that offset**, not beyond. Changes after that offset will be delivered via `new_changes` messages, ensuring each change is delivered exactly once.

```elixir
# AFTER: Consumer returns offset on subscription
def handle_call({:subscribe_materializer, pid}, _from, state) do
  {:reply, {:ok, state.latest_offset}, %{state | materializer_subscribed?: true}, ...}
end

# AFTER: Materializer uses offset to bound storage reads
{:ok, subscribed_offset} = Consumer.subscribe_materializer(stack_id, shape_handle, self())
{:noreply, %{state | subscribed_offset: subscribed_offset}, {:continue, {:read_stream, ...}}}

# In handle_continue({:read_stream, ...})
{:ok, offset, stream} = get_stream_up_to_offset(state.offset, state.subscribed_offset, storage)
```

## Test plan

- [x] Added test verifying offset coordination prevents duplicates
- [x] All existing Materializer tests pass (35 tests)
- [x] Full test suite passes (1334 tests)

### DIFF
--- a/packages/sync-service/test/electric/shapes/consumer/materializer_test.exs
+++ b/packages/sync-service/test/electric/shapes/consumer/materializer_test.exs
@@ -62,7 +62,11 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
       })
 
     respond_to_call(:await_snapshot_start, :started)
-    respond_to_call(:subscribe_materializer, :ok)
+
+    respond_to_call(
+      :subscribe_materializer,
+      {:ok, Electric.Replication.LogOffset.last_before_real_offsets()}
+    )
 
     assert Materializer.wait_until_ready(ctx) == :ok
   end
@@ -79,7 +83,11 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
       })
 
     respond_to_call(:await_snapshot_start, :started)
-    respond_to_call(:subscribe_materializer, :ok)
+
+    respond_to_call(
+      :subscribe_materializer,
+      {:ok, Electric.Replication.LogOffset.last_before_real_offsets()}
+    )
 
     assert Materializer.wait_until_ready(ctx) == :ok
 
@@ -763,7 +771,11 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
       })
 
     respond_to_call(:await_snapshot_start, :started)
-    respond_to_call(:subscribe_materializer, :ok)
+
+    respond_to_call(
+      :subscribe_materializer,
+      {:ok, Electric.Replication.LogOffset.last_before_real_offsets()}
+    )
 
     assert Materializer.wait_until_ready(ctx) == :ok
     Materializer.subscribe(ctx)
@@ -788,6 +800,104 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
     |> Enum.map(&Map.put(&1, :relation, relation))
     |> Enum.map(&Map.put(&1, :log_offset, Electric.Replication.LogOffset.first()))
     |> Enum.map(&Changes.fill_key(&1, pk_cols))
+  end
+
+  describe "startup offset coordination" do
+    test "no duplicate when offset coordination prevents overlap", ctx do
+      # This test verifies the fix: offset coordination prevents the duplicate.
+      # We use a mocked Consumer that returns an offset, and the Materializer
+      # should only read storage up to that offset, preventing duplicates.
+
+      Process.flag(:trap_exit, true)
+
+      shape_handle = "offset-test-#{System.unique_integer()}"
+
+      # Setup storage with a record at offset first()
+      storage = Storage.for_shape(shape_handle, ctx.storage)
+      Storage.start_link(storage)
+      writer = Storage.init_writer!(storage, @shape)
+      Storage.mark_snapshot_as_started(storage)
+
+      # Write a record to storage at LogOffset.first()
+      alias Electric.Replication.LogOffset
+      first_offset = LogOffset.first()
+
+      writer =
+        Storage.append_to_log!(
+          [
+            {first_offset, ~s|"public"."test_table"/"1"|, :insert,
+             ~s|{"key":"\\"public\\".\\"test_table\\"/\\"1\\"","value":{"id":"1","value":"10"},"headers":{"operation":"insert"}}|}
+          ],
+          writer
+        )
+
+      Storage.hibernate(writer)
+
+      # Mock Consumer that returns offset AND sends duplicate
+      test_pid = self()
+
+      consumer =
+        spawn(fn ->
+          receive do
+            {:"$gen_call", {from, ref}, :await_snapshot_start} ->
+              send(from, {ref, :started})
+          end
+
+          receive do
+            {:"$gen_call", {from, ref}, {:subscribe_materializer, mat_pid}} ->
+              # Return offset BEFORE the record - Materializer will read nothing from storage
+              send(from, {ref, {:ok, LogOffset.before_all()}})
+
+              # Simulate the race: send the same record via new_changes
+              # This should NOT crash because Materializer didn't read it from storage
+              GenServer.call(
+                mat_pid,
+                {:new_changes,
+                 [
+                   %Changes.NewRecord{
+                     relation: {"public", "test_table"},
+                     key: ~s|"public"."test_table"/"1"|,
+                     record: %{"id" => "1", "value" => "10"},
+                     move_tags: []
+                   }
+                 ]}
+              )
+
+              send(test_pid, :consumer_done)
+          end
+
+          Process.sleep(:infinity)
+        end)
+
+      ConsumerRegistry.register_consumer(consumer, shape_handle, ctx.stack_id)
+
+      # Start Materializer - should NOT crash
+      result =
+        Materializer.start_link(%{
+          stack_id: ctx.stack_id,
+          shape_handle: shape_handle,
+          storage: ctx.storage,
+          columns: ["value"],
+          materialized_type: {:array, :int8}
+        })
+
+      case result do
+        {:ok, pid} ->
+          # Wait for Consumer mock to finish
+          assert_receive :consumer_done, 5000
+          # Materializer should be alive
+          assert Process.alive?(pid)
+          # And should have the value
+          assert Materializer.get_link_values(%{
+                   stack_id: ctx.stack_id,
+                   shape_handle: shape_handle
+                 }) ==
+                   MapSet.new([10])
+
+        {:error, reason} ->
+          flunk("Materializer failed to start: #{inspect(reason)}")
+      end
+    end
   end
 
   describe "startup race condition handling" do
--- a/.changeset/fix-materializer-startup-race.md
+++ b/.changeset/fix-materializer-startup-race.md
@@ -1,0 +1,7 @@
+---
+'@core/sync-service': patch
+---
+
+Fix Materializer startup race condition that caused "Key already exists" crashes
+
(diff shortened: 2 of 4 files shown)


packages/sync-service/test/electric/shapes/consumer/materializer_test.exs
```diff
@@ -790,6 +802,104 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
...
+      shape_handle = "offset-test-#{System.unique_integer()}"
+
+      # Setup storage with a record at offset first()
+      storage = Storage.for_shape(shape_handle, ctx.storage)
+      Storage.start_link(storage)
+      writer = Storage.init_writer!(storage, @shape)
+      Storage.mark_snapshot_as_started(storage)
+
+      # Write a record to storage at LogOffset.first()
+      alias Electric.Replication.LogOffset
```
alco, line 822: `alias` should never be interspersed with regular code inside execution units such functions/tests. Putting it at the top of the module is best. In the worst case, it should be at the top of the `describe` block.

packages/sync-service/test/electric/shapes/consumer/materializer_test.exs
```diff
@@ -790,6 +802,104 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
...
+          writer
+        )
+
+      Storage.hibernate(writer)
+
+      # Mock Consumer that returns offset AND sends duplicate
+      test_pid = self()
+
+      consumer =
+        spawn(fn ->
```
alco, line 840: Such ad-hoc mocking is really difficult to parse for a human. This one-off process encodes a specific sequence of messages, relying on the underlying implementation of GenServer calls. It has no relation to the high-level code inside `consumer.ex`. There's no way to ensure that it stays in sync with that implementation. I'm not sure why has this approach even been considered for this particular test?

To highlight a more general problem:

I've noticed this module has a few instances of such `receive` expressions. At the same time, there's the `respond_to_call()` local function that abstracts over some instances of `receive`.

We should not allow such slip ups because they will infect the test code and will be difficult to remove after the fact. Good old test-only module or a mocked behaviour is a better way to write custom functionality for a specific component in the test env.

packages/sync-service/test/electric/shapes/consumer/materializer_test.exs
```diff
@@ -790,6 +802,104 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
...
+          assert Process.alive?(pid)
+          # And should have the value
+          assert Materializer.get_link_values(%{
+                   stack_id: ctx.stack_id,
+                   shape_handle: shape_handle
+                 }) ==
+                   MapSet.new([10])
+
+        {:error, reason} ->
+          flunk("Materializer failed to start: #{inspect(reason)}")
```
alco, line 898: I've noticed Claude's propensity to litter explicit `flunk` calls in the test code it generates. Kind of odd to have that in this case, a simple `{:ok, pid} = Materializer.start_link(...)` would have been perfectly fine.

packages/sync-service/test/electric/shapes/consumer/materializer_test.exs
```diff
@@ -790,6 +802,104 @@ defmodule Electric.Shapes.Consumer.MaterializerTest do
     |> Enum.map(&Changes.fill_key(&1, pk_cols))
   end
 
+  describe "startup offset coordination" do
+    test "no duplicate when offset coordination prevents overlap", ctx do
+      # This test verifies the fix: offset coordination prevents the duplicate.
+      # We use a mocked Consumer that returns an offset, and the Materializer
+      # should only read storage up to that offset, preventing duplicates.
+
+      Process.flag(:trap_exit, true)
```
alco, line 811: This a smell when there's no good reason for making the test process trap exits or there is a reason but no explanation of it in the code.

balegas:
That's valuable feedback. We want to maintain an high quality bar in our code. 

You mention specific code quality patterns. The agent must know we want to follow these patterns. My suggestion is that we should turn this into code quality guidelines that the agent/the reviewer agent must check whenever we create a pr. Why not creating that file or editing any coding style guidance we already have with your observations? Everytime we find something we record it

alco:
@balegas 100% agree. But I wanted to flag this up as the responsibility of the author of the PR.

We're still ways off from the workflow where all code can be written and reviewed by agents alone. Until then, we humans must remain in the loop and gatekeep any nonsense that agents try to add to the codebase. I think it's only fair for the human operator who delegates code generation to agents to make the necessary adjustments to the codified rules that we maintain to ensure code quality. Conversely, it wouldn't have been fair if I opened some PRs and expecting people who will be reviewing them to notice code that doesn't pass our quality bar and expect that they fix our rules for agents to try preventing this from happening again. The code already commited into Git must be fixed too, in addition to updating the rules for agents.

balegas:
Developer and reviewer and making sure that our tooling does the work for us. Let's start with setting up the tooling

PR #3809, opened 2026-02-02
Add POST support for subset queries to avoid URL length limits

KyleAMathews:
## Summary
Added comprehensive documentation for using POST requests with subset parameters in Electric's shape API. This addresses the `HTTP 414 Request-URI Too Long` errors that occur when WHERE clauses become too large for GET requests.

## Key Changes
- **New section**: "Using POST for subset queries" explaining when and why to use POST
- **POST body format**: Documented all subset parameters (`where`, `params`, `limit`, `offset`, `order_by`) with examples
- **Parameter separation**: Clarified which parameters belong in URL vs POST body
- **Security guidance**: Added detailed table of security-sensitive parameters that proxies must control server-side
- **Implementation example**: Provided complete TypeScript proxy handler showing dual GET/POST support with proper authorization handling
- **Client configuration**: Added examples for both `ShapeStream` and TanStack DB collections with `subsetMethod: 'POST'`
- **Backwards compatibility**: Included migration strategy for gradual rollout

## Notable Details
- Emphasizes that proxies should never trust client-provided values for `table`, `columns`, `where`, `params`, and `secret`
- Distinguishes between "shape definition" parameters (always in URL) and "subset snapshot" parameters (can be in body)
- Includes warning about GET deprecation in Electric 2.0
- Provides practical guidance on which protocol parameters are safe to pass through from clients

https://claude.ai/code/session_01MC34oGnC3Rqy4Jhsic59DY

### DIFF
--- a/website/docs/guides/auth.md
+++ b/website/docs/guides/auth.md
@@ -315,19 +315,30 @@ Electric separates parameters by purpose:
 - `params` — Parameters for the subset WHERE clause
 - `limit`, `offset`, `order_by` — Pagination controls
 
-##### Security: parameters your proxy should control
+##### Security: how Electric combines WHERE clauses
 
-When building an authorization proxy, **never trust client-provided values for security-sensitive parameters**. The proxy should set these server-side:
+Electric always combines the main shape WHERE (URL) with the subset WHERE (POST body) using `AND`:
+
+```sql
+WHERE {main_shape_where} AND ({subset_where})
+```
+
+This means **subset queries can only narrow results, never widen them**. Even if a client sends `where: "1=1"` in the POST body, the main shape WHERE still applies. The subset WHERE is validated for syntax and prohibited from containing subqueries.
+
+##### Parameters your proxy must control
+
+The proxy must set these **shape definition parameters** server-side — they define what data the client can access:
 
 | Parameter | Where | Security Consideration |
 |-----------|-------|------------------------|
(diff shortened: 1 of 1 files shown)


website/docs/guides/auth.md
```diff
@@ -259,6 +259,213 @@ Benefits:
...
+
+:::warning URL Length Limits and GET Deprecation
+GET requests with subset parameters in the URL can fail with `414 Request-URI Too Long` errors. This is common when:
+- ACL subqueries generate long WHERE clauses
+- Join queries produce large filter lists
+- Parameter arrays contain many values
+
+**Use POST to avoid this limitation.**
+
+> **Deprecation Notice:** In Electric 2.0, GET requests for subset snapshots will be deprecated and only POST will be supported. Implement POST support now to ensure forward compatibility.
```
balegas, line 274: Why are going to deprecate?

KyleAMathews: because GETs break really easily in production environments. So better to just always steer people to POSTs

evanob: how does CDN caching work with POST?

KyleAMathews: It doesn't 😀 but these are queries to postgres so they're inherently not cachable

evanob: hah 🤦 should have read a little further up! thanks

PR #3814, opened 2026-02-07
Add system wake detection for non-browser environments

KyleAMathews:
Adds system wake detection to `ShapeStream` for non-browser environments (Bun, Node.js, etc). Without this, when a daemon process wakes from OS sleep, in-flight long-poll or SSE requests hang until the OS TCP timeout (60-120+ seconds depending on platform), causing a gap in data delivery.

## Approach

Uses timer gap detection: a `setInterval` ticks every 10 seconds. If the elapsed wall-clock time since the last tick exceeds 25 seconds (10s interval + 15s threshold), the system was likely asleep. On detection, the stale request is aborted with a `SYSTEM_WAKE` reason and a fresh non-live request is issued to catch up before resuming live mode.

This is the non-browser counterpart to `visibilitychange`-based pause/resume. The two mechanisms are mutually exclusive — `#hasBrowserVisibilityAPI()` gates which one activates.

**Key invariants:**
- Wake detection only fires when `#state === 'active'` and an abort controller exists
- `#isRefreshing` is set before abort so the reconnect issues a non-live request first (same as `forceDisconnectAndRefresh`)
- `timer.unref()` prevents the interval from keeping the Node.js/Bun process alive
- Cleanup runs via `unsubscribeAll()` alongside visibility change cleanup

**Non-goals:**
- No logging on wake detection — this is normal operational behavior, not an error
- No handling of wake during paused state — if the stream is paused, there's no stale request to abort

## Verification

```bash
cd packages/typescript-client
pnpm vitest run --config vitest.unit.config.ts test/wake-detection.test.ts
```

Tests run in `@vitest-environment node` to ensure `document` is genuinely absent (rather than relying on `delete globalThis.document` under jsdom).

## Files changed

- **`src/client.ts`** — Add `#subscribeToWakeDetection()`, `#hasBrowserVisibilityAPI()`, update catch block with `isRestartAbort` for `SYSTEM_WAKE`
- **`src/constants.ts`** — Add `SYSTEM_WAKE` constant
- **`test/wake-detection.test.ts`** — New test file with node environment: timer setup, browser exclusion, and wake gap detection
- **`vitest.unit.config.ts`** — Include wake detection tests in unit config

https://claude.ai/code/session_01VhBX3nM9TfKSngu9u4C1zB

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -1595,8 +1595,8 @@ export class ShapeStream<T extends Row<unknown> = Row>
   #subscribeToWakeDetection() {
     if (this.#hasBrowserVisibilityAPI()) return
 
-    const INTERVAL_MS = 10_000
-    const WAKE_THRESHOLD_MS = 15_000
+    const INTERVAL_MS = 2_000
+    const WAKE_THRESHOLD_MS = 4_000
 
     let lastTickTime = Date.now()
 
--- a/packages/typescript-client/test/wake-detection.test.ts
+++ b/packages/typescript-client/test/wake-detection.test.ts
@@ -98,16 +98,16 @@ describe(`Wake detection`, () => {
     expect(fetchCallCount).toBeGreaterThanOrEqual(1)
     const initialFetchCount = fetchCallCount
 
-    // Advance one normal interval (10s) — should NOT trigger wake detection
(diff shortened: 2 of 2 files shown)


KyleAMathews:
Good call — changed and now merging.

K-Mistele:
awesome, thanks for this, guys! is there a new client version with this fix yet?

PR #3817, opened 2026-02-09
Add Client.poll for Elixir client

robacourt:
## Summary

- Add `Client.poll/4` as a new public API for explicit request-response polling of shape changes
- Extract `ShapeState`, `TagTracker`, and `Poll` modules from stream internals for shared use
- Refactor `Stream` to delegate to `Poll` for HTTP request handling, reducing duplication

## Motivation

Unlike `stream/3` which returns a continuous `Enumerable`, `poll/4` makes a single request and returns `{:ok, messages, new_state}`, giving callers explicit control over request timing. This has mainly been added for Integration tests, where it's useful to return when changes are received rather than wrapping changes in a stream, but has also been included in the public API as it may be useful in production use cases as well.

## Changes

- **`Client.poll/4`** — new public API accepting table names, `ShapeDefinition`, or Ecto queries
- **`ShapeState`** — extracted struct holding polling state (shape handle, offset, schema, tag tracking)
- **`TagTracker`** — extracted module for move-out tag tracking and synthetic delete generation
- **`Poll`** — extracted module handling single HTTP request/response cycle
- **`Stream`** — refactored to use `Poll` internally, removing duplicated logic
- **Tests** — unit tests for `ShapeState` and `TagTracker`

## Test Plan

- [x] Existing stream tests pass (unit tests with Bypass)
- [x] New `ShapeState` tests verify creation, resume interop, and round-trip
- [x] New `TagTracker` tests verify tag indexing and synthetic delete generation
- [x] Integration tests with running Electric server

---

### DIFF
--- a/packages/elixir-client/lib/electric/client/poll.ex
+++ b/packages/elixir-client/lib/electric/client/poll.ex
@@ -131,7 +131,7 @@ defmodule Electric.Client.Poll do
   defp handle_must_refetch(resp, client, state) do
     handle = shape_handle(resp) || "#{state.shape_handle}-next"
 
-    new_state = reset_state(state, handle)
+    new_state = ShapeState.reset(state, handle)
     new_state = handle_schema(resp, client, new_state)
 
     %{value_mapper_fun: value_mapper_fun} = new_state
@@ -142,18 +142,6 @@ defmodule Electric.Client.Poll do
       |> Enum.flat_map(&Message.parse(&1, handle, value_mapper_fun, resp.request_timestamp))
 
     {:must_refetch, messages, new_state}
-  end
-
-  defp reset_state(state, shape_handle) do
-    %{
-      state
-      | offset: Client.Offset.before_all(),
-        shape_handle: shape_handle,
-        up_to_date?: false,
-        next_cursor: nil,
-        tag_to_keys: %{},
-        key_data: %{}
-    }
   end
 
   defp process_messages(messages, state) do
--- a/packages/elixir-client/lib/electric/client/shape_state.ex
+++ b/packages/elixir-client/lib/electric/client/shape_state.ex
@@ -80,6 +80,25 @@ defmodule Electric.Client.ShapeState do
   end
 
   @doc """
+  Reset polling state for a new shape handle, preserving schema and value mapper.
+
+  Used when a 409 (must-refetch) response is received — the shape handle changes
+  but the schema remains the same.
+  """
+  @spec reset(t(), Client.shape_handle()) :: t()
+  def reset(%__MODULE__{} = state, shape_handle) do
+    %{
+      state
+      | offset: Offset.before_all(),
+        shape_handle: shape_handle,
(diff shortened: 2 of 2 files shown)


packages/elixir-client/lib/electric/client.ex
```diff
@@ -546,6 +548,84 @@ defmodule Electric.Client do
...
+
+  if Code.ensure_loaded?(Ecto) do
+    def poll(%Client{} = client, queryable, %ShapeState{} = state, opts)
+        when is_ecto_shape(queryable) do
+      shape_definition = shape!(queryable)
+      poll(client, shape_definition, state, opts)
+    end
+  end
+
+  defp do_poll(%Client{} = client, %ShapeState{} = state, opts) do
```
alco: This function looks superfluous. `Poll.request()` could be called directly.

packages/elixir-client/lib/electric/client/poll.ex
```diff
@@ -0,0 +1,243 @@
...
+  This module provides explicit request-response semantics for fetching
+  changes from Electric, as an alternative to the streaming API.
+
+  ## Usage
+
+      # Create initial state
+      state = ShapeState.new()
+
+      # Make a polling request
+      case Client.poll(client, shape, state) do
```
alco: This looks misplaced: it's the same example from the `Client` module. In this module, examples should reference `Client.Poll.<function name>` instead.

packages/elixir-client/lib/electric/client/poll.ex
```diff
@@ -0,0 +1,243 @@
...
+  ## Returns
+
+    * `{:ok, messages, new_state}` - Success, messages received
+    * `{:must_refetch, messages, new_state}` - Shape was reset (409), state has been cleared
+    * `{:error, error}` - Error occurred
+
+  ## Examples
+
+      state = ShapeState.new()
+      {:ok, messages, state} = Client.poll(client, shape, state, replica: :full)
```
alco: Again, same example as in the `Client` module. This should have been an example of invoking `Client.Pool.request()` instead.

packages/elixir-client/lib/electric/client/poll.ex
```diff
@@ -0,0 +1,243 @@
...
+
+    messages =
+      resp.body
+      |> ensure_enum()
+      |> Enum.flat_map(&Message.parse(&1, handle, value_mapper_fun, resp.request_timestamp))
+
+    {:must_refetch, messages, new_state}
+  end
+
+  defp reset_state(state, shape_handle) do
```
alco: If `state` is of type `%ShapeState{}` here, I wonder:
- does this function do anything different from `ShapeState.new()`? If yes, what and why?
- doesn't it belong inside the `ShapeState` module as `ShapeState.reset(old_state)`?

robacourt: done

PR #3823, opened 2026-02-10
fix(sync-service): Tweak behaviour of SQLite storage to handle storage latency

magnetised:
With the realisation that our supposed "nvme" attached storage in the cloud is in fact an EBS volume, this pr is a grab bag of fixes to  improve the system's performance when the storage directory is laggy and slow. 

I've split it into separate commits for ease of review. 

There's a new linux-only script `packages/sync-service/scripts/simulate-fargate-io.sh` that drops you into a shell with a `$DATA_DIR` set to a new temporary volume which is configured with a pre-defined latency and cgroup emulated IOPs limits. 

Using this I can reproduce the issues we're seeing in cloud, with timeouts creating shapes and also timeouts getting a read connection from the sqlite pool. Even with these changes it's still possible to generate timeouts creating a shape by simply saturating the system with requests. With a laggy storage backend and a timeout-based system you can always generate enough concurrent demand to fail some calls.

Having said that the system is more stable under load. A critical change is to move the retrieval of the shape from the handle from the `Consumer.init/1` callback to the `handle_continue/2` initialisation callback. This makes the starting of a consumer process an O(1) operation with no dependency on fast storage lookups in the hot-path. Previously a slow shape lookup could crash the entire stack by raising a timeout in the start_consumer call within the ShapeCache. Now a slow shape lookup will cause slightly less immediately critical problems by delaying the broadcast of changes from the shape log collector, something we should hopefully recover from more easily.

Finally I've put 2 layers of nesting in the storage directory, so that there is no single directory with e.g. 200,000 children. Instead there are 2 levels of dirs with max 100 children then the shape directories. I did this in an attempt to improve filesystem performance but I think in reality with modern ext4 it shouldn't matter how many files are in a directory. Having said that, it should improve the memory performance of recursive functions on the shape directories so might be worth keeping...

This is obviously backwards incompatible so I've bumped the storage version to invalidate any existing shape db and added a background task to remove the old-style data dirs. 

Happy to drop if we think the gain isn't worth the pain.

Future work:

- make sqlite durabilty and page cache size configurable (https://github.com/electric-sql/electric/issues/3824)

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage.ex
@@ -148,14 +148,17 @@ defmodule Electric.ShapeCache.PureFileStorage do
     shape_data_dir(shape_opts, [])
   end
 
-  defp shape_data_dir(%__MODULE__{stack_id: stack_id, shape_handle: shape_handle}, suffix) do
+  def shape_data_dir(%__MODULE__{stack_id: stack_id, shape_handle: shape_handle}, suffix) do
     {__MODULE__, stack_opts} = Storage.for_stack(stack_id)
     shape_data_dir(stack_opts.base_path, shape_handle, suffix)
   end
 
-  defp shape_data_dir(base_path, shape_handle, suffix \\ [])
-       when is_binary(base_path) and is_binary(shape_handle) do
-    Path.join([base_path, shape_handle | suffix])
+  @doc false
+  def shape_data_dir(base_path, shape_handle, suffix \\ [])
+      when is_binary(base_path) and is_binary(shape_handle) do
+    # nest storage to limit number of files per directory
+    <<p1::binary-2, p2::binary-2, _::binary>> = shape_handle
+    Path.join([base_path, p1, p2, shape_handle | suffix])
   end
 
   defp shape_log_dir(opts), do: shape_data_dir(opts, [@log_dir])
@@ -186,25 +189,20 @@ defmodule Electric.ShapeCache.PureFileStorage do
              write_concurrency: :auto
            ])
          end},
-        {Task.Supervisor, name: stack_task_supervisor(opts.stack_id)}
+        {Task.Supervisor, name: stack_task_supervisor(opts.stack_id)},
+        # TODO: remove once we're sure that no install has un-nested storage
+        # directories
+        {Task, fn -> remove_unnested_storage(opts.base_path) end}
       ],
       strategy: :one_for_one
     )
   end
 
   def get_all_stored_shape_handles(%{base_path: base_path}) do
-    case ls(base_path) do
-      {:ok, shape_handles} ->
-        shape_handles
-        |> Enum.reject(&String.starts_with?(&1, "."))
-        |> then(&{:ok, MapSet.new(&1)})
-
-      {:error, :enoent} ->
-        {:ok, MapSet.new()}
-
-      {:error, reason} ->
-        {:error, reason}
-    end
+    {:ok,
+     Path.wildcard("#{base_path}/*/*/*")
+     |> Stream.map(&Path.basename/1)
+     |> Enum.into(MapSet.new())}
   end
 
   def get_stored_shapes(stack_opts, shape_handles) do
@@ -775,7 +773,6 @@ defmodule Electric.ShapeCache.PureFileStorage do
   end
 
   defp create_directories!(%__MODULE__{} = opts) do
-    mkdir_p!(shape_data_dir(opts))
     mkdir_p!(shape_log_dir(opts))
     mkdir_p!(shape_metadata_dir(opts))
   end
@@ -1392,10 +1389,6 @@ defmodule Electric.ShapeCache.PureFileStorage do
     File.rm_rf!(path)
   end
 
-  defp ls(path) do
-    FileInfo.ls(path)
-  end
-
   defp rename!(path1, path2) do
     case FileInfo.rename(path1, path2) do
       :ok ->
@@ -1456,5 +1449,22 @@ defmodule Electric.ShapeCache.PureFileStorage do
 
   defp shape_gone?(%__MODULE__{} = opts) do
     !FileInfo.exists?(shape_data_dir(opts))
+  end
+
+  defp remove_unnested_storage(base_dir) do
+    "#{base_dir}/*"
+    |> Path.wildcard()
+    |> Enum.filter(fn path ->
+      name = Path.basename(path)
+      Regex.match?(~r/\d{6,}-\d{8,}/, name)
+    end)
+    |> tap(fn dirs ->
+      Logger.info("Removing #{length(dirs)} old storage directories")
+    end)
+    |> Enum.each(fn path ->
+      # remove with a delay - don't want this task to interfere with critical processes
+      File.rm_rf(path)
+      Process.sleep(20)
+    end)
   end
 end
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
```diff
@@ -134,10 +134,10 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.Connection do
     # for our current deployment mode synchronous = OFF would be enough (hand
     # data to kernel, don't fsync) but for oss deploys we should keep it at a
     # higher durability setting
-    "PRAGMA synchronous=NORMAL",
+    "PRAGMA synchronous=OFF",
     # Reduce page cache since hot-path lookups now use ETS cache.
     # -512 means 512KB (negative values are in KiB, default is often 2MB)
-    "PRAGMA cache_size=-512"
+    "PRAGMA cache_size=-4096"
```
msfstef, line 140: should update the comments probably? don't know how to feel about hard-tweaking OSS for how we deploy the service on a particular environment (?)

msfstef (commented):
Is this flushed at delayed cleanup to avoid any races where we clean up the entries from the buffer but they don't show up on SQLite? Is the time needed arbitrary for it to show up? It would be nice to have hard guarantees.

Additionally, I'm slightly worried with changes that are catered to the particular runtime environment - this should be more configurable right?

magnetised:
> Is this flushed at delayed cleanup to avoid any races where we clean up the entries from the buffer but they don't show up on SQLite? Is the time needed arbitrary for it to show up? It would be nice to have hard guarantees.

it's to avoid sqlite consistency issues where read-only connections don't immediately pick up writes. a read transaction starts a snapshot and if that snapshot is started just before the insert of a new shape, the read won't pick up the new shape. the old version added a verification step that used the write connection to make a last query and, as sqlite connections are always able to read their own writes, this worked at the cost of making the check contend for the write connection. 

this version provides no hard guarantees but reduces contention on the write connection. shapecache already has problems keeping up with demand under high load when storage latency is high, so not adding to that is important. pragmatically this will work even if theoretically it is not guaranteed. 

> Additionally, I'm slightly worried with changes that are catered to the particular runtime environment - this should be more configurable right?

Yes, see my description. Follow-on work will make these values configurable.

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
magnetised: will address in a follow up (see pr description)

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/query.ex
```diff
@@ -280,5 +281,11 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.Query do
     end
   end
 
-  def term_to_binary(term), do: :erlang.term_to_binary(term, [:deterministic])
+  defp term_to_binary(term), do: :erlang.term_to_binary(term, [:deterministic])
+
+  defp comparable_to_binary(commparable_shape) do
+    commparable_shape
```
robacourt:
```suggestion
  defp comparable_to_binary(comparable_shape) do
    comparable_shape
```

magnetised:
benchmark this

robacourt:
The time-based retention for flushed ETS entries is a bit fragile — the right value depends on SQLite's
  WAL visibility lag which is hard to bound, and if you get it wrong you get duplicate shapes. Since SQLite WAL visibility is
  monotonic (if a read connection can see write N, it can see everything before N), you could replace the
  time-based heuristic with a read-back verification:

  On each cleanup tick, the cleaner checks out a SQLite read connection and probes for the most recently
  flushed entry. If the read connection can see it, all flushed entries are safe to remove. If not, do
  nothing and try again next tick.

  This gives a hard guarantee — entries are only removed once provably visible to readers — regardless of
  how laggy storage is. It also cleans up faster under normal conditions (milliseconds after flush rather
  than 5 minutes).

  NimblePool checkouts get a fresh SQLite WAL snapshot each time, so verifying on one connection is
  sufficient — all subsequent checkouts on any pool connection will also see the data.
  
  What do you think?

magnetised:
> NimblePool checkouts get a fresh SQLite WAL snapshot each time, so verifying on one connection is sufficient — all subsequent checkouts on any pool connection will also see the data.

You know I don't think you can be right here. All inserts to the db are done within the shape cache process and the read check that was failing was also within the ShapeCache process, which means that that final check was definitely being performed on a freshly checked out connection, not on some simultaneous read, which wasn't seeing the write. It doesn't happen often, it's not trivial to reproduce, so there's no "that's just how it works" thing going on in terms of "fresh checkouts". The sqlite connection is just a ref AFAIK, not some process, so there's no connection between a checkout op and anything happening in the connection.

So it would need to be done on all connections, not just one. Which means that every tick we're checkout out all the connections and performing an additional operation for every shape in the cache. Under high load that would add a lot more queries -- the higher the write load the higher the verification load... 

Both you and Steph are worried about this possible (if not probable) issue. In my simulations -- which revealed the issue in the first place -- it doesn't occur and I've slammed the system pretty hard. 

I'll revert this change and have a look how it behaves. Other changes may have reduced the need.

robacourt:
> You know I don't think you can be right here. All inserts to the db are done within the shape cache process and the read check that was failing was also within the ShapeCache process, which means that that final check was definitely being performed on a freshly checked out connection, not on some simultaneous read, which wasn't seeing the write. It doesn't happen often, it's not trivial to reproduce, so there's no "that's just how it works" thing going on in terms of "fresh checkouts". The sqlite connection is just a ref AFAIK, not some process, so there's no connection between a checkout op and anything happening in the connection.
> 
> So it would need to be done on all connections, not just one. Which means that every tick we're checkout out all the connections and performing an additional operation for every shape in the cache. Under high load that would add a lot more queries -- the higher the write load the higher the verification load...

Ah, ok. Fair enough.

> Both you and Steph are worried about this possible (if not probable) issue. 

Anything that uses a wait to get out of a race condition raises alarm bells for me 😄 

> I'll revert this change and have a look how it behaves. Other changes may have reduced the need.

Yeah, that sounds worth a try

packages/sync-service/scripts/simulate-fargate-io.sh
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env bash
+#
+# Simulate AWS Fargate EBS I/O characteristics using dm-delay for latency
+# and cgroup v2 throttling for IOPS/throughput limits.
```
alco, line 4: Brilliant!

packages/sync-service/lib/electric/shape_cache/shape_status_owner.ex
```diff
@@ -21,7 +21,7 @@ defmodule Electric.ShapeCache.ShapeStatusOwner do
   end
 
   def initialize(stack_id) do
-    GenServer.call(name(stack_id), :initialize)
+    GenServer.call(name(stack_id), :initialize, :infinity)
```
alco, line 29: A comment here explaining the reasoning (and why it is impossible for the called process to hang forever) would be welcome here.

packages/sync-service/scripts/simulate-fargate-io.sh
magnetised: Can't claim too much credit! 😊 You can tell by the sheer quantity of comments that this wasn't hand-crafted

PR #3825, opened 2026-02-10
fix examples handling of response headers

KyleAMathews:
## Summary

Simplify proxy examples to forward all response headers from Electric instead of selectively managing individual headers. This fixes `MissingHeadersError` that users hit when proxies strip Electric's CORS headers.

## Root Cause

Users running Electric behind a proxy were getting `MissingHeadersError` because the example proxy code was selectively copying headers — either missing newly added Electric headers or stripping CORS headers like `access-control-expose-headers`. The yjs example had an additional subtle bug: it manually copied headers into a new `Response` while also hardcoding a list of `electric-*` headers, creating a maintenance burden every time Electric added a new header.

## Approach

Instead of maintaining an explicit list of headers to forward, the proxies now pass through **all** response headers from Electric, only removing `content-encoding` and `content-length` (which become invalid when `fetch()` decompresses the body).

Key changes:
- **tanstack proxy**: Removed CORS header filtering — Electric already sets correct CORS headers including `access-control-expose-headers`
- **yjs proxy**: Replaced ~30 lines of manual header copying with a simple `new Response()` using a mutable copy of the upstream headers. Always constructs a new `Response` (not just in the `content-encoding` branch) because Node.js `fetch()` returns immutable headers that Hono's CORS middleware can't merge into
- **Error message**: Improved `MissingHeadersError` to give actionable guidance about forwarding all headers, specifically mentioning `access-control-expose-headers`
- **Troubleshooting docs**: Updated to match the new "forward everything" guidance

### Key Invariants

- Proxies must return a `Response` with mutable headers so framework middleware (e.g., Hono CORS) can add its own headers
- Only `content-encoding` and `content-length` should be stripped, and only when the proxy decompresses the body
- Electric's own `access-control-expose-headers` header must reach the browser for the client to read `electric-*` headers

### Non-goals

- Did not change the tanstack proxy's `CORS_HEADERS` constant, which is still used for non-proxy responses (health, POST/DELETE, errors)

## Verification

Review the example proxy code to confirm all Electric headers pass through:

```bash
# Check the proxy examples work with Electric
cd examples/tanstack && pnpm dev
cd examples/yjs && pnpm dev
```

## Files changed

| File | Change |
|------|--------|
| `examples/tanstack/src/server/app.js` | Stop filtering CORS headers from Electric responses, forward all headers |
| `examples/yjs/src/server/server.ts` | Remove `exposeHeaders` CORS config and manual header copying, always create mutable Response |
| `packages/typescript-client/src/error.ts` | Improve `MissingHeadersError` with actionable proxy guidance |
| `website/docs/guides/troubleshooting.md` | Update "missing headers" solution to recommend forwarding all headers |

---

### DIFF
--- a/.changeset/fix-proxy-forward-headers.md
+++ b/.changeset/fix-proxy-forward-headers.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/client': patch
+---
+
+Improve `MissingHeadersError` message to provide clearer guidance on forwarding all response headers from Electric, including the `access-control-expose-headers` header.
--- a/examples/tanstack/src/server/app.js
+++ b/examples/tanstack/src/server/app.js
@@ -74,7 +74,8 @@ const server = http.createServer(async (req, res) => {
       try {
         const response = await fetch(originUrl)
 
-        // Copy headers, excluding problematic ones
+        // Forward all headers from Electric, removing content-encoding
+        // and content-length which become invalid after fetch() decompresses the body
         const headers = {}
         response.headers.forEach((value, key) => {
           if (
--- a/examples/tanstack/src/server/app.js
+++ b/examples/tanstack/src/server/app.js
@@ -75,14 +75,11 @@ const server = http.createServer(async (req, res) => {
         const response = await fetch(originUrl)
 
         // Copy headers, excluding problematic ones
-        const headers = { ...CORS_HEADERS }
+        const headers = {}
         response.headers.forEach((value, key) => {
           if (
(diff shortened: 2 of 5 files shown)


plutou-ben:
@balegas - Is this related to my error? If so, will I need to update anything on how the electric_query_params?

balegas:
> @balegas - Is this related to my error? If so, will I need to update anything on how the electric_query_params?

No, just consolidating examples.

balegas (approved):

PR #3827, opened 2026-02-10
feat: add PauseLock to coordinate snapshot/visibility pause interactions

KyleAMathews:
## Summary

Introduces `PauseLock`, a set-based coordination primitive that replaces the old `#activeSnapshotRequests` counter and `#pauseRequested` boolean for managing stream pause/resume across multiple independent subsystems. Fixes race conditions where a snapshot resume could override a visibility pause (or vice versa).

## Approach

**Problem**: The old code used a counter (`#activeSnapshotRequests`) and a boolean (`#pauseRequested`) to coordinate pausing between visibility changes and snapshot requests. These couldn't track *why* the stream was paused, so one subsystem's resume could inadvertently override another's pause. For example: tab goes hidden (pause), snapshot starts, snapshot finishes (resume) — but the tab is still hidden.

**Solution**: `PauseLock` uses a `Set<string>` of reason strings. The stream pauses when the first reason is acquired and only resumes when *all* reasons are released. Each snapshot gets a unique reason (`snapshot-1`, `snapshot-2`, ...) so concurrent snapshots are tracked independently.

```
lock.acquire('visibility')   // → stream pauses
lock.acquire('snapshot-1')   // → no-op, already paused
lock.release('snapshot-1')   // → no-op, 'visibility' still held
lock.release('visibility')   // → stream resumes
```

**Key design decisions**:
- **PauseLock is the single source of truth** — `client.ts` does not use `PausedState` at all. Instead of maintaining two representations of "paused" that must be synchronized, `#requestShape` simply returns when `isPaused`, and a `#resumingFromPause` flag (set by `onReleased`) handles the long-poll avoidance
- `onAcquired`/`onReleased` callbacks fire on edge transitions (empty→non-empty, non-empty→empty), not on every acquire/release
- `releaseAllMatching(prefix)` for reset paths — clears snapshot reasons without firing `onReleased`, preserving visibility state
- Snapshot lock acquired *before* waiting for stream end, immediately aborting in-flight long-polls (prevents up to 20s blocking)
- Warnings on duplicate acquire and release-of-unheld-reason for diagnosability
- Re-check `isPaused` after async setup in `#requestShape` to close the race window where the lock is acquired while the abort controller doesn't exist yet

**Also includes**:
- `#isRefreshing` boolean → `#refreshCount` counter with `try/finally` for concurrent refresh safety
- Code simplification: extracted `#teardown()` helper, early-return patterns, nullish coalescing, removed redundant comments

## Verification

```bash
cd packages/typescript-client
# 110 tests passing (includes 13 PauseLock tests)
```

## Files changed

| File | Change |
|------|--------|
| `src/pause-lock.ts` | New — set-based lock with acquire/release/releaseAllMatching |
| `src/client.ts` | Replace `#pauseRequested`/`#activeSnapshotRequests`/`#pause()`/`#resume()` with `PauseLock`; extract `#teardown()`; `#refreshCount` with try/finally; `#resumingFromPause` flag |
| `src/shape-stream-state.ts` | Idempotent `PausedState.pause()`; code simplification (parseResponseFields, constructors, handleMessageBatch) |
| `test/pause-lock.test.ts` | New — 13 tests covering edge transitions, concurrent reasons, interleaving scenarios |
| `vitest.unit.config.ts` | Add pause-lock test to unit config |

https://claude.ai/code/session_01UGPdwB6UpFkkQi9p4sjPRj


packages/typescript-client/src/client.ts
```diff
@@ -694,75 +716,61 @@ export class ShapeStream<T extends Row<unknown> = Row>
         this.#syncState = this.#syncState.toErrorState(err)
       }
 
-      // Check if onError handler wants to retry
```
kevin-dp, line 697: Hey, claude is removing all our comments. Don't do that

PR #3828, opened 2026-02-11
Fix infinite loop when local handle matches stale handle

kevin-dp:
This PR [reproduces](https://github.com/electric-sql/electric/actions/runs/21897631349/job/63217468839?pr=3828) and fixes a bug related to stale handlers.

### Bug: stale cache detection fails when client's own handle is the expired handle

When a shape handle is marked as expired (e.g. after a 409 response), the client is supposed to retry with a cache buster query parameter to bypass stale CDN/proxy caches. However, this only works when the client has **no handle** (fresh start) or a **different handle** than the expired one.

When the client resumes with a persisted handle that happens to be the same as the expired handle (`localHandle === expiredHandle`), the stale detection logic sees that the client already has a handle and returns `ignored` instead of `stale-retry`. The client logs a warning ("Ignoring the stale response") but never adds a cache buster — so it just keeps receiving the same stale cached response in an infinite loop.

### How the test reproduces it

1. Marks handle `expired-H1` as expired in the `ExpiredShapesCache`
2. Creates a `ShapeStream` that resumes with `handle: expired-H1` (simulating a client that persisted this handle from a previous session)
3. The mock backend always returns responses with that same expired handle (mimicks CDN behaviour)
4. Asserts that the client should use a `cache_buster` query parameter to escape the stale cache — which currently fails because the client takes the `ignored` path instead of `stale-retry`

### Root cause

In `checkStaleResponse` (line 311-344), the condition at line 322 is:

```typescript
if (this.#shared.handle === undefined) {
  // enter stale retry
}
// else: "We have a valid local handle — ignore this stale response"
```

This assumes that if the client has a local handle, it's a *different* handle from the expired one, so the stale response can be safely ignored. But that assumption is wrong when `localHandle === expiredHandle` — the client resumed with the same handle that was marked expired.

At this point in the code, we already know `responseHandle === expiredHandle` (line 317). The missing check is whether `this.#shared.handle` is *also* the expired handle.

### Fix

Change the condition at line 322 from:
```typescript
if (this.#shared.handle === undefined) {
```
to:
```typescript
if (this.#shared.handle === undefined || this.#shared.handle === expiredHandle) {
```
That's it — one condition added. When the client's own handle matches the expired handle, it enters `stale-retry` (gets a cache buster) instead of falling through to `ignored`. The rest of the stale-retry machinery already handles everything correctly from there.

### DIFF
--- a/packages/typescript-client/test/expired-shapes-cache.test.ts
+++ b/packages/typescript-client/test/expired-shapes-cache.test.ts
@@ -581,15 +581,9 @@ describe(`ExpiredShapesCache`, () => {
   })
 
   it(`client should retry with cache buster when local handle matches expired handle`, async () => {
-    // BUG: When the client's own persisted handle IS the expired handle,
-    // checkStaleResponse sees handle !== undefined → returns 'ignored'.
-    // But 'ignored' means "we have a valid handle, skip this stale response".
-    // The client's handle is NOT valid — it's the expired one!
-    // The client loops forever: fetch → ignored → retry → fetch → ignored...
-    // never using a cache buster to escape the stale CDN cache.
-    //
-    // Expected: The client should detect that localHandle === expiredHandle
-    // and use stale-retry (with cache buster) instead of ignored.
+    // When the client's own persisted handle IS the expired handle,
+    // the client should detect that localHandle === expiredHandle and
+    // use stale-retry (with cache buster) to bypass the stale CDN cache.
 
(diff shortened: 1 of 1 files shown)


balegas:
Nice to see this. This was the issue I was showing Sam a couple of weeks ago, but was not sure how to reproduce it. 🙏

KyleAMathews (approved):
:shipit:

PR #3831, opened 2026-02-11
feature(sync-service, electric-telmetry): Add disk usage stats

magnetised:
Re-calculate every minute by default and use cached value in between.

### DIFF
--- a/.changeset/plenty-hotels-kiss.md
+++ b/.changeset/plenty-hotels-kiss.md
@@ -1,0 +1,6 @@
+---
+'@core/electric-telemetry': patch
+'@core/sync-service': patch
+---
+
+Add disk usage telemetry to stacks.


packages/sync-service/lib/electric/stack_supervisor/telemetry.ex
```diff
@@ -59,7 +60,8 @@ defmodule Electric.StackSupervisor.Telemetry do
       [
         {__MODULE__, :count_shapes, [stack_id]},
         {__MODULE__, :report_write_buffer_size, [stack_id]},
-        {__MODULE__, :report_retained_wal_size, [stack_id, config.replication_opts[:slot_name]]}
+        {__MODULE__, :report_retained_wal_size, [stack_id, config.replication_opts[:slot_name]]},
+        {__MODULE__, :report_disk_usage, [stack_id]}
```
alco: I don't think this needs to be added by default. We need it in cloude and we can enable it in cloud

packages/electric-telemetry/lib/electric/telemetry/disk_usage/disk.ex
```diff
@@ -0,0 +1,43 @@
...
+      {:error, _} ->
+        0
+    end
+  end
+
+  defdelegate stat(path), to: :prim_file, as: :read_file_info
+
+  defp ls(path) do
+    with {:ok, list} <- :prim_file.list_dir(path) do
+      {:ok, Enum.map(list, &to_string/1)}
```
alco: We don't need to convert chartlists to binaries here:
```
iex(1)> Path.join(['foo/bar', "baz"])
"foo/bar/baz"
```

packages/electric-telemetry/lib/electric/telemetry/disk_usage.ex
```diff
@@ -0,0 +1,166 @@
...
+    {:ok, stack_id} = Keyword.fetch(args, :stack_id)
+    {:ok, storage_dir} = Keyword.fetch(args, :storage_dir)
+    update_period = Keyword.get(args, :update_period, @default_update_period)
+
+    table =
+      :ets.new(ets_table(stack_id), [
+        :named_table,
+        :set,
+        :protected,
+        read_concurrency: true
```
alco: `read_concurrencyc` seems unnecessary since we start a separate DiskUsage process per stack.

packages/electric-telemetry/lib/electric/telemetry/disk_usage.ex
```diff
@@ -0,0 +1,166 @@
...
+      table: table,
+      measurement_duration: duration
+    } = state
+
+    :ets.insert(table, {:usage_bytes, usage_bytes, updated_at, duration})
+    state
+  end
+
+  defp ets_table(stack_id) do
+    :"ElectricTelemetry.DiskUsage:#{stack_id}"
```
alco, line 110: We should also use it for the process name. Always good to have named processes for live debugging.

packages/electric-telemetry/lib/electric/telemetry/disk_usage.ex
```diff
@@ -0,0 +1,166 @@
...
+      _ ->
+        {0, nil, true}
+    end
+  end
+
+  defp save_usage!(state) do
+    cache_file = usage_cache_file(state.storage_dir)
+
+    data =
+      IO.iodata_to_binary([
```
alco, line 134: I think we should start the file with a version numbeer just in case. Makes it easier to change the format later.

packages/sync-service/lib/electric/stack_supervisor/telemetry.ex
alco: Removed this line but kept the function dinifition so that it could be referenced from Cloud.

PR #3834, opened 2026-02-12
fix: metrics from consumer

icehaunter:
the `electric.storage.transaction_stored` metric is not being reported because it was missing a stack id. I've added a test to check that it's indeed exported


robacourt (approved):

alco:
We have `stack-telemetry.lux` now that can be used to verify export of specific metrics. We should add this to that since the feedback loop is shorter compare to checking the fix in prod.

PR #3842, opened 2026-02-13
Improve concurrent request limit error messaging and documentation

KyleAMathews:
Improves the 503 error response when Electric's concurrent request limit is exceeded, replacing a generic "overloaded" message with specific, actionable information. Adds documentation for the `ELECTRIC_MAX_CONCURRENT_REQUESTS` config option and a troubleshooting guide.

## Root Cause

When too many clients connect simultaneously and exhaust the admission control permits, Electric returns a 503 with `{"code": "overloaded", "message": "Server is currently overloaded, please retry"}`. This tells the operator nothing about *which* limit was hit, what the limit is, or how to fix it — leading to unnecessary debugging.

## Approach

**Error message improvement**: The error code is now `"concurrent_request_limit_exceeded"` and the message includes the request kind (`initial` or `existing`) and the configured limit value. This lets operators immediately understand which pool is saturated and what to change.

**Documentation**: Added two new sections:
- Config docs for `ELECTRIC_MAX_CONCURRENT_REQUESTS` explaining the two-pool model (initial vs existing requests)
- Troubleshooting guide for the 503 error with three solutions: CDN with request collapsing (recommended), increasing the limit, or reducing concurrent subscriptions

**Key invariant**: The error message format must match what the documentation shows. Review caught a mismatch from an intermediate commit and corrected it.

## Verification

```bash
cd packages/sync-service && mix test test/electric/plug/router_test.exs
```

The test verifies the new error code and that the message includes `"Concurrent existing request limit exceeded"`.

## Files changed

- **`packages/sync-service/lib/electric/plug/serve_shape_plug.ex`** — Changed error code and message to include request kind and limit
- **`packages/sync-service/test/electric/plug/router_test.exs`** — Updated assertions for new error format
- **`website/docs/api/config.md`** — Added `ELECTRIC_MAX_CONCURRENT_REQUESTS` documentation
- **`website/docs/guides/troubleshooting.md`** — Added 503 concurrent request limit troubleshooting section

---

### DIFF
--- a/website/docs/api/config.md
+++ b/website/docs/api/config.md
@@ -369,17 +369,17 @@ Set `ELECTRIC_SHAPE_DB_ENABLE_MEMORY_STATS=true` to enable collection of memory 
 
 <EnvVarConfig
     name="ELECTRIC_MAX_CONCURRENT_REQUESTS"
-    defaultValue='{"initial": 300, "existing": 1000}'
-    example='{"initial": 500, "existing": 3000}'>
+    defaultValue='{"initial": 300, "existing": 10000}'
+    example='{"initial": 500, "existing": 30000}'>
 
 Maximum number of concurrent HTTP requests Electric will serve, as a JSON object with two keys:
 
 - `initial` &mdash; limit for initial sync requests (requests with `offset=-1`). Default: **300**.
-- `existing` &mdash; limit for ongoing requests (live long-polls and catch-up requests). Default: **1000**.
+- `existing` &mdash; limit for ongoing requests (live long-polls and catch-up requests). Default: **10000**.
 
 When the limit is exceeded, Electric responds with `503` and a `Retry-After` header. See the [troubleshooting guide](/docs/guides/troubleshooting#_503-mdash-concurrent-request-limit-exceeded) for details.
 
-Each `live=true` long-poll request holds a permit for up to 20 seconds, so the effective limit is determined by the number of concurrent shape subscriptions across all connected clients. Putting a [CDN with request collapsing](/docs/api/http#collapsing-live-requests) in front of Electric is the recommended way to handle high connection counts.
+Each `live=true` long-poll request holds the connection open for up to 20 seconds, so the effective limit is determined by the number of concurrent shape subscriptions across all connected clients. Putting a [CDN with request collapsing](/docs/api/http#collapsing-live-requests) in front of Electric is the recommended way to handle high connection counts.
 
 </EnvVarConfig>
 
--- a/website/docs/guides/troubleshooting.md
+++ b/website/docs/guides/troubleshooting.md
@@ -178,11 +178,11 @@ When too many clients are connected simultaneously, Electric responds with a `50
 ```json
 {
   "code": "concurrent_request_limit_exceeded",
-  "message": "Concurrent existing request limit exceeded (limit: 1000), please retry"
+  "message": "Concurrent existing request limit exceeded (limit: 10000), please retry"
 }
 ```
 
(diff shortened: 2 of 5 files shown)


website/docs/api/config.md
```diff
@@ -307,6 +307,24 @@ The amount of time a consumer process remains active without receiving transacti
...
+    example='{"initial": 500, "existing": 3000}'>
+
+Maximum number of concurrent HTTP requests Electric will serve, as a JSON object with two keys:
+
+- `initial` &mdash; limit for initial sync requests (requests with `offset=-1`). Default: **300**.
+- `existing` &mdash; limit for ongoing requests (live long-polls and catch-up requests). Default: **1000**.
+
+When the limit is exceeded, Electric responds with `503` and a `Retry-After` header. See the [troubleshooting guide](/docs/guides/troubleshooting#_503-mdash-concurrent-request-limit-exceeded) for details.
+
+Each `live=true` long-poll request holds a permit for up to 20 seconds, so the effective limit is determined by the number of concurrent shape subscriptions across all connected clients. Putting a [CDN with request collapsing](/docs/api/http#collapsing-live-requests) in front of Electric is the recommended way to handle high connection counts.
```
alco: >  long-poll request holds a permit for up to 20 seconds

This phrase is mentioned twice in this PR. "Permit" looks out of place here. The following sounds more natural, IMO

> long-poll request holds the connection open for up to 20 seconds.

website/docs/guides/troubleshooting.md
```diff
@@ -171,6 +171,35 @@ The problem will resolve itself as client/proxy caches empty. You can force this
...
+
+**Put a CDN in front of Electric** (recommended). Electric's caching headers are designed for CDN [request collapsing](/docs/api/http#collapsing-live-requests). When multiple clients poll the same shape at the same offset, the CDN collapses them into a single request to Electric and fans out the response. This dramatically reduces concurrent connections. See the [deployment guide](/docs/guides/deployment) for CDN setup.
+
+**Increase the concurrent request limit** as a stopgap. Set [`ELECTRIC_MAX_CONCURRENT_REQUESTS`](/docs/api/config#electric-max-concurrent-requests) to raise the limits:
+
+```shell
+ELECTRIC_MAX_CONCURRENT_REQUESTS='{"initial": 500, "existing": 3000}'
+```
+
+Live long-poll connections are lightweight Erlang processes, so most hardware can handle higher limits. The default of 1,000 is conservative.
```
alco: Could be a discussion for a different PR but this documentation doesn't explain `why` the limit is conservative. The rest of the narrative mentions things like "your cpu and mem could be fine" and "live connections are lightweight process". So why artificially limit the concurrency to 1000 then?

I think it should be 10k minimum.

* Partly because there's no logging on the server side when the limit is reached. The developer will only learn about it being hit if they have proper telemetry/alerting in place in their proxy between clients and Electric.

* Partly because having a CDN in front doesn't guarantee that Electric will be hit by fewer connections, when request collapsing cannot be applied to to different shape params.

* And partly because 10k simultaneous connections used to be a challenge, but at this point it should be table stakes even on modest HW.

> The term C10k was coined in 1999 by software engineer Dan Kegel, [...] serving 10,000 clients at once over 1 gigabit per second in that year.

> As of 2025, the problem has long since been solved, with the number of possible connections to a single computer being in the millions. 

https://en.wikipedia.org/wiki/C10k_problem

alco (approved):

KyleAMathews:
@alco suggestions applied!

alco:
I'm sorry, I missed your update last week. Rebased on main, fixed merge conflict. This is good to merge now.

PR #3851, opened 2026-02-16
Add `electric-has-data` response header

### LINKED ISSUE #3850: Add `electric-has-data` response header to distinguish empty live responses
## Problem

All successful live shape responses return HTTP 200, whether they contain actual change data or only an `up-to-date` control message. This makes it difficult to filter out empty responses in observability tools (access logs, Datadog, Prometheus, etc.) without parsing response bodies.

Common scenarios that return 200 with no data:
- Long-poll timeout with no changes (`no_change_response/1` in `api.ex`)
- Requests with `offset=now` (`do_serve_shape_log/1` `:now` clause)

Being able to distinguish these at the HTTP level would enable better dashboarding, alerting, and traffic analysis — e.g., filtering out noise from idle live connections.

## Proposal

Add a response header `electric-has-data` (or similar) to all shape responses:
- `electric-has-data: true` — response contains at least one data message
- `electric-has-data: false` — response contains only control messages (e.g., `up-to-date`)

This is cheap to implement because the server already tracks the `no_changes` boolean on the `Response` struct (`response.ex:37`) before headers are sent. No stream materialization is needed.

## Implementation outline

1. In `Response.put_resp_headers/2` (`response.ex`, around line 340), add the `electric-has-data` header based on `response.no_changes`:
   ```elixir
   |> put_resp_header("electric-has-data", if(response.no_changes, do: "false", else: "true"))
   ```

2. Ensure `no_changes` is correctly set for all response paths:
   - `no_change_response/1` already sets `no_changes: true`
   - The `:now` offset path and other control-only paths should also set it
   - Data-bearing responses default to `no_changes: false`

3. Expose the header in CORS `access-control-expose-headers` if needed for browser clients.

4. Optionally surface the header in the TypeScript client's response metadata for programmatic use.

## Non-goals
- Changing HTTP status codes (e.g., 204) — that would be a breaking change
- Counting the exact number of data messages — the log is a lazy stream and counting would require full materialization before sending headers, breaking the streaming architecture

balegas:
## Summary
- Adds a new `electric-has-data` response header (`"true"` / `"false"`) to all successful shape responses, allowing operators to distinguish data-bearing from empty responses without parsing bodies
- Fixes the `:now` offset path to correctly set `no_changes: true` on the response
- Auto-exposes the new header via CORS by adding it to `@electric_headers`

Fixes #3850

## Test plan
- [x] Added `refute response.no_changes` assertion on data response test in `api_test.exs`
- [x] Added `assert response.no_changes` assertion on "now" offset and long-poll timeout tests in `api_test.exs`
- [x] Added `electric-has-data` header assertions in `serve_shape_plug_test.exs` for data responses (`"true"`) and empty responses (`"false"`)
- [x] CORS test passes automatically since header is in `@electric_headers`
- [x] All 70 tests pass

### DIFF
--- a/.changeset/tidy-humans-poke.md
+++ b/.changeset/tidy-humans-poke.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Add `electric-has-data` response header to distinguish data-bearing responses from control-only responses (e.g. long-poll timeouts, `offset=now` requests).
--- a/packages/sync-service/lib/electric/shapes/api/response.ex
+++ b/packages/sync-service/lib/electric/shapes/api/response.ex
@@ -354,6 +354,10 @@ defmodule Electric.Shapes.Api.Response do
     Plug.Conn.put_resp_header(conn, "etag", etag(response))
   end
 
+  defp put_has_data_header(conn, %__MODULE__{status: status}) when status >= 400 do
+    conn
+  end
+
   defp put_has_data_header(conn, %__MODULE__{no_changes: true}) do
     Plug.Conn.put_resp_header(conn, @electric_has_data_header, "false")
   end
--- a/packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
+++ b/packages/sync-service/test/electric/plug/serve_shape_plug_test.exs
@@ -111,6 +111,8 @@ defmodule Electric.Plug.ServeShapePlugTest do
                  ]
                }
              }
+
+      assert get_resp_header(conn, "electric-has-data") == []
     end
 
     test "returns 400 for invalid offset", ctx do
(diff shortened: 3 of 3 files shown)


balegas:
@KyleAMathews I want to add this header to be able to filter out live requests with no data in OTEL. You might have opinions about it.

I don't think the latency or the amount of these requests are useful for observability and removing them would remove a lot of noise from our dashboards

balegas:
Thanks for the thorough review! Here's how I addressed each point:

**1. Missing changeset file** — ✅ Already had it, included in latest commit.

**2. `electric-has-data: true` on error responses** — ✅ Fixed. Added a guard clause to skip the header on `status >= 400`. Also added a test assertion that 400 responses don't include the header.

**3. `electric-has-data` absent from `:subset` responses** — Skipping this one. The `:subset` path is a separate code path for paginated initial data chunks that always contain data. I'd rather keep the two paths independent than couple them for consistency.

**4. No plug-level test for `offset=now`** — ✅ Added a `serve_shape_plug_test` that asserts `electric-has-data: false` and `electric-up-to-date: ""` for `offset=now` requests.

**5. TypeScript client** — Acknowledged, will track separately if needed.

All changes in 9d436eeb2.

PR #3855, opened 2026-02-17
Add fast-loop detection with automatic cache recovery

KyleAMathews:
ShapeStream now detects and self-heals tight retry loops caused by stale client-side caches. When the client is stuck making rapid requests at the same offset, it automatically clears the affected shape's cached state and retries from scratch. If the problem persists (proxy/CDN misconfiguration), it escalates through exponential backoff and eventually throws with diagnostic guidance.

## Root Cause

A user reported their Electric client creating and aborting tons of requests in a tight loop. Investigation revealed that stale `expiredShapesCache` localStorage entries were causing the client to repeatedly get 409s, reset to offset `-1`, and loop without making progress. A hard refresh (clearing JS state) fixed it — this PR makes the client self-heal automatically.

## Approach

**Offset-aware detection** — Tracks `{timestamp, offset}` pairs in a sliding 500ms window. Only flags a fast loop when 5+ requests are stuck at the **same offset**. This avoids false positives during normal rapid syncing where the offset advances with each response.

**Three-phase recovery ladder:**
1. **Count 1**: Clear the affected shape's entries from `expiredShapesCache` and `upToDateTracker`, reset the stream via `#reset()` (no handle = truly fresh start), retry immediately
2. **Counts 2-4**: Exponential backoff with full jitter (100ms base, 5s cap)
3. **Count 5**: Throw `FetchError(502)` with diagnostic message listing common proxy misconfigurations

**Scoped cache clearing** — Uses `canonicalShapeKey(this.#currentFetchUrl)` to delete only the affected shape's cache entries, leaving other active ShapeStream instances untouched.

**onError retry safety** — Fast-loop state (`consecutiveCount` + `entries`) is reset in `#start()` when `onError` returns retry options, so the retried stream gets a fresh chance instead of immediately throwing again.

## Key Invariants

- Fast-loop detection only runs on non-live requests (`!isUpToDate`); once live, counters reset
- Only same-offset requests within the window count — advancing offsets are normal syncing
- Cache clearing is scoped to the affected shape key, not global
- `onError` retry resets fast-loop state so recovery is possible

## Non-goals

- Not changing the existing 4-layer 409 defense (expired handle cache, stale response detection, fast-loop detection, prefetch protection) — this adds self-healing to layer 3
- Not adding a public API for cache clearing — this is automatic recovery

## Verification

```bash
cd packages/typescript-client
# 128 tests pass (9 test files)
```

## Files changed

- **`src/client.ts`** — Fast-loop detection with offset tracking, cache clearing + reset on first detection, `TROUBLESHOOTING_URL` constant extraction, fast-loop state reset in `#start()` retry path
- **`src/expired-shapes-cache.ts`** — Added `delete(shapeUrl)` method for scoped clearing
- **`src/up-to-date-tracker.ts`** — Added `delete(shapeKey)` method for scoped clearing
- **`test/stream.test.ts`** — 4 new tests: fast-loop detection + cache clearing, onError retry recovery, advancing-offset negative case, live-polling negative case
- **`vitest.unit.config.ts`** — Added `stream.test.ts` to unit test suite

### DIFF
--- a/packages/react-hooks/CHANGELOG.md
+++ b/packages/react-hooks/CHANGELOG.md
@@ -359,7 +359,6 @@
   Electric's TypeScript client is currently tightly coupled to PostgreSQL-specific options in its `ShapeStreamOptions` interface. As Electric plans to support multiple data sources in the future, we need to separate protocol-level options from source-specific options.
 
   ## Changes
-
   1. Created a new `PostgresParams` type to define PostgreSQL-specific parameters:
      - `table`: The root table for the shape
      - `where`: Where clauses for the shape
--- a/packages/typescript-client/CHANGELOG.md
+++ b/packages/typescript-client/CHANGELOG.md
@@ -82,7 +82,6 @@
   **Root cause:** When a 409 response arrives, the client marks the old handle as expired and fetches with a new handle. If a proxy ignores the `expired_handle` cache buster parameter and returns a stale cached response containing the old handle, the client would accept it and enter an infinite 409 loop.
 
   **The fix:**
-
   - In `#onInitialResponse`: Don't accept a shape handle from the response if it matches the expired handle in the expired shapes cache
   - In `getNextChunkUrl` (prefetch): Don't prefetch the next chunk if the response handle equals the `expired_handle` from the request URL
   - Added console warnings when this situation is detected to help developers debug proxy misconfigurations
@@ -106,7 +105,6 @@
   **Root cause:** When a page is hidden, the stream pauses and aborts in-flight prefetch requests. The aborted promises remained in the PrefetchQueue's internal Map. When the page became visible and the stream resumed, `consume()` returned the stale aborted promise, causing an AbortError to propagate to ShapeStream and stop syncing.
 
   **The fix:**
-
   - `PrefetchQueue.consume()` now checks if the request's abort signal is already aborted before returning it
   - `PrefetchQueue.abort()` now clears the internal map after aborting controllers
   - The fetch wrapper clears `prefetchQueue` after calling `abort()` to ensure fresh requests
@@ -153,7 +151,6 @@
 - b377010: Fix race condition where collections get stuck and stop reconnecting after rapid tab switching, particularly in Firefox.
 
   **Root cause:** Two race conditions in the pause/resume state machine:
-
   1. `#resume()` only checked for `paused` state, but `#pause()` sets an intermediate `pause-requested` state. When visibility changes rapidly, `#resume()` is called before the abort completes, leaving the stream stuck.
   2. Stale abort completions could overwrite the `active` state after `#resume()` has already started a new request.
 
@@ -331,7 +328,6 @@
   ```
 
   ## Common Use Cases
-
   - Authentication tokens that need to be refreshed
   - User-specific parameters that may change
   - Dynamic filtering based on current state
@@ -384,7 +380,6 @@
   ```
 
   ## Common Use Cases
-
   - Authentication tokens that need to be refreshed
   - User-specific parameters that may change
   - Dynamic filtering based on current state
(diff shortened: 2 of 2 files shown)


packages/typescript-client/CHANGELOG.md
samwillis, line 1: Not sure why this file is edited

packages/react-hooks/CHANGELOG.md
samwillis, line 1: Not sure why this file is edited

packages/typescript-client/test/stream.test.ts
```diff
@@ -17,7 +17,7 @@ describe(`ShapeStream`, () => {
     const eventTarget = new EventTarget()
     const requestArgs: Array<RequestInit | undefined> = []
     const fetchWrapper = (
-      ...args: Parameters<typeof fetch>
+      ..._args: Parameters<typeof fetch>
```
samwillis: Is this due to a lint change? (and the others below)

samwillis:
Chat GPTs review - obviously you are already looking at the CI issues.

---

## Review: fast-loop detection + self-healing cache clear

This is a really solid addition overall — the **same-offset** heuristic makes the fast-loop detector feel “safe” (normal rapid catchup advances offsets, so it shouldn’t trip), and the **first-hit self-heal** (scoped cache clear + `#reset()` fresh start) is a great UX improvement.

### ✅ What looks good

* Offset-aware detection: `{timestamp, offset}` window + `sameOffsetCount` threshold is the right guardrail.
* Only runs when `!isUpToDate` and resets tracking when live.
* Recovery ladder makes sense: **(1)** warn + clear caches + reset, **(2–4)** exponential backoff w/ jitter, **(5)** throw with concrete proxy/CDN guidance.
* Scoped clearing via `canonicalShapeKey(this.#currentFetchUrl)` (avoids nuking other active streams).

---

## 🚨 Blocking: CI is red (8 stream.test.ts timeouts)

Codecov + Blacksmith both report **8 tests timing out in `packages/typescript-client/test/stream.test.ts`** (e.g. “attach specified headers”, query param serialization, stable URL sorting, etc.). This needs resolving before merge.

My guess: adding `stream.test.ts` into the **unit** suite now runs tests that weren’t previously in that config / environment, and something is leaving a stream/timer/fetch pending (or fake timers aren’t being advanced while the new backoff uses `setTimeout`).

Suggestions to chase it down quickly:

* Ensure **every test tears down** the stream + abort controllers (`afterEach` should stop/abort and await completion).
* If any tests use **fake timers**, advance timers enough to cover any backoff path, or stub `setTimeout` / `Math.random` deterministically.
* If `stream.test.ts` is really an **integration-ish** suite, consider moving it to an integration config rather than the unit config (or split: pure URL/params tests into unit; streaming lifecycle tests elsewhere).

---

## 🟡 Non-blocking: a couple of correctness / clarity nits

1. **Cache key mismatch / naming confusion**

   * `expiredShapesCache.delete(shapeKey)` passes a canonical shape key, but the method signature is `delete(shapeUrl: string)`.
   * This might be fine if the cache is keyed by canonical key already — but if not, scoped clears won’t work.
   * At minimum I’d rename the param to `shapeKey` (or add a small comment) to avoid future footguns.

2. **Duplicate troubleshooting string in stale-cache warn**

   * In the stale-cache warning path, it currently includes the “For more information…” text twice (hardcoded URL + `${TROUBLESHOOTING_URL}`).
   * Easy tidy.

3. **Consecutive counter reset on “healthy progress”**

   * `#fastLoopConsecutiveCount` resets when the stream becomes live (or on `#start()` retry path), but not when offsets start advancing again during catchup.
   * Probably fine, but optional improvement: reset consecutive count when you observe forward progress for some period / N requests.

---

Once CI is green, I’m 👍 on the approach.

PR #3859, opened 2026-02-17
feat(sync-service): Add exclusive mode to support AWS EFS

magnetised:
NFS filesystems, such as AWS EFS, do not support SQLite WAL mode or simultaneous access by multiple processes, due to the limitiations of NFS locking semantics.

So to support AWS EFS this pr adds a new "exclusive mode" to the shape db configuration, enabled by setting `ELECTRIC_SHAPE_DB_EXCLUSIVE_MODE=true`, which multiplexes all reads and writes through a single SQLite connection and configures the database with `journal_mode=DELETE` as recommended by the [SQLite docs](https://sqlite.org/useovernet.html).

There is also a new `ELECTRIC_SHAPE_DB_STORAGE_DIR` configuration option which allows for hosting the shape db file in a different location from the shape data, e.g. on a non-networked filesystem.

These two options open up the possibility of an in-memory shape database, so enabling exclusive mode and setting `ELECTRIC_SHAPE_DB_STORAGE_DIR=:memory:` will do exactly that.

Since I was wiring up shapedb configuration, I've also added env-vars and configuration to set page cache size and durabilty (`PRAGMA cache_size=...` and `PRAGMA synchronous=...`).

Follow up work:

- Start some shape cleanup task to remove all orphaned shape data when the shape db is empty, to fix the situation when an ephemeral shape db is used alongside persistent shape data.


packages/sync-service/config/runtime.exs
```diff
@@ -277,7 +277,11 @@ config :electric,
       "ELECTRIC_REPLICATION_IDLE_TIMEOUT",
       &Electric.Config.parse_human_readable_time!/1,
       nil
-    )
+    ),
+  shape_db_exclusive_mode: env!("ELECTRIC_SHAPE_DB_EXCLUSIVE_MODE", :boolean, false),
+  shape_db_storage_dir: env!("ELECTRIC_SHAPE_DB_STORAGE_DIR", :string, nil),
+  shape_db_synchronous: env!("ELECTRIC_SHAPE_DB_SYNCHRONOUS", :string, nil),
+  shape_db_cache_size_kb: env!("ELECTRIC_SHAPE_DB_CACHE_SIZE_KB", :integer, nil)
```
alco: Instead of hardcoding units into the var name, would you like to steal my implementation of `parse_human_readable_size` from https://github.com/electric-sql/electric/pull/3274/files#diff-67d9229724ea8b9fd09358538f4cb9d4e1f1fd1f32eae1c82eda9b897229357b ?

magnetised: Nice idea. Done

PR #3862, opened 2026-02-17
Fix prefetch buffer serving GET responses to POST subset requests

KyleAMathews:
## Summary

Fix `createFetchWithChunkBuffer` incorrectly serving cached GET stream responses to POST subset/snapshot requests that share the same URL. This caused clients using POST-based subsets to receive stream chunk data instead of the expected subset snapshot JSON, triggering parse errors.

## Root Cause

The chunk prefetch queue in `createFetchWithChunkBuffer` used URL as the sole cache key when matching prefetched responses to incoming requests. It did not consider the HTTP method or request body. When a POST subset request targeted the same URL as a prefetched GET stream request, the queue would return the prefetched GET response to the POST caller — routing stream change messages into the subset handler, which expected a `{"metadata": {...}, "data": []}` snapshot object.

## Approach

Add a method guard at the top of `prefetchClient` that:
1. Extracts the HTTP method from the fetch arguments via a new `getRequestMethod` helper
2. For any non-GET request, aborts and clears the existing prefetch queue, then delegates directly to the underlying fetch client

This is the minimal fix — it scopes prefetch to GET-only (the only method that uses chunked streaming) without changing the queue's keying strategy or any other behavior.

### Key Invariants

- Prefetch queue only holds and serves GET stream responses
- A non-GET request always clears the queue to prevent stale entries from leaking into subsequent requests
- `getRequestMethod` mirrors the Fetch API spec: `init.method` takes precedence over `Request.method`, defaulting to `GET`

### Non-goals

- Changing the prefetch queue to key on method+URL (unnecessary complexity — non-GET requests don't benefit from prefetch)
- Fixing pre-existing silent `.catch(() => {})` blocks in `PrefetchQueue` (noted during review, out of scope)

## Verification

```bash
cd packages/typescript-client
pnpm vitest run test/fetch.test.ts  # 38/38 passing
```

New regression tests:
- POST with same URL as prefetched GET gets a fresh response
- POST does not trigger prefetch even with eligible response headers
- POST aborts queue so subsequent GET starts fresh
- `Request` object input correctly detected as non-GET
- Lowercase method strings handled via `.toUpperCase()`

## Files Changed

- **`packages/typescript-client/src/fetch.ts`** — Method guard in `prefetchClient`, new `getRequestMethod` helper
- **`packages/typescript-client/test/fetch.test.ts`** — 5 new regression tests for method-aware prefetch behavior

### DIFF
--- a/.changeset/fix-prefetch-post-method-mismatch.md
+++ b/.changeset/fix-prefetch-post-method-mismatch.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/client': patch
+---
+
+Fix prefetch buffer incorrectly serving cached GET responses to POST subset/snapshot requests that share the same URL, which could route stream chunks into the subset handler.


packages/typescript-client/test/fetch.test.ts
```diff
@@ -318,6 +318,126 @@ describe(`createFetchWithChunkBuffer`, () => {
...
+    mockFetch
+      .mockResolvedValueOnce(initialResponse)
+      .mockResolvedValueOnce(prefetchedGetResponse)
+      .mockResolvedValueOnce(postResponse)
+
+    await fetchWrapper(baseUrl)
+    await sleep()
+
+    const nextUrl = sortUrlParams(`${baseUrl}&handle=123&offset=456`)
+    const result = await fetchWrapper(nextUrl, { method: `POST` })
```
kevin-dp, line 344: Is this intended? I though we wanted the POST request to hit the same URL, but this one hits `nextUrl` whereas the GET request hits `baseUrl` (line 340).

KyleAMathews: Yes, this is intentional! The bug happens when a POST collides with a *prefetched* URL, not the original request URL.

Here's the sequence:

1. `fetchWrapper(baseUrl)` — initial GET, response has `handle=123` + `offset=456` headers
2. This triggers a prefetch of `nextUrl` (`baseUrl&handle=123&offset=456`) into the queue
3. `fetchWrapper(nextUrl, { method: 'POST' })` — POST arrives for the same URL that's sitting in the prefetch queue

If the POST hit `baseUrl` instead, it wouldn't collide with anything — the queue holds `nextUrl`, not `baseUrl`. The test needs to target `nextUrl` to reproduce the actual cache key collision.

PR #3868, opened 2026-02-18
fix(electric-telmetry): Handle non-existent storage-dir

magnetised:
Make sure the given `storage_dir` exists before writing the cached usage


magnetised:
> This is an extra OS call which is going to slow down things. How frequently is this called? The issue seems to be a race after a restart. Could we try to reproduce the scenario and make a more targeted fix?
> 
> I'm happy to go with it if we don't call this very frequently.

It's run every minute per stack. Don't think it will affect the system.

PR #3869, opened 2026-02-18
feat(sync-service): Add SQLite disk and memory usage telemetry

magnetised:
Include memory- and disk-usage of the shapes sqlite database in the periodic measurements.


packages/sync-service/lib/electric/stack_supervisor/telemetry.ex
```diff
@@ -52,6 +52,12 @@ defmodule Electric.StackSupervisor.Telemetry do
         ),
         Telemetry.Metrics.last_value("electric.postgres.replication.slot_confirmed_flush_lsn_lag",
           unit: :byte
+        ),
+        Telemetry.Metrics.last_value("electric.shape_db.sqlite.total_memory",
+          unit: {:byte, :kilobyte}
+        ),
+        Telemetry.Metrics.last_value("electric.shape_db.sqlite.disk_size",
+          unit: {:byte, :kilobyte}
```
msfstef: why not bytes? it's confusing to have metrics report in kilobytes when most storage metrics report in bytes

PR #3870, opened 2026-02-18
fix: use replication credentials for admin pool to fix publication management

KyleAMathews:
## Summary

Fixes publication privilege errors when `ELECTRIC_POOLED_DATABASE_URL` uses a different user than `DATABASE_URL`. The admin pool now connects with replication credentials — the same user that owns the publication — so publication management operations (adding/dropping tables) succeed regardless of the pooled connection user's privilege level.

## Root Cause

Publications are **created** on the replication connection (using `DATABASE_URL` credentials), but publication **management** (adding/dropping tables) ran on the admin pool, which uses `ELECTRIC_POOLED_DATABASE_URL` credentials. When these are different users, the ownership check in `Configuration.check_publication_status!/2` fails:

```sql
pg_get_userbyid(p.pubowner) = current_role as can_alter_publication
```

This breaks on Crunchy Bridge specifically because superusers cannot connect through PgBouncer, so the pooled URL must use a lower-privilege application user that doesn't own the publication.

## Approach

In `Connection.Manager.handle_continue(:start_connection_pool, ...)`:

1. **Validate both connection credential sets** — pooled (for snapshot queries) and replication (for admin operations) — using a `with` chain instead of a single `case`
2. **Route credentials per pool role** — `:admin` pool gets `replication_conn_opts`, `:snapshot` pool gets `pooled_conn_opts`
3. **Tag `with` clauses** so the `else` block can distinguish which validation failed and route to the correct `shutdown_or_reconnect` pid_type (`:pools` vs `:admin_pool`)

## Key Invariants

- The replication credentials are already validated and cached by the time `:start_connection_pool` runs (from the earlier `:start_replication_client` step), so no extra DB connections are made
- When `ELECTRIC_POOLED_DATABASE_URL` is not set, it defaults to `DATABASE_URL`, so both credential sets are identical — behavior is unchanged
- Error attribution is correct: pool validation failures report as `:pools`, replication validation failures report as `:admin_pool`

## Non-goals

- Improving the `can_alter_publication` SQL check itself (could be a follow-up)
- Documentation updates for Crunchy Bridge integration (separate PR)
- Handling the pre-existing bare `{:ok, pool_pid} =` pattern match on `Pool.start_link`

## Verification

The change is in connection pool initialization logic. To verify:
- Deploy with `ELECTRIC_POOLED_DATABASE_URL` using a different (lower-privilege) user than `DATABASE_URL`
- Confirm publication management operations succeed
- Confirm snapshot queries still route through the pooled connection

## Files changed

- **`packages/sync-service/lib/electric/connection/manager.ex`** — Switch admin pool from pooled connection credentials to replication credentials; tag `with` clauses for proper error attribution

### DIFF
--- a/integration-tests/tests/ipv6-to-ipv4-fallback.lux
+++ b/integration-tests/tests/ipv6-to-ipv4-fallback.lux
@@ -23,7 +23,7 @@
   ??[info] Lock acquired
   ??tcp connect (local-ipv4-only.electric-sql.dev:$pg_pooler_host_port): non-existing domain - :nxdomain
   ??[warning] Database connection failed to find valid IPv6 address for local-ipv4-only.electric-sql.dev - falling back to IPv4
-  ?+\[info\] Connection pool \[admin: $pg_host_name:$pg_pooler_host_port\] is ready
+  ?+\[info\] Connection pool \[admin: $pg_host_name:$pg_host_port\] is ready
   ?\[info\] Connection pool \[snapshot: $pg_host_name:$pg_pooler_host_port\] is ready
   ??[info] Starting replication from postgres
 


balegas:
@claude what kinds of queries do we do with the admin pool? do we need it to be large?

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -383,53 +383,68 @@ defmodule Electric.Connection.Manager do
...
+          # to have new queries queued for a longer time than failing them continuously
+          pool_queue_config =
+            if pool_role == :snapshot,
+              do: [queue_target: 5_000, queue_interval: 10_000],
+              else: []
+
+          conn_opts =
+            if pool_role == :admin,
+              do: replication_conn_opts,
+              else: pooled_conn_opts
```
alco: This looks a bit brittle. I've made a change above to show a better way.
```suggestion
```

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -383,53 +383,68 @@ defmodule Electric.Connection.Manager do
...
+           {:pool, validate_connection(pooled_connection_opts(state), :pool, state)},
+         {:replication, {:ok, replication_conn_opts, state}} <-
+           {:replication,
+            validate_connection(replication_connection_opts(state), :replication, state)} do
+      pool_sizes = pool_sizes(Keyword.get(state.pool_opts, :pool_size, 2))
+
+      state =
+        [:snapshot, :admin]
+        |> Enum.filter(fn role -> is_nil(state.pool_pids[role]) end)
+        |> Enum.reduce(state, fn pool_role, state ->
```
alco: A better approach is to zip roles with their corresponding conn_opts at the start of the loop, to avoid conditional lookups in the loop body:
```suggestion
        [{:snapshot, pooled_conn_opts}, {:admin, replication_conn_opts}]
        |> Enum.filter(fn {role, _conn_opts} -> is_nil(state.pool_pids[role]) end)
        |> Enum.reduce(state, fn {pool_role, conn_opts}, state ->
```

packages/sync-service/lib/electric/connection/manager.ex
```diff
@@ -383,53 +383,68 @@ defmodule Electric.Connection.Manager do
...
+    else
       # the ConnectionResolver process was killed, as part of the application
       # shutdown in which case we'll be killed next so just return here
-      {:error, :killed} ->
+      {_source, {:error, :killed}} ->
         {:noreply, state}
 
-      {:error, reason} ->
+      {:pool, {:error, reason}} ->
         shutdown_or_reconnect(reason, :pools, state)
```
alco:
```suggestion
        shutdown_or_reconnect(reason, :snapshot_pool, state)
```

PR #3871, opened 2026-02-18
Fix missed updates between snapshot and stream resume

KyleAMathews:
## Summary

Fixes a bug in on-demand mode where updates committed between a `requestSnapshot()` call and the stream's next live poll are silently lost. The stream now advances its internal offset/handle from the snapshot response headers, ensuring it resumes from the correct position.

## Root Cause

On-demand mode creates streams with `offset: "now"`, which makes the server immediately return up-to-date without sending any data. When `requestSnapshot()` is later called, it fetches subset data and injects it into the shape — but the stream's sync state still reflects the `"now"` position. When the pause lock releases and the stream resumes live polling, it polls from this stale `"now"` position rather than from the snapshot's position, missing any mutations committed between the snapshot response and the next live request.

The sequence:
1. Stream connects with `offset: "now"` → server creates shape, returns up-to-date
2. `requestSnapshot()` pauses the stream, fetches subset data at server position X
3. A mutation is committed at position X+1
4. Pause lock releases, stream resumes — but still at `"now"` instead of position X, skipping position X+1

## Approach

On cold starts, after `requestSnapshot()` injects the snapshot data, it now extracts the `electric-offset` and `electric-handle` response headers from the snapshot response and feeds them into the state machine via `handleResponseMetadata()`. This advances the stream's position to match the snapshot, so when the pause lock releases, the stream resumes from the right spot.

Key detail: `PausedState.handleResponseMetadata` is a no-op (delegates to base), so the code unwraps to the inner state (typically `LiveState`), applies the transition, and re-wraps in `PausedState`.

```typescript
// PausedState's handleResponseMetadata is a no-op, so unwrap to
// the inner state and re-wrap after.
const paused = this.#syncState instanceof PausedState ? this.#syncState : null
const innerState = paused?.previousState ?? this.#syncState
const transition = innerState.handleResponseMetadata({ ... })
if (transition.action === `accepted`) {
  this.#syncState = paused ? new PausedState(transition.state) : transition.state
}
```

## Key Invariants

- After `requestSnapshot()` completes, the stream's offset/handle reflects the snapshot's position
- The stream must be up-to-date before `requestSnapshot()` is called (enforced by `Shape.#awaitUpToDate()`) — this ensures the shape exists on the server and change tracking is active
- If the server doesn't return offset/handle headers, the advancement is skipped (no-op)
- If the state machine rejects the transition, a warning is logged

## Non-goals

- Does not change the state machine itself — only feeds it the correct metadata at the right time
- Does not modify how tanstack-db configures `offset: "now"` — that remains the correct default for on-demand mode

## Trade-offs

**Alternative: Change `offset: "now"` to `offset: "-1"` in tanstack-db.** This would make the stream download the full change log from the beginning, eliminating the gap. Rejected because it's expensive for large tables and unnecessary — the snapshot response already contains the position we need.

## Verification

```bash
cd packages/typescript-client && pnpm exec vitest run test/client.test.ts -t "should not miss updates"
```

The test `should not miss updates between cold-start snapshot and stream resume` exercises the exact production scenario: creates a stream with `offset: "now"` (matching on-demand mode), waits for up-to-date, takes a snapshot, mutates a row, and verifies the stream picks up the update.

## Files Changed

- **`src/client.ts`** — `fetchSnapshot` returns offset/handle from response headers; `requestSnapshot` advances sync state after snapshot; adds warning log for rejected transitions
- **`test/client.test.ts`** — Integration test verifying updates after cold-start snapshot are not missed (uses `offset: "now"` to match real on-demand behavior)
- **`.changeset/fix-snapshot-offset-advancement.md`** — Patch changeset

---

Reported on Discord: https://discord.com/channels/719702312431386674/1369767025723052134/1473804587671617597

### DIFF
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -1664,16 +1664,10 @@ export class ShapeStream<T extends Row<unknown> = Row>
       )
       this.#onMessages(dataWithEndBoundary, false)
 
-      // On cold start the stream hasn't connected yet, so its offset is
-      // still at the initial value. Advance it to the snapshot's position
-      // so the stream catches every update after the snapshot with no gaps.
+      // On cold start the stream's offset is still at "now". Advance it
+      // to the snapshot's position so no updates are missed in between.
       if (responseOffset !== null || responseHandle !== null) {
-        // PausedState's handleResponseMetadata is a no-op, so unwrap to
-        // the inner state and re-wrap after.
-        const paused =
-          this.#syncState instanceof PausedState ? this.#syncState : null
-        const innerState = paused?.previousState ?? this.#syncState
-        const transition = innerState.handleResponseMetadata({
+        const transition = this.#syncState.handleResponseMetadata({
           status: 200,
           responseHandle,
           responseOffset,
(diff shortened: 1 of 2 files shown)


KyleAMathews:
Closing this PR — investigation revealed the offset advancement isn't needed and can actually cause missed updates due to a server-side shape warmup race. The stream already handles cold-start correctly by catching up from offset `-1` through the change log, with SnapshotTracker deduplicating against snapshot data.

PR #3874, opened 2026-02-19
fix(sync-service): Add disk usage to metrics

magnetised:
Include the new disk usage metrics in the periodic collection.


packages/sync-service/lib/electric/stack_supervisor/telemetry.ex
```diff
@@ -60,7 +60,8 @@ defmodule Electric.StackSupervisor.Telemetry do
       [
         {__MODULE__, :count_shapes, [stack_id]},
         {__MODULE__, :report_write_buffer_size, [stack_id]},
-        {__MODULE__, :report_retained_wal_size, [stack_id, config.replication_opts[:slot_name]]}
+        {__MODULE__, :report_retained_wal_size, [stack_id, config.replication_opts[:slot_name]]},
+        {__MODULE__, :report_disk_usage, [stack_id]}
```
msfstef, line 64: do we need to explicitly export it as part of stack telemetry as well?

magnetised: cf discord and ^

packages/electric-telemetry/lib/electric/telemetry/stack_telemetry.ex
```diff
@@ -99,7 +99,8 @@ defmodule ElectricTelemetry.StackTelemetry do
       ),
       distribution("electric.postgres.replication.transaction_received.operations"),
       distribution("electric.storage.transaction_stored.replication_lag", unit: :millisecond),
-      last_value("electric.storage.used", unit: {:byte, :kilobyte}),
+      last_value("electric.storage.used.bytes", unit: :byte),
+      last_value("electric.storage.used.measurement_duration", unit: :millisecond),
```
msfstef: the measurement duration being last value confuses me - do we want to be exporting it and why as a last value and not e.g. a distribution?

magnetised: want to keep an eye on it -- useful to know how long its taking in prod because we don't want it to cause problems. since its only measured once a minute a distribution seems pointless.

PR #3881, opened 2026-02-19
fix(electric-telemetry): Send disk usage in bytes and use correct metric name

magnetised:


packages/electric-telemetry/lib/electric/telemetry/reporters/call_home_reporter.ex
```diff
@@ -68,7 +68,7 @@ defmodule ElectricTelemetry.Reporters.CallHomeReporter do
           stored_bytes: sum("electric.storage.transaction_stored.bytes", unit: :byte),
           stored_transactions: sum("electric.storage.transaction_stored.count"),
           stored_operations: sum("electric.storage.transaction_stored.operations"),
-          total_used_storage_kb: last_value("electric.storage.used", unit: {:byte, :kilobyte}),
+          total_used_storage_kb: last_value("electric.storage.used.bytes", unit: :byte),
```
msfstef:
```suggestion
          total_used_storage: last_value("electric.storage.used.bytes", unit: :byte),
```

magnetised: doh. we should probably leave this then, since the warning at the top makes it very clear that you can't change these. I'll keep the fix to the metric but revert the rest

PR #3888, opened 2026-02-20
Fix parameter validation for 10+ sequential params

### LINKED ISSUE #3887: Parameter validation fails for 10+ sequential params due to map key ordering
## Bug

`Electric.Shapes.Shape.Validators.validate_parameters/1` rejects valid, sequentially numbered parameters when there are 10 or more of them.

## Root cause

`all_keys_are_numbers/1` iterates over the params map using `map_while_ok/2`, which processes keys in Erlang term order. For string keys, this is lexicographic order:

```
"1", "10", "11", ..., "19", "2", "20", ...
```

The parsed integer keys are then passed to `all_keys_are_sequential/1`, which checks them positionally without sorting:

```elixir
defp all_keys_are_sequential(keys) do
  Enum.with_index(keys, fn key, index -> key == index + 1 end)
  |> Enum.all?()
  |> if(
    do: :ok,
    else: {:error, {:params, "Parameters must be numbered sequentially, starting from 1"}}
  )
end
```

With 10+ keys, the second element is `10` (from key `"10"`), but the check expects `2` → validation fails.

With fewer than 10 keys, lexicographic and numeric order happen to agree, so the bug is not triggered.

## Reproduction

```elixir
test "sequential parameters with 10+ keys are accepted", %{inspector: inspector} do
  params = Map.new(1..10, fn i -> {"#{i}", "#{i * 10}"} end)

  assert {:ok, _} =
           Shape.new("item",
             inspector: inspector,
             where:
               "value = $1 OR value = $2 OR value = $3 OR value = $4 OR value = $5 OR value = $6 OR value = $7 OR value = $8 OR value = $9 OR value = $10",
             params: params
           )
end
```

## Fix

Sort the keys numerically before checking sequentiality:

```elixir
defp all_keys_are_sequential(keys) do
  keys
  |> Enum.sort()
  |> Enum.with_index(fn key, index -> key == index + 1 end)
  |> Enum.all?()
  |> if(
    do: :ok,
    else: {:error, {:params, "Parameters must be numbered sequentially, starting from 1"}}
  )
end
```

alco:
## Summary
- Parameter validation (`all_keys_are_sequential/1`) rejected valid sequential params when there are 10 or more, because map keys are iterated in Erlang term order (lexicographic for strings: `"1", "10", "11", ..., "2"`), not numeric order
- Fix: sort parsed integer keys before checking sequentiality
- Added a regression test with 20 params

Fixes #3887

## Test plan
- [x] New test `"sequential parameters with 10+ keys are accepted"` passes
- [x] Existing `"skipped parameter positions show an error"` tests still pass

### DIFF
--- a/packages/sync-service/test/electric/shapes/shape_test.exs
+++ b/packages/sync-service/test/electric/shapes/shape_test.exs
@@ -736,13 +736,14 @@ defmodule Electric.Shapes.ShapeTest do
            "CREATE TABLE IF NOT EXISTS item (id INT PRIMARY KEY, value INT NOT NULL)"
          ]
     test "sequential parameters with 10+ keys are accepted", %{inspector: inspector} do
-      params = Map.new(1..10, fn i -> {"#{i}", "#{i * 10}"} end)
+      params = Map.new(1..20, fn i -> {"#{i}", "#{i * 10}"} end)
+
+      where = Enum.map_join(1..20, " OR ", fn i -> "value = $#{i}" end)
 
       assert {:ok, _} =
                Shape.new("item",
                  inspector: inspector,
-                 where:
-                   "value = $1 OR value = $2 OR value = $3 OR value = $4 OR value = $5 OR value = $6 OR value = $7 OR value = $8 OR value = $9 OR value = $10",
+                 where: where,
                  params: params
                )
     end


alco:
@balegas A long URL is more likely to be caused by constants such as IDs or strings embedded in the query.

PR #3894, opened 2026-02-23
fix(typescript-client): Accept fresh origin responses with matching expired handle

KyleAMathews:
Fixes an infinite retry loop (~2,750 requests/30s per shape) where the client treats fresh origin responses as stale CDN hits when the server legitimately reuses the same shape handle.

## Root Cause

When a shape's handle is marked expired (via 409), it's stored in `expiredShapesCache` (localStorage). If the server later reuses the same handle (shape definition unchanged), every response — including fresh ones directly from origin — matches the expired handle. The stale detection in `checkStaleResponse()` only checks `responseHandle === expiredHandle` and has no way to distinguish "stale CDN copy" from "fresh origin response that happens to have the same handle." This causes the client to endlessly retry with cache busters.

## Approach

Use the HTTP `age` header (RFC 7234) to detect freshness **before** passing `expiredHandle` to the state machine. CDNs set `age` to indicate how long a response has been cached — a fresh origin response has no `age` header (or `age: 0`), while a stale CDN hit has `age > 0`. This works across Cloudflare, Fastly, Vercel, etc.

The check lives in `#onInitialResponse` (client.ts), keeping the state machine untouched:

```typescript
if (expiredHandle && shapeHandle === expiredHandle) {
  const age = parseInt(headers.get(`age`) ?? ``, 10)
  if (!age) { // NaN (no header) and 0 are both falsy → fresh
    expiredShapesCache.removeExpired(shapeKey!)
    expiredHandle = null
  }
}
```

If the response is fresh, clear the expired entry and pass `expiredHandle = null` — the state machine sees no stale condition and proceeds normally.

### Key Invariants

- `age > 0` → stale CDN hit → existing stale-retry logic handles it (unchanged)
- `age` absent or `0` → fresh from origin → clear expired entry, accept the handle
- No changes to the state machine (`shape-stream-state.ts`) at all

### Non-goals

- Not adding logging/warnings for malformed `age` headers — `parseInt` returns `NaN` which is falsy, defaulting to "fresh," which is the safer failure mode (avoids infinite retries)
- Not refactoring existing stale detection logic — this is a targeted fix at the decision boundary

## Verification

```bash
cd packages/typescript-client && pnpm test
```

All 427 tests pass, including 5 new tests covering this fix.

## Files Changed

| File | Change |
|------|--------|
| `src/expired-shapes-cache.ts` | Add `removeExpired(shapeUrl)` method to delete a single cache entry |
| `src/client.ts` | Add `age` header check in `#onInitialResponse` before passing `expiredHandle` to state machine |
| `test/expired-shapes-cache.test.ts` | 5 new tests (removeExpired unit tests, fresh origin acceptance, age:0 edge case, stale CDN detection); add `age` headers to 5 existing stale CDN mocks |

---

### DIFF
--- a/.changeset/fix-fresh-origin-expired-handle.md
+++ b/.changeset/fix-fresh-origin-expired-handle.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/client': patch
+---
+
+Fix infinite retry loop when server reuses the same shape handle after a 409. Use the HTTP `age` header to distinguish fresh origin responses from stale CDN hits, preventing unnecessary cache-busted retries.


K-Mistele:
is this related to #3855 ?

KyleAMathews:
Not really — this PR is actually just wrong so closing.

PR #3901, opened 2026-02-24
fix(sync-service): prevent mid-transaction chunk offset exposure

### LINKED ISSUE #3900: Bug: Chunk boundary race can expose mid-transaction offsets to API readers
## Problem

During processing of a large transaction, chunk boundaries written to ETS become visible to concurrent API readers **before** `last_seen_txn_offset` is updated. This creates a window where a client can receive an offset from a partially-processed transaction, and then get a 400 "offset out of bounds" on the next request.

## Reproduction scenario

1. Client is subscribed to a shape on `offer_items` table
2. User applies a template that inserts many rows in a single transaction (large enough to trigger chunk boundary writes mid-transaction)
3. During transaction processing in `write_loop.ex`:
   - Chunk boundary at offset Z is written to ETS via `update_chunk_boundaries_cache` (line ~188) — **immediately visible to readers**
   - But `register_complete_txn` (line 362-376) hasn't run yet — `last_seen_txn_offset` is still at Y (previous transaction)
4. Concurrent API request reads `chunk_end_offset = Z` from chunk boundaries, returns it to client in response headers
5. Client's next request sends `offset=Z`
6. Server calls `fetch_latest_offset()` → returns Y (still not updated)
7. Y < Z → `is_out_of_bounds` guard fires → 400 error after grace period timeout

## Code references

**Chunk boundary write (visible to readers mid-transaction):**
`packages/sync-service/lib/electric/shape_cache/pure_file_storage/write_loop.ex:179-188`
```elixir
defp maybe_write_closing_chunk_boundary(...) do
  writer_acc(acc, chunk_started?: false, bytes_in_chunk: 0)
  |> ensure_chunk_file_open(state)
  |> write_to_chunk_file(ChunkIndex.make_half_entry(offset, position, 0))
  |> add_closing_chunk_boundary_to_cache(offset, position)
  |> update_chunk_boundaries_cache(opts)  # ← VISIBLE TO READERS (ETS)
  |> flush_buffer(state)
end
```

**`last_seen_txn_offset` update (happens AFTER full transaction):**
`packages/sync-service/lib/electric/shape_cache/pure_file_storage/write_loop.ex:362-376`
```elixir
defp register_complete_txn(...) do
  # This runs AFTER the entire reduce loop over all operations
  # Between chunk boundary writes and this point, readers see stale last_offset
end
```

**Out-of-bounds guard:**
`packages/sync-service/lib/electric/shapes/api.ex:44-45`
```elixir
defguardp is_out_of_bounds(request)
          when LogOffset.is_log_offset_lt(request.last_offset, request.params.offset)
```

## Customer context

Customer applying templates that create many `offer_items` rows in a single transaction. Seeing:
- 400 "offset out of bounds for this shape" errors
- `TimeoutWaitingForTxIdError` for `[sync-offer_items] Timeout waiting for txId: 738842934` — consistent with a large transaction being processed

## Proposed fix

Cap `chunk_end_offset` to never exceed `last_seen_txn_offset` when determining the response offset. In `determine_log_chunk_offset` (api.ex:412-418), add a guard:

```elixir
defp determine_log_chunk_offset(%Request{} = request) do
  chunk_end = get_chunk_end_offset(request)
  # Never return an offset beyond what's fully committed
  capped = min(chunk_end, request.last_offset)
  ...
end
```

Alternatively, defer chunk boundary ETS updates until `register_complete_txn` completes, ensuring readers never see mid-transaction state.

## Related

- #3760 — Shape invalidation race (offset regression during cleanup)
- Client-side: The client should also auto-recover from "offset out of bounds" rather than treating it as fatal (separate issue)

KyleAMathews:
Fix a race condition where chunk boundary offsets could become visible to API clients before the transaction that produced them had fully committed, potentially causing readers to observe incomplete transaction data.

Fixes #3900

## Root Cause

The `WriteLoop` published chunk boundary offsets to the shared ETS cache on every individual chunk boundary write (both opening and closing). Since a single transaction can span multiple chunks, a concurrent API reader could observe a chunk boundary pointing into the middle of an in-progress transaction. The reader would then serve data up to that offset — data that was only partially written.

Additionally, the API layer's `determine_log_chunk_offset/1` used the chunk boundary offset as-is without clamping it to the known committed offset (`last_offset`), so even a slightly stale boundary could reference an offset beyond what was fully committed.

## Approach

Two complementary fixes providing defense in depth:

**1. Deferred cache publication (write_loop.ex)**

Moved `update_chunk_boundaries_cache` out of `maybe_write_opening_chunk_boundary` and `maybe_write_closing_chunk_boundary` into `register_complete_txn`. Boundaries are now only visible in the ETS cache after the entire transaction stream has been consumed and the transaction is registered as complete.

```elixir
# register_complete_txn now handles both persistence metadata AND chunk boundaries
|> update_persistance_metadata(state, prev_persisted_txn)
|> update_chunk_boundaries_cache(opts)
```

**2. Offset clamping (api.ex)**

Added `LogOffset.min(chunk_end, last_offset)` in `determine_log_chunk_offset/1` so the API never returns a `chunk_end_offset` beyond the shape's known `last_offset`, even if the cache contains a boundary from a transaction that hasn't fully propagated to the API layer yet.

## Key Invariants

- Chunk boundaries in the ETS cache must only reference fully committed transactions
- The API must never direct clients to an offset beyond `last_offset`
- `register_complete_txn` is the single point where both persistence metadata and chunk boundaries become visible to readers

## Non-goals

- Did not add error handling for `update_chunk_boundaries_cache` ETS failures — if the shape's ETS entry is deleted during cleanup, silently discarding is correct behavior since the shape won't be recovered
- Did not address pre-existing silent error handling in the SSE path or `IO.binwrite` return value handling

## Verification

```bash
mix test test/electric/shapes/api_test.exs test/electric/shape_cache/pure_file_storage_test.exs
```

New tests:
- `pure_file_storage_test.exs`: "does not expose boundaries before transaction completes" — uses `Stream.resource` to pause mid-transaction and verify boundaries aren't visible until completion
- `api_test.exs`: "caps chunk end offset to last_offset" — verifies the API clamps a chunk boundary that exceeds `last_offset`

## Files changed

| File | Change |
|------|--------|
| `write_loop.ex` | Move `update_chunk_boundaries_cache` from per-boundary to `register_complete_txn` |
| `api.ex` | Add `LogOffset.min` clamping in `determine_log_chunk_offset/1` |
| `pure_file_storage_test.exs` | Add test verifying mid-transaction boundary invisibility |
| `api_test.exs` | Add test verifying chunk end offset clamping |

### DIFF
--- a/.changeset/fix-chunk-offset-race.md
+++ b/.changeset/fix-chunk-offset-race.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Prevent chunk boundary offsets from being exposed to clients mid-transaction, which could cause readers to observe incomplete transaction data.


alco:
Turns out we already have a code path that accepts requests with the offset slightly in the future and waits for half of the long poll timeout to see the txn commit and return a valid response. The problem is that what we see in prod shows that it doesn't see the txn commit within the 20 seconds.

That seemed implausible to me, so I looked for a different explanation and found the reason - https://github.com/electric-sql/electric/pull/3907.

I'm very wary about changing the returned offset (by capping it) since that has wide implications. Even inside the Api `chunk_end_offset` is references in different branches a shape request could take. Replacing the real chunk end offset with a capped value might cause other unintended side effects.

I'm currently coming up with a test that would reproduce the root cause explained in my PR. If it confirms the hypothesis, I think we can close this one in favour of that.

alco:
Superseded by https://github.com/electric-sql/electric/pull/3907

PR #3903, opened 2026-02-24
Make memstat SQLite extension optional in statistics collection

KyleAMathews:
## Summary
This change makes the `memstat` SQLite extension optional during statistics collection, allowing the system to gracefully degrade when the extension is unavailable while still collecting disk statistics.

## Key Changes
- **Graceful extension loading**: Modified `handle_continue/2` to catch and log errors when loading the `memstat` extension instead of failing the entire initialization
- **Conditional statistics queries**: Introduced `stats_query/1` helper function that returns different SQL queries based on whether `memstat` is available
  - With memstat: Collects both disk and memory statistics
  - Without memstat: Collects only disk statistics via dbstat
- **Safe stats extraction**: Changed `analyze_stats/2` to use `Map.get/3` with default values instead of pattern matching, preventing crashes when memory statistics are unavailable
- **State tracking**: Added `memstat_available?` flag to the GenServer state to track extension availability throughout the process lifecycle

## Implementation Details
- When `memstat` extension fails to load, a warning is logged explaining that memory statistics will not be available
- The system continues operating normally, collecting disk-level statistics from the `dbstat` virtual table
- All memory-related metrics default to 0 when the extension is unavailable, maintaining consistent state structure

https://claude.ai/code/session_012aWeWH2G2D7GnLUTYbREKH

### DIFF
--- a/.changeset/lean-soft-rude.md
+++ b/.changeset/lean-soft-rude.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Handle missing memstat SQLite extension gracefully instead of crashing on startup. When the extension is unavailable, memory statistics are simply omitted from the periodic stats collection.


alco:
@claude add the missing test mentioned in the code review feedback

alco:
Creating a unit test for this is tricky due to the need to mock the `enable_extension` call.

I've tested it manually by setting `memstat_available?` to `false`.

PR #3907, opened 2026-02-25
fix(sync-service): subscribe to shape events in out-of-bounds handler

alco:
## Summary

- Fix the root cause of the ~20s hangs observed in out-of-bounds 400 responses: the request process was not subscribed to shape events before entering `hold_until_change()`
- For `live=false` requests, `listen_for_new_changes()` skips the Registry subscription entirely. But the out-of-bounds clause in `do_serve_shape_log()` calls `handle_live_request()` unconditionally, which enters a `receive` block waiting for `:new_changes` messages that will never arrive without a subscription
- Add `ensure_subscribed()` that registers in the shape events Registry if `new_changes_ref` is nil, called before `handle_live_request()` in the out-of-bounds path

Fixes https://github.com/electric-sql/electric/issues/3900

## Context

Related to #3900 and #3901.

#3901 addresses the chunk boundary visibility race (deferring ETS updates and clamping offsets), which reduces the likelihood of hitting out-of-bounds. But even with that fix, a narrow race window remains between `register_complete_txn` updating ETS and the API reading it. When an out-of-bounds request with `live=false` slips through, the missing subscription causes a guaranteed 20s hang (`long_poll_timeout / 2`).

This PR fixes the subscription gap so that even if the race occurs, the request process will receive `:new_changes` from the consumer and recover quickly instead of waiting for the timeout.

## Test plan

- [x] Existing API tests pass (`mix test test/electric/shapes/api_test.exs`)
- [x] Add test for `live=false` out-of-bounds request recovering via subscription

### DIFF
--- a/packages/sync-service/test/electric/shapes/api_test.exs
+++ b/packages/sync-service/test/electric/shapes/api_test.exs
@@ -974,7 +974,7 @@ defmodule Electric.Shapes.ApiTest do
             {response, response_body(response)}
           end)
 
-        assert_receive {:trace, _, :call, {Api, :hold_until_change, _}}
+        assert_receive {:trace, _, :call, {Api, :hold_until_change, _}}, @receive_timeout
 
         # Simulate new changes arriving via Registry.
         # Without ensure_subscribed, this dispatch would find no registered


alco:
The new test occasionally fails with Task timeout, meaning the request process doesn't receive new changes.

Can it be explained by registration propagation delay in Registry? Looking...

alco:
The last two things Claude is complaining about are non-issues.
> Important (Should Fix): Fragile Registry.dispatch callback — still not addressed

> Suggestion (Nice to Have): Trace session not cleaned up

Both are WON'T FIX.

PR #3933, opened 2026-03-02
Replace exqlite with esqlite

magnetised:


packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/sqlite3.ex
```diff
@@ -0,0 +1,206 @@
...
+
+  ## API mapping
+
+  | Exqlite.Sqlite3                     | :esqlite3 / notes                          |
+  |-------------------------------------|--------------------------------------------|
+  | open(path, opts)                    | open(uri) – opts encoded as URI params      |
+  | close(conn)                         | close(conn)                                |
+  | execute(conn, sql)                  | exec(conn, sql)                            |
+  | prepare(conn, sql)                  | prepare(conn, sql)                         |
+  | release(conn, stmt)                 | no-op – GC'd by esqlite                    |
```
alco: GC'd at connection close? Any details that would be useful to mention here?

alco, line 18: At what point does it get GC'd? Better document this?

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/sqlite3.ex
```diff
@@ -0,0 +1,206 @@
...
+  | bind(stmt, binds)                   | bind(stmt, binds)                          |
+  | step(conn, stmt)                    | step(stmt) – conn arg dropped              |
```
alco, line 20: Where is the conn state held for this with esqlite3? Process dict or the statement itself?

magnetised: For the `step/2` function? Everything's in the nif resource.

alco: Yes, for step. I'm unsure what the NIF resource is tied to in BEAM: is its lifetime tied to the process where the connection is open?

magnetised: I've added a (dense) para that explains how prepared statements are tied to the process memory - I asked the same questions re the lack of release for the stmts and was satisfied by the answer.

alco: Thanks! That helps.

I'm not too familiar with NIF resource lifetime management and how that plays with Erlang's GC. Like, if a process dies and its whole heap can be released, does the GC still scan it looking for NIF resources that need to be `destroy`ed? It's not related to your changes, just mind-wandering.

alco (approved):
:shipit:

Issue #3928, opened 2026-02-27
Investigate potential head-of-line blocking in ShapeLogCollector for shapes with subqueries


alco:
Feedback from a user indicates that a shape query with a few nesting levels takes orders of magnitude longer to return than how long it takes to run the query against Postgres (5-8 seconds vs 20ms).

> The Electric instance is running at 1-3% CPU utilization on 4 vCPUs with 3.8% memory utilization. The write pipeline processes each transaction in 2.6ms average (8.2ms max). The SLC publish phase takes 1.8ms average (6.3ms max).

> The bottleneck is not resource exhaustion — it's architectural. The SLC processes all transactions through a single GenServer with :infinity timeout, creating head-of-line blocking proportional to active shape count and dependency layer depth.

PR #3935, opened 2026-03-02
Add Postgres oracle property tests

robacourt:
## Summary

Adds a property-based oracle testing framework that verifies Electric's shape streaming correctness by comparing materialized shape state against direct Postgres query results.

## Changes

- **Oracle harness** (`oracle_harness.ex`) — Coordinates parallel shape verification: starts shape checkers, applies batched transactions, verifies all shapes after each batch
- **Shape checker** (`shape_checker.ex`) — GenServer that materializes a single shape's stream via `Electric.Client.poll` and asserts consistency against Postgres
- **Standard schema** (`standard_schema.ex`) — 4-level hierarchy schema (`level_1` → `level_2` → `level_3` → `level_4` + tag tables) with seed data and StreamData mutation generators
- **WHERE clause generator** (`where_clause_generator.ex`) — StreamData generators for diverse SQL patterns: AND/OR compositions, NOT, LIKE, BETWEEN, IN, and 1-3 level nested subqueries including tag-based subqueries
- **Property test** (`oracle_property_test.exs`) — Generates N shapes with random WHERE clauses and M batches of random mutations, verifying Electric matches Postgres after each batch
- **Integration setup** — Added `num_clients` option to scale Finch connection pool for concurrent shape polling

Tests are excluded by default (`@moduletag :oracle`) and can be run with:
```
mix test --only oracle
```

Configurable via env vars: `SHAPE_COUNT`, `BATCH_COUNT`, `TXNS_PER_BATCH`, `MUTATIONS_PER_TXN`, `RUN_COUNT`, `LONG_POLL_TIMEOUT`.

## Test Plan
- [x] `mix test --include oracle` passes with default config (100 shapes, 10 batches)
- [ ] Verify existing tests unaffected (`mix test` without `--include oracle`)

---
Generated with [Claude Code](https://claude.com/claude-code)

Issue #3892, opened 2026-02-23
Stale rows in materialized view after multi-level subquery join path changes


alco:
## Summary

Electric's materialized view retains stale rows when mutations change the join path of a multi-level nested subquery filter. The Postgres oracle query returns the correct result, but Electric's view includes extra rows that no longer match.

## Failing shape

```sql
-- shape_20: 3-level nested subquery through junction table
SELECT id, level_3_id, value FROM level_4
WHERE level_3_id IN (
  SELECT id FROM level_3 WHERE level_2_id IN (
    SELECT id FROM level_2 WHERE level_1_id IN (
      SELECT level_1_id FROM level_1_tags WHERE tag = 'alpha'
    )
  )
)
```

## Observed behavior

After applying `txn_5` (which includes moves between parent levels and tag mutations), Electric's view contains 18 rows while the Postgres oracle returns 16 rows. The two extra rows are:

| id | level_3_id | value |
|----|-----------|-------|
| l4-3 | l3-3 | v684 |
| l4-9 | l3-3 | v682 |

Both have `level_3_id = l3-3`, suggesting Electric didn't correctly invalidate rows when the join path `level_3 → level_2 → level_1_tags` changed due to parent moves or tag operations in the transaction.

## Reproduction

Branch [`alco/oracle-replay-subquery-bug`](https://github.com/electric-sql/electric/tree/alco/oracle-replay-subquery-bug) contains a deterministic replay test extracted from a generative oracle test run (20 shapes, 40 transactions × 40 mutations).

```bash
mix test --include oracle_replay test/integration/oracle_property_test.exs
```

The failure is fully deterministic — it reproduces on every run at the same step (`txn_5`) with the same mismatch.

## Context

Found via the generative oracle property test (`mix test --include oracle`). The scenario involves a 4-level hierarchy (`level_1` → `level_2` → `level_3` → `level_4`) with junction tag tables at each level. The mutations that trigger the bug include moves (`UPDATE level_N SET level_{N-1}_id = ...`) and tag operations (`INSERT/DELETE` on `level_N_tags`) that alter which rows match the nested subquery filter.

robacourt:
I've simplified the test in [alco/oracle-replay-subquery-bug](https://github.com/electric-sql/electric/tree/alco/oracle-replay-subquery-bug)

To reproduce run
```bash
mix test --include oracle  test/integration/oracle_view_test.exs:264
```

PR #3937, opened 2026-03-02
Fix head-of-line blocking in SLC for subquery shapes via ETS link-values cache and inverted index

KyleAMathews:
Shapes with subqueries (\`IN (SELECT ...)\`) were causing 6–14s SLC stalls under load. This fixes three independent bottlenecks: O(N×D) serialised GenServer calls in the SLC filter step, O(N) other_shapes scan for dep-shape record changes, and O(D) ETS probe per shape in the hot-path membership check.

## Root Cause

Three separate hot paths were serialising work:

**Bottleneck 1 — SLC filter step (O(N×D×C) GenServer calls)**

\`get_link_values\` made a \`GenServer.call\` per invocation. In \`other_shapes_affected\`, this is called for every outer shape × every dep × every changed record. With N=300, D=3, C=20 that's ~18,000 sequential GenServer calls from a single process per transaction.

**Bottleneck 2 — Dep-shape record changes (O(N) other_shapes scan)**

When a dep-shape's own table changes (e.g. a row is inserted into the subquery's source table), \`Filter.affected_shapes\` had to scan all N outer shapes to find which ones were affected, evaluating each WHERE clause against the Materializer's current link values.

**Bottleneck 3 — Hot-path membership check (O(D) ETS probes per shape)**

\`registered_in_inverted_index?\` previously walked each shape's \`shape_dependencies_handles\` list and did an ETS lookup per dep handle to check membership — O(D) probes, called for every shape in \`other_shapes_affected\` on every change.

## Approach

**Fix 1: Per-stack ETS link-values cache**

Cache \`value_counts\` in a named ETS table (\`read_concurrency: true\`) owned by \`ConsumerRegistry\`. \`get_link_values\` does an ETS lookup (~5µs) instead of a GenServer call (~100µs + queue wait). Written after each committed transaction and after initial materialization.

**Fix 2: Inverted index in Filter**

Two new private ETS tables in \`Filter\` — \`sublink_field_table\` and \`sublink_dep_table\` — form an inverted index for dep shapes that land in non-optimisable WHERE clauses. When a dep-table record changes, the index directly returns the outer shapes whose link field matches, avoiding the O(N) scan entirely.

**Fix 3: O(1) membership check via in-struct MapSet**

Added \`sublink_shapes_set\` (a \`MapSet\` on the \`Filter\` struct) tracking which shape IDs are registered in the inverted index. \`registered_in_inverted_index?\` is now \`MapSet.member?\` — O(1) and zero ETS traffic — moved to run *before* the \`get_shape\` ETS lookup in \`other_shapes_affected\` so indexed shapes never touch ETS at all.

A guard (\`map_size(sublink_fields) > 0\`) ensures composite-PK subqueries like \`(user_id, team_id) IN (SELECT ...)\` — which produce no indexable fields — are never added to \`sublink_shapes_set\` and continue to be evaluated by \`other_shapes_affected\` normally. Without this guard they would be silently dropped.

Two new telemetry spans expose the remaining work: \`filter.sublink_candidates_count\` and \`filter.sublink_reeval.duration_µs\`.

**Key invariants:**
- ETS is written *after* each committed transaction — readers always see a consistent snapshot
- The inverted index is only built for dep shapes in top-level \`other_shapes\` (non-optimisable WHERE). Dep shapes that go through an equality index are already efficiently handled and are not registered here (preventing double-evaluation)
- \`sublink_affected_shapes\` re-evaluates the full WHERE clause for candidates to correctly handle compound conditions with non-sublink predicates
- Stale ETS entries are cleaned up in \`Consumer.terminate/2\` and the \`dependency_materializer_down\` handler
- If the link-values ETS table is missing (e.g. ConsumerRegistry restart window), \`sublink_affected_shapes\` returns an empty set rather than cascading to the expensive "return all shapes" fallback
- Composite-PK dep shapes (\`RowExpr\` left-hand sides) produce no indexable sublink fields and are never added to \`sublink_shapes_set\`; they correctly fall through to \`other_shapes_affected\`

**Non-goals:** This doesn't fix O(N) consumer fan-out in \`broadcast\` — only the O(D) Materializer serialisation within each consumer. Fixing fan-out would require a different pubsub model.

**Trade-offs:** ETS is write-once-per-txn, read-many. Stale reads between write and next commit are not a concern since \`get_link_values\` is only called during the SLC filter step, which runs *after* the commit that wrote the values.

## Verification

```sh
cd packages/sync-service

# Materializer ETS cache (43 tests)
mix test test/electric/shapes/consumer/materializer_test.exs

# ShapeLogCollector (24 tests)
mix test test/electric/shape_log_collector_test.exs

# Filter inverted index (covered by existing filter tests)
mix test test/electric/shapes/filter_test.exs

# Composite PK subquery regression
mix test test/electric/plug/router_test.exs:2671 --include slow
```

The regression test \`get_link_values reads from ETS cache and does not require the GenServer to be alive\` stops the Materializer GenServer after populating ETS, then asserts \`get_link_values\` returns the correct values — fails without the fix, passes with it.

## Files Changed

- **\`consumer/materializer.ex\`** — adds \`init_link_values_table/1\`, \`write_link_values/1\`, \`delete_link_values/2\`; updates \`get_link_values/1\` to check ETS first with GenServer fallback; adds logging when cache degrades
- **\`consumer_registry.ex\`** — creates the shared ETS table in \`new/2\` so it lives alongside the registry
- **\`consumer.ex\`** — calls \`delete_link_values\` in \`terminate/2\` and \`dependency_materializer_down\` handler to prevent stale cache entries
- **\`filter.ex\`** — adds inverted index ETS tables and \`sublink_shapes_set\` MapSet; \`register_sublink_shape/unregister_sublink_shape\` maintain both; \`sublink_affected_shapes\` does the O(K) lookup; \`registered_in_inverted_index?\` is now O(1); composite-PK guard prevents RowExpr shapes from being falsely indexed
- **\`filter/where_condition.ex\`** — moves \`registered_in_inverted_index?\` check before \`get_shape\` ETS lookup so indexed shapes are skipped without touching ETS
- **\`consumer/materializer_test.exs\`** — regression test for ETS cache
- **\`test/electric/shapes/filter_test.exs\`** — \`snapshot_filter_ets\` extended to cover both new ETS tables

---
Fixes https://github.com/electric-sql/electric/issues/3928

### DIFF
--- a/packages/sync-service/lib/electric/shapes/filter.ex
+++ b/packages/sync-service/lib/electric/shapes/filter.ex
@@ -394,9 +394,8 @@ defmodule Electric.Shapes.Filter do
       :ets.insert(filter.sublink_dep_table, {dep_handle, MapSet.put(existing_shapes, shape_id)})
     end
 
-    # Only add to the fast-path set when at least one field was indexed.
-    # Shapes with composite-key subqueries (RowExpr left-hand sides) produce no
-    # sublink_fields and must fall through to other_shapes_affected instead.
+    # RowExpr subqueries (e.g. `(a, b) IN (SELECT ...)`) produce no indexable fields;
+    # those shapes stay in other_shapes and must not be marked as indexed.
     if map_size(sublink_fields) > 0 do
       %{filter | sublink_shapes_set: MapSet.put(filter.sublink_shapes_set, shape_id)}
     else
--- a/packages/sync-service/lib/electric/shapes/filter/where_condition.ex
+++ b/packages/sync-service/lib/electric/shapes/filter/where_condition.ex
@@ -206,7 +206,7 @@ defmodule Electric.Shapes.Filter.WhereCondition do
       [shape_count: map_size(other_shapes)],
       fn ->
         for {shape_id, where} <- other_shapes,
-            # Cheap MapSet check before touching ETS — skips all sublink-indexed shapes
+            # Cheap MapSet check before touching ETS — skips shapes handled by the sublink inverted index
             not Filter.registered_in_inverted_index?(filter, shape_id),
             shape = Filter.get_shape(filter, shape_id),
             not is_nil(shape),


alco:
Not bad. Needs additional verification:

- we need a benchmark to validate the problem exists as described by the agent and that the changes resolve it
- switching from a synchronous GenServer.call to async ETS lookups need to be scrutinized for correctness in the specific code paths where `get_link_values` is called
- unleashing [Postgres oracle property tests](https://github.com/electric-sql/electric/pull/3935) on the fix could be a good way to gain confidence in the fix. But note that the oracle tests fail on `main` currently when a particular edge case is hit. We're tracking [one such edge case](https://github.com/electric-sql/electric/issues/3892), there may be others

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -235,6 +235,8 @@ defmodule Electric.Replication.ShapeLogCollector do
         registry_state: registry_state
       })
 
+    :ets.new(:SLC_lookup_counter, [:named_table, :public, :set, write_concurrency: true])
```
icehaunter: This will not work in our cloud env, where this named table will have been taken by SLC of another tenant already.

icehaunter (dismissed):

packages/sync-service/lib/electric/replication/shape_log_collector.ex
alco: It's a temporary counter I added for benchmarking purposes. Sorry for the confusion.

alco:
One thing Claude has highlighted:
>  Materializer crash: If a materializer crashes, its ETS entry persists (the table is owned by the SLC process via ConsumerRegistry). Stale values in ETS after a materializer crash are acceptable because:
>   - The materializer has restart: :temporary — it won't restart
>   - A crashed materializer triggers invalidation of dependent outer shapes
>   - The shape handle for the dead materializer won't be queried again

whilst Codex thinks it's a serious concern:
> • 1. High: stale ETS cache can suppress the existing safety fallback and cause false negatives in affected-shape routing.
>      get_link_values/1 now returns ETS data without checking materializer liveness (materializer.ex:76).
>      If a materializer dies, its ETS key is not cleared in terminate/down paths (materializer.ex:250, materializer.ex:270).
>      affected_shapes/3 relies on exceptions to fall back to “return all shapes” (where_condition.ex:167); stale ETS avoids that exception path and can make routing use old refs instead of conservative fan-out.
>      The new regression test explicitly enforces this dead-process ETS-read behavior (materializer_test.exs:407).
> 
>   Verdict: replacing GenServer.call with ETS lookups is valid for performance, but not fully correct as implemented unless you also guarantee cache invalidation/liveness coupling.
>   In steady state it is fine; under materializer death/restart races it can return stale refs and miss updates.
> 
>   I also ran:
> 
>   - mix test test/electric/shapes/consumer/materializer_test.exs test/electric/replication/shape_log_collector_test.exs (67 tests, 0 failures)
> 
>   Potential fix directions:
> 
>   1. delete ETS row on materializer shutdown/shape removal, or
>   2. include generation/version in ETS value and reject stale generations, or
>   3. on ETS hit, verify process alive for that handle before trusting cache.

alco:
@claude 
Your suggestion for

> Stale ETS entries on shape deletion

to add a `terminate()` function won't work: the Materializer process doesn't trap exits, so its `terminate()` function won't be called.

Consumer process is the one acting on materializer's life cycle events (search for `:materializer_down` and `:dependency_materializer_down` in consumer.ex), so either it or the ShapeCleaner module must clean up the stale ETS entry.


---

Also, pay attention to [this comment](https://github.com/electric-sql/electric/pull/3937#issuecomment-3991919301).

KyleAMathews:
## Latest commit: Fix stale ETS cleanup and tighten unregister guard

Addresses @alco's concern about stale ETS entries:

- Added `Materializer.delete_link_values/2` to remove ETS entries.
- Cleanup is now called in two places in `consumer.ex`:
  1. `terminate/2` — when a dep shape's consumer shuts down for any reason
  2. `handle_info({:dependency_materializer_down, handle})` — when a dep materializer crashes (before `handle_materializer_down` escalates to `stop_and_clean`)
- Also tightened `remove_shape` in `filter.ex` to only call `unregister_sublink_shape` when the shape was actually registered in the inverted index (via `registered_in_inverted_index?`).

---

**Re: CI failures** — all three reported failures (`FilterTest @>` performance, `ShapeStatusTest` high-concurrency, `ConsumerTest` txn fragments) are pre-existing issues unrelated to this PR:

- **FilterTest**: The reduction budget was calibrated for OTP 27. The repo's `.tool-versions` requests OTP 28.1.1 which has ~40% higher reduction counts for the same operations. Confirmed by checking out main branch's `filter.ex` — the test still fails identically. Passes on OTP 27.
- **ShapeStatusTest**: The slow test's 60s timeout is occasionally exceeded under CI load. Passes locally in 2.9s. The Walker appears in traces because `Shape.new/2` calls the parser, which uses Walker for protocol dispatch — this is not introduced by our changes.
- **ConsumerTest**: Flaky due to `Repatch.patch(:shared)` in concurrent tests.

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -57,11 +48,45 @@ defmodule Electric.Shapes.Consumer.Materializer do
...
+    case :ets.lookup(table, shape_handle) do
+      [{^shape_handle, values}] -> values
+      _ -> genserver_get_link_values(opts)
+    end
+  rescue
+    ArgumentError -> genserver_get_link_values(opts)
+  end
+
+  defp genserver_get_link_values(opts) do
+    GenServer.call(name(opts), :get_link_values, :infinity)
```
alco: I don't understand why this call's timeout had to be changed to `:infinity`.

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -57,11 +48,45 @@ defmodule Electric.Shapes.Consumer.Materializer do
...
+  in a stack. Called by `ConsumerRegistry` during stack initialization. Idempotent —
+  safe to call when the table already exists.
+  """
+  @spec init_link_values_table(stack_id :: term()) :: :ets.table() | :undefined
+  def init_link_values_table(stack_id) do
+    :ets.new(link_values_table_name(stack_id), [
+      :named_table,
+      :public,
+      :set,
+      read_concurrency: true
```
alco: Since this is a single shared table per stack, different materializer processes will be writing to it concurrently and the table itself becomes the new bottleneck. We need to add `write_concurrency: true` here and hope that'll alleviate lock contention.

alco:
I'm rerunning my benchmarks on the latest version of the code.

balegas:
This is wild. Well done on getting this merged so quickly.

PR #3939, opened 2026-03-03
fix(sync-service): Ignore failure to rollback

magnetised:
Ignore result of attempting to rollback a write transaction. 

Errors may happen after the txn has already been committed (or aborted) so we can ignore any errors resulting from any secondary ROLLBACK.

Helps with #3943


noudadrichem:
Legends!

PR #3945, opened 2026-03-03
chore(sync-service): Delay statistics collection and optimize via pool

magnetised:
Running locally I've seen errors collecting the statistics for the number of shapes and sqlite metrics because the telemetry poller runs before the shape db has properly started.

This pr adds an arbitrary delay to the start of metrics collection to make that situation less likely. 

It also simplifies the db optimization process by always checking out a connection from the write pool even in non-exclusive mode, rather than maintaining a separate persistent connection just for the periodic optimize call.

### DIFF
--- a/.changeset/stale-masks-reply.md
+++ b/.changeset/stale-masks-reply.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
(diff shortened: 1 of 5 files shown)


balegas:
@claude Don't we have any readiness state that we could rely on to start collecting metrics? I don't like arbitrary delays

(not sure claude with work on a closed pr)

PR #3955, opened 2026-03-04
fix(sync-service): Reduce memory retention in Bandit handler processes

alco:
## Summary

- Garbage collect before blocking in long-poll receive to release accumulated garbage
- Add `ELECTRIC_TWEAKS_HANDLER_FULLSWEEP_AFTER` config to control fullsweep frequency in handler processes

## Problem

Bandit handler processes can grow their heap to 8MB across multiple requests while live heap remains small. When these processes block in receive waiting for shape changes (up to 40s), they hold onto garbage memory.

## Solution

1. Call `:erlang.garbage_collect()` before entering the receive block
2. Allow configuring `fullsweep_after` spawn option for handler processes via `ELECTRIC_TWEAKS_HANDLER_FULLSWEEP_AFTER` env var (e.g., set to 5-10 to force more frequent full GC sweeps)

## Test plan

- [ ] Manual verification in production environment
- [ ] Monitor handler process memory with different `ELECTRIC_TWEAKS_HANDLER_FULLSWEEP_AFTER` values

### DIFF
--- a/.changeset/gc-handler-processes.md
+++ b/.changeset/gc-handler-processes.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Reduce memory retention in Bandit handler processes by calling `:erlang.garbage_collect()` before entering the long-poll receive block, and add `ELECTRIC_TWEAKS_HANDLER_FULLSWEEP_AFTER` env var to control the `fullsweep_after` GC spawn option for handler processes.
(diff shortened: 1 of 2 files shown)


packages/sync-service/lib/electric/application.ex
```diff
@@ -144,6 +144,7 @@ defmodule Electric.Application do
         shape_hibernate_after: get_env(opts, :shape_hibernate_after),
         shape_enable_suspend?: get_env(opts, :shape_enable_suspend?),
         conn_max_requests: get_env(opts, :conn_max_requests),
+        handler_fullsweep_after: get_env(opts, :handler_fullsweep_after),
```
msfstef, line 147: note: for cloud keep in mind this will separately need to be set up since we run a shared http service

alco: That's a good reminder, thanks!

PR #3960, opened 2026-03-05
perf(sync-service): Reduce ShapeDb memory usage

balegas:
## Summary

Reduces ShapeDb memory consumption by eliminating a covering index that duplicated all shape BLOBs, lowering the default SQLite page cache, and deduplicating shape data in the WriteBuffer's ETS tables. On an 8-core machine, this saves ~60 MB of SQLite page cache alone.

## Root Cause

Three independent sources of excess memory:

1. **`shapes_handle_cover_idx`** — a covering index on `(handle, shape, hash, snapshot_complete)` that included the full serialized shape BLOB. This effectively doubled all shape data in the SQLite page cache.

2. **SQLite page cache at 4 MB per connection** — with `2 * schedulers_online + 1` connections (17 on 8-core), total page cache was ~68 MB for a small metadata database.

3. **WriteBuffer stored shape data twice in ETS** — once in the shapes table for lookups, and again in the operations table for flushing to SQLite.

## Approach

**Covering index removal** — Dropped `shapes_handle_cover_idx`. The existing `shapes_handle_idx` unique index handles handle-based lookups with one extra page read to the table for the shape blob. Schema version bumped 4→5 (new DB file created on upgrade; old file is orphaned per existing convention).

**Page cache reduction** — Default reduced from 4 MB to 512 KB per connection (`ELECTRIC_SHAPE_DB_CACHE_SIZE` still available for override). Total cache drops from ~68 MB to ~8.5 MB on 8-core.

**WriteBuffer deduplication** — The shapes ETS table entry was widened from `{shape, comparable}` to `{shape, comparable, hash, relations}`. The operations table now stores only `{:add, handle}` instead of duplicating all shape data. During flush, `do_batch_write` looks up shape data from the shapes table. If a shape was removed before flush (add+remove race), the add is safely skipped since the subsequent remove op handles cleanup.

## Key Invariants

- **Add-before-remove ordering**: The operations table is an `ordered_set` keyed by monotonic time, so `:add` always precedes `:remove` for the same handle. If shape data is missing during flush (removed before flush), skipping the add is correct because the remove will follow.
- **Comparable cleanup**: On flush, comparable entries are cleaned up by looking them up from the shapes table. If `remove_shape` already cleaned them up, the lookup returns `[]` and cleanup is a no-op.
- **`handles_for_relations` ordering**: Now scans the shapes ETS table (unordered `set`) instead of the operations table (ordered `ordered_set`), so results are explicitly sorted.

## Non-goals

- **Filter ETS deduplication** — The Filter module stores a separate copy of every active shape for hot-path WHERE clause evaluation. Eliminating this copy would require SQLite lookups per incoming change, which is too slow.
- **`list_shapes` streaming** — Currently deserializes all shapes at once, but it's called rarely (partition restoration only).

## Trade-offs

The WriteBuffer dedup trades a single ETS lookup during flush for eliminating per-shape data duplication. The lookup is from the same process that owns the ETS table, so it's effectively free. The only new edge case is the add-before-flush-remove race, which is handled by skipping the add (correct behavior, logs at debug level).

## Verification

```bash
cd packages/sync-service
mix start_dev  # start postgres
mix test test/electric/shape_cache/shape_status/shape_db_test.exs  # 34 tests
mix test test/electric/shape_cache/  # 175 tests
```

All tests pass, including the ordering fix for `handles_for_relations`.

## Files changed

- **`connection.ex`** — Remove `shapes_handle_cover_idx`, bump schema version 4→5, reduce default `cache_size` from 4 MB to 512 KB
- **`write_buffer.ex`** — Widen shapes table tuple to include `hash` and `relations`; slim operations table to handle-only; update all ETS match patterns; add sort to `handles_for_relations`; look up shape data during flush instead of reading from ops entries

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/sync-service/lib/electric/shapes/shape.ex
+++ b/packages/sync-service/lib/electric/shapes/shape.ex
@@ -698,6 +698,23 @@ defmodule Electric.Shapes.Shape do
   Keeps only the fields accessed by Filter operations and resets
   large unused fields (most notably `shape_dependencies` which may
   contain full recursive Shape structs) to their defaults.
+
+  ## Maintenance note
+
+  If you add new fields to `Shape` and access them in `Electric.Shapes.Filter`
+  (e.g. in `add_shape/3`, `remove_shape/2`, `affected_shapes/2`, or
+  `is_affected_by_relation_change?/2`), you **must** also add those fields here.
+  Filter stores the result of this function in ETS, so any field not listed
+  below will be `nil` at filter evaluation time.
+
+  Fields currently accessed by Filter:
+  - `root_table` — table-condition routing and sublink index keys
+  - `root_table_id` — `is_affected_by_relation_change?/2`
+  - `where` — WHERE-clause evaluation and WhereCondition indexing
+  - `shape_dependencies_handles` — sublink dep-handle lookups
+  - `root_column_count` — `is_affected_by_relation_change?/2` column count check
+  - `flags` — `is_affected_by_relation_change?/2` selects_all_columns guard
+  - `selected_columns` — `is_affected_by_relation_change?/2` column intersection
   """
   def for_filter(%__MODULE__{} = shape) do
     %__MODULE__{


balegas:
benchmark this

balegas:
@claude address latest review comments

balegas:
benchmark this

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
```diff
@@ -209,17 +211,13 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.WriteBuffer do
...
       :ets.update_counter(shapes_table, :count_add, 1)
 
-      # Queue operation for SQLite (contains full data including hash and relations)
-      true =
-        :ets.insert(
-          ops_table,
-          {ts, {:add, handle, shape, comparable, hash, relations}, false}
-        )
+      # Queue operation for SQLite - shape data is looked up from shapes table during flush
+      true = :ets.insert(ops_table, {ts, {:add, handle}, false})
```
magnetised, line 220: i don't honestly think it's worth the effort of all this memory reduction for the write buffer. the data is only around for seconds. if you look at the stats the number of ops in the write buffer is usually < 10, only going big for large deletes.

even locally, when I create 200,000 shapes using 8,000 concurrent clients, the write buffer rarely goes above 200 entries.

packages/sync-service/lib/electric/shapes/filter.ex
```diff
@@ -87,7 +87,8 @@ defmodule Electric.Shapes.Filter do
   def add_shape(%Filter{} = filter, shape_id, shape) do
     if has_shape?(filter, shape_id), do: raise("duplicate shape #{shape_id}")
 
-    :ets.insert(filter.shapes_table, {shape_id, shape})
+    filter_shape = Shape.for_filter(shape)
```
magnetised, line 90: I think this one change is the really worthwhile addition, slimming the shape definition down to some minimum required in the filter will save a bunch of ram I think.

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
```diff
@@ -394,9 +391,11 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.Connection do
...
     # higher durability setting
     synchronous: "OFF",
-    # Default to a beefy page cache size because we want decent read performance
-    # Multiplied by 1024, because SQLite works in KiB but we configure in bytes.
-    cache_size: 4096 * 1024
+    # Default page cache size per connection. Multiplied by 1024, because
+    # SQLite works in KiB but we configure in bytes. With 2*schedulers+1
+    # connections, total cache = this * connection_count. 2 MB balances
+    # memory savings with headroom for larger deployments.
+    cache_size: 2 * 1024 * 1024
```
magnetised, line 398: this saves 10s of megabytes... i bumped it to handle the slow ebs backing volume thinking that more cache == better. plus this is an upper bound, sqlite is good at managing its caches and having a cache size of 2MiB doesn't mean that every connection will use 2MiB.

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
```diff
@@ -33,9 +33,6 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.Connection do
     CREATE INDEX shapes_comparable_cover_idx ON shapes (comparable, handle)
     """,
     """
-    CREATE INDEX shapes_handle_cover_idx ON shapes (handle, shape, hash, snapshot_complete)
```
magnetised, line 36: yeah, this was pointless and never used.

magnetised (changes requested):
I don't think the changes to the writebuffer are worth the churn and am nervous about reducing the page cache, but loving the `Shape.for_filter/1` function.

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/connection.ex
balegas: yeah, I think we can keep it at 4mb if we're more safe.

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
balegas: sure, happy to revert and still go with the patch to remove the unused index

PR #3964, opened 2026-03-06
fix(sync-service): Pass shape directly to consumer process

magnetised:
There may have been a time where we didn't have a loaded shape in the ShapeCache when starting a consumer and to avoid blocking the ShapeCache process we moved the shape load into the consumer. But that time has passed and the shape cache always has the Shape instance available when launching a consumer, even for resuming a previously suspended consumer. 

So instead of making the Consumer process load the shape, just include it in the launch opts and save a query to the ShapeDb.

Most fetches from the ShapeDb are quick, especially for a new shape where the info is probably still in the in-memory buffer and so the fetch is just a copy from ets, but there is a very long tail, especially under load. If the consumer process is not scheduled immediately the ShapeDb could have time to move the shape data out of the cache and flush it to disk, which means the shape load has to go via the disk which can be hit and miss (pun intended).

This should be an easy win as we're removing this unstable path for new shapes and reducing the shape fetch from twice to once  for resumed consumers.

I haven't gone through the tests to remove instances of starting a consumer directly, thus not including the shape, hence the `Map.get_lazy/3` call which falls back to the old shapedb fetch route.

### DIFF
--- a/.changeset/lovely-icons-sit.md
+++ b/.changeset/lovely-icons-sit.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Remove redundant ShapeDb fetch from Consumer initialization
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -358,6 +358,7 @@ defmodule Electric.ShapeCache do
     start_opts =
       opts
       |> Map.put(:shape_handle, shape_handle)
+      |> Map.put(:shape, shape)
       |> Map.put(:subqueries_enabled_for_stack?, "allow_subqueries" in feature_flags)
 
     case Shapes.DynamicConsumerSupervisor.start_shape_consumer(stack_id, start_opts) do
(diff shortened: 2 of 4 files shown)


alco:
Moving shape fetching into the consumer was done to reduce memory footprint. Putting the shape into the start opts makes it so the same shape now lives in the consumer process' as well as DynamicConsumerSupervisor's state.

> So instead of making the Consumer process load the shape, just include it in the launch opts and save a query to the ShapeDb.

How critical is this problem? I'm not against trying this out in prod but just note that we already routinely operate close to the limit of available memory on some cloud instances.

magnetised:
> Moving shape fetching into the consumer was done to reduce memory footprint. Putting the shape into the start opts makes it so the same shape now lives in the consumer process' as well as DynamicConsumerSupervisor's state.

Right, I remember the context now. Just trying for a quick fix to keep things going while we fix the wal buildup at a higher level.

balegas:
It's unclear the cause of the WAL buildup (I'm betting on a logical error). I propose we rollback this change to have less moving parts. We're constrained on memory and if we're spending more memory I'd prefer to move back to in-memory shape db.

PR #3967, opened 2026-03-06
fix(sync-service): Add a temporary shape mailbox

magnetised:
Add a single use mailbox for passing a shape to a consumer process without going through the disk

This ensures that shapes are passed cleanly to new consumer processes without increasing RAM by duplicating the shape in the supervisor's state.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -48,6 +48,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
   @spec initialize(stack_id()) :: :ok | {:error, term()}
   def initialize(stack_id) when is_stack_id(stack_id) do
     create_shape_meta_table(stack_id)
+    create_shape_mailbox(stack_id)
 
     {:ok, invalid_handles, valid_shape_count} = ShapeDb.validate_existing_shapes(stack_id)
 
@@ -141,6 +142,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
     OpenTelemetry.with_span("shape_status.remove_shape", [], stack_id, fn ->
       with :ok <- ShapeDb.remove_shape(stack_id, shape_handle) do
         :ets.delete(shape_meta_table(stack_id), shape_handle)
+        :ets.delete(shape_mailbox(stack_id), shape_handle)
         :ok
       end
     end)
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -333,6 +336,58 @@ defmodule Electric.ShapeCache.ShapeStatus do
     end
   end
 
+  @spec post_shape(stack_id(), shape_handle(), Shape.t()) :: :ok | {:error, term()}
+  def post_shape(stack_id, shape_handle, shape) do
+    if :ets.insert_new(shape_mailbox(stack_id), {shape_handle, shape}) do
+      :ok
+    else
+      {:error, "Duplicate shape in mailbox: #{inspect(shape_handle)}"}
```
alco, line 344: I don't think we need this here. ShapeStatus already handles the duplicate shape case and it runs before this code. Here we can make a best-effort of "insert into ETS if not there and return :ok otherwise".

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -333,6 +336,58 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
+          {:ok, shape} ->
+            shape
+
+          :error ->
+            raise ArgumentError,
+              message: "Unable to retrieve shape for handle #{inspect(shape_handle)}"
+        end
+
+      %Shape{} = shape ->
+        :ets.delete(table, shape_handle)
```
alco, line 368: Now I'm worried that the ETS tables becomes the bottleneck and we get another bad case of consumers failing to start on time in a loaded system.

magnetised:
> ETS read concurrency optimizes for read bursts and helps in situations where reads are more frequent than writes

I don't think the speed of ets will be our problem. These are consumers started one at a time by a single process, so we're talking about a single process reading from the ets at a time since the read is done in the `init/1` callback. Plus ets's "slow" is still probably ok! The slow we're worried about is multiple millisecond EBS slow.

packages/sync-service/lib/electric/shape_cache/shape_status.ex
magnetised: Should never happen, absolutely but no harm is just making sure of that is there?

PR #3975, opened 2026-03-07
fix(sync-service): recover from dead consumers blocking WAL flush advancement

balegas:
## Summary

Fixes a bug where crashed consumer processes permanently block WAL flush advancement (`confirmed_flush_lsn`), causing unbounded WAL growth on Postgres.

When a consumer crashes during `ConsumerRegistry.broadcast`:
1. Its dead PID stays in the ETS lookup table (consumers use `restart: :temporary`)
2. All subsequent events are silently dropped (non-suspend `{:DOWN, ...}` returns `[]`)
3. `FlushTracker` still records the shape as needing to flush (updated *after* publish)
4. Dead consumer never calls `notify_flushed` → `last_global_flushed_offset` permanently stuck
5. A **single** dead consumer blocks flush advancement for **all** shapes

Observed in production with recurring WAL spikes up to ~24 GB at 300K concurrent shapes.

### Fix 1 — Dead consumer recovery (`consumer_registry.ex`)
- `broadcast/1`: Non-suspend `{:DOWN}` now returns the handle for retry instead of silently dropping
- `publish/2`: Before retrying, checks `Process.alive?` on the PID in ETS, deletes dead entries, logs a warning, and lets the recursive call start a fresh consumer via `start_consumer!`

### Fix 2 — Nil pending_txn guard (`consumer.ex`)
- New `process_txn_fragment/2` clause for `has_begin?: false` + `pending_txn: nil`
- A replacement consumer started after a crash may receive a mid-transaction fragment without prior context — skip it instead of crashing on `nil.consider_flushed?`
- On commit fragments, calls `consider_flushed` so the FlushTracker isn't blocked

### Fix 3 — Delivery failure reporting (`consumer_registry.ex`, `shape_log_collector.ex`)
- `publish/2` now returns a `MapSet` of handles where delivery permanently failed (shape removed between event routing and delivery)
- SLC subtracts undeliverable handles from `affected_shapes` before updating the FlushTracker, preventing it from tracking shapes that will never flush

## Deferred: Periodic liveness sweep

A periodic sweep checking `Process.alive?` on all ETS entries would serve as a safety net for any edge cases not caught by Fix 1. This was deferred for human evaluation — a timeout-based staleness check is **unsafe** because legitimate consumers can be very slow:
- Synchronous `broadcast` blocks `{:writer_flushed, ...}` cast processing for the duration of the fan-out (seconds to minutes at 300K shapes)
- Storage flush is periodic, not immediate
- New shapes buffer fragments (`buffering?: true`) until snapshot info arrives (minutes under pool contention)
- SLC mailbox backlog (1M+ messages observed) can delay flush casts indefinitely

A `Process.alive?` check has zero false positives — a dead process will never flush.

## Test plan

- [x] New FlushTracker tests: "dead consumer blocks flush advancement" — verifies a single unflushed shape permanently blocks `last_global_flushed_offset`, and that `handle_shape_removed` unblocks it
- [x] New ConsumerRegistry tests: "dead consumer recovery" — verifies crashed consumers are detected, cleaned from ETS, and replaced with fresh consumers that receive the retried event
- [x] All existing FlushTracker tests pass (14)
- [x] All existing ConsumerRegistry tests pass (15)
- [x] All SLC tests pass (24)
- [x] All consumer tests pass (26)

### DIFF
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -554,25 +554,34 @@ defmodule Electric.Replication.ShapeLogCollector do
     OpenTelemetry.start_interval(:"shape_log_collector.publish.duration_µs")
     context = OpenTelemetry.get_current_context()
 
-    for layer <- DependencyLayers.get_for_handles(state.dependency_layers, affected_shapes) do
-      # Each publish is synchronous, so layers will be processed in order
-      layer_events =
-        Map.new(layer, fn handle ->
-          {handle, {:handle_event, Map.fetch!(events_by_handle, handle), context}}
-        end)
+    undeliverable =
+      for layer <- DependencyLayers.get_for_handles(state.dependency_layers, affected_shapes),
+          reduce: MapSet.new() do
+        acc ->
+          # Each publish is synchronous, so layers will be processed in order
+          layer_events =
+            Map.new(layer, fn handle ->
+              {handle, {:handle_event, Map.fetch!(events_by_handle, handle), context}}
+            end)
 
-      ConsumerRegistry.publish(layer_events, state.registry_state)
-    end
+          failed = ConsumerRegistry.publish(layer_events, state.registry_state)
+          MapSet.union(acc, failed)
+      end
 
     OpenTelemetry.start_interval(:"shape_log_collector.set_last_processed_lsn.duration_µs")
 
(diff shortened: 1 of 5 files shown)


balegas:
I've put an agent working on hypothesis. not sure if this is useful or not

alco:
Superseded by https://github.com/electric-sql/electric/pull/4011.

PR #3979, opened 2026-03-09
Fix stale PID race in ConsumerRegistry.unregister_name/1

### LINKED ISSUE #3864: Stale ConsumerRegistry and ShapeStatus entries after consumer exits with unhandled reasons
## Summary

When a consumer process dies with a "clean" exit reason that is not `{:shutdown, :cleanup}` or `{:shutdown, :suspend}`, the `ShapeCleaner.handle_writer_termination/3` catch-all clause does nothing, leaving stale entries in both `ConsumerRegistry` (ETS) and `ShapeStatus`. Since consumers use `restart: :temporary`, no automatic restart occurs, and the stale state causes downstream failures.

## Root cause

ConsumerRegistry does not implement process monitoring — the standard mechanism for custom OTP registries to detect process death and clean up (see the reference implementation in OTP's `dummy_via.erl`). Instead, the system relies entirely on `terminate/2` → `handle_writer_termination/3` for cleanup. However, `terminate/2` is not guaranteed to run (e.g. when a process is killed with an untrappable `kill` signal), and `handle_writer_termination` has a gap in its clauses.

Note: `ConsumerRegistry.unregister_name/1` is a no-op, but this is irrelevant — OTP only calls `unregister_name` during `init/1` failures, not when a process dies. The expected cleanup mechanism for custom registries is process monitoring via `DOWN` messages, which ConsumerRegistry does not implement.

`handle_writer_termination` has four clauses:

1. `{:shutdown, :cleanup}` — full cleanup via `remove_shape_async` (ConsumerRegistry, ShapeStatus, storage, ShapeLogCollector)
2. `{:shutdown, :suspend}` — removes ConsumerRegistry entry only, leaves ShapeStatus intact (intentional for lazy restart)
3. `:normal | :killed | :shutdown | {:shutdown, _}` — **does nothing** (returns `:ok`)
4. Abnormal reasons — full cleanup via `remove_shape_async`

Clause 3 is the problem. Any consumer exit matching that guard leaves:
- A stale `{shape_handle, dead_pid}` entry in ConsumerRegistry ETS
- A stale ShapeStatus entry with `snapshot_started = false`

Additionally, since `terminate/2` is not guaranteed to be called (e.g. when a process receives an untrappable `kill` signal), even clause 4 may not execute, meaning abnormal exits can also leave stale state.

## Affected code paths

1. **`ShapeCache.restore_shape_and_dependencies/4`** — trusts stale PIDs: `whereis/2` returns a dead PID from ETS without checking `Process.alive?`, causing the function to skip starting a new consumer and return the dead PID to callers.

## Previously reported paths (resolved or inaccurate)

- ~~`ShapeCache.await_snapshot_start/2` — infinite recursive loop~~ — Fixed in 7c2d1fe13b (now has a retry limit of 10 attempts with proper error handling).
- ~~`ConsumerRegistry.publish/2` — silent message loss~~ — Inaccurate: `broadcast()` monitors every consumer PID and handles `:DOWN` messages, so messages are not silently sent to dead PIDs. The crash is detected; event handling for crashed consumers is a separate concern being addressed in PR #3975.

## Context

Before `ConsumerRegistry`, there was a `Shapes.Monitor` module that used `Process.monitor` to detect consumer deaths and clean up. It was removed for performance reasons. The current approach relies on `terminate/2` callbacks, but this creates a gap where stale state can accumulate.

**Related issues/PRs:** #3844, PR #3979, PR #3975.

alco:
## Summary
- Replace the no-op `unregister_name/1` with an atomic `:ets.match_delete/2` that removes the ETS entry only if the pid matches the calling (dying) process
- A naive lookup-then-delete approach is still racey because the entry can be overwritten between the two ETS calls; `match_delete` is atomic
- Add unit tests covering: basic cleanup, skip-if-different-pid, and the full race scenario (old process unregisters after replacement has registered)

Closes #3864

## Test plan
- [x] `mix test test/electric/shapes/consumer_registry_test.exs` — 16 tests pass (3 new)
- [ ] Full test suite passes in CI

### DIFF
--- a/.changeset/fix-stale-consumer-registry.md
+++ b/.changeset/fix-stale-consumer-registry.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Fix race condition in `ConsumerRegistry.unregister_name/1` that left stale PIDs in the ETS table. Uses atomic `:ets.match_delete/2` to remove the entry only if it still belongs to the dying process, preventing accidental deletion of a replacement consumer's entry.


magnetised:
@alco the reason for the no-op unregister call was to handle race conditions between the consumer registry and the shape log collector. If you just remove the registry entry when the pid terminates but before you remove the shape from the SLC, then the SLC could receive an operation for that shape, see that the consumer process isn't there and start a new process for it. Then the SLC gets the "remove shape" for that new process and you're left with an orphan process.

PR #3994, opened 2026-03-11
fix: ignore unknown query params in parse_postgresql_uri

### LINKED ISSUE #3993: parse_url_query/1 silently ignores unknown query params when sslmode   is present but rejects them otherwise
Electric.Config.parse_url_query/1 uses Elixir map pattern matching to validate query parameters in Postgres connection URLs. The %{"sslmode" => sslmode}  pattern matches any map containing the sslmode key, regardless of other keys present. This means:
  - ?sslmode=prefer&uselibpqcompat=true → accepted (extra param silently ignored)
  - ?uselibpqcompat=true → rejected with unsupported query options: "uselibpqcompat"

  This inconsistency means the behavior depends on whether sslmode happens to be present alongside the unknown parameter, which is confusing and leads to hard-to-debug failures.

Suggested fix options (pick one):
  1. Be strict: Reject unknown params even when sslmode is present. Extract sslmode from the map, check remaining keys are empty.
  2. Be lenient: Strip/ignore unknown params before matching. Only validate the ones Electric cares about (sslmode, replication).

  Option 2 is more resilient to clients that add driver-specific params (e.g. uselibpqcompat=true from node-postgres pg-connection-string).

  Impact: In Electric Cloud (stratovolt), connection URLs stored with uselibpqcompat=true fail to parse when sslmode is absent (common for pooled connection
  URLs). This causes tenant provisioning failures — currently 5 affected tenants in production (ELECTRIC-7FH, and related).

balegas:
## Summary

`parse_postgresql_uri` now ignores unknown query parameters (like `uselibpqcompat=true`) instead of rejecting them. Previously, behavior was inconsistent: unknown params were silently ignored when `sslmode` was present but rejected when absent. This caused production failures in Electric Cloud for connection URLs containing driver-specific parameters.

## Root Cause

`parse_url_query/1` used a single `case` statement with Elixir map pattern matching:

```elixir
case URI.decode_query(query_str) do
  %{"sslmode" => sslmode} when sslmode in ~w[...] ->   # matches any map WITH sslmode
    {:ok, sslmode: ...}
  # ...
  map ->                                                # catch-all for everything else
    {:error, "unsupported query options: ..."}
end
```

`%{"sslmode" => sslmode}` is a **partial match** — it matches any map containing `sslmode`, regardless of other keys. So `?uselibpqcompat=true&sslmode=prefer` succeeded (extra key ignored), but `?uselibpqcompat=true` alone hit the catch-all and was rejected.

This inconsistency meant the same parameter was accepted or rejected depending on whether an unrelated parameter (`sslmode`) happened to be present.

## Approach

Replaced the monolithic `case` with two focused functions:

1. **`validate_no_replication_param/1`** — explicitly rejects the `replication` param (Electric manages replication connections itself, so this must remain an error)
2. **`parse_sslmode/1`** — extracts and validates `sslmode` if present, ignores everything else

This makes the intent explicit: Electric validates params it cares about and ignores the rest.

## Key Invariants

- `sslmode` validation is unchanged — same accepted values, same error messages
- `replication` param is still rejected — this is intentional, not an unknown param
- Unknown params (e.g., `uselibpqcompat`, `options`, `connect_timeout`) are silently ignored
- No change to the parsed output structure (`{:ok, keyword()}`)

## Non-goals

- **Strict validation**: We explicitly chose NOT to reject unknown params. Connection URLs come from various sources (dashboards, CLIs, provider APIs) that may add driver-specific params. Being lenient here is more robust.

## Verification

```sh
cd packages/sync-service
mix test test/electric/config_test.exs    # 37 doctests + 9 tests
mix test                                  # full suite (1835 tests)
```

## Files changed

| File | Change |
|------|--------|
| `packages/sync-service/lib/electric/config.ex` | Replace `parse_url_query` case statement with `validate_no_replication_param` + `parse_sslmode` functions; update doctests |
| `.changeset/fix-parse-url-query-unknown-params.md` | Patch changeset |

Fixes #3993

### DIFF
--- a/.changeset/fix-parse-url-query-unknown-params.md
+++ b/.changeset/fix-parse-url-query-unknown-params.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Fix `parse_postgresql_uri` to ignore unknown query parameters instead of rejecting them. Previously, unknown params like `uselibpqcompat=true` were silently ignored when `sslmode` was present (due to Elixir map pattern matching) but rejected when alone. Now unknown params are always ignored, and only `sslmode` and `replication` are validated.
--- a/packages/sync-service/lib/electric/config.ex
+++ b/packages/sync-service/lib/electric/config.ex
@@ -303,7 +303,31 @@ defmodule Electric.Config do
       {:error, "missing host"}
 
       iex> parse_postgresql_uri("postgresql://user@localhost:5433/mydb?opts=-c%20synchronous_commit%3Doff&foo=bar")
-      {:error, "unsupported query options: \"foo\", \"opts\""}
+      {:ok, [
+        hostname: "localhost",
+        port: 5433,
+        database: "mydb",
+        username: "user"
+      ]}
+
+      iex> parse_postgresql_uri("postgres://user:pass@localhost:5432/db?uselibpqcompat=true") |> deobfuscate()
+      {:ok, [
+        hostname: "localhost",
+        port: 5432,
+        database: "db",
+        username: "user",
+        password: "pass"
+      ]}
+
+      iex> parse_postgresql_uri("postgres://user:pass@localhost:5432/db?uselibpqcompat=true&sslmode=require") |> deobfuscate()
+      {:ok, [
+        hostname: "localhost",
+        port: 5432,
+        database: "db",
+        username: "user",
+        password: "pass",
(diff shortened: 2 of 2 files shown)


balegas:
I'm not sure how i feel about relaxing the verifications given that we already reject some parameters (`replication`)

PR #3996, opened 2026-03-11
fix(sync-service): Update pg_query_ex

magnetised:
v0.10.0 has updated libpg_query to the latest version based on Postgres 17, limits query size to 64KiB to prevent segfaults and includes guards for null clauses in the unparse calls again to prevent segfaults.


alco (approved):
🏆

PR #4000, opened 2026-03-12
feat: add latest processed/seen LSN broadcasting for future use

icehaunter:


packages/sync-service/lib/electric/lsn_tracker.ex
```diff
@@ -45,16 +46,42 @@ defmodule Electric.LsnTracker do
...
+      Registry.dispatch(registry, @global_lsn_updates_topic, fn entries ->
+        for {pid, _} <- entries, do: send(pid, {:global_last_seen_lsn, lsn})
+      end)
+    end
+
+    :ok
+  end
+
+  @spec subscribe_to_global_lsn_updates(stack_ref(), term()) :: {:ok, pid()} | {:error, term()}
+  def subscribe_to_global_lsn_updates(stack_ref, value \\ []) do
```
robacourt, line 77: Surely we'll need unsubscribe too?

icehaunter: Done!

PR #4002, opened 2026-03-12
fix(sync-service): Handle errors when collecting statistics

magnetised:
Telemetry is currently started before the rest of the stack, which means that it starts trying to poll information after a few seconds, but there's no guarantee that the rest of the stack is available within the given timeframe.

Wrap all telemetry collection calls in try..catch to avoid spurious, transient, errors.

https://electricsql-04.sentry.io/issues/102382605/?alert_rule_id=130196&alert_timestamp=1773223876687&alert_type=email&notification_uuid=8fa23ea6-1e04-4a07-98bd-bd0765895424&project=4508410462404688&referrer=digest_email

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
@@ -156,6 +156,8 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.WriteBuffer do
   @doc "Returns the number of pending operations in the buffer"
   def pending_operations_count(stack_id) do
     :ets.info(operations_table_name(stack_id), :size)
+  rescue
+    ArgumentError -> 0
   end
 
   @doc "Gives the change to the total count of shapes in the database once all buffered writes are applied"
--- a/packages/sync-service/lib/electric/stack_supervisor.ex
+++ b/packages/sync-service/lib/electric/stack_supervisor.ex
@@ -380,11 +380,13 @@ defmodule Electric.StackSupervisor do
     registry_partitions =
       Keyword.get(config.tweaks, :registry_partitions, System.schedulers_online())
 
-    telemetry_child = Electric.StackSupervisor.Telemetry.configure(config)
+    telemetry_spec = Electric.StackSupervisor.Telemetry.configure(config)
 
     children =
       [
-        telemetry_child,
+        # put telemetry processes first so that they are torn down last and can
+        # continue reporting while the stack terminates.
+        telemetry_spec,
         {Electric.ProcessRegistry, partitions: registry_partitions, stack_id: stack_id},
         {Electric.StackConfig,
          stack_id: stack_id,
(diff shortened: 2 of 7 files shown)


msfstef:
I'm pretty sure we had at some point moved the telemetry at the top of the tree for some reason - can we verify that this doesn't conflict with any previous decisions made and ens up causing more problems? We should have left comments on it

packages/sync-service/lib/electric/stack_supervisor.ex
```diff
@@ -380,11 +380,10 @@ defmodule Electric.StackSupervisor do
     registry_partitions =
       Keyword.get(config.tweaks, :registry_partitions, System.schedulers_online())
 
-    telemetry_child = Electric.StackSupervisor.Telemetry.configure(config)
+    telemetry_child = Electric.StackSupervisor.Telemetry.configure(config) |> dbg
```
msfstef:
```suggestion
    telemetry_child = Electric.StackSupervisor.Telemetry.configure(config)
```

magnetised: bugger!

msfstef:
Found the relevant PR https://github.com/electric-sql/electric/pull/2722 @icehaunter - it was moved to the top because we wanted telemetry to shut down last, not to start it first.

Otherwise there's various telemetry operations happening during shutdowns that don't find the telemetry processes alive to complete I think.

msfstef:
Shouldn't we just keep it at the top but make the polling resilient to the stack not being present?

packages/sync-service/lib/electric/stack_supervisor/telemetry.ex
```diff
@@ -173,16 +187,29 @@ defmodule Electric.StackSupervisor.Telemetry do
...
-      case Electric.ShapeCache.ShapeStatus.ShapeDb.statistics(stack_id) do
-        {:ok, stats} ->
-          Electric.Telemetry.OpenTelemetry.execute(
-            [:electric, :shape_db, :sqlite],
-            stats,
-            %{stack_id: stack_id}
-          )
-
-        _ ->
+      try do
```
msfstef: why not make these ShapeDb function return tuples and handle them appropriately? I would rather we don't swallow all errors in the telemetry since some of these are exclusively called here, and any actual errors we want to catch and log would now be lost and we silently stop collecting telemetry

magnetised:
@msfstef i wrote a long disagreement with you but actually I think you're right. As you say, these functions are purely for telemetry and there should be a small range of "expected" errors from them.

packages/sync-service/lib/electric/shape_cache/shape_status/shape_db/write_buffer.ex
```diff
@@ -156,6 +156,8 @@ defmodule Electric.ShapeCache.ShapeStatus.ShapeDb.WriteBuffer do
   @doc "Returns the number of pending operations in the buffer"
   def pending_operations_count(stack_id) do
     :ets.info(operations_table_name(stack_id), :size)
+  rescue
+    ArgumentError -> 0
```
msfstef, line 160: I don't remember the exact semantics of this buffer, but I assume this is an in-memory buffer so if the ETS table is not present then the operations are indeed 0 - is this correct or is there a chance of some perissted data not having been loaded and 0 being wrong?

magnetised: this is purely in-memory temporarily before being written to the db on disk, so on boot its always empty

Issue #3983, opened 2026-03-09
FlushTracker stall: consumer alive but permanently stuck (no progress detection)


alco:
Parent: #3980

## Scenario

A consumer process is alive (`Process.alive?` returns true) but has stopped making progress on flushing transactions. The FlushTracker tracks the shape, but no `handle_flush_notification` ever arrives because the consumer is stuck.

## How this can happen

- **Deadlock or infinite wait**: The consumer is blocked waiting on a resource that will never become available (e.g., a GenServer.call to a dead process without a timeout, a storage backend that hangs on I/O).
- **Infinite loop in event processing**: A bug in change handling, move-in processing, or materializer interaction causes the consumer to loop without returning from `handle_call`.
- **Message queue starvation**: The consumer's mailbox is flooded with low-priority messages that are processed before the storage `:flushed` callback, effectively starving the flush path indefinitely.

## Why this is distinct

This scenario cannot be detected by process monitoring (Option A in the parent issue) because the consumer process is alive. `Process.alive?` returns true, and no `:DOWN` message is ever sent.

Only a **progress-based detection mechanism** can catch this — e.g., tracking the last time each shape in `FlushTracker.last_flushed` advanced and treating shapes that haven't progressed within a timeout as stuck.

## Fix

Extend the liveness sweep approach (Option B in #3980) with a staleness timeout: if a shape has been in `FlushTracker.last_flushed` for longer than N seconds without its `last_flushed` offset advancing, treat it as stuck and call `handle_shape_removed` (or trigger a consumer restart).

This is the only scenario that requires timeout-based detection rather than monitor-based detection.

Issue #3982, opened 2026-03-09
FlushTracker stall: consumer crash with pending flush, handle_writer_termination clause 3 does nothing


alco:
Parent: #3980

## Scenario

A consumer accepts and processes a transaction event (replies `:ok` to `broadcast`), then later dies with an exit reason matching `handle_writer_termination` clause 3: `:normal | :killed | :shutdown | {:shutdown, _}`. The `terminate/2` callback runs, but `handle_writer_termination` does nothing — no shape removal, no FlushTracker cleanup.

## What happens

1. Transaction at LSN 5 arrives. `ConsumerRegistry.publish` delivers to the consumer. Consumer processes the event, writes to storage (buffered), replies `:ok`.
2. `FlushTracker.handle_txn_fragment` records the shape with `{last_sent: lsn5_offset, last_flushed: prev_offset}`.
3. Before the async storage flush fires (the `:flushed` message from `flush_buffer` via `send(self(), ...)` hasn't been processed yet), the consumer dies.
4. `terminate/2` runs. `terminate_writer` calls `flush_and_close_all` → `flush_buffer`, which writes data to disk and calls `send(self(), {Storage, :flushed, offset})`. But the process is exiting — this message is never processed, so `notify_flushed` is never called.
5. `handle_writer_termination` matches clause 3 and returns `:ok` — no cleanup.
6. The shape remains in `FlushTracker.last_flushed` and `min_incomplete_flush_tree` permanently.

## Causes of death in this window

The consumer can die between accepting the event and the flush callback for many reasons:

- A subsequent event causes a crash → the consumer dies, but the *previous* event's flush is lost. (Note: the crash itself is detected by `broadcast`'s `:DOWN` handler in #3975, but that only helps with the *current* event's delivery, not the previous unflushed one.)
- A linked/monitored process dies (e.g., materializer crash → consumer exits with `{:shutdown, _}`).
- External `Process.exit(pid, :shutdown)` or `:normal` exit.

## Note

The root cause — `handle_writer_termination` clause 3 doing nothing — leaves stale entries in both ConsumerRegistry (ETS) and ShapeStatus, in addition to the FlushTracker stall described above. Fixing clause 3 to perform proper cleanup would call `remove_subscription` → `FlushTracker.handle_shape_removed`, resolving this scenario.

[...]

Issue #3981, opened 2026-03-09
FlushTracker stall: consumer killed without terminate callback


alco:
Parent: #3980

## Scenario

A consumer process is killed in a way that bypasses the `terminate/2` callback:

- **`:kill` signal**: Even with `Process.flag(:trap_exit, true)` (consumer.ex:108), a `:kill` signal terminates the process immediately without calling `terminate`. From Erlang docs: "If the process receives a kill signal, it terminates, regardless of the trap_exit flag."
- **`:brutal_kill` supervisor shutdown**: If a supervisor is configured with `shutdown: :brutal_kill`, child processes receive `:kill`.

## What happens

1. A transaction arrives. `ShapeLogCollector.publish` → `ConsumerRegistry.publish` → `broadcast` delivers the event to the consumer.
2. The consumer processes the event, replies `:ok`. `FlushTracker.handle_txn_fragment` records the shape in `last_flushed` and `min_incomplete_flush_tree`.
3. The consumer is killed (`:kill` signal, etc.) before the storage flush callback fires and `notify_flushed` is called.
4. `terminate/2` does NOT run. No cleanup happens. No `handle_writer_termination`, no `remove_shape_async`, no `FlushTracker.handle_shape_removed`.
5. The shape's entry in FlushTracker becomes the permanent minimum, blocking `last_global_flushed_offset` from advancing.

## Fix

This scenario can only be addressed by an active detection mechanism in ShapeLogCollector (see parent issue #3980). The terminate callback path is insufficient by definition — no amount of improvement to `terminate` or `handle_writer_termination` can help when the callback doesn't execute.

alco:
Closing as non-issue:
- we don't have explicit kills in the code, so this option is ruled out
- consumer supervisor shutting down will also take the shape log collector with it, so the whole shapes subsystem goes down and no replication streaming can go on

Issue #3980, opened 2026-03-09
FlushTracker has no active mechanism to detect shapes that stop reporting flush progress


alco:
## Summary

The `FlushTracker` in `ShapeLogCollector` is a passive data structure that relies entirely on external notifications to advance `last_global_flushed_offset`. When a shape enters tracking via `handle_txn_fragment`, the tracker expects one of two things to eventually happen:

1. `handle_flush_notification` — the consumer reports it flushed the offset
2. `handle_shape_removed` — the shape is cleaned up

If **neither** happens, that shape's entry becomes the minimum in `min_incomplete_flush_tree` and permanently blocks WAL flush advancement for the entire stack.

`ShapeLogCollector` does not monitor consumer processes. There is no timeout, no periodic sweep, and no liveness check for shapes tracked by `FlushTracker`. The system is entirely dependent on every consumer eventually calling `notify_flushed` or being properly removed — a property that is not guaranteed across all failure modes.

## Impact

When `last_global_flushed_offset` stops advancing, the replication client never receives `{:flush_boundary_updated, lsn}`, Postgres never gets WAL flush acknowledgement past that point, and WAL accumulates indefinitely until disk is exhausted.

## What's already addressed

- **#3975** fixes the case where a consumer dies *during* `ConsumerRegistry.broadcast` — the dead consumer is detected via `:DOWN`, cleaned from ETS, and a replacement is started. Events for shapes that were removed between routing and delivery are excluded from FlushTracker entirely.

## What remains unaddressed

The child issues below describe specific failure modes that can leave shapes permanently tracked in FlushTracker with no path to resolution.

## Suggested fix

Add a defense-in-depth mechanism in `ShapeLogCollector` to actively detect when FlushTracker-tracked shapes have stopped progressing. Two approaches:

**Option A: Monitor consumer PIDs.** When `ConsumerRegistry.publish` delivers events, capture the consumer PIDs. ShapeLogCollector monitors them. On `:DOWN`, call `FlushTracker.handle_shape_removed`. This is fast (immediate detection) but requires ShapeLogCollector to handle `:DOWN` messages and maintain a PID→shape mapping.

**Option B: Periodic liveness sweep.** On each incoming transaction (or on a timer), ShapeLogCollector iterates `FlushTracker.last_flushed` and checks whether each shape still has a live consumer via `ConsumerRegistry.whereis` + `Process.alive?`. Dead shapes get `handle_shape_removed`. This is simpler but detection is delayed until the next sweep.

[...]

PR #4008, opened 2026-03-16
Fix stuck FlushTracker when consumer processes die

alco:
## Problem

When a consumer process dies (for any reason other than `{:shutdown, :cleanup}` or `{:shutdown, :suspend}`), its entry in FlushTracker is never removed. Since FlushTracker computes the global flushed offset as the minimum across all tracked shapes, a single dead consumer permanently blocks `last_global_flushed_offset` from advancing. This prevents the replication slot's `confirmed_flush_lsn` from moving forward, causing unbounded WAL growth on the Postgres side.

Related: #3980, #3981, #3982, #3983, #3864

Supersedes / builds on the work in #3975 and #3979.

## Root causes

Three independent bugs compound to create the stuck state:

1. **`ShapeCleaner.handle_writer_termination` clause 3 is a no-op.** When a consumer exits with `:normal`, `:killed`, `:shutdown`, or `{:shutdown, _}` (other than `:cleanup`/`:suspend`), the handler returns `:ok` without any cleanup. The shape entry remains in ShapeStatus, EventRouter, FlushTracker, and ConsumerRegistry.

2. **`ConsumerRegistry.unregister_name/1` is a no-op.** When a consumer process dies, OTP calls `unregister_name` in the dying process's context, but the implementation does nothing. This leaves a stale PID in the ConsumerRegistry ETS table.

3. **`ConsumerRegistry.broadcast/1` silently drops crashed consumers.** When a consumer dies during event delivery (`:DOWN` with non-suspend reason), `broadcast` returns `[]` for that handle — silently dropping the event instead of returning it for retry like it does for suspended consumers.

### 1. `ShapeCleaner.handle_writer_termination` clause 3 → `remove_shape_async`

Consumers exiting with "clean" reasons now trigger the same async cleanup as abnormal exits. This fires the full cleanup path including `FlushTracker.handle_shape_removed`, which unblocks flush advancement.

Clauses 1 (`{:shutdown, :cleanup}`) and 2 (`{:shutdown, :suspend}`) still match before clause 3 due to Elixir's clause ordering, so suspend/cleanup behavior is unaffected.

### 2. `ConsumerRegistry.unregister_name/1` → atomic `match_delete`

Uses `:ets.match_delete(table, {shape_handle, self()})` to atomically remove the entry only if it still belongs to the dying process. This avoids the TOCTOU race where a replacement consumer registers before the original's `unregister_name` fires.

### 3. `ConsumerRegistry.broadcast/1` returns crashed handles for retry

Non-suspend `:DOWN` now returns `[{handle, event}]` (same as suspend), so `publish/2` retries delivery by starting a replacement consumer.

### 4. `ConsumerRegistry.publish/2` returns undeliverable shapes

Return type changes from `:ok` to `MapSet.t(shape_handle())`. Shapes where delivery permanently fails (e.g., shape removed between routing and delivery) are reported back. Includes bounded retry (`max_retries: 3`) — when exhausted, remaining handles are marked undeliverable rather than silently dropped.

### 5. SLC excludes undeliverable shapes from FlushTracker

`ShapeLogCollector.publish/2` collects the undeliverable set from `ConsumerRegistry.publish` and subtracts it from `affected_shapes` before passing to `FlushTracker.handle_txn_fragment`. Dead shapes are never tracked, so they can never block flush advancement.

### 6. Consumer nil-pending-txn guard (defense in depth)

When a consumer dies mid-transaction and a replacement starts, the replacement may receive a subsequent fragment (`has_begin?: false`) without pending transaction context. Previously this crashed on `nil.consider_flushed?`. Now: non-commit fragments are skipped, commit fragments call `consider_flushed` to avoid blocking FlushTracker.

## Cleanup flow after fix

```
Consumer dies (any reason)
    │
    ├─ OTP calls unregister_name/1
    │  └─ atomic match_delete removes ETS entry (only if PID matches)
    │
    └─ Consumer.terminate/2 → ShapeCleaner.handle_writer_termination
        ├─ {:shutdown, :cleanup} → remove_shape_async (unchanged)
        ├─ {:shutdown, :suspend} → remove_consumer only (unchanged)
        ├─ :normal/:shutdown/:killed/{:shutdown,_} → remove_shape_async (FIXED)
        └─ abnormal → remove_shape_async (unchanged)
                │
                └─ remove_shape_async → remove_subscription
                    └─ FlushTracker.handle_shape_removed ← unblocks flush
```

At publish time (defense in depth):
```
ConsumerRegistry.publish detects dead PID (via broadcast :DOWN)
    ├─ Retries delivery (starts replacement consumer if needed)
    └─ Reports undeliverable shapes → SLC subtracts from FlushTracker
```

## Tests

- **FlushTracker unit tests** (3 new): Prove dead consumer blocks flush, `handle_shape_removed` unblocks it, multiple dead shapes compound and unblock incrementally
- **ConsumerRegistry tests** (5 new, 7 updated): `unregister_name` cleanup + race safety, `broadcast` crash retry, `publish` dead consumer recovery + undeliverable reporting
- **Consumer test** (1 new): Mid-transaction fragment with nil pending_txn doesn't crash
- **SLC integration test** (1 new): End-to-end proof that killing a consumer doesn't permanently block FlushTracker
- **ShapeCache test adjustments** (2 modified): Stop CleanupTaskSupervisor before consumers in restart tests to prevent new cleanup from deleting shape data that needs to survive restarts

## Test plan

- [x] FlushTracker unit tests pass (20 tests)
- [x] ConsumerRegistry tests pass (18 tests)
- [x] Consumer tests pass (28 tests)
- [x] SLC tests pass (27 tests)
- [x] ShapeCache tests pass (34 tests)
- [x] Combined: 127 tests, 0 failures
- [ ] CI passes
- [ ] Manual verification against production-like workload

### DIFF
--- a/packages/sync-service/test/electric/replication/shape_log_collector_test.exs
+++ b/packages/sync-service/test/electric/replication/shape_log_collector_test.exs
@@ -1127,6 +1127,108 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
     end
   end
 
+  describe "FlushTracker advancement when consumer is killed mid-transaction" do
+    setup :setup_log_collector
+
+    setup ctx do
+      parent = self()
+
+      stub_inspector(
+        load_relation_oid: fn {"public", "test_table"}, _ ->
+          {:ok, {1234, {"public", "test_table"}}}
+        end,
+        load_relation_info: fn 1234, _ ->
+          {:ok, %{id: 1234, schema: "public", name: "test_table", parent: nil, children: nil}}
+        end,
+        load_column_info: fn 1234, _ ->
+          {:ok, [%{pk_position: 0, name: "id", is_generated: false}]}
+        end
+      )
+
+      # Two consumers for two shape handles
+      consumer1 =
+        start_link_supervised!(
+          {Support.TransactionConsumer,
+           id: :alive,
+           stack_id: ctx.stack_id,
+           parent: parent,
+           shape: @shape,
+           shape_handle: "shape-alive"},
+          id: {:consumer, :alive}
+        )
+
+      consumer2 =
+        start_link_supervised!(
+          {Support.TransactionConsumer,
+           id: :doomed,
+           stack_id: ctx.stack_id,
+           parent: parent,
+           shape: @shape,
+           shape_handle: "shape-doomed"},
+          id: {:consumer, :doomed}
+        )
+
+      %{consumer_alive: consumer1, consumer_doomed: consumer2}
+    end
+
+    test "FlushTracker advances past undeliverable shapes from killed consumer", ctx do
+      # Register to receive flush boundary updates
+      {:via, Registry, {name, key}} = Electric.Postgres.ReplicationClient.name(ctx.stack_id)
+      Registry.register(name, key, nil)
+
+      # Patch start_consumer_for_handle to return :no_shape for the doomed handle
+      # This simulates the shape being gone from ShapeCache when ConsumerRegistry
+      # tries to restart the consumer.
+      Repatch.patch(
+        Electric.ShapeCache,
+        :start_consumer_for_handle,
+        [mode: :shared],
+        fn
+          "shape-doomed", _stack_id -> {:error, :no_shape}
+          _handle, _stack_id -> {:error, :no_shape}
+        end
+      )
+
+      # Kill the doomed consumer with :kill to bypass terminate/remove_shape.
+      # This leaves the shape in EventRouter but the consumer process is dead.
+      # The ETS entry may or may not be cleaned up depending on OTP internals,
+      # but ConsumerRegistry.publish will detect the dead PID either way.
+      doomed_pid = ctx.consumer_doomed
+      Process.unlink(doomed_pid)
+      ref = Process.monitor(doomed_pid)
+      Process.exit(doomed_pid, :kill)
+      assert_receive {:DOWN, ^ref, :process, ^doomed_pid, :killed}
+
+      # Now send a transaction that affects both shapes.
+      # EventRouter routes to both "shape-alive" and "shape-doomed".
+      # ConsumerRegistry.publish detects "shape-doomed" as undeliverable (dead PID, no restart).
+      # SLC should NOT track "shape-doomed" in FlushTracker.
+      lsn = Lsn.from_integer(42)
+      log_offset = LogOffset.new(lsn, 0)
+
+      txn =
+        transaction(100, lsn, [
+          %Changes.NewRecord{
+            relation: {"public", "test_table"},
+            record: %{"id" => "1"},
+            log_offset: log_offset
+          }
+        ])
+
+      assert :ok = ShapeLogCollector.handle_event(txn, ctx.stack_id)
+
+      # The alive consumer receives the transaction
+      assert_receive {Support.TransactionConsumer, {:alive, _pid}, [_txn]}, 500
+
+      # Flush only the alive consumer — FlushTracker should advance because
+      # the doomed shape was never tracked (excluded as undeliverable).
+      ShapeLogCollector.notify_flushed(ctx.stack_id, "shape-alive", log_offset)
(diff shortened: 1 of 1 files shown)


packages/sync-service/test/electric/shapes/consumer_registry_test.exs
```diff
@@ -406,4 +406,69 @@ defmodule Electric.Shapes.ConsumerRegistryTest do
       assert_receive {:broadcast, :event}
     end
   end
+
+  describe "unregister_name/1" do
```
alco, line 410: These tests need to be rewritten to exercise a real genserver getting registered and unregistered via ConsumerRegistry.

packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -149,17 +149,19 @@ defmodule Electric.Shapes.ConsumerRegistry do
           [{handle, event}]
 
         {:DOWN, ^ref, _, _, _reason} ->
-          []
+          # Crashed consumers also need retry — their handles get returned
+          # so publish/2 can start a replacement consumer.
+          [{handle, event}]
```
alco, line 154: Without a retry limit this may lead to an infinite crash loop. Need to validate this behaviour with a reproducible workload.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -553,6 +553,27 @@ defmodule Electric.Shapes.Consumer do
     process_txn_fragment(txn_fragment, state)
   end
 
+  # Guard for replacement consumer receiving mid-transaction fragment.
+  # This is a defense-in-depth measure. The primary fix (in ShapeCleaner clause 3)
+  # triggers shape removal on consumer death, so the replacement consumer will
+  # soon be killed. But in the race window before removal fires, this guard
+  # prevents a crash.
```
alco, line 560: This looks like a misguided decision. A crashed consumer process is never restarted by a supervisor. And explicitly restarting it makes no sense if its shape handle is invalidated. A new replacement shape will be created and a new consumer will be started for that. 

There should be no path leading to a consumer restarting mid-transaction and getting that transaction's fragments.

packages/sync-service/lib/electric/shapes/consumer_registry.ex
```diff
@@ -76,20 +76,46 @@ defmodule Electric.Shapes.ConsumerRegistry do
...
   end
 
-  @spec publish(%{shape_handle() => term()}, t()) :: :ok
-  def publish(events_by_handle, _registry_state) when events_by_handle == %{} do
-    :ok
+  @spec publish(%{shape_handle() => term()}, t(), non_neg_integer()) :: MapSet.t(shape_handle())
+  def publish(events_by_handle, registry_state, max_retries \\ 3)
+
+  def publish(events_by_handle, _registry_state, _max_retries) when events_by_handle == %{} do
+    MapSet.new()
```
alco, line 83: It remains to be validated whether silently dropping undelivered events causes FlushTracker to get stuck.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -548,27 +548,34 @@ defmodule Electric.Replication.ShapeLogCollector do
...
-    for layer <- DependencyLayers.get_for_handles(state.dependency_layers, affected_shapes) do
-      # Each publish is synchronous, so layers will be processed in order
-      layer_events =
-        Map.new(layer, fn handle ->
-          {handle, {:handle_event, Map.fetch!(events_by_handle, handle), context}}
-        end)
-
-      ConsumerRegistry.publish(layer_events, state.registry_state)
-    end
+    undeliverable =
```
alco, line 551: There is a missing test that would validate the premise of this change.

alco:
Superseded by https://github.com/electric-sql/electric/pull/4011.

PR #4030, opened 2026-03-19
Add support for `COALESCE` with up to 10 args on subset queries

evanob:
We would like to be able to use `COALESCE` in tanstack/db and have that work with subset queries on Electric backend. The decision to support 10 arguments is fairly arbitrary, but we can't support any number of arguments.


icehaunter:
Hey! Thanks a lot for the contribution. I've taken this as a required push to expose already-supported variadic functions into `defpostgres` to support arbitrary-count coalescing (also greatest/least and other array functions in the future). See https://github.com/electric-sql/electric/pull/4114 - if it matches your requirements, it should supersede this PR.

evanob:
that's much better, thank you @icehaunter!

PR #4041, opened 2026-03-23
Postgres-sync pricing exercise

balegas:
# **Note:** This is a proposal that has not been discussed yet. 

## Summary
- Fill in plan card content (features, target audience) for PAYG, Pro, and Scale tiers
- Split write pricing into per-service rates: **$1/1M** for Durable Streams (baseline) vs **$3/1M** for Postgres Sync (premium, due to shape filtering compute cost)
- Add max shapes limit per tier: PAYG 10K, Pro 100K, Scale/Enterprise unlimited
- Tier discounts apply to both service rates (Pro 10%, Scale 20%)

## Changes
- `website/data/pricing.yaml` — two base write rates per service
- `website/data/pricing.data.ts` — compute effective rates for both services
- `website/data/plans/*.yaml` — plan cards with real content, shapes limits
- `website/src/components/pricing/ComparisonTable.vue` — separate write rate rows + shapes row
- `website/src/components/pricing/PricingCalculator.vue` — per-service write volume inputs


msfstef:
small note: counting postgres writes as 3 write units would also be slightly simpler for billing (to avoid introducing a separate invoice line) and maintain observability and breakdown at dashboard level

balegas:
I specifically wanted to not model 1 postgres write as a cost of 3 durable stream writes because that is a bad mental model.

PR #4051, opened 2026-03-24
Support subquery moves across AND/OR/NOT WHERE clauses

robacourt:
## Summary
This PR lets shapes with boolean combinations around subqueries stay live when dependency rows move. Previously, shapes that used `AND`, `OR`, or `NOT` around a subquery would return `409` on a move, forcing the client to throw away the shape and resync it from scratch. For large shapes that could mean a very expensive full resync; this PR avoids that by reconciling those changes in-stream.

The core of the PR is a rewrite of the subquery foundation. Move-ins are now handled as exact log splices: buffer outer-table transactions, run the move-in query, and insert its results into the shape log at the precise point where they become valid. That gives subqueries a much more reliable ordering model, removes a large class of move-handling edge cases, and gives us a stable base for richer boolean `WHERE` support over subqueries.

## Compatibility
This PR changes the wire protocol.

The sync service from this branch is not compatible with `elixir-client` from before this PR. For the TanStack DB client, this requires:
- `@tanstack/db >= 0.6.2`
- `@tanstack/electric-db-collection >= 0.3.0`

The protocol change is needed because DNF-based subquery shapes need more than tags alone. Rows can now stay in the shape for multiple independent reasons, so the server sends real `active_conditions` values and position-aware `move-in` broadcast control messages. That lets clients update the truth value of the affected DNF positions for rows they already have locally, re-evaluate inclusion correctly, and avoid unnecessary deletes or full resyncs.

## Architecture
- compile each shape's `WHERE` clause into a DNF plan and use that plan consistently for routing, move planning, SQL generation, tags, and `active_conditions`
- replace the previous subquery reverse-indexing approach with a per-shape `SubqueryIndex` that each shape consumer seeds and updates from its own dependency views, keeping the reverse index exactly in sync with consumer state
- refactor `Electric.Shapes.Consumer` around explicit event handlers and ordered effects, with a single move queue and explicit buffering/splice handling
- emit DNF-aware row metadata and position-aware move broadcasts so visibility can change incrementally instead of forcing invalidation

## Secondary changes
- update the materializer and clients to understand DNF-aware `tags`, `active_conditions`, and move broadcasts
- expand coverage across DNF planning, routing, move ordering, resume behavior, and boolean/subquery edge cases
- update the oracle test harness and generators to cover the new boolean/subquery behavior more accurately


packages/sync-service/lib/electric/shape_cache/storage.ex
```diff
@@ -88,9 +88,12 @@ defmodule Electric.ShapeCache.Storage do
   @doc """
   Write a move in snapshot to the storage. Should write it alongside the main log,
   with stiching being done via a separate call `append_move_in_snapshot_to_log!`.
+
+  The stream items are `[key, tags, json]`, where `tags` is the row tag list
+  and `json` is the encoded log item body.
   """
   @callback write_move_in_snapshot!(
-              Enumerable.t({key :: String.t(), value :: Querying.json_iodata()}),
+              Enumerable.t(row()),
```
alco, line 96: The `row()` type is defined as `list()`. Can we make its definition reflect the actual value type it represents?

packages/sync-service/lib/electric/shapes/dnf_plan.ex
```diff
@@ -0,0 +1,197 @@
...
+    :dependency_polarities
+  ]
+
+  @type tag_columns :: [String.t()] | {:hash_together, [String.t()]}
+
+  @type position_info :: %{
+          ast: term(),
+          sql: String.t(),
+          is_subquery: boolean(),
+          negated: boolean(),
```
alco, line 32: Weird naming inconsistency. Both are boolean fields, should rather be named `subquery?` and `negated?`.

packages/sync-service/lib/electric/lsn_tracker.ex
```diff
@@ -71,13 +79,35 @@ defmodule Electric.LsnTracker do
...
+  Returns the most recently broadcast LSN, or 0 if none has been broadcast yet.
+  """
+  @spec get_last_broadcast_lsn(stack_ref()) :: non_neg_integer()
+  def get_last_broadcast_lsn(stack_ref) do
+    case :ets.lookup(table(stack_ref), :last_broadcast_lsn) do
+      [{:last_broadcast_lsn, lsn}] -> lsn
+      [] -> 0
+    end
+  rescue
+    ArgumentError -> 0
```
alco, line 92: Nit: since we have this `rescue` clause, we could use `:ets.lookup_element`:

```suggestion
    :ets.lookup_element(table(stack_ref), :last_broadcast_lsn, 2)
  rescue
    ArgumentError -> 0
```

packages/sync-service/lib/electric/lsn_tracker.ex
```diff
@@ -71,13 +79,35 @@ defmodule Electric.LsnTracker do
...
+    with {:ok, _} <-
+           Registry.register(
+             Electric.StackSupervisor.registry_name(stack_ref),
+             @global_lsn_updates_topic,
+             value
+           ) do
+      last_lsn = get_last_broadcast_lsn(stack_ref)
+
+      if last_lsn > 0 do
+        send(self(), {:global_last_seen_lsn, last_lsn})
```
alco, line 106: Since this is sending to `self()`, it should be mentioned in the function doc that callers are expected to process this message after a successful function call.

packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -2279,29 +2279,49 @@ defmodule Electric.Plug.RouterTest do
...
-           "CREATE TABLE child (id INT PRIMARY KEY, parent_id INT NOT NULL REFERENCES parent(id), value INT NOT NULL)",
-           "INSERT INTO parent (id, value) VALUES (1, 1), (2, 2)",
-           "INSERT INTO child (id, parent_id, value) VALUES (1, 1, 10), (2, 2, 20)"
+           "CREATE TABLE inner_table (id INT PRIMARY KEY, value INT NOT NULL)",
+           "CREATE TABLE outer_table (id INT PRIMARY KEY, inner_id INT NOT NULL REFERENCES inner_table(id), value INT NOT NULL)"
+         ]
+    test "return 400 if same subquery is used with both positive and negative polarity", %{
+      opts: opts
+    } do
+      assert %{status: 400} =
```
alco, line 2386: Does this also return a descriptive message? I think we should have that and assert against that in tests for easy reference.

packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -2940,19 +3079,33 @@ defmodule Electric.Plug.RouterTest do
       # Hard to wait exactly what we want, so this should be OK
       Process.sleep(1000)
 
-      # We're essentially guaranteed, in this test environment, to see move-out before move-in resolves.
-      # It's safe to propagate a move-out even for stuff client hasn't seen (because of hashing in the pattern)
-      # as it's just a no-op.
-      # So we should see 2 move-outs and a move-in but only for the 3rd parent. The move-in should be filtered despite
-      # being triggered for 2 moved in parents initially
+      tag2 =
+        :crypto.hash(:md5, ctx.stack_id <> req.handle <> "v:2") |> Base.encode16(case: :lower)
```
alco, line 3181: This looks like a good candidate for a helper function, to avoid repeating the same invocation with base encoding throughout the test code.

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -132,6 +134,32 @@ defmodule Electric.Shapes.Consumer do
   end
 
   @impl GenServer
+  def handle_continue({:init_consumer, config}, state) do
```
alco, line 137: This is dead code, we no longer have `:init_consumer` continuation in this module. It got replaced with the `def handle_info({:initialize_shape, shape, opts}, state)` clause some time ago.

alco (commented):
Intermediate review while I keep looking through the rest of the changes.

What exactly is a subquery dependency polarity and what does it affect?

packages/sync-service/lib/electric/replication/eval/parser.ex
```diff
@@ -1113,7 +1113,7 @@ defmodule Electric.Replication.Eval.Parser do
          comparisons = [left_comparison, right_comparison],
          {:ok, reduced} <-
            build_bool_chain(
-             %{name: "or", impl: &pg_and/2, strict?: false},
+             %{name: "and", impl: &pg_and/2, strict?: false},
```
alco, line 1116: Is this a sneaked-in bug fix for `BETWEEN`?

packages/sync-service/lib/electric/shapes/consumer/transaction_converter.ex
```diff
@@ -0,0 +1,75 @@
...
+
+        {:ok, effects} ->
+          {:cont, {:ok, [effects | acc]}}
+
+        {:error, {:truncate, _xid}} = error ->
+          {:halt, error}
+      end
+    end)
+    |> case do
+      {:ok, effects} -> {:ok, effects |> Enum.reverse() |> List.flatten()}
```
alco, line 37: Effects could alternatively be accumulated in an order-preserving way using `Stream.concat()` and here converted to a list with `Enum.to_list()`.

packages/sync-service/lib/electric/shapes/consumer/transaction_converter.ex
```diff
@@ -0,0 +1,75 @@
...
+          converted = Shape.convert_change(shape, change, opts)
+          {:cont, [converted | acc]}
+      end
+    end)
+    |> case do
+      {:error, {:truncate, _xid}} = error ->
+        error
+
+      converted ->
+        {:ok, converted |> Enum.reverse() |> List.flatten() |> mark_last_change()}
```
alco, line 59: While this code was still in Consumer and/or ChangeHandling, the last change could be marked _before_ reversing the list of changes. We can preserve that behaviour here as well, no? Instead of doing reverse + rebuild the list just to update its last entry.

alco (approved):

robacourt:
> What exactly is a subquery dependency polarity and what does it affect?

positive: `x IN subquery`
negative: `x NOT IN subquery`, `NOT(condition AND x IN subquery)`
negatives turn move outs from the subquery to move-ins for the outer subquery
also to expand the filter during move-ins queries in flight:
- for positive we need the union of the before and after move-in subquery view (include enough for before and after)
- for negative we need the intersection of the before and after subquery view (don't exclude more until after completion)

packages/sync-service/lib/electric/replication/eval/parser.ex
robacourt: Before this PR it was a latent bug: the only consumer that ran was the evaluator, which uses implementation, so BETWEEN gave correct results. This PR introduces the code paths that key off name — DNF decomposition and re-generating WHERE SQL for subquery move-in queries — at which point the mislabel becomes a real correctness bug for any shape whose WHERE clause uses BETWEEN.

packages/sync-service/lib/electric/shape_cache/storage.ex
robacourt: Addressed in https://github.com/electric-sql/electric/pull/4646/commits/8827e20b4687bd899450c3f95ceac8018c0e53b4 (PR #4646).

packages/sync-service/lib/electric/shapes/dnf_plan.ex
robacourt: Addressed in https://github.com/electric-sql/electric/pull/4646/commits/fd7748e61bc24bd4c5ea2dca6d5b7c4a17fb1934 (PR #4646).

packages/sync-service/lib/electric/lsn_tracker.ex
robacourt: Addressed in https://github.com/electric-sql/electric/pull/4646/commits/91f8f173db01b28423f1f059584c7f38924f5357 (PR #4646).

packages/sync-service/lib/electric/lsn_tracker.ex
robacourt: Addressed in https://github.com/electric-sql/electric/pull/4646/commits/dd8cb5c295bb0c53d96bbd81af26ef77f20bb1f3 (PR #4646).

packages/sync-service/lib/electric/shapes/consumer.ex
robacourt: Addressed in https://github.com/electric-sql/electric/pull/4646/commits/82ecbc20d071476dd4f351b1cb970bcd9ab3a02c (PR #4646).

packages/sync-service/test/electric/plug/router_test.exs
robacourt: Good call — the 400 returns a descriptive message under the `where` field. Added an assertion against it in https://github.com/electric-sql/electric/pull/4646/commits/1545bb3fc8bd1fca6c5150e9cf353258f9bf1012 (PR #4646).

packages/sync-service/lib/electric/shapes/consumer/transaction_converter.ex
robacourt: Done in https://github.com/electric-sql/electric/pull/4646/commits/6aa7581a20091333d9b2eb3188836f1305de0262 (PR #4646). Conversions are now flattened into the reverse-order accumulator as they're produced, so the head is the last emitted change — it's marked before a single `Enum.reverse/1`, dropping the `List.flatten/1` pass and the `List.pop_at(-1)` + `++` rebuild.

packages/sync-service/lib/electric/shapes/consumer/transaction_converter.ex
robacourt: I left this one as-is. Each `transaction_to_effects` returns 0 or 1 `Effects.AppendChanges`, so the accumulator is a list of at most single-element lists and `Enum.reverse |> List.flatten` stays cheap and clear. A `Stream.concat` chain built up in the reduce would be nested per-txn and, to my eye, less readable here without a measurable win — happy to revisit if you feel strongly.

packages/sync-service/test/electric/plug/router_test.exs
robacourt: Done in https://github.com/electric-sql/electric/pull/4646/commits/2a7fff55963ef33d180c01d465b38233520f0b2b (PR #4646). Extracted a `value_tag/3` helper and routed all 12 call sites through it. It independently mirrors `SubqueryTags.make_value_hash_raw/3` rather than calling it, so the assertions still document the expected wire format instead of reusing the code under test.

PR #4054, opened 2026-03-25
fix(typescript-client): add exponential backoff to onError-driven retries

KyleAMathews:
## Summary

Adds exponential backoff to `onError`-driven retries in `ShapeStream` to prevent tight infinite loops when `onError` returns `{}` on persistent 4xx errors (e.g., expired auth tokens returning 403).

Previously, the fetch backoff layer correctly skipped retrying 4xx errors, but when `onError` returned `{}` to signal "retry", the stream restarted **immediately with zero delay** — creating a tight loop that could hammer both Electric and the upstream database. A user reported this causing ~$200/day in Neon network egress from a development app with zero traffic.

## Root Cause

The client has two layers of error handling:
1. **Fetch backoff** (`createFetchWithBackoff`): Retries 5xx/429 with exponential backoff. Throws 4xx immediately.
2. **`onError` callback** (`#start`): Called after fetch backoff gives up. Returns `{}` to retry, `void` to stop.

When `onError` returned an object, `#start()` recursively called itself with **no delay**. The simplest "keep syncing" pattern — `onError: () => ({})` — became the most dangerous on persistent client errors.

## Approach

- **Exponential backoff with full jitter** on the `onError` retry path: 100ms base, 30s cap, same algorithm as existing fast-loop and SSE backoffs
- **Abort-aware delay**: The `setTimeout` listens for the abort signal so `stream.abort()` / component unmount tears down immediately instead of blocking up to 30s
- **Console warning on 2nd+ retry**: Logs the delay duration and error message so developers can diagnose "why is my stream not syncing?"
- **Reset on success**: `#onErrorRetryCount` resets when the stream reaches up-to-date, so a successful auth token refresh isn't penalized on the next error

## Key Invariants

- First retry: 0–100ms delay (fast enough for auth token refresh)
- Exponential growth: 200ms, 400ms, 800ms... up to 30s cap
- Abort always honored: no hanging teardown
- Fast-loop detector stays independent (its state is still cleared on onError retry)

## Non-goals

- Changing the fetch backoff layer's 4xx handling (still throws immediately, by design)
- Capping `onError` retries (the user's `onError` controls whether to give up)
- Changing the `onError` API contract (returning `{}` still means "retry")

## Verification

```bash
cd packages/typescript-client
pnpm vitest run test/stream.test.ts test/wake-detection.test.ts test/fetch.test.ts test/expired-shapes-cache.test.ts
```

All 73 unit tests pass.

## Files changed

| File | Change |
|------|--------|
| `packages/typescript-client/src/client.ts` | Add backoff fields, backoff + abort logic in `#start()` onError path, reset in `#requestShape()` success path |
| `packages/typescript-client/test/wake-detection.test.ts` | Advance fake timers by 200ms to account for new backoff delay |
| `.changeset/add-onerror-retry-backoff.md` | Changeset for patch release |

### DIFF
--- a/.changeset/add-onerror-retry-backoff.md
+++ b/.changeset/add-onerror-retry-backoff.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/client': patch
+---
+
+Add exponential backoff to onError-driven retries to prevent tight loops on persistent 4xx errors (e.g. expired auth tokens returning 403). The backoff uses jitter with a 100ms base and 30s cap, and is abort-aware so stream teardown remains responsive.
--- a/packages/typescript-client/src/client.ts
+++ b/packages/typescript-client/src/client.ts
@@ -621,6 +621,10 @@ export class ShapeStream<T extends Row<unknown> = Row>
   #fastLoopBackoffMaxMs = 5_000
   #fastLoopConsecutiveCount = 0
   #fastLoopMaxCount = 5
+  // onError retry backoff: prevent tight loops when onError always returns {}
+  #onErrorRetryCount = 0
+  #onErrorBackoffBaseMs = 100
+  #onErrorBackoffMaxMs = 30_000
   #pendingRequestShapeCacheBuster?: string
   #maxSnapshotRetries = 5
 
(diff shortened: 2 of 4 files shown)


thruflo:
Is it worth slowing down the backoff increases? Seems quite fast as in, a transitory error for a few seconds -> results in an extra few seconds delay.

If the behavior right now is zero delay then perhaps a lower gradient backoff curve solves the problem whilst still keeping the default fairly eager to re-connect?

***

I know this can be be overridden with your own onError handler and we don't like options but I feel like this being configurable could be useful DX. Like a small set of keywords for common strategies.

KyleAMathews:
@thruflo the assumption is that generally people have configured their `onError` correctly so the first request should go through — so the backoff is just a fail safe.

PR #4069, opened 2026-03-27
Add path filtering to TS workflows to skip them for Elixir-only changes

alco:
## Summary

- Skip TS workflows on PRs that only touch Elixir packages, integration tests, or READMEs
- Add `changeset-release/main` to push triggers
- Add `push` trigger to `ts_check_formatting.yml` for `main` and `changeset-release/main`
- Pushes to `main` and `changeset-release/main` always run TS CI (no path filtering)

## Approach

Uses `paths-ignore` to exclude the few non-TS paths rather than an allowlist of every TS-related path. Since the repo is majority TS, the ignore list is shorter and doesn't need updating when new TS packages or config files are added.

Ignored paths:
- `packages/sync-service/**`, `packages/electric-telemetry/**`, `packages/elixir-client/**`
- `integration-tests/**`
- `**/README.md`
- `website/**` (in `ts_tests.yml` only — formatting covers website)

## Branch protection note

GitHub treats workflows skipped due to path filtering the same as workflows where all jobs are skipped — required status checks from these workflows will not block merge for Elixir-only PRs.

## Test plan

- [x] Verify this PR triggers TS CI (workflow files are not in the ignore list)
- [ ] Verify an Elixir-only PR does not trigger TS CI
- [ ] Verify pushes to `main` always trigger TS CI


.github/workflows/ts_tests.yml
```diff
@@ -2,16 +2,29 @@ name: TS tests
...
   pull_request:
-    paths-ignore:
-      - 'website/**'
-      - '**/README.md'
-      - 'integration-tests/**'
+    paths:
+      - 'packages/typescript-client/**'
+      - 'packages/react-hooks/**'
+      - 'packages/y-electric/**'
+      - 'packages/experimental/**'
```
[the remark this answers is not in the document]
alco: This is fine.

.github/workflows/ts_tests.yml
```diff
@@ -2,16 +2,15 @@ name: TS tests
...
-      - 'integration-tests/**'
+    branches: ['main', 'changeset-release/main']
   pull_request:
     paths-ignore:
       - 'website/**'
       - '**/README.md'
       - 'integration-tests/**'
+      - 'packages/sync-service/**'
+      - 'packages/electric-telemetry/**'
+      - 'packages/elixir-client/**'
```
msfstef, line 13: TS tests include integration tests with the sync service, so I don't think it's a bad idea to be running them with sync service changes

msfstef (approved):
Approved but I would keep testing TS client on changes to the sync service

.github/workflows/ts_tests.yml
alco: This is a contentious point. In principle, we could run all tests on every changes, but it gets impractical due to the minute-long runtimes. Once a PR is opened and it then evolves via small changes, some cosmetic, some logical, having to wait for the whole CI to finish running on every push is detrimental to productivity.

I think it's a matter of discipline: if we can as a team get into the habit of shipping narrow-focused changes fast and have integration tests run at a slower cadence (cron job, release PR, pushed to `main` only, etc.) we can increase our velocity while keeping quality regressions in check.

msfstef: The problem is that _all_ TS tests run right now, but having the TS client alone run with sync service changes is neither slow (they run in parallel) nor too much

happy to skip them but I'd be worried about client-server incompatibilities sneaking in - although they'd still show up on main anyway as you said so perhaps not. that much of an issue

alco:
Superseded by https://github.com/electric-sql/electric/pull/4221.

PR #4073, opened 2026-03-31
Revert "Add `electric-has-data` response header (#3851)"

balegas:
## Summary
- Reverts #3851 which added the `electric-has-data` response header

## Test plan
- [ ] CI passes

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -607,7 +607,6 @@ defmodule Electric.Shapes.Api do
     %{
       response
       | status: 200,
-        no_changes: true,
         body: encode_log(request, [up_to_date_ctl(global_last_seen_lsn)]),
         finalized?: true
     }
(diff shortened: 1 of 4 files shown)


msfstef (commented):
do we use the `no_changes` flag anywhere else? should we just remove it from the response struct altogether? (I'm not seeing it on the PR)

balegas:
\`no_changes\` was there before #3851 — it's used in `response.ex` to generate unique ETags for empty responses (by adding `System.monotonic_time()` to the etag), which avoids infinite CDN revalidation loops. PR #3851 only added the `electric-has-data` header logic on top of it, so the revert correctly leaves `no_changes` in place.

msfstef:
Ah interesting - then given @alco 's findings we should make sure it actually works cause it seemd to have used `no_changes: false` without any changes present. I still think that we should change this to a positive boolean `has_changes`, make the default `false` and set it to `true` to the smaller number of paths that actually return data (after reading the log)

PR #4075, opened 2026-03-31
feat(sync-service): Read-only mode for seamless rolling deploys

msfstef:
## Summary

Enables Electric instances to serve existing shape data in **read-only mode** when they don't hold the Postgres advisory lock. This unblocks smoother rolling deploys by eliminating the HTTP outage window where the new instance previously blocked all requests while waiting for the lock.

### The problem

During a rolling deploy, only one instance can hold the advisory lock and stream from the replication slot. The new instance would block on `pg_advisory_lock` — during which it served zero requests, creating a 5-20 second HTTP outage window.

### The solution

The new instance now eagerly initializes shape metadata from SQLite on startup and enters read-only mode, serving existing shapes from the shared filesystem while waiting for the lock. When the lock is acquired, it refreshes metadata and transitions to full active mode.

**State machine:** `:starting` → read-only (waiting) → `:active` ↔ read-only (waiting)

- **Read-only**: Serves existing shapes from disk. New shape creation and deletion wait for active mode (preserving previous behavior for those operations).
- **Active**: Full functionality as before.
- **Transitions are cheap**: No per-shape work, just a SQLite metadata refresh on lock acquisition.

## Key changes

- **ShapeStatusOwner** eagerly initializes shape metadata from SQLite via `handle_continue` (non-blocking to supervision tree), signaling `shape_metadata_ready` to the StatusMonitor
- **StatusMonitor** gains `service_status/1` (canonical `:active | :waiting | :starting | :sleeping` mapping) and a unified `wait_until/3` API with levels `:read_only` and `:active`
- **API** uses `hold_until_stack_ready(api, require: :read_only | :active)` — shape reads default to `:read_only`, new shape creation and deletion require `:active`
- **PureFileStorage** skips ETS caching in read-only mode (`read_only` flag on storage opts via `Storage.as_read_only/1`), reading directly from disk to see latest data from the active instance
- **ShapeStatus.refresh/1** uses a generation counter for race-free ETS updates — concurrent readers never see a missing entry during the refresh
- **EtsInspector** `load_supported_features` checks ETS before calling GenServer, and persists after storing features — so read-only instances can resolve shapes without a DB connection
- **HealthCheckPlug** uses `service_status/1` — reports "waiting" (202) during read-only mode

## What's NOT changed

- `wait_until_active` keeps its existing `:ok` contract — existing callers (replication client, tests) are unaffected
- Shape handle derivation, storage format, replication protocol unchanged
- The advisory lock mechanism is unchanged
- Client-facing API contract is preserved — clients may see brief 503s for new shape creation during rolling deploy, with `retry-after` headers

## Companion PR

- #4077 moves the admin connection pool to start before lock acquisition, making the admin pool available during the lock-waiting phase. Combined with this PR, the EtsInspector can now populate its cache for unknown schemas even in read-only mode (reducing 503s on cold starts).

## Future work

- Periodic SQLite metadata refresh during read-only mode for longer-running deploys (currently a single load on startup + refresh on lock acquisition)
- The read-only state is designed to be transitory — these improvements would make longer read-only windows more robust

## Test plan

- [x] Unit tests: 1883 tests passing (new tests for `service_status/1`, `wait_until/3`, `ShapeStatus.refresh/1`, API read-only path)
- [x] Integration test: `rolling-deploy.lux` covers full story — read-only serving, fresh data visibility via shared filesystem, handover, post-handover data, completeness check
- [x] All 9 existing integration tests pass (no regressions)
- [x] `mix format --check-formatted` clean
- [x] `mix compile --warnings-as-errors` clean

### DIFF
--- a/packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
+++ b/packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
@@ -69,7 +69,9 @@ defmodule Electric.Postgres.Inspector.EtsInspector do
   @spec load_supported_features(opts :: term()) ::
           {:ok, Map.t()} | {:error, String.t() | :connection_not_available}
   def load_supported_features(opts) do
-    GenServer.call(opts[:server], :load_supported_features, :infinity)
+    with :not_in_cache <- fetch_supported_features_from_ets(opts) do
+      GenServer.call(opts[:server], :load_supported_features, :infinity)
+    end
   end
 
   @impl Inspector
@@ -153,6 +155,7 @@ defmodule Electric.Postgres.Inspector.EtsInspector do
            {:ok, features} <-
              wrap_in_db_errors(fn -> DirectInspector.load_supported_features(state.pg_pool) end) do
         store_supported_features(state, features)
+        persist_data(state)
         {:ok, features}
       end
 
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -457,13 +457,11 @@ defmodule Electric.ShapeCache do
     {descendents ++ [{handle, shape, start_shape_opts} | siblings], known}
   end
 
-  @spec fetch_latest_offset(stack_id(), shape_handle()) ::
-          {:ok, LogOffset.t()} | :error
   @spec fetch_latest_offset(stack_id(), shape_handle(), keyword()) ::
           {:ok, LogOffset.t()} | :error
   defp fetch_latest_offset(stack_id, shape_handle, opts \\ []) do
-    storage = Storage.for_shape(shape_handle, Storage.for_stack(stack_id))
-    storage = if opts[:read_only], do: Storage.as_read_only(storage), else: storage
+    storage =
+      Storage.for_shape(shape_handle, Storage.for_stack(stack_id, read_only?: opts[:read_only?]))
 
     case Storage.fetch_latest_offset(storage) do
       {:ok, offset} -> {:ok, normalize_latest_offset(offset)}
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -77,8 +77,9 @@ defmodule Electric.ShapeCache do
     end
   end
 
-  @spec resolve_shape_handle(shape_handle(), shape_def(), stack_id()) :: handle_position() | nil
-  def resolve_shape_handle(shape_handle, shape, stack_id) do
+  @spec resolve_shape_handle(shape_handle(), shape_def(), stack_id(), keyword()) ::
+          handle_position() | nil
+  def resolve_shape_handle(shape_handle, shape, stack_id, opts \\ []) do
     # Ensure that the given shape handle matches the shape using a cheap shape
     # hash check.
     # If not (or the handle has gone/changed) then try a more expensive
@@ -90,7 +91,7 @@ defmodule Electric.ShapeCache do
         else: fetch_handle_by_shape(shape, stack_id)
 
     with {:ok, resolved_handle} <- result,
-         {:ok, offset} <- fetch_latest_offset(stack_id, resolved_handle) do
+         {:ok, offset} <- fetch_latest_offset(stack_id, resolved_handle, opts) do
       {resolved_handle, offset}
     else
       _ -> nil
@@ -216,8 +217,17 @@ defmodule Electric.ShapeCache do
   @spec has_shape?(shape_handle(), Access.t()) :: boolean()
   def has_shape?(shape_handle, stack_id)
       when is_shape_handle(shape_handle) and is_stack_id(stack_id) do
-    ShapeStatus.has_shape_handle?(stack_id, shape_handle) ||
-      GenServer.call(name(stack_id), {:has_shape_handle?, shape_handle}, @call_timeout)
+    if ShapeStatus.has_shape_handle?(stack_id, shape_handle) do
+      true
+    else
+      try do
+        GenServer.call(name(stack_id), {:has_shape_handle?, shape_handle}, @call_timeout)
+      catch
+        :exit, {:noproc, _} ->
+          Logger.debug("ShapeCache GenServer not running, shape #{shape_handle} not found in ETS")
+          false
+      end
+    end
   end
 
   @spec start_consumer_for_handle(shape_handle(), stack_id()) ::
@@ -447,12 +457,15 @@ defmodule Electric.ShapeCache do
     {descendents ++ [{handle, shape, start_shape_opts} | siblings], known}
   end
 
-  @spec fetch_latest_offset(stack_id(), shape_handle()) :: {:ok, LogOffset.t()} | :error
-  defp fetch_latest_offset(stack_id, shape_handle) do
-    shape_handle
-    |> Storage.for_shape(Storage.for_stack(stack_id))
-    |> Storage.fetch_latest_offset()
-    |> case do
+  @spec fetch_latest_offset(stack_id(), shape_handle()) ::
+          {:ok, LogOffset.t()} | :error
+  @spec fetch_latest_offset(stack_id(), shape_handle(), keyword()) ::
+          {:ok, LogOffset.t()} | :error
+  defp fetch_latest_offset(stack_id, shape_handle, opts \\ []) do
+    storage = Storage.for_shape(shape_handle, Storage.for_stack(stack_id))
+    storage = if opts[:read_only], do: Storage.as_read_only(storage), else: storage
+
+    case Storage.fetch_latest_offset(storage) do
       {:ok, offset} -> {:ok, normalize_latest_offset(offset)}
       {:error, _reason} -> :error
     end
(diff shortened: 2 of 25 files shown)


packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
```diff
@@ -153,6 +155,7 @@ defmodule Electric.Postgres.Inspector.EtsInspector do
            {:ok, features} <-
              wrap_in_db_errors(fn -> DirectInspector.load_supported_features(state.pg_pool) end) do
         store_supported_features(state, features)
+        persist_data(state)
```
magnetised, line 158: it would be great to include a unit test for this - getting supported features without a db connection, which I assume is the reason for this extra persistence step since the work on getting an early admin db connection is pending

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -332,21 +366,28 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
+      _ref ->
+        :ets.delete_all_objects(table)
+        table
+    end
 
     table
   end
 
-  defp populate_shape_meta_table(stack_id) do
+  defp populate_shape_meta_table(stack_id, generation \\ 0) do
```
magnetised: nit: remove the default arg here and pass generation 0 in the `initialize/1` callback

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -73,17 +72,52 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
+             stack_id,
+             invalid_handles
+           ) do
+      # Use a generation counter to avoid clearing the table (which would race
+      # with concurrent readers). Upsert all current shapes with a new generation,
+      # then delete any entries still on the old generation.
+      generation = System.unique_integer([:positive, :monotonic])
+      populate_shape_meta_table(stack_id, generation)
+
+      :ets.select_delete(shape_meta_table(stack_id), [
```
magnetised, line 100: clever

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -73,17 +72,52 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
     OpenTelemetry.with_child_span("shape_status.add_shape", [], stack_id, fn ->
       {_, shape_handle} = Shape.generate_id(shape)
 
       # Add the lookup last as it is the one that enables clients to find the shape
       with {:ok, shape_hash} <- ShapeDb.add_shape(stack_id, shape, shape_handle) do
-        # Cache shape metadata: {handle, hash, snapshot_started, last_read_time}
         if :ets.insert_new(
              shape_meta_table(stack_id),
-             {shape_handle, shape_hash, false, nil}
+             # Generation 0 is safe here: add_shape only runs in active mode,
```
magnetised, line 117: thanks for the comment, also phew. makes it much simpler

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -332,21 +366,28 @@ defmodule Electric.ShapeCache.ShapeStatus do
   defp shape_meta_table(stack_id),
     do: :"shape_meta_table:#{stack_id}"
 
-  defp create_shape_meta_table(stack_id) do
+  defp create_or_clear_shape_meta_table(stack_id) do
```
magnetised: this is only called once on initialize so why the "create or clear" behaviour, as in when would the "clear" path happen?

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -447,12 +457,15 @@ defmodule Electric.ShapeCache do
     {descendents ++ [{handle, shape, start_shape_opts} | siblings], known}
   end
 
-  @spec fetch_latest_offset(stack_id(), shape_handle()) :: {:ok, LogOffset.t()} | :error
-  defp fetch_latest_offset(stack_id, shape_handle) do
-    shape_handle
-    |> Storage.for_shape(Storage.for_stack(stack_id))
-    |> Storage.fetch_latest_offset()
-    |> case do
+  @spec fetch_latest_offset(stack_id(), shape_handle()) ::
```
magnetised: nit: do you need this spec? v noisy. I think defaults are handled well by typespecs

packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -447,12 +457,15 @@ defmodule Electric.ShapeCache do
...
+          {:ok, LogOffset.t()} | :error
+  @spec fetch_latest_offset(stack_id(), shape_handle(), keyword()) ::
+          {:ok, LogOffset.t()} | :error
+  defp fetch_latest_offset(stack_id, shape_handle, opts \\ []) do
+    storage = Storage.for_shape(shape_handle, Storage.for_stack(stack_id))
+    storage = if opts[:read_only], do: Storage.as_read_only(storage), else: storage
```
magnetised: could we merge this read only mode check into Storage.for_stack - it's used here and in Shapes.shape_storage

packages/sync-service/lib/electric/shapes/api/request.ex
```diff
@@ -9,6 +9,7 @@ defmodule Electric.Shapes.Api.Request do
     :global_last_seen_lsn,
     :new_changes_ref,
     :new_changes_pid,
+    read_only: false,
```
magnetised:
```suggestion
    read_only?: false,
```

packages/sync-service/lib/electric/shape_cache/storage.ex
```diff
@@ -264,6 +264,10 @@ defmodule Electric.ShapeCache.Storage do
     {mod, mod.for_shape(shape_handle, opts)}
   end
 
+  @doc "Mark a storage reference for read-only disk access (no ETS caching)"
+  def as_read_only({mod, storage_opts}), do: {mod, Map.put(storage_opts, :read_only, true)}
```
magnetised:
```suggestion
  def as_read_only({mod, storage_opts}), do: {mod, Map.put(storage_opts, :read_only?, true)}
```

packages/sync-service/lib/electric/shape_cache/shape_status_owner.ex
```diff
@@ -6,6 +6,9 @@ defmodule Electric.ShapeCache.ShapeStatusOwner do
   `Electric.ShapeCache.ShapeStatus` early in the supervision tree so that
   dependent processes (e.g., shape consumers) can use a single, shared
   ShapeStatus instance regardless of their own supervisor start order.
+
+  Initialization is done asynchronously via `handle_continue` to avoid
+  blocking the supervision tree startup on potentially slow SQLite reads.
```
magnetised, line 11: with the addition of the `shape_metadata_ready` status update, right? otherwise madness ensues

magnetised (approved):
looks amazing. I've only a few minor comments

packages/sync-service/lib/electric/shape_cache.ex
msfstef: You mean as an option in the `for_stack` function? I added a `as_read_only` utility but perhaps passing the read only opt would be nicer

packages/sync-service/lib/electric/shape_cache/shape_status_owner.ex
msfstef: Normally we would block on initializing it in the connection manager when replication is established - now that still blocks with a `refresh` call.

What we've added now is that the shape status table can be used for reads, but that is gated on `shape_metadata_ready` as you mention.

packages/sync-service/lib/electric/postgres/inspector/ets_inspector.ex
msfstef: Indeed this is the reason that's there. that being said, it's already getting persisted by `persist_data` since that persist all the table info, it just wasn't getting called here, so it was inconsistent - it was getting persisted but only after a schema query

I need to think if I should explicitly avoid persisting this since the db might change, but I'm pretty sure that if the DB changes, the DB timeline changes, and we get rid of the cache - but there's definitely someting that wasn't thought through there. I will address this separately.

packages/sync-service/lib/electric/shape_cache/shape_status.ex
msfstef: I didn't have the refresh method initially so this is an artifact

packages/sync-service/lib/electric/shape_cache.ex
msfstef: I added opts to the `for_stack` call so that simplifies things a bit

msfstef:
Thanks for the thorough review! We're not going to address these suggestions as they're all non-blocking and the tradeoffs don't justify the changes:

1. **`fetch_latest_offset` private `@spec`**: magnetised's original comment was on an older version that had two `@spec` clauses to cover a default argument — that's been consolidated into a single spec now, so the noise concern no longer applies.

2. **`check_for_disk_updates` hard-codes `read_only?: true`**: This is intentional. In read-only mode we go straight to disk (bypassing ETS entirely) for freshness. In active mode, the out-of-bounds timeout path is rare and by the time it fires, any `:new_changes` message would have already been received — so the distinction is inconsequential.

3. **Duplicated pipeline**: The receive block structure makes extraction awkward, and having the steps inline keeps each clause self-contained and readable.

PR #4087, opened 2026-04-03
fix(client): self-healing for permanently stuck expired shape handles

KyleAMathews:
## Summary

Expired shape handle entries in localStorage can get permanently stuck, preventing data from ever loading for affected shapes. This adds a self-healing retry mechanism that clears the poisoned entry and retries once, allowing automatic recovery even when a proxy strips cache-buster query parameters.

Based on #4085 by @evan-liveflow — refined with additional hardening from code review.

## Root Cause

When a shape gets a 409 (handle rotation), the client stores the old handle in `localStorage['electric_expired_shapes']`. On future requests, if a response contains that handle, the client treats it as a stale cached response and retries up to 3 times with cache-buster params.

The problem: if a proxy (e.g., phoenix_sync) strips query parameters, the cache busters are ineffective. All 3 retries fail, `FetchError(502)` is thrown to `onError`, and if `onError` doesn't retry, the stream dies. The expired entry persists in localStorage, so the next session hits the same wall — permanently.

Since the server never reuses handles (now documented as **SPEC.md S0**), the expired entry becomes a false positive once the caching layer clears — but the client has no way to discover this.

## Approach

After stale cache retries exhaust (3 attempts), the client now:

1. **Always clears the expired entry** from localStorage — if cache busters didn't work, keeping the entry only poisons future sessions
2. **Attempts one self-healing retry** — resets the stream and retries without the `expired_handle` param. Since handles are never reused, the fresh response will have a new handle and won't trigger stale detection
3. **Guards against infinite loops** via `#expiredShapeRecoveryKey` (once per shape key, reset on up-to-date)

```typescript
if (transition.exceededMaxRetries) {
  if (shapeKey) {
    expiredShapesCache.delete(shapeKey)       // always clear
    if (this.#expiredShapeRecoveryKey !== shapeKey) {
      this.#expiredShapeRecoveryKey = shapeKey // remember we tried
      this.#reset()                            // fresh start
      throw new StaleCacheError(...)           // caught internally → retry
    }
  }
  throw new FetchError(502, ...)               // truly give up
}
```

### Key Invariants

- **S0**: Server handles are unique and never reused (phash2 + microsecond timestamp, SQLite UNIQUE INDEX, ETS insert_new)
- Self-healing fires at most once per shape per retry cycle (`#expiredShapeRecoveryKey` guard)
- Guard resets on up-to-date, so long-lived streams can self-heal again if CDN misbehaves later
- Expired entry is cleared on every exhaustion, regardless of whether self-healing fires

### Non-goals

- TTL on expired cache entries — the self-healing mechanism handles the failure mode without added complexity
- Changing `onError` contract — the fix works regardless of what the user's `onError` callback does

## Verification

```bash
cd packages/typescript-client
# 312 tests pass
pnpm exec tsc --noEmit
# Clean
```

## Files changed

| File | Change |
|------|--------|
| `src/client.ts` | Self-healing logic in `#onInitialResponse`, recovery key cleared on up-to-date, updated catch block comment |
| `test/expired-shapes-cache.test.ts` | Updated 2 existing tests for self-healing flow, added test for CDN-always-stale scenario |
| `SPEC.md` | Added S0 (handle uniqueness guarantee), updated L3 loop-back entry and guard table |
| `.changeset/fix-expired-shapes-self-healing.md` | Changeset for patch release |

---
Based on #4085


balegas:
## Code Review

### Overview
The core fix is correct. I traced the stale-retry → self-healing flow end-to-end and verified body cancellation, guard clearing race-freedom, and reset ordering all hold. `SPEC.md S0` is the right place to anchor the handle-uniqueness assumption the mechanism depends on.

### Suggestions

**1. The 204 guard-clearing check is looser than the comment suggests (`client.ts:1244-1246`)**

```typescript
if (transition.action === `accepted` && this.#syncState.kind === `live`) {
  this.#expiredShapeRecoveryKey = null
}
```

The comment says "e.g., 204" but the check fires whenever `accepted` + state becomes `live`. It works today only because the state machine doesn't transition to `live` at response-metadata stage for 200s — if that assumption ever changes, this will start clearing the guard prematurely. Consider either:
- Narrowing to an explicit `status === 204` check (matches the comment's intent), or
- Adding a state-machine assertion/test to lock in the assumption

Not blocking — defensive hardening.

**2. `#reset()` comment is now inaccurate (`client.ts:1770`)**

The comment on `#reset()` says `it's called from within the running stream loop (#requestShape's 409 handler)` — self-healing is now a second caller. A future edit that relies on "only 409 handler" semantics could break self-healing. One-line fix:

```
it's called from within the running stream loop (#requestShape's 409 handler
or stale-retry self-healing)
```

**3. Self-healing silently accepts stale data**

The "CDN always returns stale handle" case documents that after self-healing clears the expired entry, the client happily accepts the same stale response. The trade-off (stale data > permanent 502) is the right call, but there's no post-hoc signal to the user that they're on stale data. Consider logging a follow-up warning when a post-self-healing response arrives with the handle that was just marked-and-cleared. Low priority — could be a separate PR if telemetry becomes a need.

### Test concerns

**1. Timing-sensitive `setTimeout` waits** (`200ms`, `500ms`, `1000ms`)

The test history (`1f7a5f4`, `3c90f21`) shows the author already fought flakiness here. Consider adopting the deterministic `FetchGate` pattern from #4089 once that lands — turn-based gating eliminates all timing races. The two PRs are complementary.

**2. The 204-path test is doing too much** (`should clear recovery guard after 204 so self-healing works again`)

105 lines of mock logic with multi-phase state (`selfHealCount`, `healingDone`, mid-test `markExpired`, manual `aborter.abort()`), and asserts only `selfHealCount === 2` because the fast-loop detector may fire on slower machines. That's a signal the test is conflating concerns. Consider splitting:
- Unit test: "204 clears recovery guard" against `#onInitialResponse` directly
- Integration test: "Self-healing can fire multiple times after recovery guard clears"

**3. `markExpired` mutation inside `fetchMock.mockImplementation`** (`test/expired-shapes-cache.test.ts:418`)

Mutating `expiredShapesCache` from inside the fetch mock couples the test to an event ordering that's not obvious to future readers. At minimum, a `// Simulate a second 409 happening mid-stream` comment; ideally, restructure so the mutation happens outside the mock callback.

### Recommendation

Approve with minor suggestions. Fix #1 (tighten the 204 check) and #2 (comment update) are cheap safety wins worth doing before merge. The rest are non-blocking.

msfstef (approved):

PR #4088, opened 2026-04-03
fix(client): bound the onError retry loop to prevent unbounded retries

KyleAMathews:
## Summary

Bounds the `onError` retry loop so that persistent errors (e.g. a misconfigured proxy returning 400) no longer create unbounded recursive retries with unbounded memory growth. Adds static analysis to prevent regressions.

## Root Cause

When `onError` always returns a retry directive (as TanStack DB's `shouldRetryOnFailure` does), the `#start` catch block calls `this.#start()` recursively for every error. Two gaps made this unbounded:

1. **No retry limit** — the catch block had no counter, so retries continued forever.
2. **Fast-loop detector bypass** — `ErrorState` wrapping `LiveState` delegates `isUpToDate` → `true`, causing the fast-loop check to be skipped entirely and its counters to reset in the else branch.

The combination means a persistent 400 (which bypasses backoff in `createFetchWithBackoff`) creates a tight recursive loop that grows the heap until the process crashes.

## Approach

**Consecutive error retry counter** (`#consecutiveErrorRetries`) on `ShapeStream`:

- Incremented on each `onError` retry in `#start`'s catch block
- When it exceeds 50, the stream logs a warning, notifies subscribers of the error, and tears down
- Resets on successful data: non-empty message batch in `#onMessages`, or accepted 204 in `#onInitialResponse`

The 204-specific reset is deliberate: 204 has no body to parse, so resetting on accepted headers is safe. For normal 200s the counter only resets *after* successful batch processing — otherwise a malformed 200 (invalid JSON from a proxy) would reset the counter on every iteration and defeat the limit.

**Tail-call optimization for error recovery**: the retry path uses `return this.#start()` instead of `await this.#start()` to release the outer async frame immediately, preventing frame accumulation across multiple recovery cycles.

**ErrorState guard** at the top of `#requestShape`: if the state machine is somehow in `ErrorState` when entering the request loop, re-throw so `#start`'s catch can route through `onError` properly.

**Static analysis** (`buildUnboundedRetryReport`): walks the AST to find every recursive call inside a catch block and heuristically checks each has a recognized bounding pattern — either a counter-check (field comparison + return/throw), an error-type-guard (instanceof/status-code narrowing), or an abort signal check. This is smoke detection for regressions, not a formal proof of boundedness — novel patterns may need manual verification.

## Key Invariants

1. The retry counter is monotonically increasing during consecutive errors and strictly resets only on proven successful data flow
2. Every recursive call in a catch block across `ShapeStream` has a recognized bounding pattern (heuristically enforced by static analysis test)
3. The fast-loop detector bypass is mitigated by the retry counter firing before it matters

## Non-goals

- Making the 50-retry limit configurable (can be added later if needed)
- Fixing the `isUpToDate` delegation in `ErrorState` (the retry counter makes this safe)

## Verification

```bash
cd packages/typescript-client
# 316 tests pass
```

Key test cases:
- `onError retry loop should be bounded for persistent errors` — verifies the limit fires
- `onError retry counter resets after successful data` — 60 errors across 2 bursts of 30, separated by success
- `204 No Content responses reset the consecutive error retry counter` — 60 errors separated by 204s
- `malformed 200 responses are bounded by the retry counter` — 200 OK with invalid JSON body
- `does not report unbounded retry loops in recursive catch blocks` — static analysis guard

## Files Changed

| File | Change |
|------|--------|
| `src/client.ts` | Add retry counter fields, bound check in `#start` catch, ErrorState guard in `#requestShape`, counter resets in `#onMessages` and `#onInitialResponse` (204 only), `return this.#start()` for frame release |
| `bin/lib/shape-stream-static-analysis.mjs` | Add `buildUnboundedRetryReport` — heuristic detection of recursive calls in catch blocks without recognized bounding patterns |
| `test/stream.test.ts` | 4 new tests: bounded retry, counter reset, 204 reset, malformed 200 bound |
| `test/static-analysis.test.ts` | 1 new test: no unbounded retry findings |
| `test/expired-shapes-cache.test.ts` | Add `onError` handlers to 3 streams to prevent unhandled rejections from fast-loop 502 errors |
| `SPEC.md` | Document `#maxConsecutiveErrorRetries` guard |

### DIFF
--- a/.changeset/bound-onerror-retry-loop.md
+++ b/.changeset/bound-onerror-retry-loop.md
@@ -1,0 +1,5 @@
+---
+'@electric-sql/client': patch
+---
+
+Bound the `onError` retry loop to prevent unbounded retries and memory growth. When `onError` always returns a retry directive for a persistent error (e.g. a 400 from a misconfigured proxy), the client now limits consecutive retries to 50 before tearing down the stream and notifying subscribers. The counter resets on successful data (non-empty message batch or 204 No Content), so intermittent errors that recover do not accumulate toward the limit.
--- a/packages/typescript-client/SPEC.md
+++ b/packages/typescript-client/SPEC.md
@@ -352,22 +352,22 @@ Six sites in `client.ts` recurse or loop to issue a new fetch:
 | L2  | `#requestShape` catch → `#requestShape` | 874  | Abort with `FORCE_DISCONNECT_AND_REFRESH` or `SYSTEM_WAKE` | `isRefreshing` flag changes `canLongPoll`, affecting `live` param                   | Abort signals are discrete events                       |
 | L3  | `#requestShape` catch → `#requestShape` | 886  | `StaleCacheError` thrown by `#onInitialResponse`           | `StaleRetryState` adds `cache_buster` param                                         | `maxStaleCacheRetries` counter in state machine         |
 | L4  | `#requestShape` catch → `#requestShape` | 924  | HTTP 409 (shape rotation)                                  | `#reset()` sets offset=-1 + new handle; or request-scoped cache buster if no handle | New handle from 409 response or unique retry URL        |
-| L5  | `#start` catch → `#start`               | 782  | Exception + `onError` returns retry opts                   | Params/headers merged from `retryOpts`                                              | User-controlled; `#checkFastLoop` on next iteration     |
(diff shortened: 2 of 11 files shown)


balegas (approved):
Do we need this? I would try other fixes first and see if the problem goes away. leaving pre-approved.

KyleAMathews:
What makes you think we don't need it?

PR #4097, opened 2026-04-07
Document why shape handles get deleted

### LINKED ISSUE #2680: Feature request: document why a shape handle might get deleted
**Is your feature request related to a problem? Please describe.**
People wonder why shapes get deleted. Especially if they have very large shapes as recreating and reloading them into clients can get expensive.

If they understand why it's happening, they might be able to reduce how often shapes are recreated by e.g. preserving their shapes logs when restarting their Electric instance.

**Describe the solution you'd like**
Document the reasons for shape deletions:

1) the server is restarted and loses all its shape logs
2) pg has an error of some sort that isn't recoverable so electric deletes all the shapes
3) probably most common, there's a schema change to the table the shape uses so the log is missing the new data

fru1tworld:
## Summary
- Added a "Why shape handles get deleted" section to the shapes guide

Covers: server restart, schema changes, unrecoverable Postgres errors, LRU eviction, explicit API deletion, and client-side behavior.

Fixes #2680


alco (approved):
👍

alco:
Thanks for the contribution @fru1tworld !

PR #4101, opened 2026-04-08
fix: move admission control before shape creation

msfstef:
## Summary

- **Fixed admission control bypass** where shapes were created before admission control could reject requests, defeating its purpose under load
- **Fixed pre-existing double-decrement bug** in `handle_errors` where admission control permits could be released twice (or spuriously released when no permit was acquired)

## Problem

In `ServeShapePlug`, the plug pipeline was:

```
plug :validate_request      # ← calls Api.validate() which creates shapes
plug :check_admission       # ← admission control runs AFTER shapes already exist
plug :serve_shape_response
```

`Api.validate/2` calls `load_shape_info()` → `get_or_create_shape_handle()`, which creates shapes in ETS/SQLite and starts consumer processes — all **before** admission control runs. A flood of requests for new shapes would create all shapes and start all consumers, then admission control would reject the requests. The expensive work had already happened.

Additionally, `handle_errors` called `ensure_admission_control_release` unconditionally before calling `send_resp`. Since `check_admission` registers a `before_send` callback that also releases the permit, this caused a double-decrement when exceptions occurred after permit acquisition. If the exception occurred *before* `check_admission`, the unconditional release would spuriously decrement the counter despite no permit being acquired. The ETS counter floor of 0 masked the issue (it never went negative), but it still undercounted active requests.

## Validation

Added a deterministic integration test (`router_test.exs`) that proves the bug:

1. Pre-fills admission control slots so the next request will be rejected
2. Sends an initial request (`offset=-1`) for a new shape
3. Asserts the request is rejected with 503
4. **Asserts no shape was created** (`ShapeCache.count_shapes == 0`)

On the original code this test **fails** — the shape count is **1** (shape created despite rejection). After the fix it passes.

## Solution

Split `Api.validate/2` into two public functions:
- **`validate_params/2`** — parameter validation + `Shape.new()` only (no shape creation in cache)
- **`load_shape_info/1`** — shape creation/lookup + seek (the expensive part)

`validate/2` remains as a convenience wrapper calling both (backward-compatible for elixir-client, Phoenix.Sync, existing tests).

Reordered the plug pipeline:

```
plug :validate_request      # param validation only (no shape creation)
plug :check_admission       # admission control BEFORE shape creation
plug :load_shape            # shape creation/lookup (after admission)
plug :serve_shape_response
```

This means:
- Invalid requests still fail fast before admission control (no wasted permits)
- Shape creation only happens for admitted requests
- Permit release via `register_before_send` handles errors in `load_shape` correctly
- Existing shapes (with handle) go through cheap `resolve_shape_handle` in `load_shape`

Removed `ensure_admission_control_release` from `handle_errors` — the `before_send` callback registered in `check_admission` is the single correct mechanism for releasing permits, and it fires when `send_resp` is called in the error handler path.

## Test plan

- [x] New test fails on original code, passes after fix
- [x] All 84 router tests pass
- [x] Full test suite passes (1532 tests + 391 doctests + 8 properties, 0 failures)

### DIFF
--- a/.changeset/fix-admission-control-ordering.md
+++ b/.changeset/fix-admission-control-ordering.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix admission control bypass where shapes were created before admission control checks. Shape creation now happens after admission control, preventing resource exhaustion under load.
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -20,12 +20,13 @@ defmodule Electric.Plug.ServeShapePlug do
   plug :parse_body
 
   plug :validate_request
-  # Admission control is applied here, but note:
-  # - /v1/health bypasses this plug (routed to HealthCheckPlug)
-  # - /metrics bypasses this plug (on separate utility router/port)
-  # - / (root) bypasses this plug (handled directly in router)
-  # This ensures observability remains available under load
+  # Admission control is applied here, after parameter validation but before
+  # shape creation. This ensures invalid requests are rejected cheaply without
+  # consuming a permit, while shape creation (the expensive part) only happens
+  # for admitted requests.
+  # Note: /v1/health, /metrics, and / bypass this plug entirely.
   plug :check_admission
+  plug :load_shape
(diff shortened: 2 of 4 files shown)


icehaunter (approved):

alco (approved):
Hard approve

PR #4105, opened 2026-04-09
Fix replication connection drops from wal_sender_timeout during backpressure

msfstef:
## Problem

When the BEAM is busy or downstream event processing is slow/failing, the replication client's `apply_with_retries` blocks inside `handle_data`, preventing the gen_statem from responding to PostgreSQL's keepalive requests. After `wal_sender_timeout` (default 60s), PostgreSQL terminates the replication connection with errors like:

```
%DBConnection.ConnectionError{message: "ssl send: closed", severity: :error, reason: :closed}
```

The root cause is that PostgreSQL's replication protocol requires the client to periodically send `StandbyStatusUpdate` messages. These are **application-level** heartbeats — TCP keepalives (handled by the kernel) don't help. When the gen_statem is stuck in a blocking retry loop inside `handle_data`, it cannot send these messages, and PostgreSQL kills the connection.

## Investigation

We confirmed this is a [well-known problem](https://www.postgresql.org/message-id/CAMsr+YE2dSfHVr7iEv1GSPZihitWX-PMkD9QALEGcTYa+sdsgg@mail.gmail.com) across PostgreSQL logical replication consumers (Debezium, pgjdbc, pg_recvlogical). PostgreSQL's protocol has no flow control mechanism — keepalive handling is coupled with data processing on a single connection.

Key findings:
- **TCP keepalives don't help** — they're kernel-level and work regardless of process scheduling, but `wal_sender_timeout` checks for application-level `StandbyStatusUpdate` messages
- **SSL has no heartbeat** — TLS is just a record layer on top of TCP
- **`wal_sender_timeout`** is the culprit — it fires when the client fails to send any `StandbyStatusUpdate` within the timeout window
- **Sending `StandbyStatusUpdate` without advancing the LSN is safe** — it resets `last_reply_timestamp` without affecting the replication slot's `confirmed_flush_lsn`

## Solution

The fix addresses two requirements that are in tension:
1. **Send keepalives** while event processing is blocked, to prevent `wal_sender_timeout`
2. **Stop the replication stream** from pushing more data over the socket, to provide backpressure

### Vendored `Postgrex.ReplicationConnection` with socket pause/resume

We vendor `Postgrex.ReplicationConnection` as `Electric.Postgres.ReplicationConnection`, adding two new callback return types:

- `{:noreply_and_pause, ack, state}` — send ack messages, then **pause socket reads**. The gen_statem stops receiving new WAL data but remains responsive to `handle_info` messages (timers, notifications).
- `{:noreply_and_resume, ack, state}` — send acks, process any buffered data, resume socket reads.

The pause mechanism leverages `{active, :once}` — by not re-arming the socket after processing a batch, no new data enters the process. At most one Erlang message is buffered. TCP flow control naturally backpressures PostgreSQL's walsender when the kernel receive buffer fills (~128KB-6MB depending on OS).

### Non-blocking event dispatch in `ReplicationClient`

Instead of blocking in `handle_data` via `apply_with_retries`:

```
handle_data(XLogData) → dispatch_event(event) → {:noreply_and_pause, [], state}
```

The event is dispatched as a `{:process_event, ...}` message to self. The gen_statem returns immediately and is free to process other messages.

### Async event handler via `$gen_call` + `Process.monitor`

`apply_event` dispatches the event handler as a non-blocking `$gen_call` (the same protocol `GenServer.call` uses internally, and the same pattern as `StatusMonitor.wait_until_async`). The MFA returns a monitor ref; the gen_statem stores it and returns `{:noreply, state}` immediately. The reply (`{ref, :ok}` or `{ref, {:error, reason}}`) and crash detection (`{:DOWN, ref, ...}`) are handled in `handle_info`.

**Zero per-event process spawning** — just `Process.monitor(pid)` + `send(pid, msg)`. The existing `handle_call({:handle_event, ...})` in ShapeLogCollector handles the request unchanged.

On success, `{:noreply_and_resume, acks, state}` resumes the socket. On failure, retries are scheduled via `Process.send_after` or `StatusMonitor.wait_until_async`.

### Periodic keepalive timer

A `StandbyStatusUpdate` is sent every `min(wal_sender_timeout/3, 15s)`. The interval is derived from PostgreSQL's `wal_sender_timeout` (queried from `pg_settings` during connection setup). The 15s cap ensures responsiveness even if the timeout is set very high or changes after connection.

### Event-driven retry with `StatusMonitor.wait_until_async`

Replaces the blocking `StatusMonitor.wait_until_active(timeout: :infinity)` with a new generic `wait_until_async/2` that subscribes to status transitions and notifies the caller via a tagged message. This also replaces the previous `wait_until_conn_up_async` with a single generic mechanism.

## Changes

| File | Change |
|------|--------|
| `lib/electric/postgres/replication_connection.ex` | **New** — vendored `Postgrex.ReplicationConnection` with socket pause/resume |
| `lib/electric/postgres/replication_client.ex` | Non-blocking event dispatch via `$gen_call` + `Process.monitor`, keepalive timer, async retry |
| `lib/electric/postgres/replication_client/connection_setup.ex` | Query `wal_sender_timeout` from `pg_settings` during setup |
| `lib/electric/replication/shape_log_collector.ex` | `handle_event_async/2` (non-blocking `$gen_call` wrapper), `set_event_processing_mode/2` for integration testing |
| `lib/electric/stack_supervisor.ex` | Use `handle_event_async` MFA |
| `lib/electric/status_monitor.ex` | Generic `wait_until_async/2` (replaces `wait_until_conn_up_async`) |
| `lib/electric/connection/restarter.ex` | Use `wait_until_async` |
| `test/electric/status_monitor_test.exs` | Tests for `wait_until_async` |
| `test/electric/postgres/replication_client_test.exs` | Tests: connection survives `wal_sender_timeout` during backpressure (unavailable handler AND slow handler) |
| `test/electric/replication/shape_log_collector_test.exs` | Tests for `set_event_processing_mode` (:suspend, :crash, :hang) |
| `integration-tests/tests/replication-keepalive-during-backpressure.lux` | Lux integration test |

## Test plan

- [x] All 1551 existing unit tests pass + 3 new mode tests
- [x] 6 new `wait_until_async` tests pass
- [x] New Elixir test: connection survives `wal_sender_timeout=3s` with handler **unavailable** (6s window)
- [x] New Elixir test: connection survives `wal_sender_timeout=3s` with handler **slow** (6s delay) — **fails without async $gen_call**, passes with fix
- [x] Lux integration test: connection survives `wal_sender_timeout=5s` with 10s of suspended processing
- [x] 21/22 existing lux integration tests pass (1 pre-existing flaky failure unrelated to this change)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.changeset/fix-replication-keepalive-timeout.md
+++ b/.changeset/fix-replication-keepalive-timeout.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix replication connection drops caused by PostgreSQL's wal_sender_timeout during backpressure. The replication client now sends periodic keepalive messages while event processing is paused, preventing the connection from being killed during slow downstream processing.
--- a/integration-tests/tests/replication-keepalive-during-backpressure.lux
+++ b/integration-tests/tests/replication-keepalive-during-backpressure.lux
@@ -1,4 +1,4 @@
-[doc Verify that the replication connection survives when event processing is stuck for longer than wal_sender_timeout. When ShapeLogCollector returns not_ready, the replication client retries asynchronously while sending periodic keepalives that prevent PostgreSQL from killing the connection.]
+[doc Verify that the replication connection survives when event processing is stuck for longer than wal_sender_timeout. The replication client dispatches events via non-blocking $gen_call and sends periodic keepalives that prevent PostgreSQL from killing the connection.]
 
 [include _macros.luxinc]
 
@@ -40,21 +40,16 @@
   [invoke shape_get_snapshot items]
   ??before-block
 
-## Suspend event processing in ShapeLogCollector so it returns {:error, :not_ready}.
+## Set event processing mode to :suspend so SLC returns {:error, :not_ready}.
 ## This simulates sustained backpressure on the replication stream.
 ##
-## In the old code, apply_with_retries loops inside handle_data:
-##   sleep(50) -> wait_until_active(infinity) -> retry -> :not_ready -> repeat
-## The gen_statem never returns to its event loop, so keepalive timer messages
-## pile up unprocessed. After wal_sender_timeout, PostgreSQL kills the connection.
-##
-## With the fix, event dispatch is async: the gen_statem returns between retries,
-## allowing the keepalive timer to fire and send StandbyStatusUpdate messages.
+## The gen_statem dispatches events via a non-blocking $gen_call, keeping it
+## responsive to keepalive timers while waiting for the handler to respond.
 [shell electric]
   # Disable the fail pattern — we expect error/warning logs during the retry period
   -
 
-  !Electric.Replication.ShapeLogCollector.suspend_event_processing("single_stack")
+  !Electric.Replication.ShapeLogCollector.set_event_processing_mode("single_stack", :suspend)
   ??:ok
 
 ## Insert data to trigger event processing — this starts the :not_ready retry loop
@@ -78,7 +73,7 @@
   # or the gen_statem terminating with "tcp send: closed".
   -connection closed|tcp send: closed|ssl send: closed|$fail_pattern
 
-  !Electric.Replication.ShapeLogCollector.resume_event_processing("single_stack")
+  !Electric.Replication.ShapeLogCollector.set_event_processing_mode("single_stack", :normal)
   ??:ok
 
 ## Insert data that should flow through if the connection survived
--- a/integration-tests/tests/replication-keepalive-during-backpressure.lux
+++ b/integration-tests/tests/replication-keepalive-during-backpressure.lux
@@ -1,0 +1,100 @@
+[doc Verify that the replication connection survives when event processing is stuck for longer than wal_sender_timeout. When ShapeLogCollector returns not_ready, the replication client retries asynchronously while sending periodic keepalives that prevent PostgreSQL from killing the connection.]
+
+[include _macros.luxinc]
+
+[global pg_container_name=repl-keepalive__pg]
+
+###
+
+## Start Postgres with a low wal_sender_timeout.
+## Electric derives its keepalive interval as min(wal_sender_timeout/3, 15s).
+## With wal_sender_timeout=5s, keepalives fire every ~1.6s.
+## With the old blocking code, the replication client would be stuck inside
+## handle_data for the entire retry period — no keepalives sent — and PG would
+## kill the connection after 5s. With the fix, keepalives fire every ~1.6s.
+[invoke setup_pg_with_shell_name "pg" "" "-c wal_sender_timeout=5s" ""]
+
+## Add a test table
+[invoke start_psql]
+[shell psql]
+  """!
+  CREATE TABLE items (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    val TEXT
+  );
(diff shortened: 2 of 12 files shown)


msfstef:
Thanks for the thorough review! Addressed the actionable items in 0393830:

**Fixed:**
- **1 (catch-all for stale StatusMonitor messages)** — Added a catch-all `handle_info` clause that silently discards `{{Electric.StatusMonitor, _ref}, _result}` messages when `wait_for_active_ref` doesn't match. Prevents `FunctionClauseError` on stale notifications.
- **4 (Logger level)** — Downgraded keepalive interval log from `info` to `debug`.
- **6 (_remaining parameter)** — Added comment explaining the intentional budget reset (matches old `apply_with_retries` behavior).
- **3 (changeset)** — Added `.changeset/fix-replication-keepalive-timeout.md`.

**Not addressed (by design):**
- **2 (StatusMonitor restart)** — Not an issue. `StatusMonitor` is a child of `MonitoredCoreSupervisor` which uses `:rest_for_one` strategy, with `StatusMonitor` listed before `CoreSupervisor`. If `StatusMonitor` crashes, everything downstream (including `ReplicationClient`) is restarted. The supervision tree comment in `monitored_core_supervisor.ex` documents this explicitly.
- **5 (dead code clause)** — Already addressed in an earlier commit (42ca443) — turned it into a live defensive `Logger.warning` that fires if a second socket message arrives while paused (shouldn't happen with `{active, :once}` but guards against future OTP changes).

msfstef:
benchmark this

alco:
Wow, these benchmark results are music!

msfstef:
benchmark this

msfstef:
@alco turns out it was against 1.2.8 so I updated the reference image and ran them again! there's no regression which is good (and mostly what I expected, the previous results were unexpectedly better)

alco (approved):
Brilliant!

icehaunter (commented):
This covers the case of SLC crashing & retries, but if SLC is taking too long on a transaction for any reason (e.g. writign to all 200k consumers) then we're still blocked on a GenServer.call

msfstef:
benchmark this

msfstef (commented):
@icehaunter Good catch — you're right, the previous commit only made the dispatch async but `apply_event` still blocked on the `GenServer.call(:infinity)` to SLC.

Fixed in 7a695d1c + 3ad53afe:

**`handle_event_async/2`** replaces the blocking `GenServer.call` with a manual `$gen_call` + `Process.monitor` — the same pattern as `StatusMonitor.wait_until_async`. The gen_statem sends the message and returns immediately; the reply (or `:DOWN` on crash) is handled in `handle_info`. Zero per-event process spawning.

**New test** ("connection survives wal_sender_timeout when event handler is slow") validates a 6s handler delay against a 3s `wal_sender_timeout` — this specifically covers the case you flagged where SLC is taking a long time (e.g. writing to 200k consumers via `ConsumerRegistry.broadcast`).

Also consolidated `suspend/resume_event_processing` into `set_event_processing_mode/2` with `:suspend`, `:hang`, `:crash`, and `:normal` modes for integration testing.

alco (approved):
🤘

PR #4106, opened 2026-04-09
fix: send HTTP 102 keepalive during long-poll to prevent middlebox timeouts

msfstef:
## Summary

- Sends periodic HTTP `102 Processing` informational responses during long-poll holds to prevent network middleboxes (Cloudflare edge nodes on long paths to origin) from dropping idle connections (522 errors)
- Zero client changes required — fully backwards compatible with existing TypeScript and Elixir clients
- All 51 existing API tests pass with no modifications

## Problem

Long-poll requests to `/v1/shape?live=true` hold connections idle for up to 20 seconds inside `hold_until_change`. During this time, no bytes flow on the wire. Middleboxes on long network paths (e.g., Cloudflare colos BAH, HKG hitting ALB in us-east-1) can drop these connections, causing sporadic 522 errors.

## Approach

Uses `Plug.Conn.inform(conn, 102)` to send HTTP 1xx informational responses every 5–15 seconds during the hold. These responses:

1. **Send bytes on the wire** — keeps middlebox connections alive
2. **Don't commit the final response** — headers and status code are sent only after the hold resolves, with correct values
3. **Are invisible to HTTP clients** — `fetch()`, Req, and all standard HTTP clients transparently ignore 1xx responses per RFC 9110 §15.2
4. **Preserve CDN request collapsing** — since 1xx responses are not final responses, the CDN's coalescing window remains open for the full hold duration (see CDN section below)

### Implementation details

- A `:timer.send_interval` starts when a live non-SSE request enters the Plug path (`serve_shape_response/2` or `serve_shape_log/2`)
- A `register_before_send` callback cancels the timer and flushes stale messages when the response is sent
- The keepalive callback is stored as a closure (`on_keepalive`) on the `Request` struct, keeping `Plug.Conn` out of the domain struct
- `hold_until_change` is split into two functions: the outer sets up a `Process.send_after` timeout timer (replacing `receive...after` which would reset on keepalive re-entry), the inner `do_hold_until_change` handles the receive loop with a new `:long_poll_keepalive` clause
- Keepalive interval is `div(long_poll_timeout, 4)` clamped to 5–15 seconds

### Chunked transfer encoding with whitespace keepalive bytes

Send `Transfer-Encoding: chunked` immediately and emit `" "` (space) chunks as keepalive during the hold, then send the real JSON payload as the final chunk.

**Rejected for two reasons:**

1. **Breaks backwards compatibility.** HTTP headers are committed when chunked encoding starts, before the hold resolves. The `electric-offset` header would be stale when new data arrives during the hold. This requires a TypeScript client change (using body-based offset instead of header-based) to avoid an infinite re-fetch loop — breaking old clients on the new server.

2. **Breaks CDN request collapsing.** Live long-poll responses are cacheable (`Cache-Control: public, max-age=5, stale-while-revalidate=5`). With the current non-chunked approach, CDNs like Cloudflare hold the coalescing window open for the full hold duration — multiple clients requesting the same shape+offset are collapsed into one origin request. Starting a chunked response immediately closes this window on Cloudflare and CloudFront because the response is already "in flight." With 102, the coalescing window stays open since 1xx responses are not final responses.

### Stream.resource pattern (modeled on existing SSE keepalive)

Replace `hold_until_change` entirely with a `Stream.resource` that emits keepalive chunks and data chunks as a streaming response body.

**Rejected because** it changes the response semantics: the `Response` struct fields (`offset`, `up_to_date`, `status`) are frozen at stream creation time rather than set after the hold resolves. This broke 11 existing tests and required changing error behavior (shape rotation, out-of-bounds, stack failure all became 200 with empty body instead of their proper status codes). Also has the same CDN collapsing and backwards-compatibility issues as the chunked approach.

### Shorter long-poll timeout

Reduce `long_poll_timeout` to well under the middlebox timeout.

**Rejected because** it doubles the request rate for idle shapes without solving the root cause.

## On HTTP 102 Processing

`102 Processing` was defined in RFC 2518 (WebDAV, 1999) specifically for preventing client timeouts on long-running requests — the exact use case here. It was removed in RFC 4918 (2007) with the stated reason: **"due to lack of implementation"** — not because it was harmful or architecturally unsound.

MDN labels 102 as "deprecated," but this is an editorial judgment, not a standards-body action. No RFC has ever formally deprecated 102. The status is more accurately described as "no longer defined by any active RFC" — a state shared by many widely-used status codes (429 Too Many Requests was also defined outside RFC 9110's predecessor until recently).

### Why it's safe to use

- **RFC 9110 §15.2 requires** all HTTP/1.1+ clients to parse unknown 1xx responses and allows ignoring them — sending 102 cannot break spec-compliant implementations
- **RFC 9110 §15.1 explicitly acknowledges** status codes "outside the scope of this specification" as legitimate, provided they are IANA-registered
- **102 is permanently registered** in the [IANA HTTP Status Code Registry](https://www.iana.org/assignments/http-status-codes) with no deregistration attempts
- **Cloudflare explicitly supports and recommends it** for keepalive: ["If Cloudflare receives a 102 Processing response, it expects a final response within 120 seconds"](https://developers.cloudflare.com/support/troubleshooting/http-status-codes/1xx-informational/)
- **All major proxies forward it**: nginx, HAProxy, Envoy (since [envoyproxy/envoy#19023](https://github.com/envoyproxy/envoy/pull/19023))
- **HTTP/2 (RFC 9113 §8.8.5)** explicitly handles 1xx informational responses including 102
- **Bandit supports it** natively via `Plug.Conn.inform/2`; **Node.js** has `response.writeProcessing()` since v10; **Go** added 1xx support in Go 1.19
- **No other 1xx code fits**: `100 Continue` is for request bodies, `103 Early Hints` is for resource preloading and would confuse browsers

### Known limitations

- **Go's `net/http`** has a default limit of 5 consecutive 1xx responses. Our 3–4 responses per hold (at 5s intervals over 20s) are within this limit.
- **AWS ALB** may silently drop 1xx responses — this is harmless since ALB's idle timeout (default 60s, configurable up to 4000s) already exceeds typical long-poll durations.
- **Spring Framework 7.0** deprecated the `HttpStatus.PROCESSING` enum constant. This is the only major framework taking action, and it's the constant, not the protocol behavior.

## CDN request collapsing

Electric's live long-poll responses are cacheable (`Cache-Control: public, max-age=5, stale-while-revalidate=5`), which enables CDN request collapsing — multiple clients requesting the same shape+offset are collapsed into a single origin request.

The 102 approach preserves this because 1xx informational responses are not final responses. The CDN continues to hold the coalescing window open until the final 200 arrives. This is significant under high concurrency where many clients subscribe to the same shape.

By contrast, the alternative chunked-whitespace approach would close the coalescing window immediately by committing a 200 response before the hold resolves.

## Test plan

- [x] All 51 existing `api_test.exs` tests pass with no modifications
- [x] All 311 TypeScript client unit tests pass with no modifications
- [ ] Integration test with real HTTP server verifying 102 responses arrive on the wire
- [ ] Manual verification against Cloudflare-fronted deployment

### DIFF
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -1185,6 +1185,11 @@ defmodule Electric.Shapes.Api do
         conn
       end)
 
+    # Return value intentionally ignored — if the adapter raises (e.g.
+    # half-closed connection), the exception propagates through
+    # do_hold_until_change and the try/after block cancels the timeout
+    # timer before the process crashes. This is acceptable: Bandit will
+    # clean up the connection.
     on_keepalive = fn -> Plug.Conn.inform(conn, 102) end
 
     {conn, %{request | on_keepalive: on_keepalive}}
--- a/packages/sync-service/lib/electric/shapes/api.ex
+++ b/packages/sync-service/lib/electric/shapes/api.ex
@@ -539,7 +539,7 @@ defmodule Electric.Shapes.Api do
   end
 
   def serve_shape_response(%Plug.Conn{} = conn, %Request{} = request) do
-    {conn, request} = start_long_poll_keepalive(conn, request)
+    request = set_long_poll_keepalive(conn, request)
 
     response =
       case if_not_modified(conn, request) do
@@ -607,7 +607,7 @@ defmodule Electric.Shapes.Api do
   end
 
   def serve_shape_log(%Plug.Conn{} = conn, %Request{} = request) do
-    {conn, request} = start_long_poll_keepalive(conn, request)
+    request = set_long_poll_keepalive(conn, request)
 
     response =
       case if_not_modified(conn, request) do
@@ -859,14 +859,14 @@ defmodule Electric.Shapes.Api do
     # Combined with handler_fullsweep_after config, this helps keep handler memory usage low.
     :erlang.garbage_collect()
 
-    # Use a timer instead of `receive...after` so that the timeout is not
-    # reset when we re-enter receive to handle keepalive messages.
+    keepalive_timer = start_keepalive_timer(request)
     timeout_timer = Process.send_after(self(), {:long_poll_timeout, ref}, long_poll_timeout)
 
     try do
       do_hold_until_change(request)
     after
       Process.cancel_timer(timeout_timer)
+      cancel_keepalive_timer(keepalive_timer)
     end
   end
 
@@ -918,10 +918,6 @@ defmodule Electric.Shapes.Api do
         end
 
       :long_poll_keepalive ->
-        # Invoke the keepalive callback (sends HTTP 102 Processing) to keep
-        # the connection alive through network middleboxes that drop idle
-        # connections. Invisible to HTTP clients — they only see the final
-        # response.
         if on_keepalive, do: on_keepalive.()
         do_hold_until_change(request)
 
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -855,6 +859,25 @@ defmodule Electric.Shapes.Api do
     # Combined with handler_fullsweep_after config, this helps keep handler memory usage low.
     :erlang.garbage_collect()
 
+    # Use a timer instead of `receive...after` so that the timeout is not
+    # reset when we re-enter receive to handle keepalive messages.
```
alco: This comment doesn't make sense outside of the context of this diff. When this branch is merged, there won't be a receive...after anymore and anyway, the "re-entry" happens inside `do_hold_until_change()` while this comment is outside of that function.

packages/sync-service/lib/electric/shapes/api.ex
```diff
@@ -605,6 +607,8 @@ defmodule Electric.Shapes.Api do
   end
 
   def serve_shape_log(%Plug.Conn{} = conn, %Request{} = request) do
+    {conn, request} = start_long_poll_keepalive(conn, request)
```
alco: Shouldn't `start_log_poll_keepalive()` be called just before entering the "hold until change" receive block, same place where the long_poll_timeout timer starts?

alco (approved):

packages/sync-service/lib/electric/shapes/api.ex
msfstef: it should, the problem is we need access to the `conn` struct for register before send and creating the callback

my options were
1. add the `conn` struct to the `request` struct - which feels unidiomatic
2. thread through the `conn` argument for all these APIs up to `hold_until_change` - but that breaks the pattern for all other methods
3. populate the `request` struct with a callback for keepalives and the register before send up front

Alternatively I could also try to clean up the timer within the hold until change code - what's your take on these?

packages/sync-service/lib/electric/shapes/api.ex
msfstef: Good catch - I tend to keep these in for the agent as I'm working for easy remembering of things tried before but good to clean up

packages/sync-service/lib/electric/shapes/api.ex
msfstef: Ended up splitting into a method that populates the request struct with the keepalive and we handle the timers in the hold until change explicitly (wihtout doing register before send)

icehaunter (approved):

packages/sync-service/lib/electric/shapes/api.ex
alco: :+1: on the latest changes

PR #4110, opened 2026-04-09
chore(sync-service): instrument snapshot query timing

KyleAMathews:
Summary
- add a top-level snapshot query span around SnapshotQuery.execute_for_shape/4
- record checkout wait, setup, query, and total durations for snapshot-backed requests
- emit a telemetry event tagged with query_reason so subset, move-in, and initial snapshot paths can be separated in analysis

Verification
- mix compile in packages/sync-service
- mix test test/electric/shape_cache_test.exs in packages/sync-service (real-db cases fail here because local Postgres on localhost:54321 is unavailable in this environment)


alco:
@robacourt I have rewritten Kyle's version to use interval timings. Would appreciate it if you could confirm they are used correctly here.

robacourt:
> @robacourt I have rewritten Kyle's version to use interval timings. Would appreciate it if you could confirm they are used correctly here.

LGTM

alco (approved):

PR #4114, opened 2026-04-10
feat: add variadic function support to eval

icehaunter:
## Summary
- add variadic function support to the eval stack, including `defpostgres` definitions and overload lookup support
- lower `coalesce`, `greatest`, and `least` through the shared function parsing and execution path
- tighten variadic parsing behavior around explicit `VARIADIC` calls and precedence between exact and variadic overloads
- implement true `coalesce` short-circuiting in parser reduction and runtime execution

## Testing
- mix test test/electric/replication/eval_test.exs test/electric/replication/eval/*.exs


robacourt (approved):
LGTM

PR #4124, opened 2026-04-14
fix(telemetry): remove unused stack-level Prometheus reporter

### LINKED ISSUE #4123: Unbounded ETS table growth in stack-level Prometheus reporter
## Bug

The stack-level Prometheus reporter (`TelemetryMetricsPrometheus.Core`) accumulates distribution metric samples in an ETS `:duplicate_bag` table that is never drained, causing unbounded memory growth proportional to event throughput over the lifetime of the process.

## Cause

`StackTelemetry` starts a Prometheus reporter with a per-stack name:

```elixir
# stack_telemetry.ex:67-69
Reporters.Prometheus.child_spec(opts,
  name: :"stack_prometheus_telemetry_#{stack_id}",
  metrics: metrics
)
```

But the `/metrics` endpoint in `UtilityRouter` calls `TelemetryMetricsPrometheus.Core.scrape()` with no name argument, which defaults to `:prometheus_metrics` — the application-level reporter only:

```elixir
# utility_router.ex:9
get "/metrics", do: text_resp(conn, 200, TelemetryMetricsPrometheus.Core.scrape())
```

Inside `telemetry_metrics_prometheus_core`, distribution events are stored as raw samples in a `:duplicate_bag` ETS table. These samples are only drained by `scrape/1`, which calls `:ets.take` on the dist table. Since `scrape/1` is never called with the stack reporter's name, its dist table grows by one row per distribution event, forever.

## Affected metrics

All 8 distribution metrics in `StackTelemetry.metrics/1`, including several high-frequency ones:

- `electric.plug.serve_shape.duration` — every shape HTTP request
- `electric.postgres.replication.transaction_received.receive_lag` — every replicated transaction
- `electric.postgres.replication.transaction_received.operations` — every replicated transaction
- `electric.storage.transaction_stored.replication_lag` — every stored transaction
- `electric.shape_log_collector.transaction.affected_shape_count` — every transaction

Under moderate load (100 shape req/s + 50 txn/s), this adds ~300 ETS rows/second to the leaked table, or ~26M rows/day.

Counter, sum, and last_value metrics are unaffected — they use a `:set` table keyed by `{metric_name, labels}` and update in place.

## Proposed fix

**Option A (minimal):** Have the `/metrics` endpoint also scrape all stack-level reporters. This requires discovering the active stack reporter names, e.g. via a registry or by iterating over known stack IDs.

**Option B (remove dead code):** Don't start stack-level Prometheus reporters at all. The application-level reporter already has its own set of metrics, and the stack-level reporter's output is never served to anyone. This is the simplest fix if there is no plan to expose per-stack `/metrics` endpoints.

**Option C (defensive, in the library):** Add a periodic drain or a cardinality/size cap in `telemetry_metrics_prometheus_core` itself, so unreachable reporters don't leak. This is a broader change and wouldn't be Electric-specific.

I'd recommend **Option B** as the immediate fix: remove the `Reporters.Prometheus.child_spec` call from `StackTelemetry.exporter_child_specs/1`. The stack metrics are already exported via the OTEL, StatsD, and call-home reporters — the Prometheus reporter is the only one whose output requires an explicit pull that never happens. If per-stack Prometheus scraping is needed in the future, it can be added together with the corresponding endpoint.

alco:
## Summary

- Remove the stack-level Prometheus reporter from `StackTelemetry` that was never scraped, causing unbounded ETS table growth (~300 rows/sec under moderate load)
- The `/metrics` endpoint only scrapes the application-level reporter; stack metrics continue to be exported via OTEL, StatsD, and call-home reporters

Closes #4123

## Test plan

- [x] Verify the sync service starts without errors
- [x] Verify `/metrics` endpoint still returns application-level metrics
- [x] Confirm no `stack_prometheus_telemetry_*` ETS tables are created

### DIFF
--- a/.changeset/fix-stack-prometheus-ets-leak.md
+++ b/.changeset/fix-stack-prometheus-ets-leak.md
@@ -1,0 +1,5 @@
+---
+'@core/electric-telemetry': patch
+---
+
+Remove unused stack-level Prometheus reporter to fix unbounded ETS table growth
--- a/packages/electric-telemetry/lib/electric/telemetry/stack_telemetry.ex
+++ b/packages/electric-telemetry/lib/electric/telemetry/stack_telemetry.ex
@@ -63,10 +63,6 @@ defmodule ElectricTelemetry.StackTelemetry do
         name: :"stack_otel_telemetry_#{stack_id}",
         metrics: metrics
       ),
-      Reporters.Prometheus.child_spec(opts,
-        name: :"stack_prometheus_telemetry_#{stack_id}",
-        metrics: metrics
-      ),
       Reporters.Statsd.child_spec(opts,
         metrics: metrics ++ Reporters.Statsd.router_dispatch_metrics()
       )


robacourt (approved):
Goot spot!

PR #4125, opened 2026-04-14
feat: split shape count by indexability

icehaunter:
## Summary

Add sync-service telemetry for indexed vs unindexed shape counts so we can correlate stack behavior with how shapes are routed in `ShapeLogCollector`.

## Motivation

Some stacks have very large shape cardinality, and total shape count alone does not explain routing cost well enough. Splitting shapes by indexability makes it easier to understand when work is dominated by indexed routing versus `other_shapes` scans.

## Implementation

- emit `electric.shapes.indexed_shapes.count`
- emit `electric.shapes.unindexed_shapes.count`
- classify shapes using the same routing rules as the filter layer
- maintain cached counts in `ShapeStatus` and emit all three shape metrics (`total`, `indexed`, `unindexed`) from one O(1) snapshot rather than scanning shapes during each telemetry tick

## Verification

- added filter classification tests for indexed vs unindexed cases
- added shape-status tests for maintained counts across add/remove
- added telemetry test covering the emitted `total/indexed/unindexed/active` metrics

No linked issue; this change is operational telemetry to improve observability for large stacks.

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/shape_status.ex
+++ b/packages/sync-service/lib/electric/shape_cache/shape_status.ex
@@ -14,6 +14,7 @@ defmodule Electric.ShapeCache.ShapeStatus do
   to access the data in the ETS from anywhere, so there's an internal api,
   using the full state and an external api using just the table name.
   """
+  alias Electric.Shapes.Filter
   alias Electric.Shapes.Shape
   alias Electric.ShapeCache.ShapeStatus.ShapeDb
   alias Electric.Telemetry.OpenTelemetry
@@ -25,12 +26,18 @@ defmodule Electric.ShapeCache.ShapeStatus do
 
   @type stack_id() :: Electric.stack_id()
   @type shape_handle() :: Electric.shape_handle()
+  @type shape_counts() :: %{
+          total: non_neg_integer(),
+          indexed: non_neg_integer(),
+          unindexed: non_neg_integer()
+        }
 
   # MUST be updated when Shape.comparable/1 changes.
   @version 8
 
   # Tuple format: {handle, hash, snapshot_started, last_read_time, generation}
   @shape_last_used_time_pos 4
+  @shape_counts_key :counts
 
   @spec version() :: pos_integer()
   def version, do: @version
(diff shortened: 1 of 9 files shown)


packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -393,4 +452,74 @@ defmodule Electric.ShapeCache.ShapeStatus do
 
     :ok
   end
+
+  defp rebuild_shape_routing_state(stack_id) do
```
alco, line 447: What is the performance of going through all shapes in shapedb? Is it trivial enough for, say, 100k shapes, such that we don't need to add a tracing span around it for visibility?

packages/sync-service/lib/electric/shape_cache/shape_status.ex
```diff
@@ -393,4 +452,74 @@ defmodule Electric.ShapeCache.ShapeStatus do
...
+  defp decrement_shape_counts(stack_id, indexed?) do
+    :ets.update_counter(
+      shape_counts_table(stack_id),
+      @shape_counts_key,
+      [{2, -1}, {3, if(indexed?, do: -1, else: 0)}, {4, if(indexed?, do: 0, else: -1)}],
+      {@shape_counts_key, 0, 0, 0}
+    )
+  end
+
+  defp take_shape_indexability(stack_id, shape_handle) do
```
alco: Maybe a better name would be `lookup_if_shape_indexed?` or `shape_cached_as_indexed?`?

packages/sync-service/lib/electric/stack_supervisor/telemetry.ex
```diff
@@ -31,12 +31,25 @@ defmodule Electric.StackSupervisor.Telemetry do
...
+    with %{total: num_shapes, indexed: indexed_shapes, unindexed: unindexed_shapes} <-
+           Electric.ShapeCache.shape_counts(stack_id) do
       Electric.Telemetry.OpenTelemetry.execute(
         [:electric, :shapes, :total_shapes],
         %{count: num_shapes},
         %{stack_id: stack_id}
       )
+
+      Electric.Telemetry.OpenTelemetry.execute(
+        [:electric, :shapes, :indexed_shapes],
```
alco: Why not add these as `electric.shapes.total_shapes.count_indexed` etc.? I know the naming is a bit weird if we preserve the current `electric.shapes.total_shapes.count` unchanged. But having them as separate metrics isn't as compelling IMO.

alco (approved):

packages/sync-service/lib/electric/stack_supervisor/telemetry.ex
icehaunter: Yeah, great point, I missed that

PR #4126, opened 2026-04-14
fix(sync-service): catch write pool timeout in ShapeCache to prevent cascading failures

msfstef:
## Problem

We observed a production failure where the ShapeCache GenServer terminates due to a `NimblePool.checkout` timeout on the single SQLite write pool:

```
GenServer {:"Electric.ProcessRegistry:...", {Electric.ShapeCache, nil}} terminating
** (stop) exited in: NimblePool.checkout({:via, ..., {"stack_id", :write}})
    ** (EXIT) time out
```

This cascades into repeated failures for all `create_or_wait_shape_handle` calls, and after ~1 minute of restarts, produces `Stack not ready after 5000ms` errors.

### Root cause

The single-worker write pool (`pool_size: 1`) is shared between:
- **WriteBuffer** — flushes batched operations every 100ms via `checkout_write!`
- **`handle_for_shape_critical`** — called by `ShapeCache.maybe_create_shape` for every new shape creation

When the WriteBuffer holds the write connection (flushing a large batch or hitting slow I/O), `handle_for_shape_critical`'s `checkout_write!` times out after 10s. This exit is unhandled inside the ShapeCache GenServer's `handle_call`, killing it. Since Shapes.Supervisor uses `strategy: :one_for_all`, this restarts all shape consumers, the log collector, and the publication manager. The supervisor canary resets `supervisor_processes_ready` in the StatusMonitor, and if the stack can't re-satisfy all 8 readiness conditions within 5s, clients see "Stack not ready".

The restart itself worsens things: all shape consumers must re-create, generating more `create_or_wait_shape_handle` calls that contend on the same write pool — a feedback loop that lasts until the backlog clears.

### A. Delayed ETS cleanup after WriteBuffer flush

The `handle_for_shape_critical` path exists because of a race window: after WriteBuffer flushes a shape to SQLite and immediately deletes the ETS entry, a read-pool connection may not yet see the committed write (WAL snapshot isolation). An [earlier experiment](https://github.com/electric-sql/electric/commit/9cdc54b85) on `magnetised/lumxxlslxwzp` kept ETS entries for 60s after flush, eliminating the need for the write connection lookup entirely.

This approach works but relies on an arbitrary time-based retention window rather than a hard consistency guarantee. It was deprioritized in favor of the more conservative fixes in #3823. Worth revisiting if write pool contention continues to be a problem, but not pursued here.

### B. Catch the exit in ShapeCache (this PR)

Prevent the cascading failure by catching the exit in `maybe_create_shape` and returning a retryable error to the caller. The ShapeCache GenServer stays alive, and the API returns 503 with `retry_after: 1`.

## What this PR does

- **`shape_cache.ex`**: Wraps `maybe_create_shape` in `safe_maybe_create_shape` that catches `:exit` (NimblePool timeout, process crashes) and `:throw` (dependency creation failures). Refactors `maybe_create_shape` to return explicit `{:ok, ...} | {:error, ...}` tuples — the previous `{:ok, _pid} = start_shape(...)` match would produce a `MatchError` (a raised exception, not an exit) if the consumer failed to start, bypassing the catch entirely.

- **`shapes/api.ex`**: Adds `handle_shape_info({:error, _reason}, request)` clause returning 503 with a generic message (`"Failed to create shape, please retry"`), no internal details leaked.

- **`shape_db/supervisor.ex`** + **`shape_db.ex`**: Makes the write checkout timeout configurable via `shape_db_opts` (stored in StackConfig, read by `handle_for_shape_critical`). Production default remains 10s. This allows tests to use a short timeout (200ms) instead of waiting 10s.

- **`shape_cache_test.exs`**: Reproduction test that holds the write pool, attempts shape creation, and verifies the GenServer survives and returns an error.

## Scope and future work

The `catch :exit` is intentionally broad — it catches all exit reasons, not just timeouts. We could tighten this to only catch `:timeout` exits from NimblePool, but given the severity of the cascading failure (one timeout killing ALL shapes on the stack via `one_for_all` → supervisor restart → StatusMonitor reset → "Stack not ready"), we opted for safety. The GenServer state is pure config (`stack_id`, `feature_flags`) with no risk of corruption from a swallowed exit.

Longer term, the supervision structure itself deserves reconsideration — a single write pool timeout should not be able to bring down the entire shape serving pipeline for a stack. The `one_for_all` strategy on Shapes.Supervisor and `max_restarts: 0` on ShapeDb.Supervisor amplify transient failures into full stack outages.

## Test plan

- [x] Reproduction test: hold write pool → `get_or_create_shape_handle` returns `{:error, {:exit, _}}` → GenServer stays alive (200ms timeout via `shape_db_opts`)
- [x] All existing shape_cache tests pass (730 tests, 0 failures)
- [x] API tests pass (503 error path)

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache.ex
+++ b/packages/sync-service/lib/electric/shape_cache.ex
@@ -339,8 +339,12 @@ defmodule Electric.ShapeCache do
     :exit, reason ->
       {:error, {:exit, reason}}
 
-    :throw, {:error, _reason} = error ->
-      error
+    :error, exception ->
+      Logger.error(
+        "Failed to create shape #{inspect(shape)}: #{Exception.format(:error, exception, __STACKTRACE__)}"
+      )
+
+      {:error, exception}
   end
 
   defp maybe_create_shape(shape, %{stack_id: stack_id} = opts) do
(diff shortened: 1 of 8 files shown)


packages/sync-service/lib/electric/shape_cache.ex
```diff
@@ -328,30 +333,57 @@ defmodule Electric.ShapeCache do
     end
   end
 
+  defp safe_maybe_create_shape(shape, opts) do
+    maybe_create_shape(shape, opts)
+  catch
+    :exit, reason ->
```
msfstef, line 339: nit: I always get confused with elixir try catch - is this enough? does it cover raises? as long as we're covering the case we saw should be good

msfstef (commented):
Looks good @magnetised !

packages/sync-service/lib/electric/shape_cache.ex
magnetised: only catches the case we saw in production. easy to expand to catch exceptions too, don't like catching exceptions in general but i suppose that's better than the kind of stack explosion we're seeing.

so, will expand catch exceptions and log them as an error so we get visibility

magnetised (approved):

PR #4131, opened 2026-04-15
fix: classify admission control requests by actual shape existence

msfstef:
## Summary

> Stacked on #4130

Admission control currently classifies requests as `:initial` (strict limit, default 300) or `:existing` (generous limit, default 10,000) based solely on whether `offset == "-1"`. This is a proxy signal that breaks down in two critical scenarios:

### Problem 1: Restarts and redeploys — stale handles bypass initial limits

After a restart or redeploy with fresh storage, **all connected clients reconnect with handles pointing to shapes that no longer exist**. These requests carry a handle and a non-`-1` offset, so they are classified as `:existing` — even though the 409 redirect flow will create a brand new shape for each one. Under load, this means hundreds or thousands of shape creations happening under the permissive `:existing` bucket, which is exactly the overload scenario admission control exists to prevent.

### Problem 2: Shared shapes — initial requests penalised unnecessarily

When many clients subscribe to the same shape (e.g. a shared table), the first client creates it. Every subsequent client sends `offset=-1` (no handle yet) and gets classified as `:initial` — counted against the strict 300 limit — despite the shape already existing and the request being a cheap lookup. Under a burst of new subscribers this can trigger 503s for requests that would have been trivially served.

## Approach

Add a `resolve_existing_shape` plug between `validate_request` and `check_admission` that checks whether the shape actually exists, then classify based on reality rather than the offset string:

| Case | Check | Cost | Classification |
|---|---|---|---|
| No handle, shape exists | `fetch_handle_by_shape` (ETS + SQLite) | **Replaces** same lookup in `load_shape` — net zero | `:existing` |
| No handle, shape doesn't exist | `fetch_handle_by_shape` → `:error` | One read on the creation path (already expensive) | `:initial` |
| Handle present, shape alive | `fetch_handle_by_shape` finds shape | Same lookup, classifies accurately | `:existing` |
| Handle present, shape dead | `fetch_handle_by_shape` → `:error` | One read, correctly gates creation | `:initial` |

### Why the cost is worth it

- **No-handle case saves work**: `fetch_handle_by_shape` in `resolve_existing_shape` replaces the identical call inside `ShapeCache.get_or_create_shape_handle` that `load_shape` would do. The handle is set on the request params, so `load_shape` takes the cheap `resolve_shape_handle` path instead.
- **Handle-present case classifies accurately**: The same `fetch_handle_by_shape` call checks shape existence by definition. The alternative — not checking — means a restart/redeploy sends all reconnecting clients through the `:existing` bucket while they all trigger shape creation.
- **Handle is kept for dead shapes**: We don't clear the handle when the shape is dead — `load_shape` still needs it for the 409 redirect flow. We only use the existence check for admission classification.

## Admission release on error — approaches tried

Because `check_admission` now runs before `load_shape` (it needs to know whether the shape exists, which `resolve_existing_shape` determines up-front), any exception in `load_shape` or `serve_shape_response` must still release the permit that was acquired earlier. We iterated through three designs to get this right:

### Attempt 1 — `register_before_send` callback on the conn

`check_admission` attached a `register_before_send` callback that released the permit before `send_resp`. This fails for two reasons:

1. **Streaming**: `before_send` fires when `send_chunked` *starts* streaming, not when it finishes. The permit would be returned while chunks were still in flight, and if the ErrorHandler also ended the telemetry span via `before_send`, we'd lose `streaming_bytes_sent` and `duration` metrics.
2. **Error path**: `Plug.ErrorHandler` captures the *original* conn passed into `call/2`, not the accumulated conn. So the before_send callback attached by `check_admission` was simply absent from the conn seen by `handle_errors`, meaning exceptions in `load_shape` leaked the permit entirely.

### Attempt 2 — outer/inner `Plug.Builder` split (commit 6971b8a4e, now reverted)

Following José Valim's suggestion in plug#486: split into an outer pipeline (`ServeShapePlug`) and inner pipeline (`ServeShapePlug.Inner`). The inner module's `Plug.ErrorHandler` sees the conn with full outer state — including `conn.private[:admission_kind]` and the before_send callback. Exceptions in `load_shape` now released correctly.

Costs of this approach:
- Two modules, with five `@doc false` public callbacks so the inner module can delegate back.
- Custom `halt/1` override (can't use before_send for telemetry span end because of the streaming problem in Attempt 1 — so `halt` had to end the span explicitly).
- Four-branch `admission_kind/1` fallback chain including an offset-based last-resort for the outer ErrorHandler where neither `check_admission` nor `resolve_existing_shape` had run.
- Duplicated `@impl Plug.ErrorHandler` in both modules.

### Attempt 3 — `try/catch` + `try/after` in `call/2` (current, credit to @alco)

[@alco](https://github.com/alco) [proposed](https://github.com/electric-sql/electric/commit/be1008b1451d6f8f0cf3ba808d160de32a064a2f) collapsing back to a single module by overriding `call/2` directly:

```elixir
def call(conn, opts) do
  conn = conn |> assign(:config, opts) |> fetch_query_params() |> start_telemetry_span()

  try do
    conn =
      try do
        super(conn, opts)
      catch
        kind, reason -> handle_caught(conn, kind, reason, __STACKTRACE__)
      end

    emit_shape_telemetry(conn)
    conn
  after
    OpentelemetryTelemetry.end_telemetry_span(OpenTelemetry, %{})
    release_admission_permit()
  end
end
```

`check_admission` stashes the acquired permit in the process dictionary; the `after` clause releases it unconditionally on every exit path (success, halt, uncaught exception). The OTEL span is popped in the same `after` block.

**Why this is strictly better than Attempt 2:**
- Fewer moving parts: single module, no cross-module callbacks, no `halt/1` override, `admission_kind/1` collapses to two branches (`:shape_exists` true/false).
- Streaming is correct by construction: `super(conn, opts)` only returns once `Api.Response.send_stream/2` has synchronously drained the body, so `emit_shape_telemetry` captures the accurate `streaming_bytes_sent` and `duration` *before* the span is popped.
- Every defensive fallback we'd accumulated in Attempt 2 existed because `Plug.ErrorHandler` sees the wrong conn — Attempt 3 removes the root cause, so the fallbacks can go with it.

Net −56 lines versus Attempt 2, and strictly more correct on the streaming path.

## Test plan

- [x] New test: request for non-existing shape with `offset=-1` is classified as `:initial` (rejected when slots full)
- [x] New test: request for existing shape with `offset=-1` is classified as `:existing` (succeeds when initial slots full)
- [x] New test: request with dead handle is classified as `:initial` (rejected when initial slots full)
- [x] New test: permit released when `load_shape` raises `RuntimeError`
- [x] New test: permit released when `load_shape` raises `DBConnection.ConnectionError`
- [x] New test: correct `:existing` permit released when shape exists and offset is `-1`
- [x] New test: correct `:initial` permit released when shape does not exist
- [x] New test: release not called when exception occurs before `check_admission` runs
- [x] Full router test suite passes (87 tests)
- [x] Full serve_shape_plug test suite passes (40 tests)
- [x] Full admission-control test suite passes

### DIFF
--- a/.changeset/admission-control-check-shape-existence.md
+++ b/.changeset/admission-control-check-shape-existence.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Classify admission control requests by actual shape existence instead of offset value. Prevents shape creation storms after restarts/redeploys from bypassing initial request limits, and avoids penalising reconnecting clients to shared shapes.
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -1,51 +1,31 @@
-# Inner module is defined first because Plug.Builder resolves
-# `Electric.Plug.ServeShapePlug.Inner.init/1` at compile time.
-defmodule Electric.Plug.ServeShapePlug.Inner do
-  @moduledoc false
-  use Plug.Builder
-  use Plug.ErrorHandler
-
-  alias Electric.Plug.ServeShapePlug
-
-  plug :load_shape
-  plug :serve_shape_response
-
-  # end_telemetry_span needs to always be the last plug here.
-  plug :end_telemetry_span
-
-  defp load_shape(conn, opts), do: ServeShapePlug.do_load_shape(conn, opts)
-  defp serve_shape_response(conn, opts), do: ServeShapePlug.do_serve_shape_response(conn, opts)
-  defp end_telemetry_span(conn, opts), do: ServeShapePlug.do_end_telemetry_span(conn, opts)
-
-  # Inner ErrorHandler: the conn here carries the full outer pipeline state
-  # (shape_exists, admission_kind, before_send callbacks), so admission control
-  # release happens correctly via the before_send callback when send_resp fires.
-  @impl Plug.ErrorHandler
-  def handle_errors(conn, error), do: ServeShapePlug.do_handle_errors(conn, error)
-end
-
 defmodule Electric.Plug.ServeShapePlug do
   @moduledoc """
   Plug pipeline for serving shape requests.
 
-  Split into an outer pipeline (this module) and an inner pipeline (`Inner`)
-  following Jose Valim's recommended pattern (Plug issue #486). The outer
-  pipeline handles request setup, validation, and admission control. The inner
-  pipeline handles shape loading and response streaming.
+  `call/2` is overridden to wrap the whole pipeline in a telemetry span, an
+  explicit `try`/`catch` that invokes `handle_errors/2` on uncaught errors,
+  and a `try`/`after` that always releases the admission control permit and
+  ends the OTEL span.
 
-  This split ensures that when an exception occurs in the inner pipeline,
-  `Plug.ErrorHandler` receives the conn with all outer state intact —
-  including `conn.private[:admission_kind]` and the `register_before_send`
-  callback from `check_admission`. The outer ErrorHandler acts as a safety net
-  for exceptions in setup plugs; Plug skips it when the inner handler already
-  sent a response (via the `{:plug_conn, :sent}` mailbox check).
+  Error handling is inspired by `Plug.ErrorHandler`'s pattern but adapts to a
+  subtle limitation: `Plug.Builder` does not wrap individual plugs in
+  error-capturing frames, so the `conn` visible from the `catch` clause is
+  the one passed into `call/2` — not the accumulated conn at the point of
+  raise. This means we cannot rely on a `register_before_send` callback on
+  the error-handler's conn for admission release. Instead, `check_admission`
+  stashes the acquired permit in the process dictionary and the `after`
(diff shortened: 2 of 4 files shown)


msfstef:
Re: review suggestions

**Stale data from `fetch_handle_by_shape`**: Intentionally using the non-critical read variant. Under high write contention, a brief miss classifies the request as `:initial` — which is the correct conservative behavior when the system is under load. The non-critical variant's semantics are already documented in `shape_status.ex`.

**`Process.sleep(100)` after `clean_shape`**: Pre-existing pattern throughout the test file — not introduced here.

msfstef:
Claude's review on correctness is absolutely right - this reintroduces counters that never go down. Looking into it, double release might be a good idea here for a simple "fix".

alco (commented):
The approach looks good. Waiting on the update for the error case before taking another look.

msfstef:
@alco I have pushed a split plug approach

alco:
@codex review

robacourt (approved):
LGTM

PR #4139, opened 2026-04-20
Change form event type: React.FormEvent to React.SubmitEvent

Jaime02:
```
FormEvent is deprecated.
interface React.FormEvent<T = Element>
@deprecated
FormEvent doesn't actually exist. You probably meant to use ChangeEvent, InputEvent, SubmitEvent, or just SyntheticEvent instead depending on the event type.
```


icehaunter (approved):

icehaunter:
Thanks for the contribution!

PR #4142, opened 2026-04-21
Guard report_shape_db_stats against :noproc during deployments

alco:
## Summary

`Electric.StackSupervisor.Telemetry.report_shape_db_stats/2` calls `ShapeDb.statistics/1`, which resolves the GenServer pid and then issues a `GenServer.call`. During stack restarts the underlying server can exit between those two steps, producing an uncaught `:exit, :noproc` from the telemetry poller and flooding Sentry with errors on every deployment.

The sibling function `report_retained_wal_size` already handles this race. This change mirrors its `try/catch :exit, {:noproc, _}` pattern. Surgical, no behavioural change on the happy path.

Refs electric-sql/stratovolt#1452.

## Test plan

- [x] `mix compile --warnings-as-errors` is clean
- [x] `mix test test/electric/stack_supervisor_test.exs` passes

### DIFF
--- a/.changeset/fix-report-shape-db-stats-noproc.md
+++ b/.changeset/fix-report-shape-db-stats-noproc.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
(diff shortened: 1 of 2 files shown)


alco:
Covered by https://github.com/electric-sql/electric/pull/4148.

PR #4144, opened 2026-04-21
feat(telemetry): per-shape response_size.bytes histogram

alco:
## Summary

Adds a per-shape response size histogram so operators can attribute HTTP response payload volume to individual shapes.

- Emits a new `[:electric, :shape, :response_size]` telemetry event from `ServeShapePlug.end_telemetry_span/_`, tagged with `root_table`, `is_live` and `stack_id`.
- Registers it as a `Telemetry.Metrics.distribution` (`electric.shape.response_size.bytes`, unit `:byte`), which the existing `OtelMetricExporter` reporter publishes as an OTel histogram.
- Bytes are pulled from `conn.assigns[:streaming_bytes_sent]`, the same value already fed into the existing `[:electric, :plug, :serve_shape]` event; no new bookkeeping on the hot path.

Motivated by operational visibility work — without this, there is no way to tell which shape is responsible for a request-handler binary-memory spike.

Refs [stratovolt#1466](https://github.com/electric-sql/stratovolt/issues/1466).

## Why a new event rather than re-tagging the existing one

`[:electric, :plug, :serve_shape]` already feeds a latency distribution (with a `keep` filter dropping live requests) and span attributes. Layering high-cardinality shape tags onto it would risk affecting those consumers. A dedicated event keeps the intent and the tagging surface isolated.

## Test plan

- [x] `mix compile` in `packages/sync-service` and `packages/electric-telemetry`
- [x] `mix test test/electric/plug/serve_shape_plug_test.exs` in `packages/sync-service`
- [x] `mix test` in `packages/electric-telemetry`
- [ ] Manual check in a stack that the histogram shows up on the OTel exporter once a request has been served

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -291,11 +291,14 @@ defmodule Electric.Plug.ServeShapePlug do
   # is the place to assign them because we keep this plug last in the "plug pipeline" defined
   # in this module.
   defp end_telemetry_span(%Conn{assigns: assigns} = conn, _ \\ nil) do
+    stack_id = get_in(conn.assigns, [:config, :stack_id])
+    bytes_sent = assigns[:streaming_bytes_sent] || 0
+
     OpenTelemetry.execute(
       [:electric, :plug, :serve_shape],
       %{
         count: 1,
-        bytes: assigns[:streaming_bytes_sent] || 0,
+        bytes: bytes_sent,
         monotonic_time: System.monotonic_time(),
         duration: System.monotonic_time() - conn.private[:electric_telemetry_span][:start_time]
       },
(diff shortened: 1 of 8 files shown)


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -321,6 +338,17 @@ defmodule Electric.Plug.ServeShapePlug do
...
+  # Used as a metric label for the response-size histogram. The goal is low
+  # cardinality (one value per configured shape root table), so we only use
+  # the already-validated `table` value from request params. Unvalidated input
+  # (e.g. raw `conn.query_params["table"]`) would let an attacker create
+  # unbounded time-series cardinality by sending arbitrary table names, so we
+  # fall back to a fixed `"unknown"` label instead.
+  defp get_root_table(%{request: %{params: %{table: table}}}) when is_binary(table),
+    do: table
+
+  defp get_root_table(_assigns), do: "unknown"
```
alco: Use `nil` instead.

robacourt (approved):
LGTM!

PR #4149, opened 2026-04-22
Propagate OTel context into spawned snapshot and move-in tasks

alco:
vetted by human
ready for review

## Summary

Fixes a bug where telemetry spans defined with `OpenTelemetry.with_child_span` inside spawned tasks (initial-snapshot and move-in code paths) were silently dropped, hiding expected fine-grained spans such as `shape_snapshot.execute_for_shape`, `shape_snapshot.query_fn`, `shape_snapshot.checkout_wait`, `shape_snapshot.setup`, and `shape_snapshot.query` from Honeycomb on hosts whose traffic is dominated by initial snapshots.

### Root cause

`with_child_span/4` only creates a span when there is already a parent span in the **current Erlang process's** OTel context. `Task.Supervisor.async_nolink` / `Task.Supervisor.start_child` start new processes that do not inherit the caller's OTel context, so `in_span_context?()` returns `false` and the whole span subtree is dropped. Three spawn sites were affected:

- `Electric.Shapes.Consumer.Snapshotter.start_streaming_snapshot_from_db/4`
- `Electric.Shapes.PartialModes.query_move_in_async/5`
- `Electric.Shapes.PartialModes.query_move_in/5`

(`PartialModes.query_subset/4` is called synchronously from an HTTP-request process that already has a parent span — it was not affected.)

### Fix

Capture the context via `:otel_ctx.get_current()` before each spawn and attach it inside the task closure with `:otel_ctx.attach/1` (detached in `after`). This mirrors the pattern already used for `state.otel_ctx` in `Snapshotter.handle_continue/2`.

## Test plan

- [x] `mix compile` clean
- [x] `mix test test/electric/shapes/consumer_test.exs` — 29 passing
- [x] `mix test test/electric/shapes/consumer/move_ins_test.exs test/electric/shapes/consumer/initial_snapshot_test.exs` — 60 passing
- [ ] After deploy: verify `name = shape_snapshot.execute_for_shape AND shape.query_reason = "initial_snapshot"` returns rows in Honeycomb (previously 0 across all hosts over 24h)

Refs: https://github.com/electric-sql/alco-agent-tasks/issues/27

### DIFF
--- a/.changeset/propagate-otel-context-to-snapshot-tasks.md
+++ b/.changeset/propagate-otel-context-to-snapshot-tasks.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Propagate OpenTelemetry context into spawned snapshot and move-in tasks so that spans created via `with_child_span` (e.g. `shape_snapshot.execute_for_shape`, `shape_snapshot.query_fn`, `shape_snapshot.checkout_wait`) are linked to the originating trace instead of being silently dropped.
--- a/integration-tests/tests/otel-export.lux
+++ b/integration-tests/tests/otel-export.lux
@@ -209,15 +209,37 @@
   ?electric-offset: ([\w\d_]+)
   [global offset=$1]
 
+[shell otel_collector]
+  # Verify that the initial snapshot query span is exported from the Snapshotter path.
+  ?Span #[0-9]+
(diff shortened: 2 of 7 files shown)


robacourt (approved):
LGTM!

PR #4162, opened 2026-04-24
fix: prevent duplicate CORS headers when proxying upstream responses

balegas:
## Summary
- When the agents-server proxies requests to the streams server or Electric sync service, both the agents-server (`setHeader`) and the upstream send CORS headers, resulting in duplicate `Access-Control-Allow-Origin` values that browsers reject
- This breaks the agents-server-ui when running on a different port (e.g. Vite dev server on `:5173` hitting agents-server on `:4437`)
- Added `replaceOverlappingHeaders()` that removes any `setHeader`-set headers before `writeHead` when the upstream response provides the same header — upstream CORS headers replace (not duplicate) the defaults
- Widened `access-control-allow-headers` to include `if-none-match` and `traceparent` (supported by the Elixir sync service)

## Test plan
- [ ] Run agents-server-ui via Vite dev server on a different port and verify CORS requests succeed
- [ ] Verify proxied shape stream requests don't have duplicate `Access-Control-Allow-Origin` headers
- [ ] Verify non-proxied routes (health, UI assets) still return default CORS headers


balegas:
Not the right fix

PR #4182, opened 2026-04-27
fix(ci): add push trigger to rebuild agents-server Docker image on main

balegas:
## Summary

The `electricax/agents-server` Docker image was never rebuilt automatically. Unlike the sync-service workflow which rebuilds on every push to main, the agents-server workflow only had `workflow_call` and `workflow_dispatch` triggers.

The `workflow_call` path (from changesets release) also never triggered because `changeset tag` is chained after hex publishing via `&&` in `ci:publish` — when hex failed ([run #4174](https://github.com/electric-sql/electric/actions/runs/24899369017), [run #4168](https://github.com/electric-sql/electric/actions/runs/24898153739)), tags were never created, so `publishedPackages` was empty.

This meant features like the state explorer (#4156) were merged and published to npm but never included in the Docker image.

## Fix

Add a `push` trigger with path filtering to `agent_server_dockerhub_image.yml`, matching the pattern used by the sync-service workflow:

- Pushes to main touching `packages/agents-server/**`, `packages/agents-server-ui/**`, or `packages/agents-runtime/**` → **canary** image
- Changesets `workflow_call` → tagged **release** image (`:latest` + version)
- `workflow_dispatch` → **custom** image

## Test plan

- [ ] Merge this PR, then push a change touching any of the three agent packages
- [ ] Confirm `electricax/agents-server:canary` is published to Docker Hub
- [ ] Verify the canary image contains the state explorer from #4156

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "packageManager": "pnpm@10.12.1",
   "private": true,
   "scripts": {
-    "ci:publish": "pnpm '/^ci:publish:.+/' && pnpm exec changeset tag",
+    "ci:publish": "pnpm exec changeset tag && pnpm '/^ci:publish:.+/'",
     "ci:publish:hex-electric": "pnpm run --dir packages/sync-service publish:hex",
     "ci:publish:hex-electric-client": "pnpm run --dir packages/elixir-client publish:hex",
     "ci:publish:npm": "pnpm publish -r --filter './packages/**' --no-git-checks --access public",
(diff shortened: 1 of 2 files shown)


icehaunter (approved):

package.json
```diff
@@ -8,7 +8,7 @@
   "packageManager": "pnpm@10.12.1",
   "private": true,
   "scripts": {
-    "ci:publish": "pnpm '/^ci:publish:.+/' && pnpm exec changeset tag",
+    "ci:publish": "pnpm exec changeset tag && pnpm '/^ci:publish:.+/'",
```
icehaunter: I think the reason tagging was done after the publish was because if publishing fails, you cannot rerunthis job now because duplicate tagging will always fail

PR #4183, opened 2026-04-27
Disable compaction with subqueries

robacourt:
Compaction is not supported with subqueries, so with this PR a 400 is returned if users try to use both

### DIFF
--- a/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
+++ b/packages/sync-service/lib/electric/plug/serve_shape_plug.ex
@@ -42,8 +42,8 @@ defmodule Electric.Plug.ServeShapePlug do
   plug :put_resp_content_type, "application/json"
   plug :parse_body
   plug :validate_request
+  plug :reject_subquery_shape_compaction_request
   plug :resolve_existing_shape
-  plug :reject_unsupported_shape_creation
   plug :check_admission
   plug :load_shape
   plug :serve_shape_response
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -272,9 +274,35 @@ defmodule Electric.Plug.ServeShapePlug do
...
-    ArgumentError -> put_private(conn, :shape_exists?, false)
+    ArgumentError -> put_private(conn, :shape_exists?, nil)
   end
 
+  defp reject_unsupported_shape_creation(
+         %Conn{assigns: %{request: request}} = conn,
+         _
+       ) do
+    if conn.private[:shape_exists?] != false or
+         not unsupported_shape_creation?(request.params.shape_definition) do
```
alco: I have two thoughts about this:
1. We shouldn't take shape existence into account. Because even if it works at first, as soon as the shape churns with a 409, it won't be possible to recreate it. I think it's easier to understand the rule that when subqueries are enabled in Electric's config, any request containing the compacting parameter gets rejected with a 400 response.
2. The repeated negation is difficult to parse: `not *un*supported`. The negation in the new private function seems gratuitous? We could instead be more direct and extract the `    compaction_enabled? = fetch_field!(changeset, @tmp_compaction_flag)` line from API.Params into a helper function of the same name. Then, in this module we'd just call that and leave a comment in the code explaining that compaction is not compatible with subquery shapes.

robacourt: Good spot. Fixed both.

icehaunter (approved):

PR #4190, opened 2026-04-27
Add coder entity + let Horton spawn and converse with coders

kevin-dp:
Adds a `coder` entity — a Claude Code / Codex CLI session wrapped as a long-lived Electric Agents entity — and lets Horton spawn and keep prompting one across many turns.

## What's new

- **`coder` entity** (`@electric-ax/agents`). Drives `claude` / `codex` CLI under a fixed cwd, mirrors normalized JSONL events into the entity's `events` collection, persists a session cursor across CLI invocations. Three creation flows: create-fresh, attach to an existing on-disk session, or import (cross-agent or same-agent).
- **`spawn_coder` / `prompt_coder` Horton tools.** `spawn_coder` creates a new coder and ends Horton's turn; `prompt_coder` queues a follow-up on an existing coder. Horton wakes via `runFinished` on every coder turn with the coder's reply text in the wake payload, so it can relay it without inspecting the filesystem.
- **`ctx.recordRun()` runtime primitive** (`@electric-ax/agents-runtime`). Lets non-LLM entities (CLI wrappers, HTTP callers, etc.) bracket external operations with the same `runs`-collection events `useAgent` writes internally — satisfying the `runFinished` wake matcher and surfacing a response payload via `RunHandle.attachResponse(text)`. The coder is the first consumer.
- **Coder UI in the admin app** (`@electric-ax/agents-server-ui`). 3-tab Create / Attach / Import spawn dialog routed from the sidebar's *New session* flow, timeline that surfaces queued user prompts immediately as a "queued" bubble (paired against canonical `user_message` events by text so they swap cleanly when the CLI mirrors the JSONL back), full `nativeSessionId` in the session header for copy/paste against on-disk session files.

## Notable details

- Codex `exec` is invoked with `--skip-git-repo-check` so the coder can run in non-git, non-trusted directories.
- Claude session discovery falls back to scanning `~/.claude/projects/<sanitized-cwd>/*.jsonl` directly so non-interactive `claude -p` runs (which don't drop a `~/.claude/sessions/<pid>.json` lock file) are still findable. The complementary upstream fix in `agent-session-protocol` is in https://github.com/durable-streams/durable-streams/pull/340 — independent.
- The first prompt's text becomes the entity's display title (truncated) so the sidebar surfaces something meaningful instead of a random slug.

## Test plan

- [x] All TS unit tests green: `agents-runtime` 375/375, `agents` 42/42, `electric-ax` 61/61.
- [x] Verified end-to-end: spawn coder → send prompts → reply lands; Horton spawns a coder, wakes on `runFinished`, calls `prompt_coder` for follow-ups.


kevin-dp:
Thanks for the careful review. Addressed five of these in separate commits on top of the branch:

- **Orphan `tool_result` dead branch** (`6658ae529`). Built a separate `callsByCallId` map from `tool_call` events so the orphan check actually queries the right collection. The `OrphanToolResult` row now renders when a result lands without a matching call in the visible window (e.g. mid-session cursor attach).
- **Duplicate types in `useCodingSession.ts`** (`3b76abfc6`). The hook now imports `CodingSessionEventRow` / `CodingSessionStatus` from `@electric-ax/agents-runtime` and re-exports them, so the local copies can't drift from the runtime contract.
- **`CODING_SESSION_TYPE` literal duplicated** (`40d0921ee`). `Sidebar.tsx` and `router.tsx` now use `CODING_SESSION_ENTITY_TYPE` from `@electric-ax/agents-runtime` directly.
- **Unbounded stdout/stderr buffer** (`1550d3ea8`). Capped both at 4096 chars at the receive site — keeps the diagnostic budget without holding megabytes of CLI trace on the heap.
- **`void discoveredNativeId`** (`f9ef611b1`). The variable was dead — its only consumer was the `void` line silencing the unused-local lint. The native id is already written into `runningMeta.nativeSessionId` + `sessionMeta` in the same branch, so removed the local and the `void`.

Leaving these as-is for now:

- **`--dangerously-skip-permissions` + user-controlled `cwd`** — intentional for the "trust the agent server operator" model. Worth documenting separately when we firm up the threat model around `defaultWorkingDirectory`.
- **Wildcard CORS** — fine for the current "demo / single-tenant operator" deployment shape. Will tighten when we have a multi-tenant story.
- **Global `window.fetch` ngrok patch** — cosmetic; the header is ignored on non-ngrok hosts.
- **djb2 collision window** — ~1-in-4B for events with the same `ts` and `type`; the key is only used for dedup so a collision just means a silent skip of an already-near-identical event, not data loss.

kevin-dp:
Thanks for the second pass. Addressed all four "Important" items in separate commits on top of the branch:

- **Type divergence on `CodingSessionMetaRow.cwd`** (`5311f4c7b`). Exported a canonical `CodingSessionMetaRow` (= `CodingSessionMeta` plus the table primary key) from `@electric-ax/agents-runtime` and dropped the local copy in the UI hook. The hook now re-exports the runtime type so existing consumers still resolve. The `meta.cwd && (...)` render guards already handle the optional case.
- **Hardcoded collection-type strings in the hook** (`300ccdce6`). Added `CODING_SESSION_META_COLLECTION_TYPE` / `CODING_SESSION_CURSOR_COLLECTION_TYPE` / `CODING_SESSION_EVENT_COLLECTION_TYPE` next to `CODING_SESSION_ENTITY_TYPE` in `@electric-ax/agents-runtime`. Both the entity definition (`coding-session.ts`) and the UI hook now import them — strings live in one place.
- **Re-throw / framework-rewake contract** (`c92d67358`). Added a comment at the catch block making the contract explicit: `lastProcessedInboxKey` is advanced before the throw, so the failed prompt is consumed; the throw lets the entity bridge surface the failure to observers (Horton wakes on `runFinished` with `status=failed`, the UI flips the badge); on the next wake the for-loop resumes from the *next* queued prompt, so remaining inbox messages are deferred not dropped. Kept the re-throw — masking failures by `continue`-ing past them would hide the signal observers depend on.
- **Test coverage for session discovery** (`39feb3fd2`). New `find-new-session-after-run.test.ts` (10 tests). Each test stubs `process.env.HOME` to a per-test temp directory and drops stub `.jsonl` files at controlled mtimes, covering: empty/missing dir → null, single new jsonl → its sessionId, multiple new jsonls → newest by mtime, pre-existing ids filtered out, `getClaudeProjectDirs` adds the realpath candidate when the cwd is a symlink, `listClaudeJsonlIdsByCwd` ignores non-jsonl files, and codex falls through to `discoverNewestSession`. Marked the four helpers `export` so they're directly testable.

Skipping the two "Nice to Have" items:

- **Spin-wait in `runWithLiveMirror` finally** — agreed it's wasteful in theory, but the loop fires for at most a sub-millisecond drain in the common case. Replacing it with a `busyPromise` moves load-bearing concurrency primitives in code that's been verified end-to-end; not worth the churn here.
- **No linked issue** — convention thing for the maintainer rather than a code change.

icehaunter (approved):

PR #4204, opened 2026-04-28
website: clarify subquery docs and optimized where clause guidance

robacourt:
## Summary
- clarify `allow_subqueries` and `tagged_subqueries` docs, including rollout status and client compatibility for Electric 1.6 incremental compound subquery handling
- update the Shapes guide to treat subqueries as general-use, add a compound boolean example, and explain incremental move handling
- refresh optimized where clause and benchmark docs to cover current optimized patterns and fix cross-links and anchors

## Testing
- `pnpm build` (run from a detached `HEAD` worktree to avoid unrelated untracked local files in the current checkout)

### DIFF
--- a/website/docs/sync/guides/shapes.md
+++ b/website/docs/sync/guides/shapes.md
@@ -184,7 +184,7 @@ const stream = new ShapeStream({
 const shape = new Shape(stream)
 ```
 
-When a shape uses a subquery, Electric tracks the dependency between tables. If the data in the subquery changes (for example, a project becomes archived or a membership row is added), rows will automatically move in or out of the shape without the root row itself being modified.
+When a shape uses a subquery, Electric tracks the dependency between tables. If the data in the subquery changes (e.g., a project becomes archived), rows will automatically move in or out of the shape without the row itself being modified.
 
 Electric 1.6 keeps these moves incremental even for compound expressions that use `AND`, `OR`, and `NOT` around subqueries. In older releases those cases could return `409` and force a full resync of the shape.
 
--- a/website/docs/sync/guides/shapes.md
+++ b/website/docs/sync/guides/shapes.md
@@ -51,7 +51,7 @@ Shapes are defined by:
 A shape contains all of the rows in the table that match the where clause, if provided. If a columns clause is provided, the synced rows will only contain those selected columns.
 
 > [!Warning] Limitations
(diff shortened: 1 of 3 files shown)


website/docs/sync/guides/shapes.md
```diff
@@ -158,10 +185,12 @@ const stream = new ShapeStream({
 const shape = new Shape(stream)
 ```
 
-When a shape uses a subquery, Electric tracks the dependency between tables. If the data in the subquery changes (e.g., a project becomes archived), rows will automatically move in or out of the shape without the row itself being modified.
+When a shape uses a subquery, Electric tracks the dependency between tables. If the data in the subquery changes (for example, a project becomes archived or a membership row is added), rows will automatically move in or out of the shape without the root row itself being modified.
```
balegas: maybe a bit too deep, don't need to mention old algorithm.

website/docs/sync/guides/shapes.md
```diff
@@ -158,10 +185,12 @@ const stream = new ShapeStream({
...
+
+Electric 1.6 keeps these moves incremental even for compound expressions that use `AND`, `OR`, and `NOT` around subqueries. In older releases those cases could return `409` and force a full resync of the shape.
 
-:::warning Experimental feature
-Subqueries require enabling feature flags. Set `ELECTRIC_FEATURE_FLAGS=allow_subqueries,tagged_subqueries` to enable. See the [configuration docs](/docs/sync/api/config#allow_subqueries) for details.
+:::info Preview feature
+Subqueries are currently in preview and are enabled using `ELECTRIC_FEATURE_FLAGS=allow_subqueries,tagged_subqueries`. The flags gate availability rather than a separate syntax or API. See the [configuration docs](/docs/sync/api/config#allow_subqueries) for details.
```
balegas, line 192: Shall we drop the two flags? once we land latests fixes we could consolidate the flags.

balegas (approved):

website/docs/sync/guides/shapes.md
robacourt: I've reverted that bit

website/docs/sync/guides/shapes.md
robacourt: Yeah, I'll do that as a separate PR as currently we still need both

PR #4209, opened 2026-04-28
Add subset telemetry attrs and rows metric

robacourt:
## Summary
- add subset query span attrs for returned row count and result bytes
- record POST body params as telemetry attrs, truncating long strings to 64KB
- surface `electric.subqueries.subset_result.rows` from `ElectricTelemetry.StackTelemetry`
- add focused tests for POST body telemetry flattening/truncation and the new stack metric

## Testing
- `mix test test/electric/plug/utils_test.exs test/electric/plug/subset_telemetry_test.exs`
- `mix test test/electric/telemetry/stack_telemetry_test.exs`

### DIFF
--- a/packages/sync-service/lib/electric/plug/utils.ex
+++ b/packages/sync-service/lib/electric/plug/utils.ex
@@ -83,7 +83,22 @@ defmodule Electric.Plug.Utils do
   end
 
   alias OpenTelemetry.SemConv, as: SC
-  @max_telemetry_string_bytes 64 * 1024
+  @max_telemetry_string_bytes 256
+  @telemetry_body_root_keys ~w(
+    table
+    offset
+    handle
+    live
+    where
+    columns
+    replica
+    params
+    experimental_compaction
+    live_sse
+    experimental_live_sse
+    log
+  )
+  @telemetry_body_subset_keys ~w(where order_by limit offset params)
 
   @doc false
   def redact_query_string(nil), do: nil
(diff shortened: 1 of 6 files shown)


alco:
> record POST body params as telemetry attrs, truncating long strings to 64KB

64KB is a lot, and it's a per-value truncation limit? What's the use case for such strings in telemetry?

Normally the long strings we observe in observability data are log lines and those are generally on the order of 100 bytes.

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -115,10 +121,66 @@ defmodule Electric.Plug.Utils do
...
   end
 
+  defp telemetry_body_params(body_params) when body_params == %{}, do: %{}
+
+  defp telemetry_body_params(body_params) when is_map(body_params),
+    do: flatten_telemetry_params(body_params)
+
+  defp telemetry_body_params(_), do: %{}
+
+  defp flatten_telemetry_params(params, prefix \\ "http.body_param") do
```
alco: This is quite a complex function without any explainer.

What body params are we specifically interested in? Or is this kind of like reimplementing `URI.parse()` but for the case where all params have been POSTed in the request body? Seems odd to have a completely bespoke, convoluted parsing function for what appears to be the case of parsing query params from a different place than the URL string.

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -115,10 +121,66 @@ defmodule Electric.Plug.Utils do
...
+    |> Enum.with_index()
+    |> Enum.reduce(%{}, fn {value, index}, attrs ->
+      Map.merge(attrs, flatten_telemetry_param(value, "#{prefix}.#{index}"))
+    end)
+  end
+
+  defp flatten_telemetry_param(nil, _prefix), do: %{}
+
+  defp flatten_telemetry_param(value, prefix)
+       when is_binary(value) or is_boolean(value) or is_number(value) do
```
alco: This, in combination with ` defp maybe_truncate_telemetry_string(value) when not is_binary(value), do: value` looks like AI slop. We should need to "maybe truncate" a boolean or a number.

robacourt:
> 64KB is a lot, and it's a per-value truncation limit? What's the use case for such strings in telemetry?

The use case for POST over GET was because the URL params were getting too big, body params in a POST doesn't have that limit. So we should expect large values, including large subset where clauses (one of the main things I'm trying to expose in this PR). Honeycomb has a [64KB limit for string values](https://docs.honeycomb.io/api/events/create-an-event#:~:text=String%20Fields%3A%20Each%20string%20field%20has%20a%20maximum%20length%20of%2064KB) so I thought I'd limit it to that, but perhaps we should consider a much lower value.

alco:
@robacourt the first item in the list of Limits on that page reads

> 2,000 fields per event. The entire event must be less than 1 MB of uncompressed JSON.

We're probably nowhere near the worst case but 2K string fields 64KB each would be 125MiB.

I'd say we need to include a large enough prefix that will allow identifying the particular where expression or whatever else it is encoded in that long string. For that, 100 or a few hundred bytes should be sufficient.

It feels wrong to try to fit the whole body in a telemetry event if its total size is tens of KB. We just wouldn't use the whole payload anyway, we'd be looking for patterns.

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -115,10 +122,81 @@ defmodule Electric.Plug.Utils do
...
     |> Map.merge(Map.new(conn.req_headers, fn {k, v} -> {"http.request.header.#{k}", v} end))
     |> Map.merge(Map.new(conn.resp_headers, fn {k, v} -> {"http.response.header.#{k}", v} end))
   end
 
+  # GET requests already expose raw query params in telemetry. For POST requests,
+  # only mirror subset-defining body params so subset filters and pagination remain
+  # visible without trying to serialize the full request body.
+  defp telemetry_body_params(body_params) when body_params == %{}, do: %{}
+
+  defp telemetry_body_params(%{"subset" => subset_params}) when is_map(subset_params) do
```
alco: Is this an old format for including subset params in the POST body? In the current HTTP API spec I don't see the `subset` key being mentiond.

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -115,10 +122,81 @@ defmodule Electric.Plug.Utils do
...
+    |> trim_invalid_utf8()
+  end
+
+  defp trim_invalid_utf8(value) when value == "", do: value
+
+  defp trim_invalid_utf8(value) do
+    if String.valid?(value) do
+      value
+    else
+      trim_invalid_utf8(binary_part(value, 0, byte_size(value) - 1))
```
alco: This is an odd choice. What if the invalid character is the first character of the string?

How about this impl for `trim_invalid_utf8()`:

```
defp trim_invalid_utf8(value) do
  if String.valid?(value) do
    value
  else
    value
    |> String.chunk(:valid)
    |> Enum.filter(&String.valid?/1)
    |> Enum.join()
  end
end
```

There's no need to have a separate clause for the empty string.

packages/sync-service/test/electric/plug/subset_telemetry_test.exs
```diff
@@ -0,0 +1,117 @@
...
+
+      true
+    end)
+
+    conn =
+      conn(
+        "POST",
+        "/v1/shape?table=items&offset=-1&log=changes_only",
+        Jason.encode!(%{
+          "subset" => %{
```
alco: Not sure why body with top-level "subset" key is used throughout these new tests if the [current HTTP API spec](https://electric-sql.com/openapi.html#/paths/~1v1~1shape/post) says body params should be included in the top-level JSON object of the POST body.

packages/sync-service/lib/electric/plug/utils.ex
robacourt: It's not in the docs but it's supported in code https://github.com/electric-sql/electric/blob/4d8e452c80caa6ee31fe3d73613cb3e1fb609f84/packages/sync-service/lib/electric/plug/serve_shape_plug.ex#L234

I'm guessing it's legacy. Tanstack.DB doesn't use it. I've removed it from the telemetry.

packages/sync-service/lib/electric/plug/utils.ex
robacourt: Done

packages/sync-service/test/electric/plug/subset_telemetry_test.exs
robacourt: Yeah, I've removed it

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -134,10 +142,81 @@ defmodule Electric.Plug.Utils do
...
+
+  defp telemetry_body_params(_), do: %{}
+
+  defp subset_params_telemetry_attr(params) do
+    case Map.fetch(params, @telemetry_body_subset_params_key) do
+      {:ok, value} ->
+        %{
+          "http.body_param.subset.params" =>
+            value
+            |> inspect(limit: :infinity, printable_limit: :infinity)
```
alco: A bit dangerous. We know we're going to truncate the string anyway. I would just use default options: `limit` is 100 elements and `printable_limit` is 4096 by default, so I don't think it's strictly necessary to pass any options to this `inspect()` call.

packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -134,10 +142,81 @@ defmodule Electric.Plug.Utils do
...
+
+  defp scalar_attr_value(value) when is_binary(value) do
+    {:ok, truncate_telemetry_string(value)}
+  end
+
+  defp scalar_attr_value(value) when is_boolean(value) or is_number(value), do: {:ok, value}
+  defp scalar_attr_value(_value), do: :skip
+
+  defp truncate_telemetry_string(value)
+       when byte_size(value) <= @max_telemetry_string_bytes, do: value
```
alco: I've just noticed this branch doesn't call `trim_invalid_utf8()`. I thought that function was necessary because we don't know what a POST body might contain but we must ensure valid UTF-8 codepoints in the OTEL payload.

But if that function is only used to fix the issue where `binary_part` truncates the string mid-codepoint, then a more idiomatic solution would be to use `String.slice()`.

alco (approved):

packages/sync-service/lib/electric/plug/utils.ex
robacourt: done

packages/sync-service/lib/electric/plug/utils.ex
robacourt: done

PR #4221, opened 2026-04-29
chore: narrower CI runs

icehaunter:
Greatly simplifies CI matrices so package, example, and image jobs do less unrelated work for a given change. The main goal is to avoid testing or installing packages that are not affected by a PR, while keeping broad fallback behavior for changes that can affect the whole workspace.

What changed:

- Added `scripts/ci/affected-workspaces.mjs` to compute affected pnpm workspaces from the PR/base diff, with a full-workspace fallback for global inputs like root package metadata, lockfiles, shared tsconfig files, CI scripts, sync-service, and telemetry changes.
- Updated `ts_tests.yml` to use the affected workspace matrix instead of testing every TS package and example every time.
- Split TS package tests into packages that need the shared sync-service stack and packages that do not, so agents/start/y-electric-style packages no longer build or wait on a shared Electric image.
- Kept sync-service-backed tests for the packages that need it: `packages/typescript-client`, `packages/react-hooks`, and `packages/experimental`.
- Switched TS package, example, PR preview, and example browser-test installs to filtered `pnpm --filter ... install` where the workflow has a concrete package/example target.
- Added the `agents-runtime` test-only install exception for `@electric-ax/agents-server...`, because the runtime DSL tests dynamically import the agents server test harness.
- Added `scripts/ci/example-matrix.mjs` as the shared source of truth for deployable examples, production example tests, deploy-all, and teardown matrices.
- Simplified example deploy/test workflows to consume the generated example matrix instead of duplicating long hard-coded step lists.
- Runs browser tests directly for deployed examples when the matrix entry defines a deploy test folder, and keeps URL-only production tests for externally hosted examples.
- Removed stale/missing `nextjs` example entries from generated deploy/test matrices.
- Updated example deploy, deploy-all, teardown, and test-example installs to use filtered installs when the target example has a workspace `package.json`, with fallback full install only for non-workspace example paths.
- Scoped Lux integration tests to sync-service-related changes only: sync-service, electric-telemetry, integration test harness files, `.tool-versions`, and the workflow itself.
- Simplified sync-service Docker publishing by moving the Electric image build/publish flow onto the shared `docker_multiarch_image.yml` workflow.
- Extended the shared multi-arch Docker workflow to support tags that target multiple repositories, so the sync-service workflow can publish both `electricsql/electric` and `electricsql/electric-canary` tags through the reusable workflow.
- Kept the TypeScript package preview workflow narrower by installing only the previewed packages and their workspace dependencies.

### DIFF
--- a/.github/actions/test-example/action.yml
+++ b/.github/actions/test-example/action.yml
@@ -24,7 +24,12 @@ runs:
         cache: pnpm
     - shell: bash
       if: inputs.example_url != ''
-      run: pnpm install --frozen-lockfile
+      run: |
+        if [ -f "examples/${{ inputs.test_folder }}/package.json" ]; then
+          pnpm --filter "$(jq -r .name "examples/${{ inputs.test_folder }}/package.json")..." --filter @electric-examples/shared install --frozen-lockfile
+        else
+          pnpm install --frozen-lockfile
+        fi
     - name: Run browser tests
       if: inputs.example_url != ''
       working-directory: examples/${{ inputs.test_folder }}
--- a/.github/workflows/deploy_all_examples.yml
+++ b/.github/workflows/deploy_all_examples.yml
@@ -8,35 +8,26 @@ concurrency:
   group: ${{ github.event_name == 'push' && 'prod-deploy-group' || format('examples-pr-{0}', github.event.number) }}
 
 jobs:
+  example-matrix:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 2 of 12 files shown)


alco (approved):
Nice!

I'd love to see some short guidance inside the new .mjs files on how those paths/patterns were generated and how those lists should be updated in the future, to steer future agents in the right direction when we add/remove examples and/or packages.

PR #4231, opened 2026-04-29
fix(elixir-client):  Sync CDN-resilience fixes from the TypeScript client

magnetised:
The TS client has three behaviours around 409 (must-refetch) handling that the Elixir client was missing. All three protect against misbehaving proxies/CDNs that strip query params, cache 409s, or omit response bodies.

Changes

1. Cache-buster on every 409 — adds a fresh cache-buster query param after every must-refetch response, so a CDN that ignores expired_handle in its cache key can't pin us to a stale 409. (poll.ex handle_must_refetch/4)
2. Self-heal stuck expired_handle cache — when stale retries are exhausted, clear the cached expired handle and retry once without it before raising. Recovers from the case where a stale expired_handle value itself is what's poisoning the
  CDN cache slot. (expired_shapes_cache.ex clear_handle/1, poll.ex handle_stale_response/2, shape_state.ex self_heal_attempted? flag)
3. Synthetic must-refetch control message — always emit a must_refetch message on 409, regardless of body content, so subscribers reliably get the signal to clear local state even when a proxy strips the body.

### DIFF
--- a/.changeset/dirty-trains-fry.md
+++ b/.changeset/dirty-trains-fry.md
@@ -1,0 +1,5 @@
+---
+'@core/elixir-client': patch
+---
+
+Sync CDN-resilience fixes from the TypeScript client: cache-buster on every 409, self-heal a stuck expired handle cache and synthetic must-refetch header response for all 409s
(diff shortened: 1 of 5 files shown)


robacourt (approved):
Nice!

PR #4232, opened 2026-04-29
Make otel-export lux test reliable via normalized output and per-pattern grep

alco:
## Summary

Reworks the `otel-export.lux` integration test so its assertions no longer depend on the raw, ordering-sensitive output of the OtelCollector debug exporter.

- Adds `integration-tests/support_files/normalize-otel-output.py`, which collapses the verbose collector log into a compact, stable one-line-per-metric/span representation.
- Adds a `grep_otel_collector_output` macro in `_macros.luxinc` so each expected metric/span is asserted independently against the normalized output.
- Rewrites `otel-export.lux` to verify resource attributes for both application and stack metrics on collector startup, then assert each expected metric and span via the new macro — removing the previous coupling to alphabetical ordering of patterns.

## Test plan

- [x] `otel-export.lux` integration test passes locally
- [x] CI green

### DIFF
--- a/integration-tests/support_files/normalize-otel-output.py
+++ b/integration-tests/support_files/normalize-otel-output.py
@@ -1,0 +1,170 @@
+#!/usr/bin/env python3
+"""
+Normalizes OTel collector debug output to single-line format.
+
+Input (multi-line):
+    info	ResourceMetrics #0
+    Resource attributes:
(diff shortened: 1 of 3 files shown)


alco:
Supersedes #4226.

icehaunter (approved):

PR #4235, opened 2026-04-29
test changeset test

magnetised:


.github/workflows/check_changeset.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v5.0.0
+        with:
+          fetch-depth: 0
+      - uses: pnpm/action-setup@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: '.tool-versions'
+          cache: pnpm
+      - run: pnpm install --frozen-lockfile
```
icehaunter, line 28: We can install much less deps here if we filter to workspace root
```suggestion
      - run: pnpm install --filter . --frozen-lockfile
```

icehaunter (changes requested):
I'm OK  with this, but we could also just use https://github.com/apps/changeset-bot which was made for this purpose I think

magnetised:
> I'm OK with this, but we could also just use https://github.com/apps/changeset-bot which was made for this purpose I think

this is the actual change: https://github.com/electric-sql/electric/pull/4234

this pr was just to test the behaviour

PR #4238, opened 2026-04-29
CI: run oracle property tests

robacourt:
## Summary
- add a dedicated `oracle_property_test` job to the sync-service workflow
- keep it independent from the existing matrix job so it runs in parallel

You can see that they've run [here](https://github.com/electric-sql/electric/actions/runs/25119178438/job/73614761171?pr=4238)


magnetised (approved):
Exciting

PR #4240, opened 2026-04-29
CI: run where clause filter performance tests

robacourt:
## Summary
- add a dedicated `performance_test` job to the sync-service workflow
- run `mix test --only performance` in parallel with the existing jobs
- set `SKIP_REPATCH_PREWARM=true` as Repatch was making the tests flakey.

I've run them 5 times on CI and pass every time now that Repatch is out of the way: https://github.com/electric-sql/electric/actions/runs/25120421568/job/73619279602?pr=4240


alco (approved):
LGTM.

I've asked Codex to deduplicate the repetitive setup and build steps. Let me know if you think it is not the right approach to this.

PR #4274, opened 2026-05-05
Fix otel context propagation in a few more places missed previsouly

alco:
## Summary

Follow-up to #4149 which fixed OTel context propagation in three spawn sites. While auditing the rest of the codebase, three additional places were found where `with_child_span` calls were silently dropping spans due to missing parent context.

### 1. `Effects.query_move_in_async/4` (spawned task)

The same pattern as the sites fixed in #4149: a `Task.Supervisor.start_child` call wrapping `SnapshotQuery.execute_for_shape`, but without propagating the OTel context into the spawned task. Fixed by capturing the context before the spawn and attaching it inside the task closure.

### 2. `ShapeCache.handle_call({:create_or_wait_shape_handle, shape, otel_ctx})`

The `otel_ctx` was already being passed in the GenServer call message (from `Electric.Shapes.get_or_create_shape_handle/3`), but the handler was only forwarding it to the Snapshotter — it never set the context on the ShapeCache GenServer process itself. As a result, the `with_child_span` calls inside `ShapeStatus.fetch_handle_by_shape_critical` and `ShapeStatus.add_shape` were silently dropped.

### 3. `ShapeCache.handle_call({:start_consumer_for_handle, ...})`

This handler had a `# TODO: otel ctx from shape log collector?` comment acknowledging the gap. The caller, `ConsumerRegistry.start_consumer!/2`, wraps the GenServer call in `OpenTelemetry.with_span(...)` but the context wasn't crossing the GenServer call boundary. The public function signature was changed to take an `opts` keyword list (matching `get_or_create_shape_handle/3`), with `ConsumerRegistry` capturing the current context inside its span and passing it through.

## Test plan

- [x] `mix compile` clean
- [x] `mix test test/electric/telemetry/open_telemetry_test.exs` — 8 passing
- [ ] After deploy: verify `name = shape_status.add_shape` and `name = shape_status.fetch_handle_by_shape_critical` rows show up in Honeycomb under shape-creation traces (previously dropped).

### DIFF
--- a/.changeset/otel-context-effects-move-in.md
+++ b/.changeset/otel-context-effects-move-in.md
@@ -1,0 +1,9 @@
+---
+'@core/sync-service': patch
+---
+
+Propagate OpenTelemetry context so that child spans in ShapeStatus and SnapshotQuery are linked to originating traces:
+
+- `Effects.query_move_in_async`: propagate context into spawned task
+- `ShapeCache.handle_call({:create_or_wait_shape_handle, ...})`: set context before calling ShapeStatus functions
+- `ShapeCache.handle_call({:start_consumer_for_handle, ...})`: accept and set context from ConsumerRegistry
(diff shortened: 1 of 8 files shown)


robacourt (approved):
Nice!

PR #4281, opened 2026-05-06
feat(sync-service): add Postgres concat(VARIADIC text) to shape subset eval

abegehr:
Register concat with NULL-skipping semantics (empty string when all args are NULL), matching text concat behavior distinct from || with NULL.

Tests cover literals, refs, ilike(concat(...)), contrast with ||, and document that concat() is unsupported (no concat/0 overload).

### DIFF
--- a/.changeset/moody-maps-approve.md
+++ b/.changeset/moody-maps-approve.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Add Postgres concat(variadic text) to shape subset WHERE evaluation, with NULL arguments skipped and SqlGenerator support for concat(...) round-trips.
--- a/.changeset/moody-maps-approve.md
+++ b/.changeset/moody-maps-approve.md
@@ -2,4 +2,4 @@
 '@core/sync-service': patch
 ---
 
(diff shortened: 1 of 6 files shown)


robacourt (changes requested):
Looking good! Please could you add some tests for some of the other edge cases mentioned here: https://chatgpt.com/share/69fb490a-4c2c-8327-a6a2-74f730c95ea9 e.g. date types, variadic arrays, char(n). It's fine to just return an error when parsing so that we don't support them, but if we do parse then successfully we should handle these cases correctly.

abegehr:
added three more tests for edge-cases

.changeset/moody-maps-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Add Postgres concat(variadic text) to shape subset WHERE evaluation, with NULL arguments skipped and SqlGenerator support for concat(...) round-trips.
```
robacourt:
```suggestion
Add CONCAT(variadic text) support
```

robacourt (approved):
Great work! Approved!

PR #4282, opened 2026-05-06
Fix requestSnapshot publish ordering

KyleAMathews:
Fix `requestSnapshot()` so callers can safely read subscribed state immediately after awaiting it.

Previously, `requestSnapshot()` fetched and injected snapshot data into the stream but did not await the async subscriber publication chain. This meant consumers such as TanStack DB on-demand loading could observe stale state immediately after `await requestSnapshot()`.

## Root Cause

`requestSnapshot()` called `#onMessages(...)` without awaiting it. That method publishes through the stream’s serialized subscriber chain, so async subscribers could still be processing the injected snapshot batch after `requestSnapshot()` had already resolved.

Adding the await exposed a second edge case: if a subscriber reentrantly called and awaited `requestSnapshot()`, the nested publish could be appended behind the currently-running subscriber callback, causing a deadlock.

## Approach

- Await `#onMessages(...)` in `requestSnapshot()` so the promise resolves only after the injected snapshot batch has been delivered to subscribers.
- Make `#publish()` reentrant-safe by delivering nested publishes immediately when already inside subscriber delivery, avoiding self-deadlock.
- Strengthen tests around async subscriber completion, control boundary messages, immediate `Shape` visibility, and reentrant `requestSnapshot()` calls.

## Key Invariants

- `await requestSnapshot()` means subscriber callbacks for the injected snapshot batch have completed.
- Snapshot control messages such as `snapshot-end` and `subset-end` are delivered before `requestSnapshot()` resolves.
- Reentrant `requestSnapshot()` calls from subscriber callbacks do not deadlock.
- Non-reentrant message delivery remains serialized through the existing publish chain.

## Non-goals

- This does not change Electric shape fetch semantics or snapshot filtering.
- This does not add new public APIs.
- This does not modify the ShapeStream state-machine spec; the desired behavior was already clear, but the async publication step was not awaited.

## Trade-offs

The fix keeps the main publish path serialized while special-casing reentrant delivery to avoid deadlock. Throwing on reentrant `requestSnapshot()` would have avoided the deadlock too, but it would introduce a new API restriction. Supporting it is safer and more ergonomic for subscribers.

## Verification

```bash
cd packages/typescript-client

pnpm exec vitest --config vitest.unit.config.ts --run test/stream.test.ts -t "requestSnapshot can be awaited reentrantly"
pnpm exec vitest --config vitest.unit.config.ts --run test/stream.test.ts -t "requestSnapshot"
pnpm exec vitest --config vitest.unit.config.ts --run
```

Results:

```text
requestSnapshot can be awaited reentrantly: 1 passed
requestSnapshot tests: 2 passed
full unit config: 13 files passed, 383 tests passed
```

Also red/green verified the reentrant regression: the new reentrant test fails with the old chained publish implementation and passes with the reentrant-safe publish fix.

## Files changed

- `packages/typescript-client/src/client.ts`
  - Await snapshot message publication in `requestSnapshot()`.
  - Make `#publish()` reentrant-safe.

- `packages/typescript-client/test/stream.test.ts`
  - Add async subscriber regression coverage.
  - Assert snapshot boundary messages are delivered before resolution.
  - Add reentrant `requestSnapshot()` regression coverage.

- `packages/typescript-client/test/client.test.ts`
  - Strengthen high-level assertion so immediate reads after `await requestSnapshot()` are verified.

- `.changeset/fix-requestsnapshot-publish-ordering.md`
  - Add patch changeset for `@electric-sql/client`.


packages/typescript-client/src/client.ts
```diff
@@ -609,6 +609,12 @@ export class ShapeStream<T extends Row<unknown> = Row>
   #tickPromiseResolver?: () => void
   #tickPromiseRejecter?: (reason?: unknown) => void
   #messageChain = Promise.resolve<void[]>([]) // promise chain for incoming messages
+  // Tracks when subscriber callbacks are actively being delivered from
+  // #messageChain. requestSnapshot can inject a nested batch from inside a
+  // subscriber; in that reentrant case #publish uses this as an intentional
+  // escape hatch to deliver the nested snapshot batch immediately rather than
+  // queueing it behind the subscriber that is awaiting it.
```
msfstef, line 620:
```suggestion
```
I think we should try to avoid agent comment-creep - the flag's function is fairly clear with the comment left when used in code further down

packages/typescript-client/src/client.ts
```diff
@@ -1735,7 +1737,25 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
+
+    // We process messages asynchronously but SSE's `onmessage` handler is
+    // synchronous. Use a promise chain to ensure handlers execute sequentially
+    // in the order messages were received. If a subscriber reentrantly requests
+    // a snapshot, deliver that nested batch immediately instead of appending it
+    // behind the currently-running subscriber callback, which would deadlock
+    // when requestSnapshot awaits publication.
+    if (this.#isPublishing) {
+      return deliver()
+    }
```
msfstef: This feels a bit odd - we have a message chain promise to serialize message processing, which gets bypassed if you publish messages while a batch of messages is being processed.

Effectively, anyone calling `this.#publish` either ends up on the message chain, or if the message chain is executing it gets published out of order. In fact, since SSE's `onmessage` doesn't await the call, I would actually expect for batches of messages from the same sse connection to end up being processed/published out of order.

I'd much prefer if we could explicitly specify an API parameter to skip the strict ordering such that the `requestSnapshot` call in particular avoids it

msfstef (dismissed):
I believe that this change breaks sequential processing of SSE messages, and that either way if we want snapshot messages to make it into processing out of band we should make the API explicit about that I think

packages/typescript-client/src/client.ts
```diff
@@ -1735,7 +1737,25 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
-    )
+
+    // We process messages asynchronously but SSE's `onmessage` handler is
+    // synchronous. Use a promise chain to ensure handlers execute sequentially
+    // in the order messages were received. If a subscriber reentrantly requests
+    // a snapshot, deliver that nested batch immediately instead of appending it
+    // behind the currently-running subscriber callback, which would deadlock
+    // when requestSnapshot awaits publication.
+    if (this.#isPublishing) {
+      return deliver()
```
kevin-dp, line 1812: Since we call `deliver()` without awaiting i think we can still get "wrong" execution order.
Concrete scenario with two subscribers A and B:

1. Batch M1 arrives. `#isPublishing` flips to true, outer `deliver()` starts.
2. `Promise.all` spins up `callbackA(M1)` and `callbackB(M1)` concurrently. Both run up to their first await.
3. A awaits `requestSnapshot()`. B is still mid-await processing M1 (e.g. an async write, a fetch, awaiting downstream state).
4. `requestSnapshot()` calls `#onMessages(snapshot)` → `#publish(snapshot)` → reentrant branch fires → `deliver()` invokes **both** `callbackA(snapshot)` and `callbackB(snapshot)` immediately.
5. B's callback is now re-entered with the snapshot batch *before its M1 invocation has returned*.

So the per-subscriber invariant "my callback won't be re-entered while a previous invocation is in flight" is broken for any bystander subscriber whose M1 callback is still awaiting something when the reentrant publish fires. The messageChain-level ordering is preserved (next queued batch still waits), but B can observe `M1-start → snapshot-start → snapshot-end → M1-end` interleaving in its own callback.

In practice this may be fine — if subscribers are effectively stateless across awaits, the window is invisible. But a subscriber that mutates `this.something` across an await in its callback could see corruption.

kevin-dp (approved):

packages/typescript-client/src/client.ts
kevin-dp: Just seeing this, turns out we have the same concern :)

KyleAMathews:
## PR comment draft

I narrowed the reentrant publish bypass so it only applies to snapshot injection from `requestSnapshot()`, rather than to any nested publish while `#isPublishing` is true.

Concretely:

- `#onMessages(...)` and `#publish(...)` now accept an internal opt-in flag for reentrant bypass
- ordinary stream batches, including SSE batches, still serialize through `#messageChain`
- only `requestSnapshot()` calls `#onMessages(..., { allowReentrantPublishBypass: true })`

This preserves the original deadlock fix:

- if a subscriber handling batch `M1` does `await requestSnapshot(...)`, the injected snapshot batch can still publish immediately instead of being queued behind the subscriber that is awaiting it

But it avoids the broader regression:

- later SSE batches no longer bypass earlier in-flight publishes just because `#isPublishing === true`

I added tests for both sides of the behavior:

1. **SSE ordering regression test**
   - proves a later SSE batch is not delivered before subscriber callbacks for the earlier batch complete

2. **Bystander subscriber behavior test**
   - explicitly documents current behavior that a reentrant `requestSnapshot()` may re-enter bystander subscribers before their earlier callback completes

So the resulting contract is:

- **ordinary stream traffic remains serialized**
- **snapshot injection is the only allowed reentrant bypass**
- **bystander reentrancy during snapshot injection remains allowed and now has test coverage**

packages/typescript-client/src/client.ts
```diff
@@ -1432,7 +1436,9 @@ export class ShapeStream<T extends Row<unknown> = Row>
       return true // Always process control messages
     })
 
-    await this.#publish(messagesToProcess)
+    await this.#publish(messagesToProcess, {
+      allowReentrantBypass: opts.allowReentrantPublishBypass,
```
kevin-dp, line 1496: Could we modify the signature of the `opts` argument of `#onMessages` such that we don't need to rename the property here and can just pass `opts`?

kevin-dp (approved):
Thanks for addressing the feedback, LGTM now!

packages/typescript-client/src/client.ts
msfstef: bump on this

packages/typescript-client/src/client.ts
```diff
@@ -1735,7 +1746,24 @@ export class ShapeStream<T extends Row<unknown> = Row>
...
         })
       )
-    )
+
+    // We process messages asynchronously but SSE's `onmessage` handler is
+    // synchronous. Use a promise chain to ensure handlers execute sequentially
+    // in the order messages were received. Only requestSnapshot's injected
+    // snapshot batch is allowed to bypass the queue reentrantly; ordinary
+    // stream batches (including SSE batches) must remain serialized.
+    if (this.#isPublishing && opts.allowReentrantBypass) {
```
msfstef, line 1811: do we even need this whole `isPublishing` mechanism anymore if we have the opts flag? Why not just have the flag dictate whether the publishing needs to be queued or not - the snapshot can set the flag as not needing to be queued, and it always just publishes out of band. Makes everything simpler - am I missing something?

msfstef (commented):
If we're adding a flag for toggling out of band publishes, why not make the flag alone the only mechanism needed?

docs/requestsnapshot-pr-comment.md
```diff
@@ -0,0 +1,31 @@
+## PR comment draft
```
icehaunter: SHould be deleted probably

packages/typescript-client/src/client.ts
KyleAMathews: Yeah, the `#isPublishing` check is intentional here.

The option only authorizes a bypass, but we only actually bypass when we’re currently delivering subscriber callbacks. That preserves normal message-chain ordering for non-reentrant `requestSnapshot()` calls, while still avoiding the deadlock when `requestSnapshot()` is awaited from inside a subscriber.

So the intended behavior is:

- ordinary stream/SSE batches: always queued
- `requestSnapshot()` called outside subscriber delivery: queued
- `requestSnapshot()` called reentrantly from inside subscriber delivery: bypasses, otherwise it can deadlock

The tests added in `da2504595` cover both sides of this:

- `does not deliver a later SSE batch before subscriber callbacks for the earlier batch finish`
- `reentrant requestSnapshot may re-enter bystander subscribers before their earlier callback completes`

If we let the option alone dictate bypassing, then every `requestSnapshot()` would publish out-of-band, even when it isn’t reentrant. That seems like a broader ordering relaxation than we need.

PR #4284, opened 2026-05-06
feat(sync-service): hibernate before suspend to enable GC

alco:
## Summary

When consumer suspend is enabled, consumers now hibernate first (triggering GC) before eventually suspending (terminating). Previously, consumers went directly to suspend without hibernating, and due to a large timeout for suspend, consumer processes could hold onto garbage memory for longer than necessary.

- Add `shape_suspend_after` config (default 10 minutes) - delay between hibernation and suspension
- Consumer hibernates on `hibernate_after` timeout, schedules suspend timer
- Suspend timer fires after `suspend_after` ms, terminates process if still idle
- Any activity cancels the pending suspend timer, restarting the cycle
- Update `ConsumerRegistry.enable_suspend/4` to include suspend_after parameter

### DIFF
--- a/.changeset/hibernate-then-suspend.md
+++ b/.changeset/hibernate-then-suspend.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': minor
+---
+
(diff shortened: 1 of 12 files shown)


magnetised (approved):
Fab. Exactly the code I had in my head. Annoying how much of this is just threading the config change through...

PR #4290, opened 2026-05-07
Add queued inbox message modes for agents

samwillis:
## Summary
- Add first-class inbox message delivery modes (`queued`, `steer`, and `paused`) with pending/processed/cancelled status metadata across the agents runtime and server APIs.
- Promote queued inbox messages one at a time in the runtime, pause at edited messages, and avoid deleted/cancelled lifecycle events waking agents as runnable prompts.
- Add the agents UI queued-message drawer with edit/delete/steer/reorder controls, fractional queue positions, and inline optimistic display for idle queued sends.

## Test plan
- `pnpm -C packages/agents-runtime typecheck`
- `pnpm -C packages/agents-server typecheck`
- `pnpm -C packages/agents-server-ui typecheck`
- Browser smoke tested `http://localhost:5173/__agent_ui/` for queued sends, inline idle optimistic display, and Horton generation.

### DIFF
--- a/packages/agents-runtime/src/context-factory.ts
+++ b/packages/agents-runtime/src/context-factory.ts
@@ -150,16 +150,24 @@ function getTriggerMessageText(
   events: Array<ChangeEvent>,
   wakeOffset: string
 ): string {
-  if (wakeEvent.type === `message_received`) {
+  if (wakeEvent.type === `inbox` || wakeEvent.type === `message_received`) {
     let latestPayload: unknown = wakeEvent.payload
     for (let index = events.length - 1; index >= 0; index--) {
       const event = events[index]!
-      if (event.type !== `message_received`) {
+      if (event.type !== `inbox` && event.type !== `message_received`) {
         continue
       }
 
-      const payload = (event.value as { payload?: unknown } | undefined)
-        ?.payload
+      const value = event.value as
+        | {
+            payload?: unknown
+            status?: `pending` | `processed` | `cancelled`
+          }
+        | undefined
+      if (value?.status === `cancelled`) {
+        continue
+      }
+      const payload = value?.payload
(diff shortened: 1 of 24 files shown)


packages/agents-runtime/src/context-factory.ts
```diff
@@ -150,16 +150,24 @@ function getTriggerMessageText(
   events: Array<ChangeEvent>,
   wakeOffset: string
 ): string {
-  if (wakeEvent.type === `message_received`) {
+  if (wakeEvent.type === `inbox` || wakeEvent.type === `message_received`) {
```
KyleAMathews, line 153: let's completely remove message_recieved

packages/agents-runtime/src/entity-stream-db.ts
```diff
@@ -139,6 +139,11 @@ export function createEntityStreamDB(
     collectionNameByEventType.set(def.type, name)
     rowOffsetsByCollection.set(name, new Map())
   }
+  // Back-compat for streams written before the inbox collection's event type
```
KyleAMathews, line 142: seems fine to have no backwards compatibility?

packages/agents-runtime/src/entity-timeline.ts
```diff
@@ -109,6 +109,11 @@ export interface IncludesInboxMessage {
   from: string
   payload: unknown
   timestamp: string
+  mode?: `immediate` | `queued` | `paused` | `steer`
+  status?: `pending` | `processed` | `cancelled`
```
KyleAMathews, line 113: not necessarily needed for this PR but I'm annoyed that these types still exist as I tried to derive everything from entity-schema.ts but missed this

samwillis:
folded into #4309

PR #4303, opened 2026-05-10
Emit telemetry spans for move_in_query snapshot operations

robacourt:
## Summary

`shape_snapshot.execute_for_shape` and `shape_snapshot.query_fn` spans were missing from production traces for `move_in_query` operations, while `initial_snapshot` and `subset_query` worked fine. This wires up a parent span so the existing child spans get emitted.

## Problem

Sampling for `shape_snapshot.query_fn` is a plain `included?(_) -> true` (no rate limiting), so the absence of `move_in_query` spans wasn't a sampling issue — the spans were never being created.

### DIFF
--- a/.changeset/move-in-query-telemetry.md
+++ b/.changeset/move-in-query-telemetry.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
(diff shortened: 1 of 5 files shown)


alco:
I thought 686ebdbcafdb940d3e4fca8cae15021b8b2d13a3 had fixed that. Just now reviewing the changes, I see we have two functions named `query_move_in_async`: one in `PartialModes` and the other one in `Effects`. My commit fixed the former one (where the root span comes from the shape serving plug). 

The one defined in `Effects` is invoked in response to a `materializer_changes` callback in the Consumer module. We could start a span right in the handle_info callback instead, seems like a less ad-hoc approach then creating a span right inside the inner function.

Looks like the two implementations of `query_move_in_async` overlap. Can we deduplicate them into one?

alco:
I have confirmed that `PartialModes.query_move_in_async()` is dead code since #4051. I'm removing it completely in https://github.com/electric-sql/electric/pull/4344.

Took the liberty to clean up the changes in your branch to avoid repetition of shape attrs construction and excessive nesting.

LGTM!

alco (approved):

PR #4307, opened 2026-05-12
feat(agents-server): refactor routing for tenant library mode

icehaunter:
## Summary
- Rebuilds `agents-server` from monolithic HTTP handling into a tenant-context-driven router architecture.
- Makes the package usable as a library entrypoint through `globalRouter.fetch(request, tenantContext)`, while keeping an OSS standalone wrapper for local/server use.
- Treats the current Durable Streams subscription contract and routing adapter boundary as the source of truth for backend stream paths.

## Work Done
- Splits routing into focused modules for global, OSS, internal control-plane, entities, entity types, cron, Durable Streams proxying, and stream append handling.
- Introduces explicit `TenantContext`, tenant-aware runtime host wiring, per-tenant runtime/registry/stream clients, and tenant-scoped database schema updates.
- Adds `DurableStreamsRoutingAdapter` plus the OSS path-prefixed single-tenant adapter for mapping public/runtime stream names to backend DS paths.
- Reworks Durable Streams subscription proxying around `/v1/stream-meta/subscriptions/*`, including path mapping, webhook URL rewriting, callback storage, claim/ack/release forwarding, and response mapping.
- Moves OSS bootstrap concerns into `server.ts` and keeps routing code library-safe for external hosts that own tenancy and authentication.
- Updates wake forwarding and callback forwarding to enrich runtime wake payloads, preserve DS callback state, map ack streams, mark entities idle on completion, and notify entity bridges.
- Adds supporting runtime, scheduler, projector, UI, desktop, docs, and tests needed for the refactored server architecture.


samwillis (approved):
Merge when green

PR #4310, opened 2026-05-12
feat: add signal-driven agent controls

samwillis:
## Summary
This PR implements the agent signal controls from the [agent signals plan](https://gist.github.com/samwillis/7406b5c16272c250316e29c83eb9b2fb), with the paused-state behavior and `SIGSTOP` semantics called out below.

- Adds durable agent signals end-to-end across the runtime schema/client, server APIs, routing, conformance helpers, and ElectricAX CLI surface.
- Introduces signal-aware runtime handling for interrupt, pause/resume, graceful stop, forced kill, wake restart, and custom user signals.
- Adds focused coverage for server transitions, runtime wake behavior, abort races, ignored idle interrupts, and conformance kill semantics.
- Wires the agents UI to send signals from the entity menu and keyboard shortcuts, including `Escape` for interrupt and `Ctrl+C` for the composed stop-immediately action.

image

### `SIGINT`
**Primary intent:** Interrupt active work and continue.

**LLM generation:** Aborts the active model run via the runtime abort controller. If it arrives before the run controller exists on a running entity, the abort is remembered and applied when the controller is created. If no run is active and the entity is not running, it is recorded as ignored so it cannot abort the next run.

**Current run:** The current run is aborted and settled promptly when one exists.

**Current wake:** The wake remains open and may continue to fresh queued work.

**Entity state:** Server state usually stays `running`; the runtime records the signal as handled/aborted or handled/ignored.

**Lifecycle / queue behavior:** Non-terminal. Existing queued work can continue after the abort.

### `SIGSTOP`
**Primary intent:** Stop after the current checkpoint.

**LLM generation:** Does not directly abort generation.

**Current run:** Lets the current handler/checkpoint finish.

**Current wake:** Closes the wake cleanly after the checkpoint and releases the durable-stream claim.

**Entity state:** `idle` or `running` moves to `paused`.

**Lifecycle / queue behavior:** Non-terminal. The entity becomes quiescent: already-pending queue items do not wake it or continue draining. Following the plan's `paused` state model, the entity still exists, remains resumable/forkable, and can accept appended messages while paused. In this implementation, `SIGCONT` resumes it to `idle`, and a new non-paused message send also resumes it by transitioning `paused -> idle` before appending so that new work wakes it.

### `SIGSTOP+SIGINT` (`Stop immediately` / `Ctrl+C`)
**Primary intent:** Provide the terminal-style stop UX: abort the active generation and pause the entity.

**LLM generation:** `SIGINT` aborts the active generation.

**Current run:** The current run is aborted by `SIGINT`; `SIGSTOP` ensures pause is requested.

**Current wake:** The wake exits before promoting queued work, so the queue does not keep draining after the abort.

**Entity state:** Moves to `paused`.

**Lifecycle / queue behavior:** Composite UI behavior, not a separate protocol signal. Used by the menu `Stop immediately` action and global/input `Ctrl+C`.

### `SIGCONT`
**Primary intent:** Resume paused work.

**LLM generation:** No direct generation effect.

**Current run:** No active run is expected while paused.

**Current wake:** Runtime marks resume requested and can continue queued work.

**Entity state:** `paused` moves to `idle`.

**Lifecycle / queue behavior:** Non-terminal resume path for paused entities. The plan notes `paused -> running`, but this codebase intentionally maps resume to the scheduler-ready state, `idle`, because that is the state that can be claimed/woken for work.

### `SIGHUP`
**Primary intent:** Restart/reload wake after the current checkpoint.

**LLM generation:** Does not directly abort generation.

**Current run:** Lets the current checkpoint finish.

**Current wake:** Requests wake shutdown after the checkpoint.

**Entity state:** Server state is unchanged.

**Lifecycle / queue behavior:** Non-terminal. Intended for restart/reload style handling without killing the entity.

### `SIGTERM`
**Primary intent:** Gracefully stop.

**LLM generation:** Does not directly abort generation; allows cleanup/current checkpoint.

**Current run:** Runtime requests shutdown and gives the signal handler a chance to run.

**Current wake:** Closes after cleanup/checkpoint.

**Entity state:** `running` moves to `stopping`; `idle`/`paused` moves to `stopped`.

**Lifecycle / queue behavior:** Terminal once stopped. Wake registrations are unregistered when the server transitions directly to `stopped`.

### `SIGKILL`
**Primary intent:** Forced kill.

**LLM generation:** Aborts active generation immediately, including the case where the server has already marked the terminal signal handled and closed streams.

**Current run:** Aborts active run immediately.

**Current wake:** Requests immediate shutdown and closes streams.

**Entity state:** Moves to `killed`.

**Lifecycle / queue behavior:** Terminal/non-restartable. Unregisters wake subscriptions and closes main/error streams.

### `SIGUSR`
**Primary intent:** User-defined signal.

**LLM generation:** No built-in abort behavior.

**Current run:** Delegated to the optional signal handler.

**Current wake:** Wake continues unless custom handling decides otherwise.

**Entity state:** Server state is unchanged by default.

**Lifecycle / queue behavior:** Non-terminal extension point; runtime records handler outcome.

## Alignment With The Plan
- Follows the plan's distinction between immediate signals (`SIGINT`, `SIGKILL`, `SIGUSR`) and checkpointed/control signals (`SIGSTOP`, `SIGTERM`, `SIGHUP`, `SIGCONT`).
- Restores the original `SIGSTOP` design as a cooperative/checkpointed pause rather than making `SIGSTOP` itself abort the generation.
- Implements `paused` as the plan describes: a temporary, non-terminal, resumable state where the entity still exists, preserves state, can be forked, and does not process normal already-pending work while paused.
- Applies the agreed product clarification for waking paused entities: new normal messages resume the entity before append so they are processed promptly, while work that was already pending when the pause happened does not wake it.
- Adds the desired terminal-style UX by composing `SIGSTOP` and `SIGINT` for `Stop immediately`/`Ctrl+C`, so users can abort the current generation and pause without changing the protocol semantics of `SIGSTOP`.
- Keeps legacy `DELETE` kill behavior as a compatibility alias that now routes through signal-based `SIGKILL` semantics.
- Links webhook dispatch subscriptions before spawn `initialMessage` delivery so webhook-routed children do not miss their first wake; runner dispatch keeps the existing post-send ordering.

## Test Plan
- `pnpm -C packages/agents-runtime test --run test/process-wake.test.ts -t "SIGINT|SIGKILL|SIGSTOP"`
- `pnpm -C packages/agents-runtime test --run test/runtime-dsl.test.ts -t "(M1:|N5:)"`
- `pnpm -C packages/agents-server test --run test/dispatch-policy-routing.test.ts test/electric-agents-routes.test.ts test/electric-agents-status.test.ts`
- `pnpm -C packages/agents-runtime typecheck`
- `pnpm -C packages/agents-server typecheck`


icehaunter (approved):

packages/agents-server/src/routing/entities-router.ts
```diff
@@ -655,3 +679,31 @@ async function killEntity(
...
+
+async function signalEntity(
+  request: AgentsRouteRequest,
+  ctx: TenantContext
+): Promise<Response> {
+  const parsed = routeBody<SignalBody>(request)
+  const { entityUrl, entity } = requireExistingEntityRoute(request)
+  let signal
+  try {
+    signal = assertEntitySignal(parsed.signal)
```
icehaunter: Why do we need that instead of narrowing the type in the JSON schema to the list of known signals?

KyleAMathews:
Thanks, this is a nice direction overall — durable signals feel like the right primitive for stop/pause/kill style agent controls. I did a pass over the implementation and had a few concrete things I think we should address before merge.

### CLI only partially exposes the new signal API

`packages/electric-ax/src/index.ts`

The CLI was updated so `agents kill <url>` sends `SIGKILL` via the new `/signal` endpoint, but there’s no general signal command for the rest of the new surface.

Given this PR introduces first-class signals (`SIGINT`, `SIGSTOP`, `SIGCONT`, `SIGHUP`, `SIGTERM`, `SIGUSR`, etc.), I’d expect something like:

```sh
electric agents signal <url> <signal>
```

possibly with `--reason <text>` and `--payload <json>`.

Right now the CLI can’t send `SIGINT` to stop generation, pause/resume via `SIGSTOP`/`SIGCONT`, or send `SIGUSR`. Also, the existing `kill` command still appears to be described as “Delete an entity”, which is now misleading because it sends `SIGKILL` and transitions to `killed`.

### Missing docs updates for the new signal lifecycle API

This PR adds a pretty significant public surface:

- `POST /_electric/entities/:type/:instanceId/signal`
- new statuses: `paused`, `stopping`, `killed`
- new persisted `signal` events
- runtime `ctx.onSignal`
- client APIs `client.signal()` / `client.kill()`
- changed kill semantics from `stopped` to `killed`
- UI/CLI stop/kill behavior changes

I don’t see corresponding docs updates. We should document the signal model, status transitions, which signals are server/runtime-handled vs delivered to `ctx.onSignal`, CLI usage, and compatibility notes around kill now producing `killed` instead of `stopped`.

### `POST /signal` should reject principal entity mutations

`packages/agents-server/src/routing/entities-router.ts`

The new `signalEntity` route doesn’t appear to call the existing principal mutation guard. Other mutating routes, including legacy kill/delete, protect built-in principal entities via `rejectPrincipalEntityMutation(...)`.

As written, a request like:

```http
POST /_electric/entities/principal/:id/signal
{ "signal": "SIGKILL" }
```

can flow through `EntityManager.signal`, update the entity status to `killed`, close streams, unlink dispatch, and clear claim tokens. This seems inconsistent with the existing protection model and probably needs the same guard as the other mutating routes.

### SIGINT abort semantics need to be specified/covered

`packages/agents-runtime/src/process-wake.ts`  
`packages/agents-runtime/src/context-factory.ts`

`SIGINT` currently aborts a controller scoped to the whole handler pass, not just a single active `ctx.agent.run()` call. If the intended behavior is “cancel the current generation/handler invocation, including any in-flight or subsequent model/tool work in that invocation”, then this may be correct.

However, the current naming/logging/docs make it sound like only the active run is aborted. Also, `SIGINT` during arbitrary handler code is still recorded as `outcome: "aborted"`, and any later `ctx.agent.run()` in the same handler starts with an already-aborted signal.

Can we document the intended SIGINT scope and add tests for:

- SIGINT during non-agent handler work
- SIGINT between two `ctx.agent.run()` calls
- tool/model calls receiving and respecting the abort signal

### UI optimistic signal status does not match server signal semantics

`packages/agents-server-ui/src/lib/ElectricAgentsProvider.tsx`

`optimisticStatusForSignal()` currently maps:

```ts
SIGINT -> stopping
SIGSTOP -> paused
SIGTERM -> stopping
SIGCONT -> idle
```

But the server-side semantics are more conditional:

- `SIGINT` does not change entity status.
- `SIGSTOP` only changes `idle -> paused`; for running entities the runtime handles pause later.
- `SIGTERM` changes `idle/paused -> stopped`, but `running -> stopping`.
- `SIGCONT` only changes `paused -> idle`.

This can make the UI show incorrect status immediately after sending signals. The most visible example is chat stop: the UI sends `SIGINT`, then optimistically marks the entity `stopping`, but the server keeps the status unchanged.

The optimistic transition should either mirror `expectedSignalStatus(currentStatus, signal)` or avoid optimistic status changes for signals where the transition is state-dependent / runtime-dependent.

### Global `Ctrl+C` / `Escape` handling can hijack text input behavior

`packages/agents-server-ui/src/router.tsx`

The new signal hotkeys are registered with:

```ts
{ ignoreInputs: false, capture: true }
```

When the active tile entity is running, pressing `Ctrl+C` while focused in an input/textarea can send `SIGSTOP` + `SIGINT` and call `preventDefault()`, breaking normal copy behavior.

Likewise, `Escape` while focused inside inputs/menus/dialog-like UI can send `SIGINT` instead of allowing normal local escape behavior.

This probably needs tighter scoping, or should respect editable targets unless there is a very deliberate UI mode where those keys mean “stop agent”.

### Concurrent terminal signals can overwrite terminal status

`packages/agents-server/src/entity-manager.ts`  
`packages/agents-server/src/entity-registry.ts`

`EntityManager.signal()` checks terminal state using a prior `getEntity()`, then calls `updateStatusWithTxid()`. For terminal target statuses, the update only uses `entityWhere(entityUrl)` and does not guard against the row having become terminal between read and update.

A race like concurrent `SIGKILL` and `SIGTERM` on an idle entity could result in whichever update commits last winning, e.g. `killed` being overwritten back to `stopped`.

Terminal transitions should probably be guarded so terminal state cannot be overwritten by another terminal signal based on stale prior state.

### Guarded no-op signal updates can report success after zero-row update

`packages/agents-server/src/entity-registry.ts`

For no-op/non-terminal signal paths, guarded updates use checks like not `stopped`/`killed`, but the helper returns a txid even if zero rows were updated due to a race with a terminal transition.

That means `EntityManager.signal()` can append a signal event and return success using stale `previousState`, even though the entity is already terminal.

The registry update methods should probably verify row count / returned row and surface a conflict if the guarded update did not apply.

### `ctx.onSignal` docs overstate what is delivered

`packages/agents-runtime/src/types.ts`

The comment says:

```ts
Only catchable signals are delivered here; SIGKILL is terminal and bypasses entity code.
```

But the implementation also handles `SIGINT`, `SIGSTOP`, and `SIGCONT` internally without invoking the registered `onSignal` handler. From the current implementation, only `SIGHUP`, `SIGTERM`, and `SIGUSR` invoke the handler.

The docs should explicitly list which signals are delivered to `onSignal` and which are runtime/server-handled.

### `EntitySignal` is duplicated in runtime code

`packages/agents-runtime/src/entity-schema.ts`  
`packages/agents-runtime/src/runtime-server-client.ts`

`EntitySignal` is defined independently in both places. Since this is a public API plus persisted schema type, it would be easy for the two to drift if a signal is added/removed later. Could `runtime-server-client.ts` reuse the exported schema type instead?

Route namespace note: I don’t see a namespace mismatch. The full endpoint is correctly under `/_electric/entities/:type/:instanceId/signal`, consistent with the existing `/_electric/*` control-plane routes.

packages/agents-server/drizzle/0007_entity_signal_statuses.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "entities" DROP CONSTRAINT "chk_entities_status";
```
icehaunter: Duplicating file with 0009

packages/agents-server/src/entity-manager.ts
```diff
@@ -2354,37 +2388,177 @@ export class EntityManager {
...
+      status,
+      handled: false,
+      outcome: `ignored`,
+      unregisterWakes: false,
+    }
+  }
+
+  private async appendSignalEvent(
+    entity: ElectricAgentsEntity,
+    signalEvent: Record<string, unknown>,
```
icehaunter: Why is this just fully untyped? Makes reasoning about this harder - like how is the `errorCloseEvent` different here

KyleAMathews:
## Testing feedback

Tested the signal feature manually from the desktop app. Two issues found:

### 1. Escape/Ctrl-C keyboard shortcuts don't send SIGINT

Pressing Escape or Ctrl-C while an agent is actively generating doesn't stop it. The `useHotkey` bindings in `router.tsx` exist (lines 204-219) but the signal never reaches the server. Likely cause: `helpers.activeTile?.entityUrl` is null, so `activeEntityUrl` is null and the handler bails early at the `keyboardSignalEntityIsRunning` guard. Sending SIGINT via curl to the `/signal` endpoint works correctly and aborts the active run.

### 2. Agent doesn't see signals in its context

After being interrupted by SIGINT (via curl), the agent's next run doesn't include the signal in its message history, so it doesn't know it was interrupted. It continues as if nothing happened. Signals should be visible in the agent's context so it can acknowledge the interruption.

### 3. Signal not shown in the UI timeline

When a SIGINT is sent (even successfully via curl), there's no visual indication in the chat timeline that the generation was interrupted. The signal should be shown optimistically in the UI when sent, and also rendered in the timeline from the stream data so the user has clear feedback that their stop request was received and acted on.

PR #4312, opened 2026-05-12
fix(sync-service): make live electric-cursor strictly monotonic

balegas:
## Summary

Live SSE shape requests could enter an infinite 2-cycle refetch loop where the client bounced between two cursor values (`?cursor=A` ⇄ `?cursor=B`) at >600 req/s, observed in a customer HAR with 3,724 looping requests over 6 seconds.

Root cause: `Electric.Plug.Utils.get_next_interval_timestamp/2` could return a cursor smaller than the `prev_interval` the client sent.

`get_next_interval_timestamp/2` returned one of:
- `next_interval = ceil(now / T) * T` (a bucket boundary), or
- `next_interval + Enum.random(1..3_600)` (only when `prev_interval == next_interval`).

When the client polled with a **jittered** cursor `bucket + δ` (`δ ∈ 1..3_600`) and wall-clock had not yet crossed `bucket + δ`, the function fell through the `==` check, returned the plain `bucket`, and the response carried a cursor **smaller** than the client's input.

Combined with `Cache-Control: public, max-age=<sse_timeout - 1>` on live SSE responses (`response.ex:298-308` — added intentionally to enable CDN request collapsing), two cached entries could end up referencing each other:

| Cache key (request URL) | Cached `electric-cursor` response header |
|---|---|
| `…?cursor=A…` (clean bucket boundary, e.g. `49694000`) | `B` = `A + δ` (e.g. `49694039`) |
| `…?cursor=B…` (jittered value)                          | `A` (plain bucket, since `B != next_interval`) |

Once both entries warmed up, every client request was a cache hit returning the *other* cursor → infinite cycle within the `max-age` window. The HAR captured this exactly: every looping response shared the same `Date` header (proving cache replay), all 31 bytes (Electric's own SSE keep-alive output) with `electric-cursor: A` and `electric-cursor: B` strictly alternating.

## Fix

`get_next_interval_timestamp/2` now guarantees a strictly monotonic cursor: when bucket math would return a value `<= prev_interval`, jitter forward from `prev` instead of from the bucket. Backward cursors become structurally impossible, so a 2-cycle in cache cannot form regardless of CDN behaviour.

```elixir
prev_int = parse_prev_interval(prev_interval)

if next_interval > prev_int do
  next_interval
else
  prev_int + Enum.random(1..3_600)
end
```

This subsumes the original same-cursor case (`next_interval == prev_int` falls into the `else` branch and still jitters) that PR #1856 was patching.

## Reproduction / regression test

Added `test/electric/plug/utils_test.exs:235` "never returns a cursor smaller than prev_interval". It walks the full jitter range `1..3_600`, asserting `result > prev`. **Fails on `main`** with `cursor went backward: prev=…481, result=…480, current_bucket=…480` — the exact same pattern as the customer HAR. **Passes with the fix.**

```
mix test test/electric/plug/utils_test.exs
1 doctest, 14 tests, 0 failures

mix test test/electric/plug/serve_shape_plug_test.exs
45 tests, 0 failures
```

## Out of scope

Per discussion, this PR does **not** invalidate existing in-flight cached cycle pairs at CDN edges. They expire naturally within `sse_timeout` (≤ 60 s by default).

## Test plan

- [x] New regression test fails on `main`, passes with fix
- [x] `mix test test/electric/plug/utils_test.exs` — 14 tests + 1 doctest pass
- [x] `mix test test/electric/plug/serve_shape_plug_test.exs` — 45 tests pass (function is exercised end-to-end here)
- [ ] Manual: customer reproduction with `live=true&live_sse=true` no longer cycles after deploy

### DIFF
--- a/packages/sync-service/lib/electric/plug/utils.ex
+++ b/packages/sync-service/lib/electric/plug/utils.ex
@@ -76,11 +76,6 @@ defmodule Electric.Plug.Utils do
     next_interval = ceil(diff_in_seconds / long_poll_timeout_sec) * long_poll_timeout_sec
     prev_int = parse_prev_interval(prev_interval)
 
-    # The returned cursor MUST be strictly greater than the client's prev
-    # cursor. Otherwise CDN-cached live responses (cached for ~sse_timeout
-    # via `Cache-Control: public, max-age=<sse_timeout-1>` to enable request
-    # collapsing) can end up referencing each other in a cycle, causing the
-    # client to bounce between two cached entries at line rate.
     if next_interval > prev_int do
       next_interval
     else
--- a/packages/sync-service/lib/electric/plug/utils.ex
+++ b/packages/sync-service/lib/electric/plug/utils.ex
@@ -74,11 +74,26 @@ defmodule Electric.Plug.Utils do
     long_poll_timeout_sec = div(long_poll_timeout_ms, 1000)
     diff_in_seconds = DateTime.diff(DateTime.utc_now(), @oct9th2024, :second)
     next_interval = ceil(diff_in_seconds / long_poll_timeout_sec) * long_poll_timeout_sec
+    prev_int = parse_prev_interval(prev_interval)
 
-    if "#{next_interval}" == prev_interval do
-      next_interval + Enum.random(1..3_600)
-    else
+    # The returned cursor MUST be strictly greater than the client's prev
+    # cursor. Otherwise CDN-cached live responses (cached for ~sse_timeout
+    # via `Cache-Control: public, max-age=<sse_timeout-1>` to enable request
+    # collapsing) can end up referencing each other in a cycle, causing the
+    # client to bounce between two cached entries at line rate.
+    if next_interval > prev_int do
       next_interval
+    else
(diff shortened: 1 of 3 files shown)


packages/sync-service/lib/electric/plug/utils.ex
```diff
@@ -74,11 +74,26 @@ defmodule Electric.Plug.Utils do
...
-    if "#{next_interval}" == prev_interval do
-      next_interval + Enum.random(1..3_600)
-    else
+    prev_int = parse_prev_interval(prev_interval)
+
+    # The returned cursor MUST be strictly greater than the client's prev
+    # cursor. Otherwise CDN-cached live responses (cached for ~sse_timeout
+    # via `Cache-Control: public, max-age=<sse_timeout-1>` to enable request
+    # collapsing) can end up referencing each other in a cycle, causing the
+    # client to bounce between two cached entries at line rate.
```
msfstef: nit: we can skip the verbose comment

msfstef (approved):
LGTM - nice catch

I'd reduce amount of comments

PR #4316, opened 2026-05-12
feat(agents-desktop): Electric Cloud sign-in

kevin-dp:
## Summary
- Adds a **Settings → Account** panel to the Electric Agents desktop app that signs into Electric Cloud via GitHub or Google.
- Reuses the same loopback-redirect OAuth flow the `electric` CLI uses against `dashboard.electric-sql.cloud/api/public/auth/{github,google}/login` — no new backend work required.
- The main process opens the authorize URL in a sandboxed `BrowserWindow`, intercepts the `http://127.0.0.1:53118/callback?token=…` redirect by URL prefix (nothing actually listens on the port), validates the CSRF `state` nonce, and persists the JWT via the existing `safeStorage`-backed `SecretStore`.
- After sign-in (and on app launch when restoring an existing session) the main process hits the admin-API's `auth.whoami` to populate the user's name and workspaces; 401/403 forces a hard sign-out, other failures keep the session.
- The panel surfaces account name + email, workspace list, an **Open Electric Cloud dashboard** button (`shell.openExternal`), and a sign-out button.
- Web build hides the **Account** category — sign-in only makes sense in the desktop runtime since there's no place to safely hold the JWT.

Stacked on top of #4309.

## Implementation notes
- `ELECTRIC_DASHBOARD_URL` env-var override mirrors the CLI convention; default is `https://dashboard.electric-sql.cloud`.
- Cloud port `53118` is distinct from the MCP DCR port `53117` so the two interception prefixes never overlap.
- `auth.whoami` is called over raw `fetch` with the oRPC v1.13 wire format (`POST /api/rpc/auth/whoami` with body `{\"json\":{}}`) so we don't pull `@orpc/client` into the desktop bundle.

## Test plan
- [ ] Open **Settings → Account**, click **Sign in with GitHub**, complete the flow in the popup window, verify the panel flips to signed-in with the correct name/email/workspaces.
- [ ] Same with **Sign in with Google**.
- [ ] Click **Open Electric Cloud dashboard** and confirm the system browser opens to `dashboard.electric-sql.cloud`.
- [ ] Click **Sign out** and confirm the panel returns to the signed-out state.
- [ ] Quit and relaunch the app — the session should restore (name + workspaces re-fetched in the background).
- [ ] Cancel the OAuth popup mid-flow (close the window) — state should return to the prior signed-in or signed-out value without an error banner.


KyleAMathews:
We can just ensure an open port before picking one?

PR #4323, opened 2026-05-12
feat(agents): add principals and refactor routing for tenant library mode

KyleAMathews:
## Summary

Adds **principals** (authenticated identity) to the agents system and refactors the agents-server routing to support tenant library mode. Every API request now carries a `Principal` (`user`, `agent`, `service`, or `system`) threaded through the full lifecycle — from spawn to wake dispatch to handler context.

- Agents-server routing refactored into composable routers (entities, entity-types, runners, internal) that accept a per-request `TenantContext` with a required `principal` field
- New pull-wake runner system: runner registration, dispatch policy resolution, subscription stream linking, wake claims with claim-scoped write tokens, heartbeat liveness, and callback-forward lifecycle
- Principal context flows end-to-end: server extracts from `electric-principal` header → stored as `created_by` on entities → enriched into wake notifications → exposed as `ctx.principal` in runtime handler context
- Server UI uses asserted identity headers for dev-mode authentication

## Approach

**Routing refactor (commits 1-2):** The monolithic server handler is split into composable Hono routers. A `TenantContext` interface carries all per-request dependencies (DB, entity manager, stream client, principal). The OSS server builds this context once per request; library-mode callers can build their own and call `globalRouter.fetch(request, ctx)` directly. Pull-wake runners are a new first-class concept — remote worker processes that register with the server, tail a wake stream, and claim work through a structured REST protocol.

**Principals (commits 3-5):** A `Principal` is `{kind}:{id}` where kind is one of `user|agent|service|system`. The server resolves the principal from the trusted `electric-principal` header (set by an upstream auth proxy), with a dev-mode fallback. The principal is stored as `created_by` on entity creation and propagated through both webhook and pull-wake dispatch paths into the runtime's `ctx.principal`.

**Key invariant:** `TenantContext.principal` is non-optional — by the time any router handler runs, authentication has already succeeded. This prevents "forgot to check auth" bugs in downstream handlers.

## Key files

| File | What changed |
|------|-------------|
| `agents-server/src/principal.ts` | New — principal model, parsing, validation, `principalFromCreatedBy` |
| `agents-server/src/routing/dispatch-policy.ts` | New — dispatch policy resolution, auth assertions, subscription linking |
| `agents-server/src/routing/runners-router.ts` | New — runner REST API (register, claim, heartbeat, enable/disable) |
| `agents-server/src/routing/context.ts` | `TenantContext` with required `principal` field |
| `agents-server/src/entity-registry.ts` | Runner CRUD, dispatch state management, DB schema additions |
| `agents-server/src/entity-manager.ts` | Principal propagation in spawn/send/wake enrichment |
| `agents-server/src/stream-client.ts` | Subscription stream management, claim-scoped tokens |
| `agents-runtime/src/pull-wake-runner.ts` | New — pull-wake polling loop with heartbeat and claim dispatch |
| `agents-runtime/src/process-wake.ts` | Thread `ctx.principal` from notification into handler context |
| `agents-runtime/src/types.ts` | `RuntimePrincipal` type, `HandlerContext.principal` |
| `agents-server-ui/src/lib/assertedIdentity.ts` | New — desktop dev-mode identity for asserted auth |
| `agents/src/server.ts` | Built-in agents server converted to pull-wake dispatch |
| `agents-server/drizzle/0005_*.sql` | Runner and dispatch state tables |
| `agents-server/drizzle/0006_*.sql` | `created_by` column on entities |

## Non-goals

- **Expression language for policy constraints** — deferred to a follow-up. Policies currently support owner-scoped runner dispatch but not arbitrary claim predicates.
- **Backwards compatibility for unauthenticated requests** — agents are pre-release, so all routes require a principal (with dev-mode fallback).

## Verification

```bash
# Type check
cd packages/agents-server && pnpm exec tsc --noEmit

# Run agents-server tests
cd packages/agents-server && pnpm vitest run --pool-options.threads.maxThreads=2

# Run agents-runtime tests  
cd packages/agents-runtime && pnpm vitest run --pool-options.threads.maxThreads=2

# Run agents tests
cd packages/agents && pnpm vitest run --pool-options.threads.maxThreads=2
```


packages/agents-server/src/entity-manager.ts
```diff
@@ -328,6 +375,17 @@ export class EntityManager {
     typeName: string,
     req: TypedSpawnRequest
   ): Promise<ElectricAgentsEntity & { txid: number }> {
+    if (
+      typeName === `principal` &&
+      req.created_by !== `/${typeName}/${req.instance_id}`
```
icehaunter: why interpolation here if this can only be `principal` by the first part of the `&&`?

packages/agents-server/src/principal.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export function parsePrincipalKey(input: string): Principal {
+  const colon = input.indexOf(`:`)
+  if (colon <= 0) throw new Error(`Invalid principal key`)
+  const kind = input.slice(0, colon) as PrincipalKind
+  const id = input.slice(colon + 1)
+  if (!PRINCIPAL_KINDS.has(kind)) throw new Error(`Invalid principal kind`)
+  if (!id || id.includes(`/`)) throw new Error(`Invalid principal id`)
+  const key = `${kind}:${id}`
+  return { kind, id, key, url: `/principal/${key}` }
```
icehaunter: Given that this url will eventually be a part of an actual DS url, are we OK allowing `@` from emails? Not sure which part handles URL-encoding and what is the canonical form (same for non-ASCII chars)

packages/agents-server/src/principal.ts
```diff
@@ -0,0 +1,114 @@
...
+  const principal = parsePrincipalKey(key)
+  return {
+    url: principal.url,
+    key: principal.key,
+    kind: principal.kind,
+    id: principal.id,
+  }
+}
+
+export const principalIdentityStateSchema = {
```
icehaunter: This should be written using the typed schema thing we're using everywhere else

docs/principals-explainer.html
```diff
@@ -0,0 +1,468 @@
+<!doctype html>
```
icehaunter: Do we want this committed?

icehaunter (approved):
LTGM, mostly nitpicks apart from `@` handling because I think that needs addressing before a merge

PR #4330, opened 2026-05-13
feat(sync-service): record memory footprint on shape request spans

erik-the-implementer:
## Summary

Add process and binary memory-footprint attributes to the root `Plug_shape_get` span and its `shape_get.api.load_shape_info` child span. Each span records `memory.start.{process,binary}_bytes` on entry and `memory.end.{process,binary}_bytes` on exit.

- `process_bytes` — total memory occupied by the process (heap, stack, message queue, etc.)
- `binary_bytes` — sum of refc binary sizes referenced by the process (typically the dominant contributor in shape requests)

Memory capture is gated on the sampler decision — unsampled requests skip the `Process.info/2` calls entirely.

This makes memory evolution across a shape request queryable in Honeycomb and supports queries like "P99 memory usage per shape" by aggregating span attributes.

Refs electric-sql/stratovolt#1468.

## Test plan

- [x] `mix compile --warnings-as-errors` clean
- [x] `mix test test/electric/telemetry/open_telemetry_test.exs` — unit tests for the new helper
- [x] Existing span-emitting tests still pass

Generated with [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/.changeset/memory-footprint-span-attributes.md
+++ b/.changeset/memory-footprint-span-attributes.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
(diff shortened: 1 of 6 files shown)


robacourt (approved):
Nice!

PR #4334, opened 2026-05-14
fix: harden electric agents cli remote errors

icehaunter:
## Summary
- Harden electric-ax CLI error rendering so string-shaped server errors and empty status text still produce actionable messages.
- Add optional ELECTRIC_AGENTS_PRINCIPAL support that sends the Electric-Principal header for CLI, start-builtin, and the built-in agents entrypoint.
- Document the principal/header env vars and cover the new behavior with focused tests.

## Checks
- pnpm -C packages/electric-ax test -- --runInBand
- pnpm -C packages/electric-ax typecheck
- pnpm -C packages/agents test -- entrypoint-lib
- pnpm --filter electric-ax... -r build
- Remote smoke test: missing principal now prints Error: Missing electric-principal; setting ELECTRIC_AGENTS_PRINCIPAL lists entities successfully.


packages/electric-ax/src/index.ts
```diff
@@ -129,26 +129,45 @@ function parseElectricAgentsHeaders(
   return Object.keys(normalized).length > 0 ? normalized : undefined
 }
 
+function mergeElectricAgentsPrincipalHeader(
```
kevin-dp: This code is duplicated. If we have a shared package would be worth moving it there.

packages/electric-ax/src/index.ts
```diff
@@ -345,37 +364,59 @@ async function electricAgentsFetch(
...
+
+function fallbackResponseMessage(res: Response): string {
+  const status = res.status ? `HTTP ${res.status}` : `HTTP error`
+  return res.statusText ? `${status}: ${res.statusText}` : status
+}
+
+function stringMessage(value: unknown): string | undefined {
+  if (typeof value !== `string`) return undefined
+  const trimmed = value.trim()
+  return trimmed || undefined
```
kevin-dp: This works because `value.trim()` may return the empty string which we don't want so we return `undefined` instead. But this is implicit and not obvious at first sight. I'd make it explicit:
```ts
if (typeof value !== `string`) return undefined
const trimmed = value.trim()
if (trimmed.length === 0) return undefined
return trimmed
```

packages/electric-ax/src/start.ts
```diff
@@ -170,6 +171,29 @@ function parseAdditionalServerHeaders(
   return Object.keys(normalized).length > 0 ? normalized : undefined
 }
 
+function mergePrincipalHeader(
```
kevin-dp: Third time this function is duplicated.

kevin-dp (approved):

Issue #3895, opened 2026-02-23
Bug: onError retry path has no backoff delay


KyleAMathews:
## Problem

When `onError` returns `{}` to retry after a 4xx error, `#start()` recurses immediately with zero delay and fresh backoff state — creating a tight loop that can generate thousands of requests/second.

The inner `createFetchWithBackoff` only handles 5xx/network errors. When a 4xx triggers the `onError` callback and the user returns `{}` to retry, the code jumps straight to `this.#started = false; await this.#start()` with no delay whatsoever.

## Expected behavior

Consecutive `onError` retries should apply exponential backoff (using the user's `backoffOptions`), with the first retry remaining immediate (since the user may have just refreshed an auth token in `onError`).

## Proposed fix

- Add a `#consecutiveOnErrorRetries` counter field on `ShapeStream`
- Reset to 0 on successful response (in both `#requestShapeLongPoll` and `#requestShapeSSE`)
- Before the retry in `#start`, if `onErrorAttempt > 0`: compute exponential backoff with full jitter using user's `backoffOptions`, honor `retry-after` header as floor, await the delay
- Import `parseRetryAfterHeader` from `./fetch` to read retry-after headers from `FetchError`

## Behavioral summary

| Scenario | Before | After |
|---|---|---|
| 401 → onError refreshes token → retry | Instant | Instant (1st), then backoff |
| Persistent 4xx + `onError: () => ({})` | Tight loop, 1000s req/s | Exponential backoff to 60s |
| Successful onError retry | N/A | Resets counter to 0 |

PR #4337, opened 2026-05-14
Add exponential backoff to onError retry path

KyleAMathews:
## Summary

Adds full-jitter exponential backoff to `ShapeStream`'s `onError` retry path, preventing tight retry loops when persistent errors (e.g., 400s, exhausted network retries) trigger repeated `onError` → retry cycles. Uses the stream's existing `backoffOptions` timings for consistency with the fetch layer.

## Root Cause

Non-429 4xx errors intentionally bypass the fetch-layer backoff so `onError` handlers can repair auth tokens or params before retrying. Network failures that exhaust fetch backoff also bubble up through `onError`. Previously, when the handler returned retry opts, the retry happened immediately with no delay — creating a tight loop that could hammer a server returning persistent errors.

## Approach

A new `#backoffOnErrorRetry` method applies the same full-jitter exponential backoff strategy used by the fetch layer (L5 in SPEC.md). Key design decisions:

- **Reuses `backoffOptions`**: No new configuration surface — inherits `initialDelay`, `multiplier`, `maxDelay` from the same options that configure fetch-level backoff
- **Abort-aware**: The delay promise resolves immediately on abort signal, with a post-backoff check that tears down the stream instead of retrying
- **Consolidated state**: Three separate private fields consolidated into a single `#onErrorBackoff` object

## Key Invariants

1. The consecutive error retry guard (50 max) is still the hard limit — backoff adds pacing, not a replacement
2. The counter resets on successful message batch OR accepted 204
3. Abort during backoff → immediate teardown, no leaked timers or listeners
4. First retry gets delay in `[0, initialDelay)`, subsequent retries escalate exponentially up to `maxDelay`

## Non-goals

- No separate `onErrorBackoffOptions` config — reusing `backoffOptions` avoids API surface growth
- No debug logging during backoff — decided against to avoid log noise

## Verification

```bash
cd packages/typescript-client
pnpm test --run  # requires Electric server on :3000
```

New tests:
- `backs off before retrying when onError returns retry options` — verifies timing with mocked `Math.random`
- `tears down promptly when aborted during onError backoff` — verifies abort-aware cleanup

Existing `onError` tests updated with fast `backoffOptions` (`initialDelay: 2ms`) to stay snappy.

## Files Changed

| File | Change |
|------|--------|
| `src/client.ts` | New `#backoffOnErrorRetry` method, `#onErrorBackoff` config object, abort-aware delay at L5 call site |
| `SPEC.md` | Updated L5 description, guard mechanism docs, and all loop-back line numbers |
| `test/stream.test.ts` | 2 new tests (backoff timing, abort-during-backoff) + fast backoff options on 4 existing tests |


alco:
Fixes https://github.com/electric-sql/electric/issues/3895 ?

alco (approved):

PR #4339, opened 2026-05-16
feat(agents): pull-wake runner health check, principal rename, and lifecycle hardening

KyleAMathews:
## Summary

Adds pull-wake runner health diagnostics, renames runner ownership from `owner_user_id` to canonical `owner_principal` URLs, and hardens the pull-wake runner lifecycle around heartbeats, reconnects, shutdown, and error reporting.

This PR also fixes the local desktop send path so mutating local server requests can go through the Electron main process instead of being blocked behind Chromium's HTTP/1.1 connection limit, while keeping normal shape reads in the renderer.

## Root Cause

Pull-wake dispatch had too little observability and too many implicit lifecycle assumptions. When a runner missed wakes, got stuck reconnecting, or failed during shutdown, the server/UI had limited information to explain the state. The ownership field also used `owner_user_id`, which was misleading because principals can be users, agents, services, or system actors.

Local development exposed a separate but related issue: renderer-origin mutating requests could stall behind long-lived shape/SSE requests, delaying sends and new agent startup by seconds even when the server processed the request in milliseconds.

### Runner Health Diagnostics

- `PullWakeRunner` reports stream, heartbeat, claim, dispatch, reconnect, and error diagnostics in heartbeat payloads.
- Runtime diagnostics are stored separately from the normal runner row so the regular runners shape stays stable.
- `GET /_electric/runners/:id/health` aggregates runner state, liveness lease, sanitized client diagnostics, active claims, dispatch stats, and derived health issues.
- Health output tolerates partial/bogus diagnostics and invalid lease timestamps without producing misleading typed output.

### Principal Ownership

- Runner ownership is stored as canonical principal URLs in `owner_principal`.
- API boundaries canonicalize owner principals before storing/comparing.
- `electric-principal` request headers now accept either a principal key (`user:alice`) or a principal URL (`/principal/user%3Aalice`) through `parsePrincipalInput()` and compare internally by canonical URL.
- The migration expires active runner claims, clears runner dispatch state, deletes existing runner rows, and renames the ownership column.

### Pull-Wake Runner Lifecycle

- Adds public runner status diagnostics: `stopped | starting | connecting | streaming | reconnecting | stopping`.
- Makes `onError` reporting-only and isolates throwing reporters with a `console.error` fallback.
- `stop()` gates new claim dispatch, aborts outstanding work, waits boundedly for claim actors, drains runtime wakes, and rethrows drain errors after recording diagnostics.
- `waitForStopped()` now waits for the full stop path, not just the stream loop.
- Heartbeats are single-flight/coalesced, event-driven heartbeat scheduling uses the event throttle knob, and per-run heartbeat/reconnect counters reset on `start()`.
- Reconnects use exponential backoff and heartbeat failures can request a stream reconnect.
- Removed `maxConcurrentClaims`: it only limited claim/enqueue work, not actual wake execution, so it was a misleading invariant.

### Dispatch And Local Send Performance

- Dispatch subscription relinking is cached after the first successful link instead of doing Durable Streams round trips on every send.
- Local Electron mutating requests can be routed through the desktop main process for saved local servers, avoiding browser connection-limit stalls.
- Local `/send` remains normal `application/json`; the preflight/connection issue is handled by the desktop fetch path rather than changing API semantics.
- Ngrok warning-skip headers are only added for known ngrok hosts.

## Key Invariants

- Stored `owner_principal` values are canonical principal URLs.
- Runner auth compares canonical URL to canonical URL.
- `onError` cannot control runner lifecycle.
- Stop always gates late dispatch before runtime drain and reports drain failures to callers.
- Heartbeat health state is updated by one in-flight heartbeat at a time.
- Dispatch subscription linking is idempotent and cached per stream client/service/subscription/stream.

## Non-goals

- No broad terminology rename in this PR.
- No full module split of `pull-wake-runner.ts`; the critical invariants are fixed and tested first.
- No persistent claim-failure circuit breaker; failures are visible in diagnostics and can be escalated separately.
- No cross-principal runner listing/admin surface.

## Verification

```bash
pnpm --dir packages/agents-runtime exec vitest run test/pull-wake-runner.test.ts --reporter=dot
pnpm --dir packages/agents-server exec vitest run test/principal.test.ts test/runners-router.test.ts test/dispatch-policy-routing.test.ts --reporter=dot
pnpm --dir packages/agents-server-ui exec vitest run src/lib/auth-fetch.test.ts --reporter=dot

pnpm --dir packages/agents-runtime typecheck
pnpm --dir packages/agents-server typecheck
pnpm --dir packages/agents-server-ui typecheck
pnpm --dir packages/agents-desktop typecheck

pnpm --dir packages/agents-server-ui build
pnpm --dir packages/agents-desktop build
```


packages/agents-runtime/src/pull-wake-runner.ts
```diff
@@ -51,23 +53,93 @@ export interface PullWakeRunner {
...
 ): PullWakeRunner {
+  let state: PullWakeRunnerState = `stopped`
   let controller: AbortController | null = null
   let loop: Promise<void> | null = null
   let response: PullWakeStreamResponse | null = null
   let heartbeatTimer: ReturnType<typeof setInterval> | null = null
+  let eventHeartbeatTimer: ReturnType<typeof setTimeout> | null = null
   let currentOffset = config.offset
+  let startedAt: string | null = null
+  let streamConnected = false
```
kevin-dp, line 114: I would define `streamConnected` as a getter:
```ts
get streamConnected() {
  return !!streamConnectedSince
}
```

packages/agents-runtime/src/pull-wake-runner.ts
```diff
@@ -51,23 +53,93 @@ export interface PullWakeRunner {
...
+  reconnect_count: number
+  last_error: string | null
+  last_error_at: string | null
+  last_heartbeat_at: string | null
+  last_heartbeat_ok: boolean
+  last_claim_at: string | null
+  last_claim_result: `claimed` | `no_work` | `error` | null
+  last_dispatch_at: string | null
+  events_received: number
+  claims_succeeded: number
```
kevin-dp, line 82: Perhaps this could be a nested property:
```ts
export interface PullWakeRunnerHealth {
  // ...
  claims: {
    succeeded: number
    skipped: number
    failed: number
  }
}
```

Not sure how useful these numbers are. Perhaps an array of the actual claims that succeeded/skipped/failed would be better.

packages/agents-runtime/src/pull-wake-runner.ts
```diff
@@ -51,23 +53,93 @@ export interface PullWakeRunner {
...
+  claims_skipped: number
+  claims_failed: number
+}
+
+type PullWakeRunnerState =
+  | `stopped`
+  | `starting`
+  | `running.connecting`
```
kevin-dp, line 90: The `running.` prefixes are a bit unusual.

kevin-dp: Would this benefit from being an actual TS enum?
```ts
enum PullWakeRunnerState {
  Stopped = `stopped`,
  Starting = `starting`,
  Connecting = `running.connecting`,
  Streaming = `running.streaming`,
  Reconnecting = `running.reconnecting`,
  Stopping = `stopping`
}
```
This would make the code less string-heavy.

Or if you don't like enums:
```ts
const PullWakeRunnerState = {
  Stopped: 'stopped',
  Starting: 'starting',
  Connecting: 'running.connecting',
  Streaming: `running.streaming`,
  Reconnecting: `running.reconnecting`,
  Stopping: `stopping`
} as const

type PullWakeRunnerState = (typeof PullWakeRunnerState)[keyof typeof PullWakeRunnerState]
```

packages/agents-runtime/src/pull-wake-runner.ts
```diff
@@ -77,6 +149,49 @@ export function createPullWakeRunner(
     config.claimPath ??
     `/_electric/runners/${encodeURIComponent(config.runnerId)}/claim`
   const claimUrl = appendPathToUrl(config.baseUrl, claimPath)
+  const maxConcurrentClaims = Math.max(
+    1,
+    Math.floor(config.maxConcurrentClaims ?? DEFAULT_MAX_CONCURRENT_CLAIMS)
+  )
+
+  const toStatus = (): PullWakeRunnerStatus => {
```
kevin-dp, line 155: Why have "state" and "status" ? They are almost identical.
I would just go with the enum approach from my previous comment.
Users use it like `PullWakeRunnerState.Connecting` and it resolves to the string `"running.connecting"` so we effectively get the usage we want.

packages/agents-runtime/src/pull-wake-runner.ts
```diff
@@ -100,7 +215,24 @@ export function createPullWakeRunner(
...
+    } catch (reporterError) {
+      // onError is reporting-only; reporters must not control runner lifecycle.
+      console.error(`Pull-wake runner onError callback failed`, reporterError)
+    }
+  }
+
+  const notifyHeartbeatChange = (): void => {
+    const signal = controller?.signal
+    if (!signal || signal.aborted || heartbeatIntervalMs <= 0) return
+    if (eventHeartbeatTimer) return
```
kevin-dp, line 235: nit: we have 2 if statements for fast return, to be consistent the first one would only do the signal checks and the 2nd one would do heartbeat checks, or we would use only one if statement:
```ts
if (!signal || signal.aborted) return
if (heartbeatIntervalMs <= 0 || eventHeartbeatTimer) return
```

packages/agents-runtime/src/pull-wake-runner.ts
```diff
@@ -100,7 +215,24 @@ export function createPullWakeRunner(
...
+      config.onError?.(error)
+    } catch (reporterError) {
+      // onError is reporting-only; reporters must not control runner lifecycle.
+      console.error(`Pull-wake runner onError callback failed`, reporterError)
+    }
+  }
+
+  const notifyHeartbeatChange = (): void => {
+    const signal = controller?.signal
+    if (!signal || signal.aborted || heartbeatIntervalMs <= 0) return
```
kevin-dp: Why do we check `heartbeatIntervalMs <= 0` if it's not used in this function?
This function uses `eventHeartbeatThrottleMs` instead.

packages/agents-runtime/src/pull-wake-runner.ts
kevin-dp: We could also avoid the `eventHeartbeatTimer` check by rewriting the assignment:
```ts
eventHeartbeatTimer ??= setTimeout(() => {
  eventHeartbeatTimer = null
  void heartbeat(signal)
}, eventHeartbeatThrottleMs)
```

packages/agents-runtime/src/pull-wake-runner.ts
```diff
@@ -115,17 +247,22 @@ export function createPullWakeRunner(
...
+        lastHeartbeatOk = false
         throw new Error(
           `Pull-wake runner heartbeat failed for ${config.runnerId}: ${res.status} ${await res.text()}`
         )
       }
+      lastHeartbeatOk = true
     } catch (err) {
       if (!signal.aborted) {
-        config.onError?.(err instanceof Error ? err : new Error(String(err)))
+        lastHeartbeatOk = false
```
kevin-dp, line 282: We already set `lastHeartbeatOk = false` on L256 right before throwing and here we only set it inside `if (!signal.aborted)` but it will already be `false`. Is this the behaviour we want? If we only want it to be set to false if `!signal.aborted` then we need to not set it right before throwing.

packages/agents-runtime/src/pull-wake-runner.ts
```diff
@@ -161,107 +302,324 @@ export function createPullWakeRunner(
...
+      lastClaimResult = `claimed`
+      claimsSucceeded++
+      notifyHeartbeatChange()
+      return notification
+    } catch (err) {
+      if (signal.aborted) {
+        throw err
+      }
+      if (!claimErrorRecorded) {
+        recordClaimError()
```
kevin-dp, line 397: Do we need to set `claimErrorRecorded = true` here?
Perhaps would be better if `recordClaimError` does that (e.g. we could wrap it such that we can't forget to set this variable).

packages/agents-runtime/src/pull-wake-runner.ts
```diff
@@ -161,107 +302,324 @@ export function createPullWakeRunner(
...
+          clearTimeout(timer)
+          resolve(`settled`)
+        })
       })
+      if (result === `timeout`) return false
+    }
+    return true
+  }
+
+  const sleep = async (ms: number, signal: AbortSignal): Promise<void> => {
```
kevin-dp, line 462: This should be extracted to a utility file.

kevin-dp (commented):
I had a look through the code and left some comments.
On a higher level, i don't like the "claims" name, it collides too much with auth claims.
**The term "lease" is more standard** in distributed systems so I would rename the claims variables to: lease / activeLeaseCount / maxConcurrentLeases / getActiveLeasesForRunner.

Also, i think the code needs quite some refactoring. Currently there are ~20 variables inside one closure and they are being mutated from everywhere in this file. That's very error-prone. I'll try refactoring in a follow up PR.

kevin-dp:
Spent some time testing the lifecycle hardening claims in this PR against a running desktop app + local agents-server. Some of what's claimed works as advertised; one significant gap turned up that I think is worth flagging before merge.

## What worked

- ✅ `GET /_electric/runners/:id/health` endpoint returns the documented shape; UI and curl agree.
- ✅ Diagnostic counters update on dispatch within ~2s — event-driven heartbeat throttle path is firing as designed (`last_hb` jumped off the 30s schedule into a fresh slot ~2s after a `last_dispatch_at` change, consistent with the 2s throttle).
- ✅ Health status escalation (healthy → degraded → unhealthy) computes correctly on the server side.
- ✅ Principal rename (`owner_user_id` → `owner_principal`) and URL validation work end-to-end (after fixing one default-value alignment issue between desktop and server, which I've already pushed on `fix-pull-wake`).

## Things turned up while testing

While testing this PR I also surfaced three related reliability issues that the new diagnostics made visible — they aren't introduced by this PR (the diagnostics just exposed them), but they all interact with this work:

- #4340 — active claim is not released after dispatch (success or failure)
- #4341 — `lease_expires_at: null` on materialized claims removes the safety net for #4340
- #4342 — Local Runtime UI shows stale/contradictory state when agents-server is unreachable (compounded by #4343 since the UI reads exactly the `stream_connected: true` field that this bug pins to true)

## Suggestion

`#4343` feels like it should block merge — the headline reliability claims in the PR description don't hold under the testing I did. The four issues together are also somewhat overlapping (the UI staleness in #4342 wouldn't be nearly as bad if #4343 was fixed, because the `stream_connected` field would actually flip during the outage), so fixing #4343 has some leverage on the others.

Happy to help test once you've got a candidate fix.

kevin-dp (changes requested):
~~We should address https://github.com/electric-sql/electric/pull/4339#issuecomment-4476720088 before merging.~~

packages/agents-server/docs/superpowers/specs/2026-05-16-pull-wake-health-check-design.md
```diff
@@ -0,0 +1,214 @@
+# Pull-Wake Runner Health Check
+
+## Problem
+
+The pull-wake dispatch system is unreliable and there's no way to diagnose what's going wrong. We need:
+
+1. A **health check endpoint** (`GET /_electric/runners/:id/health`) for deep debugging — curl it to see comprehensive diagnostics about a runner's dispatch pipeline.
+2. **Rich runner state in Postgres** so that apps can sync the `runners` table via an Electric Shape and show runner status on any device (e.g. see your laptop runner's status from your phone).
+
+The `diagnostics` JSONB column on the `runners` table serves both purposes: the health endpoint reads it for the detailed response, and Shape sync delivers it reactively to any connected client.
```
icehaunter: This will make `runners` into a quite noisy shape, with every runner causing an update every 2 seconds. I suggest we split it into a separate table for heartbeats & claims, and keep `status`, `started`, `last_error` & `last_error_at` in the main table. Anyone needing full diagnostics can then either get the endpoint, or subscribe to a filtered shape for diagnostics of a particular runner. WDYT?

icehaunter (changes requested):

packages/agents-server/docs/superpowers/specs/2026-05-16-pull-wake-health-check-design.md
KyleAMathews: It *should* be a max of every two seconds (if there's lots of activity) and then when things aren't happening, updates would drop to every 30 seconds.

But yeah, fairly noisy I agree actually still. Lemme think about it.

icehaunter: We don't need the entire object on that table for "normal" UIs I think. the heartbeat endpoint can split this up easily

KyleAMathews: yeah makes sense — refactored

KyleAMathews:
Thanks for the reviews/testing. I went back through the top-level comments, inline threads, and linked issues. Current status of what has been fixed in this branch:

- Split noisy runner runtime diagnostics out of the main `runners` shape into `runner_runtime_diagnostics`, and updated the Local Runtime UI to subscribe to the filtered diagnostics shape.
- Added server-side diagnostics sanitization before storing heartbeat diagnostics and before returning health payloads.
- Hardened runner lifecycle:
  - `onError` is reporting-only and isolated from runner control flow.
  - `stop()` no longer hangs forever on stuck claim actors.
  - concurrent `stop()` calls share one shutdown sequence.
  - timed-out/stale claim actors no longer poison future claim capacity.
  - claim-capacity abort listeners are cleaned up.
- Added bounded claim concurrency and tests around shutdown/capacity behavior.
- Avoided extra event-heartbeat scheduling when the stream offset has not changed.
- Fixed duplicate dispatch subscription linking: re-linking an already-linked entity stream is now a no-op instead of issuing another add request.
- Removed the legacy runner-subscription cleanup path. No backwards-compat cleanup here.
- Fixed local desktop/dev auth alignment: desktop now defaults the local runner owner to `/principal/system%3Adev-local`, so `pnpm dev` works without `ELECTRIC_DESKTOP_PRINCIPAL="system:dev-local"`.
- Merged latest `origin/main`, pushed, and reran builds for touched packages.

Verification run locally:

- Focused runner tests
- Focused server routing/dispatch/principal tests
- `agents-server` typecheck/stylecheck/build
- `agents-runtime` typecheck/stylecheck
- `agents-desktop` typecheck/build
- changeset check
- `git diff --check`

packages/agents-server/src/routing/durable-streams-routing-adapter.ts
```diff
@@ -32,14 +31,36 @@ function appendRequestPathToStreamRoot(input: DurableStreamsRoutingInput): URL {
...
+  const match = /^(.*)\/v1\/stream(?:\/(.+))?\/?$/.exec(
+    new URL(input.durableStreamsUrl).pathname
+  )
+  if (!match) {
+    return appendRequestPathToStreamRoot(input)
+  }
+
+  const [, prefix = ``, serviceId] = match
+  const target = new URL(input.durableStreamsUrl)
+  target.pathname = `${prefix}/v1/stream-meta${incomingUrl.pathname.replace(/^\/__ds/, ``)}`
```
icehaunter: @KyleAMathews why did you undo my work here?

kevin-dp (approved):

icehaunter (approved):

Issue #4341, opened 2026-05-18
Pull-wake: heartbeat path nulls out lease_expires_at on consumer_claims


kevin-dp:
## Summary

The `lease_expires_at` column on `consumer_claims` is nulled out by the per-wake heartbeat path, leaving claims without any lease for the remainder of their lifetime. After the first heartbeat (~10s after dispatch), `lease_expires_at` becomes `null` and stays that way — meaning any lease-based reaper or expiry check cannot recover the row if `done` never arrives.

## Reproduction

Send a wake to a pull-wake runner and inspect the active claim via the health endpoint.

**At claim materialization** (`materializeActiveClaim` in `entity-registry.ts`):

```json
{
  "consumer_id": "w_ccc537e97ee8110f165930d8",
  "epoch": 1,
  "claimed_at":         "2026-05-19T11:01:11.631Z",
  "last_heartbeat_at":  null,
  "lease_expires_at":   "2026-05-19T11:01:41.631Z"
}
```

`lease_expires_at = claimed_at + 30s`. Lease is set correctly from the upstream durable-streams `lease_ttl_ms`.

**~20 seconds later**, after the runtime's first per-wake heartbeat:

```json
{
  "consumer_id": "w_ccc537e97ee8110f165930d8",
  "epoch": 1,
  "claimed_at":         "2026-05-19T11:01:11.631Z",
  "last_heartbeat_at":  "2026-05-19T11:01:31.684Z",
  "lease_expires_at":   null
}
```

`lease_expires_at` is now `null`. From this point on, no further updates ever restore the lease.

## Root cause

`packages/agents-server/src/entity-registry.ts` — `materializeHeartbeatClaim`:

```ts
async materializeHeartbeatClaim(input: MaterializeHeartbeatClaimInput): Promise<void> {
  const heartbeatAt = input.heartbeatAt ?? new Date()
  await this.db
    .update(consumerClaims)
    .set({
      lastHeartbeatAt: heartbeatAt,
      leaseExpiresAt: input.leaseExpiresAt ?? null,  // ← unconditionally writes null if not provided
      updatedAt: heartbeatAt,
    })
    .where(...)
}
```

`packages/agents-server/src/routing/internal-router.ts:606-609` — the only production caller:

```ts
await ctx.entityManager.registry.materializeHeartbeatClaim?.({
  consumerId,
  epoch,
  // no leaseExpiresAt is passed
})
```

[...]

PR #4346, opened 2026-05-18
fix(agents-server): release pull-wake claim row even when in-memory token is missing

### LINKED ISSUE #4340: Pull-wake: active claim leaks when in-memory write-token is missing (server restart, newer-wake eviction, retry)
## Summary

The pull-wake claim release path in `callback-forward` (`packages/agents-server/src/routing/internal-router.ts`) gates `materializeReleasedClaim` on the **in-memory** `ClaimWriteTokenStore`. When the in-memory token is missing or has been evicted, the `consumer_claims` row is never marked `released`, the entity status is never transitioned back from `running` to `idle`, and the row leaks indefinitely.

Surfaced while testing #4339 (pull-wake health diagnostics). The new `GET /_electric/runners/:id/health` endpoint exposes per-claim `active` rows, which made the leak observable.

The steady-state "send one message, wait" path is not affected by this bug: it releases correctly via the runtime's `sendDone` after the `idleTimeout` window (default 5 min via `packages/agents/src/bootstrap.ts:146`). The bug only fires in the failure modes documented below.

## Failure modes

The release path that genuinely *doesn't run* on the unfixed code is the one where the in-memory `ClaimWriteTokenStore` no longer holds the consumer's token at the time `done` arrives. Three scenarios reproduce this deterministically (covered by unit tests in `packages/agents-server/test/webhook-forward-routing.test.ts`):

1. **Server restart between mint and done.** The runtime calls `callback-forward` to mint a token, then runs the agent. If the agents-server restarts during the run, the in-memory store loses all tokens. When `sendDone` arrives, the lookup returns `stillOwnsClaim: false`, and the entire release block is skipped: no `materializeReleasedClaim`, no `updateStatus(idle)`. The DB row stays `active` forever.
2. **Newer wake evicts the in-memory token.** Two wakes for the same entity stream arrive close together. The second wake's `mint()` call evicts the first wake's token from the in-memory map (this is `ClaimWriteTokenStore`'s designed eviction behavior for stream→consumer mapping). When the *first* wake later calls `sendDone`, `stillOwnsClaim` is `false` for it — same skip path.
3. **Retry after `updateStatus` failure.** First done attempt successfully releases the row but `updateStatus(entity, 'idle')` throws (transient DB error, etc.). The runtime retries the done call. On retry, the row is already released, so the `entityCleared` flag is `false`; on the unfixed code the path then skips `updateStatus(idle)` again, leaving the entity stuck at `running`.

## Root cause

`internal-router.ts:611-665` in `callbackForward` has all three release actions behind a single gate:

```ts
if (entity && stillOwnsClaim) {
  await materializeReleasedClaim(...)
  await updateStatus(entity.url, 'idle')
  clearStream(...)
  await onEntityChanged(entity.url)
} else if (stillOwnsClaim) {
  clearStream(...)
} else if (entity) {
  log.info('done ignored for stale claim ...')
}
```

The `stillOwnsClaim` check is appropriate for **write authorization** (you should only be able to write to the entity's main stream if you hold the active claim), but it's the wrong gate for **claim release**, which is a DB-keyed operation on `(consumerId, epoch)`. Releasing the DB row should not require in-memory state — the DB primary key is authoritative identity.

## Suggested fix

Decouple the three concerns. Each gets its own gate:

- `materializeReleasedClaim` — runs whenever `epoch` is defined. DB identity.
- `updateStatus(entity, 'idle')` + `onEntityChanged` — runs when *either* our release just cleared the entity's active dispatch row *or* we still hold the in-memory token. The first handles server-restart + retry; the second handles the legacy "no active dispatch row exists" case in `server-claim-write-token.test.ts`.
- `clearStream` — remains gated by `stillOwnsClaim` so we never clear a newer consumer's in-memory token.

The "either" in the entity-status gate requires `materializeReleasedClaim` to report a new `entityCleared` flag indicating whether the active dispatch row was just cleared (versus a no-op because a newer wake had already taken over). Added in the registry method.

## Severity

Low-to-medium for the happy path (idle timeout naturally clears state). Higher for:
- Long-running servers that get restarted under load (claim rows accumulate, eventually exhaust `maxConcurrentClaims` for the runner)
- Entities receiving frequent wakes where the token-eviction race is likely
- Any deployment where transient DB errors during `updateStatus` aren't rare

Visibility is poor regardless — without #4339's diagnostics endpoint, operators have no way to see the orphaned rows.

## PR

Implemented in #4346 (against `fix-pull-wake`). Unit tests demonstrate all three failure modes pre-fix and pass after the fix.

## Related

- #4341 — `lease_expires_at: null` on materialized claims removes the safety net that would otherwise auto-expire abandoned rows.
- #4339 — PR whose diagnostics surfaced this issue.

kevin-dp:
## Summary

Fixes #4340 — pull-wake claims leaking in `consumer_claims` when the in-memory `ClaimWriteTokenStore` no longer holds the consumer's token at the time `sendDone` arrives.

The release path in `callback-forward` was gated by `stillOwnsClaim`, an in-memory check. When that check fails — server restart between mint and done, parallel wakes evicting each other's tokens, or a retry after a transient `updateStatus` failure — the entire release block is skipped: `materializeReleasedClaim` never runs, the entity stays stuck at `status='running'`, and the `consumer_claims` row stays `active` indefinitely.

The steady-state "send one message, wait" path is not affected by this bug: it releases correctly via the runtime's `sendDone` after the `idleTimeout` window (default 5 min via `packages/agents/src/bootstrap.ts:146`). The bug only fires in the failure modes documented in [Test scenarios](#test-scenarios) below.

## Root cause

`packages/agents-server/src/routing/internal-router.ts` had all three release actions behind the same in-memory gate:

```ts
if (entity && stillOwnsClaim) {
  await materializeReleasedClaim(...)   // DB row release
  await updateStatus(entity.url, 'idle') // entity status
  clearStream(...)                       // in-memory token cleanup
  await onEntityChanged(entity.url)
} else if (stillOwnsClaim) {
  clearStream(...)
} else if (entity) {
  log.info('done ignored for stale claim ...')
}
```

`stillOwnsClaim` is the right gate for **write authorization** during the agent run, but it's the wrong gate for releasing the **DB row**, which is keyed by `(consumerId, epoch)`. The DB primary key is authoritative identity — the in-memory token state is orthogonal.

## The fix

Three concerns, three gates:

1. **DB row release** (`materializeReleasedClaim`) — runs whenever `epoch` is defined. `(consumerId, epoch)` is the DB primary key; that's enough.
2. **Entity status → idle + `onEntityChanged`** — runs when `entityCleared || stillOwnsClaim`. `entityCleared` is a new return field from `materializeReleasedClaim`, set to `true` only when our `(consumerId, epoch)` was the active dispatch row and we just nulled it out. The `||` handles two non-trivial cases: server restart (DB has us active, token is gone) and retry after a failed `updateStatus` (state cleared on first attempt, token still held).
3. **In-memory token cleanup** (`clearStream`) — remains gated by `stillOwnsClaim` so a newer consumer's token is never cleared out from under it.

### `materializeReleasedClaim` API change

```diff
- ): Promise<ConsumerClaim | null> {
+ ): Promise<{ claim: ConsumerClaim | null; entityCleared: boolean }> {
```

Only one production caller (`internal-router.ts`); both that caller and the test mock are updated. The `.returning()` on the `entityDispatchState` UPDATE now reports whether our row was actually cleared (vs. a no-op because a newer claim is active).

## Test scenarios

The fix decouples the three concerns. Below: ✓ = action happens, × = action does NOT happen (and is correct that it doesn't).

| Scenario | `entityCleared` | `stillOwnsClaim` | DB row released | Entity → idle |
|---|---|---|---|---|
| **A. Happy path** (mint + done) | true | true | ✓ released | ✓ goes idle |
| **B. Server restart** (no in-memory token, DB row still active) | true | false | ✓ released | ✓ goes idle |
| **C. Newer wake** (wake-1 done after wake-2 takes over the stream) | false | false | ✓ wake-1's row released | × stays running — wake-2 is in flight |
| **D. Retry** (first done's `updateStatus` threw; same done retried) | false | true | ✓ no-op (already released) | ✓ goes idle |
| **E. Legacy stale-done test** (test setup never materialized active claim; token evicted) | false | false | no row to release | × stays running — newer claim conceptually in flight |

New tests in `packages/agents-server/test/webhook-forward-routing.test.ts > claim release on done callback (regression for #4340)` cover scenarios A–C. Existing tests in `server-claim-write-token.test.ts` cover D and E and continue to pass after the fix.

## Verified

- **Unit tests**: deterministic. Pre-fix, B and C fail (zero invocations of `materializeReleasedClaim`); D fails (`updateStatus` skipped on retry). Post-fix, all five scenarios produce the documented behavior.
- **Manual run-through** against a local desktop + agents-server: send one message, dispatch claims (`active_count: 0 → 1`), agent completes, runtime calls `sendDone` after `idleTimeout`, server fires the new release path, `active_count: 1 → 0`, entity status transitions back to `idle`.

## Not addressed in this PR

- **Pre-existing orphan rows**: rows that already leaked from prior runs of the unfixed code can't be released because no fresh `done` callback is coming. Would need a reaper job or admin command.
- **`lease_expires_at: null` issue (#4341)**: independent. Without a lease, even a reaper job can't time-out claims safely.

## Base branch note

This PR targets `fix-pull-wake` (#4339), not `main`, because `materializeReleasedClaim` was introduced in #4308 which is part of the `fix-pull-wake` lineage but not yet in `main`. Merge order: this → fix-pull-wake → main.

### DIFF
--- a/.changeset/release-pull-wake-claim-after-dispatch.md
+++ b/.changeset/release-pull-wake-claim-after-dispatch.md
@@ -1,0 +1,5 @@
+---
+'@electric-ax/agents-server': patch
+---
+
+Fix pull-wake claims leaking in `consumer_claims` after dispatch. The release path in `callback-forward` was gated entirely on the in-memory write-token state, so any condition that lost or evicted the token (server restart, a newer wake on the same stream) would prevent `materializeReleasedClaim` from running and leave the DB row pinned at `status='active'`. The fix decouples the durable-row release (keyed by `consumerId + epoch`) from in-memory token cleanup, and uses `entityCleared || stillOwnsClaim` to gate the entity status transition back to `idle`. Includes regression tests in `test/webhook-forward-routing.test.ts`.
--- a/packages/agents-server/src/entity-registry.ts
+++ b/packages/agents-server/src/entity-registry.ts
@@ -378,7 +378,7 @@ export class PostgresRegistry {
 
   async materializeReleasedClaim(
     input: MaterializeReleasedClaimInput
-  ): Promise<ConsumerClaim | null> {
+  ): Promise<{ claim: ConsumerClaim | null; entityCleared: boolean }> {
     const releasedAt = input.releasedAt ?? new Date()
     const rows = await this.db
       .update(consumerClaims)
@@ -398,8 +398,13 @@ export class PostgresRegistry {
       .returning()
 
     const claim = rows[0] ? this.rowToConsumerClaim(rows[0]) : null
+    let entityCleared = false
     if (claim) {
-      await this.db
+      // entityCleared distinguishes "we were the active dispatch and now it's
+      // empty" from "a newer claim was already active for this entity." The
+      // WHERE clause matches our (consumerId, epoch) so an evicted-by-newer
+      // case correctly returns zero rows.
+      const cleared = await this.db
         .update(entityDispatchState)
         .set({
           activeConsumerId: null,
@@ -419,8 +424,10 @@ export class PostgresRegistry {
             eq(entityDispatchState.activeEpoch, input.epoch)
           )
         )
+        .returning({ entityUrl: entityDispatchState.entityUrl })
+      entityCleared = cleared.length > 0
     }
-    return claim
+    return { claim, entityCleared }
   }
 
   async getActiveClaimsForRunner(
--- a/packages/agents-server/src/routing/internal-router.ts
+++ b/packages/agents-server/src/routing/internal-router.ts
@@ -624,8 +624,15 @@ async function callbackForward(
       const entity = await ctx.entityManager.registry.getEntityByStream(
         target.primaryStream
       )
-      if (entity && stillOwnsClaim) {
-        if (epoch !== undefined) {
+
+      // Release the consumer_claims row by its DB identity (consumerId,
+      // epoch). The in-memory write token is a separate concern (write
+      // authorization during the run); release of the durable row must
+      // succeed even if the token was lost (server restart) or evicted
+      // (a later wake re-minted for the same stream).
+      let entityCleared = false
+      if (epoch !== undefined) {
+        const result =
           await ctx.entityManager.registry.materializeReleasedClaim?.({
             consumerId,
             epoch,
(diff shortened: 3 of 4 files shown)


msfstef (approved):

icehaunter (approved):

KyleAMathews (approved):
Looks good — the core fix is sound. Decoupling durable claim release from in-memory write-token ownership is the right design correction here, and preserving the stillOwnsClaim gate for in-memory cleanup protects newer wakes from being disturbed. The added router-level regression coverage for held-token, missing-token, and newer-wake-took-over scenarios is also useful.\n\nMinor non-blocking follow-ups worth considering:\n\n1. Add lower-level registry tests around materializeReleasedClaim/entityCleared semantics: entityCleared=true when the active dispatch matches (consumerId, epoch), false when a newer dispatch is active, and false when no matching claim row exists.\n2. Consider the edge case where the first done attempt successfully clears the DB dispatch state but fails before updateStatus('idle'), then the server restarts before a retry. On retry, both entityCleared and stillOwnsClaim may be false, so the entity status may remain running. This seems outside the main bug fixed here, but it may be worth tracking separately if that recovery path matters.\n\nApproving — no blocking concerns from my review.

PR #4356, opened 2026-05-19
fix(agents-server): apply routing adapter to subscription paths

### LINKED ISSUE #4357: agents-server: webhook dispatch dropped on service-routed durable-streams deployments
## Symptom

On service-prefixed / service-routed durable-streams deployments (e.g. stratovolt's `cloud-agents-server`, verified against `https://agents.electric-sql.cloud/?service=svc-...`):

1. Register an entity type with `default_dispatch_policy.targets = [{ type: 'webhook', url: 'https://<public-url>/wake' }]`.
2. `PUT /_electric/entities/<type>/<id>` → 201 (entity stored with the webhook dispatch policy).
3. `POST /_electric/entities/<type>/<id>/send` → 204 (inbox event appended).
4. **No HTTP POST ever lands on the webhook URL.**

Pull-wake dispatch against the same tenant works end-to-end, so this is webhook-specific. The public webhook URL is reachable; the cloud just never calls it.

## Root cause

Two-tier path-namespace mismatch on the subscription path that doesn't exist on the data path:

- **Stream appends** are routed through the `DurableStreamsRoutingAdapter` (e.g. `serviceRoutedDurableStreamsAdapter` on the cloud), so they land in the worker keyed on the service-prefixed backend path (`<serviceId>/<type>/<id>/main`).
- **Subscription registrations** went through `StreamClient.backendSubscriptionPath` (`packages/agents-server/src/stream-client.ts:222–224`), which only slash-normalised. So the subscription's `streams` (and `pattern`, `wake_stream`, ack/release `stream`/`path` fields) were stored as the un-prefixed path (`<type>/<id>/main`).

The worker's HotBuffer lookup keyed on the append's resource id never matched the subscription link, and webhook fanout silently dropped.

Pull-wake worked only because the runner writes a wake notification directly to its own `wake_stream` and reads it back — no HotBuffer fanout coordination needed, so the path mismatch never mattered.

## Suggested fix

`StreamClient` should accept the `DurableStreamsRoutingAdapter` (and a `serviceId`) and use `toBackendStreamPath` / `toRuntimeStreamPath` on subscription payloads — symmetrically with how the per-request `durable-streams-router.ts` rewrites stream URLs. `AgentsHostTenantConfig` should gain a matching option so cloud callers can pass the adapter at tenant-registration time.

PR: #4356

## Reproduction

End-to-end repro lives in https://github.com/electric-sql/OpenFactory `packages/discord-bot` — registering a `discord-bot` entity type with a webhook dispatch target reproduces 100% on the hosted cloud tenant. Switching the same code to pull-wake (`{ type: 'runner', runnerId }`) works.

Originally filed at electric-sql/stratovolt#1519.

balegas:
## Summary

- Plumb the `DurableStreamsRoutingAdapter` through `StreamClient` so subscription path payloads (`pattern`, `streams`, `wake_stream`, ack/release `stream`/`path`) use `toBackendStreamPath`, and worker responses are mapped back via `toRuntimeStreamPath`.
- Add a matching `durableStreamsRouting` field to `AgentsHostTenantConfig` that flows the adapter through to the per-tenant `StreamClient` constructor.
- `dispatch-policy.subscriptionHasStream` now consults `ctx.durableStreamsRouting` instead of hardcoding `${ctx.service}/${path}`, so adapters with non-trivial backend conventions are honoured end-to-end.
- Default (no routing option) behaviour is unchanged — tenant-root deployments are bit-identical to the prior slash-normalising code.

## Why

Reported against the hosted cloud tenant (stratovolt's `cloud-agents-server`): webhook dispatch never fires for tenanted entities, while pull-wake works. Root cause is a two-tier path-namespace mismatch on the subscription path that doesn't exist on the data path:

- **Stream appends** are routed through the cloud's `serviceRoutedDurableStreamsAdapter`, so they land in the durable-streams worker keyed on the service-prefixed backend path (`<serviceId>/<type>/<id>/main`).
- **Subscription registrations** went through `StreamClient.backendSubscriptionPath`, which only slash-normalised. So the subscription's `streams` were stored as the un-prefixed path (`<type>/<id>/main`).

The worker's HotBuffer lookup keyed on the append's resource id never matched the subscription link, and webhook fanout silently dropped. Pull-wake worked only because the runner writes a wake notification directly to its own `wake_stream` and reads it back — no HotBuffer fanout coordination needed.

The cloud-side companion change (wire `durableStreamsRouting: serviceRoutedDurableStreamsAdapter` into `agentsHost.registerTenant`) lives in electric-sql/stratovolt#1521 and ships behind a dep bump.

## Changes

- `src/stream-client.ts` — `StreamClientOptions.subscriptionRouting?: { serviceId, adapter }`; `backendSubscriptionPath` / `runtimeSubscriptionPath` consult the adapter when set, applying `toBackendStreamPath` / `toRuntimeStreamPath` then re-normalising. This single change covers `putSubscription`, `addSubscriptionStreams`, the ack/release `subscriptionRequestBody` mapping, and the symmetric `subscriptionResponseBody` reverse-mapping.
- `src/host.ts` — `AgentsHostTenantConfig.durableStreamsRouting?: DurableStreamsRoutingAdapter`; threaded into `createStreamClient`.
- `src/index.ts` — re-exports `StreamClientSubscriptionRouting`.
- `src/routing/durable-streams-routing-adapter.ts` — documents idempotence + no-throw contract on `toBackendStreamPath` / `toRuntimeStreamPath` (subscription payloads round-trip on every refresh; unconditional prefixing would cause runaway double-prefixing).
- `src/routing/dispatch-policy.ts` — `subscriptionHasStream` uses `ctx.durableStreamsRouting?.toBackendStreamPath(...)` instead of a hardcoded service-prefix.
- `test/stream-client.test.ts` — 5 new tests covering `putSubscription`, `addSubscriptionStreams`, `ackSubscription`, `releaseSubscription`, and idempotence of already-prefixed paths, plus the runtime-namespace round-trip on responses.
- `test/host.test.ts` — 2 new regression tests: (a) `AgentsHost`-level wire-through of `durableStreamsRouting` to outgoing subscription bodies; (b) bit-identical default behaviour when the field is omitted.
- `test/dispatch-policy-routing.test.ts` — 2 existing tests with backend-prefixed fixtures (`tenant-test/...`) now set `durableStreamsRouting` explicitly; their previous implicit assumption now matches reality.

## Migration

Webhook subscriptions persisted before this fix have un-prefixed `streams` entries on disk. They are **not** auto-healed: those subscriptions were already silently dropping every webhook event (the bug this PR fixes), so the only useful operator action is to delete and recreate any affected webhook subscription after upgrading. Pre-fix pull-wake subscriptions are unaffected.

## Test plan

- [x] `pnpm test -- test/stream-client.test.ts` — 14/14 pass
- [x] `pnpm test -- test/host.test.ts` — 7/7 pass
- [x] `pnpm test -- test/dispatch-policy-routing.test.ts` — 8/8 pass
- [x] Combined sweep of `stream-client*`, `host`, `dispatch-policy*`, `electric-agents-routes`, `webhook-forward-routing` — 62/62 pass
- [x] `pnpm typecheck` — zero errors in `agents-server/src`; remaining cross-package errors are pre-existing `../agents/src/` issues not introduced by this branch
- [x] `pnpm stylecheck` — clean
- [ ] End-to-end repro against hosted tenant after stratovolt wires `durableStreamsRouting` and bumps the dep (OpenFactory discord-bot per the issue)

Fixes electric-sql/stratovolt#1519
Closes #4357

### DIFF
--- a/.changeset/agents-server-subscription-routing.md
+++ b/.changeset/agents-server-subscription-routing.md
@@ -4,6 +4,12 @@
 
 Fix webhook dispatch never firing for service-prefixed durable-streams deployments.
 
-`StreamClient` now accepts an optional `routing: { serviceId, adapter }` option that applies the `DurableStreamsRoutingAdapter` to subscription path payloads (patterns, stream paths, `wake_stream`, ack/release stream/path fields) and the inverse transform to response payloads. `AgentsHostTenantConfig` gains a `durableStreamsRouting` field that plumbs the adapter through to the per-tenant `StreamClient`.
+`StreamClient` now accepts an optional `subscriptionRouting: { serviceId, adapter }` option that applies the `DurableStreamsRoutingAdapter` to subscription path payloads (patterns, stream paths, `wake_stream`, ack/release `stream`/`path` fields) and the inverse transform to response payloads. `AgentsHostTenantConfig` gains a `durableStreamsRouting` field that plumbs the adapter through to the per-tenant `StreamClient`.
 
-Cloud / service-routed durable-streams deployments key appends on a service-prefixed backend path (`<serviceId>/<path>`), so subscriptions previously registered with the un-prefixed path never matched the HotBuffer lookup and webhook fanout silently dropped. Pull-wake dispatch was unaffected because it writes to an explicit `wake_stream` rather than relying on HotBuffer fanout. Tenant-root deployments are unaffected (no routing option passed → behavior unchanged).
+`dispatch-policy.subscriptionHasStream` now consults `ctx.durableStreamsRouting` to compute the backend-namespace comparison instead of hardcoding `${ctx.service}/${path}`, so adapters with non-trivial backend conventions are honoured.
+
+Cloud / service-routed durable-streams deployments key appends on a service-prefixed backend path (`<serviceId>/<path>`), so subscriptions previously registered with the un-prefixed path never matched the HotBuffer lookup and webhook fanout silently dropped. Pull-wake dispatch was unaffected because it writes to an explicit `wake_stream` rather than relying on HotBuffer fanout. Tenant-root deployments are unaffected (no routing option passed → behaviour unchanged, bit-identical to the prior slash-normalising code).
+
+The `DurableStreamsRoutingAdapter` interface now documents that `toBackendStreamPath` / `toRuntimeStreamPath` MUST be idempotent and side-effect free; subscription payloads round-trip through `toBackendStreamPath` on every refresh and an unconditionally-prefixing adapter would cause runaway double-prefixing across `getSubscription` → `addSubscriptionStreams` cycles.
+
+**Migration note:** webhook subscriptions persisted before this fix have un-prefixed `streams` entries on disk. They are not auto-healed: those subscriptions were already silently dropping every webhook event (the bug this PR fixes), so the only useful operator action is to delete and recreate any affected webhook subscription after upgrading. Pre-fix pull-wake subscriptions are unaffected.
--- a/.changeset/agents-server-subscription-routing.md
+++ b/.changeset/agents-server-subscription-routing.md
@@ -1,0 +1,9 @@
+---
+"@electric-ax/agents-server": patch
+---
+
+Fix webhook dispatch never firing for service-prefixed durable-streams deployments.
+
+`StreamClient` now accepts an optional `routing: { serviceId, adapter }` option that applies the `DurableStreamsRoutingAdapter` to subscription path payloads (patterns, stream paths, `wake_stream`, ack/release stream/path fields) and the inverse transform to response payloads. `AgentsHostTenantConfig` gains a `durableStreamsRouting` field that plumbs the adapter through to the per-tenant `StreamClient`.
+
+Cloud / service-routed durable-streams deployments key appends on a service-prefixed backend path (`<serviceId>/<path>`), so subscriptions previously registered with the un-prefixed path never matched the HotBuffer lookup and webhook fanout silently dropped. Pull-wake dispatch was unaffected because it writes to an explicit `wake_stream` rather than relying on HotBuffer fanout. Tenant-root deployments are unaffected (no routing option passed → behavior unchanged).
--- a/packages/agents-server/src/host.ts
+++ b/packages/agents-server/src/host.ts
@@ -325,7 +325,7 @@ export class AgentsHost {
         bearer: config.durableStreamsBearer,
         ...(config.durableStreamsRouting
           ? {
-              routing: {
+              subscriptionRouting: {
                 serviceId: config.serviceId,
                 adapter: config.durableStreamsRouting,
               },
--- a/packages/agents-server/src/host.ts
+++ b/packages/agents-server/src/host.ts
@@ -15,12 +15,6 @@ export interface AgentsHostTenantConfig {
   serviceId: string
   durableStreamsUrl?: string
   durableStreamsBearer?: DurableStreamsBearerProvider
-  /**
-   * Routing adapter applied to subscription path payloads so they match the
-   * backend stream-path namespace used for appends. Required for cloud /
-   * service-routed durable-streams deployments where appends are keyed on a
-   * service-prefixed path; safe to omit for tenant-root deployments.
-   */
   durableStreamsRouting?: DurableStreamsRoutingAdapter
   streamClient?: StreamClient
 }
--- a/packages/agents-server/src/host.ts
+++ b/packages/agents-server/src/host.ts
@@ -9,11 +9,19 @@ import { DEFAULT_TENANT_ID, UnregisteredTenantError } from './tenant.js'
 import { WakeRegistry } from './wake-registry.js'
 import type { DrizzleDB, PgClient } from './db/index.js'
 import type { DurableStreamsBearerProvider } from './stream-client.js'
+import type { DurableStreamsRoutingAdapter } from './routing/durable-streams-routing-adapter.js'
 
 export interface AgentsHostTenantConfig {
   serviceId: string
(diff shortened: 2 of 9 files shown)


balegas:
## Self-review pass

Dispatched 3 parallel review agents focused on clarity/abstractions, failure modes/races, and test coverage. Summary of findings and how each was addressed:

### Addressed in 270c1750d

**Must-fix**

1. **No integration regression test for the actual bug** *(test review, finding A)* — the 4 unit tests in the original commit exercised `StreamClient` in isolation, but `webhook-forward-routing.test.ts` and `electric-agents-routes.test.ts` pass a no-op `toBackendStreamPath` to `AgentsHost`. So the host → stream-client wire-through that this PR is plumbing had zero assertion.
   → Added `applies durableStreamsRouting adapter to per-tenant subscription payloads` and `omits the routing adapter when durableStreamsRouting is not provided` in `test/host.test.ts`. Both construct `AgentsHost`, register a tenant with a real prefix-applying adapter, and assert the per-tenant `StreamClient`'s outgoing subscription body either contains or omits the service prefix accordingly.

2. **`releaseSubscription` routing coverage** *(test review, finding B)* — the bug class (silent path mismatch) is identical to ack but `releaseSubscription` was uncovered.
   → Added `service-prefixes release stream/path fields via the routing adapter`.

3. **Pre-fix subscriptions don't auto-heal** *(failure-modes review, finding C)* — webhook subscriptions persisted before this fix have un-prefixed `streams` on disk. `subscriptionResponseBody` strips the backend prefix on response, so the runtime cannot distinguish a pre-fix un-prefixed entry from a post-fix entry round-tripped through the adapter; `subscriptionHasStream` returns true → `addSubscriptionStreams` is never called → the subscription stays broken.
   → Auto-heal would require either *not* stripping in the response (changes the runtime-namespace contract) or threading raw backend paths through a parallel API. Both are non-trivial design changes. Given that pre-fix webhook subscriptions were already silently dropping every event (the bug this PR fixes), recreating them is the same operational cost as repairing them. Documented in the changeset and PR body as the required operator action after upgrade.

**Should-fix**

4. **`dispatch-policy.ts:subscriptionHasStream` hardcoded backend transform** *(failure-modes review, finding D)* — `const backendStream = \`${ctx.service}/${normalizedStream}\`` worked coincidentally because the production adapter uses exactly this format; any adapter with a different backend convention would break the existing-stream check.
   → Replaced with `ctx.durableStreamsRouting?.toBackendStreamPath(ctx.service, normalizedStream) ?? normalizedStream`. Kept the dual-equality check as defence-in-depth for response payloads that bypass the adapter (custom `streamClient`, etc.). Existing tests that used backend-prefixed fixtures (`tenant-test/...`) now opt in to `durableStreamsRouting` — surfacing the previously-implicit assumption and adding coverage for the dispatch-policy ↔ adapter integration.

5. **Rename `routing` → `subscriptionRouting`** *(clarity review, finding E)* — `routing` is ambiguous in a class also named "stream"-related; it only affects subscription payloads, not stream URLs (those use `appendPathToBaseUrl` directly). Mirrors the typed-and-self-describing `bearer: DurableStreamsBearerProvider` on the same class.
   → Renamed on `StreamClientOptions`. `AgentsHostTenantConfig.durableStreamsRouting` stays as-is (host-level name unchanged — also matches the `TenantContext.durableStreamsRouting` field).

6. **Idempotence contract on `DurableStreamsRoutingAdapter`** *(failure-modes review, finding F)* — today's production adapter (stratovolt's `serviceRoutedDurableStreamsAdapter`) guards against double-prefixing; the interface didn't require it. A future adapter without the guard would cause runaway double-prefixing across `getSubscription` → `addSubscriptionStreams` cycles.
   → JSDoc on both `toBackendStreamPath` and `toRuntimeStreamPath` declaring them MUST be idempotent, deterministic, side-effect free, and must not throw (exceptions in ack/release paths cause re-dispatch storms).

### Also done

- Anchor comment above `backendSubscriptionPath` declaring it as the paired transform with `runtimeSubscriptionPath` *(clarity review, finding G)* — costs nothing, makes the dual nature obvious to future readers.
- Migration note in the changeset and PR body *(see item 3 above)*.

### Deferred (follow-up, not blocking)

- **Top-level `stream`/`path` field coverage in `subscriptionRequestBody`** *(test review)*: covered indirectly by the new release test (which exercises `subscriptionRequestBody.stream`), but exhaustive coverage of every branch is a follow-up.
- **Response-side `acks[]` reverse map + string-form `streams[]` element coverage** *(test review)*.
- **`removeSubscriptionStream` with routing** *(test review)*: trivial but unused by the bug repro path.
- **`createSubscription` pattern transform with routing** *(test review)*: covered indirectly through `putSubscription`.
- **`subscriptionResponseBody` null-guard on `streams` entries** *(failure-modes review)*: mirror of the server-side guard in `durable-streams-router.ts:200-218`; defensive only.
- **Asymmetric shape between `StreamClientOptions.subscriptionRouting` and `AgentsHostTenantConfig.durableStreamsRouting`** *(clarity review)*: host could accept the bundled type directly. Mechanical refactor, doesn't change behaviour.
- **Conformance-test coverage for service-prefixed deployments** *(test review)*: `agents-server-conformance-tests` has no scenario forcing a service-prefixed deployment. Worth a follow-up issue.

### Verification

- `test/stream-client.test.ts` — 14 of 14 pass
- `test/host.test.ts` — 7 of 7 pass (5 existing + 2 new wire-through regressions)
- `test/dispatch-policy-routing.test.ts` — 8 of 8 pass (after updating fixtures to opt in to `durableStreamsRouting`)
- Combined sweep of related test files — 62 of 62 pass
- `pnpm typecheck` — zero errors in `agents-server/src`
- `pnpm stylecheck` — clean

Issue #4292, opened 2026-05-08
Deployment issue: Lack of Cheap Admission Control and Adaptive Retry-After with Exponential Backoff


alco:
## Overview

During deployment events, stack transitions, or high-load scenarios, the current admission control and retry behavior can lead to thundering herd problems and suboptimal recovery. This parent issue tracks improvements to make admission control cheaper (shed load earlier) and retry behavior adaptive (server and client coordinate via `Retry-After`).

## Problem

**Server side:**
- Admission control runs too late in the request pipeline and is computationally expensive under load
- `Retry-After` values are hardcoded in multiple places with no adaptation to actual load or recovery state
- When many requests arrive simultaneously, bottlenecks form before admission control can shed load

**Client side:**
- Elixir client ignores `Retry-After` headers entirely on 503 responses
- TypeScript client honors `Retry-After` but adds no jitter when server delay dominates client backoff, causing synchronized retries
- Some error paths retry immediately with no backoff delay

## Goal

1. **Cheap admission control** — shed load as early as possible in the request pipeline with minimal per-request cost
2. **Adaptive Retry-After** — server returns dynamic values based on load/recovery state rather than fixed constants
3. **Well-behaved clients** — all clients honor `Retry-After`, add jitter to prevent synchronized retries

Issue #4266, opened 2026-05-04
ShapeCache bottlenecks under thundering herd


alco:
# ShapeCache Bottlenecks Under Thundering Herd

Analysis of where load piles up when many shape requests arrive concurrently (both `?offset=-1` initial requests and requests carrying an existing handle), specifically the load that lands *before* AdmissionControl can shed it.

## Request flow (relevant slice)

`Electric.Plug.ServeShapePlug` pipeline:

1. `:resolve_existing_shape` — runs **before** admission. SQLite read.
2. `:check_admission` — ETS counter gate (`Electric.AdmissionControl`).
3. `:load_shape` — calls into `ShapeCache`.
4. `:serve_shape_response`.

So *every* request — even the ones that will be rejected by admission — performs a SQLite read first.

## Bottleneck 1 — Pre-admission SQLite reads

`resolve_existing_shape` runs on the `Bandit` request process and queries the shape DB (`ShapeStatus.handle_for_shape/2`). Under a thundering herd this saturates the SQLite read pool *before* AdmissionControl gets a chance to reject anything. Admission control therefore can't actually shed load on the hottest path — it can only shed work that happens *after* the read.

The WriteBuffer ETS lookup (`WriteBuffer.lookup_handle/2`) short-circuits recent writes, so steady-state hits are cheap, but cold/missing shapes still fall through to a real read connection checkout.

## Bottleneck 2 — `ShapeCache` GenServer mailbox

`ShapeCache.get_or_create_shape_handle/3`:

```elixir
with {:ok, handle} <- fetch_handle_by_shape(shape, stack_id),
     {:ok, offset} <- fetch_latest_offset(stack_id, handle) do
  {handle, offset}
else
  :error ->
    GenServer.call(name(stack_id), {:create_or_wait_shape_handle, ...}, @call_timeout)
end
```

For an `offset=-1` request on a shape that doesn't exist yet, the fast path fails and every caller funnels into a single `GenServer.call` against the per-stack `ShapeCache` process. Even with internal coalescing (see below), **all calls serialize through one mailbox**: 1000 requests for the same shape become 1000 sequential mailbox entries.

## Bottleneck 3 — Write connection contention inside the GenServer

Inside the GenServer, `maybe_create_shape` calls `ShapeStatus.handle_for_shape_critical/2`, which uses the **write** connection:

```elixir
def handle_for_shape_critical(stack_id, %Shape{} = shape, timeout \\ 10_000) do
[...]

PR #4359, opened 2026-05-19
feat(sync-service): cheap admission control (Tier 0) with post-load_shape reclassification

### LINKED ISSUE #4291: AdmissionControl logic becomes expensive when hit with a thundering herd of shape requests
## Problem statement

Current admission control logic requires the stack to be up and does a shape handle lookup for each request which can become expensive at a certain number of concurrent requests.

If the stack is not ready, the request registers itself with the `StatusMonitor` and waits up to 5 seconds to be notified about stack readiness. When many requests arrive at the same time, this momentarily turns the `StatusMonitor` process into a bottleneck. All requests are let in since admission control step hasn't even been reached yet.

## Proposed solution

Add a very cheap gate before expensive shape work.

Properties:

- ETS counters or another low-overhead primitive.
- No GenServer call on the reject path.
- Coarse classification from source, request method, live flag, and offset where possible.
- Returns `503` with adaptive `Retry-After` immediately when the instance/source is saturated.

alco:
## Summary

Two changes to `ServeShapePlug`'s pipeline that together make admission control cheap to run and accurately bound the resource it's supposed to protect.

1. **Cheap Tier 0 classifier.** `:resolve_existing_shape` is deleted. The admission classifier now reads a single ETS table (`shape_meta_table` for the stack) via `ShapeStatus.has_shape_handle?/2`. No SQLite, no GenServer call. Admission moves to the front of the plug pipeline so rejected requests don't pay for validation, shape lookup, or any of the other per-request setup.
2. **Reclassify at snapshot start.** Once the Snapshotter signals that the snapshot has started (i.e. it has obtained a connection from the snapshot pool), the in-flight permit moves from `:initial` to `:existing`. The `:initial` cap now bounds requests that are in validate-and-load *or* contending for a snapshot-pool connection, not requests currently streaming their response.

## Pipeline change

Before:

```
:resolve_existing_shape   ← SQLite read (ShapeDb), runs before admission
:check_admission          ← acquires :initial or :existing permit
:put_resp_content_type
:parse_body
:validate_request
:reject_subquery_shape_compaction_request
:load_shape
:serve_shape_response     ← streams body; permit released in after-clause
```

After:

```
:put_resp_content_type                  ← runs first so admission 503s keep Content-Type
:check_admission                        ← cheap classifier (1× :ets.member)
:parse_body
:validate_request
:reject_subquery_shape_compaction_request
:load_shape
:hold_initial_until_snapshot_started    ← :initial-bucket requests park here
:reclassify_admission_kind              ← :initial → :existing swap
:serve_shape_response                   ← streams body
```

## Why this is cheaper

`:resolve_existing_shape` called into `Electric.Shapes.fetch_handle_by_shape/2`, which falls through to `ShapeDb.handle_for_shape/2` — a SQLite read on the read pool. Issue #4266 calls this out as the first bottleneck on the pre-admission path: every request, including the ones admission is about to reject, performs that read. Under thundering herd, the read pool saturates before admission can shed anything.

The new classifier is `:ets.member(shape_meta_table(stack_id), handle)`. One lookup, no pool, no GenServer.

The rescue on `ArgumentError` covers the brief window during stack startup where the per-stack meta table hasn't been created yet — those requests classify as `:initial`, which is the safer bucket.

## What this PR addresses, what it doesn't

**ShapeCache (resolved within this PR).** Before this branch, every shape request — handle or not — could end up in a `GenServer.call(ShapeCache, {:create_or_wait_shape_handle, …}, 30_000)` on a cache miss, with no upstream cap on concurrent callers. With cheap admission: `:existing` requests short-circuit on the `validate_shape_handle` ETS hash check and never touch ShapeCache; only `:initial` requests can hit `{:create_or_wait_shape_handle, …}`, and their population is hard-capped by the `:initial` bucket. ShapeCache's max instantaneous concurrent-caller count is now `max_concurrent_requests.initial` rather than "however many clients showed up." The 30s `@call_timeout` now exists only to absorb pathological filesystem tail latencies, not to throttle queue depth.

**ShapeCache mailbox-overfill on coalesced creates (related concern, tracked in #4372).** Even with the `:initial` bucket capping the population, all `:initial` requests for the same uncached shape still pile into the ShapeCache mailbox: handler 1 does the actual creation, handlers 2..N short-circuit via the SQLite critical fetch, but every caller occupies a mailbox slot for the duration of handler 1's work plus its own dispatch latency. #4372 proposes a GenServer-owned per-shape ETS lock observed by callers (via `:ets.member` before they send the call) plus a `PollWait` fallback — same pattern as #4371, with tighter backoff defaults to match shape-creation timescales.

**StatusMonitor (out of scope, tracked in #4371).** `Api.validate_params` calls `StatusMonitor.wait_until` for every request. The fast path is an ETS read and is unaffected. The slow path — stack starting, post-deploy, recovering from sleep — funnels every concurrent waiter through the StatusMonitor mailbox and replies serially on the readiness flip. The `:initial` bucket caps inflight requests across the validate+load_shape span but does not eliminate that single-process reply burst. Acceptable for now; #4371 has the per-process polling design with the adaptive congestion flag.

**EtsInspector (out of scope, tracked in #4370).** Used in `validate_request` to build `%Shape{}`. Warm cache is in-process ETS. Cold cache and PG-degraded paths serialize every concurrent request through the inspector mailbox, and crucially, failed DB lookups are **not** cached — so under PG pool exhaustion the inspector re-attempts the same failing query for each queued message. The `:initial` bucket bounds the herd size but does not stop the inspector from re-burning the DB pool on every retry.

## Why reclassification happens at snapshot start

`release_admission_permit/0` runs in `call/2`'s `after` clause — i.e., after `Api.Response.send_stream/2` has finished draining the body. For an initial snapshot of a non-trivial shape, or a long-poll, that's seconds to minutes.

So before this PR, `:initial` permits were held for the **entire** request lifetime:

```
t=0    t=5ms   t=10ms   t=50ms          t=2000ms
│      │       │        │               │
│ check_admission (:initial +1)
│      │ validate
│              │ load_shape (often coalesces; may spawn Snapshotter)
│                       │ start streaming chunks
│                                       │ done streaming, :initial -1
└────────────── :initial held for 2000ms ─────────────────────────┘
```

After this change:

```
t=0    t=5ms   t=10ms   t=50ms   t=N ms          t=2000ms
│      │       │        │        │               │
│ check_admission (:initial +1)
│      │ validate
│              │ load_shape (returns once Consumer + Snapshotter spawned)
│                       │ await_snapshot_start (Snapshotter checks out PG conn)
│                                │ try_swap (:initial -1, :existing +1)
│                                │ start streaming chunks
│                                                │ done streaming, :existing -1
└──────── :initial 50ms + Snapshotter-pool wait ─┤
                                 └────── :existing rest of lifetime ────┘
```

`:initial` residency drops from "entire request lifetime" to "validate + load_shape + Snapshotter pool checkout." Two things matter about that last span:

**1. For coalesced handlers on a popular shape, the wait is essentially free.** `ShapeCache.await_snapshot_start` short-circuits on `ShapeStatus.snapshot_started?` (one ETS lookup) when the snapshot is already running. So requests on a warm shape add at most one ETS read between `load_shape` and the swap.

**2. For requests that trigger genuine shape creation, the wait is what makes `:initial` protect the snapshot pool.** When the Snapshotter is queued waiting for a PG connection from the snapshot pool, the handler stays parked in `:initial`. Once the Snapshotter actually checks out a connection and casts `{:snapshot_started, _}`, the handler reclassifies and the next handler can enter. So `:initial` is no longer just "validate-and-load throughput" — it's also "number of requests that have caused a Snapshotter to either start or wait on the pool." Under PG-pool saturation, new requests get rejected at the admission gate (503 with `Retry-After`) instead of piling up as an unbounded Postgrex checkout queue.

The downstream `get_merged_log_stream` call no longer issues its own `await_snapshot_start`: the plug's result is propagated via `request.snapshot_status` and consumed inline. This avoids racing the consumer's `stop_and_clean` on the snapshot-failure path, which would otherwise turn legitimate `SnapshotError` 503s into `must_refetch` 409s.

## Permit-release safety against handler death

Permits are stashed in `Process.put(@admission_permit_key, …)` at acquire/reclassify time and released in `call/2`'s `after` clause. The `after` clause covers all in-process termination paths (normal return, `raise`, `throw`, `halt`, mid-stream client disconnect). The remaining concern — being killed mid-flight by an external EXIT signal — is absorbed by Thousand Island's connection handler GenServer, which sets `Process.flag(:trap_exit, true)` (`deps/thousand_island/lib/thousand_island/handler.ex:343`). Trapping converts incoming EXIT signals to mailbox messages, so the handler finishes the current request normally, releases the permit, and only then terminates on the next `receive`. The remaining gaps (`:brutal_kill` after shutdown timeout, externally-sent `:kill`) only happen during BEAM shutdown, where the ETS counters reset along with the VM.

## Related

Closes #4291 — AdmissionControl logic becomes expensive under thundering herd

- Parent #4292 — Lack of Cheap Admission Control and Adaptive Retry-After
- Addresses bottleneck 1 of #4266 — ShapeCache bottlenecks under thundering herd


packages/sync-service/lib/electric/admission_control.ex
```diff
@@ -136,6 +136,75 @@ defmodule Electric.AdmissionControl do
...
+    cap = Keyword.fetch!(opts, :max_concurrent)
+    to_pos = tuple_pos(to_kind)
+    from_pos = tuple_pos(from_kind)
+    default = {stack_id, 0, 0}
+
+    [new_to, _new_from] =
+      :ets.update_counter(
+        table_name,
+        stack_id,
+        [{to_pos, 1, cap, cap + 1}, {from_pos, -1, 0, 0}],
```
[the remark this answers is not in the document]
alco: addressed

magnetised:
> Reclassify at snapshot start. Once the Snapshotter signals that the snapshot has started (i.e. it has obtained a connection from the snapshot pool), the in-flight permit moves from :initial to :existing. The :initial cap now bounds requests that are in validate-and-load or contending for a snapshot-pool connection, not requests currently streaming their response.

@alco active snapshot is still holding a connection so shouldn't it still count towards "things contending for a snapshot-pool connection"?

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
```diff
@@ -357,6 +354,43 @@ defmodule Electric.Plug.ServeShapePlug do
...
+  defp hold_initial_until_snapshot_started(conn, _), do: conn
+
+  # Moves the handler out of :initial so the :initial bucket can admit the next
+  # validate-and-load wave while this request is still streaming the response body
+  # to the client.
+  defp reclassify_admission_kind(%Conn{assigns: %{config: config}} = conn, _) do
+    with {stack_id, :initial} <- Process.get(@admission_permit_key),
+         max = Map.fetch!(config[:api].max_concurrent_requests, :existing),
+         :ok <-
+           Electric.AdmissionControl.try_swap(stack_id, :initial, :existing, max_concurrent: max) do
```
magnetised, line 387: what does it mean if the try_swap fails?

magnetised (approved):
really appreciate the work that's gone into this. awesome stuff. have a couple of questions, one in the review one on the main pr, but am assuming that I've just not understood something.

🌟

packages/sync-service/lib/electric/plug/serve_shape_plug.ex
alco: The call to `try_swap` fails only if the `:existing` bucket is already at capacity. An unlikely case in practice since we configure it with a higher value than `:initial`.

In any case, if the swap fails, the current request keeps holding the `:initial` permit until it's released in the `after` clause of the `call` function.

alco:
> > Reclassify at snapshot start. Once the Snapshotter signals that the snapshot has started (i.e. it has obtained a connection from the snapshot pool), the in-flight permit moves from :initial to :existing. The :initial cap now bounds requests that are in validate-and-load or contending for a snapshot-pool connection, not requests currently streaming their response.
> 
> @alco active snapshot is still holding a connection so shouldn't it still count towards "things contending for a snapshot-pool connection"?

You're right.

With the changes made in this PR, the `:initial` bucket protects `ShapeCache` primarily. But if initial shape snapshots cannot keep up with the rate of new shape requests, some requests will inevitably be stuck waiting for DB connection checkout or time out on it.

So the protection of the snapshot pull is more hypothetical than practical, considering that the `:initial` bucket size can be 2 orders of magnitude larger than the connection pool size. But at the point when the shape consumer process is already running and waiting on the snapshot, ShapeCache is supposedly ready to accept the next new shape request. Which could end up being a request for an already existing shape.

In other words, it may end up being a moot point in practice, but reclassifying a shape req mid-flight has potential to make new shape creation faster (by not waiting on the full snapshot creation+serving cycle) but doesn't really make it worse in the thundering herd scenario where shape reqs are already timing out on connection pool checkout en masse.

PR #4360, opened 2026-05-19
fix(agents-runtime): realpath guard read/write/edit against symlink escape

msfstef:
## Summary

Closes the symlink-bypass class of the `read`/`write`/`edit` path guard (CVE-2025-53109/53110 class). Previously the cwd check was a string prefix on the un-resolved path, so a symlink at `<cwd>/link.txt` pointing to `/etc/hostname` was readable, and a symlinked directory could redirect writes outside the workspace. The guard now realpaths the resolved target (or its deepest existing ancestor for write/edit creating new files) and re-checks containment.

The prefix check and realpath check live together in a single `resolveInsideWorkdir(filePath, workingDirectory)` helper used by all three tools. Returns `{ ok: true, resolved }` or `{ ok: false, reason }`; each tool calls it once.

See `plans/sandboxing-investigation.md` §1.3 and §3.5.

## Known gap

Hardlinks across the cwd boundary share the inode but report a path inside cwd, so the realpath check passes them. A proper fix requires a sandbox (§3.3 of the design doc).

## Risk

Low. Could surface in setups that intentionally symlink the workspace out to a parent monorepo. Predictable error message: *"Path X resolves outside the working directory via a symlink"*.

## Test plan

- [x] `test/path-guard.test.ts` — 10 unit tests covering the helper directly: in-bounds existent/non-existent, `..` and absolute-path escape, file/directory symlinks pointing outside, intra-cwd file/directory symlinks, regular-file ancestor (ENOTDIR walk-up), workspace-itself-is-a-symlink
- [x] `test/tool-path-symlink.test.ts` — 3 integration tests proving read/write/edit honor the guard at the tool boundary
- [x] `pnpm test`, `pnpm typecheck`, `pnpm stylecheck` clean in `agents-runtime`

## Addressed from review

- ENOTDIR is now treated like ENOENT in the ancestor walk-up so a write target like `existing-file.txt/sub/new.txt` produces the natural domain error rather than re-throwing the kernel code through the outer catch.

### DIFF
--- a/.changeset/realpath-symlink-guard.md
+++ b/.changeset/realpath-symlink-guard.md
@@ -1,0 +1,5 @@
+---
+'@electric-ax/agents-runtime': patch
+---
+
+The built-in `read`, `write`, and `edit` tools now reject paths whose realpath resolves outside the working directory. Previously the cwd guard was a string prefix check on the un-resolved path, which followed symlinks transparently — a symlink at `<cwd>/link.txt` pointing to `/etc/hostname` was readable, and a symlinked directory could redirect writes outside the workspace (CVE-2025-53109/53110 class). Intra-workspace symlinks (pnpm `.pnpm` trees, etc.) keep working. For non-existent write/edit targets the guard realpaths the deepest existing ancestor instead. Known gap: hardlinks across the cwd boundary still bypass; a proper fix requires a sandbox.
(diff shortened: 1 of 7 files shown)


icehaunter (approved):

msfstef:
Closing in favour of a more thorough sandboxing work: https://github.com/electric-sql/electric/pull/4369

PR #4361, opened 2026-05-19
fix(agents-runtime): SSRF guard in fetch_url

msfstef:
## Summary

Closes the SSRF surface on the built-in `fetch_url` tool. Previously: no host policy, so `http://169.254.169.254/`, `http://127.0.0.1/`, RFC1918 ranges, etc. were all fetchable. Now: literal-IP URLs in private/loopback/link-local ranges are rejected before the network call; hostname URLs are resolved via `dns.lookup({ all: true })` and rejected if **any** returned address is private (defense-in-depth against partial-private responses).

Stacked on #4354 (the characterization tests, base branch). When that lands, this auto-rebases to `main`.

See `plans/sandboxing-investigation.md` §1.3 and §3.6.

## New option

`createFetchUrlTool({ allowedHosts: ['localhost', '127.0.0.1'] })` — case-insensitive literal hostname match; bypasses both the IP-literal check and DNS resolution for the listed hosts.

## Breaking

Any current usage that pointed `fetch_url` at `localhost`, an internal LAN host, or a metadata endpoint stops working with a clear error message. One-line mitigation:

\`\`\`ts
createFetchUrlTool({ allowedHosts: ['localhost'] })
\`\`\`

The desktop wiring needs this for developers who fetch localhost via Horton; the desktop callsite is **not** changed in this PR (separate scope).

## Known gap

**DNS rebinding** — a second resolution between the guard's `dns.lookup` and the socket connect can return a different IP. Fixing this requires a custom undici dispatcher that pins the resolved address. Out of scope here; documented in source.

## Test plan

- [x] Literal `169.254.169.254`, `127.0.0.1`, `10.0.0.1`, `192.168.1.1`, `172.16.5.5`, `[::1]` → all rejected, no fetch call
- [x] Hostname resolving to single private address → rejected
- [x] Hostname with mixed public+private → rejected (rebinding-style)
- [x] `localhost` blocked by default, allowed via `allowedHosts: ['localhost']`
- [x] `allowedHosts` doesn't leak to sibling hosts (`['localhost']` still blocks `127.0.0.1`)
- [x] Normal public hostname → allowed
- [x] Unparseable URL → rejected before DNS
- [x] All DNS calls mocked — no real network in CI

### DIFF
--- a/.changeset/fetch-url-ssrf-guard.md
+++ b/.changeset/fetch-url-ssrf-guard.md
@@ -1,0 +1,9 @@
+---
+'@electric-ax/agents-runtime': patch
+---
+
+The built-in `fetch_url` tool now refuses URLs whose host resolves to a private, loopback, link-local, or cloud-metadata IP. Literal IPs in those ranges (`127/8`, `10/8`, `172.16/12`, `192.168/16`, `169.254/16`, IPv6 `::1`, `fe80::/10`, `fc00::/7`) are rejected before any network call. Hostnames are resolved via `dns.lookup({ all: true })` and rejected if *any* returned address is private (defense-in-depth against a partial-private response).
+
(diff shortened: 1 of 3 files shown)


msfstef:
Closing as this should be/will be a general sandboxing concern - whatever sandbox strategy we implement will include this, but if unrestricted mode is being used we should not try to guard for these

PR #4362, opened 2026-05-19
feat(agents-runtime)!: bash tool runs children with an env allowlist

msfstef:
## Summary

Closes the env-var exfiltration leak in the built-in `bash` tool. Previously `bash.ts` passed `env: { ...process.env }` wholesale to `child_process.exec`, so the LLM could exfiltrate any credential-shaped variable via `echo \$ANTHROPIC_API_KEY`. Now bash spawns children with a filtered allowlist constructed once at tool creation.

See `plans/sandboxing-investigation.md` §1.3, §1.5, §3.5.

## Default allowlist

Covers shell/locale/temp/XDG basics plus categories where the absence of a variable would silently break tooling rather than improve security:

- **Terminal/UX:** `COLORTERM`, `NO_COLOR`, `FORCE_COLOR`, `CI`
- **Proxies:** `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` plus lowercase variants (curl/git/node respect both forms)
- **TLS roots:** `NODE_EXTRA_CA_CERTS`, `SSL_CERT_FILE`, `SSL_CERT_DIR` (needed in corporate-MITM setups)
- **Windows essentials:** `SYSTEMROOT`, `COMSPEC`, `WINDIR`, `APPDATA`, `LOCALAPPDATA`, `USERPROFILE` (cmd.exe and Node lookups fail without these)

Deliberately excluded from defaults despite being commonly needed — these are real injection vectors and need an explicit opt-in:

- `NPM_CONFIG_*` (auth tokens in npmrc-equivalent envs)
- `GIT_SSH_COMMAND` (arbitrary command injection)
- `NODE_OPTIONS` (`--require ./malicious.js` is code-injection)

## Extension

```ts
createBashTool(cwd, { allowedEnvKeys: ['GITHUB_TOKEN'] })
```

The list **extends** (does not replace) the default allowlist. Customers can add keys but cannot accidentally shrink the safe defaults — silently dropping `PATH` would break tooling without a useful error.

## Breaking

Anything relying on env-supplied credentials reaching bash through `process.env` — `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `GITHUB_TOKEN`, `BRAVE_SEARCH_API_KEY`, custom secrets — silently stops working. **Intended behavior.** One-line mitigation per call via `allowedEnvKeys` as above.

## Test plan

- [x] Env-leak characterization from #4354 flipped: `ANTHROPIC_API_KEY` no longer appears in child output
- [x] `PATH` still forwarded
- [x] `allowedEnvKeys: ['MY_CUSTOM']` forwards that key to the child
- [x] Extending the allowlist doesn't remove defaults — `PATH` still passes when other keys are added
- [x] `pnpm test`, `pnpm typecheck`, `pnpm stylecheck` clean

## Addressed from review

- Broadened the default allowlist with proxy/TLS/terminal/Windows vars so this PR doesn't break `git`/`pnpm`/network calls behind corporate proxies or on Windows by default.
- The merged key Set is now constructed once at `createBashTool` time instead of per `execute()` call. `process.env` is still read per-call so live mutation of the parent env is honored.

### DIFF
--- a/.changeset/bash-env-allowlist.md
+++ b/.changeset/bash-env-allowlist.md
@@ -1,0 +1,7 @@
+---
+'@electric-ax/agents-runtime': patch
+---
+
+The built-in `bash` tool now spawns children with a filtered subset of the parent environment instead of `{...process.env}`. The default allowlist covers shell/locale/temp/XDG basics, terminal hints (`COLORTERM`, `NO_COLOR`, `FORCE_COLOR`, `CI`), proxy vars (`HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` + lowercase), TLS roots (`NODE_EXTRA_CA_CERTS`, `SSL_CERT_FILE`, `SSL_CERT_DIR`), and Windows essentials (`SYSTEMROOT`, `COMSPEC`, `WINDIR`, `APPDATA`, `LOCALAPPDATA`, `USERPROFILE`). Keeps `git`, `npm`, `pnpm`, and most CLI tools functional across dev, CI, and corporate-proxy environments without leaking API keys.
+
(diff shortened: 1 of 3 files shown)


icehaunter (approved):

msfstef:
Closing in favour of a more thorough sandboxing work: https://github.com/electric-sql/electric/pull/4369

PR #4366, opened 2026-05-19
fix(agents): keep webhook paths service-agnostic

icehaunter:
## Motivation

Webhook registration and dispatch were still carrying an old assumption from the closed tenanted cloud implementation: service ids could be prepended to Durable Streams paths and then stripped before runtime delivery. In the OSS agents-server that is the wrong boundary. The service or tenant route belongs in the configured Durable Streams base URL, while subscription patterns, explicit streams, wake streams, webhook wake payloads, and callback acks should all refer to logical stream paths only.

This avoids leaking cloud tenancy details into OSS paths and fixes webhook subscriptions created through agents-server/runtimes without reintroducing service-prefixed stream names.

The end-to-end cloud path also exposed a second boundary issue: Durable Streams calls the agents-server webhook forward endpoint without the cloud bearer token, because DS authenticates the delivery with a webhook signature instead. The OSS server therefore needs to persist the DS signing secret and validate signed deliveries before forwarding them to a runtime.

## What changed

- Normalize subscription `pattern`, `streams`, `wake_stream`, and ack stream fields by stripping only leading slashes at the Durable Streams boundary.
- Stop using service-id path transforms for subscription request bodies, subscription responses, webhook wake payload adaptation, and callback forwarding.
- Keep runtime-facing/entity paths root-relative so entity lookup and `agents-runtime` webhook payloads still use paths like `/chat/one/main`.
- Treat old service-prefixed subscription streams as distinct wrong paths instead of accepting them as equivalent to the logical stream path.
- Preserve custom Durable Streams routing adapters for URL routing, including a `serviceId` input for cloud wrappers, while deprecating path conversion hooks so existing adapters can still typecheck.
- Persist the `webhook_secret` returned by Durable Streams subscription creation and verify `Webhook-Signature` before forwarding webhook deliveries to runtimes.
- Store DS subscription callbacks as subscription ids and ack them through the stream client with the DS callback token, instead of replaying the cloud `Authorization` header to Durable Streams.
- Have `agents-runtime` send callback claim tokens using `electric-claim-token` when server auth headers are configured, preserving the configured cloud `Authorization` header.
- Add regression coverage for StreamClient subscriptions, dispatch policy linking, signed webhook forwarding, callback forwarding, host/router flows, and service-scoped Durable Streams base URLs.
- Add patch changesets for `@electric-ax/agents-server` and `@electric-ax/agents-runtime`.

## Cloud compatibility

Stratovolt already builds each tenant Durable Streams URL as `.../v1/stream/{serviceId}`, which matches this boundary: service id in the base URL, logical paths in the subscription body. The cloud wrapper can keep using its routing adapter for URL routing because the adapter still receives `serviceId`; the deprecated path conversion hooks are simply no longer used by agents-server subscription rewriting.

This does need to be paired with a cloud-side auth relaxation for `POST /_electric/webhook-forward/*`: that endpoint must be routable to the tenant using the `service` routing hint without requiring the normal bearer token. The security check moves to the OSS agents-server, which validates the Durable Streams `Webhook-Signature` against the persisted subscription secret before forwarding anything to the runtime. Other cloud agents endpoints should keep the existing bearer auth.

### DIFF
--- a/.changeset/agents-server-stream-paths.md
+++ b/.changeset/agents-server-stream-paths.md
@@ -1,5 +1,8 @@
 ---
 "@electric-ax/agents-server": patch
+"@electric-ax/agents-runtime": patch
 ---
 
 Keep Durable Streams paths service-agnostic in the OSS agents-server. Subscription payloads, webhook wake paths, callback ack paths, and routing adapters now treat stream names as relative to the configured Durable Streams base URL instead of applying service-id path transforms.
+
+Persist and verify Durable Streams webhook signing secrets before forwarding webhook wakes through agents-server. Runtime handlers that use server auth headers now send Durable Streams claim tokens via `electric-claim-token`, preserving the configured server `Authorization` header for cloud callback routes.
--- a/.changeset/agents-server-stream-paths.md
+++ b/.changeset/agents-server-stream-paths.md
@@ -1,0 +1,5 @@
+---
+"@electric-ax/agents-server": patch
+---
+
+Keep Durable Streams paths service-agnostic in the OSS agents-server. Subscription payloads, webhook wake paths, callback ack paths, and routing adapters now treat stream names as relative to the configured Durable Streams base URL instead of applying service-id path transforms.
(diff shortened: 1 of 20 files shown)


balegas (approved):
Looks easy!

PR #4374, opened 2026-05-20
feat(agents): support Ed25519 webhook signing

icehaunter:
## Motivation

Durable Streams webhook subscriptions now expose Ed25519 signing metadata instead of returning per-subscription `webhook_secret` values. The signature is attached to each delivery as `Webhook-Signature: t=<timestamp>,kid=<key-id>,ed25519=<base64url-signature>`, and receivers discover public keys at the stream-root control path: `{stream-root}/__ds/jwks.json`.

Electric Agents cannot just pass the upstream Durable Streams signature through, because agents-server rewrites subscription targets and enriches webhook bodies before forwarding them to runtimes. Once the body changes, agents-server has to validate the original Durable Streams delivery, own the public signing metadata it advertises to runtimes, and re-sign each forwarded delivery.

## Changes

- Bump `@durable-streams/server` to the `eac712f` package in `agents-server` and the runtime test dependency.
- Add an agents-server webhook signing abstraction backed by Ed25519 keys.
- Serve agents-server public webhook signing keys at `GET /__ds/jwks.json` with `application/jwk-set+json` and public cache headers.
- Validate incoming Durable Streams webhook deliveries in agents-server before subscription lookup, body parsing, enrichment, or downstream forwarding.
- Rewrite webhook subscription responses so `webhook.signing.jwks_url` points at the agents-server stream root, not the upstream Durable Streams backend.
- Re-sign forwarded webhooks after agents-server adapts/enriches the body for the runtime.
- Allow runtime-bound signing to be injected per tenant through `TenantContext.webhookSigner`, or configured in the standalone server with `webhookSigner` / `webhookSigningKey` / `ELECTRIC_AGENTS_WEBHOOK_SIGNING_PRIVATE_KEY`.
- Allow upstream DS webhook verification to be customized per tenant with `TenantContext.durableStreamsWebhookSignature` or standalone `durableStreamsWebhookSignature`; default verification uses the upstream DS stream root JWKS.
- Add runtime-side Ed25519 webhook signature verification. It is enabled by default against `${baseUrl}/__ds/jwks.json` and can be customized or disabled with `webhookSignature`.
- Preserve bodyless upstream `204`/`205`/`304` subscription responses when rewrapping proxied DS responses.
- Remove `webhook_secret` from local subscription response typing and add the required changeset.

## Decisions

- The canonical JWKS URL exposed to runtimes is the agents-server stream root: `${publicUrl}/__ds/jwks.json`. This matches Durable Streams sub-root mounting rules and avoids publishing the upstream DS backend JWKS when agents-server is acting as the webhook boundary.
- Agents-server treats Durable Streams and agents-runtime as separate trust boundaries. It verifies the raw upstream DS webhook body against the upstream DS JWKS, then signs the mutated runtime body with the agents-server signer.
- Upstream DS verification defaults to `${durableStreamsUrl}/__ds/jwks.json` for normal stream-root routing, and uses the configured Durable Streams routing adapter for custom control-root routing.
- The default standalone server generates an Ed25519 signer in-process for local/dev use. Tenanted/cloud deployments should inject a stable tenant-aware signer instead of relying on that process-local default.
- Runtime verification fetches by `kid` from JWKS and validates the exact raw body bytes received by the runtime, preserving the required `timestamp.raw_body` signing contract.

## Paired tenanted-cloud update required

The tenanted cloud host should provide a `WebhookSigner` per tenant or per stream root when building `TenantContext` for agents-server routes. The signer needs to expose:

- `sign(body)` returning `t=<timestamp>,kid=<tenant-key-id>,ed25519=<signature>` over `<timestamp>.<raw_body>`.
- `jwks()` returning the tenant public key set with the matching `kid`.

For upstream Durable Streams validation, leave `TenantContext.durableStreamsWebhookSignature` unset only when the default upstream stream-root JWKS URL is correct and reachable by agents-server. If DS JWKS discovery needs tenant routing, internal auth, custom fetch plumbing, or a KMS-backed key cache, set `durableStreamsWebhookSignature` with the resolved `jwksUrl` and/or `fetchClient`. Do not set it to `false` outside trusted in-process tests.

For cloud rollout, make sure the tenant routing layer treats `GET {tenant-stream-root}/__ds/jwks.json` as an agents-server route for runtimes, not as an authenticated Durable Streams proxy pass-through. Subscription create/read responses should advertise that same tenant public URL in `webhook.signing.jwks_url`.

The agents-server process must also be able to fetch the upstream Durable Streams JWKS for the corresponding tenant stream root. If that upstream control path is behind service auth, wire the verification fetch client to include the required credentials.

The private key used by agents-server for runtime-bound signing must be stable across agent-server instances for the same tenant/stream root, or backed by a shared KMS/HSM signer. During rotation, publish old and new public keys in JWKS until all in-flight deliveries signed by the old key are outside the runtime tolerance window and any retry window you want to support.

If the cloud runtime talks to agents-server through tenant-auth headers, configure `webhookSignature.jwksUrl` or its `fetchClient` so runtime JWKS discovery can reach the tenant JWKS endpoint. Do not disable `webhookSignature` outside trusted in-process tests.

### DIFF
--- a/.changeset/healthy-webhooks-sign.md
+++ b/.changeset/healthy-webhooks-sign.md
@@ -1,0 +1,6 @@
+---
+"@electric-ax/agents-server": patch
+"@electric-ax/agents-runtime": patch
+---
+
+Update Durable Streams server webhook support to Ed25519/JWKS signatures. Agents-server now exposes its own stream-root JWKS endpoint, supports injectable webhook signing keys/signers, rewrites subscription signing metadata to the agents-server JWKS, and re-signs forwarded webhook deliveries. Agents-runtime now validates webhook signatures before dispatching wakes.
--- a/.changeset/healthy-webhooks-sign.md
+++ b/.changeset/healthy-webhooks-sign.md
@@ -3,4 +3,4 @@
 "@electric-ax/agents-runtime": patch
 ---
 
-Update Durable Streams server webhook support to Ed25519/JWKS signatures. Agents-server now exposes its own stream-root JWKS endpoint, supports injectable webhook signing keys/signers, rewrites subscription signing metadata to the agents-server JWKS, re-signs forwarded webhook deliveries, and preserves bodyless upstream 204/205/304 subscription responses. Agents-runtime now validates webhook signatures before dispatching wakes.
+Update Durable Streams server webhook support to Ed25519/JWKS signatures. Agents-server now exposes its own stream-root JWKS endpoint, supports injectable webhook signing keys/signers, validates upstream Durable Streams webhook signatures, rewrites subscription signing metadata to the agents-server JWKS, re-signs forwarded webhook deliveries, and preserves bodyless upstream 204/205/304 subscription responses. Agents-runtime now validates webhook signatures before dispatching wakes.
--- a/.changeset/healthy-webhooks-sign.md
+++ b/.changeset/healthy-webhooks-sign.md
@@ -3,4 +3,4 @@
 "@electric-ax/agents-runtime": patch
 ---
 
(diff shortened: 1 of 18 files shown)


msfstef (approved):
Looks good !

PR #4375, opened 2026-05-20
refactor(agents-runtime): move skills loader + use_skill tool into runtime

balegas:
## Summary

Move the `.md`-skill-directory loader (`createSkillsRegistry`) and the per-entity skill tool builder (`createSkillTools`) out of `@electric-ax/agents` and into the leaf `@electric-ax/agents-runtime` package, alongside the rest of the entity-runtime primitives.

No behaviour change. Same files, re-rooted to the package whose dependencies they already use.

## Why runtime is the better home

- **Dependencies already point that way.** `extract-meta.ts` imports `completeWithLowCostModel` from `agents-runtime`. `createSkillTools` takes a `HandlerContext` and returns `AgentTool[]`, both defined in `agents-runtime`. The skills code uses zero symbols defined in `@electric-ax/agents`.
- **The behaviour is entity-agnostic.** A markdown skill pack works for Horton, Worker, and anything someone builds on top of `agents-runtime` — including [`@electric-ax/discord-bot`](https://github.com/electric-sql/OpenFactory/tree/feat/discord-bot/packages/discord-bot) in OpenFactory, which is the immediate trigger for this move.
- **Symmetric with `agents-mcp`.** MCP loader is a leaf package any entity opts into. Skills are the same shape. Putting them in `agents` makes "I want skills" pull in Horton + Worker + the desktop server bootstrap as transitive context.
- Today the skills code sits in `agents` only because Horton was built first and the skills code was written next to it.

## Changes

- `src/skills/{registry,tools,preamble,extract-meta,types}.ts` → `packages/agents-runtime/src/skills/`. Renames via `git mv`, no content change apart from one import swap.
- `serverLog` (lives in `@electric-ax/agents/src/log.ts`) replaced with `runtimeLog` from `agents-runtime/src/log.ts` — same pino sink, same levels, slightly different prefix-as-first-arg call signature. Adjusted at the four call sites.
- Public exports added to `agents-runtime/src/index.ts`: `createSkillsRegistry`, `createSkillTools`, `SkillsRegistry`, `SkillMeta`.
- Internal call sites in `@electric-ax/agents` updated: `bootstrap.ts` and `agents/horton.ts` now import from `@electric-ax/agents-runtime`.
- Skill tests move alongside: `agents-runtime/test/skills/{integration,preamble,tools,registry}.test.ts`.

## Compat for embedders

`@electric-ax/agents` is currently the only external entry point that previously exposed the skills surface, and that exposure happened only via internal use (no `index.ts` re-export today). So this is **not a public-API breaking change** — embedders that already use `@electric-ax/agents` only see Horton / Worker / Server APIs, all of which still work.

If desired I can add re-export shims in `@electric-ax/agents` for `createSkillsRegistry` + `createSkillTools` to make these symbols available from both packages during a deprecation window. Tell me if you'd like that and I'll add a fixup commit.

## Verified

- Skills tests in their new home: 18/18 (`packages/agents-runtime/test/skills/{integration,preamble,tools,registry}.test.ts`).
- `@electric-ax/agents` full suite: 38/38 with the rewritten import paths.

## Follow-up

Replaces #4373 (which only re-exported from `agents`).
Once this lands and a new `@electric-ax/agents-runtime` cuts on npm, OpenFactory's discord-bot drops the vendored tarball and re-pins to the real semver.


KyleAMathews (approved):
Approved. I will say I'm not sure putting all the tool builders in the runtime is the right idea. I know I started it haha but it's premature centralization — the two options are tools are libraries you use or tools are examples you copy/paste/tweak code you use. I think tools will be often highly customized for different agents so I think in general most tool code should be examples and a few libraries. But we can figure that out later.

balegas:
I agree, in future agents will write/customize their own tools. Maybe there are some core tools that skip the pattern.

PR #4378, opened 2026-05-20
feat(agents-runtime): add send tool and make send() awaitable

KyleAMathews:
## Summary

Add a `send` tool that exposes `ctx.send()` to LLM agents (Horton and Worker), allowing them to send messages to Electric entities by URL. Change `send()` from fire-and-forget (`void`) to awaitable (`Promise<SendResult>`) across the entire stack so callers get delivery confirmation and structured error handling.

## Approach

**New `send` tool** (`packages/agents-runtime/src/tools/send.ts`)
- Factory function `createSendTool(send)` following the same pattern as `createBashTool`, `createSpawnWorkerTool`, etc.
- Typebox schema validates parameters; additional semantic check on `afterMs` (non-negative, finite)
- Catches send failures and returns structured LLM-readable results (`{ sent: false, error: true, message }`) instead of raw exceptions — consistent with `fetch_url` error handling pattern

**Return type: `void` → `Promise<SendResult>`**
- New discriminated union type: `SendResult = { sent: true; targetUrl } | { queued: true; targetUrl }`
- Propagated through `RuntimeContext.send`, `SelfHandle.send`, `EntityHandle.send`, `HandlerContext.send`, and all internal plumbing (`executeSend`, `dispatchSend`, `sendToEntity`)
- `SelfHandle.send` also gains `afterMs` in its opts, fixing a prior inconsistency

**Error handling change** (`process-wake.ts`)
- Old: `void serverClient.sendEntityMessage(...).catch(err => failBackgroundWake(err, 'SEND_FAILED'))`
- New: `await serverClient.sendEntityMessage(...)` with errors propagating to callers
- Removed `failBackgroundWake` from the send path — a transient send failure no longer poisons `liveProcessError` and terminates the entire wake cycle. The handler-level catch still records truly unhandled errors as `HANDLER_FAILED`.

## Key Invariants

- Every `send()` call site returns its promise — no fire-and-forget paths remain
- `SendResult` discriminated union ensures callers can distinguish `sent` vs `queued`
- The send tool catches all errors internally; the LLM always gets a structured response
- Send failures no longer abort the wake loop — the LLM can retry or inform the user

## Non-goals

- `EntityHandle.send` does not yet accept `opts` for `type`/`afterMs` (follow-up)
- `asToolResult` helper is duplicated between `send.ts` and `schedules.ts` — deferred extraction

## Verification

```bash
pnpm --filter @electric-ax/agents-runtime exec vitest run --pool-options.threads.maxThreads=2
pnpm --filter @electric-ax/agents exec vitest run --pool-options.threads.maxThreads=2
```

## Files Changed

| File | Change |
|------|--------|
| `agents-runtime/src/types.ts` | Add `SendResult` type; update `send` signatures from `void` to `Promise<SendResult>` |
| `agents-runtime/src/tools/send.ts` | **New** — `createSendTool` factory with validation and structured error handling |
| `agents-runtime/src/tools.ts` | Re-export `createSendTool` |
| `agents-runtime/src/process-wake.ts` | Make `executeSend` async, remove `failBackgroundWake`, return `SendResult` |
| `agents-runtime/src/setup-context.ts` | Update `dispatchSend`/`sendToEntity` to return `Promise<SendResult>` |
| `agents-runtime/src/context-factory.ts` | Update `executeSend` config and `ctx.send` return types |
| `agents-runtime/test/send-tool.test.ts` | **New** — tests for happy path, error path, and `afterMs` validation |
| `agents-runtime/test/runtime-server-client-update-metadata.test.ts` | Fix mock `executeSend` to return `Promise<SendResult>` |
| `agents/src/agents/horton.ts` | Register send tool in Horton agent |
| `agents/src/agents/worker.ts` | Register send tool in Worker agent |
| `agents/src/tools/spawn-worker.ts` | Add `send` to `WORKER_TOOL_NAMES` |
| `agents/test/horton-tool-composition.test.ts` | Add `send` to expected tool list |


samwillis (approved):
This looks good. I like making `send()` awaitable and the new tool wiring is straightforward.

Codex found two small issues around `afterMs` validation:

1. Direct runtime sends can still pass invalid delays. The send tool rejects negative/non-finite `afterMs`, but `ctx.send(..., { afterMs: -1 })` bypasses that and reaches the server. Since `entities-router` only treats `afterMs > 0` as delayed, negative values fall through to immediate delivery. I think the validation should live in the shared runtime/server path too, not only the LLM tool.

2. The send tool throws on invalid `afterMs` before its `try/catch`, so this failure mode does not return the structured `{ sent: false, error: true, ... }` result described in the PR. It would be more consistent to return the same structured tool result for validation failures.

I note that the desktop build failed on this too. I'll open a PR to fix that.

PR #4394, opened 2026-05-22
Fix malformed offset validation

robacourt:
## Summary
- validate parsed log offset parts before constructing LogOffset structs
- return 400 for malformed shape offsets like `0_-1` instead of raising FunctionClauseError
- add regression coverage for offset parsing and shape request validation

## Testing
- `mix test test/electric/replication/log_offset_test.exs test/electric/plug/serve_shape_plug_test.exs --trace`


alco (approved):
Looks good!

PR #4397, opened 2026-05-22
feat(electric-telemetry): refine coarse process_type buckets in place

erik-the-implementer:
## Summary

Refines the existing `process_type` telemetry attribute for the three coarse buckets that hide the most signal during overload (per recent investigations into long-mailbox spikes), instead of adding a separate `process_subtype` attribute.

Following review feedback (https://github.com/electric-sql/electric/pull/4397#pullrequestreview-4409634855): grouping by a bare `erlang` or `supervisor` is rarely useful, and the set of more specific names we'd replace them with is limited. So rather than carry the detail in a companion attribute, we fold it directly into `process_type`:

- `process_type = "erlang"` → replaced by the registered name; else the first named `$ancestor`; else the `initial_call` MFA string (e.g. `":erlang.apply/2"`). Falls back to `erlang` if none apply.
- `process_type = "supervisor"` → replaced by the registered name; else the first named `$ancestor`; else the `initial_call` MFA string. Falls back to `supervisor` if none apply.
- `process_type = "logger_olp"` → folded together with the handler id (its registered name) into a single value `logger_olp:<handler_id>` (e.g. `logger_olp:default`, `logger_olp:otel_log_handler`). Falls back to bare `logger_olp` for an unregistered process.

All other `process_type` values are unchanged. There is **no separate `process_subtype` attribute** — the refinement happens in place, keeping cardinality low (registered names + MFAs only; no pids, no dynamic registry tuples) while giving the drill-down those investigations needed.

Affected events: `vm.monitor.long_gc`, `vm.monitor.long_schedule`, `vm.monitor.long_message_queue`, `process.memory`, `process.bin_memory`.

## Related issues

- Source task: `electric-sql/alco-agent-tasks#46`
- SRE motivation / investigation: `electric-sql/alco-agent-tasks#45` — long-mailbox / overload investigations where the coarse `process_type` buckets (`supervisor`, `erlang`, `logger_olp`) hid the specific processes responsible.

## Implementation notes

- `ElectricTelemetry.Processes.proc_type/1` now refines the coarse type in a single `Process.info/2` call (which additionally fetches `:registered_name`). The previous `proc_type_and_subtype/1` and `proc_subtype/2` are removed.
- `sorted_groups/2` groups by `type` alone again.
- `:process_subtype` is removed from the `tags:` lists and the event metadata in `application_telemetry.ex` and `system_monitor.ex`.

## Test plan

- [x] Unit tests for each refinement bucket and its fallback paths (`packages/electric-telemetry/test/electric/telemetry/processes_test.exs`).
- [x] Full `electric-telemetry` suite passes (121/121).
- [x] Changeset updated: `@core/electric-telemetry: minor`.


robacourt (commented):
proc_type already comes from a variety of places (process label or if not initial_module) so it seems natural to extend that to check other places if the result we get back is not helpful. for example, I don't think it's every useful to have `erlang` as the proc_type and we could fall back to what you have as the subtype in that case. same for `supervisor`. 

The strong argument for a proc_sub_type would come from `{"logger_olp", handler id}` if there's an infinite amount of handler_ids, and if it's useful to group by `logger_olp`.

So I can potentially see the need for proc_subtype, but personally I wouldn't use it for `erlang` and `supervisor` because it's not particularly useful to group by `erlang` or `supervisor` and there should only be a limited number of the names we'd replace it with.

Or maybe we shouldn't have subtype at all. What are the handler ids? Are they limited and readable?

alco:
@robacourt good call. I've folded the subtype back into process_type. So this PR doesn't add a new metric tag, it simply refines the values for some process types to be more descriptive.

robacourt (approved):
Very nice!

PR #4399, opened 2026-05-22
Fix agent SIGINT abort context ordering

KyleAMathews:
## Summary
- let pi-agent-core settle aborted runs instead of immediately unsubscribing on SIGINT
- add a fallback timeout for providers/tools that ignore AbortSignal
- reorder model-facing timeline context so aborted run output appears before the SIGINT marker when server-appended signals race ahead of buffered runtime output
- add regression coverage for interrupted-run timeline ordering

## Tests
- pnpm exec vitest run test/timeline-context.test.ts
- pnpm exec eslint src/pi-adapter.ts src/timeline-context.ts src/types.ts test/timeline-context.test.ts --quiet

Note: broader agents-runtime typecheck/test runs still fail in this worktree because of existing dependency/linkage issues (missing/mismatched pino, cron-parser, @mariozechner/pi-agent-core, localOnlyCollectionOptions, etc.).

### DIFF
--- a/.changeset/agents-runtime-sigint-abort.md
+++ b/.changeset/agents-runtime-sigint-abort.md
@@ -1,0 +1,5 @@
+---
+"@electric-ax/agents-runtime": patch
+---
+
+Settle interrupted agent runs promptly when the model stream ignores abort completion, while preserving aborted run context ordering.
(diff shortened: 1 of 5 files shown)


samwillis (approved):
LGTM
that flake was annoying!

PR #4407, opened 2026-05-25
feat(agents): add webhook event source subscriptions

icehaunter:
## Motivation

Agents need a way to observe externally produced webhook streams both from normal application code and from agent-initiated subscriptions. This adds the runtime/server primitives for that without requiring the agents server to own how the event source catalog is populated. The server accepts an `eventSources` catalog from its host environment and uses that contract to resolve source keys, bucket templates, and subscriptions.

## Changes

- Add `webhook(endpointKey, { bucket })` as an observation source in `@electric-ax/agents-runtime`.
- Represent webhook rows as an `events` collection of `webhook_event` records and register default wake conditions for inserted webhook events.
- Ensure observed webhook streams exist before preload, so agents can observe known bucket paths before any events arrive.
- Add event source contract types for discoverable webhook sources, bucket templates, filters, subscription lifetimes, and hydrated wake payloads.
- Add server routes for listing event sources and upserting/deleting event source subscriptions on an entity.
- Persist dynamic subscriptions as manifest source entries and hydrate matching webhook rows into the next agent trigger message.
- Add `list_event_sources`, `list_event_source_subscriptions`, `subscribe_event_source`, and `unsubscribe_event_source` tools, and wire them into Horton.

## Programmatic webhook observation

```ts
import { createHandler, webhook } from "@electric-ax/agents-runtime"

export default createHandler(async (ctx) => {
  const ghPrComments = await ctx.observe(
    webhook("github-repo", { bucket: "prs/42" })
  )

  const events = ghPrComments.collections.events.toArray

  for (const event of events) {
    console.log(event.event_type, event.body)
  }
})
```

`webhook("github-repo")` observes `/_webhooks/github-repo`; passing `{ bucket: "prs/42" }` observes `/_webhooks/github-repo/prs/42`.

## Event source discovery contract

The agents server now accepts an externally filled `EventSourceCatalog`:

```ts
interface EventSourceCatalog {
  listEventSources(): EventSourceContract[] | Promise<EventSourceContract[]>
  getEventSource(sourceKey: string):
    | EventSourceContract
    | undefined
    | Promise<EventSourceContract | undefined>
}
```

Each `EventSourceContract` describes one discoverable webhook-backed source:

- `sourceKey`, `label`, `description`, `status`, `agentVisible`, and `revision` identify the source exposed to agents.
- `endpointKey` maps the source to the underlying webhook stream root.
- `buckets` describes named bucket templates such as `prs/:number`, the `paramsSchema` agents should satisfy, optional `eventTypes`, and optional named `filters`.

Agents use the tools against that contract:

- `list_event_sources` returns the current catalog so the model can choose `sourceKey`, `bucketKey`, `params`, and optional `filterKey`.
- `subscribe_event_source` resolves the selected bucket template, renders the durable stream path, persists a manifest-backed wake registration, and returns the subscription.
- `list_event_source_subscriptions` reports active subscriptions for the current entity.
- `unsubscribe_event_source` removes an existing subscription.

`filterKey` is recorded in the subscription and exposed in hydrated wake metadata, but server-side source filtering remains a follow-up; current wake registrations still target the resolved source/bucket stream.


.changeset/agent-event-source-tools.md
```diff
@@ -0,0 +1,7 @@
+---
+"@electric-ax/agents-runtime": minor
+"@electric-ax/agents-server": minor
+"@electric-ax/agents": minor
```
msfstef:
```suggestion
"@electric-ax/agents-runtime": patch
"@electric-ax/agents-server": patch
"@electric-ax/agents": patch
```

.changeset/generic-webhook-observation.md
```diff
@@ -0,0 +1,5 @@
+---
+"@electric-ax/agents-runtime": minor
```
msfstef:
```suggestion
"@electric-ax/agents-runtime": patch
```

KyleAMathews:
Why don't we do something like `observe(ghWebhook(options))` which returns webhook() but by wrapping it we get typed and validated options?

msfstef (approved):
Looks sane

PR #4410, opened 2026-05-26
Fork at message: re-roll a session from an earlier point in its history

kevin-dp:
## Summary

Adds first-class support for forking a session at any earlier message, rather than only at HEAD. From the UI: hover any user-message bubble (after the first response has landed) and click the GitFork icon — a new entity is created whose `main` stream is the source's history truncated **up to and including the previous completed agent response**. The new entity boots idle, ready for a different prompt.

Implements the design in `docs/fork-at-message.md` (local-only RFC).

## Phasing & commits

- `089bcb7` — initial pin to a durable-streams PR build (originally durable-streams/durable-streams#350; superseded — see "Upstream dependency" below).
- `3a496b9` — add the `EventPointer` type + lexicographic order-token helpers in agents-runtime.
- `347ea53` — widen `__electricRowOffsets` from `Map<key, string>` to `Map<key, EventPointer>`; thread through the 5 existing consumers (timeline, setup-context, context-factory).
- `b9c9ab8` — bump durable-streams pin to durable-streams/durable-streams#347 (rebased; that rebase plus a server-side fork-frame fix landed in durable-streams/durable-streams#347).
- `665e3d0e` — agents-server side: extend `streamClient.fork` with optional `forkPointer` headers, extend `forkBodySchema` with `fork_pointer: { offset, sub_offset }` (snake_case wire), thread through `entityManager.forkSubtree`, root-only pointer threading (Q2), manifest filter on the root (Q3 strict-on-root / loose-on-descendants), `readJsonWithPointers` helper.
- `f0bd4494` — agents-server-ui: hover-revealed `GitFork` button on `UserMessage` rows; `ChatView` computes the per-row fork-anchor map by walking the timeline backwards to the latest completed `runs` row and reading its pointer from `__electricRowOffsets`. Eligibility (no preceding completed run, in-flight run) suppresses the affordance.
- `9dc46f25` — bugfix: the Phase-3 manifest filter compared pointers across coordinate systems (live delivery batches vs historical single-batch reads). Switched to canonical cumulative position from each event's stable `headers.offset`. Also tightened null-anchor handling.
- `2f61c946` — bugfix: durable-streams/durable-streams#347's `Stream-Fork-Sub-Offset` addresses items within a *single log entry*, not items globally past the anchor. The runtime was minting sub-offsets across log-entry boundaries when delivery batches spanned multiple entries (typical during catch-up). Switched `entity-stream-db`'s `onBeforeBatch` to group by `item.headers.offset` so pointers are server-compatible regardless of chunking.
- `c052ed96` — UX fix: skip the all-subtree-idle wait on the source root for pointer-forks. The runtime keeps a worker warm for 5 minutes after the last `handler returned` (`status: "running"` during the keep-alive window), and the fork POST default wait is 120 s, so clicking the button right after a response landed hung for two minutes and then 409'd. Since the pointer path reads the source's `main` historically up to a frozen offset, concurrent writes past the pointer can't tear our snapshot; we only need to wait+lock kept descendants (Q3 HEAD-clones) once they're identified. HEAD-clone forks (no pointer) keep the old full-idle requirement.
- `5997c95` — flip `@durable-streams/{client,server,state}` from `pkg.pr.new/...@347` to the published versions (`^0.2.6` / `^0.3.5` / `^0.2.9`) now that durable-streams/durable-streams#347 has merged and a release has been cut.

## Upstream dependency

The upstream PR durable-streams/durable-streams#347 (`Stream-Fork-Sub-Offset`) provides the truncated-fork wire protocol this PR depends on.

It was originally rebased onto durable-streams/durable-streams#350, but that PR was deprecated by its author and split into four smaller PRs: durable-streams/durable-streams#359, durable-streams/durable-streams#360, durable-streams/durable-streams#361, and durable-streams/durable-streams#362, all of which merged into `main`. The state-package subscription hooks piece (`queryOnce`, `createTransaction`, `getStreamDBCollectionId`, `coalesce`, `concat`, `createLiveQueryCollection`, `localOnlyCollectionOptions`, `toArray`) merged via durable-streams/durable-streams#365.

durable-streams/durable-streams#347 has now **merged** (2026-06-01) and the wire protocol is available on the published packages:

- `@durable-streams/client@0.2.6`
- `@durable-streams/server@0.3.5`
- `@durable-streams/state@0.2.9`

This PR depends on those versions directly — no more `pkg.pr.new` pins.

## What's wired

- `POST /_electric/entities/<type>/<id>/fork` accepts optional `fork_pointer: { offset, sub_offset }`. Omitted → existing HEAD-clone behavior.
- Server-side validation surfaces 400 with descriptive errors on invalid offset or out-of-range sub-offset.
- Root entity's `main` is truncated at the pointer; `error` and shared-state streams clone at HEAD (Q2).
- Root's manifest filtered at the pointer; children whose manifest entry landed after X are dropped from the fork along with their subtrees (Q3 strict-on-root). Kept children come along at HEAD (Q3 loose-on-descendants).
- For pointer-forks the source root doesn't need to be idle (only the kept descendants do); HEAD-clones still wait for the full subtree to be idle.
- `createForkEntity(url, { pointer })` client helper.
- `UserMessage` per-row `onForkFromHere` prop + CSS hover/focus-within reveal. First message in a chat and in-flight runs suppress the button.

## Known gaps / follow-ups

1. **`createForkEntity` swallows errors silently** (`.catch(() => {})`). Now that pointer-forks no longer wait for the root to be idle, the 120 s hang case is gone — but other errors (manifest filter throws, 400 on bad pointer, etc.) still vanish without UI feedback. Surfacing this via a toast/inline state is a small UX follow-up.
2. **Mobile (Phase 5)** — the affordance propagates automatically through the Expo DOM embed since `ChatView` is reused, but touch devices don't fire `:hover` so the button visibility rule needs an adjustment (always-visible at reduced weight, or long-press) once the embed is tested.
3. **Stratovolt** — Stratovolt needs to bump to `@durable-streams/server@^0.3.5` before fork-at-message works against Stratovolt-hosted servers; until then it only works against local agent servers.

## Test plan

- [ ] Pull, build, run `./scripts/dev.sh start --with-agents` (needs `ANTHROPIC_API_KEY` + `ELECTRIC_AGENTS_PULL_WAKE_RUNNER_ID=builtin-agents` + `ELECTRIC_AGENTS_REGISTER_PULL_WAKE_RUNNER=true` in `.env`).
- [ ] Spawn a Horton entity, send two messages, wait for both to complete.
- [ ] Hover the second user-message bubble → GitFork button fades in.
- [ ] Click → navigates to a new entity URL ending in `-fork-<hash>` with only the first exchange in history. Works immediately even while the source is still in the post-run 5-minute keep-alive window.
- [ ] Send a new prompt in the fork → fresh response, source entity unchanged.
- [ ] Negative: first user message gets no button. While Horton is streaming, the triggering message gets no button. Button returns after `✓ done`.
- [ ] Server-side via curl: `POST /fork` with bad `fork_pointer.offset` → 400; with overshooting `sub_offset` → 400; without `fork_pointer` → 201 full clone.


packages/agents-runtime/src/entity-stream-db.ts
```diff
@@ -45,7 +46,15 @@ interface EntityWriteUtils {
 type EntityCollectionMeta = {
   __electricSourceDb?: EntityStreamDBWithActions
   __electricSourceId?: string
-  __electricRowOffsets?: Map<string | number, string>
+  /**
+   * Per-row durable-stream pointer (`{ offset, subOffset }`). Populated
+   * by the batch callback in `entity-stream-db` and consumed by code
+   * that needs to address a specific event on the stream — most
+   * notably the fork-at-message UX, which uses these pointers as
```
kevin-dp: Let's remove this docstring. There's already (and if not there should be) a docstring on the `EventPointer` type.

packages/agents-runtime/src/entity-stream-db.ts
```diff
@@ -155,6 +159,13 @@ export function createEntityStreamDB(
     rowOffsetsByCollection.set(name, new Map())
     timelineOrdersByCollection.set(name, new Map())
   }
+
+  // Tracks the END offset of the previous batch — i.e. the START
+  // offset of the next batch's items, which is the anchor we pair
+  // with each item's sub-offset to form an `EventPointer`. `null`
+  // before any batch has arrived (= "anchor at stream start"; matches
+  // PR #347's "omitted Stream-Fork-Offset" semantic).
```
kevin-dp: let's remove the reference to PR 347 from this comment

packages/agents-runtime/src/entity-stream-db.ts
```diff
@@ -285,7 +296,21 @@ export function createEntityStreamDB(
     onBeforeBatch: (batch) => {
       opts?.onBeforeBatch?.(batch)
       replayBatchOffset.current = batch.offset
-      batch.items.forEach((item, itemIndex) => {
+      // PR #347's `Stream-Fork-Sub-Offset` addresses items WITHIN A
```
kevin-dp: Remove PR 347 reference.

packages/agents-runtime/src/entity-stream-db.ts
```diff
@@ -669,11 +719,17 @@ export function createEntityStreamDB(
         event.value !== null
       ) {
         const orders = timelineOrdersByCollection.get(collectionName)
+        // applyEvent stages an in-process event (not delivered through
+        // a wire batch). It carries a single event, so the pointer's
+        // sub-offset is always 1 ("the one item past this anchor").
+        // If no real offset is available, synthesize a monotonically-
+        // increasing `local:…` token so successive applyEvent calls
```
kevin-dp: replace special ellipses character by three dots

packages/agents-runtime/src/event-pointer.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * `EventPointer` addresses a single event on a Durable Stream.
+ *
+ * It is the pair the agents-server uses on the wire when forking: it
+ * maps directly to the `Stream-Fork-Offset` + `Stream-Fork-Sub-Offset`
+ * headers defined by durable-streams PR #347.
```
kevin-dp: Let's remove the reference to PR 347. This will become part of the DS protocol so we can just reference to the DS protocol.

packages/agents-runtime/src/event-pointer.ts
```diff
@@ -0,0 +1,118 @@
...
+ *   - `subOffset` is the count of JSON messages past the anchor to
+ *     include in the fork. So sub-offset `N` includes items
+ *     `[0, 1, … N-1]` of the chunk that starts at `offset`.
+ *
+ * A pointer addresses inclusively-through-and-stops-at-item-(subOffset-1).
+ * For "fork up to and including item at batch-index `j`," set
+ * `subOffset = j + 1`.
+ *
+ * Why this shape and not the wire's opaque offset alone:
+ * PR #347 ships fork-side sub-offsets but explicitly does NOT add
```
kevin-dp: remove the PR reference

packages/agents-runtime/src/event-pointer.ts
```diff
@@ -0,0 +1,118 @@
...
+ * read-side sub-offsets — the wire never tells the client the
+ * sub-offset of a delivered item. We compute them locally by counting
+ * positions within each `JsonBatch.items` and pairing them with the
+ * batch's start offset (= the previous batch's `Stream-Next-Offset`,
+ * or `null` for the first batch of a fresh read).
+ *
+ * Limitation: local counting is correct when reads start from the
+ * beginning of the stream. If a future read-side feature ever lets us
+ * resume from an arbitrary mid-chunk cursor, our local counter would
+ * miscount items that the server already skipped past. PR #347
```
kevin-dp: Let's remove this PR reference (also check for all other occurences of this PR reference, we need to remove them all).

packages/agents-server-ui/src/components/views/ChatView.tsx
```diff
@@ -211,6 +212,40 @@ function GenericChatBody({
     })
   }, [entityUrl, generationActive, signalEntity, stopPending])
 
+  // "Fork from here" anchor map. For each user-message inbox row, the
+  // anchor is the LATEST preceding completed `runs` row — its pointer
+  // identifies "fork up to and including this response, drop everything
+  // after." See docs/fork-at-message.md Q4. Rows without a preceding
```
kevin-dp: Remove reference to this markdown doc. It won't be checked in.

packages/agents-server-ui/src/components/UserMessage.tsx
```diff
@@ -15,20 +15,24 @@ export const UserMessage = memo(function UserMessage({
...
 }: {
   section: UserMessageSection
   showStop?: boolean
   stopPending?: boolean
   onStop?: () => void
+  /**
+   * When provided, renders a hover-revealed "Fork from here" button on
+   * the bubble. The caller is responsible for eligibility — pass
+   * `undefined` for messages where no preceding completed run anchors a
+   * fork. See docs/fork-at-message.md Q4.
```
kevin-dp: Remove this reference to the doc. We won't check that doc in.

packages/agents-server-ui/src/lib/ElectricAgentsProvider.tsx
```diff
@@ -432,11 +433,24 @@ function createSignalAction(
 }
 
 function createForkEntity(baseUrl: string) {
-  return async (entityUrl: string): Promise<{ url: string }> => {
+  return async (
+    entityUrl: string,
+    opts?: { pointer?: EventPointer }
+  ): Promise<{ url: string }> => {
+    // Wire convention is snake_case; in-code TS is camelCase. See Q4
+    // "Field-naming convention" in docs/fork-at-message.md.
```
kevin-dp: Remove docs reference.

packages/agents-server/src/routing/entities-router.ts
```diff
@@ -134,6 +128,16 @@ const inboxMessageBodySchema = Type.Object({
 const forkBodySchema = Type.Object({
   instance_id: Type.Optional(Type.String()),
   waitTimeoutMs: Type.Optional(Type.Number()),
+  // Optional anchor pointing at an event on the source root's `main`
+  // stream. See docs/fork-at-message.md. Wire shape is snake_case; the
```
kevin-dp: remove doc reference

packages/agents-server/src/entity-manager.ts
```diff
@@ -119,6 +118,16 @@ type ForkSubtreeOptions = {
   rootInstanceId?: string
   waitTimeoutMs?: number
   waitPollMs?: number
+  /**
+   * Optional anchor pointing at an event on the source root's `main` stream.
+   * When set: only events at or before the pointer are kept on the root's
+   * forked `main`, and the root's manifest is filtered so that descendants
+   * spawned after the pointer are dropped from the fork (their now-orphan
+   * subtrees are not forked). Per Q2 of the fork-at-message RFC, the
```
kevin-dp: remove RFC reference

packages/agents-server/src/entity-manager.ts
```diff
@@ -744,7 +752,36 @@ export class EntityManager {
...
     try {
-      const sourceTree = await this.waitForIdleSubtree(rootUrl, opts, workLocks)
+      // For pointer-forks we read the source root HISTORICALLY at a
+      // frozen offset, so concurrent activity on the root past the
+      // pointer can't tear our snapshot — we don't need to wait for
+      // the root to be idle (which would block the "click fork right
+      // after the response landed" case, since the runtime keeps the
+      // worker warm for `idleTimeout`). We still wait+lock any kept
+      // descendants below (after `computeEffectiveSubtree` runs),
+      // since per Q3 those are HEAD-cloned and need a stable snapshot.
```
kevin-dp: remove doc/RFC ref (Q3)

packages/agents-server/src/entity-manager.ts
```diff
@@ -754,9 +791,107 @@ export class EntityManager {
         )
       }
 
-      const snapshot = await this.readForkStateSnapshot(sourceTree)
+      // When forking at a pointer, pre-read the root's main, validate the
+      // pointer against the source's true history, and materialise the
+      // root-at-pointer snapshot fragments. The pointer ONLY applies to
+      // the root's `main` stream (Q2). Descendants kept by the manifest
```
kevin-dp: Remove Q2/Q3 references. Those docs aren't checked in.

packages/agents-server/src/entity-manager.ts
```diff
@@ -932,6 +1073,73 @@ export class EntityManager {
     held.clear()
   }
 
+  /**
+   * Variant of {@link waitForIdleSubtree} that takes an explicit entity
+   * list instead of walking the registry from `rootUrl`. Used by the
+   * pointer-fork path to wait+lock only the kept descendants (Q3), since
```
kevin-dp: remove Q3 ref here and also all other refs.

packages/agents-server/src/entity-manager.ts
```diff
@@ -1026,6 +1234,116 @@ export class EntityManager {
     }
   }
 
+  /**
+   * Translate `forkPointer` into a 1-indexed CUMULATIVE position in the
+   * source's flattened history. Throws a 400 if the pointer doesn't
+   * address a real event.
+   *
+   * Semantics (mirroring PR #347's durable-streams server interpretation):
```
kevin-dp: remove PR ref

packages/agents-server/src/stream-client.ts
```diff
@@ -251,6 +256,17 @@ export class StreamClient {
         'content-type': `application/json`,
         'Stream-Forked-From': new URL(this.streamUrl(sourcePath)).pathname,
       }
+      if (opts?.forkPointer) {
+        // PR #347 returns 400 if Stream-Fork-Sub-Offset > 0 without an
```
kevin-dp: Remove PR ref

packages/agents-server/src/stream-client.ts
```diff
@@ -424,6 +440,70 @@ export class StreamClient {
...
+        headers: this.streamHeaders(),
+      })
+      const response = await handle.stream<T>({
+        offset: fromOffset ?? `-1`,
+        live: false,
+      })
+      // Per-item pointer = { offset: anchor, subOffset: position-in-batch + 1 }
+      // where anchor is the END offset of the PREVIOUS batch (null for the
+      // very first batch). Matches `entity-stream-db.ts`'s onBeforeBatch
+      // semantics so the pointers we mint here align with PR #347's
```
kevin-dp: Remove PR ref

icehaunter (approved):
Server code looks sane. I think Claude has a valid point about dead code function

icehaunter:
Also Claude's point on the error - I've added some error surfacing with a toast - we can wire that in

PR #4414, opened 2026-05-26
fix(sync-service): Allow 0 for max_requests and fullsweep_after

magnetised:
0 for max_requests means infinite...
0 for fullsweep_after means always


balegas (approved):
... very philosophical.

alco:
@magnetised Is there additional context documented somewhere behind this change?

PR #4416, opened 2026-05-26
Fix Android cloud auth callback

samwillis:
> [!WARNING]
> This PR is temporarily stacked on #4408 (`samwilli/agents-mobile-ci`) only so the agents-mobile EAS CI build runs against this fix.
>
> We should rebase this PR back onto `main` and merge it ASAP, then update the CI branch from `main` instead of keeping this fix stacked on the CI setup branch.

## Summary

Fixes the Android Electric Cloud sign-in flow getting stuck after Google / GitHub auth on real devices (and dev-mode emulators) after the latest comprehensive fix exposed two regressions.

## What was breaking

The Chrome Custom Tab redirect (`electric-agents://oauth/callback?...`) was being lost on real Android devices — especially Android 13+ release builds where the OS aggressively kills the app process while the browser is open. `openAuthSessionAsync` would return `dismiss` even though sign-in had actually succeeded, `Linking.getInitialURL()` returned `null`, the redirect was never consumed, and the user was bounced back to the welcome screen. Underlying expo issue: [expo/expo#44284](https://github.com/expo/expo/issues/44284).

The previous comprehensive fix (`e580cf9ff`) addressed those release-build cases but introduced two regressions:

* The Expo config plugin imported `@expo/config-plugins`, which pnpm doesn't hoist into agents-mobile's `node_modules`. Local prebuild worked via expo's own require context, but in EAS Build's `expo cli config --json` step the require failed and the entire Android preview build was rejected with exit code 1.
* The `/oauth/callback` route navigated with `router.replace` inside a `useEffect`. That raced with Expo Router's own intent handling in dev builds — the route would mount, the cloud-auth state would flip to `signed-in` (JWT was even persisted to AsyncStorage), the effect would fire, but the replace landed too late and the user was left stuck on the spinner. Restarting the app showed they were already signed in.

## How it's fixed now (layered defense)

1. **Native `onNewIntent` override** via a local config plugin (`plugins/with-android-on-new-intent.js`) injected into the generated `MainActivity.kt`. When Android relaunches us via the redirect intent, this forwards it through `setIntent()` so `expo-linking` / `expo-web-browser` see it via `getIntent()` instead of the stale LAUNCHER intent. Plugin now imports `expo/config-plugins` (re-exported by `expo`, which IS a direct dep and resolvable from anywhere in the package) instead of `@expo/config-plugins`.
2. **Global `Linking` listener** mounted in `CloudAuthProvider` (in addition to the `/oauth/callback` route) so cold-start redirects are consumed even before Expo Router has had a chance to navigate to the route.
3. **`Linking.parse()` in `parseCallbackUrl`** instead of `new URL()`, because the WHATWG URL parser in Hermes is unreliable for custom schemes in release builds (silently returns `null` for present query params).
4. **`waitForDeepLinkCallback`** in `signIn` — a short grace period after the browser session returns `dismiss`, so a redirect-via-deep-link still completes the flow instead of rolling state straight back to `signed-out`.
5. **Idempotent `completeCallbackUrl`** so the three concurrent handlers (browser-session result, global Linking listener, route's own handler) can't trample each other.
6. **`<Redirect>`-based navigation in `/oauth/callback`** (rendered during the render phase) instead of `router.replace` inside an effect — bulletproof against the Expo-Router intent-handling race.

Also dropped the `createTask: false` / `showInRecents: true` flags on `openAuthSessionAsync` — they were meant to keep the Custom Tab in our task on Android 13+ release builds, but in dev mode they changed how the redirect URL was delivered (it ended up at `Linking` rather than being captured by the auth session) and contributed to the route-stuck regression. The `onNewIntent` override + global Linking listener already cover the release-build kill case.

## Test plan

- [x] `pnpm --filter @electric-ax/agents-mobile typecheck`
- [x] `pnpm exec expo prebuild --platform android --clean` injects `onNewIntent` into `MainActivity.kt`
- [x] `pnpm exec expo config --json` lists the plugin and exits 0
- [ ] EAS Android preview build succeeds (CI)
- [ ] Real-device sign-in completes (Google + GitHub) on Android 13+
- [ ] Emulator sign-in completes via dev client

### DIFF
--- a/packages/agents-mobile/app/_layout.tsx
+++ b/packages/agents-mobile/app/_layout.tsx
@@ -54,14 +54,19 @@ function RootNavigator(): React.ReactElement {
   // redirect — the wizard subsumes the URL input as its step 2, and
   // dismissing it falls back to `/server-setup` only if the user
   // still hasn't configured a server.
-  if (!onboardingDismissed && pathname !== `/onboarding`) {
+  if (
+    !onboardingDismissed &&
+    pathname !== `/onboarding` &&
+    pathname !== `/oauth/callback`
+  ) {
     return <Redirect href="/onboarding" />
   }
 
   if (
     !serverUrl &&
     pathname !== `/server-setup` &&
-    pathname !== `/onboarding`
+    pathname !== `/onboarding` &&
+    pathname !== `/oauth/callback`
   ) {
     return <Redirect href="/server-setup" />
   }
--- a/packages/agents-mobile/app/_layout.tsx
+++ b/packages/agents-mobile/app/_layout.tsx
@@ -1,8 +1,6 @@
-import * as React from 'react'
 import { Redirect, Stack, usePathname } from 'expo-router'
 import { StatusBar } from 'expo-status-bar'
-import * as Linking from 'expo-linking'
-import { ActivityIndicator, AppState, StyleSheet, View } from 'react-native'
+import { ActivityIndicator, StyleSheet, View } from 'react-native'
 import { GestureHandlerRootView } from 'react-native-gesture-handler'
 import {
   SafeAreaProvider,
@@ -19,7 +17,7 @@ import {
   useMobileAppState,
 } from '../src/lib/MobileAppState'
 import { CloudAuthProvider } from '../src/lib/CloudAuthContext'
-import { debugCloudAuth, isCallbackUrl } from '../src/lib/cloudAuth'
+import { isCallbackUrl } from '../src/lib/cloudAuth'
 
 export default function RootLayout(): React.ReactElement {
   return (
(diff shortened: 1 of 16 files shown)


packages/agents-mobile/app/_layout.tsx
```diff
@@ -50,18 +54,27 @@ function RootNavigator(): React.ReactElement {
     )
   }
 
+  if (coldStartOAuthCallback) {
```
kevin-dp, line 57: Claude found this potential problem:
Looks like `coldStartOAuthCallback` could bounce between `/` and `/oauth/callback` indefinitely:

```ts
const coldStartOAuthCallback =
  !!launchUrl && isCallbackUrl(launchUrl) && pathname !== `/oauth/callback`
if (coldStartOAuthCallback) {
  return <Redirect href="/oauth/callback" />
}
```

`launchUrl` is set once in `MobileAppStateProvider` (from `Linking.getInitialURL()`) and never cleared. After the callback route succeeds and `<Redirect>`s to `/` (or `/onboarding` / `/server-setup`), the layout re-renders with `pathname !== '/oauth/callback'` and `launchUrl` still pointing at the original cold-start URL — so the condition becomes true again and bounces back to `/oauth/callback`. That second mount sees `status: 'signed-in'` and immediately `<Redirect>`s away, so it'd potentially loop.

Maybe it works in practice (Expo Router may dedupe a `<Redirect>` to the path you're already navigating away from, or there's some pathname-stability I'm missing) — but worth either adding an empirical check, or exposing a `consumeLaunchUrl()` on `MobileAppState` that the callback route calls after handing off the URL to `cloudAuth.handleDeepLink`.

kevin-dp (approved):

Issue #4381, opened 2026-05-20
Terminology cleanup: align on canonical Electric Agents vocabulary


KyleAMathews:
## Summary

The agents codebase has accumulated meaningful terminology drift across:

- `agents-runtime`
- `agents-mcp`
- `agents-server`
- `agents`
- `agents-server-ui`
- `agents-desktop`
- `agents-server-conformance-tests`

Because we are still **pre-release**, this is a good time to clean this up before we carry aliases, legacy wording, and ambiguous terms forward.

The goal of this issue is to align on a small set of naming decisions, then use those decisions to drive a cleanup pass across code, docs, tests, prompts, and UI copy.

**This thread is for agreeing on the decisions below, not for reviewing the full glossary line by line.**

## Why this matters

Terminology drift creates friction in:

- docs and onboarding
- API design
- UI copy
- tests and conformance language
- team communication
- future naming stability

Since we do **not** need backward compatibility yet, we should prefer:

- one concept → one primary term
- old synonyms removed where possible
- overloaded terms narrowed or qualified

## Decisions needed in this thread

- [ ] **Product name:** Electric Agents vs Durable Agents
- [ ] **Core runtime noun:** entity vs agent vs session/chat
- [ ] **Distributed-systems term:** claim vs lease
- [ ] **Multi-tenant naming:** tenant vs service ID
- [ ] **Schema vocabulary:** `input_schemas` / `output_schemas` vs `inbox_schemas` / `state_schemas`
- [ ] **Built-in naming:** remove `chat` alias and converge on `horton`?
- [ ] **Deprecated terminology:** remove `tagged()` terminology pre-release?
- [ ] **Stale error naming:** remove `SCHEMA_VALIDATION_ERROR` in favor of `SCHEMA_VALIDATION_FAILED`?

## Current recommendations / starting point

These are starting recommendations, not final decisions.

- **Electric Agents** should be the canonical product name.
- **entity** should remain the canonical runtime/platform term.
- **agent** should be used for model behavior and high-level product shorthand.
- **session** is probably fine as a UI alias for chat-like entities.
- **chat** should probably remain a UI metaphor or specific label, not the general platform noun.
- **tenant** should be the cross-package conceptual term.
- **service ID** should be the explicit Electric Cloud/admin identifier that maps to a tenant.
- **`inbox_schemas` / `state_schemas`** currently look like the more semantically precise schema pair.
- **`chat`** as a built-in alias looks like a good candidate for pre-release removal.
[...]

Issue #4380, opened 2026-05-20
Standardize internal agent API naming across routes, runtime client, and desktop IPC


KyleAMathews:
# Summary

The internal agent API works today, but naming is inconsistent across HTTP routes, runtime client methods, and desktop IPC channels. Similar operations use different verbs (`get`, `set`, `remove`, `register`, `pick`, `choose`), and some names mix domain concepts with transport details.

This issue proposes the naming standard we want to end up with and a direct set of renames focused on the highest-friction cases already present in the codebase.

# Problem

Current inconsistencies increase cognitive load and make the API harder to learn and extend.

Examples from the current codebase:

- Runtime client (`packages/agents-runtime/src/runtime-server-client.ts`)
  - `getEntityInfo`
  - `removeTag`
  - `registerEntitiesSource`
  - `registerWake`
  - `upsertCronSchedule`
- Desktop IPC (`packages/agents-desktop/src/{main,preload}.ts`)
  - `desktop:get-servers`
  - `desktop:cloud-auth-state`
  - `desktop:cloud-agent-servers-state`
  - `desktop:mcp-snapshot`
  - `desktop:choose-working-directory`
  - `desktop:pick-directory`
- Server routes (`packages/agents-server/src/routing/*`)
  - `POST /_electric/entities/register`
  - `POST /_electric/cron/register`
  - `POST /_electric/webhook-forward/:subscriptionId`
  - `POST /_electric/callback-forward/:consumerId`

Main drift patterns:

- reads alternate between `get*`, bare `*state`, and noun-like names
- deletions alternate between `delete` and `remove`
- `register` is overloaded for different concepts
- user-selection flows alternate between `choose` and `pick`
- some names describe transport plumbing instead of the domain action

# Proposal

Adopt a lightweight naming standard for internal APIs:

- **Reads:** `getX` for single lookups, `listX` for collections, `getXState` for snapshots/state
- **Writes:** `createX`, `updateX`, `upsertX`, `deleteX`; use `setX` only for direct assignment semantics
- **Commands:** imperative verbs for actions (`send`, `fork`, `enable`, `disable`, `connect`, `restart`)
- **Registration:** reserve `registerX` for subscriptions, watchers, or external source/wake registration

Scope for this issue:

- define the final naming standard we want across internal APIs
- prioritize runtime client and desktop IPC naming
- rename routes directly where we think the current names are not the standard we want to keep

# Suggested first-pass renames

## Runtime client

[...]

PR #4417, opened 2026-05-26
feat(agents): tenant-scope and rename pre-alpha APIs

icehaunter:
## Motivation

Hosted Agents tenanting is a stable boundary even while the Agents API is still pre-alpha and likely to change. The hosted server URL should therefore put the tenant before the API version, rooted at `/t/<service-id>/v1`, while the open-source runtime and Agents server stay tenant-unaware: they treat the configured base URL, auth headers, and any stripped prefix as the complete server surface.

This PR also folds the pre-alpha breaking rename cleanup from #4380 and #4381 into the same branch so we do not publish multiple incompatible URL/API shapes in quick succession.

The CLI now also needs a non-interactive way to inspect an entity conversation from scripts or logs. The new `electric agents view` command uses the same entity stream data as `observe`, but renders a one-shot transcript and exits instead of requiring an interactive TUI.

Desktop reloads were also able to leave work pending after it had already completed locally: the runtime intentionally skipped the DS subscription `done` callback during shutdown. This PR treats completed wake checkpoints as ackable during graceful shutdown and leaves DS subscription state as the source of truth for whether replayed pull-wake notifications still have pending work.

## Changes

- Builds desktop/cloud Agents server URLs as `/t/<tenantId>/v1`, matches cloud auth by tenant-scoped base path, and removes `?service=` / `x-electric-service` routing from desktop requests.
- Simplifies runtime and conformance URL helpers so they append protocol paths under an opaque caller-provided base URL without parsing tenant/service query parameters or constructing tenant paths.
- Moves observation backing-stream endpoints to `POST /_electric/observations/entities/ensure-stream` and `POST /_electric/observations/cron/ensure-stream`, deleting the old `/entities/register` and `/cron/register` routes.
- Renames runtime/server APIs: `getEntityInfo` -> `getEntity`, `registerEntitiesSource` -> `ensureEntitiesMembershipStream`, `registerCronSource` -> `ensureCronStream`, and `removeTag` -> `deleteTag`.
- Renames schema vocabulary from input/output to inbox/state across request fields, entity definitions, default schema exports, tests, and docs.
- Renames internal webhook/callback routes and identifiers to `subscription-webhooks` and `wake-callbacks`.
- Removes deprecated `tagged()` exports and the built-in Horton `chat` compatibility alias.
- Renames the built-in docs search tool from `search_durable_agents_docs` to `search_electric_agents_docs`.
- Adds `electric agents view <url>` as a non-interactive transcript renderer that shares the same stream preload and timeline section logic as `observe`.
- Sends the wake `done` callback during graceful shutdown after `processWake` has reached a safe ack checkpoint, preventing reloads from leaving completed DS subscription claims pending.
- Adds `packages/agents-runtime/AGENTS.md` documenting that runtime is explicitly tenant-unaware, plus the required changeset.

## Decisions

- Tenant routing belongs outside `agents-runtime` and the OSS `agents-server`. Cloud/proxy integrations provide the tenant-scoped base URL and auth headers; runtime code only appends protocol paths below that base.
- Tenanting comes before the Agents API version: `/t/<service-id>/v1`. Future Agents API version changes stay below the stable tenant boundary.
- Because this API is pre-alpha, the Electric packages do not keep compatibility aliases for the old route, schema, tag, or docs-search names.
- `electric agents view` prints observe-style conversation sections, including compact tool call results and wake markers, instead of the model-facing timeline projection so user text payloads stay readable.
- Pull-wake shutdown is no longer treated as an automatic reason to skip acking. Once cleanup reaches a safe checkpoint, the runtime should attempt `done`; replayed runner wake-stream notifications can then rely on DS subscription claim state to decide whether any work remains.

## Rollout / Rollback

- Rollout requires Stratovolt `cloud-agents-server` to route `/t/:serviceId/v1/*` and strip that prefix before handing requests to the tenant-unaware OSS server. Clients must be moved to tenant-scoped base URLs at the same time.
- Rolling back means reverting this PR and the paired Stratovolt routing/client changes together, restoring query-param tenant routing and the previous pre-alpha route names.


msfstef:
Heads-up: `agents-mobile` doesn't appear to be migrated to the new tenant-path scheme, and looks like it'll break against path-based cloud routing.

At PR head, mobile still derives the tenant from a `?service=` query param and sends the `x-electric-service` header:

- `getCloudServiceIdFromServerUrl` reads `parsed.searchParams.get('service')` — `packages/agents-mobile/src/lib/cloudAuth.ts`
- `prepareServerHeaders` injects `x-electric-service: <id>` and only exchanges the dashboard JWT for an agents token *when a `serviceId` is found* — `packages/agents-mobile/src/lib/serverHeaders.ts`

Concrete failure mode: given a tenant-scoped URL like `https://host/t/<tenant>/v1`, `getCloudServiceIdFromServerUrl` returns `null`, so `prepareServerHeaders` falls into the "local server" branch — it registers the base URL but **skips the token exchange and registers no headers**. Outbound cloud requests would then go out with no `Authorization` header (401), not merely mis-routed.

The rollout note says "Clients must be moved to tenant-scoped base URLs at the same time," but mobile is a client and isn't touched in this PR. Mobile CI going green only validates its own unchanged behavior, not compatibility with the new Stratovolt path routing.

Could you confirm which of these is the case?
- mobile cloud-agents support isn't shipped / is gated, so this is a non-issue for now;
- Stratovolt will keep honoring `?service=` / `x-electric-service` during the transition; or
- this is an oversight and mobile needs the same `/t/<tenant>/v1` migration.

_(Reviewed with Claude Code)_

msfstef:
A few nit-level observations from a closer pass — nothing blocking, but small enough that they're worth folding in while the rename is hot.

### Worth a small fix

**1. Align `extractFirstUserMessage` with the codebase's `_seq` ordering convention** — `packages/agents/src/agents/horton.ts:330-340`

The new sort uses a per-row fallback (numeric `_seq` if present, otherwise the array index), which mixes two value scales. It's safe in the current call site (`firstWake && !ctx.tags.title`, when inbox rows have just arrived via the stream and `_seq` is populated — `cleanRow` in `entity-stream-db.ts:181` strips `_seq` on the outbound persist path, not inbound). But everywhere else in `entity-timeline.ts` (lines 1086, 1127, 1147, …) the pattern is `coalesce(_seq, -1)`. Matching that would (a) avoid the implicit "`toArray` order tracks insertion" assumption and (b) keep the title path robust if it's ever called from a non-`firstWake` context.

**2. Make the new typebox schemas strict** — `packages/agents-server/src/routing/entity-types-router.ts:35-56`

Now that `input_schemas`/`output_schemas` are removed from `registerEntityTypeBodySchema` and `amendEntityTypeSchemasBodySchema`, an old client sending the legacy field names passes validation (no `additionalProperties: false`) and gets a successful no-op. Pre-alpha + no-compat-aliases is fine as a stance, but a silent 200 is worse than a 4xx for surfacing the migration. Adding `additionalProperties: false` on these two would fail closed.

**3. Finish the `removeTag → deleteTag` rename at the internal boundary** — `packages/agents-runtime/src/context-factory.ts:649`, `process-wake.ts:1914`

The public method, server method, and server-client method are all `delete*` now. Only the internal config field is still `doRemoveTag`. Since this PR is the explicit rename pass, easy to finish the job.

### Truly trivial polish

- `rejectPrincipalEntityMutation` interpolates the action into `"Principal entities are built in and cannot be ${action}"`; the new args `'tag updated'` / `'tag deleted'` produce `"cannot be tag updated"` / `"cannot be tag deleted"`. Other actions (`scheduled`, `killed`, `signaled`, …) still read fine — only these two are off.
- `internal-router.ts:204` introduces a local `appendPathToBackendUrl` that's byte-identical to the already-imported `appendPathToUrl` used two lines away. Probably an intentional semantic split between "backend URL" and "public URL" path handling — a one-line comment would clarify, otherwise consider sharing the helper.

### Things I checked that look fine

- `findCloudServerForUrl` longest-prefix matching correctly avoids `/t/a/v1` matching `/t/ab/v1/…` (the `startsWith(\`${basePath}/\`)` guard handles the path-segment boundary).
- `process-wake` shutdown ack ordering: `safeAckOffset` only advances at line 2011 after a wake completes and before the shutdown `break`, so cleanup acks only completed work; if nothing completed, `doneOffset === '-1'` and `sendDone` is called with `null` (release without ack).
- Callback path renames are consistent at all three construction sites (`dispatch-policy.ts:328`, `runners-router.ts:630`, `durable-streams-router.ts:334`) and the route definition (`internal-router.ts:131`).
- `cron-router.ts` deletion is clean — no dangling refs.
- `search_electric_agents_docs` system-prompt name matches the registered tool name at `knowledge-base.ts:931`.

_(Reviewed with Claude Code)_

msfstef (approved):

PR #4421, opened 2026-05-27
Add pg-sync observation support for agents

KyleAMathews:
## Summary

Adds **pg-sync observation** — a new way for agents to observe Postgres table changes via Electric shape streams and wake on matching row operations (insert/update/delete). This replaces the need for agents to poll or use webhooks for database change detection.

### Architecture

The feature spans three packages in a layered design:

1. **`agents-runtime`** — Defines the `pgSync()` observation source factory. Produces a `PgSyncObservationSource` with a deterministic `sourceRef` (content-addressed hash of canonical options) and registers it with the server via `registerPgSyncSource()`.

2. **`agents-server`** — Houses `PgSyncBridgeManager`, which manages the lifecycle of "bridges" — each bridge subscribes to an Electric `ShapeStream` for a given table/shape and forwards change messages into a Durable Stream. Bridges persist their cursor (`shapeHandle` + `shapeOffset`) across restarts so they resume from where they left off. The `PgSyncRouter` exposes a `POST /_electric/pg-sync/register` endpoint for runtime registration. Wake evaluation filters changes by operation type and collection.

3. **`agents`** — Exposes the `observe_pg_sync` tool to Horton agents, letting the LLM dynamically subscribe to table changes with optional operation filtering and debounce.

### Key design decisions

- **Canonical options & sourceRef**: Shape options are canonicalized (sorted params, defaulted replica, copied arrays) before hashing. This ensures two registrations for the same logical shape always produce the same `sourceRef`, preventing duplicate bridges. The canonicalization is exported from `agents-runtime` and reused by `agents-server` — single source of truth.

- **Initial snapshot skipping**: When a bridge starts fresh, it sets `skipChangesUntilUpToDate = true` to avoid treating the initial Electric snapshot as live changes. Changes before the first `up-to-date` control message are silently dropped. Resumed bridges (with a stored cursor) process all changes immediately.

- **Cursor as atomic unit**: `shapeHandle` and `shapeOffset` are modeled as a single `PgSyncCursor` object (`{ handle, offset }`), making the "both or neither" invariant unrepresentable as an illegal state.

- **Subscription error recovery**: If the ShapeStream subscription errors or the async message callback throws, the bridge attempts recovery — resuming from its last known cursor if available, or restarting from scratch in bootstrap-skip mode.

### Key invariants

- One bridge per `sourceRef` per tenant (enforced via `bridges` map + `starting` deduplication map)
- `sourceRef` is deterministic: same canonical options → same hash, regardless of field order
- Bridges survive server restarts via cursor persistence in `pg_sync_bridges` table
- Wake evaluation respects `ops` filters — an `insert`-only subscriber won't wake on `delete`
- `BigInt` values from Electric are converted to strings before JSON serialization

### Non-goals

- No automatic bridge cleanup/garbage collection (future work)
- No retry with exponential backoff on bridge failures (recovery is immediate restart)
- No UI for pg-sync observation management

## Verification

```bash
# Bridge manager (lifecycle, cursor persistence, recovery, BigInt safety)
pnpm --dir packages/agents-server exec vitest run test/pg-sync-bridge-manager.test.ts --reporter=verbose

# Wake delivery filtering by operation and collection
pnpm --dir packages/agents-server exec vitest run test/pg-sync-wake-delivery.test.ts --reporter=verbose

# Router registration, validation, 503 handling
pnpm --dir packages/agents-server exec vitest run test/pg-sync-router.test.ts --reporter=verbose

# Server routing integration
pnpm --dir packages/agents-server exec vitest run test/oss-server-router.test.ts -t 'pg-sync stream reads' --reporter=verbose

# Runtime sourceRef determinism (key-order independence, column-order, defaults)
pnpm --dir packages/agents-runtime exec vitest run test/pg-sync-source.test.ts --reporter=verbose

# Runtime server client round-trip
pnpm --dir packages/agents-runtime exec vitest run test/runtime-server-client-pg-sync.test.ts --reporter=verbose

# Process wake integration
pnpm --dir packages/agents-runtime exec vitest run test/process-wake.test.ts --reporter=verbose

# observe_pg_sync tool (including debounceMs: 0 regression)
pnpm --dir packages/agents exec vitest run test/observe-pg-sync-tool.test.ts --reporter=verbose

# Shutdown lifecycle
pnpm --dir packages/agents-server exec vitest run test/runtime-shutdown.test.ts --reporter=verbose
pnpm --dir packages/agents-server exec vitest run test/standalone-runtime.test.ts --reporter=verbose

# Type checking
pnpm --dir packages/agents-server typecheck
pnpm --dir packages/agents-runtime typecheck
pnpm --dir packages/agents typecheck
```

### `agents-runtime`
- **`observation-sources.ts`** — New `pgSync()` factory, `PgSyncOptions`/`CanonicalPgSyncConfig` types, `sourceRefForPgSync()`, canonical options export
- **`runtime-server-client.ts`** — `registerPgSyncSource()` HTTP client method
- **`process-wake.ts`** — Wire pg-sync source registration into wake processing
- **`entity-schema.ts`** — Add `value`/`oldValue` to `WakeChangeEntryValue`
- **`index.ts`** — Re-export pg-sync types and helpers

### `agents-server`
- **`pg-sync-bridge-manager.ts`** — `PgSyncBridge` (stream subscription + durable forwarding), `PgSyncBridgeManager` (lifecycle, deduplication, recovery)
- **`pg-sync-router.ts`** — HTTP registration endpoint with schema validation
- **`entity-registry.ts`** — `pg_sync_bridges` CRUD (upsert, cursor update/clear, touch, list)
- **`db/schema.ts`** — `pgSyncBridges` table definition + migration
- **`wake-registry.ts`** — pg-sync wake evaluation with operation filtering
- **`runtime.ts`** / **`standalone-runtime.ts`** — Bridge manager integration into server lifecycle
- **`routing/`** — Mount pg-sync router on global router

### `agents`
- **`tools/observe-pg-sync.ts`** — `observe_pg_sync` agent tool with TypeBox schema
- **`agents/horton.ts`** — Register tool in Horton's tool list

### Other
- **`scripts/dev.sh`** — Improved dev stack isolation and port configurability
- **`docs/agents-pg-sync-observation-plan.md`** — Design document


docs/agents-pg-sync-observation-plan.md
msfstef, line 1: nit: should we remove the plan to not commit it to the repo?

docs/agents-pg-sync-observation-plan.md
```diff
@@ -0,0 +1,673 @@
...
+
+```json
+{
+  "table": "todos",
+  "where": "priority = 'high'",
+  "params": [],
+  "columns": ["id", "text"],
+  "replica": "full"
+}
+```
```
msfstef: this doesn't include the actual database and/or db instance from which we are syncing - is this on purpose? don't we expect potential pg syncs from multiple sources?

packages/agents-server/src/pg-sync-bridge-manager.ts
```diff
@@ -0,0 +1,513 @@
...
+      initialSnapshotComplete,
+    }
+  }
+}
+
+export class PgSyncBridgeManager implements PgSyncBridgeCoordinator {
+  private bridges = new Map<string, PgSyncBridge>()
+  private starting = new Map<string, Promise<void>>()
+
+  private readonly shapeUrl: string
```
msfstef: does this mean we can have a single PG source for pg sync on an agent server?

packages/agents-server/src/pg-sync-bridge-manager.ts
```diff
@@ -0,0 +1,513 @@
...
+      rows.map((row) =>
+        this.ensureBridge(row).catch((error) => {
+          serverLog.warn(
+            `[pg-sync-bridge] failed to start ${row.sourceRef}:`,
+            error
+          )
+        })
+      )
+    )
+  }
```
msfstef, line 465: I am worried about the number of shapes a single server will start syncing - basically we are making our single server not proxy, but keep an open connection to _all_ shapes that have been registered by all runtimes on the server. And it doesn't do it lazily, when shapes are requested, it does it on bootup

Basically my concern is that the server becomes a single node for accumulation for all shapes registered, it doesn't efficiently stream the data in and out to durable streams as it materializes, parses, modifies, and writes the data as individual per-message writes to corresponding DS.

packages/agents-server/src/pg-sync-bridge-manager.ts
```diff
@@ -0,0 +1,513 @@
...
+        `pg-sync-bridge-${this.sourceRef}`
+      )
+    }
+    if (this.initialCursor) {
+      const offset = parseElectricOffset(this.initialCursor.offset)
+      if (offset) {
+        this.startStream(
+          offset,
+          this.initialCursor.handle,
+          !this.initialCursor.initialSnapshotComplete
```
msfstef, line 280: Is this right? Even if a snapshot is complete it might be that betweeen the snapshot and the up to date there are messages that have not been recorded on the durable stream?

packages/agents-server/src/pg-sync-bridge-manager.ts
```diff
@@ -0,0 +1,513 @@
...
+      shapeOffset,
+      initialSnapshotComplete
+    )
+    this.committedCursor = {
+      handle: shapeHandle,
+      offset: shapeOffset,
+      initialSnapshotComplete,
+    }
+  }
+}
```
msfstef, line 422: This is going to be _a lot_ of small writes to PG which I'm wondering whether it is necessary? Since we have the shape definition and also a durable stream can we not recover these with a network request either to the shape source and/or the backing durable stream to recover them so we can avoid PG churn?

msfstef (commented):
Main concerns:
1. this seems to be configurable for a single server configured shape source, which runners implicitly point to - was this intentional?
2. a lot of PG writing for persisting shape cursor when we already store the definition and have access to the source as well as a backing durable stream so I would expect us to be able to reuse those
3. worried about this not really being a proxy on demand, it makes the agent server as the central point for _all registered active shapes_ to continually sync into durable streams - unsure if this is going to be an issue but it feels potentially too "active" a role

docs/agents-pg-sync-observation-plan.md
KyleAMathews: done

docs/agents-pg-sync-observation-plan.md
KyleAMathews: I added the URL — that was missing yeah

packages/agents-server/src/pg-sync-bridge-manager.ts
KyleAMathews: yes — it could turn into a bottleneck at some point but a single node service can handle at least 10s of thousands of streams so we'll need to solve this eventually but we're already reading quite a few streams to e.g. wake up parents when a child run finishes and other triggers.

packages/agents-server/src/pg-sync-bridge-manager.ts
KyleAMathews: We didn't need the snapshot at all actually so changed to offset=now

packages/agents-server/src/pg-sync-bridge-manager.ts
KyleAMathews: if we don't record the offsets we'd have to read all the streams at start which would be quite slow... so we need to record the offsets somewhere & we can move cursor storage elsewhere if it proves a problem but I think we're fine for now.

msfstef: I meant that we could look at the head of the durable stream we are writing to to recover the latest info to resume the stream (and also we should probably have an endpoint on Electric to get latest info on a given shape if it exists tbh like the HEAD on durable streams)

packages/agents-server/src/pg-sync-bridge-manager.ts
```diff
@@ -0,0 +1,519 @@
...
+        this.bridges.set(row.sourceRef, bridge)
+      })().finally(() => this.starting.delete(row.sourceRef))
+      this.starting.set(row.sourceRef, start)
+    }
+    await start
+  }
+
+  private resolveSource(options: CanonicalPgSyncConfig): PgSyncResolvedSource {
+    return { url: options.url ?? this.url, secret: this.secret }
+  }
```
msfstef, line 545: Now we can pass the stream URL, but the secret is a single secret hardcoded on the server. This probably doesn't work since every source might have its own secret (like we do internally in cloud).

Also related to this is my slight concern with opening up arbitrary URLs for clients to pass through given that it's not a transparent proxy, we modify data write to pg etc. Maybe the agents server can be initialised with allowed sources and configured secrets? And then runners can only use sources that are already configured at the server level.

msfstef (approved):
Approving to unblock - had left one more comment on multiple sources issue

packages/agents-server/src/pg-sync-bridge-manager.ts
KyleAMathews: hmm HEAD to durable stream wouldn't return metadata about electric though. We don't have a way to store and retrieve metadata right?

packages/agents-server/src/pg-sync-bridge-manager.ts
KyleAMathews: An allow list perhaps makes sense in the future — for now I'm fine leaving it open. The secret was accidentally left behind (the coding agent hallucinated it when I first started and I thought I'd already removed it). I changed the security model to be you can get shapes from any electric server and security is through a proxy in front of that. The agents server appends a bunch of query parms w/ info about the principal who's requesting the info.

PR #4428, opened 2026-05-27
feat(agents-mobile): add Delete account button

kevin-dp:
## Summary

- Adds an in-app **Delete account** entry point on the Account screen, visible only when signed in to Electric Cloud.
- Tapping it opens https://electric-sql.com/about/legal/delete-account in the system browser, where the user can read what gets deleted, what may be retained, and email support to start the request.
- Pairs with #4427 (web page).

## Why

Google Play's account-deletion policy (https://support.google.com/googleplay/android-developer/answer/13327111) requires both an in-app option *and* a publicly accessible web URL describing the process. The web page alone won't pass review.

## UI placement

Sits below the existing Account / Workspaces / Open dashboard / Sign out card, in its own "Delete account" section with explanatory copy that the button just opens instructions — it doesn't perform the deletion directly. Uses the `ghost` button variant to visually de-emphasize it relative to the primary actions.

## Test plan

- [x] `pnpm --filter @electric-ax/agents-mobile run typecheck`
- [ ] Open the Account screen while signed in → "Delete account" section appears below the main card with the explanatory text + ghost button.
- [ ] Tap "Delete account…" → system browser opens to https://electric-sql.com/about/legal/delete-account.
- [ ] Open the Account screen while signed *out* → section is hidden (sign-in buttons shown instead, as before).

## Depends on

- #4427 — once that lands, the URL the button opens will be live.

### DIFF
--- a/.changeset/mobile-delete-account-button.md
+++ b/.changeset/mobile-delete-account-button.md
@@ -1,0 +1,9 @@
+---
+'@electric-ax/agents-mobile': patch
+---
+
+Add an in-app "Delete account" button to the Account screen (visible
+when signed in to Electric Cloud). Tapping the button opens the
+account-deletion instructions page at
+https://electric-sql.com/about/legal/delete-account in the system
+browser. Required for Google Play submission.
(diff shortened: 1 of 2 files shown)


samwillis (approved):
LGTM

PR #4431, opened 2026-05-27
feat(agents-desktop): add launch at login

samwillis:
## Summary
- Add an Electric Agents Desktop launch-at-login preference with background startup support so login launches keep the app in the tray instead of opening a window.
- Surface the setting in Settings and in a supported-platform onboarding Config step.
- Add a patch changeset for the desktop and server UI packages.
- Supports Mac and Widows, but not linux.

## Stack
- Stacked on #4430: https://github.com/electric-sql/electric/pull/4430

## Screenshots

image

image

## Test plan
- `source ~/.nvm/nvm.sh && nvm use 22.13.0 >/dev/null && pnpm --filter @electric-ax/agents-desktop typecheck`
- `source ~/.nvm/nvm.sh && nvm use 22.13.0 >/dev/null && pnpm --filter @electric-ax/agents-server-ui typecheck`

### DIFF
--- a/.changeset/warm-agents-login.md
+++ b/.changeset/warm-agents-login.md
@@ -1,0 +1,6 @@
+---
+'@electric-ax/agents-desktop': patch
+'@electric-ax/agents-server-ui': patch
+---
+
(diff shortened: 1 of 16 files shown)


kevin-dp (approved):
LGTM 👍

PR #4435, opened 2026-05-28
feat(agents-mobile): redesign onboarding to match desktop wizard

msfstef:
## Summary

- Reshapes mobile onboarding to mirror the desktop wizard's visual anatomy (step indicator, step header, section cards, pinned footer) while staying native on iOS/Android. Two steps only — Cloud sign-in and Server selection — since mobile has no local Horton runtime to configure.
- Cloud server picker rows now commit the connection on tap via a new `onConnect(url)` callback, fixing the "selecting a cloud server fills the URL input" bug. Manual self-hosted URL entry lives in a collapsible "Custom server" section with its own inline error display.
- Onboarding is now mandatory until a server connection is saved — both escape valves ("Don't show this again" on step 1, "Skip for now" on step 2) are gone, so `onboardingDismissed=true ⟹ serverUrl is set`. `ServerSetupScreen` has been rewritten on top of the same step-2 anatomy so the two screens stay aligned.
- `DiagnosticsScreen` gains a `__DEV__`-gated **Clear all local data** action (mirrors desktop's `GeneralPage` equivalent): wipes AsyncStorage, signs out of Cloud, and reloads the JS bundle into a fresh onboarding flow.

### Notable details
- Step indicator: numbered circles + inline labels, joined by a fixed-width hairline bar; completed steps render in `greenA3` with a check icon.
- Footer is pinned outside the `ScrollView` and respects `useSafeAreaInsets().bottom` so escape actions sit at thumb reach, never under the home indicator.
- `Sign in with GitHub` button gains the lucide GitHub icon; `Sign in with Google` uses the official Google "G" mark rendered as a single fill colour (brand-compliant in CTA contexts).
- Terms of service and privacy policy links open via `expo-web-browser`, matching the existing OAuth flow.
- Auto-advance from cloud → server is a one-shot per sign-in transition (`useRef`-tracked, seeded from `startStep` so warm restarts with a restored session don't silently re-advance on Back).
- Single `commit(url)` helper in both `OnboardingScreen` and `ServerSetupScreen` returns `Promise<string | null>` instead of re-throwing — caller routes the error to whichever state slot is contextually relevant.

## Test plan

- [ ] First-launch onboarding: signed-out path → step 1 GitHub/Google buttons → tap **Continue without Cloud** → step 2 with cloud sign-in prompt + custom server section → tap **Add custom URL**, enter `http://<lan-ip>:4437`, tap **Connect** → lands on home.
- [ ] Signed-in path: cloud sign-in succeeds → auto-advances to step 2 → tap a cloud row → row shows `Connecting…` → lands on home. URL input never visible.
- [ ] Back/Forward navigation: from step 2 tap **Back** → step 1 shows signed-in confirmation card → tap **Continue** → back on step 2. No bouncing or silent re-advance.
- [ ] Warm restart with a restored cloud session: wizard opens straight on step 2; tap **Back** → step 1 stays put (does not auto-advance back to step 2).
- [ ] Validation: open **Custom server** panel, tap **Connect** with empty input → "Enter an agents server URL." error appears under the field. Tap **Cancel** → panel collapses, error is cleared.
- [ ] Cloud-row connect failure: error appears between picker and Custom server section (not inside the panel).
- [ ] Mandatory wizard: relaunching the app before a server URL is saved drops back into the wizard — there is no skip/dismiss escape.
- [ ] `ServerSetupScreen` (Settings → Server): cloud row tap commits directly; URL input is never auto-populated by the picker.
- [ ] Dev: Diagnostics → **Clear all local data** → confirm → app restarts into onboarding.
- [ ] Light + dark themes; iPhone 14 Pro and an Android with edge-to-edge enabled.

### DIFF
--- a/.changeset/mobile-onboarding-wizard-redesign.md
+++ b/.changeset/mobile-onboarding-wizard-redesign.md
@@ -1,0 +1,28 @@
+---
+"@electric-ax/agents-mobile": patch
+---
+
+Redesign mobile onboarding to mirror the desktop wizard and make it
+mandatory until a server connection is saved.
+
+* Two-step wizard (Cloud sign-in → Server selection) sharing the desktop
+  wizard's visual anatomy — step indicator, step header, section cards,
+  pinned footer respecting safe-area insets. Mobile has no local Horton
+  runtime so the "model providers" step is omitted.
+* Cloud server picker rows commit the connection on tap via a new
+  `onConnect(url)` callback, fixing the bug where tapping a cloud row
+  populated the URL input instead of connecting. Manual self-hosted URL
+  entry lives in a collapsible "Custom server" section with inline error
+  display.
+* Onboarding is now mandatory until `onComplete` saves a URL — the
+  "Don't show this again" and "Skip for now" escape valves are removed.
+  Invariant: `onboardingDismissed=true ⟹ serverUrl is set`.
+* `ServerSetupScreen` is rewritten on top of the step-2 anatomy so the
+  Settings → Server screen and the onboarding server step stay aligned.
+* Cloud → server auto-advance is a one-shot per sign-in transition,
+  seeded from `startStep` so warm restarts with a restored session
+  don't silently re-advance when the user taps Back.
+* `DiagnosticsScreen` gains a `__DEV__`-gated **Clear all local data**
+  action that wipes AsyncStorage, signs out of Cloud, and reloads the
+  JS bundle into a fresh onboarding flow. Copy mirrors the desktop
+  Settings → General → Reset wording.
--- a/packages/agents-mobile/app/_layout.tsx
+++ b/packages/agents-mobile/app/_layout.tsx
@@ -59,9 +59,9 @@ function RootNavigator(): React.ReactElement {
   }
 
   // First-launch onboarding takes precedence over the server-setup
-  // redirect — the wizard subsumes the URL input as its step 2, and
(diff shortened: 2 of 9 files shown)


msfstef:
## Mobile onboarding
Simulator Screenshot - iPhone 14 Pro - 2026-05-28 at 16 15 11
Simulator Screenshot - iPhone 14 Pro - 2026-05-28 at 17 46 48
Simulator Screenshot - iPhone 14 Pro - 2026-05-28 at 17 47 35


## Desktop (reference)
Screenshot 2026-05-28 at 13 10 09
Screenshot 2026-05-28 at 13 42 54

msfstef:
Added a dev only clear all local data button - we can promote that to be in the non dev build as well but wasn't sure where we want it and whether we want it, currently just in diagnostics @samwillis 

Simulator Screenshot - iPhone 14 Pro - 2026-05-28 at 17 50 36

kevin-dp:
@msfstef 17:46 screenshot: “Add custom URL” button has no border and thus is inconsistent with “Sign in” button. Do we really need this “Add custom URL” button? I would rather immediately show the URL input, this also means that the explanation (left of the button) will fit on 1 line if the button is removed. But doing that it would look like the 17:47 screenshot but than without the “Cancel” button.

kevin-dp (approved):
Had a look at the code and tested out the app, working as expected 👍

PR #4437, opened 2026-05-28
fix(sync-service): match Postgres LIKE semantics for newlines and escaped wildcards

sravan27:
## Summary

`Electric.Replication.PostgresInterop.Casting.like?/3` — used by `LIKE`/`ILIKE`
in shape `where` clauses — compiles a SQL `LIKE` pattern into a regex anchored
with `^…$` and **without** the dotall flag. That diverges from Postgres `LIKE`
semantics in three ways. Because `like?/3` decides row membership for a shape
filter, each divergence can **silently include or exclude the wrong rows**
relative to Postgres.

## Divergences (reduced repros, at the function level)

| Input | Postgres | `Casting.like?` (before) |
|---|---|---|
| `like?("a\nb", "a%b")` | `true` | `false` |
| `like?("a\nb", "a_b")` | `true` | `false` |
| `like?("ab\n", "ab")` | `false` | `true` |
| `like?("hell%", "hell\\%")` | `true` | `false` |
| `like?("hell_", "hell\\_")` | `true` | `false` |

Root causes:

1. `%`→`.*` and `_`→`.`, but `.` does not match newlines without the dotall
   flag, so the wildcards don't span newlines (Postgres' do).
2. The pattern is anchored with `^…$`; in PCRE/`:re`, `$` also matches *before*
   a trailing newline, so `'ab\n' LIKE 'ab'` wrongly matches. Postgres requires
   the pattern to cover the whole value.
3. Splitting on `(?<!\\)[_%]` leaves escaped wildcards inside literal chunks, so
   `Regex.escape/1` escapes the **backslash too** — `\%` ends up matching a
   literal `\%` rather than a literal `%`.

## Fix

- Translate the pattern in a single backslash-aware pass: `\X` → literal `X`
  (so `\%`, `\_`, `\\` become literal `%`, `_`, `\`), `%` → `.*`, `_` → `.`,
  everything else → `Regex.escape/1`.
- Anchor with `\A…\z` (absolute string boundaries) instead of `^…$`.
- Compile with `:dotall` so the wildcards match newlines.

## Tests

Adds `describe "like?/2 Postgres compatibility"` in `casting_test.exs` covering
all three cases (newlines, trailing newline, escaped wildcards) plus an
`ilike?/2` case. The existing `like?` doctests are unchanged and still pass.

## Notes

- The pattern→regex mapping was cross-checked against a small, language-agnostic
  reference implementation of both the old and new behaviour against Postgres'
  documented semantics. Opened as a **draft** pending a CI / `mix test` run for
  the sync-service package.
- A trailing lone backslash is currently treated as a literal backslash;
  Postgres instead raises `LIKE pattern must not end with escape character`.
  Happy to switch to raising if you'd prefer to match that exactly.

### DIFF
--- a/.changeset/like-postgres-semantics.md
+++ b/.changeset/like-postgres-semantics.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix `LIKE`/`ILIKE` to follow Postgres semantics: `%` and `_` now match newline characters, a trailing newline in the value is no longer ignored, and a backslash-escaped `%` or `_` matches the literal character instead of the wildcard (including the backslash).
--- a/packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
+++ b/packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
@@ -94,6 +94,16 @@ defmodule Electric.Replication.PostgresInterop.Casting do
   @doc """
   LIKE function from SQL. Case sensitive by default.
 
+  Follows Postgres semantics:
+
+    * `%` matches any sequence of zero or more characters,
+    * `_` matches any single character,
+    * both wildcards also match newline characters,
+    * the pattern must match the entire string (a trailing newline in the value
+      is not ignored), and
+    * a backslash escapes the following character, so an escaped `%` or `_`
+      matches the literal character instead of acting as a wildcard.
+
   ## Examples
 
       iex> like?("hello", "hell_")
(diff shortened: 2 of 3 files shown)


icehaunter:
Hey, thanks a lot for the contribution! This is seems like a valid fix, any reason the PR is draft?

sravan27:
Thanks! It was draft only because I opened it before waiting for the full CI / sync-service validation. I moved it out of draft after the Elixir formatting, Lux integration, sync-service pg14/15/17/18, TS formatting, package typecheck/test, and Codecov checks came back clean.

I checked the remaining red job as well. The failing `ensure_sync_service_image / build_image` step is getting through the build and then failing while pushing the image:

```text
failed to push ghcr.io/electric-sql/electric/sync-service:a547c933fdf9ba9ea5456f037ad9bed0c6ed7647: denied: installation not allowed to Write organization package
```

So that looks like a fork / GitHub token / GHCR package permission issue rather than a code/test failure from this patch.

One small semantic note from the PR body is still open: a trailing lone backslash in a LIKE pattern is currently treated as a literal backslash by this patch, while Postgres raises `LIKE pattern must not end with escape character`. I left that behavior conservative, but I can switch it to raise if you want exact Postgres parity there too.

sravan27:
Thanks @icehaunter! No real reason — it just started as a draft while I reduced the repros at the `Casting.like?/3` level. It's ready for review now (un-drafted).

Each of the three divergences has a function-level reduced repro and a regression test (newline/dotall wildcards, the `$`-before-trailing-newline anchoring, and escaped-wildcard handling where `Regex.escape/1` was also escaping the backslash). Happy to adjust anything — naming, test placement, or splitting into smaller commits — just say the word.

PR #4442, opened 2026-05-29
feat(agents): add manifest-backed attachments

icehaunter:
## Motivation

Agents need user-provided files, starting with image attachments, without introducing a separate Assets abstraction. This stores attachment metadata in the entity manifest and raw bytes in private attachment streams so the runtime, server, and UI can share one model.

## Changes

- Adds manifest-backed attachment schema, stream path helpers, server routes for create/read/delete, and runtime `ctx.attachments` APIs.
- Hydrates image input attachments into pi-ai history for vision-capable models while keeping non-image attachments represented generically for a later pass.
- Adds UI attachment support in the active chat composer and landing composer: picker, dropzone, pasted images, draft previews, sent attachment previews, and image preview from manifest rows.
- Gates image attachment affordances based on registered model metadata so text-only models hide and reject image attach actions.
- Fixes Horton title generation when the first chat turn starts with attachments, including image-only starts.

## Implementation Notes

- Attachments are manifest entries with `role=input|output` and subjects for `inbox`, `run`, `text`, `tool_call`, and `context`; this leaves room for output attachments without wiring assistant image responses yet.
- Picker, drop, and paste currently accept images only, while the server model and renderers remain MIME-generic enough for PDFs and other file types later.
- pi-ai image support is treated as input capability metadata; models without `image` input do not receive attachment UI affordances because unsupported images are ignored by pi-ai.

## Changeset

Adds `.changeset/manifest-attachments.md` for patch releases of the affected agents packages.


packages/agents-server/src/entity-manager.ts
```diff
@@ -1833,6 +1975,164 @@ export class EntityManager {
...
+      })
+      await this.writeManifestEntry(
+        entityUrl,
+        manifestKey,
+        `upsert`,
+        attachment as unknown as Record<string, unknown>,
+        { txid }
+      )
+    } catch (error) {
+      await this.streamClient.delete(streamPath).catch(() => undefined)
```
samwillis: This cleanup can delete an existing attachment stream if `streamClient.create()` fails because the stream already exists for a duplicate/client-supplied id. Can we track whether this request actually created the stream before deleting it, and return a clean conflict for duplicate ids?

packages/agents-server/src/entity-manager.ts
```diff
@@ -1833,6 +1975,164 @@ export class EntityManager {
...
+    }
+    if (attachment.status !== `complete`) {
+      throw new ElectricAgentsError(
+        ErrCodeInvalidRequest,
+        `Attachment is not complete`,
+        409
+      )
+    }
+
+    const result = await this.streamClient.read(attachment.streamPath)
```
samwillis, line 2139: Before reading from `attachment.streamPath`, I think we should validate it is exactly the canonical path for this entity/id (`getEntityAttachmentStreamPath(entityUrl, id)`). Otherwise a forged attachment manifest row could point at another stream. The same check should apply before delete.

packages/agents-server/src/entity-manager.ts
```diff
@@ -1833,6 +1975,164 @@ export class EntityManager {
...
+    return {
+      attachment,
+      bytes: concatByteMessages(result.messages),
+    }
+  }
+
+  async deleteAttachment(
```
samwillis, line 2146: Delete should probably mirror create's write guards: reject stopped/non-writable entities and respect fork work locks before mutating attachment manifest/bytes.

packages/agents-server-ui/src/lib/sendMessage.ts
```diff
@@ -142,6 +147,100 @@ function readSendError(status: number, body: string): Error {
...
+  position?: string
+  attachments?: Array<File>
+  from?: string
+}): Promise<void> {
+  const url = entityApiUrl(baseUrl, entityUrl, `/send`)
+  const sender = await resolveSenderPrincipalUrl(
+    url,
+    from ?? getConfiguredActivePrincipal() ?? ``
+  )
+  await uploadMessageAttachments({
```
samwillis: This upload-then-send sequence can leave orphan attachment streams/manifests if uploads succeed and `/send` fails. For this first PR, a compensating rollback is probably enough: track uploaded attachment ids and delete them on send failure. A combined server endpoint can be follow-up work.

packages/agents-server-ui/src/components/views/NewSessionView.tsx
```diff
@@ -695,19 +714,65 @@ function DefaultAgentComposer({
...
     const cleaned: Record<string, unknown> = {}
     for (const [k, v] of Object.entries(args)) {
       if (v !== undefined && v !== ``) cleaned[k] = v
     }
-    onSubmit(trimmed, cleaned)
-  }, [args, disabled, onSubmit, submitting, value])
-
-  const isActive = Boolean(value.trim() && !disabled && !submitting)
+    onSubmit(trimmed, cleaned, files)
+    clearAttachments()
```
samwillis: This clears attachment drafts immediately after calling `onSubmit`, but the spawn/send path can still fail. Since `submitting` is also set above and not reset on failure, the new-session composer can get stuck disabled and the user loses the selected files. Can we only clear after success and reset submitting in the failure/finally path?

packages/agents-server-ui/src/components/UserMessage.tsx
```diff
@@ -73,6 +105,60 @@ export const UserMessage = memo(function UserMessage({
...
+          objectUrl={objectUrl}
+          open={previewOpen}
+          onOpenChange={setPreviewOpen}
+        />
+      </>
+    )
+  }
+
+  return (
+    <a className={styles.fileAttachment} href={attachment.url} title={label}>
```
samwillis: This raw `<a href>` bypasses `serverFetch`, so downloads won't include configured auth/cloud headers in web contexts. Can we route downloads through an authenticated blob download helper? That helper should also be usable from image previews so users have an explicit authenticated download action for original images.

packages/agents-server-ui/src/components/AttachmentImagePreviewDialog.tsx
```diff
@@ -0,0 +1,110 @@
...
+    open && !objectUrl
+  )
+  const imageUrl = objectUrl ?? fetchedObjectUrl
+  const size = formatAttachmentSize(attachment.byteLength)
+  const label = attachment.mimeType ? `${attachment.mimeType} · ${size}` : size
+
+  return (
+    <Dialog.Root open={open} onOpenChange={onOpenChange}>
+      <Dialog.Content className={styles.imagePreviewDialog}>
+        <div className={styles.imagePreviewToolbar}>
```
samwillis, line 37: Can we add a download button in this toolbar for the original image? It should use the same authenticated download helper as non-image attachments rather than a raw link, so preview and download behavior work consistently in cloud/authenticated deployments.

packages/agents-runtime/src/context-factory.ts
```diff
@@ -265,6 +311,65 @@ export function createHandlerContext<TState extends StateProxy = StateProxy>(
...
+              attachment.status !== `complete` ||
+              !attachment.mimeType.startsWith(`image/`)
+            ) {
+              return {
+                type: `text`,
+                text: `[attachment: ${attachment.filename ?? attachment.id}, type=${attachment.mimeType}, size=${attachment.byteLength ?? `unknown`}]`,
+              }
+            }
+            try {
+              const bytes = await attachmentsApi.read(block.id)
```
samwillis, line 419: This hydrates every accepted image into base64 right before the model call. Since budgeting currently treats attachment blocks very cheaply, a conversation with many/large images can produce a huge request. I don't think we need full summarisation in this PR, but we should add a simple guardrail here: e.g. newest N images and/or max total image bytes per run, with extras converted to text descriptors.

samwillis (changes requested):
Reviewed by Sam Willis in conjunction with GPT-5.5 in an interactive session.

Thanks, this is a useful first slice of image attachments. The overall shape feels right to me: keep bytes out of the entity timeline, store metadata as manifest rows, use private closed attachment streams for blobs, then hydrate image attachments at the runtime boundary.

I think there are a few things we should tighten before merging, mostly around integrity/lifecycle rather than broader context-management concerns. I do not think this PR needs to solve full summarisation, old-image retrieval, or general context management, but it should be robust for the narrow promise of sending image attachments to an agent prompt.

For this PR, I think the minimum bar is:

- No accidental data loss around duplicate attachment ids or failed creates.
- No cross-stream/cross-entity reads from forged attachment manifest metadata.
- No orphaned uploads on common send/spawn failure paths, or at least a clear rollback/cleanup path.
- No UI state where a failed attachment send leaves the composer disabled or loses the user’s selected files.
- Authenticated download behavior for attachments, including an explicit download action for image previews.
- A simple prompt-size guardrail so we do not hydrate an unbounded number/size of images into one model request.

I would explicitly leave the bigger context-management work for follow-up PRs. Important follow-ups I’d track:

- Define the long-term image context policy: recent-only, byte/token budgeting, provider-specific limits, and what happens when older images fall out of the active prompt.
- Add durable image summarisation so older images can remain useful as text context without resending pixels.
- Add an agent-facing tool to load a previous attachment/image into context on demand.
- Add a transactional spawn/send-with-attachments API so the client does not need to coordinate multi-step upload/send flows.
- Add attachment lifecycle/GC for entity deletion/kill, abandoned uploads, and failed/pending attachment states.
- Improve optimistic UI with local thumbnails, upload progress, retry/error states, and restoring drafts on failure.
- Expand beyond image inputs later: PDFs/text files/audio, output attachments, and subjects for run/text/tool_call/context.
- Add indexing/caching so attachment reads do not replay the full main stream for every download or runtime hydration.

So: I’m supportive of the design direction, and I’d keep this PR focused on making image attachments safe and predictable rather than trying to solve the whole context/summarisation story here.

icehaunter:
Addressed the actionable review items in `c8cfc9169`:

- hardened attachment create/read/delete: duplicate create conflicts no longer delete pre-existing streams, reads/deletes validate canonical stream paths, and deletes now respect entity write/fork guards
- added best-effort rollback for uploaded message attachments when upload/send fails, using client-generated attachment ids so cleanup has stable ids
- fixed composer failure handling so attachment drafts are not cleared on failed new-session sends, and active composer failures restore the draft files/text when possible
- replaced raw attachment links with authenticated blob downloads and added an authenticated original-image download button in the image preview toolbar
- added a simple image hydration guardrail: newest 4 images, max 10 MiB total; omitted images become text descriptors
- fixed the raw Durable Streams pass-through guard so non-attachment DS paths without an entity manager still proxy correctly
- added focused server tests for duplicate attachment conflicts, forged stream paths, and delete write guards

samwillis (approved):
Perfect!

PR #4443, opened 2026-05-29
fix(agents-desktop): prefer live Cloud server metadata

KyleAMathews:
## Summary
- Overlay live Electric Cloud agent server metadata onto saved desktop server entries
- Prefer the Cloud shape stream name for saved Cloud servers, with the local settings name as fallback
- Reuse the computed Cloud flag when building available server entries

## Testing
- pnpm install
- pnpm --dir packages/agents-server-ui typecheck
- git diff --check

### DIFF
--- a/.changeset/cloud-server-shape-overlay.md
+++ b/.changeset/cloud-server-shape-overlay.md
@@ -1,0 +1,5 @@
+---
+'@electric-ax/agents-server-ui': patch
+---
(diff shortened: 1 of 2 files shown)


samwillis (approved):
LGTM
Just missing a changeset.

icehaunter (approved):

PR #4444, opened 2026-05-29
fix(agents-mobile): prevent stream transition flash

KyleAMathews:
## Summary
- theme the native DOM/WebView background used by the mobile session embed
- avoid a white flash while the stream DOM bundle boots, especially in dark mode
- document that fresh worktrees need `pnpm install` before typecheck/tests

## Verification
- `pnpm install` from worktree root
- `pnpm run build` in `packages/agents-runtime`
- `pnpm run typecheck` in `packages/agents-mobile`

### DIFF
--- a/.changeset/mobile-stream-flash.md
+++ b/.changeset/mobile-stream-flash.md
@@ -1,0 +1,5 @@
+---
+"@electric-ax/agents-mobile": patch
+---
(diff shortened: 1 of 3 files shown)


samwillis (approved):
LGTM

PR #4447, opened 2026-05-29
Keep desktop runtime awake while local runtime is active

KyleAMathews:
## Summary
- prevent app suspension while the desktop local runtime is starting or running
- add a persisted desktop setting to disable that behavior
- expose the setting in Settings → General

## Testing
- pnpm --filter @electric-sql/client build
- pnpm --filter @electric-ax/agents-runtime build
- pnpm --filter @electric-ax/agents build
- cd packages/agents-desktop && pnpm typecheck
- cd ../agents-server-ui && pnpm typecheck

### DIFF
--- a/.changeset/desktop-runtime-awake.md
+++ b/.changeset/desktop-runtime-awake.md
@@ -1,0 +1,6 @@
+---
+'@electric-ax/agents-desktop': patch
+'@electric-ax/agents-server-ui': patch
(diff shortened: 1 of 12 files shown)


samwillis (approved):
Tested and LGTM.
I've added it to the onboarding config screen and the system type icon menu too.

PR #4449, opened 2026-05-29
Fix assistant history text block merging

KyleAMathews:
## Summary
- merge adjacent reconstructed assistant text messages into the previous text block instead of adding multiple text blocks
- keep assistant text + tool call grouping intact
- add a regression test for adjacent assistant text reconstruction

## Testing
- pnpm --dir packages/agents-runtime test test/pi-adapter.test.ts

This avoids OpenAI Responses duplicate fallback item ids like `msg_153` when replayed assistant history contains multiple unsigned text blocks in a single assistant message.


icehaunter:
Needs a rebase on latest main

icehaunter (approved):

PR #4451, opened 2026-06-01
feat(agents): add Kimi Moonshot provider support

icehaunter:
## Motivation

The desktop UI can configure Anthropic, OpenAI, Codex, and DeepSeek credentials for local Horton runtimes, but it could not configure Kimi / Moonshot API keys or select Moonshot-hosted models. Users who want to run Horton against Moonshot's OpenAI-compatible API needed a first-class provider path from desktop credentials through model selection and runtime execution.

## What changed

- Added a Moonshot provider definition in agents-runtime with the official Moonshot API base URL, MOONSHOT_API_KEY lookup, and supported Kimi / Moonshot model metadata.
- Taught provider detection, low-cost model selection, and the pi-ai adapter to resolve Moonshot models and pass the runtime API key explicitly.
- Added Moonshot to the built-in Horton model catalog, including model availability probing via the Moonshot /models endpoint and model config API-key handoff.
- Threaded a new moonshot credential slot through the Electron desktop main process, saved API key normalization, environment mirroring, onboarding, and Credentials settings UI.
- Updated local runtime startup messaging and dev preflight checks so MOONSHOT_API_KEY counts as a valid model provider key.
- Added a changeset covering the touched agents packages.

## Decisions

Moonshot is represented as a local provider id, moonshot, rather than using pi-ai's kimi-coding provider because the user-facing API here is Moonshot's OpenAI-compatible endpoint at https://api.moonshot.ai/v1 with MOONSHOT_API_KEY credentials. The runtime keeps model selection string-based so worker spawn arguments remain serializable, while the adapter maps Moonshot model ids to custom OpenAI-compatible model objects at execution time.

### DIFF
--- a/.changeset/kimi-moonshot-horton.md
+++ b/.changeset/kimi-moonshot-horton.md
@@ -1,0 +1,8 @@
+---
+"@electric-ax/agents-runtime": patch
+"@electric-ax/agents": patch
+"@electric-ax/agents-desktop": patch
+"@electric-ax/agents-server-ui": patch
+---
+
+Add Kimi / Moonshot API support for local Horton runtimes, including model catalog entries, runtime provider resolution, desktop credential persistence, and UI credential inputs.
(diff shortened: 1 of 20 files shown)


samwillis (approved):
LGTM

PR #4452, opened 2026-06-01
fix(agents-server-ui): surface signal failure toasts

icehaunter:
Signal and kill requests can fail after optimistic UI state has already been applied, for example when the server rejects persistence because a migration is missing. Previously those failures were easy to miss because the action callers intentionally swallowed rejected termination attempts, which made failed termination look like a no-op from the UI.

This PR adds a small server UI toast channel and mounts a global Base UI toast provider/viewport so local UI actions can surface failures without coupling every caller to notification rendering. The channel keeps the existing imperative showToast(...) API for non-React action code, then bridges those events into Base UI's toast manager.

The kill and signal optimistic action mutation functions now emit danger toasts for non-2xx responses and request-level failures, including the HTTP status and compacted server error text when available, while preserving the existing rejected mutation behavior.

The toast rendering is scoped to agents-server-ui and now uses @base-ui/react/toast primitives for provider, viewport, root, content, title, description, close, timeout handling, limits, and high-priority announcements for danger toasts. Styling stays on the existing design tokens plus lucide icons. A patch changeset is included for @electric-ax/agents-server-ui.

### DIFF
--- a/.changeset/signal-failure-toasts.md
+++ b/.changeset/signal-failure-toasts.md
@@ -1,0 +1,5 @@
+---
+"@electric-ax/agents-server-ui": patch
+---
+
+Surface failed signal and kill requests in the UI with toast notifications instead of silently swallowing persistence failures.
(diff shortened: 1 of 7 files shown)


samwillis (approved):
LGTM

PR #4454, opened 2026-06-01
feat(agents): add desktop model picker controls

icehaunter:
## Motivation

The Kimi / Moonshot provider work adds another model provider to local Horton sessions, but the session model dropdown would still expose every available model for every configured provider. That makes the picker noisy and puts provider grouping in the wrong place: users configure credentials in Settings, so the enabled-model controls should live next to those credentials and feed into Horton startup.

## What changed

- Added persisted desktop settings and IPC for enabled model values.
- Threaded enabled model values through desktop runtime startup into the built-in Horton model catalog, so disabled models are omitted from local session creation.
- Added provider headers to the new-session model dropdown and indented model rows under each provider header.
- Added Base UI collapsible model controls under each provider credential in Credentials settings, including the ChatGPT / Codex login block.
- Seeded fresh model-picker settings with the requested defaults for Anthropic, DeepSeek, Kimi / Moonshot, and OpenAI Codex, including GPT-5.5.
- Added tests for catalog filtering behavior and a changeset for the touched agents packages.

## Decisions

Enabled models are stored as provider-prefixed values, for example anthropic:claude-sonnet-4-6, so overlapping model ids from different providers remain unambiguous. The desktop status payload only exposes controls for configured providers, and the catalog falls back to the full available model list if a saved/default enabled set has no valid intersection so a bad preference cannot leave Horton without any selectable model.

This PR is stacked on #4451 because the model controls include Kimi / Moonshot as one of the provider groups.

### DIFF
--- a/.changeset/enabled-model-picker.md
+++ b/.changeset/enabled-model-picker.md
@@ -1,0 +1,7 @@
+---
+"@electric-ax/agents": patch
+"@electric-ax/agents-desktop": patch
+"@electric-ax/agents-server-ui": patch
+---
+
+Allow desktop users to choose which configured provider models appear in Horton's model picker, and group model dropdown entries by provider.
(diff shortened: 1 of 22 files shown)


samwillis (approved):
LGTM

PR #4456, opened 2026-06-01
feat(sync-service): cap logger handler OLP mailboxes

erik-the-implementer:
## What

Cap the overload-protection (OLP) mailboxes of the three logger handlers used by `sync-service` so that error/log bursts shed messages instead of blocking `Logger` callers or letting processes' mailboxes grow unbounded.

Three sites, all inside `packages/sync-service/`:

1. **Default console handler** (`config/runtime.exs`) — `sync_mode_qlen: 2000, drop_mode_qlen: 2000, flush_qlen: 5000`. Matches the stratovolt `cloud-sync-service` precedent.
2. **`OtelMetricExporter.LogHandler`** (`config/runtime.exs`) — extends the existing `:config` map with `drop_mode_qlen: 2000, flush_qlen: 5000`. `sync_mode_qlen` is intentionally **not** set: the handler module forces `sync_mode_qlen == drop_mode_qlen`, so it would be a no-op. This handler had no caps anywhere before.
3. **Sentry handler** (`lib/electric/application.ex`) — passes `discard_threshold: 2000, sync_threshold: nil` to `Electric.Telemetry.Sentry.add_logger_handler/2` (the function already accepts opts; only the call site changed). Matches the stratovolt precedent.

## Why

During the 2026-05-21 pod-replacement spikes, `logger_olp` mailboxes peaked at ~76K/~59K. Capping these mailboxes protects the leading edge of redeployment-shock overload.

`overload_kill_enable` is deliberately **not** set on either OLP handler: for the console handler it would terminate the handler and lose all subsequent logs until restart; for `OtelMetricExporter.LogHandler` it would tear down the entire `LogHandlerSupervisor`, creating a restart-amplification loop.

### DIFF
--- a/.changeset/logger-handler-caps.md
+++ b/.changeset/logger-handler-caps.md
@@ -4,6 +4,6 @@
 
 Cap the overload-protection (OLP) mailboxes of the default console, OpenTelemetry, and Sentry logger handlers so error/log bursts shed messages instead of blocking Logger callers or growing unbounded.
 
-This is **leading-edge protection only**: it shields against the early phase of a redeployment/error burst but is **not sufficient under deep scheduler starvation** — the real fix for that is upstream (request-proxy admission control and snapshot-pool sizing). See electric-sql/alco-agent-tasks#45 §3 for the analysis and caveats.
+This is **leading-edge protection only**: it shields against the early phase of a redeployment/error burst but is **not sufficient under deep scheduler starvation** — the real fix for that is upstream (request-proxy admission control and snapshot-pool sizing).
 
 Note: `sync_mode_qlen` is intentionally not set on the OpenTelemetry log handler — its module forces `sync_mode_qlen == drop_mode_qlen`, so the option would be a no-op there.
--- a/.changeset/logger-handler-caps.md
+++ b/.changeset/logger-handler-caps.md
@@ -1,0 +1,9 @@
+---
+'@core/sync-service': patch
+---
+
+Cap the overload-protection (OLP) mailboxes of the default console, OpenTelemetry, and Sentry logger handlers so error/log bursts shed messages instead of blocking Logger callers or growing unbounded.
+
+This is **leading-edge protection only**: it shields against the early phase of a redeployment/error burst but is **not sufficient under deep scheduler starvation** — the real fix for that is upstream (request-proxy admission control and snapshot-pool sizing). See electric-sql/alco-agent-tasks#45 §3 for the analysis and caveats.
+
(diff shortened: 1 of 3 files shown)


erik-the-implementer:
Thanks for the thorough review. Disposition of the items below.

### Important — "set `sync_mode_qlen: 2000` defensively on the OTel handler"

Declining this one, and it's worth spelling out why because the suggested fix would actually **break** the handler rather than be a free safety net.

Tracing `OtelMetricExporter.LogHandler.adding_handler/1` (deps `otel_metric_exporter`, pinned `0.4.4`):

1. `@olp_config_keys` (`log_handler.ex:48-58`) is the allowlist that `Map.split/2` uses to peel OLP options out of the `:config` map. **`:sync_mode_qlen` is not in that list** — only `:drop_mode_qlen`, `:flush_qlen`, the `burst_limit_*` and `overload_kill_*` keys are.
2. Anything *not* in that allowlist stays in `accumulator_config`, which is then passed to `LogAccumulator.check_config/2` → `NimbleOptions.validate/2` against a fixed schema (`log_accumulator.ex:68-77`). An unrecognized key (`sync_mode_qlen`) is rejected there, so `adding_handler/1` returns an error and **the handler fails to install**.
3. The effective `sync_mode_qlen` is set unconditionally by `prevalidate_olp/1` (`log_handler.ex:79`): `Map.put(olp_config, :sync_mode_qlen, Map.get(olp_config, :drop_mode_qlen, 200))`. It is *always* derived from `drop_mode_qlen` (which we set to 2000) and never read from user input. So there is no path where it falls back to `10` — the premise of the concern doesn't hold for this dependency.

So omitting `sync_mode_qlen` is both required (issue hard-constraint) and correct: with `drop_mode_qlen: 2000` the effective `sync_mode_qlen` is already 2000. The dependency is version-pinned via `mix.lock`, so the contract can't drift silently — a future bump is the right place to re-verify. I've left the code as-is and the inline comment documents the no-op rationale.

### Suggestions — deferred (out of scope for this change)

This issue is a deliberately verbatim port of the stratovolt precedent plus the one new OTel cap, so I'm keeping the surface minimal and not expanding scope:

- **prod-only gating of `:default_handler` caps** — stratovolt applies the same block unconditionally, and the existing `config :logger, :default_handler, level: ...` under `:test` is also unconditional; matching that precedent. The thresholds (2000) sit far above anything a test queues, so test behavior is unchanged in practice.
- **env-var overrides for the thresholds** — agreed this is reasonable, but as you note it's non-blocking; deferring until a real tuning need motivates it.
- **boot-time verification probe / cross-repo link distillation** — nice-to-haves; happy to add the probe in a follow-up if the team wants a guard against `runtime.exs` handler config silently failing to apply, but it's broader than this config tweak.

No code changes pushed since the one actionable item is declined on the grounds above.

robacourt (approved):

PR #4457, opened 2026-06-01
fix(agents): preserve volatile context source order

KyleAMathews:
## Summary

Remove a global `.sort()` by `at` timestamp in `assembleContext()` that was undoing the semantic reordering performed by `reorderInterruptedRuns()`. When SIGINT races ahead of buffered runtime writes, the interrupted assistant run now correctly appears before the interrupt marker in the model transcript.

## Root Cause

`assembleContext()` collected volatile messages from all sources, then globally sorted them by their `at` (stream offset) value. Meanwhile, `reorderInterruptedRuns()` in the timeline layer intentionally reorders aborted runs before the SIGINT signal that raced ahead of them—giving the model a coherent transcript where the partial response precedes the interruption.

The global sort undid this reordering: the SIGINT (lower offset, because it's appended directly by the server) was placed before the aborted run output (higher offset, because it's buffered through the runtime producer). The model then saw an interruption marker before any assistant output, breaking transcript coherence.

## Approach

Remove the `volatileMessages.sort()` call entirely. Volatile sources already return messages in their intended order—`timelineMessages()` projects through `materializeTimeline()` → `reorderInterruptedRuns()` to produce the correct model-facing sequence. The `at` field is preserved as metadata for budget truncation ranges and load tools, but no longer drives global ordering.

**Key invariant**: volatile messages appear in source-registration order (all messages from source A, then all from source B, etc.), with each source's internal order preserved exactly as returned.

**Trade-off**: budget truncation under pressure now drops messages from earlier-registered sources first (instead of oldest-by-timestamp first). This is the intended behavior—source ordering is more semantically meaningful than raw stream offsets.

## Non-goals

- The CI hardening commit (`2a95b8a17`) is bundled here but independent—it adds CODEOWNERS for `.github/`, fork-PR guards on benchmarking/load-test workflows, and least-privilege permission tightening.

## Verification

```bash
cd packages/agents-runtime
pnpm vitest run test/use-context-volatile-interleave.test.ts
pnpm vitest run test/timeline-context.test.ts
```

## Files changed

| File | Change |
|------|--------|
| `packages/agents-runtime/src/context-assembly.ts` | Remove `volatileMessages.sort()` call |
| `packages/agents-runtime/test/use-context-volatile-interleave.test.ts` | Update existing test expectations, add `at` assertions, three-source ordering test, and end-to-end integration test wiring `materializeTimeline` through `assembleContext` |
| `.github/CODEOWNERS` | Require admin review for `.github/` changes |
| `.github/workflows/benchmarking.yml` | Drop CONTRIBUTOR from trigger, add fork-PR guard |
| `.github/workflows/load_test.yml` | Add fork-PR guard |
| `.github/workflows/ts_tests.yml` | Downgrade `packages: write` → `read` |
| `.github/workflows/changesets_release.yml` | Remove `cache: pnpm` from setup-node |
| `.changeset/fix-sigint-context-order.md` | Changeset for `@electric-ax/agents-runtime` patch |

### DIFF
--- a/.changeset/fix-sigint-context-order.md
+++ b/.changeset/fix-sigint-context-order.md
@@ -1,0 +1,5 @@
+---
+'@electric-ax/agents-runtime': patch
+---
+
+Preserve volatile context source order in `assembleContext()` instead of globally sorting by `at` timestamp. Fixes a bug where the SIGINT reordering performed by `reorderInterruptedRuns()` was undone by a downstream sort, causing interrupted run output to appear after the interrupt marker in the model transcript.
--- a/packages/agents-runtime/src/context-assembly.ts
+++ b/packages/agents-runtime/src/context-assembly.ts
@@ -313,7 +313,12 @@ export async function assembleContext(
     }
   }
 
-  volatileMessages.sort((left, right) => left.at - right.at)
+  // Preserve volatile source order. Volatile sources such as
+  // ctx.timelineMessages() may return a model-facing transcript order that
+  // intentionally differs from raw stream-offset order, e.g. interrupted run
+  // output before the SIGINT marker that raced ahead of buffered runtime
+  // writes. `at` remains metadata for truncation ranges and load tools, but it
+  // must not globally reorder already-projected context.
 
   const remainingBudget = Math.max(0, config.sourceBudget - budgetUsed)
   const accepted: Array<VolatileMessage> = []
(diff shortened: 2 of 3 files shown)


kevin-dp (approved):
This is changing the configuration of several CI workflows that seem unrelated to the goal of this PR.

.github/workflows/changesets_release.yml
kevin-dp, line 36: Why removing this?

PR #4460, opened 2026-06-01
Fix stale Codex auth state in desktop

KyleAMathews:
## Summary
- clear stale Codex desktop auth when no usable access token can be produced
- keep desktop OAuth refresh as the only refresh path for stored credentials
- avoid showing Codex as enabled when the runtime cannot set ELECTRIC_CODEX_ACCESS_TOKEN

## Test
- pnpm --filter @electric-ax/agents-desktop typecheck

### DIFF
--- a/packages/agents-desktop/src/credentials/codex-auth.ts
+++ b/packages/agents-desktop/src/credentials/codex-auth.ts
@@ -283,19 +283,44 @@ async function refreshCodexAuthIfNeeded(
   return next
 }
 
+async function clearCodexAuth(
+  deps: Pick<
+    CodexAuthDeps,
+    `settings` | `getSecretStore` | `saveSettings` | `markCredentialsDirty`
+  >
+): Promise<void> {
+  deps.settings.codex = { enabled: false, source: null }
+  await deps.getSecretStore().delete(CODEX_AUTH_REF)
+  await deps.saveSettings()
+  deps.markCredentialsDirty()
+}
+
 export async function syncCodexEnvironment(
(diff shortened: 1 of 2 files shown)


packages/agents-desktop/src/credentials/codex-auth.ts
```diff
@@ -283,19 +283,44 @@ async function refreshCodexAuthIfNeeded(
   return next
 }
 
+async function clearCodexAuth(
+  deps: Pick<
```
kevin-dp: This `Pick`ed type is the most precise but also verbose and requires more maintenance (any time one of these props changes). I would keep it simple here: `deps: CodexAuthDeps`

packages/agents-desktop/src/credentials/codex-auth.ts
```diff
@@ -283,19 +283,44 @@ async function refreshCodexAuthIfNeeded(
...
+): Promise<void> {
+  deps.settings.codex = { enabled: false, source: null }
+  await deps.getSecretStore().delete(CODEX_AUTH_REF)
+  await deps.saveSettings()
+  deps.markCredentialsDirty()
+}
+
 export async function syncCodexEnvironment(
-  deps: Pick<CodexAuthDeps, `settings` | `getSecretStore`>
+  deps: Pick<
```
kevin-dp: Same comment as before, simplify to: `deps: CodexAuthDeps`

kevin-dp (approved):

PR #4471, opened 2026-06-02
feat(agents-desktop): Add/Edit/Remove UI for MCP servers

### LINKED ISSUE #4465: Add MCP server configuration UI to Desktop
## Summary
Bring MCP server configuration into the Desktop frontend so users can add/manage MCP servers from the app instead of manually editing the global JSON config file.

## Specific context from the check-in
MCP already works, but configuration is still mostly hidden behind existing config files. The transcript called out that this was an early pass that needs to be brought into the frontend:

- Users should be able to add an MCP server from the Desktop app.
- We should avoid requiring users to modify the global JSON config manually.
- There appears to be both global config and workspace-specific config; Desktop should respect/discover the existing behavior.
- Configured MCP servers should be available to Horton/agents.
- This work is already in progress and assigned to Kevin.

## Scope
- Add Desktop/settings UI for MCP server configuration.
- Support the existing global MCP config path.
- Support or clearly surface workspace/project-specific MCP config if that is already part of the runtime behavior.
- Persist config changes made through the UI.
- Validate config enough to catch obvious errors.
- Show useful error/status feedback when an MCP server fails to start/connect.
- Ensure configured servers are exposed to Horton/agents in new sessions.

## Nice-to-have / possible follow-up
- Bundle or one-click-enable Playwright MCP so Horton has browser automation available by default.
- Expose MCP-related actions from the composer plus menu if that fits the UI direction.

## Acceptance criteria
- A user can add an MCP server from Desktop UI without editing JSON manually.
- A user can see configured MCP servers and whether they are available/healthy.
- Config survives app restart.
- Horton/agents can use configured MCP servers.
- Existing manually configured global MCP servers continue to work.
- Workspace/project-specific MCP config behavior is either supported or explicitly documented.
- Invalid config or failed MCP startup produces actionable feedback.

kevin-dp:
Closes https://github.com/electric-sql/electric/issues/4465

## Summary

Adds a form-based **Add / Edit / Remove** flow for MCP servers in the desktop's Settings → MCP Servers page. Before this, the only way to register a server was to hand-edit `settings.json` or a workspace `mcp.json`.

The dialog supports both transports (`http` + `stdio`), all four auth modes (`none`, `apiKey`, `clientCredentials`, `authorizationCode`), and writes through to the global `settings.json mcp.servers` block. Updates flow into the live MCP registry without a runtime restart.

### `@electric-ax/agents` — `BuiltinAgentsServer.setExtraMcpServers(extras)`

Lifts the previously-closure-captured `extras` + last-known mcp.json state onto the class so a public `setExtraMcpServers()` method can replay the same merge path the workspace `mcp.json` watcher already uses. Workspace `mcp.json` still wins on name collision (same precedence rule, single code path).

### `@electric-ax/agents-desktop`

- New IPC verbs `desktop:mcp-upsert` + `desktop:mcp-remove` in `ipc/mcp.ts`.
- `runtime/mcp.ts`:
  - `upsertMcpServer` / `removeMcpServer` — persist to `settings.json mcp.servers`, fire-and-forget push to `setExtraMcpServers`, immediate snapshot rebroadcast so the dialog closes the moment the disk write is done (the transport rebuild — npm install for stdio, MCP handshake — runs in the background and the row animates `connecting → ready`).
  - `enrichSnapshot` — produces a UI-friendly snapshot that knows where each server came from (`settings` / `workspace` / `extras`) and surfaces shadow rows when a `settings.json` name collides with a workspace `mcp.json` entry.
- `settings/store.ts` — `validateMcpServerConfig`, `upsertMcpServerInSettings`, `removeMcpServerFromSettings` helpers (regex-validated against Anthropic's tool-name shape since the server name becomes the `mcp__<server>__<tool>` prefix).

### `@electric-ax/agents-server-ui`

- New `McpServerFormDialog` covering Add + Edit. Per-transport fields, per-auth-mode subforms, name regex validation, "would overwrite" warning on Add when the name already exists in `settings.json`.
- `McpServersPage` gains an Add CTA, per-row Edit + Remove on `settings` rows, "from mcp.json" + "programmatic" provenance badges, grayed-out treatment + "overridden by mcp.json" badge on shadowed `settings` rows that render next to the running workspace twin.

## Notes for reviewers

- **Secrets posture unchanged.** `apiKey.key` and `clientCredentials.clientSecret` still land in `settings.json` plaintext — same as today. Routing them through `SecretStore.encryptString` was discussed and deferred (out of scope).
- **Add only writes to `settings.json`.** The desktop never writes to workspace `mcp.json`; other tools own that file (and it can be checked into git).
- **Workspace-wins still holds.** Editing a workspace-shadowed `settings.json` entry only affects the `settings.json` copy. The row stays grayed-out until the workspace entry is removed.
- **Optimistic close.** Save resolves as soon as `settings.json` is written. Errors during the registry rebuild surface in the row's `error` status, not the dialog.

## Test plan

- [x] `pnpm --filter @electric-ax/agents test` — 48 tests pass, including a new `setExtraMcpServers re-applies merged config at runtime` case in `bootstrap-mcp.test.ts`.
- [x] `pnpm --filter @electric-ax/agents typecheck`; same for `agents-desktop`, `agents-server-ui`.
- [x] Manual smoke (local dev): Add a stdio MCP server (`npx -y @modelcontextprotocol/server-everything`) via the form → row reaches `ready` with the expected tool list → spawn a fresh Horton session → `mcp__everything__echo` resolves end-to-end.
- [ ] Reviewer should verify the shadowing path: with the smoke MCP added via the form, drop a workspace `mcp.json` containing a server with the same name and confirm the settings row goes grayed-out next to the workspace twin.

### DIFF
--- a/packages/agents-server-ui/src/components/settings/pages/McpServerFormDialog.tsx
+++ b/packages/agents-server-ui/src/components/settings/pages/McpServerFormDialog.tsx
@@ -130,7 +130,29 @@ function parseEnv(input: string): Record<string, string> | undefined {
   return Object.keys(env).length > 0 ? env : undefined
 }
 
-function buildConfig(state: FormState): {
+/**
+ * Pull the baseline auth object out of an existing config — but only if
+ * its mode matches what the form has selected. When the user switches
+ * auth modes the prior shape is structurally incompatible (different
+ * required fields, different unknowns), so we start fresh.
+ */
+function baselineAuthForMode(
+  baseline: McpServerConfigInput | undefined,
+  mode: AuthMode
+): Record<string, unknown> {
+  if (!baseline || baseline.transport !== `http`) return {}
+  const existing = baseline.auth as Record<string, unknown> | undefined
+  if (!existing || existing.mode !== mode) return {}
+  // Shallow copy — we only overwrite top-level auth fields below, so
+  // a structural clone is overkill; sub-objects (`client`, `tokens`)
+  // are preserved by reference but never mutated in place.
+  return { ...existing }
+}
+
+function buildConfig(
+  state: FormState,
+  baseline?: McpServerConfigInput
+): {
   cfg?: McpServerConfigInput
   error?: string
 } {
@@ -157,11 +179,21 @@ function buildConfig(state: FormState): {
       auth = { mode: `none` }
     } else if (state.authMode === `apiKey`) {
       if (!state.apiKey.key.trim()) return { error: `API key is required` }
-      auth = { mode: `apiKey`, key: state.apiKey.key }
+      auth = {
+        ...baselineAuthForMode(baseline, `apiKey`),
+        mode: `apiKey`,
+        key: state.apiKey.key,
+      }
       if (state.apiKey.headerName.trim()) {
         auth.headerName = state.apiKey.headerName.trim()
+      } else {
+        delete auth.headerName
       }
-      if (state.apiKey.valuePrefix) auth.valuePrefix = state.apiKey.valuePrefix
+      if (state.apiKey.valuePrefix) {
+        auth.valuePrefix = state.apiKey.valuePrefix
+      } else {
+        delete auth.valuePrefix
+      }
     } else if (state.authMode === `clientCredentials`) {
       if (
         !state.clientCredentials.tokenUrl.trim() ||
@@ -172,7 +204,11 @@ function buildConfig(state: FormState): {
           error: `Token URL, client id, and client secret are required for clientCredentials`,
         }
       }
+      // Preserve baseline `audience`, `resource`, and any other fields
+      // we don't render — the form is a strict subset of the
+      // McpServerConfig.auth schema.
       auth = {
+        ...baselineAuthForMode(baseline, `clientCredentials`),
         mode: `clientCredentials`,
         tokenUrl: state.clientCredentials.tokenUrl.trim(),
         clientId: state.clientCredentials.clientId.trim(),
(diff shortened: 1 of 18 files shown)


packages/agents-server-ui/src/components/settings/pages/McpServerFormDialog.tsx
```diff
@@ -0,0 +1,604 @@
...
+        clientSecret: state.clientCredentials.clientSecret,
+      }
+      const scopes = splitCsv(state.clientCredentials.scopes)
+      if (scopes.length > 0) auth.scopes = scopes
+    } else {
+      auth = { mode: `authorizationCode` }
+      const scopes = splitCsv(state.authorizationCode.scopes)
+      if (scopes.length > 0) auth.scopes = scopes
+    }
+    const cfg: McpServerConfigInput = {
```
samwillis: This rebuilds the submitted config only from fields rendered by the form. On edit, that can silently drop valid fields from existing hand-edited settings entries, e.g. `clientCredentials.audience`, `clientCredentials.resource`, `authorizationCode.resource`, or `authorizationCode.redirectUri`. Could we preserve the original `initial` config as the base in edit mode and patch only the fields the form owns, or expose those advanced fields explicitly?

packages/agents-server-ui/src/components/settings/pages/McpServerFormDialog.tsx
```diff
@@ -0,0 +1,604 @@
...
+              disabled={isEdit}
+              onChange={(e) =>
+                setState((s) => ({ ...s, name: e.target.value }))
+              }
+              placeholder="my-mcp-server"
+            />
+          </Field>
+
+          <Field label="Transport" required>
+            <select
```
samwillis, line 356: This should use our Base UI-backed `Select` primitive from `../../../ui` rather than a raw `<select>`. The modal itself is already using the Base UI-backed `Dialog` wrapper, so this would keep the controls consistent with the rest of the UI system. Same comment applies to the auth-mode select below.

packages/agents-server-ui/src/components/settings/pages/McpServerFormDialog.tsx
```diff
@@ -0,0 +1,604 @@
...
+      onOpenChange(false)
+    } catch (e) {
+      setError(e instanceof Error ? e.message : String(e))
+    } finally {
+      setSubmitting(false)
+    }
+  }
+
+  return (
+    <Dialog.Root open={open} onOpenChange={onOpenChange}>
```
samwillis, line 322: This dialog repeats header/close/body/footer chrome that also exists in `JsonInspectDialog` and other modal usages. Before adding more settings dialogs, I think we should extract shared pieces on top of the existing Base UI-backed `Dialog` primitive, e.g. `Dialog.Header`, `Dialog.Body`, `Dialog.Footer`, and `Dialog.CloseButton` (or a small `FormDialog` shell).

samwillis (commented):
Reviewed in an interactive session with me (Sam) and GPT-5.5.

Thanks for pushing this through. The overall shape looks good: the desktop now has a real Add/Edit/Remove path for MCP servers, `settings.json` entries are pushed into live runtimes via `setExtraMcpServers()` rather than requiring a restart, and the workspace-wins precedence is carried through into the UI with provenance/shadowing. That is the right model for this feature.

A few things I think we should tighten before merging:

1. **Preserve advanced config fields on edit.** The new form rebuilds the submitted config from only the fields it renders. That means hand-edited `settings.json` entries can lose valid MCP config such as `clientCredentials.audience`, `clientCredentials.resource`, `authorizationCode.resource`, `authorizationCode.redirectUri`, and potentially other fields we intentionally allow in the lower-level config. Since one of the migration paths here is from hand-edited config to UI-managed config, edits should either preserve unknown/unrendered fields from the original entry or expose those advanced fields explicitly.

2. **Use the Base UI-backed `Select` primitive.** The modal itself is Base UI-backed through our local `Dialog` wrapper, and most controls already use local UI primitives, but the transport/auth controls are raw `<select>` elements. We have a Base UI-backed `Select` in `src/ui`, and we should use it here for consistency with the rest of the desktop UI.

3. **Generalise the modal chrome.** `McpServerFormDialog` repeats the same header/close/body/footer structure we already hand-roll in `JsonInspectDialog` and partially in confirm-style dialogs. I do not think we need a new modal system, but it would be useful to extend the existing `Dialog` primitive with shared pieces like `Dialog.Header`, `Dialog.Body`, `Dialog.Footer`, and `Dialog.CloseButton`, or add a small `AppDialog`/`FormDialog` wrapper. Then this MCP form and the inspect dialog can share the same Base UI modal shell while keeping feature-specific content.

I am not concerned about secrets being plaintext in `settings.json` for this PR. The UI calls that out, and moving these through a secret store can be a separate hardening pass.

samwillis (approved):
Proactively approving ahead of the comments above being done so that you can merge when it's ready.

kevin-dp:
@samwillis — re point 2 (use the base-ui `Select`): I gave this a real go and dropped the change in the end. It's a three-way interaction between base-ui's Dialog modal, base-ui's Select portal, and the Dialog popup's `overflow: auto` scroll, and none of the standard workarounds held together cleanly:

- `Select modal={false}` + default Dialog modal → click registers, but base-ui Dialog's pointer-events lock on body siblings freezes the portaled Select popup (items unclickable).
- `Dialog modal="trap-focus"` + portaled Select → same symptom; `trap-focus` still treats `Select.Portal` as outside-the-dialog and dismisses it on `mouseup`, so the popup only stays open while the mouse button is held.
- `Select.Content inPortal={false}` (rendered inline) → opens and stays open, but the Positioner's floating-ui anchor calculation goes wrong inside the dialog's flex column — popup lands at the bottom of the dialog instead of under the trigger.
- `Dialog.Content` exposing its popup element via context + `Select.Portal container={dialogPopupEl}` + `positionMethod="fixed"` on the Positioner → click opens but `mouseup`-outside-press still kills it.

Pushed:
- `52dceda4c` — review 1: preserve unrendered fields on edit.
- `123bd2032` — review 3: `Dialog.Header`/`Body`/`Footer`/`CloseButton` + `McpServerFormDialog` refactored to use them.

PR #4475, opened 2026-06-02
feat(agents-server): add permission enforcement

### LINKED ISSUE #4464: Implement initial agents permissions model
## Summary
Implement the 0.1 agents permissions model agreed in the check-in: explicit per-entity grants, private-by-default Desktop sessions, workspace-level sharing, user-delegated spawned agents, and spawn-time grants. Do **not** rely on external permission webhooks for the first pass.

## Specific answer from the discussion
The direction we converged on was:

1. **Workspace is the coarse boundary for 0.1**
   - If you are outside the workspace, you should not have access.
   - For now, we can tolerate workspace-level rather than per-service roles.
   - Longer-term per-service roles can be a follow-up.

2. **Add explicit grants for sessions/entities**
   - The missing primitive is an actual grants table/model plus APIs.
   - Permissions should not be just global rules or webhook callbacks.
   - Grants are attached to the specific agent/session/entity.

3. **Desktop/Horton-created sessions are private by default**
   - If I create a session on my desktop, only I have read/write access initially.
   - Other workspace members should not automatically see or write to it.

4. **Sharing is done by writing explicit grants**
   Initial grant targets/actions should cover:
   - grant read to a specific user
   - grant read/write to a specific user, e.g. pairing
   - grant read to everyone in the workspace
   - grant read/write to everyone in the workspace, e.g. bot-created shared sessions
   - optionally public/link read access if we decide to include that now

5. **Bot/shared sessions can default to workspace grants**
   - Sessions started from shared surfaces like Discord/PR-review agents can be created with workspace-wide read/write grants.

6. **The spawn API should let the spawner specify initial grants**
   - When spawning an agent/session, the caller should be able to provide the grants that should be attached to the spawned entity.
   - This is how shared/bot-created sessions can be created with workspace-wide access from the start.
   - The spawner must not be able to grant permissions they do not themselves have authority to grant.
   - If no grants are specified, use the safe default: creator/spawner read/write only.

7. **Spawned agents act on behalf of the spawning user**
   - If user Sam spawns an agent, that agent acts as/delegates from Sam.
   - It can use Sam’s permissions but cannot elevate beyond them.
   - Chained agent spawning should preserve this “on behalf of” behavior.
   - Permission narrowing/downgrading can be designed later.

8. **Cloud should not expose service secrets to ordinary Desktop users**
   - The immediate Cloud need is a restricted role/token path so Desktop users can connect/use the agents service without receiving broad admin/service-secret access.
   - This can be workspace-level for now.

9. **Defer permission webhooks**
   - Calling webhooks on every durable-stream read/long-poll is too expensive for the first pass.
   - Webhook auth should be a later phase, likely with TTL/caching.

## Initial implementation scope
- Add a persistent grants model/table for agent/session/entity access.
- Add APIs to create/list/revoke grants.
- Extend the spawn API to accept initial grants for the spawned agent/session/entity.
- Validate spawn-time grants so the spawner cannot grant access beyond their authority.
- Apply default grants when sessions/entities are created:
  - Desktop/Horton local session: creator/spawner read/write only.
  - Shared/bot-created session: configurable workspace read/write grant, provided at spawn/create time.
- Update read/write permission checks to consult explicit grants.
- Implement user-delegated spawned-agent permission behavior.
- Add the minimal Cloud role/token behavior needed so normal Desktop users do not need service-secret access.

## Out of scope for this issue
- Full per-service Cloud roles.
- End-to-end encryption / local-only readable streams.
- External authorization webhooks.
- Webhook TTL/caching implementation.
- Fine-grained permission narrowing for spawned agents.

## Acceptance criteria
- A Desktop/Horton session created by a user is private by default.
- The creator/spawner has read/write access.
- Other workspace members cannot read/write the session unless granted.
- The creator can explicitly grant read or read/write access to another user.
- The creator can explicitly grant read or read/write access to the workspace.
- The spawn API accepts initial grants for the spawned entity/session.
- Spawn-time grants are validated against the spawner’s authority.
- Shared/bot-created sessions can be initialized with workspace-wide grants at spawn/create time.
- A user-spawned agent can access what the spawning user can access and cannot access what the user cannot access.
- Normal Desktop users can connect/use the cloud agents service without seeing or using the service secret.
- Tests cover private-by-default sessions, explicit user grants, workspace grants, denied access, spawn-time grants, invalid/escalating spawn-time grants, bot/shared defaults, and spawned-agent delegation.

icehaunter:
## Motivation

Agents Server needs authorization at every entity access path so tenants can limit which principals can spawn types and access individual entity instances. The default model should keep entity instances owner-private while still allowing common sharing flows through explicit grants, principal-kind grants, spawn-time inheritance, and Electric-visible effective permissions.

fixes #4464

## What changed

- Added permission storage for entity type spawn grants, entity instance grants, effective permissions, lineage, and shared-state links.
- Added owner-default permission checks plus middleware enforcement for entity routes, spawn, durable streams, attachments, schedules, event-source subscriptions, tags, inbox edits, and shared-state streams.
- Made new entity type registration available to any authenticated principal by default. Runtime entity definitions can publish their default spawn grants during `registerTypes()`.
- Added default built-in `spawn` grants for `principal_kind=user` on Horton and Worker.
- Materialized descendant and copy-at-spawn entity grants into `entity_effective_permissions` for point decisions and Electric visibility.
- Scoped entity list, entity type list, Electric shape predicates, and `entities(tags)` observation streams by the authenticated principal. Entity `manage` grants are included in read visibility, and entity-type `manage` grants are included in spawn/type visibility.
- Added spawn-time initial grants for the new entity. Direct root-spawn grants are allowed because the caller becomes the owner of the new entity; broad parented-spawn grants require `manage` on the parent before they can be delegated.
- Added an optional authorization hook for deployments that need custom point decisions; webhook decisions are not injected into Electric predicates and should be materialized into effective rows for shape visibility.
- Enabled Electric subqueries in `packages/agents-server/docker-compose.dev.yml` with `ELECTRIC_FEATURE_FLAGS=allow_subqueries`.

## Permissions model

Principals are still defined outside Agents Server. The server consumes the authenticated principal from the request context/header and stores grants against either a concrete principal URL, such as `/principal/user%3Aalice`, or a principal kind, such as `user`, `agent`, `service`, or `system`. Native group management is intentionally out of scope for this version; principal-kind grants cover the common broad-access cases.

Type grants control type-level actions only. A type-level `spawn` grant decides whether a principal may create an instance of a registered entity type; type-level `manage` is used for type mutation and is treated as a superset for type visibility/spawn checks. Type grants never grant access to existing entity instances. New type creation is open to authenticated principals so developers can register new entity definitions, while mutation of an existing type definition remains guarded by `manage`.

Entity grants control access to individual entity instances. Instance access defaults to `created_by` ownership, plus explicit/effective grants for verbs such as `read`, `write`, `delete`, `signal`, `fork`, `schedule`, and parent `spawn`. Entity `manage` is treated as a superset for point checks and visibility.

The permission verbs are intentionally granular. `write` covers content mutation paths such as send, tags, inbox edits, attachments, and event-source subscription edits. It does not imply `delete`, `signal`, `fork`, `schedule`, parent `spawn`, or `manage`; those require their own grants or `manage`. UI sharing presets like “read-only”, “read/write”, and “full control” should expand into the appropriate set of granular grants.

Spawn links the new entity into the permission graph:

- The caller must have a type-level `spawn` grant for the target type.
- If the spawn has a parent, the caller must also have `spawn` on that parent entity.
- The new entity records `created_by` as the caller principal.
- Parent/child lineage is recorded so descendant grants can be materialized for current and future descendants.
- Grants marked `copy_to_children` are copied into the child as direct grants during spawn.
- The spawn body may include initial direct grants for the new entity.
- For parented spawns, broad or propagating initial grants require `manage` on the parent. This covers `principal_kind` grants, `manage` grants, `propagation=descendants`, and `copy_to_children=true`.
- Manifest-discovered shared-state links connect shared-state streams back to the linked entity permissions.

## Key decisions

- Electric predicates use only `IN (SELECT ...)` subqueries with no correlated outer references.
- Entity instance access defaults to `created_by` ownership plus explicit/effective grants.
- Type grants are not used for instance visibility or instance access; entity grants are the source of truth there.
- Workspace-wide sharing is represented as `subject_kind=principal_kind` and `subject_value=user`. This assumes an Agents Server tenant/service maps to a workspace boundary, and only workspace members can authenticate as `user` principals in that tenant.
- Groups are not modeled in this version; deployments that need richer policy can use the authorization hook and/or materialize decisions into effective permission rows.
- Public or “anyone with the URL” sharing is a follow-up. The model leaves room for a later `public`/`link` subject kind, synthetic anonymous/link principals, or Cloud share-link records that resolve into an authorization context with matching Electric visibility behavior.
- Expired grants are pruned/deactivated server-side rather than using `now()` in Electric predicates.
- Pre-permission entity observation bridges are rebuilt after upgrade because old bridge rows do not include principal attribution.

## Coverage

Added focused coverage for permission service behavior, route middleware, durable stream access, shared-state authorization, Electric SQL predicate generation, principal-scoped observation streams, runtime registration grant payloads, spawn-time grant materialization and parented-spawn delegation checks, and built-in Horton/Worker spawn grants.

A local smoke run also registered a linked `/tmp` runtime, spawned an entity as `user:test`, confirmed `user:other` was denied list/send/view before grants, then confirmed `user:other` could list/send/view after explicit `read` and `write` grants using `ELECTRIC_AGENTS_SERVER_HEADERS` with `Electric-Principal`.


KyleAMathews:
I read through this against the main sharing flows we want to support in Desktop/Mobile:

1. Desktop/mobile apps spawn agents/sessions that are private to the creator by default, but overridable at spawn time.
2. A user can share a session with a specific teammate, or with everyone in the workspace, choosing granular permissions like `read`, `write`, `schedule`, `manage`, etc.
3. A user can eventually share a session globally to anyone with the URL. This likely needs a separate PR because it touches Cloud/link auth, but this PR should leave a clear path for it.

Overall this is a strong foundation: private-by-default via `created_by`, explicit entity grants, principal-kind grants, scoped lists/Electric predicates, and built-in `spawn` grants for Horton/Worker line up well with the first two flows. I think there are a few changes/clarifications we should make before merging.

## 1. Spawn-time grants need authority validation

The issue explicitly calls out validating spawn-time grants so the spawner cannot grant beyond their authority. Right now `withSpawnPermission` validates that the caller can spawn the type, and can spawn from the parent when one is supplied, but the actual initial grants are then inserted without further validation:

```ts
for (const grant of parsed.grants ?? []) {
  await ctx.entityManager.registry.createEntityPermissionGrant({
    entityUrl: entity.url,
    permission: grant.permission,
    subjectKind: grant.subject_kind,
    subjectValue: grant.subject_value,
    propagation: grant.propagation,
    copyToChildren: grant.copy_to_children,
    expiresAt: parseExpiresAt(grant.expires_at),
    createdBy: principal.url,
  })
}
```

So a caller with type-level `spawn` can currently create an entity and attach broad grants like:

```json
{
  "subject_kind": "principal_kind",
  "subject_value": "user",
  "permission": "manage",
  "propagation": "descendants",
  "copy_to_children": true
}
```

If the intended rule is “the spawner owns the newly created entity and therefore may grant any permission on it”, that should be documented and tested explicitly. But I think we likely want a validation step, especially for:

- `permission: "manage"`
- `subject_kind: "principal_kind"`
- `propagation: "descendants"`
- `copy_to_children: true`

At minimum, broad/propagating grants should require some explicit grant authority, e.g. `manage` on the parent/source context, or another clearly-defined “can delegate/share” rule. We should also add tests for invalid/escalating spawn-time grants, since that was part of the linked issue’s acceptance criteria.

## 2. Electric entity visibility should match server-side `manage` semantics

Server-side entity checks treat `manage` as a superset of the requested permission:

```ts
const permissions = [permission, `manage`] as const
```

But the Electric predicate for readable entities currently only checks:

```sql
permission = 'read'
```

That means a user with `manage` on an entity can access it through HTTP routes, but may not see it in Electric-backed entity lists unless they also have a separate `read` grant.

Suggested fix:

```sql
permission IN ('read', 'manage')
```

and add a test covering a manage-only grant being visible through both HTTP and Electric predicates.

## 3. Electric entity-type visibility should match server-side `manage` semantics

Same issue for entity types. Server-side type permission checks treat `manage` as a superset of `spawn`, but the Electric predicate only checks:

```sql
permission = 'spawn'
```

Suggested fix:

```sql
permission IN ('spawn', 'manage')
```

with a test for manage-only type grants.

## 4. Document the permission lattice / sharing semantics

The issue talks in terms of `read` and `read/write`, while this PR introduces a richer permission set:

- `read`
- `write`
- `delete`
- `signal`
- `fork`
- `schedule`
- `spawn`
- `manage`

That seems useful, but we should document what `write` does and does not imply. For example, today routes like signal/delete/fork/schedule/parent-spawn require separate permissions rather than being implied by `write`.

This matters for Desktop/Mobile UX: if the UI says “share read/write”, users may expect broader control than the server grants. Either the UI should expose the granular permissions, or we should provide a helper/API that expands common sharing presets like “read-only”, “read/write”, and “full control” into the right grants.

## 5. Workspace-wide sharing is currently `principal_kind=user`; that’s OK only if tenant == workspace

For “share with everyone in the workspace”, this PR can approximate that with:

```json
{
  "subject_kind": "principal_kind",
  "subject_value": "user",
  "permission": "read"
}
```

But this is really “all user principals in this tenant/service”, not a first-class workspace/team/group grant. That’s probably fine if Cloud maps one agents-server tenant to one workspace and only workspace members can authenticate as `user` principals in that tenant. We should document that assumption. If tenant != workspace, this is too broad.

## 6. Global URL sharing can be a separate PR, but this PR should not block it

True “anyone with the URL” sharing is not implemented here, which is fine because it likely needs Cloud/link-auth work. Current requests require a principal, and grants only target `principal` or `principal_kind`; there is no `public`, `link`, or `anonymous` subject/capability model yet.

I don’t think this PR needs to implement global URL sharing, but it should leave the model extensible for it. For example, a follow-up may need one of:

- a new grant subject kind like `public` or `link`
- a synthetic principal kind like `link`/`anonymous`
- capability/share-link records that Cloud resolves into a principal or authorization context
- matching Electric predicate behavior for link/public visibility

It would be good to avoid baking in assumptions that make that hard later, and maybe add a TODO/design note around this.

## Demo/test suggestion

Once the above is tightened up, let’s add Desktop/Mobile support for these flows so we can demo and test them end-to-end:

1. Spawn a Horton/Worker session from Desktop/Mobile and confirm it is private to the creator by default.
2. Share that session with a specific teammate using a `principal` grant and verify the teammate can see/use it according to the selected permissions.
3. Share that session with the workspace using a `principal_kind=user` grant and verify another workspace member can see/use it.
4. Verify an ungranted user cannot list/read/write the session.
5. Later, in a Cloud/link-sharing PR, add “anyone with URL” sharing and verify public/link access works through both HTTP and Electric-backed views.

samwillis (approved):

KyleAMathews:
Small clarification on global URL sharing: I’m not asking this PR to implement it. That should be a separate Cloud-coordinated PR because it likely needs share-link/capability auth at the Cloud edge.

The ask for this PR is just to make sure the model can support it later. Today grant subjects are `principal` / `principal_kind` and every request requires a principal. That’s fine for this PR, but we should avoid baking in assumptions that make a future `public`, `link`, `anonymous`, or Cloud-resolved capability subject hard to add.

Concretely, it would be good if this PR either:

- leaves the grant/predicate/auth boundaries clearly extensible for a future non-user capability subject, or
- adds a TODO/design note showing where global URL sharing would plug in later.

No need to build the Cloud/link-sharing flow in this PR.

PR #4477, opened 2026-06-02
feat(agents-mobile): unified server picker + auth refinement in kebab menu

msfstef:
## Summary

Brings the mobile three-dots (kebab) menu closer to the desktop unified server experience.

- **Persisted multi-server list** — new `savedServers` store (AsyncStorage) mirroring desktop's `ServerConfig[]`. Mobile keeps the single *active* server URL in `MobileAppState`; this store remembers every server so the user can switch without re-entering URLs. Cloud servers are tagged `electric-cloud` so they can be purged on sign-out (no token/tenantId stored — derived from the URL at request time).
- **Unified picker** — new `useAvailableServers` merges the saved list with the live Cloud agent servers (`useCloudAgentServers`), deduped by service id, into one list.
- **In-sheet server submenu** — the kebab's "Server" row now slides into a submenu listing all servers (self-hosted + Cloud, with kind icons, breadcrumbs, active check, connect-in-place for live Cloud servers, and an "Add custom server…" row that pushes to `ServerSetupScreen`). The drill animation was extracted from `SessionMenu` into a reusable `useDrillTransition` hook.
- **Account row refinement** — reflects auth status: name/email when signed in, "Sign in to Electric Cloud" when out.
- **Cloud sign-out cleanup** — purges persisted Cloud servers; if the active server was a Cloud server, falls back to a remaining self-hosted server, else clears (routing to server setup).
- **Migration** — users upgrading from the single-URL model get their active self-hosted server surfaced in the saved list.
- `ServerSetupScreen` / `CloudServerPicker` now persist into the saved list on connect.

## Design notes

- Server submenu uses the horizontal **drill + Back row** pattern, kept consistent with the existing signals submenu (swipe-down dismisses the whole sheet). Considered a dedicated dismissible sheet / swipe-to-go-back; deferred in favor of consistency.

## Testing

- 19 unit tests (vitest): `savedServers` store (upsert/dedup/cloud-purge/round-trip/merge-on-hydrate), `mergeAvailableServers` (dedup + active flag), and the `resolveActiveAfterCloudSignOut` fallback resolver.
- `tsc --noEmit` and eslint clean.
- Manual emulator pass (drill, connect a Cloud server, sign-out behavior) still TODO.

## Out of scope (follow-ups)

- **Pre-existing `inbox` live-query crash** on opening a conversation — caused by three duplicate `@tanstack/db` versions (0.6.5 / 0.6.6 / 0.6.7) failing a cross-copy `instanceof Collection` check, not by this PR. Fix likely a root `pnpm.overrides` pin. Deferred to a separate PR.
- **Generally available "Clear all local data"** in Diagnostics (currently `__DEV__`-only because the restart relies on dev-only `DevSettings.reload()`). Deferred.

### DIFF
--- a/.changeset/mobile-server-picker-refinements.md
+++ b/.changeset/mobile-server-picker-refinements.md
@@ -1,0 +1,5 @@
+---
+"@electric-ax/agents-mobile": patch
+---
+
+Harden the mobile kebab-menu server picker: surface connect failures in-sheet instead of silently swallowing them, persist a Cloud server only after the switch succeeds, and let the submenu close animation finish before resetting to the root page.
--- a/packages/agents-mobile/src/components/BottomSheet.tsx
+++ b/packages/agents-mobile/src/components/BottomSheet.tsx
@@ -192,36 +192,53 @@ export function BottomSheetSection({
 
 export function BottomSheetItem({
   label,
+  subtitle,
   icon,
   trailing,
   active,
   onPress,
   destructive,
+  disabled,
 }: {
   label: string
+  subtitle?: string
   icon?: ReactNode
   trailing?: ReactNode
   active?: boolean
   onPress: () => void
   destructive?: boolean
+  disabled?: boolean
 }): React.ReactElement {
   const tokens = useTokens()
   const styles = useMemo(() => itemStyles(tokens), [tokens])
   return (
-    <Pressable onPress={onPress}>
+    <Pressable onPress={onPress} disabled={disabled}>
       {({ pressed }) => (
-        <View style={[styles.row, pressed ? styles.pressed : null]}>
+        <View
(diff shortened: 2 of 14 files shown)


msfstef:
Screenshots of the new server picker and enhanced Account option - on signout it just redirects to the server setup/picker screen (which includes sign in buttons for cloud)

I've kept the clear all local data for __DEV__ only for now because reloading is a bit of a pain apparently

Simulator Screenshot - iPhone 14 Pro - 2026-06-02 at 17 48 06
Simulator Screenshot - iPhone 14 Pro - 2026-06-02 at 17 47 59

kevin-dp (approved):
LGTM!

PR #4478, opened 2026-06-02
Add Durable Streams fetch caching to agents runner

KyleAMathews:
## Summary
- install an Undici cache interceptor for the built-in agents local Node runner
- use a 100 MiB in-memory HTTP cache for Durable Streams fetches
- add undici as an agents package dependency

## Testing
- pnpm --filter @electric-ax/agents stylecheck


packages/agents-desktop/src/runtime/lifecycle.ts
```diff
@@ -290,6 +291,13 @@ export async function startRuntime(
   const nextRuntime = new BuiltinAgentsServer({
     agentServerUrl: activeServer.url,
     workingDirectory: deps.settings.workingDirectory ?? app.getPath(`home`),
+    durableStreamsFetchCache: {
```
samwillis, line 294: You have added a `maxCount` that can be used with the sqlite storage, but it's not configured. Not sure if it's unbounded or has an internal default?

samwillis (approved):
Other than one comment looks good.

packages/agents-desktop/src/runtime/lifecycle.ts
KyleAMathews: Good catch — I checked Undici’s `SqliteCacheStore` implementation for `7.24.7`. `maxCount` defaults to `Infinity`, and SQLite cache store doesn’t have a `maxSize` byte cap. If omitted, it only prunes expired entries and won’t evict old entries by count.

I’ve pushed a follow-up setting `maxCount: 10_000` for the desktop cache so the on-disk DB can’t grow without bound.

PR #4479, opened 2026-06-02
feat(agents): slash commands with structured input in the desktop app

samwillis:
## Summary
- Add the `composer_input` message contract, runtime validation, helper APIs, and slash-command state collections.
- Add static and dynamic slash command registration, including owned dynamic layers and proactive skill-backed command loading.
- Replace the agents UI composers with a shared ProseMirror composer that supports slash command autocomplete, structured argument slots, and consistent footer layout.

## Test plan
- Added and updated runtime/server/UI unit coverage for composer payload validation, slash command layering, skill context loading, timeline projection, and serializer behavior.
- Not run in this final PR-opening pass; only the changeset was added after the latest implementation commit.


KyleAMathews (approved):
:shipit:

PR #4480, opened 2026-06-02
Handle signed-out Cloud runtime connections

KyleAMathews:
## Summary
- show a clear sign-in-required error before starting local runtime for Electric Cloud servers
- require an active Cloud session before reusing cached agents tokens
- avoid injecting cached Cloud agents auth headers after sign-out

## Testing
- pnpm --filter @electric-ax/agents-desktop typecheck *(fails in this worktree because node_modules/workspace dependencies are not installed/linked; missing modules include electron, undici, @electric-sql/client, and workspace packages)*


samwillis:
Note to reviewer: we need to also clear the servers on the desktop app that come from cloud when the user explicitly signed out. This fix is more for if the auth fails / expires I think.

samwillis (commented):
Interactive review with GPT.

## Main issue: sign-out should delete Cloud servers

This fixes the signed-out runtime startup path, but I think explicit Cloud sign-out still leaves the main stale-state problem in place: saved `electric-cloud` servers remain in desktop settings after sign-out.

Current sign-out only clears the Cloud auth session:

```ts
ipcMain.handle(`desktop:cloud-auth-sign-out`, async () => {
  await deps.getCloudAuth().signOut()
})
```

The `CloudAuth` subscriber then stops the Cloud agent-server shape streams, but it does not remove the Cloud-derived server configs from `settings.servers`, disconnect their runtime entries, clear selections/default server, or delete their per-tenant cached agents tokens.

That means a signed-out user can still have Cloud servers in the app. This PR now prevents connecting them while signed out, which is good, but after signing into a different Cloud account the app can still reuse old Cloud server entries and cached `cloud-agents-token:<tenantId>` values from the previous account.

## Why this matters

`prepareConnection()` now correctly requires an active Cloud session before using a cached agents token:

```ts
const token = await this.cloudAuth.getToken()
if (!token) {
  throw new Error(`Not signed in to Electric Cloud`)
}

const cached = this.getAgentsToken(tenantId)
if (cached) return { url, tenantId }
```

But the cached token is still keyed only by `tenantId`, not by Cloud user/account. So once any user is signed in, a stale saved Cloud server can reuse an old agents token for that tenant without checking that the current Cloud user should still see that server.

## Suggested fix

On explicit Cloud sign-out, delete all servers with `source === "electric-cloud"` and clean up their related state:

- disconnect/stop any running runtime entries for those servers
- delete the saved Cloud server configs from `settings.servers`
- clear matching window selections and `defaultServerId`
- delete each server’s cached agents token via `forgetAgentsToken(tenantId)`
- save settings and refresh desktop state

The existing `forgetServer()` path already does most of this for one server, so the sign-out IPC could either reuse that behavior for all Cloud servers or add a dedicated “forget all Cloud servers” helper.

## Existing good parts in this PR

The new checks do improve the failure mode:

- auth header injection no longer sends cached Cloud agents auth headers while signed out
- `prepareConnection()` no longer reuses cached agents tokens unless there is an active Cloud session
- `startRuntime()` now surfaces a clear “Sign in to Electric Cloud” error before health checks / runtime startup

I’d still block on deleting Cloud servers during explicit sign-out, because otherwise the app keeps Cloud account-specific state around after the user has asked to sign out.

samwillis (approved):
Interactive review with GPT.

Approving now. The original blocker around deleting Cloud-derived servers/tokens on Cloud sign-out looks addressed in the updated branch.

I did notice some redundancy between the controller-level signed-out cleanup and the new IPC sign-out cleanup loop, but I don't think it is harmful or approval-blocking.

PR #4482, opened 2026-06-02
Scale Horton context budget by model

KyleAMathews:
## Summary
- derive Horton's context source budget from selected model metadata instead of a fixed 100k
- add model catalog helpers for context window/source budget lookup
- cover OpenAI, Anthropic, Moonshot, and unknown-model fallback cases

## Test plan
- pnpm --filter @electric-ax/agents test test/model-catalog.test.ts

### DIFF
--- a/.changeset/horton-context-budget.md
+++ b/.changeset/horton-context-budget.md
@@ -1,0 +1,5 @@
+---
+'@electric-ax/agents': patch
+---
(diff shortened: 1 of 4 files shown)


samwillis (approved):

samwillis:
Needs a changeset

PR #4487, opened 2026-06-03
feat(agents-server-ui): show elapsed time while agent is thinking

kevin-dp:
## Summary

While a response is streaming, the agent meta row now ticks `Thinking · 12s` (or just `12s` once tokens start flowing). When the response settles in-session the existing `✓ done` becomes `✓ done in 1m 5s`.

- `ElapsedTime` component (self-ticking once a second; tears down its interval on disable; uses `tabular-nums` so the digits don't jitter).
- `formatElapsedDuration(ms)` in `lib/formatTime.ts` for compact duration labels (`0s` → `1m 5s` → `2h 47m`).
- Capture pattern (`sawStreamingRef` + `finalDurationMs`) wires `✓ done in Xs` into both `AgentResponseLive` and the cached `AgentResponse`. Responses that were already complete on page load keep the bare `✓ done` — the client has no reliable completion timestamp for those (only the user message time), so subtracting `now()` would lie about the duration.

## Test plan

- [x] `pnpm typecheck` in `agents-server-ui` (no new errors in the changed files; pre-existing errors in `EntityContextDrawer.tsx` and `agents-runtime/*` are unchanged from `main`)
- [x] Web — manually verified in `http://localhost:5175/__agent_ui/` against a local agents-server; thinking ticker counts up, `✓ done in Xs` lands on completion
- [x] Electron — manually verified in the desktop dev build; same behavior
- [ ] Mobile (Expo DOM embed via `ChatLogView` → `EntityTimeline` → `AgentResponse`) — not verified in this PR. The mobile dev client hit an unrelated pre-existing crash in `agents-runtime/src/entity-timeline.ts:1225` (`Invalid source for live query: alias 'inbox' is not a Collection`) when connected to the Electric Cloud workspace, blocking validation. The change here is a pure additive UI render in shared components, so mobile should pick it up automatically once that crash is sorted; following up separately.

### DIFF
--- a/.changeset/elapsed-time-thinking-indicator.md
+++ b/.changeset/elapsed-time-thinking-indicator.md
@@ -1,0 +1,12 @@
+---
+'@electric-ax/agents-server-ui': patch
+'@electric-ax/agents-desktop': patch
+---
+
+Show elapsed time while an agent is responding. While a turn is
+streaming, the meta row now ticks `Thinking · 12s` (or just `12s` once
+tokens start flowing). When a turn settles, the bare `✓ done` becomes
+`✓ done in 1m 5s` for turns completed in-session. Historical turns
+(already complete on page load) keep the bare label, since the client
+has no reliable completion timestamp for those — only the user message
+time, and subtracting `now()` would lie about the duration.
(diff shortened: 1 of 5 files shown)


samwillis (approved):
LGTM

PR #4488, opened 2026-06-03
feat(agents-mobile): add Sentry crash/error reporting

msfstef:
## What

Adds [Sentry](https://sentry.io) crash/error reporting to `agents-mobile` — the app had no error visibility in the field.

## How

- **`src/lib/sentry.ts`** (new) — `initSentry()` calling `Sentry.init`, errors-only (`tracesSampleRate: 0`), disabled in dev (`enabled: !__DEV__`).
- **`app/_layout.tsx`** — `initSentry()` at module scope + `export default Sentry.wrap(RootLayout)` (Sentry's documented Expo Router pattern; earliest init point).
- **DSN injection** — hardcoded production default in `sentry.ts` with an `EXPO_PUBLIC_SENTRY_DSN` override, mirroring the existing `cloudAuth.ts` (`PROD_DASHBOARD_URL`) convention. The DSN is a public client identifier (shipped in every binary regardless), so it is intentionally committed — not a secret.
- **Source maps** — `withSentry` config plugin in `app.config.ts` + `getSentryExpoConfig` in `metro.config.js` (injects Debug IDs). Custom monorepo metro resolver/aliases preserved.
- **`eas.json`** — each build profile now declares an `environment` so the upload secret resolves at build time.

## Secrets

The **only** secret is `SENTRY_AUTH_TOKEN` (source-map upload). It is stored as an **EAS environment variable (Secret visibility)**, not in the repo and not a GitHub secret — EAS builds run on EAS servers, where the upload happens.

## Verification

- `pnpm --filter @electric-ax/agents-mobile run ci:check` passes (typecheck → expo-doctor → Android export).
- Exported Hermes bundle confirmed to contain the Sentry SDK runtime.
- Symbolicated stack traces will appear on the next EAS `preview`/`canary`/`production` build (when the token is present); builds still succeed without it, just with minified traces.


msfstef:
Thanks for the review — addressed in 626fbd5:

- **EU region mismatch (Important)** — fixed. The DSN is EU-region (`ingest.de.sentry.io`), so the `withSentry` upload `url` now points at `https://de.sentry.io/` to match. This makes source-map upload work regardless of whether the auth token is region-encoded.
- **Trailing newline in `.env.example` (Suggestion)** — added.
- **Navigation integration (Suggestion)** — intentionally absent; tracing is disabled (`tracesSampleRate: 0`). No action.

**Re: "No changeset needed"** — actually one *is* needed here. The repo's `.changeset/config.json` sets `"privatePackages": { "version": true }`, so private packages like `@electric-ax/agents-mobile` are versioned via changesets (that's why it's at `0.0.10`). Added a `patch` changeset.

One leftover: `.env.example` still reads "sensitive visibility" mid-file where it should say "secret" — my tooling can't edit that path, so it'll need a manual one-word tweak.

samwillis (approved):
Awesome! :shipit:

PR #4490, opened 2026-06-03
fix: dedupe @tanstack/db + set agents-mobile iOS minimum to 16.4

msfstef:
## What

Adds a root `pnpm.overrides` entry collapsing the `@tanstack/db` `0.6.x` line to a single version (`0.6.7`).

## Why

`@tanstack/db` is effectively a singleton — collections, transactions, and live queries rely on `instanceof` checks and module-level shared state, and `Collection` carries private fields, so two copies are nominally distinct types. The lockfile had drifted to multiple `0.6.x` copies at once (`0.6.5`, `0.6.6`, `0.6.7`) because `@tanstack/react-db` pins an exact `@tanstack/db` per release while `@durable-streams/state` declares an open `^0.6.0`. A single app therefore resolved two `0.6.x` copies, breaking StreamDB collections at the type and runtime level.

## Scope

The override is deliberately scoped to `>=0.6.0 <0.7.0` rather than a blanket pin, so the legacy example starters (`burn-client`, `tanstack-start-db-electric-starter`, `expo-db-electric-starter`) that still depend on `@tanstack/db` `0.0.x`/`0.5.8` via old `react-db`/`electric-db-collection` are left untouched — forcing them onto `0.6.x` would break them with API mismatches.

## Verification

- Lockfile: the `0.6.x` line now resolves to a single `0.6.7`; `0.0.27`/`0.0.33`/`0.5.8` entries from the legacy starters are unchanged.
- The full agents-mobile bundle closure (agents-mobile + agents-runtime + agents-server-ui + `@durable-streams/state`) resolves to one physical `@tanstack/db@0.6.7`.
- Affected packages build/typecheck cleanly (no new `@tanstack/db`-related errors).

## Stopgap

This is a deliberate stopgap. The durable fix is upstream — `@durable-streams/state` is moving `@tanstack/db` from a regular dependency to a peer dependency (durable-streams PR #381). Once that ships and we bump `@durable-streams/state`, the state-induced duplication resolves on its own and this override can likely be removed (the legacy-starter versions are a separate cleanup).

---

## Also in this PR: agents-mobile iOS minimum deployment target → 16.4

Separate commit (`fix(agents-mobile): set iOS minimum deployment target to 16.4`).

**What:** adds the `expo-build-properties` plugin with `ios.deploymentTarget: "16.4"`.

**Why:** the mobile chat renders inside an Expo DOM WebView whose markdown stack (`streamdown` → `remend` / `mdast-util-gfm-autolink-literal`) ships regex lookbehind `(?<=…)`. JavaScriptCore only parses lookbehind on **iOS 16.4+**; below that the entire DOM bundle fails to *parse*, so the chat renders blank with no surfaced error (reproduced on an iOS 16.2 simulator). Pinning the deployment target to 16.4 stops the app installing on engines that can't run it. Confirmed rendering correctly on iOS 17.4. These deps have shipped since the agents feature first landed, so this is a longstanding minimum-iOS reality, not a regression.

**Lockfile note:** adding `expo-build-properties` pulls in `semver ^7.7`, so pnpm deduped the existing transitive `@databases/pg-typed` `semver` from `7.6.3` → `7.7.4` to keep a single `semver@7` instance. Intentional dedup, not a wider install.

### DIFF
--- a/.changeset/dedupe-tanstack-db.md
+++ b/.changeset/dedupe-tanstack-db.md
@@ -14,3 +14,8 @@ drifted to several `0.6.x` copies, breaking StreamDB collections. Adds a root
 `>=0.6.0 <0.7.0` so the legacy example starters pinned to `0.0.x`/`0.5.8` are
 untouched. Stopgap until `@durable-streams/state` ships `@tanstack/db` as a
 peer dependency.
+
+Also raises the `agents-mobile` iOS minimum deployment target to 16.4 (via
+`expo-build-properties`). The chat renders in an Expo DOM WebView whose markdown
+stack ships regex lookbehind, which JavaScriptCore only parses on iOS 16.4+;
+below that the whole DOM bundle fails to parse and the chat renders blank.
--- a/.changeset/dedupe-tanstack-db.md
+++ b/.changeset/dedupe-tanstack-db.md
@@ -1,0 +1,16 @@
+---
+'electric-ax': patch
+'@electric-ax/agents-server-ui': patch
+'@electric-ax/agents-runtime': patch
+'@electric-ax/agents-mobile': patch
+---
+
+Dedupe `@tanstack/db` to a single instance.
+
+`@tanstack/db` is effectively a singleton (collections/transactions/live
+queries use `instanceof` checks and module-level state), but the lockfile had
+drifted to several `0.6.x` copies, breaking StreamDB collections. Adds a root
+`pnpm.overrides` entry collapsing the `0.6.x` line to `0.6.7`, scoped to
+`>=0.6.0 <0.7.0` so the legacy example starters pinned to `0.0.x`/`0.5.8` are
+untouched. Stopgap until `@durable-streams/state` ships `@tanstack/db` as a
+peer dependency.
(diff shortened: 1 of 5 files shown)


samwillis (approved):

msfstef:
Thanks — addressed the review:

- **PR scope vs description (Important):** updated the title and added an "Also in this PR" section documenting the iOS 16.4 deployment-target change. Kept it as one PR (already approved) rather than splitting.
- **Changeset body (suggestion):** `.changeset/dedupe-tanstack-db.md` now also documents the iOS 16.4 minimum, so the `agents-mobile` bump is self-explanatory.
- **Lockfile `semver` drift (suggestion):** confirmed intentional — `expo-build-properties` depends on `semver ^7.7`, so pnpm deduped the existing transitive `@databases/pg-typed` copy from `7.6.3` → `7.7.4` to keep a single `semver@7` instance. Reverting it would re-duplicate `semver`, so it's left as-is.
- **`//pnpm.overrides` doc-key coupling (suggestion):** noted; leaving the convention as-is for now.

PR #4492, opened 2026-06-03
feat(agents-server): mirror user principals into users

icehaunter:
## Motivation

User principals are already materialized as built-in principal entities, but the tenant-scoped `users` table was not populated from that flow. That left hosts without a stable user row to join against or enrich.

## Changes

- Add tenant-scoped `users` persistence for user-principal materialization.
- Mirror `user:*` principals into `users` whenever `ensurePrincipal` materializes or observes the built-in principal entity.
- Keep the mirror path insert-only so host/cloud enrichment can never be clobbered by principal creation.
- Add focused tests for manager principal mirroring and registry user persistence behavior.
- Add a package changeset.

## Decisions

OSS stays intentionally identity-provider agnostic: principal creation only guarantees a tenant-scoped user id exists. Optional fields such as display name, email, avatar, auth provider, profile, and metadata remain open for host-specific integrations to populate. There is no OSS enrichment/upsert helper in this PR, so cloud-specific metadata merge semantics live in the cloud host instead of creating a divergent dead-code helper here.

### DIFF
--- a/.changeset/principal-users-table.md
+++ b/.changeset/principal-users-table.md
@@ -1,0 +1,5 @@
+---
+"@electric-ax/agents-server": patch
+---
+
+Mirror user principals into the tenant-scoped `users` table when principal entities are materialized, while preserving any profile fields enriched by host-specific identity sync.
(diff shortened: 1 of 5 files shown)


msfstef (approved):
Nice - so any kind of integration would go off of the users table tenant id + user id to match to metadata to expose independent of the agent server code

PR #4450, opened 2026-05-30
ci: harden privileged PR workflows

KyleAMathews:
## Summary

Immediate CI supply-chain hardening following the TanStack incident audit:

- remove implicit pnpm cache from the release workflow
- restrict privileged load-test runs to same-repo PR branches
- restrict benchmark comment trigger to MEMBER/OWNER and reject fork PRs before privileged steps
- remove CROSSREPO_PAT callback payloads from benchmark requests
- downgrade TS test workflow package permission from write to read
- add CODEOWNERS protection for .github CI/CD files

## Validation

- Parsed changed workflow YAML files with Ruby YAML loader:
  - .github/workflows/changesets_release.yml
  - .github/workflows/load_test.yml
  - .github/workflows/benchmarking.yml
  - .github/workflows/ts_tests.yml

Note: this PR is from the clean branch horton/ci-immediate-supply-chain-fixes-clean and contains only the immediate CI hardening commit.

PR #4495, opened 2026-06-03
ci: fix TS tests startup failure + desktop canary electron-builder config

msfstef:
## What

Started as a fix for three CI failures observed on `main`; reviving the TS suite (the first fix) then surfaced pre-existing typecheck breakage that had been hidden while CI was dark, which this PR also cleans up.

### CI / build fixes

| Job | Verdict | Fix |
|-----|---------|-----|
| **TS tests** | 🔴 Real — startup-failing on *every* run since #4450 | Job-scoped `packages: write` on the `ensure_sync_service_image` caller |
| **Agents Desktop Canary** | 🔴 Real — fails every run since #4441 | `-c.channel` → `-c.publish.channel` |
| **Changesets** | 🟡 Flake — intermittent `agents-runtime` dts race | Promote `skills/types` to a tsdown entry |

### Typecheck fixes (pre-existing breakage exposed by re-enabling TS CI)

| File | Issue | Fix |
|------|-------|-----|
| `agents-runtime/src/pi-adapter.ts` | merged assistant `content` typed `unknown[]` (#4449) | annotate the block-array type |
| `agents-runtime/src/webhook-signature.ts` | `node:crypto` no longer exports `JsonWebKey` | cast input to `JsonWebKeyInput` |
| `agents-runtime/src/sandbox/docker.ts` | `isDockerAvailable` not on the subpath the tsconfig wildcard resolves | re-export it (cascade fix for `electric-ax` / conformance) |
| `agents-server-ui/tsconfig.json` | UI typechecked agents-runtime's node-only sandbox source via `paths` | resolve the index via built `.d.ts`; keep only the browser-safe `client` source-mapped — UI stays node-free |

## Why (CI fixes)

- **TS tests:** #4450 downgraded the workflow's top-level token to `packages: read`, but `ts_tests.yml` is the sole caller of the reusable `ensure_sync_service_image.yml`, whose job requests `packages: write` to push the sync-service image to GHCR. A called reusable workflow cannot elevate permissions above the caller's token, so GitHub failed the **entire run at startup** — meaning the TS test suite had not run on any commit (main or PR) since. Fixed by granting `packages: write` only on the caller job, keeping the top-level token at `read` per #4450's hardening.
- **Agents Desktop Canary:** electron-builder 26.8.1 rejects the config with `unknown property 'channel'`. `channel` is not a valid root property — moved under the publish provider (`-c.publish.channel=beta`, alongside the existing `-c.publish.url`).
- **Changesets:** `agents-runtime` dts build intermittently fails with `UNLOADABLE_DEPENDENCY: Could not load src/skills/types.d.ts` under CI's parallel build. Promoting `src/skills/types.ts` to a first-class tsdown entry makes its `.d.ts` a stable named output instead of a raced chunk.

## Validation

- Workflow files parse as valid YAML.
- All previously-failing typecheck packages (`agents-runtime`, `electric-ax`, `agents-server-ui`, `agents-server-conformance-tests`) verified green via CI-faithful isolated (`--frozen-lockfile`) install + build + typecheck.

## Not in scope (pre-existing, flagged separately)

- `runtime-dsl.test.ts` (92 tests, `401 UNAUTHORIZED: Principal is not allowed to spawn`) — from #4475's permission enforcement (@icehaunter); test fixtures need spawn permission seeded. Not a build/type issue.
- An `agents-mcp` dts-race flake in the conformance build (same class as the `skills/types` one).

## Note for reviewer

The canary maps channel input `canary` → publish channel `beta`. Preserved the existing value, but flagging in case it should be `canary`.

### DIFF
--- a/.changeset/agents-runtime-dts-entry.md
+++ b/.changeset/agents-runtime-dts-entry.md
@@ -1,0 +1,7 @@
+---
+'@electric-ax/agents-runtime': patch
+---
+
+Promote `skills/types` to a first-class tsdown entry so its `.d.ts` is a stable
+named output, avoiding an intermittent dts generation failure under CI's
+parallel build.
--- a/.changeset/agents-server-ui-node-free-typecheck.md
+++ b/.changeset/agents-server-ui-node-free-typecheck.md
@@ -1,0 +1,7 @@
+---
+'@electric-ax/agents-server-ui': patch
+---
+
+Typecheck against agents-runtime's built types for the package index instead of
+its source, so the UI no longer pulls node-only sandbox code into its program.
+The browser-safe `client` entry stays source-mapped (matching the vite alias).
--- a/.github/workflows/agents_desktop_build.yml
+++ b/.github/workflows/agents_desktop_build.yml
@@ -255,7 +255,7 @@ jobs:
             builder_args+=("-c.mac.identity=-")
           fi
           if [[ "${{ inputs.channel }}" == "canary" ]]; then
-            builder_args+=("-c.channel=beta")
+            builder_args+=("-c.publish.channel=beta")
             builder_args+=("-c.publish.url=https://github.com/electric-sql/electric/releases/download/agents-desktop-canary")
           fi
 
@@ -272,7 +272,7 @@ jobs:
 
           builder_args=(${{ matrix.builder_args }})
           if [[ "${{ inputs.channel }}" == "canary" ]]; then
-            builder_args+=("-c.channel=beta")
+            builder_args+=("-c.publish.channel=beta")
             builder_args+=("-c.publish.url=https://github.com/electric-sql/electric/releases/download/agents-desktop-canary")
           fi
 
--- a/.github/workflows/ts_tests.yml
+++ b/.github/workflows/ts_tests.yml
@@ -76,6 +76,11 @@ jobs:
   ensure_sync_service_image:
     needs: [detect_affected_workspaces]
     if: ${{ needs.detect_affected_workspaces.outputs.has_shared_sync_service_packages == 'true' }}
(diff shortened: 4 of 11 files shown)


samwillis (approved):

msfstef:
⚠️ **Heads up: the `TS tests` workflow has not actually run on any commit since ~June 3.**

It was failing at **startup** (`startup_failure`, ~1s) on every push and PR — main included — because #4450 capped the workflow token at `packages: read` while the reusable `ensure_sync_service_image.yml` it calls requests `packages: write` (a called workflow can't elevate above the caller). So for the past day+, *no TypeScript typechecks or unit tests were running in CI at all*. The permission fix in this PR revives the suite.

Because the suite was dark, it's now surfacing pre-existing breakage that landed on `main` unchecked:

- ✅ **Fixed in this PR** — typecheck errors in `agents-runtime` (`pi-adapter.ts` from #4449, `webhook-signature.ts` `JsonWebKey`, and a `sandbox/docker` subpath resolution issue cascading to `electric-ax` / `agents-server-ui` / conformance-tests).
- ❌ **Still failing, needs owner** — `Test packages/agents-runtime` → `runtime-dsl.test.ts` (92 tests) all fail with `401 UNAUTHORIZED: Principal is not allowed to spawn`. This comes from the permission enforcement added in #4475 (@icehaunter) — the runtime-dsl test fixtures don't seed spawn permission for the test principal. This is domain logic on the new (schema-backed) permission model, so leaving it for the author rather than fixing it blind here.

tl;dr: merging this turns TS CI back on; expect `runtime-dsl` to stay red on `main` until the #4475 permission fixtures are updated.

KyleAMathews (approved):
🙏

PR #4502, opened 2026-06-04
feat(agents-server-ui): show per-response token usage in meta row

kevin-dp:
## Summary

Adds a token-usage label to the agent response meta row, e.g.
`Thinking · 12s · 1.2k ↑ 412 ↓` while streaming and
`✓ done in 12s · 1.2k ↑ 412 ↓ · 14:18` once settled. Counter updates at
step boundaries — for a single-turn LLM call it lands once at done;
for tool-using runs it jumps as each step completes (the LLM SDK only
emits `usage` at end-of-step, so we can't tick smoothly between
streamed tokens — the elapsed-time ticker still ticks every second
alongside it).

## Plumbing

The runtime already had the token data — `pi-adapter.ts:358-359`
extracts `tokenInput`/`tokenOutput` from the provider's per-step
`usage` payload — but the bridge silently dropped them before
persistence. This PR closes that gap and surfaces them all the way to
the UI:

- `StepValue` gains optional `input_tokens` / `output_tokens` columns
  (Zod + TS). Strictly additive: events recorded before this change
  still validate (both fields optional), so no migration is needed.
- `outbound-bridge.ts:onStepEnd` now persists the values it was
  already receiving from `pi-adapter.ts`.
- `IncludesStep` / `EntityTimelineStepItem` surface the new fields,
  and the three `.select()` blocks that materialize step rows include
  them.
- The cached `agent_response` section grows a
  `tokens?: { input?, output? }` summed across the run's steps at
  section-build time, and the `fingerprintRun` cache key includes
  per-step token deltas so a late-arriving `onStepEnd` invalidates a
  stale cached section.
- New `<TokenUsage>` component in `agents-server-ui` with
  `tabular-nums` so digits don't jitter, locale-aware compact
  formatting via `Intl.NumberFormat`. Renders next to `<ElapsedTime>`
  in both the live and cached meta rows.

## Test plan

- [x] `pnpm typecheck` clean in `agents-runtime` + `agents-server-ui`
- [x] `pnpm test outbound-bridge use-chat entity-timeline` in
      `agents-runtime` (74 passed)
- [x] Full `agents-runtime` test suite: my branch matches the same
      pre-existing 401 failures observed on clean `main` (unrelated
      permission-system breakage in the test harness, not introduced
      by this PR)
- [x] Manual: launch a turn that uses tools and watch the counter
      jump at each step boundary
- [ ] Manual: pure-text turn — counter lands once at done

## Notes

- Historical responses recorded before this change have no token data
  persisted (older `steps` rows lack the columns). The `tokens` field
  is conditional on at least one step reporting a number, so those
  sections continue to render with no token row instead of "0 / 0".
- Display format `1.2k ↑ 412 ↓` chosen for compactness in the meta
  row. Open to changing to `1.2k in / 412 out` or similar if the
  arrow direction is unclear — input goes up to the model, output
  comes down.

### DIFF
--- a/.changeset/agent-token-usage.md
+++ b/.changeset/agent-token-usage.md
@@ -1,0 +1,27 @@
+---
+'@electric-ax/agents-server-ui': patch
+'@electric-ax/agents-runtime': patch
+'@electric-ax/agents-desktop': patch
+---
+
+Show per-response token usage in the agent meta row, e.g. `1.2k ↑ 412
+↓`. Updates as each step settles — for a single-turn call this lands
+once at done; for tool-using runs the counter jumps at each step
+boundary (the LLM SDK only emits `usage` at end-of-step, so we can't
+tick smoothly between tokens).
+
+Plumbing:
+
+- `StepValue` gains optional `input_tokens` / `output_tokens` columns
+  (Zod + TS). Strictly additive: events recorded before this change
+  stay valid since both fields are optional, so no migration.
+- `outbound-bridge.ts:onStepEnd` now persists the `tokenInput` /
+  `tokenOutput` it already received from `pi-adapter.ts` — previously
+  those values were accepted and silently dropped.
+- `EntityTimelineStepItem` / `IncludesStep` surface the new fields,
+  and the three `.select()` blocks that materialize steps include
+  them.
+- The cached `agent_response` section gets a `tokens?: { input?,
+  output? }` summed across the run's steps at section-build time, and
+  the section-cache fingerprint factors in step token deltas so a
+  late-arriving `onStepEnd` invalidates a stale section.
--- a/packages/agents-runtime/src/entity-schema.ts
+++ b/packages/agents-runtime/src/entity-schema.ts
@@ -154,6 +154,12 @@ type StepValue = {
   model_provider?: string
   model_id?: string
   duration_ms?: number
+  // Token usage for this step as reported by the provider's
+  // end-of-message `usage` payload. Populated on `onStepEnd` when the
+  // adapter has the data — older events without these fields stay
+  // valid (both optional), so this is a strictly additive change.
+  input_tokens?: number
+  output_tokens?: number
 }
 type TextValue = {
   key?: string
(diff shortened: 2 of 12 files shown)


samwillis (commented):
Interactive review with GPT.

Thanks for wiring this through. I traced the token usage path end-to-end:

`pi-adapter message_end.usage`
→ `bridge.onStepEnd({ tokenInput, tokenOutput })`
→ `steps.update({ input_tokens, output_tokens })`
→ timeline step rows
→ UI meta row.

Overall the stream write looks sound: token usage is attached to the step completion update, so it appears once at the end of a pure text generation step, and jumps at step boundaries for tool-using runs.

A couple of suggestions/questions:

1. Can we compute the per-run total in the timeline query/view model?

Right now `AgentResponseLive` subscribes to `run.steps` and sums `input_tokens` / `output_tokens` in React, while `buildAgentSection` separately performs the same aggregation for materialized sections.

If TanStack DB supports this shape cleanly, I think it would be better to expose a single per-run `tokens` field from `createEntityTimelineQuery` / the includes query, e.g. via a scoped aggregate over steps for the run. Then the UI only renders `run.tokens` / `section.tokens`, and the aggregation logic lives in one layer.

Since usage only lands at step end, not token-by-token, updating the parent run/timeline row at those boundaries seems acceptable to me.

2. Missing usage fields currently become real zeroes

In `pi-adapter`, when `msg.usage` exists, missing sides are coerced to `0`:

```ts
...(usage && {
  tokenInput: usage.input ?? usage.inputTokens ?? 0,
  tokenOutput: usage.output ?? usage.outputTokens ?? 0,
}),
```

Now that these values are persisted and displayed, that can make an unknown side look like a real `0`. If `pi-ai` guarantees both input and output are always present, a small regression test would be useful. Otherwise I’d preserve `undefined` for missing sides and only write numeric values.

3. Test coverage

I’d like to see a targeted regression test that proves token usage reaches the `steps.update` event, and another around the timeline/view-model total if we move the aggregation there. That would lock down the important stream contract introduced by this PR.

KyleAMathews:
👍 yeah just showing total tokens for the run at the bottom makes sense to me

samwillis (approved):
:shipit:

PR #4505, opened 2026-06-04
feat(agents): share chats with permissions

icehaunter:
## Motivation

Entity permission enforcement is now in place, and the users table is populated for tenant-scoped identity lookup. This PR exposes that capability in the agents UI so a chat manager can grant access to specific users or all users in the workspace without manually working with principal URLs or grant APIs.

It also tightens the client-side auth and affordances around those permissions so non-managers do not see sharing controls, non-writers cannot send or signal, and shared sessions are recognizable in the sidebar.

## Changes

- Adds a server-defined, tenant-scoped `users` Electric shape and a `usersCollection` in the agents UI provider for names, emails, and avatars.
- Limits synced users columns to display-safe fields: id, display name, email, avatar URL, and timestamps.
- Adds an `entity_effective_permissions` Electric shape and UI hooks for checking current-principal permissions against a chat.
- Adds a chat sharing dialog that can grant roles to a selected user principal or to all workspace users via `subject_kind: principal_kind` and `subject_value: user`.
- Maps sharing roles onto backend permissions explicitly: `View` grants `read` and `fork`; `Chat` grants `read`, `write`, `signal`, `fork`, `schedule`, and `spawn`; `Manage` grants `manage` and `delete`.
- Resolves user principals consistently from raw `user:<id>` keys and `/principal/...` URLs so the current user is filtered from individual sharing, message metadata renders `Me`, and creator checks do not depend on principal encoding.
- Displays synced user names for user principals in message metadata, share rows, sidebar creator filters, and sidebar hover cards when available.
- Adds a small `Users` shared marker to sidebar rows for sessions created by another principal, plus a `Created by` sidebar filter with `Me` pinned first.
- Gates UI actions using effective permissions: share is only visible with manage access; fork, send, stop/signal, pending-message controls, and signal menu actions are disabled when the current principal lacks the corresponding permission.
- Gives forked chats `created_by` for the principal that clicked fork, so the new fork is controlled by its creator rather than the original chat owner.
- Shows the signed-in Cloud user's principal key under Account settings for debugging.
- Injects the signed-in Cloud user as the Electric principal for Cloud-targeted desktop requests, removes the legacy `x-electric-asserted-user-id` header, removes saved Cloud servers on logout so re-login refreshes stale service tokens, and avoids logging injected auth headers.
- Adds focused coverage for principal normalization, share role mappings, users/effective-permissions shapes, fork ownership, Cloud auth header injection, and desktop server fetch auth injection.
- Adds a changeset for `@electric-ax/agents-desktop`, `@electric-ax/agents-server`, and `@electric-ax/agents-server-ui`.

## Decisions

The sharing UI creates explicit grants for every permission implied by a role instead of relying on backend permission implication. That keeps product-level roles transparent: for example, workspace “view” sharing can grant both `read` and `fork` even though `read` itself does not imply `fork`.

“All users” sharing uses the existing principal-kind permission mechanism rather than synthetic user rows. Workspaces are siloed, so `principal_kind:user` means every user principal in the current tenant, not cross-workspace sharing.

Permission gates in the UI are affordance controls, not the source of truth. Server-side permission checks still decide whether grant, signal, fork, spawn, and mutation APIs are allowed.

Cloud desktop auth headers are merged after saved server headers so the signed-in Cloud user wins over stale saved `Electric-Principal` values.

The sidebar shared marker only appears when `created_by` is present and differs from the current normalized principal. Missing creator data is left unmarked rather than inferred.


icehaunter:
Addressed Claude's important review items in `2f35ffab0`:

- Removed `auth-debug.ts` and all `logPostInjectionHeaders` call sites, so injected auth headers and bearer tokens are no longer logged.
- Trimmed the `users` Electric shape to display-safe fields only: `tenant_id`, `id`, `display_name`, `email`, `avatar_url`, `created_at`, `updated_at`.
- Matched the UI `userSchema` to the narrower synced shape.
- Updated `server-utils.test.ts` to assert that `auth_provider`, `auth_subject`, `profile`, and `metadata` are not exposed.

Verification:
- `pnpm --filter @electric-ax/agents-desktop typecheck`
- `pnpm --filter @electric-ax/agents-server typecheck`
- `node_modules/.pnpm/node_modules/.bin/tsx --test packages/agents-desktop/src/cloud/auth-injection.test.ts packages/agents-desktop/src/cloud/server-fetch.test.ts`
- `pnpm --filter @electric-ax/agents-server test -- test/server-utils.test.ts` (expanded to full server suite: 390 passed, 41 skipped)
- `git diff --check`

KyleAMathews:
Thanks for the thorough PR — the shape of this makes sense overall. A few things I think are worth double-checking:

1. **Current principal precedence in the UI vs desktop Cloud injection**

   In `useCurrentPrincipal` we currently prefer:

   ```ts
   getConfiguredActivePrincipal() ?? cloudPrincipal ?? getActivePrincipal()
   ```

   But the desktop Cloud request path now intentionally merges Cloud auth after saved server headers so the signed-in Cloud user wins over stale saved `Electric-Principal` values.

   Could this make the React UI think the current principal is the configured/saved one while actual Cloud requests are sent as the signed-in Cloud user? That seems like it could affect “Me” rendering, filtering self from sharing, creator checks, sidebar filters, and permission affordance checks.

   Should `cloudPrincipal` win for Cloud-targeted servers, or is there another mechanism that guarantees stale saved Cloud principals are removed before this hook matters?

2. **Why does the share dialog use REST-loaded grants instead of always syncing?**

   The dialog loads grant state via `GET /grants`, while effective permissions are synced through Electric. I understand the dialog needs manager-only full grant state, whereas the synced `entity_effective_permissions` shape is scoped to the current principal.

   But should manager-visible entity grants also be exposed as a synced shape, gated by `manage`, so the dialog is live and consistent with the rest of the UI data model? If REST is intentional here, can you add a short note/comment explaining why we don’t always sync grants?

3. **Role updates are not atomic**

   Updating a target’s role does parallel deletes and creates:

   ```ts
   await Promise.all([...deleteRequests, ...createRequests])
   ```

   If one request fails midway, the target could be left with a partial role. Maybe acceptable for now, but should we either serialize/retry more carefully or eventually add a server-side “set role” endpoint that atomically replaces the role permissions?

4. **`desktopServerFetch` error message seems stale**

   `desktopServerFetch` now appears to allow saved Cloud servers too, but the error still says:

   ```ts
   Desktop server fetch is only available for saved local servers
   ```

   Should that say “saved servers” or otherwise distinguish local vs Cloud?

5. **Fallback to `system:dev-local` can make UI affordances over-permissive**

   `useEntityPermission` treats built-in system principals, including `system:dev-local`, as fully allowed. Since `getActivePrincipal()` defaults to `system:dev-local`, a missing/misconfigured principal could make the UI show enabled controls even if the server later rejects the action.

   That’s probably fine for local dev, but can we verify Cloud/prod can’t fall back to `system:dev-local` in a way that creates misleading permission UI?

6. **Effective permissions are only synced for readable entities**

   The `entity_effective_permissions` shape is filtered to entities returned by `buildReadableEntitiesWhere`. That seems good for avoiding grant leakage, but it means a principal with e.g. `write`, `signal`, `schedule`, or `spawn` but no `read` would not see those permissions client-side.

   The current product roles all include `read` for View/Chat and treat `manage` as readable, so maybe this is fine. Just want to confirm the backend model guarantees “action permissions imply/read require read” enough that this won’t hide valid non-read permissions from the UI.

KyleAMathews (approved):

PR #4507, opened 2026-06-04
feat(agents-mobile): pin sessions via long-press context menu

msfstef:
## Intent

The web sidebar lets users pin sessions so frequently-used ones stay at the top regardless of grouping; mobile had no equivalent, so important sessions sink into the date buckets. This ports the pinning feature to the mobile app, adapted for touch — and folds the web's row *hover info card* into the same surface, since mobile had no way to see session metadata from the list either.

## UX

- **Long-press a root session row — or any search result** → bottom sheet with the entity info + a **Pin/Unpin** action.
  - Info is a field-for-field mirror of the web hover card (`SidebarRowInfo`): title, mono session id, `type · status · N subagents`, Runner (only when a runner is pinned), Sandbox (incl. the implicit "Local" default), Spawned (absolute), Last active (relative) — using the same shared formatters/helpers, so values match the web exactly.
- **In-session kebab menu** also gets a **Pin/Unpin** item — the mirror of the desktop tile menu's Pin/Unpin (and the ⌘K "Pin current entity" action), so a session can be pinned from inside it. Together with search long-press this closes the "searched for a session, now want to pin it" gap.
- **Pinned section** renders above the date/type/status groups; pinned roots are removed from the groups and from parent subtrees below (same de-dup as `SidebarTree`), ordered by recency like the web.
- Pinned entities **respect the type/status visibility filters** (web parity — the pinned set derives from the filtered entities).
- In tree mode pinning is **root-only** (web-sidebar parity); long-press on child subtree rows is a no-op. **Search results accept long-press at any depth** (desktop-tile-menu parity) — a pinned child hoists into the Pinned section and is filtered out of its parent's subtree.

### Deliberate divergences from the web (touch context)

| Web | Mobile | Why |
|---|---|---|
| Hover reveals pin icon + info card | Long-press sheet combining both | no hover on touch |
| Pinned section header is collapsible | Not collapsible | no mobile section is collapsible today |
| ⌘K palette groups search results into Pinned/Sessions | Search results stay a flat list (no pinned section/indicator) | search targets a known session by name; can add later |
| "Last active" has absolute-time tooltip | Relative only | no tooltip affordance |
| Per-row pinned glyph (filled, rotated) | None — section placement is the indicator | rows stay clean |

## Implementation choices

- **Storage**: per-device AsyncStorage array of entity urls (`electric-agents-mobile.pinned-entities`) — the mobile mirror of the web's `localStorage` model; not synced to the server. Module-level store like `sidebarPrefs.ts`/`savedServers.ts` (whole-set subscription — a pin toggle changes list composition, so per-url listener buckets à la `expandedTree.ts` would buy nothing), incl. the `dirty`-guarded hydration merge so a toggle during the startup window isn't clobbered. Unit-tested (vitest, gated AsyncStorage mock that genuinely exercises the pre-hydration path).
- **Synced columns extended**: entities now sync `dispatch_policy` + `sandbox`, runners sync `sandbox_profiles` (zod defs mirrored from `ElectricAgentsProvider.tsx`) so the menu can resolve runner/sandbox labels.
- **Shared-code reuse over reimplementation**: the menu reuses `formatTime` + `entityRuntime` helpers from `agents-server-ui`. The only shared-file change is type-level — `entityRuntime`'s *runner* params are loosened to a structural `Pick` subset (`RunnerLike`) since mobile's runner schema is narrower; entity params were left untouched (mobile's entity type is fully assignable).
- **Tree-mode root gating by construction**: `onLongPressRoot` isn't forwarded into the `SessionTree` recursion, so child subtree rows can't get a menu; `pinnedSet` *is* forwarded so pinned entities are filtered out of subtrees before child-count/connector geometry is computed. Search rows wire the long-press directly (flat list, any depth) — safe because the subtree filtering above already handles pinned children.
- The in-session menu reads/toggles the pin via the module store directly (no prop-drilling through `SessionScreen`), closing the sheet on toggle like the row menu does.
- The subagent count in the info sheet is pinned-filtered, matching both the rendered subtree and the web hover card.
- **Review follow-ups** (from the Claude review): the sheet re-resolves the long-pressed entity from the live query while open (snapshot fallback), so its header tracks status/title updates like the child count already did; the runners live query mounts only while the sheet has an entity instead of subscribing for the screen's lifetime; rows with a context menu expose an `accessibilityHint` so the long-press affordance is announced to screen readers.

## Testing

- `pnpm --filter @electric-ax/agents-mobile typecheck && test` — 26 passed (7 new for the pin store)
- `pnpm --filter @electric-ax/agents-server-ui typecheck && test` — 66 passed (type-only change)
- No RN component-test harness in the package, so the UI wiring was verified by review (independent parity + scope reviews) — worth a manual pass in Expo before merge: pin/unpin round-trip, persistence across relaunch, child-row no-op in tree mode, pin from a search result, pin/unpin from the in-session menu, filter interaction.

### DIFF
--- a/.changeset/agents-mobile-pin-sessions.md
+++ b/.changeset/agents-mobile-pin-sessions.md
@@ -1,0 +1,6 @@
+---
+"@electric-ax/agents-mobile": patch
+"@electric-ax/agents-server-ui": patch
+---
+
+Add session pinning to the mobile app: long-press a root session row to open a context sheet with the entity info (title, session id, type/status, subagents, runner, sandbox, spawned, last active) and a Pin/Unpin action. Pinned sessions surface in a Pinned section above the groups, persisted per-device in AsyncStorage — the mobile mirror of the web sidebar's pinning. Runner-param types in agents-server-ui's `entityRuntime` helpers are loosened to structural subsets so the mobile app can reuse them.
--- a/.changeset/agents-mobile-pin-sessions.md
+++ b/.changeset/agents-mobile-pin-sessions.md
@@ -3,4 +3,4 @@
 "@electric-ax/agents-server-ui": patch
 ---
 
-Add session pinning to the mobile app: long-press a root session row to open a context sheet with the entity info (title, session id, type/status, subagents, runner, sandbox, spawned, last active) and a Pin/Unpin action. Pinned sessions surface in a Pinned section above the groups, persisted per-device in AsyncStorage — the mobile mirror of the web sidebar's pinning. Runner-param types in agents-server-ui's `entityRuntime` helpers are loosened to structural subsets so the mobile app can reuse them.
+Add session pinning to the mobile app: long-press a root session row (or any search result) to open a context sheet with the entity info (title, session id, type/status, subagents, runner, sandbox, spawned, last active) and a Pin/Unpin action; the in-session kebab menu also gets a Pin/Unpin item, mirroring the desktop tile menu. Pinned sessions surface in a Pinned section above the groups, persisted per-device in AsyncStorage — the mobile mirror of the web sidebar's pinning. Runner-param types in agents-server-ui's `entityRuntime` helpers are loosened to structural subsets so the mobile app can reuse them.
--- a/packages/agents-mobile/src/components/Icon.tsx
+++ b/packages/agents-mobile/src/components/Icon.tsx
@@ -39,6 +39,7 @@ export type IconName =
   | `cloud`
   | `user`
   | `users`
+  | `pin`
 
 const PATHS: Record<IconName, string> = {
   back: `M15 18l-6-6 6-6`,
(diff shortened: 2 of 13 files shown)


msfstef:
<table>
  <tr>
    <td>Simulator Screenshot - iPhone 15 Pro - 2026-06-04 at 16 55 12</td>
    <td>Simulator Screenshot - iPhone 15 Pro - 2026-06-04 at 16 55 53</td>
  </tr>
  <tr>
    <td>Simulator Screenshot - iPhone 15 Pro - 2026-06-04 at 17 38 30</td>
    <td>Simulator Screenshot - iPhone 15 Pro - 2026-06-04 at 17 38 40</td>
  </tr>
</table>

samwillis:
This is great.

My one comment is can we have it open a menu that can a panel.

ChatGPT opens a menu with a few options, we should copy this. (we don't need to open a preview yet)

Open Codex

msfstef:
> My one comment is can we have it open a menu that can a panel.

@samwillis I don't think I understood this - do you mean instead of opening a bottom sheet to open a panel? If so, we can probably defer that to another PR and flesh out the UI idea a bit more

kevin-dp (approved):

PR #4508, opened 2026-06-04
feat(agents-server-ui): stream model reasoning into the UI

kevin-dp:
## Summary

While the model is "thinking" (Anthropic extended thinking, DeepSeek-R1 reasoning, Moonshot K2, OpenAI Responses summaries) the agent response now shows the reasoning text faded above the answer, with the existing `Thinking` shimmer heading plus elapsed-time ticker. Once the reasoning settles it collapses to `▸ Thought for 12s` — click to expand. Multiple reasoning rows per run render independently in order (one per LLM step in tool-using turns). UX intentionally mirrors Claude Code + OpenCode patterns.

## Implementation (end-to-end)

- **Schema** — `reasoning` row gains `run_id`, `encrypted` (Anthropic redacted-thinking opaque payload, must round-trip back to the model verbatim), and `summary_title` (extracted at write time). New `reasoningDeltas` collection mirrors `textDeltas`. Strictly additive.
- **Bridge** — `OutboundBridge` gains `onReasoningStart` / `onReasoningDelta` / `onReasoningEnd`, parallel to the text path. Reasoning counter added to `OutboundIdSeed`.
- **Adapter** — `pi-adapter.ts` routes pi-ai's `thinking_start` / `thinking_delta` / `thinking_end` events to the bridge. Parses a `**Title**\n\n<body>` heading once at write time (OpenAI Responses; no-op for Anthropic / DeepSeek / Moonshot). Defensive: handles late `thinking_delta` without a preceding `thinking_start`, and closes an open reasoning row on `message_end` (e.g. provider abort).
- **Timeline** — Live `reasoning: Collection<EntityTimelineReasoningItem>` on `EntityTimelineRunRow`, content built via the same delta-join pattern as `EntityTimelineTextItem.content`.
- **UI** — New `<ReasoningSection>` renders above items in `AgentResponseLive`:
  - **Live**: faded markdown via `Streamdown` with `ThinkingIndicator` heading + summary title + elapsed-time ticker
  - **Settled**: `▸ Thought for Ns` with click-to-expand. Closure duration snapshotted from `Date.now() - timestamp` using the same `sawStreamingRef` trick from the elapsed-time PR — accurate for in-session settles, stays a bare `Thought` for rows already settled on first mount (no real end timestamp available client-side).
  - **Redacted**: Anthropic safety-filter payloads render `⊘ Reasoning redacted by provider safety filters`. The encrypted payload is still persisted server-side so the model gets it back on the next turn.

## Reference

Patterns informed by reading OpenCode's reasoning implementation:
- 3-event streaming protocol (`reasoning-start` / `reasoning-delta` / `reasoning-end`)
- `ReasoningPart` storage shape including `encrypted` for Anthropic round-trip
- `reasoningSummary()` headline parser (5-line regex, OpenAI Responses only)
- Collapsed-by-default UX with click-to-expand

## Test plan

- [x] `pnpm typecheck` clean in `agents-runtime` + `agents-server-ui`
- [x] `pnpm test outbound-bridge pi-adapter entity-timeline` in `agents-runtime` (95 passed: 18 bridge + 21 adapter + 56 timeline)
- [x] `pnpm -C packages/agents-runtime build` — dist artifacts emit cleanly
- [ ] Manual: prompt Anthropic Claude with extended-thinking enabled; verify streaming reasoning appears faded above the answer with elapsed ticker, then collapses to `Thought for Ns` on settle
- [ ] Manual: multi-step tool-using turn; verify each step's reasoning renders as a separate collapsible row

## Notes

- Cached `AgentResponse` (the non-Live path used for old scrollback sections) doesn't yet surface reasoning — historical rows recorded before this PR lack the data anyway. Follow-up if we discover sessions where this matters.
- The pre-existing `runtime-dsl.test.ts` 401 failures (and `dispatch-policy-routing.test.ts` 500 failures) reproduce identically on clean `main` and were not introduced by this PR.


KyleAMathews (approved):
Lovely! Could you add a screenshot of the UI to the PR body?

packages/agents-runtime/src/outbound-bridge.ts
```diff
@@ -277,6 +306,56 @@ export function createOutboundBridge(
...
+      writeEvent(
+        entityStateSchema.reasoning.insert({
+          key: currentReasoningKey,
+          value: { status: `streaming`, run_id: runKey } as never,
+        }) as ChangeEvent
+      )
+    },
+
+    onReasoningDelta(delta: string) {
+      if (!currentReasoningKey) return
```
kevin-dp, line 346: These 3 functions `onReasoningStart`, `onReasoningDelta`, and `onReasoningEnd` rely on a shared `currentReasoningKey` which assumes that reasoning is sequential, i.e. there are no models that are concurrently reasoning. Is that the case? What if 2 or more Horton sessions are running in parallel, couldn't those thinking blocks interfere?

kevin-dp: It's fine. This variable is local to the closure of the `createOutboundBridge` function.
Closure-scoped per bridge, bridge is per agent.run, agent.run is per session — parallel Horton sessions can't interfere.

kevin-dp:
@KyleAMathews here are some screenshots showing how it displays while it's thinking and how it displays when it's done thinking (the "Thought for 2s" block is expandable on click).

thinking-block
thought-block

PR #4509, opened 2026-06-04
feat(agents): per-runner recent working directories synced via sessions

msfstef:
## Goal

The Horton spawn prompt needs a working directory. On desktop a native folder picker works, but when spawning onto a runner on **another machine** — or from the **mobile app** — there's no way to browse the runner's filesystem. Mobile previously hardcoded *"Mobile uses the server default working directory for now"*, and desktop recents were a per-device localStorage list with no runner awareness.

This PR gives every client a **per-runner recent working directories** picker that syncs across the user's devices, and makes mobile able to choose a sandbox profile + working directory at all.

## Design decisions

**1. Recents are derived from synced sessions, not stored.**
We originally planned to attach recents as metadata/state on the user's principal entity. Review showed everything needed is *already synced to every device*: each session row carries `spawn_args.workingDirectory` and its `dispatch_policy` runner target, scoped to the user. So recents are just a live query: filter sessions by runner → dedupe paths keeping newest `updated_at` → top 10 (`recentWorkingDirsForRunner` in `agents-server-ui/src/lib/recentWorkingDirectories.ts`). Zero new server state, no migration, no new auth surface, and existing sessions populate the list retroactively.

Accepted trade-offs: recents live only as long as the sessions that reference them, and there's no per-row "remove" (they're facts, not a curated list). The principal-entity-state design remains the follow-up if curation/durability is ever needed — the derived list could seed it.

**2. Mobile sends the sandbox profile with the working directory (bug fix).**
`args.workingDirectory` is only honoured through a sandbox-profile factory (`packages/agents/src/bootstrap.ts` `resolveCwd`); a spawn with no `sandbox` falls back to `unrestrictedSandbox(process.cwd())` and silently ignores the arg (`process-wake.ts`). Desktop always sent a profile; mobile sent none — so without this fix the chosen directory would be a no-op. Mobile now syncs `sandbox_profiles` on its runners shape, shows a profile picker (advertised order, first = default), and sends `sandbox: { profile, key: entityUrl }` like desktop.

**3. Remote profiles take no working directory.**
For local profiles (local/unrestricted, Docker) the working-directory section applies; for `remote: true` profiles the workspace lives in the provider VM, so the section is hidden and no directory is sent — same gating as desktop (`isSandboxProfileRemote`).

**4. Shared helpers extracted, not duplicated.**
`pickDefaultSandboxProfile` / `useSandboxProfileSelection` / `isSandboxProfileRemote` moved from `NewSessionView.tsx` into `agents-server-ui/src/lib/sandboxProfiles.ts`, deep-imported by mobile per the existing convention.

**5. Desktop picker becomes per-runner; localStorage hook retired.**
A desktop user targeting a remote runner needs *that runner's* recents, which localStorage can't provide. The picker now takes derived per-runner recents and defaults to the selected runner's most recent path (explicit picks win; switching runner re-derives, since paths from one machine may not exist on another). No migration needed — existing sessions already encode real recents.

## Changes

- `agents-server-ui`: new `lib/recentWorkingDirectories.ts` + tests, new `lib/sandboxProfiles.ts` (pure extraction), `WorkingDirectoryPicker` takes `recents` prop (✕-remove dropped), `NewSessionView` wires per-runner recents + default, deleted `hooks/useRecentWorkingDirectories.ts`
- `agents-mobile`: entities shape + `dispatch_policy`, runners shape + `sandbox_profiles`, `spawnEntity` gains `sandboxProfile`/`workingDirectory` (+ tests), `NewSessionScreen` gains Sandbox and Working-directory sections (recents cards, free-text absolute path, runner default)

## Testing

- New unit tests: derivation helper (5) and mobile spawn body/schemas (6); full suites pass (server-ui 71, mobile 25), typecheck + lint clean
- Not yet verified end-to-end against a live desktop runner + mobile device (spawn from mobile into a picked directory and confirm the session `pwd`s there) — flagged for review


msfstef:
Re: the two suggestions carried over in review iteration 2 — both are deliberate, documenting the reasoning here:

**#2 — mobile defaults to "Runner default" while desktop defaults to the most-recent dir.** Intentional, decided during design: mobile's cold-start flow is *runner default + free-text path*, while desktop preserves its pre-existing "resume where you last worked" behavior. Strict parity isn't the goal — mobile users typically spawn into a remote runner whose paths they may not know, so defaulting to the runner's configured directory is the safer default. If usage shows mobile users always tapping the newest recent, pre-selecting it is a one-line follow-up.

**#3 — the desktop default-selection effect interplay isn't unit-tested.** Acknowledged coverage boundary: `agents-server-ui` has no render-test harness (pure vitest, no jsdom/`renderHook`), so the `userPickedDirRef` reset/re-derive ordering relies on manual verification. The derivation logic itself (`recentWorkingDirsForRunner`) is fully covered. Adding a component-test harness to the package is out of scope for this PR.

Remaining before un-drafting: the live E2E check (spawn from mobile into a picked directory against a real runner and confirm the session `pwd`s there).

msfstef:
<table>
  <tr>
    <td>Simulator Screenshot - iPhone 15 Pro - 2026-06-04 at 18 56 56</td>
    <td>Simulator Screenshot - iPhone 15 Pro - 2026-06-04 at 18 56 51</td>
  </tr>
</table>

kevin-dp (approved):
Had a look at the code and looks good to me!

PR #4513, opened 2026-06-04
fix(agents): reclaim zombie sandbox containers and create them lazily

msfstef:
## What & why

Opening the desktop app could leave **15+ leftover `electric-sbx-*` Docker containers running** that the user never started. They piled up because, in several places, a sandbox container could outlive the work it was created for. This PR closes each of those gaps so a container only exists while something is actually using it.

## Why the containers piled up

Sandbox containers are meant to be short-lived: created for an agent's work, then torn down. Three gaps let them survive instead:

1. **We created containers we didn't need.** Every time an agent woke up — even for trivial bookkeeping (a scheduled tick with nothing to do, a drained message queue) — it started a container *before* knowing whether it would run anything. When the app reconnected and replayed a backlog of these wakes, each one spun up its own container at once.
2. **We didn't clean up on quit.** Idle containers are torn down by a short (~2-minute) delay timer. On quit, the app exits before those timers fire — so every quit left a container behind.
3. **We didn't clean up leftovers at the next startup either.** The startup cleanup deliberately skipped *running* containers (so it wouldn't kill one another process is actively using). But a container left by a crash or quit is still running — its main process is an idle loop that never exits on its own — so the cleanup never touched the very leftovers it was meant to reclaim.

Two smaller bugs made it worse: a container could be stranded if its one-time setup step failed, and a Docker error code was occasionally misread so we'd "reconnect" to a container that had already been removed.

## How we fix it

The approach attacks all three gaps so a container can't outlive its purpose:

1. **Create only when actually used (lazy creation).** A container is now started the first time an agent really uses it — runs a command, or reads/writes a file. Trivial wakes create nothing. (Edge cases preserved: an agent finishing for good still cleans up any workspace earlier runs created, and a child agent that "inherits" its parent's sandbox still makes sure that sandbox exists first.)
2. **Clean up on quit.** On shutdown we now run the pending teardowns immediately instead of letting the delay timers die with the process. This covers every exit path (desktop quit, restarts, Ctrl-C / kill) and is time-bounded so a stuck Docker daemon can't hang the quit.
3. **Reclaim true leftovers at startup.** Every container is tagged with the process that created it. At startup we check whether that process is still alive: if it's gone, the container is a genuine leftover and we reclaim it — deleting throwaway ones and stopping reusable ones (so their files survive for later reuse). Containers a live process is still using are left untouched, so we never disturb a peer.

As a quality-of-life touch, every container is also labelled so Docker Desktop groups them under one collapsible **`electric-sandboxes`** entry you can stop or delete together (`docker compose -p electric-sandboxes down`).

<details>
<summary>Implementation details (for reviewers)</summary>

### Zombie reclamation

- Containers carry a `com.electric.sandbox.owner-pid` label; the boot sweep probes it (`kill(pid, 0)`) and reclaims running orphans whose owner is dead — **remove** ephemeral, **stop** persistent (writable layer survives for reattach by key). The sweep probes + reclaims leftovers concurrently so boot latency stays flat as they accumulate.
- Labels are immutable, so reattaching a dead creator's container records adoption in an in-container marker (`/tmp/.electric-sbx-owner-pid`, tmpfs ⇒ wiped on stop); the sweep probes it before reclaiming, so a live sibling's adopted container is never swept. The sweep is now awaited at boot so it can't race the first wake's reattach.
- New `shutdownAllDockerSandboxes()` flushes pending debounced teardowns on runtime shutdown, wired through `AgentHandlerResult.shutdownSandboxes` → `BuiltinAgentsServer.stop()` (covers desktop quit, runtime restarts, CLI SIGINT/SIGTERM; bounded 5s). Live-leased entries are left to their own dispose (a sibling runtime in the same process may own them) — if the process dies first, the pid-sweep reclaims them at next boot.
- Post-start init verifies the `mkdir` exit code and force-removes the container on any failure; `isNameConflict()` now matches the daemon's actual name-conflict message instead of bare 409.

### Lazy sandbox creation

- New `lazySandbox()` wrapper (`agents-runtime/sandbox/lazy.ts`) defers the provider factory until the sandbox is actually **used** (exec/fs/fetch). The bootstrap `docker` profile returns it, so trivial wakes never create a container. Materialization is single-flight and retried on failure.
- Terminal reclaim still works without use: `reclaimDockerSandboxByKey()` wipes an earlier wake's persistent workspace by key without creating a container just to delete it (owner leases only; defers to live sibling leases — the last one draining wipes it).
- Spawn-`inherit` force-materializes the owner's workspace (`ensureSandboxMaterialized`) before spawning, so a child can attach even when the parent never ran a tool.
- Concurrent container **creations** are capped at 4 process-wide (`withCreationSlot`) — bursts queue against the daemon instead of stampeding it; reattaches/execs are unlimited, total creations unbounded.

### Grouping

All sandboxes carry `com.docker.compose.project=electric-sandboxes` (+ `com.docker.compose.service=<entity-type>`).

</details>

## Testing

- 13 new docker integration tests (running-orphan reclaim, adoption sparing, legacy-label safety, init-failure cleanup, shutdown flush, lazy composition, reclaim-by-key) and 13 unit tests for `lazySandbox` — written first, confirmed red, then green.
- Full suites: agents-runtime 803 tests / 63 files, agents 55 / 11; `tsc` and eslint clean.

## Out of scope (follow-ups)

- No cap on concurrent *wakes* (they're whole agent runs; capping would queue user-visible messages behind backlog replay — needs a product call).
- The `e2b` remote profile stays eager (working directory not statically known at profile-build time); the same wrapper applies once it is.
- Server-side orphaned-claim recovery (`dispatchRecoveryIntervalMs` is defined but unused) — expired runner leases still queue wakes forever.

### DIFF
--- a/packages/agents-runtime/src/sandbox/docker.ts
+++ b/packages/agents-runtime/src/sandbox/docker.ts
@@ -745,42 +745,48 @@ export async function sweepOrphanedDockerSandboxes(opts?: {
     return []
   }
 
-  const reclaimed: Array<string> = []
-  for (const c of listed) {
-    const name = c.Names?.[0]?.replace(/^\//, ``) ?? c.Id
-    // Held by this very process right now — its lifecycle is already managed.
-    if (sandboxContainers.has(name)) continue
-    const persistent = c.Labels?.[SANDBOX_PERSISTENT_LABEL] === `true`
-    try {
-      if (c.State === `running`) {
-        const ownerPid = Number(c.Labels?.[SANDBOX_OWNER_PID_LABEL])
-        // No (or unparsable) owner label, or the owner is alive ⇒ possibly a
-        // live peer's in-use sandbox — never touch it.
-        if (!Number.isInteger(ownerPid) || isPidAlive(ownerPid)) continue
-        // The creator is gone, but a live sibling process may have since
-        // ADOPTED the container (reattached by key) — labels are immutable, so
-        // adoption is recorded in an in-container marker. Probe it before
-        // reclaiming.
-        const adopter = await readOwnerMarker(docker.getContainer(c.Id))
-        if (adopter !== null && isPidAlive(adopter)) continue
-        if (persistent) {
-          // `t: 0` → straight to SIGKILL (PID 1 ignores SIGTERM); the writable
-          // layer survives for a later reattach by key.
-          await docker.getContainer(c.Id).stop({ t: 0 })
-        } else {
-          await docker.getContainer(c.Id).remove({ force: true, v: true })
+  // Probe + reclaim every leftover concurrently. The running-orphan path runs an
+  // exec round-trip (readOwnerMarker) and the whole sweep is awaited at boot, so
+  // a sequential loop would serialize one exec per leftover — and the motivating
+  // case is "15+ leftovers". Each container is independent (distinct name, no
+  // shared lock), so fanning out keeps boot latency flat as leftovers pile up.
+  const reclaimed = await Promise.all(
+    listed.map(async (c): Promise<string | null> => {
+      const name = c.Names?.[0]?.replace(/^\//, ``) ?? c.Id
+      // Held by this very process right now — its lifecycle is already managed.
+      if (sandboxContainers.has(name)) return null
+      const persistent = c.Labels?.[SANDBOX_PERSISTENT_LABEL] === `true`
+      try {
+        if (c.State === `running`) {
+          const ownerPid = Number(c.Labels?.[SANDBOX_OWNER_PID_LABEL])
+          // No (or unparsable) owner label, or the owner is alive ⇒ possibly a
+          // live peer's in-use sandbox — never touch it.
+          if (!Number.isInteger(ownerPid) || isPidAlive(ownerPid)) return null
+          // The creator is gone, but a live sibling process may have since
+          // ADOPTED the container (reattached by key) — labels are immutable, so
+          // adoption is recorded in an in-container marker. Probe it before
+          // reclaiming.
+          const adopter = await readOwnerMarker(docker.getContainer(c.Id))
+          if (adopter !== null && isPidAlive(adopter)) return null
+          if (persistent) {
+            // `t: 0` → straight to SIGKILL (PID 1 ignores SIGTERM); the writable
+            // layer survives for a later reattach by key.
+            await docker.getContainer(c.Id).stop({ t: 0 })
+          } else {
+            await docker.getContainer(c.Id).remove({ force: true, v: true })
+          }
+          return name
         }
-        reclaimed.push(name)
-        continue
+        if (persistent) return null
+        await docker.getContainer(c.Id).remove({ force: true, v: true })
+        return name
+      } catch {
+        /* already gone */
+        return null
       }
-      if (persistent) continue
-      await docker.getContainer(c.Id).remove({ force: true, v: true })
-      reclaimed.push(name)
-    } catch {
-      /* already gone */
-    }
-  }
-  return reclaimed
+    })
+  )
+  return reclaimed.filter((n): n is string => n !== null)
 }
 
 /**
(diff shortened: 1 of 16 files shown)


packages/agents-runtime/src/sandbox/docker.ts
```diff
@@ -209,6 +234,33 @@ function withKeyLock<T>(key: string, fn: () => Promise<T>): Promise<T> {
...
 
+/**
+ * Bound on concurrent container CREATIONS (image resolve + create + start +
+ * init), so a burst of wakes materializing sandboxes at once — e.g. a backlog
+ * replay when a runner reconnects — queues against the daemon instead of
+ * stampeding it. Reattaches and execs are not limited; total creations are
+ * unbounded, only their concurrency.
+ */
+const MAX_CONCURRENT_CREATIONS = 4
+let creationSlots = MAX_CONCURRENT_CREATIONS
```
kevin-dp, line 245: Why use a separate `creationSlots` variable whose sole purpose is counting the amount of available slots in `creationQueue`? This means we have to take care of keeping `creationSlots` in sync with `creationQueue`. It would be simpler and more robust to use only `creationQueue` and rely on its length to determine available slots. So `creationSlots` becomes `MAX_CONCURRENT_CREATIONS - creationQueue.length`. But instead of writing it like that we would probably write it like this:
```ts
if (creationQueue.length >= MAX_CONCURRENT_CREATIONS) {
  await new Promise<void>((release) => creationQueue.push(release))
}
try {
  return await fn()
} finally {
  const next = creationQueue.shift()
  next?.()
}
```

packages/agents-runtime/src/sandbox/docker.ts
```diff
@@ -496,6 +571,25 @@ export async function dockerSandbox(
...
+ * exited process — so the boot sweep can tell an adopted container from a
+ * crash leftover.
+ */
+async function writeOwnerMarker(c: DockerodeContainer): Promise<void> {
+  try {
+    await runOneOff(c, [
+      `sh`,
+      `-c`,
+      `echo ${process.pid} > ${OWNER_MARKER_PATH}`,
+    ])
```
kevin-dp: Can write this a bit more compact with:
```ts
return runOneOff(c, [...]).catch(() => {})
```

kevin-dp (approved):

packages/agents-runtime/src/sandbox/docker.ts
msfstef: It's tricky if we don't keep track of the in flight creations and blocked waiters separately. Your code doesn't work (since nothing is ever added to the creation queue), but I assume you mean that essentially we _always_ add creations to the creation queue, something like:

```ts
await new Promise<void>((release) => {
  creationQueue.push(release)
  if (creationQueue.length <= MAX_CONCURRENT_CREATIONS) release()
})
try {
  return await fn()
} finally {
  creationQueue.shift()
  // unblock any waiter entering active window
  creationQueue[MAX_CONCURRENT_CREATIONS - 1]?.() 
}
```

Or something like this - we still essentially keep track of in flight creations via the first `MAX_CONCURRENT_CREATIONS` slots of the queue, but at least to me this seems less straightforward to read and mixes the responsibility of the creation queue as both a queue and and a counter/stack.

I agree with your sentiment but I think I won't try to make this more clever. An alternative could be to just have `MAX_CONCURRENT_CREATIONS` promise queues and distribute creations between them and then that sorts itself out automatically, with the loss of a slow/stuck creation blocking everyone in that queue.

If you don't mind I might leave this as is, I think it works and is clear in its purpose, and I don't think there's a real concern about keeping track of that variable given the conciseness and limited role it plays - perhaps encapsulating it more would make it less scary.

kevin-dp: Aha i see. I was a bit confused by what the code actually does. It's basically a counting semaphore. I'm fine keeping it like this for now, probably not worth creating a counting semaphore abstraction for it right now.

PR #4522, opened 2026-06-05
Fix agent send attribution auth

KyleAMathews:
## Summary
- include both from_principal/from_agent plus the active wake write token on runtime-originated sends
- allow server send attribution when from_agent is backed by a valid active agent write token
- keep spoofed from_agent rejection and add coverage for the write-token path

## Tests
- pnpm --filter @electric-ax/agents-mcp build
- pnpm --filter @electric-sql/client build
- pnpm --filter @electric-ax/agents-runtime build
- cd packages/agents-server && pnpm vitest run test/electric-agents-routes.test.ts -t "send endpoint"
- cd packages/agents-runtime && pnpm vitest run test/process-wake.test.ts -t "executes send via server API"
- pnpm --filter @electric-ax/agents-runtime typecheck
- pnpm --filter @electric-ax/agents-server typecheck

### DIFF
--- a/.changeset/fix-agent-send-auth.md
+++ b/.changeset/fix-agent-send-auth.md
@@ -1,0 +1,6 @@
+---
+"@electric-ax/agents-runtime": patch
+"@electric-ax/agents-server": patch
+---
+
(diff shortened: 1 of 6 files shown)


samwillis (approved):
Reviewed the auth/runtime attribution changes and surrounding send permission path. No issues found; targeted runtime and server send tests pass.

PR #4524, opened 2026-06-06
Preserve Cloud auth dispatcher when enabling fetch cache

KyleAMathews:
## Summary
- compose the Durable Streams fetch cache on top of the existing undici global dispatcher instead of replacing it
- preserve Electric Agents Desktop's Cloud auth header injector after the built-in agents runtime starts
- add a regression test for dispatcher composition

## Testing
- pnpm --filter @electric-ax/agents test test/durable-streams-cache.test.ts
- pnpm --filter @electric-ax/agents build

## Notes
Debugging showed Cloud setup succeeded, but after the built-in agents runtime installed the Durable Streams cache, subsequent Cloud requests lost Authorization and returned 401 {"error":"Missing bearer token"}.

### DIFF
--- a/.changeset/fix-desktop-cloud-auth-dispatcher.md
+++ b/.changeset/fix-desktop-cloud-auth-dispatcher.md
@@ -1,0 +1,5 @@
+---
+"@electric-ax/agents": patch
+---
+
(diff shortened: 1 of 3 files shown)


samwillis (approved):
LGTM

PR #4527, opened 2026-06-07
feat(agents-desktop): UI improvements

samwillis:
## Summary

This PR collects several agents desktop/server UI improvements:

- Adds a nested Horton model settings menu that groups model, reasoning, and speed choices, with the closed trigger showing the model and lighter thinking-level text.
- Fixes select/menu presentation issues, including cropped descenders in select triggers and excess dropdown bottom spacing.
- Adds and corrects tooltips for spawn form/chat timeline controls, including attach, model, working directory, send, runner, and sandbox controls.
- Renames the local sandbox option to `No sandbox`.
- Refines macOS desktop sidebar vibrancy by using transparent chrome with backdrop-filter tinting, visible backdrop-aware hover/active states for sidebar rows and controls, blended full-width dividers, and light-mode-only content edge shadows.
- Moves the timeline “Fork from here” action from a hover-only historical prompt button to an always-visible assistant response footer action beside copy response.

## Validation

- `git diff --check`
- `asdf exec corepack pnpm --filter @electric-ax/agents-server-ui typecheck`

### DIFF
--- a/.changeset/fuzzy-agents-ui.md
+++ b/.changeset/fuzzy-agents-ui.md
@@ -1,0 +1,7 @@
+---
+"@electric-ax/agents-server-ui": patch
+---
+
+Polish the agents UI with improved spawn-form model controls, tooltips,
+macOS sidebar vibrancy styling, select sizing fixes, and a response-footer
+fork action in the timeline.
--- a/packages/agents-server-ui/src/components/AgentResponse.module.css
+++ b/packages/agents-server-ui/src/components/AgentResponse.module.css
@@ -36,6 +36,10 @@
   gap: 2px;
 }
 
+.tooltipTrigger {
+  display: inline-flex;
+}
+
 .metaActionButton {
   color: var(--ds-text-4);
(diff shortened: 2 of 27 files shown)


KyleAMathews:
Found one small issue:

**P2 — Disabled fork action tooltip won’t show**

In `packages/agents-server-ui/src/components/AgentResponse.tsx`, `ResponseMetaActions` wraps the fork `IconButton` directly in `Tooltip`, but when `forkDisabled` is true the underlying button is disabled. Native disabled buttons do not emit hover/focus/pointer events, so the “Fork permission required” tooltip will not appear — exactly the state where the tooltip is most useful.

Elsewhere in this PR, disabled send buttons are wrapped in a `<span className={styles.tooltipTrigger}>…</span>` before applying `Tooltip`; using the same pattern here, or otherwise rendering a non-disabled tooltip trigger wrapper around the disabled button, should fix it.

KyleAMathews (approved):

PR #4530, opened 2026-06-08
Add collaborative markdown document tools

samwillis:
## Summary

Stacked on #4555. Adds collaborative markdown documents for Electric Agents. This PR consumes the streaming tool-call hook from #4555, but keeps the markdown document implementation in a separate layer.

What this adds:

- A `document` manifest entry for markdown docs with:
  - content MIME: `text/markdown`
  - transport MIME: `application/vnd.electric-agents.markdown-yjs`
  - provider: `y-durable-streams`
  - Y.Text name: `markdown`
- Horton/worker markdown tools for creating, reading, replacing, editing, and streaming inserts into markdown docs.
- Runtime-side document sessions that materialize a wake-local `Y.Doc`, keep it synced through `@durable-streams/y-durable-streams`, and append binary Yjs updates to durable streams.
- Raw Yjs update and awareness stream support on the server; the server creates streams and serves metadata but does not semantically edit document content.
- CodeMirror/Yjs markdown editor UI with open and split-right actions from timeline and manifest surfaces.
- Agent and user presence plumbing based on Yjs awareness, with agent cursor placement during streamed insert/replace operations.
- Worker sharing of markdown document refs so spawned workers can collaborate on the same document.

Architecture notes:

- The agent runtime owns semantic document edits. Tools mutate the local wake-scoped `Y.Doc`; the Yjs provider writes updates to the durable stream.
- The agents server stays deliberately dumb for document contents: it creates/locates streams, proxies durable stream access, and stores the manifest row.
- Presence follows the same split: clients and agents publish raw Yjs awareness updates; the server does not interpret cursor positions.
- Basic replace/edit tools remain available for non-streaming edits; streamed insert/replace builds on #4555's `onArgsDelta` hook.

## Validation

- `pnpm --filter @electric-ax/agents typecheck`
- `pnpm --filter @electric-ax/agents-server typecheck`
- `pnpm --filter @electric-ax/agents-runtime exec vitest run test/markdown-docs-tools.test.ts test/outbound-bridge.test.ts test/pi-adapter.test.ts`
- `pnpm --filter @electric-ax/agents-runtime exec vitest run test/runtime-dsl.test.ts`
- `pnpm --filter @electric-ax/agents exec vitest run test/spawn-worker-tool.test.ts test/horton-tool-composition.test.ts test/horton-system-prompt.test.ts`
- `pnpm --filter @electric-ax/agents-server exec vitest run test/electric-agents-routes.test.ts test/electric-agents-status.test.ts test/electric-agents-manager-write-validation.test.ts`
- `pnpm --filter @electric-ax/agents-server-ui exec vitest run src/components/views/MarkdownDocumentView.test.ts src/lib/workspace/workspaceReducer.test.ts`
- `pnpm --filter @electric-examples/yjs run typecheck`
- `pnpm --filter @electric-ax/agents-mobile run ci:check`
- `git diff --check`


kevin-dp (changes requested):
I checked out this branch to give it a spin. I asked Horton to create a markdown document which it did. Then i opened it in the pane on the right and i added some text to the document. Then i asked Horton to tell me what i added to the document. However, it told me that the document is unchanged. And if i open the document again the changes are indeed missing. So it seems that my changes aren't automatically saving.

I asked Claude about it and it told me this is a bug:

Found it — **this is a real bug in the PR, not a missing save step**. Your edits are being silently rejected by the server. Here's the full chain, verified empirically against your running stack:

## What's happening

1. There is no save button by design — the editor streams every keystroke to a server-side "updates stream" for the doc.
2. But when Horton's turn (wake) that created the document finished, the agent runtime closed its document session. `YjsProvider.disconnect()` ends by calling `producer.close()`, which sends a final append with a **`stream-closed: true`** header — and that **permanently closes the document's update stream on the server**.
3. From that moment, every write from anyone — your editor, future agent edits — gets **HTTP 409 "Stream is closed"**. The client library swallows the error (its `flush()` even resolves successfully), so the UI keeps showing "Connected" and your text stays in local memory only.
4. When Horton re-read the doc, it correctly saw only the original server-side content — your sentence never made it there.

## How I verified it

- Fetched your doc's raw stream from the server: it contains Horton's original markdown but no "SYNCINGG" — your edit never landed.
- Replayed the editor's exact write path with a script: appends to your doc return `409 Stream is closed`.
- Controlled experiment on a fresh doc: session A writes fine (200), disconnects (final append carries `stream-closed: true`, server answers 204), then session B's writes all fail with 409. Reproduces every time.

## Where the fix belongs

The closing call is in `@durable-streams/y-durable-streams` (v0.2.7, latest): `disconnect()` → `closeUpdatesProducer()` → `producer.close()`, and in `@durable-streams/client` a producer's `close()` means "close the stream", not "leave the session". For a multi-writer collaborative doc, no single participant's disconnect should close the shared stream. Either the provider needs a flush-without-close disconnect, the PR's wake-scoped session `close()` (`packages/agents-runtime/src/markdown-document-session.ts:100`) shouldn't disconnect this way, or the agents-server needs to implement the stream-rollover the protocol seems to expect. A second, smaller issue: the UI shows "Connected" while writes 409 — sync failures should surface in that status indicator.

There's no workaround for testing user→agent edits right now — every doc's stream gets closed as soon as the creating wake ends.

PR #4531, opened 2026-06-08
Add queryable_columns shape allow-list

robacourt:
## Summary

Adds `queryable_columns` as a server-side shape allow-list for columns that may be referenced by shape queries. This decouples the security boundary from the `columns` sync projection, so proxies can strictly control which columns are queryable while still allowing clients to request a narrower synced projection.

References security advisory: https://github.com/electric-sql/electric/security/advisories/GHSA-c82q-v86f-c87f

## Changes

- Parse and validate the new `queryable_columns` shape parameter.
- Restrict top-level `where`, subset filters/orderings, and `columns` projections to queryable columns.
- Default synced columns to `queryable_columns` when no `columns` projection is provided.
- Add router and shape tests for normal and subset behavior.
- Update HTTP docs, auth/shapes guides, OpenAPI, and add a changeset.

### DIFF
--- a/packages/sync-service/test/electric/plug/router_test.exs
+++ b/packages/sync-service/test/electric/plug/router_test.exs
@@ -1061,6 +1061,95 @@ defmodule Electric.Plug.RouterTest do
       assert [%{"key" => ^key, "value" => ^value}, _] = Jason.decode!(conn.resp_body)
     end
 
+    @tag with_sql: [
+           "CREATE TABLE queryable_users (id BIGINT PRIMARY KEY, name TEXT NOT NULL, secret_token TEXT NOT NULL)",
+           "INSERT INTO queryable_users VALUES (1, 'Alice', 'supersecret123')"
+         ]
+    test "GET allows where clauses on queryable columns that are not synced", %{opts: opts} do
+      conn =
+        conn("GET", "/v1/shape", %{
+          table: "queryable_users",
+          offset: "-1",
+          queryable_columns: "id,name,secret_token",
+          columns: "id,name",
+          where: "secret_token LIKE 'super%'"
+        })
+        |> Router.call(opts)
+
+      assert %{status: 200} = conn
+
+      assert [
+               %{
+                 "headers" => %{"operation" => "insert"},
+                 "value" => %{"id" => "1", "name" => "Alice"}
+               },
+               %{"headers" => %{"control" => "snapshot-end"}}
+             ] = Jason.decode!(conn.resp_body)
+    end
+
+    @tag with_sql: [
+           "CREATE TABLE queryable_users (id BIGINT PRIMARY KEY, name TEXT NOT NULL, secret_token TEXT NOT NULL)",
+           "INSERT INTO queryable_users VALUES (1, 'Alice', 'supersecret123')"
+         ]
+    test "GET rejects where clauses and synced columns outside queryable_columns", %{opts: opts} do
+      conn =
+        conn("GET", "/v1/shape", %{
+          table: "queryable_users",
+          offset: "-1",
+          queryable_columns: "id,name",
+          columns: "id,name",
+          where: "secret_token LIKE 'super%'"
+        })
+        |> Router.call(opts)
+
+      assert %{status: 400} = conn
+      assert %{"errors" => %{"where" => [message]}} = Jason.decode!(conn.resp_body)
+      assert message =~ "unknown reference secret_token"
+
+      conn =
+        conn("GET", "/v1/shape", %{
+          table: "queryable_users",
+          offset: "-1",
+          queryable_columns: "id,name",
+          columns: "id,name,secret_token"
+        })
+        |> Router.call(opts)
+
+      assert %{status: 400} = conn
+
+      assert %{
+               "errors" => %{
+                 "columns" => ["The following columns are not queryable: secret_token"]
+               }
+             } = Jason.decode!(conn.resp_body)
+    end
+
+    @tag with_sql: [
+           "CREATE TABLE queryable_users (id BIGINT PRIMARY KEY, name TEXT NOT NULL, secret_token TEXT NOT NULL)",
+           "INSERT INTO queryable_users VALUES (1, 'Alice', 'supersecret123')"
+         ]
+    test "GET defaults synced columns to queryable_columns when columns are omitted", %{
+      opts: opts
+    } do
+      conn =
+        conn("GET", "/v1/shape", %{
+          table: "queryable_users",
+          offset: "-1",
+          queryable_columns: "id,name"
+        })
+        |> Router.call(opts)
+
+      assert %{status: 200} = conn
+
+      assert [
+               %{
+                 "headers" => %{"operation" => "insert"},
+                 "value" => %{"id" => "1", "name" => "Alice"}
+               },
+               %{"headers" => %{"control" => "snapshot-end"}}
+             ] = Jason.decode!(conn.resp_body)
+    end
+
     test "GET works when there are changes not related to the shape in the same txn", %{
       opts: opts,
       db_conn: db_conn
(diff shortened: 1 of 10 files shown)


website/docs/sync/api/http.md
```diff
@@ -77,6 +77,8 @@ When you make an initial sync request, with `offset=-1`, you're telling the serv
 
 When a shape is first requested, Electric queries Postgres for the data and populates the log by turning the query results into insert operations. This allows you to sync shapes without having to pre-define them. Electric then streams out the log data in the response.
 
+The `columns` query parameter controls which columns are synced in the response. The `queryable_columns` query parameter controls which columns may be referenced by the shape `where` clause, subset filters, subset ordering, and the `columns` projection. If `queryable_columns` is set and `columns` is omitted, Electric syncs the queryable columns by default.
```
robacourt: This feels like two much detail here

website/docs/sync/guides/auth.md
```diff
@@ -333,7 +334,8 @@ The proxy must set these **shape definition parameters** server-side — they de
 | Parameter | Where | Security Consideration |
 |-----------|-------|------------------------|
 | `table` | URL | **Must be set server-side.** Letting clients specify the table allows access to any table. |
-| `columns` | URL | **Should be set server-side.** Clients could request sensitive columns. |
+| `queryable_columns` | URL | **Should be set server-side.** This is the column allow-list for WHERE clauses, subset filters, ordering, and synced projections. |
+| `columns` | URL | Optional sync projection. If clients can choose it, `queryable_columns` must also be set server-side so clients cannot sync columns outside the allow-list. |
```
robacourt: why?

website/docs/sync/guides/auth.md
```diff
@@ -356,15 +358,15 @@ These parameters are safe because they either can't widen data access or are nee
 | `order_by` | POST body | Sorting for pagination |
 
 :::tip Key Principle
-Your proxy is an **authorization layer** that controls the **shape definition** (table, columns, main WHERE). Clients can freely use subset parameters to filter and paginate within that shape — Electric ensures they can only narrow results, never escape the main WHERE clause.
+Your proxy is an **authorization layer** that controls the **shape definition** (table, queryable columns, main WHERE). Clients can freely use subset parameters to filter and paginate within that shape — Electric ensures they can only narrow results and can only reference queryable columns.
```
robacourt: never escape the main WHERE clause AND can only reference queryable columns

website/docs/sync/guides/auth.md
```diff
@@ -333,7 +334,8 @@ The proxy must set these **shape definition parameters** server-side — they de
 | Parameter | Where | Security Consideration |
 |-----------|-------|------------------------|
 | `table` | URL | **Must be set server-side.** Letting clients specify the table allows access to any table. |
-| `columns` | URL | **Should be set server-side.** Clients could request sensitive columns. |
+| `queryable_columns` | URL | **Should be set server-side.** This is the column allow-list for WHERE clauses, subset filters, ordering, and synced projections. |
```
robacourt: put `columns` first

website/docs/sync/guides/shapes.md
```diff
@@ -237,7 +238,7 @@ If you need an operator that isn't supported yet, please [raise a feature reques
 
 ### Columns
 
-This is an optional list of columns to select. When specified, only the columns listed are synced. When not specified all columns are synced.
+This is an optional list of columns to sync to the client. It is a projection setting for reducing the amount of data sent over the wire. When specified, only the listed columns are synced. When not specified, all columns are synced, unless [`queryable_columns`](#queryable-columns) is set, in which case the synced columns default to the queryable columns.
```
robacourt: The original wording was clearer

website/docs/sync/guides/auth.md
```diff
@@ -333,7 +334,8 @@ The proxy must set these **shape definition parameters** server-side — they de
 | Parameter | Where | Security Consideration |
 |-----------|-------|------------------------|
 | `table` | URL | **Must be set server-side.** Letting clients specify the table allows access to any table. |
-| `columns` | URL | **Should be set server-side.** Clients could request sensitive columns. |
+| `columns` | URL | Optional sync projection. If clients can choose it, use `queryable_columns` to restrict which columns they can sync. |
```
robacourt: Maybe remove `columns` from this table as it's not a security consideration

website/docs/sync/guides/auth.md
```diff
@@ -333,7 +334,8 @@ The proxy must set these **shape definition parameters** server-side — they de
...
+| `queryable_columns` | URL | **Should be set server-side.** Column allow-list for WHERE clauses, subset filters, ordering, and synced projections. |
```
robacourt:
```suggestion
| `queryable_columns` | URL | **Must be set server-side if the table has sensitive columns.** Column allow-list for WHERE clauses, subset filters, ordering, and synced projections. |
```

packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -1061,6 +1061,95 @@ defmodule Electric.Plug.RouterTest do
       assert [%{"key" => ^key, "value" => ^value}, _] = Jason.decode!(conn.resp_body)
     end
 
+    @tag with_sql: [
+           "CREATE TABLE queryable_users (id BIGINT PRIMARY KEY, name TEXT NOT NULL, secret_token TEXT NOT NULL)",
+           "INSERT INTO queryable_users VALUES (1, 'Alice', 'supersecret123')"
+         ]
+    test "GET allows where clauses on queryable columns that are not synced", %{opts: opts} do
```
alco: What does "are not synced" means here?

packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -1061,6 +1061,95 @@ defmodule Electric.Plug.RouterTest do
...
+               },
+               %{"headers" => %{"control" => "snapshot-end"}}
+             ] = Jason.decode!(conn.resp_body)
+    end
+
+    @tag with_sql: [
+           "CREATE TABLE queryable_users (id BIGINT PRIMARY KEY, name TEXT NOT NULL, secret_token TEXT NOT NULL)",
+           "INSERT INTO queryable_users VALUES (1, 'Alice', 'supersecret123')"
+         ]
+    test "GET rejects where clauses and synced columns outside queryable_columns", %{opts: opts} do
```
alco: Maybe `s/synced columns/selected columns/`?

packages/sync-service/test/electric/plug/router_test.exs
```diff
@@ -1061,6 +1061,95 @@ defmodule Electric.Plug.RouterTest do
...
+          queryable_columns: "id,name",
+          columns: "id,name,secret_token"
+        })
+        |> Router.call(opts)
+
+      assert %{status: 400} = conn
+
+      assert %{
+               "errors" => %{
+                 "columns" => ["The following columns are not queryable: secret_token"]
```
alco: Shouldn't the error be the same as when the column doesn't exist on the table? This different error message allows one to know if the columns is actually explicitly excluded from the queryable set.

alco (approved):
👍

packages/sync-service/test/electric/plug/router_test.exs
robacourt: good idea. thanks

packages/sync-service/test/electric/plug/router_test.exs
robacourt: Good spot. Done

alco:
Took a focused pass looking for missed corner-cases in the new allow-list boundary. Two issues worth addressing, plus a docs nit.

### 1. Same-table subqueries bypass `queryable_columns` (when `allow_subqueries` is enabled)

The top-level WHERE refs are correctly filtered to the allow-list:

```elixir
refs = column_info |> filter_columns(queryable_columns) |> Inspector.columns_to_expr()
```

…but subqueries are handled separately by `build_shape_dependencies/2`, which drops the allow-list:

```elixir
shared_opts = Map.drop(opts, [:where, :columns, :queryable_columns, :relation])
#                                                  ^^^^^^^^^^^^^^^^^ dropped
```

Each sublink is rebuilt as its own dependency shape via `new()` with `queryable_columns: nil`, i.e. **no restriction**, and there's no guard that the subquery's `FROM` table differs from the root table. So a self-referential sublink re-opens the exact boolean-oracle the advisory describes:

```
where = id IN (SELECT id FROM users WHERE secret_token LIKE 'super%')
```

The outer `id` resolves against the filtered refs (fine), but the inner `secret_token` predicate is validated against the root table's **full** column set. Row inclusion in the outer result then leaks whether `secret_token` matches — even though it's neither queryable nor synced — which allows character-by-character enumeration again.

This is gated behind the `allow_subqueries` feature flag (not the default), so it's not a default-config hole — but since this PR *is* the security boundary, I think the subquery path deserves an explicit decision rather than silently inheriting the unrestricted column set.

Fix options:
- Propagate `queryable_columns` into `shared_opts` for dependency shapes whose relation equals the root table (a sibling table's allow-list is effectively "all columns", so passing it through only constrains same-table subqueries); or
- Reject sublinks targeting the root table when `queryable_columns` is set; or
- At minimum, document the limitation and add a test pinning the chosen behaviour.

### 2. `queryable_columns: []` empty-check is dead code

In `validate_queryable_columns/3` the `cond` tests `missing_pk_cols` first:

```elixir
missing_pk_cols = pk_cols -- queryable_columns   # for [] this is the full PK list
...
queryable_columns == [] -> "The list of queryable columns must not be empty"
```

For any table with a primary key, `[]` trips the `missing_pk_cols` branch first, so the user gets *"must include all primary key columns"* and the dedicated empty-list message is never reachable. Not a security problem — just a misleading error and an unreachable branch. Either move the empty check first or drop it.

### Docs nit

Worth an explicit line that a proxy's own server-side `where` must only reference columns it also lists in `queryable_columns`, otherwise its legitimate auth filter gets rejected. The `shapes.md` / `auth.md` examples already do the right thing (`org_id` is in the list), but calling it out would save a footgun.

### Suggested test for issue 1

General shape — a router test mirroring the existing `queryable_users` cases, but with the subquery feature flag enabled and a sublink on the same table referencing a non-queryable column:

```elixir
@tag with_sql: [
       "CREATE TABLE queryable_users (id BIGINT PRIMARY KEY, name TEXT NOT NULL, secret_token TEXT NOT NULL)",
       "INSERT INTO queryable_users VALUES (1, 'Alice', 'supersecret123')",
       "INSERT INTO queryable_users VALUES (2, 'Bob', 'public')"
     ]
test "rejects subqueries referencing non-queryable columns on the root table", %{opts: opts} do
  conn =
    conn("GET", "/v1/shape", %{
      table: "queryable_users",
      offset: "-1",
      queryable_columns: "id,name",
      columns: "id,name",
      where: "id IN (SELECT id FROM queryable_users WHERE secret_token LIKE 'super%')"
    })
    # enable the allow_subqueries feature flag for this request
    |> Router.call(opts)

  assert %{status: 400} = conn
  assert %{"errors" => %{"where" => [message]}} = Jason.decode!(conn.resp_body)
  assert message =~ "unknown reference secret_token"
end
```

As written today this would return `200` and leak the oracle rather than `400`, so it doubles as the failing repro.

PR #4533, opened 2026-06-08
feat(agents-mobile): native slash-command composer for the Horton prompt

msfstef:
## What this is

This brings the **mobile (Expo / React Native) Horton prompt** up to feature/experience parity with the **desktop ProseMirror composer**: native slash-command autocomplete and structured `composer_input` payloads — on **both** the in-session and new-session inputs.

It's built as a **native composer** (a real RN `TextInput`), deliberately **not** a WebView embedding the desktop ProseMirror editor — `prosemirror-view` needs a real DOM and can't run under `react-native-web`, so the ProseMirror path would *require* a WebView, not merely allow one. The design, alternatives, and trade-offs are documented below.

> 🤖 Developed with Claude (hence the label). Manually verified on iOS + Android simulators; typecheck, lint, and the unit suites are green (45 mobile tests + the runtime composer-input suite).

---

## Why native, not a WebView

The desktop composer is ProseMirror, which is DOM-only. We could have embedded it in a WebView / Expo DOM component, but the only reuse that actually matters — the `composer_input` wire contract and the slash-command grammar — is **pure logic we can share losslessly**. A WebView typing surface would import WKWebView caret/keyboard/IME friction and force an awkward native-popover-over-WebView bridge, all for editor code we don't need. So: **native input, shared logic.**

---

## What's in here

**1. Shared the slash grammar (no desktop behaviour change)**
Lifted the serializer + slash-command grammar out of the desktop-only `ComposerEditor` into `agents-runtime/src/composer-input.ts` (exported via `/client`), so desktop and mobile derive nodes / highlighting / autocomplete from one source of truth. The desktop call sites just **repoint their imports** — behaviour-neutral for desktop — and the serializer tests moved into the runtime suite.

**2. Native in-session composer**
`NativeComposer.tsx` drives slash autocomplete on the existing `SessionScreen` `TextInput`: trigger detection from the shared grammar (no caret coordinates), a native keyboard-docked suggestion menu, and sending via `createSendComposerInputAction`. Discovery uses the built-in `db.collections.slashCommands` the screen already holds, falling back to the entity type's static `slash_commands` (static commands aren't materialised into the per-entity collection — the same fallback desktop uses).

**3. New-session spawn parity + forward-compatible plumbing**
The same autocomplete on `NewSessionScreen`, spawning with a `composer_input` initial message. **No server change** — the spawn endpoint already accepts a structured `initialMessage` + `initialMessageType` + an `args` record; we only widened the mobile `spawnEntity` wrapper and **generalised its `args` channel** so future schema-driven spawn args / model settings / attachments slot in additively rather than needing a re-plumb.

**4. Inline command "badges"**
Recognised `/command` tokens (and their declared argument words) are highlighted inline — the command in bold accent, the argument in a quieter accent, sharing a faint tint — mirroring the desktop badge's command + argument slot.

**5. Platform fixes found while testing on device**
- **iOS auto-grow:** switching the input to child `<Text>` (needed for inline highlighting) breaks iOS's `onContentSizeChange` ([RN #13732](https://github.com/facebook/react-native/issues/13732)), so the input now sizes to content intrinsically.
- **Android keyboard:** Android's default `resize` mode already lifts the bottom-anchored composer above the keyboard; the manual `translateY` double-offset it (sending it toward the top of the screen) — now applied on iOS only.

---

## Key decisions & trade-offs

- **One shared grammar, pinned by a test.** Mobile (regex over the source string) and desktop (walking the ProseMirror doc) can diverge at the edges — e.g. the grammar is intentionally lowercase-only — so a runtime test pins that boundary to stop the two silently drifting.
- **The popover is native and docked above the keyboard, not caret-anchored.** RN `TextInput` exposes selection *indices* but not caret x/y (there's no `coordsAtPos` analog), and an above-keyboard list is the Slack/Discord/iMessage pattern anyway.
- **Badges are a flat tint + font weight, not a rounded chip.** The shipped "pills" differentiate command vs. argument with an inline text background plus weight/colour (bold accent command, quieter accent argument). An editable RN `TextInput` can't pad or round an inline text background ([RN #10807](https://github.com/facebook/react-native/issues/10807)), and `react-native-live-markdown` can't either (same text rendering) — a true rounded chip needs a native rich-text widget or a WebView, both ruled out. We evaluated the `react-native-live-markdown` route, found it wouldn't deliver real chips, and stayed dependency-free. This is the platform ceiling and the final approach here, not a stopgap.
- **Editing a queued message flattens it to `{ text }`.** This is desktop's *existing* behaviour, not a mobile regression.

---

## Known limitations (consciously accepted)

- **Typing a multi-line message flashes the conversation behind the composer.** As the composer grows a line, the bottom inset changes and the *virtualised* conversation timeline re-pads and reflows — that reflow is the flash. We chased the native-side prop churn and the CSS (mask) causes and reverted all of it once they didn't pan out; the only genuine fix is restructuring how the native composer and the WebView-embedded conversation share space, which isn't worth it for a brief multi-line-typing transient.
- **Inline highlighting is a flat tint, not a chip** (see decisions above).
- **Future spawn features** (schema-form args, model/reasoning/speed, attachments) are **not** in this PR — but the `spawnEntity` / `args` plumbing is shaped so they're additive.

---

## How to review

- **Start with *Why native, not a WebView* and *Key decisions & trade-offs* above** for the design rationale and decision log.
- **Shared extraction:** `agents-runtime/src/composer-input.ts` + the desktop repoints in `agents-server-ui` (import changes only — desktop behaviour is unchanged).
- **Mobile:** `NativeComposer.tsx`, `lib/slashAutocomplete.ts` (pure + unit-tested), `SessionScreen.tsx`, `NewSessionScreen.tsx`, `lib/agentsClient.ts`.

**Try it:** open a Horton session, type `/` → a native menu of its commands appears; pick one → it inserts and highlights; send → Horton runs the command. Same flow on the new-session prompt.

---

Opened as a **draft** for early eyes — happy to walk through any of the decisions.


msfstef:
<table>
<tr>
<td>Simulator Screenshot - iPhone 15 Pro - 2026-06-08 at 19 26 38</td>
<td>Simulator Screenshot - iPhone 15 Pro - 2026-06-08 at 19 27 13</td>
</tr>
</table>

packages/agents-mobile/src/screens/SessionScreen.tsx
```diff
@@ -473,12 +482,44 @@ function NativeMessageComposer({
...
+  // commands — statically-declared ones are never materialised into it. So, like
+  // the desktop composer (ChatView -> MessageInput), fall back to the entity
+  // type's declarations when the live collection is empty.
+  const { data: liveSlashCommands = [] } = useLiveQuery(
+    (q) =>
+      db
+        ? q
+            .from({ slashCommand: db.collections.slashCommands as any })
+            .orderBy(({ slashCommand }: any) => slashCommand.name, `asc`)
+        : undefined,
```
kevin-dp, line 497: Is this valid?
The docs type `useLiveQuery`'s `queryFn` argument as `(q) => QueryBuilder<TContext>`. So it should return a `QueryBuilder` but this may return `undefined`.

packages/agents-mobile/src/screens/SessionScreen.tsx
```diff
@@ -473,12 +482,44 @@ function NativeMessageComposer({
...
+    (q) =>
+      entity
+        ? q
+            .from({ type: entityTypesCollection })
+            .where(({ type }) => eq(type.name, entity.type))
+        : undefined,
+    [entityTypesCollection, entity?.type]
+  )
+  const slashCommands = useMemo<Array<SlashCommandRow>>(() => {
+    if (liveSlashCommands.length > 0) {
```
kevin-dp, line 510: This if test isn't needed. We can remove it and just return `(matchingTypes[0]?.slash_commands ?? []).map((command) => ({ ... })`.

kevin-dp (approved):
Good work 👍

packages/agents-mobile/src/screens/SessionScreen.tsx
msfstef: Apparently the docs don't mention the overload for returning null or undefined

https://github.com/TanStack/db/blob/ca8aa0d4772d0e3f7b90f066071767e67bbe89f7/packages/react-db/src/useLiveQuery.ts#L100-L117

Quoting from agent:
> Returning undefined/null is the library's supported "query not ready yet" signal — data comes back undefined, which the code coalesces to [] via data: liveSlashCommands = []. The desktop codebase already relies on this in two places: MessageInput.tsx:118 and ChatView.tsx:209.

packages/agents-mobile/src/screens/SessionScreen.tsx
msfstef: This is checking if `liveSlashCommands` is available (which is dynamically registered slash commands) and returns those otherwise it falls back to the statically declared slash commands - this is mirroring the desktop UI:
https://github.com/electric-sql/electric/blob/55c063a3034f09cc881b83b499ef4af28269c157/packages/agents-server-ui/src/components/MessageInput.tsx#L111-L123

PR #4534, opened 2026-06-08
(sync-service) queryable_columns to only restrict columns changeable by the client

robacourt:
PR #4531 made `queryable_columns` apply to normal shape `where` clauses, but as https://github.com/electric-sql/electric/pull/4531#issuecomment-4650457200 pointed out, sensitive columns could still be accessed via subqueries. Making arbitrary `where` clauses with subqueries secure would require more than a root-table column allow-list: we would also need a table allow-list plus per-table `queryable_columns` for every table referenced by a subquery.

However, normal where clauses are supposed to be restricted and only set server-side. There's no need to restrict them at all. The are to restrict is the subset snapshots, where clients can change subset `where` and `order_by`. 

This PR makes the behavior consistent by allowing any server-defined main `where` clause while keeping `queryable_columns` enforcement for subset `where` and subset `order_by`. It also updates the docs to make this distinction clear too.

## Summary
- allow main shape `where` clauses to reference any table column even when `queryable_columns` is set
- keep `queryable_columns` enforcement for synced projections, subset `where`, and subset `order_by`
- update docs, OpenAPI, and the existing changeset to describe the narrower boundary

### DIFF
--- a/.changeset/queryable-columns.md
+++ b/.changeset/queryable-columns.md
@@ -2,4 +2,4 @@
 "@core/sync-service": patch
 ---
 
-Add `queryable_columns` as a server-side shape allow-list for columns that may be queried by subset snapshots or synced. This lets proxies decouple the `columns` sync projection from the subset-query security boundary while preventing subset filters and ordering, and projected columns, from referencing non-queryable columns.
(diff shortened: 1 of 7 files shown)


alco (approved):
👍

PR #4537, opened 2026-06-09
fix(sync-service): Stop the publication manager from getting flooded with duplicate casts

### LINKED ISSUE #4396: RelationTracker: cast-issuance dedup gate is too permissive; PublicationManager.Configurator mailbox grows unboundedly under shape-arrival bursts
## Summary

`Electric.Replication.PublicationManager.RelationTracker.update_needed?/1` fires a fresh `Configurator.configure_publication` cast on every `add_shape` / `remove_shape` call as long as `committed_relation_filters` hasn't caught up to `submitted_relation_filters` — even when the in-flight configuration chain is already going to commit exactly the relations being asked for. Under a burst of shape arrivals, the `publication_manager_configurator` mailbox grows to hundreds of thousands of entries carrying identical filter sets, even though the actual SQL work is bounded.

## Observation

During the 2026-04-28 Ohm production deployment (Stratovolt source `svc-free-armadillo-64bp5xmfga`, sole tenant on the `ohm` instance), the new instance's BEAM message-queue snapshot showed `publication_manager_configurator` at **705,925** queued messages — #3 by mailbox depth, behind only `shape_cache` (4.5M) and `RequestProxy` (1.15M). The configurator stayed backlogged for the full ~51-minute recovery window (13:49 → 14:40 UTC).

Honeycomb breakdown: https://ui.honeycomb.io/electric-sql-06/environments/production/datasets/electric-region/result/9EezctBcZLk

A pair of spikes on 2026-05-21 deployments shows the same shape.

This source had **far fewer than 100 distinct relations**, so the 705k figure cannot be explained by per-relation work. It is the cast-issuance rate that is wrong.

## Root cause

`relation_tracker.ex:345`:

```elixir
defp update_needed?(%__MODULE__{
       prepared_relation_filters: prepared,
       submitted_relation_filters: submitted,
       committed_relation_filters: committed
     }) do
  not MapSet.equal?(prepared, submitted) or not MapSet.equal?(submitted, committed)
end
```

The second disjunct (`submitted != committed`) returns `true` for the entire window between "Configurator cast was sent" and "every relation in that submission has been confirmed via `:relation_configuration_result {:ok, :configured}`". During that window, **every** `add_shape` / `remove_shape` call that lands in RelationTracker — and every internal recompute — calls `update_publication_if_necessary`, which now fires a fresh cast carrying the **same** `prepared_relation_filters` MapSet that the Configurator is already working on.

The Configurator's own dedup layers downstream do work as designed:

- `handle_cast({:update_publication, …})` (`configurator.ex:86–93`) coalesces by overwriting `scheduled_filters`. Only one debounce timer is armed per window, so at most one chain runs per debounce period.
- `determine_publication_relation_actions/2` (called from `handle_continue({:configure_filters, filters}, state)`) diffs PG's actual publication state against the target filter set. If PG already has the relations, `to_add` is empty and **no SQL runs**.

So the SQL is bounded by unique relation transitions, as expected for a source with <100 relations. **What is unbounded is the cast count itself**, because:

1. The receive loop is suspended during the chain (handle_continue chains run to completion before re-entering receive). Casts accumulate as raw `{:update_publication, full_filter_set}` tuples in the mailbox.
2. Each cast allocates a fresh `MapSet` via `expand_oids/2` on the RelationTracker side.
3. Layer 2 coalescing only kicks in once the receive loop unsuspends — at which point N casts collapse to one `scheduled_filters` slot, but N copies of the MapSet have already been allocated and queued.

Under the Ohm recovery scenario, RelationTracker was driven by ~700k shape arrivals while the Configurator was stuck waiting on table locks against an exhausted admin pool. Each of those shape arrivals minted a cast that was about to be thrown away.

## Primary fix

**Track an `in_flight_relation_filters` snapshot on the RelationTracker side and suppress cast issuance while it's set and matches `prepared`.**

```elixir
defstruct [
  …,
  in_flight_relation_filters: nil,   # MapSet | nil
  …
]
```

- `update_publication/1`: before issuing the cast, set `in_flight = prepared`.
- `handle_cast({:relation_configuration_result, …, {:ok, _}})` and `handle_cast({:configuration_error, …})`: when `committed == in_flight` (or on global error), clear `in_flight = nil`.
- `update_needed?/1`:

```elixir
defp update_needed?(%__MODULE__{
       prepared: prepared,
       submitted: submitted,
       committed: committed,
       in_flight: in_flight
     }) do
  prepared_changed = not MapSet.equal?(prepared, submitted)
  retry_needed     = in_flight == nil and not MapSet.equal?(submitted, committed)
  prepared_changed or retry_needed
end
```

The committed-lag check stays — it's the legitimate "previous attempt didn't fully commit, retry" signal — but it is **gated on `in_flight == nil`**, so it can no longer fire while a chain is already going to commit the relations in question. The existing 60s `publication_refresh_period` timer (`relation_tracker.ex:307–317`) is the fallback retry if a chain crashes without clearing `in_flight` cleanly.

Properties after the fix:

- 700k concurrent `add_shape` calls produce **exactly one** Configurator cast as long as `prepared_relation_filters` doesn't change.
- If `prepared` *does* change mid-chain (a new relation appears in a new shape), exactly one fresh cast is fired; subsequent calls observing the same new `prepared` produce no further casts until the chain completes.
- Layer 2 coalescing and Layer 3 per-relation diffing remain unchanged and continue to do their jobs.

Estimated patch size: ~15 lines in RelationTracker plus tests.

## Secondary hardening

These don't address the cast-rate explosion but are independently good defenses against related failure modes. Recommended to land after the primary fix, separately.

1. **Yield between `perform_relation_actions` chunks.** Replace the per-relation `{:continue, …}` tail-recursion in `configurator.ex:216` with either a self-message after each chunk of K relations or a batched `handle_info` loop, so the receive loop drains between chunks. Doesn't change behaviour after the primary fix, but caps mailbox depth at the size of one chunk's worth of arrivals even if some other future caller pattern produces excess casts.

2. **Explicit per-action DB timeout.** `do_publication_update/3` calls `Configuration.*` which run through `Postgrex.transaction` with no `timeout:`, inheriting the 15s default. Under admin-pool exhaustion this is the wrong knob — a per-action ceiling of a few seconds bounds the per-failure cost without giving up correctness.

3. **Refresh `scheduled_filters` mid-chain.** After processing K relations, before continuing, re-read `state.scheduled_filters` and recompute `determine_publication_relation_actions/2` against the new target if it has changed. Avoids reapplying stale ADD/DROP decisions when the desired set has moved.

## Related

- #4266 — ShapeCache bottlenecks under thundering herd (umbrella for this family)
- #4370 — EtsInspector: prevent mailbox overload and avoid serving orphaned replies
- #4371 — StatusMonitor: replace mailbox-based `wait_until` with adaptive per-process polling
- #4372 — ShapeCache: coalesce concurrent `create_or_wait` callers
- #1785 (closed) — Slow to respond concurrent requests for new shapes

The above are about **callers parking in the mailbox of a single GenServer**. This issue is the inverse pattern: **fire-and-forget casts piling up because the producer's dedup gate is too permissive**. The fixes are not interchangeable.

erik-the-implementer:
## Summary

Fixes #4396.

Adding or removing a shape can make `RelationTracker` ask the `Configurator` to reconfigure the Postgres publication. The bug: while one such reconfiguration was still in progress, *every* further `add_shape`/`remove_shape` fired off another, redundant request. This happened because the check that decides whether a new request is needed (`update_needed?/1`) treated "request sent but not yet fully committed" as "still needs work", so it kept saying yes for the whole duration of the in-flight request. The Configurator already merges these requests and skips redundant SQL, so the *database* work stayed bounded — but the redundant request *messages* piled up in its mailbox. During the 2026-04-28 Ohm deploy this reached **705,925** queued messages.

The fix gives `RelationTracker` a memory of the request it currently has outstanding (`in_flight_relation_filters`):

- When it sends a reconfigure request, it records which relations that request covers.
- While a request is outstanding, it no longer sends a duplicate just because the previous one hasn't finished committing yet. So a burst of `add_shape`/`remove_shape` calls produces at most one request. (A genuinely new relation is still a real change, so it still triggers exactly one fresh request.)
- It forgets the outstanding request as soon as the Configurator reports back — whether that's a successful commit, a per-relation error, or a global error — which frees it to send the next one.
- If a request silently dies without ever reporting back, a 60s timer still kicks in and retries, so nothing gets stuck forever.

### Note on forgetting the request after a per-relation error

The issue suggested forgetting the outstanding request only on a successful commit or a global error. This PR also forgets it when a *single* relation fails, because the existing behaviour ("keep retrying a relation that keeps failing") depends on it.

This is safe and does **not** bring back the message flood. We only forget the request once the Configurator has actually reported a result back — and getting a result, even an error, means the request ran and made progress. In the production incident there were no results at all (the work was stuck waiting on database locks), so the request would have correctly stayed remembered the whole time. Either way, the number of requests sent is capped by how fast results come back plus the 60s timer — never by how long a single request takes to finish.

### DIFF
--- a/.changeset/suppress-publication-config-casts.md
+++ b/.changeset/suppress-publication-config-casts.md
@@ -1,0 +1,5 @@
+---
+'@core/sync-service': patch
+---
+
+Suppress redundant publication-configuration casts while a submission is already in flight. Under a burst of shape arrivals the publication manager's RelationTracker no longer mints one Configurator cast per add_shape/remove_shape, preventing the `publication_manager_configurator` mailbox from growing unboundedly (issue #4396).
--- a/.changeset/suppress-publication-config-casts.md
+++ b/.changeset/suppress-publication-config-casts.md
@@ -2,4 +2,4 @@
 '@core/sync-service': patch
 ---
 
-Suppress redundant publication-configuration casts while a submission is already in flight. Under a burst of shape arrivals the publication manager's RelationTracker no longer mints one Configurator cast per add_shape/remove_shape, preventing the `publication_manager_configurator` mailbox from growing unboundedly (issue #4396).
+Avoid sending duplicate publication-configuration requests while one is already in progress. Under a burst of shape arrivals, the publication manager no longer issues a separate request for every shape added or removed, preventing the configurator's message queue from growing without bound (issue #4396).
--- a/packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
+++ b/packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
@@ -307,11 +307,12 @@ defmodule Electric.Replication.PublicationManager.RelationTracker do
       )
     end
 
-    {:noreply, state, state.publication_refresh_period}
+    {:noreply, %{state | in_flight_relation_filters: nil}, state.publication_refresh_period}
   end
 
   def handle_cast({:configuration_error, {:error, error}}, state) do
-    {:noreply, reply_to_all_waiters({:error, error}, state), state.publication_refresh_period}
+    state = %{reply_to_all_waiters({:error, error}, state) | in_flight_relation_filters: nil}
+    {:noreply, state, state.publication_refresh_period}
   end
 
   @impl true
--- a/packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
+++ b/packages/sync-service/lib/electric/replication/publication_manager/relation_tracker.ex
@@ -29,11 +29,12 @@ defmodule Electric.Replication.PublicationManager.RelationTracker do
     prepared_relation_filters: MapSet.new(),
     submitted_relation_filters: MapSet.new(),
     committed_relation_filters: MapSet.new(),
-    # Snapshot of the filter set carried by the cast currently being processed
-    # by the Configurator. While set, the "committed lags submitted" retry in
-    # update_needed?/1 is suppressed so a burst of add_shape/remove_shape calls
-    # cannot mint one redundant cast each. Cleared on a terminal signal from the
-    # Configurator (full commit, per-relation error, or global error).
+    # The relations covered by the configuration request the Configurator is
+    # currently working on, or nil when none is outstanding. While set, the
+    # "committed lags submitted, retry" branch in update_needed?/1 is held off,
+    # so a burst of add_shape/remove_shape calls can't each send another
+    # duplicate request. Reset to nil once the Configurator reports back (a full
+    # commit, a per-relation error, or a global error).
     in_flight_relation_filters: nil,
     waiters: %{},
     # start with optimistic assumption about what the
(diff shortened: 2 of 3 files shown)


magnetised (approved):
Nice

PR #4538, opened 2026-06-09
Custom collections on entity streams + session comments

balegas:
## Summary

Adds a generic mechanism for custom collections on entity streams and wires session comments on top of it. Comments are an opt-in feature on `horton` and `worker` entity types; other entity types stay comment-free. Replaces the bespoke `POST /comments` route + built-in `comments` runtime collection from #4529 with a layered design that's reusable for any other custom collection (e.g. samwillis's planned yjs doc comments).

## How it works

**Declaration.** Entity types declare the custom collections they accept via `custom_collection_schemas: { <name>: <JSON Schema> }`. `EntityDefinition.customCollectionSchemas` carries this through the typed registration API; the runtime forwards it via `create-handler.ts` to the server, where it's persisted on `entity_types` (new JSONB column, migration 0015) and snapshotted onto each entity at spawn time. `horton` and `worker` declare a `comment` schema; the schema itself lives in `packages/agents/src/agents/comment-schema.ts` as a Zod definition.

**Write path.** A single generic endpoint, `POST /_electric/entities/:type/:id/collections/:name`, accepts opaque JSON, looks up the schema declared for `<name>` on the entity's type via `getEffectiveSchemas`, and rejects writes (422) when no schema is declared or the value doesn't validate. Reserved built-in collection names (`BUILT_IN_COLLECTION_TYPES`, exported from `@electric-ax/agents-runtime`) are rejected too so the runtime stays the sole writer of agent-managed collections. `amendSchemas` supports additive amendments of custom collection schemas via a shared `mergeAdditiveSchemaMap` helper that also DRYs the existing inbox/state merges.

**Read path.** `createEntityTimelineQuery` takes an optional `customSource` query-builder branch shaped to `EntityTimelineCustomRow` (`{ collection, key, order, value }`) and splices it into the same unionAll/orderBy pipeline as the built-in timeline collections. Consumers get one ordered live query instead of merging a second source client-side — no two-source ordering race. Multi-collection callers can union their sources client-side and use the `collection` field as a discriminator.

**UI.** `useEntityTimeline` registers a `comments` collection in the StreamDB customState (with a passthrough Standard Schema for the runtime row type), composes a `customSource` for it, hands it to `createEntityTimelineQuery`, and reshapes envelope rows into a UI-friendly `comment` variant via a small `projectRow` that spreads `row.custom.value` and overrides `key`/`order`. Comment surface — bubbles, reply affordances on user messages, agent responses, tool calls, and inline events, the comments-only tile view, and the reply composer — is all gated on whether the entity's type opted in to a `comment` schema. `ElectricEntity` / `ElectricEntityType` carry `custom_collection_schemas` through their zod schemas, and the `entity_types` Electric shape's column allow-list (`server-utils.ts`) was updated to expose the new column to UI clients.

## Notable design points

- **`EntityDefinition.customCollectionSchemas`** is typed as `Record<string, StandardSchemaV1 | Readonly<Record<string, unknown>>>` rather than the strict `Record<string, StandardJSONSchemaV1>` used by `inboxSchemas`/`stateSchemas`. The runtime's `toJsonSchema()` accepts Standard Schemas (Zod), raw JSON Schema objects, and anything with a `~standard.jsonSchema` converter — so the looser union is honest about what works in practice. The strict variant has no callers in the repo (the conformance tests send raw JSON Schema over the wire, going around the typed API).
- **The envelope shape** (`{ collection, key, order, value }`) keeps the caller's `.select(...)` to a single `value: row` line — adding a field to `CommentRow` later only needs to land in `projectRow`, not in two places. The runtime guarantees `key` and `order` (the minimum it needs for the union and orderBy); the rest rides through `value`.

## Test plan

- [x] `pnpm --filter @electric-ax/agents-runtime run typecheck && run test` — 57 timeline tests incl. 2 new custom-source tests
- [x] `pnpm --filter @electric-ax/agents-server run typecheck && run test` — 55 targeted tests incl. write-validation + route delegation (happy path, schema mismatch, undeclared collection, reserved-name rejection, invalid name, non-object value, stopped-entity, route-layer delegation)
- [x] `pnpm --filter @electric-ax/agents-server-ui run typecheck && run test` — 93 tests incl. 3 new comments-action tests, 3 ChatView tests, 1 InlineEventCard test
- [x] `pnpm --filter @electric-ax/agents run typecheck`
- [x] Manual smoke on a fresh stack: spawned a horton session, verified inline reply icons appear next to copy/edit on user messages, agent responses, tool calls, and inline events; sent a comment via the reply composer; reply preview scrolls to the target row when clicked; comments-only tile view shows only comment rows; on entity types without `custom_collection_schemas.comment` declared, the reply icons and comments-only view are correctly hidden.

## Notes

- Migration `0015_entity_custom_collection_schemas.sql` adds a nullable JSONB column on both `entity_types` and `entities` — additive, no backfill needed.
- The `entity_types` shape's column allow-list at `packages/agents-server/src/utils/server-utils.ts:138` was updated to include `custom_collection_schemas` — without this the UI gate would silently fail closed.
- No public API was removed. The runtime exports `BUILT_IN_COLLECTION_TYPES` and the `EntityTimelineCustomRow` / `EntityTimelineCustomSource` types as new public surface.


packages/agents-runtime/src/create-handler.ts
```diff
@@ -503,6 +503,11 @@ export function createRuntimeRouter(
         ...(definition.inboxSchemas && {
           inbox_schemas: mapSchemas(definition.inboxSchemas),
         }),
+        ...(definition.customCollectionSchemas && {
```
KyleAMathews, line 506: you can already define custom schemas so we don't need this

PR #4539, opened 2026-06-09
feat(consumer): bound Shape.Consumer heap growth via spawn opts + adaptive GC

### LINKED ISSUE #4476: Shape consumers accumulate multi-MB heaps of unreclaimed floating garbage (default fullsweep_after, rarely hibernate)
## Summary

On a busy multi-tenant production node, `Electric.Shapes.Consumer` processes accumulate **~7–9 MB resident heaps each, while their actual live state is ~8 KB**. The heap is floating garbage on the old generation that is never reclaimed, because (a) consumers rarely hibernate and (b) they spawn with the BEAM default `fullsweep_after` (`65535`), so generational GC never performs a shrinking full sweep. With several thousand such processes per node this reached **~36 GB** of unreclaimable heap on a single node. A manual `:erlang.garbage_collect/1` on any consumer drops its heap ~90%, confirming the retained memory is unreachable garbage rather than live state.

## Evidence (single production node)

- **5,043** consumers; **4,455** with heap `> 4 MB`; aggregate **~36 GB**.
- `hibernating` = **0 / 5043** (none in `:gen_server.loop_hibernate` / `:erlang.hibernate`).
- Representative per-consumer snapshot (`:sys.get_state/1` + `Process.info/2`):
  - `total_heap_size` ≈ **6.7–9.2 MB**
  - `:erts_debug.size/1` of the **entire** GenServer state ≈ **8 KB** (shape ~3 KB, writer ~1 KB; `buffer`, `txn_offset_mapping`, `transaction_builder` all empty; `pending_txn: nil`)
  - `hibernate_after` = **600000 ms** (10 min)
  - `garbage_collection[:fullsweep_after]` = **65535** (BEAM default)
  - `garbage_collection[:minor_gcs]` ≈ **40–50** (≪ 65535 ⇒ a full sweep has never run)
- Manual `:erlang.garbage_collect(pid)` reclaims ~90% (heap → ~0.2 MB).

The shapes involved are correctly indexed (equality on a high-cardinality column) and process only a few dozen tiny transactions per day each, so this is **not** driven by transaction size, throughput, or unindexed WHERE evaluation — it is purely a GC-policy issue against a large fleet of long-lived, mostly-idle processes.

## Root cause

`fullsweep_after` defaults to `65535`: the old heap is only fully collected on a full sweep, which won't happen until `fullsweep_after` minor GCs **or** a hibernation. Consumers:

1. **rarely hibernate** — `shape_hibernate_after` is large (10 min in this deployment) and any received message re-arms the GenServer inactivity timeout, so active-ish consumers seldom idle that long; and
2. **never approach 65535 minor GCs**.

So promoted floating garbage accumulates on the old heap indefinitely. This is the same class of problem already addressed for Bandit handler processes via `handler_fullsweep_after` (whose doc comment in `config.ex` describes exactly this old-heap accumulation), but `Electric.Shapes.Consumer` was never wired into a `fullsweep_after` spawn option — `Consumer.start_link/1` is a plain `GenServer.start_link/3` with no `:spawn_opt`.

## Proposed fix

Spawn consumers with a tunable `fullsweep_after`, mirroring `ShapeLogCollector`, which already does `spawn_opt: Electric.StackConfig.spawn_opts(stack_id, :shape_log_collector)`. Thread a `:consumer` entry through the existing `process_spawn_opts` config (or add a dedicated knob analogous to `handler_fullsweep_after`) into `Consumer.start_link/1`:

```elixir
GenServer.start_link(__MODULE__, init_arg,
  name: name(stack_id, shape_handle),
  spawn_opt: Electric.StackConfig.spawn_opts(stack_id, :consumer)
)
```

A modest `fullsweep_after` bounds each consumer's heap to its true working set with negligible extra GC cost (live state is ~8 KB, so full sweeps are cheap). Consider applying the same to `Consumer.Snapshotter` and `Consumer.Materializer`.

Note on tuning: the per-process garbage-vs-minor-GC ratio observed here is high (~8 MB accumulated over only ~40–50 minor GCs), so the effective `fullsweep_after` should be **low** (order of 10–20) to actually bound the heap — values in the hundreds/thousands still let substantial garbage accumulate between sweeps for these low-allocation processes.

## Operational mitigation (no deploy)

Periodically running `:erlang.garbage_collect/1` over all registered consumers reclaims the garbage immediately (heaps slowly refill until the spawn-opt fix ships).

## Affected code

- `packages/sync-service/lib/electric/shapes/consumer.ex` — `start_link/1` (no `:spawn_opt`)
- `packages/sync-service/lib/electric/replication/shape_log_collector.ex` — existing `spawn_opts` pattern to mirror
- `packages/sync-service/lib/electric/config.ex` — `handler_fullsweep_after` / `process_spawn_opts` precedent

erik-the-implementer:
Closes #4476

## Summary

Bounds runaway old-heap growth of `Shape.Consumer` processes (issue #4476). On a production node with 5,043 consumers, ~36 GB was unreclaimable garbage (7–9 MB heaps each holding ~8 KB of live state) because consumers inherit the BEAM default `fullsweep_after: 65535` and rarely hibernate, so a full sweep is effectively never triggered.

Two complementary, independently-tunable levers:

1. **Per-process spawn opts (env-driven).** `Consumer`, `Consumer.Snapshotter`, and `Consumer.Materializer` now pass `spawn_opt: Electric.StackConfig.spawn_opts(stack_id, key)` (keys `:consumer`, `:consumer_snapshotter`, `:consumer_materializer`), mirroring `ShapeLogCollector`. This lets `ELECTRIC_PROCESS_SPAWN_OPTS` set `fullsweep_after` per process type. No baked-in defaults — purely env-configured.

2. **Adaptive GC, opt-in.** After processing a transaction fragment (including during the startup buffer drain), the consumer forces `:erlang.garbage_collect()` only if its heap exceeds `consumer_gc_heap_threshold` (bytes; `nil` = off, the default). The threshold is cached per-consumer at startup (`State.new/2`), so changing it affects consumers started afterwards. The forced sweep runs **off the reply path** via a `{:continue, :maybe_gc}` step, so it never blocks the `ShapeLogCollector`'s synchronous publish call, and a hysteresis interval (`@gc_min_interval_ms`) caps how much CPU a busy consumer spends sweeping. See `Electric.Shapes.Consumer.set_gc_heap_threshold/2`.

Request-handler (`GET /v1/shape`) processes are intentionally untouched — they already expose `fullsweep_after` via `ELECTRIC_TWEAKS_HANDLER_FULLSWEEP_AFTER` and force GC before long-poll blocking.

## Config

| Env var | Meaning | Default |
|---|---|---|
| `ELECTRIC_PROCESS_SPAWN_OPTS` | JSON map; now accepts `consumer`/`consumer_snapshotter`/`consumer_materializer` keys (e.g. `{"consumer":{"fullsweep_after":4}}`) | `%{}` |
| `ELECTRIC_CONSUMER_GC_HEAP_THRESHOLD` | adaptive-GC heap threshold in bytes; `nil` disables | `nil` |

### DIFF
--- a/.changeset/consumer-heap-gc.md
+++ b/.changeset/consumer-heap-gc.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Bound Shape.Consumer heap growth: make the consumer family's process spawn options (incl. `fullsweep_after`) configurable per process via `ELECTRIC_PROCESS_SPAWN_OPTS`, and add an opt-in adaptive GC that runs after a transaction fragment when the consumer's heap exceeds the runtime-tunable `ELECTRIC_CONSUMER_GC_HEAP_THRESHOLD` (off by default).
--- a/packages/sync-service/config/runtime.exs
+++ b/packages/sync-service/config/runtime.exs
@@ -291,6 +291,7 @@ config :electric,
   process_registry_partitions: env!("ELECTRIC_TWEAKS_PROCESS_REGISTRY_PARTITIONS", :integer, nil),
   process_spawn_opts:
     env!("ELECTRIC_PROCESS_SPAWN_OPTS", &Electric.Config.parse_spawn_opts!/1, %{}),
+  consumer_gc_heap_threshold: env!("ELECTRIC_CONSUMER_GC_HEAP_THRESHOLD", :integer, nil),
   http_api_num_acceptors: env!("ELECTRIC_TWEAKS_HTTP_API_NUM_ACCEPTORS", :integer, 100),
   conn_max_requests: env!("ELECTRIC_TWEAKS_CONN_MAX_REQUESTS", :integer, nil),
   handler_fullsweep_after: env!("ELECTRIC_TWEAKS_HANDLER_FULLSWEEP_AFTER", :integer, nil),
--- a/packages/sync-service/lib/electric/application.ex
+++ b/packages/sync-service/lib/electric/application.ex
@@ -156,7 +156,8 @@ defmodule Electric.Application do
         shape_suspend_after: get_env(opts, :shape_suspend_after),
         conn_max_requests: get_env(opts, :conn_max_requests),
         handler_fullsweep_after: get_env(opts, :handler_fullsweep_after),
-        process_spawn_opts: get_env(opts, :process_spawn_opts)
+        process_spawn_opts: get_env(opts, :process_spawn_opts),
+        consumer_gc_heap_threshold: get_env(opts, :consumer_gc_heap_threshold)
       ],
       manual_table_publishing?: get_env(opts, :manual_table_publishing?),
       shape_db_opts: [
--- a/packages/sync-service/lib/electric/config.ex
+++ b/packages/sync-service/lib/electric/config.ex
@@ -115,8 +115,7 @@ defmodule Electric.Config do
     # e.g. %{shape_log_collector: [min_heap_size: 1024 * 1024, min_bin_vheap_size: 1024 * 1024]}
     process_spawn_opts: %{},
     # Heap-size threshold (in BYTES) above which a consumer runs :erlang.garbage_collect()
-    # after processing a transaction fragment. nil disables adaptive GC. Looked up at
-    # runtime via StackConfig so it can be changed from a live IEx shell.
+    # after processing a transaction fragment.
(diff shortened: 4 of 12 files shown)


robacourt (approved):
🚀

PR #4542, opened 2026-06-09
fix(agents): pull-wake runner resilience — fetch cache stacking, hung connections, xstate lifecycle

KyleAMathews:
## Summary

Fixes two bugs that left the desktop agents runtime unable to pick up sessions until a full app restart, then ports the pull-wake runner's lifecycle to an xstate v5 state machine so this class of bug can't recur silently. No public API changes — all 21 pre-existing behavioral tests pass unmodified.

## Root cause

Two independent failures combined into "runner looks alive but never picks up work, and restarting the runtime doesn't help":

1. **Hung connection couldn't be aborted.** The heartbeat-failure → stream-reset path (`requestStreamReconnect`) was a no-op unless `streamConnected` was true. If the wake stream hung while *(re)connecting* (e.g. after an agents-server restart), heartbeat failures counted up but nothing could cancel the stuck `streamFactory` call — the runner sat in `connecting` forever.
2. **Runtime restart stacked cache interceptors.** Every `BuiltinAgentsServer.start()` composed a new undici cache interceptor onto the global dispatcher without a guard. Restarting the runtime in-process stacked SQLite-backed cache layers over the same file; only a full app restart got a clean dispatcher.

## Approach

- `installDurableStreamsFetchCache` gets a module-level installed guard (warns on repeat calls).
- The runner lifecycle moves into an xstate machine (`pull-wake-machine.ts`): `stopped → running.{connecting, streaming, reconnecting} → stopping`. The structural win: **invoked actors are auto-aborted on state exit**, so a `STREAM_RESET` event during `connecting` cancels the in-flight connect via the promise actor's signal — no manual `connectAbort` bookkeeping, which is how the original bug slipped through. Backoff timers (`after`) cancel the same way.
- Diagnostics, heartbeat coalescing, and claim processing stay in the runner closure as effects the machine triggers (`PullWakeMachineEffects`). The machine decides *when*; the closure does *what*.

## Key invariants

- Every (state × event) pair is pinned by an exhaustive 35-case transition matrix test — adding a state or event forces a deliberate decision.
- Repeated heartbeat failures (≥2) reset the stream from **any** running substate.
- One shutdown sequence regardless of concurrent `stop()` calls; `stop()` rejects with drain errors.
- Claim actors survive stream reconnects and only drain (1s grace) at stop.

## Non-goals

- No behavioral changes to claiming, heartbeat cadence, or backoff timing (1s → ×2 → 30s cap preserved).
- Heartbeat send/coalescing logic was not moved into the machine — it's effect plumbing, not lifecycle.

## Verification

```bash
pnpm --filter @electric-ax/agents-runtime exec vitest run test/pull-wake-runner.test.ts   # 61 tests
pnpm --filter @electric-ax/agents exec vitest run test/durable-streams-cache.test.ts      # 2 tests
```

Full agents-runtime suite: 852 passed; 2 pre-existing environment failures unrelated to this PR (`sandbox-docker` Docker-timing assertion, `tool-providers` unbuilt workspace dep).

## Files changed

- `packages/agents-runtime/src/pull-wake-machine.ts` — new: xstate v5 lifecycle machine + effects interface
- `packages/agents-runtime/src/pull-wake-runner.ts` — rewritten as a thin adapter: same public API, closure keeps diagnostics/heartbeat/claims, machine owns lifecycle
- `packages/agents-runtime/test/pull-wake-runner.test.ts` — +1 hung-connection regression test, +40 machine transition tests; 21 original tests untouched
- `packages/agents/src/durable-streams-cache.ts` — idempotency guard + warning
- `packages/agents/test/durable-streams-cache.test.ts` — idempotency test
- `packages/agents-runtime/package.json` — adds `xstate` (zero runtime deps)
- `.changeset/pull-wake-runner-resilience.md` — patch bumps for both packages

### DIFF
--- a/packages/agents/src/durable-streams-cache.ts
+++ b/packages/agents/src/durable-streams-cache.ts
@@ -27,8 +27,6 @@ export function installDurableStreamsFetchCache(
     )
     return
   }
-  installed = true
-
   const store =
     options.store === `sqlite` || options.sqliteLocation
       ? new cacheStores.SqliteCacheStore({
@@ -51,4 +49,6 @@ export function installDurableStreamsFetchCache(
       })
     )
   )
+
+  installed = true
 }
--- a/packages/agents/src/durable-streams-cache.ts
+++ b/packages/agents/src/durable-streams-cache.ts
@@ -7,6 +7,8 @@ import {
 
 const MEMORY_CACHE_SIZE_BYTES = 100 * 1024 * 1024
 
+let installed = false
+
 export type DurableStreamsFetchCacheOptions =
   | false
   | {
@@ -19,6 +21,8 @@ export function installDurableStreamsFetchCache(
   options: DurableStreamsFetchCacheOptions = {}
 ): void {
   if (options === false) return
+  if (installed) return
+  installed = true
 
   const store =
     options.store === `sqlite` || options.sqliteLocation
--- a/packages/agents/src/durable-streams-cache.ts
+++ b/packages/agents/src/durable-streams-cache.ts
@@ -21,7 +21,12 @@ export function installDurableStreamsFetchCache(
   options: DurableStreamsFetchCacheOptions = {}
 ): void {
   if (options === false) return
-  if (installed) return
+  if (installed) {
+    console.warn(
+      `[agents] installDurableStreamsFetchCache called more than once; ignoring`
+    )
+    return
+  }
   installed = true
 
   const store =
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -793,7 +793,7 @@ importers:
         version: link:../../packages/react-hooks
       next:
         specifier: ^14.2.5
-        version: 14.2.17(@babel/core@7.29.0)(@opentelemetry/api@1.9.1)(@playwright/test@1.52.0)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+        version: 14.2.17(@opentelemetry/api@1.9.1)(@playwright/test@1.52.0)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       pg:
         specifier: ^8.12.0
         version: 8.13.1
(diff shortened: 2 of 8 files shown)


packages/agents/src/durable-streams-cache.ts
```diff
@@ -19,6 +21,13 @@ export function installDurableStreamsFetchCache(
   options: DurableStreamsFetchCacheOptions = {}
 ): void {
   if (options === false) return
+  if (installed) {
+    console.warn(
+      `[agents] installDurableStreamsFetchCache called more than once; ignoring`
+    )
+    return
+  }
+  installed = true
```
samwillis: Small resilience edge: this marks the cache as installed before the store and global dispatcher setup have actually succeeded. If `SqliteCacheStore`, `compose`, or `setGlobalDispatcher` throws on the first runtime start, later `BuiltinAgentsServer.start()` retries in the same process will warn and skip cache installation entirely. Could we move `installed = true` after `setGlobalDispatcher(...)` succeeds, or reset it in a `catch`?

pnpm-lock.yaml
```diff
@@ -793,7 +793,7 @@ importers:
         version: link:../../packages/react-hooks
       next:
         specifier: ^14.2.5
-        version: 14.2.17(@babel/core@7.29.0)(@opentelemetry/api@1.9.1)(@playwright/test@1.52.0)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
+        version: 14.2.17(@opentelemetry/api@1.9.1)(@playwright/test@1.52.0)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
```
samwillis: This lockfile update includes unrelated resolution churn beyond adding `xstate` (for example React peer snapshots and the Next/styled-jsx optional `@babel/core` metadata). Could you regenerate with the repo's pinned pnpm version or revert the unrelated lockfile entries so this PR only introduces the `xstate` lock changes?

samwillis (approved):
Looks good overall. I reviewed this interactively with GPT.

The xstate refactor preserves the pull-wake runner contract well from what I can see: start/stop idempotency, claim draining, heartbeat diagnostics, reconnect backoff, and the hung-connection regression all have targeted coverage.

Approving so this does not block merging, with the assumption that the inline follow-ups are addressed before/around merge.

PR #4546, opened 2026-06-09
feat(agents): editable session titles with txid-aware sync

KyleAMathews:
## Summary

Add editable session titles — users can click a session title in the UI to rename it inline, and Horton gets a `set_title` tool to programmatically set titles. Both paths use a new txid-propagation pattern so tag mutations await sync consistency before resolving.

## Approach

Three layers of change:

**1. `set_title` tool** (`packages/agents/src/tools/set-title.ts`)
New Horton tool that calls `ctx.setTag('title', trimmedTitle)`. Validates non-empty input, returns structured error responses (doesn't throw) so the LLM can react to failures.

**2. Click-to-edit UI** (`packages/agents-server-ui/src/components/EntityHeader.tsx`)
Title text becomes a `<button>` that switches to an `<input>` on click. Enter delegates to `onBlur` (single save path to prevent double-fire). Escape cancels. Optimistic update via `createSetEntityTitleAction` with rollback on persistence failure. Error toast on failure.

**3. txid propagation for tag/send/inbox mutations**
The core change that makes this work reliably. Tag mutations (`setTag`/`deleteTag`) now return the Postgres transaction ID (`pg_current_xact_id()`) so clients can call `awaitTxId()` to wait for the write to round-trip through Electric sync before proceeding. The same pattern was added to `send`, `updateInboxMessage`, and `deleteInboxMessage` (using `crypto.randomUUID()` as stream-event txids). The UI-side optimistic actions (`spawn`, `kill`, `signal`, `setEntityTitle`) all await the txid before resolving.

## Key Invariants

- `readTxidResponse` accepts both string and numeric txids (Postgres returns numbers per Electric convention; inbox UUIDs are strings)
- Tag `awaitTxId` only fires when a write actually occurred (`result.txid` is present); no-op tag writes skip it
- `onBlur` is the single save path for title edits — `onKeyDown(Enter)` blurs the input rather than calling save directly
- All mutation actions show error toasts on failure (spawn was missing one; now consistent)

## Non-goals

- **awaitTxId timeout handling**: All 12 `awaitTxId` call sites will reject on timeout, causing optimistic rollback even though the server succeeded. This pre-dates this PR and warrants its own design decision.
- **Parallel arrays refactor**: `uploadMessageAttachments` returns `{ ids, txids }` — could be `Array<{ id, txid }>` but that's a separate cleanup.

## Verification

```bash
pnpm vitest run packages/agents-runtime/test/runtime-server-client-update-metadata.test.ts
pnpm vitest run packages/agents/test/horton-tool-composition.test.ts
pnpm vitest run packages/agents-server/test/electric-agents-routes.test.ts
pnpm vitest run packages/agents-server/test/dispatch-policy-routing.test.ts
```

## Files Changed

**New:**
- `packages/agents/src/tools/set-title.ts` — `set_title` tool implementation

**Feature:**
- `packages/agents/src/agents/horton.ts` — Register tool, add to system prompt
- `packages/agents-server-ui/src/components/EntityHeader.tsx` — Click-to-edit inline title
- `packages/agents-server-ui/src/components/EntityHeader.module.css` — Styles for title input and button
- `packages/agents-server-ui/src/lib/ElectricAgentsProvider.tsx` — `createSetEntityTitleAction`, spawn/title error toasts, `awaitTxId` calls

**txid propagation:**
- `packages/agents-server/src/entity-registry.ts` — Return `pg_current_xact_id()` from tag mutations
- `packages/agents-server/src/entity-manager.ts` — Propagate txid through send/inbox/tag operations
- `packages/agents-server/src/routing/entities-router.ts` — Return txid in tag/send/inbox responses (204→200+JSON)
- `packages/agents-runtime/src/runtime-server-client.ts` — `readTxidResponse` accepts string+number txids
- `packages/agents-runtime/src/process-wake.ts` — `awaitTxId` after tag operations
- `packages/agents-server-ui/src/lib/sendMessage.ts` — `readRequiredTxid` helper, txid propagation in send/upload/inbox

**Tests:**
- `packages/agents-runtime/test/runtime-server-client-update-metadata.test.ts` — deleteTag coverage, numeric txid coercion, non-JSON fallback
- `packages/agents-server/test/electric-agents-routes.test.ts` — Tag endpoint tests with txid
- `packages/agents-server/test/dispatch-policy-routing.test.ts` — Updated mocks for new return types
- `packages/agents/test/horton-tool-composition.test.ts` — set_title happy/empty/error paths


packages/agents-server/src/entity-manager.ts
```diff
@@ -2687,14 +2710,14 @@ export class EntityManager {
...
-  ): Promise<ElectricAgentsEntity> {
+    token?: string
+  ): Promise<ElectricAgentsEntity & { txid?: number }> {
     const entity = await this.registry.getEntity(entityUrl)
     if (!entity) {
       throw new ElectricAgentsError(ErrCodeNotFound, `Entity not found`, 404)
     }
 
-    if (!this.isValidWriteToken(entity, token)) {
+    if (!this.isValidWriteToken(entity, token ?? ``)) {
```
samwillis: This still requires an entity write token even though the route now gates `/tags/*` with `withEntityPermission('write')`. The new UI title edit path calls `/tags/title` via `serverFetch` with principal auth, not the entity write token, so a principal-authorized caller will pass the route middleware and then get `401 Invalid write token` here. I think the route-authorized path needs to bypass this legacy token check, or the manager API needs a separate principal-authorized tag mutation path. A test that exercises no `Authorization`/`electric-claim-token` but valid write permission would catch it.

samwillis (approved):
Approving so this does not block merging, assuming the auth/token issue called out inline is addressed before merge.

Overall this looks good to me: the txid propagation is consistent across the new optimistic paths, the inline title editor avoids the common Enter/blur double-save trap, and the set_title tool is scoped and handles error cases cleanly.

Note: this was an interactive review with GPT.

PR #4547, opened 2026-06-09
Render standalone entity stream errors

KyleAMathews:
## Summary
- include standalone entity stream errors in the main timeline query
- render errors with their error code and message in the timeline UI
- keep existing run-attached error/fallback rendering unchanged

## Verification
- pnpm --dir packages/agents-server-ui typecheck
- pnpm --dir packages/agents-runtime test entity-timeline --run

Note: full packages/agents-runtime typecheck still fails on existing unrelated missing exports in test/pg-sync-source.test.ts.

### DIFF
--- a/packages/agents-runtime/src/entity-timeline.ts
+++ b/packages/agents-runtime/src/entity-timeline.ts
@@ -6,6 +6,7 @@ import {
   createLiveQueryCollection,
   eq,
   isNull,
+  isUndefined,
   like,
   localOnlyCollectionOptions,
   or,
@@ -1304,7 +1305,7 @@ function buildEntityTimelineQuery(
 
   const errorSource = q
     .from({ error: db.collections.errors })
-    .where(({ error }) => isNull(error.run_id))
+    .where(({ error }) => or(isNull(error.run_id), isUndefined(error.run_id)))
     .select(({ error }) => ({
       key: error.key,
       order: coalesce(error._timeline_order, `~`),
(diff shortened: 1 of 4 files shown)


packages/agents-runtime/src/entity-timeline.ts
```diff
@@ -1285,6 +1302,17 @@ function buildEntityTimelineQuery(
       new_state: signal.new_state,
     }))
 
+  const errorSource = q
+    .from({ error: db.collections.errors })
+    .where(({ error }) => isNull(error.run_id))
```
samwillis: This looks like it will miss the standalone errors actually written by `process-wake.ts`: those inserts omit `run_id` rather than setting it to `null`. I checked TanStack DB behavior in the worktree and `isNull(error.run_id)` only matched explicit `null`, not missing/`undefined`, so these runtime-created standalone errors would still be filtered out. Could this use `isUndefined(error.run_id)` or `or(isNull(error.run_id), isUndefined(error.run_id))`? The same consideration applies to `createEntityErrorsQuery` below.

samwillis (approved):
Interactive GPT review. Approving so this does not block merging, with the assumption that the inline issue around matching omitted standalone error `run_id`s will be addressed before merge.

Verification run locally in the worktree:
- `pnpm --dir packages/agents-runtime test entity-timeline --run`
- `pnpm --dir packages/agents-runtime build && pnpm --dir packages/agents-server-ui typecheck`

PR #4548, opened 2026-06-09
fix(agents-server): sanitize attachment filename fallback

KyleAMathews:
## Summary
- sanitize the legacy `Content-Disposition` filename fallback to ASCII-only before setting attachment response headers
- preserve the full Unicode filename in the RFC 5987 `filename*` parameter
- add a regression test for screenshot-style filenames containing a narrow no-break space

## Test plan
- `pnpm --filter @electric-ax/agents-runtime build`
- `pnpm --dir packages/agents-server exec vitest run test/electric-agents-routes.test.ts`

### DIFF
--- a/.changeset/attachment-filename-fallback.md
+++ b/.changeset/attachment-filename-fallback.md
@@ -1,0 +1,5 @@
+---
+"@electric-ax/agents-server": patch
+---
+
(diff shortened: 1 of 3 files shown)


samwillis (approved):
Reviewed interactively with GPT. No blockers found; happy to approve.

PR #4549, opened 2026-06-09
Secure Electric shape proxy: default-deny tables + validate client where

KyleAMathews:
## Summary

Hardens the agents-server Electric shape proxy (`/_electric/electric/v1/shape`) against two access-control bypasses. Previously, any authenticated principal could read **any** table — and sidestep per-tenant/per-principal row scoping — through the proxy. Now unscoped tables are denied and client `where` clauses that could escape the enforced scope are rejected.

## Root Cause

The proxy injects the privileged Electric API secret into every `/v1/shape` request, then applied scoping only via an `if/else-if` chain over ~9 known tables. Two gaps:

1. **Default-allow fall-through.** Any table *not* in the chain fell through to `return target` — forwarded upstream with the secret attached and **no row/column filter**. A low-privilege principal could request `table=entity_permission_grants`, `subscription_webhooks`, etc. and read the whole table (cross-tenant in a shared deployment).
2. **`where`-clause break-out.** Scoping was combined as `enforced AND (clientWhere)`. Since the client controls the text inside the parens and SQL `AND` binds tighter than `OR`, a payload like `1=1) OR (1=1` collapses to `(enforced AND (1=1)) OR (1=1)` → always true, defeating the scoping.

## Approach

- **Default-deny:** the table `if/else-if` now ends in an `else` that throws `ElectricProxyError(403, TABLE_NOT_ALLOWED)`. Only the explicitly column+row-scoped tables (the exact set the UI, mobile, and runtime sync) can be proxied; a missing `table` is also rejected. Verified no legitimate consumer requests a table outside this set (`wake_registrations` connects to Electric directly, not via the proxy).
- **`where` validation:** `isSelfContainedWhereClause` rejects (`400 INVALID_WHERE`) any client clause whose parentheses dip below the top level or are unbalanced, has unterminated `'`/`"` literals, or contains `--` / `/* */` comment markers — a single-pass scanner that ignores characters inside string/identifier literals (with `''`/`""` escapes).
- **Wiring:** `ElectricProxyError` lives in `server-utils.ts` (kept free of the heavy `entity-manager` import graph) and is mapped to an HTTP response by the global `errorMapper`, which now also logs rejections for intrusion detection.

## Key Invariants

- Every proxied `/v1/shape` request resolves to an explicitly scoped table, or it is rejected before forwarding (the secret is never attached to an unscoped request that reaches upstream).
- A client `where` that passes validation cannot break out of the `(...)` group it is wrapped in — so `enforced AND (client)` always keeps `enforced` in force.
- The validator only ever over-rejects (fail-safe); it never lets an unsafe clause through. Dollar-quoted / `E''` strings are intentionally unmodeled and fall into over-rejection.

## Non-goals

- Not converting the (pre-existing) `if/else-if` table chain to a `switch` — out of scope, would bury the security change.
- Not parsing/validating the *semantics* of the client `where` beyond structural self-containment; the upstream Postgres parser remains the authority on validity.
- Not constant-time secret comparison or other unrelated proxy hardening.

## Trade-offs

A structural scanner (vs. a full SQL parser) is simpler and safe because the only way to escape the wrapping parens is an unbalanced top-level `)`, which the depth check catches; comment markers are rejected outright to avoid the trailing-paren footgun. Reusing `ElectricAgentsError` was rejected to avoid pulling the heavy `entity-manager` graph into a utils module (keeps the unit test fast); the class instead mirrors its shape and is mapped uniformly in `errorMapper`.

## Verification

```bash
cd packages/agents-server
pnpm vitest run test/server-utils.test.ts test/routing-hooks.test.ts
# 32 passed — default-deny + missing table, break-out payloads (balanced
# 1=1) OR (1=1 and depth-underflow a=1) OR (b=2), trailing comment, unterminated
# '/" literals, allowed cases (OR-wrapping, parens/comments inside literals, ''
# escapes, double-quoted identifiers), and the 403/400 errorMapper mapping.
```

## Files changed

- `packages/agents-server/src/utils/server-utils.ts` — `ElectricProxyError` class, `isSelfContainedWhereClause` validator, default-deny branch + `where` validation in `buildElectricProxyTarget`.
- `packages/agents-server/src/routing/hooks.ts` — `errorMapper` maps `ElectricProxyError` to an HTTP response and logs the rejection.
- `packages/agents-server/test/server-utils.test.ts`, `test/routing-hooks.test.ts` — security test coverage.

### DIFF
--- a/packages/agents-server/src/utils/server-utils.ts
+++ b/packages/agents-server/src/utils/server-utils.ts
@@ -145,7 +145,20 @@ export function buildElectricProxyTarget(options: {
     throw new ElectricProxyError(`INVALID_WHERE`, `Invalid where clause`, 400)
   }
 
-  const table = options.incomingUrl.searchParams.get(`table`)
+  const tableParams = options.incomingUrl.searchParams.getAll(`table`)
+  if (tableParams.length !== 1) {
+    throw new ElectricProxyError(
+      `TABLE_NOT_ALLOWED`,
+      `Table is not available through the Electric proxy`,
+      403
+    )
+  }
+
+  const table = tableParams[0]
+  // Canonicalise the upstream table after validation so duplicate client query
+  // params cannot be interpreted differently by Electric or intermediaries.
+  target.searchParams.set(`table`, table)
+
   if (table === `entities`) {
     target.searchParams.set(
       `columns`,
--- a/packages/agents-server/src/utils/server-utils.ts
+++ b/packages/agents-server/src/utils/server-utils.ts
@@ -90,6 +90,25 @@ export async function fileExists(filePath: string): Promise<boolean> {
   }
 }
 
+/**
+ * Raised when an Electric shape proxy request must be rejected for security
+ * reasons (an un-scoped table, or a client `where` clause that could escape the
+ * enforced per-tenant/per-principal scoping). The global `errorMapper` hook
+ * maps this to an HTTP error response. Defined here (rather than reusing
+ * `ElectricAgentsError`) to keep this module free of the heavy entity-manager
+ * import graph.
+ */
+export class ElectricProxyError extends Error {
+  constructor(
+    readonly code: `INVALID_WHERE` | `TABLE_NOT_ALLOWED`,
+    message: string,
+    readonly status: number
+  ) {
+    super(message)
+    this.name = `ElectricProxyError`
+  }
+}
+
 export function buildElectricProxyTarget(options: {
   incomingUrl: URL
   electricUrl: string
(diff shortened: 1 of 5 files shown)


packages/agents-server/src/utils/server-utils.ts
```diff
@@ -117,6 +136,15 @@ export function buildElectricProxyTarget(options: {
     target.searchParams.set(`secret`, options.electricSecret)
   }
 
+  // The enforced scoping `where` is AND-combined with the client's own `where`.
+  // A client clause that is not self-contained (e.g. `1=1) OR (1=1`) could
+  // break out of its parenthesised group and neutralise the scoping under SQL
+  // operator precedence, so reject anything that isn't balanced.
+  const clientWhere = options.incomingUrl.searchParams.get(`where`)
```
samwillis, line 143: One remaining access-control edge case: duplicate `table` params are still forwarded upstream. The proxy authorizes only `incomingUrl.searchParams.get("table")`, but it appends the full original query string first, so `table=users&table=subscription_webhooks&offset=-1` is accepted as `users` here while the generated upstream URL still contains both `table` values plus the privileged `secret`. If Electric or an intermediary resolves duplicate params differently, this can preserve the bypass this PR is trying to close. Please reject `incomingUrl.searchParams.getAll("table").length !== 1` and/or force the canonical allowlisted value on the upstream target with `target.searchParams.set("table", table)` after validation.

samwillis (approved):
Reviewed interactively with GPT.

Approving on the assumption that the duplicate-query-parameter issue in the inline comment will be addressed before/around merge; I do not want this review state to block merging.

Verification performed:
- pnpm install with Node v22.13.0
- pnpm --dir packages/agents-server exec vitest run test/server-utils.test.ts test/routing-hooks.test.ts
- Initial routing-hooks run needed @electric-ax/agents-runtime to be built in the fresh worktree; after pnpm --dir packages/agents-runtime build, routing-hooks.test.ts passed.

PR #4551, opened 2026-06-10
feat(agents): generic externally-writable custom collections (comments as first consumer)

balegas:
## Summary

An alternative to #4529 that delivers the same session-comments feature, but built on a **generic, extensible interface** instead of hardcoding `comments` into the entity schema. Comments becomes one *custom collection* among potentially many; the mechanism underneath is reusable.

Three capabilities the agent runtime didn't have before:

1. **Opt-in externally-writable collections.** Entity state is agent-owned. A collection becomes writable from the HTTP router only when its definition sets `externallyWritable` — everything else stays agent-only by default. (Every collection is always writable *by the agent*; only an opt-in subset is writable *externally*, hence the name.)
2. **Authenticated, auditable writes.** Router writes require authentication. The server stamps the authenticated principal into the **change-event header** (provenance, outside the user payload), and the client materializes it into a read-only **virtual column** (`_principal`). A client can't spoof it via `value`.
3. **Schema-validated writes.** Each collection carries a schema; router writes validate `value` server-side via the existing `validateWriteEvent` before append.

## Design

**Runtime** (`@electric-ax/agents-runtime`)
- `CollectionDefinition.externallyWritable?: boolean`.
- At registration, writable collections are emitted as `externally_writable_collections` (`{ type }` per collection).
- The entity stream DB materializes `headers.principal` into the fixed `_principal` virtual column, mirroring `_timeline_order`, and strips it (with `_seq`/`_timeline_order`) before client write-back.
- `createEntityTimelineQuery` accepts `customSources` — consumer-provided sources unioned into the timeline query. The runtime itself knows nothing about specific custom collections, and the agent's LLM context never includes them.

**Server** (`@electric-ax/agents-server`)
- `externally_writable_collections` persisted as a jsonb column on `entity_types` (migration 0016) and resolved via `getEffectiveSchemas`. Legacy `principalColumn` in registration payloads is accepted and ignored for version-skew tolerance.
- `EntityManager.writeCollection` gates on it (403 if not writable), enforces entity-status/fork-lock, stamps `headers.principal = { url, kind, id }`, validates `value`, and appends.
- `POST /:type/:instanceId/collections/:collection` exposes it (auth + `write` permission, same middleware chain as `send`). Single POST, `operation` in the body, 201 insert / 200 update·delete.

**Comments consumer** (`@electric-ax/agents` + `@electric-ax/agents-server-ui`)
- `commentsCollection` (schema ported from #4529, minus `from_principal`) declared as `state` on Horton and worker.
- The UI projects comments into the timeline by passing a comments `customSources` entry to the timeline query (author resolved from `_principal`); comments are a UI-level concern, hardcoded only in `agents-server-ui`.
- The #4529 comments UI is cloned and re-sourced: writes via an optimistic TanStack action backed by the authenticated `/collections/comments` endpoint; reads `comment.from`.

## Test Plan

- [x] `@electric-ax/agents-runtime` — typecheck + 839 tests (incl. principal virtual-column materialization, write-back stripping, registration emit, timeline projection)
- [x] `@electric-ax/agents-server` — typecheck + suite (incl. `writeCollection` 403/409 gating, principal-header stamping, schema validation, `externally_writable_collections` jsonb round-trip)
- [x] `@electric-ax/agents` — typecheck + 60 tests (comments declared on Horton/worker)
- [x] `@electric-ax/agents-server-ui` — typecheck + 88 tests (optimistic comment write to `/collections/comments`, author from `_principal`)
- [ ] Manual: post a comment from the UI, confirm it syncs, right-aligns for the author, and the principal is recorded in the change-event header


.changeset/generic-externally-writable-collections.md
```diff
@@ -0,0 +1,8 @@
+---
+'@electric-ax/agents-runtime': minor
+'@electric-ax/agents-server': minor
+'@electric-ax/agents-server-ui': minor
+'@electric-ax/agents': minor
```
msfstef: We generally do patch changes still even for breaking changes

docs/superpowers/plans/2026-06-10-generic-writable-collections.md
msfstef, line 1: Are we committing plans or not?

docs/superpowers/specs/2026-06-10-generic-writable-collections-design.md
msfstef, line 1: Are we committing the specs? Also, the spec is explicitly text-wrapped, not sure if we follow some convention.

msfstef:
**Design question: is `comments` meant to be per-agent or platform-level?**

The externally-writable safeguard itself makes sense to me — default-deny, agent owns its state. But the PR seems ambiguous about whether the *comments collection's existence* is meant to be optional:

- **Write side:** comments only exists because horton/worker explicitly declare `state: { comments: commentsCollection }`. Any agent definition that omits that line silently loses commenting.
- **Read side:** the UI assumes comments everywhere — `UI_ENTITY_CUSTOM_STATE` registers it on every entity stream, the Comment tab and comments-only view are offered for all chat entities, and posting to a type that didn't declare comments fails only at submit time with a 403.

The design spec (dropped in 13050d02) justifies the *writability* opt-in, but I couldn't find a rationale for making *existence* opt-in per agent — it seems to just fall out of the mechanism.

If comments is a platform feature, could it join the default layer instead — e.g. merged into every registration alongside `DEFAULT_STATE_SCHEMAS`, with its `externally_writable_collections` entry — so the server-side truth matches what the UI already assumes? If it's genuinely per-agent, the UI should gate the comment affordances on the type's `externally_writable_collections` (already synced to the client). As-is we have per-agent boilerplate on the write side and an unconditional promise on the read side.

msfstef:
**Provenance is "last touch", which the comments feature can't actually be built on**

The header-principal design answers *who performed this write* — but comments need *who authored this row*, and for any operation other than `insert` those are different questions:

- There's no stored author field (`from_principal` was deliberately dropped); authorship exists only as `_principal`, materialized from the **latest** event's header.
- Updates are full row replaces, and the router stamps the *caller's* principal on every event. So B updating A's comment replaces the content **and flips the displayed author to B**. A's authorship survives only in raw stream history, which nothing reads.
- Agent-side updates are worse: no principal header + full replace means the row comes out with no `_principal` at all.
- The schema's `edited_at`/`deleted_by` fields imply edits/soft-deletes are intended — but `deleted_by` is client-supplied in `value` (spoofable), and the provenance model can't represent "created by A, edited by B" at all.

Combined with no ownership checks (the dropped spec records this as a decision: any principal with entity `write` may perform any operation), anyone who can prompt the session can delete or rewrite anyone's comment — and rewriting silently reattributes it.

Insert-only is the case where the provenance model is actually *sufficient*, so concretely: could `ExternallyWritableCollectionConfig` grow an operations allowlist (e.g. `{ type, operations: ['insert'] }`), with comments shipping insert-only? That closes the moderation/reattribution hole without the server needing to materialize state for row-level ownership checks, and update/delete can be enabled later once there's a durable-authorship story (server-echoed creator, ownership-gated ops, or an explicit author field).

.changeset/generic-externally-writable-collections.md
balegas: Done — switched all four entries to patch in 96731120a.

balegas:
> Design question: is `comments` meant to be per-agent or platform-level?

We decided that comments would only be a feature for horton and hard-code it to the ui. I think this is limiting, so I introduced contracts and entities can opt-in for them. The ui will render comments for any entity with that capability.

- The canonical `commentsCollection` declares a `contract: 'comments/v1'` marker, forwarded in type registration.
- The server reserves the `comments` collection name for that contract (and vice versa), so an agent's unrelated writable collection can never be mistaken for platform comments.
- The entity GET response now includes the type's `externally_writable_collections`, and the UI registers the comments collection on the entity stream only when the contract is advertised — the Comments view, composer Comment tab, timeline merge, and display toggle are all gated on it, so the submit-time 403 is no longer reachable through the UI.

docs/superpowers/plans/2026-06-10-generic-writable-collections.md
balegas: Not committing them — the plan was dropped from the PR in a8df99a04.

docs/superpowers/specs/2026-06-10-generic-writable-collections-design.md
balegas: Same as the plan — the spec was dropped from the PR in 13050d02c (looks like that push crossed paths with your comment).

balegas:
> Provenance is "last touch", which the comments feature can't actually be built on

Provenance here is intentionally attributed to the mutation: `_principal` answers "who performed the write that produced the current value", which is the correct attribution for materializing the latest value of a row — it isn't trying to be a durable authorship record. You're right that durable authorship ("created by A, edited by B") isn't representable in the materialized row today, and that anyone with entity `write` can rewrite or remove a comment — we're consciously accepting that for now: comments inherit entity-level write semantics.

The fuller treatment, if/when we want it, is additive on top of this design rather than a rework of it:

- **Comment history** — every revision (with its principal header) is already durably retained in the stream; surfacing it is a read-side feature, no write-path changes needed.
- **Permissions on the state contract** — an operations allowlist like the one you sketched (or ownership rules) hangs naturally off the collection config / the `comments/v1` contract added for the per-agent gating.

So: deferring both and leaving the scope of this PR as is. Happy to open a follow-up issue capturing the moderation/authorship gaps so they don't get lost.

msfstef:
A few smaller items — two worth addressing in this PR, one fine to defer:

**1. Client comment keys can collide across users (silent data loss).** `createClientCommentKey` mints `client-comment-${Date.now()}-${pendingOrderIndex}`, and the server uses the client key verbatim. `pendingOrderIndex` is a module-level counter seeded at the same constant in every client, so two users' first comments get the same index — if they land in the same millisecond the keys are byte-identical. Since an `insert` with an existing key is coerced to `update` in the stream-db, the second comment silently overwrites the first (body **and** `_principal`), so one user's comment vanishes with no error and is misattributed.

The client must mint the key (optimistic↔synced reconciliation needs it), so the fix is just to make it globally unique — `crypto.randomUUID()` for the key — while keeping `pendingOrderIndex` for the optimistic `order` field. Identity and ordering are independent concerns that got conflated here; order collisions are harmless (real `_timeline_order` takes over once synced), key collisions are durable.

**2. Comments can't be posted to stopped/completed sessions — design question.** `writeCollection` rejects via `rejectsNormalWrites` (`stopping`/`stopped`/`killed` → 409), because comments are events on the entity's main stream which may be closed once the entity terminates. But the affordances this PR ships — reply-to-any-row, the comments-only view, jump-to-target — are a post-hoc *review* toolkit, and review most naturally happens on a *finished* session. So the highest-value moment to comment is the one state the endpoint refuses. I don't think this is a quick fix (it's the stream-lifecycle coupling, not a policy check), so mainly flagging it as a product decision to make explicitly: is v1 intentionally "comment on live sessions only", and is that consistent with shipping a review-oriented comments view? If review of completed sessions is the point, the stream-lifecycle tension needs a plan even if deferred.

**3. (defer) No `agents-mobile` implementation.** Comments will sync into mobile entity streams but render nowhere. Fine to leave for a follow-up — just flagging so it's tracked.

---

**Note — still open from the provenance thread.** I'm happy to accept deferring durable authorship to a follow-up, but two of those points are cheap and arguably belong in this PR rather than the follow-up issue, since they ship live in the meantime:
- **Insert-only external ops** for the comments contract (the operations allowlist you sketched) — without it, update/delete are open at the endpoint, so the moderation/reattribution hole is live in shipped code, not actually deferred.
- **`deleted_by` is client-supplied in `value`** — spoofable, and it contradicts the header-provenance principle the PR is built on. Either derive it from the principal header or drop the field until the soft-delete flow exists.

Fine if you'd rather fold these into the follow-up issue too — just want them tracked explicitly rather than implied by "deferred".

msfstef (approved):

balegas:
Thanks — went through these.

**1. Client key collision (fixed).** Good catch — this was a real silent-overwrite. `createClientCommentKey` now mints `client-comment-${crypto.randomUUID()}`; `pendingOrderIndex` stays purely for the optimistic `order`, so identity and ordering are no longer conflated.

**2. Commenting on stopped/completed sessions.** Intentional for v1: commenting is scoped to live sessions. Sessions here are long-lived, so the "review happens on a finished session" framing doesn't apply the way it would for short runs — review/commenting happens against the live session while it's active. Decoupling comments from the main-stream lifecycle would be separate work; not planning it for this PR.

**3. No `agents-mobile` impl.** Agreed — deferring to a follow-up.

**4. Insert-only external ops (fixed, in this PR).** Added a per-collection `operations` allowlist on the externally-writable contract, enforced server-side (403 on disallowed ops), defaulting to insert-only. Comments declare `['insert']`. So update/delete are no longer open at the endpoint — the moderation/reattribution hole is closed in shipped code, not deferred.

**5. `deleted_by` spoofable (fixed, in this PR).** Dropped `deleted_by`, `deleted_at`, and `edited_at` from the `comments/v1` schema entirely — under insert-only they're unsettable dead fields, and `deleted_by`-in-`value` directly contradicted the header-provenance model. When the edit/soft-delete flow lands, `deleted_by` will be derived from the principal header (same path as `_principal`), never from client `value`. Removed the corresponding dead delete/edit rendering from the UI too.

Durable authorship / edit / soft-delete remain deferred to the follow-up.

PR #4553, opened 2026-06-10
feat(agents-mobile): schema-driven spawn args, model controls & image attachments (desktop parity)

msfstef:
## Summary

Brings `agents-mobile` to parity with the desktop agents app for four composer / spawn capabilities, building on #4533 (the native mobile slash-command composer):

- **Schema-driven new-session args** — render an agent's `creation_schema` as native controls (feature 7)
- **Model / reasoning / speed controls at spawn** (feature 8)
- **Image attachments in the in-session chat composer** (feature 6)
- **Image attachments at spawn time** (feature 9)

**No server / runtime changes were required.** Every wire contract already exists; this PR makes the mobile client produce the same payloads desktop does, plus a small set of shared-UI fixes that device testing surfaced.

## Why

#4533 reached slash-command composer parity on mobile, but the composer was otherwise text-only. The four gaps above are all implemented on desktop in `agents-server-ui`; this closes them on mobile.

## Key insight: 4 features → 2 capabilities, 0 server changes

1. **Features 7 + 8 are the same mechanism.** Model / reasoning / speed are *not* special server concepts — they're ordinary `enum` properties of the entity type's `creation_schema`, detected by field-name convention. `creation_schema` is already synced to mobile and `spawnEntity` already forwards `args`, so this is purely additive native UI over data mobile already has.
2. **Features 6 + 9 share one path.** The shared send action already threads attachments end-to-end; the only blocker was that `uploadMessageAttachments` used browser `File` / `FormData.set`, which React Native lacks.

### Capability A — schema-driven spawn args + model/reasoning/speed (7, 8)
- New mobile: `lib/spawnArgs.ts`, `lib/lastPickedModel.ts`, `components/SchemaArgsControls.tsx`, wired into `NewSessionScreen`.
- Renders the synced `creation_schema` natively, full parity with desktop `SchemaForm`: enum → BottomSheet picker pill (model grouped by provider, last pick remembered in AsyncStorage), boolean → `Switch`, string/number → text field, string-array → comma field, object → JSON field. Field labels are humanized (`reasoningEffort` → "Reasoning Effort") and provider names mapped (`openai` → "OpenAI"). Required fields gate "Start session"; server still validates against the schema.
- **"Start session" is pinned to the bottom** of the new-session screen (outside the scroll) so it stays reachable as the schema / model / sandbox sections stack up. The bar reports its height via `onLayout` and the scroll content is padded by that height so every section still scrolls clear; the bar owns its safe-area bottom inset (`Screen` leaves the bottom to screen-specific controls, matching the in-session composer).

### Capability B — image attachments (6, 9)
- New mobile: `lib/attachments.ts` + attach button / thumbnail tray in `NativeComposer`; `image`/`camera` icons.
- Deps: `expo-image-picker` (library + camera; **iOS uses a native `ActionSheetIOS`**), `expo-image-manipulator` (transcode to JPEG).
- In-session reuses the shared `createSendComposerInputAction({ attachments })`. At spawn it mirrors desktop `doSpawn` (spawn without `initialMessage`, then send `immediate` with attachments). Gated on whether the session's model supports image input (`schemaModelSupportsImageInput`).
- **Display is handled by the existing desktop chat-log WebView embed** — no native timeline rendering was needed; once a mobile-sent attachment syncs, the embedded renderer shows the thumbnail.

### Shared / cross-cutting
- `agents-server-ui/lib/schemaProperties.ts` (new): the DOM-free schema-classification helpers (inline props, model/reasoning/speed detection + grouping, string-array/JSON parsing, object-schema guards) extracted from `SchemaForm`/`NewSessionView` so desktop and the native mobile composer share one source of truth — now including the model-provider label map + `provider:model` id parsing that were otherwise mirrored on each platform. Behaviour-preserving for desktop.
- `agents-server-ui/lib/sendMessage.ts`: `uploadMessageAttachments` accepts `File | NativeFileDescriptor` and branches `FormData.append` (RN) vs `.set` (web). No wire-format change.
- Embed / timeline fixes (device testing): `WorkspaceProvider` around the embed router so the state-inspector view's `useWorkspace` resolves; image-preview dialog respects the composer inset; iOS timeline thumbnail containment.
- `agents/horton.ts`: robust session-title generation for image/attachment messages.

## Key decisions (the "why", for future sessions)

- **Extract shared schema helpers rather than duplicate.** Mobile reuses `agents-server-ui` pure logic + `agents-runtime/client`, and builds native UI itself (the established pattern). The desktop side of the refactor is mostly deletion.
- **Attachment display via the existing WebView embed**, not native rendering — the desktop chat log already renders attachments; mobile just needs to *send* them.
- **iOS picker via `ActionSheetIOS`, not our `BottomSheet`.** Presenting the native image picker over an RN `Modal` freezes / fails to open on iOS; the native action sheet has no such conflict. Android keeps the `BottomSheet` (intent-based picker, no conflict).
- **Transcode every picked image to JPEG.** iOS returns HEIC, which Anthropic/OpenAI vision models reject (the agent run errored with `finish_reason=error`); `expo-image-manipulator` normalizes to JPEG.
- **Title generation hardening.** The low-cost, text-only title model went conversational on image messages it couldn't see ("I'm sorry but no images were actually shared…") and that became the title — firmer prompt + a guard that rejects sentence-like responses and falls back to the local title.

## Known limitations / follow-ups

- **Android keyboard:** the composer is `position: absolute; bottom: 0` and relies on Android window resize to lift above the IME; the iOS-style translate is intentionally **not** applied on Android (doing so double-offsets — device testing showed a 2× lift, proving resize is active). If a device ever shows the composer covered under the keyboard, the deterministic fix is `android:windowSoftInputMode="adjustNothing"` (config plugin) + translate-only — that needs a native rebuild. Rationale is in the `SessionScreen` composer-transform comment.
- **Image-preview bottom gap** is a tuned constant in `AttachmentImagePreviewDialog.module.css` (`36px` = 16 gap + 20 `CHAT_COMPOSER_OVERLAP` compensation); nudge if a device shows too much / too little gap.
- **Title fix** only affects newly-created chats (existing bad titles don't change retroactively).
- Attachment policy is **image-only** (desktop parity); arbitrary documents are out of scope.

## Testing

- **Unit:** `schemaProperties` (agents-server-ui), `spawnArgs` (agents-mobile), `generate-title` (agents) — all green. Typecheck clean across `agents-server-ui`, `agents-mobile`, `agents-desktop`, `agents`; `expo-doctor` 18/18.
- **Manual (iOS + Android dev builds):** schema controls + model/reasoning pickers, required-field gating, attach from library/camera, spawn-with-attachments + in-session send, HEIC handling, image-input gating, timeline display, image-preview layering, keyboard behaviour.

## Build / deploy notes

- **Native modules added** (`expo-image-picker`, `expo-image-manipulator`) + an `expo-image-picker` config plugin (iOS photo/camera usage strings) → requires a **native dev build** (not Expo Go; a Metro reload is not enough).
- **Embed / CSS changes** (`agents-server-ui`) ship in the WebView DOM bundle → Metro reload.
- **Title generation** runs server-side in `agents` (the horton handler) → restart / redeploy the agents server / runner.


msfstef:
## Review follow-up (8446ef4)

Thanks for the review. Addressed both actionable items; documenting the two suggestions we intentionally left as-is.

### Addressed
- **Changeset coverage for `@electric-ax/agents`** — added the `agents` patch entry + a paragraph describing the title-generation hardening. The `horton.ts` change is a real behavioral change in a publishable package, so it now ships with a version bump and changelog entry instead of silently.
- **`buildInitialSpawnArgs` now respects `omitKeys`** — it was the only function in the spawn-args pipeline that didn't. Threaded `omitKeys` through both seeding loops, passed `SCHEMA_OMIT_KEYS` at the call site, and added a test covering an inline enum and a defaulted text field. Harmless today (horton's `workingDirectory` has no default), but removes the latent leak you described.

### Deliberately not changed

- **`creationSchema` omitted from the seeding `useEffect` deps** — this is intentional, not an oversight. Keying the reseed on `activeTypeName` (not the schema object) is exactly what stops a types-collection re-sync from clobbering in-progress edits; adding `creationSchema` to the deps would reintroduce that bug. The trade-off you noted (a server-side schema edit while the same type stays selected won't reseed) is the lesser evil and acceptable. Left a comment at the call site already states this intent. We confirmed lint is not flagging it.

- **Required-field gating passing on raw invalid-JSON text** — intentionally matched to desktop `SchemaForm`: any non-empty value passes the client gate and the server is the source of truth for schema validation. Diverging here would put mobile out of parity and duplicate validation logic the server already owns. Noted as parity-consistent in the review, and we agree it should stay that way.

### Note on the "missing changeset / unrelated 0.4.15 entries"
The unrelated `agents` 0.4.15 CHANGELOG entries (undici/fork/schedule/docker) seen in earlier diffs were a stale-local-`main` artifact, not something this branch introduces — current `main` already contains that release-publish commit as this PR's base (the branch is stacked directly on #4533). Verified: vs current `main`, this branch's only `packages/agents/` changes are `horton.ts` + `generate-title.test.ts`.

msfstef:
https://github.com/user-attachments/assets/7b454432-41a9-4d26-8b45-38cd5055af3e

[android-photo.webm](https://github.com/user-attachments/assets/50e0bed6-cb02-4317-9144-42553f0ce7bb)

kevin-dp (approved):

PR #4560, opened 2026-06-11
fix(sync-service): drop duplicate relation messages before routing

robacourt:
Relation messages can be emitted by Postgres when tables autoanalyze and autovacuum. Previously we checked all shapes to see if the Relation message affected them, which scales O(n) with the number of shapes (taking approx 25ms with 20K shapes). This PR checks to see if the Relation message has actually changed, and only checks the shapes if it has, saving that expensive operation.

## Summary
- classify relation messages as new, unchanged, changed, or identity changed
- drop unchanged relation messages before shape routing so they do not call Filter.affected_shapes/2
- keep inspector cache cleanup and partition semantics

### DIFF
--- a/.changeset/drop-duplicate-relations.md
+++ b/.changeset/drop-duplicate-relations.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Drop duplicate relation messages before shape routing when Postgres resends unchanged table metadata.
--- a/packages/sync-service/lib/electric/replication/shape_log_collector.ex
+++ b/packages/sync-service/lib/electric/replication/shape_log_collector.ex
@@ -633,9 +633,7 @@ defmodule Electric.Replication.ShapeLogCollector do
   end
 
   defp handle_relation(state, rel) do
-    OpenTelemetry.add_span_attributes("rel.is_dropped": false)
-
-    {updated_rel, tracker_state} =
+    {relation_status, updated_rel, tracker_state} =
       AffectedColumns.transform_relation(rel, state.tracked_relations)
 
     # PG doesn't send all the details in the relation message (in particular, nullability), but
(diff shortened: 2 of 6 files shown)


erik-the-implementer:
> [!NOTE]
> This review was generated by Claude on request by @alco.

Dropping `:unchanged` relation messages is only safe under one invariant: the tracker state in `tracked_relations` is always consistent with the stored identity of every live shape. Routing "unchanged" relations through `Filter.affected_shapes/2` was never a no-op by accident — it was the recovery mechanism for when that invariant breaks. Each shape compared *its own* `root_table_id` / name / column count against the message, so a shape that missed an earlier relation change got terminated on the next (identical) message. This PR removes that net, and there are two concrete windows where the invariant breaks:

**1. Retry after `connection_not_available` (in-memory tracker advanced too early)**

`tracked_relations` is committed into state (`shape_log_collector.ex:653`) *before* `Partitions.handle_relation/2`. When that returns `{:error, :connection_not_available}`, the error path returns the already-advanced state, and the replication client retries the same event (`replication_client.ex:371-380`). On retry the relation matches the tracker → `:unchanged` → dropped. A real schema change never reaches consumers.

**2. Restart with the early-persisted tracker (not fixed by addressing #1 alone)**

`PersistentReplicationState.set_tracked_relations/2` persists the advanced tracker *before* partition handling and *before* `publish/2`. That ordering is pre-existing, but it used to be harmless precisely because unchanged relations were still routed. Now:

- a `:changed` relation arrives → tracker persisted → Electric crashes/redeploys before consumers process the publish (or the handler errors and the stack restarts during the retry window — the `:DOWN` retry path in `replication_client.ex` shows handler crashes are expected and retried);
- on restart, shapes reload stale, the tracker reloads already containing the new relation (`shape_log_collector.ex:229`);
- PG re-sends the relation message on the new session → matches persisted tracker → `:unchanged` → dropped.

The stale shapes (select-all shape with old column count, shape on a renamed table, …) are never terminated and keep serving the wrong schema until the next actual schema change. Previously this exact sequence self-healed via the per-shape recheck.

**Fix**

Commit the tracker — both in-memory and persisted — only after `Partitions.handle_relation/2` succeeds *and* `publish/2` returns (publish is synchronous per dependency layer, so this is a straightforward reorder):

```elixir
case Partitions.handle_relation(state.partitions, updated_rel) do
  {:ok, partitions} ->
    state = %{state | partitions: partitions}

    with {:ok, state} <- publish_relation(state, updated_rel, relation_status) do
      :ok =
        PersistentReplicationState.set_tracked_relations(
          tracker_state,
          state.persistent_replication_data_opts
        )

      {:ok, %{state | tracked_relations: tracker_state}}
    end

  {:error, :connection_not_available} ->
    {{:error, :connection_not_available}, state}
end
```

This makes the failure direction safe: a crash after publish but before persist means the relation re-classifies as `:changed` on replay and is re-routed — redundant but harmless — instead of silently dropped.

**Tests**

Both windows are currently uncovered, and they're exactly where this PR changes recovery semantics:

- have `Partitions.handle_relation/2` return `{:error, :connection_not_available}` for a changed relation, re-send it, assert it *is* routed;
- persist a changed relation into `PersistentReplicationState`, restart the collector with shapes whose stored identity predates the change, re-send the same relation message, assert the affected shape is terminated.

(For the record: the steady-state optimization itself is sound. Info that's missing from relation messages entirely — nullability etc. — was never detectable by the shape recheck anyway; that's handled by `Inspector.clean/2`, which this PR correctly preserves for the `:unchanged` case.)

robacourt:
Addressed the tracker-ordering review feedback in b96a3a90d.

Changes:
- tracked_relations is no longer persisted or installed in memory before partition handling and relation publish complete.
- On :connection_not_available, the collector returns the original tracker state so the retry reclassifies changed relations correctly.
- Added regression coverage for retry after partition inspection failure and for not persisting a changed relation before routing completes.

Reran:
- mix test test/electric/replication/shape_log_collector_test.exs
- mix test test/electric/replication/shape_log_collector/affected_columns_test.exs
- mix test test/electric/shapes/consumer_test.exs
- git diff --check && git diff --cached --check

alco (approved):

PR #4561, opened 2026-06-11
Flatten 1:1 shape GET child spans into Plug_shape_get root-span attributes

erik-the-implementer:
## Motivation

We are over our Honeycomb event budget, and two spans are an outsized contributor: `shape_get.api.load_shape_info` and `shape_get.plug.serve_shape_log` together produce ~147M events/week — roughly 33% of the org's event budget.

Both spans are strictly 1:1 with the `Plug_shape_get` root span (one `load_shape_info` per request; one `serve_shape_log` per request on the log-serving path), and neither carries information that can't live on the root span. `load_shape_info` is a leaf span; `serve_shape_log`'s only children are the `shape_get.plug.stream_chunk` spans, which are untouched by this PR and now parent directly to the root.

## What changed

- New helper `Electric.Telemetry.OpenTelemetry.with_flattened_span/2`: runs a function and records its duration and process-memory footprint as attributes on the **current** span instead of emitting a child span.
- `Electric.Shapes.Api.do_load_shape_info/1` and `serve_shape_log/1` now use it instead of `with_span/4`. No other spans were changed (`shape_get.plug.serve_subset_response` and `shape_get.plug.stream_chunk` are left as-is).
- Changeset: patch bump for `@core/sync-service`.

## What's preserved, and where

All numeric data the child spans used to carry now lands on the `Plug_shape_get` root span:

| Was (child span) | Now (root-span attribute) |
|---|---|
| `shape_get.api.load_shape_info` duration | `load_shape_info.duration_ms` (float ms, sub-ms precision) |
| its `memory.start.{process,binary}_bytes` / `memory.end.{process,binary}_bytes` | `load_shape_info.memory.start.*` / `load_shape_info.memory.end.*` |
| `shape_get.plug.serve_shape_log` duration | `serve_shape_log.duration_ms` |
| its `memory.start.*` / `memory.end.*` | `serve_shape_log.memory.start.*` / `serve_shape_log.memory.end.*` |

The memory attributes are prefixed with the flattened span's short name so they don't collide with the root span's own `memory.start.*` / `memory.end.*` attributes.

Error semantics are unchanged: attributes (including duration) are recorded in an `after` block even when the wrapped code raises, and the exception propagates up to `ServeShapePlug`'s error handler, which records the exception event and sets error status on the root span exactly as before (the child spans never carried error status themselves — `:otel_tracer.with_span/4` only ends the span on raise).

## Trade-offs

- Trace waterfalls lose the visual split between "load shape info" and "serve shape log" inside a request — but no numeric data is lost; the same timings are queryable as root-span attributes.
- The `[:electric, :shape_get, :api, :load_shape_info, :start|:stop|:exception]` and `[:electric, :shape_get, :plug, :serve_shape_log, ...]` `:telemetry.span` events are no longer emitted. A repo-wide search found no handlers attached to them.
- In embedded (elixir-client) usage, `Api.serve_shape_log/1` previously created a root span of its own; with no enclosing span the helper is now a no-op (it just runs the function), so embedded calls produce no span — consistent with the unsampled-request path.

## Verification

- `mix test test/electric/telemetry/open_telemetry_test.exs test/electric/shapes/api_test.exs test/electric/plug/serve_shape_plug_test.exs test/electric/plug/serve_shape_plug_logging_test.exs` — 121 tests, 0 failures.
- New unit tests for `with_flattened_span/2` cover attribute recording, the raise path, and the no-span-context no-op.

### DIFF
--- a/.changeset/flatten-shape-get-spans.md
+++ b/.changeset/flatten-shape-get-spans.md
@@ -3,3 +3,5 @@
 ---
 
 Reduce OTel span volume for shape GET requests: the `shape_get.api.load_shape_info` and `shape_get.plug.serve_shape_log` child spans (strictly 1:1 with the root request span) are no longer emitted. Their timing and process-memory information is preserved as attributes on the `Plug_shape_get` root span instead (`load_shape_info.duration_ms`, `load_shape_info.memory.{start,end}.*`, `serve_shape_log.duration_ms`, `serve_shape_log.memory.{start,end}.*`).
+
+Note for operators: since these two names no longer exist as spans, listing them in `ELECTRIC_EXCLUDE_SPANS` (the `Sampler` `:exclude_spans` setting) no longer has any effect — the data is recorded as root-span attributes whenever the request trace is sampled. In embedded usage (elixir-client / Phoenix.Sync) `Api.serve_shape_log/1` previously created a standalone root span; it now records attributes onto the caller's current span if one exists (e.g. the Phoenix request span in OTel-instrumented apps) and emits nothing otherwise.
--- a/.changeset/flatten-shape-get-spans.md
+++ b/.changeset/flatten-shape-get-spans.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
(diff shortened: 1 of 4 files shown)


robacourt (approved):
Nice!

PR #4562, opened 2026-06-11
Read and apply the SampleRate attribute to shape spans from the tracestate request header

erik-the-implementer:
## Motivation

The Cloudflare worker in front of Electric Cloud head-samples successful requests at 1:20 and keeps **all** >= 500 responses at sampleRate 1, weighting its exported spans accordingly. Electric only ever received the sampled/not-sampled bit via `traceparent`, which causes an **~20x undercount**: Electric spans carry no `SampleRate` attribute, so aggregates over electric-region spans under-report worker traffic ~20x relative to the worker's own (weighted) spans.

Server-side errors are already captured as Sentry events, so this PR deliberately does not create extra spans to carry error attributes — it only stamps the sampling weight onto spans that are already being recorded.

The worker now sends its sampling rate alongside `traceparent` on every proxied request (regardless of the sampled flag) via the W3C `tracestate` header:

```
tracestate: electric=rate:<positive-integer>    # currently rate:20
```

Tracing backends that understand sampling weights natively read an integer span attribute named `SampleRate` and weight aggregates by it.

## Behavior matrix (per shape GET request)

| Remote parent | Rate hint | Response status | Behavior |
|---|---|---|---|
| none (direct traffic) | — | any | **Unchanged** — spans recorded, no `SampleRate` attribute |
| sampled=1 | `rate:N` | < 500 | Spans exported as today **plus** `SampleRate = N` on the root span and the `stream_chunk` child spans |
| sampled=1 | `rate:N` | >= 500 | Spans exported as today **plus** `SampleRate = 1` (error responses ignore the hint — mirrors the worker's keep-all-errors-at-rate-1 semantics) |
| sampled=0 | any | any | **Unchanged** — no spans exported |
| any | unparseable / `rate:<1` | any | Hint ignored — behaves as if no hint was sent |

## Implementation

- **`Electric.Plug.TraceContextPlug`** — after the existing `traceparent` extraction, parses the `electric=rate:N` member out of the (already-decoded) tracestate carried by the extracted span context, and stores `%{sample_rate_hint: N}` in `conn.private`. Exposes `trace_context/1` and `sample_rate_attrs/2` for downstream plugs.
- **`Electric.Plug.ServeShapePlug`** — `add_span_attrs_from_conn/1` (run at span start and again from `emit_shape_telemetry/1` once the final response status is known) merges the `SampleRate` attribute into the conn-derived attributes and stamps them onto the current root span via `add_span_attributes/1`. When the remote parent was not sampled the parent-based sampler left no recording span, so the stamp is a no-op and nothing is exported — the deliberate volume win for unsampled traffic.
- **`Electric.Shapes.Api.Response.send_stream/2`** — the per-request `shape_get.plug.stream_chunk` child spans (which survive the flattening in #4561 because they are 1:n, not 1:1) get the same `SampleRate` attribute at creation, from the stashed hint and the response status.

### Notes / scope decisions

- `Sampler`'s `exclude_spans` semantics and the flattening work from the base branch are untouched.
- Embedded usage (no `MIX_TARGET=application`, no OTel SDK): the propagator no-ops, so no trace context is ever stored and every path is unchanged.

### Tests

- Unit tests cover the tracestate parsing: valid `electric=rate:N` members, unparseable values, and `rate:<1` all map to the expected hint / ignored behavior.


packages/sync-service/lib/electric/telemetry/open_telemetry.ex
```diff
@@ -376,6 +384,22 @@ defmodule Electric.Telemetry.OpenTelemetry do
...
+  by the given span context.
+
+  Returns the member's value as a string, or `nil` when the tracestate has no such
+  member.
+  """
+  @spec tracestate_value(span_ctx(), String.t()) :: String.t() | nil
+  def tracestate_value(span_ctx, key) when Record.is_record(span_ctx, :span_ctx) do
+    case :otel_tracestate.get(key, span_ctx_record(span_ctx, :tracestate)) do
+      value when is_binary(value) and value != "" -> value
+      [_ | _] = value -> List.to_string(value)
```
magnetised:
```suggestion
      [_ | _] = value -> IO.iodata_to_binary(value)
```

just in case it's an improper list or something

magnetised (approved):

PR #4564, opened 2026-06-11
feat(agents-mobile): session sharing, access management & copy/share links (desktop parity)

msfstef:
## Summary

Brings session sharing to `agents-mobile`, closing the parity gap with the desktop `ShareEntityDialog` while using mobile-native UX paradigms. Mobile previously only *consumed* effective permissions to gate actions; it can now manage who has access to a session, copy the session id, and share a browser-openable session link.

Part of the agents-mobile ↔ agents-desktop parity track (follows #4553, #4546).

### 1. Share session screen (`ShareSessionScreen` + `app/session-share.tsx` modal route)
Opened from the session kebab menu's **Share** entry. Single-column layout, top to bottom:
- **Session title + copyable id** (tap id → copy, copy→check icon swap).
- **Session link pill** — abbreviated web URL (scheme stripped, middle-ellipsized) with a share glyph; one tap opens the **native OS share sheet** (which includes Copy, so there is no separate in-app "Copy link").
- **People with access** — pinned, non-removable **Owner** row derived from `entity.created_by` (omitted when null), then one row per grantee (avatar/initials, name/email, role pill). Desktop doesn't show the owner; mobile adds it so the list is never confusingly empty.
- **General access** — the workspace-wide *All users* grant (`subject_kind: principal_kind / user`) as its own section (Google-Drive style) rather than mixed into the people list like desktop.
- **Add people** — search-first user list (synced `users` table), excluding self/owner/existing grantees.
- **Role picker** — tapping any row opens a bottom sheet with **View / Chat / Manage** (same permission sets and eye/message/shield glyphs as desktop) plus a destructive **Remove access** (Alert-confirmed). Roles commit immediately per row — no desktop-style deferred Grant/Update button.

### 2. Copy session id
The session menu's status header now shows the id (in place of the agent type label) with a tap-to-copy affordance; same treatment on the long-press row sheet's id line. Mirrors the desktop `EntityHeader` copy-id pattern (1.2s icon swap, via the new `useCopyFeedback` hook + `expo-clipboard`).

### 3. Session web links
`sessionWebUrl(serverUrl, entityUrl)` builds `{serverUrl}/__agent_ui/#/entity/{id}`.

## Key decisions & why

| Decision | Rationale |
|---|---|
| Grants come from REST `GET /_electric/entities/:type/:id/grants`, not sync | The synced `entity_effective_permissions` shape is scoped to the **current principal only** (`server-utils.ts` `buildCurrentPrincipalEntityEffectivePermissionsWhere`), so it cannot list other people's access. Same approach as desktop's `loadGrants()`. Refetch-after-mutation (no optimistic grant state) because the server owns grant ids needed for subsequent diffs. |
| Web URL targets `/__agent_ui/` directly, not the server root | The root `302` redirect uses an absolute `location: /__agent_ui/`, which would drop a Cloud `/t/<service-id>/v1` tenant prefix. The session id stays **un-encoded** to match the web UI's hash splat route (`/entity/$`). |
| Share screen reachable by **everyone** (no `manage` gate on the menu entry) | Link sharing must work for non-managers. The grant sections self-gate: the manage-protected endpoint returns 401/403 → the screen shows a manage-required message (this also gracefully handles revoking your own manage access mid-session). |
| Single share affordance (link pill → native sheet), no in-app Copy link | The native sheet includes Copy; two buttons for the same URL was duplicative. The pill shows the URL being shared (Meet/Zoom pattern). In-app clipboard copy remains only for the session **id**. |
| One **Share** menu item next to Pin | Earlier iteration had Copy id / Copy link / Share… / Share & access — collapsed after review feedback. |
| `userDisplay()`/`initials()` extracted to `agents-server-ui/src/lib/userDisplay.ts` | Mobile deep-imports shared lib code (existing pattern: `principals`, `sharePermissions`, `entity-api`, `auth-fetch`); extraction avoids drift vs. duplicating. Internal lib file only — no public API change, desktop dialog updated to import it. |
| Inbound links **out of scope** | Universal links (https → app) need associated-domain setup; planned follow-up. The web URL format was chosen so those links can later deep-link into the app once a domain is verified. |

## Architecture notes (for future sessions on this PR)

- `src/lib/entityGrants.ts` — all grant logic, pure where possible: REST wrappers (`listEntityGrants`/`createEntityGrant`/`deleteEntityGrant`, errors throw `GrantsRequestError` carrying `status`), `diffGrantsForRole` (set-based create/delete diff, ignores non-share custom permissions, handles duplicate rows), `grantIdsForRemoval`, `buildShareAccessModel` (groups grants → all-users + per-user entries with roles, drops current user / system principals / role-less subjects), `setSubjectRole` / `removeSubjectAccess` orchestration. Unit-tested with the repo's hoisted `serverFetch` mock pattern.
- `src/lib/sessionLinks.ts` — `sessionIdFromEntityUrl`, `sessionWebUrl` (pure, unit-tested incl. Cloud prefix + malformed-URL fallback).
- Role semantics live in shared `agents-server-ui/src/lib/sharePermissions.ts`: View=[read,fork], Chat=[read,write,signal,fork,schedule,spawn], Manage=[manage,delete]; `roleFromGrants` / `rolePermissionsMatchGrants` reused.
- `SessionMenu` takes an optional `onShare` prop threaded from `app/session.tsx` → `SessionScreen` (router push to `/session-share?entityUrl=…`).
- `BottomSheetItem` always reserves a 22px icon slot — give items icons or they look indented (why the role options carry glyphs).
- Search fields must mirror `SearchBar` metrics (fixed 36px height, zero-padding input) or the placeholder sits off-centre on Android.

## Testing

- ✅ 22 new vitest tests (`sessionLinks.test.ts`, `entityGrants.test.ts`) — written test-first; full mobile suite 83/83, `agents-server-ui` 88/88, both packages typecheck clean.
- ⬜ Manual QA needed (see below).

### Manual QA checklist
- [ ] **Rebuild dev clients** — `expo-clipboard` is a new native module; old dev binaries won't load this JS (the one build-impacting change).
- [ ] Open a copied/shared link in a browser against a **local** server and a **Cloud** server (verifies `/__agent_ui/#/entity/{id}` resolves under the `/t/<svc>/v1` prefix — the one assumption not verifiable statically).
- [ ] iOS + Android: copy id from menu + long-press sheet; link pill opens native sheet; cancel is silent.
- [ ] Manager flow: add user → change role View↔Chat↔Manage → remove → All-users grant add/change/remove; cross-check grants in the desktop ShareEntityDialog.
- [ ] Non-manager: share screen shows link actions + manage-required message; no crash.
- [ ] Self-revocation: remove the All-users *Manage* grant that granted you manage → screen degrades to access-denied state.

## Out of scope / follow-ups
- Universal links / inbound https session links opening the app (needs associated domains + per-deployment config).
- Haptic feedback on copy (would add `expo-haptics`).
- Sharing for non-session entity types (no mobile detail surfaces exist for them yet).

### DIFF
--- a/.changeset/agents-mobile-session-sharing.md
+++ b/.changeset/agents-mobile-session-sharing.md
@@ -1,0 +1,12 @@
+---
+"@electric-ax/agents-mobile": patch
+"@electric-ax/agents-server-ui": patch
+---
+
+Bring session sharing to mobile (desktop `ShareEntityDialog` parity, mobile-first UX):
+
+- **Share session screen.** A modal route opened from the session menu's new **Share** entry. It exposes a link pill (abbreviated session web URL — one tap opens the native OS share sheet, which includes Copy), a "People with access" list with a pinned Owner row, a Google-Drive-style "General access" section for the workspace-wide *All users* grant, and a search-first "Add people" section. Roles (View / Chat / Manage, same permission sets and glyphs as desktop) commit per row through a bottom-sheet picker with a destructive *Remove access* action — no deferred Grant/Update button. The grant list comes from the manage-protected REST `GET /grants` endpoint (the synced effective-permissions shape is scoped to the current principal, so it can't list other people's access); non-managers still get the link actions and see a manage-required message below.
+- **Copy session id.** The session menu's status header and the long-press row sheet now render the id with a tap-to-copy affordance (copy→check icon swap, mirroring the desktop entity header), via a new `expo-clipboard` dependency.
+- **Session web links.** `sessionWebUrl()` builds `{serverUrl}/__agent_ui/#/entity/{id}` directly — targeting the web UI path rather than the server root, whose absolute-path redirect would drop a Cloud `/t/<service-id>/v1` tenant prefix.
+
+The desktop dialog's `userDisplay()`/`initials()` helpers move into `agents-server-ui`'s `lib/userDisplay.ts` so mobile deep-imports them instead of duplicating. Grant-diffing, removal, and access-model grouping logic is ported into a pure, unit-tested `entityGrants` module. No server API changes.
--- a/packages/agents-mobile/app/session-share.tsx
+++ b/packages/agents-mobile/app/session-share.tsx
@@ -1,0 +1,27 @@
+import { Stack, useLocalSearchParams, useRouter } from 'expo-router'
+import { ShareSessionScreen } from '../src/screens/ShareSessionScreen'
+import { useAgentsRouteGuard } from '../src/lib/useAgentsRouteGuard'
+
+export default function SessionShareRoute(): React.ReactElement | null {
+  const params = useLocalSearchParams<{ entityUrl?: string }>()
+  const router = useRouter()
+  const guard = useAgentsRouteGuard()
+  if (guard) return guard
+
+  const entityUrl = Array.isArray(params.entityUrl)
+    ? params.entityUrl[0]
+    : (params.entityUrl ?? ``)
+
+  return (
+    <>
+      <Stack.Screen options={{ presentation: `modal` }} />
+      <ShareSessionScreen
+        entityUrl={entityUrl}
+        onBack={() => {
+          if (router.canGoBack()) router.back()
+          else router.replace(`/`)
+        }}
+      />
+    </>
+  )
+}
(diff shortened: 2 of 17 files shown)


msfstef:
https://github.com/user-attachments/assets/d13ca285-362f-4821-b0ac-06c1d1f97d74

kevin-dp (approved):

Issue #4563, opened 2026-06-11
Unhandled inspector errors crash ShapeLogCollector; ReplicationClient then logs a full error every 50ms for up to 10 minutes


erik-the-implementer:
## Summary

In Electric Cloud production (observed across sync-service 1.6.0–1.6.10) we see a steady stream of errors of the form:

```
Error processing replication event (NNNms retry budget left): ...
Error dispatching replication event (NNNms retry budget left): ** (exit) exited in: Electric.Replication.ShapeLogCollector.handle_event_async(...)
    ** (EXIT) no process: the process is not alive ...
```

Investigation shows these are one mechanism with several underlying gaps:

1. The hot replication path (`ShapeLogCollector` → `Partitions` → `Inspector`) does synchronous Postgres introspection but only handles one error value, `{:error, :connection_not_available}`. Every other legal return of the inspector contract crashes the `ShapeLogCollector`.
2. The SLC crash takes down the whole `Electric.Shapes.Supervisor` (`:one_for_all`), and while it restarts (which includes shape restore, potentially seconds to minutes), `ReplicationClient` retries the pending event **every 50 ms**, logging a full error — including the entire inspected event payload — on every attempt. A single incident can emit up to ~12,000 error log lines (10-minute budget / 50 ms), which flood logs and any error tracker attached to the logger.

### 1. `Partitions.handle_relation` is missing clauses for `:table_not_found` and generic `{:error, reason}`

`packages/sync-service/lib/electric/shapes/partitions.ex` — the `case Inspector.load_relation_info(...)` in `handle_relation/2` (~L126) only matches `{:ok, %{parent: ...}}`, `{:ok, _}`, and `{:error, :connection_not_available}`.

But the inspector legitimately returns:

- bare `:table_not_found` — e.g. when a table is dropped/renamed (or dropped+recreated with a new OID) between the WAL record being written and Electric introspecting it, such as while replaying backlog after a reconnect. Observed: `{:case_clause, :table_not_found}` crashes. Note `add_shape/3` in the same module *does* handle `:table_not_found` (~L67), so this looks like an omission.
[...]

PR #4565, opened 2026-06-11
Don't crash ShapeLogCollector on introspection failures in the replication path

erik-the-implementer:
Fixes the crash family from #4563: the replication hot path (`Partitions`, pk-column lookup, shape restore) only handled `{:error, :connection_not_available}` from the inspector, so every other legal inspector return crashed the ShapeLogCollector, took down the `:one_for_all` shapes supervision tree, and put the replication client into its 10-minute noproc retry loop.

## Changes

- **`Partitions.handle_relation/2`**: handles `:table_not_found` (table dropped/renamed between the WAL record and introspection — nothing left to track, mirroring the existing `add_shape` behavior) and propagates other errors instead of crashing with a `CaseClauseError`.
- **`Partitions.add_shape/3`**: returns `{:error, reason}` instead of raising a `RuntimeError` on unexpected introspection errors; the shape-registration path propagates the error to the caller.
- **`pk_cols_of_relation`**: when the table was dropped before introspection, changes are keyed on the full record (the same fallback used for tables without a primary key) instead of crashing in `Utils.map_while_ok`; handling of the dropped table is left to the affected shapes further down the stack.
- **ShapeLogCollector**: unexpected introspection failures in the relation and transaction paths log a warning with the real reason and reply with `{:error, :connection_not_available}` — the one error the replication client knows how to recover from — pausing replication until introspection succeeds instead of crashing the collector.
- **Shape restore**: retries introspection in place (100ms × 100) when the connection pool isn't available instead of crashing with a `MatchError` in a restart loop. Skipping the shape was not an option: nothing re-registers restored shapes later, so a skipped shape would silently stop receiving updates. If the error persists past ~10s the restore still fails, but with a descriptive error.
- **`DirectInspector`**: connection-class errors returned in-band by Postgrex (e.g. `"ssl recv: closed"`) are classified as `:connection_not_available` via the existing `Electric.DbConnectionError` classifier, so they take the established recovery path instead of leaking through as strings.

## Production context

In Electric Cloud these crashes account for most of the "retry budget" Sentry noise: `{:case_clause, :table_not_found}`, `{:case_clause, {:error, "ERROR 53200 (out_of_memory) ..."}}`, `query_canceled`, `"ssl recv: closed"`, the restore `MatchError`, and — downstream of all of them — thousands of `(EXIT) no process` dispatch errors per incident while the supervision tree restarts. See #4563 for the breakdown.

## Notes for review

- All new behavior is covered by tests written before the fixes (9 new tests across `partitions_test`, `shape_log_collector_test` and a new `direct_inspector_test`).
- One known rough edge: a persistently failing introspection now produces one ShapeLogCollector warning per replication-client retry cycle (throttled only by the introspection query duration). Rate-limiting that warning could be a follow-up; the companion PR throttles the replication client's own retry logging.

### DIFF
--- a/.changeset/replication-introspection-error-handling.md
+++ b/.changeset/replication-introspection-error-handling.md
@@ -1,0 +1,12 @@
+---
+"@core/sync-service": patch
+---
+
+Stop the ShapeLogCollector from crashing when table introspection fails on the hot replication path.
+
+Previously the replication pipeline only handled `{:error, :connection_not_available}` from the inspector; every other legal inspector return crashed the ShapeLogCollector and took the whole shapes supervision tree down with it, putting the replication client into a 10-minute noproc retry loop:
+
+- `:table_not_found` (table dropped or renamed between the WAL record and introspection) crashed `Partitions.handle_relation/2` and the pk-column lookup with a `CaseClauseError`. Relation messages for dropped tables are now ignored and changes for dropped tables are keyed on the full record, the same fallback used for tables without a primary key.
+- In-band database errors (e.g. out-of-memory, `statement_timeout` cancelling the catalog query) crashed `Partitions.handle_relation/2` with a `CaseClauseError` and made `Partitions.add_shape/3` raise. They are now propagated as `{:error, reason}`; the collector logs a warning and pauses replication via the existing retry path until introspection succeeds.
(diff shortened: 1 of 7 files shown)


robacourt (approved):
Nice!

PR #4567, opened 2026-06-11
fix(agents-server-ui): show uncached input tokens in the meta row

kevin-dp:
## Summary

Follow-up to #4502. The input side of the per-response token-usage label summed `input + cacheRead + cacheWrite`, but `cacheRead` re-counts the entire conversation history on every warm-cache step — so the label ballooned into a cumulative number that said nothing about the work the response actually did.

The adapter now surfaces only the **uncached** input side: fresh prompt tokens plus cache writes, with prompt-cache reads excluded. `cacheWrite` is counted because cache-enabled providers report newly appended prompt tokens there (with `input` collapsing to ~0); excluding it would surface tiny "3 ↑" labels instead. This matches the accounting the #4552 goal-budget progress bar uses, so the two displays stay consistent.

## Changes

- `pi-adapter.ts`: drop `usage.cacheRead` from the input sum (legacy flat `inputTokens` fallback unchanged — no cache split means the whole side counts as uncached)
- Comments updated across `outbound-bridge.ts`, `entity-schema.ts`, `entity-timeline.ts` and `TokenUsage.tsx` to document the uncached semantics of `input_tokens`
- Test updated: `input: 50, cacheRead: 1200, cacheWrite: 100` now persists `150`, not `1350`

## Notes

- Steps recorded before this change keep their stored cache-inclusive totals — both fields are optional and the display sums what's persisted, so no migration is needed.
- #4552 touches the same `pi-adapter.ts` block (it splits total vs. uncached for budget enforcement) and will need a small rebase after this merges.

### DIFF
--- a/.changeset/uncached-input-tokens.md
+++ b/.changeset/uncached-input-tokens.md
@@ -1,0 +1,19 @@
+---
+'@electric-ax/agents-server-ui': patch
+'@electric-ax/agents-runtime': patch
+---
+
+Show only uncached input tokens in the per-response token usage label.
+
+The input side previously summed `input + cacheRead + cacheWrite`, so
+on warm-cache turns the meta row re-counted the entire conversation on
+every step and ballooned into a cumulative number that said nothing
+about the work the response actually did. The adapter now surfaces the
+uncached side only — fresh prompt tokens plus cache writes, with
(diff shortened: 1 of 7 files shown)


samwillis (approved):
LGTM

PR #4569, opened 2026-06-11
feat(agents): embedder docker-sandbox options + modelCatalog on AgentHandlerResult

balegas:
## What

- `BuiltinAgentHandlerOptions.dockerSandbox?: { image, allowFloatingTag, env, extraMounts }` — threads into the built-in `docker` sandbox profile's `dockerSandbox()` call (which already supports all of these). Custom mounts append after the working-directory mount.
- `AgentHandlerResult.modelCatalog` — exposes the catalog `createBuiltinAgentHandler` already builds.
- Exports: `resolveBuiltinModelConfig`, `resolveDockerSandboxOpts`, and types `BuiltinModelCatalog`, `BuiltinAgentModelConfig`, `BuiltinDockerSandboxOptions`, `BuiltinDockerSandboxMount`.

## Why

OpenFactory registers a custom Discord-facing agent entity alongside horton/worker. It needs (a) a sandbox image with git + gh + a GitHub-App token helper baked in, plus the GH_APP env/key mount inside the container, and (b) the same model-arg resolution the built-in agents use, without re-implementing the catalog.

## Tests

`packages/agents/test/docker-sandbox-options.test.ts` covers the option merge (cwd-mount only, full passthrough, mount append, empty).

## Notes for reviewers

- `BuiltinDockerSandboxMount` intentionally duplicates the runtime's mount item shape instead of importing `DockerSandboxOpts` — that type only exists on the `@electric-ax/agents-runtime/sandbox/docker` subpath, which this package deliberately keeps out of its top-level import graph (docker code is reached via dynamic import only).
- `env` / `extraMounts` extend the existing embedder/operator trust boundary: same trust level as the embedder process itself, and `extraMounts` stays behind the runtime's docker-socket guard.

### DIFF
--- a/.changeset/openfactory-sandbox-options.md
+++ b/.changeset/openfactory-sandbox-options.md
@@ -6,4 +6,4 @@ Embedder customization hooks for the built-in agents:
 
 - `BuiltinAgentHandlerOptions.dockerSandbox` ({ image, allowFloatingTag, env, extraMounts }) threads into the built-in `docker` sandbox profile. These are embedder/operator-trust inputs: `extraMounts` is subject to the runtime's docker-socket guard and `env` is passed verbatim into the container.
 - `AgentHandlerResult.modelCatalog` exposes the resolved model catalog so embedders can register sibling agent types with the same model resolution.
-- New exports: `resolveBuiltinModelConfig`, `resolveDockerSandboxOpts`, and types `BuiltinModelCatalog`, `BuiltinAgentModelConfig`, `BuiltinDockerSandboxOptions`, `BuiltinDockerSandboxMount`.
+- New exports: `resolveBuiltinModelConfig`, and types `BuiltinModelCatalog`, `BuiltinAgentModelConfig`, `BuiltinDockerSandboxOptions`, `BuiltinDockerSandboxMount`.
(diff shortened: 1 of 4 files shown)


balegas:
The failing build was the **Prettier formatting** check (not a compile error): the review-fix added a drift-guard whose long `NonNullable<import(...)>` type line gets wrapped by Prettier, which then split the `import()` away from its `eslint-disable-next-line quotes` comment.

Fixed in 58d5492 by moving the disable comment directly above the wrapped `import()` line, so Prettier, ESLint (`quotes`), and `tsc` are all satisfied. Verified locally:

- `prettier --check packages/agents/src/bootstrap.ts` → clean
- `pnpm -C packages/agents stylecheck` → clean
- `pnpm -C packages/agents typecheck` → clean
- `pnpm -C packages/agents test docker-sandbox-options` → 4 passed

For context, the review suggestions were addressed in the prior commits: dropped `resolveDockerSandboxOpts` from the public `index.ts`/changeset (internal helper; the test imports from source), documented the `/work` mount-collision footgun, and added the compile-time mount-shape drift guard.

samwillis (approved):

PR #4573, opened 2026-06-11
Fix persisted run error rendering

KyleAMathews:
Fixes run error rendering in the agents timeline so handler failures stay attached to the failed run after a refresh. Long or multiline error messages now render using the same expandable event-card pattern as tool calls.

## Root Cause

Handler failures were written to the entity `errors` collection without a `run_id`:

```ts
entityStateSchema.errors.insert({
  value: {
    error_code: errCode,
    message: errMsg,
  },
})
```

The live UI could briefly surface the failure while the failed run was still in memory, but after a refresh the run row only reads errors associated with that run. Without `run_id`, the error became a standalone stream error and disappeared from the failed run’s footer.

Separately, the UI rendered the full error text inline. Provider/runtime errors can be long or multiline, so the footer could truncate important details with no way to expand them.

## Approach

- Snapshot the set of run keys before invoking each handler pass.
- If the handler fails, find the last run created during that pass and write the handler error with that run’s `run_id`.
- Reuse `InlineEventCard` for long or multiline run errors so they follow the same expandable pattern as tool calls:
  - short errors stay inline and truncated in the response metadata row
  - long or multiline errors render as a collapsible `run error` card with the full message in the card body

## Key Invariants

- A handler error should attach only to a run created during the same handler pass.
- Existing historical runs must not receive errors from later failed handler passes.
- Standalone/non-run errors should remain standalone when no run was created.
- Runtime-created run rows are appended sequentially, so the last newly-created run is the run to associate with a handler failure.
- The UI should preserve full error text while keeping the collapsed timeline readable.

## Non-goals

- This does not change standalone entity error rows in the timeline.
- This does not change model provider error classification or error-code generation.
- This does not introduce a second expandable styling system for run errors.

## Trade-offs

The runtime infers the failed run by comparing the pre-handler run-key snapshot to the post-failure run collection, rather than threading a current run id through every handler/agent path. That keeps the change localized to `processWake` and preserves existing agent/run APIs, at the cost of selecting the last newly-created run when a handler creates multiple runs before failing.

Using `InlineEventCard` makes long run errors visually heavier than a small inline footer, but it reuses the existing expandable UI pattern instead of adding bespoke details styling.

## Verification

```bash
pnpm install
pnpm --filter @electric-ax/agents-runtime exec vitest run test/process-wake.test.ts --reporter=dot
pnpm --filter @electric-ax/agents-server-ui run typecheck
```

## Files changed

- `packages/agents-runtime/src/process-wake.ts`
  - Tracks runs created during each handler pass and attaches handler errors to the last new run.

- `packages/agents-server-ui/src/components/AgentResponse.tsx`
  - Adds shared run error rendering for both live and cached agent responses.
  - Uses `InlineEventCard` for long or multiline error messages.

- `.changeset/run-error-rendering.md`
  - Adds patch changeset entries for the affected agents packages.

### DIFF
--- a/packages/agents-runtime/src/process-wake.ts
+++ b/packages/agents-runtime/src/process-wake.ts
@@ -146,6 +146,35 @@ function toError(err: unknown): Error {
   return err instanceof Error ? err : new Error(String(err))
 }
 
+function compareRunOrder(
+  left: { key: string; _timeline_order?: string; _seq?: number },
+  right: { key: string; _timeline_order?: string; _seq?: number }
+): number {
+  if (left._timeline_order && right._timeline_order) {
+    return left._timeline_order.localeCompare(right._timeline_order)
+  }
+  if (left._timeline_order) return 1
+  if (right._timeline_order) return -1
+
+  if (left._seq !== undefined && right._seq !== undefined) {
+    return left._seq - right._seq
+  }
+  if (left._seq !== undefined) return 1
+  if (right._seq !== undefined) return -1
+
+  return left.key.localeCompare(right.key)
+}
+
+function latestNewRunKey(
+  db: EntityStreamDBWithActions,
+  existingRunKeys: ReadonlySet<string>
+): string | undefined {
+  return db.collections.runs.toArray
+    .filter((run) => !existingRunKeys.has(run.key))
+    .sort(compareRunOrder)
+    .at(-1)?.key
+}
+
 async function resolveHeadersProvider(
   provider: ProcessWakeConfig[`claimHeaders`]
 ): Promise<Record<string, string> | undefined> {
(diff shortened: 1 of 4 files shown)


packages/agents-runtime/src/process-wake.ts
```diff
@@ -146,6 +146,15 @@ function toError(err: unknown): Error {
   return err instanceof Error ? err : new Error(String(err))
 }
 
+function latestNewRunKey(
+  db: EntityStreamDBWithActions,
+  existingRunKeys: ReadonlySet<string>
+): string | undefined {
+  return db.collections.runs.toArray
+    .filter((run) => !existingRunKeys.has(run.key))
+    .at(-1)?.key
```
samwillis: Could be a query once, but this is ok

samwillis (approved):
LGTM
one nit, but likely fine

PR #4575, opened 2026-06-11
fix: harden desktop send response parsing

KyleAMathews:
## Summary
- preserve configured auth headers when local Electron requests fall back from desktop IPC transport to browser fetch
- replace raw `res.json()` calls in desktop send/session mutation paths with optional/required JSON helpers
- add a regression test for local Electron FormData fallback header injection

## Testing
- pnpm --filter @electric-ax/agents-server-ui test -- --run src/lib/auth-fetch.test.ts

### DIFF
--- a/scripts/dev.sh
+++ b/scripts/dev.sh
@@ -373,8 +373,6 @@ cmd_isolated() {
   done
 
   cd "$REPO_ROOT"
-  [[ -f .env ]] || die ".env not found at repo root. Create one with ANTHROPIC_API_KEY or OPENAI_API_KEY."
-  grep -qE '^(ANTHROPIC_API_KEY|OPENAI_API_KEY)=' .env     || die ".env is missing ANTHROPIC_API_KEY or OPENAI_API_KEY."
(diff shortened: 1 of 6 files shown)


scripts/dev.sh
```diff
@@ -373,8 +373,6 @@ cmd_isolated() {
   done
 
   cd "$REPO_ROOT"
-  [[ -f .env ]] || die ".env not found at repo root. Create one with ANTHROPIC_API_KEY or OPENAI_API_KEY."
-  grep -qE '^(ANTHROPIC_API_KEY|OPENAI_API_KEY)=' .env     || die ".env is missing ANTHROPIC_API_KEY or OPENAI_API_KEY."
```
samwillis, line 377: Is this an intentional part of this PR?

samwillis (approved):
LGTM

scripts/dev.sh
KyleAMathews: yes — it was from back when we required a .env file

PR #4576, opened 2026-06-11
Fix pg-sync observations: required source URLs, per-principal identity, unobserve tool

KyleAMathews:
## Summary

pg-sync observations (Horton's `observe_pg_sync` tool) registered "successfully" but never delivered row changes to the agent. This PR makes them work end-to-end: an explicit Electric shape endpoint URL is now **required and validated at registration time**, source identity is **stable across wakes but correctly scoped per principal**, a duplicate registration no longer corrupts a running bridge's bootstrap state, and a new **`unobserve_pg_sync`** tool lets an agent stop observing without disturbing other observers.

## Root cause

Four defects made pg-sync observations silently dead or fragile:

1. **Per-request metadata polluted the source identity.** `sourceRefForPgSync()` hashed the *entire* canonical options object, including the `metadata` field (principal, tenant, `wakeId`, `runtimeConsumerId`). Since `wakeId` is unique per wake, every re-observation minted a brand-new `sourceRef` → a brand-new empty durable stream → a brand-new bridge that subscribed from `offset: now`. The change that triggered the wake lived on the *previous* wake's stream, so the agent's handle DB was always empty. Old bridges leaked, one ShapeStream per wake.

2. **Registration succeeded before anything touched Electric.** The server returned `200` and the tool reported success before the bridge ever connected. The real connection happened later inside the bridge's `ShapeStream` subscription, and any failure (wrong URL, unreachable host, nonexistent table) landed in an infinite `recoverStream()` backoff loop with only a server-side `warn`. A bare host URL like `http://localhost:30000` is especially deadly: Electric answers `200` with an empty body on its root path, so even an "ok" check passed while the shape API actually lives at `/v1/shape`.

*Found in external review:*

3. **A duplicate registration reset a running bridge's bootstrap state.** The registry upsert conflict path cleared `initialSnapshotComplete` to `false` while leaving the persisted `shapeHandle`/`shapeOffset` intact. Once `sourceRef` is stable across wakes (defect 1's fix), re-registration is the *common* path — so after a re-register + server restart, the bridge resumed from the saved cursor but with `skipChangesUntilUpToDate = true`, treating real changes that arrived during downtime as bootstrap rows and dropping them until the next `up-to-date`.

4. **Identity ignored the principal entirely.** The minimal fix for defect 1 stripped *all* metadata from the hash — but the live Electric request still forwards `electric_agents_principal_*`/`tenant`/`entity` params (sourced from the authenticated registrant). With identity blind to those, two different principals observing the same shape collapse onto one shared bridge bound to whoever registered first, making auth/scoping order-dependent.

## Approach

**Stable but principal-scoped identity.** `sourceRefForPgSync()` now folds identity-bearing metadata (tenant, principal, observing entity) into the hash and excludes only ephemeral per-request fields:

```ts
const EPHEMERAL_PG_SYNC_METADATA = new Set([`wakeId`, `runtimeConsumerId`, `streamPath`])
// hash(shape options + identity metadata, minus ephemerals)
```

The same agent reuses one bridge/stream across its wakes (its principal/entity are stable; `wakeId` is excluded), while distinct principals each get their own correctly-scoped, individually-probed stream. The server is authoritative: it recomputes `sourceRef` with the authenticated principal and rebinds the agent's manifest entry + wake to the returned `sourceRef`/`streamUrl` (`withRegisteredManifestEntry`).

**Registration-time probe.** `PgSyncBridgeManager.register()` fetches the shape log once *before* persisting a registry row or starting a bridge:

```ts
const probeUrl = buildShapeProbeUrl(source.url, options) // offset=now, same param encoding as the bridge
const response = await fetchFn(probeUrl, { signal: AbortSignal.timeout(this.probeTimeoutMs) })
if (!response.ok) throw new PgSyncSourceValidationError(/* surfaces Electric's status + body */)
if (!response.headers.get('electric-handle')) throw new PgSyncSourceValidationError(/* suggests /v1/shape */)
```

A genuine shape response always carries the `electric-handle` header — that's what distinguishes a real shape endpoint from Electric's root path. `PgSyncSourceValidationError` maps to HTTP `400` in the router, so the failure reaches the agent as actionable feedback rather than dying in a retry loop. The probe runs only for not-yet-running sources, so the hot path stays a single round-trip.

**Cursor preservation.** The upsert conflict path no longer resets `initialSnapshotComplete`. A `sourceRef` conflict means the shape-identity options are byte-identical, so the persisted cursor and bootstrap state are always still valid — clearing them was unconditionally wrong.

**Unobserve.** `unobserve_pg_sync` drops *only this agent's wake*: it writes a `delete` manifest event for `source:pgSync:<sourceRef>` and calls `wakeRegistry.unregisterByManifestKey` (via `entityManager.deletePgSyncObservation` → `DELETE /pg-sync-observations/:sourceRef` → `serverClient.removePgSyncObservation` → new `ctx.unobserve`). The shared bridge is keyed by `sourceRef`, not by subscriber, so it keeps running for any other observers. Mirrors the existing webhook-source unsubscribe path.

**Prompt guidance.** The Horton system prompt and tool descriptions tell the model the `url` is an HTTP(S) shape endpoint (with example), not a `postgres://` string, that there's no default, and to ask the user rather than guess — plus how to stop with `unobserve_pg_sync`.

## Key invariants

- Same shape **and** same principal ⇒ same `sourceRef` ⇒ same bridge/stream, across wakes.
- Different principals ⇒ different `sourceRef` ⇒ isolated, individually-probed streams (no order-dependent scoping).
- A successful registration implies the endpoint was reachable and returned a real shape log (had `electric-handle`).
- A `sourceRef` upsert conflict never invalidates the persisted cursor/bootstrap state.
- No partial registration: a failed probe writes no registry row, ensures no stream, constructs no `ShapeStream`.
- `unobserve` removes only the calling entity's wake; the bridge survives for other observers.

## Non-goals

- **No backwards compatibility.** The URL was previously optional with a server-side default; that default is removed. Registry rows persisted before this change (without a `url`) can never resume, so server startup now *deletes* them with one clear log line instead of warn-spamming every boot — affected agents must re-observe.
- **No cross-principal stream sharing (YAGNI).** Per-principal streams are the safe default and never *incorrect*, only potentially redundant when distinct principals observe a byte-identical shape (likely rare). If that ever costs something, the planned extension is an opt-in `scope: 'principal' | 'tenant'` flag — `'tenant'` would hash on shape + tenant only and drop principal from the live params, which is only safe if the Electric proxy does not scope rows per principal. Deferred until there's a measured need.
- The probe is a registration-time liveness/validity check, not a permanent guarantee — the bridge's existing retry loop still handles Electric going down later.

## Trade-offs

- **Per-principal stream fan-out.** Distinct principals observing an identical shape get separate Electric streams. Bounded: a given agent always reuses its own stream across wakes, so extra streams only appear across *different* principals with byte-identical shapes. The safe-by-default choice; sharing is the opt-in optimization above.
- **Probe param encoding is a parallel implementation.** `buildShapeProbeUrl` re-derives the Electric TS client's query encoding (comma-joined arrays, `params[n]`). Unlike the client it does not quote column identifiers, so probe and stream encoding can diverge for exotic column names. Acceptable because the probe is a validity check, not the data path.
- **`electric-handle` heuristic.** A proxy/CDN that strips response headers would fail the probe even if the shape works. Low risk, and the error points at `/v1/shape`.

## Verification

```bash
# Identity scoping (per-principal, ephemeral-insensitive) + cursor preservation
pnpm -F @electric-ax/agents-runtime exec vitest run test/pg-sync-source.test.ts
pnpm -F @electric-ax/agents-server  exec vitest run test/pg-sync-bridge-manager.test.ts test/pg-sync-router.test.ts

# Unobserve tool + route
pnpm -F @electric-ax/agents         exec vitest run test/observe-pg-sync-tool.test.ts
pnpm -F @electric-ax/agents-server  exec vitest run test/webhook-source-subscriptions-route.test.ts

# Typecheck
pnpm -F @electric-ax/agents-server -F @electric-ax/agents-runtime -F @electric-ax/agents exec tsc --noEmit
```

The cursor-preservation and per-principal-identity fixes were each verified red→green: a real-Postgres registry test fails on the old reset, and `sourceRefForPgSync` returns identical hashes for distinct principals before the fix. Manually verified end-to-end against a live Electric instance: a bare URL (`http://localhost:30000`) fails fast with a `/v1/shape` suggestion, the agent retries with the correct URL, registration succeeds, and row changes flow through to wakes.

## Files changed

**Identity & bridge**
- `agents-runtime/src/observation-sources.ts` — `sourceRefForPgSync` hashes shape + identity metadata, excluding ephemeral fields (`wakeId`, `runtimeConsumerId`, `streamPath`).
- `agents-server/src/entity-registry.ts` — stop resetting `initialSnapshotComplete` on upsert conflict; `deletePgSyncBridge`.
- `agents-server/src/pg-sync-bridge-manager.ts` — `buildShapeProbeUrl`, `probeSource`, `PgSyncSourceValidationError`, `electric-handle` check; delete legacy url-less rows at startup.
- `agents-server/src/routing/pg-sync-router.ts` — require `url`; map `PgSyncSourceValidationError` → 400; log unexpected 500s.
- `agents-runtime/src/process-wake.ts` — `withRegisteredManifestEntry` rebinds entities/pgSync sources to their server-assigned ref/streamUrl.
- `agents-server/src/manifest-side-effects.ts`, `wake-registry.ts`, `agents-runtime/src/setup-context.ts`, `types.ts` — resolve pgSync stream URL from `config.streamUrl`; propagate `streamUrl`; map `old_value`.

**Unobserve tool (new)**
- `agents/src/tools/unobserve-pg-sync.ts` — `unobserve_pg_sync`: unobserve by `sourceRef`/`table`, or list active observations.
- `agents-server/src/entity-manager.ts` — `deletePgSyncObservation` (manifest delete + `unregisterByManifestKey`).
- `agents-server/src/routing/entities-router.ts` — `DELETE /pg-sync-observations/:sourceRef`.
- `agents-runtime/src/runtime-server-client.ts` — `removePgSyncObservation`.
- `agents-runtime/src/context-factory.ts`, `types.ts` — `ctx.unobserve`.
- `agents/src/agents/horton.ts` — register the tool; "Observing Postgres tables" prompt (observe + unobserve).

**Tests** — per-principal/ephemeral identity, real-DB cursor preservation, probe failure taxonomy, unobserve tool resolution + DELETE route, router 400/500 mapping.


packages/agents-server/src/pg-sync-bridge-manager.ts
```diff
@@ -478,6 +529,9 @@ export class PgSyncBridgeManager implements PgSyncBridgeCoordinator {
     const resolvedSource = this.resolveSource(canonicalOptions)
     const sourceRef = sourceRefForPgSync(canonicalOptions)
     const streamUrl = getPgSyncStreamPath(sourceRef, this.registry?.tenantId)
+    if (!this.bridges.has(sourceRef) && !this.starting.has(sourceRef)) {
+      await this.probeSource(resolvedSource, canonicalOptions)
+    }
     const row = await this.registry?.upsertPgSyncBridge({
```
samwillis, line 535: This still upserts the bridge row even when `this.bridges` already has `sourceRef`. In `PostgresRegistry.upsertPgSyncBridge`, the conflict path resets `initialSnapshotComplete` to `false` while leaving any persisted `shapeHandle`/`shapeOffset` intact. After a duplicate registration and server restart, `PgSyncBridge.start()` resumes from that cursor but passes `skipChangesUntilUpToDate = true`, so changes that happened after the saved cursor can be treated as bootstrap rows and dropped until the first `up-to-date`.

With this PR, duplicate registration is the expected path because `sourceRef` is stable across wake metadata. Could we avoid rewriting `initialSnapshotComplete` for an existing running bridge, or only clear it when we also clear the stored cursor / rebuild the bridge?

packages/agents-runtime/src/observation-sources.ts
```diff
@@ -107,7 +107,10 @@ export function canonicalPgSyncOptions(
 }
 
 export function sourceRefForPgSync(options: PgSyncOptions): string {
-  return hashString(JSON.stringify(canonicalPgSyncOptions(options)))
+  // metadata is per-request context (principal, wakeId, ...) and must not
+  // change the identity of the observed shape.
+  const { metadata: _metadata, ...identity } = canonicalPgSyncOptions(options)
```
samwillis: Making the identity ignore `metadata` is right for `wakeId`, but the server still passes metadata into `buildElectricShapeParams()` for both the probe and the `ShapeStream`. That means the first registration to start a bridge fixes the long-lived Electric request's `electric_agents_*` params, while later registrations with different principal/entity metadata reuse that same bridge and skip the probe.

If the Electric proxy uses those params for auth, tenant scoping, filtering, or audit decisions, behavior becomes order-dependent. I think we either need to keep only truly non-semantic metadata out of the identity while identity-relevant metadata remains part of it, or stop forwarding per-request metadata into the shared ShapeStream request.

samwillis (commented):
## Interactive GPT Review

Reviewed interactively with GPT against PR #4576 in an isolated worktree.

I found two issues that look worth addressing before merge:

1. Re-registering an already-running pg-sync bridge can reset persisted bootstrap state while preserving the cursor, which can cause a restart to drop real changes until `up-to-date`.
2. `sourceRef` now ignores request metadata, but the live Electric request still includes metadata-derived params, so bridge reuse can make authorization/scoping behavior depend on whichever registration started the bridge first.

I left inline comments on the relevant code paths.

packages/agents-server/src/pg-sync-bridge-manager.ts
KyleAMathews: Good catch — fixed. The `onConflictDoUpdate` set in `PostgresRegistry.upsertPgSyncBridge` no longer clears `initialSnapshotComplete` (it only touches `options`/`streamUrl`/timestamps now). The reasoning: a `sourceRef` conflict means the shape-identity options are byte-identical, so the persisted `shapeHandle`/`shapeOffset` and bootstrap flag are always still valid — resetting the flag while keeping the cursor was unconditionally wrong, and exactly the `{cursor present, initialSnapshotComplete=false}` corruption you described.

Verified red→green with a real-Postgres test (`pg-sync bridge registry (real DB)` in `pg-sync-bridge-manager.test.ts`): upsert → commit cursor with `initialSnapshotComplete=true` → re-upsert → assert the cursor *and* the flag survive. Fails on the old code, passes now. (`590e83194`/current `569f47987`.)

packages/agents-runtime/src/observation-sources.ts
KyleAMathews: Agreed, and fixed in that direction. `sourceRefForPgSync` now keeps identity-relevant metadata *in* the hash and excludes only genuinely-ephemeral fields:

```ts
const EPHEMERAL_PG_SYNC_METADATA = new Set([`wakeId`, `runtimeConsumerId`, `streamPath`])
// hash(shape options + identity metadata, minus ephemerals)
```

So tenant/principal/entity now affect identity: different principals get distinct, **individually-probed** bridges (each shape request authorized under its own credentials), while the same agent still reuses one bridge across wakes since `wakeId` is excluded — which keeps the original re-registration fix intact. That removes the order-dependence you flagged: no bridge is ever bound to "whoever registered first."

We deliberately did **not** take the other branch (stop forwarding per-request metadata into the shared ShapeStream). Per-principal is the safe default regardless of how the Electric proxy scopes rows; cross-principal sharing would only be correct if the proxy *doesn't* scope by principal, and we did not want to bet on that. If the fan-out (distinct principals observing byte-identical shapes) ever costs something, the planned extension is an opt-in `scope: 'principal' | 'tenant'` flag rather than changing the default — noted as a non-goal in the PR description (YAGNI for now).

Verified red→green: `sourceRefForPgSync({...base, metadata:{principalId:'a'}})` vs `{principalId:'b'}` hashed identically before, differ now; the ephemeral fields still collapse to the same ref (`sourceRefForPgSync identity metadata` tests in `pg-sync-source.test.ts`).

samwillis (approved):

PR #4577, opened 2026-06-15
feat(agents-server-ui): hide token usage label below a threshold

kevin-dp:
## Summary

Follow-up to #4567. The per-response token-usage label in the meta row is shown on every step, including trivial ones — a quick tool-only step or a one-line reply renders noise like \`47 ↑ 12 ↓\` that says nothing useful.

This gates the label on the **combined** input + output count: it's only rendered when the total is at or above a threshold. The cutoff lives in a single named constant, `SHOW_USAGE_THRESHOLD` (currently `1000`), so it's trivial to tune later.

\`1000\` is a deliberate choice — it's the same point where `formatTokenCount` switches from raw digits to compact `1.2k` notation, so every label we *do* render reads in the same compact style. Thresholding the combined total (rather than per-side) matters because after #4567 the input side collapses to ~0 on warm-cache turns, with all the signal in `output`.

## Changes

- `TokenUsage.tsx`: add `SHOW_USAGE_THRESHOLD` constant and an early `return null` when `(input ?? 0) + (output ?? 0)` is below it.

## Notes

- Stacked on #4567 (`kevin/uncached-input-tokens`) since both edit `TokenUsage.tsx`. Retarget to `main` once #4567 merges.
- No test added: this package has no DOM test env (jsdom/testing-library aren't deps) and no component tests — consistent with #4567 placing its test in `agents-runtime`. Happy to extract the gate into a pure `shouldShowUsage` helper and unit-test it if preferred.

### DIFF
--- a/.changeset/token-usage-threshold.md
+++ b/.changeset/token-usage-threshold.md
@@ -1,0 +1,9 @@
+---
+'@electric-ax/agents-server-ui': patch
+---
+
+Hide the per-response token-usage label when the combined input + output
+count falls below a threshold (`SHOW_USAGE_THRESHOLD`, currently 1000).
+Tiny tool-only steps and one-line replies no longer clutter the meta row
+with noise like `47 ↑ 12 ↓`; the threshold lives in a single constant so
+it's easy to tune.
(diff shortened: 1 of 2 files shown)


samwillis (approved):
LGTM

Issue #3675, opened 2026-01-06
Support multiple subqueries at the same subquery level (e.g. x IN subquery OR y IN subquery)


robacourt:
## Problem
  We want to support shapes with multiple subqueries at the same **subquery level** (outer query = level 1, subquery = level 2, nested subquery = level 3, etc). Example: `x IN (subquery) OR y IN (subquery)` where
  both subqueries are level 2. Today this isn’t possible because:

  - `DependencyLayers` assumes a shape has 0 or 1 dependency handle, so a shape with multiple same-level subquery dependencies can’t be layered.
    `lib/electric/shapes/dependency_layers.ex`
  - Move-in/out handling assumes a single subquery/tag. There are TODOs and explicit single-subquery assumptions in:
    - `lib/electric/shapes/consumer/move_ins.ex`
    - `lib/electric/shapes/consumer/materializer.ex`
    - `lib/electric/shapes/shape/subquery_moves.ex`
  - If we allow multiple subqueries at the same subquery level, we also need to detect that situation and 409 on any move-ins/outs for that shape (see `lib/electric/shapes/consumer.ex`).

  ## Current behavior
  - Shapes with multiple same-level subqueries don’t layer correctly (only the first dependency is handled).
  - Move-in/out handling is only correct for a single subquery dependency.

  ## Proposed solution
  1) **Support multiple dependency handles in `DependencyLayers`.**
     Replace the current single-dep recursion with a max-depth layering strategy:
     - Compute layer index as `1 + max(layer(dep_handle))` for all dependencies of the shape.
     - Insert the shape handle into that layer.
     - Keep `get_for_handles/2` behavior unchanged.

  2) **Detect “multiple subqueries at the same subquery level.”**
     During shape construction (or where clause parsing), walk the `where` AST and detect if there are 2+ subqueries at the same level (e.g., multiple `sublink_membership_check` nodes whose parent query level is
  equal). Flag the shape (e.g. `shape.same_level_subqueries? = true` or `shape.subquery_level_counts`).

  3) **Short-term safety: invalidate on move-ins/outs if flagged.**
     In `Consumer.handle_info({:materializer_changes, ...})`, if the shape is flagged as having multiple same-level subqueries, return 409 (stop + clean) on any move-in/out. This avoids incorrect incremental
  handling until multi-tag support exists.

  ## Acceptance criteria
  - Detection of multiple same-level subqueries is reliable.
[...]

PR #4579, opened 2026-06-15
Fix same-table subquery dependency deduplication

robacourt:
This PR fixes the bugs with where clauses such as `id IN (SELECT a FROM rel ...) OR id IN (SELECT b FROM rel ...)` where both subqueries are treated as the same even though there selected columns differ, as raised by [vibl](https://github.com/electric-sql/electric/issues/3675#issuecomment-4701113417)

## Summary
- include explicitly selected columns in shape comparability so same-table subqueries projecting different PK columns remain distinct dependencies
- bump the shape status version because shape hash semantics changed

### DIFF
--- a/.changeset/fix-subquery-dependency-dedup.md
+++ b/.changeset/fix-subquery-dependency-dedup.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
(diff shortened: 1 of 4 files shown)


alco (approved):
👍

PR #4582, opened 2026-06-15
feat(agents-mobile): bring fork capability to parity with desktop

msfstef:
## Summary

Brings the **fork** capability on agents-mobile to parity with desktop. Mobile previously had no native fork action, and the embedded per-message "Fork from here" button rode the WebView's `fetch` (slow, and double-tappable into multiple stray forks). This PR adds both desktop fork affordances to mobile and hardens the existing button.

## What changed

**1. Native whole-subtree fork (the kebab "Fork subtree")**
- `agentsClient.ts`: `forkEntity` (POST `…/fork`, empty body = whole-subtree HEAD clone) + a `ForkPointer` type, built on the shared `entityApiUrl` helper. Exposed through `AgentsProvider`. `SESSION_PERMISSIONS` gains `fork`.
- `SessionScreen`/`SessionMenu`: a gated **Fork subtree** kebab item — root-only (`!entity.parent`), disabled when terminal (stopped/killed) or lacking the `fork` permission, with a single-flight guard, a pending spinner, an inline error, and navigation to the new root on success. Adds a `git-fork` glyph to the native `Icon` set.

**2. Route the embedded per-message button through native**
- `createForkEntity` gains an optional transport; the embed injects a marshalled `onRequestForkEntity` so the per-message fork's POST runs over native RN networking instead of the WebView's `fetch`. The shared failure-toast behaviour is preserved by keeping the wrapper in `createForkEntity`.

**3. Harden the embedded button**
- Extracted a small, tested `singleFlight` primitive; added a spinner + disabled state on the button; mounted a `ToastProvider` inside the embed so fork failures surface in the WebView (previously they fired into a bus with no listener and vanished).

## Key decisions

- **One mobile kebab covers both desktop whole-subtree entry points.** Desktop exposes whole-subtree fork from the tile menu *and* the ⌘K command palette; mobile has neither concept, so the single kebab "Fork subtree" provides the capability. The per-message pointer fork comes free via the shared timeline embed.
- **Why route the embed fork through native at all?** The mobile timeline is the *same* `agents-server-ui` `ChatLogView` rendered in an Expo DOM WebView; the per-message fork is the **only** mutation that originates *inside* the WebView. Every other mobile mutation (send, signal/stop, steer/edit/delete/reorder, spawn, and the kebab fork) originates in native chrome and already runs over RN networking. So routing the button native makes the embed's lone write *consistent* with everything else (and mirrors how desktop routes embed POSTs through Electron IPC) — leaving it on WebView `fetch` would be the inconsistent split. This was the conclusion of an explicit altitude review; it holds independent of the (locally-shaky) latency argument.
- **Failure UX** matches each surface's convention: the per-message button shows a toast (desktop parity); the kebab shows an inline error, consistent with how mobile already surfaces `signalError`.
- **Reuse over re-implementation**: `forkEntity` uses the shared `entityApiUrl` (matching desktop `createForkEntity` and mobile `entityGrants.ts`); `singleFlight` is extracted and shared with the desktop button.
- Mobile adds a **single-flight guard the desktop fork lacks**, since the whole-subtree fork can take several seconds (it waits for descendants to idle).

## Architecture notes (for future sessions)

- **Mobile = native chrome + a timeline-only WebView.** `SessionScreen` (RN) owns the header, composer, and kebab; the WebView embed (`SessionChatLogDomEmbed` → `EmbedApp` → `ChatLogView`) renders only the timeline. Message *rendering* (bubbles, reasoning, tool calls, attachment previews, token usage) is shared and free on mobile.
- **`serverFetch` in the RN context is just native `fetch`** (no `electronAPI`), so native-chrome mutations are already off the WebView transport.
- **Expo DOM marshalled function props support return values and error propagation** (verified in `expo/src/dom/marshal.tsx` — `invokeNativeAction` resolves with `result`, rejects with the reconstructed error). That is what lets the embed get `{ url }` back and keep its existing `.then(navigate)` chain.

## Testing

- New: `forkEntity` wire tests (`agentsClient.test.ts`) and `singleFlight` unit tests.
- Full suites green (88 agents-mobile, 92 agents-server-ui), `tsc --noEmit` clean both packages, eslint clean.
- ⚠️ **Caveat:** the native-transport round-trip is verified from the Expo DOM source + types, but should be confirmed once on a device/simulator build.

## Follow-up parity gaps (out of scope here)

A broader mobile↔desktop audit surfaced these for separate PRs: **rename / `setEntityTitle`** (no mobile equivalent — closest analog to this work), **kill confirmation** (mobile kill via SIGKILL fires with no confirm dialog), attachment-manifest preview in the context drawer, and session-list refinements (group-by/filter by runner/workingDir, list-row actions). Fork itself is fully at parity.

### DIFF
--- a/packages/agents-mobile/app/session.tsx
+++ b/packages/agents-mobile/app/session.tsx
@@ -36,6 +36,12 @@ type SessionDomEmbedProps = {
   inlineQueuedMessages?: Array<OptimisticInboxMessage>
   bottomInset?: number
   onRequestOpenEntity: (entityUrl: string) => Promise<void>
+  // Marshalled so the per-message fork runs over native networking; pointer
+  // shape matches the runtime's `EventPointer`.
+  onRequestForkEntity?: (
+    entityUrl: string,
+    opts?: { pointer?: { offset: string | null; subOffset: number } }
+  ) => Promise<{ url: string }>
   style?: StyleProp<ViewStyle>
   matchContents?: boolean
   serverHeaders?: { url: string; headers: Record<string, string> } | null
@@ -68,7 +74,7 @@ function SessionRouteInner({
   }
   router: ReturnType<typeof useRouter>
 }): React.ReactElement {
-  const { serverUrl } = useAgents()
+  const { serverUrl, forkEntity } = useAgents()
   const tokens = useTokens()
   const scheme = useColorSchemeMode()
   const insets = useSafeAreaInsets()
@@ -153,6 +159,9 @@ function SessionRouteInner({
             bottomInset={composerInset}
             serverHeaders={serverHeaders}
             onRequestOpenEntity={async (target) => openSession(target)}
+            onRequestForkEntity={(targetUrl, opts) =>
+              forkEntity({ entityUrl: targetUrl, pointer: opts?.pointer })
+            }
             dom={domOptions(styles, embedSize, tokens.bg)}
           />
         ) : (
@@ -186,6 +195,7 @@ function SessionRouteInner({
           entityUrl={entityUrl}
           onBack={goBack}
           onSetView={setView}
+          onOpenEntity={openSession}
           onShare={openShare}
         />
       )}
--- a/packages/agents-mobile/app/session.tsx
+++ b/packages/agents-mobile/app/session.tsx
@@ -12,6 +12,7 @@ import {
 import { useLocalSearchParams, useRouter } from 'expo-router'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import { useAgents } from '../src/lib/AgentsProvider'
+import type { ForkPointer } from '../src/lib/agentsClient'
 import { useAgentsRouteGuard } from '../src/lib/useAgentsRouteGuard'
 import { useColorSchemeMode, useTokens } from '../src/lib/ThemeProvider'
 import {
@@ -36,11 +37,10 @@ type SessionDomEmbedProps = {
   inlineQueuedMessages?: Array<OptimisticInboxMessage>
   bottomInset?: number
   onRequestOpenEntity: (entityUrl: string) => Promise<void>
-  // Marshalled so the per-message fork runs over native networking; pointer
-  // shape matches the runtime's `EventPointer`.
+  // Marshalled so the per-message fork runs over native networking.
   onRequestForkEntity?: (
     entityUrl: string,
-    opts?: { pointer?: { offset: string | null; subOffset: number } }
+    opts?: { pointer?: ForkPointer }
   ) => Promise<{ url: string }>
   style?: StyleProp<ViewStyle>
   matchContents?: boolean
--- a/.changeset/agents-mobile-fork.md
+++ b/.changeset/agents-mobile-fork.md
@@ -1,0 +1,12 @@
+---
+"@electric-ax/agents-mobile": patch
+"@electric-ax/agents-server-ui": patch
+---
+
+Bring the fork capability to mobile at parity with desktop:
+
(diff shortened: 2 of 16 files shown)


msfstef:
@claude please review this PR. Focus on the cross-package fork wiring (the marshalled `onRequestForkEntity` transport in `EmbedApp`/`ElectricAgentsProvider`), the native `forkEntity` + kebab gating in `agents-mobile`, and whether the decisions described above are sound. Post your findings as a review comment — do not push any changes.

msfstef:
Thanks — addressed in `3e4dc0418`:

- **🟠 State-based kebab guard → synchronous latch.** Correct catch. `forkSubtree` now flips a `forkInFlightRef` before the `await` (cleared in `finally`) instead of relying on the `forkPending` state, so a same-tick double-tap can't fire two forks — matching the per-message button's `singleFlight`. Dropped `forkPending` from the `useCallback` deps and corrected the comment so it no longer claims parity it didn't have.
- **Nit 1 — disabled kebab gave no reason.** The disabled "Fork subtree" item now shows a `Fork permission required` subtitle when the caller lacks the `fork` permission, mirroring the per-message button's label.
- **Nit 2 — `onOpenEntity` optional on `SessionScreen`.** Left as-is: it's defensive optionality consistent with the screen's other optional callbacks, and both `app/session.tsx` call sites pass it. Noted for future callers.

The on-device/simulator round-trip check for the native fork transport remains the one thing to confirm before un-drafting.

msfstef:
Cleaned up and rebased onto latest `main` (force-pushed; history is now one squashed commit on top of main).

**Comment/diff cleanup:** trimmed the multi-line explanatory comments added across the fork work to lean one-liners matching sibling density (collapsed the duplicated single-flight-latch rationale, shortened the ForkPointer / forkEntity / createForkEntity / forkEntityImpl docs, dropped comments that restated the code).

**Rebase — convergent refactor resolved:** `main` landed `useForkFromHere` (a shared hook extracting the per-message "Fork from here" logic), `useEntityPermission`, `fork` in `CHAT_VIEW_PERMISSIONS`, and a comments/reply feature — overlapping the same files. Resolution kept **both** intents:
- Adopted main's `useForkFromHere` hook in `ChatView` (dropped my now-redundant inline version), so my `ChatView` delta disappears entirely.
- My button-hardening moved into the hook: `useForkFromHere`'s `onFork` now **returns** the fork promise (was fire-and-forget), so the `singleFlight` spinner in `AgentResponse` can track in-flight state.
- Merged `AgentResponse` so main's Reply/error-rendering meta-actions and my fork spinner/`singleFlight` coexist (kept main's `!showCopy && !showFork && !onReply` early-return after my hooks).
- The native re-route is unaffected — it lives at the provider level (`forkEntityImpl` → `createForkEntity`), and `useForkFromHere` calls `useElectricAgents().forkEntity`, so the marshalled native transport flows through transparently.

Verified post-rebase: both packages typecheck + lint clean; 88 agents-mobile and 105 agents-server-ui tests pass. The PR is now conflict-free (`MERGEABLE`).

msfstef:
https://github.com/user-attachments/assets/da8effaa-a659-4250-8a44-4452ce43f308

packages/agents-mobile/app/session.tsx
```diff
@@ -36,6 +36,12 @@ type SessionDomEmbedProps = {
   inlineQueuedMessages?: Array<OptimisticInboxMessage>
   bottomInset?: number
   onRequestOpenEntity: (entityUrl: string) => Promise<void>
+  // Marshalled so the per-message fork runs over native networking; pointer
+  // shape matches the runtime's `EventPointer`.
+  onRequestForkEntity?: (
+    entityUrl: string,
+    opts?: { pointer?: { offset: string | null; subOffset: number } }
```
kevin-dp: Can we export the `ForkPointer` type from `packages/agents-mobile/src/lib/agentsClient.ts` instead of inlining it here?

kevin-dp (approved):

PR #4586, opened 2026-06-16
Add electric-agents:// deep links to open sessions in desktop & mobile apps

### LINKED ISSUE #4584: Support deep links for opening sessions directly in Electric Agents desktop & mobile apps
## Summary

Support deep linking so users can open a session in the Electric Agents desktop or mobile app directly via a link (from email, docs, chat, etc). No dependence on a deployed web UI for routing or fallback in this initial implementation.

## Scope

- **Session Link Format:**
  - Use a custom protocol scheme: `electric://session/{id}`
  - All shared links and documentation should use this format for now.

- **Desktop (Electron):**
  - Register the `electric://` protocol handler at install/launch.
  - When a link is opened:
    - If app is running, bring to foreground and route to the requested session.
    - If app is not running, launch & route to the requested session.
    - Support deep linking for sessions only (future-proof path for other entity types).

- **Mobile (Expo/React Native):**
  - Register as handler for the `electric://` protocol on iOS & Android (intent filters, URL types).
  - On link open:
    - If user is not signed in or server isn’t configured, show onboarding/setup, then navigate to target session.
    - If app is running, navigate directly to session screen.
  - Generate and handle only custom protocol links for now (no web fallback).

- **Testing:**
  - Validate functionality on all platforms (Windows, macOS, iOS, Android).
  - Test with app running, backgrounded, or closed.

## Out of Scope
- No standard web session links or server-side redirects (as no deployed web UI exists yet).
- No fallback to opening a web browser if the app is not installed.

## Acceptance Criteria
- Users can click/tap `electric://session/{id}` links anywhere (browser, email, chat, etc) and the agents app opens to the correct session on desktop or mobile.
- Appropriate routing for onboarding states (setup/sign-in first, then session).
- Documentation updated to guide users to share/consume `electric://` links.

---

See thread discussion for additional details and related code context.

balegas:
Closes #4584

## Summary

Adds deep links that open a specific agent session directly in the Electric Agents **desktop (Electron)** and **mobile (Expo/React Native)** apps.

Link format (self-contained — a session is identified by *server + entityUrl*, so both are carried):

```
electric-agents://open-session?server=<url-encoded server base URL>&entity=<url-encoded entityUrl>
```

- **Scheme** reuses `electric-agents` (already registered for OAuth on mobile); desktop now registers it too.
- **Host** is `open-session` (not `session`) so expo-router doesn't auto-route it to the internal `/session` screen — a dedicated landing route handles it. The same link works on both platforms.

### Desktop (Electron)
- Registers the `electric-agents` scheme (`setAsDefaultProtocolClient` + `electron-builder.yml` `protocols`).
- Handles all three OS delivery paths: macOS `open-url` (with a pre-ready queue), Windows/Linux cold-start `argv`, and warm-start `second-instance`.
- New `desktop:open-session` IPC channel → controller resolves the link's server against saved servers → renderer switches to that server and navigates to the session. Unknown server → a toast ("you haven't added this server").

### Mobile (Expo/RN)
- Captures incoming links (cold start via initial URL, warm via a global `Linking` listener) as a `pendingSessionLink`.
- A dedicated `app/open-session.tsx` landing route switches to the link's server when needed (self-hosted; Cloud servers need sign-in and are not silently switched), then opens the session.
- Routed after the onboarding/server-setup gates, so a link arriving mid-onboarding opens once setup completes.

### Sharing (generate)
- **Mobile** `ShareSessionScreen`: the Session link now emits the `electric-agents://open-session` app link (replacing the web link), per the issue's "custom protocol only".
- **Desktop** `ShareEntityDialog`: new **Copy session link** action.

### Docs
- `website/docs/agents/usage/sharing-and-deep-links.md` (+ sidebar entry).

## Out of scope (v1)
Web/https session links & server-side redirects; browser fallback when the app isn't installed; auto-adding an unknown server from a link; deep links for non-session entity types (the format leaves room).

## Design / plan
Committed under `docs/superpowers/specs/` and `docs/superpowers/plans/` for reviewers.

## Test plan
- [x] Unit tests for the link build/parse/match helpers — mobile (`agents-mobile`, vitest, in CI), desktop renderer builder (`agents-server-ui`, vitest, in CI), desktop main parser (`agents-desktop`, node:test — matches the package's existing convention; typechecked in CI).
- [x] `pnpm typecheck` green across `agents-mobile`, `agents-server-ui`, `agents-desktop`; full mobile (92) and server-ui (104) suites pass.
- [ ] **Manual platform matrix** (not CI-automatable): app **running / backgrounded / closed** × **macOS / Windows / iOS / Android**:
  - `open "electric-agents://open-session?server=…&entity=…"` (macOS)
  - `xcrun simctl openurl booted "electric-agents://open-session?…"` (iOS)
  - `adb shell am start -a android.intent.action.VIEW -d "electric-agents://open-session?…"` (Android)
  - Verify: known server opens the session; unknown server shows the inform path; the share UIs produce a link that round-trips.


balegas:
Thanks for the thorough review — addressed all of it in e56570596.

### Important

**1. Desktop cold-start delivery + wrong-server routing**
The open-session payload now flows through a single *pull* (`desktop:get-pending-session`) rather than a pushed payload that races React mounting its listener. Main stashes the payload; the renderer pulls it on mount (cold start) and on an `onOpenSession` wake-up *signal* (warm start), and pulling clears it so it's consumed exactly once. While here I also caught that `main.ts` runs `createWindow()` *before* `openSessionFromDeepLink`, so the "existing window" branch was actually taken on cold start too — the signal+pull design covers both. The renderer also now waits for `servers` to hydrate the target id before `setActiveServer` + navigate, so it never routes against whatever server happened to be active.

**2. Mobile silently adding an unknown server**
Mobile now refuses like desktop: it only switches to servers the user has already added; an unknown self-hosted server gets a "you haven't added this server" message instead of a silent `addSavedServer` + switch. The route's branching is extracted into a pure `decideOpenSession` state machine with unit tests (cloud-abandon / activeMatches / known-switch / unknown-refuse / trailing-slash).

### Suggestions

- **Loose host match** — tightened to a boundary check (`open-session` must be followed by `?`, `/`, or end); `open-sessionfoo` no longer matches. Covered by tests.
- **Duplicated helpers** — extracted one shared module, `@electric-ax/agents-runtime/session-links` (build/parse/match/argv), re-exported by desktop, mobile and web. The parser is now plain string ops so it behaves identically across Node/browser/RN and transparently handles the single-slash Android variant. `hostOf` (3 copies) collapsed into `serverHost.ts`. Added a shared round-trip/fixtures suite plus per-package wiring tests.
- **Dispatch-layer coverage** — added tests for `resolveOpenSessionPayload` (desktop server-matching) and `decideOpenSession` (mobile state machine).
- **Large planning docs** — removed from the tree.

Also updated the docs page so the mobile "unknown server" description matches the new refuse behavior.

All four packages typecheck and lint clean, and the new/updated unit suites pass (48 tests).

packages/agents-mobile/src/lib/openSessionDecision.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (!target) return { kind: `abandon` }
+
+  const targetServer = normalizeServerUrl(target.serverUrl)
+  const activeServer =
+    activeServerUrl !== null ? normalizeServerUrl(activeServerUrl) : null
+  const activeMatches = targetServer === activeServer
+
+  if (activeMatches) return { kind: `route`, entityUrl: target.entityUrl }
+  // A Cloud server we're not already signed into can't be switched to silently.
+  if (isCloudServer(targetServer)) return { kind: `abandon` }
```
msfstef, line 48: **Should-fix: mobile silently drops links to *saved* Cloud servers, and diverges from desktop here.**

Because `isCloudServer` is checked before `isSavedServer`, any non-active Cloud server returns `abandon` even when the user has already added and signed into it. And `abandon` redirects to `/` with no message (`open-session.tsx`) — strictly worse feedback than the explanatory `refuse` screen an *unknown self-hosted* server gets. Desktop, meanwhile, resolves saved Cloud servers via `findSavedServerForUrl` (no `source` filter, unlike `findCloudServerForUrl`) and *does* switch + open them. So the same shared link opens on desktop but silently no-ops on mobile.

The "Cloud needs interactive sign-in" rationale doesn't hold for an *already-saved* server: `HomeMenu.selectServer` switches to a Cloud server with just `prepareServerHeaders(url)` + `saveServerUrl(url)` — no prompt; the token is derived from the URL service-id + the stored dashboard JWT, which is exactly what the `switch` decision triggers. An expired token just 401s in-session, same as a manual switch.

Suggested reorder so the saved-server check handles cloud + self-hosted uniformly:

```ts
if (activeMatches) return { kind: `route`, entityUrl: target.entityUrl }
if (isSavedServer(targetServer)) return { kind: `switch`, serverUrl: targetServer }
// Only an *unsaved* Cloud server needs an interactive sign-in we can't do from a link.
if (isCloudServer(targetServer)) return { kind: `abandon` } // ideally surface a "sign in first" message instead of a silent redirect
return { kind: `refuse`, host: hostOf(targetServer) }
```

This makes mobile match desktop. Note the cloud-abandon unit test (`openSessionDecision.test.ts:38-45`, which hardcodes `isSavedServer: () => false`) and the comments at `openSessionDecision.ts:9` / `:47` encode the old premise and would need updating alongside.

msfstef (commented):
Reviewed the deep-links work end to end — overall it's a careful, well-tested implementation. The pull-not-push IPC on desktop (stash in main, renderer pulls + clears via `getPendingSession`) correctly sidesteps the cold-start listener race, and extracting the pure helpers (`resolveOpenSessionPayload`, `decideOpenSession`, `session-links`) so the branchy parts are unit-testable is the right call.

Two things flagged inline — one should-fix (mobile Cloud links), one non-blocking discussion point (mobile share link format) — plus one housekeeping note:

**Rebase + regenerate the lockfile.** `pnpm-lock.yaml` is ~830 lines smaller and includes a batch of unrelated babel/transitive **downgrades** (e.g. `@babel/compat-data` 7.29.7 → 7.29.3, several `@babel/plugin-*` 7.29.7 → 7.27.1/7.28.x). That looks like a `pnpm install` against a stale tree rather than anything this feature needs. Could you rebase on `main` and regenerate the lockfile with a fresh `pnpm install` so the PR doesn't carry an accidental dependency rollback?

packages/agents-mobile/src/screens/ShareSessionScreen.tsx
```diff
@@ -236,7 +236,7 @@ export function ShareSessionScreen({
   }
 
   const shareLink = async (): Promise<void> => {
-    const url = sessionWebUrl(serverUrl, entityUrl)
+    const url = sessionAppUrl(serverUrl, entityUrl)
```
msfstef, line 239: **Non-blocking — flagging for discussion, not requesting a change.**

This swaps the mobile share link from an `https` web link (`sessionWebUrl`) to the custom-scheme `electric-agents://` link. Worth being deliberate about the trade-off, because custom-scheme links are a known weak spot for *sharing* specifically: many clients won't make them clickable (Gmail strips them outright; Slack/Notion/etc. render them as inert text), and if the recipient doesn't have the app installed there's no fallback — vs. an https/universal link that can open a web page. Desktop's custom scheme is fine (that's the standard Electron pattern); it's the mobile *share* path where this narrows where a pasted link actually works.

Deferring full Universal/App Links (which need hosted `apple-app-site-association` / `assetlinks.json` + domain control) to a later pass is totally reasonable. The low-effort middle ground, if we want to keep paste-into-chat working now, is to have the share sheet emit an `https` link to a small redirect page that attempts the `electric-agents://` handoff and falls back gracefully — and those same https links upgrade into seamless universal links later. Not blocking; just want it to be a conscious call rather than a side effect of "custom protocol only."

PR #4587, opened 2026-06-16
feat(agents-mobile): session comments (desktop parity)

balegas:
## Summary

Brings session **comments** to `agents-mobile`, closing the parity gap with the desktop UI (comments shipped in #4551). Part of the agents-mobile ↔ agents-desktop parity track (follows #4564, #4553).

The mobile session timeline is a shared web component (`SessionChatLogDomEmbed`) rendered inside an Expo-DOM WebView, while the composer is native. Because the embed already merges comments via `useEntityTimeline` and renders `CommentBubble`, **reading comment bubbles already worked on mobile** — this PR adds the missing *write*, *reply*, and *dedicated-view* surface.

## What's included

- **Comment mode in the native composer.** A compact `Message / Comment` toggle (top of the composer), shown only when the entity type declares the comments collection and the user has `write` access. Comment mode reuses the desktop's `createSendCommentAction` verbatim (optimistic insert + authenticated `POST /collections/comments`), disables attachments/slash, and posts plain-text comments. Editing a queued message resets to prompt mode (mirrors desktop) so the comment path can't hijack an edit.
- **Tap-a-row-to-reply.** A new optional Expo-DOM callback `onRequestReplyToComment(target)` is threaded `SessionChatLogDomEmbed → EmbedApp → ChatLogView → EntityTimeline.onReplyToRow`. Tapping "reply" on a timeline row forwards the `SelectedCommentTarget` (plain JSON) to the native shell, which switches the composer to comment mode, focuses the input, and shows a reply banner.
- **Optimistic-comment bridge.** The composer (native) and timeline (embed) run in separate JS contexts with separate stream `db`s, so the optimistic row inserted on the native side isn't visible in the embed on its own. Still-pending comments are forwarded across the boundary and projected into the embed timeline (deduped by key against synced rows), mirroring the existing `inlineQueuedMessages` bridge — so a posted comment renders immediately. Their `~pending` orders keep them in the same bottom band the shared query uses, so ordering matches desktop.
- **Comments-only view.** A new `comments` `EmbedViewId` renders `buildCommentsTimeline` (via a `commentsOnly` prop on the chat-log embed) with a comment-locked native composer, reachable from a new gated **Comments** entry in the session menu.

The shared `agents-server-ui` changes are additive and optional — desktop callers of `ChatLogView` (embed-only) are unaffected. **No server / runtime changes.**

## Test plan

- [x] `@electric-ax/agents-mobile` + `@electric-ax/agents-server-ui` — typecheck clean
- [x] Full suites pass (server-ui 102, mobile 84), incl. the reused `comments.test.ts` (send action, `buildCommentsTimeline`, target encode/decode)
- [x] ESLint + Prettier clean

> No new automated tests: the new logic lives in deeply-integrated screens/providers, and neither package has a render harness for them (mobile has no component-render setup; server-ui only renders self-contained components). The reused comment primitives stay covered by `comments.test.ts`.

## Known limitation

- **Comments-only view target-click.** Tapping a reply's *timeline-row* target snapshot doesn't scroll (that row isn't in the filtered comment list); tapping a *comment* target works. Target-click in the main chat view works fully. A "switch to full timeline + focus" bridge would close this.


balegas:
Thanks for the review — addressed in 7f02888d4.

**Fixed**
- **Important #1 (`showStop` in comment mode).** Added `!commentMode` to the `showStop` condition so the Post button on the comments surface (and the comments-only view) never becomes a *Stop generating* control while the agent is running — mirrors desktop `MessageInput.tsx`.
- **Suggestion #1 (duplicated `formatReplyBannerLabel`).** Moved it into shared `lib/comments.ts` (no React-Native dep) and import it on both desktop and mobile; removed both local copies. Added a unit test for it (lowercase-first-char + `Reply` fallback).

**Deferred (reasoning)**
- **#2 silent no-op on unavailable send** — unreachable in practice: comment mode is gated on `commentsEnabled`, and the composer is disabled when `!db`. A surfaced error would be cosmetic only.
- **#3 effect-race focus** — benign as noted; net state is correct (prompt, no banner). Not worth the extra guard.
- **#4 render harness / broader extraction** — neither package has a component-render harness; reused primitives stay covered by `comments.test.ts`. The cheap part (extracting the pure label helper) is done + tested.

Verified locally: both packages typecheck, ESLint + Prettier clean, full suites pass (agents-server-ui 104, agents-mobile 84).

msfstef (commented):
Looks reasonable but it needs a rebase and also I would appreciate screenshots/video for the feature

PR #4590, opened 2026-06-16
feat(agents-desktop): give pull-wake runner a distinguishable label

msfstef:
## What

The agents-desktop app registers itself as a pull-wake "runner" that shows up in the mobile/desktop runner picker. Every desktop instance registered with the same hardcoded label `Electric Agents Desktop`, making it impossible to tell runners apart when selecting where to run an agent.

This makes the label distinguishable. It now defaults to:

```
<identity> · <hostname>
```

where `<identity>` is the signed-in Electric Cloud **name**, falling back to **email** (available earlier, before `whoami` resolves), then to `Electric Desktop` for local/manual servers. Examples:

- `Stefanos · stefs-mbp.local` (Cloud, signed in)
- `stef@example.com · stefs-mbp.local` (Cloud, name not yet resolved)
- `Electric Desktop · stefs-mbp.local` (local/manual server)

## Overrides

Resolution precedence (mirrors the existing `pullWakeRunnerId` pattern):

1. `ELECTRIC_DESKTOP_PULL_WAKE_RUNNER_LABEL` env var
2. `pullWakeRunnerLabel` in `settings.json`
3. auto `<identity> · <hostname>`

## Notes

- No schema/server/UI changes — registration already upserts the `label` on the stable runner id, so existing runners pick up the new label automatically on next launch. The selection UIs already render `label`, so the improvement flows straight through to both mobile and desktop pickers.
- Cloud auth state is now captured once in `startRuntime` and reused for both owner-principal derivation and the label (no extra `getCloudAuthState()` call); owner-principal behavior is unchanged.

## Test plan

- [x] `tsc --noEmit`, prettier, eslint pass
- [ ] Launch desktop signed into Cloud → runner appears as `<name> · <hostname>` in the picker
- [ ] Launch against a local server → runner appears as `Electric Desktop · <hostname>`
- [ ] Set `pullWakeRunnerLabel` / `ELECTRIC_DESKTOP_PULL_WAKE_RUNNER_LABEL` → overrides win on next launch

### DIFF
--- a/.changeset/desktop-unique-runner-label.md
+++ b/.changeset/desktop-unique-runner-label.md
@@ -1,0 +1,13 @@
+---
+"@electric-ax/agents-desktop": patch
+---
+
+Give the desktop pull-wake runner a distinguishable label instead of the
+hardcoded `Electric Agents Desktop`, so multiple runners are easy to tell
+apart in the mobile/desktop runner picker. The label now defaults to
+`<identity> · <hostname>`, where identity is the signed-in Cloud name
+(falling back to email, then `Electric Desktop`). It can be overridden via
+the `pullWakeRunnerLabel` setting in `settings.json` or the
+`ELECTRIC_DESKTOP_PULL_WAKE_RUNNER_LABEL` env var. Existing runners pick up
+the new label automatically on next launch (registration upserts on the
+stable runner id).
--- a/packages/agents-desktop/src/runtime/lifecycle.ts
+++ b/packages/agents-desktop/src/runtime/lifecycle.ts
@@ -1,4 +1,5 @@
 import path from 'node:path'
+import os from 'node:os'
 import { app, powerSaveBlocker } from 'electron'
 import { AGENT_SKILLS_DIR } from '../shared/paths'
 import {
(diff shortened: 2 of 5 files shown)


kevin-dp (approved):
Great quality of life improvement 👍

msfstef:
Thanks for the review! Addressed both suggestions:

**1. Empty-string identity → dangling separator** — Fixed in 9681caaaa. The auto prefix now uses `||` with `trim()` (`name?.trim() || email?.trim() || "Electric Desktop"`), so a blank or whitespace-only Cloud name/email falls through instead of producing a leading `" · <hostname>"`. This also makes the auto path consistent with the `.trim()` guard already in `store.ts`.

**2. Test coverage for label resolution** — Intentionally skipping for now. `lifecycle.ts` can't be unit-tested as-is (Electron `app` dependency), so a test would require extracting the resolution into a pure helper — an abstraction we deliberately kept out for a single-call-site, minimal change. Happy to revisit (helper in `shared/headers.ts` alongside the `runnerOwnerPrincipal*` helpers + a table-driven `node:test`) if we want to lock in the precedence/fallback behavior later.

PR #4594, opened 2026-06-16
Fix O(n) shape removal from the subquery index

### LINKED ISSUE #4279: Removing shape from Subquery Index is O(n) in v1.6
Removing a shape from Subquery Index in the Where Clause Filter has been found to be O(n), with n being the number of shapes in that index, in the sync-service v1.6. Adding and removing shapes needs to be O(1) to allow the system to scale, which v1.5 adhered to. Since shape removal currently blocks replication processing, slow removal can cause WAL lag.

## Issue in production

This issue caused WAL lag for Autoarc. Below you can see v1.6 on the left and v1.5 on the right. In the top graph you can see the WAL lag 1.6 caused. In the 2nd graph you can see that this was no due to replication processing. In the bottom graph you can see that shape removal was significantly slower in 1.6, enough to cause the WAL lag.

Image

[source](https://ui.honeycomb.io/electric-sql-06/environments/production/datasets/electric-region/board-query/9npN3NxtB4y/result/m9DPHXvcNsr?tab=overview)

The trigger for what's causing the shapes to be removed is "Materializer shape invalidated" which was present in v1.5 and will be dealt with on a separate issue.

## Benchmarks

Microbenchmarks around the SubqueryIndex have confirmed shape removal is O(n) and account for the delays seen by AutoArc.

## Cause

The SubqueryIndex introduced in v1.6 is a reverse index managed by the shape consumers themselves and is roughly in the form `value -> shape_handle` where evaluating which shape handles are affected by what values is O(1) but shape removal involves a scan through the ETS table to find the shape handles. A naive solution would be to also hold `shape_handle -> values` however this would increase memory footprint on what is already a highly memory inefficient structure. Since we want to reduce memory footprint, we should consider a redesign of how the index works.

robacourt:
## Summary

Removing a shape from the subquery index in the where-clause filter had become **O(n)** in the number of shapes — and **O(shapes × values)** once subqueries were seeded. Shape removal runs on the replication path and blocks it, so slow removal caused WAL lag in production (the "Materializer shape invalidated" path). This restores **O(1)** removal (as in v1.5).

Fixes #4279

## Problem

The subquery index is a single ETS `:bag` table. Several deletions used `:ets.match_delete/2` with a **wildcard in the key position** (e.g. `{:membership, handle, :_, :_}`, `{:node_positive_member, node, :_}`). A `:bag` (hash) table can only use its key index when the key is **fully bound**, so a partially-bound key forces a **full-table scan** that grows with the number of shapes. The worst offender was `delete_node_members`, which scanned every shape's per-`(node, value)` routing row on each removal — O(shapes × values).

## Solution

Switch the table from `:bag` to **`:ordered_set`** and lift the discriminating fields (`value`, `shape`, `branch`, `next_cond`) out of the value tuples into the **keys**. In an `:ordered_set`, a delete whose key has a bound **prefix** is range-limited (not a full scan), and a point delete by a fully-bound key is O(log n). Concretely:

- `unregister_shape/2` replaces its full-table `match_delete` scans with prefix-bounded `select_delete`s.
- `remove_shape/5` point-deletes the shape's node rows, and **derives** its `node_*_member` deletions from the removed shape's own membership rows (scoped by `subquery_ref`), applying exact point-deletes. This touches only the removed shape's rows — never the other shapes sharing a value.

Removal is now **O(V_shape · log n)** — proportional to the removed shape's own view size, independent of the total number of shapes, the number of shapes on a node, and the number of shapes sharing a value.

This is a pure internal storage change: every public function and all externally-observable behavior is preserved. The per-change routing hot path (`affected_shapes`) and the exact-evaluation `member?` path are equal-or-faster (the `:ordered_set` prefix `select` projects in C); table memory is roughly flat (measured slightly lower in a microbenchmark).

> Note: a dedicated **memory** reduction (interning / row-shrinking) is intentionally **out of scope** here — this PR is the removal-complexity fix only, kept minimal and behavior-preserving. Memory can be tackled separately.

## Test plan

- [x] `:performance` **flatness** tests prove removal is independent of each dimension — total shapes, shapes-on-node, and shapes-per-value — by measuring removal at two well-separated sizes and asserting the delta stays within noise (observed delta 0 at 1k vs 20k).
- [x] **Positive control**: removal cost *does* scale with the removed shape's own view size (pins the complexity class as O(V_shape), so a regression can't hide under a flat budget).
- [x] **No-orphan guard**: a full `Filter.remove_shape` of a seeded shape, and draining a shared node one-by-one, leaves the subquery index byte-identical to its pre-add state.
- [x] Existing `subquery_index` / `filter` / `consumer` / `shape_log_collector` suites green; full `mix test` green.


alco (approved):
Brilliant!

PR #4596, opened 2026-06-16
feat(agents): context-usage ring gauge + composition breakdown

kevin-dp:
Improves the composer-footer context-usage indicator (built on the Phase 0 gauge from the base branch). Two commits:

### 1. Circular ring gauge
Replaces the flat dot before the `X%` label with a small SVG donut whose arc fills proportionally to context-window usage. Track + progress arc both use `currentColor`, so the existing normal/warning/critical level colour tints it for free.

### 2. Composition breakdown popover
Hovering the indicator reveals a per-part breakdown of the prompt — a stacked bar + legend showing how much of the window the **system prompt**, **tool definitions**, **conversation messages**, and **free space** each occupy (modelled on Claude Code's `/context`).

**How the data works:**
- The runtime estimates the *stable* request parts (`{ system, tools }`, char/4 via `approxTokens`) once per model call and persists them on the step as a new additive `context_breakdown` column, next to the real cache-inclusive `context_input_tokens`.
- The UI derives the **Messages** bucket as `real total − system − tools`, and **Free** as `window − total`, so the segments always sum to the gauge even though the system/tools figures are approximations (the panel notes this).
- Shared, testable helpers `computeContextBreakdown` / `parseContextBreakdown` keep the math in `token-accountant` and out of the component.

### Files
- `entity-schema`: additive `context_breakdown` string column on steps.
- `outbound-bridge` / `pi-adapter`: compute + persist the estimate.
- `token-accountant`: breakdown helpers + types, exported from the client entry (+ 4 unit tests).
- UI: `ContextUsageRing`, `ContextUsageDetails` (HoverCard popover), wired into `ContextUsageIndicator`.

### Testing
- Full runtime suite green (1024 passed); UI `tsc` clean.
- Verified live in the desktop app — ring fills correctly and the hover breakdown renders.

### Note on base
Targets `feat/context-compaction` because it extends the Phase 0 gauge that only exists there. Re-target to `main` once the compaction PR merges.

### DIFF
--- a/.changeset/context-usage-breakdown.md
+++ b/.changeset/context-usage-breakdown.md
@@ -1,0 +1,13 @@
+---
+'@electric-ax/agents-runtime': minor
+'@electric-ax/agents-server-ui': patch
+---
+
+Context-usage indicator: a circular ring gauge plus a hover breakdown popover
+showing how the prompt decomposes across system prompt, tools, messages, and
+free space (à la Claude Code's `/context`). The runtime now persists an
+approximate `context_breakdown` of the stable request parts (system + tools) on
+each step alongside the cache-inclusive total, and exports new
+`computeContextBreakdown` / `parseContextBreakdown` helpers from the client
+entry; the UI derives the "messages" bucket as the real remainder so the
+segments always sum to the gauge.
--- a/.changeset/context-usage-breakdown.md
+++ b/.changeset/context-usage-breakdown.md
@@ -1,5 +1,5 @@
 ---
-'@electric-ax/agents-runtime': minor
+'@electric-ax/agents-runtime': patch
 '@electric-ax/agents-server-ui': patch
 ---
 
--- a/packages/agents-runtime/src/client.ts
+++ b/packages/agents-runtime/src/client.ts
@@ -50,6 +50,8 @@ export {
   CONTEXT_USAGE_BACKGROUND_START,
   CONTEXT_USAGE_HARD_CEILING,
   computeContextUsage,
+  computeContextBreakdown,
+  parseContextBreakdown,
   contextUsageLevel,
   formatContextUsagePercent,
 } from './token-accountant'
@@ -57,6 +59,9 @@ export type {
   ContextUsage,
   ContextUsageInput,
   ContextUsageLevel,
+  ContextBreakdownKey,
+  ContextBreakdownParts,
+  ContextBreakdownSegment,
 } from './token-accountant'
 export type { GoalCommand } from './goal-command'
 
--- a/packages/agents-runtime/src/entity-schema.ts
+++ b/packages/agents-runtime/src/entity-schema.ts
@@ -165,11 +165,8 @@ type StepValue = {
   context_input_tokens?: number
   // The model's context window for this step.
   context_window?: number
-  // JSON-encoded estimate of how the prompt decomposes across the stable
-  // request parts — `{ system, tools }` token counts (approximate; char/4).
-  // The UI derives the remaining "messages" bucket as the real cache-inclusive
-  // total minus these, so the breakdown always sums to the gauge. Optional and
-  // additive; older events without it stay valid.
+  // JSON estimate of the prompt's stable parts — `{ system, tools }` token
+  // counts (char/4). The UI derives "messages" as the real total minus these.
   context_breakdown?: string
 }
 type TextValue = {
--- a/packages/agents-runtime/src/entity-schema.ts
+++ b/packages/agents-runtime/src/entity-schema.ts
@@ -165,6 +165,12 @@ type StepValue = {
   context_input_tokens?: number
   // The model's context window for this step.
   context_window?: number
+  // JSON-encoded estimate of how the prompt decomposes across the stable
+  // request parts — `{ system, tools }` token counts (approximate; char/4).
+  // The UI derives the remaining "messages" bucket as the real cache-inclusive
+  // total minus these, so the breakdown always sums to the gauge. Optional and
+  // additive; older events without it stay valid.
+  context_breakdown?: string
 }
 type TextValue = {
   key?: string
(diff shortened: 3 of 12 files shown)


kevin-dp:
## 🤖 Automated review — context-usage ring gauge + composition breakdown

_Generated by a review agent. Severity-ranked; cite `file:line`._

### Findings

**[medium] — `approxTokens(opts.tools)` produces a near-meaningless tools estimate** — `pi-adapter.ts` via `token-budget.ts`
`opts.tools` is `Array<AgentTool>`, and the array branch of `approxTokens` only does char/4 on blocks whose `type === 'text'`; every other object contributes a flat `64`. So "Tools" is effectively `64 × toolCount` regardless of how large each tool's name/description/JSON-schema actually is — a real tool set is typically thousands of tokens, so this badly understates the tools bucket, which then silently inflates the "Messages" remainder. Suggest `approxTokens(JSON.stringify(opts.tools))` (or serialize each tool's name + description + parameters). Worth fixing since the feature's whole value is the honesty of this split.

**[low] — legend percents may not sum to 100%** — `ContextUsageDetails.tsx` + `formatContextUsagePercent`
Per-segment percents are each `Math.round`ed, so the four legend rows won't necessarily sum to 100%, and a non-zero segment under 0.5% renders a colored swatch + bar sliver labeled "0%". Cosmetic; consider a sub-percent format or a `<1%` floor for non-zero segments.

**[low] — duplicated clamp logic** — `ContextUsageDetails.tsx`
The headline recomputes `usedRatio = min(1, used/window)` locally while the trigger uses `usage.ratio`. They agree, but passing/using `usage.ratio` directly avoids drift.

**[nit] — schema/persistence verified clean** — `entity-schema.ts` / `outbound-bridge.ts`
`context_breakdown` column is additive (`z.string().optional()`, conditional spread on write) → backward-compatible; `parseContextBreakdown` is appropriately tolerant (malformed JSON → `{}`, non-number fields dropped).

**[nit] — SVG ring + accessibility verified clean** — `ContextUsageRing.tsx`
dasharray/offset math correct (`rotate(-90)` starts at 12 o'clock, sweeps clockwise; ratio 0 renders nothing, 1 fills, pre-clamped to [0,1]); SVG is `aria-hidden` with the parent span carrying the `aria-label`; the stacked bar has `role="img"` + label.

### Verified correct
- `computeContextBreakdown` clamping is sound (`used`→[0,window], `system`→[0,used], `tools`→[0,used−system], `messages=max(0,…)`, `free=window−used`); segments provably sum to the window and the used trio to `usedTokens`; `window>0 ? … : 0` guard prevents NaN. Negative-clamp and missing-parts cases tested.
- React: `useMemo` over `[steps]` recomputes correctly; stable unique `key`; returns `null` cleanly when there's no usage; HoverCard standalone usage matches the primitive's API; only one consumer (`MessageInput.tsx`).

### Test gaps
- No test for `usedTokens > contextWindow` (over-window clamp) or the tools-cap path where `system` alone exceeds `used`. No UI tests for the ring/details (consistent with the area).

### Verification
- `npx vitest run token-accountant` → 25 passed. `agents-server-ui` `tsc --noEmit` → clean.

### Overall
Mergeable after addressing the tools-estimate issue. The math, clamping, SVG geometry, schema additivity, and parse tolerance are all correct and well-tested; the React/accessibility wiring is solid. The biggest real risk is the **[medium]**: `approxTokens` on the tool array degenerates to `64 × toolCount`, so the "Tools" segment is essentially a constant and its shortfall is absorbed into "Messages" — the breakdown can look plausible while being materially wrong about where context is going, defeating the feature's purpose. Everything else is low/nit.

kevin-dp:
Thanks — addressed the actionable findings, one commit each:

- **[medium] tools estimate** → `5712883e7`. Confirmed: `approxTokens(opts.tools)` hit the array branch's flat ~64-per-block fallback, so "Tools" was ~64×toolCount regardless of real schema size (and the shortfall inflated "Messages"). Now `approxTokens(JSON.stringify(opts.tools))`, which captures each tool's name + description + parameter schema — what actually occupies the prompt.

- **[low] tiny non-zero segment labelled "0%"** → `922ff76c0`. A non-zero segment that rounds to 0% now shows `<1%`, so a visible swatch/bar isn't labelled "0%". (Left exact sum-to-100% alone — that needs largest-remainder apportionment, overkill for a composition popover.)

- **[low] duplicated clamp** → `8d95012c4`. The header now uses `usage.ratio` (already `min(1, used/window)` from `computeContextUsage`) instead of recomputing it.

- **Test gap (over-window clamp)** → `fac33da3a`. Added a `computeContextBreakdown` case for `usedTokens > contextWindow` (used clamped to window, `free` = 0, segments still sum to the window). The tools-cap path (`system ≥ used`) was already covered by the existing "never produces negative messages" test.

The two **[nit]s** (schema additivity, SVG/accessibility) were verified-clean, so no changes there.

Full runtime suite green; UI typecheck + tests pass.

kevin-dp:
Thanks for the two reviews — both flagged no critical/important issues. Addressed the two actionable suggestions, one commit each:

- **aria-label dropped the token/model detail** (both reviews) → `c252bba92`. Moving to a hover-only HoverCard left the trigger's `aria-label` as just the percent, a small regression for keyboard/screen-reader users vs. the old Tooltip. The trigger now carries `Context used: X% (<used> / <window> tokens · <model>) — hover for breakdown`, so the essential numbers are available without hovering.

- **Zero-token legend rows** (review 2) → `3caa02967`. The stacked bar already skips zero-ratio segments, but the legend rendered all four rows — so older steps with no persisted breakdown showed noisy `System prompt — 0 — 0%` rows. Gated the legend on `tokens > 0` to match the bar.

The "legend percents needn't sum to 100%" note is cosmetic and already mitigated by the `<1%` floor — left as-is per the earlier call (exact apportionment is overkill here).

UI typecheck + tests pass (111).

samwillis (approved):
Reviewed the current head against feat/context-compaction and the previous review thread. The earlier findings are addressed (tools estimate serialization, ratio reuse, <1% tiny segments, over-window clamp test, aria-label detail, and empty legend buckets), and I didn’t find any blocking regressions. CI is green. Approving.

KyleAMathews (approved):

PR #4598, opened 2026-06-16
Add read tool output guardrails

KyleAMathews:
Adds bounded, line-numbered output to the agents runtime `read` tool so file reads no longer dump up to 512KB of raw text into model context. Users can page through files with `offset`/`limit`, get explicit continuation guidance, and the edit tool no longer requires a prior read before applying an exact string replacement.

## Root Cause

The read tool had a 512KB hard file-size cap, but successful reads under that limit returned the entire file. That made medium-sized files expensive in context and gave agents no first-class way to inspect specific line ranges except by falling back to shell commands.

The edit tool also enforced a same-wake prior-read check. That added friction without much safety: an edit is already constrained by whether the exact `old_string` patch applies uniquely, or by `replace_all=true` when requested.

## Approach

- Add optional `offset` and `limit` parameters to `read`.
- Render file contents with line numbers and XML-ish path/content wrappers.
- Bound returned file content to 2000 lines and 50KB by default.
- Truncate individual lines longer than 2000 characters.
- Include continuation hints such as `Use offset=2001 to continue` when output is partial.
- Preserve existing safeguards for sandbox containment, binary files, and files over 512KB.
- Remove the edit tool's read-set gate while keeping exact-match and multi-match protections.
- Update Horton prompt/tool descriptions so agents are no longer told they must read before editing.

## Key Invariants

- A single read call should have a predictable context footprint.
- Truncation must be explicit, never silent.
- Edits should only apply when `old_string` is found exactly once, unless `replace_all=true` is provided.
- Existing binary and oversized-file rejection behavior should remain intact.

## Non-goals

- This does not add true streaming/ranged filesystem reads; the current sandbox API still reads the file after the existing 512KB stat gate.
- This does not add `.hortonignore`/secret deny rules.
- This does not broaden edit semantics beyond existing exact string replacement behavior.

## Trade-offs

The implementation keeps the existing 512KB file-size gate instead of introducing sandbox-level streaming. That makes the change smaller and safer, while still reducing the returned tool output from as much as 512KB to a bounded page. A future improvement could move truncation earlier into the sandbox read path for larger-file previews.

Removing the prior-read gate relies on the patch application itself for safety: if the exact text is absent or ambiguous, the edit is rejected. This is simpler and avoids forcing unnecessary reads solely to satisfy runtime state.

## Verification

```bash
cd packages/agents-runtime
pnpm exec vitest run test/read-file-tool.test.ts test/write-edit-roundtrip.test.ts test/edit-tool-read-guard.test.ts test/sandbox-tool-refactor.test.ts test/readset-isolation.test.ts test/sandbox-tool-symlink-safety.test.ts
pnpm run typecheck

cd ../agents
pnpm exec vitest run test/horton-system-prompt.test.ts
pnpm run typecheck
```

Also attempted broader agents tests; unrelated bootstrap/pull-wake tests failed locally with undici cache `invalid onError method` errors after building runtime dist.

## Files changed

- `packages/agents-runtime/src/tools/read-file.ts` — adds offset/limit params, line-numbered bounded output, truncation metadata, and empty-file/trailing-newline handling.
- `packages/agents-runtime/src/tools/edit.ts` — removes the same-session prior-read gate while preserving exact-match replacement behavior.
- `packages/agents-runtime/test/read-file-tool.test.ts` — adds focused coverage for default paging, offset/limit, byte/line truncation, and empty/trailing-newline handling.
- `packages/agents-runtime/test/edit-tool-read-guard.test.ts` and `readset-isolation.test.ts` — update coverage for edit behavior without read-set enforcement.
- `packages/agents/src/agents/horton.ts` — removes prompt guidance claiming edits require prior reads.
- `.changeset/horton-read-file-guardrails.md` — patch changeset for affected agents packages.


packages/agents-runtime/src/tools/edit.ts
```diff
@@ -6,17 +6,11 @@ import { SandboxError } from '../sandbox/types'
...
-export function createEditTool(
-  sandbox: Sandbox,
-  readSet: Set<string>
-): AgentTool {
+export function createEditTool(sandbox: Sandbox): AgentTool {
   return {
     name: `edit`,
     label: `Edit File`,
-    description: `Replace text in a file. The file must have been read with the read tool earlier in this session. By default the old_string must occur exactly once; set replace_all to true to replace every occurrence.`,
+    description: `Replace text in a file. By default the old_string must occur exactly once; set replace_all to true to replace every occurrence.`,
```
balegas, line 13: **Important: line-number prefixes will collide with `old_string`.**

`read` now prefixes every line with `` `${index + 1}: ` `` (`read-file.ts:90`), e.g. `42: const x = 1`, but `edit` matches `old_string` against the raw file bytes (`edit.ts:48,90`), which have no prefix. An agent following the natural read-then-edit flow will copy `42: ` into `old_string` and hit `Error: old_string not found`.

Nothing in either tool description warns about this, and the bounded-read change makes the read→edit path more common. Suggest clarifying the description, e.g.:

> The literal text to find, exactly as it appears in the file. Do **not** include the `N: ` line-number prefixes shown by the read tool — those are display-only.

packages/agents-runtime/src/tools/read-file.ts
```diff
@@ -52,15 +189,28 @@ export function createReadFileTool(
...
+            details: { charCount: 0, totalBytes: fileStat.size, binary: true },
           }
         }
 
         const text = buffer.toString(`utf-8`)
+        const formatted = formatReadOutput(filePath, text, {
+          offset: positiveInteger(rawOffset, 1),
+          limit: positiveInteger(rawLimit, DEFAULT_READ_LIMIT_LINES),
+        })
         readSet?.add(resolve(sandbox.workingDirectory, filePath))
```
balegas: **Important: `readSet` is now dead state.**

`edit` was the only consumer of `readSet`. With the guard removed, `readSet` is still produced here and in `write.ts:46`, and threaded through `horton.ts` and `worker.ts`, but nothing ever calls `readSet.has(...)` anymore.

This is dead plumbing — the `new Set()` allocations and the optional `readSet?` params on read/write can be removed. Leaving it invites a future reader to assume it still gates something. Either remove it in this PR or mark it with a `// TODO: readSet is now unused`.

Relatedly, `readset-isolation.test.ts` no longer constructs any `readSet` and no longer tests isolation, so its filename/`describe` are now misleading — worth renaming or folding into the edit test.

balegas (approved):

packages/agents-runtime/src/tools/edit.ts
KyleAMathews: Fixed in 26a46c102 — updated the `old_string` parameter description to call out that read tool line-number prefixes like `42: ` are display-only and must not be included.

packages/agents-runtime/src/tools/read-file.ts
KyleAMathews: Fixed in 26a46c102 — removed the stale `readSet` plumbing entirely from read/write/Horton/Worker and updated docs/tests. Also removed the stale isolation/read-guard test rather than keeping a renamed version. Verified `rg "readSet" .` returns no matches.

PR #4601, opened 2026-06-16
fix(agents-mobile): eliminate chat timeline flashing and fix auto-scroll

msfstef:
## Intent

Eliminate the visible "flashing" of the agents-mobile chat timeline during normal use — sending a message, growing the composer (multiline / attachments), and queuing messages — **while preserving** the dynamic bottom inset, instant optimistic sending, and reliable pin-to-bottom. Also fix inconsistent auto-scroll when the composer grows.

## Architecture context (read this first)

The mobile chat screen is **two separate UIs**:
- A **native composer + queue drawer** (`agents-mobile/src/screens/SessionScreen.tsx`).
- A **chat timeline rendered inside an Expo DOM (`'use dom'`) WebView embed** (`agents-server-ui/src/embed/SessionChatLogDomEmbed.tsx` → `EmbedApp.tsx` → `ChatView.tsx`'s `ChatLogView` → `EntityTimeline.tsx`), hosted by `agents-mobile/app/session.tsx`.

These run in **separate JS contexts**. The native side and the embed communicate via props (serialized across the Expo DOM bridge) and an imperative handle.

## Root cause (the key finding)

**A value delivered over the Expo DOM prop bridge forces a full, memoization-busting re-render of the embed's React tree every time it changes — and with a tree as heavy as the timeline, that re-render *is* the visible "flash."**

Confirmed against the Expo SDK 54 source (`expo/src/dom/webview-wrapper.tsx` + `dom-entry.tsx`) — it is **not** a native WebView reload or remount:
- The WebView `source` is derived from the component's file path only, never from props, so a prop change never reloads the page; there's no `key` change either, so no remount.
- Prop updates are sent as a message: the native wrapper `injectJavaScript`s a `$$props` payload and the page-side runtime handles it with a plain `setProps(...)` state update on a root created once. So a prop update is mechanically just a `setState` inside the WebView.
- **But the payload is JSON — re-serialized and re-deserialized — so every data prop arrives as a brand-new reference on every update.** The page renders `<App {...freshlyDeserializedProps} {...actions} />`, busting every downstream `useMemo` / `React.memo` / `useCallback` that depends on a prop. Nothing memo-skips, so the whole heavy subtree re-does its work (virtualized-list measurement, markdown rendering, CSS masks) → a dropped frame = the flash.
- The native wrapper only emits `$$props` when its marshalled props **change identity**. So a parent re-render that passes new *inline* references (`style={[...]}`, `dom={...}`, inline callbacks) ships a fresh-reference update and flashes too — which is why the earlier "froze the CSS / native no-op" tests still flashed until the embed stopped re-rendering. It was never a true no-op; the references were changing.

You cannot make a bridge-crossed prop update cheap (the fresh references are unavoidable). The only lever is to **not send frequently-changing values as props at all** — keep props referentially stable so the native side emits nothing, and push dynamic updates through the imperative handle (which never crosses the bridge).

So every flash traced back to *something that emitted a `$$props` update (a changed/fresh prop reference) or re-rendered the embed*:
1. **Drawer flash on send** — the optimistic message was tracked via `entity.status === 'running'` and pruned from inline tracking while still `pending`, so it briefly fell into the queue drawer.
2. **Composer-resize flash** — the composer height re-renders the host route on every multiline/attachment change, re-rendering the (non-memoized) embed.
3. **Inset prop flash** — the bottom inset was passed as a live prop that changed on every resize.
4. **Queued-send flash** — `scrollToBottomSignal` and `inlineQueuedMessages` were live props that changed on send.
5. **Auto-scroll bug** — the pin-to-bottom `ResizeObserver` watched the **content-box**, but the inset is applied as **bottom padding**; a padding-only change leaves the content-box untouched, so multiline growth didn't trigger a re-pin (it only scrolled when content also changed — hence intermittent).

## Approach

**Nothing that changes frequently crosses the bridge as a prop.**
- **Memoize** both DOM embeds and keep every native-passed prop reference-stable (memoized `style`/`dom`, value-memoized `serverHeaders`, stable callbacks, frozen initial inset).
- **Deliver dynamic values imperatively** via a `useDOMImperativeHandle` ref (`SessionChatLogDomRef`):
  - `setBottomInset(px)` → direct CSS-var write (`--mobile-chat-bottom-inset`), no React.
  - `scrollToBottom()` → direct `scrollTop`, no React.
  - `setInlineQueuedMessages(messages)` → updates the embed's **internal** `useState`. The render still cascades, but every *surrounding* reference stays stable, so memoized subtrees (timeline, markdown, router) are skipped — cheap, no flash — unlike a bridge prop update where all references are fresh.
- The imperative handle registers a beat after the WebView boots, so the native side pushes via a small `usePushToEmbed` hook that retries on animation frames until the handle exists.
- Fix the drawer flash by keying off `generationActive` (matching desktop) and keeping a message in inline tracking until it actually leaves the pending inbox.
- Fix auto-scroll by observing the **border-box** in the pin-to-bottom `ResizeObserver`, plus a synchronous `scrollTop` correction before the rAF re-pin (ResizeObserver fires before paint, avoiding one misaligned frame).

## Key decisions & trade-offs

- **Internal state for inline queued messages re-renders the embed root.** That's fine *only because* `onNavigatePathname` is `useCallback`'d — the embed's router is `useMemo`'d on it, and an unstable identity would remount (and flash) the timeline. Props are unchanged on an internal re-render, so the callback keeps its identity.
- **`SessionChatLogDomRef` lives in a plain `.ts` module**, not the `'use dom'` embed file: named exports from `'use dom'` files aren't resolvable by the consuming app (single-default-export only).
- **`expo/dom` ambient shim** (`agents-server-ui/src/types/expo-dom.d.ts`) + an **inline `DOMImperativeFactory` index signature** in `sessionChatLogDomRef.ts`: agents-server-ui doesn't depend on expo (the embed is only bundled by the mobile app), so `expo/dom` isn't resolvable there. This is deliberate — a previous attempt to `import`/`extends` from `expo/dom` failed mobile tsc with "Cannot find module 'expo/dom'". **Do not "simplify" this back to importing from `expo/dom`.**

### Deliberately NOT done (so we don't go in circles)

- **Do not pass the inset/queued messages as live props** "since the embed is memoized." `React.memo` re-renders when a prop *value* changes — live props reintroduce the flash. The frozen-initial + imperative-update split is intentional.
- **Do not freeze `serverHeaders` with `useMemo(..., [])`.** It must update if auth headers refresh (token rotation); hence the `JSON.stringify` memo key (stable identity, updates on real change). The stringify is on a tiny object — negligible.
- **Did not unify the three imperative methods into one generic `updateState` channel** or replace the rAF retry with a "ready" event — Expo exposes no handle-ready signal, and three distinct semantic methods read clearer. Duplication was removed via the shared `usePushToEmbed` hook instead.
- **Did not push the `ResizeObserver` behavior behind a prop / create a `ChatLogViewMobile` wrapper.** `ChatLogView` is already the mobile-embed-only view, and border-box + sync-scroll are harmless/general for desktop.
- **Did not extract the inline-pending projection shared by `ChatLogView` and `SessionScreen`** — it's pre-existing, cross-package, and array-vs-map; extracting would over-abstract.

## Files changed

- `agents-mobile/app/session.tsx` — memoize embeds; stabilize props; `usePushToEmbed` hook + imperative inset/queued-message delivery; imperative `scrollToBottom` on send; `async openSession`.
- `agents-mobile/src/screens/SessionScreen.tsx` — `generationActive` gating; inline-tracking prune guard (keep while still pending).
- `agents-server-ui/src/embed/SessionChatLogDomEmbed.tsx` — imperative handle (`setBottomInset`/`scrollToBottom`/`setInlineQueuedMessages`); internal state for queued messages; stable `onNavigatePathname`.
- `agents-server-ui/src/embed/sessionChatLogDomRef.ts` *(new)* — shared imperative-handle contract + rationale.
- `agents-server-ui/src/types/expo-dom.d.ts` *(new)* — ambient `expo/dom` shim for agents-server-ui's tsc.
- `agents-server-ui/src/embed/EmbedApp.tsx` — inset var seeded on document root; removed dead `scrollToBottomSignal` plumbing.
- `agents-server-ui/src/components/views/ChatView.tsx` — `ChatLogView` uses `generationActive`/`entityStopped`; dropped `scrollToBottomSignal` from `cacheKey` and props.
- `agents-server-ui/src/components/EntityTimeline.tsx` — pin-to-bottom observes border-box + synchronous scroll correction.

## Testing / how to verify

Manually verified on device by the author: no flash on send, multiline growth, attachments, queue add, or first message to an idle agent; auto-scroll follows composer growth; inset and instant optimistic send still work.

**Important:** webview-side files (`SessionChatLogDomEmbed`, `sessionChatLogDomRef`, `EmbedApp`, `ChatView`, `EntityTimeline`) require a **DOM-bundle rebuild** to take effect; native files (`session.tsx`, `SessionScreen.tsx`) hot-reload.

Both `agents-mobile` and `agents-server-ui` typecheck and lint clean.

## Rebase notes

Rebased onto `main` after the **"fork subtree"** mobile feature merged (it touched the same files). Integration points to be aware of:
- `session.tsx` now passes `onRequestForkEntity` to the embed. It is wrapped in a **`useCallback` (`handleForkEntity`)** rather than the inline closure main used — an inline function would be an unstable prop and defeat the embed's `memo` (reintroducing the flash). Keep it stable.
- `EmbedApp.tsx` merged cleanly: main's `ToastProvider` + `forkEntityImpl` wiring coexists with this PR's removed `scrollToBottomSignal`/inline-inset-style and dropped `CSSProperties` import.
- `SessionScreen.tsx` auto-merged: main's fork state/menu wiring is independent of this PR's `generationActive` gating + inline-prune guard.

## Status / follow-ups

- Draft. No automated test covers the flash behavior (it's a WebView-layer visual artifact); verification is manual.
- The simplification pass (reuse/efficiency/altitude/comment review) is already applied — the diff is intended to be the final, consolidated form, not a stack of patches.

### DIFF
--- a/.changeset/mobile-chat-timeline-flash-fix.md
+++ b/.changeset/mobile-chat-timeline-flash-fix.md
@@ -1,0 +1,6 @@
+---
+'@electric-ax/agents-mobile': patch
+'@electric-ax/agents-server-ui': patch
+---
+
+Fix mobile chat timeline flashing when sending messages, resizing the composer (multiline/attachments) and queuing messages, and fix inconsistent auto-scroll as the composer grows. The timeline WebView embed now receives dynamic updates (bottom inset, inline queued messages, scroll) imperatively instead of via props that re-render and flash it.
--- a/.changeset/mobile-chat-timeline-flash-fix.md
+++ b/.changeset/mobile-chat-timeline-flash-fix.md
@@ -3,4 +3,4 @@
 '@electric-ax/agents-server-ui': patch
 ---
 
-Fix mobile chat timeline flashing when sending messages, resizing the composer (multiline/attachments) and queuing messages, and fix inconsistent auto-scroll as the composer grows. The timeline WebView embed now receives dynamic updates (bottom inset, inline queued messages, scroll) imperatively instead of via props that re-render and flash it.
+Fix mobile chat timeline flashing when sending messages, resizing the composer (multiline/attachments) and queuing messages, and fix inconsistent auto-scroll as the composer grows. The timeline WebView embed now receives dynamic updates (bottom inset, inline queued messages, scroll) imperatively instead of via props that re-render and flash it. Also fix the composer occasionally not clearing after send when typing quickly.
--- a/packages/agents-mobile/app/session.tsx
+++ b/packages/agents-mobile/app/session.tsx
@@ -1,4 +1,13 @@
-import { useEffect, useMemo, useState, type ComponentType } from 'react'
+import {
+  memo,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+  type ComponentType,
+  type Ref,
+} from 'react'
 import {
   Keyboard,
   Platform,
@@ -23,6 +32,7 @@ import {
 } from '../src/screens/SessionScreen'
 import type { EmbedViewId } from '../src/lib/embedView'
 import SessionChatLogDomEmbedModule from '@electric-ax/agents-server-ui/src/embed/SessionChatLogDomEmbed'
+import type { SessionChatLogDomRef } from '@electric-ax/agents-server-ui/src/embed/sessionChatLogDomRef'
 import SessionStateInspectorDomEmbedModule from '@electric-ax/agents-server-ui/src/embed/SessionStateInspectorDomEmbed'
 import { getActiveServerHeadersSnapshot } from '@electric-ax/agents-server-ui/src/lib/auth-fetch'
 import type { OptimisticInboxMessage } from '@electric-ax/agents-server-ui/src/lib/sendMessage'
@@ -33,8 +43,6 @@ type SessionDomEmbedProps = {
   serverUrl: string
   entityUrl: string
   theme: `light` | `dark`
-  scrollToBottomSignal?: number
-  inlineQueuedMessages?: Array<OptimisticInboxMessage>
   bottomInset?: number
   onRequestOpenEntity: (entityUrl: string) => Promise<void>
   // Marshalled so the per-message fork runs over native networking.
@@ -45,13 +53,20 @@ type SessionDomEmbedProps = {
   style?: StyleProp<ViewStyle>
   matchContents?: boolean
   serverHeaders?: { url: string; headers: Record<string, string> } | null
+  ref?: Ref<SessionChatLogDomRef>
   dom?: unknown
 }
 
-const SessionChatLogDomEmbed =
+// The timeline is an Expo DOM (WebView) embed that re-posts — and visibly
+// flickers — on any prop change or re-render. So we memo it, keep every prop
+// below reference-stable, and push the values that actually change (bottom
+// inset, queued messages, scroll) imperatively via the ref, not as props.
+const SessionChatLogDomEmbed = memo(
   SessionChatLogDomEmbedModule as ComponentType<SessionDomEmbedProps>
-const SessionStateInspectorDomEmbed =
+)
+const SessionStateInspectorDomEmbed = memo(
   SessionStateInspectorDomEmbedModule as ComponentType<SessionDomEmbedProps>
+)
 
 export default function SessionRoute(): React.ReactElement | null {
   const params = useLocalSearchParams<{
@@ -83,7 +98,6 @@ function SessionRouteInner({
   const [chatComposerHeight, setChatComposerHeight] = useState(
     CHAT_COMPOSER_BASE_HEIGHT + insets.bottom
   )
-  const [chatLogScrollSignal, setChatLogScrollSignal] = useState(0)
   const [inlineQueuedMessages, setInlineQueuedMessages] = useState<
     Array<OptimisticInboxMessage>
   >([])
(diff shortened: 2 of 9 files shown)


msfstef:
Thanks for the thorough review. Responding to the three optional suggestions — no code changes this round, with reasoning below (happy to act on #1 if we want to close the edge):

**1. Asymmetric `projectedPendingMessage` gating (`entityStopped`).** Confirmed the divergence is real but narrow. The embed `ChatLogView` mirrors desktop `GenericChatBody`, which suppresses the inline pending bubble when `entityStopped`; the native `SessionScreen.projectedPendingMessage` only drives drawer de-duplication and gates on `generationActive` only. The only way to hit the gap: queue a message to an **idle** agent, then have it transition to `stopped` within the 15s inline-tracking window before the message syncs. In that window the embed won't render it inline (stopped) and the native drawer still hides it (via `inlineQueuedKeysRef`), so it's briefly invisible — then self-corrects when the 15s timeout releases tracking (it reappears in the drawer) or the message processes. Note a one-line gate change isn't a full fix: even with `projectedPendingMessage` nulled, `inlineQueuedKeysRef` still filters the key out of the drawer, so closing this properly means surfacing tracked-but-stopped messages in the drawer (matching desktop's suppress-inline-show-in-drawer behavior). Given it's a transient stop-edge (not a flash), I've left it as-is; can do the alignment as a follow-up if we want it closed.

**2. `scrollToBottom()` single rAF racing the inline render.** Agreed and intentional. The single rAF pins to the current `scrollHeight` as a best-effort first pin; the optimistic row arrives on a separate inner-render schedule, and the border-box pin-to-bottom `ResizeObserver` re-pins once the row lays out. The observer is deliberately the load-bearing mechanism, so the rAF racing the row layout is harmless.

**3. `SessionChatLogDomRef` index signature.** Accepted trade-off as documented — it's required to satisfy expo's `DOMImperativeFactory` without importing `expo/dom` from agents-server-ui (which can't resolve it). The three named methods below the index signature keep the real surface honest at the definition site; call sites lose arity checking, which we accept for this three-method handle.

msfstef:
https://github.com/user-attachments/assets/72c40839-67d2-4221-adaf-2002f26de249

kevin-dp (approved):

PR #4605, opened 2026-06-17
feat(agents): context compaction (budget notices, mid-turn floor, background compaction)

kevin-dp:
Adds context compaction to the agents runtime, which previously only truncated at the window limit. Modelled on OpenAI Codex's summarization but adapted to our event-sourced timeline: a compaction checkpoint is a durable `context_inserted` row placed at a stored **watermark**, so history reconstruction folds everything up to the watermark into a summary.

### What's included
- **Token-usage gauge** — persist cache-inclusive `context_input_tokens` + `context_window` per step; `ContextUsageIndicator` shows "X% used" in the composer footer.
- **Budget notices** — inject a `<token_budget>` message into the model context at 25 / 50 / 75% usage (synthesized at the model-call seam, not persisted).
- **Tool-output truncation** — cap any single oversized `tool_result` with a placeholder.
- **Mid-turn synchronous floor** — at the 90% hard ceiling, compact before the model step via the adapter's `transformContext` hook (so a single tool-heavy turn can't blow the window).
- **Background (turn-end) compaction** — non-blocking: at 85% a detached summarize runs off the critical path; its checkpoint is applied at the next turn's start, or immediately if it finishes while idle. Each generation uses a watermark-unique checkpoint id so a new run can't supersede a prior completed one, and summarize calls are bounded by a hard timeout.
- **UI** — a "Compacting…" indicator (distinct blocking vs. background styling) and a collapsible "Context compacted" entry in the conversation timeline.

### Design notes
- Reconstruction places the summary **at** the stored watermark, so a prompt+answer that arrived while a background summary ran (physically after the checkpoint, logically after the watermark) are kept verbatim **after** the summary.
- Only `complete` checkpoints act as a reconstruction watermark; `running`/`failed` are UI-only and never hide history (crash-safe).
- Thresholds are env-tunable: `ELECTRIC_AGENTS_COMPACT_CEILING`, `ELECTRIC_AGENTS_COMPACT_BG_CEILING`, `ELECTRIC_AGENTS_COMPACT_MIN_TOKENS`.

### Testing
- Full runtime suite green (1020 passing), including dedicated compaction, reconstruction, mid-turn, and background unit tests.
- Verified live in the desktop app end-to-end (compaction applies, context shrinks, indicator clears).

### Note
PR #4596 (context-usage ring gauge + breakdown) is stacked on this branch; re-target it to `main` after this merges.


kevin-dp:
## 🤖 Automated review — context compaction

_Generated by a review agent. Severity-ranked; cite `file:line`._

### Findings

**[high] — mid-turn checkpoint over-covers and drops the kept tail on the next reconstruction** — `context-factory.ts` (mid-turn `writeCheckpoint`) + `timeline-context.ts` `compactionWatermarkOf`
The mid-turn compactor keeps `messages.slice(coveredCount)` (the last `COMPACT_KEEP_TAIL=6` messages) verbatim for the *current* model call, but the persisted `complete` checkpoint is written with the fixed id `compaction` and **no `attrs.watermark`**. So `compactionWatermarkOf` falls back to the checkpoint row's own stream order, which is the latest order at write time. On the next turn, reconstruction drops every timeline item with `at <= checkpoint.at` — including the 6 tail messages, which are neither in the summary (it only covered `0..coveredCount`) nor kept → lost context. Background compaction avoids this by storing `watermark: head`; the mid-turn path should likewise persist the boundary watermark it actually summarized. No test reconstructs across a mid-turn compaction.

**[low] — an orphaned `running` checkpoint pins the UI indicator forever** — `CompactionIndicator.tsx`
The indicator picks the globally-newest `kind:"compaction"` row and spins if it is `running`. Reconstruction correctly ignores orphaned `running` rows, but the UI has no equivalent guard: if a process crashes after writing a background `running` row and before its terminal `complete`/`failed`, the spinner stays forever (no later row supersedes it, since each generation has a unique id). Cosmetic.

**[low] — stale "95%" comments** — `compaction.ts`, `token-accountant.ts` JSDoc, `context-factory.ts`
Several comments still say 95% / "critical at 95%" but `CONTEXT_USAGE_HARD_CEILING = 0.9`. Stale only.

**[nit] — `pi-adapter.ts` `pendingRequestMessageCount = messages.length`** records the pre-compaction length while the request actually sent is the shorter compacted list. Because `anchorTokens` is re-anchored to real cache-inclusive usage each step-end, the drift is bounded and only affects *when* the mid-turn trigger fires — not correctness.

### Verified correct
- Tool-pair invariant in reconstruction (each run emits its `tool_call`+`tool_result` together under one `at`, so a watermark can't split a within-run pair).
- Background generation ids (`compaction-bg-<watermark>`) correctly prevent a new `running` from superseding a prior `complete`.
- `running`/crashed checkpoints never hide history.
- Idle-apply path: no double-apply (`pendingBackgroundCompaction` nulled after applying; pre-handler and idle apply are mutually exclusive); the settle-just-before-idle race is handled; the `onSettled` slot-identity guard prevents stale callbacks.
- Summarize timeout: `call.catch(()=>{})` attached synchronously before the race (no unhandled rejection), timer cleared in `finally`. Tested for stall/abort/empty-summary.
- Schema fields additive/optional; reconstruction tolerates missing fields.

### Test gaps
- No test reconstructs across a **mid-turn** checkpoint (exactly the `[high]` path).
- No orphaned-`running` UI indicator test.
- Idle-apply loop and "settle just before idle" race not unit-tested at the process-wake level.

### Overall
The architecture is sound and unusually well-commented; the background lifecycle, supersession, watermark ordering, and timeout race are carefully designed and well-tested. The one finding I'd block on is the **mid-turn over-coverage** (`[high]`): unlike background compaction, the synchronous floor doesn't persist a watermark, so its checkpoint silently drops the ~6 verbatim tail messages on the following turn's reconstruction — a real (if bounded) context-loss bug on the 90% path. I'd want that fixed (persist the boundary watermark) plus a reconstruction test before merge; everything else is low/nit/cosmetic.

kevin-dp:
Addressed the review (8ce00aa96):

**[high] mid-turn tail-loss — fixed.** The mid-turn checkpoint was persisted with no `attrs.watermark`, so reconstruction fell back to the checkpoint row's own (latest) position and dropped the verbatim tail the summary had excluded. Fix: the mid-turn compactor now folds the **whole** context into the summary (Codex-style, no verbatim pre-compaction tail) and the checkpoint is stamped with `watermark = current timeline head`. Summary and watermark now agree, so reconstruction folds exactly what was summarized and keeps everything appended afterward. This also removed the `keepTail` + tool-pair orphan-guard machinery. Added the missing reconstruction test (`timeline-compaction.test.ts` → "mid-turn checkpoint keeps messages produced after compaction") and rewrote `compaction-midturn.test.ts` for the new behavior. Within-turn recency is preserved via the sticky view (summary + messages appended since).

**[low] stale "95%" comments — fixed** (hard ceiling is 90%).

**[low] orphaned `running` UI spinner & [nit] `pendingRequestMessageCount`** — acknowledged, left as-is: both are cosmetic/bounded and not context-affecting; happy to follow up separately if wanted.

Full runtime suite green (tsc clean).

kevin-dp:
Circled back and addressed the remaining two as well, each in its own commit:

- **[low] orphaned `running` spinner** — c313578fe: a summarize always writes a terminal row within its ~120s timeout, so a `running` checkpoint older than 150s is treated as orphaned (crashed) and no longer rendered, with a self-clearing timer so it disappears even with no further events.
- **[nit] `pendingRequestMessageCount`** — 0bc94adb2: added a comment clarifying it intentionally records the incoming (uncompacted) count, since pi-agent passes transformContext the full conversation each step (so the count indexes that array and the next step's trailing slice is exact).

kevin-dp:
Thanks for the careful review. Went through each point:

**Addressed (one commit each):**

- **Suggestion 1 — mid-turn watermark captured after the await** → `b130e42a0`. Agreed, real (if narrow) timing-dependence. The watermark is now snapshotted when the `running` row is written — *before* the summarize await — so an event materializing into the StreamDB during the slow summarize can't bump the head past what the summary covered. Mirrors how background compaction snapshots its head up front.

- **Important — untested orphan-clearing logic** → `445c7552e`. Agreed. This package has no React-render harness (no jsdom/testing-library; `passWithNoTests`), so rather than pull that in I extracted the decision into a pure `isRunningCheckpointOrphaned(timestamp, now)` helper (+ `STALE_RUNNING_MS`) and unit-tested it: fresh → shown, just under deadline → shown, at/past deadline → hidden, and **missing/unparseable timestamp → shown** (the exact NaN path you flagged, now locked in). The component imports the helper, so the tested logic is what runs.

**Declined (with rationale):**

- **Suggestion 2 — reuse `selectLatestContextUsage` in `ContextUsageIndicator`.** The usage *math* is already shared (the component calls `computeContextUsage`); the only duplication is the "latest step by `_seq`" loop, and the component needs the full latest *row* (its `model_id`, and on the stacked #4596 the `context_breakdown` too) — so a single pass returning the row is cleaner here than `selectLatestContextUsage` + a second lookup. Happy to revisit if you'd prefer a shared `selectLatestContextStep` row-returning helper.

- **Suggestion 3 — `as any` on the query collection.** It's an established pattern across the existing UI query sources (e.g. `lib/compaction.ts`); properly typing `q.from(db.collections.x)` is a worthwhile but separate cross-cutting effort rather than something to one-off here.

Full runtime suite green; UI typechecks + the new helper test pass.

KyleAMathews (approved):

PR #4608, opened 2026-06-17
Store UUID filter values as compact binaries

robacourt:
## Summary
- Store parsed UUID values as 16-byte binaries (rather than 36 byte strings) inside the sync-service filter/eval representation, converting back to canonical UUID text at API and SQL output boundaries. This reduces memory use across the system, but especially in where-clause filter indexes, such as the equality and subquery index keys, by avoiding full canonical UUID strings for indexed values.
- Bump the ShapeStatus metadata version so persisted shape comparable terms created with the old UUID string representation are ignored and rebuilt under the new representation.
- Add a changeset for `@core/sync-service`.

## Testing
- `mix test test/electric/shape_cache/shape_status_test.exs test/electric/shape_cache/shape_status/shape_db_test.exs`
- Temporarily patched the standard oracle schema from text/int IDs to UUID PK/FK columns and ran `mix test --only oracle test/integration/oracle_property_test.exs` successfully. This patch was not committed.
- Temporarily ran the larger UUID oracle stress check successfully: `CHECK_TIMEOUT=60000 SHAPE_COUNT=800 MUTATIONS_PER_TXN=10 TXNS_PER_BATCH=10 BATCH_COUNT=200 SKIP_REPATCH_PREWARM=true mix test --seed 8 --only oracle test/integration/oracle_property_test.exs`. This patch was not committed.
- Temporarily added route/subset coverage for `id = $1::uuid` and `id = CAST($1 AS uuid)` against the `items.id` UUID column; the four targeted tests passed and were not committed.

### DIFF
--- a/packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
+++ b/packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
@@ -43,8 +43,21 @@ defmodule Electric.Replication.PostgresInterop.Casting do
   def parse_bool(x) when x in ~w|f fa fal fals false|, do: false
 
   def parse_uuid(maybe_uuid) do
-    {:ok, value} = Ecto.UUID.dump(maybe_uuid)
-    Ecto.UUID.load!(value)
+    case maybe_uuid do
+      <<_::128>> = uuid ->
+        uuid
+
+      _ ->
+        {:ok, value} = Ecto.UUID.dump(maybe_uuid)
+        value
+    end
+  end
+
+  def uuid_to_string(<<_::128>> = uuid), do: Ecto.UUID.load!(uuid)
+
+  def uuid_to_string(maybe_uuid) do
+    {:ok, uuid} = Ecto.UUID.cast(maybe_uuid)
+    uuid
   end
 
   def parse_date("epoch"), do: Date.from_iso8601!("1970-01-01")
--- a/packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
+++ b/packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
@@ -43,14 +43,8 @@ defmodule Electric.Replication.PostgresInterop.Casting do
   def parse_bool(x) when x in ~w|f fa fal fals false|, do: false
 
   def parse_uuid(maybe_uuid) do
-    case maybe_uuid do
-      <<_::128>> = uuid ->
-        uuid
-
-      _ ->
-        {:ok, value} = Ecto.UUID.dump(maybe_uuid)
-        value
-    end
+    {:ok, value} = Ecto.UUID.dump(maybe_uuid)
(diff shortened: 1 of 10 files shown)


packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
```diff
@@ -43,8 +43,21 @@ defmodule Electric.Replication.PostgresInterop.Casting do
...
+      <<_::128>> = uuid ->
+        uuid
+
+      _ ->
+        {:ok, value} = Ecto.UUID.dump(maybe_uuid)
+        value
+    end
+  end
+
+  def uuid_to_string(<<_::128>> = uuid), do: Ecto.UUID.load!(uuid)
```
alco: How is this different from `Casting.uuid_to_string(value)`?

packages/sync-service/test/electric/shapes/filter_test.exs
```diff
@@ -90,6 +90,38 @@ defmodule Electric.Shapes.FilterTest do
...
+
+      filter =
+        Filter.new()
+        |> Filter.add_shape(
+          "s1",
+          Shape.new!("uuid_table", where: "id = '#{uuid1}'::uuid", inspector: inspector)
+        )
+        |> Filter.add_shape(
+          "s2",
+          Shape.new!("uuid_table", where: "id = '#{uuid2}'::uuid", inspector: inspector)
```
magnetised: what happens if we don't cast the value to `uuid`? If I were writing that where filter I wouldn't cast to uuid...

alco (approved):

magnetised (approved):

packages/sync-service/test/electric/shapes/filter_test.exs
robacourt: Good point. The uncast form works: the unknown string literal is inferred from the UUID column side of the equality, and the equality index still stores the 16-byte UUID key.

I updated this test to use `id = '...'` rather than `id = '...'::uuid`, and verified it with:

```sh
mix test test/electric/shapes/filter_test.exs
```

packages/sync-service/lib/electric/replication/postgres_interop/casting.ex
robacourt: It is not meaningfully different here: `Ecto.UUID.cast/1` accepts both canonical UUID strings and raw 16-byte UUID binaries.

I removed the `<<_::128>>`/`Ecto.UUID.load!/1` special case so `uuid_to_string/1` uses the single `Ecto.UUID.cast/1` path for both forms.

Verified with:

```sh
mix test test/electric/replication/eval/env_test.exs test/electric/replication/eval/sql_generator_test.exs test/electric/replication/eval/runner_test.exs test/electric/shapes/filter_test.exs
```

PR #4613, opened 2026-06-17
Fix batched child completion wakes

KyleAMathews:
## Summary

Fix a child-wake regression where parent agents could miss child completion wake rows when multiple observed children finished close together. The runtime now batches queued wake rows into one `wake_batch` payload so the parent gets every child result in a single handler run, avoiding extra latency and token usage from serial re-wakes.

## Root Cause

Pending live batches can contain more than one `wake` row. The wake-draining path previously selected a single runnable wake from the fresh window and then acknowledged the whole window. When two child completion wake rows were coalesced into the same pending window, one wake could be surfaced to the handler while the sibling wake was effectively consumed by the ack.

## Approach

- Add `combineWakeEvents(...)` for wake-only pending windows.
- Preserve the single-wake path by delegating to `changeEventToWakeEvent(...)` when there is only one wake row.
- When multiple wake rows are present, create one `WakeEvent` with:
  - `type: "wake"`
  - `eventCount` set to the number of wake rows
  - `payload.type: "wake_batch"`
  - `payload.sources` for the child/source URLs
  - `payload.wakes` containing the original wake messages
  - `payload.changes` flattened from the wake messages
- Use the last wake row offset for the selected wake offset so the batched handler pass advances through the full coalesced wake window.
- Keep non-wake fresh work using the existing `selectWakeFromEvents(...)` path.

## Key Invariants

- A pending wake-only window must not drop any wake rows when acknowledged.
- Multiple wake rows should be delivered to the agent in one handler run, not serialized into extra LLM calls.
- Single wake rows should preserve existing behavior and payload shape.
- The ack offset should still cover the drained pending window after the batched wake is handled.

## Non-goals

- This does not change the server wake registry or child wake production logic.
- This does not introduce multiple handler passes for sibling child wakes.
- This does not change inbox-triggered wake selection.

## Trade-offs

An alternative fix was to split coalesced wake rows into separate in-process wake windows. That would preserve one wake per handler run, but it adds latency and can waste LLM tokens. Batching keeps the runtime efficient and lets the agent reason over all child completions together.

## Verification

```bash
pnpm --filter @electric-ax/agents-runtime test test/process-wake.test.ts -t "combines multiple pending wake events" --run
pnpm --filter @electric-ax/agents-runtime typecheck
```

## Files changed

- `packages/agents-runtime/src/process-wake.ts` — batch multiple pending wake rows into one wake payload before acking the window.
- `packages/agents-runtime/test/process-wake.test.ts` — add a focused regression test that coalesces two pending wake rows and verifies one batched handler wake plus acking at the last offset.
- `.changeset/batch-child-completion-wakes.md` — patch changeset for `@electric-ax/agents-runtime`.

### DIFF
--- a/.changeset/batch-child-completion-wakes.md
+++ b/.changeset/batch-child-completion-wakes.md
@@ -1,0 +1,5 @@
+---
+'@electric-ax/agents-runtime': patch
+---
+
+Batch queued child completion wakes into a single wake payload so parent agents receive every child result without extra handler runs.
--- a/packages/agents-runtime/src/process-wake.ts
+++ b/packages/agents-runtime/src/process-wake.ts
@@ -367,10 +367,9 @@ function combineWakeEvents(
   }
 }
 
-function selectWakeFromEvents(
+function selectNextInboxWakeFromEvents(
   events: Array<ChangeEvent>,
-  fallbackSource: string,
-  preferredKind?: FreshKind
+  fallbackSource: string
 ): { wakeEvent: WakeEvent; offset: string | null } | null {
   const cancelledInboxKeys = new Set<string>()
   for (let i = events.length - 1; i >= 0; i--) {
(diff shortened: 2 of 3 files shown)


samwillis (approved):
LGTM and worked when merged into my test brnach

PR #4616, opened 2026-06-17
Fix agent wake claim token races

KyleAMathews:
Fixes agent wake handling so concurrent desktop sessions no longer invalidate each other's claim write tokens, and same-stream duplicate wakes are retried after the active wake drains instead of being dropped.

This should stop the recent local desktop stall pattern where an unrelated wake or SIGINT caused another active session to start receiving `Invalid write token` 401s. It also narrows failed-parent child cleanup so a parent handler failure detaches the parent wake registration instead of sending `SIGKILL` to a child that may already be running.

## Root Cause

The claim write token store treated a consumer as owning at most one stream. When the same desktop runner/consumer claimed a wake for a second agent stream, it deleted the first stream's token, so the still-active wake failed later writes with `UNAUTHORIZED: Invalid write token`.

A second race existed for duplicate same-stream wake events. The runner could claim another wake for a stream while the previous wake was still active, minting a replacement token for that same stream. The first fix avoided concurrent same-stream claims, but review caught that a simple local skip could strand pending work if the skipped event was the only retry signal.

Separately, parent handler failure cleanup used `deleteEntity()` for children spawned before the failure. That client method actually sends semantic `SIGKILL`, which is too broad for rolling back a failed parent manifest entry.

## Approach

- Scope claim write tokens by stream, while keeping a reverse index so `clearConsumer()` still clears all streams owned by that consumer.
- Track active wake stream paths in the runtime router, and let the pull wake runner ask whether a stream is already in flight.
- Prevent concurrent claims for the same stream while a claim or wake is active.
- Defer locally skipped same-stream wake events and retry them after the active stream wake drains.
- Treat `SIGINT` of an active or just-starting run as a shutdown request, so the wake closes instead of returning to idle and immediately processing fresh input in the aborted invocation.
- Add a narrow internal wake unregister endpoint and use it to detach failed-parent child wake registrations without signaling the child entity.

## Key Invariants

- A consumer may hold claim write tokens for multiple streams at the same time.
- A stream has at most one active claim token at a time.
- A duplicate wake-stream event must not mint a replacement token while that stream is active.
- A locally skipped wake must be retried after the active wake drains so pending work is not stranded.
- `SIGINT` aborts the current handler and closes that wake; later work runs in a later wake.
- Failed parent cleanup may detach wake registrations created by that failed run, but must not terminate the child entity.

## Non-goals

- This does not change Durable Streams lease/heartbeat semantics.
- This does not rewrite the runtime as per-agent XState actors; it adds guardrails to the existing promise-based wake dispatch model.
- This does not remove the explicit public/legacy `SIGKILL` APIs; it stops using them for failed-parent child cleanup.

## Verification

```bash
pnpm --filter @electric-ax/agents-runtime exec vitest run test/pull-wake-runner.test.ts test/process-wake.test.ts --pool-options.threads.maxThreads=2
pnpm --filter @electric-ax/agents-server exec vitest run test/claim-write-token-store.test.ts test/subscription-webhooks-routing.test.ts --maxWorkers=2
pnpm --filter @electric-ax/agents-runtime typecheck
pnpm --filter @electric-ax/agents-server typecheck
```

## Files Changed

- `packages/agents-server/src/claim-write-token-store.ts`: allow one consumer to hold multiple stream-scoped claim tokens.
- `packages/agents-runtime/src/create-handler.ts`: expose stream-path active wake tracking to the pull runner.
- `packages/agents-runtime/src/pull-wake-runner.ts`: prevent concurrent same-stream claims and defer/retry skipped active-stream wake events.
- `packages/agents-runtime/src/process-wake.ts`: close the wake after active-run `SIGINT` aborts, and detach child wake registrations on parent handler failure without killing children.
- `packages/agents-runtime/src/runtime-server-client.ts` and `packages/agents-server/src/routing/internal-router.ts`: add a narrow internal wake unregister call.
- Runtime and server tests: cover multi-stream consumers, same-stream duplicate deferral, retry after active wake drains, SIGINT wake closure, and failed-parent child wake detach.
- `.changeset/fix-agent-wake-claim-token-races.md`: patch changeset for the affected agents packages.

### DIFF
--- a/packages/agents-server/src/claim-write-token-store.ts
+++ b/packages/agents-server/src/claim-write-token-store.ts
@@ -7,26 +7,22 @@ interface ActiveClaimWriteToken {
 
 export class ClaimWriteTokenStore {
   private readonly claimsByStream = new Map<string, ActiveClaimWriteToken>()
-  private readonly streamByConsumer = new Map<string, string>()
+  private readonly streamKeysByConsumer = new Map<string, Set<string>>()
 
   mint(service: string, streamPath: string, consumerId: string): string {
     const streamKey = this.streamKey(service, streamPath)
     const consumerKey = this.consumerKey(service, consumerId)
     const previousClaimForStream = this.claimsByStream.get(streamKey)
     if (previousClaimForStream) {
-      this.streamByConsumer.delete(
-        this.consumerKey(service, previousClaimForStream.consumerId)
+      this.removeConsumerStream(
+        this.consumerKey(service, previousClaimForStream.consumerId),
+        streamKey
       )
-    }
-
-    const previousStreamForConsumer = this.streamByConsumer.get(consumerKey)
-    if (previousStreamForConsumer) {
-      this.claimsByStream.delete(previousStreamForConsumer)
     }
 
     const token = randomUUID()
     this.claimsByStream.set(streamKey, { token, consumerId })
-    this.streamByConsumer.set(consumerKey, streamKey)
+    this.addConsumerStream(consumerKey, streamKey)
     return token
   }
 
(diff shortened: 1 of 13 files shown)


packages/agents-server/src/claim-write-token-store.ts
```diff
@@ -7,26 +7,22 @@ interface ActiveClaimWriteToken {
 
 export class ClaimWriteTokenStore {
   private readonly claimsByStream = new Map<string, ActiveClaimWriteToken>()
-  private readonly streamByConsumer = new Map<string, string>()
+  private readonly streamKeysByConsumer = new Map<string, Set<string>>()
```
samwillis, line 10: This fixes the in-memory side of multi-stream ownership, but the callback release path still appears to be keyed through a single mutable `consumer_callbacks.primaryStream` per consumer. If the same consumer claims A then B, A's later `done` can resolve to B's stream and clear/update the wrong active token/status. The release path should use the released `(consumerId, epoch)` claim's `stream_path`/`entity_url`, or callback state needs to be scoped by claim/stream too.

samwillis (changes requested):
## Review

Interactive review with GPT-5.5.

### Blocking Finding

The token store now correctly allows a single consumer to own multiple stream-scoped write tokens, but the surrounding callback routing still stores only one `primaryStream` per `(tenantId, consumerId)` in `consumer_callbacks`. If the same consumer claims stream A, then stream B before A sends `done`, A's later `done` will look up B's `primaryStream`, check/clear B's in-memory token, and potentially mark B idle while leaving A's token/state stale. That can recreate the invalid-token/stale-state race this PR is trying to remove.

I think the fix is to make the release path use the released claim identity rather than the mutable callback row's current `primaryStream`: either key callback state by `(consumerId, epoch)` / stream, or use `materializeReleasedClaim()`'s returned `claim.stream_path` and `claim.entity_url` for token ownership, token clearing, and status updates.

### Verification

I attempted the focused PR tests, but the worktree shell is on Node `v18.20.5` and pnpm refused to run because the repo requires `>=20.19.0 || >=22.12.0`.

samwillis (approved):
Follow-up interactive review with GPT-5.5.

The latest update addresses my previous blocking concern: the wake `done` path now uses the released claim's stream identity instead of the mutable callback row's current primary stream, and the new regression coverage exercises the multi-stream single-consumer case.

Verification run locally in the review worktree:
- `@electric-ax/agents-runtime`: `test/pull-wake-runner.test.ts` and `test/process-wake.test.ts` passed, 115 tests
- `@electric-ax/agents-server`: `test/claim-write-token-store.test.ts` and `test/subscription-webhooks-routing.test.ts` passed, 19 tests

Approved.

PR #4621, opened 2026-06-17
Require meaningful slugs for spawned workers

KyleAMathews:
Require `spawn_worker` calls to include a meaningful slug, then append a short random suffix so spawned worker URLs are readable and generally unique. This makes child worker paths easier to identify in wake messages, logs, and the UI.

## Root Cause

Workers previously used an opaque `nanoid(10)` id, so spawned worker URLs were unique but not meaningful. When multiple workers were dispatched, agents and humans had to infer purpose from surrounding context instead of the worker path itself.

## Approach

- Add a required `slug` parameter to the `spawn_worker` tool schema.
- Normalize the slug into a safe path prefix:
  - trim whitespace
  - lowercase
  - replace unsupported characters with `-`
  - collapse repeated hyphens
  - trim leading/trailing hyphens
  - cap at 48 characters
- Append a short random hex suffix to avoid routine collisions:

```ts
const id = `${normalizedSlug}-${randomBytes(3).toString(`hex`)}`
```

- Reject slugs that normalize to no meaningful letter/number content before spawning.
- Update Horton’s worker-spawning instructions to tell it to provide short meaningful slugs.
- Remove stale fork-tool wording that said fork ids mirror the old `spawn_worker` id parameter.
- Add a changeset for `@electric-ax/agents`.

## Key Invariants

- Every spawned worker id starts with a meaningful slug.
- Worker ids remain generally unique via the random suffix.
- Invalid slugs do not call `ctx.spawn`.
- Existing worker behavior remains otherwise unchanged:
  - selected tools are forwarded
  - model config is forwarded
  - initial message is forwarded
  - `runFinished` wake with response remains enabled
  - sandbox inheritance remains enabled

## Non-goals

- This does not require globally collision-proof worker ids. If an unlikely slug/suffix collision occurs, spawning fails and the agent can retry.
- This does not restrict slugs to hyphen-only text; dots and underscores remain allowed because worker ids become URLs and can be encoded.
- This does not change the fork id generation behavior beyond removing stale documentation.

## Trade-offs

The implementation uses a small random suffix rather than UUID/nanoid-level entropy to keep worker URLs compact. That preserves readability while still avoiding collisions in normal use. Slugs are capped at 48 characters to avoid unwieldy URLs, at the cost of truncating very long model-provided labels.

## Verification

```bash
pnpm --filter @electric-ax/agents typecheck
pnpm --filter @electric-ax/agents exec vitest run test/spawn-worker-tool.test.ts
```

## Files changed

- `.changeset/meaningful-worker-slugs.md`
  - Adds a patch changeset for `@electric-ax/agents`.

- `packages/agents/src/tools/spawn-worker.ts`
  - Requires `slug`.
  - Normalizes and caps slug values.
  - Appends a random hex suffix to form the worker id.
  - Rejects non-meaningful slugs before spawning.

- `packages/agents/test/spawn-worker-tool.test.ts`
  - Updates existing spawn-worker tests to pass slugs.
  - Adds coverage for slug normalization, slug length capping, and invalid slug rejection.

- `packages/agents/src/agents/horton.ts`
  - Updates Horton’s instructions to provide meaningful slugs when spawning workers.

- `packages/agents/src/tools/fork.ts`
  - Removes stale wording tying fork ids to `spawn_worker` id behavior.

### DIFF
--- a/.changeset/meaningful-worker-slugs.md
+++ b/.changeset/meaningful-worker-slugs.md
@@ -1,0 +1,5 @@
+---
+'@electric-ax/agents': patch
+---
+
+Require meaningful slugs when spawning workers and append a random suffix so worker URLs are easier to identify while remaining generally unique.
--- a/packages/agents/src/agents/horton.ts
+++ b/packages/agents/src/agents/horton.ts
@@ -333,7 +333,7 @@ Dispatch a worker when:
 - The subtask is independent and can run in parallel with other work.
 - You need an isolated context (e.g., focused coding on one file without pulling its full content into our chat).
 
-When you spawn a worker, write its system prompt the way you'd brief a colleague who just walked in: include file paths, line numbers, what specifically to do, and what form of answer you want back. The system prompt sets the worker's persona and constraints; the required initialMessage is the concrete task you're handing off — that's what kicks the worker off, so without it the worker sits idle.
+When you spawn a worker, provide a short meaningful slug for the worker path (for example, \`auth-audit\` or \`api-test-fix\`) and write its system prompt the way you'd brief a colleague who just walked in: include file paths, line numbers, what specifically to do, and what form of answer you want back. The system prompt sets the worker's persona and constraints; the required initialMessage is the concrete task you're handing off — that's what kicks the worker off, so without it the worker sits idle.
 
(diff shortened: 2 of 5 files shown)


samwillis (approved):
LGTM

PR #4624, opened 2026-06-18
Use compact runtime shape refs internally

robacourt:
## Summary

Introduces an **in-memory numeric `shape_id`** (the "runtime shape ref") that replaces the binary `shape_handle` for all internal routing, filtering, and identity. Shapes are held millions of times across ETS tables, maps, and sets; each `shape_handle` is a ~30-byte binary, so swapping the internal key to a small integer is a substantial memory win — the largest amplifier being the subquery index, where the key was duplicated 4–6× per node per shape.

`shape_handle` is preserved at every external/durable boundary: the HTTP API, on-disk + storage-ETS keys, the `ShapeStatus` authority's public face, the client change-notification `Registry`, and the subquery tag-hash. The `shape_id` is **in-memory only — never persisted, freshly minted each boot.**

## How to review

The change is deliberately structured so most of it is trivial to review. It breaks into four parts:

### 1. Mechanical changes (the bulk — ~17 of 23 lib files are 90–100% mechanical)
Pure `shape_handle` → `shape_id` renames: function params, `@spec`s, ETS tuple keys, message tuples, local vars, and `is_shape_handle` → `is_integer` guard swaps. No behaviour change. These modules are key swaps only, **no algorithm change**:
`consumer_registry.ex`, `event_router.ex`, `filter.ex`, `filter/indexes/subquery_index.ex`, `materializer.ex`, `where_clause.ex`, `request_batcher.ex`, `partitions.ex`, `dependency_layers.ex` (algorithm unchanged — only the caller now supplies resolved ids), `ref_resolver.ex`, `state.ex`, `effects.ex`, `setup_effects.ex`, `event_handler/subqueries/{steady,buffering}.ex`, `dynamic_consumer_supervisor.ex` (partition hash now over id — behaviour-equivalent).

### 2. Core logic — `shape_status.ex` (the one place to read closely)
This is the feature. ShapeStatus becomes the authority that mints ids and holds the bidirectional map:
- `shape_id_table` ETS (`id → handle` reverse map) + `mint_id/1` (atomic `:ets.update_counter` on a `:seq` row).
- `id_for_handle/2`, `handle_for_id/2`, `fetch_shape_by_id/2`, and `shape_handle_for_log/2` (returns `"unknown, id: N"` on miss so log/telemetry callers can resolve unconditionally).
- The `shape_meta_table` row grows to a 6-tuple (id appended); ids are minted both in `add_shape/2` (new shapes) and `populate_shape_meta_table/2` (restore/boot), and the reverse-map row is cleaned up in `remove_shape/2`.

The id is never written to SQLite — the `shape_db`/`connection`/`query` persistence layer is untouched.

### 3. Planned boundary (handle ↔ id resolution, as designed)
Resolution happens **once at each edge**, never on the steady-state hot path:
- `shape_log_collector.ex` `dependency_ids/2` and `event_handler_builder.ex` `dep_id_for_handle!/2` — resolve a shape's persisted dependency handles → ids at consumer setup.
- `consumer.ex` `consumer_pid_for_handle/2` — the single handle→id resolver for cold-path/handle-boundary callers; routing-identity functions (`name`, `stop`, `consumer_pid`) are id-only.
- `shape_handle_for_log/2` used at id-only log/telemetry sites.

**Deliberately kept on `shape_handle`:** the API/plug boundary, all `Storage.*` and on-disk paths, the client Elixir `Registry` pub-sub (`register_for_changes`/`Registry.dispatch`), `PublicationManager`/`RelationTracker` (one-entry-per-shape, low memory value), and the subquery tag-hash (`subquery_tags.ex`/`move_broadcast.ex` are unchanged — their `md5("#{stack_id}#{shape_handle}…")` is streamed/persisted in `:tags` headers and must byte-match the Postgres-side hash in `querying.ex`).

### 4. Implementation-time additions (the only non-mechanical extras — review these)
Almost all are the unavoidable consequence of introducing a lookup: once a path resolves handle→id, it must handle the lookup *missing* (a shape concurrently deleted):
- **Deletion-race branches** around new lookups: `shape_cache.ex` restore path (`:error -> halt`), `shape_cleaner.ex` (resolve id **before** `ShapeStatus.remove_shape` deletes the mapping; nil-guard helpers `stop_consumer/3`, `remove_from_shape_log_collector/2`; suspend-path `:error -> :ok`), `consumer.ex` `all_materializers_alive?`.
- **`shape_cache.ex` `start_shape/4` + `start_materializers/3`** restructured to a `with` chain: a materializer-id miss now aborts and purges *before* starting the consumer, instead of starting both regardless (symmetric graceful handling).
- **`snapshotter.ex`** looks the consumer up via `ConsumerRegistry.whereis(id)` directly so a snapshot-failure report still reaches it after cleanup has removed the handle→id mapping.
- **`shape_status.ex` `put_handle_id/3`** — a `Mix.env() == :test`-gated ETS seeder for routing tests (compile-gated; cannot reach prod).

## Guarantees verified
- `shape_id` never reaches the wire (no `shape_id` in `api/`, `plug/`, `log_items.ex`, or `:tags`/`electric-handle` headers).
- id is in-memory only and never persisted (re-minted each boot).
- Subquery tag-hash stays handle-keyed and byte-compatible with the Postgres side.
- The two registries stay distinct (custom `ConsumerRegistry` = id; client Elixir `Registry` = handle).
- No internal function accepts both a handle and an id.

## Testing
Full suite green: 391 doctests, 9 properties, 1833 tests, 0 failures. Subquery move-in/move-out integration and `electric-handle` header round-trip covered by existing suites.


.changeset/small-shape-refs.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Use compact runtime shape refs internally instead of shape handle binaries in the sync service's in-memory routing, consumer, materializer, and filter paths. Public APIs, persisted shape metadata, storage paths, and client-visible handles remain unchanged. The service now logs each runtime shape ref assignment with its public shape handle so production logs can correlate compact refs back to handles.
```
robacourt:
```suggestion
Reduce memory use by using compact runtime shape refs internally.
```

packages/sync-service/lib/electric/replication/shape_log_collector/request_batcher.ex
```diff
@@ -53,24 +58,48 @@ defmodule Electric.Replication.ShapeLogCollector.RequestBatcher do
           Electric.Shapes.Shape.t(),
           :create | :restore
         ) :: :ok | {:error, any()}
+  def add_shape(_stack_id, _shape_handle, _shape, :restore) do
+    :ok
+  end
+
+  def add_shape(stack_id, shape_handle, shape, :create)
+      when is_stack_id(stack_id) and is_shape_handle(shape_handle) do
+    GenServer.call(
```
alco: This should just call the local `add_shape()` function to avoid duplicating `GenServer.call()` and the timeout magic number.

packages/sync-service/lib/electric/replication/shape_log_collector/request_batcher.ex
```diff
@@ -108,18 +137,34 @@ defmodule Electric.Replication.ShapeLogCollector.RequestBatcher do
...
+  end
+
+  def handle_call({:add_shape, shape_ref, shape_handle, shape}, from, state) do
     {:noreply,
      %{
        state
-       | to_add: Map.put(state.to_add, shape_handle, shape),
-         to_remove: MapSet.delete(state.to_remove, shape_handle),
-         to_schedule_waiters: Map.put(state.to_schedule_waiters, shape_handle, from)
+       | to_add: Map.put(state.to_add, shape_ref, {shape_handle, shape}),
```
alco: Hmm, so these maps and sets may now have to shapes of values? I feel like this change is missing a better in-code documentation clarifying the tradeoffs. For an uninformed reader it looks odd to have similar handle_call clauses where one has the additional `shape_handle` param.

packages/sync-service/lib/electric/replication/shape_log_collector/request_batcher.ex
```diff
@@ -193,4 +238,22 @@ defmodule Electric.Replication.ShapeLogCollector.RequestBatcher do
...
   end
+
+  defp shape_ref_for_handle(stack_id, shape_handle) do
+    case ShapeStatus.shape_id_for_handle(stack_id, shape_handle) do
+      {:ok, shape_id} -> shape_id
+      :error -> shape_handle
+    end
+  end
+
+  defp shape_handle_for_ref(stack_id, shape_id) when is_shape_id(shape_id) do
```
alco: Why the double naming of `shape_ref` and `shape_id` referring to the same thing?

alco (commented):
I have only reviewed a part of it.

So far, the addition of optionality between `shape_ref` and `shape_handle` in multiple modules/functions looks like added complexity we would like to avoid. What's the heuristic I should use when reading the code to know whether a given function can end up taking a shape ref or a shape handle at runtime?

robacourt:
> I have only reviewed a part of it.
> 
> So far, the addition of optionality between `shape_ref` and `shape_handle` in multiple modules/functions looks like added complexity we would like to avoid. What's the heuristic I should use when reading the code to know whether a given function can end up taking a shape ref or a shape handle at runtime?

Great points. I've totally re-written it to remove ambiguity and to make the boundary clear.

magnetised:
@robacourt why have a handle -> id lookup at all. why not just replace shape handles with an integer (system_boot_time (microseconds) + atomic read and increment)? seems like we could lose the external handle vs internal id split and just use an integer everywhere (that's stringified in the url, de-stringified in the request params normalisation). we never use the fact that the handle is composed of the shape hash + timestamp. it's completely opaque so there's no need to keep that handle construction, we could just generate an integer (in a thread safe way). did you consider this? is there something I'm missing?

robacourt:
> @robacourt why have a handle -> id lookup at all. why not just replace shape handles with an integer (system_boot_time (microseconds) + atomic read and increment)? seems like we could lose the external handle vs internal id split and just use an integer everywhere (that's stringified in the url, de-stringified in the request params normalisation). we never use the fact that the handle is composed of the shape hash + timestamp. it's completely opaque so there's no need to keep that handle construction, we could just generate an integer (in a thread safe way). did you consider this? is there something I'm missing?

I'd prefer this approach, but I'm struggling to find a way of fitting it into an 8 byte number (this PR manages to make it an 8 byte number by keeping it as an internal ID). You can only user 60bits before it starts using 24 bytes (boxed). You could squeeze boot time into 42 bits if it were milliseconds (rather than microseconds) and you used a custom epoch, that would give you ~139 years to play with. Then you'd have 18 bits left so if you went over 262K shapes you'd be into 24 byte numbers again. You also have the problem of two servers started in the same millisecond, but perhaps you could link it to when the database lock is acquired as no two servers can have that at the same time.

24 bytes would still smaller than our current shape handles which work out as 64 bytes once boxed. In reducing the memory footprint of the where clause filter you'd probably end up wanting a 8 byte ref to the 24 byte number which is what this PR was trying to avoid, but maybe that complexity is simpler than the complexity of this PR. What do you think @magnetised ?

magnetised:
@robacourt I don't think a 60 bit integer would limit us to ~300k shapes. I've thought about this a bit here and there and clauded this integer-only design:

  Overview

  Each ID is a 59-bit non-negative integer composed of a fixed boot prefix and a serial counter. Both prefix components are computed once at server boot; ID generation thereafter is a single atomic increment. Every ID is a single-word Erlang fixnum (no bignum allocation).

  Bit layout

```
  [ 8 bits: boot-year ][ 51 bits: serial counter ]
    └ years since 2026 (0–255)      └ atomic, seeded at boot
```
 
 Total width 59 bits, within the positive range of a 64-bit BEAM fixnum (max 2⁵⁹−1).

  Boot procedure

  At startup the server computes and freezes two values:

  1. Year prefix = boot_year − 2026 (8 bits, range 0–255 → 2026–2281).
  2. Counter seed = microseconds elapsed since the start of the boot year, written as the counter's initial value.

  The seed spaces independent boots within the same year into different regions of the counter range, preserving rough chronological ordering across boots and reboots.

  ID generation

  id = (year_prefix << 51) | atomic_increment(counter)

  A single atomic fetch-and-add per ID. No clock read after boot.

  Behaviour past the boot year

  The year field is the boot year, not the current wall-clock year — it is a fixed epoch tag, not a live clock. When a server runs past the calendar year boundary:

  - The prefix stays at its boot value; the counter simply continues climbing in the same bucket.
  - Nothing special happens at the boundary — no rollover, no reseed. IDs remain unique and monotonic.
  - The only ceiling is exhaustion of the 51-bit counter (see capacity below), which takes decades at any realistic rate. A long-running server is unaffected.
  - A reboot adopts the then-current year as its new prefix, moving to a fresh bucket and a fresh seed.

  Consequently a server may run indefinitely across many calendar years on a single boot prefix; the prefix distinguishes boot epochs rather than tracking the wall clock.

  Capacity

  - Per boot bucket: 2⁵¹ = 2,251,799,813,685,248 (~2.25 quadrillion) IDs.
  - Worst case — booting at the end of a year, seed ≈ a full year of µs (3.15×10¹³): ~2.22 quadrillion IDs still available. The bucket dwarfs a year of microseconds, so even an end-of-year boot consumes ~1.4% of the range.
  - Aggregate across all 256 year prefixes: 2⁵⁹ = 5.76 × 10¹⁷ IDs.

  Throughput

  - Bucket-exhaustion ceiling: 2⁵¹ ÷ seconds-per-year ≈ 71.4 million IDs/sec sustained for a full year before a bucket is exhausted (~70× real time).
  - Collision-avoidance ceiling: seeds are microsecond-spaced, so boots one second apart start 1,000,000 apart. A boot stays clear of a later same-year boot's region as long as it averages under 1 million IDs/sec.

  Constants
```
  ┌────────────────────────────────┬───────────────────────────────┐
  │           Parameter            │             Value             │
  ├────────────────────────────────┼───────────────────────────────┤
  │ Epoch                          │ start of 2026                 │
  ├────────────────────────────────┼───────────────────────────────┤
  │ Total ID width                 │ 59 bits (positive fixnum)     │
  ├────────────────────────────────┼───────────────────────────────┤
  │ Year field                     │ 8 bits (2026–2281, 256 years) │
  ├────────────────────────────────┼───────────────────────────────┤
  │ Counter field                  │ 51 bits                       │
  ├────────────────────────────────┼───────────────────────────────┤
  │ IDs per boot bucket            │ 2,251,799,813,685,248         │
  ├────────────────────────────────┼───────────────────────────────┤
  │ Sustained rate (1-year bucket) │ ~71.4 M/sec                   │
  ├────────────────────────────────┼───────────────────────────────┤
  │ Safe rate (collision spacing)  │ ~1 M/sec                      │
  └────────────────────────────────┴───────────────────────────────┘
```

robacourt:
> @robacourt I don't think a 60 bit integer would limit us to ~300k shapes. I've thought about this a bit here and there and clauded this integer-only design:

Love this! I'm still trying to work out how we can guarantee there's no collisions though. If we do have a collision, users could see data they're not supposed to see (currently a hash of the shape is included in the handle which prevents this). Collisions sound highly unlikely but there are a few scenarios they could happen:
1. two servers could start in the same microsecond - but they can't get the lock at the same microsecond, so maybe it could be the time we acquire the lock
2. two servers could acquire the lock at different times but think it's the same microsecond because their clocks are out of sync (I was thinking we could get the microsecond time from postgres once we have the lock, but there doesn't seem to be a way to do this that's immune to system clock jumps)
3. The above two scenarios don't have to be the same microsecond, they can be close and the counter from one can increment to the timestamp of the other

robacourt:
Closed in favour of #4643

PR #4629, opened 2026-06-18
Fix reasoning section elapsed timers

KyleAMathews:
Fix reasoning-section elapsed timers so each thinking block measures from when that block starts, not from the beginning of the run. This prevents later “Thought for …” sections in long tool-using runs from showing inflated durations that grow across the whole run.

## Root Cause

`ReasoningBlock` received the parent run timestamp and used it for both the live elapsed timer and the final collapsed `Thought for …` label. In multi-step runs, later reasoning rows can start well after the run begins — for example after a tool call returns — so anchoring every row to the run start made each new reasoning section inherit all prior run time.

## Approach

- Make `ReasoningBlock` own its timing anchor locally.
- Capture the first render where that reasoning row is live.
- Use that captured timestamp for the live `ElapsedTime` display.
- Snapshot the final duration when the row transitions from `streaming` to `completed`.
- Remove the misleading parent `timestamp` prop from the reasoning block API.

Rows that mount already completed still render as bare `Thought`, because the UI does not know their true start/end duration.

## Key Invariants

- A reasoning block’s elapsed time is scoped to that block, not the whole run.
- Later reasoning blocks must not inherit time spent in previous thinking, text, or tool-call sections.
- Historical/completed reasoning rows should not invent durations from incomplete client-side timing data.
- The parent response timestamp remains available for response-level metadata, but not for per-reasoning timing.

## Non-goals

- This does not persist reasoning start/end timestamps in the backend schema.
- This does not change response-level `✓ done in …` timing.
- This does not alter reasoning ordering, markdown rendering, or redacted reasoning behavior.

## Trade-offs

This uses the client-observed first-live-render time rather than a persisted provider/server timestamp. That is intentionally scoped: it fixes the live UI inflation bug without adding schema or event-model changes. Historical rows still avoid showing made-up durations.

## Verification

```bash
cd packages/agents-server-ui
pnpm test src/components/ReasoningSection.test.tsx
pnpm typecheck
```

Also validated changeset coverage:

```bash
```

## Files changed

- `.changeset/fix-reasoning-timers.md`
  - Adds a patch changeset for `@electric-ax/agents-server-ui`.

- `packages/agents-server-ui/src/components/AgentResponse.tsx`
  - Stops passing the parent run timestamp into `ReasoningBlock`.

- `packages/agents-server-ui/src/components/ReasoningSection.tsx`
  - Captures per-reasoning-row live start time.
  - Uses that local start time for live and settled reasoning duration labels.
  - Removes the `timestamp` prop from `ReasoningBlock`.

- `packages/agents-server-ui/src/components/ReasoningSection.test.tsx`
  - Adds regression coverage proving live reasoning timers do not start from the run timestamp.
  - Adds client-rendered coverage for the live → completed transition producing `Thought for Ns`.

### DIFF
--- a/.changeset/fix-reasoning-timers.md
+++ b/.changeset/fix-reasoning-timers.md
@@ -1,0 +1,5 @@
+---
+'@electric-ax/agents-server-ui': patch
+---
+
+Fix reasoning section elapsed timers so each thinking block measures from when that block starts, rather than from the beginning of the run.
--- a/packages/agents-server-ui/src/components/AgentResponse.tsx
+++ b/packages/agents-server-ui/src/components/AgentResponse.tsx
@@ -666,7 +666,6 @@ export const AgentResponseLive = memo(function AgentResponseLive({
               key={entry.key}
               entry={entry.reasoning}
               isStreaming={isStreaming}
-              timestamp={timestamp}
               expanded={Boolean(expandedReasoning[entry.key])}
               onToggle={toggleReasoning}
             />
--- a/packages/agents-server-ui/src/components/ReasoningSection.test.tsx
+++ b/packages/agents-server-ui/src/components/ReasoningSection.test.tsx
@@ -1,0 +1,92 @@
+// @vitest-environment jsdom
+
+import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest'
+import { act } from 'react'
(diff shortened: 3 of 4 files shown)


kevin-dp (approved):
Nice improvement. Always love a one line fix to the logic 💯

PR #4636, opened 2026-06-19
Emit per-individual-table ETS memory/size telemetry

erik-the-implementer:
## What

Adds per-individual-table ETS memory/size telemetry so we can drill into *which specific tables* dominate ETS memory in Honeycomb, beyond the existing per-`table_type` aggregate.

## Details

- New `[:ets, :table]` telemetry event emitting **`ets.table.memory`** and **`ets.table.size`**, tagged by `table_name` + `table_type`, for the top N tables by memory.
- N is the new `top_ets_individual_count` option (default 10).
- Reuses the already-present `EtsTables.top_tables/1` (no new gathering code); the existing `top_by_type` / `ets.memory.total` path is untouched.
- The per-table emitter does its own `:ets.all()` walk (kept separate from the type rollup so the two counts stay independently configurable — a cheap second scan, not worth disturbing the working type path to merge).

Two follow-up cleanup commits remove redundant `to_string/1` calls on telemetry tag values for the ETS and process emitters. Every reporter that carries these tagged metrics (OTel, Prometheus, StatsD) stringifies tag values itself, and the CallHomeReporter doesn't carry them at all — so the call-site `to_string/1` was a no-op. Tag values are now passed through raw (atom table name, string type).

## Testing

- Added an emission test asserting `[:ets, :table]` fires with `memory`/`size` measurements and `table_name`/`table_type` metadata.
- Full `electric-telemetry` suite: 124 tests, 0 failures; clean `mix compile --warnings-as-errors`.

Changeset included (`@core/electric-telemetry`, patch).

### DIFF
--- a/.changeset/ets-per-table-memory-metrics.md
+++ b/.changeset/ets-per-table-memory-metrics.md
@@ -1,0 +1,5 @@
+---
+'@core/electric-telemetry': patch
+---
+
+Emit per-individual-table ETS memory/size telemetry (`ets.table.memory` and `ets.table.size`, tagged by `table_name` and `table_type`) for the top N tables by memory, controlled by the new `top_ets_individual_count` option (default 10). Complements the existing per-`table_type` `ets.memory.total` aggregate.
--- a/packages/electric-telemetry/lib/electric/telemetry/application_telemetry.ex
+++ b/packages/electric-telemetry/lib/electric/telemetry/application_telemetry.ex
@@ -43,9 +43,10 @@ defmodule ElectricTelemetry.ApplicationTelemetry do
     metrics = metrics(opts)
 
     # `ets.table.*` are high-cardinality per-table gauges tagged by the raw table name
-    # (which embeds per-shape/stack ids). That's the intended trade-off for OTel/Honeycomb
-    # and is harmless for StatsD (series are cleared between OTel exports and sent per-scrape
-    # over StatsD), but `TelemetryMetricsPrometheus.Core` keeps every series in its registry
+    # (which embeds per-shape/stack ids). That rotating top-N set is the intended trade-off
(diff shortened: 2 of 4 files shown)


erik-the-implementer:
Thanks — addressed the cardinality issue and verified the rest.

**Important #1 (Prometheus cardinality) — fixed** in `e653bf570`. `ets.table.*` are now excluded from the Prometheus reporter's metric list and flow to OTel + StatsD only; the bounded `ets.memory.total` (by `table_type`) stays on Prometheus. Added a routing test asserting exactly that. I confirmed the asymmetry is correct: the OTel exporter clears its series between exports (`metric_store.ex` `:ets.match_delete` on the exported generation), and StatsD is per-scrape, so a rotating top-N set is harmless there — only `TelemetryMetricsPrometheus.Core` keeps series with no TTL, which is the leak/stale-gauge you flagged.

**Suggestion #1 (double `:ets.all()` scan) — kept as-is, deliberately.** `top_memory_stats/2` does exist and would do one scan with independent counts; I'm keeping the two emitters separate to avoid rewriting the working `ets_memory/1` path. Noted for revisiting if poller latency shows up on very-high-shape nodes.

**Suggestion #2 (to_string at reporter boundary) — verified by source.** Prometheus `exporter.escape/1` begins with `to_string/1`, OTel `OtlpUtils.to_kv_value/1` has an `is_atom` clause, and both StatsD formatters stringify. With `ets.table.*` now off Prometheus, the only atom tag reaching Prometheus is `process_type`, which `escape/1` handles.

**Suggestion #3 (option naming) — left as-is** to avoid churning the existing `top_ets_table_count`; both are default-only opts.

No linked issue — this is a small telemetry enhancement rather than a tracked bug.

PR #4643, opened 2026-06-23
Shrink subquery filter index memory via integer interning

robacourt:
## Summary

Shrinks the memory footprint of the **subquery filter index** (`Electric.Shapes.Filter.Indexes.SubqueryIndex`). The per-value ETS rows that back subquery routing and exact membership repeated several boxed terms in their keys; each is now a compact integer id, leaving only the actual typed value boxed. **No change to routing behaviour** — purely a representation change.

End to end this cuts the cost per seeded subquery membership value by **~51%** (e.g. **~568 MiB → ~275 MiB for 1,000,000 seeded values**).

## How to review

The work is split into one commit per independent idea — **reviewing commit-by-commit is the easiest path**, each is self-contained and verified by the memory test:

1. **`Add memory test`** — adds a `:memory`-tagged test in `filter_test.exs` that seeds the index with realistic shape handles (built like `Shape.generate_id`, ~25-byte binaries that occupy 64 B per ETS row) and UUID values, and prints a per-tag memory breakdown. This is the measurement harness the rest of the PR is graded against. Run with `mix test --include memory test/electric/shapes/filter_test.exs`.
2. **`Use integer condition ids instead of make_ref()`** — condition ids are only allocated by the single filter-owning process, so a filter-local integer counter replaces `make_ref/0` (a boxed ref costs ~3 words per row that embeds one). Also shrinks eq/incl index rows.
3. **`Key subquery membership rows by dependency index, not the ref list`** — the `["$sublink","<dep>"]` ref list (two boxed binaries) was the heaviest term in membership rows. The dependency index is already passed to every writer and encoded in the ref, so we key by it and recover it on the two read paths — no mapping table, no new concurrency surface.
4. **`Intern shape handles to integer ids`** — each handle is interned to a small integer via `:erlang.unique_integer/1` (forward-only, no reverse map; `insert_new` makes it idempotent + race-safe), and the handle string is stored once in a `{:handle_id, handle}` row instead of in every per-value row.
5. **`Intern subquery node ids to integers`** — the `{condition_id, field}` node id (a tuple with a boxed `field` binary) repeated in every per-value node-member row is replaced by a node integer. It's allocated in `ensure_node_meta` and read back from the `node_meta` row the routing path already reads, so **no extra hot-path lookup**.

The last commit is the changeset.

After these, the two dominant per-value rows hold exactly one boxed term — the typed value (the actual data):
- `:membership` → `{{:membership, sid, dep_index, value}, true}` — 240 → 96 B
- `:node_positive_member` → `{{tag, node_int, value, sid, ncid}, true}` — 264 → 104 B

## Benchmarks

Real `:ets.info(:memory)` for the subquery index, before (pre-optimization) vs after, across scale and shapes:values ratio:

| scenario | seeded values | before | after | per-value before → after |
|---|---:|---:|---:|---|
| 1k shapes × 50 | 50,000 | 28.9 MiB | 14.1 MiB | 606.6 → 296.2 B (−51%) |
| 5k shapes × 50 | 250,000 | 144.6 MiB | 70.6 MiB | 606.6 → 296.2 B (−51%) |
| 10k shapes × 100 | 1,000,000 | 567.7 MiB | 274.7 MiB | 595.3 → 288.1 B (−52%) |
| 1k shapes × 1000 | 1,000,000 | 558.0 MiB | 267.8 MiB | 585.1 → 280.8 B (−52%) |
| 200 shapes × 5000 | 1,000,000 | 557.2 MiB | 267.2 MiB | 584.2 → 280.2 B (−52%) |

Per-row deltas (50k scenario): `:membership` 240 → 96 B, `:node_positive_member` 264 → 104 B.

**Scale doesn't change the per-value saving** (cost is linear in total seeded values). The shapes:values ratio has a small effect — more values per shape amortizes the one-per-shape `{:handle_id, handle}` row, so high-fan-out shapes land slightly better (~52% vs ~51%).

---

### DIFF
--- a/.changeset/shrink-subquery-index.md
+++ b/.changeset/shrink-subquery-index.md
@@ -1,0 +1,12 @@
+---
+'@core/sync-service': patch
+---
+
+Reduce the memory footprint of the subquery filter index. The per-value ETS
+rows that back subquery routing and exact membership repeated several boxed
+terms in their keys — the shape handle, `make_ref/0` condition references, the
+canonical `["$sublink", "<dep>"]` subquery ref, and the `{condition_id, field}`
+node id. Each of those is now a compact integer id, leaving only the actual
+typed value boxed. This cuts the cost per seeded subquery membership value by
+~51% (e.g. ~568 MiB → ~275 MiB for 1,000,000 seeded values) with no change to
+routing behaviour.
--- a/packages/sync-service/lib/electric/shapes/filter.ex
+++ b/packages/sync-service/lib/electric/shapes/filter.ex
@@ -33,22 +33,48 @@ defmodule Electric.Shapes.Filter do
     :where_cond_table,
     :eq_index_table,
     :incl_index_table,
-    :subquery_index
+    :subquery_index,
+    :counters_table
   ]
 
   @type t :: %Filter{}
   @type shape_id :: any()
 
+  # Condition ids identify nodes in the where-condition tree (table conditions,
+  # equality/inclusion/subquery sub-conditions). They were `make_ref/0`s, but a
+  # boxed reference costs ~3 words on every ETS row that embeds one. They are
+  # only ever allocated by the single process that owns the filter (the
+  # EventRouter, via add_shape/remove_shape), so a filter-local integer counter
+  # gives the same uniqueness as a reference for ~0 bytes (small ints are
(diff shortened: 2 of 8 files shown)


magnetised (approved):
Nice work.

PR #4650, opened 2026-06-24
Reset replication metadata caches on new/temporary slot shape purge

robacourt:
## Summary

When Electric purges all shapes because the replication slot was (re)created — e.g. a **temporary** slot recreated on restart with `CLEANUP_REPLICATION_SLOTS_ON_SHUTDOWN=true` — or because the timeline changed, it now also resets the two replication-metadata caches that previously survived and were read back on startup.

## Problem

With a temporary slot, every restart drops the old slot and creates a new one, which triggers a full shape purge (storage `cleanup_all!` + `ShapeStatus.reset`). But two persistent caches were left untouched and **read on startup**, so they could be inconsistent with the freshly-created slot (WAL continuity was intentionally discarded):

- `<stack_id>:tracked_relations` — read by `ShapeLogCollector` on init via `PersistentReplicationState.get_tracked_relations/1`. It was only reset on a **timeline change**, not on the new-slot purge path.
- `<stack_id>:ets_inspector_state` — restored by `EtsInspector` on init. Never reset on either path.

## Changes

- `connection/manager.ex` — move `PersistentReplicationState.reset/1` (tracked_relations) out of the timeline-only block into the **shared purge block**, so it also runs on the new-slot / temporary-slot path; add `Electric.Postgres.Inspector.reset(state.inspector)` to clear the inspector cache in the same block.
- `postgres/inspector/ets_inspector.ex` — new public `reset/1` + `handle_call(:reset, …)` that empties the in-memory ETS table and re-persists empty state so the cleared cache survives a restart.
- `postgres/inspector.ex` — new `reset/1` behaviour callback + `{module, opts}` dispatch wrapper (mirrors `clean/2`).
- `postgres/inspector/direct_inspector.ex` + `test/support/stub_inspector.ex` — no-op `reset/1` (no cache).

On a new slot, pgoutput resends `Relation` messages before any DML, so clearing `tracked_relations` is safe (it repopulates); the inspector is a write-through cache reconciled against Postgres, so clearing is conservative.

## Test Plan

- [x] `ets_inspector_test.exs` — `reset/1` drops the entire ETS cache, and a restart does not restore the cleared cache.
- [x] `connection/manager_test.exs` — seeds the inspector cache + non-empty tracked_relations + a non-matching timeline, then asserts both are reset once the manager comes up (exercises the shared purge block the new-slot path uses).
- [x] Affected suites green: manager, inspector, shape_log_collector (104 tests).

---

### DIFF
--- a/.changeset/wipe-replication-metadata-on-slot-purge.md
+++ b/.changeset/wipe-replication-metadata-on-slot-purge.md
@@ -1,0 +1,12 @@
+---
+'@core/sync-service': patch
+---
+
+Reset replication metadata caches when shapes are purged on a new or temporary
+replication slot. Previously, when Electric purged all shapes after the
+replication slot was (re)created — e.g. a temporary slot recreated on restart
+with `CLEANUP_REPLICATION_SLOTS_ON_SHUTDOWN=true` — the persisted relation
+tracking (`tracked_relations`) and the inspector cache (`ets_inspector_state`)
+survived and were read back on startup, even though WAL continuity had been
+lost. These caches are now reset as part of the same purge, so no stale
+pre-restart metadata is reused.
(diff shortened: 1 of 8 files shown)


alco (approved):
LGTM!

I would like to echo Claude's piece of feedback regarding the missing test coverage for the specific case of "reset in case of slot invalidation":
> Test exercises the timeline-change path as a proxy for the new-slot path.

PR #4648, opened 2026-06-24
(DRAFT) Oracle restore tests - and example fixes

robacourt:
## What this does

Adds a restart-aware mode to the oracle property test, which restarts the
`Electric.StackSupervisor` mid-test to exercise restore-from-file. Wiring up
`RESTART_SERVER_EVERY` surfaced a number of real restore/restart bugs in the
sync service; this PR includes deterministic regression tests
(`test/integration/oracle_restore_test.exs`) plus example fixes for several of
them (subquery materializer restore, multi-chunk log replay, duplicate/orphan
ops after restart, mid-restart cleanup, coordinated shutdown).

## Bug doc

Each issue uncovered by the new test — symptom, root cause, fix, and regression
test — is written up in [`packages/sync-service/bugs.md`](https://github.com/electric-sql/electric/pull/4648/changes#diff-c2744ab82dd6dac5f05bb541c439ff01b1a385e811828fccd46267b77cedb737).
Start there for the full triage.

PR #4651, opened 2026-06-24
Restore subquery materializer state from disk after server restart

robacourt:
## Summary

Fixes **bug 1** from the restart-aware oracle property test triage (see #4648 / `bugs.md`): subquery shapes diverge from Postgres after a `StackSupervisor` restart that restores from disk.

A shape whose where clause contains a subquery — e.g. `project_id IN (SELECT id FROM projects WHERE active = true)` — could, after a server restart, return stale results (missing rows) or a `409 must-refetch` on an `optimized: true` shape.

## Problem

Two compounding root causes:

1. **The dependency materializer only replayed the first chunk of the source shape's persisted history on startup.** `Storage.get_log_stream/3` returns at most one snapshot chunk per call; the materializer called it once and treated that as the full history, silently dropping later snapshot chunks and persisted log entries. Its `value_counts` was wrong from the moment it came up.

2. **Dependent (outer) subquery consumers were not restarted after a stack restart.** The router only delivers events to a consumer when its own `root_table` changes, so a shape whose movement is driven entirely by its dependency stayed dormant — its materializer was never started, the SubqueryIndex never seeded, and its on-disk view never updated when the dependency changed.

## Solution

- **`materializer.ex`** — replace the single-call `get_stream_up_to_offset/3` with `read_history_up_to_subscribed/2`, which iterates snapshot chunks (via `Storage.get_chunk_end_log_offset/2`) up to the subscribed offset, stopping the moment the read steps into the main log so it never re-applies already-applied entries.
- **`shape_cache.ex`** — in `handle_continue(:wait_for_restore, …)`, eagerly restart consumers for every shape with `shape_dependencies != []` (before `ShapeLogCollector.mark_as_ready`), re-establishing the materializer subscription before live events flow.

## Test Plan

- [x] New integration repro `test/integration/oracle_restore_test.exs` (`oracle_restore_bug_1`) — fails on `main` (client view missing the reactivated project's issues), passes with the fix. Uses a readable `projects → issues` domain schema.
- [x] New materializer unit test (`startup history replay`) — directly pins root cause #1; fails on the pre-fix single-chunk read (`value_counts = {10}` instead of `{99}`).
- [x] New `shape_cache_test` describe block (`wait_for_restore eager subquery consumer start`) — asserts subquery shapes are eagerly restored while non-subquery shapes are not; existing `start_consumer_for_handle` test updated for the new eager behavior.
- [x] `mix test test/electric/shapes/` — 572 tests, 0 failures.

Run the focused repro:

```sh
cd packages/sync-service
mix test --only oracle test/integration/oracle_restore_test.exs
```

---

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
@@ -225,6 +225,13 @@ defmodule Electric.Shapes.Consumer.Materializer do
             is_log_offset_lte(next_offset, state.offset) ->
               # Defensive: chunk_end did not advance. Stop to avoid an
               # infinite loop. This shouldn't happen in normal operation.
+              Logger.warning(
+                "Materializer chunk iteration did not advance past " <>
+                  "#{inspect(state.offset)} (chunk_end=#{inspect(next_offset)}); " <>
+                  "stopping replay at subscribed_offset to avoid an infinite loop",
+                shape_handle: state.shape_handle
+              )
+
               %{state | offset: state.subscribed_offset}
 
             is_log_offset_lte(state.subscribed_offset, next_offset) ->
--- a/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
@@ -171,33 +171,77 @@ defmodule Electric.Shapes.Consumer.Materializer do
   end
 
   def handle_continue({:read_stream, storage}, state) do
-    {:ok, offset, stream} =
-      get_stream_up_to_offset(state.offset, state.subscribed_offset, storage)
-
-    {state, _} =
-      stream
-      |> decode_json_stream()
-      |> apply_changes(state)
-
+    state = read_history_up_to_subscribed(state, storage)
     write_link_values(state)
-
-    {:noreply, %{state | offset: offset}}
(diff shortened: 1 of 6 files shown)


packages/sync-service/lib/electric/shapes/consumer/materializer.ex
```diff
@@ -171,33 +171,77 @@ defmodule Electric.Shapes.Consumer.Materializer do
...
+          next_offset = Storage.get_chunk_end_log_offset(state.offset, storage)
+
+          cond do
+            is_nil(next_offset) ->
+              # No further chunks past this offset — we've reached the end.
+              %{state | offset: state.subscribed_offset}
+
+            is_log_offset_lte(next_offset, state.offset) ->
+              # Defensive: chunk_end did not advance. Stop to avoid an
+              # infinite loop. This shouldn't happen in normal operation.
```
magnetised, line 227: Logger.warning...?

magnetised (approved):
Cool

packages/sync-service/lib/electric/shapes/consumer/materializer.ex
robacourt: Done — added a `Logger.warning` in this defensive branch (fe34afd57) so the non-advancing chunk_end condition is surfaced if it ever happens.

PR #4654, opened 2026-06-25
Fix wake reconnects without abort reasons

### LINKED ISSUE #4497: Client does not resume syncing when a react native application comes back to the foreground
**Versions**

- Client: react-native 0.8X.X
- Electric: 1.5.16

**Bug description**
When a react native using @electric-sql/client is backgrounded the underlying http connections are killed. When the application is brought back to the foreground the @electric-sql/client instance does not resume syncing.

**Expected behavior**
A react native client coming back to the foreground should trigger @electric-sql/client to resume syncing so the client sees fresh data.

KyleAMathews:
Fixes a ShapeStream reconnect path so system-wake and force-refresh aborts restart even in runtimes that do not preserve `AbortSignal.reason`. This should make React Native/mobile foreground resumes more reliable and prevent affected shape streams from silently stopping until app reload.

Fixes https://github.com/electric-sql/electric/issues/4497

## Root Cause

`ShapeStream` intentionally aborts an in-flight live request when it needs to reconnect, including after wake detection sees a long timer gap:

```ts
controller.abort(SYSTEM_WAKE)
```

The request loop then decided whether to recurse/restart by checking `requestAbortController.signal.reason` for `SYSTEM_WAKE` or `FORCE_DISCONNECT_AND_REFRESH`.

Some non-browser runtimes, notably React Native/fetch polyfill combinations, may abort the request but not preserve `AbortSignal.reason`. In that case the fetch rejects as aborted, but the client does not recognize it as a restart-worthy abort. The loop can then return normally, leaving the stream subscribed but no longer issuing HTTP requests. That manifests as missing backend changes and `awaitTxId` timeouts until a full app reload creates fresh streams.

## Approach

Track restart intent on the `AbortController` instance before calling `abort(...)`, instead of relying solely on the runtime preserving `AbortSignal.reason`.

```ts
this.#restartAbortControllers.add(requestAbortController)
requestAbortController.abort(SYSTEM_WAKE)
```

The catch path now treats an abort as restart-worthy when either:

- the controller was marked in the internal `WeakSet`, or
- the runtime preserved the historical abort reason strings.

The wake path only marks a controller that is not already aborted. This preserves the previous semantics where an already-aborted request cannot be retroactively reclassified as a wake restart.

## Key Invariants

- `forceDisconnectAndRefresh()` and wake detection must restart the request loop after aborting a live request.
- Pause/snapshot/external aborts must remain ordinary interruptions and must not be mistaken for reconnect requests.
- Wake detection must not mark an already-aborted controller as restart-worthy.
- Existing runtimes that preserve `AbortSignal.reason` must continue to work.
- The restart marker should not retain controllers after the catch path observes them.

## Non-goals

- This does not add a general request timeout for native fetch implementations that never settle after abort.
- This does not expose `forceDisconnectAndRefresh()` through TanStack DB collection utilities.
- This does not change Electric protocol parameters, shape semantics, or txid handling.

## Trade-offs

An alternative was to polyfill `AbortSignal.reason` or broaden restart detection to any aborted request. A polyfill would not help all consumers, and treating all aborted requests as restarts risks resuming after pause/snapshot/external interruptions. Marking the specific controllers that Electric aborts for restart is narrow and runtime-independent.

## Verification

```bash
cd packages/typescript-client
pnpm run typecheck
pnpm exec vitest --config vitest.unit.config.ts --run test/wake-detection.test.ts
```

Also verified changeset coverage:

```bash
```

All passed locally.

## Files changed

- `.changeset/rn-wake-restart.md` — patch changeset for `@electric-sql/client`.
- `packages/typescript-client/src/client.ts` — tracks restart-intent controllers, uses that marker in the abort catch path, and avoids marking already-aborted wake requests.
- `packages/typescript-client/test/wake-detection.test.ts` — adds a regression test simulating runtimes that drop abort reasons.
- `packages/typescript-client/SPEC.md` — updates the request-loop restart path documentation.


kevin-dp (approved):
LGTM!

PR #4656, opened 2026-06-26
chore(durable-streams-rust): rename folder, drop Intel macOS build, fix README

balegas:
## What

Three related cleanups to the Rust Durable Streams server packaging:

1. **Rename the folder** `packages/server-rust` → `packages/durable-streams-rust`.
2. **Drop the Intel macOS (`macos-13`) build** from the publish matrix.
3. **Fix the README** write-throughput figure and remove the per-stream results table.

## Does this affect CI / publishing?

No — the rename is path-only. The **npm package name** (`@electric-ax/durable-streams-server-rust`), the **crate name** (`durable-streams`), and the **binary name** (`durable-streams-server`) are all unchanged, so published artifacts and their consumers are unaffected. Updated every path reference:

- CI workflows (`server_rust_tests`, `server_rust_publish`, `server_rust_dockerhub_image`, `changesets_release`)
- `Dockerfile` COPY paths, npm package templates' `repository.directory`, conformance test comments
- `pnpm-lock.yaml` importer key (regenerated via `pnpm install --lockfile-only`)

`pnpm-workspace.yaml` globs `packages/*`, so workspace resolution is transparent to the move. `git mv` preserved file history (tracked as renames).

## Why drop macos-13

The Intel macOS runner pool is scarce and stalls releases (a publish run sat queued waiting for it). That binary only feeds the **currently-disabled npm packages**, so it's not on any shipping path today. Re-add it when npm publishing is re-enabled. `cargo-publish` stays gated on the `build` job, which now runs the two Linux targets plus the readily-available Apple-Silicon `macos-latest`.

## README

`~0.86M append/s` → `860,000 append/s`; removed the per-stream results table (kept the prose summary).


samwillis (approved):
LGTM

PR #4658, opened 2026-06-28
fix(electric-telemetry):  Fix process_type parsing for request processes that have a request id that is not the standard 20 bytes.

robacourt:
## Summary

Fixes an unbounded `process_type` cardinality leak in telemetry for request-handling processes.

## Problem

`process_type` for request-handling processes is derived from the process label set by `Electric.Plug.LabelProcessPlug`, which looks like:

```
Request F-jPUudNHxbD8lIAABQG - GET /v1/shape?table=users&offset=-1
```

`parse_binary_label/1` is meant to strip the request id and query, collapsing this to a bounded `GET /v1/shape`. But it did so with a fixed-width binary match that assumes the request id is exactly 20 bytes:

```elixir
"Request " <> <<_req_id::binary-20, " - ", rest::binary>> -> ...
```

That holds for ids `Plug.RequestId` self-generates, but `Plug.RequestId` reuses an incoming `x-request-id` header when a client/proxy supplies one. A UUID (36 chars), an Envoy/nginx id, or an LB trace id is not 20 bytes, so the clause falls through to the generic branch that returns the first 20 bytes of the whole label — i.e. `Request <first 12 chars of the id>`. Every distinct incoming request id that GCs slowly then adds a permanent new series to `prometheus_metrics_dist` for events like `vm.monitor.long_gc` and `vm.monitor.long_schedule`. Request handlers building large shape responses are exactly the processes that trip those monitors, so the table grows without bound when a fronting proxy injects `x-request-id`.

## Solution

Keep the fast fixed-width binary match for the common 20-byte case (`Plug.RequestId`'s self-generated ids), and add a slow path that only runs when the id is a different length: it splits on the `" - "` delimiter regardless of request-id length, then strips the query string. All requests to a route collapse to a single `process_type` (e.g. `GET /v1/shape`) regardless of request-id source or length. A `"Request "` label with no delimiter falls back to truncation.

---

## Also includes: unrelated CI fix

CI for this PR was failing on `Electric.ShapeCacheTest` (in `sync-service`, unrelated to the telemetry change). The cause is a semantic merge collision already on `main`: #4635 changed `wait_until`'s signature to `wait_until(fun, timeout_ms)` (function first), while #4651 added a new call site at `shape_cache_test.exs:1582` using the old arg order `wait_until(1000, fn -> ... end)`. The two merged cleanly in git but not in meaning, so the slow suite raised a `FunctionClauseError` on every run. This PR swaps the arguments to match the current signature, unblocking both `main` and this PR.


packages/electric-telemetry/lib/electric/telemetry/processes.ex
```diff
@@ -262,26 +262,43 @@ defmodule ElectricTelemetry.Processes do
...
+          _ -> truncate_label(label)
+        end
 
-        # Cut off the URL query part, leaving only the `<method> <path>` prefix.
-        part_len =
-          case :binary.match(rest, "?") do
-            {pos, _} ->
-              pos
+      _ ->
+        truncate_label(label)
```
robacourt, line 280: I'm wondering if we should have a fixed label, say "Unparseable Request" and to log a warning with the full label, to prevent an unbounded process type in this scenario

alco: What specific problems does unbounded process type cardinality cause? Is it controlled by the client, i.e. does it have a DoS potential?

alco (approved):
👍

packages/electric-telemetry/lib/electric/telemetry/processes.ex
robacourt: It came up as DoS potential when investigating promethius issues, but it turns out the issue was to do with promethus library not aggregating until it is drained. It doesn't matter whether the label is unbounded or not, if the prometheus metrics are not aggregated, they grow unbounded. So I'll leave this as is.

.changeset/process-type-request-id-cardinality.md
```diff
@@ -0,0 +1,5 @@
+---
+"@core/electric-telemetry": patch
+---
+
+Fix unbounded `process_type` cardinality on request-handling processes. `parse_binary_label/1` only stripped the request id from process labels when it was exactly 20 bytes (the length `Plug.RequestId` self-generates). A proxy- or load-balancer-supplied `x-request-id` (UUID, Envoy/nginx trace id, etc.) has a different length, so the raw per-request id leaked into `process_type`, adding a permanent new series to telemetry events like `vm.monitor.long_gc` and `vm.monitor.long_schedule` for every distinct id. The label is now parsed by splitting on the `" - "` delimiter regardless of request-id length, collapsing all requests to a route back to a single `process_type` (e.g. `GET /v1/shape`).
```
robacourt:
```suggestion
Fix `process_type` parsing for request processes that have a request id that is not the standard 20 bytes.
```

PR #4660, opened 2026-06-29
docs: warn that Prometheus /metrics must be scraped when enabled

robacourt:
## Summary

If `ELECTRIC_PROMETHEUS_PORT` is set but nothing ever scrapes the `/metrics` endpoint, Electric's memory grows without bound and can eventually OOM-kill the service. This came out of debugging a customer whose `prometheus_metrics_dist` ETS table had grown to ~8 GB on an OTel-only deployment that had the Prometheus port enabled but unscraped.

### Why it happens

`telemetry_metrics_prometheus_core` stores **distribution (histogram)** metrics in a `:duplicate_bag` and appends one row per observation (`distribution.ex` `:ets.insert`), only aggregating into buckets — and draining the bag via `:ets.take` — **at scrape time** (documented in the library's moduledoc: *"aggregations for distributions (histogram) only occur at scrape time"*). So if the endpoint is never scraped, the bag is never drained.

Electric's default Prometheus metric set includes `electric.postgres.replication.transaction_received.receive_lag`, a distribution emitted roughly once per replication transaction, so the table accretes continuously under load (~0.55 GB/day in the customer's case). Counters/sums/last-values are unaffected — they use bounded `:set` storage.

This is purely a docs change to make the "must be scraped" requirement explicit everywhere we mention Prometheus.

## Changes

- **Telemetry reference** (`website/docs/sync/reference/telemetry.md`) — warning callout in the Metrics section.
- **Config reference** (`website/docs/sync/api/config.md`) — note on `ELECTRIC_PROMETHEUS_PORT`.
- **Deployment guide** (`website/docs/sync/guides/deployment.md`) — warning in the Observability section.
- **`prometheus_port` docstring** (`packages/sync-service/lib/electric.ex`).

## Follow-up (not in this PR)

Worth considering a code-level guard so an enabled-but-unscraped endpoint can't OOM the service (e.g. move high-frequency distributions like `receive_lag` out of the default Prometheus set, or run an internal periodic drain).

I have raised a PR with the library maintainer for a potential fix: https://github.com/beam-telemetry/telemetry_metrics_prometheus_core/pull/77

### DIFF
--- a/packages/sync-service/lib/electric.ex
+++ b/packages/sync-service/lib/electric.ex
@@ -103,7 +103,7 @@ defmodule Electric do
     enable tracking of instance usage metrics across restarts, otherwise will be
     randomly generated at boot (default: a randomly generated UUID).
   - `telemetry_statsd_host` (`t:String.t/0`) - If set, send telemetry data to the given StatsD reporting endpoint (default: `#{default.(:telemetry_statsd_host)}`)
-  - `prometheus_port` (`t:integer/0`) - If set, expose a prometheus reporter for telemetry data on the specified port (default: `#{default.(:prometheus_port)}`)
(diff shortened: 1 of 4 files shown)


alco (approved):
Great find!

PR #4663, opened 2026-06-29
fix(telemetry): bound Prometheus distribution buffer (fork pin)

robacourt:
## Summary

Fixes unbounded ETS growth in the Prometheus metrics reporter. When `ELECTRIC_PROMETHEUS_PORT` is set but the `/metrics` endpoint is scraped infrequently — or never, as on OpenTelemetry-only deployments — `telemetry_metrics_prometheus_core` buffers **one ETS row per distribution observation** and only drains it on scrape. The per-transaction `receive_lag` distribution dominates, so the `prometheus_metrics_dist` table grows without bound. A customer hit an **8 GB+** ETS table and eventual OOM this way.

This pins `telemetry_metrics_prometheus_core` to a fork that bounds the buffer by aggregating automatically — on a size threshold (default 10k buffered samples) and a time fallback (default 60s) — in addition to on scrape, and serializes aggregation in the registry process (also fixing a pre-existing overlapping-scrape race).

Upstream PR: https://github.com/beam-telemetry/telemetry_metrics_prometheus_core/pull/77

## Why a git dep is safe here

The telemetry deps (the `electric_telemetry` path dep and its transitive `telemetry_metrics_prometheus_core`) are gated behind `MIX_TARGET=application` in `sync-service/mix.exs` (`telemetry_deps(_) -> []` otherwise). The Hex-published `electric` package is built with the default target, so this dependency is **absent from the published dep tree** — the git pin cannot affect `mix hex.publish`. It only affects the standalone sync-service / Docker / CI build, where `mix deps.get` fetches git deps fine (git is installed in the builder image). `electric-telemetry` itself is not published to Hex.

## Changes

- `packages/electric-telemetry/mix.exs` — pin to the fork branch, with a comment to revert to `~> 1.2` once upstream lands on Hex.
- `packages/electric-telemetry/mix.lock` + `packages/sync-service/mix.lock` — locked to the fork commit (only that one entry changes).
- Changeset (`@core/sync-service` patch).

Verified `electric-telemetry` compiles against the fork. **Revert this pin to the Hex release after #77 merges and ships.**


robacourt:
Closed in favour of #4660 . We'll wait for https://github.com/beam-telemetry/telemetry_metrics_prometheus_core/pull/77 to be merged before using it.

PR #4666, opened 2026-06-30
Prevent shape removal during server restart

robacourt:
## Summary

Fixes **bug 6** from the restart-aware oracle property-test triage (#4648 / `bugs.md`): a healthy subquery shape returning a `409 (must-refetch)` after a `StackSupervisor` restart.

A single targeted fix, with a **deterministic regression test that fails without the fix and passes with it**. This is the part of the fix that was battle-tested in `rob/restore-subqueries-bug-5`.

## Problem

Bug 6 is a shutdown *race*. During a restart's stack teardown, a dependency consumer's inline `GenServer.call` into its materializer (`notify_materializer_of_new_changes/3`) races the materializer's death and exits with `:noproc`. The consumer crashes with a non-shutdown reason, which routes through `handle_writer_termination` and removes the shape from disk. Because the `ShapeLogCollector` is already gone mid-shutdown, the removal half-completes — the handle is deleted from SQLite but cleanup fails — so after the restart the handle is missing and a fresh poll on that `optimized: true` shape gets a 409.

## Solution

- **`consumer.ex`** — `notify_materializer_of_new_changes/3` catches the `:exit` (`:noproc`, and transient `:normal`/`:shutdown`) from its inline materializer call. The pending monitored `:DOWN` then drives a clean stop instead of the abnormal-shutdown path that removes the shape.

## Test Plan

- [x] `consumer_test.exs` — "dependency consumer survives a :noproc from its materializer without removing the shape". Without the fix: the consumer crashes with `{:noproc, …}` and logs *"Removing shape …"*.

---

### DIFF
--- a/.changeset/restore-shutdown-shape-removal.md
+++ b/.changeset/restore-shutdown-shape-removal.md
@@ -1,0 +1,9 @@
+---
+'@core/sync-service': patch
+---
+
+Stop subquery shapes from being spuriously removed during a server restart. When
+a dependency consumer's inline call to its materializer raced the materializer's
+shutdown, the resulting `:noproc` exit crashed the consumer and removed the shape
+from disk, causing a `409 must-refetch` after the restart. The consumer now
+absorbs that exit and lets the monitored `:DOWN` drive a clean stop.
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -1063,6 +1063,18 @@ defmodule Electric.Shapes.Consumer do
          opts
        ) do
     Materializer.new_changes(Map.take(state, [:stack_id, :shape_handle]), changes_or_bounds, opts)
+  catch
+    # The consumer monitors the materializer; if the materializer died the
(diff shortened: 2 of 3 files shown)


.changeset/restore-shutdown-shape-removal.md
```diff
@@ -0,0 +1,13 @@
...
+
+Stop subquery shapes from being spuriously removed (and held requests from
+crashing) during a server restart. When a dependency consumer's inline call to
+its materializer raced the materializer's shutdown, the resulting `:noproc`
+exit crashed the consumer and removed the shape from disk — leaving it
+half-removed and returning a `409 must-refetch` after the restart. The consumer
+now absorbs that exit and lets the monitored `:DOWN` drive a clean stop.
+Additionally, `validate_shape_handle/3` and the API's disk-update check now
+tolerate the brief window during a restart where the per-stack ETS tables are
+absent, falling back instead of raising an `ArgumentError`.
```
robacourt:
```suggestion
Stop subquery shapes from being spuriously removed (and held requests from
crashing) during a server restart.
```

alco (approved):
Nice one!

PR #4667, opened 2026-06-30
perf(durable-streams-rust): serve caught-up long-polls from the epoll reactor

balegas:
## What

Generalizes the Linux epoll reactor (renamed `sse_reactor` → `reactor`) to serve **two** reader modes from one shared slab/epoll/flush/wake core. A caught-up long-poll on a reactor-eligible stream now hands its socket to the reactor instead of parking the whole connection-task future for the up-to-30s wait — the same per-subscriber memory win the SSE reactor already gives live tails.

- The reactor writes the single long-poll response on data / close / timeout, then **hands the socket back for keep-alive** so polling clients reuse the connection (no per-poll churn). When the next request is another same-stream long-poll, it's served **in place** without leaving the reactor.
- `Kind::{Sse, LongPoll}` splits only what differs (what to produce on a wake, what to do when finished); slab, epoll arm/flush/backpressure, cross-thread wake and lifetime sweep are shared.
- Long-poll handlers were refactored into sync, reusable builders (`long_poll_data_resp`, `long_poll_close`, `long_poll_timeout`) so the inline path and the reactor emit byte-identical responses.
- Linux only; non-Linux and cold/forked/tiered streams keep the proven inline wait. SSE behaviour is unchanged.

## Socket-closure (`Connection: close`) check

The base SSE PR's final fix marks SSE responses `Connection: close` because SSE *unilaterally* closes the socket (close-instead-of-FIN → RST → discarded in-flight response on a pooling client). I verified the long-poll reactor does **not** have this class of bug, by construction:

- The advertised `Connection` header is tied to the **same** `keep_alive` flag that decides close-vs-handback (`write_head` emits `connection: close` iff `!keep_alive`; `finish` keeps the socket iff `keep_alive`). The disposition can never disagree with what was advertised.
- Every `close()` of a keep-alive socket happens with a drained recv buffer (the in-place reader reads to `EAGAIN` before deciding; parked closes carry no in-flight body), so it's a clean FIN, never an RST.

The reactor's SSE path inherits the `Connection: close` fix automatically (its head is built from `handle_sse`'s headers).

## Testing

- Linux cross-compile (`x86_64-unknown-linux-gnu`) and native builds pass.
- Reactor unit tests cover long-poll serialization (200 data wrap, 204 close → `connection: close`, 204 timeout → keep-alive).
- Conformance suite covers both SSE and long-poll against the containerized Linux server.


balegas:
## Simplified: dropped the in-reactor keep-alive request reader (design 1)

Per review, removed the ~200-line in-reactor HTTP request reader. After writing a keep-alive long-poll response the reactor now hands the socket back to the async `conn_loop`, which reads and dispatches the client's next request (re-registering with the reactor if it's another caught-up poll). This **reuses the existing, proven async keep-alive path** instead of maintaining a second request reader in the reactor.

**Net effect:**
- `reactor.rs` 1272 → 1073 lines (−199); removed `read_request`, `enter_reading_request`, `arm/disarm_epollin`, the `LpPhase` enum, the `EPOLLIN` path, and `parse_long_poll_continuation` (handlers).
- Diff vs `main` shrank from +1494/−709 to +661/−129.
- The only thing left mode-specific to long-poll is the response build (`longpoll_produce`/`tick`/`finalize`/`serialize`) + `handback`; slab/epoll/flush/backpressure/wake/lifetime-sweep are fully shared with SSE.

**Validation (Linux):**
- 94 unit tests pass; native build + clippy (`--all-targets`) clean.
- Conformance suite: **11 consecutive full runs, 326 passed each, 0 failures** (the closure/connection-reuse race historically reproduced ~1/19), including the long-poll "wait for new data" and SSE stream-closure tests.

PR #4668, opened 2026-07-01
Skip already-processed transactions on restart

robacourt:
## Summary

On restart, the persistent replication slot can replay transactions the consumer has already applied and persisted. The consumer deduplicated these on its multi-fragment path, but the complete-transaction **fast paths short-circuited before reaching that check** — so a replayed transaction was re-written to the shape log (duplicating ops) and re-notified to dependent subquery materializers, which re-applied it and crashed with `Key ... already exists`.

This is **bug 2** from the restart-aware oracle property-test triage (see #4648 / `packages/sync-service/bugs.md`).

## Problem

After a `StackSupervisor` restart, the source consumer reconnects to the persistent slot and replays from its confirmed LSN, which can lag the shape's on-disk offset. Complete single-fragment transactions (the common case — one statement per txn) took the fast path and were re-processed:

- **Normal shapes:** duplicate / orphan operations appended to the shape log.
- **Subquery shapes:** the re-notified duplicate is re-applied by the dependency materializer, which is a stateful reducer and crashes on the duplicate insert (`"Key ... already exists"`). That crash then cascades into tearing down and *removing* the healthy dependent shapes and returning `409 must-refetch` to clients.

## Solution

Consolidate the offset dedup into a **single `handle_txn_fragment/2` clause** that skips any fragment already at or below `latest_offset` (routing to the existing `skip_txn_fragment`), placed before the complete-transaction fast paths. The dedup now lives in one place and covers single- and multi-fragment paths uniformly, replacing the separate `fragment_already_processed?/2` check. Normal operation is untouched: new transactions have `offset > latest_offset`, so the guard is false and they fall through to the existing clauses.

This relies on a transaction's fragments being entirely at-or-below or entirely above `latest_offset` (which only advances at commit boundaries) — the same assumption the previous per-fragment check already made.

## Test Plan

- [x] `test/electric/shapes/consumer_test.exs` — two regression tests (single-fragment and multi-fragment) replay an already-applied transaction straight to the consumer, bypassing the log collector's own dedup (as happens on restart with a fresh collector), and assert it is skipped: no `append_to_log!`, no `:new_changes` notification, log unchanged. Fail without the fix.
- [x] `test/electric/shapes/` + `test/electric/replication/` — 937 tests, 100 doctests, 6 properties, 0 failures.

---

### DIFF
--- a/.changeset/consumer-skip-replayed-txns.md
+++ b/.changeset/consumer-skip-replayed-txns.md
@@ -1,0 +1,12 @@
+---
+'@core/sync-service': patch
+---
+
+Fix shapes processing duplicate operations after a server restart. On restart the
+persistent replication slot can replay transactions the consumer has already
+applied and persisted. The multi-fragment path skipped these via
+`fragment_already_processed?/2`, but the single-fragment complete-transaction fast
+path did not — so a replayed transaction was re-written to the shape log
+(duplicating ops) and re-notified to dependent subquery materializers, which
+re-applied it and crashed. The consumer now skips a complete transaction already
+at or below its persisted `latest_offset` on every path.
--- a/.changeset/consumer-skip-replayed-txns.md
+++ b/.changeset/consumer-skip-replayed-txns.md
@@ -4,9 +4,7 @@
 
 Fix shapes processing duplicate operations after a server restart. On restart the
 persistent replication slot can replay transactions the consumer has already
-applied and persisted. The multi-fragment path skipped these via
-`fragment_already_processed?/2`, but the single-fragment complete-transaction fast
-path did not — so a replayed transaction was re-written to the shape log
-(duplicating ops) and re-notified to dependent subquery materializers, which
-re-applied it and crashed. The consumer now skips a complete transaction already
-at or below its persisted `latest_offset` on every path.
+applied and persisted. The consumer now skips any transaction already at or below
+its persisted offset, so a replayed transaction is no longer re-written to the
+shape log (duplicating ops) or re-notified to dependent subquery materializers
+(which would re-apply it and crash).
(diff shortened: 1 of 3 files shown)


.changeset/consumer-skip-replayed-txns.md
```diff
@@ -0,0 +1,10 @@
+---
+'@core/sync-service': patch
+---
+
+Fix shapes processing duplicate operations after a server restart. On restart the
+persistent replication slot can replay transactions the consumer has already
+applied and persisted. The consumer now skips any transaction already at or below
+its persisted offset, so a replayed transaction is no longer re-written to the
+shape log (duplicating ops) or re-notified to dependent subquery materializers
+(which would re-apply it and crash).
```
robacourt:
```suggestion
Fix shapes processing duplicate operations after a server restart.
```

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -587,6 +587,20 @@ defmodule Electric.Shapes.Consumer do
...
 
+  # A transaction already applied and persisted (e.g. replayed from the persistent
+  # replication slot on restart) sits at or below `latest_offset`. Skip it here so
+  # the dedup covers every path — including the complete-transaction fast paths
+  # below, which short-circuit before reaching `process_txn_fragment/2`. Re-applying
+  # such a transaction would duplicate ops in the shape log and, for dependent
+  # subquery materializers, re-apply and crash them.
+  #
+  # This relies on a transaction's fragments being entirely at-or-below or entirely
+  # above `latest_offset` (it only advances at commit boundaries).
```
robacourt:
```suggestion
  # Skip transactions already applied and persisted (e.g. replayed from the persistent
  # replication slot on restart) - ones at or below `latest_offset`. 
  #
  # This relies on a transaction's fragments being entirely at-or-below or entirely
  # above `latest_offset` (it only advances at commit boundaries).
```

alco (approved):

PR #4671, opened 2026-07-01
fix(telemetry): reduce SystemMonitor heap growth

icehaunter:
## Summary

This reduces avoidable heap growth in `ElectricTelemetry.SystemMonitor` when the VM monitor emits long GC, long schedule, or long message queue events.

- Classify monitor-event processes with a lightweight `Process.info/2` call that only requests the label, initial call, and the single `$initial_call` process dictionary entry.
- Keep memory aggregation behavior intact while avoiding full process dictionary copies there as well.
- Add an hourly self-GC timer to `SystemMonitor` so large transient terms copied while handling monitor events do not remain resident indefinitely.
- Add a patch changeset for `@core/electric-telemetry`.

## Motivation

In production, `ElectricTelemetry.SystemMonitor` could report more than 1GB of process memory while its GenServer state and message queue were tiny. Forcing `:erlang.garbage_collect(pid)` dropped the process memory back to a few KB, showing the memory was transient garbage retained on the process heap rather than live state.

The expensive path was process classification for monitor events. `proc_type/1` requested full process dictionary, memory, and binary metadata from arbitrary target processes even though monitor-event tagging only needs the process label and initial-call fallback. On nodes with many large request processes, those copied terms can inflate the SystemMonitor heap until a full collection happens.

## Decisions

The public `proc_type/1` path now uses the narrow metadata request. The top memory/bin-memory poller path still asks for memory and binary metadata because those metrics need it, but it now requests only the `$initial_call` dictionary entry instead of the full dictionary.

The SystemMonitor GC is periodic rather than per-event to avoid adding garbage-collection work to every monitor notification while still bounding retained transient heap growth.

### DIFF
--- a/.changeset/fix-telemetry-system-monitor-memory.md
+++ b/.changeset/fix-telemetry-system-monitor-memory.md
@@ -1,0 +1,5 @@
+---
+"@core/electric-telemetry": patch
+---
+
+Reduce SystemMonitor heap growth by avoiding full process dictionary, memory, and binary metadata when classifying monitor events, and periodically garbage collect the monitor process.
--- a/packages/electric-telemetry/lib/electric/telemetry/processes.ex
+++ b/packages/electric-telemetry/lib/electric/telemetry/processes.ex
@@ -4,6 +4,16 @@ defmodule ElectricTelemetry.Processes do
 
   @default_count 5
   @default_limit {:count, @default_count}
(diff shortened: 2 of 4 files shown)


msfstef (approved):
nice

robacourt (approved):
Nice!

PR #4672, opened 2026-07-01
fix(sync-service): shed live shape subscribers that stop draining their mailbox

robacourt:
## Summary

Fixes an unbounded-memory / OOM failure mode in the live shape-streaming path. A live `GET /v1/shape` subscriber whose client can't keep up would grow its process mailbox without bound until the node ran out of memory.

## Problem

A live request registers as a change subscriber. On every transaction the consumer notifies each subscriber with a fire-and-forget `send/2` (`Consumer.notify_clients_of_new_changes/2` → `Registry.dispatch`), with **no backpressure to the producer and no cap on the subscriber's mailbox**. The only backpressure is the socket write (`Plug.Conn.chunk`), which blocks the request process on a stalled/dead client — so the process stops draining its mailbox while the producer keeps enqueuing `:new_changes`. The message queue, and the reference-counted binary each queued message pins, grows without bound.

Observed in production (Honeycomb):
- A single `GET /v1/shape` process with `vm.monitor.long_message_queue.length` ≈ **200k** (`MAX == SUM` for that `process_type` → one process).
Screenshot 2026-07-01 at 17 30 58
- `vm.memory.binary` climbing 0 → **3.9 GB** in lock-step, dragging `vm.memory.system`/`total` to ~5.2 GB → **OOM**.
- Corroborating `Error while streaming response: :timeout` / `socket closed` warnings, and near-zero completed `shape_get.plug.stream_chunk` spans (the process was parked inside one blocked `chunk` write).
- Node was otherwise near-idle on request volume (~78 shape requests in 5h), ruling out load; footprint fixes in 1.7.x don't address it because it's a missing-backpressure bug, not a memory-footprint one.

## Solution

Producer-side mailbox watermark. Before notifying each subscriber, the consumer checks its `message_queue_len`; if it exceeds `:slow_subscriber_max_queue_len`, the subscriber is **terminated** instead of notified. The client reconnects and resumes from its last offset (the live protocol is resumable), so shedding is safe.

This is the only mechanism that catches a handler blocked inside `Plug.Conn.chunk` — it can't wake to check itself, but the producer inspects it externally. The watermark is **always on** (a positive integer, default **10,000**; non-positive values are rejected at stack startup) and tunable via `ELECTRIC_SLOW_SUBSCRIBER_MAX_QUEUE_LEN`.

## Changes

- `lib/electric/shapes/consumer.ex` — watermark check + shed in the notify loop.
- Config wiring for `ELECTRIC_SLOW_SUBSCRIBER_MAX_QUEUE_LEN` (`runtime.exs`, `config.ex`, `application.ex`, `stack_supervisor.ex`, `stack_config.ex`), mirroring the existing `ELECTRIC_CONSUMER_GC_HEAP_THRESHOLD` path.
- Changeset (patch).

## Test Plan

- [x] New reproduction test: a non-draining subscriber is shed once its mailbox crosses the watermark (fails before the fix, passes after).
- [x] New guard test: a healthy, draining subscriber whose backlog stays under the watermark is never shed.
- [x] `consumer_test.exs`, `config_test.exs`, `stack_supervisor_test.exs`, `shape_cache_test.exs`, `shape_log_collector_test.exs`, `partitioned_tables_test.exs` all pass (no regressions on the notify path).

> Note: this is producer-side shedding (mechanism A) only. Notification coalescing and a tighter socket send-timeout were discussed as defense-in-depth and deferred; A alone bounds the mailbox and stops the OOM.

---

### DIFF
--- a/packages/sync-service/lib/electric/shapes/consumer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer.ex
@@ -45,6 +45,16 @@ defmodule Electric.Shapes.Consumer do
   # fragment rate.
   @gc_min_interval_ms 1_000
 
+  # Safety net for live subscribers that stop draining their mailbox (e.g. a
+  # stalled or dead client socket). A healthy subscriber drains :new_changes
+  # promptly and sits near zero; this default only sheds pathologically
+  # backed-up subscribers well before they can OOM the node. Always a positive
+  # integer — shedding cannot be disabled. Tunable via
+  # ELECTRIC_SLOW_SUBSCRIBER_MAX_QUEUE_LEN (stack config key
+  # :slow_subscriber_max_queue_len). Used only as a fallback default when the
+  # stack config has not been seeded (e.g. in unit tests).
+  @default_slow_subscriber_max_queue_len 10_000
+
   @type initialize_shape_opts() :: %{
           :action => :create | :restore,
           optional(:otel_ctx) => OpenTelemetry.otel_ctx() | nil,
@@ -1012,6 +1022,13 @@ defmodule Electric.Shapes.Consumer do
           latest_log_offset :: LogOffset.t()
         ) :: :ok
   defp notify_clients_of_new_changes(state, latest_log_offset) do
+    max_queue_len =
+      Electric.StackConfig.lookup(
+        state.stack_id,
+        :slow_subscriber_max_queue_len,
+        @default_slow_subscriber_max_queue_len
+      )
+
     Registry.dispatch(
       Electric.StackSupervisor.registry_name(state.stack_id),
       state.shape_handle,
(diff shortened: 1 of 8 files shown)


packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -1053,10 +1037,39 @@ defmodule Electric.Shapes.Consumer do
...
+
+    :ok
+  end
+
+  # A live subscriber (e.g. a `GET /v1/shape` request) that can't drain its
+  # mailbox — a stalled or dead client socket while changes keep streaming —
+  # would otherwise accumulate one `:new_changes` message per transaction
+  # without bound, pinning reference-counted binary until the node runs out of
+  # memory. Once its mailbox crosses the watermark we stop notifying it and
+  # terminate it; the client reconnects and resumes from its last offset.
```
alco, line 1058: Here's what looks suspicios to me in this approach:

When processing a shape request, the request handler process enters the `receive` state until it receive **a single** `:new_changes` message. At that point it serves the shape log from storage until the end and closes the HTTP response. There's no looping and there's no expectation that the request handler process can do anything with the 2nd or any subsequent `:new_changes` message.

So how does a handler process end up with 10k messages in its inbox? To me the more likely explanation is that this is a bug, partly caused by a leaky abstraction: the subscription for changes is initiated by a call to `Registry.register()` which subscribes the current *process*. A single *request*'s lifetime is usually smaller than that of a *process* because the same Bandit process may handle multiple requests. So the process ends up receiving messages from a consumer after it has already finished processing the request and has returned to Bandit's pool of request handlers.

This also highlights another problem: once the current request finishes and the same process picks up a new request, it will get a new subscription `ref` and will use it for pattern matching in its `receive` expression. So any messages from the previous subscription will remain in the mailbox and will cause repeated compute tax on the request handler process while it is doing selective receive using a different `ref` value.

alco (changes requested):

packages/sync-service/lib/electric/shapes/consumer.ex
robacourt: Thanks @alco , I should have spotted that! As for your proposed alternative cause, I'm looking into it.

robacourt:
Closed in favour of #4708 and #4711

PR #4673, opened 2026-07-01
Recover live requests when mobile fetch hangs

KyleAMathews:
Adds a live ShapeStream request watchdog and makes fetch retry backoff abortable. This should help React Native/mobile clients recover when an in-flight live fetch hangs or delays reconnects across background/foreground or network transitions instead of leaving sync stalled until app reload.

## Root Cause

PR #4654 fixed the case where React Native aborts a request but drops `AbortSignal.reason`: Electric now tracks restart intent independently of the abort reason.

However, React Native/`whatwg-fetch` can fail in a harder way around app lifecycle transitions: the in-flight fetch promise may not promptly resolve or reject after background/foreground, even if Electric aborts the request. When that happens, `ShapeStream` stays stuck awaiting the live long-poll promise and never reaches the catch/retry path that issues the next request.

A related issue is retry backoff: `TypeError: Network request failed` is retryable, but the backoff sleep was a plain `setTimeout`. A wake/reconnect abort during that sleep would not interrupt the delay, so reconnect could be delayed until the backoff completed.

## Approach

Add a live-request watchdog for long-poll requests:

```ts
liveRequestTimeoutMs?: number | false
```

By default, live long-poll requests time out after 45s. On timeout, Electric:

1. marks the request controller as restart-worthy,
2. aborts the in-flight request, and
3. rejects independently via `Promise.race`, so the request loop continues even if platform fetch ignores abort and never settles.

The watchdog applies only to live long-poll requests (`live=true`), not initial snapshot/catch-up requests. It can be disabled with `liveRequestTimeoutMs: false`.

Fetch retry backoff now uses an abort-aware sleep, so stream-level aborts such as system-wake reconnects can interrupt retry delays immediately.

## Key Invariants

- A wedged live fetch must not block the ShapeStream request loop forever.
- Live long-poll retries may reuse the same URL; non-live stuck-loop guards remain unchanged.
- User/external aborts must still stop the stream rather than being reclassified as reconnects.
- Retry backoff remains indefinite by default for offline scenarios, but aborts can interrupt the sleep.
- Initial/catch-up requests are not subject to the live long-poll watchdog.

## Non-goals

- This does not add direct React Native `AppState` integration.
- This does not change Electric protocol semantics or txid matching.
- This does not expose TanStack collection stream internals.
- This does not make `TypeError: Network request failed` fatal; it remains retryable.

## Trade-offs

The alternative is asking apps to wrap `fetch` or recreate collections on foreground. Handling this in `@electric-sql/client` is preferable because it protects all consumers and avoids app-specific lifecycle workarounds.

The default timeout is intentionally above normal Electric live long-poll cadence, so healthy long-polls should complete naturally while stuck mobile requests self-heal.

## Verification

Red/green evidence:

- Added a failing regression test where a live fetch never resolves and ignores abort; it failed before the watchdog and passes after the fix.
- Added a fetch backoff test that aborts during a retry delay.

Local verification:

```bash
cd packages/typescript-client
pnpm run typecheck
pnpm exec vitest --config vitest.unit.config.ts --run test/fetch.test.ts test/wake-detection.test.ts
```

Both passed locally.

## Files changed

- `.changeset/rn-fetch-watchdog.md` — patch changeset for `@electric-sql/client`.
- `packages/typescript-client/src/client.ts` — adds `liveRequestTimeoutMs` and live long-poll watchdog logic.
- `packages/typescript-client/src/fetch.ts` — makes retry backoff sleep abortable.
- `packages/typescript-client/test/wake-detection.test.ts` — regression test for a hung live fetch that ignores abort.
- `packages/typescript-client/test/fetch.test.ts` — regression test for aborting during network-error backoff.
- `packages/typescript-client/vitest.unit.config.ts` — includes `fetch.test.ts` in the unit-test config.
- `packages/typescript-client/SPEC.md` — documents the watchdog loop-back path and guard.


KyleAMathews:
Closing in favor of fresh PR #4674 based on current main.

PR #4677, opened 2026-07-02
test(oracle): graceful/brutal/rolling restart strategies for postgres oracle tests

robacourt:
## Summary

Extends the oracle property test's restart coverage (`RESTART_SERVER_EVERY`) from a single hard-coded graceful shutdown to three interchangeable restart strategies, selected via `RESTART_TYPE`, plus an opt-in flag to retry transient poll errors so the test never fails on a non-production availability blip.

This is **test-only** — no product code changes, nothing ships to consumers, so no changeset.

### Restart strategies (`RESTART_TYPE`)

A `Support.OracleHarness.RestartStrategy` behaviour with three implementations, resolved by `for_type/1`:

- **`graceful`** (default) — clean `stop_supervised` + restart the same `stack_id`, restoring from disk. Identical to the previous behaviour.
- **`brutal`** — `Process.exit(pid, :kill)` crash (no terminate callbacks), reconcile the ExUnit child spec, wait for the tree to fully terminate, then recover the same `stack_id`/storage.
- **`rolling`** — faithful rolling deploy: a new stack (own `stack_id`/storage/HTTP server) contends on the **same** replication slot via the Postgres advisory lock, boots to `read_only`, then the old stack is gracefully stopped (releasing the lock; persistent slot + publication survive) and the new stack takes over as the single active writer.

Supporting refactors:
- `ComponentSetup.start_stack_supervisor!` made public with `:id`/`:await_ready?` opts (behaviour-preserving defaults).
- Bandit setup factored into `IntegrationSetup.start_bandit_client/2` with a shared Finch pool so a second HTTP server can coexist during a rolling deploy.

### Opt-in transient-error retry (`RETRY_TRANSIENT_ERRORS`, default off)

When enabled, the shape checker retries a transient poll error (5xx response or a connection-level failure) within the check timeout instead of failing immediately. These are availability blips that production hides from clients during a restart/deploy (HTTP drains before teardown; the LB switches after the new instance is ready), so they are not data-consistency signals.

Default behaviour is unchanged. Regardless of the flag, `409`/must-refetch, `4xx`, and data mismatches always fail, so real bugs are never masked.

## Notes for reviewers

- The restart oracle path is opt-in and **not run in CI** (CI runs the oracle test without `RESTART_SERVER_EVERY`). These strategies are a developer tool for hunting restart/deploy consistency bugs.
- Verified against a live Postgres: `graceful` is behaviour-identical to the previous path; `brutal` recovers cleanly; `rolling` performs the full lock hand-off and serves all shapes from the new stack's fresh resnapshot. The transient-error classifier was unit-checked (`{5xx, 4xx, transport}` -> `{retry, fail, retry}`) and confirmed not to mask the real 409 signal.

## Test Plan

- [ ] `RESTART_TYPE=graceful RESTART_SERVER_EVERY=2 mix test test/integration/oracle_property_test.exs --include oracle`
- [ ] `RESTART_TYPE=brutal RESTART_SERVER_EVERY=2 mix test test/integration/oracle_property_test.exs --include oracle`
- [ ] `RESTART_TYPE=rolling RESTART_SERVER_EVERY=2 mix test test/integration/oracle_property_test.exs --include oracle`
- [ ] Default run (no `RESTART_TYPE`) unaffected

---


packages/sync-service/test/support/component_setup.ex
```diff
@@ -481,17 +483,217 @@ defmodule Support.ComponentSetup do
...
+
+    cond do
+      down? ->
+        :ok
+
+      System.monotonic_time(:millisecond) > deadline ->
+        flunk("stack #{stack_id} did not fully terminate after brutal kill")
+
+      true ->
+        Process.sleep(10)
```
alco: I'd pick a larger timeout, say 100, to avoid spinning too much while the stack gets shut down

alco (approved):

packages/sync-service/test/support/component_setup.ex
robacourt:
```suggestion
        Process.sleep(100)
```

PR #4680, opened 2026-07-02
fix(sync-service): reduce live request wait memory

icehaunter:
## Motivation

Live shape requests spend most of their long-poll lifetime parked in a receive while waiting for either a shape change or timeout. Under high concurrency, parking inside the larger request-handling function kept more heap live than the small wait state actually needed, which made the per-request memory floor significant.

## Changes

- Move the blocking live-request receive into a narrow `wait_for_live_change/2` helper that only holds the change reference and timeout while parked.
- Convert receive outcomes into small event values and handle response construction after the wait returns.
- Keep garbage collection immediately before the parked receive, but run it inside the narrow helper so collection happens with a smaller root set.
- Remove a dead `SystemMonitor` timer fallback branch that Elixir 1.20 reports as unreachable under warnings-as-errors.
- Add changesets for sync-service and electric-telemetry.

## Decisions

This keeps the waiter in the existing request process instead of introducing a central waiter registry. That avoids adding a new register/deregister bottleneck while still reducing the heap retained by parked live requests. The existing behavior is preserved for new changes, shape rotation, out-of-bounds checks, read-only timeout handling, and ordinary no-change responses.

### DIFF
--- a/.changeset/reduce-live-request-wait-memory.md
+++ b/.changeset/reduce-live-request-wait-memory.md
@@ -1,0 +1,8 @@
+---
+'@core/sync-service': patch
+'@core/electric-telemetry': patch
+---
+
+Reduce memory retained by live shape requests while they wait for changes.
+
+Remove a dead SystemMonitor timer branch that fails warnings-as-errors on Elixir 1.20.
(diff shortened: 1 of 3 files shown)


robacourt (approved):
LGTM

PR #4681, opened 2026-07-02
Restore subquery shapes correctly after a server restart (materializer move replay)

robacourt:
## Summary

Optimized streaming subquery shapes could diverge from Postgres after a graceful server restart — permanently losing or retaining stale rows — including the **nested** case where the inner subquery is itself an optimized subquery. This fixes that. A representative shape: `issues WHERE project_id IN (SELECT id FROM projects WHERE active = true)`.

## Why it diverges

A subquery shape is served by a **materializer** that tracks which rows the inner query selects and drives **moves** (move-in / move-out) into the outer shape as that set changes. A move is asynchronous and multi-stage: a move-in fires a Postgres query and only splices its rows into the shape log once that query returns.

A graceful restart can interrupt a move at any stage — in flight, queued behind another move, sitting unprocessed in the consumer's mailbox, or computed by the materializer but not yet emitted. When it does, the materializer rebuilds its in-memory *view* from disk on restart (so it "knows" the move happened), but the outer shape's *storage* never received it and nothing re-drives it. The shape is left permanently out of sync.

## The approach: replay missed moves from a durable position

Rather than making each of those stages individually crash-safe — a fragile, case-by-case effort — this makes the materializer the single point that re-drives moves, so every timing class is handled by one mechanism.

Each emitted move is tagged with its source LSN, and each outer consumer persists, per dependency, the position up to which it has applied moves. On restart the consumer re-subscribes and the materializer **replays exactly the moves after that position**; the consumer re-applies them. Because the consumer just replays whatever it is handed, in-flight / queued / mailbox / not-yet-emitted all reduce to "the persisted position is behind" — covered by construction rather than case by case.

For that to be safe across a crash the persisted position must never run ahead of durable storage, so it is **coupled to the writer flush**: a move's position is staged when the move pipeline drains and only committed once the writer confirms the flush has passed that move's spliced rows (and unconditionally at terminate, after the writer is flushed).

## Making the same mechanism correct for nested subqueries

When the dependency is itself an optimized subquery, its log contains move-in/move-out **control messages**, not just data rows — and two properties of those messages break the naive replay:

- they carry no `lsn`/`op_position` in their JSON headers; and
- being dependency-driven, they have no WAL LSN of their own, so they all share a single `tx_offset` (only `op_offset` increments).

Replay originally reconstructed each item's offset from its JSON headers and grouped items into batches by `tx_offset`. Control messages therefore couldn't be positioned (no header offset) or separated (shared `tx_offset`), so the replayed **seed** collapsed to the pre-log snapshot state and the corrective move was never re-emitted. Two changes fix this while leaving the mechanism above intact:

- read each item with its **authoritative storage offset** (`Storage.get_log_stream_with_offsets/3`) instead of re-parsing headers, so control messages are positioned correctly; and
- chunk replay batches by the **full `(tx_offset, op_offset)` offset** — the space `from_lsn`/positions actually live in — so op-distinguished dependency moves are delimited per move.

## The commits

Each fix lands with the failing test that motivates it:

1. **Add failing subquery-restore oracle test** — a deterministic restart reproduction.
2. **Replay missed moves after restart** — the core replay mechanism.
3. **Couple move-position persistence to durable flush** — make the persisted position crash-safe.
4. **Add failing test for nested subquery replay dropping control messages** — the nested reproduction.
5. **Replay control-message moves with authoritative storage offsets** — position every item by its storage offset.
6. **Chunk move replay by full offset, not by transaction** — restore per-move granularity for op-distinguished moves.
7. **Test replay of a value toggled multiple times within one transaction** — lock in that per-op replay is safe (never a move-in and move-out for the same value in one payload).
8. **Drop redundant materializer replay-skip guard** — rely on #4668 for restart de-duplication.

## Testing

- [x] `:oracle_restore_bug_1` and `:oracle_restore_optimized_refetch` (single-level) — deterministic across repeated runs.
- [x] `:oracle_restore_nested_subquery` — was ~4% divergent, now deterministic (80/80).
- [x] Deterministic materializer unit coverage for control-message replay and for multi-toggle-in-one-transaction ordering.
- [x] materializer + consumer + subquery suites — 0 failures.

---

### DIFF
--- a/packages/sync-service/test/integration/oracle_restore_test.exs
+++ b/packages/sync-service/test/integration/oracle_restore_test.exs
@@ -35,7 +35,12 @@ defmodule Electric.Integration.OracleRestoreTest do
   setup ctx do
     ctx =
       with_electric_client(ctx,
-        router_opts: [long_poll_timeout: 100],
+        # A realistic long-poll timeout. A very short one (e.g. 100ms) trips a
+        # separate post-restart long-poll readiness race (bug 3): the poll times
+        # out before replication has caught up after the restart, yielding a
+        # spurious 409. That is independent of the subquery-restore behaviour
+        # under test here.
+        router_opts: [long_poll_timeout: 5000],
         num_clients: 1
       )
 
--- a/packages/sync-service/test/integration/oracle_restore_test.exs
+++ b/packages/sync-service/test/integration/oracle_restore_test.exs
@@ -1,8 +1,8 @@
 defmodule Electric.Integration.OracleRestoreTest do
   @moduledoc """
-  Targeted regression tests for restore-from-file. Each test exercises a
-  scenario from `bugs.md` with a deterministic, minimal mutation sequence,
-  reusing `Support.OracleHarness.test_against_oracle/4`.
+  Targeted regression tests for subquery-shape restore across a server restart,
+  each with a deterministic, minimal mutation sequence, reusing
+  `Support.OracleHarness.test_against_oracle/4`.
 
   These tests use a small, readable "issue tracker" domain schema rather than
   the abstract `level_N` hierarchy from `Support.OracleHarness.StandardSchema`:
@@ -12,9 +12,6 @@ defmodule Electric.Integration.OracleRestoreTest do
 
   The shape under test is "issues belonging to an active project", expressed
   as a subquery over `projects.active`.
-
-  These tests are expected to fail until the underlying Electric bugs are
-  fixed.
   """
 
   use ExUnit.Case, async: false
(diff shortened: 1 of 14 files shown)


packages/sync-service/test/integration/oracle_restore_test.exs
```diff
@@ -129,4 +129,70 @@ defmodule Electric.Integration.OracleRestoreTest do
 
     OracleHarness.test_against_oracle(ctx, shapes, batches, restart_server_every: 1)
   end
+
+  @tag :oracle_restore_optimized_refetch
+  # Build a large persisted backlog on the `projects` source shape: 200 toggles
+  # under a small `chunk_bytes_threshold` so its log spans many chunks. After the
+  # restart the persistent replication slot has to replay that backlog.
+  @tag chunk_bytes_threshold: 200
+  test "optimized subquery shape must-refetches after restart during slot catch-up replay",
```
alco: I'm curious how you run this test. I've run `oracle_restore_test.exs` 10 times and every time it finished quickly with both tests passing.

alco (commented):
I have noticed "Claude Code Opus 4.8" in commit messages so I fed this PR to Fable 5 which spawned dozens of agents to do a thorough review.

https://electric-sql.github.io/alco-agent-tasks/artifacts/html/electric-pr-4681-review.html

robacourt:
Closed in favour of #4735

PR #4704, opened 2026-07-13
perf(durable-streams-rust): per-shard checkpoint triggers — interval knob + retained-WAL size budget

balegas:
Stacked on #4697. Replaces the hardcoded 3s all-shards checkpoint wave with a per-shard dual trigger:

- `--wal-checkpoint-interval-ms` — time trigger, default 3000 (= old behavior).
- `--wal-checkpoint-wal-bytes` — size trigger (0 = off): checkpoint a shard as soon as ITS retained WAL exceeds the budget.

### Why

The checkpoint's only job is bounding retained-WAL size = crash-replay time: acks never gate on it and reads never touch the WAL (reads are sendfile from the data file's page cache, which is written before the WAL ack barrier). So cadence is purely a replay-time budget — the size trigger makes that budget explicit, and shards self-stagger by their own write rates instead of storming the device in a synchronized wave. Each shard's clock restarts when it finishes; an in-flight guard prevents re-firing a slow shard; due shards still checkpoint concurrently.

### Results (c4d-standard-64-lssd, split-lane layout, `--wal-shards 5 --wal-checkpoint-syncfs on`; suite `wal-sizetrigger`)

| arm | 10k streams | 100k streams |
|---|---|---|
| ref-3s (timer, control) | 293.9k | 275.3k |
| **size-1g** (`--wal-checkpoint-wal-bytes 1GiB`, 60s fallback) | **309.6k** | **303.2k** |
| checkpoint-off ceiling (wal-splitlane) | 305.1k | 306.0k |

The size trigger recovers the entire 7–11% checkpoint cost — throughput at the ckpt-off ceiling with replay bounded at ≤1 GiB retained WAL per shard.

### Verification

- 56/56 wal tests pass (checkpoint body untouched — recoverability contract unchanged; the syncfs crash-recovery e2e from #4697 covers the durability ordering).
- Defaults preserve current behavior exactly (3s cadence, size trigger off).


balegas:
Stacked validation complete (`wal-stacked-1m`): this PR's size trigger + exclusive pinned cores + split-lane layout together = **382.7k ops/s @100k streams** (vs 10.4k pre-campaign, 37×; memory-mode ceiling 512k). WAL_TUNING.md in this branch documents the full ideal configuration.

Extreme-cardinality check on the same config: 244k @500k, **56k @1M** — a separate, new mechanism appears near 1M streams (fd-per-stream vs nofile ceiling / 1M-file directory / map working set — not the checkpoint storm this stack fixed). Tracked as the next investigation.

PR #4708, opened 2026-07-13
fix(sync-service): bound memory pinned by serves to slow or stalled clients

robacourt:
## Summary

Bounds the server-side memory pinned by a shape response being served to a client that stops (or slows) reading. Previously every such serve pinned its entire in-flight log chunk (~10 MB by default) — twice — for as long as the serve lived, with nothing to bound or reap it. Accumulated stalled serves exhausted a production node.

## The production incident (2026-07-01)

A customer's primary Electric node (1.7.0) OOM'd after `vm.memory(:binary)` grew 0 → **3.9 GB** in ~65 minutes. Root-caused from their telemetry plus reproduction on both 1.7.0 and current main:

- A bulk write (~12:36 UTC) woke a fleet of parked live long-pollers into large multi-chunk catch-up serves.
- Serves whose clients stalled never complete — and **emit no telemetry** (spans and request counters are completion-based), so the node looked near-idle while dying. The stalled population was visible only as a ~+400 rise in BEAM process count tracking the memory curve.
- Each stalled serve pinned **~19.6 MB**: the whole ~10 MB chunk held by the handler process, plus the same chunk's iodata queued in the socket's driver queue (`port_command` enqueues the entire iodata; busy-port only gates the *next* write — driver queues are invisible to per-process memory attribution).
- The references are **live**, so no GC tuning helps: a forced global GC in the reproduction reclaims ~0.3 MB of hundreds pinned.
- Measured linear scaling: 5 stalled clients ≈ 150 MB, 50 ≈ 602 MB. ~400 × ~10 MB ≈ the 3.9 GB observed.

The stalled clients themselves were streams hung mid-body in `@electric-sql/client` before #4674. **#4674 (client-side watchdog, `@electric-sql/client` ≥ 1.5.23) prevents this accumulation at its source** — a hung fetch is aborted and reconnected, which also frees the server side. But the server must not depend on well-behaved clients: any client that opens a shape request and stops reading (buggy consumer, drip-feeding proxy, hostile actor) could pin ~2× chunk-size per connection indefinitely. This PR adds the server-side guard.

## Why each stalled serve pinned ~2× the chunk

1. **`LogFile.stream_jsons` read eagerly**: one whole-range `pread` (~the full 10 MB chunk as a single binary), returning an eager *list* of entries sliced from it as sub-binaries — pinning the chunk once as the read binary and again via the entry list, for the lifetime of the response.
2. **`Encoder.JSON` batched by item count only** (`chunk_every(500)`): with large rows a single body element grew to multi-MB, held in full by the request process and, once written, by the socket driver queue.

## Solution

1. `stream_jsons` is now a lazy `Stream.resource` reading **64 KiB blocks** (mirroring the existing `stream_jsons_until_offset`, reusing `extract_jsons_from_binary`'s partial-input handling). Each emitted entry pins at most its read block.
2. Encoder batches are additionally capped at **256 KiB** (count cap unchanged).

Per stalled connection: **19.6 MB → well under 1 MB**, regardless of `chunk_bytes_threshold`.

Trade-off for review: a 10 MB chunk is now ~160 `pread`s instead of 1 (the old comment chose the eager read deliberately). This is noise relative to JSON encoding and socket writes; `@read_block_bytes` is a module attribute if tuning is wanted.

## Test plan

- [x] New integration test (`stalled_serve_memory_test.exs`): 10 live long-pollers on never-read sockets, one ~50 MB transaction, 25s grace; asserts per-connection pinned response memory (deduplicated handler-held refc binary + owned port queue bytes) ≤ 2 MB/connection. RED on main (19.6 MB/conn), GREEN with this fix. The assertion is implementation-agnostic: bounding the write unit or reaping stalled serves both satisfy it.
- [x] No regressions: `pure_file_storage` + storage suites, `router_test`, `low_privilege_router_test`, `api_test`, all of `test/integration/` — 361 tests passing.

## Follow-up (not in this PR)

A progress deadline for serves (reap connections making no forward progress) — defense-in-depth so even sub-MB pinning can't accumulate without bound across pathological connection counts.

---

### DIFF
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
@@ -179,9 +179,8 @@ defmodule Electric.ShapeCache.PureFileStorage.LogFile do
   # Every emitted json is a sub-binary of the block it was read from, and the
   # response-streaming pipeline can hold an in-flight element (and everything
   # it references) for the whole lifetime of a serve — including one stalled
-  # on a slow client. An eager whole-range read pinned the entire ~10MB chunk
-  # (once as the read binary, again as the full list of entries) per stalled
-  # connection; bounding the read block bounds that to ~the block size.
+  # on a slow client. Bounding the read block bounds the memory each stalled
+  # connection can pin.
   @read_block_bytes 64 * 1024
 
   def stream_jsons(
--- a/packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
+++ b/packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
@@ -174,6 +174,16 @@ defmodule Electric.ShapeCache.PureFileStorage.LogFile do
     )
   end
 
+  # Read the log range lazily in bounded blocks rather than all at once.
+  #
+  # Every emitted json is a sub-binary of the block it was read from, and the
+  # response-streaming pipeline can hold an in-flight element (and everything
+  # it references) for the whole lifetime of a serve — including one stalled
+  # on a slow client. An eager whole-range read pinned the entire ~10MB chunk
+  # (once as the read binary, again as the full list of entries) per stalled
+  # connection; bounding the read block bounds that to ~the block size.
+  @read_block_bytes 64 * 1024
+
   def stream_jsons(
         %PFS{} = opts,
         log_file_path,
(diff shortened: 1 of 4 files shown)


packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
```diff
@@ -174,37 +174,68 @@ defmodule Electric.ShapeCache.PureFileStorage.LogFile do
     )
   end
 
+  # Read the log range lazily in bounded blocks rather than all at once.
+  #
+  # Every emitted json is a sub-binary of the block it was read from, and the
+  # response-streaming pipeline can hold an in-flight element (and everything
+  # it references) for the whole lifetime of a serve — including one stalled
+  # on a slow client. An eager whole-range read pinned the entire ~10MB chunk
```
alco: Claude just loves embedding the current state of the implementation in code comments and commit messages. But it goes stale very fast.

I find it is usually a good idea to chop off such casual mentions since the more of them there are, the more the muddy the understanding of the code.

packages/sync-service/test/integration/stalled_serve_memory_test.exs
```diff
@@ -0,0 +1,155 @@
...
+  import Support.DbSetup
+  import Support.DbStructureSetup
+  import Support.IntegrationSetup
+
+  alias Electric.Plug.Router
+
+  @moduletag :tmp_dir
+
+  # Stalled connections and the per-connection pinned-memory budget the server
+  # must stay within regardless of chunk size. Today each stalled serve pins
```
alco: This "Today each stalled serve pins..." comment is already outdated in this branch.

alco (approved):
Nice!

packages/sync-service/lib/electric/shape_cache/pure_file_storage/log_file.ex
robacourt: Fixed in 0b41e5774 — trimmed the old-implementation narration here (and the same pattern in encoder.ex), keeping only the rationale for the bound.

packages/sync-service/test/integration/stalled_serve_memory_test.exs
robacourt: Fixed in fe417f7c9 — dropped the pre-fix pinning numbers; the budget constant stands on its own.

PR #4713, opened 2026-07-14
test(sync-service): reproduce FlushTracker stall when a tracked consumer dies without cleanup

balegas:
## Summary

Deterministic reproduction of the long-standing "system stops flushing the WAL" stall (replication slot `confirmed_flush_lsn` stops advancing → unbounded WAL growth on Postgres). **This PR adds a test only — no fix.** The test passes because it asserts the current (buggy) behavior, documenting the stall.

## The stall

`ShapeLogCollector`'s `FlushTracker` tracks each delivered shape at commit until the consumer reports its storage flush. The global flush boundary is the **minimum across incomplete entries**, so a single entry that never completes pins `confirmed_flush_lsn` for the entire stack.

Existing cleanup paths for a dead consumer are:

1. `DOWN` detection in `ConsumerRegistry.publish/2` — but only when a later event is routed **to that same shape**.
2. `ShapeCleaner.handle_writer_termination/3` from the consumer's `terminate/2` — but `terminate` never runs on an untrappable `:kill` (and is a no-op for `:normal` / `:shutdown` / `{:shutdown, other}` reasons).

A consumer killed after its shape is tracked but before it flushes, on a shape whose table then goes quiet, escapes both paths: the FlushTracker entry is permanent and the WAL flush boundary never advances again — even though every live consumer is fully flushed.

## The test

`FlushTracker stall when tracked consumer dies without cleanup on a quiet shape` in `shape_log_collector_test.exs`:

- txn 1 (lsn 42) touches two tables; both consumers process it and both shapes are tracked
- the consumer for `other_table` is killed with `:kill` before flushing
- the surviving consumer flushes everything; boundary advances only to 40 (just below the dead shape's tracked position)
- txn 2 (lsn 50) touches only `test_table`, so publish-time DOWN detection never fires; after the surviving consumer flushes it, **no boundary update is emitted** — the stall
- root-cause proof: explicitly calling `ShapeLogCollector.remove_shape/2` for the dead shape immediately unblocks the boundary to 50

## Fix direction (follow-up)

Reconcile FlushTracker entries against consumer liveness in the SLC (monitor consumer pids, or periodically sweep entries whose consumer is dead) instead of relying on publish-time detection and consumer self-reporting. A stall watchdog logging the blocking shape handles + consumer liveness would also make future occurrences diagnosable in production.

## Test plan

- [x] `mix test test/electric/replication/shape_log_collector_test.exs` — 38 tests, 0 failures

### DIFF
--- a/packages/sync-service/test/electric/replication/shape_log_collector_test.exs
+++ b/packages/sync-service/test/electric/replication/shape_log_collector_test.exs
@@ -1665,6 +1665,138 @@ defmodule Electric.Replication.ShapeLogCollectorTest do
     end
   end
 
+  # Reproduction of the "system stops flushing the WAL" stall:
+  #
+  # A consumer that dies WITHOUT running its terminate callback (e.g. an
+  # untrappable `:kill` exit) after its shape has been tracked in the
+  # FlushTracker leaves a permanently incomplete entry. The only two cleanup
+  # paths — DOWN detection during `ConsumerRegistry.publish/2` for that same
+  # shape, and `ShapeCleaner`-driven `ShapeLogCollector.remove_shape/2` from the
+  # consumer's `terminate/2` — both require either new traffic routed to the
+  # dead shape or a terminate that never ran. If the shape's table stays quiet,
+  # the FlushTracker's global minimum is pinned forever and
+  # `confirmed_flush_lsn` never advances for the whole stack, even though every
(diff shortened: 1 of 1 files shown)


alco:
This test has been adapted into https://github.com/electric-sql/electric/pull/4730, with the failure logic reversed: the adapted test asserts that a missing flush from a consumer *doesn't* wedge FlushTracker.

PR #4716, opened 2026-07-15
Add ELECTRIC_TCP_READ_TIMEOUT for the HTTP keep-alive idle timeout

alco:
## What

Adds `ELECTRIC_TCP_READ_TIMEOUT` to configure ThousandIsland's `read_timeout` (default 60s), which doubles as Bandit's HTTP keep-alive idle timeout: connections that have waited this long for their next request are closed. The option mirrors `ELECTRIC_TCP_SEND_TIMEOUT` (human-readable duration) and threads through `tcp_read_timeout` → `thousand_island_options/1`. Default is unchanged (`nil` → ThousandIsland's 60s).

Includes a docs entry in `website/docs/sync/api/config.md` and a changeset.

## Why

When Electric runs behind a connection-pooling proxy, the target's keep-alive timeout must **exceed** the proxy's idle timeout. If it doesn't, the proxy can reuse a pooled connection at the same moment Bandit's idle timer closes it (the close is a bare FIN — nothing announces it in-band), and the resulting TCP reset surfaces to clients as an intermittent `502 Bad Gateway`.

### DIFF
--- a/.changeset/tidy-donuts-brush.md
+++ b/.changeset/tidy-donuts-brush.md
@@ -1,0 +1,9 @@
+---
+"@core/sync-service": patch
+---
+
+Add `ELECTRIC_TCP_READ_TIMEOUT` to configure the socket read / HTTP keep-alive
+idle timeout (ThousandIsland's `read_timeout`, default 60s). When Electric runs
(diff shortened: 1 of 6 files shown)


robacourt (approved):
👍

PR #4725, opened 2026-07-15
feat(burn): switch agents to latest Haiku model

balegas:
## What

- Adds a `:haiku` model alias pointing at `claude-haiku-4-5-20251001` and switches all three Burn agents (Sarah, Jerry, Frankie) plus the adapter default to it.
- Mirrors the alias in the test config (unchanged cheap test model).

## Why

The burn demo (https://burn.examples.electric-sql.com) is currently down: the production ECS service crash-loops because the Anthropic API key it deployed with was revoked (401 → agent GenServers crash → app shutdown → ALB 502/503).

The `ANTHROPIC_KEY` repo secret has already been rotated to a valid key (verified against the API). **Merging this PR triggers the examples deploy for burn**, which picks up both the new key and the Haiku model.

## Notes

Follow-up worth considering: a 401 from Anthropic currently takes the whole app down via a `MatchError` in `Burn.Agents.do_instruct/5` (`lib/burn/agents.ex:53`) — handling the error tuple gracefully would keep the demo up (degraded) instead of crash-looping.

### DIFF
--- a/examples/burn/config/config.exs
+++ b/examples/burn/config/config.exs
@@ -14,9 +14,10 @@ config :burn,
 config :burn, Burn.Adapters.Anthropic,
   api_url: "https://api.anthropic.com/v1/messages",
   api_version: "2023-06-01",
+  # Only Haiku is resolvable in the deployed demo — any code path that
+  # asks for another model alias fails fast rather than spending on a
+  # more expensive model.
   models: [
-    opus: "claude-opus-4-20250514",
-    sonnet: "claude-sonnet-4-20250514",
     haiku: "claude-haiku-4-5-20251001"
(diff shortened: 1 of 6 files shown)


balegas:
Update: removed the `opus`/`sonnet` aliases from the model config entirely. The model was already never user-selectable (compile-time `@model` atoms only — verified there is no web-layer path that accepts a model parameter), but now `claude-haiku-4-5-20251001` is the *only* resolvable model in the deployed demo: any code path referencing another alias fails fast via `Keyword.fetch!` instead of silently spending on a more expensive model.

thruflo (approved):

PR #4729, opened 2026-07-16
fix(sync-service): cold nested subquery shape times out subscribing to a dependency materializer (#4715)

### LINKED ISSUE #4715: Cold nested subquery shape can time out subscribing to an intermediate dependency materializer
## Summary

Electric 1.7.7 deterministically returns an initial HTTP 500 for a cold nested dependency shape. Retrying the same shape succeeds after the dependency snapshots finish.

## Minimal topology

```sql
assets
WHERE id NOT IN (
  SELECT asset_id
  FROM container_rels
  WHERE asset_id IS NOT NULL
    AND asset_id NOT IN (
      SELECT asset_id
      FROM container_rels
      WHERE asset_id IS NOT NULL
        AND container_id IN (
          SELECT id FROM containers WHERE <per-user access predicate>
        )
    )
)
```

The equivalent historical positive-`IN` topology did not fail in the matched run.

## Controlled results on 1.7.7

- Concurrency 1: one current nested shape returned 500 after 5.191 seconds; retry returned 200 after 2.966 seconds and reached `0_inf`.
- Bounded concurrency 4: four current nested shapes produced exactly four first-request 500s; all recovered on retry.
- All five recovered clients had valid handles, reached up-to-date, returned the expected accessible row set, and excluded the inaccessible probe row.
- Five matched historical positive-`IN` shapes returned 200 on their first request.
- Service logs contained five shape removals caused by `GenServer.call(intermediate_materializer, :subscribe, 5000)` timing out and five outer `await_snapshot_start(..., 45000)` request failures.
- No restart, OOM, missing-materializer, Stack-not-ready, snapshot-failure, or iodata signature occurred in the bounded window.

## Expected

A single cold nested shape waits for dependency materializers and completes without an externally visible 500.

## Actual

The outer consumer times out after five seconds while subscribing to the intermediate dependency materializer. Electric removes the outer shape and returns 500. A later client retry succeeds.

## Relevant source

In 1.7.7:

- `packages/sync-service/lib/electric/shapes/consumer/materializer.ex`: `subscribe/1` uses the default five-second `GenServer.call` timeout.
- `packages/sync-service/lib/electric/shapes/consumer.ex`: `all_materializers_alive?/1` calls `Materializer.subscribe/1` during outer consumer initialization.

A minimal read-only client harness can be provided. It creates distinct cold shape keys with a no-op predicate nonce and records status, latency, handle, offset, up-to-date state, and row-set correctness. It does not require schema or data mutation.

Retry is a client mitigation, not a resolution: every cold current-topology request in the matched run still exposed an initial 500.

robacourt:
## Summary
A cold nested subquery shape could return an initial HTTP 500 while its dependency snapshots were still in progress. A retry succeeded once the dependency snapshots finished.

Fixes #4715

## Problem
When an outer shape's consumer initializes, `Consumer.all_materializers_alive?/1` subscribes to each dependency materializer via `Materializer.subscribe/1`. That was the **only** materializer call using the default 5s `GenServer.call` timeout — `wait_until_ready/1` and `new_changes/3` both use `:infinity`.

A materializer can't answer any call until `handle_continue(:start_materializer)` returns, and that blocks on `Consumer.await_snapshot_start(..., :infinity)` until its own snapshot starts. For a cold nested topology, the intermediate materializer itself waits on a deeper dependency, so its snapshot can start later than 5s after the outer consumer subscribes. The subscribe then times out, Electric removes the outer shape, and the client's first request 500s:

```
Removing shape "..." due to abnormal shutdown: {:timeout, {GenServer, :call, [#PID, :subscribe, 5000]}}
  ... consumer.ex: Electric.Shapes.Consumer.all_materializers_alive?/1
  ... consumer.ex: Electric.Shapes.Consumer.finish_initialization/3
```

This matches the logs reported in #4715.

## Solution
`Materializer.subscribe/*` now waits with `:infinity`, consistent with the other materializer calls. This matches the intended semantics from the issue — *"a single cold nested shape waits for dependency materializers and completes without an externally visible 500."*

Why this is safe:
- **Liveness** is already handled by the caller: `all_materializers_alive?/1` does `Process.monitor(pid, ...)` *before* subscribing, so a dead materializer surfaces as a call exit rather than being masked by a short timeout.
- **No deadlock**: the shape-dependency graph is a DAG (outer → inner), so no dependency can transitively wait on the outer consumer.
- **Replication is not stalled** by the longer block: the outer shape is registered with the `ShapeLogCollector`'s `EventRouter` (via `SetupEffects` → `ShapeLogCollector.add_shape`) only *after* `all_materializers_alive?/1` returns, inside `initialize_event_handler`. While the consumer is blocked in the subscribe it is not yet a routing target, so the collector never broadcasts a transaction to it. The overall wait is bounded by the request-level `await_snapshot_start(outer, 45000)` budget.

## Reproduction / Test Plan
Added a `:slow` router test (`test/electric/plug/router_test.exs`) that deterministically reproduces the bug by stalling **only** the dependency shape's snapshot past the 5s subscribe window (via an injected `:create_snapshot_fn`), then asserting the cold nested shape still serves 200.

- [x] Test fails on `main` with the `{:timeout, {GenServer, :call, [_, :subscribe, 5000]}}` signature
- [x] Test passes with the fix
- [x] `materializer_test.exs`, `consumer_test.exs`, and the `/v1/shapes - subqueries` router tests pass

Run the reproduction:
```sh
mix test test/electric/plug/router_test.exs --include slow --only line:3505
```

---

### DIFF
--- a/.changeset/cold-nested-subquery-subscribe-timeout.md
+++ b/.changeset/cold-nested-subquery-subscribe-timeout.md
@@ -1,0 +1,5 @@
+---
+"@core/sync-service": patch
+---
+
+Fix a cold nested subquery shape returning an initial HTTP 500 while its dependency snapshots were still in progress. When an outer shape's consumer initialized, it subscribed to each dependency materializer using `GenServer.call` with the default 5s timeout. A dependency materializer stays blocked in start-up until its own snapshot starts, so if that snapshot took longer than 5s the subscribe timed out, the outer shape was removed, and the client's first request 500'd (a retry succeeded once the snapshot finished). The subscribe now waits with `:infinity`, consistent with the other materializer calls, so a cold nested shape waits for its dependency materializers and completes without an externally visible 500. Liveness is unaffected: the caller already monitors the materializer, so a dead dependency surfaces as a call exit rather than being masked by a short timeout.
--- a/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
+++ b/packages/sync-service/lib/electric/shapes/consumer/materializer.ex
@@ -103,9 +103,15 @@ defmodule Electric.Shapes.Consumer.Materializer do
     end)
   end
 
-  def subscribe(pid) when is_pid(pid), do: GenServer.call(pid, :subscribe)
+  # A materializer cannot answer any call until `handle_continue(:start_materializer)`
+  # returns, and that blocks on `await_snapshot_start(:infinity)` until its own snapshot
+  # starts. A cold dependency snapshot can take longer than the default 5s `GenServer.call`
+  # timeout, so we wait with `:infinity` (consistent with `wait_until_ready/1` and
+  # `new_changes/3`). Liveness is handled by the caller monitoring the materializer, so a
+  # dead materializer surfaces as a call exit rather than being masked by a short timeout.
+  def subscribe(pid) when is_pid(pid), do: GenServer.call(pid, :subscribe, :infinity)
 
-  def subscribe(opts) when is_map(opts), do: GenServer.call(name(opts), :subscribe)
+  def subscribe(opts) when is_map(opts), do: GenServer.call(name(opts), :subscribe, :infinity)
 
   def subscribe(stack_id, shape_handle),
     do: subscribe(%{stack_id: stack_id, shape_handle: shape_handle})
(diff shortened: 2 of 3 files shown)


alco (approved):
Reasonable!

PR #4730, opened 2026-07-17
fix(sync-service): prevent permanent flush-boundary stalls from unexpected consumer lifecycle transitions

alco:
## Problem

The FlushTracker pins `confirmed_flush_lsn` for the whole stack on the minimum incomplete flush entry, and cleanup of entries is purely event-driven: either new traffic routed to the same shape (publish-time DOWN detection) or an explicit `ShapeLogCollector.remove_shape/2` arriving via the ShapeCleaner chain. Several real-world paths break both — a storage error that crashes the consumer *and* aborts its `terminate/2` before `handle_writer_termination` runs, a removal chain that dies between the `ShapeStatus` removal and the SLC removal (unretryable: the retry short-circuits at `{:error, :data_removed}`), untagged `{:shutdown, x}` exit reasons classified as benign, and a consumer that stays alive but stops flushing. On a subsequently-quiet table any of these pins the WAL ack forever, causing unbounded WAL growth on Postgres. PR #4713 has a minimal reproduction of the stall.

## Solution

The SLC becomes the authority on consumer-death handling for flush accounting. Invariant: an incomplete FlushTracker entry always has a live monitor watching the pid responsible for completing it, and a timestamp bounding how long it may sit without flush progress.

- **Writer monitors**: the SLC monitors every consumer with an incomplete FlushTracker entry (created when a commit-fragment publish tracks the shape, released when the entry completes or the shape is removed). DOWNs are classified by exit reason: crashes unpin the flush boundary immediately and trigger shape invalidation; deliberate-cleanup DOWNs (`{:shutdown, :cleanup}`) unpin only; bare `:shutdown`/`:killed` are assumed to be supervisor teardown and left alone.
- **Stall detection**: incomplete entries carry a `last_progress_at` timestamp refreshed only by flush notifications. A periodic tick finds shapes whose entries make no flush progress for `ELECTRIC_FLUSH_STALL_GRACE_PERIOD` (default 60s) and invalidates them via the normal ShapeCleaner path, re-arming on each fire so a lost cleanup chain retries instead of pinning forever. Misclassified DOWNs self-heal through the same mechanism.
- **Stall challenge-response**: a stalled entry whose writer is still alive is challenged (`:verify_flush_progress`) before being acted on. A consumer answers only while legitimately deferring flush notifications (buffering ahead of PG snapshot info, subquery move-in awaiting splice); the answer re-arms the entry's grace period so healthy deferral phases are not misread as stalls. Only a suspect still stalled and silent at the next check — or a shape whose writer is already dead and demonitored, which skips the challenge — is invalidated. Consumers stay silent in steady state: no periodic messaging, and a wedged consumer can't answer, so the backstop stays armed.
- **Reason normalization**: the benign clause of `handle_writer_termination` shrinks to exactly `:normal`/`:killed`/`:shutdown`, so untagged `{:shutdown, x}` reasons now trigger removal; tagged dependency-materializer shutdowns map to `stop_and_clean`; suspend additionally requires all txn state flushed and notified (guaranteeing a suspended consumer never leaves an incomplete entry behind); `Consumer.stop` escalates a wedged consumer to a kill after the stop call times out.
- ShapeCleaner re-issues the idempotent SLC removal even when the shape is already gone from `ShapeStatus`, healing pins left by removal chains that died partway through.

Telemetry: new `[:electric, :flush_tracker, :writer_down]` and `[:electric, :flush_tracker, :stall_detected]` events.


packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
```diff
@@ -95,15 +103,25 @@ defmodule Electric.Replication.ShapeLogCollector.FlushTracker do
     last_flushed == %{} and :gb_trees.is_empty(tree)
   end
 
-  @spec handle_txn_fragment(t(), TransactionFragment.t(), Enumerable.t(shape_id())) :: t()
+  @doc """
+  Check whether the shape has an incomplete flush entry (completed entries are deleted).
+  """
+  @spec tracked?(t(), shape_id()) :: boolean()
+  def tracked?(%__MODULE__{last_flushed: last_flushed}, shape_id) do
```
magnetised, line 110: if the doc is correct this is a confusing name, e.g. `incomplete?` would make more sense

if the doc is incorrect, well...

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -488,6 +539,199 @@ defmodule Electric.Replication.ShapeLogCollector do
...
+      {:ok, {_old_pid, _old_ref}} ->
+        # A different pid now owns this shape's entry (resumed consumer): swap the
+        # monitor over to it.
+        state |> demonitor_writer(shape_handle) |> monitor_writer(shape_handle, pid)
+
+      :error ->
+        # Monitoring an already-dead pid yields an immediate :noproc DOWN, so a
+        # writer that dies before this call is never lost: the DOWN is classified
+        # as teardown and the stall check picks the entry up if that was wrong.
+        ref = Process.monitor(pid)
```
magnetised: could we drop the `writer_monitor_refs` reverse mapping if we send the shape handle in the down message? e.g. `Process.monitor(pid, tag: {:down, shape_handle})` will send `{{:down, shape_handle}, ...}` rather than `{:DOWN, ...}`

packages/sync-service/lib/electric/replication/shape_log_collector.ex
```diff
@@ -647,45 +896,78 @@ defmodule Electric.Replication.ShapeLogCollector do
...
 
-        _ ->
-          flush_tracker
-      end
+        flush_tracker =
+          FlushTracker.handle_txn_fragment(
+            state.flush_tracker,
+            event,
+            delivered_shapes,
+            System.monotonic_time(:millisecond)
```
magnetised: a simple private `now()` function would unify all these monotonic_time calls

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -102,7 +102,28 @@ defmodule Electric.Shapes.Consumer do
...
+
+  defp kill_if_alive(pid) do
+    if Process.alive?(pid) do
+      ref = Process.monitor(pid)
+      Process.exit(pid, :kill)
+
+      receive do
+        {:DOWN, ^ref, :process, ^pid, _reason} -> :ok
+      after
+        5_000 ->
```
magnetised: uncomfortable with this path - we just killed the process, we should see that. if we don't WTF is going on and is the process still alive? I think either a retry loop leading to an error or a direct error here is appropriate

packages/sync-service/lib/electric/shapes/consumer.ex
```diff
@@ -1251,12 +1299,22 @@ defmodule Electric.Shapes.Consumer do
     Inspector.clean(table_oid, inspector)
   end
 
+  # Invariant: consumers must only ever stop with :normal/:shutdown or one of the two
```
magnetised:
```suggestion
  # Consumers must only ever stop with :normal/:shutdown or one of the two
```

machines writing like machines

magnetised (approved):
Awesome. Couple of minor queries but great.

Was initially unhappy with yet *more* responsibilities being crammed into the SLC but it makes sense.

packages/sync-service/lib/electric/replication/shape_log_collector/flush_tracker.ex
alco: I think the function name is fine. The module name is `FlushTracker` and it does track whether shapes have flushed new data to storage. I'll update the doc string.

packages/sync-service/lib/electric/replication/shape_log_collector.ex
alco: Great idea!

packages/sync-service/lib/electric/replication/shape_log_collector.ex
alco: Fair enough.

packages/sync-service/lib/electric/shapes/consumer.ex
alco: Good catch! It's not like there's any logic hanging off this monitor message. It's pure slop. We just kill and move on.

